PR #3344, opened 2026-05-05
[Model] realtime tool calling for Qwen3 Omni

raghavm243512:
## Purpose

Add multi-turn conversation, system instructions, and tool calling for
speech-to-speech Qwen3-Omni over the OpenAI-compatible `/v1/realtime`
WebSocket API.

The connection now constructs a Qwen3 chat-template prompt around the
audio per turn, enabling:

- A system instruction supplied via `session.update`.
- Multi-turn history retained across turns within a session, with
  user-side transcripts produced by a side-channel ASR pass that
  continuous-batches alongside the chat thinker on stage 0.
- Tool calls detected mid-generation in the audio pass — the engine
  aborts before the talker spins on `<tool_call>` XML, the call is
  emitted to the client, and the model resumes with audio once the
  client returns the `function_call_output`.

Three modes are selectable per session: plain (no tools/instructions),
instructions-only, and tools. All three share the multi-turn machinery;
tool-call detection / abort is gated on `self.tools`.

Add the Qwen-specific chat-template surface
(constants, `format_tools_schema`, `parse_tool_call`, `render_history`,
`build_audio_pass_prompt`, `build_transcription_prompt`,
`strip_special_tokens`) onto `Qwen3OmniMoeForConditionalGeneration`, so
`RealtimeConnection` only routes through `self.model_cls`.

A few adjacent fixes ride along:
- `compute_talker_prompt_ids_length` skips text-only user blocks (was
  oversizing the talker prefill and hanging stage-1 on tool-response
  turns).
- `talker2code2wav` guards against empty `multimodal_output` (the audio
  pass abort can leave the talker output empty) and warns when codec
  output approaches `max_num_batched_tokens`.
- `AsyncOmniEngine` hard-exits with code 1 on post-startup orchestrator
  crash so deployments can detect the failure instead of running
  silently broken.
- Async chunking working over realtime endpoint. See orchestrator.py changes

### Unit test modifications

Two existing test files updated; no new test files added.

**`tests/distributed/omni_connectors/test_adapter_and_flow.py`** — added
`TestComputeTalkerPromptIdsLength` (9 cases): empty input, system /
text-only user / non-last assistant blocks skipped; audio / image /
video user blocks counted; last assistant adds 9; full conversation
mix. Locks in the new "skip text-only user block" behavior.

**`tests/entrypoints/test_realtime_connection_helpers.py`** — added
`TestQwen3OmniRealtimeChatFormatting` (~20 cases) covering the
realtime classmethods now living on
[...]

PR #5555, opened 2026-07-29
[Frontend] Add tool/function calling support to /v1/realtime for Qwen3-Omni

### LINKED ISSUE #5522: /v1/realtime has no tool/function-calling support
**Environment:** vllm-omni 0.24.0, serving `Qwen/Qwen3-Omni-30B-A3B-Instruct` via `vllm serve Qwen/Qwen3-Omni-30B-A3B-Instruct --omni --no-async-chunk`.

**What we're trying to do:** Use the self-hosted `/v1/realtime` websocket as a full speech-to-speech "brain" for a LiveKit Agents voice worker, including tool/function calling (the model deciding to call e.g. a weather-lookup or inventory-check tool mid-conversation) — the same pattern OpenAI's and Gemini's realtime APIs support.

**What we found:** The realtime protocol has no wire mechanism for tools at all, at either the vLLM or vllm-omni layer:
- `SessionUpdate` (`vllm/entrypoints/speech_to_text/realtime/protocol.py`) only declares a `model` field.
- `RealtimeConnection.handle_event` (`vllm/entrypoints/speech_to_text/realtime/connection.py`) only reads `event.get("model")` from `session.update` — nothing else is parsed or threaded into generation.
- vllm-omni's `RealtimeConnection` override (`vllm_omni/entrypoints/openai/realtime_connection.py`) only adds audio-delta emission on top of that; it doesn't add tool support either.
- There's no server→client event for a function-call request, and no client→server event to submit a tool result.

**Impact:** A realtime session can never be given tool definitions, so the model has no way to know it has any tools available — it just verbally declines ("I can't provide real-time weather updates...") rather than invoking anything, even though the same checkpoint supports tool-calling fine via `/v1/chat/completions` with `tools=[...]`.

**Request:** Add tool/function-calling support to the realtime protocol — e.g. a `tools` field on `SessionUpdate`, plus request/response event types for function calls (mirroring OpenAI's `response.function_call_arguments.*` / `conversation.item.create` shape, or vLLM's own `tool_parsers` used for chat completions). Happy to help test against a patch if useful.

TheCodeWrangler:
## Summary

Fixes #5522.

`/v1/realtime` had no wire mechanism for tools at all - no field on `session.update`, no function-call events in either direction - so a self-hosted Qwen3-Omni realtime session could never be given tools, even though the same checkpoint supports tool-calling fine via `/v1/chat/completions`. This adds it, mirroring OpenAI's Realtime API event shapes:

- `session.update` gains an optional `tools` field.
- `buffer_realtime_audio()` renders through the model's own chat template with `tools=` when present, so the thinker gets the `<tools>...</tools>` system preamble and `<tool_call>` output format it was already trained on (same format `/v1/chat/completions` uses). No-op when `tools` is empty - verified byte-identical rendering to the prior hardcoded prompt string.
- New `realtime_tool_calls.py`: a small, self-contained `<tool_call>` XML-tag parser (streaming name/arguments diffing via bracket-depth JSON matching, not naive string trimming).
- New server events: `response.output_item.added`, `response.function_call_arguments.delta`, `response.function_call_arguments.done`.
- New client event: `conversation.item.create` with a `function_call_output` item submits a tool result, which splices onto the raw token sequence and continues generation - producing the model's actual spoken reply, recursing if it chains another tool call.

This PR is part of a 3-PR split of a broader `/v1/realtime` spec-conformance audit (#5558): this one covers tool-calling, #TBD covers voice/speaker selection, #TBD covers `instructions` (system prompt). Split into independent PRs so each can be reviewed/merged on its own.

## A real vLLM template-resolution bug found along the way

`safe_apply_chat_template`'s own auto-resolution (`resolve_chat_template` in `vllm/renderers/hf.py`) explicitly skips the AutoProcessor-based template lookup whenever `tools` is passed (that path is gated on `tools is None`). For this multimodal checkpoint the tool-aware chat template genuinely lives on the processor, not the raw tokenizer, so with `tools` set it silently fell through to vLLM's generic non-tool-aware fallback template - no crash, no error, just a plausible-looking wrong answer. Worked around here by passing the processor's `chat_template` explicitly (`resolve_chat_template`'s 1st priority always wins and still respects `tools` when applying it). Happy to also raise this separately against vLLM proper if maintainers think it's worth a dedicated fix upstream there.

## Known limitation (by design, not an oversight)

Audio is not forwarded to the client for a turn that turns out to be a tool call, but the 3-stage pipeline (thinker→talker→code2wav) still runs end-to-end for it under the hood. Properly skipping talker/code2wav for a tool-call segment would require teaching the shared orchestrator (`engine/orchestrator.py`, used by every model in the repo) a new "finish successfully after stage 0 only" capability - a much larger, riskier change for a first PR. This trades a small amount of wasted compute on an already lighter-loaded GPU (talker/code2wav run on a separate GPU from the thinker in the standard 2-GPU deploy config) for a change contained entirely to this feature's own files. Happy to discuss tackling proper pipeline-level gating as a follow-up if there's interest.

## Robustness bounds

Two limits on the tool-call loop, both with unit tests:

- **`MAX_TOOL_ROUNDS = 8`.** `_await_tool_results_and_continue` re-enters `_run_generation`, so a model that keeps re-emitting tool calls instead of answering would recurse without bound -- growing the stack and holding each round's result generator open, since every round is awaited inside the outer `try`. Exceeding the cap emits an `error` event with code `tool_call_loop` and abandons the turn. This is the failure mode hit during bring-up when the spliced prompt was malformed (see `_close_assistant_turn`).
- **The tool-result wait releases when the client goes away.** It previously checked `_is_connected` only at the top of the loop and then blocked in `Queue.get()` indefinitely, so a client that vanished mid-tool-call parked the generation task forever. The wait is now bounded so liveness is re-checked; a slow-but-live client is unaffected.

A new user turn also resets the round budget and drains any stale `function_call_output`, so a late result from a previous turn can't be consumed as if it answered one of this turn's calls.

## Test plan

- [x] New L1 unit tests for the `<tool_call>` parser (`tests/entrypoints/test_realtime_tool_calls.py`) - 10 tests, all passing, including a regression test for a streaming-diff edge case that left invalid trailing `}\n` in the accumulated arguments JSON.
- [x] New L1 unit tests for `handle_event`'s tool-call event routing (`tests/entrypoints/test_realtime_connection_helpers.py`) - 5 tests, all passing, using `mocker`/`AsyncMock` per repo convention.
- [x] New L3 online e2e test (`tests/entrypoints/openai_api/test_qwen3_omni_realtime_websocket.py::test_tool_calling_round_trip`) following the exact pattern of its two neighbors in the same file. **Run green through the real `omni_server` fixture**: `pytest -sv '...::test_tool_calling_round_trip[sync]' --run-level advanced_model` -> `1 passed in 156.66s`, on 2x H100 with real weights.
- [x] Manually verified end-to-end on real GPUs against `Qwen3-Omni-30B-A3B-Instruct`: full tool-call round trip (model calls `get_weather`, streams valid incremental JSON arguments, client submits a result, generation resumes and speaks the final answer incorporating it) and the no-tools path confirmed byte-for-byte unchanged from before this PR.
- [x] All 25 tests across both new/changed test files actually executed (not just collected) against the real `vllm`/`vllm_omni` install inside the serving container, all passing.
- No CI YAML changes needed - `tests/entrypoints/` is already swept unconditionally by `test-ready.yml` (L1) and `test-merge.yml` (L3 H100/L4), so both new test files are picked up automatically.

**Note on the new L3 test**: now verified end to end through the `omni_server` fixture (`--run-level advanced_model`, 2x H100, real `Qwen3-Omni-30B-A3B-Instruct` weights): the model emits a `get_weather` call, the client submits a `function_call_output`, and generation resumes and speaks a final reply incorporating the result. Note the test needs `espeak-ng` + `pyttsx3` to synthesize its input audio; the CI images install both (`docker/Dockerfile.ci`), a plain serving image does not.

### DIFF
--- a/vllm_omni/entrypoints/openai/realtime_connection.py
+++ b/vllm_omni/entrypoints/openai/realtime_connection.py
@@ -58,6 +58,13 @@ class RealtimeConnection(VllmRealtimeConnection):
         self.engine = cast(AsyncOmni, self.serving.engine_client)
         self._realtime_audio_ref: np.ndarray | None = None
         self._tools: list[dict[str, Any]] | None = None
+        # The current turn's prompt as handed to the engine BEFORE multimodal
+        # expansion: un-expanded `prompt_token_ids` (one `<|audio_pad|>`) plus the
+        # audio in `multi_modal_data`. Tool-call continuations rebuild from this
+        # rather than from the engine's post-expansion `output.prompt_token_ids`,
+        # because those contain the expanded audio placeholder run with no way to
+        # re-attach the audio - see _await_tool_results_and_continue.
+        self._turn_prompt: dict[str, Any] | None = None
         # index (parser-assigned, per generation) -> {"call_id", "name", "arguments"}
         self._pending_tool_calls: dict[int, dict[str, Any]] = {}
         self._tool_result_queue: asyncio.Queue[dict[str, Any]] = asyncio.Queue()
@@ -109,10 +116,25 @@ class RealtimeConnection(VllmRealtimeConnection):
             audio_stream, input_stream, self.serving.model_config, tools=self._tools
         )
         async for prompt in stream_input_iter:
+            # Remember the pre-expansion prompt so tool-call continuations can
+            # re-anchor on the user's audio (see self._turn_prompt).
+            if isinstance(prompt, dict):
+                self._turn_prompt = dict(prompt)
             yield await self._render_prompt(prompt)
 
-    async def _render_token_prompt(self, prompt_token_ids: list[int]) -> AsyncGenerator[StreamingInput, None]:
-        yield await self._render_prompt(TokensPrompt(prompt_token_ids=prompt_token_ids))
+    async def _render_token_prompt(
+        self,
+        prompt_token_ids: list[int],
+        multi_modal_data: dict[str, Any] | None = None,
+    ) -> AsyncGenerator[StreamingInput, None]:
+        token_prompt = TokensPrompt(prompt_token_ids=prompt_token_ids)
+        # Tool-call continuation: re-attach the turn's audio. Without it the
+        # `<|audio_pad|>` placeholder still sits in the token ids but has no
+        # encoder output behind it, so the thinker cannot see what the user asked
+        # and free-associates unrelated tool calls instead of answering.
+        if multi_modal_data:
+            token_prompt["multi_modal_data"] = multi_modal_data
+        yield await self._render_prompt(token_prompt)
 
     @staticmethod
     def _tensor_to_numpy(value) -> np.ndarray | None:
@@ -397,12 +419,27 @@ class RealtimeConnection(VllmRealtimeConnection):
             add_generation_prompt=True,
             tokenize=False,
         )
-        continuation_token_ids = (
-            list(prior_prompt_token_ids) + list(assistant_token_ids) + tokenizer.encode(suffix_text)
-        )
+        # Splice onto the turn's PRE-expansion prompt, not the engine's
+        # post-expansion `output.prompt_token_ids`. The latter carries the expanded
+        # `<|audio_pad|>` run, and re-submitting it as a bare TokensPrompt drops the
+        # audio itself: the thinker then sees placeholder tokens with no encoder
+        # output, loses the user's question entirely, and answers by inventing
+        # further tool calls (unrelated cities/items) instead of replying. Falling
+        # back to `prior_prompt_token_ids` only matters for a continuation that never
+        # went through buffer_realtime_audio (no audio to lose in that case).
+        base_prompt = self._turn_prompt or {}
+        base_token_ids = list(base_prompt.get("prompt_token_ids") or prior_prompt_token_ids)
+        multi_modal_data = base_prompt.get("multi_modal_data")
+        continuation_token_ids = base_token_ids + list(assistant_token_ids) + tokenizer.encode(suffix_text)
+
+        # Advance the base so a chained tool call next round splices onto this
+        # turn's full history while the audio stays attached exactly once, at the
+        # front, still un-expanded.
+        if self._turn_prompt is not None:
+            self._turn_prompt = {**base_prompt, "prompt_token_ids": continuation_token_ids}
 
         input_stream: asyncio.Queue[list[int]] = asyncio.Queue()
-        await self._run_generation(self._render_token_prompt(continuation_token_ids), input_stream)
+        await self._run_generation(self._render_token_prompt(continuation_token_ids, multi_modal_data), input_stream)
 
     async def send_json(self, payload: dict):
         try:
--- a/vllm_omni/entrypoints/openai/realtime_connection.py
+++ b/vllm_omni/entrypoints/openai/realtime_connection.py
@@ -4,6 +4,7 @@ import asyncio
 import base64
 import json
 from collections.abc import AsyncGenerator, Mapping
+from dataclasses import dataclass
 from typing import Any, cast
 from uuid import uuid4
 
@@ -20,10 +21,25 @@ from vllm.tokenizers import cached_tokenizer_from_config
 from vllm.transformers_utils.processor import cached_processor_from_config
 
 from vllm_omni.entrypoints.async_omni import AsyncOmni
-from vllm_omni.entrypoints.openai.realtime_tool_calls import ToolCallStreamState, extract_deltas
+from vllm_omni.entrypoints.openai.realtime_tool_calls import ToolCallDelta, ToolCallStreamState, extract_deltas
 from vllm_omni.entrypoints.utils import coerce_param_message_types
 
 logger = init_logger(__name__)
+
+# How long to block on a client tool result before re-checking the connection.
+# Only bounds the wait between liveness checks, not the total wait: a tool may
+# legitimately take a long time, but a client that has gone away must not leave
+# the generation task parked forever.
+_TOOL_RESULT_POLL_S = 0.5
+
+
+@dataclass
+class _PendingToolCall:
+    """A tool call being accumulated from the model's streamed output."""
+
+    call_id: str
+    name: str
+    arguments: str = ""
 
 
 class RealtimeConnection(VllmRealtimeConnection):
@@ -46,12 +62,20 @@ class RealtimeConnection(VllmRealtimeConnection):
       - generation then continues automatically with the tool result appended,
         streaming the model's actual spoken reply as normal.
 
-    No audio is synthesized/forwarded for a turn that turns out to be a tool
-    call - the underlying 3-stage pipeline (thinker->talker->code2wav) still
-    runs end to end for it (there is no clean lower-level hook to skip talker/
-    code2wav without changing the shared orchestrator - see PR description),
-    but the resulting audio bytes are simply not sent to the client.
+    Audio for a tool-call turn is not forwarded to the client once the
+    `<tool_call>` tag has been parsed - the underlying 3-stage pipeline
+    (thinker->talker->code2wav) still synthesizes it (there is no clean
+    lower-level hook to skip talker/code2wav without changing the shared
+    orchestrator - see PR description), the bytes are just dropped. Any audio
+    that arrived before the tag was recognized has already been sent; in
+    practice the tag appears in the thinker's text well ahead of the
+    corresponding synthesized audio.
+
+    A chain of tool calls is bounded by MAX_TOOL_ROUNDS.
     """
+
+    # Upper bound on consecutive tool-call rounds within one user turn.
+    MAX_TOOL_ROUNDS = 8
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
@@ -65,9 +89,11 @@ class RealtimeConnection(VllmRealtimeConnection):
         # because those contain the expanded audio placeholder run with no way to
         # re-attach the audio - see _await_tool_results_and_continue.
         self._turn_prompt: dict[str, Any] | None = None
-        # index (parser-assigned, per generation) -> {"call_id", "name", "arguments"}
-        self._pending_tool_calls: dict[int, dict[str, Any]] = {}
+        # parser-assigned index (per generation) -> the call being accumulated
+        self._pending_tool_calls: dict[int, _PendingToolCall] = {}
         self._tool_result_queue: asyncio.Queue[dict[str, Any]] = asyncio.Queue()
+        # Consecutive tool-call rounds in this turn, bounded by MAX_TOOL_ROUNDS.
+        self._tool_rounds = 0
 
     async def handle_event(self, event: dict):
         event_type = event.get("type")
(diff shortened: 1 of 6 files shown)


vllm_omni/entrypoints/openai/realtime_connection.py
```diff
@@ -208,14 +395,130 @@ async def _run_generation(
...
+        # comment in Qwen3OmniMoeForConditionalGeneration.buffer_realtime_audio
+        # for why relying on safe_apply_chat_template's own auto-resolution
+        # is unsafe for this checkpoint.
+        processor = cached_processor_from_config(model_config)
+        # Multiple tool calls in one turn -> one combined tool-role message,
+        # matching how Qwen's own chat template batches consecutive tool
+        # results under a single <|im_start|>user block. No `tools=` here:
+        # this is a continuation of a conversation that already has the
+        # tools system preamble in its token history, not a fresh turn.
+        combined_result_text = "\n".join(results_by_index[i] for i in sorted(results_by_index))
```
hsliuustc0106: This breaks parallel tool calls. The Qwen3-Omni chat template emits one `<tool_response>` block per `role="tool"` message (while grouping consecutive tool messages under the same user turn). Joining the results creates one `<tool_response>` containing both outputs, so the model cannot associate each result with its call. Please pass one tool message per result in call order and add a two-call round-trip test. See the [model template](https://huggingface.co/Qwen/Qwen3-Omni-30B-A3B-Instruct/blob/refs%2Fpr%2F36/chat_template.json).

vllm_omni/entrypoints/openai/realtime_connection.py
```diff
@@ -4,35 +4,170 @@
...
+        event_type = event.get("type")
+        if event_type == "session.update":
+            tools = event.get("tools")
+            if tools is not None:
+                self._tools = tools
+            await super().handle_event(event)
+        elif event_type == "conversation.item.create":
+            item = event.get("item") or {}
+            if item.get("type") == "function_call_output":
+                self._tool_result_queue.put_nowait(item)
```
hsliuustc0106: Please validate this event before enqueueing it. Any dict with `type="function_call_output"` is accepted, including missing or non-string `call_id`/`output` fields. Later, unknown IDs are only logged and ignored, while `output` is coerced with `str()`; a client typo can therefore leave generation waiting without a client-visible error. Validate the event shape and make missing/unknown IDs and non-string outputs explicit protocol errors, with tests for each case.

vllm_omni/entrypoints/openai/realtime_connection.py
```diff
@@ -4,35 +4,170 @@
...
+        seam for extra per-connection state like tools, so this reimplements
+        its (short) body directly rather than patching upstream vLLM."""
+        stream_input_iter = self.serving.model_cls.buffer_realtime_audio(
+            audio_stream, input_stream, self.serving.model_config, tools=self._tools
+        )
+        async for prompt in stream_input_iter:
+            # Remember the pre-expansion prompt so tool-call continuations can
+            # re-anchor on the user's audio (see self._turn_prompt).
+            if isinstance(prompt, dict):
+                self._turn_prompt = dict(prompt)
```
hsliuustc0106, line 220: This assignment loses earlier audio when `async_chunk=True`. In that mode Qwen's buffer yields one `TokensPrompt` per segment, so the continuation reattaches only the final segment through `_turn_prompt`; the model no longer sees the beginning of the user's utterance. Please aggregate the audio-bearing segments (or explicitly reject/document `async_chunk` for tool calls) and add regression coverage.

hsliuustc0106 (changes requested):
Request changes.

The happy path is promising, but the following correctness issues need to be addressed (see inline comments):

1. Preserve one tool result per `role="tool"` message for parallel calls.
2. Validate malformed and unknown tool results and return client-visible errors instead of waiting silently.
3. Preserve all audio when `async_chunk=True`, or explicitly bound and document that limitation.

Scope note: the issue thread also explicitly listed `tool_choice`, Pydantic event models, and the example-client flags as direct additions for this work; they are not in this diff. Please either include them or narrow the issue/PR scope, and document clearly that this implementation covers the non-duplex path only.

vllm_omni/entrypoints/openai/realtime_connection.py
TheCodeWrangler: You're right, rendered both forms through the checkpoint's own `chat_template` to confirm:

```
# joined (before)
<|im_start|>user
<tool_response>
sunny 72
rainy 55
</tool_response><|im_end|>

# one message per result (after)
<|im_start|>user
<tool_response>
sunny 72
</tool_response>
<tool_response>
rainy 55
</tool_response><|im_end|>
```

Fixed in 181ccce2 — one `role="tool"` message per result, in call order. `sorted()` is over the parser-assigned index, and `extract_deltas` appends `tool_call_starts` in the order `<tool_call>` appears in the generated text, so call order holds even when the client returns the results out of order.

**Two-call round-trip test added** in d5392f42 — `test_two_call_tool_round_trip[sync]`, same `@hardware_test(H100, num_cards=2)` shape as its neighbour. One utterance naming two cities, and the client answers each call with a *different* result (Boston -> "sunny and 72 degrees", Denver -> "snowing and 28 degrees"), so the test asserts that both 72 and 28 reach the spoken reply.

Verified on 2x H100 with real weights (`--run-level advanced_model`): `1 passed in 168.67s`.

Writing it also exposed a latent bug in the existing e2e driver, fixed in the same commit: `tool_call_name` was a single variable and `tool_call_args` a single accumulator, so with parallel calls the interleaved `response.output_item.added` / `.function_call_arguments.delta` / `.done` events overwrote and concatenated across calls. The driver now keys per `call_id` and preserves the order the model requested them in.

Unit coverage alongside it — `TestParallelToolResultsRenderSeparateBlocks`: two calls, two calls with the results submitted in reverse, and a duplicate result for one call.

vllm_omni/entrypoints/openai/realtime_connection.py
TheCodeWrangler: Fixed in 181ccce2, with a gate at each of the two points you identified.

**At ingress** — a new `_enqueue_tool_result` runs before anything is queued:

- `call_id` missing, not a string, or empty -> `invalid_function_call_output`
- `output` not a string -> `invalid_function_call_output` (naming the type it got)

Nothing reaches the queue unless both hold, so a malformed event can no longer influence generation at all.

**At consume** — an unknown `call_id` now sends `unknown_tool_call_id` to the client, listing the ids that were actually pending, instead of only logging. And the `str()` coercion is gone: `output` is guaranteed a string by the time it is read, so the read is now `item["output"]` and a non-string can no longer be silently stringified into the prompt.

`TestToolResultValidation` has a case per shape you named, each asserting the client-visible error code:

| case | expected |
|---|---|
| `call_id` missing | `invalid_function_call_output`, nothing queued |
| `call_id` non-string | `invalid_function_call_output`, nothing queued |
| `output` non-string | `invalid_function_call_output`, nothing queued |
| `call_id` unknown | `unknown_tool_call_id` |
| well-formed | no error, queued |

The first four fail without the fix; the last one pins behaviour that must not regress.

One deliberate decision worth stating, since it is the residue of your point rather than a fix for it: on an unknown `call_id` the turn reports the error and **keeps waiting**. Aborting on a typo would throw away a tool exchange that is still legitimately in flight, and the correct result may well arrive right after the mistaken one. So the client learns immediately and can retry, while abandonment is bounded by disconnect detection rather than a deadline — a slow tool and an absent one are indistinguishable from the server's side, so I did not want to invent a timeout that would break genuinely slow tools. That trade-off is now written on the class docstring rather than left implicit. Happy to add a deadline if you would rather have one, but it felt like a policy choice worth surfacing rather than picking silently.

Note on overlap: @cgloriacc offered pydantic models for the tool events, which would replace these explicit checks rather than sit beside them — the `_enqueue_tool_result` docstring says so. I added the checks now rather than waiting, since the hole you found is live either way.

vllm_omni/entrypoints/openai/realtime_connection.py
TheCodeWrangler: Went with rejecting rather than aggregating, because aggregating on its own isn't enough.

I pulled the gate out temporarily and ran the existing tool round trip against an `--async-chunk` server to see what actually happens. It passes, but for a single-city utterance the model emits the call twice:

```
calls: [{'name': 'get_weather', 'arguments': '{"city": "Boston"}'},
        {'name': 'get_weather', 'arguments': '{"city": "Boston"}'}]
```

The thinker generates per yielded segment, so the accumulated text carries `<tool_call>` twice and the client is asked to run the tool twice. Aggregating the audio doesn't touch that — the duplication comes from generating per prompt, not from which audio is attached.

The `_turn_prompt` line is unchanged for a related reason: the mid-stream yield in `buffer_realtime_audio` only happens under `async_chunk`, so with it off there's exactly one prompt per turn and nothing to lose.

So in 181ccce2, `session.update` carrying `tools` is refused under `async_chunk` with a `tools_require_no_async_chunk` error pointing at `--no-async-chunk`, documented on the class, and covered both ways by `TestToolsRejectedUnderAsyncChunk`. `async_chunk` defaults to off, so this only bites deployments that opted in — though they do lose the lower time-to-first-audio, and I haven't measured how much that is.

The duplicate emission looks like an engine-level thing rather than a tool-calling one, and it reproduces in a couple of minutes.

Sy0307:
Tracked in #6494 as the focused legacy `/v1/realtime` extension stack: #5555 tools, #5565 voice, and #5566 instructions. #3344 is a broader overlapping alternative that also carries multi-turn history. Please resolve the existing tool terminal-event review first and decide which prompt/history implementation is authoritative before merging cumulative branches. The PR body can also replace the two `#TBD` references with #5565 and #5566.

vllm_omni/entrypoints/openai/realtime_connection.py
```diff
@@ -153,35 +382,58 @@ async def _run_generation(
...
+                        if not request_prompt_token_ids:
+                            request_prompt_token_ids = list(output.prompt_token_ids)
 
                     delta_text = first_output.text or ""
                     full_text += delta_text
                     completion_tokens_len += len(new_token_ids)
 
                     if delta_text:
-                        await self.send(TranscriptionDelta(delta=delta_text))
+                        content_delta, tool_deltas = extract_deltas(full_text, tool_state)
```
lishunyang12, line 400: Only parse tool calls when this session has tools configured.

vllm_omni/entrypoints/openai/realtime_tool_calls.py
```diff
@@ -0,0 +1,164 @@
...
+    else:
+        content_end = len(current_text) - _partial_tag_overlap(current_text, _TOOL_CALL_START)
+    content_delta = ""
+    if content_end > state.sent_content_idx:
+        content_delta = current_text[state.sent_content_idx : content_end]
+        state.sent_content_idx = content_end
+
+    for i, start in enumerate(state.tool_call_starts):
+        block_start = start + len(_TOOL_CALL_START)
+        end_idx = current_text.find(_TOOL_CALL_END, block_start)
```
lishunyang12, line 127: Ignore closing tags inside JSON strings so valid tool arguments stay intact.

PR #5384, opened 2026-07-24
[Refactor]: Intrduce x_to_text.py in examples

princepride:
## Purpose

This PR adds a shared `examples/offline_inference/x_to_text/x_to_text.py` entrypoint for text-output inference, supporting T2T and I2T for:

  - `ByteDance-Seed/BAGEL-7B-MoT`
  - `tencent/HunyuanImage-3.0-Instruct`
  - `bytedance-research/MammothModa2-Preview`

It detects the model family, applies the corresponding prompt format, and selects the appropriate AR deploy config. The recipes are updated to use this shared example, and the superseded MammothModa2-specific understanding example is removed.
 ## Test and Results

Reference Image

image

  ### BAGEL T2T

  ```bash
  python examples/offline_inference/x_to_text/x_to_text.py \
    --model ByteDance-Seed/BAGEL-7B-MoT \
    --prompt "Explain multimodal inference in three concise sentences."
```

  Result:

```
  Multimodal inference is the process of combining information from multiple sources or
  modalities (such as text, images, audio, and video) to draw a more comprehensive
  understanding of a concept or situation. It involves integrating data from different senses
  and perspectives to create a more accurate and nuanced interpretation. This approach can be
  particularly useful in fields like natural language processing, computer vision, and human-
  computer interaction, where it can help improve the performance of AI systems by providing
  them with richer and more contextually relevant information.
```

  ### BAGEL I2T

```
  python examples/offline_inference/x_to_text/x_to_text.py \
    --model ByteDance-Seed/BAGEL-7B-MoT \
    --image image.png \
    --prompt "Describe this image in detail."
```

  Result:

```
  The image is a close-up photograph of a person with long, straight brown hair that falls
  past the shoulders. The individual appears to be wearing a white collared shirt and a dark
  tie or scarf. The background is blurred, suggesting an outdoor setting with natural light.
  The focus is on the upper part of the person's body, capturing the details of their attire
  and hair. There are no visible texts or distinctive features that provide additional
  context about the location or event. The style of the image is realistic, with a shallow
  depth of field emphasizing the subject against a softly blurred backdrop.
```

  ### HunyuanImage-3 T2T

```
  python examples/offline_inference/x_to_text/x_to_text.py \
    --model tencent/HunyuanImage-3.0-Instruct \
    --prompt "Explain multimodal inference in three concise sentences."
```

  Result:

```
  The response explained multimodal inference through a funnel-style visual metaphor: text,
  image, and audio inputs are combined and distilled into a single inference result.
```

[...]

PR #5238, opened 2026-07-20
[BugFix][Diffusion] Preserve image payloads with reasoning metadata

akshatvishu:
Resolves #5209

### Purpose

Fix a bug in `vllm_omni/diffusion/output_formatter.py` where image generation models returning auxiliary reasoning or thinking text in metadata (such as SenseNova-U1, BAGEL and HunyuanImage-3.0 when `--think` or CoT is enabled) were incorrectly misclassified as text-only outputs (`final_output_type="text"`). This caused generated image payloads to be discarded (`images=[]`), leading to `AssertionError: No images generated` in CI offline inference tests (`vllm-omni_build_12389_diffusion-star-h100-star-single-gpu.log`).

### Changes

Updated `is_text_output` in `vllm_omni/diffusion/output_formatter.py` to check `postprocess_output.primary_key == "text"` exclusively:
```diff
diff --git a/vllm_omni/diffusion/output_formatter.py b/vllm_omni/diffusion/output_formatter.py
index a36305fc..35f897b9 100644
--- a/vllm_omni/diffusion/output_formatter.py
+++ b/vllm_omni/diffusion/output_formatter.py
@@ -149,10 +149,7 @@ def format_diffusion_outputs(
         "postprocess_time_ms": timings.postprocess_time_s * 1000,
     }

-    # Detect text output: when the pipeline returns a string (e.g.,
-    # SenseNova-U1 / BAGEL single-stage img2text / text2text), wrap it
-    # as a text-type response instead of an image.
-    is_text_output = postprocess_output.primary_key == "text" or "text" in postprocess_output.metadata
+    is_text_output = postprocess_output.primary_key == "text"
```

When a model generates an image payload (`payload={"image": img}`) alongside thinking text metadata (`metadata={"text": {"think_text": ...}}`), `primary_key` is `"image"`. Checking `primary_key == "text"` ensures image payloads are preserved (`images=[img]`) while dedicated text-generation modes (`img2text` / `text2text`) continue to be formatted correctly.

### Testplan

1. **Unit Tests**:
   - Added `test_formatter_preserves_image_with_reasoning_metadata` in `tests/diffusion/test_diffusion_output_formatter.py`, parametrized for SenseNova-U1/BAGEL thinking metadata (`{"think_text": ...}`) and HunyuanImage-3.0 CoT metadata (`{"ar_generated_text": ...}`).
   - Command:
     ```bash
     pytest tests/diffusion/test_diffusion_output_formatter.py
     ```

[...]

Issue #4539, opened 2026-06-18
[RFC]: Unify example model usage — migrate all special-usage examples to standard task examples + declarative model_extras


princepride:
## Motivation

PR #4225 introduced a unified parameter-declaration contract for DiT pipelines: per-model request/response params are declared centrally under `vllm_omni/model_extras/<model>.py` (registered in `registry.py`) via `EXTRA_BODY_PARAMS` / `EXTRA_OUTPUT_PARAMS`, routed into the engine through `extra_body` / `custom_output`, and the model's bespoke example script is deleted in favor of the shared task examples (`text_to_image`, `image_to_image`, …).

#4225 migrated **BagelPipeline** and **SenseNovaU1Pipeline** as the reference. Every other model still ships its own special-usage example directory, which causes:

- New users must learn a different invocation per model instead of one standard task-oriented entrypoint.
- Model-specific knobs are hardcoded in scattered scripts instead of declared once and validated centrally.
- Example scripts drift from the engine API and are frequently untested.

**Goal:** migrate every remaining special-usage model example onto the standard task examples + a `model_extras` declaration, until no model requires bespoke example code.

## Proposed Change

For each model in *Sub-tasks*, the migration mirrors #4225:

1. **Declare** the model's extra request/response params in a new `vllm_omni/model_extras/<model>.py` and register it in `vllm_omni/model_extras/registry.py`.
2. **Route** the pipeline's previously-hardcoded knobs through `extra_body` / `custom_output` (keep a legacy fallback during the transition).
3. **Delete** the dedicated example dir(s) under `examples/offline_inference/<model>/` and `examples/online_serving/<model>/`, folding usage into the relevant shared task example (`text_to_image`, `image_to_image`, `image_to_video`, `text_to_video`, `text_to_audio`, `text_to_speech`, …) — adding only an `extra_body` recipe where the model needs non-default params.
4. **Update e2e tests** so the model is exercised through the standard path (and remove tests tied to the deleted bespoke script).
5. **Update markdown/README** — remove the per-model README and document the model's `extra_body` knobs in the relevant task README / recipe.

### ⚠️ Contribution requirements (mandatory for every sub-task PR)

A sub-task PR will **not be merged** unless it includes **all** of the following — this is the single most important rule of this RFC:

[...]

PR #4225, opened 2026-06-07
[Refactor]Base class for dit pipelines with unified parameter declaration[1/N]

princepride:
## Purpose
#3572
This PR introduces a shared base-class contract for DiT diffusion pipelines so model-specific request and response parameters can be declared in one place:

- Add `DiffusionPipelineBase` with `EXTRA_BODY_PARAMS` and `EXTRA_OUTPUT_PARAMS`.
- Route pipeline-declared `extra_body` fields into `OmniDiffusionSamplingParams.extra_args` for both online serving and shared offline examples.
- Surface pipeline-declared `custom_output` fields in OpenAI-compatible response metrics.
- Migrate BAGEL and SenseNova-U1 pipelines to the new declaration pattern while keeping legacy fallback lookup during migration.
- Remove dedicated BAGEL online/offline example clients and document BAGEL usage through the shared text-to-image examples and recipe.

### Offline Test
```
python examples/offline_inference/text_to_image/text_to_image.py \
  --model ByteDance-Seed/BAGEL-7B-MoT \
  --stage-configs-path vllm_omni/deploy/bagel_single_stage.yaml \
  --prompt "A beautiful sunset over mountains" \
  --height 1024 --width 1024 \
  --num-inference-steps 50 --seed 42 \
  --extra-body '{"timestep_shift": 3.0, "cfg_text_scale": 4.0, "cfg_interval": [0.4, 1.0], "cfg_renorm_type": "global", "cfg_renorm_min": 0.0}' \
  --output /tmp/bagel_t2i_single.png
```

```
python examples/offline_inference/text_to_image/text_to_image.py \
  --model ByteDance-Seed/BAGEL-7B-MoT \
  --stage-configs-path vllm_omni/deploy/bagel.yaml \
  --prompt "A beautiful sunset over mountains" \
  --height 1024 --width 1024 \
  --num-inference-steps 50 --seed 42 \
  --negative-prompt "blurry, low quality" \
  --extra-body '{"timestep_shift": 3.0, "cfg_text_scale": 4.0, "cfg_img_scale": 1.5, "cfg_interval": [0.4, 1.0], "cfg_renorm_type": "global", "cfg_renorm_min": 0.0, "think": false}' \
  --output /tmp/bagel_t2i_multi.png

```

### Online Test
```
vllm serve ByteDance-Seed/BAGEL-7B-MoT --omni --port 8091 \
  --deploy-config vllm_omni/deploy/bagel_single_stage.yaml

```

```
[...]

PR #5559, opened 2026-07-29
[Example] Migrate HunyuanImage-3.0 to model_extras + shared task examples

### LINKED ISSUE #4544: [RFC sub-task] Migrate hunyuan_image3 example to standard task example + model_extras
Sub-task of #4539.

Migrate the **`hunyuan_image3`** model off its dedicated special-usage example onto the standard task example + a declarative `model_extras` registration, following the reference migration in #4225 (BagelPipeline / SenseNovaU1Pipeline).

### Scope
- Special-usage example dir(s): `examples/offline_inference/hunyuan_image3/`

### Steps
1. Add `vllm_omni/model_extras/hunyuan_image3.py` declaring `EXTRA_BODY_PARAMS` / `EXTRA_OUTPUT_PARAMS`, and register it in `vllm_omni/model_extras/registry.py`.
2. Route the pipeline's previously-hardcoded knobs through `extra_body` / `custom_output` (keep a legacy fallback during the transition).
3. Delete the bespoke example dir(s) above; fold usage into the relevant standard task example (`text_to_image` / `image_to_image` / `text_to_video` / `text_to_audio` / etc.) with an `extra_body` recipe for any non-default params.
4. Update e2e test(s) to exercise `hunyuan_image3` through the standard path and pass in CI.
5. Update markdown/README: remove the per-model README, document the knobs in the task README/recipe.

### ⚠️ Mandatory for the PR (will be sent back if missing)
- [ ] A complete, copy-pasteable, reproducible run script (model id + all `extra_body` params + env vars)
- [ ] The **actual run results/output** pasted into the PR (generated image/video/audio + logs; request+response for serving)
- [ ] Updated e2e test passing in CI
- [ ] Updated markdown/README

### Definition of done
- [ ] `vllm_omni/model_extras/hunyuan_image3.py` added + registered
- [ ] Pipeline knobs routed via `extra_body` / `custom_output`
- [ ] Bespoke example dir deleted; usage folded into standard task example
- [ ] e2e test updated + passing
- [ ] README/markdown updated
- [ ] Reproducible script + actual results in PR description

suyanli220:
# [Example] Migrate HunyuanImage-3.0 to model_extras + shared task examples

Closes #4544 (sub-task of RFC #4539).

## Summary

Moves HunyuanImage-3.0-Instruct's standard four-modality path off bespoke
per-model scripts onto declarative `model_extras` registration + the shared
task examples (`text_to_image.py` / `image_edit.py` / `x_to_text.py`),
following the MammothModa2 precedent from #4225.

- Adds `vllm_omni/model_extras/hunyuan_image3.py`: declares
  `EXTRA_BODY_PARAMS` (`bot_task` / `use_system_prompt` / `system_prompt` /
  `negative_prompt`), the text/image prompt builders, and an
  `ar_input_builder` hook.
- Adds `prompt_utils.build_ar_prompt_inputs()`: one seam wrapping
  `build_prompt_tokens` + `resolve_stop_token_ids` (byte-for-byte HF parity,
  with a string fallback when no tokenizer is available), reused by the
  examples and the OpenAI server.
- Wires the shared `text_to_image.py` / `image_edit.py` to invoke a model's
  declared `ar_input_builder` generically (guarded; models without one are
  untouched).
- Adds `--trust-remote-code` to the shared scripts (opt-in, off by default).
- AR-only comprehension (i2t / t2t) routes through the shared
  `examples/offline_inference/x_to_text/x_to_text.py` entrypoint (see
  "Rebase onto upstream/main" below).
- Removes `examples/offline_inference/hunyuan_image3/` entirely (`end2end.py`,
  `reproduce.sh`, README, and its auto-generated docs page); consolidates all
  usage documentation into `recipes/Tencent/HunyuanImage-3.0-Instruct.md`.

RFC #4539 asks for model-specific example directories to collapse into
task-level examples. All four modalities now do: t2i/it2i via
`text_to_image.py` / `image_edit.py`, i2t/t2t via the shared `x_to_text.py`,
and `examples/offline_inference/hunyuan_image3/` no longer has any dedicated
script or docs page. `end2end.py`'s streaming chain-of-thought display and
diffusion KV-cache dtype / skip-steps controls have no equivalent in the
shared examples yet -- accepted as a known gap rather than a reason to keep
the directory, matching every other already-migrated model in the RFC (see
"Second rebase" below for how this played out across review).
`recipes/Tencent/HunyuanImage-3.0-Instruct.md` is the single documentation
home for run commands, deploy-config selection, and hardware notes.

## A real bug found and fixed during GPU verification

Found by actually running the migrated path end-to-end on 4×H200 — not
visible from reading the diff alone.

`registry.py` originally registered HunyuanImage3's `extra_body_params` /
`ar_input_builder` under `"HunyuanImage3Pipeline"` — the diffusers pipeline
class name. But `od_config.model_class_name` at runtime is
`"HunyuanImage3ForCausalMM"` (the diffusion-registry arch key set by
`vllm_omni/model_executor/models/hunyuan_image3/pipeline.py` and consumed via
`vllm_omni/config/stage_config.py`'s `engine_args.setdefault("model_class_name", ps.model_arch)`).
The old key never matched at runtime, so `extra_body` routing
(`bot_task` / `use_system_prompt` / `negative_prompt`) and the
`ar_input_builder` lookup were silent no-ops — the entire declarative
routing this migration adds would never have fired.

Fixed by re-keying the spec to `"HunyuanImage3ForCausalMM"`, keeping
`"HunyuanImage3Pipeline"` as a compatibility alias for any path that
surfaces the pipeline class name instead. Covered by a new regression test:
`test_hunyuan_image3_extra_registry_matches_runtime_model_class_name`.

### Also found (already fixed on main — no change carried here)

GPU verification also hit `ValueError: No images found in request_output`
on every think-mode t2i/it2i request, traced to
`output_formatter.format_diffusion_outputs()`:
`is_text_output = postprocess_output.primary_key == "text" or "text" in postprocess_output.metadata`.
HunyuanImage-3.0's DiT `post_decode()` attaches the AR's chain-of-thought
text to `metadata["text"]` alongside the real image payload (same pattern
as Bagel / SenseNova-U1 "think" mode), so the `or "text" in metadata` clause
misclassified the response as text-only and discarded the image.
Checked current `main` and this is already fixed there in #5238
(`is_text_output` keyed off `primary_key` alone), so no fix is carried in
this PR. Verification below was run with that fix applied locally to
unblock testing.

Also removed a speculative `_recover_images_from_outputs` /
`_tensor_to_pil` deep-scan fallback that had been added to
`text_to_image.py` / `image_edit.py` while chasing the bug above before its
root cause was found — once fixed, HunyuanImage-3.0's DiT output populates
`OmniRequestOutput.images` through the exact same default
`output_formatter.py` path every other diffusers-style model uses, so the
fallback never had a live code path and only added ~150 lines of unverified
heuristics.

## Response to review

- **`--trust-remote-code` opt-in was bypassed**: `_apply_ar_stage_inputs()`
  hardcoded `AutoTokenizer.from_pretrained(model, trust_remote_code=True)`
  regardless of the CLI flag. Fixed: threaded the resolved flag through in
  both `text_to_image.py` and `image_edit.py`.
- **E2E didn't cover this PR's actual new path**: added
  `tests/model_extras/test_shared_script_ar_integration.py` (CPU, calls the
  shared scripts' real `_apply_ar_stage_inputs` against the real
  `model_extras` registry — the path where the registry-key bug above would
  have gone undetected) and `test_shared_script_ar_path_reaches_generation`
  in the GPU e2e file (drives `get_model_class_name` → `get_ar_input_builder`
  → `_apply_ar_stage_inputs` → `Omni.generate()` through the real registry,
  reusing the `en_recaption` reference embedding to also check output parity
  with the hand-constructed path the existing tests use).
- **`stop_token_ids` broadcast too broadly**: was applied to every
  non-diffusion-typed stage by inference. `ARStageInputs` now declares
  `stage_indices` explicitly; the shared scripts apply `stop_token_ids` by
  that index instead of type-based inference.
- **Special-token ids are tied to a specific tokenizer revision**: added a
  registry-declared `ar_tokenizer_validator` hook
  (`get_ar_tokenizer_validator`) that checks a loaded real tokenizer's
  special-token ids against `HUNYUAN_IMAGE3_SPECIAL_TOKEN_IDS` and raises on
  drift.
- **AR tokenizer load failure was silently swallowed**: `_apply_ar_stage_inputs`
  now re-raises by default instead of degrading to the string-prompt form
  (which can complete "successfully" with subtly wrong stop tokens). Added
  an explicit `--allow-tokenizer-fallback` opt-in on both scripts for a
  deliberate offline-compat run.
- **Directory convergence with RFC #4539 incomplete**: documented above
  rather than silently diverging.

**Deferred to a follow-up PR** (re-assessed each individually rather than
batching them; these three specifically need either a live GPU re-run to
verify or a larger, separate design):
- Reusing the engine's already-loaded tokenizer in `_apply_ar_stage_inputs`
  instead of a second independent `AutoTokenizer.from_pretrained` call.
  Concretely: `AsyncOmniEngine._bootstrap_orchestrator` creates its event
  loop as a local variable, never stored on `self`, so nothing outside the
  orchestrator's background thread can currently reach it; exposing
  `get_tokenizer()` to the sync `Omni` wrapper needs that loop stored (e.g.
  `self._loop`) and a `run_coroutine_threadsafe`-based sync wrapper. That
  touches core engine bootstrap/threading code every model goes through,
  not just HunyuanImage3, so it needs a live run to confirm it doesn't
  introduce a deadlock or shutdown-ordering issue -- unlike the fixes above,
  which are scoped to one function with exactly one current caller.
- Pinning a verified model revision for the shared-script commands above —
  blocked on `Omni()` not having a `revision` passthrough today; that'd need
  to land first. (Previously tracked against `reproduce.sh`, since removed
  along with the rest of the directory -- see "Second rebase" below.)

## Rebase onto `upstream/main`

Rebased to pick up 143 commits, including upstream's own independent
`x_to_text.py` shared example + `build_x_to_text_prompt`/
`get_x_to_text_model_family` seam (#5384) — landed after this PR's original
i2t/t2t design, so it made the "generic shared understanding example"
follow-up item above moot and left our dedicated
`run_hunyuan_image3_understanding.py` fully duplicated. Resolved by:

- Dropping `run_hunyuan_image3_understanding.py`; i2t/t2t now go through
  the shared `x_to_text.py`, matching t2i/it2i's use of the shared
  `text_to_image.py` / `image_edit.py`.
- Refactoring `build_x_to_text_prompt` (upstream's function, in the same
  `model_extras/hunyuan_image3.py` file this PR already owns) to delegate to
  `build_ar_stage_inputs` instead of re-deriving the AR prefill / stop-token
  logic inline — so all four modalities share one seam down to
  `prompt_utils.build_ar_prompt_inputs`. Verified this is behavior-preserving:
  for `t2t`/`i2t`, `bot_task` resolves to `None` whether passed explicitly
  (the old code's behavior) or omitted from `extra_body` (the new code's
  behavior; `_normalize_task_and_bot_task`'s legacy-preset table maps both
  tasks' omitted-`bot_task` sentinel to `None`).
- Regenerated the example README + docs page for the new entrypoint, and
  the CPU test that pins "no duplicated tokenization logic"
  (`test_example_routes_through_shared_seam`) to check the seam function
  itself rather than the now-deleted script.

## Second rebase — directory removal, doc consolidation

Rebased again onto a further-diverged `upstream/main`. This raised three
review points from @princepride: (1) fully deleting
`examples/offline_inference/hunyuan_image3/`, (2) folding
`--trust-remote-code` / `--allow-tokenizer-fallback` into a generic
`--extra-args` mechanism instead of dedicated flags, (3) a tokenizer
docstring's clarity.

While resolving the rebase conflict on `end2end.py` (this PR's earlier
commit had deleted it; `upstream/main` had kept it), I found upstream had
independently kept extending that exact file with real, unique
functionality (streaming chain-of-thought display, diffusion KV-cache
dtype / skip-steps control) that has no equivalent in the shared task
examples yet, and initially restored it from `upstream/main` rather than
re-deleting it to avoid silently regressing that functionality. @princepride
weighed that tradeoff explicitly and confirmed the RFC's directory-removal
goal still applies regardless -- that gap is accepted as known, not a reason
to special-case this model. Resolved by fully removing
`examples/offline_inference/hunyuan_image3/` (`end2end.py`, `reproduce.sh`,
README, and its auto-generated docs page) and consolidating all usage
documentation into `recipes/Tencent/HunyuanImage-3.0-Instruct.md`, matching
every other already-migrated model in the RFC. Also, upstream's own
"canonical image task prompts" refactor (#6049) made per-model
`text_to_image_prompt_builder` hooks optional; HunyuanImage3's builder did
nothing beyond reproducing the shared script's own canonical envelope, so
it (plus its registry entry and tests) is deleted, matching the
cosmos3/LingBot precedent from that same upstream commit.

The two other review points (`--extra-args` folding, tokenizer docstring)
were addressed via threaded replies rather than code changes — the CLI
flags are opt-in safety switches with a small closed set of values, not
open-ended model tuning knobs, so keeping them as dedicated flags (with
clear `--help` text) reads better than moving them behind an opaque JSON
blob; and the docstring reply pointed at `validate_ar_tokenizer`'s existing
wording, which already explains *why* it exists (catching an AR/tokenizer
revision drift that would otherwise silently produce wrong stop-token ids)
rather than just what it does.

A separate, earlier reviewer (@hsliuustc0106) had left four inline comments
that were still open with no replies: two turned out to already be fixed in
code from the review-response round above (a test-argv bug, and the
tokenizer validator's base-vs-Instruct false positive), and two were
genuinely unaddressed until this round — `serving_chat.py` calling
`resolve_stop_token_ids` with a different effective `bot_task` than
`build_prompt_tokens`/`build_prompt` used for the same request (fixed, with
a regression test), and `test_system_prompt_scores` only printing its CLIP
score instead of asserting a threshold (fixed, reusing the same
`MIN_SEMANTIC_SIMILARITY` floor `test_shared_script_ar_path_reaches_generation`
already used).

### Round 1 — 4×H200, `tencent/HunyuanImage-3.0-Instruct` (pre-rebase)

Ran `examples/offline_inference/hunyuan_image3/reproduce.sh` end-to-end
(`CUDA_VISIBLE_DEVICES=0,1,2,3`, AR on devices 0-1 / DiT on devices 2-3, per
`vllm_omni/deploy/hunyuan_image_3_moe.yaml`):

| Modality | Result |
|---|---|
| t2i  | ✅ correct, high-quality 1024×1024 output |
| it2i | ✅ completes correctly, extra_args routed correctly, edit applied (image quality artifact noted below) |
| i2t  | ✅ accurate description of the generated image |
| t2t  | ⚠️ unexpected model behavior — see below |

**t2i** ("A brown and white dog is running on the grass"):
t2i_after_fix

**it2i** ("Make the dog neon pink", edited from the t2i output above):
it2i_after_fix

Recolor applied correctly, but the output has visible noise/oversaturation
(see "Known limitations" below — pre-existing, not from this PR's changes).

Unit tests: 51 passed on this pass (`tests/model_extras/test_model_extras.py`,
`tests/diffusion/test_diffusion_output_formatter.py`).

### Round 2 — 4×H200, post-rebase, all four modalities re-run

After the review-response round and the rebase onto `upstream/main` (see
below), re-ran `reproduce.sh` end-to-end with **all four modalities**,
including i2t/t2t for the first time since they were re-routed onto the
shared `x_to_text.py`:

| Modality | Result |
|---|---|
| t2i  | ✅ 36s, correct 1024×1024 output, no artifacts |
| it2i | ✅ 41s, edit correctly applied (neon-pink flowers on a different source image than round 1); same pre-existing noise/oversaturation artifact reproduces on this unrelated input too, reinforcing it's model-inherent, not input-specific |
| i2t  | ✅ accurate, coherent description of the input image ("The image features a vibrant scene with a prominent tower, likely the Tokyo Skytree, set against a clear blue sky...") |
| t2t  | ✅ reproduces the exact known hallucination (model describes an imagined image containing the question text instead of answering "Paris" directly) — confirms this is a stable, pre-existing model behavior, not something the migration changed |

  **t2i** ("A brown and white dog is running on the grass"):
 
image

  **it2i** ("Make the flowers neon pink", edited from a different source image):
image

KV cache transfer (AR→DiT, Mooncake connector, TP=2 both stages) succeeded
cleanly on both TP ranks for both t2i and it2i, and the AR-only i2t/t2t path
(TP=4, `hunyuan_image3_ar.yaml`) initialized and generated without incident.

Environment notes specific to this round (all environment/infra, none in
this PR's code):
- Same `DIFFUSION_ATTENTION_BACKEND=TORCH_SDPA` requirement as round 1 (see
  "Known limitations").
- `Omni()`'s default `init_timeout`/`stage_init_timeout` (600s/300s) was too
  tight for a cold start on a fresh container (no kernel cache) — confirmed
  via logs that the AR worker was still making genuine progress toward
  becoming ready when the timeout fired and tore it down. Not a hang; a
  cold-start budget issue. Not changed in this PR's shipped code — worth a
  follow-up if others hit it on fresh hardware.
- Hit and worked around a pre-existing bug in the (unrelated,
  `vllm_omni/distributed/omni_connectors/`) Mooncake KV-transfer connector:
  its ZMQ port allocation didn't reliably differentiate between a stage's
  TP ranks on a cold multiprocessing spawn, causing an `EADDRINUSE` bind
  race. Out of scope for this PR (core connector infra, not
  `model_extras`/HunyuanImage3 code) — filing a separate issue upstream with
  the repro.

Unit tests: the 51 from round 1 still pass. The review-response round's CPU
tests (`tests/model_extras/test_shared_script_ar_integration.py`, plus
additions to `test_model_extras.py`/`test_prompt_utils.py`) are
syntax-checked and manually traced against real behavior but still not
executed via `pytest` in this environment (no working Python env with
network access to install `pytest` locally). The GPU e2e pytest additions
(`test_shared_script_ar_path_reaches_generation`) require 8 GPUs
(pre-existing `skipif` gate in this test file, unrelated to this PR) and
were not run either round — however, round 2 exercised the *exact same
production code path* that test targets
(`get_model_class_name` → `get_ar_input_builder` → `_apply_ar_stage_inputs`
→ `Omni.generate()`) via the real `reproduce.sh` CLI invocation, which is
the load-bearing evidence for that code path now.

### Known limitations (pre-existing, not introduced by this PR)

- **it2i image quality**: edited images show visible noise/oversaturation.
  Isolated to *not* be caused by the environment's `TORCH_SDPA` attention
  fallback (see below) — t2i, running through the identical DiT model and
  backend, is clean. Root cause likely sits in the pre-existing DiT
  image-conditioning / VAE path, not in code this PR touches. Round 2
  reproduced the identical artifact on a different source image and edit
  prompt, reinforcing that it's model/pipeline-inherent rather than tied to
  one specific input.

- **t2t hallucination**: the model answers a plain-text question by
  describing an imagined image containing that text, instead of answering
  directly. Round 2 reproduced this verbatim in kind (again describing an
  imagined image containing the literal question text) on a live rerun,
  reinforcing the code-level argument below with a second live data point.
  **Confirmed not a regression**: the pre-migration `end2end.py`
  (`_MODALITY_TASK_MAP["text2text"] = ("t2t", None)`) resolves to the exact
  same `task="t2t", bot_task=None` → `en_unified` system-prompt path as the
  migrated code (`build_ar_stage_inputs`/`build_x_to_text_prompt` with
  `task="t2t"`, `bot_task` resolving to `None`). Both call the same
  `build_ar_prompt_inputs()` seam with identical inputs, so this is a
  pre-existing model/prompt-template behavior, not something this PR
  introduced. (A live pre-migration GPU rerun was attempted for a visual
  side-by-side but hit unrelated environment issues — legacy `transformers`
  trust_remote_code prompt handling on the deleted script, vllm/driver
  version skew — that aren't worth resolving for a script this PR removes;
  the code-level comparison above is the load-bearing evidence.)
- This environment's CUDA driver (575.57.08, max CUDA 12.9) required
  installing a `+cu129` wheel variant instead of the PyPI default
  (round 1: vllm 0.25.0; round 2: vllm 0.26.0, matching this repo's
  `docker/Dockerfile.cuda` base image as of the round-2 rebase); the driver
  also predates a flash-attn build exposing the non-varlen `flash_attn_func`
  HunyuanImage3's piecewise-attention path calls (`fa3-fwd` only provides
  `flash_attn_varlen_func`), so both rounds ran with
  `DIFFUSION_ATTENTION_BACKEND=TORCH_SDPA`. Unrelated to this PR's code.

## Repro

`examples/offline_inference/hunyuan_image3/reproduce.sh` is removed along
with the rest of the directory (see "Second rebase" above); the four
modalities now run directly through the shared task examples --
copy-pasteable commands for all four are in
`recipes/Tencent/HunyuanImage-3.0-Instruct.md`. Summarized:

```bash
export MODEL=tencent/HunyuanImage-3.0-Instruct

# t2i
  --model "$MODEL" --deploy-config vllm_omni/deploy/hunyuan_image_3_moe.yaml \
  --trust-remote-code --prompt "A brown and white dog is running on the grass" \
  --height 1024 --width 1024 --guidance-scale 5.0 --num-inference-steps 50 --seed 42 \
  --extra-body '{"bot_task": "think", "use_system_prompt": "en_recaption"}' \
  --output t2i.png

# it2i (set EDIT_IMAGE to a real reference image path)
python examples/offline_inference/image_to_image/image_edit.py \
  --model "$MODEL" --deploy-config vllm_omni/deploy/hunyuan_image_3_moe.yaml \
  --trust-remote-code --image "$EDIT_IMAGE" --prompt "Make the flowers neon pink" \
  --guidance-scale 5.0 --num-inference-steps 50 --seed 42 \
  --extra-args '{"bot_task": "think"}' --output it2i.png

# i2t / t2t (AR-only deploy config, auto-selected)
python examples/offline_inference/x_to_text/x_to_text.py \
  --model "$MODEL" --image "$EDIT_IMAGE" --prompt "Describe the content of this image."
python examples/offline_inference/x_to_text/x_to_text.py \
  --model "$MODEL" --prompt "Explain why the sky appears blue."
```


tests/model_extras/test_shared_script_ar_integration.py
```diff
@@ -0,0 +1,329 @@
...
+) -> None:
+    """The other tests in this file call _apply_ar_stage_inputs directly with
+    an already-resolved trust_remote_code bool, which never exercises
+    argparse itself. Closing that specific gap doesn't need a GPU or a model
+    load -- parse_args() only touches sys.argv -- so cover it here instead of
+    the (expensive, one-model-load-per-invocation) full CLI subprocess a
+    stricter GPU e2e test would need."""
+    module = _load_module(module_name, relpath)
+    monkeypatch.setattr(sys, "argv", ["prog", *argv])
+    args = module.parse_args()
```
hsliuustc0106, line 262: [P1] Seed the required arguments for the image-edit parameterization. `image_edit.parse_args()` declares both `--image` and `--prompt` as required, but the two image-edit cases reach this call with only `[]` or `--trust-remote-code`. `argparse` therefore raises `SystemExit(2)` and the assertion below is never reached. Please add minimal module-specific argv (for example, `--image dummy.png --prompt x`) before appending the flag under test.

vllm_omni/model_extras/registry.py
```diff
@@ -170,6 +187,20 @@ def default_image_to_video_prompt(
...
+    # "HunyuanImage3Pipeline" alias is kept for any path that surfaces the
+    # pipeline class name instead.
+    "HunyuanImage3ForCausalMM": {
+        "extra_body_params": HUNYUAN_IMAGE3_EXTRA_BODY_PARAMS,
+        "extra_output_params": HUNYUAN_IMAGE3_EXTRA_OUTPUT_PARAMS,
+        "init_extra_args_for_non_diffusion_stages": HUNYUAN_IMAGE3_INIT_EXTRA_ARGS_FOR_NON_DIFFUSION_STAGES,
+        "text_to_image_prompt_builder": build_hunyuan_image3_text_to_image_prompt,
+        "image_to_image_prompt_builder": build_hunyuan_image3_image_to_image_prompt,
+        "ar_input_builder": build_hunyuan_image3_ar_stage_inputs,
+        "ar_tokenizer_validator": validate_hunyuan_image3_ar_tokenizer,
```
hsliuustc0106, line 263: [P1] This validator is keyed only by architecture, but both the base and Instruct checkpoints report `HunyuanImage3ForCausalMM`. The hardcoded table contains Instruct-only `<img_ratio_33>`–`<img_ratio_36>` tokens: they are absent from the [base tokenizer](https://huggingface.co/tencent/HunyuanImage-3.0/blob/main/tokenizer.json) and present in the [Instruct tokenizer](https://huggingface.co/tencent/HunyuanImage-3.0-Instruct/blob/main/tokenizer.json). Consequently the shared base-model path fails validation; the new E2E hides this by passing `validate_tokenizer=None`. Please make the hook checkpoint/topology-aware (or validate only the applicable token set) and cover the real production wiring.

vllm_omni/diffusion/models/hunyuan_image3/prompt_utils.py
```diff
@@ -317,12 +345,94 @@ def build_prompt_tokens(
...
+    sys_type: str | None = None,
+    custom_system_prompt: str | None = None,
+    num_images: int = 1,
+    tokenizer: Any | None = None,
+    image_size: str | None = None,
+) -> ARPromptInputs:
+    """Build the complete AR-stage inputs for a HunyuanImage3 request.
+
+    This is the single seam shared by the offline task examples and the
+    OpenAI server: it wraps :func:`build_prompt_tokens` (byte-for-byte HF
```
hsliuustc0106: [P2] The OpenAI server does not currently use this seam: `serving_chat.py` still calls `build_prompt`, `build_prompt_tokens`, and `resolve_stop_token_ids` independently. That can already disagree for I2I when `bot_task` is omitted—the prompt path defaults it, while stop-token resolution receives explicit `None`. Please migrate the server call site to this helper, or narrow this contract/documentation rather than claiming unified behavior.

tests/e2e/offline_inference/test_hunyuanimage3.py
```diff
@@ -341,3 +346,92 @@ def test_system_prompt_scores(
...
+        # MODEL_NAME is the base "tencent/HunyuanImage-3.0" -- if the two tokenizers
+        # differ even slightly, validating here would fail for a reason unrelated to
+        # what this test checks. The validator has its own dedicated coverage in
+        # test_prompt_utils.py / test_shared_script_ar_integration.py.
+        validate_tokenizer=None,
+    )
+
+    outputs = omni.generate(prompt_dict, sampling_params_list=sampling_params_list)
+    generated_image = _extract_generated_image(outputs)
+    score = compare_semantic(SYSTEM_EN_RECAPTION, generated_image, clip_model, clip_processor)
```
hsliuustc0106, line 454: [P2] Please assert the parity criterion here. The test currently computes and prints the CLIP score, so any successfully generated image passes regardless of semantic similarity. Add a justified threshold or tolerance against the reference path so this test can detect a migration regression.

hsliuustc0106 (commented):
Inline findings from review of `c5ad2e00`. Separately, this PR is currently dirty/non-rebaseable against main. Main now contains the shared Hunyuan `x_to_text.py` path and its own `model_extras/hunyuan_image3.py`; please preserve `build_x_to_text_prompt` while resolving the add/add conflict and route comprehension through the shared x-to-text entrypoint rather than retaining the now-duplicated dedicated script.

suyanli220:
> Inline findings from review of `c5ad2e00`. Separately, this PR is currently dirty/non-rebaseable against main. Main now contains the shared Hunyuan `x_to_text.py` path and its own `model_extras/hunyuan_image3.py`; please preserve `build_x_to_text_prompt` while resolving the add/add conflict and route comprehension through the shared x-to-text entrypoint rather than retaining the now-duplicated dedicated script.

@hsliuustc0106 Thank you for the detailed review. I verified the findings.

Current progress:

The P1 issues have been addressed. trust_remote_code now correctly follows the user-provided CLI flag, and I added both CPU integration coverage and GPU E2E coverage for the previously untested get_ar_input_builder → build_ar_stage_inputs → _apply_ar_stage_inputs path.
The P2 items, including overly broad stop_token_ids propagation, token-ID drift validation, and the RFC directory-structure differences, have either been fixed or explicitly documented in the PR.

I am rebasing onto the latest main, which includes PR #5384’s shared x_to_text.py path and introduces an add/add conflict in model_extras/hunyuan_image3.py. I will preserve build_x_to_text_prompt and migrate the HunyuanImage3 I2T/T2T flows to the shared entrypoint instead of retaining the duplicated model-specific script.

After resolving the rebase and conflicts, I will rerun all affected tests in the GPU environment and update the PR with the results.

Thanks again.

suyanli220:
@hsliuustc0106 Following up on my earlier reply — rebase is done and the GPU re-run is complete.

Rebased onto latest `upstream/main` (adopted the shared `x_to_text.py` entrypoint per your comment, preserving `build_x_to_text_prompt` and routing it through `build_ar_stage_inputs`). Re-ran all four modalities end-to-end on 4×H200 (round 2 in the PR description above):

- **t2i**: ✅ correct, clean 1024×1024 output
- **it2i**: ✅ edit applied correctly; same pre-existing noise/oversaturation artifact from round 1 reproduces on an unrelated input, confirming it's model-inherent
- **i2t**: ✅ accurate, coherent description of the input image
- **t2t**: ✅ reproduces the known hallucination behavior verbatim — confirms it's a stable pre-existing model quirk, not something the migration changed

KV cache transfer (AR→DiT, Mooncake connector) succeeded on both TP ranks for t2i/it2i, and the AR-only i2t/t2t path (TP=4) initialized and generated without incident.

Full details, environment notes, and the remaining known limitations (all pre-existing, unrelated to this PR's code) are in the "Verification" section of the PR description.

One honest caveat: the CPU tests from the review-response round are syntax-checked and manually traced but not executed via `pytest` in this environment (no network access to install `pytest` locally), and the GPU e2e pytest addition needs 8 GPUs (pre-existing `skipif` gate, unrelated to this PR) which I didn't have available. Round 2's `reproduce.sh` run does exercise the exact same production code path that GPU e2e test targets, though, which is the load-bearing evidence for it.

princepride:
Please resolve conflicts first.😊

examples/offline_inference/hunyuan_image3/README.md
princepride, line 1: Can we remove the `examples/offline_inference/hunyuan_image3` folder? I think we should let hunyuan image's recipe include it.

examples/offline_inference/image_to_image/image_edit.py
```diff
@@ -420,6 +507,22 @@ def parse_args() -> argparse.Namespace:
...
+        "--allow-tokenizer-fallback",
+        action="store_true",
+        help=(
+            "For models with a declared ar_input_builder (e.g. HunyuanImage-3.0): if loading "
+            "the AR tokenizer fails, degrade to the string-prompt form (no BPE parity) instead "
+            "of failing the run. Off by default -- a tokenizer load failure usually means a "
+            "missing --trust-remote-code or a network/cache issue that's worth surfacing, not "
+            "silently masking with a possibly-wrong prompt."
+        ),
+    )
```
princepride, line 525: Can we let extra-params include these arguments?

vllm_omni/model_extras/hunyuan_image3.py
```diff
@@ -3,35 +3,220 @@
...
+        # HUNYUAN_IMAGE3_PIPELINE / _AR_PIPELINE topologies).
+        stage_indices=[0],
+    )
+
+
+def validate_ar_tokenizer(tokenizer: Any) -> None:
+    """Registry-declared ``ar_tokenizer_validator`` hook for HunyuanImage3.
+
+    Called by the shared task examples right after they load a *real*
+    tokenizer for the AR stage, so a model/tokenizer revision bump that
```
princepride, line 173: Do you mean HunyuanImage-3.0 have changed their tokenizer?

princepride (changes requested):

vllm_omni/model_extras/hunyuan_image3.py
suyanli220: No — the tokenizer hasn't changed. This is a defensive/forward-looking check, not a fix for something that already happened.

examples/offline_inference/image_to_image/image_edit.py
suyanli220: These two flags are a little different from bot_task / use_system_prompt.

bot_task and use_system_prompt affect how a request is built for a specific model. These two instead affect the shared script itself: trust_remote_code controls tokenizer loading, and the fallback flag controls whether tokenizer load failures are tolerated or raised.

Since that behavior is the same for any model using ar_input_builder, I think they fit better as top-level CLI flags than model-specific extra-params.

I’d keep them there unless you had a different separation in mind.

examples/offline_inference/hunyuan_image3/README.md
suyanli220: Good catch — I checked recipes/Tencent/HunyuanImage-3.0-Instruct.md and it already references the shared x_to_text.py example, so the recipe convention here is a good fit. I'll fold the remaining content (the four-modality routing table and reproduce.sh) into that recipe file and remove the examples/offline_inference/hunyuan_image3/ folder entirely, so there's a single documentation home for this model instead of two. Will follow up with that change.

suyanli220: Correction to my earlier reply — after rebasing onto the latest upstream/main, I realized we probably shouldn’t remove this folder completely.

end2end.py has continued to evolve upstream. It now supports things like streaming CoT output and diffusion KV-cache dtype / skip-step controls, which the shared text_to_image.py / image_edit.py path does not support yet. So deleting it would remove existing functionality, not just clean up duplicated scripts.

I’d suggest:

Keep the latest upstream version of end2end.py.
Move the README’s four-modality routing examples into recipes/Tencent/HunyuanImage-3.0-Instruct.md.
Trim reproduce.sh so it only documents the shared-script path.

That leaves a clearer split: the shared scripts are the standard path for the four modalities, while end2end.py remains for the extra features it currently supports, such as streaming and KV-cache controls.

If the goal is still to remove end2end.py entirely, I think we should first port those missing features to the shared scripts. That feels like a separate, larger change.

examples/offline_inference/image_to_image/image_edit.py
princepride: Okay, can you create a sub-task in the following issue to investigate whether Bagel/GLM-Image/Minimax-H3(maybe we have more ar-input model, feel free to add them😊) need these arguments?

vllm_omni/model_extras/hunyuan_image3.py
princepride: If we don't observe this error, I think we can hold off on adding this code to avoid excessive defensive programming.

suyanli220: Got it 👌

examples/offline_inference/image_to_image/image_edit.py
suyanli220: I checked the code directly. Right now, none of the other three models need these flags.

--trust-remote-code and --allow-tokenizer-fallback are only needed for HunyuanImage3 because its ar_input_builder loads the tokenizer again in the shared example script, before the request reaches the engine. That client-side tokenizer load is used to prepare the AR prefill tokens / stop-token IDs for its AR → DiT KV-cache reuse path.

For the other models:

Bagel: its prompt builder only builds strings. It has no ar_input_builder and does not load a tokenizer here.
GLM-Image: it is also AR + DiT, but the tokenizer is loaded inside the engine-side AR model code, which already uses the normal engine trust_remote_code setting. There is no extra client-side tokenizer load.
Minimax-H3: it is a single-stage diffusion pipeline, so this path does not apply.

I also checked model_extras/*.py: HunyuanImage3 is currently the only model that registers an ar_input_builder / ar_tokenizer_validator.

tests/model_extras/test_shared_script_ar_integration.py
suyanli220: Fixed — the image-edit test cases now seed `--image dummy.png --prompt x` before appending the flag under test, so `parse_args()` no longer raises `SystemExit(2)` before the assertion runs.

vllm_omni/model_extras/registry.py
suyanli220: Fixed — `validate_special_token_ids` (in `prompt_utils.py`) now treats a token that the loaded tokenizer doesn't recognize at all (`convert_tokens_to_ids` returning `unk_token_id`/`None`) as "not part of this checkpoint's vocab," not drift. So the base checkpoint's missing `<img_ratio_33>`-`<img_ratio_36>` tokens no longer trigger a false failure; only an id that resolves to something *else* is flagged. The GPU E2E test also now wires the real validator (`get_ar_tokenizer_validator`) instead of passing `None`.

vllm_omni/diffusion/models/hunyuan_image3/prompt_utils.py
suyanli220: Good catch — confirmed and fixed. `serving_chat.py` was calling `resolve_stop_token_ids` with the raw outer `bot_task` variable instead of the same (possibly-omitted) value `build_prompt_tokens`/`build_prompt` used, so an omitted `bot_task` on an it2i request with an explicit size could resolve to a different stop-token set than the prompt that was actually built. Fixed by reusing `build_kwargs`'s own `bot_task` entry for both calls, and added a regression test (`test_image_edits_omitted_bot_task_stop_tokens_match_prompt_default`) asserting they now agree.

tests/e2e/offline_inference/test_hunyuanimage3.py
suyanli220: Fixed — `test_system_prompt_scores` now asserts `score >= MIN_SEMANTIC_SIMILARITY` (0.5, same loose floor `test_shared_script_ar_path_reaches_generation` uses) instead of only printing the score, so a migration regression producing a wrong/blank/garbage image will actually fail the test.

suyanli220:
> Please resolve conflicts first.😊

@princepride All three review points are addressed now, could you take another look when you have a chance?

princepride (commented):
This PR aims to completely remove the `examples/offline_inference/hunyuan_image3` directory and consolidate all usage recipes into `recipes/Tencent/HunyuanImage-3.0-Instruct.md`. 😂

suyanli220:
@princepride Directory fully removed and docs consolidated as discussed. Everything's addressed now. Let me know if there's anything else, otherwise would appreciate an approval when you get a chance!

recipes/Tencent/HunyuanImage-3.0-Instruct.md
```diff
@@ -31,8 +31,10 @@ FP8/NVFP4 configuration:
 ## References
 
 - Model: <https://huggingface.co/tencent/HunyuanImage-3.0-Instruct>
-- Existing model-specific offline example:
-  [`examples/offline_inference/hunyuan_image3`](../../examples/offline_inference/hunyuan_image3)
+- Shared T2I example:
+  [`examples/offline_inference/text_to_image/text_to_image.py`](../../examples/offline_inference/text_to_image/text_to_image.py)
+- Shared IT2I example:
+  [`examples/offline_inference/image_to_image/image_edit.py`](../../examples/offline_inference/image_to_image/image_edit.py)
```
princepride, line 37: I think we can remove these reference files and add `x_to_image.py` examples

princepride (changes requested):
A little recommend, ptal.

recipes/Tencent/HunyuanImage-3.0-Instruct.md
suyanli220: Makes sense as a follow-up to the `x_to_text.py` precedent, but I don't think it belongs in this PR: `text_to_image.py` / `image_edit.py` are ~800 lines each and referenced from 36+ other recipes/docs across the project, so merging them into `x_to_image.py` is a repo-wide shared-infrastructure change touching every model that does t2i/it2i, not something scoped to HunyuanImage3.

Would it be okay to file this as a separate follow-up issue instead, so it gets its own scoped PR (and doesn't block this one)? Happy to open it if so.

princepride (approved):
LGTM

recipes/Tencent/HunyuanImage-3.0-Instruct.md
suyanli220: Filed as #6256.

vllm_omni/model_extras/hunyuan_image3.py
```diff
@@ -3,35 +3,199 @@
...
+            "prompt_token_ids": ar_inputs.prompt_token_ids,
+            "modalities": ar_inputs.modalities,
+            "use_system_prompt": ar_inputs.use_system_prompt,
         },
-        resolve_stop_token_ids(task=task, bot_task=None, tokenizer=tokenizer, image_size="auto"),
+        ar_inputs.stop_token_ids,
+    )
+
+
+def build_image_to_image_prompt(
```
hsliuustc0106: can we reuse from the registry.py?

vllm_omni/model_extras/hunyuan_image3.py
```diff
@@ -3,35 +3,199 @@
...
 from __future__ import annotations
 
+from collections.abc import Mapping
+from dataclasses import dataclass, field
 from typing import Any
 
+from PIL import Image
+
 
 def build_x_to_text_prompt(
```
hsliuustc0106, line 11: same here? can we reuse?

vllm_omni/model_extras/hunyuan_image3.py
suyanli220: Yes, good catch. HunyuanImage3's `build_image_to_image_prompt` only adds `modalities=["image"]` on top of the default builder, and `_apply_ar_stage_inputs` sets the modalities from the AR builder immediately afterwards anyway.

I'll drop the Hunyuan-specific image-to-image builder and let it fall back to `default_image_to_image_prompt` from `registry.py`.

vllm_omni/model_extras/hunyuan_image3.py
suyanli220: This one is a bit different. `registry.py::build_x_to_text_prompt` is already the dispatcher that routes the `hunyuan_image3` family to this function, so calling it from here would recurse back into the same builder.

The generic `default_x_to_text_prompt` isn't sufficient for HunyuanImage3 either, since this path needs the model-specific AR prompt tokenization and stop-token IDs from `build_ar_stage_inputs`.

So I think this Hunyuan-specific builder still needs to stay. If you meant reusing a different helper from `registry.py`, let me know and I can adjust it.

PR #5561, opened 2026-07-29
[Bugfix][Diffusion] Apply MagCache residual once per skipped step

### LINKED ISSUE #5560: [Bug]: MagCache applies the cached residual at every hooked block on a skipped step (output = x + N·residual)
### Your current environment

<details>

```text
vLLM-Omni: main @ fa0658eabb06baea5e278ef5407e84605c3e03f4
vLLM:      v0.26.0
```

Hook forward numerics only — accelerator, model weights and torch version do
not matter. The repro below runs on CPU.

</details>

### Your code version

<details>

```text
v0.26.0
```
</details>
<details>

```text
fa0658eabb06baea5e278ef5407e84605c3e03f4 (main)
```
</details>

### 🐛 Describe the bug

With `cache_backend="mag_cache"`, a skipped step applies the cached residual
once per hooked block instead of once total, producing `x + N·residual`
(N = 57 for Flux). Nothing crashes or warns; outputs are just silently wrong
whenever MagCache actually skips.

The cached residual spans the whole block stack (`tail_output -
head_block_input`, [L405-L431](https://github.com/vllm-project/vllm-omni/blob/fa0658eabb06baea5e278ef5407e84605c3e03f4/vllm_omni/diffusion/cache/magcache/hook.py#L405-L431))
and `MagCacheHeadHook` already applies it on a cache hit
([L143-L201](https://github.com/vllm-project/vllm-omni/blob/fa0658eabb06baea5e278ef5407e84605c3e03f4/vllm_omni/diffusion/cache/magcache/hook.py#L143-L201)).
`MagCacheBlockHook`, registered on every non-head block
([L568-L578](https://github.com/vllm-project/vllm-omni/blob/fa0658eabb06baea5e278ef5407e84605c3e03f4/vllm_omni/diffusion/cache/magcache/hook.py#L568-L578)),
adds it again in its own skip path
([L379-L403](https://github.com/vllm-project/vllm-omni/blob/fa0658eabb06baea5e278ef5407e84605c3e03f4/vllm_omni/diffusion/cache/magcache/hook.py#L379-L403)).
In `diffusers/hooks/mag_cache.py` — cited at the top of this file — that skip
path returns its inputs unchanged.

CPU repro: a 3-block toy transformer with `num_inference_steps=2`,
`retention_ratio=0.5`, `mag_ratios=[1.0, 1.0]` (step 0 computes, step 1 is a
guaranteed cache hit).

```python
import torch, torch.nn as nn
from vllm_omni.diffusion.cache.magcache.config import MagCacheConfig
from vllm_omni.diffusion.cache.magcache.hook import apply_mag_cache_hook

class Block(nn.Module):
    def forward(self, hidden_states, encoder_hidden_states):
        return encoder_hidden_states, hidden_states + 1.0

class Toy(nn.Module):
    def __init__(self):
        super().__init__()
        self.blocks = nn.ModuleList(Block() for _ in range(3))
    def forward(self, h, e):
        for b in self.blocks:
            e, h = b(h, e)
        return h

model = Toy()
apply_mag_cache_hook(model, MagCacheConfig(
    transformer_type="Toy", threshold=0.24, max_skip_steps=5,
    retention_ratio=0.5, num_inference_steps=2,
    mag_ratios=torch.tensor([1.0, 1.0])))

x, e = torch.zeros(1, 4, 8), torch.zeros(1, 2, 8)
print("computed step:", model(x, e).flatten()[0].item())
print("skipped  step:", model(x, e).flatten()[0].item())
```

```text
computed step: 3.0
skipped  step: 9.0     # expected 3.0; residual applied 3x (head + 2 blocks)
```

Related, same divergence: in the single-block configuration the head hook is
also the tail ([L562-L566](https://github.com/vllm-project/vllm-omni/blob/fa0658eabb06baea5e278ef5407e84605c3e03f4/vllm_omni/diffusion/cache/magcache/hook.py#L562-L566)),
and nothing advances `step_index` on a skip, so the `mag_ratios` schedule
stalls and stale state (including `previous_residual`) leaks into the next run.

`tests/diffusion/cache/test_cache_backends.py` mocks `apply_mag_cache_hook`,
so neither is covered.

### Proposed fix

Follow the reference: return inputs unchanged in `MagCacheBlockHook`'s skip
path (keeping the tail-side `advance_step`), and advance the schedule in
`MagCacheHeadHook`'s skip path when it is also the tail. PR with CPU
regression tests that fail on main: #5561.

yashkgp:
## Purpose

Fixes #5560.

The cached MagCache residual is the residual of the whole block stack — the
tail hook stores `tail_output - head_block_input` — and `MagCacheHeadHook`
already applies it once on a cache hit, matching the reference implementation
this file cites (`diffusers/hooks/mag_cache.py`). But `MagCacheBlockHook`,
registered on every non-head block, applied the same full-stack residual
*again* in its own skip path. A skipped step therefore produced
`x + N·residual` instead of `x + residual`, where N is the number of hooked
blocks (57 for `FluxTransformer2DModel`). Every generation in which
`cache_backend="mag_cache"` actually skipped a step silently produced
corrupted latents — no crash, no warning, just "MagCache degrades quality".

This makes `MagCacheBlockHook.new_forward`'s skip path pass its inputs through
unchanged, exactly like the diffusers reference (which returns
`hidden_states` / the `(encoder_hidden_states, hidden_states)` tuple
unmodified), keeping the tail-side `advance_step`. The now-unused residual
fetch and the `apply_residual`/`apply_residual_tuple` calls in that path are
removed; the strategy methods are still used by the head hook, which remains
the single application point.

Second, in the single-block configuration the head hook doubles as the tail
(`_apply_mag_cache_head_hook(..., is_tail=True)`), but its skip path returned
without advancing the step schedule and no `MagCacheBlockHook` exists to do
it. `mag_ratios[current_step]` stalled on the skipped index and the
end-of-run reset never fired, leaking stale skip state (including
`previous_residual`) into the next inference run. The skip path now advances
the schedule when the head hook is also the tail.

## Test Plan

Two CPU regression tests in `tests/diffusion/cache/test_magcache_hook.py`,
marked `core_model` and `cpu` like the neighbouring
`test_cache_backends.py`, so the existing L1 lane picks them up with no
Buildkite changes:

- `test_skipped_step_applies_full_stack_residual_once` — a 3-block toy
  dual-stream transformer with a guaranteed cache hit on step 1
  (`num_inference_steps=2`, `retention_ratio=0.5`, `mag_ratios=[1.0, 1.0]`);
  asserts no block computes on the skipped step **and** the skipped output
  equals the computed one. On main the output is 3× the correct value.
- `test_single_block_advances_schedule_on_skip` — a 1-block model whose head
  hook is also the tail; asserts the skipped final step resets the schedule
  (`step_index == 0`, `previous_residual is None`) and that the next run
  computes instead of replaying the stale residual. On main the schedule
  stalls and the next run silently skips.

```bash
pytest -sv tests/diffusion/cache/test_magcache_hook.py -m 'core_model and cpu'
```

Standalone repro of the original behaviour (no accelerator needed) is in
#5560 — a toy 3-block transformer prints `9.0` on the skipped step on main
where the computed step prints `3.0`.

**vLLM Version:** v0.26.0

**vLLM-Omni Commit:** fa0658eabb06baea5e278ef5407e84605c3e03f4

## Test Result

New tests on this branch:

```text
tests/diffusion/cache/test_magcache_hook.py::test_skipped_step_applies_full_stack_residual_once PASSED
tests/diffusion/cache/test_magcache_hook.py::test_single_block_advances_schedule_on_skip PASSED
2 passed in 0.30s
```

The same tests against main's `hook.py` (fix stashed):

```text
tests/diffusion/cache/test_magcache_hook.py::test_skipped_step_applies_full_stack_residual_once FAILED
tests/diffusion/cache/test_magcache_hook.py::test_single_block_advances_schedule_on_skip FAILED
2 failed in 0.35s
```

Neighbouring suite unaffected: `tests/diffusion/cache/test_cache_backends.py`
→ `28 passed`. `ruff check`, `ruff format --check` and
`tools/pre_commit/check_test_marks.py` are clean on both changed files.

Runs were done on a CPU-only macOS box (no CUDA available); the toy-model
tests exercise the hook forward numerics directly and need no GPU. GPU e2e
with a real MagCache model needs CI.


Gaohan123 (commented):
Please post real local generation results before and after the PR as a comparison. And also the performance result. Thanks

yashkgp:
Thanks! I don't have a CUDA box, so I ran this on CPU against a real pretrained
model: `facebook/DiT-XL-2-256` (28 transformer blocks, no text encoder). MagCache
is enabled the normal way through `MagCacheBackend.enable()`, and the "before"
numbers run main's actual hook code — I pull both `new_forward` methods out of
`git show origin/main:` and bind them on in the same process, so nothing is
retyped.

50 steps, 27 of them skipped by the cache, compared against the same pipeline
with caching off:

| | PSNR vs uncached | pixels off by >0.1 |
|---|---|---|
| main | 11.3 dB | 72% |
| this PR | 29.6 dB | 1.5% |

| this PR (fixed) | main (buggy) | reference (caching off) |
|:---:|:---:|:---:|
| this PR (fixed) | main (buggy) | reference, caching off |
| matches the reference | posterized, colour blown out, pose and background lost | ground truth |

A second seed and class gives the same story: 11.1 dB → 24.6 dB. On main the
image comes out posterized with blown-out colour and the subject loses its pose
and background; with the fix it matches the uncached run.

The baseline is verified rather than assumed: a calibrate-mode run never skips,
and it comes out bit-identical to a completely hook-free run.

I checked the Flux path separately, since that's where the in-tree strategy is
and it takes the tuple branch. The error grows with block count, which is the
fingerprint of this bug: 5.8× worse than the fix at 2 blocks, 21.5× worse at
Flux.1-dev's 57 blocks. That's why full-size Flux is the worst case.

Environment: M4 Pro CPU, torch 2.11.0, diffusers 0.38.0, vLLM 0.26.0.

Gaohan123 (approved):
LGTM. Thanks

PR #5483, opened 2026-07-28
[Test] Give the logprobs scheduler doubles the vLLM 0.26 contracts

tzhouam:
## Summary

`test_omni_ar_scheduler_logprobs` was missed by the 0.26 mock sweep (`bc46e54a`/`763aa865` in #5443): its `_Request` double lacks `num_in_flight_tokens` (read by the `update_from_output` settling loop at `omni_ar_scheduler.py:399`) and its `_free_request` stub returns `None` instead of the `(kv_xfer_params, ec_xfer_params)` 2-tuple, so `test_invalid_logprobs_finish_only_the_affected_scheduler_request` fails on any 0.26-based tree (reproduced on current `main` with the v0.26 environment; passes CI only when deselected/ordered around).

## Test

`pytest tests/core/sched/test_omni_ar_scheduler_logprobs.py` — 6 passed (was 1 failed, 5 passed).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Related issues

- Fixes #5484 (Merge&Ready CI: `test_invalid_logprobs_finish_only_the_affected_scheduler_request` — `AttributeError: '_Request' object has no attribute 'num_in_flight_tokens'`; this PR also fixes the second failure layer behind it, the `_free_request` stub returning `None` instead of the 0.26 2-tuple).

PR #5273, opened 2026-07-21
[bugfix] validate AR sampled-token logprob alignment

hbhflw2000:
## Summary

This PR hardens the AR scheduler logprob contract for RL callers.

When a request asks for logprobs, the scheduler now slices the request-local output and verifies that each sampled-token logprob row corresponds exactly to the generated token at the same position.

## Motivation

RL training consumes rollout sampled-token logprobs as old policy logprobs. Returning shifted, incomplete, non-finite, or token-mismatched rows silently corrupts PPO/GRPO statistics and makes downstream parity diagnostics misleading.

The scheduler previously forwarded `logprobs.slice_request(...)` without validating this contract.

## Changes

Add `_slice_sampled_logprobs(...)` in `OmniARScheduler` and use it for AR requests that generated tokens and requested logprobs.

The helper validates:

- model-runner logprobs are present;
- token-id and logprob arrays are rank-2;
- row counts match generated token count;
- sampled-token columns are non-empty;
- the first returned token id in every row equals the generated token id;
- sampled-token logprob values are finite.

Invalid output now fails fast with a descriptive `RuntimeError`.

## Tests

Added unit coverage for:

- request-local slicing with cumulative generated-token offsets;
- preservation of sampled-token values;
- missing model-runner logprobs;
- generated-token/logprob row-count mismatch;
- sampled-token ID mismatch;
- non-finite sampled-token logprobs.

Also validated the AR path locally with TP1 and TP4. The returned sampled-token rows matched generated tokens with no missing, duplicated, or non-finite values.

Repository pre-commit checks passed for the changed files, including ruff, formatting, typos, and pickle-import checks.

## Compatibility

This does not change sampling, logits, or behavior for requests that do not ask for logprobs.

For requests that explicitly request logprobs, invalid runner output now raises immediately instead of propagating corrupted RL statistics.

PR #5564, opened 2026-07-29
[Bugfix] Trim AR sampled logprob rows with new_token_ids on mid-step stop

### LINKED ISSUE #5563: [Bug]: AR scheduler emits untrimmed logprob rows when a mid-step stop trims new_token_ids (spec decode)
### Your current environment

<details>

```text
vLLM-Omni: main @ fa0658eabb06baea5e278ef5407e84605c3e03f4
vLLM:      v0.26.0
```

Scheduler bookkeeping only, so it does not depend on the accelerator, model,
or torch version. Reproduced with the CPU-only scheduler test harness.

</details>

### Your code version

<details>

```text
v0.26.0
```
</details>
<details>

```text
fa0658eabb06baea5e278ef5407e84605c3e03f4 (main)
```
</details>

### 🐛 Describe the bug

`OmniARScheduler.update_from_output` slices the sampled-token logprobs for a
request **before** `_update_request_with_output` runs, but that call can trim
`new_token_ids` when a stop is hit mid-step (speculative decoding samples
tokens past EOS / a stop token / `max_model_len`). The `EngineCoreOutput` then
pairs the trimmed token list with the untrimmed logprob rows.

The early slice was introduced by #5273 so that a malformed runner payload
terminates only the affected request — that part is fine, but it slices with
the full sampled length:

https://github.com/vllm-project/vllm-omni/blob/fa0658eabb06baea5e278ef5407e84605c3e03f4/vllm_omni/core/sched/omni_ar_scheduler.py#L415-L431

`_update_request_with_output` (inherited from vLLM's scheduler) then trims the
token list in place when a stop hits before the last sampled token —
`del new_token_ids[num_new:]`:

https://github.com/vllm-project/vllm/blob/v0.26.0/vllm/v1/core/sched/scheduler.py#L2006-L2022

and the output is emitted with the stale, longer `new_logprobs`:

https://github.com/vllm-project/vllm-omni/blob/fa0658eabb06baea5e278ef5407e84605c3e03f4/vllm_omni/core/sched/omni_ar_scheduler.py#L544-L565

Upstream vLLM does not have this problem because it slices the logprobs
**after** the trim, with `len(new_token_ids)`:

https://github.com/vllm-project/vllm/blob/v0.26.0/vllm/v1/core/sched/scheduler.py#L1822-L1828

**User-visible impact.** `LogprobsProcessor._update_sample_logprobs` appends
one output position per received row and adds every row's sampled logprob to
`cumulative_logprob`, with no cross-check against the emitted token count:

https://github.com/vllm-project/vllm/blob/v0.26.0/vllm/v1/engine/logprobs.py#L69-L119

So for a spec-decode request that stops at token 1 of e.g. 3 sampled tokens,
any AR stage serving `logprobs` returns:

- `len(choice.logprobs.content) == 3` while `len(token_ids) == 1` — the
  OpenAI-contract invariant of one logprob entry per generated token is
  violated, and the extra entries describe tokens that were never emitted;
- `cumulative_logprob` polluted by the logprobs of the trimmed
  (rejected/past-EOS) tokens.

**CPU repro** (scheduler-level, mirrors the harness style of
`tests/core/sched/test_omni_ar_scheduler_logprobs.py`): schedule one request
with 2 spec tokens, have the runner return 3 sampled tokens `[7, 8, 9]` with 3
logprob rows, and stop on the first token (EOS) in
`_update_request_with_output`, i.e. `del token_ids[1:]`. The emitted
`EngineCoreOutput` has `new_token_ids == [7]` but
`new_logprobs.logprob_token_ids.shape[0] == 3`.

A ready-to-run regression test is included in the linked PR; on main it fails
with:

```text
FAILED tests/core/sched/test_omni_ar_scheduler_logprobs.py::test_mid_step_stop_trims_logprob_rows_with_token_ids
assert 3 == 1  # logprob rows vs emitted tokens
```

### Proposed fix

Keep the early validation slice from #5273 (it is what isolates a bad runner
payload to one request), and re-slice with the trimmed length after
`_update_request_with_output` when the token list got shorter — the same
post-trim slice upstream performs:

```python
if new_token_ids:
    num_sampled_tokens = len(new_token_ids)
    new_token_ids, stopped = self._update_request_with_output(request, new_token_ids)
    if new_logprobs is not None and len(new_token_ids) < num_sampled_tokens:
        new_logprobs = logprobs.slice_request(req_index, len(new_token_ids))
```

`LogprobsLists.slice_request(req_index, n)` returns the first `n` rows of the
request's segment, which matches the head-trim semantics of
`del new_token_ids[num_new:]`, so the kept rows stay aligned with the kept
tokens. PR with the fix + regression test incoming.

yashkgp:
## Purpose

Fixes #5563.

`OmniARScheduler.update_from_output` slices the sampled-token logprobs before
`_update_request_with_output` so that a malformed runner payload terminates
only the affected request (#5273). But `_update_request_with_output` trims
`new_token_ids` in place when a stop is hit mid-step — with speculative
decoding, tokens are sampled past EOS / a stop token / `max_model_len` and
`del new_token_ids[num_new:]` cuts them. The `EngineCoreOutput` was then
emitted with the trimmed token list paired against the untrimmed logprob rows.

Downstream, `LogprobsProcessor._update_sample_logprobs` appends one logprob
position per received row and folds every row into `cumulative_logprob`, so an
AR request that stopped at token 1 of 3 sampled tokens returned 3 logprob
entries for 1 emitted token, and a `cumulative_logprob` polluted by tokens
that were never emitted.

The fix keeps the early validation slice (its purpose is request-local error
isolation) and re-slices with the trimmed length when
`_update_request_with_output` shortened the token list — the same post-trim
`logprobs.slice_request(req_index, len(new_token_ids))` that upstream vLLM
performs after stop handling (`vllm/v1/core/sched/scheduler.py` L1822–L1828 at
v0.26.0). `LogprobsLists.slice_request(req_index, n)` returns the first `n`
rows of the request's segment, matching the head-trim semantics of the stop
path, so the kept rows stay aligned with the kept tokens.

## Test Plan

**vLLM Version:** 0.26.0

**vLLM-Omni Commit:** fa0658eabb06baea5e278ef5407e84605c3e03f4 (main)

New regression test
`test_mid_step_stop_trims_logprob_rows_with_token_ids` in
`tests/core/sched/test_omni_ar_scheduler_logprobs.py` (marked
`core_model` + `cpu`, same harness as the neighbouring #5483 doubles):
one request scheduled with 2 spec tokens, runner returns 3 sampled tokens
with 3 logprob rows, `_update_request_with_output` stops on the first token
and trims the rest. Asserts the emitted output pairs 1 token with exactly
1 logprob row.

```bash
pytest tests/core/sched/test_omni_ar_scheduler_logprobs.py -v
```

## Test Result

On main (fix reverted, regression test kept):

```text
FAILED tests/core/sched/test_omni_ar_scheduler_logprobs.py::test_mid_step_stop_trims_logprob_rows_with_token_ids
assert token_rows.shape[0] == len(output.new_token_ids)  # 3 != 1
1 failed, 6 passed
```

With this PR:

```text
7 passed
```

Full scheduler suite:

```text
$ pytest tests/core/sched/ -q
84 passed, 14 warnings in 0.11s
```

`tools/pre_commit/check_test_marks.py`, `ruff check`, and `ruff format
--check` all pass on the touched files.


Gaohan123 (approved):
LGTM. Thanks

Issue #5558, opened 2026-07-29
[RFC]: /v1/realtime OpenAI-spec conformance gap audit


TheCodeWrangler:
## Motivation

While adding tool-calling to `/v1/realtime` for Qwen3-Omni (#5522, #5555), I read the full upstream protocol (`vllm/entrypoints/speech_to_text/realtime/{protocol,connection}.py`) and vllm-omni's override side by side against OpenAI's actual Realtime API event/field list. This is a systematic diff, not a report of one bug — several of these are already tracked, several aren't tracked anywhere I could find. Filing this as one place to see the whole picture and avoid duplicate issues.

**Status**: three of these gaps now have PRs, split into independent, individually-reviewable PRs rather than one bundled change:
- Tool-calling → #5555
- Voice/speaker selection → #5565
- `instructions` (system prompt) → #5566

Two big items here are **already covered** by existing RFCs and I'm not asking for new work on them, just cross-referencing:
- No server-side VAD, no barge-in/interrupt (`response.cancel`) → #5179, #3745
- No persistent multi-turn session / one-connection-per-turn → #3745 (root-caused in detail there: KV freed on request finish, streaming re-prefills each segment)

Everything else below is, as far as I can tell, **not filed anywhere yet**.

## Client → server events

| OpenAI event | Status |
|---|---|
| `session.update` | Partial: `model` ✅, `tools` → #5555, `voice`/`speaker` → #5565, `instructions` → #5566. Still missing: `modalities`, `input_audio_format`/`output_audio_format`, `input_audio_transcription`, `turn_detection`, `tool_choice`, `temperature`, `max_response_output_tokens` |
| `input_audio_buffer.append` | ✅ |
| `input_audio_buffer.commit` | Reinterpreted: the `final` flag conflates "close the buffer" with "generate a reply now." OpenAI's commit only finalizes a user-turn item; a response is never auto-triggered by commit alone |
| `input_audio_buffer.clear` | 🔴 Missing |
| `conversation.item.create` | Only accepts `function_call_output` (via #5555); any other item type errors |
| `conversation.item.truncate` | 🔴 Missing — no way to inform the server that a user interrupted mid-playback |
| `conversation.item.delete` | 🔴 Missing |
| `response.create` | 🔴 Missing — no way to trigger a reply independent of a buffer commit (e.g. proactive/"speak first" turns, or re-requesting a response) |
| `response.cancel` | 🔴 Missing — tracked via #5179/#3745 |

[...]

PR #5565, opened 2026-07-29
[Frontend] Add voice/speaker selection to /v1/realtime for Qwen3-Omni

TheCodeWrangler:
> [!IMPORTANT]
> **Stacked on #5555.** GitHub only allows `base: main` for a
>
> This PR's own change is **+159/-7 across 3 files** (1 commit).
> Best reviewed once #5555 has landed.

## Summary

`/v1/realtime` had no way to select a TTS voice at all - every session silently used whichever key HF's `talker_config.speaker_id` lists first for the checkpoint (e.g. `chelsie` for `Qwen3-Omni-30B-A3B-Instruct`), with no way to request a different one, even though `/v1/chat/completions` already supports a `speaker` field for this exact purpose.

- `session.update` gains an optional `voice` (or `speaker`) field.
- Threaded through `buffer_realtime_audio()` as a new `speaker` param, set on the yielded prompt's `additional_information={"speaker": [...]}` - the exact same shape `/v1/chat/completions` already uses (`serving_chat.py`), read back out by `talker_preprocess_prefill` via `payload.get("speaker")`.
- `_render_prompt` reapplies `additional_information` to the *rendered* `engine_input` after `render_cmpl_async` returns: that renderer's internal pipeline (`BaseRenderer.process_for_engine_async`) only carries over fields it explicitly knows about and silently drops anything else, so setting `additional_information` on the pre-render prompt alone is not sufficient - this mirrors exactly how `serving_chat.py._preprocess_chat` applies it post-render for `/v1/chat/completions`.

This PR is part of a 3-PR split of a broader `/v1/realtime` spec-conformance audit (#5558): #5555 covers tool-calling, this one covers voice/speaker selection, #TBD covers `instructions` (system prompt). Split into independent PRs so each can be reviewed/merged on its own - this one has no dependency on the others.

## Caveat found and fixed before landing here

My first attempt at this got the plumbing "right" by every check I had (code review, passing unit tests) but was a complete no-op in production - `render_cmpl_async` silently drops any prompt field it doesn't explicitly recognize, so `additional_information` never reached the engine and every turn kept using the default speaker regardless of what was requested. Caught only by actually listening to the resulting audio, not by any automated check. This PR includes the fix from the start (`_render_prompt` reapplies it post-render) rather than shipping the broken version first.

## Test plan

- [x] New L1 unit tests for `handle_event`'s voice/speaker capture (`tests/entrypoints/test_realtime_connection_helpers.py::TestRealtimeConnectionSpeakerRouting`) - 4 tests.
- [x] New L1 regression test for the `render_cmpl_async` drop bug (`tests/entrypoints/test_realtime_connection_helpers.py::TestRenderPromptPropagatesAdditionalInformation`) - 2 tests, asserting `_render_prompt` reapplies `additional_information` to the rendered output when the renderer's mocked return doesn't already include it (simulating the real pipeline's behavior).
- [x] All 44 tests across this file and `test_realtime_tool_calls.py` actually executed (not just collected) against the real `vllm`/`vllm_omni` install inside the serving container, all passing.
- [x] Manually verified end-to-end on a live server: requesting different voices (`aiden` vs `chelsie`) for identical input audio over `/v1/realtime` produces genuinely different, audibly distinct output, confirmed by a human listener recognizing each as the correct named voice - not just "different bytes," which the broken first attempt would also have technically satisfied in some cases.
- No CI YAML changes needed - `tests/entrypoints/` is already swept unconditionally by `test-ready.yml`.

### DIFF
--- a/tests/entrypoints/openai_api/test_qwen3_omni_realtime_websocket.py
+++ b/tests/entrypoints/openai_api/test_qwen3_omni_realtime_websocket.py
@@ -208,6 +208,37 @@ WEATHER_TOOL = {
 }
 
 
+# Two cities in one utterance, to make the model request two calls in one turn.
+TOOL_CALLING_TWO_CITY_PHRASE_TEXT = "What is the weather like in Boston and in Denver right now?"
+
+# Distinct results per city, so the final reply shows each result reached the call
+# that asked for it rather than both being collapsed into one <tool_response>.
+_TOOL_RESULT_BY_CITY = {
+    "boston": "sunny and 72 degrees",
+    "denver": "snowing and 28 degrees",
+}
+
+
+def _tool_result_for(arguments: str) -> str:
+    """Answer a get_weather call based on the city it asked about."""
+    city = (json.loads(arguments).get("city") or "").lower()
+    for key, result in _TOOL_RESULT_BY_CITY.items():
+        if key in city:
+            return result
+    raise AssertionError(f"Tool call asked about an unexpected city: {arguments!r}")
+
+
+def _two_city_tool_calling_pcm16_input() -> bytes:
+    syn = generate_synthetic_audio(
+        10,
+        1,
+        sample_rate=16000,
+        phrase_text=TOOL_CALLING_TWO_CITY_PHRASE_TEXT,
+    )
+    wav_bytes = base64.b64decode(syn["base64"])
+    return _pcm16_mono_16k_from_wav_bytes(wav_bytes)
+
+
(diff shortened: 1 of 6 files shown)


Sy0307:
Tracked in #6494 as the focused voice-selection part of the legacy Qwen `/v1/realtime` extension stack (#5555 → #5565 → #5566). Review should wait for the shared base/prompt-history direction to be selected against broader alternative #3344 and GA-route RFC #6592, so voice plumbing is not duplicated across incompatible connection owners.

PR #5566, opened 2026-07-29
[Frontend] Add instructions (system prompt) support to /v1/realtime

TheCodeWrangler:
> [!IMPORTANT]
> **Stacked on #5555 → #5565.** GitHub only allows `base: main` for a
>
> This PR's own change is **+81/-14 across 3 files** (1 commit).
> Best reviewed once #5565 has landed.

## Summary

`/v1/realtime` previously had no way to set a system prompt at all, so a client could only shape model behavior via whatever the model's own chat template auto-generates - no way to constrain scope, set persona, or otherwise steer behavior via instructions the way `/v1/chat/completions` already supports via a system-role message.

- `session.update` gains an optional `instructions` field, rendered as a system-role message ahead of the audio-input user turn.
- Threaded through `buffer_realtime_audio()` as a new `instructions` param; when set, renders via the model's own chat template (system + user messages) instead of the hardcoded prompt string. No-op when unset - byte-identical rendering to before this change.

This PR is part of a 3-PR split of a broader `/v1/realtime` spec-conformance audit (#5558): #5555 covers tool-calling, #5565 covers voice/speaker selection, this one covers `instructions`. Split into independent PRs so each can be reviewed/merged on its own - this one has no dependency on the others.

## Test plan

- [x] New L1 unit tests for `handle_event`'s `instructions` capture (`tests/entrypoints/test_realtime_connection_helpers.py::TestRealtimeConnectionInstructionsRouting`) - 3 tests.
- [x] All 47 tests across this file and `test_realtime_tool_calls.py` actually executed (not just collected) against the real `vllm`/`vllm_omni` install inside the serving container, all passing.
- [x] Verified mechanically (temporary debug logging against a live server, since removed) that the instructions text reaches the rendered prompt as a proper system-role message.
- No CI YAML changes needed - `tests/entrypoints/` is already swept unconditionally by `test-ready.yml`.

### DIFF
--- a/tests/entrypoints/openai_api/test_qwen3_omni_realtime_websocket.py
+++ b/tests/entrypoints/openai_api/test_qwen3_omni_realtime_websocket.py
@@ -208,6 +208,37 @@ WEATHER_TOOL = {
 }
 
 
+# Two cities in one utterance, to make the model request two calls in one turn.
+TOOL_CALLING_TWO_CITY_PHRASE_TEXT = "What is the weather like in Boston and in Denver right now?"
+
+# Distinct results per city, so the final reply shows each result reached the call
+# that asked for it rather than both being collapsed into one <tool_response>.
+_TOOL_RESULT_BY_CITY = {
+    "boston": "sunny and 72 degrees",
+    "denver": "snowing and 28 degrees",
+}
+
+
+def _tool_result_for(arguments: str) -> str:
+    """Answer a get_weather call based on the city it asked about."""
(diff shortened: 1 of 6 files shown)


Sy0307:
Tracked in #6494 as the focused instructions part of the legacy Qwen `/v1/realtime` extension stack (#5555 → #5565 → #5566). The remaining coordination gate is to choose the authoritative prompt/history owner against broader #3344 and GA-route RFC #6592 before landing this cumulative branch.

PR #774, opened 2026-01-13
[Hardware] Support platforms and plugin system

gcanlin:
PLEASE FILL IN THE PR DESCRIPTION HERE ENSURING ALL CHECKLIST ITEMS (AT THE BOTTOM) HAVE BEEN CONSIDERED.

## Purpose

Close #702.

~The current stage is POC, so let's discuss the design first.~

The PR is almost ready. Feel free to review, thanks!

For better review, this change includes five parts:
1. Platform interfaces design and hardware implementation. **Need to check whether the implementing methods of kinds of hardware correct**.  cc @faaany @tjtanaa 
2. Plugin system is similar to vLLM, **only need to review the loading point whether feasible**.
3. Make diffusion worker hardware-agnostic 
4. Extract diffusion attention interface as one example. cc @ZJY0516 
5. Collect all function about device into platform and replaced by `current_omni_platform.xxx`.

Also cc @Yikun @wangxiyuan @hsliuustc0106 @Gaohan123 @david6666666 @tzhouam or anyone who's interested in it.

TODO(following-up): 
- Make ring attention hardware-agnostic in the following-up PR.
- Remove offload cuda hardcode.

## Test Plan

- [x] CUDA/ROCm: CI

- [ ] NPU:
   - [x] Omni: Qwen2.5-Omni
   - [x] Omni: Qwen3-Omni
   - [x] Diffusion-Only: Qwen-Image

- [ ] XPU

## Test Result

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft.
</details>

**BEFORE SUBMITTING, PLEASE READ <https://github.com/vllm-project/vllm-omni/blob/main/CONTRIBUTING.md>** (anything written below this line will be removed by GitHub Actions)

PR #5568, opened 2026-07-29
[Bugfix][XPU] Add MoTRMSNorm.forward_xpu delegating to forward_native

Joshna-Medisetty:
## Summary

`MoTRMSNorm` crashes on Intel XPU with `NotImplementedError`, breaking MoT models like BAGEL-7B-MoT.

This is a regression: XPU used to land on `forward_native` until #774 added an `is_xpu()` dispatch branch whose base `forward_xpu` raises. Every `CustomOp` subclass then needed an identical `forward_xpu`; `MoTRMSNorm` was missed.

## Fix

Make `CustomOp.forward_xpu` delegate to `forward_native`, consistent with the existing `forward_hip` / `forward_musa` fallbacks. vLLM core fixed the same bug the same way in [vllm-project/vllm#7670](https://github.com/vllm-project/vllm/pull/7670).

## Cleanup

Removes seven now-redundant overrides that only called `forward_native`: `SiluAndMul`, `RotaryEmbedding`, `LayerNorm`, `RMSNorm`, `RMSNormVAE`, `AdaLayerNorm`, `_RMSNorm`.

`_RotaryEmbedding.forward_xpu` is kept — it selects the cached cos/sin path, not the native math. `forward_xpu` under `diffusion/attention/backends/` belongs to `AttentionImpl` and is untouched.

## Test

Verified on a 4-card Intel XPU host.

- BAGEL-7B-MoT (1024x1024, 30 steps, layer-wise offload, TP=1): raises `NotImplementedError` without this change, completes in 84s with it.
- Z-Image-Turbo (25 steps, CPU offload, TP=1): passes, 82s.
- `tests/diffusion/layers/`: 86 passed; the only failures are four pre-existing CUDA-only tests that can't run on XPU (identical on `main`).
- Adds `tests/diffusion/layers/test_custom_op.py`..


vllm_omni/diffusion/layers/mot/mot_layernorm.py
```diff
@@ -82,6 +82,17 @@ def forward_cuda(
             eps=self.variance_epsilon,
         )
 
+    # ------------------------------------------------------------------
+    # XPU fast-path (no XPU RMSNorm kernel; use the pure-PyTorch path)
+    # ------------------------------------------------------------------
+    def forward_xpu(
```
xuechendi, line 88: I would assume forward_native is default option, right? If we don't provide xpu specific op, we don't need forward_xpu explicitly, can you confirm?

Joshna-Medisetty: forward_native is not the default on XPU. For specialized backends, dispatch_forward() in custom_op.py picks a platform-specific method at init time; on XPU it binds forward_xpu, not forward_native [custom_op.py](https://github.com/vllm-project/vllm-omni/blob/main/vllm_omni/diffusion/layers/custom_op.py).

xuechendi:
@Joshna-Medisetty , do you have a reference link from Vllm side?

Joshna-Medisetty:
@xuechendi. I added reference in the description but here it is: [#7670](https://github.com/vllm-project/vllm/pull/7670)

xuechendi (approved):
LGTM.

xuechendi:
@Joshna-Medisetty please fix conflict

PR #5569, opened 2026-07-29
[XPU][Bugfix] Make Omni AR async output path device-agnostic

Joshna-Medisetty:
## Summary
The AR async output builder in `gpu_ar_model_runner.py` uses hardcoded `torch.cuda.*` (streams, events, `set_device`), which breaks the warm-up path on XPU and other non-CUDA accelerators.

## Fix
Replace them with the `torch.accelerator.*` / generic `torch.Event` / `torch.Stream` APIs the repo already mandates :
- `torch.cuda.current_stream()` + `with torch.cuda.stream(s)` → `torch.accelerator.current_stream()` + `torch.accelerator.set_stream()` with `try/finally` restore (same idiom as `vllm_omni/diffusion/ipc.py`)
- `torch.cuda.Event`/`Stream` → `torch.Event`/`torch.Stream`
- `torch.cuda.set_device()` → `torch.accelerator.set_device_index()`

No behavior change on CUDA; unblocks AR-stage models (e.g. Qwen3-TTS) on XPU.

## Test
Qwen3-TTS-12Hz-1.7B runs to completion on Intel XPU with this change.


vllm_omni/worker/gpu_ar_model_runner.py
```diff
@@ -104,7 +104,7 @@ class _AsyncCPUPayloadSnapshot:
     def __init__(
         self,
         payload: Any,
-        ready_event: torch.cuda.Event | None,
```
xuechendi, line 112: I think we have done monkey patch in vLLM side, 
Please follow same method - https://github.com/vllm-project/vllm/blob/main/vllm/v1/worker/xpu_model_runner.py#L43

don't modify vllm_omni/worker/gpu_ar_model_runner.py

xuechendi: update to https://github.com/vllm-project/vllm-omni/blob/main/vllm_omni/platforms/xpu/worker/xpu_ar_model_runner.py accordingly

Joshna-Medisetty: Reverted gpu_ar_model_runner.py and moved the fix into platforms/xpu/utils.py via torch_cuda_wrapper() (same pattern as vLLM xpu_model_runner.py). Added set_device and the missing stream/event/graph patches so the Omni async output path works on XPU without changing the shared GPU runner.

xuechendi (approved):

PR #4981, opened 2026-07-09
Enable Diffusion Image models asynchronous output 

BLANKETusers:
## Summary

This feature moves the D2H (Device→Host) copy and SHM packaging completed by the Worker GPU forward from the main thread of the Worker to the background daemon thread. Enable the default stream of the GPU to immediately start the forward of the next request, eliminating the GPU bubble. This feature is now enabled by default. In the RPC methods, `execute_model_batch `and `execute_model `will use asynchronous output, while `execute_stepwise `and other methods will follow the original path.

## Test Plan
(1) HunyuanImage3 performance test(step=8,concurrency=2)
`pytest -s -v tests/dfx/perf/scripts/run_diffusion_benchmark.py   --test-config-file tests/dfx/perf/tests/test_hunyuan_image_tp4.json`
(2) HunyuanImage3 accuracy test
```
export ENABLE_ASYNC_DIFFUSION_OUTPUT=1
pytest -s -v tests/e2e/accuracy/test_hunyuan_image3.py
```
(3) Helios-Distilled E2E accuracy test
```
export ENABLE_ASYNC_DIFFUSION_OUTPUT=1
pytest tests/e2e/accuracy/test_video_streaming_output_similarity.py -v -s \
    --run-level full_model
```
(4)QwenImage performance test(step=8,concurrency=2)
`pytest -s -v tests/dfx/perf/scripts/run_diffusion_benchmark.py --test-config-file tests/dfx/perf/tests/test_qwen_image_vllm_omni.json `

(5) UT
```
pytest tests/diffusion/test_async_output_data.py \
       tests/diffusion/test_ipc_async.py \
       tests/diffusion/test_result_pump.py \
       tests/diffusion/test_async_output_worker.py \
       tests/diffusion/test_request_scheduler_async.py \
       tests/diffusion/models/hunyuan_image3/test_hunyuan_image3_post_process.py \
       tests/platforms/test_record_device_event.py -v -x
```

### (1) performance
Enable Async Diffusion Output
668f8bb5c42ea695d10a545c06857360

Disable Async Diffusion Output
c58ba42c9aa9b0d2139c595f477e7641

### (2) accuracy
Enable Async Diffusion Output
accracy

### (3)Helios-Distilled E2E accuracy test
e2e

### (4)QwenImage performance test
Enable Async Diffusion Output
9df23741b99a94724307dd6ab44ac930

Disable Async Diffusion Output
6fe326e61cfea7e8f5e9ebf6e2f84936

[...]

PR #5571, opened 2026-07-30
[Bugfix] Record device-agnostic torch.Event on XPU to fix async D2H image corruption [XPU]

tthakkal:
Images generated on XPU can show corrupted pixels at the bottom after #4981. The root cause is that the event gating the async D2H copy never actually establishes a dependency: `torch.Stream.wait_event()` silently no-ops when passed a `torch.xpu.Event`, so the side stream begins copying while the compute stream is still writing the tensor.

## Purpose
#4981 introduced an async D2H flow that records a device event and uses it to gate a side D2H stream:

`XPUOmniPlatform.record_device_event()` returned a `torch.xpu.Event`. The consumer waits on a **generic** `torch.Stream`, and that combination is the one pairing that does not work: `torch.Stream.wait_event` is a C-level binding that silently drops the dependency for a `torch.xpu.Event` instead of enqueuing it. No error is raised — the wait is simply not honored, so the D2H copy races the still-running forward and the host reads a partially written tensor. Because the copy is issued in submission order, the unwritten region is the tail of the buffer, which surfaces as garbage rows at the bottom of the image.

### Fix

Record a device-agnostic `torch.Event`, which dispatches through the accelerator hooks so the wait is honored:

```python
# Before
event = torch.xpu.Event()
# After
event = torch.Event()
```

**This also aligns XPU with the CUDA and ROCm platforms, which already use `torch.Event()`.**

### Files Changed

- `vllm_omni/platforms/xpu/platform.py`

Before this change:
image

After change:
image

## Test Plan

**vLLM Version:**

**vLLM-Omni Commit:**

## Test Result


dsocek (approved):
LGTM

dsocek:
@xuechendi, @zhenwei-intel   This bug is present in any XPU image generation, causing bad image quality for other models too. This fix from @tthakkal works even for simple z-image example:

```py
from vllm_omni.inputs.data import OmniDiffusionSamplingParams

if __name__ == "__main__":
    omni = Omni(model="Tongyi-MAI/Z-Image-Turbo")
    params = OmniDiffusionSamplingParams(
        height=1024,
        width=1024,
        num_inference_steps=9,
        guidance_scale=0.0,
        seed=42,
    )
    outputs = omni.generate("a red apple on a white table", params)
    outputs[0].request_output.images[0].save("zimage.png")
```

Before fix:
image

After fix:
image

Can you please help review or delegate for quick turnaround?

xuechendi:
I don't think it is a good design to add `xpu.synchronize()` here and there as hot fix. Need to plan for a more consistent solution. Since this is TP issue, can we add to collectives?

dsocek:
Thanks for quick comment @xuechendi, this is not TP issue per se, the example I shared is single card case.

xuechendi:
@Joshna-Medisetty , can you check, does our weekly validation shows such issue?

vllm_omni/platforms/xpu/platform.py
```diff
@@ -101,8 +101,20 @@ def synchronize(cls) -> None:
...
+        and that C-level binding silently no-ops for a ``torch.xpu.Event``
+        instead of enqueuing the dependency — the side stream then starts its
+        D2H copy while the compute stream is still writing the tensor, so the
+        host reads a partially-written image (garbage rows at the bottom of the
+        output). ``torch.Event`` dispatches through the accelerator hooks and
+        the wait is honored, which is the actual fix.
+        """
         try:
-            event = torch.xpu.Event()
+            event = torch.Event()
```
tthakkal, line 118: @xuechendi found a better fix that aligns XPU with the CUDA and ROCm platforms, which already use `torch.Event()`. updated PR description as well

Joshna-Medisetty: @xuechendi @hsliuustc0106 
LGTM

xuechendi (approved):
LGTM

tthakkal:
@xuechendi  thanks for the approval! could you kick off the tests and merge when they're green!!

tthakkal:
@xuechendi Failures look network issue, could you re trigger

tthakkal:
@xuechendi Could you re trigger now

hsliuustc0106:
I think we need to add a tag like xpu-test/rocm-test, cc @yenuo26 @khluu

tthakkal:
@xuechendi @hsliuustc0106 current build failures are timeout and amd related tests, could you help me here https://buildkite.com/vllm/vllm-omni-amd-ci/builds/10414/list
image

tthakkal:
@hsliuustc0106 @xuechendi — could you help get this one through? It's approved and all GitHub checks are green, but the merge is still blocked. Would really appreciate you pushing it over the line. Thanks!

PR #5579, opened 2026-07-30
[Perf][Docker][ROCm] Cache vLLM-Omni Dependencies Before Source Copy

### LINKED ISSUE #5576: [Perf][Docker][ROCm] Cache vLLM-Omni Dependencies Before Source Copy

The current AMD Dockerfile reruns Python dependency installation after source changes because it copies the full repo before it runs `uv pip install`.

#### Evidence

In `docker/Dockerfile.rocm`:

- lines 62 to 63 run `COPY . ${COMMON_WORKDIR}/vllm-omni`
- lines 82 to 85 run `uv pip install ... --no-cache-dir ".[dev]" --no-build-isolation`

That order means a source change invalidates the `COPY .` layer and forces the later install layer to run again.

In [vllm-omni-amd-ci_build_10021_amd-docker-build-image.log](https://github.com/user-attachments/files/30529229/vllm-omni-amd-ci_build_10021_amd-docker-build-image.log):

- steps 7 to 12 are `CACHED`
- step 13 runs `COPY . /app/vllm-omni`
- step 15 reruns `uv pip install`
- step 15 takes 11.3 seconds
- step 17 exports layers in 7.6 seconds
- the push output includes two large uploaded layers, reported as 621.7 MB and 1.786 GB

This proves that, in a warm-host case, early layers can be cached but the source copy and Python install still rerun.

#### Recommended change

Use the same dependency-first layout that `docker/Dockerfile.ci` already uses, while keeping the AMD-specific steps in `docker/Dockerfile.rocm`.

The main optimization is the reorder:

- copy `pyproject.toml`, `setup.py`, and `requirements/` before install
- run editable install before the full source copy
- keep the full repo copy in a later layer

The uv cache mount is a secondary optimization. It can reduce repeated downloads when the install layer still runs, but it does not replace the reorder.

The AMD-specific steps that must stay are:

- the TorchCodec build before source copy
- the root `conftest.py` workaround written during image build
- the `python3 -c "from torchcodec.decoders import VideoDecoder"` validation

#### Additional context

This is a starting diff for `docker/Dockerfile.rocm`. It is not a tested patch.

```diff
diff --git a/docker/Dockerfile.rocm b/docker/Dockerfile.rocm
index <old>..<new> 100644
--- a/docker/Dockerfile.rocm
+++ b/docker/Dockerfile.rocm
@@
-# Step 3: Copy vllm-omni code and install without uv
-RUN mkdir -p ${COMMON_WORKDIR}/vllm-omni
-COPY . ${COMMON_WORKDIR}/vllm-omni
+# Install dependencies before the full source copy so source-only changes
+# do not invalidate the dependency install layer.
+RUN mkdir -p ${COMMON_WORKDIR}/vllm-omni/vllm_omni && \
+    touch ${COMMON_WORKDIR}/vllm-omni/vllm_omni/__init__.py
+WORKDIR ${COMMON_WORKDIR}/vllm-omni
+COPY pyproject.toml setup.py ./
+COPY requirements/ requirements/
+ENV SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0
+RUN --mount=type=cache,target=/root/.cache/uv \
+    uv pip install --python "$(python3 -c 'import sys; print(sys.executable)')" \
+        --no-build-isolation -e ".[dev]"
+
+# Keep the full repo copy after the dependency layer.
+COPY . ${COMMON_WORKDIR}/vllm-omni
@@
-RUN cd ${COMMON_WORKDIR}/vllm-omni && \
-    uv pip install --python "$(python3 -c 'import sys; print(sys.executable)')" \
-        --no-cache-dir ".[dev]" --no-build-isolation && \
-    python3 -c "from torchcodec.decoders import VideoDecoder"
+# Keep TorchCodec validation, but do not reinstall dependencies here.
+RUN python3 -c "from torchcodec.decoders import VideoDecoder"
```

akshatvishu:
Resolves #5576

### DIFF
--- a/docker/Dockerfile.rocm
+++ b/docker/Dockerfile.rocm
@@ -58,8 +58,18 @@ RUN --mount=type=cache,target=/root/.cache/pip \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
-# Step 3: Copy vllm-omni code and install without uv
-RUN mkdir -p ${COMMON_WORKDIR}/vllm-omni
+# Step 3: Install vllm-omni dependencies before the full source copy so
+# source-only edits do not invalidate the dependency layer.
+RUN mkdir -p ${COMMON_WORKDIR}/vllm-omni/vllm_omni && \
+    touch ${COMMON_WORKDIR}/vllm-omni/vllm_omni/__init__.py
+WORKDIR ${COMMON_WORKDIR}/vllm-omni
+COPY pyproject.toml setup.py ./
+COPY requirements/ requirements/
+RUN --mount=type=cache,target=/root/.cache/uv \
+    SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0 \
+    uv pip install --python "$(python3 -c 'import sys; print(sys.executable)')" \
+        --no-build-isolation -e ".[dev]"
+
 COPY . ${COMMON_WORKDIR}/vllm-omni
 
 # This is a workaround to ensure pytest exits with the correct status code in CI tests.
@@ -79,11 +89,9 @@ RUN printf '%s\n' \
     '    os._exit(_exit_code)' \
     > ${COMMON_WORKDIR}/vllm-omni/conftest.py
 
-RUN cd ${COMMON_WORKDIR}/vllm-omni && \
-    uv pip install --python "$(python3 -c 'import sys; print(sys.executable)')" \
-        --no-cache-dir ".[dev]" --no-build-isolation && \
-    python3 -c "from torchcodec.decoders import VideoDecoder"
+RUN python3 -c "from torchcodec.decoders import VideoDecoder"
 
+WORKDIR ${COMMON_WORKDIR}
 RUN ln -sf /usr/bin/python3 /usr/bin/python
 
 FROM base AS test


akshatvishu:
Potentially, we're changing the Docker CI workflow for AMD so this has to land after that!

docker/Dockerfile.rocm
```diff
@@ -79,11 +89,9 @@ RUN printf '%s\n' \
     '    os._exit(_exit_code)' \
     > ${COMMON_WORKDIR}/vllm-omni/conftest.py
 
-RUN cd ${COMMON_WORKDIR}/vllm-omni && \
-    uv pip install --python "$(python3 -c 'import sys; print(sys.executable)')" \
-        --no-cache-dir ".[dev]" --no-build-isolation && \
-    python3 -c "from torchcodec.decoders import VideoDecoder"
+RUN python3 -c "from torchcodec.decoders import VideoDecoder"
```
alex-jw-brooks, line 92: I think this should be moved to before the `COPY . ${COMMON_WORKDIR}/vllm-omni`, because then it'll cache this RUN if the non dep files change :)

Also, do you know if there is a reason that we test the torchcodec import again here? The install script already does this [here](https://github.com/akshatvishu/vllm-omni/blob/da0e0d31ecd5945fc5eb08e7cccc66759dc2a015/tools/install_torchcodec_rocm.sh#L120), so a bit curious why it is needed unless Omni's dep installs can break it somehow

alex-jw-brooks (changes requested):
Thanks for working on this! One suggestion

PR #5311, opened 2026-07-22
[diffusion][feature] Add LingBot-Video T2I and TI2V generation modes

wtz2333:
# [diffusion][feature] Add LingBot-Video T2I and TI2V generation modes

## Summary

This PR extends the LingBot-Video Base support introduced by #5035 with two
additional generation modes:

- text-to-image (T2I);
- text-image-to-video (TI2V) with one first-frame image.

Both `robbyant/lingbot-video-dense-1.3b` and
`robbyant/lingbot-video-moe-30b-a3b` use the same
`LingBotVideoPipeline`. The existing Dense/MoE transformer implementation is
unchanged. Existing T2V requests are normalized through the same request
contract for consistency, but T2V support itself is not introduced by this PR.

### Generation modes and request handling

- Added a shared request contract for T2I, T2V, and TI2V mode selection.
- Added normalization and validation for plain or structured captions,
  negative prompts, frame count, FPS/duration, dimensions, and official
  resolution presets.
- Preserved the difference between an omitted negative prompt and an
  explicitly empty negative prompt.
- Resolved `num_frames` via the `num_frames != 1` heuristic (`1` is the
  image-model default/omitted sentinel, matching Wan2.2-S2V); explicit
  `num_frames` takes precedence over `seconds`/`duration`, and LingBot video
  lengths are aligned upward to the causal VAE `4n+1` grid. Full
  `seconds`/`fps` normalization is deferred to a follow-up.
- Output dimensions flow through the standard `width`/`height`/`size` request
  fields (no model-specific dimension resolver or `extra_params` aliases);
  LingBot aligns them to its 16-multiple grid internally.
- Kept the legacy modality-free T2V fallback while using explicit output
  modalities in all new T2I and TI2V paths.

### T2I and TI2V pipeline support

- Added one-frame latent generation and image output for T2I.
- Added TI2V conditioning through both Qwen3-VL image tokens and a VAE clean
  temporal prefix.
- Applied the official aspect-preserving resize and center-crop policy once,
  then reused the aligned image for both conditioning paths.
- Restored the clean temporal prefix before the first denoise step and after
  every scheduler step.
- Reused the existing denoising loop for sequential CFG, batched CFG, and
  CFG-disabled execution.
- Preserved explicit `image` and `video` output keys through post-processing so
  the offline and OpenAI-compatible formatters retain the output modality.

### Offline and online serving

- Extended the shared offline T2I and image-to-video entry points with LingBot
  prompt builders instead of adding duplicate model-specific scripts.
[...]

PR #5035, opened 2026-07-11
[diffusion][model] Add LingBot Video dense and MoE support

wtz2333:
## Summary

This PR adds native LingBot Video text-to-video support for both checkpoints:

- `robbyant/lingbot-video-dense-1.3b`
- `robbyant/lingbot-video-moe-30b-a3b`

It incorporates the dense foundation originally proposed in [PR #4996](https://github.com/vllm-project/vllm-omni/pull/4996) and adds the routed-MoE implementation, serving coverage, nightly coverage, and reproducible numerical-parity tools. Because this PR targets `vllm-project:main`, both the dense and MoE paths are in scope for review; #4996 is not a merge prerequisite for this PR.

The supported public path is native single-GPU BF16 text-to-video inference through offline generation and the OpenAI-compatible `/v1/videos` API.

### Dense LingBot Video foundation

- Added the native `LingBotVideoPipeline`, FlowUniPC scheduler integration, prompt encoding, latent preparation, classifier-free guidance, denoising loop, and VAE decode path.
- Added the checkpoint-driven LingBot transformer with dense MLP blocks, 3D RoPE, packed-CFG attention fallback, and the required FP32 precision policy for sensitive normalization and modulation paths.
- Registered `LingBotVideoPipeline` in the diffusion registry and added LingBot model extras for serving defaults and supported sampling parameters.
- Added the dense offline example, `/v1/videos` serving tests, supported-model documentation, recipe, DFX configuration, and dense end-to-end parity benchmark.
- Kept the runtime native to vLLM-Omni; it does not import the upstream `lingbot_video` Python package.

### Routed MoE transformer

- Added the checkpoint-compatible token-choice router.
- Preserved the reference routing asymmetry:
  - bias-corrected scores select experts;
  - bias-free scores provide expert weights.
- Added group-limited top-k routing for the 128-expert, top-8 checkpoint configuration.
- Kept router weights and correction bias in FP32 while loading routed and shared experts in BF16.
- Added checkpoint-native grouped expert weights:
  - `w1 [E, I, H]`
  - `w2 [E, H, I]`
  - `w3 [E, I, H]`
- Added vectorized per-expert token padding and CUDA `torch._grouped_mm` execution.
- Added a per-expert PyTorch fallback for CPU/unit validation.
- Restored token order with the reference FP32 scatter-weighted reduction.
- Added shared-expert execution and padding-mask handling.
- Selected dense or sparse FFNs from checkpoint config while preserving dense-checkpoint behavior.

### Serving behavior

[...]

PR #5581, opened 2026-07-30
[diffusion][feature] Add LingBot-Video native loading, HSDP, optional Refiner, and Cache-DiT

wtz2333:
# [diffusion][feature] Add LingBot-Video native loading, HSDP, optional Refiner, and Cache-DiT

## Summary

This PR extends the LingBot-Video Base support introduced by #5035 and the
T2I/TI2V request contract added by #5311 with the deployment, capacity, and
acceleration support required by the official package:

- native Base and Refiner weight loading;
- model-level and layerwise CPU offload;
- reusable pipeline stages and diffusion stage profiling;
- HSDP capacity support for LingBot's mixed-dtype Dense/MoE transformers;
- optional Base-to-Refiner execution for T2V and TI2V;
- Cache-DiT acceleration for the Base transformer and optional Refiner.

Both `robbyant/lingbot-video-dense-1.3b` and
`robbyant/lingbot-video-moe-30b-a3b` continue to use
`LingBotVideoPipeline`. Refiner and Cache-DiT remain opt-in. Existing
Base-only T2I/T2V/TI2V behavior is unchanged when they are disabled.

### Native loading and CPU offload

- Constructs LingBot transformers from config and loads strict native Base and
  Refiner sources through `AutoWeightsLoader`, without eagerly materializing
  weights through `from_pretrained()`.
- Separates construction/load placement from the final execution device.
- Preserves BF16 bulk parameters and the existing FP32 norm, AdaLN, and router
  islands.
- Declares the DiT, text encoder, and VAE components required by generic
  component discovery.
- Supports model-level sequential CPU offload and ordinary layerwise offload.
- Keeps VAE encode/decode device transitions explicit and restores component
  placement after successful execution, latent returns, and exceptions.

### Pipeline stages and profiling

- Adds immutable execution and Refiner option objects and validates them before
  model execution.
- Splits condition preparation, denoising, canonical decode, Refiner handoff,
  Refiner conditioning, and public output formatting into reusable stages.
- Reuses one denoising primitive for Base and Refiner.
- Preserves T2I, T2V, TI2V, sequential CFG, batched CFG, and CFG-disabled
  execution when Cache-DiT is disabled.
- Integrates `DiffusionPipelineProfilerMixin` for Base condition/diffuse, VAE
  decode, Refiner handoff/condition, and Refiner diffuse stages. Internal stage
  durations are returned only when profiling is enabled.

### HSDP capacity support

- Uses complete `LingBotVideoBlock` instances as HSDP shard boundaries and
  includes the root module for parameters outside the block stack.
- Keeps grouped MoE expert tensors with their owning transformer block.
- Preserves LingBot's mixed parameter dtypes during FSDP all-gather without
  changing the default behavior for other models.
- Uses the generic outermost-DiT discovery contract to shard both Base and
  Refiner when the Refiner is loaded.
- Keeps HSDP and TP mutually exclusive.
- Keeps the existing single-GPU MoE T2V/TI2V nightly smoke unchanged.
- Adds a separate HSDP2 T2V/TI2V case marked `slow`, so the existing weekly
  H100 multi-GPU sweep collects it without a new job.

### Optional official Refiner

- Adds immutable startup configuration under
  `model_config.lingbot_refiner`.
- Loads the official `refiner/` package through a second strict native source
  with `prefix="refiner_transformer."`.
- Keeps Refiner disabled by default and Base-only as the per-request default
  when Refiner is loaded with `default_run=false`.
- Rejects invalid Refiner requests before Base compute and keeps T2I on the
  Base-only path.
- Implements the official in-memory T2V/TI2V handoff:

  ```text
  Base latent
  -> shared VAE decode
  -> temporal sample and bicubic resize
  -> shared VAE encode
  -> official x_up/noise mixture
  -> Refiner denoise
  -> final shared VAE decode
  ```

- Shares Qwen3-VL, the processor, and VAE while keeping Base/Refiner weight
  sources, schedulers, generators, sigma schedules, and denoising state
  independent.
- Reinjects the geometry-aligned clean TI2V first-frame latent after every
  Refiner scheduler step.
- Keeps Base and Refiner mutually exclusive on GPU under model-level CPU
  offload and returns only the final refined output.

### Cache-DiT

- Declares `LingBotVideoBlock` as Cache-DiT
  `ForwardPattern.Pattern_3`.
- Uses separate conditional/unconditional cache state for LingBot's sequential
  two-pass CFG.
- Installs one adapter for Base and the optional Refiner, then refreshes the two
  cache contexts independently after each scheduler resolves its actual
  timestep count.
- Refreshes cache state for every request so one request's schedule cannot leak
  into the next.
- Tracks the multi-transformer adapter as the exact teardown target and clears
  Base/Refiner stage refreshers when request-scoped caching is disabled.
- Supports model-level CPU offload and ordinary layerwise offload.
- Rejects unvalidated combinations before model prefetch: PP, TP, EP, SP,
  CFG parallelism, VAE patch parallelism, HSDP, and distributed layerwise
  offload.
- Requires sequential two-pass CFG on every active denoising stage. The engine
  dummy warmup is exempt, and the next real request refreshes the context.
- Adds a balanced Dense T2V preset to the dedicated offline example and online
  documentation: `Fn=4`, `Bn=0`, eight warmup steps, at most 16 cached
  steps, residual threshold `0.12`, medium dynamic SCM, and no consecutive
  cached steps. The offline example accepts the official structured prompt
  JSON directly; Cache-DiT remains configurable for other workloads.

### Online TI2V size contract

Real HSDP2 validation exposed an existing serving gap: `/v1/videos` forwards
model-specific `size` through sampling `extra_args`, while LingBot request
normalization previously read only prompt fields. TI2V therefore fell back to
480x480 even when the request specified 320x192.

The fix only merges `height`, `width`, `size`, `resolution`, and
`ratio` from `extra_args` into the existing dimension resolver. A CPU
regression and the MoE/HSDP2 online T2V+TI2V case cover this behavior.

### Tests, examples, and documentation

- Adds Cache-DiT metadata, installation, refresh, sequential-CFG, Base/Refiner,
  dummy-run, offload, and early-rejection regression coverage.
- Retains the existing loader/offload, HSDP, Refiner math/lifecycle, metadata,
  profiler, and failure-path tests.
- Extends the LingBot offline example and online guide with Cache-DiT usage,
  structured prompt JSON input, and the validated balanced preset.
- Updates the diffusion feature matrix.
- Leaves `.buildkite/cuda/test-nightly.yml` identical to current `main`.
- Routes the new H100 feature coverage through the existing weekly marker
  sweeps: Dense Cache-DiT + profiler, MoE CPU offload + Cache-DiT + Refiner,
  and two-GPU HSDP.

## Relationship with earlier PRs

| PR | Responsibility |
| --- | --- |
| #5035 | Shared LingBot Base pipeline and Dense/MoE T2V transformer support |
| #5311 | T2I/TI2V request contract, conditioning, output handling, and serving integration |
| This PR | Native loading, offload, profiling, HSDP capacity support, optional official Refiner, and Cache-DiT |

This PR builds the deployment and acceleration layer around the existing
generation contracts; it does not replace their transformer math or public
T2I/T2V/TI2V behavior.

## Scope and compatibility

| Mode | Base only | Base -> Refiner | Cache-DiT |
| --- | --- | --- | --- |
| T2I | Supported | Rejected | Base |
| T2V | Supported | Optional | Base and loaded Refiner |
| TI2V | Supported | Optional | Base and loaded Refiner |

Refiner and Cache-DiT are startup features, not dynamically loaded per request.
Cache-DiT currently requires an unsharded DiT stage and sequential two-pass
CFG. Unsupported combinations fail before expensive model loading.

## Validation environment

- Rebased onto `vllm-project/vllm-omni` `main` at
  `8e44e78038766b44832d6f9fd7000909517cd1cd` (2026-08-10)
- PR head: `af41085c9273f2e08d6602e5f887365d28c5ef6c`
- Python 3.12.13
- PyTorch 2.11.0+cu130
- vLLM 0.26.0
- Diffusers 0.37.0
- NVIDIA L20X, 143771 MiB, driver 570.133.20
- Local official LingBot Dense 1.3B and MoE 30B-A3B packages

The environment reports a vLLM/vLLM-Omni version mismatch and a non-blocking
bitsandbytes `libnvJitLink.so.13` warning. Neither blocked the tested paths.

### Automated regression

| Validation | Result |
| --- | --- |
| Review regression suite: fixed-order Cache-DiT, hub prefetch, HSDP, all LingBot CPU tests, and Model Extras | 271 passed, 25 warnings in 15.55 s |
| Post-rebase LingBot Cache-DiT regression | 19 passed, 17 warnings in 5.54 s |
| Existing MoE/HSDP2 online T2V+TI2V case | 1 passed, 17 warnings in 93.85 s |
| Weekly H100 marker routing | 6 single-GPU + 1 multi-GPU LingBot cases collected |
| Existing nightly marker routing | 1 unchanged single-GPU MoE case collected |
| Review-scoped and final three-file pre-commit, including the TTS adapter ratchet | Passed |
| `git diff --check` | Passed |

The nine PR commits were replayed exactly once on current `main`; the
now-merged PR #5311 dependency history remains outside this PR diff. The final
rebase added two upstream documentation commits and completed without
conflicts. The focused Cache-DiT regression and pre-commit were rerun after the
rebase. The offload and profiler GPU artifacts were retained because the replay
did not change those execution paths. The official-prompt Cache-DiT candidates
in this revision were generated after the rebase and are compared with the
matching fixed-seed baselines described below. The official-prompt Refiner
pair was also generated after the rebase.
The online HSDP2 case
generated real T2V and TI2V H.264 videos at 320x192, 24 fps, and nine frames;
the TI2V first-frame contract passed.

### Official Dense T2V with the Cache-DiT default policy

The Dense comparison uses the
[official LingBot-Video single-GPU T2V example](https://github.com/Robbyant/lingbot-video/blob/f348dca454c8f1e5707721614dbac10fa5290fed/scripts/single-gpu/run_dense_t2v.sh)
and its structured
[`example_1` prompt](https://github.com/Robbyant/lingbot-video/blob/f348dca454c8f1e5707721614dbac10fa5290fed/assets/cases/t2v/example_1/prompt.json).
Both independent engines use eager BF16 on one L20X, seed 42, sequential CFG,
guidance 3, flow shift 3, the model default negative prompt, 480x832, 121
frames at 24 fps, and 40 denoising steps. Each engine runs one warmup request
followed by two measured requests.

This validation deliberately uses the upstream Cache-DiT default DBCache
policy rather than the balanced preset in the example: `Fn=8`, `Bn=0`, eight
warmup steps, residual threshold `0.08`, no SCM mask, and no limits on total
or consecutive cached steps.

| Metric | Baseline | Cache-DiT default |
| --- | ---: | ---: |
| Measured E2E samples | 106.138 s, 106.204 s | 78.880 s, 78.491 s |
| Median E2E | 106.171 s | **78.685 s** |
| Speedup / latency reduction | - | **1.349x / 25.9%** |
| Peak reserved | 25780 MiB | 26864 MiB |
| Mean / minimum frame SSIM | - | 0.9200 / 0.8937 |
| Global PSNR / MAE | - | 29.42 dB / 4.070 |

The two sequential CFG passes cache 16 of 40 steps each, or 32 of 80
transformer forwards in total. Dense has 24 blocks; each cache hit computes
the first eight and reuses the remaining 16, corresponding to a theoretical
26.7% reduction in block computation. This is consistent with the measured
25.9% E2E latency reduction.

Both outputs are real 832x480, 121-frame H.264 MP4 files. The subject and
overall composition remain intact, but facial details, clothing edges, pose,
and motion diverge visibly from the baseline. Mean SSIM falls to 0.9200 and
the last 30 frames average 0.9123, so the upstream default is a speed-first
setting for this workload rather than a near-lossless recommendation.

dense_official_cache_dit_upstream_default_comparison

### Official MoE T2V with the Cache-DiT default policy

The MoE comparison uses the same official `example_1` prompt, default negative
prompt, 480x832 resolution, 121 frames, 24 fps, 40 steps, seed 42,
sequential CFG, guidance 3, and flow shift 3. Cache-DiT and HSDP are currently
an explicitly rejected combination, so this paired validation uses one L20X,
model-level CPU offload, and the pipeline profiler. Each independent engine
runs one measured request without an additional warmup; these numbers are
fixed-prompt latency evidence rather than a statistical throughput claim.

The Cache-DiT engine uses the same upstream default DBCache policy as the
Dense run: `Fn=8`, `Bn=0`, eight warmup steps, residual threshold `0.08`, no
SCM mask, and no total or consecutive cache limits.

| Metric | Baseline | Cache-DiT default |
| --- | ---: | ---: |
| Request E2E | 310.253 s | **203.448 s** |
| Speedup / latency reduction | - | **1.525x / 34.4%** |
| Denoising | 276.187 s | **166.795 s** |
| Denoising speedup | - | **1.656x** |
| Condition preparation | 28.551 s | 31.216 s |
| VAE decode | 4.186 s | 4.162 s |
| Peak reserved | 81488 MiB | 81488 MiB |
| Mean / minimum frame SSIM | - | 0.8986 / 0.8596 |
| Global PSNR / MAE | - | 27.06 dB / 4.729 |

Both sequential CFG passes cache 19 of 40 steps, or 38 of 80 transformer
forwards. MoE has 48 blocks; each hit computes eight blocks and reuses 40,
for a theoretical 39.6% reduction in block computation. The measured
denoising time falls by the same 39.6%, which explains the larger speedup than
the Dense run.

The generated video remains semantically correct, but the face, hair contour,
arm position, and motion trajectory increasingly diverge from the baseline.
The final 30 frames average 0.8764 SSIM. The default policy therefore caches
too aggressively for MoE quality-sensitive use despite its substantial
speedup.

moe_official_cache_dit_upstream_default_comparison

### CPU offload comparison

The paired Dense T2V run used the same 192x320, nine-frame, 40-step workload
with Cache-DiT and internal profiling disabled.

| Metric | Baseline | Model CPU offload | Change |
| --- | ---: | ---: | ---: |
| Median E2E | 2.293 s | 2.728 s | +19.0% |
| Steady-state peak reserved | 20256 MiB | 9224 MiB | -54.5% |
| Initialization | 31.17 s | 38.39 s | +7.22 s |

The generated MP4 SHA-256 is identical on both paths.

### Profiler comparison

Profiler-off and profiler-on Dense T2V outputs have the same SHA-256. Median
E2E was 2.293 s without internal profiling and 2.275 s with it (a -0.8%
difference, within run-to-run noise). The profiler-on median breakdown was:

| Stage | Median |
| --- | ---: |
| Base condition preparation | 55.25 ms |
| Base diffuse | 2.145 s |
| VAE decode | 55.37 ms |

### Official-prompt HSDP2 Base-to-Refiner comparison

This rerun uses the official
[`example_1` structured prompt](https://github.com/Robbyant/lingbot-video/blob/f348dca454c8f1e5707721614dbac10fa5290fed/assets/cases/t2v/example_1/prompt.json)
and default negative prompt on two L20X GPUs. Both independent engines use
eager BF16, sequential CFG, HSDP shard size 2 with one replica, and the
pipeline profiler. CPU offload, Cache-DiT, SP, and VAE tiling are disabled.
The Base workload remains 480x832, 121 frames at 24 fps, 40 steps, seed 42,
and guidance/shift 3. The Refiner keeps the official eight-step schedule,
`t_thresh=0.85`, `sigma_tail=2`, and guidance/shift 3. Logs confirm that both
48-block outermost DiTs plus their roots are HSDP-sharded.

The exact official 1088x1920 Refiner output was attempted first. Its Base
condition (0.10 s), Base denoise (290.89 s), Base decode (4.18 s), and
Refiner handoff (10.39 s) completed, but the first Refiner denoise pass failed
in the MoE FFN while requesting another 15.48 GiB. Rank 0 had 15.03 GiB free
and rank 1 only 67.94 MiB. HSDP shards weights, not the 1080p activations. The
[official multi-GPU recipe](https://github.com/Robbyant/lingbot-video/blob/f348dca454c8f1e5707721614dbac10fa5290fed/docs/en/dit_inference.md)
also uses context parallelism, releases Base before Refiner, and enables
Refiner VAE tiling; this isolated HSDP2 run intentionally does not. The
successful paired run therefore changes only the Refiner output to 720x1280.

| Metric | HSDP2 Base only | HSDP2 Base -> Refiner (720p) |
| --- | ---: | ---: |
| Engine initialization | 53.68 s | 122.89 s |
| Model memory after load / rank | 37.52 GiB | 65.62 GiB |
| Request E2E | 296.28 s | 506.48 s |
| Peak reserved / rank | 66920 MiB | 124380 MiB |
| Base condition | 0.099 s | 0.092 s |
| Base diffuse | 290.78 s | 290.90 s |
| Refiner handoff | - | 4.47 s |
| Refiner diffuse | - | 194.24 s |
| Decode total | 4.21 s | 13.99 s |

The Refiner adds 210.20 s (+71.0%) and 57460 MiB of peak reserved memory per
rank in this one-request comparison. Against the earlier one-GPU CPU-offload
run at the same 720p Refiner resolution, HSDP2/no-offload reduces E2E from
548.25 s to 506.48 s (-7.6%) and Refiner denoise from 221.40 s to 194.24 s
(-12.3%). HSDP all-gather raises Base denoise from 276.37 s to 290.90 s
(+5.3%), while keeping the text encoder resident reduces condition preparation
from 28.93 s to 0.092 s. Initialization rises from 60.66 s to 122.89 s.

Both HSDP2 outputs are byte-identical to their earlier CPU-offload
counterparts at the same resolutions. They are real 121-frame H.264 videos:
Base-only is 832x480 and Base-to-Refiner is 1280x720, both 5.04 seconds at
24 fps. This is fixed-prompt functional and stage-cost evidence, not a
statistically meaningful throughput or universal subjective-quality claim.

moe_t2v_official_base_vs_refiner_720p_hsdp2

Artifacts:

- Base-only MP4: `gpu/official_refiner_hsdp2/outputs/moe_t2v_official_base_only_hsdp2.mp4`
- Base-to-Refiner MP4: `gpu/official_refiner_hsdp2/outputs/moe_t2v_official_with_refiner_720p_hsdp2.mp4`
- Performance JSON: `gpu/official_refiner_hsdp2/metrics/`
- Contact sheet: `gpu/official_refiner_hsdp2/comparisons/moe_t2v_official_base_vs_refiner_720p_hsdp2.png`

### HSDP capacity evidence

The original PR validation on H200 used the same MoE 320x192, nine-frame,
two-step workload:

| Metric | Single GPU | HSDP2 | HSDP4 |
| --- | ---: | ---: | ---: |
| Process GPU memory after load / rank | 65.62 GiB | 37.52 GiB | 23.45 GiB |
| Worker peak reserved / rank | 73.03 GiB | 49.74 GiB | 35.67 GiB |
| Peak-memory reduction | - | 31.9% | 51.2% |
| Total generation | 600.67 ms | 1215.76 ms | 1303.64 ms |

HSDP is a capacity feature here. The small workload is slower because every
transformer block requires an all-gather; these numbers do not claim reduced
system host memory. The current L20X regression additionally reran the
existing HSDP2 online T2V/TI2V case successfully.

## Not covered

- Cache-DiT combined with PP, TP, EP, SP, CFG parallelism, VAE patch
  parallelism, HSDP, or distributed layerwise offload; these combinations are
  rejected before model loading.
- Multi-prompt perceptual evaluation or human blind testing.
- Statistically meaningful Refiner throughput or subjective-quality claims;
  the official-prompt paired result is a functional and stage-cost comparison.
- Multi-request and multi-sample batching.
- New nightly coverage; intentionally not added. Added GPU functional coverage
  is routed through the existing weekly H100 sweeps.


vllm_omni/diffusion/distributed/hsdp.py
```diff
@@ -35,6 +35,22 @@ def _unshardable_parameters(model: nn.Module) -> set[nn.Parameter]:
     }
 
 
+def _resolve_hsdp_param_dtype(
```
wtz2333, line 38: Why add this function?

hsliuustc0106:
@wtz2333 review the agent codes please

vllm_omni/diffusion/models/lingbot_video/pipeline_lingbot_video.py
```diff
@@ -853,22 +1599,44 @@ def forward(self, req: DiffusionRequestBatch) -> DiffusionOutput:
...
+            sampling.width = execution_options.refiner.width
+            sampling.num_frames = sample_frames
+            sampling.fps = execution_options.refiner.output_fps
+            sampling.frame_rate = float(execution_options.refiner.output_fps)
+            output = {
+                "payload": {"video": frames},
+                "metadata": {
+                    "video": {
+                        "fps": execution_options.refiner.output_fps,
+                        "refined": True,
```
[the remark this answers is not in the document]
wtz2333: fixed it

vllm_omni/diffusion/distributed/hsdp.py
wtz2333: This helper isolates the `param_dtype` decision because the previous inline logic only handled FP8. LingBot intentionally has mixed parameter dtypes: the bulk weights are BF16, while AdaLN, normalization, and MoE router parameters remain FP32. Passing `param_dtype=torch.bfloat16` to FSDP would make the unsharded parameters used by forward uniformly BF16, defeating those FP32 islands. LingBot therefore opts in via `_hsdp_preserve_param_dtype`, and returning `None` tells FSDP to preserve each parameter's loaded dtype. Existing non-FP8 models still return `configured_dtype`, and the existing FP8 behavior is unchanged. Keeping this in a helper also lets us test those three cases directly.

tests/e2e/online_serving/test_lingbot_video_expansion.py
```diff
@@ -42,6 +42,25 @@ def _get_diffusion_feature_cases(model: str):
...
+                    "--enforce-eager",
+                ],
+            ),
+            id="layerwise-offload",
+            marks=SINGLE_CARD_FEATURE_MARKS,
+        ),
+    ]
+
+
+@pytest.mark.full_model
```
yenuo26, line 64: maybe slow mark

tests/e2e/online_serving/test_lingbot_video_moe.py
```diff
@@ -97,3 +177,86 @@ def test_video_generation_modes_moe(omni_server: OmniServer, openai_client: Open
...
+        responses[0].videos[0],
+        synthetic_image["np_array"],
+        max_mean_absolute_error=0.25,
+    )
+
+
+@pytest.mark.slow
+@pytest.mark.diffusion
+@pytest.mark.parametrize("omni_server", _get_cache_dit_refiner_cases(MODEL), indirect=True)
+def test_cpu_offload_cache_dit_refiner_t2v_moe(omni_server: OmniServer, openai_client: OpenAIClientHandler) -> None:
```
yenuo26: After cleanup, a round of naming rectification was carried out. Can you modify openai_client to online_client?

vllm_omni/diffusion/distributed/hsdp.py
```diff
@@ -35,6 +35,22 @@ def _unshardable_parameters(model: nn.Module) -> set[nn.Parameter]:
...
+) -> torch.dtype | None:
+    """Resolve whether FSDP should cast parameters during all-gather.
+
+    Most diffusion transformers use one storage dtype and benefit from the
+    configured mixed-precision policy. Models with intentional mixed-dtype
+    parameters can opt out so FSDP preserves each parameter's loaded dtype.
+    FP8 models require the same behavior for their GEMM kernels.
+    """
+    preserve_param_dtype = bool(getattr(model, "_hsdp_preserve_param_dtype", False))
+    has_fp8_params = any(p.dtype in (torch.float8_e4m3fn, torch.float8_e5m2) for p in model.parameters())
```
RuixiangMa, line 47: any() iterates all parameters unnecessarily when preserve_param_dtype is already True.

vllm_omni/diffusion/models/lingbot_video/request_utils.py
```diff
@@ -98,6 +102,205 @@ class LingBotRequestConfig:
...
+    )
+    refiner_output_fps = _positive_int_option(
+        extra_args.get("refiner_output_fps", 24),
+        "refiner_output_fps",
+    )
+    raw_max_frames = extra_args.get("refiner_max_video_frames")
+    refiner_max_frames = (
+        None if raw_max_frames is None else _positive_int_option(raw_max_frames, "refiner_max_video_frames")
+    )
+    if refiner_max_frames is not None and refiner_max_frames != 1 and (refiner_max_frames - 1) % 4 != 0:
```
RuixiangMa: hardcodes % 4 while refiner_utils.py:163 parameterizes it as vae_temporal_factor. If the VAE temporal factor ever changes, request validation and execution validation will silently diverge — users pass request validation but fail at execution time with a confusing error. Pass the VAE temporal factor into request validation or assert they match.

wtz2333: I removed the hardcoded temporal factor from request validation by adding a vae_temporal_factor parameter to normalize_lingbot_execution_options. The pipeline now passes its actual self.vae_scale_factor_temporal, so request‑time and execution‑time validation use the same value.

vllm_omni/diffusion/distributed/hsdp.py
wtz2333: I changed `_resolve_hsdp_param_dtype` to return immediately when `_hsdp_preserve_param_dtype` is enabled, so it no longer calls model.parameters() on that path.

tests/e2e/online_serving/test_lingbot_video_expansion.py
wtz2333: Added pytest.mark.slow to this full‑model test so it is collected by the existing weekly H100 slow‑test sweep.

tests/e2e/online_serving/test_lingbot_video_moe.py
wtz2333: Updated this test to use the new online_client fixture and renamed the corresponding request calls to match the helper cleanup and current test naming convention.

vllm_omni/diffusion/models/lingbot_video/refiner_utils.py
```diff
@@ -0,0 +1,262 @@
...
+        raise ValueError(f"Refiner sample_fps must be positive, got {sample_fps}.")
+    if vae_temporal_factor <= 0:
+        raise ValueError(f"Refiner VAE temporal factor must be positive, got {vae_temporal_factor}.")
+    if max_frames is not None and (
+        max_frames <= 0 or (max_frames != 1 and (max_frames - 1) % vae_temporal_factor != 0)
+    ):
+        raise ValueError(f"Refiner max_frames must be 1 or aligned to the VAE temporal factor, got {max_frames}.")
+
+    if source_fps > sample_fps:
+        raw = int(num_source_frames / source_fps * sample_fps)
```
wtomin, line 180: I am a little confused. What are the difference between `source_fps` and `sample_fps`?

vllm_omni/diffusion/models/lingbot_video/pipeline_lingbot_video.py
```diff
@@ -335,9 +497,19 @@ def __init__(self, *, od_config: OmniDiffusionConfig, prefix: str = ""):
         self._crop_start: int | None = None
 
         model = od_config.model
-        local_files_only = os.path.exists(model)
+        revision = od_config.revision
+        local_files_only = os.path.isdir(model)
+        load_device = torch.get_default_device()
```
wtomin, line 502: Is it likely that ` torch.get_default_device()` and `get_local_device()` give different devices, so that `load_device` is not the same as `self.devices`?

PR #4474, opened 2026-06-16
[BugFix][Qwen-Image] align txt_seq_lens RoPE width with padded embeds

SamitHuang:
## Summary

Fixes #4443.

Qwen-Image pipelines previously derived `txt_seq_lens` from `prompt_embeds_mask.sum(dim=1)`, which counts valid tokens only. Diffusers (and the RL trainer) derive text RoPE length from the **padded encoder hidden-state width** (`prompt_embeds.shape[1]`). When prompts are right-padded for request batching or fixed-width collation, vLLM-Omni built a too-short RoPE table and attention numerics diverged from the reference.

This PR:
- Adds `txt_seq_lens_from_embeds()` and uses it in all Qwen-Image pipeline variants (`pipeline_qwen_image*.py`).
- Refreshes per-request `txt_seq_lens` / `negative_txt_seq_lens` on the continuous-batching path after in-place prompt padding in `input_batch.py`.

## Why this is not duplicating an existing PR

Checked open PRs for `txt_seq_lens` / RoPE / #4443 — no overlapping fix found.

### 1. Previous failure case (inputs)

Representative **request-batch / CB padding** scenario (two requests fused to a shared width):

| Field | Request A | Request B |
|---|---|---|
| Valid tokens (`mask.sum()`) | 10 | 20 |
| Padded width (`prompt_embeds.shape[1]`) | 32 | 32 |

**Pre-PR metadata (buggy):**
```python
txt_seq_lens = prompt_embeds_mask.sum(dim=1).tolist()  # [10, 20]
max(txt_seq_lens) = 20  <  shape[1] = 32  # RoPE table too short
```

**Post-PR metadata (fixed):**
```python
txt_seq_lens = txt_seq_lens_from_embeds(prompt_embeds)  # [32, 32]
max(txt_seq_lens) = 32  ==  shape[1] = 32
```

**RoPE table length (`QwenEmbedRope`, CPU):**

| `txt_seq_lens` | Text freq rows | Matches padded width? |
|---|---|---|
| `[10, 20]` (pre-PR) | **20** | no (20 < 32) |
| `[32, 32]` (post-PR) | **32** | yes |

Single-request CB padding is also covered by `test_input_batch_txt_seq_lens.py` (valid=10, pad to 32).

### 2. Automated tests (CI, CPU)

```bash
python3 -m pytest \
  tests/diffusion/models/qwen_image/test_qwen_image_rope_utils.py \
[...]

Issue #4443, opened 2026-06-15
[Bug] QwenImage RoPE `txt_seq_lens` uses `mask.sum()` instead of padded embed width (misaligned with diffusers)


zhtmike:
## Summary

vLLM-Omni QwenImage pipelines set `txt_seq_lens` from `prompt_embeds_mask.sum(dim=1)` when preparing generation context / `prepare_encode()`. Diffusers (latest `main`) instead derives RoPE text length from the **padded encoder hidden-states width** (`encoder_hidden_states.shape[1]`) inside `QwenImageTransformer2DModel.forward()` via `compute_text_seq_len_from_mask()`.

When prompt embeddings are wider than the number of valid (non-padding) tokens — common under **continuous batching (CB)** after in-place padding, RL training with fixed-width collation, or any caller that supplies pre-padded embeds — vLLM-Omni builds a **too-short** text RoPE table. That changes attention numerics relative to diffusers/PyTorch reference and can inflate rollout–trainer KL in RL setups.

Verified on **vllm-omni `main` @ `a693ae67`** against **diffusers `main` @ `d1f8e55c3`**.

## Affected code (vllm-omni)

All QwenImage pipeline variants set `txt_seq_lens` from `mask.sum()` in their generation-context helpers:

- `vllm_omni/diffusion/models/qwen_image/pipeline_qwen_image.py` (`_prepare_generation_context`, used by `forward()`, `diffuse()`, and `prepare_encode()`)
- `vllm_omni/diffusion/models/qwen_image/pipeline_qwen_image_edit.py`
- `vllm_omni/diffusion/models/qwen_image/pipeline_qwen_image_edit_plus.py`
- `vllm_omni/diffusion/models/qwen_image/pipeline_qwen_image_layered.py`

```python
# Current (vllm-omni):
txt_seq_lens = prompt_embeds_mask.sum(dim=1).tolist() if prompt_embeds_mask is not None else None
negative_txt_seq_lens = (
    negative_prompt_embeds_mask.sum(dim=1).tolist() if negative_prompt_embeds_mask is not None else None
)
```

`prepare_encode()` copies these values onto `DiffusionRequestState.txt_seq_lens` / `negative_txt_seq_lens`. The stepwise CB path (`InputBatch`) reuses them but **does not refresh them** after `_prepare_request_prompt_field()` pads `prompt_embeds` / masks on the request state in place.

## Reference behavior (diffusers latest main)

[...]

PR #3588, opened 2026-05-14
[Quant][AutoRound] Add W4A16 support for Qwen-Image

yiliu30:
## Summary

Add W4A16 AutoRound quantization support for the Qwen-Image transformer:

- Fix weight prefix propagation in `QwenImageTransformerBlock`, `QwenImageCrossAttention`, and `FeedForward` so quantized packed weights resolve correctly.
- Add `_resolve_txt_mod_quant_config()` to selectively propagate quant_config for INC-based configs only (preserving FP8 safeguards per #2728).
- Improve weight loading with `_resolve_qwen_image_lookup_name()` and `_normalize_qwen_image_weight_name()` helpers.
- Add e2e load test for the W4A16 quantized Qwen-Image model.

###

- Performance (2× RTX 5090 D, TP=2, CPU offload)

| Config | FP16 | W4A16 | Speedup | FP16 Mem | W4A16 Mem | Mem Δ |
|---|---|---|---|---|---|---|
| 512×512, 20 steps | 8.58s | 6.80s | **+26%** | 27.4 GB | 15.0 GB | **-45%** |
| 512×512, 50 steps | 15.71s | 13.17s | **+19%** | 27.4 GB | 15.0 GB | **-45%** |
| 1024×1024, 25 steps | 24.41s | 22.64s | **+8%** | 30.8 GB | 16.4 GB | **-47%** |
| 1024×1024, 50 steps | 44.54s | 42.26s | **+5%** | 30.8 GB | 16.4 GB | **-47%** |

Latency improves 5-26% (largest gain at low resolution); memory drops 45-47% across all configs. FP16 cannot run on a single 5090 at any resolution; W4A16 fits at 512×512 on one GPU.

- Quality (TIIF-Bench × CLIP Score, 512×512, 28 steps)

| Dimension | N | FP16 CLIP | W4A16 CLIP | Δ | Cross-CLIP |
|---|---|---|---|---|---|
| text rendering | 30 | 33.05 | 32.26 | -0.80 | 85.39 |
| color + spatial | 13 | 33.86 | 32.81 | -1.05 | 95.64 |
| **Overall** | **43** | **33.46** | **32.53** | **-0.92** | **90.52** |

- A CLIP score delta of -0.92 (on 0-100 scale) is negligible — W4A16 preserves prompt-image alignment comparable to FP16. Cross-CLIP of 90.52 confirms both models produce semantically similar outputs for the same prompt.

## Test
[...]

PR #5586, opened 2026-07-30
[BugFix][Qwen-Image-Edit] restore txt_seq_lens RoPE width after #3588 regression

NumberWan:
## Summary
#3588 accidentally reverted the #4474 RoPE length fix in pipeline_qwen_image_edit.py back to mask.sum(). Restore txt_seq_lens_from_embeds and add a source guard covering all Qwen-Image pipeline variants to prevent another silent regression.

## Purpose
- Restores [#4474](https://github.com/vllm-project/vllm-omni/pull/4474) behavior in `pipeline_qwen_image_edit.py`: derive `txt_seq_lens` / `negative_txt_seq_lens` from padded embed width via `txt_seq_lens_from_embeds()`, not `prompt_embeds_mask.sum()`.
- [#3588](https://github.com/vllm-project/vllm-omni/pull/3588) (AutoRound W4A16, merged **after** #4474) accidentally reintroduced `mask.sum()` in the Edit pipeline only; `pipeline_qwen_image.py` / Edit-Plus / Layered were unaffected.
- Tests:
  - `test_qwen_image_rope_utils.py` — helper + RoPE table length use padded width (includes the assert formerly in `test_qwen_image_rope_table_failure_case.py`)
  - `test_input_batch_txt_seq_lens.py` — CB padding refreshes `state.txt_seq_lens`
  - `test_qwen_image_pipelines_txt_seq_lens.py` — ban `txt_seq_lens = ...mask.sum(` in `pipeline_qwen_image*.py`

## Test Plan and Result
- [x] `pytest tests/diffusion/models/qwen_image/test_qwen_image_pipelines_txt_seq_lens.py tests/diffusion/models/qwen_image/test_qwen_image_rope_utils.py tests/diffusion/test_input_batch_txt_seq_lens.py -q` (`10 passed`)

## Related
- Fixes regression of #4443 / #4474 on the Qwen-Image-Edit path (introduced by #3588)
- 


NumberWan:
@fhfuih

fhfuih:
Thanks. The production code change looks good to me

- Any reason you deleted the `tests/diffusion/models/qwen_image/test_qwen_image_rope_table_failure_case.py`? Didn't check the detail, but it seems to be introduced in #4474 , not #3588 
- The 1st added test case (`tests/diffusion/models/qwen_image/test_qwen_image_pipelines_txt_seq_lens.py::test_qwen_image_pipeline_uses_txt_seq_lens_from_embeds`) is very harsh. It does regression check though, but by checking if the pipeline import the specific function from the specific python file. __I am not saying such a test is 100% ugly---this may be the best way to do it if we need such a test. But I'd still like to draw attention to others and question whether such a harsh regression test is needed in our codebase.__
- Does `test_txt_seq_lens_from_embeds_builds_rope_table_for_padded_width` check the similar thing? (By checking the produced `txt_seq_lens` matched the expected/improved pattern). If so, keeping this test is enough to guard similar regression to #3588 , and the above test case can be removed.
  - e.g., if this function is moved or refactored in the future.
- The 2nd added test case `test_expected_qwen_image_pipeline_set_is_complete` is useless. Consider removing it.
- If to not remove the first test case, you can keep a note in the file-level docstring, that if there is future change to this function/Qwen pipelines, feel free to modify or remove this test (to reassure future developers)

NumberWan:
> Thanks. The production code change looks good to me
> 
> * Any reason you deleted the `tests/diffusion/models/qwen_image/test_qwen_image_rope_table_failure_case.py`? Didn't check the detail, but it seems to be introduced in [[BugFix][Qwen-Image] align txt_seq_lens RoPE width with padded embeds #4474](https://github.com/vllm-project/vllm-omni/pull/4474) , not [[Quant][AutoRound] Add W4A16 support for Qwen-Image #3588](https://github.com/vllm-project/vllm-omni/pull/3588)
> * The 1st added test case (`tests/diffusion/models/qwen_image/test_qwen_image_pipelines_txt_seq_lens.py::test_qwen_image_pipeline_uses_txt_seq_lens_from_embeds`) is very harsh. It does regression check though, but by checking if the pipeline import the specific function from the specific python file. **I am not saying such a test is 100% ugly---this may be the best way to do it if we need such a test. But I'd still like to draw attention to others and question whether such a harsh regression test is needed in our codebase.**
> * Does `test_txt_seq_lens_from_embeds_builds_rope_table_for_padded_width` check the similar thing? (By checking the produced `txt_seq_lens` matched the expected/improved pattern). If so, keeping this test is enough to guard similar regression to [[Quant][AutoRound] Add W4A16 support for Qwen-Image #3588](https://github.com/vllm-project/vllm-omni/pull/3588) , and the above test case can be removed.
>   
>   * e.g., if this function is moved or refactored in the future.
> * The 2nd added test case `test_expected_qwen_image_pipeline_set_is_complete` is useless. Consider removing it.
> * If to not remove the first test case, you can keep a note in the file-level docstring, that if there is future change to this function/Qwen pipelines, feel free to modify or remove this test (to reassure future developers)


Thanks for the review.
**1.) test_qwen_image_rope_table_failure_case.py**
Removed after folding the useful RoPE table-width assert into test_qwen_image_rope_utils.py (builds_rope_table_for_padded_width). Coverage kept, standalone file dropped.
**2) Harsh AST / import check**
Agreed it was too brittle. Replaced with a lighter call-site guard that only bans (negative_)?txt_seq_lens = ...mask.sum( in pipeline_qwen_image*.py. No required helper import; docstring notes it can be changed/removed if wiring changes.
**3)Is builds_rope_table_for_padded_width enough alone?**
Not for this regression: it exercises the helper, but #3588 broke the Edit pipeline call site while helper tests still passed. Keeping both the helper/RoPE test and the mask.sum ban.
**4) About test_expected_qwen_image_pipeline_set_is_complete**
I agreed with your opinion, the file has been removed.

NumberWan:
@Gaohan123 PTAL

tests/diffusion/models/qwen_image/test_qwen_image_pipelines_txt_seq_lens.py
```diff
@@ -0,0 +1,36 @@
...
+    "pipeline_path",
+    sorted(QWEN_IMAGE_DIR.glob("pipeline_qwen_image*.py")),
+    ids=lambda path: path.name,
+)
+def test_qwen_image_pipeline_does_not_set_txt_seq_lens_from_mask_sum(pipeline_path: Path):
+    source = pipeline_path.read_text(encoding="utf-8")
+    match = _MASK_SUM_TXT_SEQ_LENS_RE.search(source)
+    assert match is None, (
+        f"{pipeline_path.name} must not derive txt_seq_lens from mask.sum(); found: {match.group(0)!r}"
+    )
```
fhfuih: This regression test seems quite hardcode-ish. You are basically testing the source code literally. And static linting/checking tools cannot capture the usage of `txt_seq_lens_from_embeds()` (or the non-usage of `.sum`, if you'd like to write it that way). So if relevant variables or functions is to be renamed in the future, the author will not realize this by "finding references" or even "global search (by text)" in IDE---only after a `pytest` fail.

I think a better way is to mock/patch the `txt_seq_lens_from_embeds()` function, import and run the outer function/call-site, then assert this inner function is called twice (once for each prompt).

But I admit this method I mention may be cumbersome to write nowadays, because `txt_seq_lens_from_embeds()` is called in `forward()`, with may other expensive function calls above it and many different inputs & heavy properties. I can only think of mocking every heavy stuff above `txt_seq_lens_from_embeds()` calls, and mock `txt_seq_lens_from_embeds()` themselves to raise (to stop `forward()` early and everything below them is not called either), then finally assert the number of times it is called. So there will still be many hardcoded assumption about the function routine.

So I cannot give a very confident suggestion for now 😂. Feel free to consult AI if there is a better way to write a test, or to insist that your current version is better.

fhfuih: Asked AI, and it suggested a better test design like follows:

1. `pipeline = object.__new__(QwenImageEditPipeline)` + minimal attrs (This skips the creation of heavy attr initialization)
2. Fake `encode_prompt` to return padded embeds + a shorter mask (this is the primary heavy function call inside `forward()`, skip it is enough)
3. Capture args in fake `diffuse` (this is called right after the critical lines that computes `txt_seq_lens`)
4. Assert `txt_seq_lens == [padded_width, …]` (and neg path if exercised)

NumberWan: Thanks, that makes sense. The source regex was too brittle — it tested spelling, not the padded-width invariant, and IDE rename/find-refs wouldn't catch it.

I replaced it with a behavioral check on the Edit forward() call site (the path #3588 actually broke):

skip heavy __init__ with object.__new__
stub encode_prompt to return padded embeds + a shorter mask
stub prepare_latents / prepare_timesteps so we don't run the expensive path
capture args in a fake diffuse (right after txt_seq_lens is computed)
assert txt_seq_lens == [padded_width] (and the negative CFG path when exercised)
Helper tests in test_qwen_image_rope_utils.py stay as they are — they still wouldn't have caught #3588 alone.

congw729 (approved):
LGTM. I will trigger the tests and merge this one after the CI passes

PR #5188, opened 2026-07-18
[Model] Optimize MiniCPM-o multimodal encoder memory

xsmccc:
## Purpose

Reduce MiniCPM-o 4.5 multimodal encoder peak memory without changing SigLIP,
FlashAttention, or encoder-cache internals. This addresses the A.3 profiling
scope in #5069.

The active vision path previously sent every image slice or decoded video frame
to SigLIP in one batch, even though the model configuration already defines
`vision_batch_size=16`. The audio path also requested every Whisper hidden
state when the deployed configuration only consumes the final layer.

This PR:

- bounds active-path VPM calls by `config.vision_batch_size`, concatenates their
  outputs, and invokes the resampler once over the full batch;
- requests only Whisper `last_hidden_state` for `audio_encoder_layer=-1`, while
  preserving all hidden states for non-final layer selection;
- adds focused CPU tests for the vision and audio control-flow changes.

The production change does not modify SigLIP/FlashAttention kernels and does
not overlap the MiniCPM-o encoder-cache files changed by #4773.

### RTX PRO 6000 Blackwell results

Environment: one 97,887 MiB RTX PRO 6000 Blackwell, driver 590.44.01,
CUDA 13.1, Python 3.12.3, PyTorch 2.11.0+cu130, vLLM 0.25.0,
Transformers 5.14.1, and Triton 3.6.0. Before/after runs use the same BF16
checkpoint, deploy config, input preprocessing, and `VLLM_USE_FLASHINFER_SAMPLER=0`.
Direct vision/audio timings use one warmup plus 10/20 measured runs. Server
latencies use one warmup plus five independent cache-miss media variants.
The measurements compare baseline `d09f549e` with the candidate diff; the
MiniCPM-o encoder source at the final rebased base (`4fbbdfd2`) is byte-identical
to that baseline.

| Input / metric | Before | After | Change |
| --- | ---: | ---: | ---: |
| 32-frame video server request peak | 6,262 MiB | 2,606 MiB | -58.4% |
| Video standalone peak allocated | 5,958,894,592 B | 3,032,646,656 B | -49.1% |
| GPU KV-cache capacity | 210,864 tokens | 264,672 tokens | +25.5% |
| Video encoder latency, 10-run mean | 458.98 ms | 464.39 ms | +1.18% |
| Video TTFT, 5-run mean | 1,328.18 ms | 1,289.92 ms | -2.88% |
| Whisper standalone peak allocated | 56,558,592 B | 23,391,232 B | -58.6% |
[...]

Issue #5069, opened 2026-07-13
[RFC]: MiniCPM-o 4.5 Performance Acceleration


amy-why-3459:
## MiniCPM-o 4.5 progress ↔ Issue/PR mapping (updated 2026-09-11)

Legend: ✅ merged/done · ⏳ open/in-progress PR · ⚠️ partial/limited · ❌ TODO (no issue/PR yet) · 🚫 closed unmerged

Tracking issue for **MiniCPM-o 4.5 performance acceleration** after the original RFC (2026-07-13). Most of the original plan has landed on `main`: **3-stage** Thinker → Talker → Code2Wav, default `async_chunk`, CUDA graphs on stages 0/1, **inner HiFT + CFM DiT graphs on stage 2**, Talker batched codec sampling, `max_num_seqs: 4`, per-request isolation, MiniCPM-o accuracy/perf CI on the ready gate, and **NPUGraph replay for Code2Wav**. Duplex serving graduation has now merged; shipping YAMLs enable the duplex session mode. Remaining work includes Talker AR migration, cache parity/regression coverage, concurrency scale-out, duplex hardening, and published baselines.

Format follows [#5700](https://github.com/vllm-project/vllm-omni/issues/5700) (roadmap Status / Ref / Note tables).

### Delta since 2026-08-23

Status checked against GitHub on 2026-09-11. PR status is not a fresh hardware qualification or performance measurement; benchmark figures below retain their original workload/platform scope.

| Status | Ref | Note |
| --- | --- | --- |
| ✅ | [#5464](https://github.com/vllm-project/vllm-omni/pull/5464), [#5730](https://github.com/vllm-project/vllm-omni/pull/5730) | Offline/online tests and audio_embeds fixes merged. |
| ✅ | [#6021](https://github.com/vllm-project/vllm-omni/pull/6021), [#6170](https://github.com/vllm-project/vllm-omni/pull/6170), [#6318](https://github.com/vllm-project/vllm-omni/pull/6318) | Configurable duplex sessions, handoff/barge-in isolation, and locked-session instruction validation merged. |
| ✅ | [#6196](https://github.com/vllm-project/vllm-omni/pull/6196), [#6619](https://github.com/vllm-project/vllm-omni/pull/6619) | Duplex serving graduation and shipping-YAML enablement merged; this does not close all productization work. |
[...]

PR #5587, opened 2026-07-30
[Model] Optimize MiniCPM-o streaming audio encoder memory

xsmccc:
Related to #5069.

This is the focused streaming-audio follow-up discussed in
https://github.com/vllm-project/vllm-omni/issues/5069#issuecomment-5119179172
and https://github.com/vllm-project/vllm-omni/pull/5188#issuecomment-5033947603.

### Problem

`get_audio_embedding_streaming()` always requested
`output_hidden_states=True` and selected `hidden_states[-1]`, even when
`audio_encoder_layer == -1` means that only Whisper's final encoder output is
needed. This retains the embedding output and every intermediate encoder-layer
state on every cached streaming chunk.

### Changes

- Use `last_hidden_state` and set `output_hidden_states=False` for the default
  final-layer selection (`audio_encoder_layer == -1`).
- Preserve the existing `hidden_states[audio_encoder_layer]` behavior for
  non-final layer selection, including Whisper's embedding-output-first hidden
  state ordering.
- Leave the streaming KV-cache, attention mask, extra-context arguments,
  projection, and pooling paths unchanged.
- Add focused CPU regression tests for:
  - final-layer selection without retaining intermediate hidden states;
  - non-final layer selection with hidden states enabled;
  - two cached chunks with cache reuse, growing attention-mask/cache lengths,
    and unchanged extra-context arguments.

The production diff is intentionally small. The one-off GPU comparison harness
and result artifacts are not included in this PR, following the review guidance
from #5188.

## Test Plan

**vLLM Version:** 0.25.0

**vLLM-Omni Commit:** `ec760d9bd8448c1c7a8b34778c62408880496076`

**Base Commit:** `1d6c0816a4fbd9079878d62bbe28c1a12f146a1f`

The GPU evidence was captured before the final rebase. Git object verification
shows that the MiniCPM-o production file is byte-identical across the old and
current bases (blob `577ce5aa`) and across the benchmarked and current
candidates (blob `d37aa0d1`).

**Environment:**

- Python 3.12.3
- PyTorch 2.11.0+cu130
- Transformers 5.14.1
- CUDA runtime reported by PyTorch: 13.0
- NVIDIA driver 590.44.01
- NVIDIA RTX PRO 6000 Blackwell Server Edition, 96 GiB
- Model: `OpenBMB/MiniCPM-o-4_5`, BF16, eager attention

Commands:

```bash
pytest -q tests/model_executor/models/minicpmo_4_5/test_audio_streaming.py
pytest -q \
  tests/model_executor/models/minicpmo_4_5 \
  tests/worker/test_native_duplex_hooks.py

pre-commit run debug-statements --from-ref origin/main --to-ref HEAD
pre-commit run end-of-file-fixer --from-ref origin/main --to-ref HEAD
pre-commit run mixed-line-ending --from-ref origin/main --to-ref HEAD
pre-commit run trailing-whitespace --from-ref origin/main --to-ref HEAD
pre-commit run ruff-check --from-ref origin/main --to-ref HEAD
pre-commit run ruff-format --from-ref origin/main --to-ref HEAD
pre-commit run typos --from-ref origin/main --to-ref HEAD
pre-commit run check-test-ci-coverage --from-ref origin/main --to-ref HEAD
pre-commit run check-pickle-imports --from-ref origin/main --to-ref HEAD

PATH=/usr/bin:$PATH bash tools/run_ready_jobs.sh --dry-run
```

The GPU comparison used a fixed 8.03-second, 16 kHz audio prefix split into
eight approximately one-second cached chunks. Each process ran one unmeasured
warm-up streaming session, one peak-memory session, and then latency
repetitions. The complete baseline/candidate pair was run twice independently.

### CPU and repository checks

- Focused streaming tests: `3 passed, 14 warnings in 0.40s`
- MiniCPM-o model-executor tests plus native duplex hooks:
  `173 passed, 1 skipped, 17 warnings in 2.68s`
- All listed pre-commit hooks passed.
- `check-test-ci-coverage` passed, and `tools/run_ready_jobs.sh --dry-run`
  places the new tests in the `Simple Model Executor Test` CPU job.

### Cached streaming numerical equivalence

Both independent baseline/candidate pairs produced:

- 327,680 output elements bit-exact (`max_abs_error = 0`,
  `mean_abs_error = 0`, cosine similarity `1.0`);
- all 48 final KV-cache tensors bit-exact;
- matching cache lengths of 50, 100, ..., 400 tokens;
- reuse of the same cache object after the first chunk.

### Peak GPU memory

| Run | Baseline peak allocated delta | PR peak allocated delta | Saving | Baseline peak reserved delta | PR peak reserved delta | Saving |
| 1 | 41.72 MiB | 40.76 MiB | 0.96 MiB | 54 MiB | 50 MiB | 4 MiB |
| 2 | 41.72 MiB | 40.76 MiB | 0.96 MiB | 56 MiB | 50 MiB | 6 MiB |

As a non-regression sanity check, mean eight-chunk latency was
63.38 ms vs. 62.63 ms over 20 repetitions in run 1, and
66.11 ms vs. 62.63 ms over 5 repetitions in run 2. These sequential local
measurements are not presented as a latency improvement claim.

### Local environment limitation

Transformers 5.14.1 has cache-interface differences from the copied Whisper
cache contract used by this checkout. The out-of-tree GPU harness therefore
applied identical compatibility adapters to baseline and candidate for
`DynamicCache.get_usable_length`, `EncoderDecoderCache` indexing, and the
Whisper attention cache call/return contract. The relative numerical and memory
comparison remains controlled; no compatibility adapter is part of this PR.

The broader local command
`pytest -q tests/model_executor -m "core_model and cpu"` is also blocked during
collection in unchanged Qwen model tests because the installed vLLM 0.25.0 does
not export `vllm.multimodal.utils.set_mm_embedding_modality`. The focused and
related MiniCPM-o suites above complete successfully; project CI should run the
full split in its pinned environment.


amy-why-3459:
Docs build is failing, please run cd docs && make html to locate and fix the error.

xsmccc:
> 文档构建失败了，请运行 CD 文档并制作 HTML 来定位并修复错误。

Thanks for flagging this. I checked the exact PR head, `8c417ddfca490b60bb608a24c59fe92bfd0b58ec`.

The Read the Docs build did not reach the MkDocs build step. Its detailed notification reports:

`Error while checking out the repository`  
`Failed to checkout revision: 8c417ddfca490b60bb608a24c59fe92bfd0b58ec`

On the current branch, documentation is configured through the root `.readthedocs.yml` and `mkdocs.yml`; there is no `docs/Makefile`, so I ran the equivalent strict build on the exact PR head:

`python -m mkdocs build --strict`

It completed successfully: `Documentation built in 205.83 seconds`.

The PR changes only the MiniCPM-o model implementation and its test file, with no documentation or documentation-configuration changes. Since the PR head revision is now available, this appears to be a transient checkout/ref-propagation failure after syncing `main`. Could someone with Read the Docs access please rerun that build?

Issue #5582, opened 2026-07-30
[RFC]: Establish an operator boundary for vLLM-Omni diffusion


hsliuustc0106:
> **Status:** Draft for discussion
>
> **vLLM-Omni snapshot:** [`92534733`](https://github.com/vllm-project/vllm-omni/commit/925347332f8190d172dfa8ac0c307a5cfa2320cb)
>
> **Upstream vLLM snapshot:** [`fa2a2589`](https://github.com/vllm-project/vllm/commit/fa2a2589bd2a1fce0851df7fd42ffb54b6195f04)

## Summary

vLLM has an operator boundary, most visibly through `_custom_ops.py` and its
kernel-provider interfaces. Model code can use optimized kernels without
depending on native extension names, packed layouts, allocation details, or
platform-specific dispatch.

vLLM-Omni already reuses much of that machinery. However, diffusion support has
also accumulated local Triton kernels, `torch.library` custom ops, platform
calls, third-party providers, and compatibility patches. These cover semantics
that are sometimes different from autoregressive inference:

- N-dimensional and non-contiguous activations;
- routed mixture-of-transformers (MoT) computation;
- diffusion-specific attention and quantized attention;
- platform-specific microscaling formats; and
- stateful diffusion cache operations.

This RFC asks whether vLLM-Omni should define an explicit diffusion operator
boundary. “Independent” means independently defined semantics, not independent
reimplementation: generic kernels should continue to come from upstream vLLM.

Approval would only establish ownership principles. It would not approve a
refactor, public API, file layout, or migration plan.

### Diffusion needs a stronger tensor contract

Diffusion linears commonly receive tensors shaped `(..., K)` and must return
`(..., N)`, preserving all leading dimensions. Inputs may also be
non-contiguous.

The current code adapts upstream FP8 paths by:

- making ScaledMM activations contiguous before a two-dimensional `view`; and
- restoring leading dimensions when the FlashInfer provider returns a
  two-dimensional result.

Several quantization methods independently perform similar flatten/restore
logic. These are provider-contract requirements, not new GEMM algorithms.

### Some operations are genuinely diffusion-specific

Examples include:

- MoT kernels that route text and VAE tokens to different weights inside one
  fused GEMM;
- diffusion attention with its own layouts, masks, parallel strategies, and
  optional quantization;
- NPU FP8 attention that rotates and block-quantizes Q/K/V on each denoising
  step; and
- experimental paged attention that writes into a process-owned diffusion KV
[...]

PR #5528, opened 2026-07-29
[ROCm] [CI] Fix ROCm CI for v0.26.0

tjtanaa:
PLEASE FILL IN THE PR DESCRIPTION HERE.

## Purpose

Fix all the CI failures in the ROCm CI.

## Test Plan

Make sure that the ready CI and merge CI are passing.

**vLLM Version:**

**vLLM-Omni Commit:**

## Test Result

Ready CI Result: https://buildkite.com/vllm/vllm-omni-amd-ci/builds/10008/list
Merge CI Result: https://buildkite.com/vllm/vllm-omni-amd-ci/builds/10003/list

**BEFORE SUBMITTING:** read [CONTRIBUTING.md](https://github.com/vllm-project/vllm-omni/blob/main/CONTRIBUTING.md) and run the [precheck-pr skill](https://github.com/vllm-project/vllm-omni/blob/main/.claude/skills/precheck-pr/SKILL.md) with the code agent for a self-check against project conventions.
(anything written below this line will be removed by GitHub Actions)

PR #5329, opened 2026-07-22
 [BugFix][Diffusion] Add CPU LAPACK fallback for FlowUniPC

akshatvishu:
## Purpose

Resolves #5327
The ROCm PyTorch wheels used in CI are built without CPU LAPACK support. Consequently, `FlowUniPCMultistepScheduler` raises a `RuntimeError` when `torch.linalg.solve` receives CPU tensors.

The initial fix routed the Cosmos3 mock pipeline to CUDA on ROCm, using hipSOLVER and avoiding the CPU LAPACK requirement. This only worked around the issue in  the test.

Now, I have instead fixes the underlying scheduler behavior:

  - Adds `safe_linalg_solve`, which falls back to NumPy when solving CPU tensors on a PyTorch build without LAPACK.
  - Uses the helper at both UniPC matrix-solve call sites.
  - Removes the ROCm-specific GPU workaround from `test_flow_unipc_reproducible_with_same_seed`.

GPU solves and CPU solves on PyTorch builds with LAPACK continue using `torch.linalg.solve`.

## Test Plan
Run the test suite on a ROCm platform:
```bash
pytest tests/diffusion/models/cosmos3/test_cosmos3_pipeline.py -k "test_flow_unipc_reproducible_with_same_seed"
```
image

**vLLM Version:** 0.25.0

**vLLM-Omni Commit:** 23547fd4f143e9b7526f365e0c046eee876901be

PR #5592, opened 2026-07-30
[ROCm] Dispatch AR-Diffusion paged KV attention to AITER

ZJLi2013:
## Purpose

On ROCm, every GPU parametrization of `tests/diffusion/ar_diffusion/test_paged_attention.py` fails:

```
ImportError: vllm.vllm_flash_attn requires the CUDA flash attention extensions
(_vllm_fa2_C or _vllm_fa3_C). On ROCm, use upstream flash_attn.
```

`ar_diffusion_paged_attention()` picks its backend with `query_flat.is_cuda`, which is `True` for HIP
tensors. ROCm therefore skips the CPU reference path and lands on the unguarded CUDA-only import.
The diffusion tower's attention path already tries aiter first on ROCm
(`diffusion/attention/backends/utils/fa.py:29-37`); this path never got the equivalent branch.

This PR adds that branch, between the existing CPU and CUDA ones and touching neither, so CUDA
behaviour is unchanged:

- `_rocm_varlen_fa()` resolves aiter's varlen attention once per process, through vLLM's
  `_aiter_ops` provider.
- `_pack_paged_kv()` maps `(block_table, seq_lens)` into aiter's packed varlen layout, vectorized.
- `_build_packed_kv_index()` builds that mapping's gather index, once per forward from `prepare()`.

It gathers KV instead of passing `block_table` to aiter because neither aiter paged entry point fits
this cache: `flash_attn_varlen_func(block_table=...)` faults the GPU on gfx950, and
`mha_batch_prefill_func` only ships bf16 kernels for `page_size ∈ {16, 1024}` while AR-Diffusion
pages at `tokens_per_frame` (1560 at 480p). The copy runs at HBM bandwidth (3.4-4.9 TiB/s) and is
~6% of attention time.

The test guard needed fixing too: `_cuda_flash_attn_usable()` returns `True` on ROCm because its
`ldd` probe raises, which is why these tests fail rather than skip. Without also accepting a
ROCm+aiter platform they would flip to *skipped* instead of passing.

What this PR does not do is route the dispatch itself through the diffusion attention registry and its
per-platform `get_diffusion_attn_backend_cls()` overrides; the branch stays inline. That refactor would
touch the CUDA path, which reviewers without AMD hardware cannot validate, so it belongs with #5582's
outcome rather than a platform fix.

Scope: this is the platform branch only, not end-to-end DreamZero on ROCm — a full rollout still
stops at FlowUniPC's `torch.linalg.solve` (no LAPACK), the blocker #5528 skips a test for and #5329
addresses.

## Test Plan

```bash
pytest -sv tests/diffusion/ar_diffusion/test_paged_attention.py
```

Oracle is the file's own `_dense_attention` reference (fp16, `rtol=atol=2e-2`), platform-independent.
Run on MI350X (gfx950), single GPU, image
`vllm/vllm-openai-rocm:nightly-49f31d7cee425a6d38f8c5bc76877986daf832ed`, torch 2.11.0+gitd0c8b1f.

**vLLM Version:** 0.23.1rc1.dev1533+g49f31d7ce.rocm723

**vLLM-Omni Commit:** 1d6c0816a4fbd9079878d62bbe28c1a12f146a1f

## Test Result

Before and after on the same base, toggled only by stashing the patch.

Before:

```
8 failed, 20 passed, 18 warnings in 1.59s
```

All 8 are `test_paged_attention_matches_dense_reference_gpu`, all the same ImportError. The other 20
already passed, so this file holds no platform assumption beyond that one missing branch.

After:

```
28 passed, 18 warnings in 68.22s
```

### DIFF
--- a/vllm_omni/experimental/ar_diffusion/kv_cache/paged_attention.py
+++ b/vllm_omni/experimental/ar_diffusion/kv_cache/paged_attention.py
@@ -25,6 +25,25 @@ def set_current_paged_kv_cache(kv_cache: Any) -> None:
     _CURRENT_PAGED_KV_CACHE = kv_cache
 
 
+# Packed-varlen gather index published by prepare() on the aiter path, same lifetime as
+# the pool registry above. It depends only on (block_table, seq_lens, page), which every
+# layer of one forward shares, so one build serves all of them. Kept next to the tensors
+# it was built from because ar_diffusion_paged_attention() is also reachable without a
+# prepare() (tests call it directly), and those calls must not pick up a stale index.
+_CURRENT_PACKED_KV_INDEX: tuple[torch.Tensor, torch.Tensor, int, torch.Tensor, torch.Tensor] | None = None
+
+
+def set_current_packed_kv_index(
+    block_table: torch.Tensor,
+    seq_lens: torch.Tensor,
+    page: int,
+    flat: torch.Tensor,
+    cu_seqlens_k: torch.Tensor,
+) -> None:
+    global _CURRENT_PACKED_KV_INDEX
+    _CURRENT_PACKED_KV_INDEX = (block_table, seq_lens, page, flat, cu_seqlens_k)
+
+
 _LAYER_IDX_TENSORS: dict[int, torch.Tensor] = {}
 
 
@@ -217,6 +236,11 @@ class ARDiffusionPagedForwardContext:
         the padded block-table metadata ONCE for all layers, and publishes the
         pool registry for the fused custom op. The compiled per-layer code then
         only consumes prebuilt tensors via ``ARDiffusionPagedLayerInputs``.
+
+        On the aiter path it also builds the packed-varlen gather index here, for
+        the same reason and with the same reach: it is a function of the block
+        table, so rebuilding it per layer costs 40x its 22 kernels and 3 host
+        syncs for an identical answer.
         """
         if getattr(self, "_prepared", False):
             return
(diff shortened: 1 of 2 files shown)


vllm_omni/experimental/ar_diffusion/kv_cache/paged_attention.py
```diff
@@ -365,6 +366,65 @@ def _resolve_fa_version(head_size: int) -> int:
     return version
 
 
+_ROCM_VARLEN_FA: Callable[..., Any] | None = None
+_ROCM_VARLEN_FA_RESOLVED = False
+
+
+def _rocm_varlen_fa() -> Callable[..., Any] | None:
```
tjtanaa, line 399: is using `@cache` better?

tjtanaa:
@ZJLi2013 could you share some data about the perf improvement?

vllm_omni/experimental/ar_diffusion/kv_cache/paged_attention.py
ZJLi2013: applied

ZJLi2013:
> @ZJLi2013 could you share some data about the perf improvement?

hi, @tjtanaa thanks for the review. This PR is an enablement fix, not a performance PR. Before it, every GPU parametrization of test_paged_attention.py on ROCm crashes with ImportError .

ZJLi2013:
Closing: main already resolves AITER and gathers paged KV into packed tensors before the call, so what this PR argues for has landed by another route.

Issue #5332, opened 2026-07-23
[RFC]: CI improvement plan — E2E layering, Acc, pinned deps, job matrix


yenuo26:
﻿## Description

> **Amendment:** Static quality tracks (**pre-commit**, **CodeQL**, **SonarQube Cloud**) moved to a dedicated RFC — see **#5992**. Existing E2E / Acc / pin / coverage / job-matrix plan unchanged below.

### Motivation.

| Issue                                | Current State                                                                                                                                                                                                                                                                                                               | Target State                                                                                                                                                                               |
| ------------------------------------ | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| Split L2 tiny framework              | [#4197](https://github.com/vllm-project/vllm-omni/issues/4197) partially landed under tests/model_tests/diffusion/ (DIFFUSION_TEST_SETTINGS, builders, offline/online/alignment), while most L2/L3/L4 e2e still lives as one-off modules under tests/e2e/{online_serving,offline_inference}/ (~100+ test_*.py)      | One layered layout under tests/e2e/models/; shared launch / request / assert logic lives in tests/helpers/                                                                                 |
| Boilerplate explosion                | Each model adds a near-copy of OmniServerParams / OmniRunner param lists, prompt builders, and send_*_request + assert glue (e.g. test_*_expansion.py)                                                                                                                                                               | L3: one online runner + one offline runner; cases are JSON. L2: one registry entry (DiffusionModelTestOpts / future Omni/TTS opts)                                                         |
| Unclear CI level ownership           | L2 tiny smoke vs L3/L4 full-weight e2e are mixed by filename convention (test_foo.py vs test_foo_expansion.py) and markers, but not by directory                                                                                                                                                                         | Directory + marker contract: tiny_model/ → L2; online_serving/ + offline_inference/ JSON → L3–L4 smoke                                                                                 |
| Helper drift                         | Assertions / runtime already centralize in tests/helpers/{runtime,assertions,mark}.py, but model-specific builders, step expansion, and case→fixture mapping still scatter in test modules                                                                                                                                  | Canonical helpers for builders, runner/server param expansion, run_level-aware validation                                                                                                |
[...]

PR #1120, opened 2026-01-31
[CI] Add CI branch coverage calculation,  fix statement coverage results and add log before test for buildkite  log group

yenuo26:
PLEASE FILL IN THE PR DESCRIPTION HERE ENSURING ALL CHECKLIST ITEMS (AT THE BOTTOM) HAVE BEEN CONSIDERED.

## Purpose
1.Add CI branch coverage calculation
2.fix statement coverage results.
3.Add log before test for buildkite Group
## Test Plan
For coverage calculation:
```
 pytest -v -s tests/entrypoints/
 pytest -v -s tests/diffusion/cache/ --cov-append
 pytest -v -s tests/diffusion/lora/ --cov-append
 pytest -v -s tests/model_executor/models/qwen2_5_omni/test_audio_length.py --cov-append
 pytest -v -s tests/worker/ --cov-append
 pytest -v -s tests/distributed/omni_connectors/test_kv_flow.py --cov-append
```
For buildkite log:
run CI
## Test Result
1.coverage calculation
```
=================================================================================== tests coverage ====================================================================================

Name                                                                                               Stmts   Miss Branch BrPart  Cover   Missing
TOTAL                                                                                              33274  29484   9658    195    10%
```
2.buildkite log
image
image

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft.
</details>

**BEFORE SUBMITTING, PLEASE READ <https://github.com/vllm-project/vllm-omni/blob/main/CONTRIBUTING.md>** (anything written below this line will be removed by GitHub Actions)

PR #5593, opened 2026-07-30
[CI] Per-model, per-entry-mode code coverage collection

ShengleiFu:
## Purpose

Implements PR 10 of the CI plan in #5332: collect coverage per model **and per entry mode**, uploaded as Buildkite artifacts.

CI collects repo-wide `--cov=vllm_omni` on two CPU jobs today, with no per-model breakdown. Online serving drives the whole `entrypoints/openai` stack in a separate server process; offline inference never touches it. That split is currently invisible.

This PR:

- Adds `.buildkite/common/scripts/run_cov_split.sh`, so opting a model in is one call with its model id and test paths — no per-job bash. It runs each mode, names the reports, uploads them, and fails the step if any half or the upload failed. Everything after `--` is forwarded verbatim to both runs with quoting preserved, so a job passes its own `-m '...' --run-level '...'` (or anything else) unchanged. `--offline`/`--online` are repeatable and individually optional, so a job with tests in one mode runs just that half.
- Applies it to two Merge-tier jobs, each writing `coverage-<model_id>-<mode>-<step_id>.xml.gz`. Both pilots are single-GPU so they are cheap to reproduce: `Diffusion · Bagel Test` (`h100_1`) and `TTS · Qwen3-TTS Base Test` (`l4_1`).
- On scheduled `main` Nightly builds the Ready and Merge E2E groups upload with `--e2e`, which bypasses diff-gating (`upload_pipeline.py` sets `changed_files = None`), so these two jobs run daily without duplicating them into `test-nightly.yml`. On ordinary PR/main builds they stay `source_file_dependencies`-gated, so coverage is not paid on every merge.
- Enables `[tool.coverage.run] patch = ["subprocess"]` and `sigterm = true`. Online tests start the server with `subprocess.Popen` and stop it with `SIGTERM`; without these, pytest-cov 7.x measures only the pytest parent. Also adopts `core = "sysmon"`, `source`, `concurrency` and `[tool.coverage.report]` from #5870 (thanks @NickCao); `sigterm` is kept, since that block without it measures only the parent.
- Sets `mount-buildkite-agent: true` on the `l4_*` presets (thanks @yenuo26). They run under the `docker` plugin, which leaves `buildkite-agent` off `PATH` inside the container, so the helper's upload failed there while the same code path passed on `h100_1` (`kubernetes`). Every artifact-uploading job in the repo today is on a `kubernetes` preset, so nothing had exercised this.
- Gzips each report before upload. `--cov=vllm_omni` sets coverage's `source`, so the XML lists every statement in the package, including files the model's tests never import: a Bagel report is 7,134,953 bytes, of which 96.5% is `<line>` elements and 77% belongs to the 840 of 1169 files that run never touched — a run covering zero lines produces the same 7 MB. That repetition gzips to 411,617 bytes (17×) with no data loss, so the reports are compressed rather than narrowing what is measured (which would cost the whole-package denominator). `gunzip` before feeding one to a Cobertura consumer.
- Folds the artifact-upload exit code into the step result — it previously ran after `EXIT` was computed under `set +e`, so a failed upload reported green.
- Raises `timeout_in_minutes` for the second model load a split implies (Bagel 20 → 40, Qwen3-TTS 40 → 60), sized from each job's original budget rather than from local timings on faster GPUs. That is the only runtime bound: an inner per-half `timeout` was dropped, since `timeout` sends only SIGTERM and a process that ignores it would block the call and never reach the upload.
- Documents the naming convention, the opt-in recipe, and two traps: a test whose `num_cards` exceeds its `mirror_hardwares` preset is silently skipped so its artifact measures nothing, and an empty half exits 5.

`Backward compatibility:` CI config, docs and `.gitignore` only; no product code. Non-pilot jobs untouched. `[tool.coverage.run]` is repo-wide, so the existing CPU coverage jobs also gain subprocess measurement.

`Not a duplicate:` no open PR references #5332; #1120 (merged) added repo-wide coverage, not per-model.

`Note:` both jobs list `run_cov_split.sh` and `pyproject.toml` in `source_file_dependencies` (thanks @congw729, @hsliuustc0106) — both change what the jobs measure, so a change there schedules them rather than failing a later nightly. A selector test in `tests/buildkite/test_upload_pipeline.py` pins that, and asserts an unrelated change still selects neither. Since this PR adds the helper, its own diff schedules both pilots. The results below come from running the exact job commands on GPU hosts.

`AI assistance:` developed with AI assistance (Claude); all findings below were reproduced on real hardware before submitting.

## Test Plan

**vLLM Version:** v0.26.0
**vLLM-Omni Commit:** 18be2ba3 (the base the GPU results below were produced on)

- [x] `python -c "import yaml; yaml.safe_load(open('.buildkite/cuda/test-merge.yml'))"`
- [x] `pytest tests/buildkite/test_upload_pipeline.py tests/buildkite/test_skip_ci.py` → 45 passed
- [x] Coverage config A/B on a `subprocess.Popen` child stopped by `SIGTERM` (coverage 7.15.4 / pytest-cov 7.1.0), 11 measurable lines: `sysmon` + `patch` + `sigterm` → 11/11, #5870's block as written → 2/11, no config → 2/11. On Python 3.11 coverage warns `Can't use core=sysmon … using default core` and still measures 11/11, so the `>=3.10` floor holds
- [x] Selector test verified RED/GREEN: dropping either dependency fails it, restoring both passes
- [x] Helper behaviour: single-mode runs, repeatable paths, missing-both rejected, `--` pass-through preserving quoting (`-m 'a and b'`, `-k 'foo bar'` arrive as single arguments), exit code non-zero when a half or the upload fails
- [x] Helper end-to-end locally: stale `.xml.gz` from an earlier run is cleared, report gzips 7,133,315 → 388,960 bytes, `gunzip` round-trips to the identical XML (`lines-valid=191447`, 1169 files), and a failed upload still exits non-zero
- [x] `pre-commit` clean on changed files
- [x] Each half collects ≥1 test (`--collect-only`): Bagel 2 + 2, Qwen3-TTS 1 + 2
- [x] Both pilots, both modes, with `--cov=vllm_omni --cov-report=xml:coverage-<model>-<mode>.xml`
- [x] Confirmed the online artifact contains server-subprocess coverage, not just import-time lines
- [x] Re-ran Qwen3-TTS through the helper after extracting it: `lines-covered` 14513 / 18972 vs 14512 / 18972 inline, i.e. identical online

## Test Result

| Pilot | offline | online |
| --- | --- | --- |
| Bagel | 2 passed (271s), 16318 lines | 2 passed (491s), 21792 lines |
| Qwen3-TTS Base | 1 passed (376s), 14512 lines | 2 passed (879s), 18972 lines |

Attribution runs in both directions — line-level set differences between the two artifacts:

| | covered only online | covered only offline |
| --- | --- | --- |
| Bagel | 5555 lines | 81 lines |
| Qwen3-TTS Base | 5162 lines | 702 lines |

Online-only is the serving stack that runs in the server child process — `openai/api_server.py` (395 / 404 lines), `serving_speech.py` (303 / 633), `cli/serve.py` (141 / 140), all zero offline. Offline-only includes `worker/omni_connector_model_runner_mixin.py` (260) and `core/sched/omni_scheduling_coordinator.py` (74). A union is derivable from the two reports; the attribution is not recoverable from a union.

Most usefully, Qwen3-TTS shows a **model-specific** gap: `model_executor/models/qwen3_tts/cuda_graph_decoder_wrapper.py` gets 158 lines online and **zero** offline. Repo-wide coverage cannot show that.

Two checks that the numbers mean what they claim:

- `engine/stage_engine_core_proc.py` is covered identically in both modes, so subprocess measurement works for the stage procs each mode spawns — the delta really is the HTTP layer.
- Without `patch`/`sigterm` those lines disappear, verified with a minimal repro (a `SIGTERM`-ed child went from only `import`/`def` lines measured to fully measured).

Counts are `<line hits!="0">` per `<class>` in the Cobertura XML; per-file sums reproduce coverage.py's own `lines-covered` exactly for all four artifacts.


yenuo26:
- Copying the same code into every job is too cumbersome.
Currently, opting in requires each model job to copy: splitting into two pytest runs, stitching EXIT_*, and then artifact upload.
Suggestion: Extract a common script/helper first (e.g., run_cov_split.sh <model_id> …), so the YAML only needs to pass the model ID and test path, rather than replicating a dozen lines of bash in every step.

- Coverage on Merge alone may not be comprehensive enough. Model tests are scattered across ready / merge / nightly / weekly. Many L2 models are not included in Ready. The full_model paths in Nightly / Weekly are often more meaningful for 'real coverage.'Since we already run both ready and merge E2E test groups in nightly, I think we could just add an additional sample test job to test-nightly for coverage collection. Also, given that computing coverage itself adds time overhead, I don't think we need to run it on every PR merge—once a day might be sufficient.

yenuo26:
Although I mentioned in the RFC using online and offline as separate entry points, I would also like to explore whether it's necessary to split collection between online and offline, or if it's more appropriate to collect per model as a single unit.

ShengleiFu:
> I would also like to explore whether it's necessary to split collection between online and offline, or if it's more appropriate to collect per model as a single unit.

Hi @yenuo26 , for separate versus combined reports, the pilot data shows that mode attribution is bidirectional at line granularity. Although neither pilot has a **file covered exclusively** offline, shared files contain **substantial sets of lines** reached by only one mode:

| | covered only online | covered only offline |
| --- | --- | --- |
| Bagel | 5555 lines | 81 lines |
| Qwen3-TTS Base | 5162 lines | 702 lines |

For example, online-only coverage includes 158 lines in qwen3_tts/cuda_graph_decoder_wrapper.py and the entrypoints/openai serving stack. Offline-only coverage includes 260 lines in worker/omni_connector_model_runner_mixin.py and 74 lines in core/sched/omni_scheduling_coordinator.py.

A combined per-model report can be derived by taking the union of the two mode reports, but the mode attribution cannot be reconstructed from that union. Since each mode uniquely exercises code that the other does not, a combined artifact would preserve total coverage while erasing both the online-only and offline-only signals.

For that reason, I would prefer to retain the per-mode artifacts and let the dashboard aggregate them when it needs a single per-model number. wdyt?

> Copying the same code into every job is too cumbersome.
Currently, opting in requires each model job to copy: splitting into two pytest runs, stitching EXIT_*, and then artifact upload.
Suggestion: Extract a common script/helper first (e.g., run_cov_split.sh <model_id> …), so the YAML only needs to pass the model ID and test path, rather than replicating a dozen lines of bash in every step.
Coverage on Merge alone may not be comprehensive enough. Model tests are scattered across ready / merge / nightly / weekly. Many L2 models are not included in Ready. The full_model paths in Nightly / Weekly are often more meaningful for 'real coverage.'Since we already run both ready and merge E2E test groups in nightly, I think we could just add an additional sample test job to test-nightly for coverage collection. Also, given that computing coverage itself adds time overhead, I don't think we need to run it on every PR merge—once a day might be sufficient.

Thanks for the valuable feedback, for those two, I will update accordingly.

yenuo26:
> > I would also like to explore whether it's necessary to split collection between online and offline, or if it's more appropriate to collect per model as a single unit.
> 
> Hi @yenuo26 , for separate versus combined reports, the pilot data shows that mode attribution is bidirectional at line granularity. Although neither pilot has a **file covered exclusively** offline, shared files contain **substantial sets of lines** reached by only one mode:
> 
> covered only online	covered only offline
> Bagel	5555 lines	81 lines
> Qwen3-TTS Base	5162 lines	702 lines
> For example, online-only coverage includes 158 lines in qwen3_tts/cuda_graph_decoder_wrapper.py and the entrypoints/openai serving stack. Offline-only coverage includes 260 lines in worker/omni_connector_model_runner_mixin.py and 74 lines in core/sched/omni_scheduling_coordinator.py.
> 
> A combined per-model report can be derived by taking the union of the two mode reports, but the mode attribution cannot be reconstructed from that union. Since each mode uniquely exercises code that the other does not, a combined artifact would preserve total coverage while erasing both the online-only and offline-only signals.
> 
> For that reason, I would prefer to retain the per-mode artifacts and let the dashboard aggregate them when it needs a single per-model number. wdyt?
> 
I think you make a very good point. With that in mind, keeping separate entry points for online and offline may be the better approach.

ShengleiFu:
Hi @yenuo26 , updated the pr based on the feedback, could you take another look, thanks!

.buildkite/common/scripts/run_cov_split.sh
congw729, line 1: Could we add .buildkite/common/scripts/run_cov_split.sh to the source_file_dependencies of both pilot jobs? Otherwise, changes to this shared helper will not trigger either job on regular PR builds, so regressions may only be discovered by the next nightly run.

ShengleiFu: good catch, done, thanks!

.buildkite/cuda/test-merge.yml
```diff
@@ -94,13 +90,21 @@ steps:
           - vllm_omni/model_executor/models/qwen3_tts/
           - vllm_omni/model_executor/stage_input_processors/qwen3_tts.py
           - vllm_omni/deploy/qwen3_tts.yaml
+          - .buildkite/common/scripts/run_cov_split.sh
+        # Must exceed both --timeout budgets plus upload margin, or this cap
+        # kills a half before its own timeout fires and the artifacts are lost.
+        timeout_in_minutes: 85
```
yenuo26: If the actual script timeout is 40 minutes, I don't think we need to extend it to 85 minutes here.
And To simplify the logic, I think you can just remove the --timeout parameter and rely solely on timeout_in_minutes for control.

.buildkite/common/scripts/run_cov_split.sh
```diff
@@ -0,0 +1,69 @@
...
+        --offline)   OFFLINE="$2"; shift 2 ;;
+        --online)    ONLINE="$2"; shift 2 ;;
+        --markers)   MARKERS="$2"; shift 2 ;;
+        --run-level) RUN_LEVEL="$2"; shift 2 ;;
+        --timeout)   TIMEOUT="$2"; shift 2 ;;
+        *) echo "run_cov_split.sh: unknown argument: $1" >&2; exit 2 ;;
+    esac
+done
+
+for required in MODEL_ID OFFLINE ONLINE MARKERS RUN_LEVEL; do
```
yenuo26: Not every job actually has both offline and online test cases. I think it would be more reasonable to change the logic here so that either online or offline is sufficient.

.buildkite/common/scripts/run_cov_split.sh
```diff
@@ -0,0 +1,69 @@
...
+RUN_LEVEL=""
+TIMEOUT="40m"
+
+while [[ $# -gt 0 ]]; do
+    case "$1" in
+        --model-id)  MODEL_ID="$2"; shift 2 ;;
+        --offline)   OFFLINE="$2"; shift 2 ;;
+        --online)    ONLINE="$2"; shift 2 ;;
+        --markers)   MARKERS="$2"; shift 2 ;;
+        --run-level) RUN_LEVEL="$2"; shift 2 ;;
```
yenuo26: I'd suggest generalizing this to a common command-line arguments placeholder rather than restricting it specifically to run-level. In practice, other nightly tests may pass additional arguments as well, such as --test-config-file.

.buildkite/common/scripts/run_cov_split.sh
```diff
@@ -0,0 +1,69 @@
...
+done
+
+rm -f "coverage-${MODEL_ID}-offline.xml" "coverage-${MODEL_ID}-online.xml"
+
+run_half() {
+    local mode="$1"
+    shift
+    echo "--- coverage: ${MODEL_ID} ${mode}"
+    # shellcheck disable=SC2086  # paths are intentionally word-split
+    timeout "${TIMEOUT}" pytest -s -v $* \
```
hsliuustc0106: `timeout` only sends SIGTERM. If pytest or its server does not exit after that signal, this call stays blocked and we never reach the second half or artifact upload; the Buildkite step cap then kills the whole step. Please add a kill-after/process-tree policy and budget its grace period.

.buildkite/cuda/test-merge.yml
```diff
@@ -90,13 +90,21 @@ steps:
           - vllm_omni/model_executor/models/qwen3_tts/
           - vllm_omni/model_executor/stage_input_processors/qwen3_tts.py
           - vllm_omni/deploy/qwen3_tts.yaml
+          - .buildkite/common/scripts/run_cov_split.sh
```
hsliuustc0106, line 97: `pyproject.toml` is now a direct input to these runs (`patch`/`sigterm`), but it is not a source dependency. With only that file changed, `_process_test_steps` drops both pilots on normal PR builds. Please add it to both lists and cover this selector case.

.buildkite/common/scripts/run_cov_split.sh
ShengleiFu: As per the above comment from @yenuo26 , `timeout` has been removed.

.buildkite/cuda/test-merge.yml
ShengleiFu: Make sense, added it

.buildkite/common/scripts/run_cov_split.sh
ShengleiFu: Good point, updated it

.buildkite/common/scripts/run_cov_split.sh
ShengleiFu: updated, thanks!

.buildkite/cuda/test-merge.yml
ShengleiFu: removed

ShengleiFu:
Thank you both for the review! I just addressed all of them and plz take another look, thanks! @yenuo26 @hsliuustc0106

.buildkite/common/scripts/run_cov_split.sh
```diff
@@ -0,0 +1,66 @@
...
+    echo "run_cov_split.sh: need at least one of --offline / --online" >&2
+    exit 2
+fi
+
+run_half() {
+    local mode="$1"
+    shift
+    echo "--- coverage: ${MODEL_ID} ${mode}"
+    pytest -s -v "$@" "${PYTEST_ARGS[@]}" \
+        --cov=vllm_omni --cov-report="xml:coverage-${MODEL_ID}-${mode}.xml"
```
yenuo26: I think the filename here may need a random suffix. In nightly CI, different test types/jobs may use the same model.
Other than that, I have no further questions.

ShengleiFu: Good point!

Would `BUILDKITE_STEP_ID` work for you as the suffix, rather than a random value?

    coverage-<model_id>-<mode>-<step_id>.xml

Two reasons I'd lean that way:

- It's Buildkite metadata, so a report can be traced back to the job that produced it. A
  random value is an orphan the moment it's written — if one model's numbers ever look
  wrong, there's nothing to trace.
- It stays the same across retries, which matters because the upload here is
  unconditional: a half that fails still gets its partial report uploaded. With a
  per-attempt value (a random one, or `BUILDKITE_JOB_ID`) the failed attempt's partial XML
  and the retry's full one end up side by side under different names, both matched by a
  glob. Sharing one path avoids that — the step id remains the logical step identifier
  across retries, while Buildkite's default artifact lookup selects only the latest
  attempt.

If you'd prefer readable filenames over a UUID, the alternative is giving the opted-in
steps a `key:` and using `BUILDKITE_STEP_KEY` with the step id as fallback — happy to go
that way instead.

Plz let me know @yenuo26 , thanks!

ShengleiFu: Hi @yenuo26 , gentle ping on this. Thanks!

yenuo26: I apologize for the delayed response. I think BUILDKITE_STEP_ID works well for this use case.

yenuo26 (approved):
LGTM

yenuo26:
@hsliuustc0106 @Gaohan123 @NickCao please help to add merge-test label

ShengleiFu:
CI failed, both halves passed and both reports were written; only the upload failed:

Coverage XML written to coverage-qwen3_tts-offline-019fef92-….xml
Coverage XML written to coverage-qwen3_tts-online-019fef92-….xml
run_cov_split.sh: line 74: buildkite-agent: command not found

`l4_*` presets run under the `docker` plugin, which does not put `buildkite-agent` on
`PATH` inside the container. `h100_*` presets run under `kubernetes`, which does — so
"Diffusion · Bagel Test" (`h100_1`) passed on the same code path. Every artifact-uploading
job in the repo today is on an `h100_*` preset, so nothing had exercised this before.

Proposal:

- writes reports to `coverage-reports/` and publishes them with
  `artifact_paths: "coverage-reports/*.xml"` on each opted-in step; the helper no longer
  uploads anything itself
- recreates that directory on entry, so on the docker presets — where one checkout is
  reused across jobs — an earlier job's report cannot be published as this run's
- adds a `tests/buildkite/` check that every step invoking the helper declares
  `artifact_paths`, so the next model to opt in cannot repeat this
- records the constraint in `ci_settings.md`

Pros: `artifact_paths` is run by the agent after the command, so the reports now
survive a run that dies before the helper's last line — previously an early crash lost
both.

Cons: this moves the upload out of the helper and into each step's YAML

@yenuo26 plz let me know if you have any feedback or other suggestions, thanks!

yenuo26:
> CI failed, both halves passed and both reports were written; only the upload failed:
> 
> Coverage XML written to coverage-qwen3_tts-offline-019fef92-….xml Coverage XML written to coverage-qwen3_tts-online-019fef92-….xml run_cov_split.sh: line 74: buildkite-agent: command not found
> 
> `l4_*` presets run under the `docker` plugin, which does not put `buildkite-agent` on `PATH` inside the container. `h100_*` presets run under `kubernetes`, which does — so "Diffusion · Bagel Test" (`h100_1`) passed on the same code path. Every artifact-uploading job in the repo today is on an `h100_*` preset, so nothing had exercised this before.
> 
> Proposal:
> 
> * writes reports to `coverage-reports/` and publishes them with
>   `artifact_paths: "coverage-reports/*.xml"` on each opted-in step; the helper no longer
>   uploads anything itself
> * recreates that directory on entry, so on the docker presets — where one checkout is
>   reused across jobs — an earlier job's report cannot be published as this run's
> * adds a `tests/buildkite/` check that every step invoking the helper declares
>   `artifact_paths`, so the next model to opt in cannot repeat this
> * records the constraint in `ci_settings.md`
> 
> Pros: `artifact_paths` is run by the agent after the command, so the reports now survive a run that dies before the helper's last line — previously an early crash lost both.
> 
> Cons: this moves the upload out of the helper and into each step's YAML

Add mount-buildkite-agent: true  in .buildkite/common/ci_mirror_hardwares.yml maybe can fix it. Like
```
l4_1:
  agents:
    queue: gpu_1_queue
  plugins:
    - docker#v5.2.0:
        image: public.ecr.aws/q9t5s3a7/vllm-ci-test-repo:$BUILDKITE_COMMIT
        always-pull: true
        propagate-environment: true
        mount-buildkite-agent: true   # ← Add
        shm-size: "8gb"
        environment:
          - HF_HOME=/fsx/hf_cache
          - HF_TOKEN
        volumes:
          - /fsx/hf_cache:/fsx/hf_cache
```
I think this is a fairly small change, but we probably need to verify whether it actually works.

pyproject.toml
```diff
@@ -280,6 +280,14 @@ filterwarnings = [
     "ignore:.*does not have '__bases__' attribute.*:UserWarning",
 ]
 
+[tool.coverage.run]
```
NickCao, line 304: Could you add the changes from https://github.com/vllm-project/vllm-omni/pull/5870/changes? `core = "sysmon"` greatly improves the performance.

ShengleiFu: Added, PTAL, thanks!

ShengleiFu:
> Add mount-buildkite-agent: true in .buildkite/common/ci_mirror_hardwares.yml maybe can fix it.

This looks neat, updated. We need to re-apply `merge-test` label to trigger the validation.

ShengleiFu:
Update for artifact size — each report is 6.8 MB, which is too large

The size is driven by what gets *listed*, not what gets covered. `--cov=vllm_omni` sets `source`, so coverage walks the package and emits one `<line>` element per statement for every file, including ones the test never imports. Measured on the real `coverage-bagel-offline-*.xml`:

- 191,447 `<line>` elements = 96.5% of the bytes; only 17,176 are covered
- 840 of the 1,169 files were never imported — 77% of the file
- a local run covering **zero** lines produces the identical 191,447-line inventory

| | bytes | gzipped |
| --- | --- | --- |
| today | 7,134,953 | 411,617 |
| only files actually imported (`--cov` + `[run] include`) | 1,668,295 | 110,330 |

- `skip_empty = True` → 7,134,953 → 7,116,950 bytes (**−0.25%**). It skips files with *no executable statements* (105 of them here), not files at 0% coverage — and those files contribute no `<line>` elements to begin with.
- `show_missing = False` → **byte-identical**. It only affects the text/`term` report; the XML writer ignores it.

Also worth flagging: these cannot go in a new `.coveragerc`. Coverage reads a *single* config file and `.coveragerc` outranks `pyproject.toml` — adding one silently drops `[tool.coverage.run]`, i.e. `patch`/`sigterm`, which takes online coverage back to parent-only (verified 11/11 → 2/11 on the repro above). `.coveragerc.ini` is not a filename coverage looks for at all, so it has no effect either way.

Proposal: gzip on upload in this PR (17× smaller, no semantic change) 

cc: @yenuo26

yenuo26 (approved):
LGTM

yenuo26:
@hsliuustc0106 @Gaohan123 @amy-why-3459 please help to check whether this can be merged

PR #5604, opened 2026-07-30
[MiniCPM-o][NPU] Enable MiniCPM-o Code2Wav NPUGraph replay

Zhou248:
## Summary

Enable Ascend NPUGraph capture/replay for the MiniCPM-o 4.5 Stage 2 Code2Wav CFM DiT estimator.

This is a scoped implementation of the Stage 2 graph item discussed in #5069. It keeps the Python-heavy and dynamic parts of Code2Wav eager while replaying the largest deterministic estimator kernel for exact-shape buckets.

## Why this boundary

The complete Stage 2 forward cannot be captured safely:

| Path | Execution mode | Reason |
| --- | --- | --- |
| CFM DiT estimator `blocks_forward_chunk` | NPUGraph | deterministic tensor-only compute with exact-shape inputs |
| Timestep embedding | eager | upstream code creates CPU frequency tensors and performs H2D copies |
| Flow encoder | eager | dynamic streaming/cache boundary |
| HiFT | eager | random phase generation and waveform overlap state |
| Request parsing/state commit | eager | Python metadata and request-owned dictionaries |

Stage 2 therefore intentionally keeps `enforce_eager: true` at the vLLM generation-runner level. The inner graph is managed by the Code2Wav backend and does not conflict with Stage 0/1 PIECEWISE graph execution.

## Implementation

- Add an exact-signature graph cache keyed by operation, cache mode, tensor shape, dtype, and full device.
- Warm each new signature eagerly once, capture it with the vLLM global graph pool, and replay from the next matching call.
- Copy current values into static graph inputs before replay.
- Clone persistent graph outputs before storing request-owned streaming caches so later replay cannot overwrite another request's state.
- Precompute CFM timestep embeddings outside capture and pass them into the estimator graph.
- Keep cached and uncached estimator paths in separate graph buckets.
- Cap the cache with `code2wav_max_npu_graphs` (default: 32); unseen shapes fall back to eager after the cap.
- Treat capture failure as process-fatal because torch-npu may leave allocator/capture state invalid.
- Force graph-compatible Ascend runtime settings before loading Token2Wav:
  - `torch.npu.config.allow_internal_format = False`
  - `torch.npu.set_compile_mode(jit_compile=False)`
  - reject `ASCEND_LAUNCH_BLOCKING=1`
- Require the MATH SDPA backend during graph mode and stop swallowing exceptions raised by the Token2Wav body.
- Enable Stage 2 inner NPUGraph in the bundled 1-GPU, 2-GPU, and 3-GPU MiniCPM-o 4.5 deploy configs. It can be disabled with `code2wav_enable_npu_graph: false`.

## Correctness and tests

Added CPU regression coverage for:

- eager-equivalent timestep scheduling
- exact-shape capture/replay dispatch
- separate cached/uncached graph buckets
- use of the vLLM global graph pool
- persistent-output clone ownership
- eval-mode and runtime preconditions
- fail-stop behavior after capture failure
- SDPA context exception propagation and required MATH selection

Added an A3 hardware test that captures a real `cosyvoice2.flow.decoder_dit.DiT` estimator and compares eager vs NPUGraph outputs for both uncached and cached attention/CNN state. The test is wired into `.buildkite/npu/test-npu-ready.yml`.

User-side A/B validation on the same Stage 2 workload showed that enabling the inner NPUGraph reduced Stage 2 TTFT by approximately 15% compared with eager execution.

Local checks completed:

- `ruff format --check`: passed
- `ruff check`: passed
- `python -m compileall`: passed
- Buildkite/deploy YAML parse: passed
- `git diff --check`: passed

Local pytest could not collect because the available macOS Anaconda PyTorch installation is missing `libtorch_cpu.dylib`. The real NPUGraph test also requires an Ascend A3 environment and is intentionally delegated to NPU CI.

## Performance validation

User-side A/B validation measured approximately 15% lower Stage 2 TTFT with inner NPUGraph replay than with eager execution. Before marking ready, this result should be accompanied by the exact A/B command, A3 hardware/software versions, warmup and measured rounds, concurrency, latency/RTF range, and peak NPU memory.

## Risk and rollback

- Each captured signature consumes graph-pool memory; `code2wav_max_npu_graphs` bounds this growth.
- New shapes run eagerly after the limit.
- Operators can disable only Stage 2 capture with `code2wav_enable_npu_graph: false` while retaining Stage 0/1 graph execution.
- Capture failures emit an actionable restart/disable error rather than continuing with potentially corrupted capture state.

Related to #5069.

### DIFF
--- a/.buildkite/npu/test-npu-ready.yml
+++ b/.buildkite/npu/test-npu-ready.yml
@@ -11,6 +11,7 @@ steps:
         commands:
           - >-
             pytest -s -v
+            tests/platforms/npu/test_graph_tools.py
             tests/platforms/npu/test_minicpmo_code2wav_npugraph.py
             tests/e2e/offline_inference/test_minicpmo_4_5.py
             tests/e2e/online_serving/test_minicpmo_4_5.py
--- a/.buildkite/npu/test-npu-ready.yml
+++ b/.buildkite/npu/test-npu-ready.yml
@@ -1,4 +1,15 @@
 steps:
+  - label: "Platforms Test"
+    key: ready-platforms-test
+    depends_on: upload-ready-pipeline
+    mirror_hardwares: a3_npu_2
+    commands:
+      - >-
+        pytest -s -v
+        tests/platforms/npu/test_minicpmo_code2wav_npugraph.py
+        -m 'core_model and npu and A3 and omni'
+        --run-level 'core_model'
+
   - group: ":card_index_dividers: Omni Model Test"
     key: ready-omni-group
     depends_on: upload-ready-pipeline
@@ -9,13 +20,7 @@ steps:
           VLLM_WORKER_MULTIPROC_METHOD: spawn
           VLLM_ALLOW_LONG_MAX_MODEL_LEN: "1"
         commands:
-          - >-
-            pytest -s -v
-            tests/platforms/npu/test_minicpmo_code2wav_npugraph.py
-            tests/e2e/offline_inference/test_minicpmo_4_5.py
-            tests/e2e/online_serving/test_minicpmo_4_5.py
-            -m 'core_model and npu and A3 and omni'
-            --run-level 'core_model'
+          - pytest -s -v tests/e2e/offline_inference/test_minicpmo_4_5.py tests/e2e/online_serving/test_minicpmo_4_5.py -m 'core_model and npu and A3 and omni' --run-level 'core_model'
       - label: "Omni · MiniCPM-o 4.5 Duplex Test"
         mirror_hardwares: a3_npu_2
         env:
--- a/.buildkite/npu/test-npu-ready.yml
+++ b/.buildkite/npu/test-npu-ready.yml
@@ -9,7 +9,13 @@ steps:
           VLLM_WORKER_MULTIPROC_METHOD: spawn
           VLLM_ALLOW_LONG_MAX_MODEL_LEN: "1"
         commands:
-          - pytest -s -v tests/e2e/offline_inference/test_minicpmo_4_5.py tests/e2e/online_serving/test_minicpmo_4_5.py -m 'core_model and npu and A3 and omni' --run-level 'core_model'
+          - >-
+            pytest -s -v
+            tests/platforms/npu/test_minicpmo_code2wav_npugraph.py
+            tests/e2e/offline_inference/test_minicpmo_4_5.py
+            tests/e2e/online_serving/test_minicpmo_4_5.py
+            -m 'core_model and npu and A3 and omni'
+            --run-level 'core_model'
       - label: "Omni · MiniCPM-o 4.5 Duplex Test"
         mirror_hardwares: a3_npu_2
         env:
--- a/.buildkite/npu/test-npu-ready.yml
+++ b/.buildkite/npu/test-npu-ready.yml
@@ -11,7 +11,6 @@ steps:
         commands:
           - >-
             pytest -s -v
-            tests/platforms/npu/test_graph_tools.py
             tests/platforms/npu/test_minicpmo_code2wav_npugraph.py
             tests/e2e/offline_inference/test_minicpmo_4_5.py
             tests/e2e/online_serving/test_minicpmo_4_5.py
--- a/recipes/OpenBMB/MiniCPM-o-4_5.md
+++ b/recipes/OpenBMB/MiniCPM-o-4_5.md
@@ -92,9 +92,10 @@ to 32 entries by default, after which unseen shapes run eagerly.
 An ACL graph capture failure is fatal to that Stage-2 process because older
 torch-npu releases can leave allocator and RNG capture state invalid. Restart
 the service after a capture failure. To run without capture, set
-`code2wav_enable_npu_graph: false` before startup. Tune the cache limit with
-`code2wav_max_npu_graphs` in the connector `extra` config. Graph mode also
-requires `ASCEND_LAUNCH_BLOCKING` to be unset or set to `0`.
+`code2wav_enable_npu_graph: false` under the Stage-2
+`platforms.npu.stages[].additional_config` block before startup. Tune the cache
+limit there with `code2wav_max_npu_graphs`. Graph mode also requires
+`ASCEND_LAUNCH_BLOCKING` to be unset or set to `0`.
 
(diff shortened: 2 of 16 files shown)


amy-why-3459:
**Gate status**: Merge conflict unresolved (`mergeable: false, dirty`), please rebase first.

**Overall assessment**: High-quality code, correct boundary design, thorough test coverage. Two blocking issues to address first, plus two non-blocking suggestions.

---

### Blocking

**1. Merge conflict unresolved**

The branch has conflicts with `main` and cannot be merged. Please rebase first before proceeding with review.

**2. Incomplete performance data**

The PR description mentions "approximately 15% lower Stage 2 TTFT" but immediately acknowledges that the data is incomplete — missing the exact A/B command, A3 hardware/software versions, warmup and measurement rounds, concurrency, latency/RTF range, and peak NPU memory. While the title is `[Model]` rather than `[Perf]`, the core value of NPUGraph is the performance gain, and approving without reproducible, complete performance data is weak.

Please provide:
- Exact test commands
- A3 environment info (NPU model, driver version, torch-npu version)
- At least 3 measurement runs, mean ± stddev, with clear warmup round count
- Concurrency and throughput data
- Full latency distribution (not just mean)
- Peak NPU memory

---

### Non-blocking suggestions

**3. NPU graph logic is too tightly coupled with `BatchedToken2Wav`**

`batched_token2wav.py` has roughly 244 new lines in this diff, ~70% of which are NPU graph-related methods and attributes (`_npu_graph_supported`, `_npu_stream_is_capturing`, `_npu_graph_eligible`, `_capture_npu_graph`, `_run_graphable`, etc., all with `npu` in the name). This inflates the responsibilities of `BatchedToken2Wav` (a model-driven class) and introduces direct dependencies on the `torch.npu` namespace.

A better approach would be to extract NPUGraph capture/replay logic into a standalone platform module (e.g. `vllm_omni/platforms/npu/graph_tools.py`), with `BatchedToken2Wav` driving it through an injected graph runner object and platform checks staying in `_build_backend`. This would:
- Keep `BatchedToken2Wav` clean, without direct NPU namespace dependencies
- Allow the graph runner to be reused by CosyVoice3 and other Code2Wav-like models
- Keep platform code in the platform layer, consistent with the `vllm_omni/platforms/npu/` architecture convention

**4. Fragmented H2D copies in `_apply_batched_repetition_penalty`**

```python
for row, history in enumerate(histories):
    recent = history.reshape(-1)[-window_size:].to(device=logits.device, dtype=torch.long)
```

Each request does a `history.to(device)` inside the loop. Since the Talker stage codec sampling batch size is usually small (≤4), the real-world impact is negligible. If the batch size grows in the future, consider moving the most recent windows of all histories to device in a single batched operation.

---

amy-why-3459:
Hi @Zhou248 — this branch is conflicting with `main` again and isn't mergeable.

Please rebase onto current `main` and resolve the conflicts. Two related PRs landed after this branch last moved and will overlap the Code2Wav graph path:

- #5869 HiFT CUDA Graph (`enable_hift_graph`)
- #6082 CFM DiT CUDA Graph (`enable_cfm_graph` / `cfm_max_graphs`)

Reuse the in-tree graph wrapper / yaml knobs where they already cover the same surface, and keep this PR scoped to the NPUGraph estimator replay. After the rebase, ping here and we can continue review.

Zhou248:
Rebased this PR onto the current `main` and resolved the conflicts.

The resolution preserves the CUDA graph work from #5869 and #6082 (`enable_hift_graph`, `enable_cfm_graph`, and `cfm_max_graphs`) while keeping the Ascend NPUGraph implementation scoped to `vllm_omni/platforms/npu/`. The shared MiniCPM-o `model_executor` files and in-tree Token2Wav adapter now remain unchanged from `main`; the NPU platform patch adapts to the current `_estimator_step(..., time=...)` interface and keeps timestep embedding outside capture.

Current head: `1b464e8f`. GitHub reports the PR as mergeable. `pre-commit` and `Build Wheel` are passing; Read the Docs is still pending.

.buildkite/npu/test-npu-ready.yml
```diff
@@ -9,7 +9,13 @@ steps:
           VLLM_WORKER_MULTIPROC_METHOD: spawn
           VLLM_ALLOW_LONG_MAX_MODEL_LEN: "1"
         commands:
-          - pytest -s -v tests/e2e/offline_inference/test_minicpmo_4_5.py tests/e2e/online_serving/test_minicpmo_4_5.py -m 'core_model and npu and A3 and omni' --run-level 'core_model'
+          - >-
+            pytest -s -v
```
amy-why-3459: @yenuo26 PTAL

.buildkite/npu/test-npu-ready.yml
```diff
@@ -9,7 +9,13 @@ steps:
...
+            tests/platforms/npu/test_minicpmo_code2wav_npugraph.py
```
yenuo26: Please don't add this to the existing E2E job. Instead, you can create a new job named 'Platforms Test'.

Zhou248: Addressed in `4a6be60c`. The NPUGraph hardware test now runs in a standalone top-level job labeled `Platforms Test`, and the existing MiniCPM-o E2E job has been restored to its original offline/online test scope. I also verified the rendered pipeline expands the new job to the A3 NPU runner.

Zhou248:
Fixed the PR-attributable CI failures in commit `4a1b878c`:

- Lazily load `NPUOmniPlatform` so CPU/ROCm jobs can collect the lightweight NPU helper tests without requiring the optional `vllm_ascend` package.
- Restore the stage 0/1 `PIECEWISE` NPU graph overrides in `minicpmo_4_5_3gpu.yaml`, matching the other MiniCPM deployment configs.
- Fix the NPUGraph hardware test's time embedding shape to match the production `t_embedder(time).unsqueeze(1)` path (`[B, 1, H]`).

GitHub Actions (pre-commit and wheel builds), DCO, and Read the Docs pass on the new head. The external Buildkite contexts have not been reported for `4a1b878c` yet; please retrigger them if they do not start automatically.

The AMD LTX2 `rocm_unquantized_gemm` failure seen on the previous head is also present on the adjacent upstream-main build (#10645), so it is unrelated to this PR.

amy-why-3459 (approved):

PR #5610, opened 2026-07-30
[Doc] Document async Omni output materialization

fake0fan:
## Purpose

Document async Omni output materialization and how it removes CPU-side payload construction from the AR decode critical path.

This PR:

- explains the synchronous bottleneck, safe tensor snapshots, asynchronous D2H copies, and background output builder;
- includes the Qwen3-Omni blog figure that illustrates the Talker decode-step gap reduction;
- records the controlled Qwen3-Omni performance results from the optimization blog;
- provides activation examples for Qwen3-Omni and Qwen3-TTS, including required runtime guards and fallback behavior; and
- adds the feature page to the documentation navigation and design index.

## Test Plan

- Validate Markdown anchors, local links, image references, and fenced code blocks.
- Parse the vendored SVG and documentation navigation YAML.
- Run repository pre-commit hooks against the staged documentation changes.
- Run `git diff --check`.
- Run `mkdocs build --strict` when the documentation dependency set resolves.

**vLLM Version:** N/A (documentation-only change)

**vLLM-Omni Commit:** `b3f4fbf9e217ab6174fc0c475e33a478006ac34c` (`origin/main` base)

## Test Result

- Markdown links, anchors, image references, and code fences: passed.
- SVG XML parsing and local figure resolution: passed.
- `docs/.nav.yml` YAML parsing and navigation target validation: passed.
- Pre-commit hooks: passed.
- `git diff --check`: passed.
- `mkdocs build --strict`: blocked before MkDocs startup by the existing dependency-resolution conflict where `qwen-asr==0.0.6` pins `transformers==4.57.6` while the project requires `transformers>=5.5.3`.

### DIFF
--- a/docs/design/feature/omni_async_output_materialization.md
+++ b/docs/design/feature/omni_async_output_materialization.md
@@ -55,6 +55,13 @@ background output path
 This allows payload construction for step `N` to overlap with GPU decode work
 for step `N + 1`. It does not change model computation or generated output.
 
+!!! note "Platform scope"
+    This design currently applies to the CUDA and ROCm AR runners, which use
+    `GPUARModelRunner`. Ascend NPU uses a separate `NPUARModelRunner` path that
+    still constructs `OmniModelRunnerOutput` and drains connector output
+    synchronously before wrapping the result as an asynchronous runner output.
+    XPU and MUSA are not currently covered or validated by this design.
+
 <p align="center">
   <img
     alt="Decode step gap before and after async output materialization"
@@ -67,7 +74,7 @@ for step `N + 1`. It does not change model computation or generated output.
   <em>
     Figure 1: Moving payload construction off the decode path reduces the
     observed gap between consecutive Talker steps from about 2.8 ms to 41 µs.
-    Source:
+    Adapted from the
     <a href="https://vllm.ai/blog/2026-07-01-qwen3-omni-optimization">
       Qwen3-Omni optimization blog
     </a>.
@@ -158,7 +165,7 @@ next decode step                                +-- accumulate full payloads
 ### Safe State Snapshots
 
 The next scheduler step can mutate runner state while the background builder is
-still active. The builder therefore consumes snapshots rather than live fields.
+still active. Most step-local builder inputs are therefore detached snapshots.
 The snapshotted state includes:
 
 - scheduler token counts and speculative-token metadata
@@ -166,12 +173,30 @@ The snapshotted state includes:
 - sampled token IDs, logprobs, and prompt logprobs
 - query start locations and scheduled-token spans
 - hidden states and multimodal output tensors
-- KV, encoder-cache, and Omni connector output state
+- KV-connector and encoder-cache outputs captured for the step
 
 CUDA tensor cloning is required because CUDA Graph and model output buffers can
 be reused by the next forward pass. The dedicated copy stream and pinned host
 buffers make the D2H transfer asynchronous; the snapshot retains its cloned
 CUDA sources until the transfer event completes.
+
+### Live Runner State and Ownership
+
+The complete builder input is not a detached snapshot. Two parts deliberately
+remain runner-owned:
+
+- Full-payload accumulation looks up each request in `self.requests` and
+  updates the runner's pending full-payload accumulator.
+- `get_omni_connector_output()` drains live, per-cycle connector signals after
+  payload accumulation. Connector output state itself is not snapshotted.
+
+Correctness therefore depends on the output lifecycle: request entries and the
+full-payload accumulator must remain valid until the background builder
+finishes, and that builder must be the only consumer that drains connector
+signals for its output cycle. Receive-side connector state written by the
+background receiver is coordinated by the connector mixin's `_lock`.
+`OmniAsyncGPUModelRunnerOutput.get_output()` joins the builder and is the
+completion and exception boundary before the materialized output is consumed.
 
 ### Output Builder
 
@@ -182,7 +207,7 @@ The background builder performs the work that previously ran inline:
 - applies model-specific payload processing
 - accumulates full payloads when required
 - creates the tensor-only `multimodal_outputs` wire payload
-- reads connector readiness after payload accumulation
+- drains live connector readiness signals after payload accumulation
 - constructs the final `OmniModelRunnerOutput`
 
 Input-side connector operations remain synchronous with model execution. In
@@ -206,8 +231,9 @@ uses the normal generation-stage path.
 
 There is no separate `async_omni_output` command-line or YAML option. The
 feature is selected automatically for model stages that opt in when the runtime
-conditions are safe. For the supported Qwen3-Omni and Qwen3-TTS pipelines,
-using the bundled deployment profile is enough to turn it on.
+conditions are safe. On CUDA and ROCm, using the bundled deployment profile is
+enough to turn it on for the supported Qwen3-Omni and Qwen3-TTS pipelines. The
+same profile does not enable background Omni materialization on Ascend NPU.
 
 ### Example 1: Qwen3-Omni
 
@@ -294,6 +320,7 @@ The async output path is used only when all of the following conditions hold:
 
 | Requirement | Reason |
 |---|---|
+| The platform uses the CUDA or ROCm `GPUARModelRunner` path | Ascend NPU materializes the Omni output synchronously; XPU and MUSA are not covered or validated by this design |
 | AR async scheduling is enabled | The optimization relies on the scheduler advancing while the prior output is materialized |
 | `async_chunk` is enabled | The feature targets incremental downstream Omni payloads |
 | The model stage opts in with `use_async_omni_output` | Models must declare that their output lifecycle is safe to defer |
@@ -302,9 +329,12 @@ The async output path is used only when all of the following conditions hold:
 | Routed-expert output is disabled | Routed-expert extraction currently requires the synchronous path |
 | Postprocess is absent or explicitly runs eagerly | State needed by the next decode step must be updated before the runner returns |
 
-These checks are evaluated per stage. An unsupported combination does not
-prevent serving; it only falls back to synchronous output construction for that
-stage.
+On CUDA and ROCm, these checks are evaluated per stage. An unsupported
+combination does not prevent serving; it only falls back to synchronous output
+construction for that stage. On Ascend NPU, `NPUARModelRunner.sample_tokens()`
+fully constructs `OmniModelRunnerOutput`, calls
+`get_omni_connector_output()`, and only then creates the asynchronous wrapper,
+so Omni payload materialization remains synchronous.
 
 !!! note
     `use_async_omni_output`,
@@ -316,6 +346,8 @@ stage.
 
 - `vllm_omni/worker/gpu_ar_model_runner.py`: Async output object, safe tensor
   snapshots, compatibility guards, and deferred Omni output builder.
+- `vllm_omni/platforms/npu/worker/npu_ar_model_runner.py`: Separate Ascend NPU
+  runner, where Omni output materialization remains synchronous.
 - `vllm_omni/model_executor/models/qwen3_omni/qwen3_omni.py`: Thinker and
   Talker opt-in behavior.
 - `vllm_omni/deploy/qwen3_omni_moe.yaml`: Default Qwen3-Omni deployment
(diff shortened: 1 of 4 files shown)


docs/design/feature/omni_async_output_materialization.md
```diff
@@ -0,0 +1,332 @@
...
+  <img
+    alt="Decode step gap before and after async output materialization"
+    src="../../source/architecture/qwen3-omni-async-output-step-gap.svg"
+    width="100%"
+  >
+</p>
+
+<p align="center">
+  <em>
+    Figure 1: Moving payload construction off the decode path reduces the
```
hsliuustc0106, line 80: this svg need some polishment to illustrate what's inside each block, e.g., which stage and what operations...

docs/design/feature/omni_async_output_materialization.md
```diff
@@ -0,0 +1,332 @@
...
+postprocess runs eagerly because the next decode step needs its updated state,
+while codec payload construction runs asynchronously. Its Code2Wav stage also
+uses the normal generation-stage path.
+
+## Enabling the Feature
+
+There is no separate `async_omni_output` command-line or YAML option. The
+feature is selected automatically for model stages that opt in when the runtime
+conditions are safe. For the supported Qwen3-Omni and Qwen3-TTS pipelines,
+using the bundled deployment profile is enough to turn it on.
```
hsliuustc0106: Please make the platform scope explicit. A bundled profile is not sufficient on NPU: `NPUARModelRunner.sample_tokens()` fully constructs `OmniModelRunnerOutput` and calls `get_omni_connector_output()` before wrapping it in `AsyncGPUModelRunnerOutput`, so Omni materialization is still synchronous. If this page describes only CUDA/ROCm, say so here and in the compatibility table; if NPU is intended, runtime support/guards are missing. Scoping the docs is the smaller fix for this PR.

docs/design/feature/omni_async_output_materialization.md
```diff
@@ -0,0 +1,332 @@
...
+The next scheduler step can mutate runner state while the background builder is
+still active. The builder therefore consumes snapshots rather than live fields.
+The snapshotted state includes:
+
+- scheduler token counts and speculative-token metadata
+- request IDs and request-to-batch-index mappings
+- sampled token IDs, logprobs, and prompt logprobs
+- query start locations and scheduled-token spans
+- hidden states and multimodal output tensors
+- KV, encoder-cache, and Omni connector output state
```
hsliuustc0106: `Omni connector output state` is not snapshotted by the current implementation. `_build_omni_model_runner_output_from_snapshot()` calls `get_omni_connector_output()` in the background, and that method drains live per-step signal sets; the builder also reads `self.requests` during full-payload accumulation. Please remove it from this list and document these live-state ownership/synchronization requirements, or move the connector drain into a synchronous, step-owned snapshot.

fake0fan:
Updated based on the reviews. Please check again. @hsliuustc0106

hsliuustc0106 (approved):
lgtm, @amy-why-3459 PTAL

amy-why-3459:
### 1. Hardware Compatibility Scope
- The documentation primarily describes the feature for NVIDIA CUDA/ROCm platforms, but support for NPUs is unclear. 
- The current implementation does not include runtime guards for NPU platforms. For example, `NPUARModelRunner.sample_tokens()` constructs outputs synchronously, which contradicts the goal of asynchronous materialization.
- **Recommendation**: 
  - Clearly document the platform compatibility (e.g., CUDA/ROCm vs. NPU).
  - If NPU support is planned, implement and document the necessary runtime guards for the feature and adjust the compatibility tables accordingly.

---

### 2. Synchronization of Live-State Data
- The `Omni connector output state` is currently not snapshotted. It relies on `get_omni_connector_output()` being called in the background, which drains live-state signals and uses `self.requests` during full-payload accumulation.
- These live-state dependencies and ownership are not documented, which might lead to incorrect usage.
- **Recommendation**: 
  - Remove "Omni connector output state" from the snapshot list if the current implementation does not support snapshotting it.
  - Alternatively, document the synchronization and ownership requirements to avoid ambiguities.

---

### 3. Dependency Conflicts
- The documentation build process is blocked because of a dependency conflict. Specifically:
  - `qwen-asr==0.0.6` pins `transformers==4.57.6`.
  - However, the project requires `transformers>=5.5.3`, resulting in an unresolved dependency.
- **Recommendation**: 
  - Highlight this conflict in the PR comments or raise a dedicated issue explaining the problem.
  - Provide contributors with a workaround to resolve the dependency conflict locally, like using a specific pip command, until a formal fix is available.

---

### 4. Polishing Figures
- The SVG illustration for the Talker decode-step gap reduction is helpful but has room for improvement. Reviewers pointed out that the internal stages and operations within each block are unclear.
- **Recommendation**: 
  - Add detailed labels or annotations indicating the operations or processes within each block to make the diagram more informative and easier to understand.

---

### 5. Documentation Validation with MkDocs
- The dependency conflict currently blocks the validation step: `mkdocs build --strict`.
- **Recommendation**: Add documentation or comments about addressing this issue in the validation workflow, ensuring contributors understand how to resolve the conflict and successfully validate their changes.

---

These comments are intended to address the raised concerns and improve the clarity, functionality, and maintainability of the documentation. Let me know if you need further assistance or have additional feedback!

fake0fan:
Thanks, @amy-why-3459. I addressed and revalidated the points in your review summary on the latest rebased head (`529ac8fa`):

1. **Hardware scope:** The Overview now has an explicit **Platform scope** note limiting this design to the CUDA/ROCm `GPUARModelRunner` path. The compatibility table and fallback text also state that Ascend `NPUARModelRunner` still materializes `OmniModelRunnerOutput` and drains connector output synchronously; XPU/MUSA are not currently covered or validated.
2. **Live-state synchronization:** The snapshot list no longer claims that Omni connector state is snapshotted. The new **Live Runner State and Ownership** section documents the live `self.requests` lookup, full-payload accumulator ownership, connector signal drain, connector `_lock`, single-consumer requirement, and `get_output()` completion/error boundary.
3. **Dependency conflict / local workaround:** `qwen-asr` belongs to the optional `forced-aligner` extra. Project-wide `uv run` resolution attempts to co-resolve that extra with `transformers>=5.5.3`, but the docs build does not require it. A fresh docs environment avoids that conflict:

   ```bash
   uv venv --python 3.12 --seed
   source .venv/bin/activate
   uv pip install -e ".[docs]"
   mkdocs build --strict
   ```

   I confirmed that `uv pip install -e ".[docs]"` resolves without `qwen-asr`, and a clean-worktree `mkdocs build --strict` passes.
4. **Figure detail:** The SVG now labels the Talker stage and `forward -> sample` work, the synchronous/background Omni output paths, D2H, per-request payload construction, full-payload accumulation, connector drain, and the operations that remain inline at the ~41 us boundary.
5. **Validation:** Pre-commit, SVG rendering, and the clean-worktree strict MkDocs build all pass.

The branch is also rebased onto current `origin/main` (`a4ea67a2`).

docs/design/feature/omni_async_output_materialization.md
```diff
@@ -0,0 +1,364 @@
...
+    This design currently applies to the CUDA and ROCm AR runners, which use
+    `GPUARModelRunner`. Ascend NPU uses a separate `NPUARModelRunner` path that
+    still constructs `OmniModelRunnerOutput` and drains connector output
+    synchronously before wrapping the result as an asynchronous runner output.
+    XPU and MUSA are not currently covered or validated by this design.
+
+<p align="center">
+  <img
+    alt="Decode step gap before and after async output materialization"
+    src="../../source/architecture/qwen3-omni-async-output-step-gap.svg"
```
hsliuustc0106: With MkDocs directory URLs, this resolves to `/design/source/...`, while the asset is emitted at `/source/architecture/...`, so Figure 1 404s. Please use `../../../source/architecture/qwen3-omni-async-output-step-gap.svg` so it remains safe under the `/vllm-omni/` base path.

docs/design/feature/omni_async_output_materialization.md
```diff
@@ -0,0 +1,364 @@
...
+    output construction. Enabling async chunk alone cannot activate the feature
+    for a model that has not opted into async Omni output.
+
+## Compatibility and Fallbacks
+
+The async output path is used only when all of the following conditions hold:
+
+| Requirement | Reason |
+|---|---|
+| The platform uses the CUDA or ROCm `GPUARModelRunner` path | Ascend NPU materializes the Omni output synchronously; XPU and MUSA are not covered or validated by this design |
```
hsliuustc0106: This is not an enforced platform condition: `_should_use_async_omni_output()` has no platform check, XPU inherits `GPUARModelRunner`, and MUSA selects `GPUARWorker`. They can take the background path when the remaining conditions pass. Please describe CUDA/ROCm as validation scope only, or add a guard; otherwise the fallback promise is incorrect.

docs/source/architecture/qwen3-omni-async-output-step-gap.svg
```diff
@@ -0,0 +1,150 @@
...
+
+  <rect x="540" y="226" width="235" height="54" rx="5" class="talker"/>
+  <text x="658" y="246" text-anchor="middle" class="block-title">TALKER STAGE &#183; STEP N+1</text>
+  <text x="658" y="267" text-anchor="middle" class="block-op">forward &#8594; sample</text>
+
+  <path d="M 535 280 L 548 298" class="handoff"/>
+  <rect x="540" y="298" width="360" height="58" rx="5" class="async-output"/>
+  <text x="720" y="315" text-anchor="middle" class="block-title">BACKGROUND OMNI OUTPUT &#183; STEP N</text>
+  <text x="720" y="334" text-anchor="middle" class="block-op">wait for D2H &#8594; per-request slice / payload build</text>
+  <text x="720" y="349" text-anchor="middle" class="block-op">full-payload accumulate &#8594; connector drain</text>
```
hsliuustc0106: The enabled async path cannot perform full-payload accumulation: async output requires `async_chunk`, while `should_accumulate_full_payload_output()` returns false whenever `async_chunk` is enabled. Please replace this with the streaming payload handoff, or label it as synchronous-only.

docs/design/feature/omni_async_output_materialization.md
```diff
@@ -0,0 +1,364 @@
...
+```yaml
+async_chunk: true
+
+stages:
+  - stage_id: 0  # Thinker
+    enable_prefix_caching: false
+    # async_scheduling defaults to true for AR stages
+
+  - stage_id: 1  # Talker
+    enable_prefix_caching: false
```
hsliuustc0106, line 269: could this work with prefix cache?

docs/design/feature/omni_async_output_materialization.md
fake0fan: Fixed in `50126edf`. The image source now uses `../../../source/architecture/qwen3-omni-async-output-step-gap.svg`. A clean-worktree `mkdocs build --strict` passed, and I also verified that the SVG is emitted under `site/source/architecture/` and that the rendered page contains the corrected path.

docs/design/feature/omni_async_output_materialization.md
fake0fan: Fixed in `50126edf`. I removed CUDA/ROCm from the runtime-requirements table and now describe them only as the validated scope. The text explicitly notes that XPU inherits `GPUARModelRunner` and MUSA selects `GPUARWorker`, so either may enter this path when the actual runtime guards pass, although this optimization is not yet validated on those backends. Ascend NPU remains documented separately as the synchronous `NPUARModelRunner` path.

docs/source/architecture/qwen3-omni-async-output-step-gap.svg
fake0fan: Fixed in `50126edf`. I replaced full-payload accumulation in both SVG lanes and the accessibility description with the async-chunk streaming payload handoff and connector-signal drain. The architecture section now explains that the shared full-payload branch reads `self.requests` but is unreachable for this feature because async Omni output requires `async_chunk`, which makes `should_accumulate_full_payload_output()` return `False`; the enabled path partitions each step into streaming inter-stage and client handoffs.

docs/design/feature/omni_async_output_materialization.md
fake0fan: Not currently. I made this explicit in `50126edf`: `_should_use_async_omni_output()` returns `False` whenever `self.omni_prefix_cache` is present, so enabling Omni prefix caching selects synchronous output materialization. Supporting both requires snapshotting or otherwise synchronizing prefix-cache merge and update state before the background builder can consume it safely.

hsliuustc0106: OK, please mark it and we will make them compatible in follow-up PRs

PR #5613, opened 2026-07-30
[Bugfix][Full Duplex] Make runtime_control payloads JSON serializable

### LINKED ISSUE #5612: [Bug]: Duplex session.created fails — DuplexFence is not JSON serializable in runtime_control
### Your current environment

<details>

```text
vllm-omni : main @ 1b718f7b
vllm      : 0.26.0 (base image vllm/vllm-openai:v0.26.0, per docker/Dockerfile.cuda)
model     : Qwen/Qwen3-Omni-30B-A3B-Instruct (3-stage: thinker -> talker -> code2wav)
GPUs      : 2x H100 80GB
endpoint  : WS /v1/duplex
```

The minimal reproduction below needs **no GPU, no model and no deploy config** — it is
pure Python against `main`.

</details>

### 🐛 Describe the bug

`_redact_runtime_control_result` returns non-`dict` / non-`list` values unchanged, but
duplex control results carry typed values straight from the engine: `fence` and
`accepted_fence` are `DuplexFence` **dataclass instances**, inserted by
`DuplexControlClient.execute`:

https://github.com/vllm-project/vllm-omni/blob/1b718f7b/vllm_omni/experimental/fullduplex/engine/duplex_control_client.py#L74
https://github.com/vllm-project/vllm-omni/blob/1b718f7b/vllm_omni/experimental/fullduplex/engine/duplex_control_client.py#L81

That redacted value is then put on the wire — `session.created` sets `runtime_control`
from it:

https://github.com/vllm-project/vllm-omni/blob/1b718f7b/vllm_omni/experimental/fullduplex/openai/session_runner.py#L695

`WebSocket.send_json` therefore raises and **the duplex session dies during the
handshake**, before any audio can be appended. The client receives:

```json
{"type": "error", "code": "internal_error",
 "error": "Object of type DuplexFence is not JSON serializable", "server_event_seq": 1}
```

Server side:

```
ERROR [session_runner.py:1861] Duplex session failed: Object of type DuplexFence is not JSON serializable
  File ".../openai/session_runner.py", line 696, in handle_session
    await emit_event(created_payload)
  File ".../openai/session_runner.py", line 76, in attachment_send
    await websocket.send_json(payload)
  File ".../starlette/websockets.py", line 174, in send_json
    text = json.dumps(data, separators=(",", ":"), ensure_ascii=False)
TypeError: Object of type DuplexFence is not JSON serializable
```

#### Scope

This is not model specific. The `runtime_control` key is only attached when
`open_result` is a `dict`, which happens for any adapter whose capabilities declare
`implementation_level="model_native_duplex"` (`runtime_bridge.py:88`).

I hit it while bringing up a new native duplex adapter for Qwen3-Omni. **I have not
tested MiniCPM-o 4.5**, so I can't say whether it is affected — but I'd guess not in
practice, because its validated path is the OpenAI Realtime projection
(`examples/online_serving/minicpmo/realtime_duplex_demo.py` defaults to
`/v1/realtime?duplex=1`) rather than the raw `/v1/duplex` protocol. That would make
`/v1/duplex` the less-exercised path and explain why this survived. Maintainers can
confirm that far faster than I can.

Worth noting for other model authors: a model that already owns `/v1/realtime` (as
Qwen3-Omni does, for its existing half-duplex endpoint) has no choice but to use
`/v1/duplex`, so it hits this immediately.

### 🔁 Reproduction

Pure Python on `main` — no GPU, no model, no adapter:

```python
import json
from vllm_omni.experimental.fullduplex.engine.messages import DuplexFence
from vllm_omni.experimental.fullduplex.openai.runtime_bridge import NativeRuntimeBridgeMixin

# Shape produced by DuplexControlClient.execute() (duplex_control_client.py:74,81)
control_result = {
    "ok": True,
    "operation": "open",
    "session_id": "sess-1",
    "fence": DuplexFence("sess-1", epoch=0, turn_id=0, response_seq=0, incarnation=0),
    "accepted_fence": DuplexFence("sess-1", epoch=0, turn_id=0, response_seq=0, incarnation=0),
    "stage_results": [{"stage_id": -1, "replica_id": -1, "result": {"supported": True}}],
}

redacted = NativeRuntimeBridgeMixin._redact_runtime_control_result(control_result)
print("redacted['fence'] type:", type(redacted["fence"]).__name__)

# session_runner.py:695 -> created_payload["runtime_control"], then send_json()
json.dumps({"type": "session.created", "runtime_control": redacted})
```

Actual output on `main`:

```
redacted['fence'] type: DuplexFence
TypeError: Object of type DuplexFence is not JSON serializable
```

Expected: `redacted["fence"]` is a plain `dict` and the payload serializes.

To reproduce against a live server: stand up any 3-stage duplex deployment whose
adapter reports `implementation_level="model_native_duplex"`, connect to `/v1/duplex`,
and send `session.create`. The session errors out before `session.created` is
delivered.

### 💡 Proposed fix

Make the redactor's output wire-safe by construction. It already exists to produce a
client-visible payload, so it seems like the right place to guarantee JSON-safety
rather than relying on every producer to pre-serialize.

```diff
--- a/vllm_omni/experimental/fullduplex/openai/runtime_bridge.py
+++ b/vllm_omni/experimental/fullduplex/openai/runtime_bridge.py
@@
 import asyncio
 import base64
 import inspect
 from collections.abc import Mapping
+from dataclasses import asdict, is_dataclass
+from enum import Enum
@@ def _redact_runtime_control_result(cls, value: object) -> object:
             return redacted
         if isinstance(value, list | tuple):
             return [cls._redact_runtime_control_result(item) for item in value]
+        if is_dataclass(value) and not isinstance(value, type):
+            # Control results carry typed values straight from the engine --
+            # notably `fence` / `accepted_fence` are `DuplexFence` instances
+            # (duplex_control_client.py:74,81). This payload goes on the wire
+            # via `session.created.runtime_control` (session_runner.py:695),
+            # where `send_json` would raise TypeError.
+            return {key: cls._redact_runtime_control_result(child) for key, child in asdict(value).items()}
+        if isinstance(value, Enum):
+            return value.value
         return value
```

With this applied, the reproduction above prints `redacted['fence'] type: dict` and
serializes cleanly. I confirmed both the failure and the fix by running that snippet
in a vLLM 0.26.0 container, and separately confirmed the live `/v1/duplex` handshake
completes afterwards.

The `Enum` branch is defensive rather than something I hit — enums are equally
unserializable and control results carry several (`mode`, `session_mode`), so it
seemed worth closing at the same time. Happy to drop it if you'd rather keep the
change minimal.

#### Alternatives considered

1. **Serialize at the producer** (`DuplexControlClient.execute` emits fence dicts).
   Arguably cleaner layering, but `accepted_fence` is consumed as a typed
   `DuplexFence` elsewhere (`duplex_control_client.py:41-42`), so this would need a
   separate wire-shaped copy.
2. **Drop `fence` / `accepted_fence` in the redactor**, as is already done for
   `stage_handoff` and friends. Simplest, but the fence values look genuinely useful
   to clients for debugging epoch/turn behaviour, so dropping them seemed worse than
   encoding them.

Happy to open a PR with whichever direction you prefer.

I'm working on a Qwen3-Omni native duplex adapter against #3745 and will report
anything else I hit on the way.

TheCodeWrangler:
Fixes #5612.

## The bug

`_redact_runtime_control_result` returned non-`dict`, non-`list` values unchanged, but
duplex control results carry typed values straight from the engine: `fence` and
`accepted_fence` are `DuplexFence` **dataclass instances** added by
`DuplexControlClient.execute`
([duplex_control_client.py#L74](https://github.com/vllm-project/vllm-omni/blob/1b718f7b/vllm_omni/experimental/fullduplex/engine/duplex_control_client.py#L74),
[#L81](https://github.com/vllm-project/vllm-omni/blob/1b718f7b/vllm_omni/experimental/fullduplex/engine/duplex_control_client.py#L81)).

That redacted value goes on the wire — `session.created` sets `runtime_control` from it
([session_runner.py#L695](https://github.com/vllm-project/vllm-omni/blob/1b718f7b/vllm_omni/experimental/fullduplex/openai/session_runner.py#L695))
— so `WebSocket.send_json` raised:

```
TypeError: Object of type DuplexFence is not JSON serializable
```

and the duplex session **died during the handshake**, before any audio could be
appended. The client received only `{"type": "error", "code": "internal_error"}`.

Not model specific: `runtime_control` is attached whenever the open result is a `dict`,
which is the case for any adapter declaring
`implementation_level="model_native_duplex"` (`runtime_bridge.py:88`).

## The fix

The redactor exists to produce a client-visible payload, so it seemed like the right
place to guarantee JSON-safety rather than requiring every producer to pre-serialize.
Dataclasses are converted with `asdict` and recursed; enums are reduced to their values.

## Why the conversion is safe

I checked every consumer of the redacted value before changing its type:

| call site | consumer |
|---|---|
| `runtime_bridge.py:84` | `send_json` |
| `runtime_bridge.py:487` | `logger.warning` |
| `runtime_bridge.py:565` | `send_json` |
| `runtime_bridge.py:1296` | `send_json` |
| `session_runner.py:695` | `created_payload` → `send_json` |

All wire or log; nothing consumes it as typed data. The one programmatic reader of
`accepted_fence` — `DuplexControlRequestError.__init__`
([duplex_control_client.py#L41-L42](https://github.com/vllm-project/vllm-omni/blob/1b718f7b/vllm_omni/experimental/fullduplex/engine/duplex_control_client.py#L41-L42))
— reads the **raw** result dict, which this PR does not touch.

The `Enum` branch is defensive rather than something I hit; control results carry
several (`mode`, `session_mode`) and they are equally unserializable. Happy to drop it
if you'd prefer a more minimal change.

## Testing

New `tests/e2e/features/fullduplex/openai/test_runtime_control_redaction.py` covers
top-level fences, fences nested inside `stage_results`, enum reduction, preservation of
the existing heavy-payload key dropping (`stage_handoff`, `tts_handoff`, …), plain-value
pass-through, and that dataclass *types* are not mistaken for instances.

Verified the tests actually catch the bug rather than merely passing. Full
`tests/e2e/features/fullduplex` suite, same environment, toggling only this patch:

```
base (this patch's source change reverted):   3 failed, 170 passed
with this patch:                              0 failed, 173 passed
```

The 3 failures on the base are exactly the new tests in this PR, each on the
`DuplexFence` serialization path:

```
test_redacted_control_result_is_json_serializable
test_fences_are_converted_to_plain_dicts
test_nested_dataclasses_are_converted
```

Also confirmed end to end: with this applied, a live Qwen3-Omni 3-stage duplex server
completes the `/v1/duplex` handshake and delivers `session.created`, where previously
the session errored out during the handshake.

## Alternatives considered

1. **Serialize at the producer** (`DuplexControlClient.execute` emits fence dicts) —
   arguably cleaner layering, but `accepted_fence` is consumed as a typed `DuplexFence`
   in the error path, so it would need a separate wire-shaped copy.
2. **Drop `fence` / `accepted_fence` in the redactor**, as is already done for
   `stage_handoff` and friends — simplest, but those values look genuinely useful to
   clients for debugging epoch/turn behaviour.

Happy to switch to either if maintainers prefer.

### DIFF
--- a/tests/e2e/features/fullduplex/openai/test_runtime_control_redaction.py
+++ b/tests/e2e/features/fullduplex/openai/test_runtime_control_redaction.py
@@ -1,0 +1,115 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+
+"""Regression tests for duplex ``runtime_control`` payload redaction.
+
+The redacted result is emitted on the WebSocket (for example
+``session.created.runtime_control``), so it must always survive
+``json.dumps``. Control results carry typed values straight from the engine --
+``fence`` and ``accepted_fence`` are ``DuplexFence`` instances added by
+``DuplexControlClient.execute`` -- which previously reached
+``WebSocket.send_json`` unconverted and killed the session at handshake.
+"""
+
+from __future__ import annotations
+
+import json
+from dataclasses import dataclass
+from enum import Enum
+
+import pytest
+
+from vllm_omni.experimental.fullduplex.engine.messages import DuplexFence
+from vllm_omni.experimental.fullduplex.openai.runtime_bridge import NativeRuntimeBridgeMixin
+
+pytestmark = [pytest.mark.core_model, pytest.mark.cpu]
+
+_redact = NativeRuntimeBridgeMixin._redact_runtime_control_result
+
+
+def _control_result() -> dict[str, object]:
+    """The shape ``DuplexControlClient.execute`` returns for an open."""
+    return {
+        "ok": True,
+        "operation": "open",
+        "session_id": "sess-1",
+        "fence": DuplexFence("sess-1", epoch=0, turn_id=0, response_seq=0, incarnation=0),
+        "accepted_fence": DuplexFence("sess-1", epoch=1, turn_id=2, response_seq=3, incarnation=4),
+        "stage_results": [{"stage_id": -1, "replica_id": -1, "result": {"supported": True}}],
+    }
+
+
+def test_redacted_control_result_is_json_serializable() -> None:
+    """The whole payload must serialize the way send_json does."""
+    redacted = _redact(_control_result())
+    # Must not raise.
+    json.dumps({"type": "session.created", "runtime_control": redacted})
+
+
+def test_fences_are_converted_to_plain_dicts() -> None:
+    redacted = _redact(_control_result())
+    assert redacted["fence"] == {
+        "session_id": "sess-1",
+        "epoch": 0,
+        "turn_id": 0,
+        "response_seq": 0,
+        "incarnation": 0,
+    }
+    # Field values are preserved, not just the type.
+    assert redacted["accepted_fence"]["epoch"] == 1
+    assert redacted["accepted_fence"]["turn_id"] == 2
+    assert redacted["accepted_fence"]["response_seq"] == 3
+    assert redacted["accepted_fence"]["incarnation"] == 4
+
+
+def test_nested_dataclasses_are_converted() -> None:
+    """Fences can appear inside stage results, not only at the top level."""
+    redacted = _redact(
+        {
+            "stage_results": [
+                {"stage_id": 0, "result": {"fence": DuplexFence("sess-2", epoch=7)}},
+            ]
+        }
+    )
+    json.dumps(redacted)
+    assert redacted["stage_results"][0]["result"]["fence"]["epoch"] == 7
+
+
+def test_enums_are_reduced_to_values() -> None:
+    class _Mode(str, Enum):
+        APPEND = "append_audio_chunk"
+
+    redacted = _redact({"mode": _Mode.APPEND})
+    json.dumps(redacted)
+    assert redacted["mode"] == "append_audio_chunk"
+
+
+def test_redaction_still_drops_heavy_payload_keys() -> None:
+    """The pre-existing contract must be preserved."""
+    redacted = _redact(
+        {
+            "ok": True,
+            "stage_handoff": object(),
+            "tts_handoff": object(),
+            "omni_payload": object(),
+            "tts_hidden_states": object(),
+            "tts_token_ids": object(),
+            "traceback": "long trace",
+        }
+    )
+    assert redacted == {"ok": True}
+
+
+def test_plain_values_pass_through_unchanged() -> None:
+    payload = {"ok": True, "count": 3, "name": "x", "missing": None, "items": [1, "two"]}
+    assert _redact(payload) == payload
+
+
+def test_dataclass_types_are_not_treated_as_instances() -> None:
+    """A class object is not a dataclass instance and must not be asdict'd."""
+
+    @dataclass
+    class _Shape:
+        x: int = 1
+
+    assert _redact(_Shape) is _Shape
--- a/vllm_omni/experimental/fullduplex/openai/runtime_bridge.py
+++ b/vllm_omni/experimental/fullduplex/openai/runtime_bridge.py
@@ -4,6 +4,8 @@ import asyncio
 import base64
 import inspect
 from collections.abc import Mapping
+from dataclasses import asdict, is_dataclass
+from enum import Enum
 
 import numpy as np
 from vllm.logger import init_logger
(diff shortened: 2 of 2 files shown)


Gaohan123 (approved):
LGTM. Thanks

PR #5615, opened 2026-07-30
[Bugfix] Fix Wan2.2 omitted guidance scale resolution

### LINKED ISSUE #5415: [Bug]: Inaccurate x2v benchmark results
### Your current environment

<details>

```text
Your output of `python collect_env.py` here
```

</details>

### Your code version

<details>

```text

```
</details>
<details>

```text

```
</details>

### 🐛 Describe the bug

The `diffusion_benchmark_serving.py` script does not enable the guidance scale parameter. The default guidance scale parameter causes the benchmark to only run positive tests during the low-noise phase, omitting negative tests, resulting in inaccurate latency measurements.

lishunyang12:
## Purpose

Resolves #5415.

Wan2.2 requests that omit `guidance_scale_2` could resolve the two denoising stages to different scales. The low-noise stage treated the internally filled fallback as user-provided and could skip its negative CFG pass.

This change resolves both T2V and I2V guidance scales from the explicit `*_provided` flags and adds coverage for omitted, explicit single-scale, and explicit dual-scale requests.

cc @xiongjingwu @ZacheryAU @Gaohan123

## Test Plan

**vLLM Version:** 0.26.0

**vLLM-Omni Commit:** `d1443f52`

```bash
python -m pytest -q \
  tests/diffusion/test_diffusion_request.py \
  tests/diffusion/models/wan2_2/test_wan22_pipeline_diffuse.py \
  tests/diffusion/models/wan2_2/test_wan22_i2v_pipeline.py
```

The regression test exercises a real `OmniDiffusionRequest`; on the parent commit, the omitted-scale case fails with `assert 1.0 == 4.0`.

## Test Result

```text
13 passed
```

The full Wan2.2 CPU subset also passed (`35 passed`), excluding one collection-only environment failure caused by missing `pytest-mock`. File-scoped pre-commit hooks and `git diff --check` pass.


lishunyang12:
Validated the default-guidance fix end-to-end with
vLLM 0.26.0, PyTorch 2.11.0+cu130, and CUDA 13.0.

Wan2.2 T2V settings: 256x256, 9 frames, 40 steps, seed 0, latent output.

| revision / arguments | latent SHA256 | result |
|---|---|---|
| parent, scales omitted | `d86f9f0cc0dde009335d8efc070c0dde386ee042e937d9a32906a3494caaca62` | identical to explicit `4.0/1.0` |
| parent, explicit `4.0/1.0` | `d86f9f0cc0dde009335d8efc070c0dde386ee042e937d9a32906a3494caaca62` | old behavior |
| PR, scales omitted | `91210b61259ed6329e96107ab798c7277300827f350a036250057bad90ad4fcd` | identical to explicit `4.0/4.0` |
| PR, explicit `4.0/4.0` | `91210b61259ed6329e96107ab798c7277300827f350a036250057bad90ad4fcd` | intended behavior |

All four runs produced finite tensors with shape `(1, 3, 32, 32, 16)`.
This confirms the regression and that the patch changes only the omitted
low-noise guidance default to the documented value.

vllm_omni/diffusion/models/wan2_2/pipeline_wan2_2.py
```diff
@@ -84,6 +84,24 @@ def resolve_wan_flow_shift(req: OmniDiffusionRequest, od_config: OmniDiffusionCo
         raise ValueError(f"Invalid flow_shift={raw_flow_shift!r}. flow_shift must be a float.") from exc
 
 
+def resolve_wan_guidance_scales(
+    sampling_params: OmniDiffusionSamplingParams,
+    default_guidance_scale: float,
+) -> tuple[float, float]:
+    guidance_scale = (
+        sampling_params.guidance_scale if sampling_params.guidance_scale_provided else default_guidance_scale
```
hsliuustc0106, line 92: `guidance_scale_provided` is false for an explicit `0.0` because `OmniDiffusionRequest.__post_init__` uses truthiness and rewrites zero to `1.0`. This therefore resolves an API-valid zero to Wan’s default (`4.0`, or `5.0` for I2V) and unexpectedly enables CFG. Please preserve omission provenance with `None`/a sentinel and add a regression test through a real `OmniDiffusionRequest`.

Gaohan123 (approved):
LGTM. Thanks

PR #5616, opened 2026-07-30
[Bugfix] Route ngram_gpu through sampled GPU tokens

### LINKED ISSUE #5348: [Bug]: ngram_gpu speculative decoding on the Qwen3-Omni thinker stage crashes the stage engine at the first inference (`expand(CUDABoolType{[1, 1]}, size=[0])` in `update_token_ids_ngram`)
## Summary

Enabling draft-free GPU n-gram speculative decoding (`method: ngram_gpu`) on
stage 0 (thinker) of the Qwen3-Omni-30B pipeline kills the stage engine on
the very first request. The engine boots cleanly (weights load, KV cache
created, cudagraphs captured); on the first generation, at the first decode
step-with-draft, vLLM's `NgramProposerGPU.update_token_ids_ngram` raises

```
RuntimeError: expand(CUDABoolType{[1, 1]}, size=[0]): the number of sizes
provided (1) must be greater or equal to the number of dimensions in the
tensor (2)
```

The `StageEngineCoreProc` treats this as fatal, the whole `AsyncOmni` engine
dies (`OmniEngineDeadError` surfaced to the caller), and every replica
restart reproduces it identically (3/3 attempts in our run, same first-step
crash each time). 100% deterministic on our stack.

The crash is reached through vllm-omni's AR model runner
(`vllm_omni/worker/gpu_ar_model_runner.py::sample_tokens` →
`propose_draft_token_ids`), which delegates into vLLM's
`gpu_model_runner.propose_draft_token_ids` → `drafter.update_token_ids_ngram`.
The same `ngram_gpu` config on a plain vLLM text model does not crash for
us, so this looks like an integration contract mismatch between the omni AR
runner and the GPU n-gram drafter rather than a bug in the drafter alone.

## Environment

- vllm-omni 0.24.0 (PyPI), vLLM 0.24.0
- Model: Qwen/Qwen3-Omni-30B-A3B-Instruct, bf16, 3-stage pipeline
  (thinker / talker / code2wav), all stages on ONE GPU
- GPU: 1x NVIDIA RTX PRO 6000 Blackwell (96 GB, sm_120), CUDA 12.9 stack
- Python 3.11, Ray 2.56 (single-node Ray Serve deployment of `AsyncOmni`
  via the deploy-config path)
- Stage 0 runs with `async_scheduling: true` (the 0.24 default for LLM_AR
  stages; the crash traceback goes through `step_with_batch_queue`),
  `enforce_eager: false`, `enable_chunked_prefill: true`,
  `max_num_seqs: 4`, `max_model_len: 32768`

## Reproduction

Deploy-config delta vs the bundled `qwen3_omni_moe.yaml` — one added block
on stage 0 (everything else unchanged; our full yaml is a tuned variant of
the upstream example):

```yaml
stages:
  - stage_id: 0
    # ... (thinker stage as in vllm_omni/deploy/qwen3_omni_moe.yaml) ...
    speculative_config:
      method: ngram_gpu
      num_speculative_tokens: 3
      prompt_lookup_min: 8
      prompt_lookup_max: 16
```

The config is accepted and applied — boot log confirms:

```
Initializing a V1 LLM engine (v0.24.0) with config:
model='Qwen/Qwen3-Omni-30B-A3B-Instruct',
speculative_config=SpeculativeConfig(method='ngram_gpu',
model='Qwen/Qwen3-Omni-30B-A3B-Instruct', num_spec_tokens=3), ...
```

Then submit ONE generation request (ours: 31 prompt tokens + one ~100-frame
audio input, `output_modalities=text,audio`, sampling temperature 0.4 /
top_k 1 / max_tokens 256). The stage-0 engine crashes within seconds, before
yielding any output (`yields=0 text_chunks=0 audio_chunks=0`).

## Full stage-0 worker traceback

```
ERROR 07-22 18:23:56 [multiproc_executor.py:1000] WorkerProc hit an exception.
ERROR 07-22 18:23:56 [multiproc_executor.py:1000] Traceback (most recent call last):
ERROR 07-22 18:23:56 [multiproc_executor.py:1000]   File ".../vllm/v1/executor/multiproc_executor.py", line 992, in worker_busy_loop
ERROR 07-22 18:23:56 [multiproc_executor.py:1000]     output = func(*args, **kwargs)
ERROR 07-22 18:23:56 [multiproc_executor.py:1000]   File ".../torch/utils/_contextlib.py", line 124, in decorate_context
ERROR 07-22 18:23:56 [multiproc_executor.py:1000]     return func(*args, **kwargs)
ERROR 07-22 18:23:56 [multiproc_executor.py:1000]   File ".../vllm/v1/worker/gpu_worker.py", line 833, in sample_tokens
ERROR 07-22 18:23:56 [multiproc_executor.py:1000]     return self.model_runner.sample_tokens(grammar_output)
ERROR 07-22 18:23:56 [multiproc_executor.py:1000]   File ".../torch/utils/_contextlib.py", line 124, in decorate_context
ERROR 07-22 18:23:56 [multiproc_executor.py:1000]     return func(*args, **kwargs)
ERROR 07-22 18:23:56 [multiproc_executor.py:1000]   File ".../vllm_omni/worker/gpu_ar_model_runner.py", line 1975, in sample_tokens
ERROR 07-22 18:23:56 [multiproc_executor.py:1000]     propose_draft_token_ids(valid_sampled_token_ids)
ERROR 07-22 18:23:56 [multiproc_executor.py:1000]   File ".../vllm_omni/worker/gpu_ar_model_runner.py", line 1908, in propose_draft_token_ids
ERROR 07-22 18:23:56 [multiproc_executor.py:1000]     self._draft_token_ids = self.propose_draft_token_ids(
ERROR 07-22 18:23:56 [multiproc_executor.py:1000]   File ".../vllm/v1/worker/gpu_model_runner.py", line 4908, in propose_draft_token_ids
ERROR 07-22 18:23:56 [multiproc_executor.py:1000]     ) = self.drafter.update_token_ids_ngram(
ERROR 07-22 18:23:56 [multiproc_executor.py:1000]   File ".../vllm/v1/spec_decode/ngram_proposer_gpu.py", line 438, in update_token_ids_ngram
ERROR 07-22 18:23:56 [multiproc_executor.py:1000]     valid_sampled_token_ids_gpu.masked_fill_(discard_mask_expanded, -1)
ERROR 07-22 18:23:56 [multiproc_executor.py:1000] RuntimeError: expand(CUDABoolType{[1, 1]}, size=[0]): the number of sizes provided (1) must be greater or equal to the number of dimensions in the tensor (2)
```

Escalation in the stage engine core (same timestamp):

```
(StageEngineCoreProc_stage0_replica0) ERROR [stage_engine_core_proc.py:178] StageEngineCoreProc encountered a fatal error.
  File ".../vllm_omni/engine/stage_engine_core_proc.py", line 169, in run_stage_core
    engine_core.run_busy_loop()
  File ".../vllm/v1/engine/core.py", line 1265, in run_busy_loop
    self._process_engine_step()
  File ".../vllm/v1/engine/core.py", line 1304, in _process_engine_step
    outputs, model_executed = self.step_fn()
  File ".../vllm/v1/engine/core.py", line 595, in step_with_batch_queue
    model_output = future.result()
  ...
RuntimeError: Worker failed with error 'expand(CUDABoolType{[1, 1]}, size=[0]):
the number of sizes provided (1) must be greater or equal to the number of
dimensions in the tensor (2)', please check the stack trace above for the
root cause
```

The accompanying `dump_input.py` dump shows this is literally the first
scheduler step of the first request:

```
OmniSchedulerOutput(scheduled_new_reqs=[NewRequestData(
  req_id=warmup-...-91ac549f, prompt_token_ids_len=31, ...
  mm_features=[... modality='audio' ... audio_feature_lengths: tensor(100) ...])],
  num_scheduled_tokens={warmup-...: 31}, total_num_scheduled_tokens=31,
  scheduled_spec_decode_tokens={}, num_invalid_spec_tokens=null,
  num_spec_tokens_to_schedule=3, ...)
SchedulerStats(num_running_reqs=1, num_waiting_reqs=0, step_counter=0, ...)
```

## Expected vs actual

- **Expected:** either (a) n-gram speculative decoding works on the thinker
  stage (it is a standard AR text-decode loop; draft-free n-gram lookup has
  no draft-model requirement), or (b) the configuration is rejected at
  startup with a clear "speculative decoding not supported for omni AR
  stages" error.
- **Actual:** the config is accepted, the engine boots cleanly, and the
  stage engine core dies with an opaque tensor-shape RuntimeError on the
  first inference. Every Ray Serve replica restart reproduces it (3/3),
  so the deployment can never become healthy.

## What we ruled out

- Not our infra or the rest of the config: paired control arms in the same
  benchmark wave (identical machine type, image, model, deploy yaml except
  the one lever — an unmodified baseline and an fp8 arm) booted and served
  the same first request cleanly. The `speculative_config` block is the
  only delta on the failing arm.
- Not flaky: deterministic across 3 consecutive fresh replicas (fresh
  process, fresh CUDA context each time), same first-step crash.
- Not a warmup artifact of our harness: the failing request is an ordinary
  single `generate()` (text+audio-in, text+audio-out) — the first inference
  of any kind after boot.

## Status on current main / novelty check (2026-07-23)

- The crash region is unchanged on current vLLM main: the
  `discard_mask_expanded = discard_request_mask[:num_reqs].unsqueeze(1)` ->
  `masked_fill_` sequence at `ngram_proposer_gpu.py:441-442` is byte-identical
  to v0.24.0, so 0.25 will reproduce this.
- We found no existing vLLM or vllm-omni issue/PR matching
  `update_token_ids_ngram` or this `expand(CUDABoolType{[1, 1]}, size=[0])`
  signature (nearest ngram_gpu items are unrelated: vllm#42533, PRs
  vllm#44054/#44056).
- Mechanism, as far as we traced it: a `[1, 1]` discard mask is broadcast
  against a ZERO-ROW `sampled_token_ids` tensor — the omni AR runner can hand
  the drafter an empty sampled-ids tensor on the first decode step, and
  nothing in `update_token_ids_ngram` guards `numel() == 0`.
- Workaround for omni users: `method: ngram` (the CPU proposer) takes a
  different code path and does not crash on the same pipeline.

## Suggested investigation direction (unverified hypothesis)

The failing line expands a `[1, 1]` bool discard mask to `size=[0]`, i.e.
`valid_sampled_token_ids_gpu` arrives at `update_token_ids_ngram` with 0
rows while the mask side still reflects 1 request × 1 token. That points at
the boundary in `vllm_omni/worker/gpu_ar_model_runner.py:1975`, which passes
its own `valid_sampled_token_ids` into vLLM's
`gpu_model_runner.propose_draft_token_ids`: under async scheduling
(`step_with_batch_queue`) the omni AR runner's first post-prefill
`sample_tokens` call appears to hand the drafter an empty/differently-shaped
sampled-ids tensor than vLLM's own runner would at the same point (vLLM's
runner guards / shapes this before calling the drafter). Worth checking:

1. whether `gpu_ar_model_runner.sample_tokens` should skip
   `propose_draft_token_ids` when `valid_sampled_token_ids` is empty
   (mirror of the guard in vLLM's `gpu_model_runner`), and
2. whether the omni AR runner keeps `valid_sampled_token_ids` as a CPU
   list-of-lists vs the GPU tensor layout `update_token_ids_ngram` expects
   at v0.24.0.

We have not root-caused it further; happy to run patched wheels on the same
hardware to validate a fix.

## Related context

We previously reported vllm-omni #5170 (per-stage vllm:* metrics never
report data) from the same 3-stage Qwen3-Omni production stack; this report
comes from the same deployment harness.

lishunyang12:
## Purpose

Resolves #5348.

`ngram_gpu` speculative decoding on an Omni AR stage could crash on the first request because `GPUARModelRunner.sample_tokens` invoked the proposer after async-scheduling bookkeeping and passed an empty CPU result instead of the sampler's GPU token tensor.

This change ports the vLLM 0.26 pre-bookkeeping `ngram_gpu` dispatch: it passes `sampler_output.sampled_token_ids` directly when the input fits and preserves the padded-drafter update/count path otherwise.

The fix was validated end-to-end on one NVIDIA B300 with Qwen3-Omni-30B-A3B-Instruct. The parent commit crashes the first request in `NgramProposerGPU.update_token_ids_ngram()` and ends in `EngineDeadError`; this PR completes the same request normally with 11 streamed chunks.

This fixes the immediate engine crash. End-to-end accepted-token streaming through Qwen3-Omni's downstream async-chunk path remains a separate validation item.

cc @pratapyash @tzhouam @amy-why-3459

## Test Plan

**vLLM Version:** 0.26.0

```bash
python -m pytest -q tests/worker/test_gpu_ar_model_runner.py
python -m pytest -q tests/engine/test_orchestrator_error_handling.py
```

The worker test file checks the surrounding runner behavior. The GPU regression itself is covered by the B300 end-to-end reproduction above instead of a mock-heavy unit test.

## Test Result

```text
tests/worker/test_gpu_ar_model_runner.py: 23 passed
tests/engine/test_orchestrator_error_handling.py: 7 passed
orchestrator fatal-error race stress run: 100 passed
```

File-scoped pre-commit hooks and `git diff --check` pass.


lishunyang12:
Validated the fix end-to-end on one NVIDIA B300 (SM103) with
Qwen3-Omni-30B-A3B-Instruct, vLLM 0.26.0, PyTorch 2.11.0+cu130, and CUDA 13.0.

Both runs used the same three-stage config with `ngram_gpu`,
`num_speculative_tokens=3`, asynchronous scheduling on the thinker, greedy
sampling, and a text-only request.

- Parent commit: the first request crashes `StageEngineCoreProc_stage0` in
  `NgramProposerGPU.update_token_ids_ngram()` with
  `expand(CUDABoolType{[1, 1]}, size=[0])`, followed by `EngineDeadError`.
- This PR: the request completes normally with 11 streamed chunks and the text
  `Predicts multiple tokens ahead to speed up generation.`

This reproduces the reported GPU failure and confirms that routing the sampled
tokens through the sampler's GPU tensor before speculative bookkeeping prevents
the engine crash.

vllm_omni/worker/gpu_ar_model_runner.py
```diff
@@ -2016,6 +2017,21 @@ def propose_draft_token_ids(sampled_token_ids):
                         len(self.input_batch.req_ids), self.num_spec_tokens
                     )
                     self._copy_draft_token_ids_to_cpu(scheduler_output, zeros_only=True)
+            elif spec_config.use_ngram_gpu() and not spec_config.disable_padded_drafter_batch:
+                assert isinstance(self.drafter, NgramProposerGPU)
+                sampled_token_ids = sampler_output.sampled_token_ids
+                if input_fits_in_drafter:
+                    propose_draft_token_ids(sampled_token_ids)
+                elif self.valid_sampled_token_count_event is not None:
```
hsliuustc0106, line 2019: When the input no longer fits, this branch updates the valid-token count but omits vLLM 0.26’s subsequent zeroing of `_draft_token_ids` and `_copy_draft_token_ids_to_cpu(..., zeros_only=True)`. Under async scheduling, stale optimistic drafts can survive near the model-length boundary and corrupt the next step. Please port the adjacent false-fit cleanup and test `_input_fits_in_drafter=False` with async scheduling.

Gaohan123 (approved):
LGTM. Thanks

Issue #5611, opened 2026-07-30
[Bug]: Ring attention fails on B200/Blackwell (Hopper FA3 kernel not available)


yenuo26:
### Your current environment

<details>
<summary>The output of <code>python collect_env.py</code></summary>

```text
Hardware: NVIDIA B200 (Blackwell, sm_100) - observed in vLLM-Omni CUDA nightly / B200 mirror CI (MIRROR_HW=b200).
OS / container: vLLM CI test image (vllm-ci-test-repo:$BUILDKITE_COMMIT).
Note: full collect_env.py dump was not captured on the failing agent; environment is the standard Buildkite B200 diffusion job.
```

</details>

### Your code version

<details>
<summary>The commit id or version of vllm</summary>

```text
CI pin / image-bundled vLLM (same as CUDA nightly Buildkite image at failure time)
```
</details>
<details>
<summary>The commit id or version of vllm-omni</summary>

```text
main / B200 nightly branch tip around the CUDA B200 migration work
(reproducible on current main: ring selector still prefers HAS_FA3 over arch-aware fallback)
```
</details>

### 🐛 Describe the bug

On B200 / Blackwell, ring-related diffusion attention paths crash (Hopper FA3 / unavailable SDPA kernels during warmup).

#### Reproduction steps

1. Use a B200 machine (or B200 CI image with `MIRROR_HW=b200`).
2. Run:

```bash
export VLLM_OMNI_LOGGING_LEVEL=DEBUG
pytest -s -v tests/examples/offline_inference/test_image_to_video.py::test_image_to_video[more_cli_examples_004]
```

3. Observe crash during orchestrator stage init / diffusion engine dummy warmup (see logs below).

Ring-enabled SP cases (`ring_degree > 1` / `--ring`) hit the same class of B200 attention backend failure (Hopper FA3 / unavailable kernels).

#### Logs

```text
  0%|          | 0/2 [00:00<?, ?it/s]/usr/local/lib/python3.12/dist-packages/torch/_dynamo/utils.py:3895: UserWarning: Memory efficient kernel not used because: (Triggered internally at /pytorch/aten/src/ATen/native/transformers/cuda/sdp_utils.cpp:986.)
  return node.target(*args, **kwargs)  # type: ignore[operator]
[...]

PR #5543, opened 2026-07-29
[CI] Select mirror_hardwares presets via MIRROR_HW (H100/B200)

yenuo26:
PLEASE FILL IN THE PR DESCRIPTION HERE.

## Purpose

Same CUDA L4 YAML for **nightly H100** and **pre-release extra SKUs** (B200 via `MIRROR_HW=b200`). Uploader expands `mirror_hardwares` at upload time. GitHub labels (`ready` / `merge-test` / `nightly-test` / `weekly-test`) and env names (`NIGHTLY`, `WEEKLY`) are unchanged.

CUDA pipeline **filenames stay** `.buildkite/cuda/test-ready.yml` / `test-merge.yml` / `test-nightly.yml` / `test-weekly.yml` (conceptual L2–L5). Upload keys stay `upload-ready-pipeline` … `upload-weekly-pipeline`. Local runners: `tools/run_ready_jobs.sh`, `tools/run_merge_jobs.sh`, `tools/nightly/run_nightly_jobs.sh`.

Rebased/merged `upstream/main` through `a65e89cb` (`0ba29ede`).

```mermaid
flowchart TD
  test["Test: hardware_test / -m"]
  yml["L4 YAML job"]
  mark["pytest mark includes B200<br/>if the case should run on B200"]
  omit["omit mirror_hardwares"]
  force["mirror_hardwares: h100_2"]
  up["upload_pipeline.py"]
  env{"MIRROR_HW"}
  infer["Match chip from -m<br/>then compose chip_N"]
  skip["Omit step"]
  pool["Agent pool"]

  test --> mark
  yml --> omit
  yml --> force
  mark --> up
  omit --> up
  force --> up
  up --> env
  env -->|"unset: H100 else L4"| infer
  env -->|"b200 and B200 in -m"| infer
  env -->|"no matching SKU"| skip
  infer --> pool
  force -->|"string preset; skip CUDA non-b200 when MIRROR_HW=b200"| pool
```

`MIRROR_HW` must be empty or `b200` (case-insensitive). Typos such as `b20o` **fail the upload**. A name that is not a key in `ci_mirror_hardwares.yml` (including leftover `mirror_hardwares: 2`) **fails the upload**. NPU presets ignore `MIRROR_HW`.

### Test side: add a `B200` mark

B200 is **not** auto-attached. A case that should still run when the job lands on B200 must declare it explicitly:

```python
@hardware_test(res={"cuda": ["H100", "B200"]}, num_cards=2)
```

[...]

PR #5617, opened 2026-07-30
[Bugfix] Fall back from unsupported FA3 ring kernels

lishunyang12:
## Purpose

Addresses the ring-attention failure reported in #5611.

The pinned `fa3-fwd==0.0.3` module imports on Blackwell, but its extension contains only SM80 and SM90a cubins. The import-only availability check therefore selected FA3 on B200/B300 and failed at execution with `no kernel image is available for execution on the device`.

This change records the architecture contract for the pinned wheel, checks the actual query device before selecting FA3 or FA2, and falls back to the existing ring SDPA implementation when neither installed backend supports that device. The separate non-ring LTX2 SDPA traceback in #5611 is outside this patch.

cc @yenuo26 @david6666666

## Test Plan

**vLLM Version:** 0.26.0

**vLLM-Omni Commit:** `bf928460`

```bash
pytest -q \
  tests/diffusion/attention/test_ring_backend_selection.py \
  tests/diffusion/attention/test_attention_sp.py
```

GPU validation uses NVIDIA B300 (SM103), PyTorch 2.11.0+cu130, and CUDA 13.0, including a real two-rank NCCL ring.

## Test Result

```text
22 passed
```

- Parent commit: imports `fa3-fwd`, selects FA3, and reproduces `no kernel image is available for execution on the device`.
- This PR, one GPU: selects SDPA and returns a finite BF16 tensor of shape `(1, 32, 8, 64)`.
- This PR, two GPUs: both NCCL ranks return finite BF16 tensors of the same shape.
- `cuobjdump --list-elf` confirms the pinned extension has SM80/SM90a cubins and no Blackwell cubin.

File-scoped pre-commit hooks pass.

### DIFF
--- a/vllm_omni/diffusion/attention/parallel/ring.py
+++ b/vllm_omni/diffusion/attention/parallel/ring.py
@@ -10,7 +10,12 @@ import torch
 from vllm.logger import init_logger
 
 # import torch.distributed as dist # Not used directly here, but good practice if needed
-from vllm_omni.diffusion.attention.backends.ring.ring_globals import HAS_AITER, HAS_FA3, HAS_FLASH_ATTN
+from vllm_omni.diffusion.attention.backends.ring.ring_globals import (
+    FA3_SUPPORTED_CUDA_MAJORS,
+    HAS_AITER,
+    HAS_FA3,
+    HAS_FLASH_ATTN,
+)
 from vllm_omni.diffusion.attention.backends.ring.ring_selector import AttnType
 from vllm_omni.diffusion.attention.parallel.base import (
     ParallelAttentionContext,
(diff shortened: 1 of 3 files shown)


yenuo26:
I think if you're willing, you can directly push your changes to #5543. After adding the corresponding diffusion-x2v-test label, you can run end-to-end validation on B200. Also, during my debugging on that PR, I've made some overlapping changes to the code in vllm_omni to address these issues—feel free to replace my changes with your fixes.

vllm_omni/diffusion/attention/parallel/ring.py
```diff
@@ -24,6 +29,34 @@
...
+def _can_use_fa3(device: torch.device) -> bool:
+    """Return whether the installed FA3 kernels support ``device``.
+
+    Importing an extension only proves that its Python module is installed.
+    When the source publishes a supported-major contract, also check it before
+    entering a CUDA launcher that may abort instead of raising an exception.
+    """
+    if not HAS_FA3 or device.type != "cuda":
+        return False
+    if FA3_SUPPORTED_CUDA_MAJORS is None:
```
hsliuustc0106, line 41: This leaves the `flash_attn_interface` source-build branch unguarded. It sets `HAS_FA3` but never populates `FA3_SUPPORTED_CUDA_MAJORS`, so SM10x returns true here and selects FA3 again. That implementation is Hopper-only; give this branch an explicit `{9}` contract (or a capability probe) and cover it.

lishunyang12:
> #5543

Noted, I will add my changes on top of your PR.

david6666666:
Folded into #5543 at [`bc8bb338`](https://github.com/vllm-project/vllm-omni/commit/bc8bb33811d3ebdeff44676cfafd6f9c6611b39c).

The architecture-aware FA2/FA3 selection, source-build Hopper contract, pinned `fa3-fwd==0.0.3` contract, ring-SDPA fallback, and regression tests now live on the `B200` branch. The #5543 description and cuDNN comments were also updated to reflect that the former LTX-2 + `CUDNN_ATTN` crash was already fixed by `f9057564`; `CUDNN_ATTN` now remains cuDNN-only and does not silently fall back to MATH.

Closing this PR as superseded by #5543.

PR #5618, opened 2026-07-30
[Bugfix] Preserve short channel-last video tensors

### LINKED ISSUE #5417: [Bug]: Video encoding corrupts channel-last tensors with 3 or 4 frames
### Your current environment

Current `main` at `2bb618903afc9dc29a1f29578dcac31caa047ad5`.

### Your code version

vLLM-Omni `main`.

### Describe the bug

The OpenAI-compatible video encoding path misclassifies channel-last Torch tensors when the generated video contains exactly 3 or 4 frames.

`_normalize_video_tensor()` treats every rank-4 tensor whose first dimension is 3 or 4 as `[C, F, H, W]`. However, generated video tensors are also accepted in the standard channel-last layout `[F, H, W, C]`. For a 3- or 4-frame RGB video, both the frame count and channel count are 3 or 4, so the current first-dimension-only check transposes an already channel-last tensor.

For example, a tensor with shape `[4, 2, 5, 3]` is transformed into `[2, 5, 3, 4]`. The downstream encoder then interprets it as two RGBA frames and drops the fourth channel, producing the wrong frame count and spatial shape.

This affects both synchronous MP4 encoding and fragmented MP4 streaming because both paths call `_coerce_video_to_uint8_frames()`.

Minimal reproduction:

```python
import torch
from vllm_omni.entrypoints.openai.video_api_utils import _coerce_video_to_uint8_frames

video = torch.zeros((4, 2, 5, 3), dtype=torch.uint8)
frames = _coerce_video_to_uint8_frames(video)
assert frames.shape == (4, 2, 5, 3)
```

Current result: `frames.shape == (2, 5, 3, 3)`.

Expected result: channel-last tensors preserve their frame count and shape. Channel-first `[C, F, H, W]` tensors should remain supported.

### Proposed fix

Use the shared NumPy layout normalization for Torch tensors as well. That helper already gives an explicit channel-last dimension precedence in ambiguous shapes and only converts channel-first layouts when the last dimension is not a valid channel count.

lishunyang12:
## Purpose

Resolves #5417.

`_normalize_video_tensor()` transposed every rank-4 Torch tensor whose first dimension was 3 or 4. That is ambiguous with the supported `[frames, height, width, channels]` layout, so three- and four-frame RGB videos were corrupted before encoding.

This change removes the Torch-only early transpose and lets the shared array normalizer resolve layouts. A valid channel-last dimension takes precedence, while unambiguous `[channels, frames, height, width]` tensors are still converted.

cc @pxljs

## Test Plan

**vLLM Version:** 0.26.0

**vLLM-Omni Commit:** `18882c6e`

```bash
pytest -q tests/entrypoints/openai_api/test_video_api_utils.py
```

CUDA validation uses NVIDIA B300 (SM103), PyTorch 2.11.0+cu130, and CUDA 13.0.

## Test Result

```text
10 passed
```

| input layout | parent result | PR result |
|---|---|---|
| `(3, 2, 5, 3)` | `(2, 5, 3, 3)` | `(3, 2, 5, 3)` |
| `(4, 2, 5, 3)` | `(2, 5, 3, 3)` | `(4, 2, 5, 3)` |

Both PR outputs are byte-for-byte equal to their original channel-last inputs. An additional regression test confirms unambiguous channel-first tensors are still converted. File-scoped pre-commit hooks pass.

### DIFF
--- a/tests/entrypoints/openai_api/test_video_api_utils.py
+++ b/tests/entrypoints/openai_api/test_video_api_utils.py
@@ -48,6 +48,25 @@ def test_encode_video_bytes_exports_frames_without_interpolation(monkeypatch):
     assert mux_calls[0]["audio"] is None
 
 
+@pytest.mark.parametrize("frame_count", [3, 4])
+def test_coerce_video_preserves_ambiguous_channel_last_tensor(frame_count):
+    video = torch.arange(frame_count * 2 * 5 * 3, dtype=torch.uint8).reshape(frame_count, 2, 5, 3)
+
+    frames = video_api_utils._coerce_video_to_uint8_frames(video)
+
+    assert frames.shape == (frame_count, 2, 5, 3)
+    np.testing.assert_array_equal(frames, video.numpy())
+
+
+def test_coerce_video_converts_channel_first_tensor():
+    video = torch.arange(3 * 4 * 2 * 5, dtype=torch.uint8).reshape(3, 4, 2, 5)
+
+    frames = video_api_utils._coerce_video_to_uint8_frames(video)
+
+    assert frames.shape == (4, 2, 5, 3)
+    np.testing.assert_array_equal(frames, video.permute(1, 2, 3, 0).numpy())
+
+
 def test_fragmented_mp4_video_encoder_reuses_single_muxer(monkeypatch):
     muxers = []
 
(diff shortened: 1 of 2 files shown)


lishunyang12:
Already fixed by the issue author.

PR #5619, opened 2026-07-30
[Bugfix] Match image file metadata to output format

### LINKED ISSUE #5421: [Bug]: /v1/images/generations with response_format="file" always returns image/png, ignoring output_format
<details>

```text
        System Info
OS                           : macOS 15.3.1 (arm64)
GCC version                  : Could not collect
Clang version                : 17.0.0 (clang-1700.0.13.5)
CMake version                : version 4.1.1
Libc version                 : N/A

       PyTorch Info
PyTorch version              : 2.11.0
CUDA used to build PyTorch   : None

      Python Environment
Python version               : 3.13.5 (64-bit runtime)
Python platform              : macOS-15.3.1-arm64-arm-64bit-Mach-O

       CUDA / GPU Info
Is CUDA available            : False
CUDA runtime version         : No CUDA
GPU models and configuration : No CUDA
Nvidia driver version        : No CUDA

          CPU Info
Apple M4 Pro

[pip3] numpy==2.4.6
[pip3] pyzmq==27.1.0
[pip3] torch==2.11.0
[pip3] torchaudio==2.11.0
[pip3] torchcodec==0.15.0
[pip3] torchsde==0.2.6
[pip3] torchvision==0.26.0
[pip3] transformers==5.14.1
[pip3] x-transformers==2.24.0

         vLLM Info
vLLM-Omni Version            : dev (git sha: 2bb61890)
vLLM Build Flags:
  CUDA Archs: Not Set; ROCm: Disabled
GPU Topology:
  Could not collect
```

This is a CPU-only macOS box (vLLM built from source with `VLLM_TARGET_DEVICE=cpu`).
The bug is in the HTTP response layer and never reaches the engine, so it does not
depend on the platform, backend or model. The second repro below loads no model at all.

</details>

### Your code version

<details>

```text
0.26.0
```
</details>
<details>

```text
main @ 2bb618903afc9dc29a1f29578dcac31caa047ad5
```
</details>

### 🐛 Describe the bug

`POST /v1/images/generations` with `response_format: "file"` always returns
`Content-Type: image/png` and a `.png` filename, even when the request asks for
`output_format: "jpeg"` or `"webp"`.

The bytes themselves are encoded correctly, so what you get back is a real JPEG or
WebP file labelled as a PNG. Anything that trusts the content type or the file
extension (browsers, `curl -OJ`, image libraries that pick a decoder from the
filename) will either fail to open it or only work because the decoder sniffs the
magic bytes and ignores the label.

#### Root cause

`ImageGenerationResponse.stream_response()` hardcodes PNG, even though the response
model already holds the real format in `self.output_format`
([images.py:194](https://github.com/vllm-project/vllm-omni/blob/main/vllm_omni/entrypoints/openai/protocol/images.py#L194)):

```python
if len(self.data) == 1:
    image_bytes = base64.b64decode(self.data[0].b64_json)
    filename = f"image_{uuid.uuid4().hex[:8]}.png"      # always .png
    return StreamingResponse(
        io.BytesIO(image_bytes),
        media_type="image/png",                          # always image/png
        ...
```

The multi-image ZIP branch just below has the same issue, every entry is named
`image_{idx}.png` whatever the format:

```python
zf.writestr(f"image_{idx}.png", base64.b64decode(item.b64_json))
```

The encoding side handles the format properly, in
[api_server.py:1924](https://github.com/vllm-project/vllm-omni/blob/main/vllm_omni/entrypoints/openai/api_server.py#L1924):

```python
output_format = _choose_output_format(request.output_format or "png", None)
image_data = [
    ImageData(b64_json=encode_image_base64_with_compression(img, format=output_format), ...)
    for img in images
]
response = ImageGenerationResponse(..., output_format=output_format)
if request.response_format != ResponseFormat.FILE:
    return response
return response.stream_response()          # format is dropped here
```

So `output_format` makes it all the way into the response model and is then ignored
by the one branch that needs it. Only `response_format: "file"` is affected. The
default `b64_json` path is fine, since the client is told the format through the
`output_format` field in the JSON response.

#### How to reproduce

Against a running server with any diffusion model:

```bash
curl -sS -D - -o out.png http://localhost:8000/v1/images/generations \
  -H 'Content-Type: application/json' \
  -d '{"prompt": "a red square", "response_format": "file", "output_format": "webp"}' \
  | grep -i -E 'content-type|content-disposition'
file out.png
```

Got:

```text
content-type: image/png
content-disposition: attachment; filename="image_108c8469.png"
out.png: RIFF (little-endian) data, Web/P image
```

Expected `content-type: image/webp` and a `.webp` filename.

Since none of this touches the engine, it also reproduces with no server, no model
and no accelerator:

```python
import base64
import PIL.Image

from vllm_omni.entrypoints.openai.image_api_utils import encode_image_base64_with_compression
from vllm_omni.entrypoints.openai.protocol.images import ImageData, ImageGenerationResponse

img = PIL.Image.new("RGB", (8, 8), (255, 0, 0))
for fmt in ["png", "jpeg", "webp"]:
    b64 = encode_image_base64_with_compression(img, format=fmt)
    raw = base64.b64decode(b64)
    resp = ImageGenerationResponse(
        created=0, data=[ImageData(b64_json=b64)], output_format=fmt, size="8x8"
    ).stream_response()
    print(
        f"requested={fmt:5s} actual_bytes={raw[:4]!r:22s} "
        f"media_type={resp.media_type:10s} "
        f"filename={resp.headers.get('content-disposition')}"
    )
```

On `main`:

```text
requested=png   actual_bytes=b'\x89PNG'             media_type=image/png  filename=attachment; filename="image_a02622f8.png"
requested=jpeg  actual_bytes=b'\xff\xd8\xff\xe0'    media_type=image/png  filename=attachment; filename="image_f80ebea9.png"
requested=webp  actual_bytes=b'RIFF'                media_type=image/png  filename=attachment; filename="image_108c8469.png"
```

The last two rows are JPEG (`\xff\xd8\xff`) and WebP (`RIFF`) payloads going out as
`image/png`.

#### Possible fix

Pick the extension and MIME type from `self.output_format` in `stream_response()`,
mapping `jpg`/`jpeg` to `image/jpeg` and `webp` to `image/webp`, and use the same
extension for the ZIP entries. A `png` fallback keeps the current behaviour when
`output_format` is unset. That fallback does get hit in practice, because
`ImageGenerationResponse.output_format` is declared as `str = Field(None, ...)` at
[images.py:186](https://github.com/vllm-project/vllm-omni/blob/main/vllm_omni/entrypoints/openai/protocol/images.py#L186),
which is a non-optional annotation with a `None` default.

I could not find a test that asserts the content type of a `response_format: "file"`
response. `tests/entrypoints/openai_api/test_image_server.py:1305` checks the `n>1`
ZIP shape but not the format, so a regression test covering the three formats would
be worth adding along with the fix.

Happy to send a PR for this if nobody else is already on it.

lishunyang12:
## Purpose

Resolves #5421.

`ImageGenerationResponse.stream_response()` hardcoded `image/png` and `.png` even when the encoded bytes were JPEG or WebP; multi-image ZIP entries had the same mismatch.

This change maps `png`, `jpg`/`jpeg`, and `webp` to matching MIME types and extensions for single-file and ZIP responses, preserves PNG as the metadata fallback, and makes the accepted `jpg` alias encodable by Pillow as JPEG.

cc @yashkgp

## Test Plan

**vLLM Version:** 0.26.0

**vLLM-Omni Commit:** `6f4992c1`

```bash
pytest -q tests/entrypoints/openai_api/test_image_server.py
```

The parent commit reproduces mismatched metadata for JPEG and WebP. Parametrized regression tests cover single-image responses and multi-image ZIP entries.

## Test Result

```text
77 passed
```

| format | parent metadata | PR metadata |
|---|---|---|
| PNG | `image/png`, `.png` | `image/png`, `.png` |
| JPEG | `image/png`, `.png` | `image/jpeg`, `.jpg` |
| WebP | `image/png`, `.png` | `image/webp`, `.webp` |

The encoded bytes, MIME type, and extension match for `png`, `jpg`, `jpeg`, and `webp`. File-scoped pre-commit hooks pass. This is an HTTP-layer fix; the same path was also exercised in the CUDA 13.0/PyTorch 2.11.0+cu130 B300 environment, but it does not invoke a GPU kernel.


tests/entrypoints/openai_api/test_image_server.py
```diff
@@ -1301,14 +1301,19 @@ def test_generate_images_rejects_model_mismatch(test_client):
     assert "model mismatch" in response.json()["detail"].lower()
 
 
-def test_image_file_response_format_multiple(test_client):
+@pytest.mark.parametrize(
+    ("output_format", "expected_pil_format"),
+    [("png", "PNG"), ("jpeg", "JPEG"), ("webp", "WEBP")],
+)
+def test_image_file_response_format_multiple(test_client, output_format, expected_pil_format):
```
hsliuustc0106, line 1308: `test_client` only exercises the single-stage dispatcher. The supported multi-stage path returns before `stream_response()` and always base64-encodes PNG, so `response_format="file"` still returns JSON for multi-stage models (including JPEG/WebP requests). Please add these cases with `async_omni_test_client` and route that dispatcher through the common format/file response builder.

Gaohan123 (approved):
LGTM. Thanks

PR #5228, opened 2026-07-20
[perf] Minicpm o4.5 optimization

R2-Y:
PLEASE FILL IN THE PR DESCRIPTION HERE.

## Purpose
1.Split MiniCPM-o 4.5 into Thinker, Talker AR, and batched Code2Wav stages.
2. Stream request-isolated codec chunks through async connectors.
3. Add exact-shape Code2Wav batching and lifecycle cleanup.
4. Generalize connector routing and runner metadata capabilities.

## Test Plan
daily-omni benchmark test

## Test Result
image
image
image

metric | PR #5233 | this pr | delta
-- | -- | -- | --
Successful / Failed | 1197 / 0 | 1197 / 0 | -
Benchmark duration (s) | 1851.74 | 1009.61 | ↓45.5%
Request throughput (req/s) | 0.65 | 1.19 | ↑83%
Peak concurrent requests | 12 | 15 | ↑25%
Mean E2EL (ms) | 15437.34 | 8378.54 | ↓45.7%
Output token throughput (tok/s) | 1.93 | 3.55 | ↑84%
Peak output tok/s | 9.00 | 14.00 | ↑56%
Total Token throughput (tok/s) | 1761.98 | 3231.65 | ↑83%
Mean TTFT (ms) | 1123.81 | 1297.28 | ↑15.4%（deteriorate）
Mean TPOT (ms) | 11.02 | 44.08 | ↑300%（deteriorate）
Mean ITL (ms) | 11.01 | 44.21 | ↑302%（deteriorate）
Audio throughput (audio-s/s) | 4.11 | 8.31 | ↑102%
Mean AUDIO_TTFP (ms) | 15435.91 | 3864.27 | ↓75%
Mean AUDIO_RTF | 4.36 | 1.72 | ↓61%

**BEFORE SUBMITTING:** read [CONTRIBUTING.md](https://github.com/vllm-project/vllm-omni/blob/main/CONTRIBUTING.md) and run the [precheck-pr skill](https://github.com/vllm-project/vllm-omni/blob/main/.claude/skills/precheck-pr/SKILL.md) with the code agent for a self-check against project conventions.
(anything written below this line will be removed by GitHub Actions)

PR #5622, opened 2026-07-30
[Bugfix][MiniCPM-o] Fix offline example for the three-stage pipeline

eval-dev:
## Purpose

`examples/offline_inference/minicpmo/end2end.py` was written against the fused
two-stage MiniCPM-o 4.5 topology and was not updated when the Talker/Code2Wav
split landed in #5228. On current `main` the documented smoke test
(`run_single_prompt.sh`) cannot produce correct audio.

Two independent problems:

**1. Sampling-params list is one entry short.** `sampling_params_list` was built
from two entries and then sliced with `[: omni.num_stages]`. With
`num_stages == 3` the slice is a no-op, so `generate()` raises:

```
ValueError: Expected 3 sampling params, got 2
```

**2. The Talker stops after one chunk.** The Talker used `max_tokens=1`, with the
comment *"max_tokens=1 satisfies the scheduler; waveform is produced in-process
by Token2Wav"*. That was correct while Token2Wav ran inside stage 1 and a single
decode step produced the whole waveform. After the split, each step advances the
codec stream, so `max_tokens=1` stops it after one chunk and yields ~0.16 s of
audio instead of ~10 s.

This second one **fails silently** — the run still exits 0 and writes a valid
24 kHz WAV, just a truncated one. It is only visible if you check the duration.

The fix mirrors stage 1 and stage 2 of the MiniCPM-o 4.5 deploy configs
(`min_tokens: 50`, `max_tokens: 2048`, `stop_token_ids: [1]`), and brings this
example in line with `examples/offline_inference/qwen3_omni/end2end.py`, which
already uses the three-entry + `[:num_stages]` pattern (including
`stop_token_ids` on its talker). No library code is touched.

## Test Plan

```bash
python examples/offline_inference/minicpmo/end2end.py \
    --deploy-config vllm_omni/deploy/minicpmo_4_5_3gpu.yaml \
    --query-type text \
    --output-wav ./smoke_out
```

Then check the emitted WAV, since the Talker bug does not surface in the exit code:

```bash
python -c "
import wave, glob
w = wave.open(glob.glob('./smoke_out/output_*.wav')[0])
print('ch=%d rate=%d dur=%.2fs' % (w.getnchannels(), w.getframerate(),
                                   w.getnframes() / w.getframerate()))"
```

**vLLM Version:** 0.26.0

**vLLM-Omni Commit:** a951dd0c

## Test Result

Hardware: 3x NVIDIA H800 80GB. Deploy config: `minicpmo_4_5_3gpu.yaml`.

| | Result |
|---|---|
| `main` | `ValueError: Expected 3 sampling params, got 2` |
| with fix 1 only | exits 0, but `ch=1 rate=24000 dur=0.16s` |
| this PR | exits 0, `ch=1 rate=24000 dur=10.24s` |

Text output is unchanged by this PR:

```
Hello! I am vLLM Omni, an AI model designed to assist with various tasks by
understanding context across multiple modalities like vision and language.
```

Scope note: verified with `--query-type text`. The change is to sampling
parameters only, which do not vary by query type, but the other query types
(`use_image`, `use_audio`, `use_video`, `use_mixed_modalities`) were not
exercised in this run.

Note on the single-GPU config: `minicpmo_4_5_3gpu.yaml` was used because the
default `minicpmo_4_5.yaml` co-locates all three stages on GPU 0 at
`0.55 + 0.22 + 0.22 = 0.99` utilization. Once the Talker emits a real codec
stream, Code2Wav's HiFi-GAN `conv_transpose1d` then fails with
`RuntimeError: GET was unable to find an engine to execute this computation` —
cuDNN cannot obtain workspace outside vLLM's pools. The upsampling layers
themselves are fine (reproduced standalone in fp32 and bf16, cuDNN on and off,
all pass). That is a separate deploy-config concern and is not addressed here.


amy-why-3459:
LGTM for the sampling-params fix — matches the post-#5228 three-stage deploy configs and the silent `max_tokens=1` truncation is a real footgun.

Can you also update `examples/offline_inference/minicpmo/README.md` (and the `end2end.py` module docstring) so they no longer describe the fused two-stage Talker+Token2Wav topology?

Issue #3745, opened 2026-05-19
[RFC] Full-Duplex Session Architecture for vLLM-OMNI


lishunyang12:
## Motivation

A new class of speech models — MiniCPM-o 4.5 (Omni-Flow / TDM), Nemotron VoiceChat (`nemotron_duplex_h`), SoulX-Duplug, Moshi— are **full-duplex by design**: they perceive and respond *concurrently*. The user can speak while the model is speaking, interrupt mid-sentence, and the model keeps a single conversation context alive for minutes. This is the "Doubao / Gemini Live / GPT-4o voice" experience: no push-to-talk, no turn button, sub-300ms barge-in.

vLLM-OMNI cannot serve this today. Three model PRs (#3642, #3512, #1967) have independently hit the **same wall**: the runtime is request-oriented (`prompt → output → done`, KV freed at finish), the models are stream-oriented (continuous audio in *while* generating out, conversation-lifetime KV cache). Each PR is at risk of growing its own ad-hoc streaming path that we later have to reconcile. **This RFC defines the duplex session primitive once, and enumerates how each model conforms.**

This is also explicitly shared infrastructure with World Models (#1987): *"client pushes observations, server returns predictions, repeat"* is the same persistent bidirectional loop as *"client pushes audio, server returns audio."*

### Use Cases

| Use Case | Input cadence | Barge-in | Session length | Latency target |
|----------|---------------|----------|----------------|----------------|
| Live voice assistant (Doubao-style) | 80–160 ms audio | Required | minutes | < 400 ms v2v |
| Real-time interpreter / dubbing | continuous | Soft | minutes | < 1 s |
| Embodied agent (#1987 world model) | obs stream | Required | unbounded | < 500 ms |
| Call-center / IVR | continuous | Required | minutes | < 700 ms |
| Accessibility narration | continuous | Optional | unbounded | < 1 s |

---

## The Impedance Mismatch (grounded in current code)

```mermaid
flowchart LR
    subgraph REQ["vLLM-OMNI today — request-oriented"]
        A1[add_request] --> A2[prefill prompt]
        A2 --> A3[decode loop]
        A3 --> A4[finish]
        A4 --> A5["_free_request → _free_blocks<br/>KV returned to manager"]
    end
    subgraph DUP["Duplex models — stream-oriented"]
        B1[open session] --> B2[continuous chunks in]
        B2 --> B3[generate out concurrently]
        B3 --> B2
[...]

PR #5626, opened 2026-07-31
[Bugfix][Full Duplex] Report the real error from the duplex chat fallback

TheCodeWrangler:
## Purpose

`ErrorResponse` nests its detail under `error: ErrorInfo` and has no flat `message` / `type` of its own, so the error branch in `_run_response` raised:

```
AttributeError: 'ErrorResponse' object has no attribute 'message'
  chat_fallback.py:38 in _run_response
    await send_json({"type": "error", "error": result.message, "code": result.type or "chat_error"})
```

An error handler that fails while reporting a failure. The outer `except Exception` catches the `AttributeError` and reports *that* to the client with `code: response_error`, so the original cause is lost.

Observed live: the underlying error was `ValueError: Invalid or unsupported audio file.` from `vllm/multimodal/media/audio.py`. That is genuinely useful — it says the fallback was handed audio it could not decode — and the crash hid it completely.

For reference, the actual schema:

```python
ErrorInfo.model_fields      # message: str (req), type: str (req), param: str | None, code: int (req)
ErrorResponse.model_fields  # error: ErrorInfo
```

Found while bringing up a native duplex adapter for Qwen3-Omni against `experimental/fullduplex/`. Not Qwen3-Omni specific — latent for MiniCPM-o 4.5 on any fallback error.

## Repro

```bash
git fetch origin fix/duplex-second-model-blockers && git checkout FETCH_HEAD
git checkout upstream/main -- vllm_omni/experimental/fullduplex/openai/chat_fallback.py
pytest tests/e2e/features/fullduplex/openai/test_chat_fallback_error.py
```

```
>       assert errors[0]["error"] == "boom"
E       assert "'ErrorRespon...ute 'message'" == 'boom'
ERROR [chat_fallback.py:63] Duplex response failed: 'ErrorResponse' object has no attribute 'message'
```

Restore the file and it passes.

## Changes

Read the nested `ErrorInfo`. Four lines, `experimental/` only.

This matches how the rest of the tree already reads an `ErrorResponse`: `video_stream_session.py:394` (`generator.error.message`) and `duplex_control_client.py:86` (`result_message.error.message`).
## Test Plan

New `tests/e2e/features/fullduplex/openai/test_chat_fallback_error.py` (`core_model`, `cpu`) drives the real `_run_response` with a chat service that returns an `ErrorResponse`.

**Without the source fix** the test collects and fails with exactly the reported symptom:

```
>       assert errors[0]["error"] == "boom"
E       assert "'ErrorRespon...ute 'message'" == 'boom'

ERROR [chat_fallback.py:63] Duplex response failed: 'ErrorResponse' object has no attribute 'message'
ERROR [chat_fallback.py:63] AttributeError: 'ErrorResponse' object has no attribute 'message'
```

That log line is the mechanism: the outer `except Exception` catches the `AttributeError` and reports *it* to the client, so the real cause never arrives.

**With the fix**, full suite in a CPU-only container (`vllm 0.26.0`):

```
tests/e2e/features/fullduplex -> 174 passed
```

## Note on scope

This PR previously bundled a second, unrelated fix (Qwen3-Omni advertising `talker_mtp` on the thinker stage). That one has no test on `main` — it needs a thinker stage with a preprocess hook, which does not exist yet — so it has been dropped from here and will come with the work that makes it reproducible, rather than asking a reviewer to take it on description alone.

## Related

- #5613 — `DuplexFence` not JSON serializable in `runtime_control`, same class of defect found on the same bring-up (merged).
- #5658 — streaming segment state aliased across stages, also found on this bring-up.
- #3745 — the full-duplex RFC.


amy-why-3459:
@Sy0307 @natureofnature PTAL

Issue #5487, opened 2026-07-28
[Bug]: Nightly CI, FAILED tests/e2e/online_serving/test_bagel_expansion.py::test_bagel[tp_vae_patch_parallel_2]


congw729:
### Your current environment

<details>
<summary>The output of <code>python collect_env.py</code></summary>

```text
ci env
```

</details>

### Your code version

<details>
<summary>The commit id or version of vllm</summary>

```text
0.26.0
```
</details>
<details>
<summary>The commit id or version of vllm-omni</summary>

```text
d4a5eaac9081cb8345aa535aba4eb4218d8730dc
```
</details>

### 🐛 Describe the bug

https://buildkite.com/vllm/vllm-omni/builds/12788/canvas?sid=019fa4bf-aeb3-4fa6-b856-45ad0c17115f&tab=output

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://vllm-omni.readthedocs.io), which can answer lots of frequently asked questions.

PR #5482, opened 2026-07-28
[BugFix] Never dispatch diffusion with an incomplete CFG companion bundle

tzhouam:
## Summary

Nightly Bagel failures (builds 2892/2918: `test_bagel_shared_memory_connector` pixel mismatches and the multi-GPU `test_bagel[tp_vae_patch_parallel_2]` NCCL watchdog kill) trace to a CFG companion-output race in the orchestrator: `_handle_kv_ready_raw_outputs` marked a companion **done on `kv_ready` alone**, before its processed output was stashed, so the parent could pass the `all_companions_done` gate and dispatch to the diffusion stage with 0/N companion outputs — the recurring `only 0/2 CFG companion outputs arrived` warning. That silently degrades CFG conditioning, and after an abort leaves ranks running divergent branches (observed as mismatched all_reduce sizes 7168 vs 71680 on the same PG/SeqNum in `bagel_transformer.forward_cache_update_text`, hanging NCCL for 600s until the watchdog kills the worker).

## Lifecycle now enforces one invariant — diffusion dispatch never happens with an incomplete companion bundle

- A companion counts as done **only after** `set_companion_output` stashes its processed output (`_route_output` is the sole done-marking site); the `kv_ready` fast path no longer done-marks.
- The forward gate checks completeness non-destructively and re-defers the parent instead of dispatching. Bundles are **peek-not-pop** (`get_companion_outputs`): a streaming re-submission bundles the complete set again, never `[parent]` alone.
- `_cleanup_request_ids` is CFG-aware — every teardown path (stage error, abort, replica loss, membership unregister) releases parent tracker state, folds companions into the batch, and fails a deferred parent whose companion was aborted/lost before its output arrived. A late abort of an already-completed companion does **not** cancel its parent.

Deferred to a follow-up (needs a liveness owner in the orchestrator loop): a deadline on deferred parents and detection of a companion replica dying silently.

## Test

- 16 new deterministic unit tests in `tests/engine/test_cfg_companion_lifecycle.py` covering ordering (parent-first/companion-first/duplicates), the forward gate (re-defer, fail-on-lost-bundle, repeated forward captured at the diffusion input hook), abort semantics (before/after companion completion, parent-abort expansion), and replica-loss cleanup.
- Full `tests/engine` orchestrator/tracker/abort suites pass (74 tests).
- Reviewed over three rounds by GPT-5.6 Sol (cursor-agent); final verdict APPROVE.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

PR #5636, opened 2026-07-31
[BugFix] Make pure-TP KV receive fallback all-or-nothing across ranks

### LINKED ISSUE #5627: [Bug]: Nightly CI, BAGEL tp_vae_patch_parallel_2 fails again with NCCL ALLREDUCE timeout
### Your current environment

<details>

```text
CI env (2 x NVIDIA H100 80GB HBM3)
```

</details>

### Your code version

<details>

```text
0.26.0
```
</details>
<details>

```text
a951dd0cb4eeb9a6c28b2a075e62c2144b1bc272
```
</details>

### 🐛 Describe the bug

The nightly CI failure from #5487 has recurred in the same test:

```text
FAILED tests/e2e/online_serving/test_bagel_expansion.py::test_bagel[tp_vae_patch_parallel_2]
openai.InternalServerError: Error code: 500 - EngineCore encountered an issue
```

The server log shows rank 1 timing out in an ALLREDUCE after 600 seconds:

```text
[Rank 1] Watchdog caught collective operation timeout:
WorkNCCL(SeqNum=115, OpType=ALLREDUCE, NumelIn=71680, NumelOut=71680,
Timeout(ms)=600000) ran for 600035 milliseconds before timing out.
RuntimeError: DiffusionExecutor is closed.
```

Other BAGEL configurations passed; only `tp_vae_patch_parallel_2` failed.

- Build: https://buildkite.com/vllm/vllm-omni/builds/12981
- Failed attempt: https://buildkite.com/vllm/vllm-omni/builds/12981/list?jid=019fb431-f35a-4661-b79f-30d7de8636f6
- Previous occurrence: #5487

Expected: the TP + VAE patch parallel configuration completes without an NCCL collective timeout.

tzhouam:
## Purpose

Fixes #5627 (recurrence of #5487): nightly `test_bagel[tp_vae_patch_parallel_2]` dying with a 600s NCCL ALLREDUCE watchdog timeout followed by `RuntimeError: DiffusionExecutor is closed`.

## Root cause

The hung collective is not a VAE collective: NumelIn=71680 = ~20 tokens × 3584 hidden — a TP allreduce inside Bagel's `forward_cache_update_text` prefill. The chain:

1. `tp_vae_patch_parallel_2` is **pure TP** (VAE patch parallelism is neither SP nor CFG-parallel), so `_TransferTopoConfig` assigns role `LOCAL` to every rank and `distribute_kv_cache` is a no-op — each TP rank independently fetches its own per-rank KV key with `recv_timeout=30s`.
2. A receive timeout **returns silently** (`receive_multi_kv_cache` → `(None, 0)`, no raise).
3. The rank that missed falls into `pipeline_bagel.py`'s `injected_kv is None` branch and re-encodes text locally **through the TP-sharded LM**, entering allreduces the KV-holding peer never joins.
4. NCCL watchdog kills the worker after 600s → executor marked dead → every subsequent request 500s with "DiffusionExecutor is closed".

This is why only the pure-TP Bagel config fails (others don't TP the LM or use leader→follower KV distribution), why it's flaky (stage-1 rank 0 shares GPU 0 with the AR stage, so per-rank timing skew is built in), and why #5482 — which fixed the orchestrator-level CFG bundle race — didn't stop the recurrence: the nightly that failed ran at #5482's own merge commit.

## Fix

After the per-rank receive in pure-TP topology, ranks exchange their receive flags (rank 0 collects via the world group's CPU object plane and broadcasts the verdict). If any rank missed, ranks that received **drop their applied KV** (cleared symmetrically with what apply/broadcast set) so every rank takes the identical no-KV prefill path — the TP collective schedule can never diverge. The skip guards depend only on topology/config/request-id values that are identical on every rank, so ranks always agree on whether the exchange happens; a rank whose connector failed to initialize still participates and reports a miss instead of deadlocking.

Hooked into both entry points: `receive_multi_kv_cache_distributed` and `consume_and_distribute_kv_cache` (prefetch path).

## Test Plan

- New `tests/distributed/omni_connectors/test_kv_recv_tp_consensus.py` (9 CPU tests): skip guards (non-TP topologies, single-rank group, `need_recv_cache=False`, dummy warmup), rank-0 verdict collection + KV drop on peer miss, follower flag report + verdict honoring, and wiring from both entry points.
- Updated `test_tp_without_cfg_keeps_independent_receive_path` to the new contract: per-rank fetch is preserved, plus the flag exchange.

## Test Result

- All 116 tests across the touched connector test files pass locally; full `tests/distributed/omni_connectors/` failures are identical to clean-main baseline (pre-existing RDMA/mooncake env failures only).


vllm_omni/distributed/omni_connectors/kv_transfer_manager.py
```diff
@@ -1747,6 +1761,58 @@ def _recv_kv_payload(self, group: Any, src: int, device: torch.device) -> dict[s
...
+        watchdog kills the worker (#5627). Exchange per-rank receive flags;
+        if any rank missed, every rank drops its KV so all take the same
+        fallback path together.
+
+        The skip guards below depend only on topology, config, and the
+        request id — identical on every rank — so ranks always agree on
+        whether the exchange happens.
+        """
+        pt = self.topo_config
+        group = pt.world
```
Gaohan123: [P1] Scope the receive consensus to the TP group

This consensus must only include ranks that jointly execute the same TP-sharded request. `pt.world` also contains DP/PP ranks, while the `LOCAL` role currently only excludes CFG/SP topologies. In TP+DP execution, different DP replicas can process different request IDs, so this code mixes receive flags from unrelated requests; a miss in one replica unnecessarily invalidates another request. More seriously, if one replica skips the exchange—for example because it is handling a dummy or invalid request—while another enters it, the world-group send/broadcast sequence can deadlock.

Please use the tensor-parallel group (for example, `get_tp_group()`) for this exchange and add a TP=2, DP=2 test where the two DP replicas handle different request IDs and receive outcomes.

vllm_omni/distributed/omni_connectors/kv_transfer_manager.py
```diff
@@ -1757,6 +1823,7 @@ def receive_multi_kv_cache_distributed(
         received = False
         if not self.topo_config.is_follower:
             received = self.receive_multi_kv_cache(req, cfg_kv_collect_func, target_device)
+        received = self._tp_local_receive_consensus(req, received)
```
Gaohan123, line 1856: [P1] Include CFG companion KV results in the consensus verdict

At this point, `received` only represents the primary KV result: `receive_multi_kv_cache()` returns `primary_ok` even when one or more `cfg_kv_collect_func` companion receives fail. Bagel's collector reports such failures by returning a partial dictionary, so all ranks can report `received=True` while only some ranks have `cfg_text` or `cfg_img` KV. Those TP ranks then continue with different branch caches and potentially different cache lengths, producing inconsistent TP-sharded computation and possibly another collective mismatch.

This is relevant to the failing Bagel test because it supplies a negative prompt and therefore creates a `cfg_text` companion request. Please make the consensus cover every expected companion role as well as the primary KV, and clear the complete payload on all TP ranks if any required role is missing on any rank. A regression test should cover primary success on every rank with a companion timeout on one rank.

vllm_omni/distributed/omni_connectors/kv_transfer_manager.py
tzhouam: Good catch — agreed on both counts, and the deadlock half was the serious one: a DP replica skipping the exchange on a dummy request while another entered it would hang the world group. Fixed in bac3eafc:

- `_TransferTopoConfig` now carries the stage's tensor-model-parallel group (`vllm.distributed.parallel_state.get_tp_group()`, `None` when vLLM TP isn't initialized) and the consensus runs exclusively on it; `pt.world` is never touched by the exchange. Within the TP group every skip guard (topology, `need_recv_cache`, dummy-request id) is symmetric, since all TP ranks execute the same request.
- Tests: `test_world_group_is_never_used_for_the_exchange` models the TP=2×DP=2 layout you described (world of 4, TP group of 2) and asserts the exchange happens on the TP group with world receiving zero calls; the existing rank-role tests were moved onto the TP group. A true multi-process TP×DP e2e isn't in CI's current matrix, but the deadlock precondition — asymmetric entry into the exchange — is now structurally impossible because no guard depends on state that can differ within the TP group.

Local `test_bagel[tp_vae_patch_parallel_2]` e2e re-passed on the updated code (9 min, L20X×2, vLLM 0.26).

vllm_omni/distributed/omni_connectors/kv_transfer_manager.py
tzhouam: Confirmed — `receive_multi_kv_cache` returns `primary_ok` only and companion collection is best-effort (partial dicts applied, exceptions swallowed), so ranks could agree on `received=True` while diverging on `cfg_text`/`cfg_img`. Fixed in bac3eafc:

- Ranks now exchange a state signature instead of a flag: `(primary_received, tuple(sorted(_collect_request_kv_payload(req))))`. Reusing `_collect_request_kv_payload` means the signature automatically tracks every applied role — primary, metadata, and all `cfg_*` branch fields — with no separate role list to maintain. Verdict is "identical signature on every TP rank"; any divergence clears the full payload everywhere (via the same enumeration) and falls back together.
- One deliberate deviation from "clear if any required role is missing on any rank": if a role is missing on *all* ranks symmetrically, the state is kept. Symmetric absence is consistent — for `cfg_text` it's Bagel's documented empty-unconditional-branch path (original BAGEL text2img behavior) — so failing it would degrade requests that succeed today. Only cross-rank *divergence* can desync collectives, and that's exactly what the signature comparison rejects. Happy to tighten to strict completeness if you see a case symmetric absence breaks.
- Tests: `test_rank0_drops_kv_on_companion_role_divergence` (primary OK everywhere, `cfg_text` on one rank only → cleared on all) and `test_rank0_keeps_symmetric_partial_state`.

Live validation: in the local e2e rerun of `test_bagel[tp_vae_patch_parallel_2]` (which sends a negative prompt, so `cfg_text` is exercised), the new path caught a real divergence — rank 0 reported `['kv_metadata', 'past_key_values', 'sp.cfg_kv_request_ids', ...]`, rank 1 only `['sp.cfg_kv_request_ids']` — both ranks dropped and fell back together, and the test passed.

Gaohan123 (approved):
LGTM. Thanks

Issue #5442, opened 2026-07-27
[Bug] MiniCPM-o 4.5 code2wav resolves assets against the HF repo-id string — kills all MiniCPM e2e in H100 CI and cascades into qwen3 startup OOMs


linyueqian:
### What happens

Every MiniCPM-o 4.5 e2e test in the H100 `full_model` CI sweep dies at server startup:

```
FileNotFoundError: MiniCPM-o Code2Wav prompt audio not found: openbmb/MiniCPM-o-4_5/assets/HT_ref_audio.wav
```

`minicpmo_4_5_code2wav.py` builds its asset paths from the raw model string:

- `model_path = str(vllm_config.model_config.model)` (L120)
- default prompt wav = `Path(model_path) / "assets" / "HT_ref_audio.wav"` (L129-133)
- `token2wav_path = Path(model_path) / "assets" / "token2wav"` (L606) would fail the same way next

When the model is passed as a local directory this works, which is why local runs never see it. CI passes the HF repo id, so `openbmb/MiniCPM-o-4_5/assets/...` is treated as a relative filesystem path that doesn't exist.

### Why it looks like an OOM

The failed startups leave their sibling stage procs running (the e2e harness doesn't reap them on startup failure). By the time the qwen3 tests in the same sweep start, the card is gone — the torch error names the leftovers directly:

```
GPU 0 has a total capacity of 79.18 GiB of which 51.44 MiB is free.
Process 7774 has 43.61 GiB memory in use. Process 11222 has 33.96 GiB memory in use.
```

so all 39 `test_qwen3_omni_expansion` params then error with `Stage 1 replica 0 exited with code 0 before API server became ready`. Final tally for the step: 1 passed, 18 skipped, 52 errors — every error traces back to this path bug plus the leak.

Seen in build 12693 (PR #5414) and identically in build 12700 (main, commit 2bb6189), step `Omni · Function Test with H100`.

### Suggested fix

When `model_config.model` is not an existing directory, resolve it first, e.g.

```python
model_ref = vllm_config.model_config.model
if not Path(model_ref).is_dir():
    from huggingface_hub import snapshot_download
    model_ref = snapshot_download(model_ref, allow_patterns=["assets/*"])
self.model_path = str(model_ref)
```

(cache hit is instant since the weights are already downloaded). Separately, the harness reaping leaked stage procs on failed startup would stop one broken model from taking down the rest of the sweep — happy to send a PR for either.

PR #5638, opened 2026-07-31
[Perf][MiniCPM-o] TensorRT acceleration for the Code2Wav vocoder (DiT estimator + campplus)

yuekaizhang:
## Usage

One toggle enables both engines (CUDA only; other platforms ignore it):

```yaml
# deploy config, Code2Wav stage extra
token2wav_trt: true
token2wav_trt_dtype: fp16   # optional, fp16 (default) | fp32
token2wav_trt_max_batch: 16 # optional, = 8 requests x CFG
```

or `MINICPMO_TOKEN2WAV_TRT=1` in the launching shell. First start exports
ONNX and builds the engine (~1 min on H100); subsequent starts reuse the
plan cached under `$MINICPMO_TOKEN2WAV_TRT_CACHE` (default
`$HF_HOME/vllm_omni/token2wav_trt`).

## Performance

Daily-Omni open-ended QA, text+audio (`use_tts_template=true`), 10 requests,
concurrency 2, 2xH100 (`minicpmo_4_5_2gpu.yaml`), warm server (second run):

| Metric | torch | TRT | Δ |
| --- | ---: | ---: | ---: |
| Median audio RTF | 0.39 | **0.28** | **-28%** |
| Median audio TTFP | 1411 ms | **1156 ms** | -18% |
| Median E2EL | 3247 ms | **2167 ms** | -33% |
| Audio throughput | 5.10 s/s | **7.34 s/s** | **+44%** |
| Median TTFT (text) | 863 ms | 876 ms | unchanged |
| Streaming continuity OK | 100% | 100% | — |

Standalone vocoder microbench (bs=1, 3.6 s clip): DiT step RTF 0.143 → 0.062
(**2.31x**), consistent with the upstream PR's ~2x on L20. campplus:
50.8 ms → 3.6 ms per call (runs once per unique prompt WAV, so it shows up
in voice-clone first-chunk latency rather than this default-voice benchmark).


Sy0307:
Look nice. I wonder whether we can support TRT backend as general vocoder backend, because I used to find out TRT backend can take more performance benefit than current path. Your thought? Also cc @linyueqian

amy-why-3459:
Overall this is a clean, well-scoped change. Swapping TRT in at
`BatchedToken2Wav._estimator_step` instead of patching the pip `cosyvoice2.flow.flow_matching`
(as the upstream PR does) is the right call: minimal blast radius, the torch
encoder/HiFT/cache layout stays untouched, and the pip package stays pristine.
The serving-level numbers (E2EL -33%, audio throughput +44%) are convincing.
I'm inclined to **approve**, with a few items I'd like clarified/hardened first.
### Strengths
- **Correct integration layer.** TRT is a single pre-branch in `_estimator_step`; when the toggle is off there is zero overhead and the original path is byte-for-byte unchanged.
- **Strongly-typed precision handled correctly.** Precision comes from the fp16 ONNX graph itself (TRT >= 10/11 removed `BuilderFlag.FP16` / `EXPLICIT_BATCH`), and `copy.deepcopy(dit)` ensures the live torch weights keep their original dtype.
- **Same-stream, sync-free execution.** Running on the caller's current CUDA stream via `execute_async_v3(stream.cuda_stream)` avoids the private-stream + `synchronize()` double round-trip. The comment about a private stream producing silent garbage audio is a nice, hard-won note.
- **Sensible platform handling.** Non-CUDA platforms ignore the toggle instead of raising, which is correct for a shared deploy config.
- **Dependency cleanup.** Dropping `stepaudio2-minicpmo` and unifying CUDA/NPU on the in-tree adapter removes an external dependency.
- **campplus reuse.** Reusing CosyVoice3's `get_campplus_trt` (process-wide cache + internal lock) matches the `[T, 80] -> [1, 192]` contract.
### Blocking / please clarify
**1. Engine cache key does not include GPU architecture (important).**
The plan filename is keyed on `dtype/depth/batch/trt_version`, but a serialized TRT
plan is SM/architecture-specific. When the same `$HF_HOME` cache is shared across
machines with different SMs (e.g. an H100/L20 mix — both are referenced in the PR
description), deserialization can fail or misbehave. Please add
`torch.cuda.get_device_capability()` (or the device name) to the plan key.
**2. `max_batch` overflow is silently opaque (important).**
The engine's optimization profile caps at `max_batch` (default 16 = 8 requests x
CFG). If `x.shape[0]` exceeds it at runtime, `set_input_shape` / `execute_async_v3`
just fails with `"TRT DiT execution failed"`, which is hard to diagnose. Please add
an explicit bound check in `TrtDiTStepper.step` (batch vs `max_batch`, chunk vs
`max_len`, cache_len vs `max_cache_len`) with a readable error, or fall back to the
torch path.
### Non-blocking suggestions
**3. First-build concurrency race.**
When multiple ranks/processes hit an unbuilt plan simultaneously, each exports ONNX
and builds the engine, all writing to the same fixed `.tmp` path before `os.replace`.
`os.replace` is atomic, but concurrent builds waste ~1 min x N and can clobber each
other's `.tmp`. Consider a pid/uuid-unique temp name or a file lock.
**4. Dead `dtype` parameter in `convert_onnx_to_trt`.**
The function takes `dtype` and immediately `del dtype`s it (precision comes from the
ONNX graph). This is correct behavior, but keeping the parameter — and still passing
it from `build_dit_trt_stepper` — is misleading. Please remove it or add a one-line
note that it's placeholder/logging-only.
**5. dtype consistency in `enable_trt_spk_embedding`.**
The TRT branch of `_forward_spk_embedding` returns `spk_model(spk_feat).to(self.device)`
without a `.to(dtype)`. The onnxruntime path previously produced an fp32 tensor.
The cosine similarity of 0.999997 shows the values are fine, but please confirm the
output dtype matches what the downstream flow conditioning expects, to avoid an
implicit cast outside autocast.
**6. campplus path resolution.**
`enable_trt_spk_embedding` uses `f"{self.model_path}/campplus.onnx"`, consistent with
the onnxruntime loader. Please confirm `model_path` is already a local directory at
that point (a hub repo id would make `open()` fail inside the TRT wrapper) — worth
one voice-clone first-chunk run to verify.

yuekaizhang:
Thanks for the thorough review! Items 1–4 are addressed in 5b63372a; 5–6 are clarified below.

**1. GPU architecture in the plan key — fixed.** Plan filenames now embed
`torch.cuda.get_device_capability()` (e.g. `...b16.sm90.trt11.1.0.106.plan`),
so a cache dir shared across heterogeneous machines builds one plan per arch
instead of reusing a foreign one.

**2. `max_batch` overflow — fixed.** `TrtDiTStepper.step` now validates batch /
chunk frames / att-cache frames against the optimization-profile bounds the
engine was built with, and raises a `ValueError` that names
`token2wav_trt_max_batch` (and notes CFG is included in the batch dim) instead
of an opaque `execute_async_v3` failure. Verified: batch=34 against a b16
engine raises the readable error.

**3. First-build race — fixed.** Both the ONNX export and the serialized plan
now write to a pid-unique temp file followed by an atomic `os.replace`, so
concurrent first builds can no longer clobber each other's partial writes
(the duplicate-build cost for simultaneous cold starts remains, but the result
is always a complete file; last-writer-wins with identical content).

**4. Dead `dtype` parameter — removed.** `convert_onnx_to_trt` no longer takes
`dtype`; the docstring states that precision is strongly typed from the
exported ONNX graph.

**5. spk-embedding output dtype — no change needed.** `CampplusTRT.__call__`
allocates its output explicitly as `torch.float32`
(`speaker_embedding_trt.py`), same as the onnxruntime path, so downstream flow
conditioning sees an identical fp32 tensor either way.

**6. campplus path — always local at that point.** In the Code2Wav path,
`_build_backend` runs after `_resolve_model_dir()` (hub ids are materialized
to a local snapshot before any asset access, #5442), and the core's
`model_path` is the local `assets/token2wav` directory string. A hub repo id
can never reach `enable_trt_spk_embedding`.

Re-verified after the changes: engine rebuilds under the new sm-keyed name,
vocoder output is ASR-identical to torch, and the standalone DiT step speedup
is 2.5x.

yuekaizhang:
> Look nice. I wonder whether we can support TRT backend as general vocoder backend, because I used to find out TRT backend can take more performance benefit than current path. Your thought? Also cc @linyueqian

Thanks, I agree with this plan. There are two separate TRT integrations already in the repo: CosyVoice3 flow-estimator + Campplus, alongside the DiT stepper from this PR. Around 80% of their build, cache and execution logic overlaps. We can consolidate these shared codes into a unified file like vllm_omni/utils/trt.py.

Regarding the general vocoder backend, I plan to benchmark HiFi-GAN and other vocoders to check if TensorRT can bring tangible gains, then submit an RFC afterwards.

cc: @Sy0307 @linyueqian

amy-why-3459 (approved):

amy-why-3459:
Thanks, LGTM.

Issue #4237, opened 2026-06-07
[Feature]: Optimize  KV Cache Transfer Latency for Hunyuan Image AR+DiT


Bounty-hunter:
## Problem Statement

Hunyuan Image-3 uses a two-stage architecture (AR stage generates KV cache → DiT stage receives and renders). Currently, receiving KV cache on the DiT side incurs a **~200ms blocking delay**.

**Profiling breakdown (inside `receive_kv_cache_for_request`):**

<img width="988" height="464" alt="Image" src="https://github.com/user-attachments/assets/d3aed250-378f-4a5b-8356-c608423043fc" />

| Component | Time | Description |
|-----------|------|-------------|
| `connector.get()` | **102ms** | ZMQ handshake + RDMA transfer + GPU sync |
| `_clone_received_payload_tensors()` | **40ms** | CPU-side tensor.clone() |
| `aten::copy_` (device transfer) | **51ms** | CPU→GPU tensor copy |
| **Total** | **~193ms** | Complete receive pipeline |

> **Note:** The 102ms in `connector.get()` is expected to reduce to **~20ms** after properly configuring RDMA (currently falling back to TCP).

---

## Architecture Layers
```
DiffusionModelRunner (caller)
    ↓ synchronous call
OmniKVTransferManager (high-level wrapper)
    receive_kv_cache_for_request()
    - poll connector.get()
    - deserialize, clone, merge multi-rank data
    - device transfer (CPU→GPU)
    ↓ calls low-level transport
Connector (low-level transport primitive)
    .get() — synchronous blocking, ZMQ req/resp cannot be split
```
---

### Phase 1: Remove Redundant Clone (P0) — 40ms Guaranteed Gain

**Modification:** `kv_transfer_manager.py:1080`
```python
# Before:
data = KVCacheTransferData.from_bytes(memoryview(buf_tensor.numpy()))
data = self._clone_received_payload_tensors(data)  # 40ms
raw_data.release()

# After:
data = KVCacheTransferData.from_bytes(memoryview(buf_tensor.numpy()))
raw_data.release()  # ManagedBuffer is one-time use, no clone needed
```

**Gain:** 40ms → 0ms, zero risk
---

### Phase 2: Isolated CUDA Stream (P1) — 51ms Overlappable

**Problem:** CPU→GPU copy executes on default stream, blocking main thread for 51ms

**Modification:** `kv_transfer_manager.py`
```python
# In __init__:
[...]

PR #5550, opened 2026-07-29
[Diffusion] Add Worker contracts and RPC plumbing for Scheduler-managed paged KV cache

Acerak01-fy:
## Purpose

This PR adds the Worker-facing request DTO and RPC plumbing for Scheduler-managed diffusion KV cache. It is the W0 follow-up to merged PR #5541 and the W0 control-plane slice of [RFC #5244](https://github.com/vllm-project/vllm-omni/issues/5244).

The scope is intentionally narrow: this PR defines how a Scheduler allocation snapshot can travel with a newly scheduled diffusion request, while preserving the existing dense execution path. It does not yet create the Scheduler allocator, initialize physical KV pages, install Worker BlockTables, or consume the block IDs in model attention.

## Ownership and data flow

PR #5541 introduced Scheduler-owned logical KV state:

```text
DiffusionKVRequest / DiffusionKVContext
  owner: SchedulerRequestState
  lifetime: mutable across scheduling and native KVCacheManager operations
  rule: never sent to a Worker
```

This PR introduces the corresponding request-time wire snapshot:

```text
DiffusionKVMetadata
  owner: one newly scheduled request envelope
  lifetime: one allocation snapshot
  contents: allocation generation, sequence/context logical bounds,
            native cache-group block IDs
```

The transport path is:

```text
[W1 Scheduler allocation producer]
  -> DiffusionKVMetadata
  -> NewRequestData.diffusion_kv_metadata
  -> MultiprocDiffusionExecutor
       request mode: optional fourth execute_model argument
       batch/step: metadata remains inside DiffusionSchedulerOutput
  -> Diffusion Worker / ModelRunner boundary
  -> [W1 native BlockTable installer]
```

`request.prepared_layout` remains a separate model-owned Worker payload. The generic KV transport does not derive block IDs from it and does not send mutable `DiffusionKVRequest` objects to Workers.

## W0 capability boundary

Metadata is optional in this PR.

Merged PR #5541 stores authoritative `DiffusionKVRequest` objects in `SchedulerRequestState`, removes them from the request sent to the Worker, and does not yet allocate blocks or produce `DiffusionKVMetadata`. Requiring metadata before those producer/installer capabilities exist would make `paged_scheduler` unusable or require fabricated block allocations.

W0 therefore treats these dataclasses as trusted internal DTOs and performs only checks backed by information available at the current execution boundary:

- dense mode rejects unexpected diffusion KV metadata;
- paged mode accepts metadata when supplied but does not require it yet;
[...]

PR #5541, opened 2026-07-29
[Diffusion] Prepare HunyuanImage3 for Scheduler-managed paged KV cache

zwhzzz0821:
## Purpose

This PR introduces the request-side control-plane contracts needed for future Scheduler-managed diffusion paged KV cache support described in RFC #5244, with HunyuanImage3 as the first integration.

The PR establishes three boundaries:

- model preprocessing produces the exact logical KV shape of a request;
- the Scheduler owns persistent logical KV request state; and
- physical cache geometry, block allocation, Worker BlockTables, and paged
  attention remain native vLLM/follow-up responsibilities.

It does not enable paged attention, allocate KV blocks, or change the existing dense Hunyuan execution path.

### Design and architecture

Diagram labels:

- `[EXISTING]`: existing behavior retained;
- `[EXTENDED IN THIS PR]`: an existing object or step with a new field/path;
- `[NEW IN THIS PR]`: a new request-side contract or helper; and
- `[FOLLOW-UP]`: Scheduler allocation or Worker data-plane work.

```text
                         Request preprocessing
                                  |
                                  v
              [EXISTING] raw OmniDiffusionRequest
                                  |
                                  v
           [EXTENDED IN THIS PR] Hunyuan pre_process_func
                                  |
                   +--------------+--------------+
                   |                             |
                   v                             v
    [NEW IN THIS PR] prepare_hunyuan_layout()    |
                   |                             |
                   v                             |
    [NEW IN THIS PR] HunyuanPreparedLayout       |
    tokens / image geometry / RoPE layout        |
                   |                             |
                   |       [NEW IN THIS PR]
                   |       build_hunyuan_diffusion_kv_requests()
                   |                             |
                   |                             v
                   |       one DiffusionKVRequest per CFG row
                   |         primary: [prefix | target]
                   |         kv_contexts=() for Hunyuan
                   |                             |
                   +--------------+--------------+
                                  |
                                  v
              [EXTENDED IN THIS PR] BaseScheduler
                                  |
                   +--------------+--------------+
                   |                             |
                   v                             v
    SchedulerRequestState                 NewRequestData
      .diffusion_kv_requests                .req.prepared_layout
      Scheduler-only                       Worker-bound
                   |                             |
                   v                             v
       [FOLLOW-UP] DiffusionKVCacheManager  Hunyuan Worker reuses
                   |                       prepared tokens/layout
                   v
       native vLLM KVCacheManager
                   |
                   v
       [FOLLOW-UP] DiffusionKVMetadata
[...]

PR #5640, opened 2026-07-31
[Core] Make KV prefetch H2D copies event-driven

EchoHayate:
## Purpose

The existing asynchronous diffusion KV prefetch runs on a background thread and a dedicated device stream, but its H2D tensor moves are synchronous. The prefetch future therefore cannot complete until the copy finishes, preventing the consumer from expressing the dependency as a device-stream wait.

This change:

- uses non-blocking H2D copies for prefetched KV tensors;
- records a completion event on the copy stream and lets the background future return immediately after enqueue;
- makes the consumer stream wait on that event and calls `record_stream` on destination tensors;
- retains source tensors and connector pool buffers until the copy event completes;
- synchronizes and releases resources safely for stale futures, failures, misses, and shutdown;
- adds a reproducible manager-level H2D benchmark.

Related to #4237.

## Performance

Review follow-up in `279830b3`: the benchmark now starts timing before
`_prefetch_executor.submit(...)` and waits for `future.result()`, so
`event_future_ready` includes executor submission, background-thread
scheduling, and the event-based H2D enqueue. The earlier `0.090 ms` result
called `_prefetch_payload(...)` directly and therefore excluded executor
scheduling; that number is superseded and is no longer used as the current
performance claim.

Refreshed A100 measurement:

- GPU: 1x NVIDIA A100 80GB PCIe
- PyTorch: 2.11.0
- CUDA runtime: 13.1
- vLLM: 0.26.0
- Transfer: 256 MiB pinned host tensor
- Warmup/measured iterations: 3/20

```bash
PYTHONPATH=. python benchmarks/distributed/omni_connectors/benchmark_async_prefetch_h2d.py   --device cuda:0   --size-mib 256   --warmup 3   --iterations 20
```

| Future-ready behavior | Median | P95 | Range |
|---|---:|---:|---:|
| Blocking until copy event | 10.730 ms | 10.868 ms | 10.704-10.868 ms |
| Executor-submitted event return | 0.219 ms | 0.342 ms | 0.200-0.342 ms |

Including executor submission and background-thread scheduling, median
future-ready latency is 49.00x lower (97.96% reduction). The copy event was
still incomplete at return in 23/23 runs. Peak device memory was 256.0 MiB.

This benchmark measures future availability, not end-to-end request
throughput. It separately reports the time until the background future is
ready and the time until the same copy event completes.

## Test Plan

- Unit tests cover event recording, consumer `wait_event`, destination `record_stream`, deferred source release, stale futures, misses/fallback, event failures, copy failures, and shutdown cleanup.
- Real A100 integration used a 16 MiB serialized `ManagedBuffer`, delayed the copy stream, verified that prefetch and consume returned while the event was incomplete, then verified data integrity after release.
- Compared the broad connector suite with an unmodified baseline checkout. Both have the same 18 macOS `/dev/shm` failures and 2 CPU cleanup teardown errors; the changed branch adds 12 passing tests.

**vLLM Version:** 0.26.0

**vLLM-Omni Commit:** `279830b382b5d5e1044ad814b0a34ad7700a5b35`

## Test Result

```text
55 passed, 15 warnings in 0.87s
Ruff check: passed
Ruff format --check: passed
compileall: passed
git diff --check: passed

Review follow-up:
34 passed, 16 warnings in 1.13s
Focused full-path consume/apply regression: 1 passed

Broad connector suite (changed): 18 failed, 200 passed, 28 skipped, 2 errors
Broad connector suite (baseline): 18 failed, 188 passed, 28 skipped, 2 errors
Failure/error set is identical and environment-specific (/dev/shm unavailable on macOS; CPU fixture calls empty_cache).

A100 integration:
ASYNC_KV_EVENT_INTEGRATION_OK bytes=16777400 prefetch_submit_ms=1.251 consume_submit_ms=0.197 shape=(2048, 2048)
```

### DIFF
--- a/benchmarks/distributed/omni_connectors/benchmark_async_prefetch_h2d.py
+++ b/benchmarks/distributed/omni_connectors/benchmark_async_prefetch_h2d.py
@@ -48,6 +48,7 @@ def main() -> None:
         OmniKVCacheConfig(need_recv_cache=True),
         async_prefetch=True,
     )
+    assert manager._prefetch_executor is not None
 
     def fake_receive(
         request_id,
@@ -77,12 +78,14 @@ def main() -> None:
         torch.accelerator.synchronize()
         torch.accelerator.reset_peak_memory_stats()
         started = time.perf_counter()
-        result = manager._prefetch_payload(
+        future = manager._prefetch_executor.submit(
+            manager._prefetch_payload,
             f"bench-{iteration}",
             None,
             device,
         )
-        submitted = time.perf_counter()
+        result = future.result()
+        future_ready = time.perf_counter()
         event_incomplete += int(not result.ready_event.query())
         result.ready_event.synchronize()
         completed = time.perf_counter()
@@ -90,7 +93,7 @@ def main() -> None:
         manager._release_prefetch_result(result)
 
         if iteration >= args.warmup:
-            submit_ms.append((submitted - started) * 1000)
+            submit_ms.append((future_ready - started) * 1000)
             blocking_ready_ms.append((completed - started) * 1000)
             peak_memory_mib.append(peak_mib)
         del result
--- a/benchmarks/distributed/omni_connectors/benchmark_async_prefetch_h2d.py
+++ b/benchmarks/distributed/omni_connectors/benchmark_async_prefetch_h2d.py
@@ -1,0 +1,109 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+
+"""Measure future-ready latency for asynchronous KV prefetch H2D copies."""
+
+import argparse
+import statistics
+import time
+
+import torch
+
+from vllm_omni.distributed.omni_connectors.kv_transfer_manager import (
+    OmniKVCacheConfig,
+    OmniKVTransferManager,
+)
+
+
+def _percentile(values: list[float], fraction: float) -> float:
+    ordered = sorted(values)
+    index = min(int(len(ordered) * fraction), len(ordered) - 1)
+    return ordered[index]
+
+
+def _format_metrics(name: str, values: list[float]) -> str:
+    return (
+        f"{name}: median={statistics.median(values):.3f} ms, "
+        f"p95={_percentile(values, 0.95):.3f} ms, "
+        f"range=[{min(values):.3f}, {max(values):.3f}] ms"
+    )
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--device", default="cuda:0")
+    parser.add_argument("--size-mib", type=int, default=256)
+    parser.add_argument("--warmup", type=int, default=3)
+    parser.add_argument("--iterations", type=int, default=20)
+    args = parser.parse_args()
+
+    device = torch.device(args.device)
+    torch.accelerator.set_device_index(device.index)
+    element_size = torch.empty((), dtype=torch.float32).element_size()
+    element_count = args.size_mib * 1024 * 1024 // element_size
+    source = torch.empty(element_count, dtype=torch.float32, pin_memory=True)
+    source.normal_()
+
+    manager = OmniKVTransferManager(
+        OmniKVCacheConfig(need_recv_cache=True),
+        async_prefetch=True,
+    )
+
+    def fake_receive(
+        request_id,
+        target_device=None,
+        *,
+        sender_info=None,
+        non_blocking_copy=False,
+        copy_stream=None,
+        deferred_copy_sources=None,
+        deferred_pool_buffers=None,
+    ):
+        assert non_blocking_copy
+        assert copy_stream is not None
+        assert deferred_copy_sources is not None
+        assert deferred_pool_buffers is not None
+        destination = source.to(target_device, non_blocking=True)
+        deferred_copy_sources.append(source)
+        return {"layer_blocks": {"key_cache": [destination], "value_cache": []}}, source.nbytes
+
+    manager.receive_kv_cache_for_request = fake_receive
+    submit_ms: list[float] = []
+    blocking_ready_ms: list[float] = []
+    peak_memory_mib: list[float] = []
+    event_incomplete = 0
+
+    for iteration in range(args.warmup + args.iterations):
+        torch.accelerator.synchronize()
+        torch.accelerator.reset_peak_memory_stats()
+        started = time.perf_counter()
+        result = manager._prefetch_payload(
+            f"bench-{iteration}",
+            None,
+            device,
+        )
+        submitted = time.perf_counter()
+        event_incomplete += int(not result.ready_event.query())
+        result.ready_event.synchronize()
+        completed = time.perf_counter()
+        peak_mib = torch.accelerator.max_memory_allocated() / (1024 * 1024)
+        manager._release_prefetch_result(result)
+
+        if iteration >= args.warmup:
+            submit_ms.append((submitted - started) * 1000)
+            blocking_ready_ms.append((completed - started) * 1000)
+            peak_memory_mib.append(peak_mib)
+        del result
+
+    manager.shutdown_prefetch()
+    print(f"device={device}, size={args.size_mib} MiB, warmup={args.warmup}, iterations={args.iterations}")
+    print(_format_metrics("event_future_ready", submit_ms))
+    print(_format_metrics("blocking_future_ready", blocking_ready_ms))
+    print(
+        f"event_incomplete_at_return={event_incomplete}/{args.warmup + args.iterations}, "
+        f"peak_memory={max(peak_memory_mib):.1f} MiB"
+    )
+
+
+if __name__ == "__main__":
+    main()
--- a/tests/distributed/omni_connectors/test_kv_async_prefetch.py
+++ b/tests/distributed/omni_connectors/test_kv_async_prefetch.py
@@ -358,6 +358,41 @@ def test_consume_and_distribute_attaches_prefetched_payload():
     assert req.past_key_values is not None
 
 
+def test_deserialization_failure_after_get_is_non_retryable(monkeypatch):
+    _, receiver, connector = _make_sender_receiver()
+    connector.get = lambda **_kwargs: (b"consumed-payload", 16)
+
+    def fail_deserialization(_payload):
+        raise RuntimeError("decode failed")
+
+    monkeypatch.setattr(
+        "vllm_omni.distributed.omni_connectors.kv_transfer_manager.KVCacheTransferData.from_bytes",
+        fail_deserialization,
+    )
+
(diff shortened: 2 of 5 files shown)


hsliuustc0106:
@Bounty-hunter @BLANKETusers PTAL

benchmarks/distributed/omni_connectors/benchmark_async_prefetch_h2d.py
```diff
@@ -0,0 +1,109 @@
...
+
+    for iteration in range(args.warmup + args.iterations):
+        torch.accelerator.synchronize()
+        torch.accelerator.reset_peak_memory_stats()
+        started = time.perf_counter()
+        result = manager._prefetch_payload(
+            f"bench-{iteration}",
+            None,
+            device,
+        )
```
yuanheng-zhao: `OmniKVTransferManager._prefetch_executor.submit` to submit `_prefetch_payload` to be executed, rather than calling `OmniKVTransferManager._prefetch_payload` directly. This might affect cost metrics. Please correct me if I understand wrong

tests/distributed/omni_connectors/test_kv_async_prefetch.py
```diff
@@ -354,3 +360,339 @@ def test_consume_then_apply_attaches_payload():
     receiver._record_stream_for_prefetched(data)
```
BLANKETusers: **🟠 P1: `_record_stream_for_prefetched` was removed from `consume_and_distribute_kv_cache` because `consume_prefetched_kv` now handles it internally (line 1353). However, `test_consume_then_apply_attaches_payload` still directly calls `receiver._record_stream_for_prefetched(data)`, which is inconsistent with the new logic.

Suggested fix: update the test to go through `consume_prefetched_kv` full path instead of manually calling `_record_stream_for_prefetched`, to verify the internal `record_stream` logic works correctly.

BLANKETusers (commented):
One minor inconsistency found in the test.

benchmarks/distributed/omni_connectors/benchmark_async_prefetch_h2d.py
EchoHayate: Your understanding is correct. The previous benchmark called `_prefetch_payload(...)` directly, so its future-ready metric excluded executor submission and background-thread scheduling. Fixed in `279830b3`: timing now starts before `_prefetch_executor.submit(...)` and stops after `future.result()`, and the README documents that scope. I also marked the previous `0.090 ms` A100 number as superseded in the PR description rather than presenting it as current evidence; the corrected benchmark needs a fresh GPU rerun.

tests/distributed/omni_connectors/test_kv_async_prefetch.py
EchoHayate: Fixed in `279830b3`. `test_consume_then_apply_attaches_payload` is now `test_consume_and_distribute_attaches_prefetched_payload` and calls the production `consume_and_distribute_kv_cache()` path directly, without manually invoking `_record_stream_for_prefetched`. The focused regression passed, and the complete async-prefetch test file reports `34 passed, 16 warnings`.

AbelSara:
Could you supplement the end-to-end throughput comparison on hunyuan-image_it2i?

benchmarks/distributed/omni_connectors/benchmark_async_prefetch_h2d.py
EchoHayate: Refreshed the corrected benchmark on 1x A100 80GB with PyTorch 2.11.0 / CUDA 13.1 / vLLM 0.26.0. With executor submission and background-thread scheduling included, `event_future_ready` is median `0.219 ms` (p95 `0.342 ms`) versus blocking median `10.730 ms` (p95 `10.868 ms`), a `49.00x` median reduction (`97.96%`). The copy event was incomplete at return in `23/23` runs and peak device memory remained `256.0 MiB`. The PR description now contains the refreshed table and keeps the earlier `0.090 ms` result explicitly superseded.

EchoHayate:
Thanks for the suggestion. I completed a controlled end-to-end `hunyuan-image_it2i` comparison.

### Setup

- Baseline: merge-base `499d930fe599486a55f1ccc6fb695983aef52ef9`
- PR head: `fb4ac9f881b3cb273c217886224a96492bb923ad`
- 8x A100 80GB PCIe
- AR TP4 on GPUs 0-3 + DiT TP4 on GPUs 4-7
- 12 requests per run, client concurrency 4, 8 denoise steps, fixed input / seed 42 / guidance 1.0
- Run order: baseline-1 -> head-1 -> head-2 -> baseline-2
- Per-layer diffusion profiling was disabled for the measurement

The validation image does not include a usable Mooncake runtime, so this used the single-node `SharedMemoryConnector`. The same A100-only validation adaptations were applied byte-identically to both sides: AR VAE spatial tiling, a fixed 512 MiB AR KV-cache budget with `max_model_len=8192`, one startup-only allocator cache release, and an SHM TP-rank fix so each TP rank uses a distinct key. Therefore, this is an A100 validation result rather than a replacement for the H100/Mooncake nightly topology.

### Results

| Run | Throughput (req/s) | Mean latency (s) | P99 (s) | Completed |
| baseline-1 | 0.046289 | 78.495 | 92.428 | 12/12 |
| head-1 | 0.046778 | 77.520 | 93.844 | 12/12 |
| head-2 | 0.045943 | 78.910 | 93.118 | 12/12 |
| baseline-2 | 0.045999 | 78.664 | 94.480 | 12/12 |

Aggregated means:

- Baseline: `0.046144 req/s`, mean latency `78.580 s`, P99 `93.454 s`
- PR head: `0.046361 req/s`, mean latency `78.215 s`, P99 `93.481 s`
- Delta: throughput `+0.469%`, mean latency `-0.464%`, P99 `+0.029%`

All four runs completed with zero failed requests, zero CUDA OOMs, and zero KV receive timeouts. The `+0.469%` throughput difference is smaller than the run-to-run variation, so I do **not** consider it a statistically meaningful end-to-end speedup. The defensible conclusion is that this workload is effectively throughput/latency neutral and shows no regression.

The corrected 256 MiB manager benchmark still shows the intended local effect (`10.730 ms` blocking-ready vs `0.219 ms` executor-submitted event-ready, `49.00x` earlier future availability), but the full image pipeline is dominated by AR, KV serialization/transport, DiT, and VAE work, so that local improvement is not measurable as an end-to-end throughput gain in this A100 setup.

hsliuustc0106:
I reproduced the targeted tests with the local vllm-0.26 environment:

- `vllm 0.26.0`
- `torch 2.11.0+cu130`
- CUDA-visible host

Command:
```
python -m pytest tests/distributed/omni_connectors/test_kv_async_prefetch.py tests/distributed/omni_connectors/test_kv_flow.py -q
```

Result: **56 passed, 1 failed, 15 warnings**.

The failure is:
```
tests/distributed/omni_connectors/test_kv_async_prefetch.py::test_consumed_prefetch_releases_pool_buffer_after_event_completes
```

The test installs a fake event exposing only `completed` and `query()`, but `consume_prefetched_kv()` uses the real current stream's `wait_event()`. On this host that raises `AttributeError: '_Event' object has no attribute 'wait'`; the cleanup path then also expects `synchronize()`. The test is marked `cpu`, but that marker does not stub `current_omni_platform.current_stream()`, and the test still fails when CUDA is disabled because the real current-stream lookup attempts CUDA initialization.

As a control, excluding only this test gives **56 passed**. This looks like a test fixture/platform-isolation issue rather than a production-path failure. The fixture could monkeypatch `current_omni_platform.current_stream` to a fake compute stream (as the neighboring wait/record test does), or provide `wait()` and `synchronize()` on the fake event.

Bounty-hunter:
As noted in the comment, the +0.469% throughput improvement is within run-to-run variance and is not statistically significant. Given the marginal performance benefit, the additional complexity introduced by this approach may not be justified.

We are currently implementing scheduler-allocated KV blocks for the diffusion stage (#5550/#5541). Once that is in place, prefetch can write directly into pre-allocated device blocks, which is closer to the upstream vLLM design and avoids the need for this additional machinery. I suggest aligning with that approach. It may be better to consider refactoring the prefetch-related components directly after diffusion kv manager lands, rather than investing further in the current implementation.

EchoHayate:
@hsliuustc0106 Thanks for reproducing this. Your diagnosis is correct: this was a test fixture/platform-isolation issue, not a production-path failure.

Fixed in `8fc99258`. The CPU-marked deferred-release test now stubs `current_omni_platform.current_stream()` with a fake compute stream, exercises the production `consume_prefetched_kv()` path, and asserts that `wait_event()` receives the ready event. It no longer depends on a real CUDA stream interpreting the fake event.

I reproduced the failure before the fix on an A100-visible host with:

- `vllm 0.26.0`
- `torch 2.11.0+cu131`
- Python 3.11

Before the fix, the focused test failed with the same missing `_Event.wait()` error, followed by the missing `synchronize()` cleanup error.

After the fix:

```text
python -m pytest   tests/distributed/omni_connectors/test_kv_async_prefetch.py   tests/distributed/omni_connectors/test_kv_flow.py -q

57 passed, 18 warnings in 0.95s
```

Additional checks:

```text
python -m py_compile tests/distributed/omni_connectors/test_kv_async_prefetch.py  passed
ruff check tests/distributed/omni_connectors/test_kv_async_prefetch.py          passed
ruff format --check tests/distributed/omni_connectors/test_kv_async_prefetch.py passed
git diff --check                                                               passed
```

Only the test fixture changed; production code is unchanged.

EchoHayate:
@Bounty-hunter Thanks for the architectural context. I agree that the measured `+0.469%` throughput difference is within run-to-run variance and is not a statistically reliable end-to-end speedup; I am not treating it as an E2E performance claim.

I also agree that scheduler-owned paged KV is the preferred long-term direction: the receiver Scheduler should allocate destination device blocks first, and the transfer path should write directly into those blocks, eventually enabling missing-only transfer, cache-aware admission, and page-level reuse.

This PR is intentionally limited to the current dense legacy transfer path. In that path, the receiver retrieves a request-level payload and materializes model-specific dense KV tensors before DiT consumes them. The change only makes that existing H2D handoff event-driven and preserves source/pool-buffer lifetimes; it does not change cache ownership, allocation, or compete with the #5550/#5541 design.

I will not expand this PR into a separate paged-KV architecture. I will keep the scope bounded and leave whether this interim dense-path improvement is worth merging, deferring, or closing to the maintainers as #5550/#5541 progress.

EchoHayate:
Closing this implementation in favor of the scheduler-owned paged-KV direction discussed above. The dependent foundation PRs #5541 and #5550 have since merged, while the controlled 8xA100 HunyuanImage3 A/B showed the current dense-path patch was E2E-neutral within run-to-run variance (+0.469% throughput). The event/lifetime analysis and benchmark remain useful historical evidence, but I do not plan to keep rebasing this parallel mechanism. Any follow-up will target direct writes into scheduler-allocated device blocks instead of reviving this branch.

PR #5645, opened 2026-07-31
[Bugfix][Diffusion] Drop outputs for cancelled async waiters

EchoHayate:
## Purpose

Refresh the cancelled async-output cleanup fix on current `main`.

When an async diffusion caller times out or is cancelled, its registered
`Future` becomes done/cancelled. If `OUTPUT_READY` arrives later, the result
pump removes that waiter but currently treats it the same as "no waiter was
ever registered" and recreates an entry in `_completed_outputs`.

That leaves stale per-request output state after cancellation. The same state
transition exists for both single-request outputs and request-level batch
splits.

## Reproduction and root cause

The focused regressions exercise both affected paths:

```bash
pytest -q \
  tests/diffusion/test_result_pump.py::TestResultPumpDispatch::test_output_ready_drops_result_after_waiter_cancelled \
  tests/diffusion/test_result_pump.py::TestBatchSplitDelivery::test_batch_output_drops_cancelled_waiter_and_resolves_active_waiter
```

Before the fix:

- a cancelled single-request waiter was removed and its late result was
  reinserted into `_completed_outputs`;
- in a mixed batch, the cancelled member was reinserted while the active
  member resolved normally.

The root cause was collapsing three lifecycle states into one
`pending is not None and not pending.done()` condition:

1. no waiter was registered -> cache the legitimate early output;
2. a live waiter exists -> resolve it;
3. a waiter existed but is already done/cancelled -> drop the late output.

## Current-main context

Current `main` already contains:

- #5983, which prevents the result-pump thread from crashing when cancellation
  races with `set_result()` / `set_exception()`;
- #6023, which makes early batch-output registration atomic.

Those fixes are preserved. This PR only closes the remaining lifecycle gap.
Successful SHM payloads are still unpacked before the delivery decision, so
dropping a late output does not skip shared-memory cleanup.

## Fix

- Do not recreate `_completed_outputs` entries when a known single-request
  waiter is already done/cancelled.
- Apply the same rule to per-request futures produced by a batch output.
- Preserve early-output caching when no waiter has ever registered.
- Preserve `try_set_result()` / `try_set_exception()` handling for the
  check-then-set cancellation race.

## Validation

Tested base: `main@f67cc52f7449d5f5dbb9d7790c7a049fa6aa2ba1`

Head: `6e46fc484bb57c97b89dd64021c74ef6ac03491a`

```text
tests/diffusion/test_result_pump.py

93 passed, 18 warnings in 5.90s
```

Also passed:

- changed-file pre-commit, including Ruff, mypy, SPDX, forbidden-import,
  test-marker, and `torch.cuda` checks;
- `python3 -m compileall` for both changed files;
- `git diff origin/main...HEAD --check`;
- diff-scoped static review with no P0-P2 findings.

This is an L1 CPU lifecycle/correctness fix. It requires no model weights or
GPU validation and makes no performance claim.

## AI assistance

Used Codex to help inspect the current-main lifecycle paths, refresh the
branch, run tests and static checks, and draft the PR update. I reviewed the
two-file production/test diff and remain responsible for the implementation,
claims, and follow-up.


EchoHayate:
Rebuilt this fix on current `main` (`da4a08b6`) as `6f49b67b`.

This remains a small lifecycle gap after #5983/#6023: those changes prevent `InvalidStateError` and preserve legitimate early-output delivery, while a result arriving after its waiter is already cancelled/done can still be inserted into `_completed_outputs` with no future consumer. The refreshed patch keeps SHM unpack/cleanup and early-output caching intact, but drops only those abandoned late results (including mixed active/cancelled batch output).

Validation: the two new regressions fail on current main and pass with the patch; `test_result_pump.py` plus `test_multiproc_engine_concurrency.py` are `84 passed`; changed-file pre-commit/mypy/Ruff/SPDX and all GitHub checks are green.

@Gaohan123 @SamitHuang PTAL when convenient.

EchoHayate:
Refreshed again on the vLLM 0.28 mainline (`main@94b1546c`, new head `0584ad61`). The patch remains limited to the cancelled/done-waiter late-output cleanup and its focused regressions; legitimate early outputs and SHM unpack/unlink behavior are unchanged.

Fresh validation against vLLM 0.28.0:
- `tests/diffusion/test_result_pump.py` + `tests/diffusion/test_multiproc_engine_concurrency.py`: **84 passed**
- changed-file pre-commit: passed
- `git diff --check`: passed

@Gaohan123 @SamitHuang PTAL when convenient.

EchoHayate:
## Author self-review

- Scope/invariants checked: the production diff only changes late
  `OUTPUT_READY` delivery in `MultiprocDiffusionExecutor`. It preserves the
  three distinct states: cache when no waiter has registered, resolve a live
  waiter, and drop a result only when the known waiter is already
  done/cancelled. Single-output and batch-split paths follow the same rule.
- Cleanup boundary checked: successful SHM payloads are unpacked before the
  resolve/cache/drop decision, so cancelled requests still release their
  shared-memory payload. Existing #5983 race-safe setters and #6023 atomic
  early-output behavior remain intact.
- Tests and static checks: the two focused cancellation regressions pass;
  `test_result_pump.py` plus `test_multiproc_engine_concurrency.py` are
  **93 passed, 18 warnings**. Changed-file pre-commit (Ruff, mypy, SPDX,
  forbidden imports, test markers, and related hooks), compileall, and
  `git diff origin/main...HEAD --check` pass.
- Runtime evidence: L1 CPU lifecycle coverage only; no model weights or GPU
  are required for this state-machine fix.
- Evidence and claim boundaries: this PR makes no throughput, latency, or
  model-level correctness claim. GitHub CI for this new head is still the
  authoritative package/build check.
- AI assistance: Codex assisted with current-main inspection, branch refresh,
  validation, static review, and drafting this update. I reviewed the complete
  two-file diff and take responsibility for the code and claims.

Current base: `f67cc52f7449d5f5dbb9d7790c7a049fa6aa2ba1`.
Current head reviewed: `6e46fc484bb57c97b89dd64021c74ef6ac03491a`.

@Isotr0py @princepride @SamitHuang PTAL when convenient.

PR #5414, opened 2026-07-26
[Bugfix][Connectors] Flush processor tail on terminal async-chunk sends

linyueqian:
## Purpose

Fixes #5413 (Qwen3-Omni async-chunk nightly failures after the duplex end-condition change).

Mechanism: #5383 correctly made terminal stops stop counting as segment boundaries (`is_segment_finished = not finished`), but `ChunkTransferAdapter._send_single_request` passes only `is_segment_finished` as the producer processor's flush signal (the inline comment says it itself: *"Existing processors use is_finished as a flush signal"*). A one-shot async-chunk request now reaches its terminal send with the flush signal `False`, so processors like `thinker2talker_async_chunk` take their accumulate path (`if not is_finished: …`) and never emit the accumulated tail. `meta.finished` is stamped `True` after the processor either way — so the downstream stage receives the finished marker **with the final chunk's payload missing**, which is exactly the shape of the `test_qwen3_omni_expansion.py[async_chunk]` failures (first request fails, server wedges, remaining tests get connection errors).

Fix: call the processor with `is_finished=is_segment_finished or is_finished` — flush on segment end *or* terminal — while leaving the wire's `meta.is_segment_finished` semantics from #5383 untouched (terminal still stamps it `False`, which #5380's duplex path depends on).

## Test Plan

New regression test `test_send_single_request_terminal_chunk_still_flushes_processor`: a terminal task (`is_finished=True, is_segment_finished=False`) must deliver the flush signal to the processor and stamp `meta.finished=True` / `meta.is_segment_finished=False` on the wire.

## Test Result

`pytest tests/distributed/omni_connectors/test_chunk_transfer_adapter.py` → 46 passed (45 existing + 1 new).

Could not run the Qwen3-Omni nightly e2e locally (no weights on the available box) — would appreciate a nightly re-trigger on this branch to confirm.

cc @Gaohan123 @Sy0307

---

**Update:** this PR now also carries the fix for the CI breakage that was blocking it (#5442, see the log analysis in the comments): `minicpmo_4_5_code2wav.py` resolved asset paths from the raw `model_config.model` string, which is a repo id in CI, so every MiniCPM-o e2e test died at stage-2 startup and the leaked stage procs starved the qwen3 tests into startup OOMs. Non-directory model refs now resolve through `snapshot_download` (assets only) first. Unit tests added; `test_code2wav_model_dir.py` + adapter suite → 49 passed.

Fixes #5442

Issue #5395, opened 2026-07-25
[Bug] Qwen3-Omni talker garbles one-word answers (~13-27%/request, seed-deterministic, pre-existing on main); Whisper-small judge inflates test_one_word_prompt_001 failures ~2x


tzhouam:
### Summary

`tests/e2e/online_serving/test_qwen3_omni_expansion.py::test_one_word_prompt_001` fails intermittently in nightly CI (e.g. `vllm-omni-rebase` build 2861, `Omni · Function Test with H100`, 10/10 retry attempts failed). We investigated this end-to-end — including 30-sample serial single-request measurements on both `dev/vllm-align` (vLLM `dcfebf93f4`) and `main` (vLLM v0.25.0), audio dumps, spectrogram inspection, and three independent ASR reads of every failed sample. The conclusion: **this is not a rebase/branch regression.** It is a combination of (a) a real, pre-existing talker weakness on one-word outputs and (b) a test-side ASR measurement problem that roughly doubles the apparent failure rate.

### Measurements (30 serial single requests each, 2×L20X, seed=0)

| Configuration | Whisper-small verdict (current CI judge) |
|---|---|
| `dev/vllm-align` (vLLM `dcfebf93f4`), `--async-chunk` | 9/30 garbled |
| `dev/vllm-align`, sync (`--no-async-chunk`) | 12/30 garbled |
| `main` + vLLM v0.25.0, `--async-chunk` | 6/30 garbled |

Statistically indistinguishable across branches and across sync/async — the "async_chunk regression" framing does not hold. Failing sample indices are **byte-for-byte reproducible across server restarts** (seed=0 makes the pipeline deterministic per request order), which also means the test's 10 retries re-roll much less randomness than intended.

### Audio-level analysis of the 9 failed samples (rebase, async_chunk)

We saved all 30 WAVs and judged each failure with Whisper-small (CI's judge), Whisper large-v3 (auto and forced-English), and wav2vec2 CTC (English chars + espeak IPA phonemes, which has no language-model prior to "autocorrect" toward *London*):

| sample | whisper-small | large-v3 (EN) | wav2vec2 phonemes | verdict |
|---|---|---|---|---|
| 4 | 'Longan' | London | `l ʌ ŋ d ʌ n` | correct (ASR false alarm) |
| 3 | 'Wondon' | London | `w a ŋ d ʌ n` | degraded: /l/ onset → [w] |
[...]

PR #5383, opened 2026-07-24
[Bugfix][Frontend] Remove async-chunk guard from realtime API

Sy0307:
## Purpose

Fixes #5363.

Qwen3-Omni Realtime requests should be allowed to use the standard `/v1/realtime` handler when the server is started with `--async-chunk`. Full-duplex requests continue to use the separate handler selected by `?duplex=true`.

## What broke

The route-level guard added in #3907 rejected every standard Realtime WebSocket request whenever `async_chunk` was enabled. The request was closed before the existing realtime handler or model pipeline could run.

Removing that guard restored the API entry point, but it also exposed a latent streaming lifecycle bug in the downstream Stage 1 session path. The guard masked this bug; it did not cause it.

In a resumable streaming request, `EngineCoreOutput.finished` can describe two different events:

1. the current streaming **segment** stopped and the request will resume when the next connector payload arrives; or
2. the entire **session** reached a terminal state.

Before this change, the scheduler and orchestrator did not preserve that distinction consistently:

- `OmniARScheduler` marked every stopped request as `is_segment_finished=True`, including terminal requests.
- `OmniGenerationScheduler` did not mark resumable generation stops as segment boundaries.
- `Orchestrator` counted any `finished=True` output from a final-output stage as session completion, even when it represented only a streaming segment.
- A terminal AR request with no tensor payload could also miss the final connector notification.

As a result, the orchestrator could finalize/clean up a session too early, or the downstream stage could resume without the correct segment boundary. Stage 1 then reused request state at the wrong lifecycle point, producing timing-sensitive scheduler or prefill/decode handoff inconsistencies.

This is separate from the async placeholder accounting fixed by #5183. This branch is based on `93786c9c`, which already contains #5183.

### API routing

- Remove the global `async_chunk` rejection from `/v1/realtime`.
- Re-enable the existing paced Qwen3-Omni async-chunk H100 E2E test.

### Streaming lifecycle

- Derive `is_segment_finished` from `_handle_stopped_request()`:
  - resumable stop (`finished=False`) → segment boundary;
  - terminal stop (`finished=True`) → session terminal, not a segment boundary.
- Mark terminal AR requests as non-resumable.
- Emit the terminal connector notification even when no inter-stage tensor payload is present.
- Propagate segment-boundary metadata from `OmniGenerationScheduler`.
- Prevent `Orchestrator` from counting a streaming segment boundary as completion of a final-output stage.

[...]

PR #5646, opened 2026-07-31
[CI/Build] Let audio assertions pin the transcript language

ShengleiFu:
## Purpose

Test-side half of #5395. `test_one_word_prompt_001` flakes because whisper's language
auto-detection is unstable on the sub-second clip a one-word answer produces: correct
"London" audio is transcribed as `런던` / `ランデ` / `梁敦`, and the short-text containment
check in `assert_omni_response` then fails on good output.

This PR:

- Threads an optional `language` through `convert_audio_bytes_to_text` →
  `convert_audio_file_to_text` → `_whisper_transcribe_in_current_process` →
  `model.transcribe(...)`. Default stays `None` (= whisper auto-detection), so no
  existing caller changes behaviour.
- Lets a request declare `transcript_language`, resolved in `_resolve_audio_transcript`
  the same way `similarity_threshold` / `transcript_escalation_model` already are.
- Carries the same language into the escalated pass in `_assert_transcript_matches`, so a
  request declaring both `transcript_language` and `transcript_escalation_model` does not
  get English on the first transcription and auto-detection on the retry.
- Pins `"en"` for `test_one_word_prompt_001` only.
- Adds unit coverage for the whole parameter chain and for the auto-detection default.

## Measurements

90 serial single requests, sync arm (`--no-async-chunk`), 2× RTX PRO 6000, vLLM 0.25.0,
stock `qwen3_omni_moe.yaml`, `--seed 0`. All four cells judge the **same** 90 WAVs with
CI's containment rule:

| judge | auto language | `language="en"` |
|---|---|---|
| whisper-`small` | 22/90 fail | **7/90 fail** |
| whisper-`large-v3` | 35/90 fail | 7/90 fail |

Of `small`'s 22 failures, forcing English recovers 15: 13 are non-Latin-script
renderings of correct audio (`런덴`×6, `런던`, `롱댄`, `ランデ`×2, `ランデン`, `ランダン`,
`梁敦`) and 2 are nearby Latin-script recognitions (`Longan`, `Ronden`). One of the 13,
`런던`, is literally "London" in Hangul.

**#5395 suggests upgrading the judge to `large-v3`; the data says not to.** Upgrading
without pinning the language is *worse* (35/90 — a more multilingual model drifts more
readily on a 0.7 s clip). With the language pinned, both models report the same aggregate
failure count (7/90), although they disagree on three samples in each direction.
Upgrading to `large-v3` therefore does not reduce the observed failure rate, while adding
a 2.9 GB download and a slower judge. Language pinning accounts for the entire aggregate
failure-rate reduction in this corpus; model size changes which individual clips pass, but
not the 7/90 total.

### Independent CTC cross-check (no external LM)

whisper is an encoder-decoder whose decoder is itself a language model, so it can snap a
mushy clip to a high-prior word — and "London" is high-prior here. Re-judging the same 90
clips with wav2vec2 CTC (`facebook/wav2vec2-large-960h-lv60-self`: per-frame characters, no
external LM) gives a readout that cannot autocorrect toward the expected word:

| judge | CTC mismatches caught | Whisper-only mismatches | CTC mismatches missed |
|---|---|---|---|
| `small` + `language="en"` | 6 | 1 | 4 |
| `large-v3` + `language="en"` | 5 | 2 | 5 |

`large-v3` transcribes five clips independently flagged by CTC as "London" (CTC reads them
`LONG DAY`, `LANDING`, `DON'T THEN`, `LANDED`, `ONE DAY`) and rejects two that CTC reads as
`LONDON`. So once the language is pinned, `small` agrees slightly more often with the
independent CTC readout on this corpus than `large-v3` does — which is a further reason not
to make the model swap, on top of the identical aggregate count.

CTC produces a non-London transcription for 10 of the 90 clips. That is a statement about
CTC's output, not a measured defect rate: wav2vec2 has no external LM (the property needed
here) but it still carries its own training priors, and on a separate 90-clip run it
rendered "London" as `LANDON` / `LANDE` / `LONDA` on clips that **both** whisper
configurations accepted. It is therefore useful as a tie-breaker on individual disputed
clips, and unreliable as a substring-matched rate estimator. Treat it as an independent
second opinion, not a verdict.

Seed determinism from the issue also reproduced: re-running after a full server restart
produced byte-identical WAVs for all 30 overlapping samples.

## Scope / what this does not fix

The residual is out of scope here and belongs to the model-side half of #5395. Four clips
are rejected by every whisper configuration (`I'm on that.` / `Today, Jordan` /
`Young that.` / `Long day.`) and CTC also reads them as non-London, making those the
strongest model-side candidates — though none has been confirmed by blinded listening.

Correcting an earlier revision of this description: two clips I had listed as failures only
`large-v3` detects (`Linden`, `LENDEN`) are ones CTC reads as `LONDON`, so I no longer
present them as `small` masking a real failure. Only one such clip is corroborated by CTC
(`One Dan`, CTC `ONE DEN`), and `small` transcribes it as "London" so containment passes;
escalation only fires on failure, so no judge change here surfaces it. This PR reduces the
flake rate of the gate; it does not make the judge correct.

## Not a duplicate

No open PR references #5395, and none touches `convert_audio_file_to_text`,
`transcript_escalation_model`, `one_word_prompt` or `assert_omni_response`. #5395 is
assigned to me.

Unrelated but worth flagging for anyone re-measuring this test on an older tree: the
`--async-chunk` path had a terminal-flush regression that truncated a one-word answer
from ~0.78 s to 0.297 s (only the initial 4-frame chunk survived), which would swamp any
talker-quality measurement.

- introduced by `e043818a` — [#5383](https://github.com/vllm-project/vllm-omni/pull/5383),
  2026-07-25, which made `is_segment_finished` false on a terminal stop;
- fixed by `8f115164` — [#5414](https://github.com/vllm-project/vllm-omni/pull/5414),
  2026-07-27, one line in `chunk_transfer_adapter.py`:
  `is_finished=is_segment_finished` → `is_finished=is_segment_finished or is_finished`.

Current `main` is well past the fix; the measurements above were taken on the sync arm,
which never enters that path.

## Test Plan

- [x] `pytest tests/helpers/tests/test_media.py tests/helpers/tests/test_assertions.py`
- [x] `ruff check` + `ruff format --check` (v0.14.10, as pinned in `.pre-commit-config.yaml`)
- [x] `tools/pre_commit/check_test_marks.py` on every touched test file
- [x] Audited all existing callers of the three helpers: each passes `output_path` /
      `raw_bytes` positionally and anything else by keyword, so appending `language` as a
      trailing optional parameter cannot change an existing call's meaning.
- [x] `pytest tests/e2e/online_serving/test_qwen3_omni_expansion.py::test_one_word_prompt_001`
      on 2× RTX PRO 6000 with vLLM 0.26.0 and the stock `qwen3_omni_moe.yaml` (no local
      edits): **2 passed in 887s**, both the `default` and `async_chunk` parametrizations.
      (This needs two cards: the stage-CLI fixture initialises all three stages
      concurrently, so folding them onto one 96 GB card races stage 0's KV-cache profiling
      — `Available KV cache memory: 0.0 GiB (profiling fallback)` — and the server never
      boots.)

## Test Result

```
$ pytest tests/helpers/tests/test_media.py tests/helpers/tests/test_assertions.py -q
8 passed in 0.20s
```

Each new test was confirmed to fail against unmodified `main` before the change, for the
intended reason:

```
test_transcribe_forwards_requested_language
  TypeError: _whisper_transcribe_in_current_process() got an unexpected keyword argument 'language'
test_resolve_transcript_honors_declared_language
  AssertionError: assert None == 'en'
test_bytes_entrypoint_forwards_language_to_subprocess
  TypeError: convert_audio_bytes_to_text() takes from 1 to 2 positional arguments but 3 were given
test_escalated_transcript_keeps_declared_language
  TypeError: _assert_transcript_matches() got an unexpected keyword argument 'language'
```

The two pre-existing tests in `test_assertions.py` are untouched and still pass. The two
default-behaviour tests pass before and after, pinning auto-detection as the default.

The 2-GPU e2e run is worth reporting in full rather than just as a pass. Across both
parametrizations it consumed 4 of the test's 10 retries, and the 21 logged transcripts were:

```
London     x17   (containment passed)
Langdon    x1
Long den   x1
Why?       x2
```

So the gate is not deterministic after this change — but every residual failure is a
plausible talker artefact, and **none** is a script-drift mistranscription of the
`런던` / `ランデ` / `梁敦` kind this PR targets. That is the intended outcome: the
judge-induced failures are gone, the model-side ones remain and belong to #5395's other
half.

`tests/helpers/tests/` also contains `test_media_pydub_removal.py` (2 of its cases need
`vllm.multimodal.media.audio`) and `test_runtime_stage_cli.py` (needs `torch`); neither is
runnable in the local env used above and neither references the helpers changed here —
verified by re-running them with this branch's changes reverted and getting identical
results.

---

AI assistance was used for this change (investigation, measurement harness and drafting).
I have reviewed every changed line and run the tests reported above.

### DIFF
--- a/tests/e2e/online_serving/test_qwen3_omni_expansion.py
+++ b/tests/e2e/online_serving/test_qwen3_omni_expansion.py
@@ -468,6 +468,10 @@ def test_one_word_prompt_001(omni_server, openai_client) -> None:
         "messages": messages,
         "stream": True,
         "key_words": {"text": ["london"]},
+        # A one-word English clip is too short for whisper's language detection to
+        # be stable; unpinned it renders "London" as 런던 / ランデ / 梁敦 and the
+        # containment check then fails on correct audio.
+        "transcript_language": "en",
     }
 
     # Retry only when assert_omni_response fails on text/audio cosine similarity (see tests/helpers/assertions.py).
--- a/tests/helpers/assertions.py
+++ b/tests/helpers/assertions.py
@@ -632,6 +632,7 @@ def _assert_transcript_matches(
     *,
     threshold: float,
     escalation_model: str | None = None,
+    language: str | None = None,
 ) -> None:
     """Assert spoken audio matches ``expected_text``.
 
@@ -647,7 +648,8 @@ def _assert_transcript_matches(
     clip is re-transcribed with that stronger ASR and the assertion is decided on
     its verdict -- so a weak whisper-``small`` mishear on a short clip does not
     flake the gate, while a genuine model artifact still fails (the strong ASR
-    mismatches too).
+    mismatches too). ``language`` applies to that escalated pass too, so a request
+    that pins a language does not silently fall back to auto-detection on retry.
     """
     expected = str(expected_text).strip().lower()
     similarity = cosine_similarity_text(transcript.strip().lower(), expected)
@@ -664,7 +666,7 @@ def _assert_transcript_matches(
             f"whisper-small below threshold ({similarity:.2f} <= {threshold}); "
             f"escalating to whisper-{escalation_model} to rule out an ASR mishear"
         )
-        strong_transcript = convert_audio_bytes_to_text(audio_bytes, model_size=escalation_model)
+        strong_transcript = convert_audio_bytes_to_text(audio_bytes, model_size=escalation_model, language=language)
         strong_similarity = cosine_similarity_text(strong_transcript.strip().lower(), expected)
         print(
             f"audio content (whisper-{escalation_model}): {strong_transcript}\n"
@@ -741,6 +743,7 @@ def assert_audio_speech_response(response: Any, request_config: dict[str, Any], 
                     expected_text,
                     threshold=0.9,
                     escalation_model=request_config.get("transcript_escalation_model"),
+                    language=request_config.get("transcript_language"),
                 )
         _assert_preset_voice_gender_from_audio(
             response.audio_bytes,
--- a/tests/helpers/assertions.py
+++ b/tests/helpers/assertions.py
@@ -527,7 +527,7 @@ def _resolve_audio_transcript(
     audio_bytes = getattr(response, "audio_bytes", None)
     if not audio_bytes:
         return None
-    return convert_audio_bytes_to_text(audio_bytes)
+    return convert_audio_bytes_to_text(audio_bytes, language=request_config.get("transcript_language"))
 
 
 def assert_omni_response(response: Any, request_config: dict[str, Any], run_level):
--- a/tests/helpers/media.py
+++ b/tests/helpers/media.py
@@ -652,7 +652,9 @@ def _serialize_whisper_model_download(model_size: str = "small"):
         f.close()
 
 
-def _whisper_transcribe_in_current_process(output_path: str, model_size: str = "small") -> str:
+def _whisper_transcribe_in_current_process(
+    output_path: str, model_size: str = "small", language: str | None = None
+) -> str:
     import whisper
 
     device_index = None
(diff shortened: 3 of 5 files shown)


Gaohan123 (approved):
LGTM. Thanks

PR #5648, opened 2026-07-31
[Perf][Ming-TTS] Batch output snapshot transfers

EchoHayate:
## Summary

- resolve the active Ming-TTS output row once per stage-output snapshot
- batch the decode-step and stop-reason D2H transfer
- reuse the same snapshot logic for async-chunk and full-payload paths
- preserve sparse masks, empty masks, Python/list metadata, and 2D/3D latent inputs

## Problem

`llm2audio_vae_async_chunk` previously called `_extract_last_value` twice and
`_extract_last_patch` once for every generated latent patch. Each helper
independently scanned `ming_has_patch`, and the scalar helpers converted the
active index and selected value with `.item()`.

On CUDA output tensors this repeated the same mask work three times and placed
multiple GPU-to-host synchronizations on the streaming output critical path.

## Root cause

The patch, final decode step, and stop reason are fields of one logical model
output, but they were extracted independently. The helpers therefore could not
share the active row or batch metadata transfers.

## Design

`_extract_ming_output_snapshot` now:

1. resolves the last active row once;
2. uses a fixed-shape reverse-mask reduction for the per-step last-row path;
3. transfers decode-step and stop-reason metadata together;
4. selects the latent patch using the same row;
5. supports `all_patches=True` for the full-payload path.

The list/Python metadata fallback remains supported, while the normal tensor
path avoids repeated scalar extraction.

### Focused extraction microbenchmark

Environment:

- NVIDIA A100 80GB PCIe
- PyTorch 2.7.0+cu126
- latent shape per row: `[4, 64]`
- 100 warmup calls
- 1,000 iterations per round, 7 rounds, median
- each case asserts old/new patch and metadata equality before timing

| Batch | Before (us) | After (us) | Speedup | Time saved |
|---:|---:|---:|---:|---:|
| 1 | 246.284 | 175.314 | 1.40x | 28.8% |
| 8 | 254.055 | 171.922 | 1.48x | 32.3% |
| 32 | 245.826 | 173.739 | 1.41x | 29.3% |

PyTorch profiler counts for batch 32:

| Operation | Before | After |
|---|---:|---:|
| `aten::nonzero` | 3 | 0 |
| `cudaMemcpyAsync` | 9 | 3 |
| `cudaStreamSynchronize` | 9 | 3 |

### Real-model end-to-end validation

I also ran the full two-stage streaming pipeline with the public checkpoint,
not a source-extraction harness:

- model: `inclusionAI/Ming-omni-tts-0.5B`
- checkpoint revision: `9154772e7fbc585907b6237e3190790676f28975`
- checkpoint: 12 files, 3.1 GB cache, 1,181 readable safetensors tensors
- hardware: NVIDIA A100 80GB PCIe
- case: `style`
- mode: `--streaming --enforce-eager`
- base: direct parent `226cb4a38cee37d2cc6a3d509703b5473f815e81`
- optimized implementation: `6de07c561ec18e47374a08bcb22654abebeae1af`

| Metric | Base | Optimized | Delta |
|---|---:|---:|---:|
| Pipeline E2E wall time | 22,441.339 ms | 21,960.589 ms | -2.14% |
| Stage 0 generation | 21,637.977 ms | 21,285.235 ms | -1.63% |
| Stage 1 generation | 22,427.015 ms | 21,942.719 ms | -2.16% |
| Time to first audio | 21.910 s | 21.463 s | -2.04% |
| Mean inter-chunk time | 0.5320 s | 0.4986 s | -6.28% |
| Peak GPU memory above pre-run baseline | 39,082 MiB | 39,084 MiB | +2 MiB |

This is one sequential base/head E2E pair, so the latency deltas are
directional evidence rather than a statistically stable throughput claim. The
focused benchmark and profiler counts isolate the intended synchronization
reduction.

## Correctness

The real-model base and optimized runs both produced:

- mono 44.1 kHz PCM16 audio
- 352,800 samples / 8.0 seconds
- non-silent waveform (`RMS = 0.07408`, max amplitude `0.61943`)
- no clipped samples

The two output WAV files are bitwise identical:

`SHA256 6f5ed321f222cd0bb282c5e3c6225de5ac6050c491015d01233130640d2707c4`

Added focused coverage for:

- selecting the last row from a sparse active mask;
- selecting all active rows for the full-payload path;
- no-active-row behavior;
- Python/list metadata fallback;
- full-payload patch and terminal metadata propagation.

Repository-level validation in the same full vLLM-Omni environment:

- `8 passed, 17 warnings` for
  `tests/model_executor/stage_input_processors/test_ming_tts_async_chunk.py`
- Python compile check passed
- `git diff --check` passed

The first repository-level run exposed that the new full-payload test used
attribute access for the real dict-like `OmniTokensPrompt` result. Commit
`4c15fff7` corrects the test to exercise the repository's actual mapping
contract. The validation image does not include `ruff`, so the repository CI
pre-commit job remains the formatting/lint gate.

### DIFF
--- a/tests/model_executor/stage_input_processors/test_ming_tts_async_chunk.py
+++ b/tests/model_executor/stage_input_processors/test_ming_tts_async_chunk.py
@@ -155,7 +155,7 @@ def test_ming_full_payload_reuses_snapshot_for_patches_and_metadata():
     outputs = llm2audio_vae([stage_output])
 
     assert len(outputs) == 1
-    additional_information = outputs[0].additional_information
+    additional_information = outputs[0]["additional_information"]
     torch.testing.assert_close(
         additional_information["ming_latent_patches"],
         patches[[0, 2]],
--- a/tests/model_executor/stage_input_processors/test_ming_tts_async_chunk.py
+++ b/tests/model_executor/stage_input_processors/test_ming_tts_async_chunk.py
@@ -10,6 +10,10 @@ import torch
 from vllm_omni.model_executor.models.ming_tts.constants import LATENT_DIM, PATCH_SIZE
 from vllm_omni.model_executor.stage_input_processors.ming_tts import (
     MING_EMIT_PATCH_COUNT_KEY,
+    MING_FINAL_DECODE_STEP_KEY,
+    MING_STOP_REASON_KEY,
+    _extract_ming_output_snapshot,
+    llm2audio_vae,
     llm2audio_vae_async_chunk,
 )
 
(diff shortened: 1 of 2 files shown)


EchoHayate:
Added full-environment validation and fixed the test contract in `4c15fff7`.

- repository-level focused pytest: `8 passed, 17 warnings`
- real `inclusionAI/Ming-omni-tts-0.5B` two-stage streaming E2E succeeded on both base and head
- generated 8.0 s, 44.1 kHz, non-silent WAVs are bitwise identical
- single-pair E2E was 22.44 s -> 21.96 s with effectively unchanged peak GPU memory; this is reported as directional, not a stable throughput claim

The PR description now includes the exact checkpoint revision, run configuration, correctness audit, performance data, and validation boundaries.

PR #5455, opened 2026-07-27
[BugFix][MiniCPM-o] Restore audio output without async chunk

R2-Y:
PLEASE FILL IN THE PR DESCRIPTION HERE.

Fix MiniCPM-o 4.5 returning empty audio when `async_chunk=false`. The synchronous Talker-to-Code2Wav path now transfers the complete codec payload and waits for it before decoding.

## Purpose
- Add a synchronous full-payload producer for the Talker-to-Code2Wav handoff.
- Add a token-only placeholder so Code2Wav reserves the required input slots without duplicating codec data.
- Preserve reference audio, duplex metadata, silence context, and request routing in the synchronous payload.
- Initialize sender and receiver connectors for full-payload mode, not only async-chunk mode.
- Register MiniCPM-o Code2Wav with the full-payload input coordinator so decoding waits for connector delivery.
- Keep the existing async-chunk path unchanged.
Call chain:
`Talker output → full-payload accumulator → SharedMemoryConnector → Code2Wav wait gate → waveform output`

## Test Plan
- Run focused stage processor and pipeline topology tests.
- Run connector role and engine initialization tests for both async and synchronous modes.
- Run scheduling coordinator tests.
- Start MiniCPM-o 4.5 with `async_chunk=false`.
- Send a non-streaming OpenAI-compatible text-and-audio request.
- Verify the audio choice contains a valid WAV payload.

**vLLM Version:** `0.25.0`

**vLLM-Omni Commit:** `cc05be9a3` + this change

## Test Result

- `92 passed`
- Ruff checks passed.
- End-to-end Stage 2 output:
  - `audio_generated_frames`: `297600`
  - `output_unit_count`: `297600`
  - WAV size: `568364` bytes
  - Header: valid `RIFF/WAVE`
- Audio is returned in the dedicated audio choice: `choices[1].message.audio.data`.

non_async_chunk:
image
async_chunk:
image

**BEFORE SUBMITTING:** read [CONTRIBUTING.md](https://github.com/vllm-project/vllm-omni/blob/main/CONTRIBUTING.md) and run the [precheck-pr skill](https://github.com/vllm-project/vllm-omni/blob/main/.claude/skills/precheck-pr/SKILL.md) with the code agent for a self-check against project conventions.
(anything written below this line will be removed by GitHub Actions)

PR #5650, opened 2026-07-31
[Bug fix ] fix kv connector with unexpected local_rank

### LINKED ISSUE #5595: [Bug]: Hunyuan nightly: /v1/images/edits stalls with Mooncake KV-transfer timeouts; engine shutdown leaks orphaned workers; fixed ZMQ port 50051 causes cross-test collisions
### Summary

The nightly local Hunyuan CI (H200 dev machine, 8x NVIDIA L20X, shared box) has been failing every night since 2026-07-28. After four consecutive reproductions (Jul 28 manual + Jul 28/29/30 cron + Jul 31 manual re-run under a cleaned environment), the failures decompose into **three distinct product problems**. Environment-side noise (OpenBLAS thread exhaustion causing a SIGSEGV on the shared host) has been ruled out: with thread caps applied (`OPENBLAS_NUM_THREADS=32`, `OMP_NUM_THREADS=32`, `RAYON_NUM_THREADS=4`) the Jul 31 run shows zero `pthread_create` errors, yet all three problems below persist.

### Environment

- H200 dev machine, 8x NVIDIA L20X (shared), tests run on 4 GPUs
- vLLM `0.26.0`; vllm-omni `0.1.dev2381+g273fc8eb1` (latest main at each run; commits `f2c5152d` → `273fc8eb` all affected)
- Python 3.12.13, `HF_HOME=/models`

---

### Problem 1: `/v1/images/edits` requests stall — Mooncake KV-transfer timeouts

`tests/e2e/accuracy/test_hunyuan_image3.py::test_image_to_image_alignment_online` has failed in **every run since Jul 28** (recurrence of #5249, which was closed on Jul 24 as "not reproduced").

On the cleanest reproduction (Jul 31, no zombies, no thread exhaustion), the client POSTs to `/v1/images/edits` and the server never responds; the test fails after the 600 s read timeout:

```text
E  TimeoutError: timed out        # urllib3 read timeout, 600 s, POST /v1/images/edits
```

Server-side, the request is stuck in the stage-0 → stage-1 KV handoff; the cron runs show the same signature repeatedly:

```text
ERROR [kv_transfer_manager.py:1497] Timeout waiting for KV cache for request 0_3fee3cbb-... after 30.0s
```

Note: a `ValueError: Cannot use chat template functions...` appears at DEBUG level in `renderers/hf.py:283` around request time, but it looks like renderer capability probing rather than the root cause; the KV-transfer timeout is the hard evidence.

### Problem 2: engine shutdown leaks orphaned workers (zombie processes)

When the offline `test_image_to_image_alignment` crashes mid-run (worker SIGTERM + TCPStore shutdown), cleanup never completes and the spawned children (`resource_tracker`, `spawn_main`, 4x `vLLM-Omni::DiffusionWorker-*`) are orphaned to PPID 1, spinning in `shm_broadcast` poll loops for 8+ hours:

```text
WARNING [diffusion_engine.py:969] Worker thread did not terminate within 10s; scheduler and executor shutdown will be deferred.
DEBUG [shm_broadcast.py:208] Poller timed out    (repeats forever)
```

The workers even detect the parent's death (`multiproc_executor.py:793 Parent process exited, terminating worker queues`) but still never exit.

Proposed fix (bounded escalation instead of deferral):
1. Set `prctl(PR_SET_PDEATHSIG, SIGKILL)` in worker bootstrap — kernel-level guarantee that orphaning is impossible; the poll-loop shutdown flag then only serves as a second line of defense.
2. Replace "shutdown deferred" with `join(timeout)` → `terminate()` (SIGTERM) → short grace → `kill()` (SIGKILL) → `os._exit()` after killing the whole child process tree. The executor already knows all child PIDs, so a `kill_process_tree` on shutdown-timeout is feasible.
3. Make the `shm_broadcast` poll loop exit after N consecutive timeouts once a shutdown/parent-dead flag is set.

### Problem 3: fixed Mooncake ZMQ port (50051) causes cross-test collisions

The Mooncake connector binds a fixed ZMQ port from config. Two independent failure modes observed:

- **Zombie cascade (Jul 28-30):** orphans from Problem 2 hold the port overnight; the it2i perf benchmark then dies at startup:

```text
(Worker) ERROR [mooncake_transfer_engine_connector.py:1025] ZMQ bind failed on ...:50051: Address already in use
(APIServer) RuntimeError: Engine core initialization failed.
```

- **Self-collision without any zombies (Jul 31):** with a verified-clean host (port bindable, zero orphans), the accuracy test's own server had barely finished tearing down when the it2i benchmark started ~3 minutes later — and the bind still failed (dying listener / TIME_WAIT on the same fixed port).

The second mode shows this is not just a cleanup issue: any back-to-back runs sharing the fixed port can collide. Proposed fix: close the socket with `linger=0` in a `finally` block, and bind to an ephemeral port (bind to port 0, read the actual port) propagated to the peer via the existing stage handshake, keeping the configured port only as a default.

---

### Impact

- No usable online-accuracy signal for Hunyuan since Jul 28 (Problem 1).
- No it2i perf data for three consecutive nights (Problems 2+3); the three t2i perf benchmarks (tp4 / tp2_sp2 / tp2_cfgp2) pass in the same runs.

### Harness-side mitigations already in place (not fixes)

- Accuracy pytest runs under `setsid` with whole-process-group kill on timeout/exit.
- Orphan sweep + wait-until-port-free (up to 5 min) before the perf phase.

These protect the perf results but do not fix Problems 1-3 themselves.

### Logs

On the nightly machine:
- `/rebase/vllm-omni/logs/nightly_jobs_local_20260728_115104` (manual, SIGSEGV variant)
- `/rebase/vllm-omni/logs/nightly_jobs_local_20260728_160527` (cron, hang)
- `/rebase/vllm-omni/logs/nightly_jobs_local_20260729_170105` (cron, hang + cascade)
- `/rebase/vllm-omni/logs/nightly_jobs_local_20260730_170035` (cron, hang + cascade)
- `/rebase/vllm-omni/logs/nightly_jobs_local_20260731_100520` (manual, clean env, Problems 1+3 still reproduce)

Bounty-hunter:

## Purpose
https://github.com/vllm-project/vllm-omni/issues/5595#issuecomment-5138555004

**The bug is**: 
(1) In ar + dit,  the dit rank 0 success to recive kv, but dit rank1 failed to recive kv.
(2) kv manager connector bind an already used port.
(3) Incomplete RDMA configuration in the CI environment.
```
ls -l /sys/class/infiniband/mlx5_bond_1/device/net/
total 0
```
### Root Cause
https://github.com/vllm-project/vllm-omni/pull/5455 change the behavior that  AR stage enter `init_omni_connectors`, which is only called when async_chunk enable. 

The problem is that there are few problem in omni connector.
(1) for first bug, the  reason is: In `init_omni_connectors` , it init self._kv_transfer_manager.kv_send_key_builder, and it use local rank to build kv key. However, `_parse_rank_mapping` init `local_rank` from env value `LOCAL_RANK`. But in ar worker process, it never set `LOCAL_RANK`, so both tp1,tp2 set local_rank to 0, their kv key are both base on rank0. The Dit stage so that can't recived the kv which identify by rank1.

(2) for second bug, the reason is:  
<1> omni connector overlooked the case `tp > 1`,  all rank  used the same port for rdma， so it conflict
**<2>** After <1> fix，there are confict between  `omni connector` and `kv manager connector`, they both try to combine the connector info in yaml so they conflict. 

I believe there is a conflict in the current connector design. OmniConnect is responsible for transferring non-KV data between stages, while the KV Manager handles KV transfer. Since both read from the same connector configuration, they can end up competing with each other, which makes them difficult to use together safely. Addressing this properly would likely require a connector-system refactor, so for now I am presenting the changes that revert the problematic implementation.

### Fix

(1) restore the `stage_init_utils.py` change by https://github.com/vllm-project/vllm-omni/pull/5455
(2) set tcp transfer
## Test Plan
`pytest -s -v tests/e2e/accuracy/test_hunyuan_image3.py`
`pytest -s -v tests/dfx/perf/scripts/run_diffusion_benchmark.py \
  --test-config-file tests/dfx/perf/tests/test_hunyuan_image3_it2i.json`

**vLLM Version:**

**vLLM-Omni Commit:**

## Test Result
image

image

### DIFF
--- a/tests/dfx/perf/tests/test_hunyuan_image3_it2i.json
+++ b/tests/dfx/perf/tests/test_hunyuan_image3_it2i.json
@@ -27,16 +27,8 @@
             "async_chunk": false,
             "trust_remote_code": true,
             "connectors": {
-              "rdma_connector": {
-                "name": "MooncakeTransferEngineConnector",
-                "extra": {
-                  "host": "auto",
-                  "zmq_port": 50051,
-                  "protocol": "rdma",
-                  "device_name": "",
-                  "memory_pool_size": 17179869184,
-                  "memory_pool_device": "cpu"
-                }
+              "shared_memory_connector": {
+                  "name": "SharedMemoryConnector"
               }
             },
             "stages": [
@@ -66,7 +58,7 @@
                   "need_send_cache": true
                 },
                 "output_connectors": {
-                  "to_stage_1": "rdma_connector"
+                  "to_stage_1": "shared_memory_connector"
                 },
                 "default_sampling_params": {
                   "temperature": 0.0,
@@ -87,7 +79,8 @@
                 "devices": "2,3",
                 "distributed_executor_backend": "mp",
                 "omni_kv_config": {
-                  "need_recv_cache": true
+                  "need_recv_cache": true,
+                  "enable_kv_async_prefetch": true
                 },
                 "parallel_config": {
                   "tensor_parallel_size": 2,
@@ -96,7 +89,7 @@
                 },
                 "vae_use_tiling": true,
                 "input_connectors": {
-                  "from_stage_0": "rdma_connector"
+                  "from_stage_0": "shared_memory_connector"
                 },
                 "default_sampling_params": {
                   "num_inference_steps": 50,
--- a/vllm_omni/engine/stage_init_utils.py
+++ b/vllm_omni/engine/stage_init_utils.py
@@ -1164,17 +1164,19 @@ def get_stage_connector_spec(
     stage_id: int,
     async_chunk: bool,
 ) -> dict[str, Any]:
-    """Return the first connector spec for a stage data-plane edge."""
+    """Return the first connector spec for the stage when async chunking is enabled."""
     from vllm_omni.distributed.omni_connectors import get_stage_connector_config
+
+    if not async_chunk:
+        return {}
 
     stage_connectors_cfg = get_stage_connector_config(omni_transfer_config, stage_id)
     for cfg in stage_connectors_cfg.values():
         return dict(cfg.get("spec", {}))
 
-    # A producer does not consume connector data itself. Keep its connector
-    # for both async-chunk and terminal full-payload sends, but mark it
-    # sender-only so the scheduler does not park orchestrator-provided inputs
-    # waiting for an upstream payload.
+    # A stage can be an async producer without consuming chunks itself. Keep
+    # its connector for save_async(), but mark it sender-only so the scheduler
+    # does not park normal orchestrator-provided inputs waiting for a chunk.
     target_stage = str(stage_id)
     for (from_stage, _to_stage), spec in getattr(omni_transfer_config, "connectors", {}).items():
         if from_stage == target_stage:


Bounty-hunter:
@Gaohan123 @R2-Y PTAL

R2-Y:
reverts the #5455 stage_init_utils change that injected stage connector specs even when async_chunk=false. That revert is intended to fix AR→DiT KV transfer regressions (bad local_rank keying and OmniConnector vs KV TransferManager port/config contention after non-async stages started calling init_omni_connectors). However, MiniCPM-o 4.5 non-async audio depends on that same injection: Talker→Code2Wav still moves the full codec payload over SharedMemory, and MiniCPM runners only initialize OmniConnector when an explicit sender/receiver role is present. Restoring the async_chunk-only gate would again leave Stage 2 without a connector wait path and regress to empty audio WAV output.

PR #5653, opened 2026-07-31
[Bugfix] Preserve channel-last video tensor layout

### LINKED ISSUE #5417: [Bug]: Video encoding corrupts channel-last tensors with 3 or 4 frames

whisper-la:
## Purpose

Fix video encoding for channel-last PyTorch tensors containing exactly 3 or 4 frames.

Previously, `_normalize_video_tensor` treated any 4D tensor whose first dimension was 3 or 4 as channel-first. Therefore, valid `[F, H, W, C]` tensors with 3 or 4 frames were incorrectly permuted and encoded with the wrong frame or spatial dimensions.

This change delegates tensor layout detection to the existing shared NumPy normalization logic. It preserves channel-last tensors while retaining support for channel-first `[C, F, H, W]` tensors.

Fixes #5417.

## Test Plan

Run the targeted regression tests:

```bash
python -m pytest -s -v \
  tests/entrypoints/openai_api/test_video_api_utils.py::test_coerce_video_preserves_channel_last_tensor_with_few_frames \
  tests/entrypoints/openai_api/test_video_api_utils.py::test_coerce_video_converts_channel_first_tensor
```
  Run the complete related test file:
```bash
python -m pytest -s -v \
  tests/entrypoints/openai_api/test_video_api_utils.py
```

Hardware: NVIDIA H100
vLLM Version: 0.26.0
vLLM-Omni Commit: 1b9f7ea882325e50ccb0aad1fecbb15a5854328f

## Test Result
Targeted video layout regression tests: 3 passed
Complete video API utility test suite: 10 passed
No test failures
The targeted tests verify that:
Channel-last RGB tensors with 3 frames preserve their original layout and values.
Channel-last RGB tensors with 4 frames preserve their original layout and values.
Channel-first tensors are still converted correctly to channel-last format.


lishunyang12:
Thanks for your contribution. Already fixed by the issue author.

PR #5654, opened 2026-07-31
[Model] Optimize MammothModa2 MoE token routing

EchoHayate:
### Problem

MammothModa2 creates one `gen_token_mask` per model forward, but every MoE-enabled decoder layer independently classified the same tokens. The mixed-token path repeated `any`, `all`, two `where` calls, `sum().item()`, concatenation, `argsort`, splitting, and an inverse gather.

### Root cause

The routing metadata was represented only as a boolean mask. Each layer therefore rebuilt generation/understanding positions and performed CUDA scalar reads. On an 8-layer routing-only profile this produced 24 `item` / `_local_scalar_dense` calls and 8 sorts, introducing repeated host synchronization in the decoder hot path.

### Design

- Build `(generation_positions, understanding_positions)` once per pipeline stage.
- Reuse the same routing metadata in generation-aware embeddings and every local Qwen2/Qwen3 decoder layer.
- Route mixed tokens with `index_select` and restore order with `index_copy_`, removing permutation concatenation, sorting, and inverse gather.
- Preserve the existing `gen_token_mask` API as a compatibility fallback for direct callers.
- Keep pipeline-parallel mask propagation unchanged; each downstream stage builds its local routing tuple once.

No model configuration or serving configuration changes are required.

## Test Plan

**Hardware:** 1x NVIDIA A100 80GB PCIe, driver 535.261.03

**Software:** PyTorch 2.5.1+cu121, CUDA 12.1

**vLLM Environment:** repository validation image with the full vLLM/vLLM-Omni import path; the earlier source-extracted harness is retained only as local routing evidence

**vLLM-Omni Commit:** `c885caf5` (base `4956a41f`)

Commands:

```bash
python3 -m py_compile \
  vllm_omni/model_executor/models/mammoth_moda2/mammoth_moda2.py \
  tests/model_executor/models/test_mammoth_moda2_moe_routing.py

uvx --offline ruff check \
  vllm_omni/model_executor/models/mammoth_moda2/mammoth_moda2.py \
  tests/model_executor/models/test_mammoth_moda2_moe_routing.py

uvx --offline ruff format --check \
  vllm_omni/model_executor/models/mammoth_moda2/mammoth_moda2.py \
  tests/model_executor/models/test_mammoth_moda2_moe_routing.py

python mammoth_moe_remote_validate.py \
  vllm_omni/model_executor/models/mammoth_moda2/mammoth_moda2.py
```

The benchmark uses 5 warmup iterations and 15 measured iterations per case, reporting the median and observed range. It isolates routing overhead with identity experts; it is not an end-to-end model throughput claim.

## Test Result

Correctness:

- CPU source-extracted parity: passed
- CUDA source-extracted parity: passed
- 2D and 3D mixed routing, all-generation, all-understanding, legacy mask fallback, and invalid mask/routing sizes are covered
- Ruff check: passed
- Ruff format check: passed
- Python compilation: passed

Profiler, 256 tokens and 8 local layers:

| Operation | Before | After |
|---|---:|---:|
| `any` / `all` | 8 / 8 | 0 / 0 |
| `item` / `_local_scalar_dense` | 24 / 24 | 0 / 0 |
| `where` | 16 | 2 |
| `argsort` / `sort` | 8 / 8 | 0 / 0 |

Routing-only latency:

| Tokens | Local layers | Before median (range), ms | After median (range), ms | Speedup |
|---:|---:|---:|---:|---:|
| 32 | 1 | 0.4663 (0.4508-0.4958) | 0.2343 (0.2229-0.2749) | 1.99x |
| 32 | 8 | 3.4137 (3.3158-4.6182) | 0.7320 (0.6982-0.7929) | 4.66x |
| 32 | 32 | 13.3233 (13.0059-18.8020) | 2.4468 (2.4045-2.5491) | 5.45x |
| 256 | 1 | 0.4991 (0.4876-0.5142) | 0.2476 (0.2396-0.2810) | 2.02x |
| 256 | 8 | 3.5932 (3.5065-3.6934) | 0.7422 (0.7288-0.7683) | 4.84x |
| 256 | 32 | 14.5576 (13.3345-18.7424) | 2.3470 (2.2930-3.8484) | 6.20x |
| 1024 | 1 | 0.6496 (0.6320-0.6573) | 0.3015 (0.2937-0.3526) | 2.15x |
| 1024 | 8 | 4.3922 (3.5989-4.7838) | 1.1249 (0.9173-1.1355) | 3.90x |
| 1024 | 32 | 14.8636 (14.2016-19.0052) | 3.2024 (3.1956-3.7451) | 4.64x |

Peak temporary CUDA allocation for 1024 tokens / 32 layers decreased from 32.02 MiB to 24.01 MiB (25.0%).

### Full repository pytest

The committed test now also runs through the full repository import path (not AST/source extraction):

```text
pytest -q tests/model_executor/models/test_mammoth_moda2_moe_routing.py
7 passed, 17 warnings in 0.10s
```

### Real MammothModa2 AR -> DiT A/B

**Checkpoint:** `bytedance-research/MammothModa2-Preview` at revision `ef5a5e41dbf0de1ef6275586b7580f0d4248b4c6` (34.5 GiB, 8 safetensors shards, all 1,391 indexed tensors opened successfully).

**Hardware/topology:** 2x NVIDIA A100 80GB PCIe, with the real AR stage on one GPU and the real DiT stage on the other. The evidence deploy config keeps `max_model_len=8192` and eager execution, raises the per-stage memory budget to `0.9`, and separates the stages because both full models cannot remain resident on one 80GB GPU. Base and head used the identical topology.

**Workload:** 256x256 T2I, prompt `A cat sitting on a laptop keyboard`, deterministic AR sampling (`temperature=0`, `top_k=1`), 273 generated AR tokens, 2 DiT steps, `text_guidance_scale=1.0`, fixed seed. Both commits used the same local checkpoint and offline mode.

| Metric | Base `4956a41f` | Head `c885caf5` | Head vs base |
|---|---:|---:|---:|
| Generation time | 7.7207 s | 7.1137 s | -7.9% |
| AR stage generation | 7,335.2 ms | 6,867.1 ms | -6.4% |
| DiT stage generation | 243.4 ms | 150.5 ms | directional only |
| Peak GPU memory, AR GPU | 80,009 MiB | 80,009 MiB | unchanged |
| Peak GPU memory, DiT GPU | 8,318 MiB | 8,318 MiB | unchanged |

Correctness/result checks:

- Both runs completed the full two-stage pipeline and produced a 256x256 RGB PNG.
- Base and head PNG files are byte-for-byte identical: SHA-256 `f3458df9e0be54439a6d807a25d4cba8ab7e11eb5924faf16c38352e64c2f0ea`.
- The output contains 13,143 unique RGB colors (pixel range 21-200), so this is not an empty/constant artifact.
- Both runs generated the same 273 AR tokens and reported the same peak GPU memory.

This is a single cold-start A/B on a shared machine, so the latency delta is directional evidence, not a statistically stable E2E throughput claim. The routing microbenchmark remains the controlled evidence for the hot-path operation reduction; the real-model run establishes end-to-end correctness and that the optimization is exercised by a mixed text/visual generation workload.


EchoHayate:
Added full-repository and real-model validation to the PR description.

- focused pytest through the full repository import path: **7 passed**
- real `bytedance-research/MammothModa2-Preview` checkpoint: 8 shards / 1,391 indexed tensors verified
- real two-stage 256x256 AR -> DiT base/head run: **273 AR tokens + 2 DiT steps**
- base/head PNGs are byte-for-byte identical (`f3458df9...f0ea`)
- peak memory unchanged (AR GPU 80,009 MiB; DiT GPU 8,318 MiB)
- one cold-run generation time was 7.7207 s -> 7.1137 s; I am treating this as directional only, not a stable E2E throughput claim

This replaces the earlier limitation that only the AST/source-extracted harness had been exercised.

EchoHayate:
## Author self-review

- Scope/invariants checked: current head `7d6c564cae173d70fb98a260f8db1b28dfe154ea` has a two-file net diff. It computes generation/understanding token positions once per pipeline stage, reuses them for embedding and decoder-layer MoE routing, preserves the existing mask-based fallback, skips an unused expert for homogeneous inputs, validates routing cardinality, and keeps output order/shape/contiguity unchanged.
- Current-main status: the branch was refreshed onto `f67cc52f7449d5f5dbb9d7790c7a049fa6aa2ba1`. Upstream has since advanced to `2bae550c04318269d6f672a670da4d0776307248`; those newer commits do not touch either changed file, a merge-tree check is clean, and GitHub currently reports the PR as mergeable.
- Tests and static checks: `7 passed, 15 warnings`; changed-file pre-commit passed; Python compile checks passed; `git diff origin/main...HEAD --check` passed. A diff-scoped review of the production change found no P0-P2 issue.
- Runtime evidence: the existing 2×A100 MammothModa2 256×256 AR→DiT run belongs to the earlier identical functional patch at `c885caf5`, not to this refreshed head. It produced byte-identical PNGs and a directional single cold-run generation result of `7.7207 s → 7.1137 s`; this is not claimed as a stable throughput result or a current-head GPU rerun.
- Evidence boundary: current-head validation is CPU/unit/static only. A repeated current-head GPU E2E remains desirable before treating the historical timing as a performance conclusion.
- AI assistance: I used Codex to help inspect the current-main diff, draft/refine tests and this review summary, and run the reported local checks. I reviewed the resulting changes and claims and remain responsible for the submission.

Current head reviewed: `7d6c564cae173d70fb98a260f8db1b28dfe154ea`.

Kare0638:
Peer review under the #7075 cross-review request (focus: TP/token-routing
constraints, generated-token and hidden-state correctness).

**Reviewed head:** `7d6c564cae173d70fb98a260f8db1b28dfe154ea`
**Base:** `main` at `f67cc52f7449d5f5dbb9d7790c7a049fa6aa2ba1`
**Reviewer's dependency context:** AR TP validation #7114; harness measured at
`51cf142b`; recipe measurements in #7403.

Existing threads read first. The author's 09-09 self-review already states the
evidence boundary, so this does not repeat it; what follows is what I measured
against it.

## Evidence boundary: checked

The self-review states the runtime evidence belongs to `c885caf5`, not this head.

Between the evidence base `4956a41f` and current `main`, the file was deleted by
#6353 and restored by #6694 as a 1,032-line whole-file add. `4956a41f` and
current `main` differ by one line in this file: `is_interleaved` moved into the
`vllm_omni.model_executor.models.utils` import
(`vllm_omni/model_executor/models/mammoth_moda2/mammoth_moda2.py:60`). That line
is outside `moe_forward`, `get_input_embeddings` and the decoder-layer `forward`.

## Measurements

### Base/head comparison at TP=1/2/4

Base `f67cc52f7449` and head `7d6c564cae17` differ by this PR's two files only
(`git diff --stat` reports 214/36 over them; `git merge-base` of the two is the
PR base itself). Both builds were run with an identical copy of the AR TP parity
harness from #7114 (`sha256:503e8ccf49a7476a…`).

| | TP=1 | TP=2 | TP=4 |
| --- | --- | --- | --- |
| generated token ids | identical | identical | identical |
| AR→DiT hidden states | identical | identical | identical |
| VAE image | identical | identical | identical |

18 of 18 pairs (3 degrees x 3 prompts x 2 engine starts) bitwise identical,
`torch.equal` on all three tensors. Per-build results: `50 passed, 1 failed` on
each side; the failure is the same case with the same value on both
(`cosine 0.99859744`, a threshold in the reviewer's harness that reproduces on
base). Full per-pair output, commands and environment in the following comment.

TP=8 not run — 4 GPUs was the maximum available in the storage region used.

### Operation census of the production diff

Added operations: `index_select` (x4), `index_copy_` (x4), `torch.where` (x2),
`torch.empty` (x1). Zero arithmetic operations.

### Call-site census

`moe_forward` has two production call sites at this head,
`mammoth_moda2.py:302` and `:669`. Both pass `token_routing`. `token_routing` is
`None` exactly when `gen_token_mask` is `None`, so the branch at
`mammoth_moda2.py:149-151` converts `None` to `None` in production. The
remaining callers are the five in the new test file.

### CI wiring

`.buildkite/cuda/test-merge.yml:17` runs
`pytest -sv tests/model_executor -m 'core_model and cpu'`; the new test carries
both markers.

## Code observations

**`mammoth_moda2.py:194-202`** — the output buffer takes `dtype` and `device`
from `und_out`, then `index_copy_` writes `gen_out` into it. `index_copy_`
requires matching dtypes; the previous `torch.cat([gen_out, und_out])` promoted
instead. Both experts are the same precision at this head. #7334 and #7436 add
FP8 coverage to this model.

**`mammoth_moda2.py:174-178`** — the validation is
`gen_pos.numel() + und_pos.numel() == total_tokens`, a cardinality check. The
buffer at `:196` is `torch.empty`, filled by two `index_copy_` calls. A routing
tuple with a duplicate index and a gap satisfies the cardinality check and
leaves the uncovered row uninitialized. `_build_moe_token_routing` produces a
partition, so this is unreachable through the in-tree path.

**`mammoth_moda2.py:149-151`** — the PR body describes the `gen_token_mask` path
as "a compatibility fallback for direct callers". See the call-site census
above; two of the seven new tests cover this path.

**`mammoth_moda2.py:505`** — the previous comment recorded that in the vLLM v1
path only `inputs_embeds` may be provided with `input_ids=None`, and that
generation tokens then fall back to `und_expert`. The replacement documents the
routing mechanism; the caveat is not carried over. The behaviour itself is
unchanged.

## A/B and profiling cross-check

| Required by the request | Present in the PR |
| --- | --- |
| Fixed baseline / head / checkpoint | Yes — `4956a41f` / `c885caf5`, checkpoint `ef5a5e41` |
| Hardware and placement | Yes — 2x A100 80GB PCIe, AR and DiT on separate GPUs |
| Workload / seed / resolution / steps / CFG | Yes — 256x256, 273 AR tokens, 2 DiT steps, guidance 1.0, `temperature=0`, `top_k=1`, fixed seed |
| Feature / cache state | Eager stated; prefix-caching and chunked-prefill state not stated |
| Preparation/startup/warmup separated | Micro-benchmark yes (5 warmup, 15 measured). E2E is one cold run |
| Sample counts and variability | Micro-benchmark yes. E2E n=1 |
| Correctness / quality | Byte-identical PNG with SHA-256, identical 273 AR tokens |
| Per-device memory | Reported — see below |
| AR / transfer / DiT / VAE / E2E attribution | AR, DiT and E2E; transfer and VAE not separated |

Two observations on the numbers:

The E2E memory rows were taken with `gpu_memory_utilization` raised to 0.9,
against the shipped 0.5 for stage 0 and 0.3 for stage 1
(`vllm_omni/deploy/mammoth_moda2.yaml:11,19`). At 0.9 of an 80 GB card the
reservation is ~73 GB, and the reported figure is 80,009 MiB on the AR GPU, so
the row reflects the budget rather than the change's allocation. The
micro-benchmark's 32.02 → 24.01 MiB temporary allocation is measured under
isolated conditions.

The `-7.9%` generation-time delta is a single cold run; the PR labels it
directional and does not derive a throughput claim from it.

## Tests run

See the following comment for full output, commands and environment.

## Not covered

- TP=8.
- Pipeline parallelism — the non-first-rank branch builds its own routing tuple
  from the propagated mask; not exercised by the new tests or the E2E run.
- Chunked prefill and prefix caching enabled.
- Mixed precision between the two experts.
- Which `enable_prefix_caching` / chunked-prefill state the PR's E2E A/B used.

Kare0638:
Raw results for the TP comparison in my review above, since I asked for
runnable commands and raw output and should supply my own.

## Pairwise capture comparison — full output

`torch.equal` on all three tensors, not a tolerance comparison.

```
base: /workspace/tp-validation/results-BASE     (f67cc52f7449)
head: /workspace/tp-validation/results-HEAD     (7d6c564cae17)

 tp prompt         run        tokens   hidden   image   verdict
  1 cat_keyboard   run1         True     True    True   identical
  1 cat_keyboard   -repeat1     True     True    True   identical
  1 terse          run1         True     True    True   identical
  1 terse          -repeat1     True     True    True   identical
  1 long_scene     run1         True     True    True   identical
  1 long_scene     -repeat1     True     True    True   identical
  2 cat_keyboard   run1         True     True    True   identical
  2 cat_keyboard   -repeat1     True     True    True   identical
  2 terse          run1         True     True    True   identical
  2 terse          -repeat1     True     True    True   identical
  2 long_scene     run1         True     True    True   identical
  2 long_scene     -repeat1     True     True    True   identical
  4 cat_keyboard   run1         True     True    True   identical
  4 cat_keyboard   -repeat1     True     True    True   identical
  4 terse          run1         True     True    True   identical
  4 terse          -repeat1     True     True    True   identical
  4 long_scene     run1         True     True    True   identical
  4 long_scene     -repeat1     True     True    True   identical

All 18 pairs bitwise identical across builds at TP=1/2/4.
```

Each row is one `(TP degree, prompt, engine start)` triple. `tokens` compares the
AR stage's generated ids, `hidden` the float32 AR→DiT stage output
`(num_tokens, 3584)`, `image` the VAE output `(3, 256, 256)`.

## Commands

Both builds were prepared identically; only the checkout differs.

```bash
# base
git checkout f67cc52f7449d5f5dbb9d7790c7a049fa6aa2ba1
# head
git fetch origin pull/5654/head:refs/remotes/pr/5654
git checkout refs/remotes/pr/5654          # 7d6c564cae173d70fb98a260f8db1b28dfe154ea

# same harness overlaid on both (sha256:503e8ccf49a7476a…)
cp harness.py tests/e2e/offline_inference/test_mammoth_moda2_tp_parity.py

pytest tests/e2e/offline_inference/test_mammoth_moda2_tp_parity.py \
       -k "tp1 or tp2 or tp4" -vv -s
```

Sanity check that the comparison is clean:

```console
$ git diff --stat f67cc52f7449 refs/remotes/pr/5654
 .../models/test_mammoth_moda2_moe_routing.py       | 114 +++++++++++++++++
 .../models/mammoth_moda2/mammoth_moda2.py          | 136 +++++++++++++++------
 2 files changed, 214 insertions(+), 36 deletions(-)

$ git merge-base refs/remotes/pr/5654 f67cc52f7449
f67cc52f7449d5f5dbb9d7790c7a049fa6aa2ba1
```

The merge base is the PR base itself, so the two builds differ by this PR and
nothing else.

## Per-build results and timings

```
base f67cc52f7449:  1 failed, 50 passed, 25 deselected in 3962.81s (1:06:02)
head 7d6c564cae17:  1 failed, 50 passed, 25 deselected in 3937.39s (1:05:37)
```

Capture wall-clock per engine start, three prompts each:

| engine start | base (s) | head (s) |
| --- | ---: | ---: |
| TP=1 run 1 | 259.9 | 236.0 |
| TP=2 run 1 | 861.5 | 881.1 |
| TP=4 run 1 | 887.9 | 866.4 |
| TP=1 run 2 | 182.8 | 185.7 |
| TP=2 run 2 | 889.5 | 875.0 |
| TP=4 run 2 | 869.5 | 885.8 |

These are capture wall-clock including engine startup and teardown, not a
latency measurement — I am reporting them only so the run is auditable. They are
not comparable to the PR's A/B numbers and should not be read as one: this host
is markedly slower at TP>1 than the machines my earlier runs used (TP=2 at
~880 s here against ~285 s elsewhere), which I assume is topology.

The single failure is the same case with the same value on both sides:

```
FAILED test_ar_stage_output_parity_on_identical_inputs[tp2-terse]
  cosine 0.99859744 < 0.999, median relative error 1.322e-02
```

It is a threshold question in my own harness — cosine is magnitude-weighted and
this prompt drives the massive-activation channels harder — and it reproduces
identically on base, so it is unrelated to this PR.

## Environment

| | |
| --- | --- |
| Hardware | 4x H100 80GB HBM3, single node |
| Driver / CUDA | 570.211.01 / 13.0 |
| Software | vLLM 0.28.0, torch 2.13.0+cu130, Python 3.12.3 |
| Checkpoint | `bytedance-research/MammothModa2-Preview` @ `ef5a5e41dbf0de1ef6275586b7580f0d4248b4c6` |
| Placement | stage 0 sharded over devices `0..N-1`; stage 1 at TP=1 on device 0 in every column |
| Workload | 3 prompts (prefill 30 / 34 / 70 tokens), 256x256, 2 DiT steps, guidance 1.0, seed 42, `temperature=0`, `top_k=1` |
| Feature state | `enforce_eager: true`, `enable_prefix_caching: false`, chunked prefill off, `max_num_seqs: 1` |
| Repetitions | 2 independent engine starts per degree |

Stage 1 is held at TP=1 on device 0 in every column, so it is a control: what
varies between columns is only stage 0's shard count.

## Raw artifacts

36 captures retained (18 per build): generated token ids, float32 hidden states,
and the VAE output per `(degree, prompt, run)`. Happy to attach any subset, or
the per-prompt drift statistics the harness prints, if useful for the record.

## What this does and does not establish

Establishes: on this workload and checkpoint, the code paths reached produce
bit-identical AR output, stage output and image at TP=1, 2 and 4. The production
diff adds `index_select`, `index_copy_`, `where` and `empty` and no arithmetic
operations. Which paths were and were not reached is detailed in the following
comment.

Does not establish: TP=8 (4 GPUs was the most I could get); pipeline
parallelism; behaviour with prefix caching or chunked prefill enabled; mixed
precision between the two experts, which is finding 2 in my review and becomes
reachable with #7334 / #7436.

Kare0638:
Coverage note on the measurement above.

**The 18/18 bitwise comparison did not execute the mixed-token path** — the
`index_select` / `index_copy_` code this PR rewrites.

## Control flow

`mammoth_moda2.py:180-183`:

```python
if gen_pos.numel() == 0:
    return und_expert(hidden_states)
if und_pos.numel() == 0:
    return gen_expert(hidden_states)

gen_pos = gen_pos.to(device=flat_hid.device)      # rewritten path begins here
gen_hid = flat_hid.index_select(0, gen_pos)
```

`get_input_embeddings` at `:460-470` has the same pair of early returns.

## Token composition of the measured runs

| | value |
| --- | --- |
| prompt tokens | 34 / 30 / 70 by prompt, all below `gen_vocab_start_index` (152064) |
| generated tokens | 272, of which 0 are below 152064 |
| `enable_chunked_prefill` | `false` — the prompt is one prefill forward |
| `max_num_seqs` | 1 — decode is one token per forward |

Prefill therefore sees an all-understanding batch and returns at `:180`; each
decode step sees a single generation token and returns at `:182`.

## Applies to the PR's E2E A/B as well

The PR's A/B uses `256x256 T2I, prompt "A cat sitting on a laptop keyboard"`
with 273 generated AR tokens — the same token composition, so the same early
returns apply. The byte-identical PNG and matching token sequence were produced
without reaching `index_select`.

On current evidence the mixed path is covered by the seven new CPU unit tests,
which use identity experts (`hidden_states + 10`, `hidden_states * 2`) and a
mixed mask checked against a `torch.where` reference, in 2D and 3D.

## What the base/head comparison does cover

All of it executed on both builds and was bitwise identical at TP=1/2/4:

- routing construction — two `torch.where` per pipeline stage, hoisted out of the
  per-layer loop
- both early-return branches, including their relocation after the reshape block
- `get_input_embeddings` with the routing tuple threaded through
- per-layer parameter plumbing across 3 TP degrees

## Workloads that would reach the mixed path

1. Image-conditioned generation — a prompt already containing generated visual
   tokens, so prefill is mixed.
2. Chunked prefill enabled with a chunk boundary inside the visual span.
3. `max_num_seqs > 1` with an understanding and a generation request in flight,
   if the batch reaches `moe_forward` as one call.

The same base/head comparison can be run on any of these on request — same
method, same host, roughly one hour.

PR #5662, opened 2026-08-01
[Bugfix][Core] Resync num_waiting_for_streaming_input so an abort cannot park a stage

TheCodeWrangler:
## Purpose

A multi-stage duplex pipeline stops scheduling one of its stages after the
first session, and the stage never recovers. The server stays healthy, accepts
connections, accepts input, and silently produces nothing.

Root cause is an accounting leak in `num_waiting_for_streaming_input`.

Upstream computes

```python
# Scheduler.get_num_unfinished_requests
num_waiting = len(self.waiting) + len(self.skipped_waiting) - self.num_waiting_for_streaming_input
return num_waiting + len(self.running)
```

and maintains that counter incrementally: `+1` when a resumable request runs out
of input and parks as `WAITING_FOR_STREAMING_REQ`, `-1` in
`_update_request_as_session` when the next streaming update arrives, or in
`finish_requests` when the request is removed **while still in that status**.

Omni rewrites `request.status` outside both of those hooks — the chunk-transfer
adapter parks/restores requests via `requests_origin_status`, and
`_realign_request_status_to_queues` rewrites status to match queue membership.
When either lands on a request that is currently counted, the request leaves
`WAITING_FOR_STREAMING_REQ` without the matching decrement, and nothing in
either path ever recomputes the counter.

**One phantom is enough to park the stage permanently.** `EngineCore.has_work()`
consults `has_requests()`, which reads through that counter. Observed with a
single queued request:

```
(1 waiting + 0 skipped - 1 phantom) + 0 running == 0   ->  has_work() == False
```

so the engine blocked in `input_queue.get()` and never called `schedule()`
again, while holding a live request in `waiting`.

### Fix

Rather than chase every status write, restate the invariant the counter is
supposed to encode — how many requests in the waiting queues are parked for
streaming input — and recompute it from that ground truth.

`OmniSchedulerMixin._resync_streaming_input_counter()` is a no-op whenever
upstream's incremental arithmetic is correct (the two agree by construction),
and self-heals when a status rewrite has desynced it. It is `O(len(waiting))`,
bounded by `max_num_seqs`.

Called from:
- both schedulers' `finish_requests` — an abort is the one leak that cannot
  recover on its own, since an inflated counter is precisely what stops
  `schedule()` from running again;
- `OmniARScheduler.schedule`, after the chunk-transfer park/restore surgery and
  while `self.waiting` is still the real queue.

### Note for reviewers

The counter is upstream vLLM's, but the desync is caused by omni-side status
rewrites, so the repair belongs here rather than in vLLM. If you would rather
see the adapter and realign paths decrement explicitly instead, I am happy to
rework it — I chose the recompute because it is invariant-based and cannot be
defeated by a future call site that rewrites status without knowing about the
counter, which is exactly how this arose.

Found while bringing up a second native full-duplex model; nothing in the fix
or the tests is model-specific.

## Test Plan

New: `tests/core/sched/test_omni_scheduler_streaming_input_counter.py`, run
against both `OmniARScheduler` and `OmniGenerationScheduler`:

- the abort case that leaks (counted request whose status was stomped to `RUNNING`)
- the consequence stated as the engine sees it — a fresh request queued behind
  the leak must read as unfinished work
- the resync must **not** over-correct a genuinely parked request
- `skipped_waiting` is counted too, since `get_num_unfinished_requests` sums both
- no-op when upstream's accounting is already correct
- tolerates a scheduler without the attribute
- placement in `schedule()`: resync happens after chunk-transfer surgery and
  before delegating upstream

Plus an end-to-end run on a real 3-stage duplex pipeline, driven by a scripted
client replaying a browser demo's exact wire protocol.

**vLLM Version:** 0.26.0

**vLLM-Omni Commit:** 896f05e7

## Test Result

```
tests/core/sched: 96 passed
```

The 6 new behavioural tests fail without the fix and pass with it (verified by
stubbing `_resync_streaming_input_counter` to a no-op).

End to end, on a freshly booted 3-stage duplex server:

| | before | after |
|---|---|---|
| consecutive sessions answering | 1 / 2 | 5 / 5 |
| turns producing audio (4 sessions x 3 turns) | — | 12 / 12 |

Diagnosis note, since the failure is invisible in logs: every layer looked
healthy and the parked stage logged nothing at all. `py-spy dump` on that
stage's process showed the main loop in `_process_input_queue`, which is only
reachable when `has_work()` is false — i.e. the engine believed it was idle
while holding a queued request.

### DIFF
--- a/tests/core/sched/test_omni_scheduler_streaming_input_counter.py
+++ b/tests/core/sched/test_omni_scheduler_streaming_input_counter.py
@@ -1,0 +1,254 @@
+"""Tests for the ``num_waiting_for_streaming_input`` resync placed in
+``OmniARScheduler.finish_requests`` / ``OmniGenerationScheduler.finish_requests``
+and in ``OmniARScheduler.schedule``.
+
+Upstream keeps that counter incrementally (``+1`` when a resumable request
+parks in a waiting queue as ``WAITING_FOR_STREAMING_REQ``, ``-1`` when the
+next update arrives or the request is finished) and
+``Scheduler.get_num_unfinished_requests`` *subtracts* it from the waiting
+queues. Omni rewrites ``request.status`` outside both hooks -- the
+chunk-transfer adapter's park/restore and
+``_realign_request_status_to_queues`` -- so a counted request can leave
+``WAITING_FOR_STREAMING_REQ`` without the matching decrement and inflate the
+counter for the lifetime of the process.
+
+An inflated counter does not degrade throughput, it hangs the stage:
+``EngineCore.has_work()`` reads false while a live request sits in
+``waiting``, so the engine blocks in ``input_queue.get()`` and never calls
+``schedule()`` again. Regression for the Qwen3-Omni duplex symptom where the
+server answered the first conversation after a boot and was silent from the
+second on while still reporting healthy.
+"""
+
+from __future__ import annotations
+
+import pytest
+from vllm.v1.request import RequestStatus
+
+import vllm_omni.core.sched.omni_ar_scheduler as ar_sched_mod
+import vllm_omni.core.sched.omni_generation_scheduler as gen_sched_mod
+from vllm_omni.core.sched.omni_ar_scheduler import OmniARScheduler
+from vllm_omni.core.sched.omni_generation_scheduler import OmniGenerationScheduler
+
+pytestmark = [pytest.mark.core_model, pytest.mark.cpu]
+
+
+class _StubRequest:
+    def __init__(self, request_id: str, status: RequestStatus) -> None:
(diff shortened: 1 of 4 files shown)


Sy0307:
This PR is still blocked by pre-commit (`N818` at `tests/core/sched/test_omni_scheduler_streaming_input_counter.py:246`, plus ruff-format changes). The tests also monkeypatch upstream `finish_requests()` instead of exercising the real adapter/status-rewrite -> Omni scheduler -> `get_num_unfinished_requests()` path. Please rebase onto the current scheduler stack, fix CI, and add one real abort-to-next-session regression.

Issue #5303, opened 2026-07-22
[New Model]: Gepard-1.0 (nineninesix) — Qwen3.5-backbone streaming TTS over FSQ NanoCodec


mjZhaoElaine:
### The model to consider.

**Gepard-1.0** (nineninesix) — a small, streaming-native autoregressive TTS model
aimed at real-time dialogue synthesis.

- Model: https://huggingface.co/nineninesix/gepard-1.0
- Reference vLLM serving code (author's): https://github.com/nineninesix-ai/gepard-vllm
- Base model: https://huggingface.co/nineninesix/qwen3_5-full-attn-only-14

**Architecture (~555.7M params total):**
- **AR backbone**: Qwen3.5 full-attention decoder — 14 layers, hidden 1024, 8 heads (~500M)
- **Audio codec**: NVIDIA NeMo **NanoCodec** (FSQ-based, 22.05 kHz, 21.5 frames/s).
  Each audio frame = **32 orthogonal FSQ channels sampled in a single step** (one clean
  pass per frame, no sequential codebook decode).
- **Generation**: single-stage autoregressive over FSQ codec tokens; optional CFG
  refinement; zero-shot voice cloning from reference audio.
- **Output**: streaming-native; author reports ~25× real-time on a single RTX 5090.

**Specs:**
- Languages: en, es, pt, nl
- License: **Apache-2.0** (model weights + code). The NanoCodec dependency is under the
  **NVIDIA Open Model License**, which explicitly permits both commercial and
  non-commercial use — so it can be shipped as an optional dependency without licensing
  concerns.
- Traction: ~8.6k downloads / 120+ likes on the Hub in its first month.

### The closest model vllm-omni already supports.

**Qwen3-TTS** — both use a Qwen decoder backbone driving codec-token generation, so much
of the existing talker/serving scaffolding applies. Key differences:

- Gepard uses an **FSQ NanoCodec** where the full frame (32 orthogonal channels) is
  sampled in **one pass per frame**, versus Qwen3-TTS's talker + code-predictor / MTP
  token-by-token pattern. This maps onto the **single-stage-AR** path described in the
  repo's `add-tts-model` skill rather than the multi-stage talker path.
- Different codec (NanoCodec / NeMo) and sample rate (22.05 kHz).

### What's your difficulty of supporting the model you want?

- **Codec dependency**: NanoCodec ships via NeMo
[...]

PR #5146, opened 2026-07-16
[Refactor][OutputProcessor 2/3]: OmniRequestOutput should inherit RequestOutput and no nested wrap-up

bowieshi:
PLEASE FILL IN THE PR DESCRIPTION HERE.

## Purpose

OmniRequestOutput taks RequestOutput as a field. However, RequestOutput is a class that contains the output of a LLM output. Our OmniRequestOutput is designed to be a unified output class for both LLM and diffusion models, which should inherit from RequestOutput instead of containing it as a field. This will allow us to use polymorphism and treat OmniRequestOutput as a RequestOutput when needed, while still having the additional fields for diffusion models.

Related code snippets:
`vllm_omni/outputs.py`: line 62 ~ line 111
before modification:
```python
@dataclass
class OmniRequestOutput:
    request_id: str = ""
    finished: bool = True
    # Pipeline stage fields
    stage_id: int | None = None
    final_output_type: str = "text"
    request_output: RequestOutput | None = None
    # Diffusion model fields
    images: list[Image.Image] = field(default_factory=list)
    prompt: OmniPromptType | None = None
    latents: torch.Tensor | None = None
    trajectory_latents: torch.Tensor | None = None
    trajectory_timesteps: torch.Tensor | None = None
    trajectory_log_probs: torch.Tensor | None = None
    trajectory_decoded: list | None = None
    metrics: dict[str, Any] = field(default_factory=dict)
    _multimodal_output: dict[str, Any] = field(default_factory=dict)
    _custom_output: dict[str, Any] = field(default_factory=dict)
    # profiling data
    stage_durations: dict[str, float] = field(default_factory=dict)
    # memory usage info
    peak_memory_mb: float = 0.0
    # error handling
    error: str | None = None
    error_status_code: int | None = None
    error_type: str | None = None
```

After modification: 
```python
@dataclass
class OmniRequestOutput(RequestOutput):
    # --- Inherited RequestOutput attributes (redeclared as fields) ---
    request_id: str = ""
    prompt: OmniPromptType | None = None
    prompt_token_ids: list[int] | None = None
    prompt_logprobs: Any = None
    outputs: list[CompletionOutput] = field(default_factory=list)
    finished: bool = True
    metrics: Any = field(default_factory=dict)
    lora_request: Any = None
    encoder_prompt: str | None = None
    encoder_prompt_token_ids: list[int] | None = None
    num_cached_tokens: int | None = None
    kv_transfer_params: dict[str, Any] | None = None

[...]

PR #4970, opened 2026-07-08
[Perf][Qwen3-TTS] Drop default seed from qwen3_tts.yaml to restore batched MTP sampling

linyueqian:
## Purpose

Mitigates the Qwen3-TTS side of #4963 (nightly perf regression, failing since the 2026-07-04 run).

Root cause chain:

1. `qwen3_tts.yaml` has carried `seed: 42` in both stages' default sampling params since #1161 (introduced for the disaggregated pipeline, carried into the deploy schema by #2383). It never influenced residual-MTP sampling.
2. #4869 started propagating the deploy-level default seed into every request's `extra_args["tts_local_seed"]`. That put **all** traffic on the explicitly-seeded talker-MTP path, whose multi-row fallback ran one scalar forward per row — the 2026-07-04 nightly collapsed to ~11 audio_throughput at every concurrency (also #4883's timeouts, build 12043 at `0798b0bf`).
3. #4889 kept seeded batches batched via per-row generators, but per-row `torch.multinomial` still costs ~40% audio_throughput and ~2x TTFP/RTF at c>=16 vs the unseeded batched multinomial. That residual gap is what #4963 flags (e.g. base c=16: TTFP 658ms -> 1134ms, RTF 0.365 -> 0.635, audio_throughput 40.3 -> 24.2 comparing the 07-03 and 07-07 nightlies).

## Changes

Remove `seed: 42` from both stages of `vllm_omni/deploy/qwen3_tts.yaml`, matching what #4869 already did for `qwen3_tts_high_concurrency.yaml`. With no default seed, no request carries `tts_local_seed`, so the code predictor returns to the single batched-multinomial fast path — the exact pre-07-04 hot path.

- The nightly perf job launches the server with only the model name (this default profile) and never sends a per-request seed; its WER/quality phase is `enabled: false`, so nothing in CI needs the default determinism.
- Deterministic sampling remains fully available per request via the `seed` field, which stays batched through #4889's per-row generator path.
- Making the *seeded* path itself cheap is complementary follow-up work: #4261 (Gumbel-max sampling) plus potential per-step noise pre-draw.

VoxCPM2's TTFP regression in #4963 starts the same night but is a separate mechanism (the seed propagation is qwen3_tts-gated); prime suspects there are #4606/#4878 and it needs its own bisect.

## Test Plan

**vLLM Version:** 0.24.0

- 2-line YAML deletion; verified the file parses and no test references the deploy-level seed.
[...]

PR #5666, opened 2026-08-01
[Model] Add Gepard-1.0 native-AR FSQ/NanoCodec TTS (offline inference)

mjZhaoElaine:
Adds [**Gepard-1.0**](https://huggingface.co/nineninesix/gepard-1.0) as a single-stage `LLM_AR` pipeline — a native-AR TTS where each step samples one 32-channel FSQ frame and the NeMo NanoCodec decodes the accumulated frames to 22.05 kHz mono. Generation ends on a learned binary stop head, not an EOS token. Part of #5303; serving follows separately.

It rides the native-AR runner hooks rather than patching vLLM: `preprocess` feeds the previous frame's 32 codes back as the next input embedding, `forward` + `compute_logits` do the 32-head Gumbel-max sampling and the stop head, `make_omni_output` drains the codec.

#### Results — offline example, `enforce_eager`, `gpu_memory_utilization=0.3`

Zero-shot, using the model's trained `null_prefix` default voice.

| Card | Samples | Frames | Duration | Peak | Ended on |
|---|---:|---:|---:|---:|---|
| A40 | 47104 | 46 | 2.14 s | 0.343 | stop head |
| RTX PRO 4500 | 43008 | 42 | 1.95 s | 0.277 | stop head |

Every waveform is an exact multiple of 1024 samples — the model's own frame ledger — and every request ends on the stop head rather than running to `max_tokens`. The checkpoint is 1.08 GiB and loads in 1.8 s; AR LM plus codec fit in ~2 GiB, so an L4 is enough.

#### Status

- [x] Offline inference, zero-shot
- [x] Concurrent requests, per-request isolation
- [x] Per-request seed reaches the in-model sampler
- [ ] Serving (`/v1/audio/speech` adapter) — separate PR
- [ ] Voice cloning (`ref_compressor` + codec encode) — separate PR
- [ ] CUDA graph + codec warmup — perf follow-up

#### Tests

- `test_gepard_window.py` — CPU. Decode-window arithmetic against a shift-invariant codec, so "chunked emissions concatenate to exactly one whole-history decode" is an exact assertion.
- `test_gepard_wiring.py` — CPU. Drives the real `preprocess → forward → make_omni_output` through a stand-in for `GPUARModelRunner.execute_model`, and decides delivery with the runner's own `_resolve_sparse_mm_routing`.
- `test_gepard_tts.py` — GPU + `nemo_toolkit[tts]`. 4 passed in 168 s on an RTX PRO 4500: waveform shape, frame conservation, stop-head duration, transcript, four-way concurrent isolation.

**CI tier: weekly only.** `.buildkite/cuda/test-weekly.yml` gains a `TTS · Gepard-1.0 · L4` step that installs `nemo_toolkit[tts]` and then runs the whole file at `--run-level full_model`. The merge and nightly tiers do **not** cover it, so a PR touching Gepard gets no signal from CI before merge — the two CPU files above are what runs there. Its own step rather than a row in the existing mark-selected `TTS · L4`, because the image has no NanoCodec and the module import-skips out of any shared collection.

No accuracy comparison against the reference implementation: its server needs Blackwell/Hopper + CUDA 13 + Postgres, which was not reproducible on the cards used here. The transcript assertion stands in for it, and a real comparison belongs with the serving PR, where one harness can drive both.

#### Dependencies

`nemo_toolkit[tts]`, for the NanoCodec decoder only, imported lazily so nothing else pays for it and `load_format=dummy` still works. Not in `pyproject.toml`: it pins `transformers~=4.57.0` against this repo's `>=5.5.3`, so an extra would be unsatisfiable rather than opt-in. Install line is in the example README, and the weekly step runs that same line with `--overrides`.

The README also notes `VLLM_USE_FLASHINFER_SAMPLER=0` for hosts whose CUDA toolkit cannot build the sampler kernels — not Gepard-specific, but it presents as a model failure inside `profile_run`, and the Qwen3-TTS and JoyAI-VL docs already carry the same note.

<details>
<summary>Two non-obvious things in the diff</summary>

**`meta.sparse_audio` is required, not decorative.** Audio is decoded once per chunk, so most steps produce none — and on a step with an empty payload the AR runner ships the pooler payload instead, whose hidden states the output processor files under this stage's `"audio"` key, straight into the waveform. The marker routes through `_resolve_sparse_mm_routing`, which ships only for the request ids listed in `meta.req_id`.

**The streaming decode window carries lookback.** NanoCodec is convolutional, so decoding a bare chunk boundary is discontinuous; `decode_stream` decodes `[start_idx - lookback, frame_count)` and trims afterwards. `samples_per_frame` is derived from each decode rather than assumed — the real hop is exactly 1024 (the advertised 21.5 fps is 22050/1024 rounded).

**Routing.** The checkpoint self-identifies as `qwen3_5_text`, so `model_type` inference never reaches the gepard registry entry. `gepard.yaml` pins `pipeline: gepard` plus a stage-level `hf_overrides.architectures`; overriding `model_type` does **not** work — the backbone's own `Qwen3_5DecoderLayer` checks it.
</details>

**vLLM Version:** 0.27.0


mjZhaoElaine:
**Question on the optional dependency.**

`nemo_toolkit[tts]` is needed for the NanoCodec decoder. I did not add it to the `dev` extra because it currently constrains `transformers<5`, which conflicts with vLLM 0.25's requirement — installing it into a dev environment breaks `import vllm` until `transformers>=5` is forced back (and that in turn pulls numpy 2.5, which numba rejects, so `numpy<2.5` has to go in the same command).

So for now the e2e module `importorskip`s it and the install is documented in the example README. Happy to add it to `dev`, or to a dedicated extra, if you would rather have it declared — just flagging the conflict first so it is your call rather than a surprise in someone's dev install.

mjZhaoElaine:
Offline inference is ready and CI is green. @linyueqian PTAL when you get a chance — serving will follow in a separate PR.

examples/offline_inference/text_to_speech/gepard/end2end.py
```diff
@@ -0,0 +1,143 @@
...
+        "--output-dir",
+        default=os.path.join(
+            os.environ.get("XDG_CACHE_HOME", os.path.join(os.path.expanduser("~"), ".cache")),
+            "gepard_output",
+        ),
+        help="Directory for WAV outputs (default: ~/.cache/gepard_output).",
+    )
+    parser.add_argument(
+        "--deploy-config",
+        default="gepard.yaml",
```
linyueqian: [blocking] This default does not resolve, so the documented quick start fails.

`load_deploy_config` hands the string straight to `OmegaConf.load`, which opens a bare name relative to the current working directory. Running the README command from the repo root gives:

```
FileNotFoundError: [Errno 2] No such file or directory: '<repo>/gepard.yaml'
```

The help text on the next line says a bare name resolves against `vllm_omni/deploy/`, but nothing on that code path does that. The e2e test avoids it by going through `tests.helpers.stage_config.get_deploy_config_path`, which is test-only.

Since the pipeline pin makes the explicit YAML mandatory here, resolving it in the example seems simplest:

```python
from pathlib import Path
import vllm_omni

_DEFAULT_DEPLOY = str(Path(vllm_omni.__file__).resolve().parent / "deploy" / "gepard.yaml")
```

Everything else I ran below used an absolute path to work around this.

vllm_omni/model_executor/models/gepard/gepard_talker.py
```diff
@@ -0,0 +1,485 @@
...
+    def _flush_deferred_cleanup(self) -> None:
+        for req_id in self._deferred_cleanup_ids:
+            # A request can finish without our STOP ever firing (max_tokens,
+            # abort). Drain whatever frames it accumulated before dropping the
+            # state, else the tail of the clip is silently lost.
+            state = self._active_states.get(req_id)
+            if state is not None:
+                self._emit_audio(state, is_final=True)
+            self._active_states.pop(req_id, None)
+        self._deferred_cleanup_ids.clear()
```
linyueqian, line 231: [blocking] This flush never runs for the request it is meant to save, so the tail is not delayed, it is dropped.

`on_requests_finished` only queues the id, and `_flush_deferred_cleanup` is called at the end of `forward` (line 417). For the last in-flight request there is no next forward, so this loop never executes at all. And when there is a next step, the id is no longer in `req_ids_output_copy`, so `_resolve_sparse_mm_routing` filters the payload out anyway.

Reproduced by lowering `max_tokens` in `gepard.yaml` and changing nothing else:

| max_tokens | samples returned | frames |
|---|---|---|
| 30 | 10240 | 10 |
| 100 | 98304 | 96 |
| 150 | 98304 | 96 |

96 is `FIRST_CHUNK_FRAMES + CHUNK_FRAMES`. Output is always cut back to the last cadence boundary, so worst case is 86 frames, roughly 4 seconds of speech, with no error or warning. I added a log line right here and confirmed it never fires.

VoxCPM2 gets away with the same deferred-cleanup shape because it emits every step, so its worst case is one frame. The 86-frame chunk is what turns the same latency into seconds. Whatever the fix, it has to emit while the request is still live rather than waiting on the finished notification, for example by tracking the per-request token budget inside `forward`.

tests/model_executor/models/test_gepard_window.py
```diff
@@ -0,0 +1,265 @@
...
+    num_frames = FIRST_CHUNK_FRAMES + CHUNK_FRAMES + 5
+    talker._active_states["req-0"] = state
+
+    _generate(talker, state, num_frames, stop_at_end=False)
+    assert state.emitted_frames < num_frames, "a partial chunk should still be pending"
+
+    talker._deferred_cleanup_ids.add("req-0")  # on_requests_finished's effect
+    talker._flush_deferred_cleanup()
+
+    torch.testing.assert_close(_drain(talker), _expected_samples(num_frames))
```
linyueqian, line 205: [important] This passes even though the path it is named after does not work in the engine.

It calls `_flush_deferred_cleanup` directly and reads `_audio_queue` by hand, so all it really checks is that `_emit_audio` computes the right samples once someone calls it. The two things that actually break are whether anything calls the flush, and whether a payload for an already-finished request survives routing. Neither is exercised.

Either rename it to what it covers, or drive it through the runner so it goes red until the issue above is fixed.

vllm_omni/model_executor/models/gepard/gepard_talker.py
```diff
@@ -0,0 +1,485 @@
...
+            gumbel = -torch.log(-torch.log(u))
+            heads = (gathered / self.temperature + gumbel).argmax(dim=-1)  # (N,31)
+
+        h0_logits = fused[:, : self.head0_vocab].float()  # (N,8)
+        if self._greedy:
+            head0 = h0_logits.argmax(dim=-1)
+        else:
+            u0 = torch.rand_like(h0_logits).clamp_min_(1e-20)
+            g0 = -torch.log(-torch.log(u0))
+            head0 = (h0_logits / self.temperature + g0).argmax(dim=-1)  # (N,)
```
linyueqian: [important] These draw from the global RNG, so no seed a caller sets ever reaches the audio. `SamplingParams.seed` only feeds vLLM's sampler, which is already deterministic here because the pipeline pins temperature to 0.

Two identical requests in the same process gave me 40 frames at peak 0.288, then 49 frames at peak 0.380. Separate process launches do match, but only because the global RNG starts from the same state, which is not something a server can lean on.

The practical problem is that `seed: 42` in `gepard.yaml` reads like it makes generation reproducible. Either thread the per-request seed into a `torch.Generator` here, or drop the seed from the YAML and say in the README that output is not reproducible yet.

examples/offline_inference/text_to_speech/README.md
```diff
@@ -401,6 +402,28 @@ python examples/offline_inference/text_to_speech/qwen3_tts/end2end.py \
 
 ---
 
+## Gepard-1.0
+
+Single-stage native AR TTS at 22.05 kHz. Pipeline: `Qwen3.5 backbone → 32 FSQ codebook heads (one frame/step) → NeMo NanoCodec`. Backbone runs under vLLM paged attention; the 32-head sampling + learned embedding feedback ride the native-AR runner hooks. `enforce_eager` (CUDA graph is a perf follow-up).
+
+### Prerequisites
+```bash
+pip install nemo_toolkit[tts]   # NanoCodec decoder (NVIDIA Open Model License)
```
linyueqian: [important] This command breaks the environment it runs in, which matches what you flagged in the PR comment. `nemo_toolkit` pins `transformers~=4.57.0`, so pip resolves it by downgrading transformers 5.14.1 to 4.57.6, and `import vllm` fails afterwards. Confirmed with a dry run against a clean vLLM environment.

On where to declare it, I would keep it exactly where you have it, out of `pyproject.toml`. There is already precedent a few hundred lines up in this same file: SoulX-Singer English SVS documents `pip install "nemo_toolkit[asr]==2.6.1"` with the same clash, since that release pins `transformers~=4.53.0`. The two alternatives do not actually work:

- `dev` extra: `requirements/common.txt` pins `transformers >= 5.5.3`, so every dev install downgrades and breaks vLLM.
- dedicated extra: extras resolve together with the base deps, so `pip install vllm-omni[gepard]` is an unsatisfiable resolution rather than an opt-in.

The useful part is that the NeMo pin turns out not to be load-bearing for the decode path. I ran the full e2e with `nemo_toolkit` 2.7.3 and transformers 5.14.1 held in place, and the codec import, the decode and the audio were all fine. So I would pin the version and document the constraint instead of the bare install:

```bash
pip install "nemo_toolkit[tts]==2.7.3" --constraint <(echo "transformers>=5.5.3")
```

examples/offline_inference/text_to_speech/gepard/end2end.py
```diff
@@ -0,0 +1,143 @@
...
+                save_audio(audio.cpu(), str(output_dir / f"output_{i}_{j}.wav"), sr)
+
+    print("Done.")
+
+
+def parse_args():
+    parser = TrackingArgumentParser(description="Gepard-1.0 offline TTS inference (zero-shot)")
+    parser.add_argument("--text", default="Hello, this is Gepard speaking.", help="Text to synthesize.")
+    parser.add_argument("--max-new-frames", type=int, default=1000, help="Max AR frames (21.5 fps).")
+    parser.add_argument("--seed", type=int, default=None, help="Random seed.")
```
linyueqian: [suggestion] Both of these are dead. They get packed into `additional_information` at line 54 and nothing under `vllm_omni/model_executor/models/gepard` reads either key. `config.max_frames` is unused as well. The real cap is `max_tokens` in `gepard.yaml`, and per the talker comment, hitting it truncates the clip without saying so.

Worth wiring them up or dropping them, otherwise `--max-new-frames 200` looks like it does something.

linyueqian (commented):
Ran this end to end at `bc1471b9f` on a single H200-class card. The native-AR hook usage reads clean, and the two non-obvious things you called out in the description are both doing real work: every waveform that came back was an exact multiple of 1024 samples, and concurrent requests stayed isolated.

What passed:

- `tests/model_executor/models/test_gepard_window.py`: 20 passed
- `tests/e2e/offline_inference/test_gepard_tts.py`: 2 passed
- offline example: 40960 samples (40 frames), 1.86 s, peak 0.288; whisper-small reads it back as `"Hello, this is Gopard speaking."`
- three concurrent requests: no cross-talk, each transcript matches its own prompt, all whole-frame

Two things I think want fixing before this lands. The quick start in the README does not run as written, and audio is silently truncated whenever a request finishes without your STOP head firing. Both reproduce every time, details inline. Your dependency question is answered inline on the README too, short version is that I would keep it exactly where you put it.

One environment note that is not on you: main needs vLLM 0.26.0, not the 0.25.0 in the description. `omni_ar_scheduler.py` reads `request.num_in_flight_tokens`, which landed in `cbab1d4d8`, so on 0.25.0 the engine dies on the first scheduler step.

examples/offline_inference/text_to_speech/README.md
```diff
@@ -16,6 +16,7 @@ list of supported architectures across all modalities, see
 |---|---|---|---|---|---|---|
 | CosyVoice3 | `FunAudioLLM/Fun-CosyVoice3-0.5B-2512` | 2 (talker + code2wav) | ✓ | ✓ | — | 24 kHz |
 | Fish Speech S2 Pro | `fishaudio/s2-pro` | dual-AR | ✓ | ✓ | — | 44.1 kHz |
+| Gepard-1.0 | `nineninesix/gepard-1.0` | single (native AR + NanoCodec) | — (zero-shot; cloning WIP) | ✓ (online) | zero-shot | 22.05 kHz |
```
linyueqian: [suggestion] The streaming column claims online support, but this PR is offline only and serving is the follow-up. A dash until that lands would keep the table from promising an endpoint that is not there yet.

vllm_omni/model_executor/models/gepard/prompt.py
```diff
@@ -0,0 +1,44 @@
...
+    *,
+    start_of_text: int,
+    end_of_text: int,
+    start_of_speech: int,
+    speaker_token_base: int,
+    num_speaker_prefix: int,
+    target_text_tokens: int = 16,
+    apply_below: int = 13,
+    max_repeats: int = 8,
+    repetition_enabled: bool = True,
```
linyueqian: [suggestion] `gepard_config.json` carries a `text_repetition` block with exactly these numbers (`target_text_tokens: 16`, `apply_below: 13`, `max_repeats: 8`, `enabled: true`), and it is the one sidecar section `GepardConfig` does not parse. The values match today's checkpoint so nothing is wrong right now, but the docstring says this has to match the training layout or WER collapses, which is a good argument for reading it from the config rather than duplicating it here.

vllm_omni/model_executor/models/gepard/gepard_talker.py
```diff
@@ -0,0 +1,485 @@
...
+        cfg = vllm_config.model_config.hf_config
+        if not isinstance(cfg, GepardConfig):
+            # hf_overrides re-routes the model class but patches the loaded
+            # config instance rather than re-classing it, so we get the
+            # checkpoint's backbone config. Rebuild the real GepardConfig.
+            cfg = GepardConfig.from_checkpoint(vllm_config.model_config.model, backbone_config=cfg.to_dict())
+        self.config = cfg
+        self._device = current_omni_platform.get_torch_device()
+        # Deterministic argmax sampling, for reproducible comparisons.
+        self._greedy = os.environ.get("TTS_GREEDY", "0") == "1"
```
linyueqian: [suggestion] `TTS_GREEDY` is unscoped while the other four knobs in this file and in `nanocodec.py` all use `VLLM_GEPARD_*`. `VLLM_GEPARD_GREEDY` would keep it consistent and avoid picking up someone else's variable in a shared environment.

mjZhaoElaine:
Four questions before this lands. The first three are platform-level and outlive this PR.

**1. Should `sampling_constraints` merge over caller-supplied `SamplingParams` rather than be replaced by them?**
`resolve_sampling_params_list` (`entrypoints/omni_base.py:269`) returns caller params verbatim; constraints only reach the stage defaults (`config/stage_config.py:873`). So any caller-supplied `SamplingParams` — even just a seed — drops Gepard's `stop_token_ids`, and the request runs to `max_tokens` (~46 s) instead of stopping. 23 pipelines declare `sampling_constraints`, so this is not Gepard-specific. Deliberate, or worth a separate PR?

**2. Where should `nemo_toolkit[tts]` be declared, and where should the e2e test run?**
It pins `transformers~=4.57.0` against the repo's `>=5.5.3`, so a `[gepard]` extra would be unsatisfiable rather than opt-in; it is documented in the example README instead, as SoulX-Singer does for `nemo_toolkit[asr]`. The test therefore skips at import and I added no `test-merge.yml` step, since an empty file-scoped collection exits 5. Which do you want: (a) the image gains the dep and I add the merge step here, (b) nightly/weekly only, and the `advanced_model` mark comes off, or (c) leave it uncovered with the `TO CONFIRM` note?

**3. Should `gepard.yaml` keep `seed: 42`?**
Gepard samples all 32 heads in `forward`, so this PR threads the seed into a per-request generator. The YAML default therefore makes every request byte-identical and forces a per-row Gumbel draw instead of a batched one — the same cost #4970 cited when dropping this line from `qwen3_tts.yaml`. Keep or drop?

**4. Recompute preemption breaks native-AR TTS generally — per-model warning, or a scheduler-level opt-out?**
A frame's other 31 codes never enter the token stream, so a re-prefill cannot rebuild the KV behind audio already delivered and generation restarts from frame 0. Not fixable inside the model. This PR detects it and warns; voxcpm2 has the same exposure.

mjZhaoElaine:
All nine addressed as of `5ddbc76e`, in your order.

**Deploy-config default** — resolved to the packaged path exactly as you wrote it. The underlying bug survives, though: `AsyncOmniEngine.__init__` hands the raw string to `load_deploy_config` without going through `_resolve_deploy_path` (`config/omni_config.py:220`), which already does this resolution. Any model taking a bare name hits it. Left alone as out of scope.

**Dropped tail** — fixed the way you suggested. The runner threads `max_tokens` into `preprocess`, and `forward` flushes on the request's final step, while it is still in the output batch.

**Window test** — replaced by `tests/model_executor/models/test_gepard_wiring.py`, which drives the real `preprocess → forward → make_omni_output` through a stand-in for `GPUARModelRunner.execute_model` and decides delivery with the runner's own `_resolve_sparse_mm_routing`. It goes red on the old behaviour.

**Seed** — threaded into a per-request `torch.Generator`; a seeded row draws its own Gumbel noise. Whether the YAML default should survive that is question 3 in my comment below.

**NeMo install** — pinned to 2.7.3 and kept out of `pyproject.toml`, as you advised. One correction: `--constraint` cannot express this. `nemo_toolkit[tts]` *declares* `transformers~=4.57.0`, and a constraint narrows a range rather than overriding a declared dependency, so that form dies with `ResolutionImpossible`. `--overrides` does work, and `huggingface-hub` needs naming too — transformers 5.x pulling it back is otherwise an accident of the first override:

```bash
uv pip install "nemo_toolkit[tts]==2.7.3" \
    --overrides <(printf '%s\n' "transformers>=5.5.3" "huggingface-hub>=1.0")
```

**Dead CLI flags** — dropped. **Streaming column** — now a dash. **`text_repetition`** — parsed from the sidecar; `prompt.py` takes the config, and the example and e2e tests build prompts through `GepardConfig.from_checkpoint()`, so both sides read one source. **`TTS_GREEDY`** — now `VLLM_GEPARD_GREEDY`.

Description corrected to 0.26.0 — thanks for catching that, and for running it end to end.

mjZhaoElaine:
The failing docs check is not from this PR, and it will hit every new build.

`.readthedocs.yml`'s `pre_install` passes `--index-url https://download.pytorch.org/whl/cpu`, which *replaces* PyPI rather than adding to it. pip now discards the `typing-extensions` 4.15.0 wheel — `inconsistent Name: expected 'typing-extensions', but metadata has 'typing_extensions'` — falls back to the sdist, and cannot resolve its build dependency:

```
ERROR: Could not find a version that satisfies the requirement flit_core<4,>=3.11 (from versions: none)
ERROR: No matching distribution found for flit_core<4,>=3.11
```

Same command, same error on #6124 ([build 34038580](https://app.readthedocs.org/projects/vllm-omni/builds/34038580/)) and here ([build 34038761](https://app.readthedocs.org/projects/vllm-omni/builds/34038761/)). 

Here's the separate PR to fix the CI failure in main: https://github.com/vllm-project/vllm-omni/pull/6129

tests/e2e/offline_inference/test_gepard_tts.py
```diff
@@ -0,0 +1,348 @@
...
+    defaults rather than merge over them, dropping the pipeline's stop token.
+    """
+    outputs = omni.generate([_build_request(t) for t in texts])
+    assert len(outputs) == len(texts), f"expected {len(texts)} outputs, got {len(outputs)}"
+
+    by_index: dict[int, _Clip] = {}
+    for stage_outputs in outputs:
+        index = int(str(stage_outputs.request_id).split("_", 1)[0])
+        # OmniRequestOutput.request_output is a single RequestOutput, not a list.
+        req_output = stage_outputs.request_output
```
linyueqian: [blocking] `OmniRequestOutput` no longer wraps a nested `RequestOutput`, so every e2e test fails at this head:

```
AttributeError: 'OmniRequestOutput' object has no attribute 'request_output'
```

Main's `fcf108539` ("[Refactor][OutputProcessor 2/8]: OmniRequestOutput should inherit RequestOutput and no nested wrap-up", #5146) landed on 2026-08-12, one day before this head, and the branch rebased past it without migrating. `OmniRequestOutput` now subclasses `RequestOutput` directly, so `.outputs` is on the object itself. `voxcpm2/end2end.py` already reads it the flat way.

Running the suite at `8bc4644ed`:

```
3 failed, 1 deselected in 201.26s
```

all three with that same AttributeError. Changing this line to `req_output = stage_outputs` (and the same in the example) gives `3 passed in 161.05s`, so the stale accessor is the only thing wrong underneath. The description's "4 passed in 291 s" must predate the rebase.

Worth re-running the suite after the fix rather than trusting the earlier number, since this is exactly the class of breakage a rebase introduces silently.

examples/offline_inference/text_to_speech/gepard/end2end.py
```diff
@@ -0,0 +1,142 @@
...
+    # defaults rather than merge over them, dropping the pipeline's stop token.
+    output_dir = Path(args.output_dir)
+    output_dir.mkdir(parents=True, exist_ok=True)
+
+    print(f"Synthesizing: {args.text!r}")
+    inputs = build_request(text=args.text)
+
+    for i, stage_outputs in enumerate(omni.generate(inputs)):
+        # OmniRequestOutput.request_output is a single RequestOutput, not a list.
+        req_output = stage_outputs.request_output
```
linyueqian: [blocking] Same `.request_output` removal as the test file, so the shipped example also dies here before writing any audio. `req_output = stage_outputs` fixes it; the `if req_output is not None` guard below can go too, since `generate` always yields an output object now.

Once both are migrated the example runs clean: 41984 samples, 1.90 s, and whisper reads it back as the prompt.

.buildkite/cuda/test-merge.yml
```diff
@@ -103,6 +103,10 @@ steps:
             "
         mirror_hardwares: l4_1
 
+      # TO CONFIRM: should we add Gepard-1.0 to nightly/weekly step? If yes,
+      # will need nemo_toolkit[tts] to the image. Currently no Gepard-1.0 step
+      # here
```
linyueqian: [question] This asks a question of the reviewers from inside a CI config, so it will get merged and then forgotten. Better answered in the PR thread and dropped from the file.

For what it is worth, the answer matters more than it looks: with no Gepard step anywhere, the offline tests run in no tier at all, which is why the `.request_output` breakage above reached this head undetected. The image genuinely lacks `nemo_toolkit[tts]` and I would not block the PR on adding it, but it is worth deciding now whether that install lands in the nightly image or whether these tests stay local-only for a while, and saying so in the description so nobody later assumes CI is covering them.

linyueqian (commented):
Round 2 at `8bc4644ed`, on a single H200-class card with vLLM 0.27.0, transformers 5.15.0 and nemo_toolkit 2.7.3.

All nine findings from the last round are closed, and I re-ran the two blocking ones on hardware rather than taking the diff's word for it:

- The quick start now works with no `--deploy-config` at all: 41984 samples, 1.90 s.
- Budget-truncated requests deliver everything. `max_tokens` 30, 100 and 150 return exactly 30, 100 and 150 frames. Last round the same three returned 10, 96 and 96.
- A seeded request is reproducible: runs 2 through N in one process are bit-identical. Only the very first request in a fresh process differs, by 1.8e-4 max abs on a peak-0.28 signal, which looks like first-call kernel selection rather than seeding. Worth knowing if someone tests the docstring's claim.
- Your NeMo install line works verbatim: transformers stayed at 5.15.0 and `import vllm` still worked afterwards.

`_MiniEngine` is the part I'd single out. I reverted just `is_final=do_stop or is_last_token` back to the old behaviour and 8 tests went red, including the exact 30/100/150 budgets I measured on the GPU. That is a real regression test for the bug, not a test written around the fix. Same for `_flush_deferred_cleanup` no longer pretending it can save the tail: cleanup plus a warning for the cases that genuinely cannot be recovered is the honest shape.

One blocker left, and it is a rebase casualty rather than anything you designed wrong. Details inline.

Small thing on the description: it says vLLM 0.26.0, but the head is rebased past `2ce15ab7 [Rebase] Rebase to vllm 0.27.0 (#5976)`, so 0.27.0 is what it needs.

.buildkite/cuda/test-merge.yml
linyueqian: we can add it in the ci weekly test

congw729: > we can add it in the ci weekly test

Agree.

tests/e2e/offline_inference/test_gepard_tts.py
mjZhaoElaine: Fixed at `ddf1a9cf`. Re-ran rather than trusting the pre-rebase number: `4 passed in 168.22s`, `load_format=auto`, on an RTX PRO 4500.

Dropped the `assert req_output is not None` along with the accessor — `outputs` is a plain list now, and the empty case is already covered by the `no audio output produced` assertion below it.

examples/offline_inference/text_to_speech/gepard/end2end.py
mjZhaoElaine: Fixed, and the guard dropped as you suggested. Example output: `(44032,), 22050 Hz, 2.00s, peak 0.273` — 43 whole frames.

.buildkite/cuda/test-merge.yml
mjZhaoElaine: Note dropped, and weekly it is. `.buildkite/cuda/test-weekly.yml` gains `TTS · Gepard-1.0 · L4`: installs `nemo_toolkit[tts]` with the README overrides, then runs the whole file at `--run-level full_model`. Its own step rather than a row in `TTS · L4`, since the image has no NanoCodec and the module import-skips out of that step's mark-based collection.

Ran the step's four commands verbatim on an RTX PRO 4500 — not in CI, so this is a local stand-in for the weekly job: the install audits clean against an existing env, then `4 passed`.

The description now says weekly-only in as many words, including that merge and nightly do not cover the file, so nobody later reads pre-merge CI as coverage.

linyueqian (approved):
Both round-2 items are fixed and I re-verified them at `ddf1a9cf2` on a single H200-class card, vLLM 0.27.0 with transformers 5.15.0 and nemo_toolkit 2.7.3.

The flat `OmniRequestOutput` read lands correctly: the e2e suite is `4 passed in 334 s` with nothing patched on my side, and the shipped example runs straight through to 41984 samples at 1.90 s. CPU suites stay at 53 passed. The weekly step is the right call given the image has no NanoCodec, and installing it inside the step with the same overrides the README documents keeps the two in sync rather than leaving a second recipe to drift.

Adding up across the three rounds: eleven findings, all closed, and the two that mattered were closed properly rather than papered over. The budget-tail fix is the one I would point at. `_MiniEngine` reproduces the exact orderings the bug lived in, and reverting the one-line fix turns 8 tests red, including the same 30/100/150 budgets I measured on hardware. That is a regression test that would have caught the original bug, which is a higher bar than most model PRs clear.

Approving. Adding `ready` to trigger Buildkite, which is worth watching since the offline tests have never run in CI on this branch.

PR #5667, opened 2026-08-01
[Perf][ERNIE-Image] Avoid timestep scalar synchronization

EchoHayate:
## Purpose

Remove an avoidable host/device synchronization from the ERNIE-Image
denoising loop.

Each scheduler timestep is already a device scalar. The previous path called
`t.item()` and then constructed a new device tensor with
`torch.full(...)`. Reading the scalar through Python invokes
`aten::_local_scalar_dense`, which synchronizes the CUDA stream once per
denoising step.

## Root cause

The timestep batch only needs to repeat one scalar for the current transformer
input batch:

1. the scheduler creates `t` on the execution device;
2. `t.item()` copies/synchronizes it back to the host;
3. `torch.full(...)` sends the same value back to the device.

This happens for both CFG and non-CFG inference.

## Solution

- Add `expand_scalar_to_batch()`, which expands a scalar tensor directly on
  its current device and converts the dtype without a Python scalar read.
- Derive the timestep batch size from `latent_model_input.shape[0]`, covering
  both CFG and non-CFG paths.
- Add a regression test that makes every `Tensor.item()` call fail and checks
  the output shape, dtype, and values.

The implementation follows the device-resident timestep expansion pattern
already used by the Krea2 diffusion pipeline.

## Historical performance evidence

The following results were measured for the unchanged production patch at
commit `4553f1707ac7f5e66c9f2c8340afdf6c4baa8603`. They are not a rerun of the
current refreshed head.

### Targeted microbenchmark

- NVIDIA A100 80GB PCIe
- PyTorch 2.5.1+cu121
- 2,000 expansions per trial, 100 warmups, 10 trials
- median host-observed time per expansion, including a final CUDA synchronize

| Batch | Before | After | Speedup |
| ---: | ---: | ---: | ---: |
| 1 | 20.2573 us | 11.5145 us | 1.76x |
| 2 | 19.6727 us | 11.9433 us | 1.65x |
| 8 | 20.9347 us | 11.8510 us | 1.77x |
| 64 | 19.8597 us | 12.4419 us | 1.60x |

Profiler count over 50 expansions:

| Metric | Before | After |
| --- | ---: | ---: |
| `aten::_local_scalar_dense` | 50 | 0 |

### Controlled ERNIE-Image E2E

- 2x NVIDIA A100 80GB PCIe
- `baidu/ERNIE-Image`, revision
  `5346b31d68c9c23758ba56ef8be5e9dc174c7f99`
- 512x512, 2 denoising steps, guidance scale 4.0, seed 42
- tensor parallel size 2, eager mode, CPU model offload
- prompt enhancement disabled with `apply_pe=false` to make the model input
  deterministic
- five generations in one loaded process per revision

| Metric | Base | PR |
| --- | ---: | ---: |
| Steady-state median, iterations 2-5 | 4.9644 s | 4.7839 s |
| Directional median change | - | -3.63% |
| Peak GPU 1 memory | 23,036 MiB | 23,012 MiB |
| Peak GPU 2 memory | 19,142 MiB | 19,142 MiB |

All five base/head PNG pairs were byte-identical and pixel-identical. The
short two-step workload is noisy, so this is supporting model-level evidence,
not a broad throughput claim.

## Current-main refresh

Current head: `cf7374759f1087b1b557bdc1f7ace01e1f2a9f93`

Merged upstream main
`f67cc52f7449d5f5dbb9d7790c7a049fa6aa2ba1` without changing the production
optimization. The current-main net diff remains three focused files.

Local validation on the refreshed head:

```bash
PYTHONPATH="$PWD:/Users/bytedance/Worktrees/vllm/v0.28.0" \
  /Users/bytedance/Desktop/vllm-omni/.venv/bin/python -m pytest \
  -o addopts='' tests/diffusion/utils/test_tensor_utils.py -q

uvx --from pre-commit pre-commit run --files \
  tests/diffusion/utils/test_tensor_utils.py \
  vllm_omni/diffusion/utils/tensor_utils.py \
  vllm_omni/diffusion/models/ernie_image/pipeline_ernie_image.py

python3 -m compileall -q \
  tests/diffusion/utils/test_tensor_utils.py \
  vllm_omni/diffusion/utils/tensor_utils.py \
  vllm_omni/diffusion/models/ernie_image/pipeline_ernie_image.py

git diff origin/main...HEAD --check
```

Results:

- focused regression: `1 passed, 15 warnings`
- changed-file pre-commit: passed
- compileall: passed
- diff check: passed
- diff-scoped static review: no P0-P2 findings

## Evidence boundary

The production patch is unchanged from the revision used for the historical
2xA100 E2E run, but the merge commit and surrounding current-main code have
not been rerun on GPU. The current-head claim is therefore limited to local
regression/static validation plus historical identical-patch GPU evidence.

## AI assistance

Codex assisted with the current-main refresh, test execution, diff-scoped
self-review, and drafting this update. I reviewed the resulting diff and
remain responsible for the implementation and claims above.


EchoHayate:
## Full-model validation update (2026-08-11)

I reran this change with the real `baidu/ERNIE-Image` checkpoint instead of only the scalar-expansion microbenchmark.

### Environment and source pins

- GPU: 2x NVIDIA A100 80GB PCIe
- Model revision: `5346b31d68c9c23758ba56ef8be5e9dc174c7f99`
- Base commit: `185e3d6df60276166349e69fdc1171fd0e50c79f`
- PR commit: `4553f1707ac7f5e66c9f2c8340afdf6c4baa8603`
- Checkpoint validation: 24 snapshot entries, 0 broken links, 5 safetensors headers readable, 14 JSON files parsed; the four large PE/text-encoder/transformer blobs were checked against exact sizes and SHA-256 values.

### Workload

- Prompt: `A photo of a cat sitting on a laptop`
- 512x512, 2 denoising steps, guidance scale 4.0, seed 42
- tensor parallel size 2, eager mode, CPU model offload
- 5 generations in one loaded process for each commit

ERNIE-Image prompt enhancement samples text by default. An initial run therefore produced different enhanced prompts for base and head, so those images were not a valid correctness comparison. I reran both commits with the supported request option `apply_pe=false`, keeping the actual diffusion/text-encoder/VAE path unchanged while making the model input deterministic.

### Controlled results (`apply_pe=false`)

| Metric | Base | PR |
| --- | ---: | ---: |
| All 5 runs completed | yes | yes |
| Steady-state median, iterations 2-5 | 4.9644 s | 4.7839 s |
| Observed median change | - | -3.63% |
| Fastest iteration | 4.1638 s | 4.5230 s |
| Peak GPU 5 memory | 23,036 MiB | 23,012 MiB |
| Peak GPU 7 memory | 19,142 MiB | 19,142 MiB |

Correctness:

- Base and PR produced the same PNG SHA-256 on all 5 iterations: `6b9b5b0663c0f9a8d0a5d5788c809f5baa3332f989fe93475ded057ef10d7d30`.
- All 5 base/head pairs were byte-identical and pixel-identical with the same shape.

This closes the previous full-model validation gap. The run confirms real-checkpoint image parity and no material memory increase. The measured ~3.6% steady-state median improvement is consistent with removing the scalar synchronization, but this short 2-step workload is noisy (one base iteration was an outlier), so I am treating it as supporting E2E evidence rather than a broad throughput claim.

EchoHayate:
## Author self-review

- Scope/invariants checked: the net diff remains limited to device-resident
  timestep expansion for ERNIE-Image plus its focused regression test; CFG and
  non-CFG both derive the timestep batch from the actual transformer input.
- Unsupported combinations: this change does not alter model support,
  parallelism, offload, or checkpoint-loading behavior.
- Tests and static checks: focused regression `1 passed, 15 warnings`;
  changed-file pre-commit, compileall, and `git diff origin/main...HEAD
  --check` passed. A diff-scoped review found no P0-P2 issues.
- Runtime evidence: the unchanged production patch previously completed a
  controlled 2xA100 ERNIE-Image E2E with five byte/pixel-identical output
  pairs and a directional steady-state median change of `-3.63%`.
- Evidence boundary: the historical GPU run was at `4553f170`; the refreshed
  merge head was not rerun on GPU, so I am not presenting it as current-head
  hardware evidence.
- AI assistance: Codex assisted with the refresh, verification, review, and
  wording. I reviewed the diff and validation results.

Current head reviewed: `cf7374759f1087b1b557bdc1f7ace01e1f2a9f93`.

PR #5670, opened 2026-08-01
[Bugfix] Invalidate stale ref_audio cache on local file modification (#4873)

### LINKED ISSUE #4873: [Bug]:  Stale Reference Audio Resolve Cache When `ref_audio` Locator Content Changes
### Your current environment

any env

### Your code version

master

### 🐛 Describe the bug

# Stale Reference Audio Resolve Cache When `ref_audio` Locator Content Changes

## Summary

`_resolve_ref_audio()` currently caches decoded reference audio by hashing the original `ref_audio` string. This works well for immutable `data:` URLs, but it can return stale waveform data when a stable locator, such as a local file path or URL, points to changed audio content.

This is independent from the MOSS reference-code cache. The reference-code cache is content-addressed after audio resolution, but it cannot detect updates if `_resolve_ref_audio()` returns an already-cached old waveform before the new content is read.

## Affected Area

- `vllm_omni/entrypoints/openai/serving_speech.py`
- `OmniOpenAIServingSpeech._resolve_ref_audio`
- Speech models that consume `request.ref_audio`, including MOSS-TTS Local voice cloning

## Current Behavior

`_resolve_ref_audio()` uses the original `ref_audio_str` as the cache identity:

```python
cache_key = hashlib.sha1(ref_audio_str.encode("utf-8")).hexdigest()
cached = self._ref_audio_resolve_cache.get(cache_key)
if cached is not None:
    return wav_list, sr
```

For locator-style inputs, the string identifies where to load audio from, not necessarily the audio content itself.

For example:

1. Request uses `ref_audio="file:///tmp/speaker.wav"`.
2. The file currently contains audio content `A`.
3. `_resolve_ref_audio()` decodes and caches `wav(A)` under `sha1("file:///tmp/speaker.wav")`.
4. The file is overwritten with audio content `B`.
5. A later request uses the same `ref_audio="file:///tmp/speaker.wav"`.
6. `_resolve_ref_audio()` hits the old string-keyed cache and returns `wav(A)`.
7. Downstream reference-code caches only see `wav(A)`, so they cannot detect that the file now contains `B`.

## Expected Behavior

When a mutable reference locator points to changed audio content, the server should avoid returning stale decoded waveform data.

For local files, cache identity should account for file content or a safe invalidation signal. For remote URLs, either the existing string-keyed semantics should be documented as best-effort, or the cache should be bypassed/revalidated depending on the chosen policy.

## Why This Matters

This can cause voice-cloning requests to use an old speaker reference even after the user updates the referenced file or URL content.

The issue is subtle because downstream MOSS reference-code caching may be content-addressed correctly, yet it receives stale input from the earlier resolve cache.

## Relationship To MOSS Reference-Code Cache

The MOSS reference-code cache and `_resolve_ref_audio()` cache operate at different layers:

```text
ref_audio locator
  -> _resolve_ref_audio cache
     stores decoded waveform
  -> MOSS reference encoder cache
     stores encoded reference codes
```

The new MOSS reference-code cache does not duplicate `_resolve_ref_audio()` and does not introduce this stale-hit behavior. It only makes the existing resolve-cache freshness semantics more visible.

### Option 1: Use Content-Aware Cache Keys For Local Files

For `file://` or local-path references, derive the resolve-cache key from file metadata plus a content sentinel or full content hash. This mirrors the approach used by sglang-omni's reference-audio path cache.

Pros:

- Fixes the common mutable local-file case.
- Preserves resolve-cache benefits for stable files.
- Keeps behavior deterministic for local serving workflows.

Cons:

- Adds file stat/hash overhead before cache lookup.
- Needs care around TOCTOU changes while decoding.

### Option 2: Bypass `_resolve_ref_audio` Cache For Mutable Locators

Keep caching for `data:` URLs, but bypass or greatly restrict caching for local paths and HTTP URLs.

Pros:

- Simple and conservative.
- Avoids stale decoded audio for mutable locators.

Cons:

- Loses useful decode/download caching for repeated local-file requests.
- May regress latency for workloads that repeatedly use stable paths.

### Option 3: Keep Current Behavior But Document Semantics

Document that `_resolve_ref_audio()` is keyed by the original locator string and assumes locator content is immutable during server lifetime.

Pros:

- No behavioral risk.
- Clarifies expectations for users and future cache layers.

Cons:

- Does not fix stale-hit behavior.
- Easy for users to hit in local development and dynamic file workflows.

## Proposed Direction

Prefer Option 1 for local files, with a clear follow-up decision for HTTP URLs.

For local files, use a content-aware or revalidated key before returning cached decoded waveform data. For remote URLs, either bypass caching by default or preserve current best-effort string-keyed behavior behind explicit documentation.

## Reproduction Sketch

1. Start a vLLM-Omni speech server with a voice-cloning model.
2. Send a request with:

```json
{
  "input": "hello",
  "ref_audio": "file:///tmp/ref.wav"
}
```

3. Replace `/tmp/ref.wav` with a different speaker while keeping the same path.
4. Send the same request again.

Actual result:

- The server may reuse the old decoded waveform from `_ref_audio_resolve_cache`.

Expected result:

- The second request should use the updated file content, or the cache semantics should explicitly require immutable referenced files.

## Notes

This issue is separate from MOSS-TTS-Local reference-code caching, single-flight, and batched reference encoding. Those optimizations should remain content-addressed downstream of audio resolution.

pranavthakur0-0:
Fixes #4873

## Purpose

When a user updates a local reference audio file but keeps the same file path, `_resolve_ref_audio()` was previously returning a stale decoded audio waveform because the cache key was only hashing the `ref_audio_str` locator path.

This PR implements Option 1 from the issue discussion by updating `_get_ref_audio_cache_key` to be content-aware for local files. The cache key now incorporates the file's `mtime` and `size`, automatically detecting file modifications without needing a server restart, while preserving the existing string-hash behavior for URLs and immutable data.

## Test Plan

- Created unit tests in `tests/test_ref_audio_cache.py` verifying cache invalidation behavior using temporary files.
- Ran an end-to-end integration test locally with `OpenMOSS-Team/MOSS-TTS-Nano` in `voice_clone` mode by downloading a sample reference audio, querying the `/v1/audio/speech` endpoint, overwriting the audio file on disk, and querying the endpoint again.

**vLLM Version:**
0.24.0

**vLLM-Omni Commit:**
32032b87b7c103f2c7d3ab0454ed9de5c7de5f84

## Test Result

- **Unit test**: Passed (asserted cache keys correctly mismatch upon file edit).
- **End-to-End**: Passed (Server dynamically invalidated the `_ref_audio_resolve_cache`, reloaded the decoded waveform, and generated the second audio request with the new cloned speaker voice).


vllm_omni/entrypoints/openai/serving_speech.py
```diff
@@ -1995,14 +1995,31 @@ async def _resolve_higgs_audio_v3_adapter(self):
...
+    def _get_ref_audio_cache_key(ref_audio_str: str) -> str:
+        """Compute the cache key hash for a given ref_audio string, incorporating file metadata for local files."""
+        cache_key_source = ref_audio_str
+        if not ref_audio_str.startswith(("http://", "https://", "data:")):
+            try:
+                path = ref_audio_str[7:] if ref_audio_str.startswith("file://") else ref_audio_str
+                st = os.stat(path)
+                cache_key_source = f"{ref_audio_str}:{st.st_mtime}:{st.st_size}"
+            except Exception:
+                pass
```
tzhouam: silent failure should be fixed

vllm_omni/entrypoints/openai/serving_speech.py
```diff
@@ -1995,14 +1995,31 @@ async def _resolve_higgs_audio_v3_adapter(self):
         self._higgs_audio_v3_adapter = adapter
         return adapter
 
+    @staticmethod
+    def _get_ref_audio_cache_key(ref_audio_str: str) -> str:
+        """Compute the cache key hash for a given ref_audio string, incorporating file metadata for local files."""
+        cache_key_source = ref_audio_str
+        if not ref_audio_str.startswith(("http://", "https://", "data:")):
+            try:
+                path = ref_audio_str[7:] if ref_audio_str.startswith("file://") else ref_audio_str
```
tzhouam: Derive the path the same way MediaConnector._load_file_url does (urlparse + url2pathname) instead of ref_audio_str[7:], otherwise percent-encoded paths (e.g. %20) stat a nonexistent/wrong file and silently fall back to the stale-cache behavior this PR is fixing.

tzhouam (changes requested):

pranavthakur0-0:
Hi @tzhouam,
Thanks for the feedback! I've updated the PR to address both comments:

- Added `urlparse` + `url2pathname` for percent-encoded file URIs.
- Replaced silent error handling with proper warning logging.

Could you please review again? Thanks!

tzhouam:
please fix the precommit, thanks

tests/test_ref_audio_cache.py
```diff
@@ -0,0 +1,98 @@
...
+    """A missing file must fall back to the URI-string hash AND emit a warning.
+
+    The reviewer's core complaint was 'silent failure should be fixed' — so we
+    must assert that the warning is actually logged, not just that the fallback
+    key value is correct.
+    """
+    import logging
+
+    file_uri = "file:///path/to/nonexistent/file.wav"
+    with caplog.at_level(logging.WARNING):
```
linyueqian: [High] I do not think this test can pass in this repo. `vllm_omni/logger.py` reparents the `vllm_omni` logger under `vllm` (`vllm_omni_root.parent = vllm_root`, and that module is imported via `vllm_omni/patch.py`), and vLLM sets `propagate = False` on the `vllm` logger. Records emitted by `vllm_omni.entrypoints.openai.serving_speech` therefore stop at `vllm` and never reach the Python root logger, which is where a bare `caplog.at_level(...)` installs its handler, so `caplog.records` should come back empty and the `assert any("stale cache" in ...)` fails.

This is already known here: see the note in `tests/diffusion/offloader/test_module_collector.py` (lines 16-18) and the working pattern in `tests/model_executor/stage_input_processors/test_mimo_audio_llm2code2wav.py` (lines 37-47), which attaches `caplog.handler` to the module logger directly and restores it in a `finally`. Please switch to that, and use `rec.getMessage()` rather than `rec.message` while you are there.

The PR says the unit tests passed, so if this does pass for you please paste the pytest output, I would like to understand why.

vllm_omni/entrypoints/openai/serving_speech.py
```diff
@@ -1995,14 +1997,41 @@ async def _resolve_higgs_audio_v3_adapter(self):
...
+            try:
+                if ref_audio_str.startswith("file://"):
+                    parsed = urlparse(ref_audio_str)
+                    path = url2pathname(
+                        (parsed.netloc or "") + (parsed.path or "")
+                    )
+                else:
+                    path = ref_audio_str
+                st = os.stat(path)
+                cache_key_source = f"{ref_audio_str}:{st.st_mtime}:{st.st_size}"
```
linyueqian: [Med] Use `st.st_mtime_ns` instead of `st.st_mtime`. `st_mtime` is a float, and at epoch ~1.75e9 a float64 only resolves to roughly 238 ns, so it already discards the nanosecond precision that ext4/xfs/apfs record. More practically, any overwrite that lands inside one mtime tick and keeps the same byte size is invisible to this key, which is exactly the "user re-exports the same clip" case the issue is about. `st_mtime_ns` is a one-word change that closes most of that window.

tests/test_ref_audio_cache.py
```diff
@@ -0,0 +1,98 @@
...
+        file_uri = f"file://{test_file}"
+
+        key1 = OmniOpenAIServingSpeech._get_ref_audio_cache_key(file_uri)
+        print(f"  [1] Key for original file: {key1}")
+
+        # Wait slightly to ensure the mtime changes on the filesystem
+        time.sleep(0.01)
+
+        # 2. Modify the file
+        test_file.write_text("dummy audio data 2 modified")
```
linyueqian: [Med] This rewrite changes the byte length as well as the mtime (`"dummy audio data 1"` is 18 bytes, `"dummy audio data 2 modified"` is 27), so the test passes even if only `st_size` were hashed and it never actually exercises the mtime half of the key. Please keep the replacement the same length, for example `"dummy audio data 2"`, so the assertion depends on mtime alone. Pairs with the `st_mtime_ns` comment on the implementation.

vllm_omni/entrypoints/openai/serving_speech.py
```diff
@@ -2067,7 +2096,7 @@ def _make_ref_audio_artifact_cache_key(wav: np.ndarray, sr: int) -> str:
         return h.hexdigest()
 
     def _get_resolved_ref_audio_artifact_key(self, ref_audio_str: str) -> str | None:
-        source_key = hashlib.sha1(ref_audio_str.encode("utf-8")).hexdigest()
+        source_key = self._get_ref_audio_cache_key(ref_audio_str)
```
linyueqian: [Med] Both call sites now stat the file twice and the two results can disagree. `_resolve_ref_audio(x)` is immediately followed by `_get_resolved_ref_audio_artifact_key(x)` at lines 2216-2217 (higgs_audio_v3) and 3636-3637 (qwen3_tts). Before this PR both hashed the same pure string so they were guaranteed to match; now if the file is rewritten between the two calls the second stat produces a different key, the lookup misses, and this returns `None`.

The failure is graceful rather than wrong (Qwen3-TTS silently loses the artifact-only path, higgs v3 loses the ref-code cache) but it is a silent perf cliff with no log, and it also doubles the syscalls on the hot path for every speech request. Cleaner fix: have `_resolve_ref_audio` return the key it actually used, or stash it per request, so both sides share one stat.

vllm_omni/entrypoints/openai/serving_speech.py
```diff
@@ -1995,14 +1997,41 @@ async def _resolve_higgs_audio_v3_adapter(self):
...
+                    path = url2pathname(
+                        (parsed.netloc or "") + (parsed.path or "")
+                    )
+                else:
+                    path = ref_audio_str
+                st = os.stat(path)
+                cache_key_source = f"{ref_audio_str}:{st.st_mtime}:{st.st_size}"
+            except OSError:
+                logger.warning(
+                    "Failed to stat ref_audio path %r; falling back to "
```
linyueqian: [Med] This logs the full `ref_audio` value with `%r`, at warning level, on every request that fails to stat. `ref_audio` is client-controlled and `_validate_ref_audio_format` (line 1615) is only wired into some model validators, so a client that sends a bare base64 blob without the `data:` prefix would dump the entire audio payload into the server log on every single call, and a client with a typo'd path would repeat the warning indefinitely. Please truncate the value (a prefix plus length is enough to debug) and consider debug level or a log-once guard.

vllm_omni/entrypoints/openai/serving_speech.py
```diff
@@ -1995,14 +1997,41 @@ async def _resolve_higgs_audio_v3_adapter(self):
...
+        if not ref_audio_str.startswith(("http://", "https://", "data:")):
+            try:
+                if ref_audio_str.startswith("file://"):
+                    parsed = urlparse(ref_audio_str)
+                    path = url2pathname(
+                        (parsed.netloc or "") + (parsed.path or "")
+                    )
+                else:
+                    path = ref_audio_str
+                st = os.stat(path)
```
linyueqian, line 1803: [Low] This stats a client-supplied path before the `--allowed-local-media-path` gate runs. The allowlist is enforced inside `MediaConnector` in `_resolve_ref_audio`, which only happens after the cache key is computed, so the server now touches arbitrary paths outside the allowlist. `stat` returns no content and the result is not observable to the caller, so the exposure is small, but on an automounted or NFS path a stat can block and this is an async handler. Consider restricting the stat to paths under `allowed_local_media_path`, or computing the key after the connector has validated the path.

vllm_omni/entrypoints/openai/serving_speech.py
```diff
@@ -1995,14 +1997,41 @@ async def _resolve_higgs_audio_v3_adapter(self):
...
+    @staticmethod
+    def _get_ref_audio_cache_key(ref_audio_str: str) -> str:
+        """Compute the cache key hash for a given ref_audio string, incorporating file metadata for local files."""
+        cache_key_source = ref_audio_str
+        if not ref_audio_str.startswith(("http://", "https://", "data:")):
+            try:
+                if ref_audio_str.startswith("file://"):
+                    parsed = urlparse(ref_audio_str)
+                    path = url2pathname(
+                        (parsed.netloc or "") + (parsed.path or "")
```
linyueqian: [Low] Concatenating netloc and path produces a relative path for the authority form of the scheme: `file://localhost/tmp/speaker.wav` becomes `localhost/tmp/speaker.wav`, which stats relative to the server's cwd and falls back to the string-only key. Either drop the netloc when it is empty or `localhost` (the only values RFC 8089 allows for a local file), or reject a non-empty netloc explicitly.

tests/test_ref_audio_cache.py
```diff
@@ -0,0 +1,98 @@
...
+    key2 = OmniOpenAIServingSpeech._get_ref_audio_cache_key(url)
+
+    assert key1 == key2, "Cache key for URLs should remain constant!"
+    print("  ✓ Passed: URLs correctly maintain a constant cache key.")
+
+
+def test_missing_file_cache_key(caplog):
+    """A missing file must fall back to the URI-string hash AND emit a warning.
+
+    The reviewer's core complaint was 'silent failure should be fixed' — so we
```
linyueqian: [Nit] This docstring quotes a code-review exchange, which should not ship in the committed test. State the behavior being asserted instead, for example "a missing file falls back to the URI-string hash and emits a warning so the stale-cache risk is visible in the logs".

tests/test_ref_audio_cache.py
```diff
@@ -0,0 +1,98 @@
...
+        # so it should differ from a plain SHA-1 of the URI string.
+        string_only_key = hashlib.sha1(encoded_uri.encode("utf-8")).hexdigest()
+        assert key != string_only_key, (
+            "Percent-encoded URI should be decoded and stat'd, "
+            "not fall back to string-only cache key!"
+        )
+        print("  ✓ Passed: Percent-encoded file URIs are decoded correctly.")
+
+
+if __name__ == "__main__":
```
linyueqian: [Nit] The `__main__` block can go, along with the `print()` calls in the tests above. Pytest already captures and reports per-test status, and the block is inconsistent by construction since it has to skip `test_missing_file_cache_key` for needing a fixture.

linyueqian (commented):
Thanks for picking this up. The approach is right and matches Option 1 in #4873, and folding the key computation into a single helper so `_get_resolved_ref_audio_artifact_key` stays consistent with `_resolve_ref_audio` was the correct instinct. Documenting the remote-URL semantics in the docstring also covers the "document as best-effort" half of the issue, and it would be worth mirroring one line of that into `docs/serving/speech_api.md` so users know URL-backed references are not revalidated.

Two things I would want resolved before this lands: the new logging test looks like it cannot pass in this repo (details inline), and the invalidation signal is weaker than it needs to be. The rest are smaller.

One more, too diffuse to attach to a line: when the key changes, the previous entry stays in `_ref_audio_resolve_cache` until LRU eviction, so `_discard_ref_audio_artifact_ready_if_unreferenced` never fires for the file that was replaced and its artifact key stays in `_ref_audio_model_artifact_ready`. Not incorrect, but the cache now holds both versions of every edited file and the stale artifact stays marked ready. Worth a comment at minimum.

I reviewed by reading, I did not run the suite (no vllm in my checkout), so please paste the actual pytest output for the new tests.

pranavthakur0-0:
Hi @tzhouam and @linyueqian,

Thanks for the feedback! I've updated the PR to address all comments

linyueqian (commented):
Round 2. Thanks for the thorough pass, all nine points from the previous review are addressed: the caplog test now uses the module-logger handler pattern with `getMessage()` and a DEBUG level that matches the implementation, `st_mtime_ns` is in, the invalidation test writes a same-length replacement so it genuinely exercises mtime rather than size, the warning is truncated and demoted, `allowed_local_media_path` is threaded through, the non-local `file://` authority is rejected, and the test file picked up an SPDX header and lost the prints, the `__main__` block and the quoted review exchange. The docstring note about the previous entry surviving until LRU eviction is a good addition.

The one place I would still push back is the fix for the double stat. Swapping it for `_last_ref_audio_source_key` removes the redundant syscall but introduces a worse failure mode, details inline. The shape I had in mind was narrower: have `_resolve_ref_audio` return the key it used (or expose an internal variant that returns `(wav_list, sr, cache_key)`), and let the two call sites that need the artifact key pass it straight into `_get_resolved_ref_audio_artifact_key`. No shared mutable state, no lifetime question, and the key is computed exactly once by construction. Keying the stash on `request_id` instead of the locator string would also work.

vllm_omni/entrypoints/openai/serving_speech.py
```diff
@@ -1995,14 +2002,87 @@ async def _resolve_higgs_audio_v3_adapter(self):
...
-        cache_key = hashlib.sha1(ref_audio_str.encode("utf-8")).hexdigest()
+        # Pass the allowed-local-media-path so the stat is restricted to
+        # paths the server operator has explicitly permitted.
+        allowed_path = None
+        if not self._diffusion_mode:
+            allowed_path = getattr(self.model_config, "allowed_local_media_path", None)
+        cache_key = self._get_ref_audio_cache_key(ref_audio_str, allowed_local_media_path=allowed_path)
+        # Stash so _get_resolved_ref_audio_artifact_key can reuse without
+        # re-stat-ing.
+        self._last_ref_audio_source_key[ref_audio_str] = cache_key
```
linyueqian: [High] This dict is written on every `_resolve_ref_audio` call but only drained by `_get_resolved_ref_audio_artifact_key`, and only the higgs_audio_v3 path (line 2216) and the qwen3_tts path (line 3276) call that getter. Every other caller never pops: `_resolve_ref_audio_many`, the MOSS/CosyVoice/Voxtral branches in this file, and the OmniVoice, IndexTTS2, Fish Speech and Ming adapters under `entrypoints/openai/tts_adapters/`. Those entries are never removed and the cache has no bound.

The key is the raw `ref_audio_str`, which for an inline request is the entire `data:audio/wav;base64,...` payload, so a server doing inline voice cloning accumulates a few MB of dead string per distinct reference clip for the lifetime of the process, driven entirely by client input. That is a worse outcome than the duplicate `os.stat` it replaces.

Suggest dropping the shared dict: return the cache key from `_resolve_ref_audio` (or add an internal variant returning `(wav_list, sr, cache_key)`) and pass it into `_get_resolved_ref_audio_artifact_key` at the two call sites. If you would rather keep a side table, key it on `request_id` and clear it in the same place the other per-request state is cleaned up, so its lifetime is bounded by the request.

vllm_omni/entrypoints/openai/serving_speech.py
```diff
@@ -2067,7 +2147,11 @@ def _make_ref_audio_artifact_cache_key(wav: np.ndarray, sr: int) -> str:
         return h.hexdigest()
 
     def _get_resolved_ref_audio_artifact_key(self, ref_audio_str: str) -> str | None:
-        source_key = hashlib.sha1(ref_audio_str.encode("utf-8")).hexdigest()
+        # Reuse the cache key stashed by the preceding _resolve_ref_audio call
+        # to avoid a second os.stat (TOCTOU race + wasted syscall).
+        source_key = self._last_ref_audio_source_key.pop(ref_audio_str, None)
```
linyueqian: [Med] The stash is keyed on the locator, not the request, and there is an `await connector.fetch_audio_async(...)` between the write at line 2085 and this pop, so two in-flight requests for the same `ref_audio` interleave. Request B overwrites A's entry, then A pops B's key. If the file was rewritten between the two, A resolved the old waveform but receives the new file's artifact key, and `_resolve_higgs_audio_v3_ref_codes` will then serve reference codes that do not correspond to the audio A actually decoded.

It needs a concurrent request plus a mid-flight file edit, so it is narrow, but that is precisely the scenario this PR exists to handle, and it is a silent wrong-output rather than a stale-cache miss. Passing the key explicitly, or keying the stash on `request_id`, removes the window entirely.

vllm_omni/entrypoints/openai/serving_speech.py
```diff
@@ -2067,7 +2147,11 @@ def _make_ref_audio_artifact_cache_key(wav: np.ndarray, sr: int) -> str:
...
+        if source_key is None:
+            source_key = self._get_ref_audio_cache_key(ref_audio_str)
```
linyueqian: [Low] The fallback omits `allowed_local_media_path`, so when the stash misses, the stat runs unrestricted and re-opens the case the allowlist parameter was added to close. It is only reachable through the interleaving described above, so it is low impact today, but if the stash stays it should forward the same `allowed_path` that `_resolve_ref_audio` computes at line 2079.

vllm_omni/entrypoints/openai/serving_speech.py
```diff
@@ -1995,14 +2002,87 @@ async def _resolve_higgs_audio_v3_adapter(self):
...
+                        raise OSError(f"file:// URI with non-local authority {netloc!r} cannot be stat'd")
+                else:
+                    path = ref_audio_str
+                # Only stat paths under the configured allowlist to avoid
+                # touching arbitrary client-supplied paths (NFS automounts,
+                # etc.) before MediaConnector validates the request.
+                if allowed_local_media_path is not None:
+                    resolved = os.path.realpath(path)
+                    allowed_base = os.path.realpath(allowed_local_media_path)
+                    if os.path.commonpath([resolved, allowed_base]) != allowed_base:
```
linyueqian: [Low] `os.path.commonpath` raises `ValueError`, not `OSError`, when the inputs mix absolute and relative paths or sit on different drives, so on Windows a `ref_audio` on a different drive letter from `allowed_local_media_path` escapes the `except OSError` and propagates out of the key helper into the request. Either widen the handler to `(OSError, ValueError)` or use `Path(resolved).is_relative_to(allowed_base)`, which returns `False` instead of raising.

pranavthakur0-0:
Hi @linyueqian,
Thanks for the review! I've pushed the fixes for the remaining issues. 
Could you please review?

pranavthakur0-0:
hey @linyueqian, @tzhouam 
can you please review when you get the time?

vllm_omni/model_executor/models/moss_tts/reference_encoder.py
```diff
@@ -76,11 +75,15 @@ async def encode_reference_codes(
     Returns:
         The reference RVQ codes tensor (CPU), ready to pass to the processor.
     """
+    # Resolve the audio first. This uses the serving layer's resolve cache,
+    # which correctly handles mtime/size invalidation for local files.
+    wav_list, sr, resolve_cache_key = await resolve_ref_audio(ref_str)
```
linyueqian: [Med] Hoisting the resolve above the cache check costs work in the `voice_name` branch that it cannot repay. When `voice_name` is set, `speaker_name` is the voice and `created_at` is `voice_created_at`; `resolve_cache_key` is never read. Re-upload already bumps `created_at` and `_evict_existing_upload` calls `_speaker_cache.clear(voice_name_lower)`, so that branch never needed a content-addressed key. Every named-voice request now resolves before it is allowed to hit the cache that exists to skip exactly that.

Whether the resolve is cheap depends on `_ref_audio_resolve_cache`, and the two caches are sized very differently. `_put_resolved_ref_audio` books `size = len(wav_list) * 40` against a 256 MB budget, so one 10 s clip at 24 kHz takes 9.6 MB and the cache holds **27** of them; at the 30 s `_REF_AUDIO_MAX_DURATION` it holds **9**. `SpeakerEmbeddingCache` has 512 MiB for small code tensors and stays warm far longer. Past a couple of dozen voices in rotation, speaker-hit-with-resolve-miss becomes the steady state, and each of those now re-runs the base64 + WAV decode of the uploaded clip — `_apply_uploaded_speaker` sets `ref_audio` to the data URL from `_get_uploaded_audio_data`, so it is a decode rather than a network fetch. The encode is still skipped, so this is not the dominant cost; it is avoidable work put back into the path that was optimised to remove it.

To be explicit about what I am *not* asking for: in the anonymous branch, resolving first is correct and I would like it left alone. It is what makes the key and the waveform come from one stat, which is the property I asked for last round. Please keep that and give the named branch its own early return:

```python
if voice_name:
    speaker_name, created_at = voice_name, int(voice_created_at)
    cache_key = speaker_cache.make_cache_key(speaker_name, model_type=..., created_at=created_at)
    cached = speaker_cache.get(cache_key)
    if cached is not None:
        return cached["codes"].clone()
    wav_list, sr, _ = await resolve_ref_audio(ref_str)
else:
    wav_list, sr, resolve_cache_key = await resolve_ref_audio(ref_str)
    speaker_name, created_at = "ref:" + resolve_cache_key, 0
    cache_key = speaker_cache.make_cache_key(...)
    cached = speaker_cache.get(cache_key)
    if cached is not None:
        return cached["codes"].clone()
```

linyueqian (commented):
Round 3. Every item from the previous two rounds is addressed, and I checked them against the code rather than the commit messages: the shared `_last_ref_audio_source_key` dict is gone in favour of `_resolve_ref_audio` returning `(wav_list, sr, cache_key)`, which is the shape I asked for — the key and the audio now come from a single stat by construction, so the interleaving window and the unbounded side table both disappear. `commonpath`'s `ValueError` is caught, the `file://` authority is rejected, the allowlist is threaded through, and the log line is truncated and demoted.

I also ran the four new tests. I still have no vllm in my checkout, so I extracted `_get_ref_audio_cache_key` verbatim into a standalone module and transcribed the tests against it — all four pass, including the percent-encoded case and the same-byte-length rewrite that now genuinely exercises `st_mtime_ns` rather than `st_size`. That is the helper's logic only; it does not exercise the real logger hierarchy, so if the lane has run the new test file I would still like to see that output for `test_missing_file_cache_key`.

What is new since my last pass is `ef25742b4`, and that is where my remaining comments are. The change itself is right and I want to be clear about that: without it the resolve cache would hand back a fresh waveform while MOSS's speaker cache kept serving codes keyed on `sha1(ref_str)`, so a delay-family voice clone would still reproduce the old speaker. Fixing only the first layer would have left the user-visible bug in place for that family. My concerns are with how it is ordered and that nothing tests it.

Two smaller things that are not worth their own inline comment:

`ref_str=None` now raises `AttributeError` inside `_get_ref_audio_cache_key`, where the old code had an explicit `(ref_str or "")` guard. `_build_moss_tts_params` passes `request.ref_audio` straight through at the `tts`/`local`/`ttsd` branches, so the value can be `None`. It is a narrow regression rather than a new crash — the old code also died on the resolve when the speaker cache missed, and `_apply_uploaded_speaker` fills `ref_audio` in for uploaded voices — but a named voice with a warm speaker cache and no `ref_audio` used to return codes and now throws. A one-line guard restores the old behaviour.

On the log level: @tzhouam asked for the silent failure to be fixed, I asked for truncation and suggested debug, and the result is `logger.debug`. That may be quieter than the original request intended. @tzhouam, your `CHANGES_REQUESTED` from 4 Aug is still outstanding — worth confirming this lands where you wanted it.

The required `buildkite/vllm-omni` lane has never run on this head; the five green checks are all GitHub Actions. Whatever we conclude here, that lane is what should gate the merge.

vllm_omni/model_executor/models/moss_tts/reference_encoder.py
```diff
@@ -76,11 +75,15 @@ async def encode_reference_codes(
...
+
     if voice_name:
         speaker_name = voice_name
         created_at = int(voice_created_at)
     else:
-        speaker_name = "ref:" + hashlib.sha1((ref_str or "").encode("utf-8")).hexdigest()
+        speaker_name = "ref:" + resolve_cache_key
```
linyueqian, line 109: [Med] This line is the whole point of `ef25742b4` and nothing tests it. `git grep encode_reference_codes tests/` returns nothing, so the function has no unit coverage at all, and the new `tests/test_ref_audio_cache.py` only exercises the static key helper in isolation.

The e2e in the PR description does not reach here either. It used `MOSS-TTS-Nano`, and nano is `v is None`, which returns at `serving_speech.py:1573` with `prompt_audio_array` before `encode_reference_codes` is ever called. The description also still pins `32032b87b`, which predates this commit — so the test plan as written covers the resolve-cache half of the fix and not this half.

That matters because this is the half a user would actually notice: with the resolve cache fixed but the speaker cache still keyed on `sha1(ref_str)`, a delay-family clone keeps reproducing the old speaker even though the fresh waveform was decoded. A test that calls `encode_reference_codes` twice with a stub `resolve_ref_audio` returning two different keys, asserting the second call re-encodes rather than serving the cached tensor, would pin it without needing a model. Please also refresh the Test Plan for the current head.

vllm_omni/entrypoints/openai/serving_speech.py
```diff
@@ -1894,14 +1896,88 @@ async def _resolve_higgs_audio_v3_adapter(self):
...
+                    if netloc.lower() in ("", "localhost"):
+                        path = url2pathname(parsed.path or "")
+                    else:
+                        raise OSError(f"file:// URI with non-local authority {netloc!r} cannot be stat'd")
+                else:
+                    path = ref_audio_str
+                # Only stat paths under the configured allowlist to avoid
+                # touching arbitrary client-supplied paths (NFS automounts,
+                # etc.) before MediaConnector validates the request.
+                if allowed_local_media_path is not None:
```
linyueqian: [Low] The guard is inert in the configuration that needs it. It only engages when `allowed_local_media_path` is not `None`, but `None` is the default, and `_resolve_ref_audio` also forces it to `None` in diffusion mode. In both cases `MediaConnector` refuses to load local files at all, so the stat can never produce a usable key — it is pure exposure for zero benefit, which is the case I raised originally.

I checked this rather than inferring it. Extracting the helper verbatim and calling `_get_ref_audio_cache_key(<path outside any allowlist>, allowed_local_media_path=None)` returns a key that differs from `sha1(path)`, i.e. the stat ran. With the allowlist set, the same call correctly falls back to the string-only key.

Inverting the condition covers both and is shorter than the current form:

```python
if allowed_local_media_path is None:
    raise OSError("local media not permitted; skipping stat")
resolved = os.path.realpath(path)
...
```

tests/test_ref_audio_cache.py
```diff
@@ -0,0 +1,107 @@
...
+# Stub out the server-entry modules that are pulled in by
+# vllm_omni/entrypoints/openai/__init__.py before the real import.
+# Those modules transitively import vllm.entrypoints.chat_utils symbols
+# (e.g. get_history_tool_calls_cnt) that may not exist in all installed
+# vllm versions.  The static method under test only uses stdlib, so it
+# loads cleanly once the chain is broken.
+import sys
+from unittest.mock import MagicMock
+
+sys.modules.setdefault("vllm_omni.entrypoints.openai.serving_chat", MagicMock())
```
linyueqian: [Low] These two `setdefault` calls are global side effects at import time, and one of them replaces `vllm_omni.entrypoints.openai.api_server` with a `MagicMock` for the rest of the process. `tests/entrypoints/openai_api/test_serving_speech.py:27` imports that module for real. Today the ordering saves you — `entrypoints` sorts before `test_ref_audio_cache.py` inside `tests/`, so the real module is already in `sys.modules` and `setdefault` no-ops — but that is collection order doing the work, not isolation, and it will not hold under sharding or a subset run.

The stub is only needed because the file sits on its own. `tests/` has no other top-level test file, and `test_serving_speech.py` imports `OmniOpenAIServingSpeech` with no stubbing whatsoever. Moving these four tests into that file drops the `setdefault` pair, the `# ruff: noqa: E402`, and the comment explaining them, and puts them next to `test_resolve_ref_audio_reuses_decoded_audio_for_same_source`, which this PR already touches.

linyueqian (commented):
Round 4 on `9ef62e901`. The named-voice early cache check, the allowlist guard rework and the test relocation all landed. Moving the four cases into `tests/entrypoints/openai_api/test_serving_speech.py` takes the process-global `sys.modules` stub and the `E402` waiver with them, and puts the coverage in the step that owns this code (`.buildkite/cuda/test-ready.yml:23`) instead of the misc step. The stash is gone for good and all 12 `_resolve_ref_audio` call sites plus `_resolve_ref_audio_many` and the MOSS callable unpack the third element.

Three things left. The first is the one I would want resolved before this lands: the decode cache is content-aware now, but the prefix-cache salt is not, so #4873 can still reproduce end to end. Details inline.

vllm_omni/entrypoints/openai/serving_speech.py
```diff
@@ -3088,8 +3174,8 @@ async def _build_qwen3_tts_request(
             # Uploaded voice: ref_audio was auto-set as [base64_data_url]
             ref_audio_source = tts_params["ref_audio"][0]
         if ref_audio_source is not None and isinstance(ref_audio_source, str):
-            wav_list, sr = await self._resolve_ref_audio(ref_audio_source)
-            artifact_key = self._get_resolved_ref_audio_artifact_key(ref_audio_source)
+            wav_list, sr, cache_key = await self._resolve_ref_audio(ref_audio_source)
```
linyueqian, line 2931: [High] `_resolve_ref_audio` is content-aware now, but the prefix-cache salt is not, so the symptom in #4873 survives end to end.

`_conditioning_cache_salt` (line 262) hashes `request.ref_audio` as a raw string. Its own docstring states the invariant it exists to hold: "identical conditioning may share the prefix, any difference never does." A local file edit breaks that, and the resolve-cache fix does not reach it.

Qwen3-TTS is the clean repro. Line 3191 builds the prompt as `tokens_input(prompt_token_ids=[1] * ph_len)`, and `ph_len` comes from the text plus `ref_code_length`, which is derived from the decoded waveform. Send the same `input` twice against `file:///data/spk.wav`, overwriting the file between the two calls with a same-duration re-export (the case the issue describes). The second request decodes the new audio correctly, but `ph_len` is unchanged so the token ids are the identical `[1] * ph_len`, and the salt is unchanged because the locator string did not move. vLLM returns a full prefix hit and the request runs on KV built from the previous reference. Prefix caching is on by default; only a few deploy YAMLs turn it off.

Same path for MOSS (`tts_adapters/moss_tts.py:104`) and Fish Speech (`tts_adapters/fish_speech.py:54`). IndexTTS2 has its own variant at `tts_adapters/indextts2.py:84`.

This is pre-existing rather than something the PR introduced, but it sits inside "Fixes #4873" and you now compute exactly the value that closes it. Smallest fix: stash `cache_key` into `tts_params` here, and fold it in alongside `voice_created_at` in the second loop at lines 272-278. That reuses the existing "resolved conditioning the raw request cannot see" mechanism instead of adding another one.

vllm_omni/model_executor/models/moss_tts/reference_encoder.py
```diff
@@ -76,24 +81,51 @@ async def encode_reference_codes(
...
+    model_type = f"moss_tts_{variant}_nq{n_vq}"
+
+    # ---- Named-voice branch: check cache *before* resolving ----
+    # When voice_name is set, the cache key is (voice_name, created_at) which
+    # does not depend on the resolved audio content.  Re-upload already bumps
+    # created_at and the upload path clears the speaker cache slot, so the key
+    # is always fresh.  Skipping the resolve on cache hits avoids the base64
+    # decode / WAV decode cost that would otherwise dominate once the resolve
+    # cache (27 slots at 10 s) is full.
     if voice_name:
```
linyueqian, line 90: [Med] This branch still bypasses the content-aware key whenever `voice` arrives alongside an inline `ref_audio`, and the early return now makes that path cheaper rather than fixing it.

The comment above says the key is always fresh because "re-upload already bumps created_at and the upload path clears the speaker cache slot". That holds for real uploaded speakers, but `voice_name` is not restricted to those. `serving_speech.py:1631-1633` reads `_voice` straight off `request.voice` with no membership or inline check, `_apply_uploaded_speaker` returns early at `serving_speech.py:1109` as soon as `ref_audio is not None`, and `moss_tts` is absent from the built-in rejection list at `serving_speech.py:1114-1122`. So `voice="bob"` plus `ref_audio="file:///data/spk.wav"` validates cleanly, with `_voice_created_at("bob")` returning 0.

Edit the file and the request takes this branch, hits on `("bob", moss_tts_*, 0)` and returns the old RVQ codes. Before this commit the resolve at least ran first and was discarded; now it is skipped outright.

The adapter already has the guard you want: `tts_adapters/moss_tts.py:92` only sets `voice_name` when `voice_lower in server.uploaded_speakers and not has_inline_ref_audio`. Applying that same condition where `_voice` is computed at `serving_speech.py:1631` lines the two up and routes this case onto the anonymous branch you fixed.

vllm_omni/entrypoints/openai/serving_speech.py
```diff
@@ -1894,14 +1896,91 @@ async def _resolve_higgs_audio_v3_adapter(self):
...
+                    else:
+                        raise OSError(f"file:// URI with non-local authority {netloc!r} cannot be stat'd")
+                else:
+                    path = ref_audio_str
+                # Only stat paths the server operator has explicitly permitted
+                # via --allowed-local-media-path.  When no allowlist is
+                # configured (None), MediaConnector refuses local file loads
+                # anyway, so the stat would produce a usable key for a request
+                # that will fail — pure exposure for zero benefit.
+                if allowed_local_media_path is None:
```
linyueqian: [Low] This does not fire in the configuration it is written for. vLLM's default is `allowed_local_media_path: str = ""` (`vllm/config/model.py:190`), not `None`, so on a server started without the flag line 1982 yields `""`, `"" is None` is False, and execution falls through to `os.path.realpath("")`, which resolves to the process cwd.

The guard then means "must be under the server's working directory" and the stat still runs for anything under it, which is the exposure the comment says it closes.

`if not allowed_local_media_path:` covers both `""` and `None`.

vllm_omni/model_executor/models/moss_tts/reference_encoder.py
```diff
@@ -76,24 +81,51 @@ async def encode_reference_codes(
     Returns:
         The reference RVQ codes tensor (CPU), ready to pass to the processor.
     """
+    # Guard against None ref_str — the old code had ``(ref_str or "")`` and
+    # callers in the sound_effect / voice_generator branches may not supply it.
+    ref_str = ref_str or ""
```
linyueqian: [Nit] The premise in the comment does not hold. `_encode_ref` is only called from the `tts`/`local` and `ttsd` branches (`serving_speech.py:1650`, `1652`, `1654`); `sound_effect` and `voice_generator` never reach it, and the adapter's `validate` rejects a missing `ref_audio` on all three that do (`tts_adapters/moss_tts.py:43-65`). `ref_str` cannot be `None` here.

It is also not a safe default: `resolve_ref_audio("")` stats `""`, falls back to the string-only key, then fails inside `MediaConnector.fetch_audio_async("")`, so an empty string converts a clear failure into a confusing media-fetch error. Dropping the line and keeping `ref_str: str` is cleaner. If you would rather keep the widened type, the `resolve_ref_audio` annotation on line 45 still reads `Callable[[str], ...]`.

pranavthakur0-0:
Hey @linyueqian — thanks for Round 4.

You listed Qwen3, MOSS, Fish, and IndexTTS2 for the salt fix, so I left Higgs out of this PR. I did peek at it though: v3 looks like the same leftover (prefix caching on, placeholder tokens, no cache_salt), while v2 should be fine since prefix caching is off by default.

Would you rather I add v3 here, or follow up after this lands?

linyueqian:
Please add v3 here as well. Thanks!

pranavthakur0-0:
Hey @linyueqian 
added the v3 updates can you review ? Thanks!

tzhouam:
@vllm-omni-review-bot

tzhouam (approved):
LGTM

pranavthakur0-0:
> LGTM

thanks!!!

tzhouam:
@vllm-omni-review-bot check why npu ci is failed

TaffyOfficial:
@vllm-omni-review-bot check why npu ci is failed

tzhouam:
@pranavthakur0-0 please check the review of the bot and debug the npu ci.

pranavthakur0-0:
@tzhouam 
So I fixed the issues pointed out by @vllm-omni-review-bot 
But the  NPU CI issue is from this pr https://github.com/vllm-project/vllm-omni/pull/6274
and someone raised this issue https://github.com/vllm-project/vllm-omni/issues/6366

Should I fix it in this pr ?

Gaohan123:
> @tzhouam So I fixed the issues pointed out by @vllm-omni-review-bot But the NPU CI issue is from this pr #6274 and someone raised this issue #6366
> 
> Should I fix it in this pr ?

@pranavthakur0-0 Thanks. I suggest you can fix it in the PR directly.

pranavthakur0-0:
> > @tzhouam So I fixed the issues pointed out by @vllm-omni-review-bot But the NPU CI issue is from this pr #6274 and someone raised this issue #6366
> > Should I fix it in this pr ?
> 
> @pranavthakur0-0 Thanks. I suggest you can fix it in the PR directly.

It's already fixed 
I think this pr will pass all CI test

Issue #2347, opened 2026-03-31
[Roadmap][Feature] Support Moore Threads (MUSA) GPUs


yeahdongcn:
Enable first-class support for [Moore Threads](https://en.mthreads.com/) MUSA GPUs in vLLM-Omni, with efficient execution on MTGPU hardware while preserving upstream compatibility.

## Background

[Moore Threads](https://en.mthreads.com/) **MUSA** (Meta-computing Unified System Architecture) SDK provides a CUDA-compatible execution environment with vendor-specific runtime and software ecosystem. This design enables CUDA-based projects to port to MUSA with limited source-level changes while supporting vendor-specific optimizations.

## Reference Projects

| Project | Description |
|---------|-------------|
| [vllm-musa](https://github.com/MooreThreads/vllm-musa) | vLLM with MUSA support |
| [torchada](https://github.com/MooreThreads/torchada) | CUDA-to-MUSA PyTorch extension adapter |
| [torch_musa](https://github.com/MooreThreads/torch_musa) | PyTorch extension for MTGPU |
| [mate](https://github.com/MooreThreads/mate) | MUSA AI Tensor Engine (FlashAttention-compatible) |

### 1. Core Features

- [x] **MUSA Platform Integration** ([#2337](https://github.com/vllm-project/vllm-omni/pull/2337))
  - [x] Environment info collection
  - [x] OOT platform support
  - [x] Installation documentation
- [ ] Flash Attention 3 via MATE
- [ ] `torch.accelerator` support (blocked on Torch 2.9)

### 2. Model Support

- [ ] Qwen2.5-Omni
- [ ] Qwen3-Omni

### 3. Performance

- [ ] Baseline benchmarks
- [ ] Advanced optimizations

### 4. Testing & CI

- [ ] Provision CI runner (self-hosted) (8×S5000)
- [ ] MUSA CI pipeline integration

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://vllm-omni.readthedocs.io), which can answer lots of frequently asked questions.

gcanlin:
FYI: vLLM-Omni release plan: https://docs.google.com/document/d/1OY_11S0FdOzY5txBdLrtPp4NleEhr7S3OE1YnpOWu9w/edit?tab=t.0

yeahdongcn:
[...]

PR #5503, opened 2026-07-28
[Perf] Cache RoPE cos/sin tables in code predictor

l-wave:
# Cache RoPE cos/sin tables in the code predictor
## What & why
The code predictor's `_RotaryEmbedding`（`vllm_omni/model_executor/models/common/qwen3_code_predictor.py`）recomputed its rotary tables on every forward pass.
Each call executed the following chain:
`outer(positions, inv_freq) → matmul → cat → cos → sin → cast`

Namely, one matmul plus two trigonometric kernels per autoregressive step, accompanied by host-device synchronizations introduced by these kernels.
Since the code predictor repeatedly reprefills a short growing sequence (length `2 … num_code_groups+1`) at every AR step, massive redundant computation occurs continuously.

This PR aligns with the mainstream vLLM `RotaryEmbedding` implementation pattern:
1. Prebuild `cos_cached` / `sin_cached` **once inside `__init__`**
2. In forward, only index the precomputed tables and cast to input dtype

The NPU path already adopted cached tables. After this change, CUDA / CPU / XPU share the same lookup logic via a unified `_lookup` helper function.

## Numerical equivalence
The cached computation maintains bit-identical results compared with the original on-the-fly calculation:
`emb = cat(freqs, freqs)` is still constructed from the same `inv_freq` in float32, and cast to input dtype at the final stage, consistent with legacy logic.

Numerical verification: `allclose` check yields maximum difference = 0 across fp16 / bf16 / fp32.
All existing dtype test cases pass:
```
tests/model_executor/models/qwen3_tts/test_code_predictor_dtype.py ... 21 passed
```

## Benchmark
`bench/bench_rope_cache.py` isolates pure `_RotaryEmbedding.forward` logic to compare **online computation** vs **cached lookup**.
Timing uses CUDA Events + NVTX ranges, compatible with `nsys profile` tracing.

Benchmark configuration mirrors real Qwen3-TTS code predictor workload:
- `num_code_groups=32` (max_seq=33)
- `head_dim=128`
- `rope_theta=10000`
- Sweep: `seq_len ∈ {2,4,8,16,24,33}`, `bsz ∈ {1,8,32}`

Hardware: NVIDIA H20, torch 2.11, 1000 iterations, reported as median per-call GPU time.

| dtype     | live (ms) | cache (ms) | speedup |
|-----------|-----------|------------|---------|
| bfloat16  | 0.0957    | 0.0324     | 2.95x   |
| float16   | 0.0960    | 0.0319     | 3.01x   |
[...]

PR #5671, opened 2026-08-01
[Hardware][MUSA] Enable Qwen3-Omni ModelOpt FP8 inference

yeahdongcn:
## Purpose

Enable the supported Qwen3-Omni ModelOpt FP8 checkpoint on MUSA while keeping
shared vendor-neutral model paths unchanged.

Related issue: https://github.com/vllm-project/vllm-omni/issues/2347

Companion MUSA runtime support:
https://github.com/MooreThreads/vllm-musa/pull/161

### MUSA code-predictor path

- Add a dedicated `forward_musa()` implementation for the Qwen3 code
  predictor.
- Keep the shared `forward_native()` implementation unchanged for other
  vendors.
- Reuse the existing code-predictor helpers and preserve the output dtype.

This avoids the muDNN batched-matmul failure caused by the degenerate leading
dimension used by the shared native path on MUSA.

### Talker MTP graph capability

- Add `supports_talker_mtp_graph_capture()` to the Omni platform interface.
- Keep the default capability enabled and disable it only on MUSA.
- Let Qwen3-Omni declare `talker_mtp_graph_safe` from that capability.
- Preserve the legacy graph-wrapping behavior for models that do not declare
  the capability, while allowing an explicit `False` to skip the dedicated
  talker MTP FULL graph.

### ModelOpt FP8 stage configuration

The supported ModelOpt checkpoint quantizes the thinker, while talker and
code2wav remain BF16. Under the Qwen3-Omni MUSA deploy profile, stages 1 and 2
now set:

```yaml
hf_overrides:
  quantization_config: null
```

This uses vLLM's existing HF override flow before quantization auto-detection.
There is no generic `EngineArgs`/`OmniModelConfig` quantization rewrite in this
PR.

### CPU and configuration tests

```bash
python3 -m pytest -q tests/config/test_config_factory.py
# 284 passed

python3 -m pytest -q \
  tests/model_executor/models/qwen3_tts/test_code_predictor_dtype.py \
  tests/worker/test_omni_gpu_model_runner.py
# 51 passed
```

The focused pre-commit hooks also passed for every changed file.

Coverage includes:

- MUSA deploy overrides are present only for talker/code2wav;
- the non-MUSA deploy profile is unchanged;
- MUSA code-predictor dtype behavior;
- explicit unsafe, legacy-unspecified, and explicit-safe talker graph paths.

### MUSA end-to-end

Environment:

- 2x MTT S5000, isolated lease;
- MUSA driver/runtime 5.2;
- `torch==2.9.1.post1+musa5.2.0`;
- `vllm==0.1.dev1+gee0da84ab.d20260805`;
- image:
  `registry.mthreads.com/mcconline/inference/vllm-omni:minimax-h3-20260806`
  (`sha256:b9c5305ce399...`);
- normal compiled/cudagraph serving path with MCCL; no eager or compile-disabled
  fallback.

The current vLLM-MUSA production stack is based on vLLM 0.24. To prove that
the new deploy YAML is sufficient on that supported stack, the E2E run used
the full compatibility image and explicitly removed the previous generic
`_scope_prequantized_model_config` helper from its image-native `arg_utils.py`.
Only the final MUSA deploy YAML supplied the stage quantization overrides.

Observed engine configuration:

| Stage | Quantization | KV-cache dtype |
|---|---|---|
| thinker | `modelopt` | `fp8_e4m3` |
| talker | `None` | `auto` |
| code2wav | `None` | `auto` |

A `modalities=["text", "audio"]` request completed through all three stages:

| Check | Result |
|---|---|
| HTTP/server health | `200` |
| Text output | `Beijing` |
| Audio bytes | `33,494` |
| Sample rate | `24,000 Hz` |
| Duration | `0.696875 s` |
| Samples finite | `true` |
| Audio RMS | `0.0588463` |

The server log contained no NaN, `EngineCoreError`, request failure, or stage
failure.

## Commit hygiene

- One signed-off commit.
- No vLLM-MUSA source override is carried by this PR.


gcanlin:
cc @lishunyang12 @david6666666

vllm_omni/engine/arg_utils.py
```diff
@@ -33,6 +35,53 @@
 }
 
 
+def _stage_hf_config(
+    hf_config: Any,
```
hsliuustc0106: do not use any type

vllm_omni/engine/arg_utils.py
```diff
@@ -33,6 +35,53 @@
...
+    *,
+    model_stage: str | None,
+    hf_config_name: str | None,
+) -> Any:
```
hsliuustc0106: why we need the fucntions in this page for MUSA?

vllm_omni/worker/gpu_model_runner.py
```diff
@@ -206,8 +206,14 @@ def _init_talker_mtp(self) -> None:
         cudagraph_mode = self.compilation_config.cudagraph_mode
         assert cudagraph_mode is not None
         has_separate_talker = getattr(self.model, "talker", None) is not None
-        talker_mtp_graph_safe = getattr(self.model, "talker_mtp_graph_safe", False)
-        if cudagraph_mode.has_full_cudagraphs() and (has_separate_talker or talker_mtp_graph_safe):
+        talker_mtp_graph_safe = getattr(self.model, "talker_mtp_graph_safe", None)
```
hsliuustc0106, line 209: why it needs to change this

vllm_omni/engine/arg_utils.py
yeahdongcn: This has been removed from the PR.

vllm_omni/engine/arg_utils.py
yeahdongcn: Removed these helpers as well. The MUSA-specific stage overrides now live in the deploy YAML, where talker and code2wav clear the checkpoint-level quantization metadata.

vllm_omni/worker/gpu_model_runner.py
yeahdongcn: This allows Qwen3-Omni to skip the dedicated talker MTP graph on MUSA. The previous `has_separate_talker or ...` condition always enabled it. `None` keeps the existing behavior for other models, while an explicit `False` is now respected.

hsliuustc0106:
fix ci failures

yeahdongcn:
> fix ci failures

@hsliuustc0106 It should be fixed by the latest commit. Could you retrigger the CI?

hsliuustc0106:
@amy-why-3459 PTAL

amy-why-3459:
Thanks for moving quantization scoping into platforms.musa deploy overlays and for scoping talker_mtp_graph_safe to the talker stage only — that addresses the earlier arg_utils concern cleanly.

Main ask: CustomOp.forward_musa already delegates to forward_cuda, and _RotaryEmbedding.forward_cuda uses _lookup. The new forward_musa reimplements on-the-fly RoPE and overrides that default. Prefer deleting the override or implementing forward_musa as return self._lookup(...). If _lookup fails on MUSA, please document the failure mode — the current “native matmul ldb=0” rationale applies to forward_native, which MUSA should not hit via CustomOp dispatch.

Also please note in the PR why MUSA must disable talker MTP FULL graph capture (failure mode), and whether CUDA ModelOpt FP8 needs the same quantization_config: null for stages 1/2.

yeahdongcn:
@amy-why-3459 Thanks for the detailed review!

My initial MUSA validation was performed on a v0.24.1-based integration branch, before the cached RoPE implementation from #5503 was present. On that older path, RoPE used the expanded matmul and reproduced the muDNN `ldb=0` failure, which was why the original MUSA workaround was added.

You are right that on current main, the base `CustomOp.forward_musa` already delegates to `forward_cuda`, and `_RotaryEmbedding.forward_cuda` uses `_lookup`, so MUSA should not enter `forward_native`. I have removed the redundant `forward_musa` override and the related code-predictor test changes. I also revalidated the default path on MUSA: Qwen3-Omni ModelOpt FP8 completed code-predictor warmup, `torch.compile`, graph capture, and end-to-end inference successfully on 2× S5000.

For talker-MTP graph capture, we reproduced a concrete MUSA failure. With the dedicated talker-MTP FULL graph enabled, stage 1 reaches `torch.multinomial` during graph capture and fails with `RuntimeError: MUSA error: operation not permitted when stream is capturing`. This prevents stage 1 from initializing. The MUSA capability therefore disables only the dedicated talker-MTP FULL-graph wrapper; the normal compiled and captured paths for the rest of the stage remain enabled.

For quantization, this ModelOpt checkpoint quantizes only the thinker, while talker and code2wav remain BF16. On MUSA, the checkpoint-level quantization metadata is otherwise inherited by the audio stages during quantization auto-detection, which is why stages 1 and 2 set `quantization_config: null` in the MUSA deploy overlay. And I assume CUDA is working at the moment, as I couldn't find any existing issues indicating otherwise.

amy-why-3459 (approved):

PR #5674, opened 2026-08-01
[Bugfix] Reset duplex segment watermark

0xtoward:
## Purpose

Resumable duplex requests reuse the same external request ID across input segments. After a segment ends, token accounting restarts from zero while `requests_num_chunks_sent` still contains the previous segment's count. The adapter then treats the boundary and subsequent chunks as stale, which can stop audio forwarding in later turns.

## Root Cause

The dedup watermark was cleared by the background sender. A new segment could reach `save_async` before that sender processed the queued boundary.

## Fix

Reset the producer watermark when the segment boundary is queued. The send queue remains FIFO, so the boundary is still delivered before chunks from the next segment.

## Test Plan

```bash
pytest -q tests/distributed/omni_connectors/test_chunk_transfer_adapter.py
```

The regression test queues a boundary and the next segment before the background sender runs, then checks that both chunks are sent in order.

**vLLM Version:** `0.1.dev18616+ge5588e49b`

**vLLM-Omni Commit:** `75b93e0c67586e64e455d820e9620777d293599d` (`minicpm-challenge`)

## Test Result

- Connector tests: `50 passed, 14 warnings`
- Ascend 910C, MiniCPM-o 4.5 BF16, three stages, `async_chunk=true`:

| | Before | After |
|---|---:|---:|
| Completed input segments | 3 / 6 | 6 / 6 |
| Stale watermark warnings | 18 | 0 |
| Completed audio responses | 2 | 4 |
| Result | timed out | passed |

The model selected listen mode for the other two segments in the passing run.

### DIFF
--- a/vllm_omni/distributed/omni_connectors/transfer_adapter/chunk_transfer_adapter.py
+++ b/vllm_omni/distributed/omni_connectors/transfer_adapter/chunk_transfer_adapter.py
@@ -191,24 +191,31 @@ class OmniChunkTransferAdapter(OmniTransferAdapterBase):
 
         confirmed_num_computed_tokens = self._confirmed_num_computed_tokens(request)
 
-        # If the request is preempted, skip the already saved chunks.
-        if confirmed_num_computed_tokens < self.requests_num_chunks_sent.get(request.external_req_id, 0):
+        external_req_id = request.external_req_id
+        previous_num_computed_tokens = self.requests_num_chunks_sent.get(external_req_id, 0)
+
+        # Segment boundaries may restart token accounting from zero.
+        if not is_segment_finished and confirmed_num_computed_tokens < previous_num_computed_tokens:
             logger.warning(
-                f"Enqueue save_async for request {request.external_req_id}, "
+                f"Skip stale save_async for request {external_req_id}, "
                 f"request.num_computed_tokens={request.num_computed_tokens}, "
                 f"request.num_output_placeholders={getattr(request, 'num_output_placeholders', 0)}, "
-                f"previous_chunks_sent={self.requests_num_chunks_sent.get(request.external_req_id, 0)}"
+                f"previous_chunks_sent={previous_num_computed_tokens}"
             )
             return
 
-        self.requests_num_chunks_sent[request.external_req_id] = confirmed_num_computed_tokens
         task = {
             "multimodal_output": multimodal_output,
             "request": request,
             "is_finished": is_finished,
             "is_segment_finished": is_segment_finished,
         }
-        self._pending_save_reqs.append(task)
+        if is_segment_finished:
+            self._pending_save_reqs.append(task)
+            self.requests_num_chunks_sent.pop(external_req_id, None)
+        else:
+            self.requests_num_chunks_sent[external_req_id] = confirmed_num_computed_tokens
+            self._pending_save_reqs.append(task)
         with self._save_cond:
             self._save_cond.notify()
 
@@ -425,7 +432,6 @@ class OmniChunkTransferAdapter(OmniTransferAdapterBase):
 
         if is_segment_finished:
             self.code_prompt_token_ids.pop(external_req_id, None)
-            self.requests_num_chunks_sent.pop(external_req_id, None)
             self.ramp_chunk_count.pop(external_req_id, None)
             cached_ic = getattr(self, "_cached_ic", None)
             if cached_ic is not None:
--- a/tests/core/sched/test_omni_ar_scheduler_streaming.py
+++ b/tests/core/sched/test_omni_ar_scheduler_streaming.py
@@ -15,6 +15,7 @@ import pytest
 # isort: off
 import vllm_omni  # noqa: F401 - import for side effects (patch vLLM)
 from vllm.sampling_params import SamplingParams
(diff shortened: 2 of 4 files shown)


yuanheng-zhao:
This seems to be a bug on `main` as well. PTAL @Sy0307 @Shirley125 @Gaohan123

0xtoward:
> This seems to be a bug on `main` as well. PTAL @Sy0307 @Shirley125 @Gaohan123

Got it! I just rebased onto main.

amy-why-3459:
Please rebase onto main first — the PR is currently unmergeable (`dirty`).

The race fix LGTM: clearing `requests_num_chunks_sent` when the segment boundary is enqueued (instead of in `_send_single_request`) is the right place, and the new regression test covers the exact ordering bug.

0xtoward:
Rebased onto the latest `main` as a single commit. The PR is mergeable now, and all checks pass.

congw729 (commented):
Fix and regression test both look correct. One async-scheduling interleaving to confirm before approving (see inline), plus two nits.

vllm_omni/distributed/omni_connectors/transfer_adapter/chunk_transfer_adapter.py
```diff
@@ -183,24 +183,30 @@ def save_async(
...
         task = {
             "multimodal_output": multimodal_output,
             "request": request,
             "is_finished": is_finished,
             "is_segment_finished": is_segment_finished,
         }
         self._pending_save_reqs.append(task)
+        if is_segment_finished:
+            # Do not block the next segment on the background sender.
+            self.requests_num_chunks_sent.pop(external_req_id, None)
```
congw729, line 207: With async scheduling, a late in-flight output from the old segment can reach `save_async(..., is_segment_finished=False)` with `confirmed == C_old` after this pop. It passes the L190 check and re-sets the watermark to `C_old` at L199; with the sender-side pop removed, nothing clears it, so the next segment's chunks get dropped persistently — the same symptom this PR fixes. Can the runner emit `inter_stage_output` for a discarded in-flight step? If not, this is fine as-is; if yes, consider skipping that `save_async` at the scheduler call site.

vllm_omni/distributed/omni_connectors/transfer_adapter/chunk_transfer_adapter.py
```diff
@@ -183,24 +183,30 @@ def save_async(
...
-        # If the request is preempted, skip the already saved chunks.
-        if confirmed_num_computed_tokens < self.requests_num_chunks_sent.get(request.external_req_id, 0):
+        external_req_id = request.external_req_id
+        previous_num_computed_tokens = self.requests_num_chunks_sent.get(external_req_id, 0)
+
+        # Segment boundaries may restart token accounting from zero.
+        if not is_segment_finished and confirmed_num_computed_tokens < previous_num_computed_tokens:
             logger.warning(
-                f"Enqueue save_async for request {request.external_req_id}, "
+                f"Enqueue save_async for request {external_req_id}, "
```
congw729: nit: this branch skips the enqueue, but the message still says "Enqueue save_async" — consider "Skipping stale save_async".

vllm_omni/distributed/omni_connectors/transfer_adapter/chunk_transfer_adapter.py
```diff
@@ -183,24 +183,30 @@ def save_async(
...
                 f"request.num_computed_tokens={request.num_computed_tokens}, "
                 f"request.num_output_placeholders={getattr(request, 'num_output_placeholders', 0)}, "
-                f"previous_chunks_sent={self.requests_num_chunks_sent.get(request.external_req_id, 0)}"
+                f"previous_chunks_sent={previous_num_computed_tokens}"
             )
             return
 
-        self.requests_num_chunks_sent[request.external_req_id] = confirmed_num_computed_tokens
+        self.requests_num_chunks_sent[external_req_id] = confirmed_num_computed_tokens
```
congw729: nit: for boundary tasks this write is immediately popped at L207-209; an explicit if/else would avoid the dead store. Non-blocking.

vllm_omni/distributed/omni_connectors/transfer_adapter/chunk_transfer_adapter.py
0xtoward: Hi @congw729 , thanks for catching this. I’ve updated the PR, could you take another look?

0xtoward:
Hi @congw729 , could you take another look? Thanks!

Sy0307:
The enqueue-time reset fixes the first ordering race, but a late old-segment frame can still reach `save_async()` after the reset and write the old high watermark back. The current guard only handles non-empty `generated_token_ids`; a tokenless frame with `inter_stage_output` bypasses it. A valid next-segment count such as 3 is then dropped against the restored watermark 111. Please drop the whole stale frame before connector submission and add this scheduler-to-connector regression.

0xtoward:
Superseded by #6529, which integrated the segment-generation fence and retained this contribution via `Co-authored-by`. The cross-model follow-up landed in #6834. Closing this now-conflicting PR.

PR #5172, opened 2026-07-17
[Refactor] Reject unknown diffusion config fields

TaffyOfficial:
﻿## Purpose

`_DiffusionConfigProjection.from_kwargs()` and
`OmniDiffusionConfig.from_kwargs()` previously filtered by dataclass field
names before construction. A typo such as `enable_sleep_mod` therefore
disappeared and the runtime silently used the default.

This PR implements **PR 1: strict diffusion payload** from #5160.

## Design

```text
mixed stage arguments
    -> normalize supported diffusion aliases
    -> validate the complete key set against diffusion and other stage owners
    -> partition the diffusion-owned payload
    -> construct the strict projection / OmniDiffusionConfig
```

- Unknown diffusion fields now fail with the field name and stage id, including
  unknown fields whose value is `None`.
- Supported legacy aliases are normalized before validation and emit
  `FutureWarning`. Supplying both non-`None` legacy and canonical values is a
  conflict.
- Direct diffusion config construction and structured/legacy startup reuse the
  same normalization and ownership-validation primitives.
- Mixed shared engine fields and stage metadata are recognized before the
  diffusion payload is partitioned, so they are not misreported as unknown
  diffusion fields.
- Global shared `kv_cache_dtype` and `seed` stay outside diffusion stage config.
  Explicit `stage_N_kv_cache_dtype` / `stage_N_seed` and stage-local YAML values
  fail because diffusion has no consumer for them.
- Direct `OmniDiffusionConfig.from_kwargs(kv_cache_dtype=...)` remains a
  deprecated compatibility alias for `diffusion_kv_cache_dtype`.
- Strict construction no longer contains a final whitelist filter that can
  silently discard unowned keys.

## Scope

This PR intentionally does **not** change:

- flat versus nested parallel configuration or parallel precedence;
- OmniKV deploy/topology precedence;
- cache-backend fallback behavior;
- `enable_ar_profiler` routing;
- other fields without consumers or broader runtime ownership cleanup;
- diffusion algorithms, quantization behavior, or request-time extras.

Those changes require independent ownership and parity review and are not part
of RFC #5160 PR 1.

## Files

- `vllm_omni/diffusion/data.py`: alias normalization and pure strict validation.
- `vllm_omni/config/omni_config.py`: mixed-stage ownership validation,
  partitioning, and strict structured projection.
- `vllm_omni/engine/stage_init_utils.py`: startup consumes the validated
  diffusion-owned payload.
[...]

Issue #4021, opened 2026-05-31
[RFC]: Unified VllmOmniConfig


wuhang2014:
### Motivation.

| Problem | Impact |
|---------|--------|
| **`StagePipelineConfig` → `StageConfig` → `OmegaConf` → `dict`** | 4 representations for the same data, each hop re-serializes |
| **Legacy `ModelPipeline`/`_load_pipeline`** | Dead code (no `pipeline.yaml` files exist); 3 models still on `stage_configs/` YAML path |
| **`OmniDiffusionConfig` has 60+ flat fields** | Violates single responsibility; `enable_sleep_mode` declared twice; no shared interface with AR config |
| **Diffusion vs AR config creation are completely separate code paths** | No reuse of field resolution, normalization, or validation |

### Proposed Change.

Updates in [comment](https://github.com/vllm-project/vllm-omni/issues/4021#issuecomment-4714305383)

### Architecture

Constructed once at startup, passed to every stage engine during initialization. Exactly three top-level fields:

```
VllmOmniConfig
├── pipeline_config                ← existing PipelineConfig, 1 instance (model_type, stages, ...)
├── stage_configs            ← tuple of n per-stage configs
└── orchestrator_config      ← 1 instance, orchestrator-only
```

All other data lives in sub-config dataclasses, each owning a single concern.

```
pipeline_config: PipelineConfig  (existing class from stage_config.py)
├── model_type, model_arch           ← model identity
├── stages: (StagePipelineConfig,)   ← per-stage frozen topology tuple
├── hf_architectures                 ← HF architecture aliases for disambiguation
└── diffusers_class_name             ← for diffusers-style model_index.json lookup

stage_configs[i]: VllmOmniStageConfig
├── model_config             ← per-stage behavior: session defaults, compilation, weight loading
├── stage_pipeline_config    ← frozen pipeline topology (from pipeline.py)
├── load_config              ← weight loading (per-stage, can differ)
├── cache_config             ← GPU memory / KV cache (thinker 0.9, code2wav 0.1)
├── scheduler_config         ← batching / scheduling (thinker 64, diffusion 1)
├── connector_config         ← inter-stage wiring (stage-specific)
├── runtime_config           ← devices, replicas, env (always per-stage)
├── parallel_config          ← TP, SP, HSDP (stage 0 TP=4, stage 1 TP=1)
├── diffusion_config         ← diffusion-only, None for AR stages
└──quantization_config      ← per-stage (pipeline default overridable)

[...]

PR #5140, opened 2026-07-16
[Refactor 1/N] Centralize Omni config resolution

fake0fan:
## Summary

This PR introduces `resolve_omni_config()` as the single production configuration-resolution entrypoint for both `AsyncOmniEngine` and headless stage startup.

It has been rebased onto current `main` after the migrations tracked by #5266 landed, including removal of `--stage-configs-path`, internal `stage_configs_path` plumbing, and the legacy `stage_args` YAML loader. The implementation now reuses the registry + deploy-config path already present on main instead of carrying a parallel compatibility resolver.

The current OmegaConf-compatible stage shape remains a temporary runtime bridge while RFC #4021 moves consumers to typed `VllmOmniConfig.stage_configs`.

Progresses #4021.

## What changed

- Add `OmniConfigResolution` and `resolve_omni_config()` under `vllm_omni.config`.
- Route standard and headless startup through the same resolver contract.
- Resolve registered pipelines through `StageConfigFactory.create_from_model()` and reuse its existing runtime compatibility bridge.
- Return the effective deploy path, pipeline metadata, stage configs, and strategy-derived load-balancing policy together.
- Select generic single-stage diffusion only after registry resolution fails.
- Move generic diffusion normalization to `OmniDiffusionConfig` and parallel normalization to `DiffusionParallelConfig`.
- Remove engine-owned default diffusion stage synthesis and the second engine-side merge of LoRA, attention, quantization, cache, and profiler overrides.
- Remove obsolete source selection, model-config path guessing, mode filtering, and loader wrappers from `entrypoints/utils.py`.
- Parse `--stage-overrides` once at the CLI boundary and pass the typed mapping to both startup paths.
- Keep removed Python arguments (`stage_configs_path` and `stage_configs`) as explicit errors rather than silently restoring legacy behavior.

## Configuration flow

```text
model + PipelineConfig + deploy YAML + CLI/Python overrides
                              |
                              v
                    resolve_omni_config()
                    |-- structured config
                    |-- runtime compatibility stages
                    |-- generic diffusion fallback
                    |-- strategy-derived LB policy
                    '-- effective deploy path
                              |
                 +------------+------------+
                 |                         |
                 v                         v
          AsyncOmniEngine             headless stage
          consumes all stages         stage_by_id()
```

## Scope and compatibility

- No new user-facing config format is introduced.
- Registry-backed pipelines, deploy YAML, composable-parallel strategy input, per-stage overrides, and generic diffusion remain supported.
[...]

PR #5678, opened 2026-08-02
[Config] Read engine args from VllmOmniConfig

Acerak01-fy:
This is PR7 for RFC #4021. It adds the structured engine-argument projection from `VllmOmniConfig` while intentionally leaving current runtime inputs on the existing legacy `list[StageConfig]` / OmegaConf shape. The subsequent AR engine startup (PR8) can switch the consumer path as part of the coordinated stage-init cutover.

## Purpose

`build_engine_args_dict_from_omni_stage_config()` accepts one selected `BaseVllmOmniStageConfig` and builds the flat arguments expected by `OmniEngineArgs` or `OmniDiffusionConfig` directly from its typed sub-configs.It projects model, load, cache, scheduler, runtime, connector, parallel, quantization, and diffusion values without reconstructing them through legacy `engine_args`, OmegaConf, or dict/object compatibility access.

Engine inputs that would otherwise be lost during structured construction are retained by their concern-owned configs. This includes resolved model and tokenizer inputs, worker and platform overrides, connector state, diffusion VAE parallelism, `retains_state_across_chunks`, `logits_processors`, and `enable_session_state_manager`. The structured stage does not gain a catch-all legacy `engine_args` field.

Representation-independent finalization is shared by the legacy and typed implementations. This preserves model/tokenizer path precedence, Audex checkpoint and tokenizer resolution, CLI tokenizer fallback, connector-spec injection, diffusion attention normalization, worker resolution, generation cache defaults, sampling-extra state, and model/platform adjustments.

The engine startup/runtime path is not migrated in this PR. The existing implementation is named `build_legacy_engine_args_dict()`, while the stable `build_engine_args_dict()` entry point preserves its existing signature and
delegates to that legacy implementation. Current production consumers continue to call the stable entry point with legacy stage objects. The typed projection is prepared and tested independently until effective strategy state and the remaining stage-init inputs are ready for the coordinated cutover.

This also updates the related unit tests to cover:

- Typed engine arguments for representative AR, generation, and diffusion stages matching the existing legacy projection.
- Every currently registered pipeline retaining all backend-consumed legacy fields in the typed projection.
- Model and tokenizer subdirectory precedence, CLI tokenizer fallback, and Audex checkpoint behavior.
- Async connector injection, worker resolution, generation cache defaults, sampling-extra state, and diffusion attention normalization.
- Platform-sensitive ROCm attention defaults and diffusion parallel settings.
- The stable compatibility API matching the explicitly named legacy implementation without mutating legacy stage config state.
- Current runtime, headless, and diffusion startup consumers remaining on the stable compatibility entry point.

## Test Plan

vLLM Version: 0.25.0

vLLM-Omni Commit: 41540327

### Test Result

Static checks:

```bash
../.venv-vllm-omni-main/bin/ruff check \
  tests/engine/test_stage_engine_args.py \
  tests/entrypoints/test_serve.py \
  vllm_omni/config/omni_config.py \
  vllm_omni/engine/stage_init_utils.py \
  vllm_omni/engine/stage_runtime.py \
  vllm_omni/entrypoints/cli/serve.py

../.venv-vllm-omni-main/bin/ruff format --check \
  tests/engine/test_stage_engine_args.py \
  tests/entrypoints/test_serve.py \
  vllm_omni/config/omni_config.py \
  vllm_omni/engine/stage_init_utils.py \
  vllm_omni/engine/stage_runtime.py \
  vllm_omni/entrypoints/cli/serve.py
```

Result:

```text
All checks passed!
7 files already formatted
```

Unit tests:

```bash
../.venv-vllm-omni-main/bin/python -m pytest \
  tests/engine/test_arg_utils.py \
  tests/engine/test_async_omni_engine_stage_init.py \
  tests/engine/test_single_stage_mode.py \
  tests/engine/test_stage_engine_args.py \
  tests/engine/test_stage_metadata.py \
  tests/entrypoints/test_serve.py -q
```

Result:

```text
266 passed, 2 skipped, 22 warnings in 45.93s
```

Compatibility check against PR #5172 at `4a5487ba`:

```text
Automatic merge completed without conflicts.
312 passed, 2 skipped, 22 warnings in 41.12s
```


vllm_omni/engine/stage_init_utils.py
```diff
@@ -748,25 +765,112 @@ def stage_runtime_env(stage_id: int, runtime_cfg: Any) -> Generator[None, None,
...
-    if "tensor_parallel_size" in engine_args and engine_args["tensor_parallel_size"] is None:
-        del engine_args["tensor_parallel_size"]
+    """Copy defined typed config fields into backend adapter kwargs."""
+    projected: dict[str, Any] = {}
+    for config_field in fields(config):
+        name = config_field.name
+        if name in exclude:
+            continue
+        value = getattr(config, name)
+        if value is not None:
```
hsliuustc0106, line 780: This projects structured defaults that the legacy adapter leaves unset. For ming_flash_omni_image, legacy omits max_num_seqs (so OmniDiffusionConfig uses 1), but this emits 128 and startup rejects MingImagePipeline because it does not support request batching. The same mismatch changes the vLLM defaults for prefix caching, chunked prefill, hybrid-cache management, and GPU utilization. Please preserve the unset/backend state and compare effective parity in both directions.

vllm_omni/config/omni_config.py
```diff
@@ -751,7 +797,7 @@ def _global_stage_cli_fields() -> frozenset[str]:
         frozenset(f.name for f in fields(OmniEngineArgs))
```
hsliuustc0106, line 901: This accepts inherited OmniEngineArgs options globally, but the later per-concern allowlists silently discard options without an owner. I reproduced revision, tokenizer_revision, code_revision, seed, and download_dir surviving the legacy AR path and disappearing from the typed path. Please make ownership exhaustive or reject unowned fields instead of accepting and dropping them.

vllm_omni/config/omni_config.py
```diff
@@ -1116,12 +1196,27 @@ def _build_quantization_config(
...
+        kwargs["trust_remote_code"] = _copy_value(deploy.trust_remote_code)
+    if "dtype" not in kwargs and deploy.dtype is not None:
+        kwargs["dtype"] = _copy_value(deploy.dtype)
+    if "active_stream_window" not in kwargs:
+        kwargs["active_stream_window"] = _copy_value(deploy.active_stream_window)
+    if "custom_voice_dir" not in kwargs and deploy.custom_voice_dir is not None:
+        kwargs["custom_voice_dir"] = _copy_value(deploy.custom_voice_dir)
     if "has_sampling_extra_args" not in kwargs:
         kwargs["has_sampling_extra_args"] = bool((default_sampling_params or {}).get("extra_args"))
     if "model_subdir" not in kwargs and topology.model_subdir is not None:
```
hsliuustc0106, line 1354: The legacy builder lets stage engine_extras override topology model_subdir and tokenizer_subdir, but those keys never reach engine here, so this fallback restores the topology values. A direct reproduction selected /tmp/override-model through legacy and /tmp/top-model through typed. Please preserve deploy precedence and cover both subdirectory overrides.

vllm_omni/config/omni_config.py
```diff
@@ -1158,21 +1264,35 @@ def _build_scheduler_config(
...
-def _build_connector_config(stage_deploy: StageDeployConfig | None) -> OmniStageConnectorConfig:
+def _build_connector_config(
+    deploy: DeployConfig,
+    topology: StagePipelineConfig,
+    stage_deploy: StageDeployConfig | None,
+    engine: _ConnectorEngineOverrides,
+) -> OmniStageConnectorConfig:
     output_connectors = stage_deploy.output_connectors if stage_deploy is not None else None
     input_connectors = stage_deploy.input_connectors if stage_deploy is not None else None
+    omni_kv_config = _first_defined(
```
hsliuustc0106: This reverses the legacy omni_kv_config precedence: legacy applies topology after stage extras, while this selects the stage value first. With topology need_recv_cache=True and deploy need_send_cache=True, legacy emits receive and typed emits send. Please keep the legacy ordering; if the precedence change is intentional, document and test it as a behavior change.

Acerak01-fy: Done.

vllm_omni/config/omni_config.py
Acerak01-fy: Done.

vllm_omni/config/omni_config.py
Acerak01-fy: Done.

vllm_omni/engine/stage_init_utils.py
Acerak01-fy: Done.

vllm_omni/config/omni_config.py
```diff
@@ -739,21 +814,51 @@ def enrich_config(self) -> None:
...
+    | _PARALLEL_ENGINE_FIELDS
+    | _DIFFUSION_STAGE_ENGINE_FIELDS
+)
+
+
+def _validate_stage_cli_override_ownership(
+    stage_id: int,
+    overrides: Mapping[str, Any],
+) -> None:
+    unowned_fields = set(overrides) - _STAGE_ENGINE_FIELDS
```
hsliuustc0106: This validates only CLI inputs and uses an all-stage union. Deploy `engine_extras` are never checked, and LLM validation accepts diffusion-only fields: deploy `enable_lora=True` becomes `False`, while LLM CLI `kv_cache_dtype=fp8` becomes `auto` through the typed path. Please validate the merged input against execution-type-specific owners before partitioning.

tests/engine/test_stage_engine_args.py
```diff
@@ -0,0 +1,525 @@
...
+
+        comparable_fields = (legacy_args.keys() & backend_fields) - {
+            "cache_config",
+            "parallel_config",
+        }
+        assert {name: typed_args[name] for name in comparable_fields} == {
+            name: legacy_args[name] for name in comparable_fields
+        }
+
+        deferred_default_fields = backend_fields & _BACKEND_DEFERRED_DEFAULT_FIELDS
```
hsliuustc0106: This only compares the six hard-coded deferred defaults. Adding typed-only `skip_tokenizer_init=True` still lets the parity tests pass. Please compare effective values across all backend fields, or reject unexpected typed-only keys.

hsliuustc0106:
Can we align this PR with [#5140’s final target](https://github.com/vllm-project/vllm-omni/pull/5140#final-target-layered-configuration-structure), or document why this PR supersedes it?

#5140 gives each resolved stage Omni-owned topology/runtime/connector/request metadata plus one `engine_spec`: `EngineArgs` for AR/generation or an un-enriched `OmniDiffusionConfig` for diffusion, materialized in the owning process.

This PR instead splits selected `EngineArgs` fields across manual `OmniStage*Config` trees and `_..._FIELDS` sets, then flattens them back. Those buckets already differ from upstream ownership—`max_model_len`, tokenizer fields, and hybrid-cache settings are examples—and the inline threads show the result: accepted fields can be lost, execution-type validation can use the wrong owner set, and parity tests can miss typed-only defaults.

Since runtime has not cut over, please either rework the projector around one complete engine spec, or make the divergence explicit in #4021 and add a schema-derived exhaustive support check against the pinned upstream `EngineArgs`.

tests/engine/test_stage_engine_args.py
Acerak01-fy: Done

hsliuustc0106 (approved):
lgtm

PR #5681, opened 2026-08-02
[CI/Build] Gate one-word pronunciation on a success rate, not per request

### LINKED ISSUE #5395: [Bug] Qwen3-Omni talker garbles one-word answers (~13-27%/request, seed-deterministic, pre-existing on main); Whisper-small judge inflates test_one_word_prompt_001 failures ~2x

ShengleiFu:
## Purpose

Second test-side follow-up to #5395 

Whether a one-word answer is pronounced intelligibly is sampled, not guaranteed — the
talker runs at `temperature: 0.9`, which is the upstream default (it is also what HF's
`generate()` uses). Asserting it per request gates on a random variable, and the 10-retry
loop was compensating for exactly that. It also hid degradation: at a 20% per-request
failure rate a 5-request attempt still passes 33% of the time, so ten retries leave the
job green ~98% of the time.

This PR:

- Sends a fixed 40 requests per execution mode and requires at least 29 to transcribe as
  the expected word, instead of requiring every request to pass with up to 10 retries.
- Keeps the request shape: 8 batches of 5 concurrent requests, both `default` and
  `async_chunk` parametrizations, same prompt, same judge. Counting per-request outcomes
  does mean the assertion now runs inside each worker, so transcription is concurrent too
  where the previous `request_num=5` path sent concurrently but judged serially.
- Counts **only** an audio mismatch against the budget. An HTTP failure, missing audio or a
  text-keyword miss still fails immediately, so a broken server cannot hide inside the 11
  tolerated failures.
- Does no retries and no early exit, since both would bias the rate being measured.
- Reports the count, the Wilson interval and every failing transcript when the gate trips,
  so an on-call reader can tell sampling noise from a real drop.

## Why 29/40

85.6% is a **predeclared minimum acceptable success probability**, taken from a measured
HF reference run (77/90); vLLM-Omni itself measured 92.2% (83/90). The threshold was fixed
from that in advance, not tuned until CI went green. Assuming 40 independent Bernoulli
trials:

| gate | rejection at p=.856 | power at p=.60 | power at p=.70 |
|---|---:|---:|---:|
| >=29/40 | 0.9% | 92.9% | 55.9% |
| >=30/40 | 2.3% | 96.5% | 69.1% |
| >=31/40 | 5.4% | 98.4% | 80.4% |

These are nominal probabilities **conditional on p**, not a measured false-failure rate for
this CI. 77/90 is a single point estimate with a Wilson interval of 76.8%-91.4%; if the
true healthy rate sat at that lower endpoint, >=29/40 would reject about 20% of runs.
Establishing a real rate would need repeated report-only runs on the target hardware, and
the HF baseline was also measured serially rather than at this test's five-way concurrency.

30/40 looks stricter but the two parametrizations multiply: 2.3% per mode is 4.6% per
nightly at p=.856, which would trade one flaky gate for another. 29/40 keeps that at 1.8%.

**This is a smoke gate for gross pipeline breakage, not a quality tracker.** A drop to 70%
is caught only about half the time; resolving moderate drift needs ~80+ samples and belongs
in an accuracy benchmark rather than a per-request CI gate. The docstring says so, so the
gate is not later mistaken for quality coverage.

## Second commit: sharing the contract instead of matching it by hand

The success-rate loop has to tell an audio mismatch apart from a serving failure, and the
test was doing that by re-declaring the assertion's wording in the test file. Had that
wording drifted, the gate would have silently reclassified every quality failure as a hard
failure, with nothing to flag it.

`AUDIO_MISMATCH_MESSAGE` is now exported and used at both assert sites, alongside
`is_audio_mismatch()` and `wilson_interval()`. Two of the new unit tests drive the real
`assert_omni_response` to failure and classify what it actually raised, so a future
rewording fails there instead of at the gate.

I did not extract the batching loop into a general helper: with a single caller that would
be guessing at the right interface. Worth doing when a second gate wants the same shape.

## Test Plan

- [x] `pytest tests/helpers/tests/test_success_rate_helpers.py` — 12 passed, CPU only
- [x] Confirmed those tests have teeth: reintroducing a hardcoded message at one assert
      site turns `test_real_audio_mismatch_failure_is_classified` red
- [x] `ruff check` + `ruff format --check` (v0.14.10, as pinned) and
      `tools/pre_commit/check_test_marks.py`
- [x] `pytest tests/e2e/online_serving/test_qwen3_omni_expansion.py::test_one_word_prompt_001`
      on 2× RTX PRO 6000 with vLLM 0.26.0 and the stock `qwen3_omni_moe.yaml`, no local edits
- [x] Verified the power figures above against the binomial CDF rather than reusing them

## Test Result

```
$ pytest .../test_qwen3_omni_expansion.py::test_one_word_prompt_001 --run-level=full_model
one-word intelligibility: 38/40 succeeded     # default (sync)
one-word intelligibility: 37/40 succeeded     # async_chunk
2 passed in 972.61s (0:16:12)
```

95% and 92.5% against a 72.5% gate, so margins of 9 and 8 requests. `async_chunk` is not
meaningfully worse than `default`, which is why a single threshold covers both modes.

CI cost is close to unchanged: this run issues 80 requests where the previous shape issued
about 10, and total wall time went from 887s to 973s (+10%). The ~5 minute server start
dominates, and the worst case is now a fixed 40 requests per mode instead of up to 50.

```
$ pytest tests/helpers/tests/test_success_rate_helpers.py -q
12 passed in 0.11s
```

## Not a duplicate

No open PR touches `test_one_word_prompt_001` or the assertion helpers. #5395 is assigned
to me and its test-side half shipped as #5646.

---

AI assistance was used for this change (measurement, analysis and drafting). I have
reviewed every changed line and run the tests reported above.


tests/helpers/assertions.py
```diff
@@ -24,6 +25,30 @@
...
+def is_audio_mismatch(exc: BaseException) -> bool:
+    """True when a failure is audio not matching the text, rather than a serving error.
+
+    A test that tolerates a few mismatches while measuring a success rate must not
+    tolerate an HTTP failure, missing audio or a text-keyword miss as well.
+    """
+    return isinstance(exc, AssertionError) and AUDIO_MISMATCH_MESSAGE in str(exc)
+
+
+def wilson_interval(successes: int, total: int, z: float = 1.96) -> str:
```
yenuo26, line 125: If we add this check to improve stability, I actually think all the logic can be handled entirely inside the helper's assert — without requiring the test cases to explicitly reference it. Besides one_word, other test cases also suffer from instability due to this kind of interference. I think we could make this a general-purpose feature.

tests/helpers/tests/test_success_rate_helpers.py
```diff
@@ -0,0 +1,84 @@
...
+    AUDIO_MISMATCH_MESSAGE,
+    assert_omni_response,
+    is_audio_mismatch,
+    wilson_interval,
+)
+
+pytestmark = [pytest.mark.core_model, pytest.mark.cpu]
+
+
+def test_audio_mismatch_is_recognised_with_detail_appended():
```
yenuo26: Also, I think adding too many UT for test helper functions would be redundant. As long as the complex and critical logic is covered, that should be sufficient.

tests/helpers/assertions.py
ShengleiFu: Good point, updated, ptal, thanks!

tests/helpers/tests/test_success_rate_helpers.py
ShengleiFu: Removed unnecessary ones

tests/e2e/online_serving/test_qwen3_omni_expansion.py
```diff
@@ -474,17 +494,13 @@ def test_one_word_prompt_001(omni_server, openai_client) -> None:
...
-        except AssertionError as e:
-            if _similarity_assert_msg not in str(e) or attempt == _max_retries - 1:
-                raise
-            print(f"Similarity assertion failed, retrying {attempt + 2}/{_max_retries}: {e!r}")
+    # The judge runs per request so mismatches can be counted, which also transcribes
+    # concurrently where the previous request_num=5 path judged serially -- worth watching
+    # on tighter cards, since each judge loads whisper on the same device as the talker and
+    # code2wav stages.
+    openai_client.send_omni_request(
+        request_config, request_num=40, min_successes=29, max_concurrency=get_max_batch_size()
```
yenuo26, line 502: I'm not sure if 40 requests add too much overhead. If so, maybe we can scale it down proportionally.

yenuo26:
@vllm-omni-review-bot

tests/e2e/online_serving/test_qwen3_omni_expansion.py
ShengleiFu: I measured one before/after run on 2 x RTX PRO 6000, timing the request phase separately for both parametrizations:

| | before (up to 10 attempts) | after (fixed 40) |
|---|---:|---:|
| `default` | 25 requests, 248s | 40 requests, 184s |
| `async_chunk` | 5 requests, 66s | 40 requests, 185s |
| combined request phase | 314s | 369s |

The request phase increased by 55s (17.5%) across the two parametrizations in this run. The full test wall time happened to decrease from 887.83s to 644.77s, although startup and server setup variance means I would not attribute that reduction directly to this change.

The cost is now predictable: the old implementation used between 5 and 50 requests per parametrization depending on when an attempt passed, while the new implementation always uses 40. In this run, `default` needed five attempts and was slower than the fixed gate; `async_chunk` previously passed on its first attempt and was faster.

The effective request-phase time per sample also dropped from 9.9‚Äì13.2s to about 4.6s because outcomes are now judged concurrently, whereas the old `request_num=5` path transcribed them serially.

Reducing the sample count has a noticeable statistical cost. Under the same independent Bernoulli assumption and baseline \(p=.856\), using 15/20 would give about an 11.1% probability that at least one of the two healthy parametrizations fails, versus about 1.7% with 29/40. Lowering the threshold to 14/20 reduces per-parametrization detection power for a drop to \(p=.70\) from 55.9% to 39.2%, while still giving about a 3.5% combined healthy-failure probability.

Given the measured 55s request-phase increase and the loss of discrimination at \(n=20\), my preference is to keep 29/40. If CI time is the higher priority, I can scale it down.

A separate optimization would be to reuse the Whisper process/model: `convert_audio_file_to_text` currently starts a subprocess and reloads the model for each transcription. Since that helper is shared, I would handle it in a separate PR.

yenuo26 (approved):
LGTM

ShengleiFu:
Hi @amy-why-3459, CI is done, could you help merge it, thanks for applying `ready` label!

PR #5453, opened 2026-07-27
[Refactor] P0.2: Migrate API server helpers out of api_server

herotai214:
# [Refactor] P0.2: Migrate API server helpers out of api_server

cc @linyueqian  
See if/how a `P0.1: Add API server guardrails` should be added prior merging this PR.
This PR tgt with P0.3 serves as the fundamental structure of our whole refactor project.

## Summary

This is `Refactor [P0.2]` for #5227.

It moves non-route helpers and request models out of `vllm_omni/entrypoints/openai/api_server.py` into endpoint-owned or server-owned modules. The goal is to stop `api_server.py` from being the place other code imports helper implementation from.

This PR does **not** migrate endpoint route bodies yet. That is intentional: Phase 0 is split into:

- `P0.1`: API server guardrails/tests
- `P0.2`: helper migration, this PR
- `P0.3`: endpoint route-owner extraction, immediately after this PR

## What Changed

Moved API-server helpers into clearer owners:

- `openai/app_state.py` for app-state accessors
- `openai/chat_template.py` for chat-template bootstrap helpers
- `openai/diffusion.py` for shared diffusion-stage generation/sampling helpers
- `openai/lora.py` for OpenAI LoRA request parsing
- `openai/images/helpers.py` for image-only helpers
- `openai/models/serving.py` for diffusion-only `/v1/models` serving shim
- `serve/utils/errors.py` for server-level engine exception handling
- `serve/utils/routes.py` for app/router route-table mutation helpers
- `serve/profile/protocol.py` and `serve/profile/utils.py` for profiler route support
- `serve/omni_control/protocol.py` for sleep/wakeup request models

This follows upstream vLLM's current ownership direction where practical:

- server/app mechanics live under `entrypoints/serve/utils`;
- OpenAI-compatible response shaping stays under `entrypoints/openai`;
- model-list serving helpers live near `entrypoints/openai/models`;
- endpoint-family helpers move toward their owning OpenAI endpoint packages instead of remaining in the server composition root.

Updated tests and serving imports to use the new helper locations.

## Why `api_server.py` Still Imports These Helpers

This PR intentionally does not move endpoint bodies yet, so `api_server.py` temporarily imports helpers from their new owners.

For example:

```python
from vllm_omni.entrypoints.openai.images.helpers import (
    _check_max_generated_image_size,
    _extract_images_from_result,
    _load_input_images,
)
from vllm_omni.entrypoints.openai.lora import _parse_lora_request
[...]

PR #5272, opened 2026-07-21
[Migrate] Move sampling parameter overrides from legacy dispatches to TTS adapters

sphinxkkkbc:
PLEASE FILL IN THE PR DESCRIPTION HERE.

## Purpose

M3 of #4855. This PR moves model-specific `sampling_params_list` mutations from `serving_speech` into the corresponding TTS adapters through`apply_sampling_overrides`. Removed model-specific dispatch logic in `serving_speech` while preserving the existing behavior for legacy dispatch paths.

## Compatibility Note

The legacy `glm_tts` and `cosyvoice3` paths previously applied their `sampling_params_list` overrides before merging `request.extra_params`. With the unified adapter flow, the order is now: `stream coercion -> extra_params -> apply_sampling_overrides -> seed`, matches the contract documented in the adapter base class.

These adapters do not read or modify `sampling_params_list[0].extra_args`, so changing the order of these two independent operations does not change their behavior.

For `cosyvoice3`, the legacy flow applied the dynamic token limits in two steps:

```python
sampling_params_list[0].min_tokens = max(1, int(text_token_len * min_ratio))
sampling_params_list[0].max_tokens = min(2048, int(text_token_len * max_ratio))
```

When `request.max_new_tokens` was provided, it then applied the request-level cap:

```python
sampling_params_list[0].min_tokens = min(
    getattr(sampling_params_list[0], "min_tokens", 0),
    request.max_new_tokens,
)
sampling_params_list[0].max_tokens = request.max_new_tokens
```

The migrated `CosyVoice3Adapter.apply_sampling_overrides()` consolidates these operations and preserves the previous behavior.

## Test Plan

`pytest -v tests/entrypoints/openai_api/test_serving_speech.py`
`pytest -v tests/worker/test_gpu_ar_model_runner.py `

**vLLM Version:** 0.26.0

**vLLM-Omni Commit:** 4fc8892996fca97a652d05092793e0fba11e3c08

## Test Result

`204 passed, 17 warnings in 1.44s `
`23 passed, 16 warnings in 1.14s`

**BEFORE SUBMITTING:** read [CONTRIBUTING.md](https://github.com/vllm-project/vllm-omni/blob/main/CONTRIBUTING.md) and run the [precheck-pr skill](https://github.com/vllm-project/vllm-omni/blob/main/.claude/skills/precheck-pr/SKILL.md) with the code agent for a self-check against project conventions.

(anything written below this line will be removed by GitHub Actions)

PR #4976, opened 2026-07-09
[New Model] support nemotron Audex

yuekaizhang:
PLEASE FILL IN THE PR DESCRIPTION HERE.

## Test Plan

**vLLM Version:**

**vLLM-Omni Commit:**

## Test Result

**BEFORE SUBMITTING:** read [CONTRIBUTING.md](https://github.com/vllm-project/vllm-omni/blob/main/CONTRIBUTING.md) and run the [precheck-pr skill](https://github.com/vllm-project/vllm-omni/blob/main/.claude/skills/precheck-pr/SKILL.md) with the code agent for a self-check against project conventions.
(anything written below this line will be removed by GitHub Actions)

Issue #4855, opened 2026-07-03
[Refactor][TTS]: Problems and improvement directions (async-chunk reliability, serving_speech, runtime hot path)


linyueqian:
Part of #4843. TTS module self-review for the Q3 refactor. Line numbers are against main `5c51782c`. Perf numbers come from runs on an H20 (nsys CUDA traces, py-spy at function granularity, a per-step scheduler probe, concurrency sweeps c=1 to c=128); the async-hang findings were verified path-by-path in the code. Happy to share the profiling harness, raw profiles, and the patches mentioned below.

Three problems, in the order we would fix them:

| # | Problem | Kind | Evidence status |
|---|---------|------|-----------------|
| 1 | Async-chunk streaming has no hang safety net; engine failure paths can wedge or over-kill | reliability | every wait traced in code; matches 7 reported issues (1 still open) |
| 2 | `serving_speech.py` adapter migration stopped halfway; 41 model branches remain | maintainability | branch inventory + adapter coverage counted on main |
| 3 | Per-step host work in the worker keeps the GPU ~87% idle under load | performance | nsys + py-spy attribution; fix A/B'd on main, +13% throughput |

---

## Status

Kept here rather than in a comment so it does not scroll away. Every state below was checked against the API on 2026-08-13, not from memory. **Please check this table before starting an item** — four of these were picked up independently and I only noticed after writing a duplicate of one myself (see the note under R1.1).

| Item | State | Where |
|------|-------|-------|
| R1.1 chunk-wait deadline | in review | #3944 (@Ronnie-Rui) — approved, DCO outstanding |
| R1.2 send-failure reporting | in review | #6033 |
| R1.3 prewarm skip fails | in review | #6033 |
| R1.4 timeout validation | in review | #6033 |
| R2.1 non-blocking `sync_q.put` | **open** | — |
| R2.2 per-replica fault isolation | **landed** | #4583 (@ShengleiFu), merged 11 Aug |
| M1 label ladder | **landed** | #4862 (@sphinxkkkbc), merged 4 Jul |
| M2 leaf validators into adapters | **open** | — |
| M3 sampling overrides | in review | #5272 (@sphinxkkkbc) |
| M4 capability metadata | in review | #6138 (@sphinxkkkbc) |
| M5 batch path through adapters | **open** | — |
| M6 delete detection flags | **open** | detection ladder landed in #5682; the flag deletion and CI flip remain |
[...]

PR #5682, opened 2026-08-02
[Refactor][TTS] Derive TTS model detection from adapter metadata

linyueqian:
## Purpose

Every TTS adapter already declares the `model_stage` key(s) its model is deployed under:

```python
# tts_adapters/cosyvoice3.py
stage_keys = frozenset({"cosyvoice3_talker"})
```

But that field was documented as decoration —

```python
# tts_adapters/base.py:103 (before)
#: Engine ``model_stage`` key(s) this model uses, for documentation only.
```

— while `serving_speech.py` re-declared the identical sets as 20 module constants and re-implemented the mapping as a 47-line ladder:

```python
_COSYVOICE3_TTS_MODEL_STAGES = {"cosyvoice3_talker"}
...
if model_stage in _COSYVOICE3_TTS_MODEL_STAGES:
    return "cosyvoice3"
```

Two declarations of the same fact, in two files, that had to be edited together — and only one of them was load-bearing. Adding a TTS model meant editing the shared serving module in three separate places before the adapter was even reachable.

This makes the adapter metadata the single source of truth. Part of the `serving_speech.py` migration tracked in #4855 (extends the adapter framework from #4327).

## What changes

**`TTSModelAdapter`** gains the metadata detection needs:

| Field | Purpose |
|---|---|
| `model_archs` | architectures identifying this model, for the few whose stage key is ambiguous or absent |
| `arch_identifies_entry_stage` | model owns no stage key; stage discovery must find it by architecture |
| `detect_priority` | resolution order; only set to break a genuine overlap |
| `matches()` | the match rule, overridable when it is not set membership |
| `stage_serves_speech()` | for models speech-capable only in some deployment topologies |

**`tts_adapters`** derives the mapping: `detect_tts_model_type()`, `all_tts_stage_keys()`, `tts_entry_stage_archs()`.

**`serving_speech.py`** drops the 20 constants and the ladder. `_detect_tts_model_type` is now three lines, and `_find_tts_stage` asks the registry instead of hardcoding Audex's topology rule and Ming's architecture.

Only three adapters needed more than the `stage_keys` they already had: VoxCPM2 (talker architecture is authoritative), Ming dense (no stage key of its own), CoVo (shares the generic `fused_thinker_talker` key, so the architecture disambiguates).

### Architecture matching is a fallback

Ming dense deploys its AR stage as the generic `model_stage="llm"` (`ming_tts/pipeline.py:24`), while Ming *flash-omni* owns the `ming_tts` stage key. So architecture matching runs **after** every adapter that owns a real stage key (`detect_priority = 200`). The reverse order would let an architecture claim stage keys other adapters own.

`ming_flash_omni_tts` has no adapter yet, so it stays detectable through an explicit `LEGACY_TTS_DETECTORS` entry rather than an implicit branch — visible, counted, and deletable when it migrates.

## Behaviour

Unchanged. `test_tts_detection.py` keeps a verbatim copy of the deleted ladder and asserts the new implementation agrees with it across all **43 `model_stage` values any in-tree pipeline emits** (enumerated, with a test that fails if an adapter ever claims a key absent from that list), against every architecture no adapter claims — plus each claimed architecture against the stage keys it really ships with. It also asserts that `all_tts_stage_keys()` equals the old `_TTS_MODEL_STAGES` exactly, that resolution order is total and independent of import order, that no two same-priority detectors can match one input, and that detection can only name an adapter-backed or explicitly declared type.

**One deliberate divergence**, pinned in its own test with the rationale: a Ming-architecture checkpoint deployed under *another* model's stage key. The ladder tested Ming's architecture at a fixed position partway down, so the answer depended on whether the other model's stage key happened to sit above or below it; detection now always gives the stage key priority. No pipeline can produce that combination — a checkpoint's architecture and a pipeline's stage key cannot be mixed across models.

Writing that oracle is what caught this: my first implementation gave `model_archs` unconditional precedence, and the test failed on `audex_omni` + Ming architecture.

Also drops `_is_fish_speech` and `_is_cosyvoice3`, which had no production readers left after #4833.

## Keeping it this way

Adding a TTS model no longer requires editing `serving_speech.py`. Two things hold that:

- **`tools/pre_commit/check_tts_adapter.py`** ratchets the remaining `self._tts_model_type == ...` comparisons (29 today) and the number of unmigrated model types. Both may only decrease; the hook insists you lower the budget after removing branches, so the recorded numbers cannot drift upward silently. Raising one means editing the file, which makes it a reviewable decision. It counts direct comparisons, `getattr` reads, `match` subjects and single-hop local aliases. `tests/tools/test_check_tts_adapter.py` pins each of those and also pins table dispatch as a **known gap** — the gate guards against drift in the shape people actually write, not against deliberate evasion, and I would rather that limit be asserted than assumed.
- **The docs stop teaching the anti-pattern.** `docs/contributing/model/adding_tts_model.md` and `.claude/skills/add-tts-model/SKILL.md` both instructed contributors to add a stage constant, a union entry, a detection branch, a validation-dispatch branch, and per-model methods to `serving_speech.py` — five edits to the shared module, described as mandatory and "in a single commit". Both now describe writing one adapter. This is the actual reason the migration kept losing ground: new models were adding branches faster than the migration removed them.

The old "rebase conflicts in `serving_speech.py` are expected, always keep both sides" guidance goes away with it — a new adapter is a new file, so the only shared line is its import.

## One test-marking fix, folded in

The Audex topology rule — `audex_omni` serves speech only when the speech decoder is deployed alongside it — moves out of a hardcoded branch in `_find_tts_stage` and onto `AudexAdapter.stage_serves_speech()`. Its regression coverage is `tests/entrypoints/openai_api/test_audex_serving_guards.py`, which turns out to have carried **no `pytestmark`** since it landed with #4976, so the CPU lane (`-m "core_model and cpu"`) deselects the entire file. `check_test_marks.py` flags it today; the hook was bypassed when the file was added.

Since this PR moves the very behaviour those tests protect, marking them here rather than in a separate PR: 49 deselected tests become 49 running ones, no test bodies changed.

While checking that, I swept the tree with the same hook: **23 of 648 test files are unmarked** (18 missing hardware, 4 missing both, 1 missing level) and correspondingly under-collected. The rest each need a judgement call about which lane they belong to — mostly GPU suites where enabling them may surface real failures — so they are out of scope here. Happy to file a tracking issue with the list.

## Test

```
pytest tests/entrypoints tests/tools/test_check_tts_adapter.py -m "core_model and cpu"
1382 passed, 2 skipped, 0 failed
```

`ruff check` and `ruff format --check` clean. (`tests/tools/test_generate_nightly_perf_excel.py` fails collection on this checkout; it does so identically on `upstream/main`, so it is untouched by this PR.)

Also booted VoxCPM2 (`--deploy vllm_omni/deploy/voxcpm2.yaml`) on an L20X, which exercises the new architecture-first path end to end: the warmup request is gated on `_tts_model_type == "voxcpm2"` and fired, so detection resolved through the registry on a real deployment. That box has a pre-existing vLLM version skew (`AttributeError: 'OmniRequest' object has no attribute 'num_in_flight_tokens'`) that kills the engine core before audio comes out; I A/B'd it against `upstream/main` on the same box and it reproduces identically there, so it is unrelated to this change.

## Second review

I had this reviewed adversarially before opening it. It found no deployment-breaking bug and confirmed the equivalence argument independently — including the point that CoVo's inherited `code2wav` stage resolves to `None` under both implementations, so its absence from the pair list hides nothing. Two low-severity findings, both fixed here:

- The ratchet was bypassable by aliasing, `getattr`, or `match`. Now counted, with tests.
- My first draft of the docs told contributors to use `DiffusionTTSAdapter` for diffusion-engine models. That would have been actively misleading: `create_speech()` short-circuits to `_create_diffusion_speech()` under `for_diffusion()`, so a diffusion adapter's `validate()`/`build()` never run, and there is no production subclass. Both docs now say so explicitly.

Writing the equivalence oracle also turned up something worth knowing independently of this PR: **`omnivoice_generator` is claimed by the OmniVoice adapter and was in the deleted ladder, but no pipeline emits it.** OmniVoice is served through the diffusion path, which sets the model type directly and never consults detection. The key is vestigial in both the old and new code; this PR preserves it as-is rather than quietly changing it.

## Notes for reviewers

- No overlap with #5453 (P0.2), which touches `api_server.py` and its new packages.
- Touches `tts_adapters/base.py` and three adapter files, so it will need a trivial rebase against #5272 (sampling overrides), which adds methods to the same classes. Happy to land in either order.

Refs #4327, #4855

### DIFF
--- a/.claude/skills/add-tts-model/SKILL.md
+++ b/.claude/skills/add-tts-model/SKILL.md
@@ -220,65 +220,70 @@ See `plan/voxcpm2_native_ar_design.md`.
 
 ### Steps
 
-1. **Register in `serving_speech.py`** — add all 5 points in a **single commit**;
-   partial integration causes hard-to-debug failures. This file is modified by every
-   model PR and is the most common source of rebase conflicts — see conflict note below.
+1. **Write one adapter** under `vllm_omni/entrypoints/openai/tts_adapters/`.
+   `serving_speech.py` should not need an edit — detection, stage discovery and
+   dispatch are all derived from what the adapter declares.
 
-   **Point 1** — stage constant (near the top, alongside the other `_*_TTS_MODEL_STAGES` sets):
+   Create `vllm_omni/entrypoints/openai/tts_adapters/your_model.py`:
    ```python
-   _YOUR_MODEL_TTS_MODEL_STAGES = {"your_stage_key"}
+   @register_tts_adapter
+   class YourModelAdapter(ARTTSAdapter):
+       name = "your_model"                              # registry key + log label
+       stage_keys = frozenset({"your_stage_key"})       # the deploy yaml's model_stage
+
+       def validate(self, request) -> str | None:
+           if not request.input or not request.input.strip():
+               return "Input text cannot be empty"
+           return None
+
+       async def build(self, request, sampling_params_list, has_inline_ref_audio):
+           params = {"text": [request.input]}
+           if request.voice is not None:
+               params["voice"] = [request.voice]
+           return PreparedRequest(
+               prompt={"prompt": request.input},
+               tts_params=params,
+               model_type=self.name,
+           )
    ```
 
-   **Point 2** — union into `_TTS_MODEL_STAGES`:
-   ```python
-   _TTS_MODEL_STAGES: set[str] = (
-       ...
-       | _YOUR_MODEL_TTS_MODEL_STAGES
-   )
-   ```
+   Then add the module to the import block at the bottom of
+   `tts_adapters/__init__.py` so it registers. That import line is the only shared
+   file a new model touches — which is also why the old rebase-conflict hotspot is
+   gone.
 
-   **Point 3** — model type detection in `_detect_tts_model_type()`:
-   ```python
-   if model_stage in _YOUR_MODEL_TTS_MODEL_STAGES:
-       return "your_model"
-   ```
+   > **Pure-diffusion TTS does not go through adapters yet.** Under
+   > `for_diffusion()`, `create_speech()` routes straight to
+   > `_create_diffusion_speech()` and never calls `validate()`/`build()`.
+   > `DiffusionTTSAdapter` is scaffolding with no production subclass, so logic
+   > placed in one would silently never run. Diffusion-engine models follow the
+   > existing diffusion path; wiring it through adapters is open work (#4855).
 
-   **Point 4** — validation dispatch in `_validate_tts_request()`:
-   ```python
-   if self._tts_model_type == "your_model":
-       return self._validate_your_model_request(request)
-   ```
+   **If a stage key alone cannot identify the model**, declare
+   `model_archs = frozenset({"YourModelForConditionalGeneration"})`; add
+   `arch_identifies_entry_stage = True` when the model owns no stage key at all
+   (Ming dense). For a rule that is not set membership, override `matches()` (see
+   `covo_audio.py`). For a genuine overlap with another adapter, give one an
+   explicit `detect_priority` — `test_tts_detection.py` fails on an unordered
+   overlap. Models that only serve speech in some topologies override
+   `stage_serves_speech()` (see `audex.py`).
 
-   **Point 5** — validation + parameter-builder methods:
-   ```python
-   def _validate_your_model_request(self, request) -> str | None:
-       if not request.input or not request.input.strip():
-           return "Input text cannot be empty"
-       return None
+   **Reuse shared helpers via `self.ctx.server`** rather than reimplementing them:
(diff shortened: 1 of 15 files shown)


sphinxkkkbc:
Opened #5746 to migrate `ming_flash_omni_tts` to the adapter framework. Once that PR is merged, the corresponding `LegacyDetector` entry will no longer be needed and can be removed.

hsliuustc0106 (approved):
No blocking findings at `2c9201405139dc5981d232d66e8639b45530b233`. The registry parity, model-specific detection ordering, Audex topology guard, documentation, and regression coverage are coherent. AMD CI is still pending.

Issue #5386, opened 2026-07-24
vLLM-Omni Q3 Config / Entrypoints Roadmap


alex-jw-brooks:
Tracking some of the ongoing discussions around config cleanup and entrypoints, since we have had some relevant discussions around it in the English community syncs, and it would be helpful to link subissues to as we progress to ensure things are moving towards a better state + help coordinate PRs.

### VllmOmniModelConfig
- See details in the RFC: https://github.com/vllm-project/vllm-omni/issues/4021; most PRs are linked there for now since its pretty comprehensive
    - [ ] Config resolver: https://github.com/vllm-project/vllm-omni/pull/5140 (@fake0fan)

Milestones:
- [x] Initial OmniModelConfig implementation
- [ ] OmniModelConfig is cleanly integrated with composable parallel strategy config
- [ ] OmniModelConfig is active on the production path

Mostly being worked on by @zwhzzz0821 @Acerak01-fy @alex-jw-brooks @lishunyang12

### Legacy Migration / Deprecation Policies
- [ ] Migrate all legacy state configs to deploy configs (@zwhzzz0821 @lishunyang12)
- [ ] Remove legacy stage configs path (@zwhzzz0821)
- [ ] Add more formal deprecation policies and timeline (i.e., analogous to [this](https://docs.vllm.ai/en/stable/contributing/deprecation_policy) for vLLM).
    - For this, we should have consistent infrastructure for logging when deprecated things will be removed by so that we can ensure that they are actually removed. 
- [ ] Remove OmegaConfig integration (@alex-jw-brooks)

### Code Health
- [ ] Handle environment variables in a way that better aligns with vLLM's patterns
- [ ] Standardize naming conventions for pipelines with multiple variants
- [ ] Remove hardcoded kwarg string lists, since these are hard to maintain
- Move away from passing **kwargs / dicts in different places and add/reuse better data structures to prevent silently ignored options
- Remove overly defensive fallbacks (i.e., `getattr`s / `.get` defaults), add better type hints
    - [X] https://github.com/vllm-project/vllm-omni/pull/5199 (@princepride)
- Misc. fixes / improved behaviors
    - [X] https://github.com/vllm-project/vllm-omni/pull/4115

### Entrypoints / Config Auditing
- [ ] Test entrypoints to see which flags are actually being applied correctly in configs, and which aren't; some are silently ignored or are confusing (@akshatvishu)
[...]

PR #5685, opened 2026-08-02
[Core] Add typed diffusion request option specs

lfqaEWFC:
# Add Typed Diffusion Request Option Specs

## Summary

Part of the config/entrypoints roadmap tracked in #5386 (replacing loose kwargs/dict plumbing with typed declarations). This PR introduces typed, extensible request-option specifications for diffusion models while preserving compatibility with existing dictionary-based declarations.

```text
OpenAI / offline request
    (extra_body / extra_params)
            │
            ▼
normalized request options
            │
            ▼
OmniDiffusionSamplingParams.extra_args
            │
            ▼
model-specific consumers
```

The goal is to provide a typed and extensible foundation for model-specific request options without changing existing runtime behavior.

## Changes

* Introduce typed `ModelExtraSpec` for declaring model-specific request and output options, with shared frozen instances where model aliases overlap.
* Add typed `DiffusionRequestOptionSpec` for request normalization, replacing three loose kwargs with a single required, typed declaration.
* Preserve compatibility with existing dictionary-based model declarations.
* Route declared root-level and `extra_body` request options into `OmniDiffusionSamplingParams.extra_args`.
* Emit once-per-process warnings for unrecognized request-root and `extra_body` fields (filtered against declared, registered, and aliased fields) to flag likely typos.
* Bugfix: fields consumed directly by the serving layer (`return_stage_metrics`) are exempt from the warning; the exemption affects diagnostics only, not routing.
* Preserve pass-through of keys nested inside `extra_args` / `extra_params` containers for backward compatibility.
* Document the new warning behavior in `docs/serving/chat_completions_api.md`.
* Add focused unit tests covering typed and legacy declarations, declared option routing, typo warnings, exemptions, and pass-through compatibility.

## Compatibility

This change is fully backward compatible.

Existing dictionary-based declarations continue to work without modification, and undeclared request options nested inside `extra_args` / `extra_params` continue to be forwarded exactly as before.

## Out of Scope

This PR intentionally does **not** modify:

* CLI or environment argument propagation
* `DeployConfig` or `StageDeployConfig`
* runtime payloads such as `submit_kwargs` or `EngineQueueMessage.kwargs`
* executor sampling behavior
* repository-wide model keyword cleanup

These areas are planned as independent follow-up work to keep review scope focused.

## Validation

Validated with:

* **115 tests passed**

  * `tests/model_extras/test_model_extras.py`
  * `tests/entrypoints/openai_api/test_diffusion_request_utils.py`
  * `tests/entrypoints/openai_api/test_serving_chat_sampling_params.py`
* pre-commit hooks passed (ruff check/format, typos, sign-off)
* `git diff --check` passed


hsliuustc0106:
@alex-jw-brooks @NickCao PTAL

vllm_omni/entrypoints/openai/diffusion_request_utils.py
```diff
@@ -22,16 +49,43 @@ def _request_mapping(name: str, value: object | None) -> dict[str, object]:
...
 
 
+def _warn_for_close_unknown_fields(
+    args: Mapping[str, object],
+    *,
+    source: str,
+    allowed_fields: Collection[str],
+) -> None:
+    allowed = set(allowed_fields) | _EXTRA_ARG_CONTAINER_FIELDS
+    choices = sorted(allowed_fields)
```
hsliuustc0106: `extra_args` and `extra_params` are valid routes but are not candidate choices here. `extra_agrs={"solver": "euler"}` is accepted as a request extra, then silently dropped without a warning. Please include the container names in `choices` and add a regression test.

lfqaEWFC: Thanks, good catch. I updated the suggestion candidates to include the valid container routes (`extra_args` and `extra_params`), not just the declared model fields. I also added a regression test for the `extra_agrs={"solver": "euler"}` case to ensure it now emits a warning suggesting `extra_args` instead of being silently ignored.

tests/entrypoints/openai_api/test_serving_chat_sampling_params.py
```diff
@@ -133,6 +134,54 @@ def test_unknown_root_extra_does_not_claim_canonical_extra(serving_chat):
...
+    assert diffusion_request_args == {}
+    assert not any("cfg_text_sclae" in record.message for record in caplog.records)
+
+
+def test_misspelled_extra_args_container_warns(serving_chat, caplog):
+    serving_chat._diffusion_extra_body_params = frozenset({"cfg_text_scale"})
+    request = ChatCompletionRequest(
+        model="test",
+        messages=[],
+        extra_args={"solver": "euler"},
```
alex-jw-brooks: I think this is test is incorrect, right?

vllm_omni/model_extras/registry.py
```diff
@@ -98,6 +99,40 @@
...
+        extra_body_params=frozenset(extra_body_params),
+        extra_output_params=frozenset(extra_output_params),
+        init_extra_args_for_non_diffusion_stages=init_extra_args_for_non_diffusion_stages,
+        text_to_image_prompt_builder=text_to_image_prompt_builder,
+        image_to_image_prompt_builder=image_to_image_prompt_builder,
+        image_to_video_prompt_builder=image_to_video_prompt_builder,
+    )
+
+
+ModelExtraSpecLike = ModelExtraSpec | Mapping[str, Any]
```
alex-jw-brooks: IMO we should actually avoid backwards compatibility here if we are migrating to a well-defined `ModelExtraSpec`  and can port all of the existing configs. Otherwise we risk supporting both for awhile even though rebasing in-flight PRs to this pattern is straightforward

vllm_omni/model_extras/registry.py
```diff
@@ -98,6 +99,40 @@
...
+    extra_body_params: Collection[str] = (),
+    extra_output_params: Collection[str] = (),
+    init_extra_args_for_non_diffusion_stages: bool = False,
+    text_to_image_prompt_builder: TextToImagePromptBuilder | None = None,
+    image_to_image_prompt_builder: ImageToImagePromptBuilder | None = None,
+    image_to_video_prompt_builder: ImageToVideoPromptBuilder | None = None,
+) -> ModelExtraSpec:
+    return ModelExtraSpec(
+        extra_body_params=frozenset(extra_body_params),
+        extra_output_params=frozenset(extra_output_params),
```
alex-jw-brooks: Can you remove the extra `model_extra_spec` wrapper and just create the `ModelExtraSpec` directly with the dataclass?

vllm_omni/model_extras/registry.py
```diff
@@ -260,10 +295,24 @@ def default_image_to_video_prompt(
 _EXTRA_SPECS["Mammothmoda2Model"] = _EXTRA_SPECS["MammothModa2DiTPipeline"]
 
 
-def _get_spec(model_class_name: str | None) -> dict[str, Any] | None:
+def _coerce_model_extra_spec(spec: ModelExtraSpecLike) -> ModelExtraSpec:
```
alex-jw-brooks: Related to above, I think we should remove support for dicts and just enforce that everything is a ModelExtraSpec where it should be so that we don't need coercion, especially since it can add another layer of defaults

vllm_omni/model_extras/registry.py
```diff
@@ -303,11 +352,9 @@ def build_text_to_image_prompt(
...
 ) -> dict[str, Any]:
     spec = _get_spec(model_class_name)
-    builder: TextToImagePromptBuilder = (
-        spec.get("text_to_image_prompt_builder", default_text_to_image_prompt)
-        if spec is not None
-        else default_text_to_image_prompt
-    )
+    builder = spec.text_to_image_prompt_builder if spec is not None else None
+    if builder is None:
+        builder = default_text_to_image_prompt
```
alex-jw-brooks: nit: I think 
```python
 builder = spec.text_to_image_prompt_builder if spec is not None else default_text_to_image_prompt
```

or 
```python
builder = default_text_to_image_prompt
if spec is not None:
    builder = spec.text_to_image_prompt_builder
```
would be a bit more readable. Same with the image/video builders

vllm_omni/entrypoints/openai/diffusion_request_utils.py
```diff
@@ -22,16 +49,43 @@ def _request_mapping(name: str, value: object | None) -> dict[str, object]:
...
+    *,
+    source: str,
+    allowed_fields: Collection[str],
+) -> None:
+    allowed = set(allowed_fields) | _EXTRA_ARG_CONTAINER_FIELDS
+    choices = sorted(allowed)
+    for key in sorted(args):
+        if key in allowed:
+            continue
+        matches = get_close_matches(key, choices, n=1, cutoff=0.82)
```
alex-jw-brooks: While I do appreciate the fuzzy matching here, I think it may actually be better to just log the names of ignored arguments as a warning instead (if anything), because the log will show on the server side, so it won't be easily visible to the client in most cases.

Also, since fuzzy matching is computationally expensive, I think there could be a (sort of obscure) scenario where this could be used maliciously a bit depending on where it is called. E.g., by sending intentional invalid giant `args` dicts that take longer to run fuzzy matching than most cases

vllm_omni/entrypoints/openai/diffusion_request_utils.py
```diff
@@ -4,13 +4,40 @@
...
+
+    serving_root_fields: frozenset[str] = field(default_factory=frozenset)
+    registered_extra_fields: frozenset[str] = field(default_factory=frozenset)
+    root_field_aliases: Mapping[str, str] = field(default_factory=dict)
+
+    @classmethod
+    def from_fields(
+        cls,
+        *,
+        serving_root_fields: Collection[str] = (),
```
alex-jw-brooks: Can you remove the default values here, since the only caller passes them?

alex-jw-brooks (changes requested):
Thanks for looking at this! I think it's a good direction. Some suggestions

lfqaEWFC:
> 感谢您的关注！我认为这是个好方向。以下是一些建议。

@alex-jw-brooks ,Thanks for the review! All comments are addressed in the latest commits:

1. Incorrect test — removed test_misspelled_extra_args_container_warns; undeclared keys inside extra_args always pass through, which is already covered by test_unknown_extra_args_keys_still_pass_through. 2–4. Typed-only registry — removed ModelExtraSpecLike, the model_extra_spec() wrapper, and the coercion layer; _EXTRA_SPECS is now dict[str, ModelExtraSpec] constructed directly with the dataclass.
2. Builder selection — all three prompt builders now assign the default first and override only when the spec provides a builder.
3. Typo detection — dropped difflib.get_close_matches entirely; _warn_for_unknown_fields logs the ignored field names as a simple warning (scoped to free-form model_extra/extra_body sources so explicit pydantic fields are not flagged).
4. from_fields defaults — removed; all three parameters are required and the only caller passes them explicitly.

Validation: 115 targeted tests pass, and pre-commit / build (3.11, 3.12) / DCO are green on the head commit. Note the warning behavior is now a plain name list rather than a "did you mean" suggestion, so the PR description was updated accordingly.

Issue #4676, opened 2026-06-24
[RFC]: MOSS-TTS-Local-v1.5 Serving Optimization


gcanlin:
# RFC: MOSS-TTS-Local-v1.5 Serving Optimization

## Background

MOSS-TTS-Local-v1.5 uses a Global Transformer + Local Transformer architecture. The Global Transformer is a Qwen3-style backbone that models text, reference-audio codes, and previous-frame audio feedback; the Local Transformer expands each global hidden state into one acoustic frame.

Each acoustic frame contains 12 RVQ codebooks, and codebooks inside the frame are generated sequentially. This means every audio frame requires not only one backbone step, but also a local continue/stop decision and 12 codebook sampling steps.

MOSS-Audio-Tokenizer-v2 handles both reference-audio encoding and waveform decoding, targets 48 kHz stereo output, and supports native incremental decoding. This gives MOSS Local high-quality audio and a natural streaming path, but it also makes the serving path heavier than many 24 kHz mono or lightweight-vocoder TTS models.

The current vLLM-Omni MOSS Local path is functionally usable, but it does not yet fully use batching, GPU-resident state, CUDA Graph, tensor payloads, or codec streaming. The goal is to reduce RTF, restore true low-TTFP streaming, and make high-concurrency behavior observable, recoverable, and tunable.

## Goals

- Reduce single-request and concurrent RTF by removing Python hot-path work, CPU syncs, small-kernel launches, and repeated tensor allocations.
- Let Stage 1 consume codec frames before Stage 0 finishes the full utterance, reducing TTFP.
- Preserve offline quality, stop semantics, and voice-clone behavior while adding batched, graphed, and streaming fast paths.
- Keep the optimization path compatible with vLLM-Omni multi-stage serving, benchmark, stats, and the OpenAI-compatible speech API.

### Model Initial Support

- [x] https://github.com/vllm-project/vllm-omni/pull/4664
- [x] https://github.com/vllm-project/vllm-omni/pull/3420

### Reference Audio Preprocessing

- [ ] Add a content-addressed cache for reference-audio codes to avoid repeatedly encoding the same speaker audio. Cache values should be stored on CPU in a compact dtype and returned as clones to avoid accidental mutation by downstream code.

- [ ] Add single-flight for the same uncached reference audio so concurrent requests trigger only one real encoding job. This prevents bursts of voice-clone traffic from overloading the codec encoder with duplicate work.

[...]

PR #4982, opened 2026-07-09
[Feature][Moss-TTS] Reference Audio Preprocessing

BruceLoveDecimal:
## Purpose

This PR implements the reference-audio caching / coalescing design from
**RFC #4676**. It lifts the previous stateless `encode_reference_codes()` helper
into a per-server `MossReferenceEncoder` that adds three layers on top of the
existing speaker cache:

1. **Content-addressed caching** — anonymous references are keyed by the
   resolved-audio content hash (reusing the server's existing
   `_get_resolved_ref_audio_artifact_key`), so the *same clip arriving via
   different URLs/paths* shares one cache entry. Named voices remain keyed by
   name + `created_at` (re-upload invalidates). Cached codes are downcast to
   **int32** for a compact footprint, and every returned tensor is an
   independent int64 copy (no aliasing of the cached buffer).
2. **Single-flight** — concurrent requests for the same uncached reference join
   one in-flight encode (and share the upstream resolve/download) instead of each
   paying the full cost. The shared flight is `asyncio.shield`-ed so a caller
   cancelling its own request cannot tear down the encode for every other waiter,
   with an identity guard on slot cleanup.
3. **Micro-batched encoding** — cold encodes are coalesced within a small time
   window (`10 ms`, `max_batch=8`) into a single batched processor forward on a
   worker thread, with per-clip prep-failure isolation and a per-item fallback if
   the batch forward fails. The `ttsd` two-speaker path now encodes both refs
   concurrently (`asyncio.gather`) so they land in the same batch window.

## A/B benchmark: base vs head, five scenarios

Following up on the review request for a per-mechanism A/B: below is base `7174838` (current main) vs head `c7e9a3a4` (this branch; differs from base by exactly this PR's commits), same machine, same venv, same dataset, same seed, one column per claimed mechanism.

### Environment & method

| Item | Value |
|---|---|
| GPU | NVIDIA H20 (96 GB) |
| Python / PyTorch / vLLM | 3.12 / 2.13.0+cu130 / 0.28.0 (identical venv for both sides; only `PYTHONPATH` switches the vllm-omni tree) |
| Model | MOSS-TTS-Local-Transformer-v1.5 + MOSS-Audio-Tokenizer-v2 (local weights, offline) |
| Dataset | `zhaochenyang20/seed-tts-eval` (en), 100 rows sampled with fixed seed 42 → 98 distinct reference clips |
[...]

PR #5686, opened 2026-08-02
[Perf][TTS] Single-flight reference-audio encoding for MOSS-TTS

skyloevil:
## Purpose

Part of #4676: add single-flight for identical uncached MOSS-TTS reference audio.

The speaker cache is populated only after encoding finishes. Concurrent requests that miss the same cold entry can therefore start duplicate CPU codec work before the first result is stored. This change makes eight concurrent requests sharing one resolved reference perform one codec encode instead of eight.

## What changed

In `reference_encoder.py`:

- Added an in-flight task table keyed by the same `(speaker, model variant/n_vq, created_at)` key as `SpeakerEmbeddingCache`.
- Preserved main's resolver-derived anonymous-reference identity. Anonymous requests resolve first, obtain the resolver's mtime/size content-snapshot key, and only then join or create the encode task for that final key. They intentionally still perform one resolve per request; only the expensive codec encode is collapsed.
- Kept the named-voice fast path: its stable voice/timestamp key is known before resolution, so a cold burst shares resolution and encoding, while a warm hit skips resolution.
- Used `asyncio.shield` so cancellation of one waiter does not cancel the shared work. Completion removes the in-flight entry; failures are not cached and a later request retries.
- Stored a detached CPU tensor and returned a clone to every caller so downstream mutation cannot affect the cache or another response.
- Kept the blocking codec call in `asyncio.to_thread`.

The lookup/create/register sequence contains no `await`; on the API server's single event loop, only one leader can be registered for a cache key.

## Test plan

Added nine CPU-only, weight-free tests in `tests/model_executor/models/moss_tts/test_reference_encoder_single_flight.py` covering:

1. Eight anonymous cold callers: one encode and eight resolver calls.
2. Different resolver keys remain independent.
3. Different locators with the same resolver key collapse.
4. Warm anonymous cache hits skip encoding.
5. A changed anonymous resolver key forces re-encoding.
6. A warm named cache hit skips resolution.
7. A shared encode failure is not cached and retry succeeds.
8. Cancelling one waiter does not cancel the shared named task.
9. Every caller receives independent tensor storage.

The file is marked `core_model`, `cpu`, and `tts`, and is collected by the existing ready/merge command:

```bash
pytest -sv tests/model_executor -m 'core_model and cpu'
```

No pipeline change is required.

## Test result

Validated at rebased head `58a92f655617e2f54f295694fec78bfc02c43208` on `upstream/main` `f7f48929db8dcd60c8ce538f4057bec5a789057c`:

- Focused import-light suite against the real production module: `9 passed` on Python 3.12.9 / Torch 2.3.1 and `9 passed` on Python 3.12.12 / Torch 2.7.0.
- File-scoped `pre-commit`: all applicable hooks passed, including Ruff, formatting, mypy, CI marks, SPDX, forbidden imports, and CUDA API checks.
- Regression proof against the exact base production implementation: the new concurrent-cold test observes eight codec calls and fails its `== 1` assertion; the rebased head passes with one codec call.

The local environments cannot boot the repository-wide pytest configuration because their installed vLLM/pyzmq stack is incompatible, so the focused local run used an import-light pytest harness. The normal marked suite is left to repository CI. No wall-clock hardware benchmark is claimed; the verified performance invariant is the codec-work count reduction from eight to one for the tested cold burst.

### DIFF
--- a/tests/model_executor/models/moss_tts/test_reference_encoder_single_flight.py
+++ b/tests/model_executor/models/moss_tts/test_reference_encoder_single_flight.py
@@ -1,0 +1,273 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM-Omni project
+"""Regression tests for single-flight reference-audio encoding in MOSS-TTS.
+
+Guards the duplicate-encode class from RFC #4676 ("Add single-flight for the
+same uncached reference audio so concurrent requests trigger only one real
+encoding job"). The speaker cache is written only *after* an encode finishes,
+so a burst of voice-clone requests arriving on one cold reference — the common
+shape: one speaker, many utterances — used to run N multi-second CPU codec
+passes before the first one populated the cache.
+
+The invariants under test are the ones that make single-flight safe to put in
+front of a shared cache: exactly one real encode per cold reference, distinct
+references never collapsed, failures never cached (the next request retries),
+one caller giving up never taking the others down, and every caller getting an
+independent tensor rather than a view of the cached one.
+
+CPU-only and weight-free: the MOSS processor and the audio resolver are both
+injected, so nothing here touches the codec, the talker, or torchaudio.
+"""
+
+from __future__ import annotations
+
+import asyncio
+
+import pytest
+
+pytestmark = [pytest.mark.core_model, pytest.mark.cpu, pytest.mark.tts]
+
+
+torch = pytest.importorskip("torch")
+
+SR = 24000
+N_VQ = 4
+
+
+@pytest.fixture(autouse=True)
+def _clear_inflight():
+    """Keep the module-global in-flight table from leaking between tests."""
+    from vllm_omni.model_executor.models.moss_tts import reference_encoder
+
+    reference_encoder._INFLIGHT.clear()
+    yield
+    reference_encoder._INFLIGHT.clear()
+
+
+class _FakeProcessor:
+    """Counts real encodes and returns a code tensor derived from the waveform."""
+
+    def __init__(self):
+        self.calls = 0
+        self.fail_with: Exception | None = None
+
+    def encode_audios_from_wav(self, wavs, sampling_rate, n_vq):
+        self.calls += 1
+        if self.fail_with is not None:
+            raise self.fail_with
+        # First waveform sample doubles as a fingerprint, so a test can prove
+        # which reference a returned tensor actually came from.
+        marker = float(wavs[0].reshape(-1)[0])
+        return [torch.full((n_vq, 3), marker)]
+
+
+class _Gate:
+    """An async ``resolve_ref_audio`` the test can hold open and count."""
+
+    def __init__(self):
+        self.event = asyncio.Event()
+        self.calls = 0
+        self.fail_with: Exception | None = None
+        self.cache_key: str | None = None
+        self.waveform_marker: float | None = None
+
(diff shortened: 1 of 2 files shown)


skyloevil:
@gcanlin  @hsliuustc0106  @linyueqian  PTAL.Thanks~

linyueqian (changes requested):
Reviewed at `8636ecce` (whole footprint). The single-flight implementation is coherent against merge-base bfaaa3c2c: concurrent cold requests share one shielded task, failures retire the slot, and callers receive cloned tensors. However, the PR cannot merge. A static merge-tree confirms a conflict in reference_encoder.py because current main changed anonymous-reference cache identity and the resolver contract in the same function. That conflict must be resolved while preserving main's newer resolver-derived cache key behavior.

The reported test results are also stale for this head. The description identifies 67c54777 as the tested vLLM-Omni commit, which is the parent of the feature commit and predates the later main merge that produced 8636eccea. The inlined CI for the reviewed head covers builds, documentation, and pre-commit, but not the stated pytest suites.

Validation: static reads only. I confirmed the merge conflict independently with `git merge-tree` against current main.

skyloevil:
> Reviewed at `8636ecce` (whole footprint). The single-flight implementation is coherent against merge-base [bfaaa3c](https://github.com/vllm-project/vllm-omni/commit/bfaaa3c2c3fb41f32f3015dc20958cc770eb5f38): concurrent cold requests share one shielded task, failures retire the slot, and callers receive cloned tensors. However, the PR cannot merge. A static merge-tree confirms a conflict in reference_encoder.py because current main changed anonymous-reference cache identity and the resolver contract in the same function. That conflict must be resolved while preserving main's newer resolver-derived cache key behavior.
> 
> The reported test results are also stale for this head. The description identifies [67c5477](https://github.com/vllm-project/vllm-omni/commit/67c54777bb22e9e7e08fdf7c47a64f06b566fc47) as the tested vLLM-Omni commit, which is the parent of the feature commit and predates the later main merge that produced [8636ecc](https://github.com/vllm-project/vllm-omni/commit/8636eccea8fb268b47a24fea9ccbd91f4ebf14ec). The inlined CI for the reviewed head covers builds, documentation, and pre-commit, but not the stated pytest suites.
> 
> Validation: static reads only. I confirmed the merge conflict independently with `git merge-tree` against current main.

Thanks for the detailed review! I’ve noted both the merge conflict and the stale test results. I’m working on the fix now, making sure to preserve the latest resolver-derived cache key behavior from main. I’ll also rerun the relevant pytest suites on the updated head and update the PR accordingly.

skyloevil:
Rebased onto current main and updated the head to 58a92f65. The reference_encoder.py conflict is resolved while preserving main’s resolver-derived anonymous cache identity: anonymous callers resolve independently, then join the codec task using the final resolver key, giving 8 resolves but 1 encode in the concurrent regression test.
The PR now includes 9 focused tests covering resolver-key identity and invalidation, named/anonymous paths, failure retry, cancellation isolation, and tensor ownership. The concurrent test reproduces 8 codec calls against the exact base implementation and passes with 1 on the updated head. 
Could you please re-review the updated head?

skyloevil:
Closing this PR as its primary single-flight functionality is now covered by #4982, which also adds content-addressed caching and micro-batched encoding. Any remaining cross-locator concurrent deduplication improvement can be addressed separately in a focused follow-up. Thanks for the review!

PR #5687, opened 2026-08-02
[Bugfix] Fix Qwen3-Omni AWQ quantization name mapping

### LINKED ISSUE #5573: [Bug]: Qwen3-Omni AWQ (compressed-tensors) thinker produces degenerate text under --omni and in offline end2end; same weights correct without --omni
# [Bug]: Qwen3-Omni AWQ (compressed-tensors) thinker decodes a constant token under --omni; same weights correct without --omni

## Environment

- Image: `vllm/vllm-omni:v0.24.0` (Docker Hub), unmodified
- Model: `cyankiwi/Qwen3-Omni-30B-A3B-Instruct-AWQ-4bit` (compressed-tensors, pack-quantized W4A16 g32; thinker MoE body quantized, everything else BF16/F16 per its `ignore` list)
- GPUs tried: 1× L40S and 1× A100-80GB (identical results)
- Stage config: single-GPU variant of the bundled `qwen3_omni_moe.yaml` (all stages `devices: "0"`, utilizations 0.11/0.05/0.02, stage-0 `max_model_len` 32768). All stages initialize and serve.

## Symptom

Under `vllm serve <model> --omni`, every chat completion returns the same
degenerate text: the token `"0"` repeated to `max_tokens` (`finish_reason:
"length"`), at temperature 0 and at their example's sampling settings alike.
Speech output renders that text — the WAV transcribes as babble.

The **same checkpoint through the same image without `--omni`** (text-only
serving) answers correctly, makes tool calls, and reads OCR — content-verified.

## The prompt is not the problem

With `--enable-log-requests` + `VLLM_LOGGING_LEVEL=DEBUG`, the request logger
shows a byte-perfect rendered prompt at the API boundary:

```
prompt: '<|im_start|>user\nWhat is 7 times 6? Answer in one short sentence.<|im_end|>\n<|im_start|>assistant\n'
prompt_token_ids: [151644, 872, 198, 3838, 374, 220, 22, 3039, 220, 21, 30, 21806, 304, 825, 2805, 11652, 13, 151645, 198, 151644, 77091, 198]
SamplingParams(temperature=0.0, max_tokens=64, ...)
```

Response content: `'0000000000000000000000000000000000000000000000000000000000000000'`.

## Exonerated by bisection (all reproduce the constant token)

- request shape: plain string content, content-parts, and the exact shape in
  `examples/online_serving/qwen3_omni/run_curl_multimodal_generation.sh`
  (system + parts + `sampling_params_list` with `stop_token_ids`)
- `modalities`: absent, `["text"]`, `["text","audio"]`
- `--chat-template` passed explicitly vs omitted (the repo's jinja is
  byte-identical to the official `chat_template.json`)
- `--enable-auto-tool-choice --tool-call-parser hermes` on vs off
- torch.compile cache: cold and warm, and across node generations
- GPU: L40S (sm89) vs A100-80 (sm80)
- checkpoint aux files: config.json (minus quant block), tokenizer files,
  templates, `processor_class` — all verified identical to
  `Qwen/Qwen3-Omni-30B-A3B-Instruct`

One checkpoint-side fix was needed just to load: the repo's quantization
`ignore` list is spelled in HF names (`talker.model.*`), while the talker
stage renames modules to `talker.language_model.model.*` before
compressed-tensors ignore matching runs — without adding
`re:talker\..*`/`re:code2wav\..*` to `ignore`, stage 1 dies with
`KeyError: 'layers.0.mlp.experts.routed_experts.w2_weight'`. That may be
worth a fix of its own (match ignore against both spellings, as
a6024209 did for the thinker lm_head prefix).

## Offline result — reproduces in your own example

`examples/offline_inference/qwen3_omni/end2end.py --query-type text`
(byte-exact from the v0.24.0 tag), same checkpoint, same stage config:

```
Prompt: <|im_start|>system\nYou are Qwen, ... <|im_start|>assistant\n
vllm_text_output:
orm党的tasmofnow0mac·iess.
ly antele.aspx
```

Incoherent token soup — the same broken logits the online path shows; greedy
decoding just collapses it to the constant token. So this is not the serving
layer: the omni engine path produces degenerate thinker output for this
checkpoint in both the offline and online lanes.

## Scope notes (untested dimensions)

- BF16 official checkpoint under `--omni` was NOT tested (cost); we cannot say
  whether the bug is quant-specific or affects all checkpoints.
- The bundled 2-GPU topology was NOT tested; all repro runs place the three
  stages on one device (single-GPU stage config attached below). If stage
  co-residency is load-bearing here, that would be useful to know — nothing in
  the docs marks it unsupported.

## Why we think it's the omni engine loop

The prompt is correct at the API layer and the weights are proven by the
non-omni path, so what remains is what `--omni` changes for stage 0: the
OmniARAsyncScheduler / async omni output path, hidden-state capture for the
talker connector, and the stage engine-core handoff. The output being a
CONSTANT low-id token (id 15 = "0") from step 1, deterministic across
hardware, looks like sampling from a wrong/cleared buffer rather than a bad
prompt or bad weights.

## Single-GPU stage config used (replaces vllm_omni/deploy/qwen3_omni_moe.yaml)

```yaml
async_chunk: true
connectors:
  connector_of_shared_memory:
    name: SharedMemoryConnector
    extra:
      initial_codec_chunk_frames: 4
      codec_chunk_frames: 25
      codec_left_context_frames: 25
stages:
  - stage_id: 0
    max_num_batched_tokens: 32768
    max_num_seqs: 64
    max_model_len: 32768
    gpu_memory_utilization: 0.11
    trust_remote_code: true
    enable_prefix_caching: false
    devices: "0"
    default_sampling_params: {temperature: 0.0, max_tokens: 2048}
  - stage_id: 1
    max_num_batched_tokens: 32768
    max_num_seqs: 64
    gpu_memory_utilization: 0.05
    trust_remote_code: true
    enable_prefix_caching: false
    devices: "0"
    input_connectors: {from_stage_0: connector_of_shared_memory}
    default_sampling_params: {temperature: 0.9, top_k: 50, max_tokens: 4096, repetition_penalty: 1.05}
  - stage_id: 2
    max_num_batched_tokens: 65536
    max_num_seqs: 64
    gpu_memory_utilization: 0.02
    enforce_eager: false
    trust_remote_code: true
    enable_prefix_caching: false
    enable_chunked_prefill: false
    async_scheduling: false
    devices: "0"
    input_connectors: {from_stage_1: connector_of_shared_memory}
    default_sampling_params: {temperature: 0.0, top_p: 1.0, top_k: -1, max_tokens: 65536, repetition_penalty: 1.1}
```

lishunyang12:
## Purpose

Fixes #5573.

Qwen3-Omni compressed-tensors checkpoints store their `ignore` entries with Hugging Face names such as `thinker.model.layers.0.mlp.gate`. The unified Omni wrapper inserts an additional `language_model` module, but quantization method selection previously ran before that outer rename was represented in the quantization config.

On v0.24.0 this could silently construct the BF16 MoE routers as zero-initialized quantized layers and produce degenerate repeated tokens. On current `main`, the same mismatch fails during weight loading with:

```text
ValueError: There is no module or parameter named 'layers.0.mlp.gate.weight' in Qwen3MoeLLMModel.
AttributeError: 'RoutedExperts' object has no attribute 'w2_weight'
```

This PR:

- exposes the final thinker/talker module paths on the unified Qwen3-Omni wrapper;
- applies quantization-config name mapping once at that outer boundary;
- keeps the nested stage models from applying their weight mappers a second time, while preserving packed-module metadata.

The change only runs during model construction; it does not add work to the inference hot path or change quantization kernels.

cc @Gaohan123 @YIHONG-JIN @shinyobjectz

## Test Plan

**Hardware:** 2x NVIDIA B300 SXM6 AC (275040 MiB each)

**Driver / CUDA:** 610.43.02 / CUDA 13.0

**PyTorch / Transformers:** 2.11.0+cu130 / 5.14.1

**vLLM Version:** 0.26.0

**vLLM-Omni Commit:** 67c54777

Reproduce and validate with the checkpoint from #5573:

```bash
PYTHONPATH=$PWD \
CUDA_VISIBLE_DEVICES=0,1 \
TOKENIZERS_PARALLELISM=false \
python -u examples/offline_inference/qwen3_omni/end2end.py \
  --model cyankiwi/Qwen3-Omni-30B-A3B-Instruct-AWQ-4bit \
  --query-type text \
  --modalities text \
  --num-prompts 1 \
  --stage-init-timeout 600 \
  --init-timeout 900 \
  --output-wav /tmp/qwen3-omni-awq-output
```

The increased initialization limits allow a cold B300 FlashInfer MoE JIT build to finish; cached runs initialize faster.

Run repository checks for every changed file:

```bash
pre-commit run --files \
  vllm_omni/model_executor/models/qwen3_omni/qwen3_omni.py \
  vllm_omni/model_executor/models/qwen3_omni/qwen3_omni_code2wav.py \
  vllm_omni/model_executor/models/qwen3_omni/qwen3_omni_moe_talker.py \
  vllm_omni/model_executor/models/qwen3_omni/qwen3_omni_moe_thinker.py \
  vllm_omni/model_executor/models/qwen3_omni/quantization.py
```

## Test Result

- Current `main`: thinker and talker fail with the two weight-loading errors above.
- This PR: thinker, talker, and code2wav load 1567, 339, and 230 weights respectively; all three stages become ready.
- The same text request completes successfully with coherent output:

```text
Distributed system with microservices, queuing, parallel processing, GPU clusters, and auto-scaling for high-throughput, low-latency audio generation.
```

All changed-file pre-commit hooks pass.


vllm_omni/model_executor/models/qwen3_omni/quantization.py
```diff
@@ -0,0 +1,36 @@
...
+from vllm.model_executor.models.interfaces import SupportsQuant
+
+_OUTER_MAPPING_APPLIED = "_qwen3_omni_outer_mapping_applied"
+
+
+def apply_outer_quant_config_mapping(model: SupportsQuant) -> None:
+    """Map checkpoint paths once at the unified model boundary."""
+    SupportsQuant._maybe_apply_model_mapping(model)
+    if model.quant_config is not None:
+        setattr(model.quant_config, _OUTER_MAPPING_APPLIED, True)
```
Gaohan123, line 14: [P2] Do not mark component configs as mapped before mapping their children

When `model.quant_config` is a `ComponentQuantizationConfig`, the call above does not propagate the mapper to its `_components` or `_default` configs because `ComponentQuantizationConfig` does not override `apply_vllm_mapper()` and the inherited base implementation is a no-op. Nevertheless, this line marks the wrapper as already mapped, so nested thinker/talker models skip their own mapping. After `resolve("thinker")` or `resolve("talker")`, the underlying compressed-tensors config therefore still contains HF names such as `talker.model.*`, which can reproduce the incorrect quantization or weight-loading failure this PR is intended to fix.

Please recursively apply the mapper to every non-None component/default config before setting this marker (preferably by implementing `ComponentQuantizationConfig.apply_vllm_mapper()`), and add a regression test covering a component-wrapped compressed-tensors config.

david6666666:
Addressed the component-config review feedback in `49b7a36c8`:

- Added `ComponentQuantizationConfig.apply_vllm_mapper()` so the mapper is applied recursively to every non-`None` component and the default config before the Qwen3-Omni outer-mapping marker is set.
- Added a regression test with real component-wrapped `CompressedTensorsConfig` instances, covering component/default `ignore` remapping and a `None` component.

Validation:

- `pytest -q tests/diffusion/quantization/test_component_routing.py` — **40 passed**.
- `pre-commit run --files <all 7 changed files>` — **all hooks passed**.
- Qwen3-Omni AWQ end-to-end on **2x NVIDIA B300 SXM6 AC**, driver **610.43.02 / CUDA 13.0**, PyTorch **2.11.0+cu130**, Transformers **5.14.1**, vLLM **0.26.0**.

For the end-to-end test I used the same checkpoint and request as the PR description, with `--stage-init-timeout 1200 --init-timeout 1800`. On the first cold-cache run, the original 900-second global timeout fired while FlashInfer was still compiling the SM103 fused-MoE kernel; that compile completed about 43 seconds later. There were no quantization mapping or weight-loading errors. The cached rerun completed successfully with exit code 0:

- thinker: **1567 weights**
- talker: **339 weights**, using the unquantized BF16 MoE backend
- code2wav: **230 weights**
- all three stages ready; engine initialized in **222.77s**

Generated output:

```text
Distributed system with microservices, queuing, parallel processing, GPU clusters, and auto-scaling for high-throughput, low-latency audio generation.
```

CI note: [ReadTheDocs build #34164407](https://app.readthedocs.org/projects/vllm-omni/builds/34164407/) failed before installing this project. Its seeded pip 23.1 rejected the `typing-extensions` wheel metadata and then could not find `flit_core` on the PyTorch CPU-only index. None of the changed project files had been executed, so this is unrelated to the patch.

Gaohan123 (approved):
LGTM. Thanks

PR #3799, opened 2026-05-21
[Bugfix] Fix audio_ttfp goodput and audio_duration help

Ronnie-Rui:
## Purpose

This PR fixes two small benchmark metric inconsistencies.

First, the goodput calculation used `audio_ttft` for audio first-packet latency, while the benchmark metric is documented and reported as `audio_ttfp`. This makes `audio_ttfp` goodput SLOs silently ignored.

Second, `audio_duration` is already supported as a percentile metric, but it was missing from the `--percentile-metrics` help text for `vllm bench serve`.

PR #5689, opened 2026-08-02
[Bugfix] Enable audio TTFP goodput SLOs

### LINKED ISSUE #5668: [Bug]: audio_ttfp goodput SLO is rejected by the CLI and not applied by Omni benchmark metrics
### Your current environment

<details>

```text
CI env
```

</details>

### Your code version

<details>

```text
0.26.0
```
</details>
<details>

```text
185e3d6df60276166349e69fdc1171fd0e50c79f
```
</details>

### 🐛 Describe the bug

The Omni serving benchmark reports `audio_ttfp` as an audio latency metric, but users cannot use it as a request-level goodput SLO through the normal CLI.

There are two separate problems in the current architecture.

#### 1. The CLI rejects `audio_ttfp`

The vLLM-Omni benchmark CLI delegates its base argument registration and goodput validation to upstream vLLM.

The upstream `check_goodput_args()` currently accepts only:

```python
["ttft", "tpot", "e2el"]
```

Therefore, a command containing the following option is rejected before the Omni metric calculation is reached:

```bash
--goodput audio_ttfp:500
```

#### 2. The Omni calculation checks the wrong metric name

The Omni goodput calculation checks the misspelled key `audio_ttft` instead of the canonical metric name `audio_ttfp`:

```python
if "audio_ttft" in goodput_config_dict:
    valid_metrics.append(audio_ttfps)
    slo_values.append(
        goodput_config_dict["audio_ttft"]
        / MILLISECONDS_TO_SECONDS_CONVERSION
    )
```

Relevant code:

https://github.com/vllm-project/vllm-omni/blob/185e3d6df60276166349e69fdc1171fd0e50c79f/vllm_omni/benchmarks/metrics/metrics.py#L791-L810

The canonical metric name is defined as:

```python
AUDIO_TTFP = "audio_ttfp"
```

Relevant definition:

https://github.com/vllm-project/vllm-omni/blob/185e3d6df60276166349e69fdc1171fd0e50c79f/vllm_omni/metrics/definitions.py#L29-L42

#### Normal CLI path

Running an Omni benchmark with:

```bash
vllm bench serve --omni \
    ... \
    --goodput audio_ttfp:500
```

fails during goodput argument validation because `audio_ttfp` is not included in the allowed goodput metric names.

#### Programmatic calculation path

If CLI validation is bypassed and `calculate_metrics()` receives:

```python
goodput_config_dict = {
    "audio_ttfp": 500,
}
```

the configured threshold is not added to `valid_metrics`, because the implementation checks for `audio_ttft`.

This produces two incorrect cases:

1. If `audio_ttfp` is combined with another recognized SLO, such as `ttft`, only the recognized SLO affects `request_goodput`. The reported goodput may therefore be higher than the requested combined SLO permits.

2. If `audio_ttfp` is the only programmatically supplied SLO, no metric is added to `valid_metrics`. The resulting goodput remains zero instead of being calculated from the audio threshold.

The legacy spelling `audio_ttft` is not a usable workaround because the normal CLI rejects it as well.

### Reproduction

Start an Omni-compatible serving endpoint, then run a benchmark with an audio TTFP goodput SLO:

```bash
vllm bench serve --omni \
    --backend openai-chat-omni \
    --model <model> \
    --endpoint /v1/chat/completions \
    --dataset-name <audio-output-dataset> \
    --num-prompts 2 \
    --goodput audio_ttfp:500
```

Observe that the benchmark rejects `audio_ttfp` as an invalid goodput metric before requests are benchmarked.

The calculation-layer problem can also be reproduced with a unit test containing two successful outputs:

```python
fast_output.audio_ttfp = 0.2
slow_output.audio_ttfp = 0.8

metrics, _ = calculate_metrics(
    ...,
    outputs=[fast_output, slow_output],
    dur_s=1.0,
    goodput_config_dict={"audio_ttfp": 500},
)
```

With a threshold of 500 milliseconds:

- `fast_output` should satisfy the SLO;
- `slow_output` should fail the SLO;
- `metrics.request_goodput` should therefore be `1.0` request/second.

The current implementation does not perform this comparison.

### Expected behavior

The following option should be accepted by the Omni benchmark CLI:

```bash
--goodput audio_ttfp:500
```

For each successful request:

- the recorded `audio_ttfp` value should be compared with the configured threshold;
- the configured value should be interpreted in milliseconds, consistently with the existing goodput interface;
- the request should count toward goodput only if all configured request-level SLOs are satisfied.

For example, with an `audio_ttfp` SLO of 500 milliseconds:

- a request with `audio_ttfp = 0.2` seconds should pass;
- a request with `audio_ttfp = 0.8` seconds should fail.

When multiple SLOs are supplied, such as:

```bash
--goodput ttft:1000 audio_ttfp:500
```

a request should count toward goodput only if both thresholds are satisfied.

### Why this remains relevant under the latest architecture

The benchmark CLI has been refactored to reuse upstream vLLM argument registration and validation, while Omni-specific metric calculation remains in `vllm_omni`.

Consequently, replacing `audio_ttft` with `audio_ttfp` in `metrics.py` alone is no longer sufficient.

A complete fix must address both:

1. the upstream-derived CLI validation path; and
2. the Omni-specific goodput calculation.

### Proposed fix

1. Add an Omni-specific goodput validation path that accepts `audio_ttfp` while preserving the upstream `ttft`, `tpot`, and `e2el` metrics.

2. Replace the hard-coded `audio_ttft` key in the metric calculation with the canonical definition:

   ```python
   defs.AUDIO_TTFP
   ```

3. Update the `--goodput` help text and benchmark documentation to list `audio_ttfp` as a supported request-level SLO.

4. Validate programmatically supplied goodput keys explicitly so that unsupported or misspelled keys do not silently produce an empty metric set.

5. Add regression tests for both CLI validation and goodput calculation.

### Acceptance criteria

- [ ] `vllm bench serve --omni --goodput audio_ttfp:<milliseconds>` is accepted.
- [ ] The `audio_ttfp` threshold is applied to request-level goodput calculation.
- [ ] A request above the configured `audio_ttfp` threshold is excluded from goodput.
- [ ] A request below the configured threshold is included in goodput.
- [ ] Combined SLOs such as `ttft` and `audio_ttfp` require both conditions to pass.
- [ ] Existing `ttft`, `tpot`, and `e2el` behavior remains unchanged.
- [ ] CLI help and documentation list `audio_ttfp` as a supported goodput metric.
- [ ] Unsupported programmatic goodput keys produce a clear validation error.
- [ ] Regression tests cover both the normal CLI path and direct metric calculation.

- [x] I searched for related issues.
- [x] I verified the behavior against the latest `main` branch.

16:52

lishunyang12:
## Purpose

Fixes #5668.

`vllm bench serve --omni --goodput audio_ttfp:500` was unusable for two independent reasons:

1. vLLM-Omni reuses vLLM's `check_goodput_args()`, which only accepts `ttft`, `tpot`, and `e2el`, so the CLI rejected `audio_ttfp` before the benchmark started.
2. Omni's metric calculation checked the misspelled key `audio_ttft`, so a directly supplied `audio_ttfp` SLO was silently ignored.

This PR adds an Omni goodput validator that preserves the existing vLLM SLOs and accepts the canonical `audio_ttfp` metric. The same validator is used by direct metric calculation, so unsupported keys now fail clearly. It also applies the audio threshold using `defs.AUDIO_TTFP` and documents the new CLI option. Multiple SLOs keep the existing AND semantics: every configured threshold must pass for a request to contribute to goodput.

This completes the intent of #3799 for the current benchmark architecture by also covering the upstream-derived CLI validation path and direct-config validation.

cc @Gaohan123 @ZacheryAU @Ronnie-Rui

## Test Plan

**vLLM Version:** 0.26.0

**vLLM-Omni Commit:** 8e92d464

Reproduce the CLI validation path without launching a server:

```bash
PYTHONPATH=. python - <<'PY'
from argparse import Namespace

import vllm_omni.benchmarks.patch.patch  # noqa: F401
from vllm.benchmarks import serve

print(serve.check_goodput_args(Namespace(goodput=["audio_ttfp:500"])))
PY
```

On `main`, this raises `ValueError: Invalid metric name found, audio_ttfp: 500.0`. With this PR, it prints:

```text
{'audio_ttfp': 500.0}
```

Run the focused CLI and metric regression tests:

```bash
PYTHONPATH=. python -m pytest -q \
  tests/benchmarks/test_serve_cli.py \
  tests/benchmarks/metrics/test_metrics.py
```

### DIFF
--- a/vllm_omni/benchmarks/metrics/metrics.py
+++ b/vllm_omni/benchmarks/metrics/metrics.py
@@ -13,6 +13,23 @@ from vllm_omni.metrics import definitions as defs
 _PERCENTILE_ROWS_TYPE = list[tuple[float, float]] | None
 _FLOAT_LIST_TYPE = list[float]
 _INT_LIST_TYPE = list[int]
+VALID_GOODPUT_METRICS = ("ttft", "tpot", "e2el", defs.AUDIO_TTFP)
+
+
+def validate_goodput_config(goodput_config_dict: dict[str, float]) -> None:
+    """Validate request-level goodput SLO names and values."""
+    for slo_name, slo_val in goodput_config_dict.items():
+        if slo_name not in VALID_GOODPUT_METRICS:
+            raise ValueError(
+                f"Invalid metric name found, {slo_name}: {slo_val}. "
+                "The service level objective name should be one of "
+                f"{list(VALID_GOODPUT_METRICS)}. "
+            )
+        if slo_val < 0:
+            raise ValueError(
+                f"Invalid value found, {slo_name}: {slo_val}. The service level objective value should be non-negative."
+            )
+
 
 _MULTIMODAL_BENCHMARK_FIELDS = [
     (defs.MEAN_AUDIO_TTFP_MS, float, field(default=0.0)),
@@ -712,6 +729,8 @@ def calculate_metrics(
     Returns:
         A tuple of the benchmark metrics and the actual output lengths.
     """
+    validate_goodput_config(goodput_config_dict)
+
     actual_output_lens: list[int] = []
     total_input = 0
     completed = 0
@@ -795,9 +814,9 @@ def calculate_metrics(
         if "ttft" in goodput_config_dict:
             valid_metrics.append(ttfts)
             slo_values.append(goodput_config_dict["ttft"] / MILLISECONDS_TO_SECONDS_CONVERSION)
-        if "audio_ttft" in goodput_config_dict:
+        if defs.AUDIO_TTFP in goodput_config_dict:
             valid_metrics.append(audio_ttfps)
-            slo_values.append(goodput_config_dict["audio_ttft"] / MILLISECONDS_TO_SECONDS_CONVERSION)
+            slo_values.append(goodput_config_dict[defs.AUDIO_TTFP] / MILLISECONDS_TO_SECONDS_CONVERSION)
         if "tpot" in goodput_config_dict:
             valid_metrics.append(all_tpots)
             slo_values.append(goodput_config_dict["tpot"] / MILLISECONDS_TO_SECONDS_CONVERSION)
--- a/docs/cli/bench/serve.md
+++ b/docs/cli/bench/serve.md
@@ -55,6 +55,12 @@ You can use `vllm bench serve --omni --help=all` to get descriptions of all para
                     'Allowed metric names are "ttft", "tpot", "itl", "ttfc", "tpoc", "icl", '
                     '"tpop", "e2el", "audio_ttfp", "audio_rtf", "audio_duration". '
 
+- `--goodput`
+  Specify request-level service objectives as space-separated `KEY:VALUE` pairs, with values in milliseconds.
+  Supported metrics are `ttft`, `tpot`, `e2el`, and `audio_ttfp`. When multiple objectives are supplied, a
+  request contributes to goodput only when it satisfies every objective, for example
+  `--goodput ttft:500 audio_ttfp:1000`.
+
 - `--print-stage`
 Print per-stage benchmark metrics for --omni serving when stage metrics are returned by the server. Disabled by default.
 
--- a/tests/benchmarks/metrics/test_metrics.py
+++ b/tests/benchmarks/metrics/test_metrics.py
@@ -66,6 +66,63 @@ def _make_output(prompt_len: int, output_tokens: int = 10) -> MixRequestFuncOutp
     return output
 
 
+def test_audio_ttfp_goodput_and_combined_slos() -> None:
+    fast = _make_output(100)
+    fast.audio_ttfp = 0.2
+    slow_audio = _make_output(100)
+    slow_audio.audio_ttfp = 0.8
+
+    audio_metrics, _ = calculate_metrics(
+        input_requests=[],
+        outputs=[fast, slow_audio],
+        dur_s=1.0,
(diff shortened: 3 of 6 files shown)


vllm_omni/benchmarks/metrics/metrics.py
```diff
@@ -795,9 +814,9 @@ def calculate_metrics(
         if "ttft" in goodput_config_dict:
             valid_metrics.append(ttfts)
             slo_values.append(goodput_config_dict["ttft"] / MILLISECONDS_TO_SECONDS_CONVERSION)
-        if "audio_ttft" in goodput_config_dict:
+        if defs.AUDIO_TTFP in goodput_config_dict:
```
hsliuustc0106, line 817: [P1] Treat missing audio TTFP as failing this SLO

`MixRequestFuncOutput.audio_ttfp` defaults to `0.0`, so text-only (or successful audio responses with no packet) requests are appended as zero and pass `audio_ttfp:500` (`0.0 <= 0.5`). In mixed benchmarks this overstates goodput. Please track whether audio TTFP is available and make missing values fail, or reject this SLO for non-audio requests.

lishunyang12:
Status update (2026-08-23): The PR has one unresolved P1 review finding: missing audio TTFP currently defaults to 0.0 and can incorrectly pass the SLO. Moving back to draft until missing-value semantics and a regression test are added.

PR #5704, opened 2026-08-03
[Bugfix][CosyVoice3] Fix DiT streaming chunk attention mask

BeckYang26:
## Purpose

CosyVoice3 streaming code2wav requires chunk attention masks so each frame only attends within its allowed time window. The current `cosyvoice3_dit.py` uses `DiffusionAttention`, which does not pass the mask into softmax—it only zeroes outputs afterward via `masked_fill`. That breaks streaming semantics.

This PR:
- Switches `DiTAttention` to `F.scaled_dot_product_attention(..., attn_mask=...)` so masks take effect before softmax
- Adds `subsequent_chunk_mask` / `add_optional_chunk_mask` (aligned with CosyVoice3_split)
- Forwards `streaming` from CFM to the PyTorch DiT estimator (TRT path unchanged)

## Test Plan

- [x] Tests covering the three changed files: `DiTAttention`, `DiTBlock`, `DiT`, `subsequent_chunk_mask` / `add_optional_chunk_mask`, and CFM `streaming` passthrough
- [x] `pre-commit run --files` on the three changed files

**vLLM Version:** N/A (CosyVoice3 DiT mask fix only)

**vLLM-Omni Commit:** 67c54777 (base main)

## Test Result

**Environment:** NVIDIA GeForce RTX 4090, Python 3.12.13, torch 2.11.0+cu130, vllm 0.22.0, diffusers 0.38.0

**Tests for changed code:** 24 passed, 0 failed

Covers `test_cosyvoice3_components.py` (`TestDiTAttention`, `TestDiTBlock`, `TestDiT`) and `test_cosyvoice3_utils.py` (padding / chunk-mask helpers), with PR sources overlaid on the test runtime.

**pre-commit (Windows, 3 changed files):** Passed (ruff check/format, typos, whitespace)

<details>
<summary>Full-suite runs (failures are outside this PR's diff)</summary>

We also ran the broader CosyVoice3 suites on the same machine. This PR only touches `cosyvoice3_dit.py`, `utils.py` (chunk-mask helpers), and `cfm.py` (streaming flag passthrough). The failures below are in other modules and reproduce without these changes:

| Test | Failure | Why unrelated |
|------|---------|---------------|
| `test_forward_reuses_streaming_cache_state_between_chunks` | `AttributeError: CosyVoice3Model has no attribute '_stage1_chunk_counter'` | Bug/missing field in `cosyvoice3.py` talker→code2wav path; file not modified |
| `test_forward_clears_streaming_cache_on_terminal_chunk` | Same `_stage1_chunk_counter` error | Same as above |
| `test_sample_uses_ras_rejection_for_recent_repetition` | RAS sampler returned `[[1]]` instead of expected `[[2]]` | Talker RAS sampling logic; no sampler changes in this PR |
| `test_talker2code2wav_async_chunk_uses_initial_codec_chunk_frames_for_first_emit` | First async-chunk payload not `None` as test expects | Async-chunk framing in `stage_input_processors/cosyvoice3.py`; not modified (this PR only forwards `streaming` inside CFM→DiT) |

</details>

### DIFF
--- a/vllm_omni/model_executor/models/cosyvoice3/utils.py
+++ b/vllm_omni/model_executor/models/cosyvoice3/utils.py
@@ -273,3 +273,77 @@ def make_pad_mask(lengths: torch.Tensor, max_len: int = 0) -> torch.Tensor:
     seq_length_expand = lengths.unsqueeze(-1)
     mask = seq_range_expand >= seq_length_expand
     return mask
+
+
+def subsequent_chunk_mask(
+    size: int,
+    chunk_size: int,
+    num_left_chunks: int = -1,
+    device: torch.device = torch.device("cpu"),
+) -> torch.Tensor:
+    """Create chunk-wise causal mask ``(size, size)`` for streaming DiT/encoder.
+
+    Aligned with CosyVoice3_split ``cosyvoice/utils/mask.py``.
+    ``True`` means the query position may attend to that key position.
+
+    Example (size=4, chunk_size=2)::
+
+        [[1, 1, 0, 0],
+         [1, 1, 0, 0],
+         [1, 1, 1, 1],
+         [1, 1, 1, 1]]
+    """
+    del num_left_chunks  # CosyVoice ONNX-friendly impl does not use left-chunk limit.
+    pos_idx = torch.arange(size, device=device)
+    block_value = (torch.div(pos_idx, chunk_size, rounding_mode="trunc") + 1) * chunk_size
+    return pos_idx.unsqueeze(0) < block_value.unsqueeze(1)
+
+
+def add_optional_chunk_mask(
+    xs: torch.Tensor,
+    masks: torch.Tensor,
+    use_dynamic_chunk: bool,
+    use_dynamic_left_chunk: bool,
+    decoding_chunk_size: int,
+    static_chunk_size: int,
+    num_decoding_left_chunks: int,
+    enable_full_context: bool = True,
+) -> torch.Tensor:
+    """Apply optional chunk / pad mask for CosyVoice DiT attention.
+
+    Aligned with CosyVoice3_split ``add_optional_chunk_mask``.
+    Inference path used by CosyVoice3 DiT:
+
+    - ``streaming=True`` → ``static_chunk_size > 0`` chunk mask
+    - ``streaming=False`` → ``static_chunk_size=0`` → pad mask only
+    """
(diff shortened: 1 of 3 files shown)


linyueqian (changes requested):
Reviewed at `c3d06236`. Moving the mask into attention is the right core fix, but the default accelerated path still bypasses it and the fallback changes backend behavior without committed coverage. I am requesting changes.

The existing GitHub checks are green, but this PR has no per-PR Buildkite result.

vllm_omni/model_executor/models/cosyvoice3/utils.py
```diff
@@ -273,3 +273,77 @@ def make_pad_mask(lengths: torch.Tensor, max_len: int = 0) -> torch.Tensor:
...
+        chunk_masks = chunk_masks.unsqueeze(0)
+        chunk_masks = masks & chunk_masks
+    elif static_chunk_size > 0:
+        chunk_masks = subsequent_chunk_mask(xs.size(1), static_chunk_size, num_decoding_left_chunks, xs.device)
+        chunk_masks = chunk_masks.unsqueeze(0)
+        chunk_masks = masks & chunk_masks
+    else:
+        chunk_masks = masks
+    assert chunk_masks.dtype == torch.bool
+    if (chunk_masks.sum(dim=-1) == 0).sum().item() != 0:
```
lishunyang12: Use torch.where for empty rows; this .item() synchronizes every diffusion step.

BeckYang26:
> Reviewed at `c3d06236`. Moving the mask into attention is the right core fix, but the default accelerated path still bypasses it and the fallback changes backend behavior without committed coverage. I am requesting changes.
> 
> The existing GitHub checks are green, but this PR has no per-PR Buildkite result.

Addressed in 607df1fb:

1. Streaming chunk maps now go through MATH SDPA, with `attn_mask` applied before softmax.
2. 2D padding still uses the default DiffusionAttention path, so FLASH is unchanged for the non-streaming case.
3. I ran the new L1 CPU tests locally: 8 passed (pre-softmax chunk constraint, MATH pin, and CFM `streaming` passthrough).

vllm_omni/model_executor/models/cosyvoice3/utils.py
BeckYang26: > Use torch.where for empty rows; this .item() synchronizes every diffusion step.

Addressed in 607df1fb. Empty all-false rows are now filled with `torch.where`, so `.item()` is no longer used.

PR #5699, opened 2026-08-03
[Model] [Ascend] Add soundfile fallback for MiniMax H3 audio loading for NPU support

brandneway:
### 1. Soundfile fallback for MiniMax H3 audio loading

torchaudio 2.6+ routes `load()` through TorchCodec, whose aarch64 wheels are
built against a CUDA torch and fail to load (`libnvrtc.so.13` missing) under a
CPU-only torch (e.g. Ascend/NPU environments). This breaks MiniMax H3 ref2va
video+audio input with:

    RuntimeError: TorchCodec is required for load_with_torchcodec

This PR routes both audio load sites (`reference_video.load_video_audio` and
pipeline `_load_audio`) through a new `load_audio_file()` helper that tries
torchaudio first, then falls back to soundfile, then to an ffmpeg-demux
fallback — so wav/mp3/m4a/mp4 inputs all work at native sample rate without
requiring TorchCodec.

Behavior is unchanged on platforms where torchaudio/TorchCodec works; the
fallback chain only activates when the primary path raises.

### 2. Ascend NPU recipe for MiniMax H3

Adds `recipes/MiniMaxAI/MiniMax-H3-NPU.md`, an Ascend NPU deployment guide
adapted from the GPU recipe (`recipes/MiniMaxAI/MiniMax-H3.md`):

- Environment: Atlas 800I A3, aarch64 CPU-only PyTorch 2.10.0 + torch_npu
  2.10.0.post2, CANN 9.0.1; no CUDA runtime required
- Validated multi-NPU configuration: 8x NPU with Ulysses 8, text-encoder
  TP 8, native tiled VAE patch parallelism 8, and layerwise offload
- Documents that audio inputs work without TorchCodec via the fallback
  introduced in this PR
- Reuses the GPU recipe's HTTP API examples and parameter tables by
  reference to avoid duplicated content

## Test Plan

**vLLM Version:** 0.26.0

**vLLM-Omni Commit:** &lt;推送后的最新 commit sha&gt;

1. On an Atlas 800I A3 (8x NPU) with CPU-only torch (no TorchCodec), follow
   the new NPU recipe to serve the `FL2VA` and `Ref2VA` partitions and run
   T2VA and Ref2VA (prompt + video) requests at 1344x768.
2. Exercise the audio fallback chain with wav / mp3 / m4a inputs and with a
   reference video carrying an embedded audio track (mp4).
3. Regression: on a CUDA torch environment where TorchCodec loads normally,
[...]

PR #5706, opened 2026-08-03
[NPU][Quantization] Add RainFusion attention and INT8 online quantization for MiniMax H3

Huangzjun:
## Purpose

Follow-up to #5699 — this PR adds two Ascend NPU performance optimizations for MiniMax H3 DiT inference.

### 1. RainFusion block-sparse attention

Adds the `RAINFUSION_ATTN` backend for MiniMax H3 using MindIE-SD's `rf_v2` block-sparse kernel.

Only the video segment of the packed multimodal sequence is sparsified. Prefix tokens, including text, visual conditions, and audio, remain dense.

The backend automatically falls back to `FLASH_ATTN` during warmup steps, on skipped layers, for short video sequences, when valid video geometry is unavailable, or when the video segment is not a multiple of 128 rows (unaligned resolutions still run, just without a sparsity speedup). MiniMax H3 publishes the required geometry through `AttentionMetadata.video_layout` (`VideoTokenLayout`: `prefix_len` + `latent_grid`).

On top of the 8×NPU serve command in `recipes/MiniMaxAI/MiniMax-H3-NPU.md`, replace the attention backend line with either:

```bash
--diffusion-attention-backend RAINFUSION_ATTN
```

(defaults: `sparsity=0.8`, `start_step=0`), or for explicit tuning:

```bash
--diffusion-attention-config '{"default":{"backend":"RAINFUSION_ATTN","block_sparse":{"sparsity":0.8,"start_step":12}}}'
```

Sparsity requires `latent_t × (height/32) × (width/32) % 128 == 0` (e.g. 1536×768 sparse, 1344×768 dense fallback at 209 frames). Logs report `RAINFUSION_ATTN active` vs `RAINFUSION_ATTN staying dense`. See `docs/user_guide/diffusion/attention_backends.md` for the full `block_sparse` reference.

### 2. INT8 online dynamic quantization

Enables the existing `--diffusion-quantization-config int8` path for the MiniMax H3 DiT transformer.

Weights are loaded from the original BF16 checkpoint and quantized layer by layer at load time, so no pre-quantized checkpoint is required. The text encoder and VAEs remain unquantized.

This PR also:

* Falls back to unquantized execution for projections exceeding the `QuantBatchMatmulV3` output-dimension limit (checked in the NPU INT8 quant method using per-partition shapes).
* Constructs VAE components on CPU to avoid CPU-only operations running under the NPU default-device context.
* Moves quantized layers back to CPU immediately after loading when layerwise offload is enabled, via an explicit quant-method capability, reducing peak NPU memory usage.

On top of the same recipe serve command, add:

```bash
--diffusion-quantization-config int8 \
--enable-layerwise-offload
```

RainFusion and INT8 may be enabled together by combining the attention change from §1 with these two flags.

## Test Plan

**vLLM Version:** `0.26.0`

**vLLM-Omni Commit:** `a20b900d`

1. Run MiniMax H3 T2VA at 1536×768 with `RAINFUSION_ATTN` (sparse path) and at 1344×768 (dense fallback when video rows are not block-aligned).
2. Enable INT8 online quantization with layerwise offload and verify that the model loads without OOM.
3. Enable RainFusion and INT8 quantization together and verify end-to-end T2VA generation.
4. Run the following unit tests:

```bash
pytest \
  tests/diffusion/attention/test_attention_config.py \
  tests/diffusion/attention/test_rainfusion_plan.py \
  tests/diffusion/quantization/test_int8_config.py
```

## Test Result

Validated on an Atlas 800I A3 server with 8 NPUs, CANN 9.0.1, PyTorch 2.10.0+cpu, and torch_npu 2.10.0.post2.

Common configuration:

* T2VA, 209 frames
* Text encoder TP8
* Ulysses 8
* VPP8 tiled VAE
* Layerwise offload

### Prompt

A red motorcycle races along a coastal highway at sunset, camera tracking alongside.

### RainFusion

`RAINFUSION_ATTN`, `sparsity=0.8`, `start_step=0`, BF16:

* 1536×768 — sparse (`latent_t×h×w = 71424`, multiple of 128)

https://github.com/user-attachments/assets/f9d4a98d-a360-42b6-b228-fcde2f0a1677

* 1344×768 — dense fallback (`62496` video rows, not block-aligned)

https://github.com/user-attachments/assets/09d2fa80-2e64-46cb-82bb-860a7ee748be

### INT8 online quantization

`FLASH_ATTN` with INT8 online quantization:

https://github.com/user-attachments/assets/a36a1bdb-a40e-463b-9cef-f942765d4eac

### RainFusion + INT8 online quantization

`RAINFUSION_ATTN`, `sparsity=0.8`, `start_step=0`, with INT8 online quantization:

https://github.com/user-attachments/assets/f2e051ce-84c5-4f2d-8db7-0621a1981c6d

## Related PR

* [Model] [Ascend] Add soundfile fallback for MiniMax H3 audio loading for NPU support #5699


hsliuustc0106:
modify npu receips

vllm_omni/diffusion/attention/backends/rainfusion_attn.py
```diff
@@ -0,0 +1,391 @@
...
+                )
+
+        self.dense_fallback = FlashAttentionBackend.get_impl_cls()(
+            num_heads=num_heads,
+            head_size=head_size,
+            softmax_scale=softmax_scale,
+            causal=causal,
+            num_kv_heads=num_kv_heads,
+            prefix=prefix,
+            qkv_layout=qkv_layout,
```
hsliuustc0106, line 178: `qkv_layout` is `None` for MiniMax H3, but RainFusion receives `[B, S, N, D]` and the NPU Flash fallback defaults `None` to `BNSD`. The token refiner always takes this dense path because it does not publish RainFusion geometry, so this can fail or interpret the sequence and head axes incorrectly before any sparse block runs. Please pass `BSND` explicitly and cover a real dense fallback path.

vllm_omni/diffusion/attention/backends/registry.py
```diff
@@ -65,6 +65,7 @@ class MyCustomBackend:
     FLASH_ATTN_HUB = "vllm_omni.diffusion.attention.backends.flash_attn_hub.FlashAttentionHubBackend"
     FLASH_ATTN_3_HUB = "vllm_omni.diffusion.attention.backends.flash_attn_hub.FlashAttention3HubBackend"
     TRTLLM_ATTN = "vllm_omni.diffusion.attention.backends.trtllm_attn.TrtllmAttentionBackend"
+    RAINFUSION_ATTN = "vllm_omni.diffusion.attention.backends.rainfusion_attn.RainFusionAttentionBackend"
```
hsliuustc0106, line 68: This makes an NPU-only backend globally selectable, while the CUDA, ROCm, XPU, and MUSA explicit-selection paths accept any enum member. Those platforms instantiate the model and fail only on the first forward; NPU also defers the `mindiesd` probe. Please make platform resolution reject unsupported hardware and validate the dependency before model construction.

vllm_omni/diffusion/attention/backends/rainfusion_attn.py
```diff
@@ -0,0 +1,391 @@
...
+            # contiguous and keep their (t, h, w) order. rf_v2 ignores attn_mask
+            # and reads actual_seq_lengths off the tensor, so these rows are
+            # attended: zeroing the values keeps them out of the numerator, and
+            # at a couple hundred rows against tens of thousands the share of
+            # the softmax denominator they take stays negligible. Measured on
+            # Ascend 910 at sparsity=0, where the mask is fully populated and
+            # the kernel must reproduce dense attention: padded grids land at
+            # 0.06-0.16% mean relative error against 0.06% for grids that need
+            # no pad, and 4-5% for the unpadded mismatch.
+            head = query.new_zeros((query.shape[0], pad, *query.shape[2:]))
```
hsliuustc0106: These rows are still attended because `rf_v2` ignores the mask, so zero K/V entries change every real query softmax denominator. The documented 1344x768 case takes this path with 155 rows, and the comment records additional numerical error. Please keep unaligned geometry dense until a kernel path can exclude padding, and add a fully populated-mask parity test.

vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
```diff
@@ -543,6 +559,16 @@ def __init__(
         self.expand_ratio = expand_ratio
         self.modality_num = modality_num
         self.hidden_size = arch.hidden_size
+        if quant_config is not None and out_features > _NPU_QUANT_MATMUL_MAX_OUT_FEATURES:
```
hsliuustc0106: This applies an Ascend `QuantBatchMatmulV3` limit to every quantization config, so CUDA INT8, FP8, and other methods also lose roughly 43% of the quantized weights. It also checks global `out_features`; `ColumnParallelLinear` invokes the kernel on `output_size_per_partition`, which is below this limit at TP2 and above. Move this capability check into the NPU quant/platform implementation and use the local shape.

vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
```diff
@@ -628,11 +659,14 @@ def __init__(
...
         quant_config: QuantizationConfig | None,
+        *,
+        layer_idx: int,
     ) -> None:
         super().__init__()
         self.norm1 = _norm(arch.hidden_size, eps=arch.norm_eps)
         self.norm2 = _norm(arch.hidden_size, eps=arch.norm_eps)
-        self.attn = MiniMaxH3Attention(arch, quant_config)
+        # The prefix is how RAINFUSION_ATTN recovers the block index for skip_layers.
+        self.attn = MiniMaxH3Attention(arch, quant_config, prefix=f"blocks.{layer_idx}.attn")
```
hsliuustc0106: The new prefix only reaches the attention backend; the QKV/output, MLP, AdaLN, and condition `LinearBase` constructors still use the default empty prefix. That prevents `ignored_layers` and component quantization routing from matching these layers. Please carry complete weight-owner prefixes into every quantized linear and add routing assertions.

vllm_omni/diffusion/model_loader/diffusers_loader.py
```diff
@@ -406,6 +412,28 @@ def load_model(
...
+        offload means none of it is supposed to be resident at inference time.
+
+        Only layers whose quant method materializes weights lazily can do this,
+        since they are the ones that know when a layer is finished.
+        """
+        marked = 0
+        for module in model.modules():
+            quant_method = getattr(module, "quant_method", None)
+            if getattr(quant_method, "uses_meta_device", False):
+                module._offload_after_quant = True
```
hsliuustc0106: `uses_meta_device` does not define support for this private attribute. Only the local INT8 mixin reads `_offload_after_quant`; upstream online FP8 also advertises `uses_meta_device` but ignores it, while the loader logs that all marked layers were offloaded. Please replace this mutation with an explicit quant-method capability or hook and only report supported implementations.

vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
```diff
@@ -402,15 +415,16 @@ def _run_packed_attention(
...
-                "max_seqlen_k": max_seqlen,
-            },
-        )
+        extra: dict[str, Any] = {
+            "cu_seqlens_q": cu_seqlens,
+            "cu_seqlens_k": cu_seqlens,
+            "max_seqlen_q": max_seqlen,
+            "max_seqlen_k": max_seqlen,
+        }
+        if sparse_params is not None:
```
hsliuustc0106: this will not work for GPU/ROCm...

vllm_omni/diffusion/data.py
```diff
@@ -1455,20 +1455,47 @@ def enabled(self) -> bool:
...
+
+
 @dataclass
 class AttentionSpec:
     """Specifies a backend and its typed backend-specific config for one attention role."""
 
     backend: str
     skip_softmax: SkipSoftmaxSpec | None = None
     quant: AttnQuantSpec | None = None
+    rainfusion: RainFusionSpec | None = None
```
hsliuustc0106: the attentionspec should not point to specific hardware fusion kernels

vllm_omni/diffusion/attention/backends/rainfusion_attn.py
Huangzjun: MiniMaxH3Attention now sets qkv_layout="BSND" explicitly (shared by DiT blocks and the token refiner); RainFusion also validates layout. With MindIE-SD's default head_first=False, BSND inputs were already handled correctly via internal transposes — explicit BSND removes four redundant transposes per call. Dense fallback is covered by test_missing_video_layout_falls_back_to_dense and by the refiner path in the NPU runs.

vllm_omni/diffusion/attention/backends/registry.py
Huangzjun: Added supported_platforms and validate_available() on AttentionBackend; RainFusion declares ("npu",) and probes mindiesd. All five platforms call OmniPlatform.validate_diffusion_attn_backend() on explicit backend selection, so unsupported hardware or a missing dependency fails before model construction rather than on the first forward.

vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
Huangzjun: Geometry no longer goes through vendor-specific keys in extra; it is now AttentionMetadata.video_layout: VideoTokenLayout | None (prefix_len + latent_grid), published unconditionally by the model — other backends simply ignore it. Together with the platform gate, selecting RAINFUSION_ATTN on non-NPU hardware fails during backend resolution.

vllm_omni/diffusion/model_loader/diffusers_loader.py
Huangzjun: The loader no longer sets _offload_after_quant based on uses_meta_device; LazyWeightMixin exposes supports_offload_after_quant and enable_offload_after_quant(), and the loader only marks methods that advertise support, avoiding false counts for FP8 and other methods that do not implement this behavior.

vllm_omni/diffusion/data.py
Huangzjun: Renamed to BlockSparseSpec / AttentionSpec.block_sparse, validated against BLOCK_SPARSE_BACKENDS for the selected backend; inner_precise is removed from the user-facing schema and kept internal to the backend. Docs and test_attention_config.py updated accordingly.

vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
Huangzjun: Removed _NPU_QUANT_MATMUL_MAX_OUT_FEATURES from the model; the check now lives in NPUInt8LinearMethod / NPUInt8OnlineLinearMethod.create_weights in int8_config.py and applies to NPU INT8 only. It uses sum(output_partition_sizes) (per-partition width) and delegates to UnquantizedLinearMethod when over the limit. TestNPUQuantMatmulShapeFallback covers wide-layer fallback and recovery after TP sharding.

vllm_omni/diffusion/attention/backends/rainfusion_attn.py
Huangzjun: Removed the prefix-padding path; when latent_t * h * w is not a multiple of 128, _resolve_plan returns None and dense fallback runs (1344×768 is this case). Added test_misaligned_video_segment_stays_dense and test_fully_populated_mask_reproduces_dense_attention (sparsity=0 parity); docs and the recipe now document the alignment rule, with 1536×768 as the sparse example.

vllm_omni/diffusion/attention/backends/rainfusion_attn.py
```diff
@@ -0,0 +1,353 @@
...
+
+        if self.rainfusion.enabled:
+            self._validate_parallel_config()
+            if causal:
+                raise ValueError(
+                    "RAINFUSION_ATTN does not support causal attention: rf_v2 selects key "
+                    "blocks by pooled relevance and cannot express a causal mask. Select "
+                    "FLASH_ATTN for causal roles."
+                )
+            layout = (qkv_layout or _INPUT_LAYOUT).upper()
```
david6666666: If `qkv_layout` is None this check passes by assuming BSND, but the dense fallback below gets the raw value and FlashAttention defaults None to BNSD -- so the sparse and dense paths would disagree for any model that does not declare the layout explicitly. Maybe require an explicit BSND here, or pass the resolved layout to the fallback.

vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
david6666666: This is addressed in the current head -- prefix is now threaded into the qkv/out, fc1/fc2, adaln and condition projections, with a routing test (test_fused_layer_can_be_ignored_by_its_prefix) to back it. Resolving.

vllm_omni/diffusion/attention/backends/rainfusion_attn.py
Huangzjun: the old (qkv_layout or _INPUT_LAYOUT) check could let the sparse path assume BSND while the dense fallback still saw None. That assumption is removed: an undeclared layout now skips sparsity in _resolve_plan (with a one-time warning), so both paths stay dense and the fallback sees the same raw None as plain FLASH_ATTN. Explicit non-BSND layouts are still rejected at construction. MiniMax H3 already declares qkv_layout="BSND", so this only affects layers that leave it unset.

david6666666 (approved):
LGTM

PR #5708, opened 2026-08-03
[diffusion][feature] Enable online FP8 for MiniMax-H3

lishunyang12:
## Purpose

Enable dynamic online FP8 quantization for the MiniMax-H3 DiT while preserving
the checkpoint's mixed BF16/FP32 precision policy.

This change:

- gives every quantizable DiT linear a stable full module prefix;
- keeps the video/audio patch projections, timestep MLP, and final video/audio
  heads in FP32;
- moves the H3-specific QKV reorder and fused gate/up split into
  `load_weights`, so vLLM's online-quantization weight-loader wrapper is not
  overwritten;
- keeps AdaLN activations in BF16 instead of casting them to the FP8 weight
  dtype; and
- documents the supported online FP8 scope and adds focused regression tests.

Online FP8 is approximate. The Qwen3-VL text encoder and both VAEs remain at
checkpoint precision.

## Test Plan

```bash
python -m pytest tests/diffusion/models/minimax_h3 -q \
  --ignore=tests/diffusion/models/minimax_h3/test_minimax_h3_e2e.py

pre-commit run --files \
  benchmarks/diffusion/benchmark_minimax_h3_fp8.py \
  docs/user_guide/quantization/fp8.md \
  tests/diffusion/models/minimax_h3/test_minimax_h3_parallel.py \
  vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
```

Reproduce the 4x B300 BF16/FP8 comparison (the defaults are T2VA, 1344x768,
124 frames at 24 FPS, 50 inference steps, 5 seconds, seed 1101, one warmup,
and one measured request):

```bash
CUDA_VISIBLE_DEVICES=0,1,2,3 \
python benchmarks/diffusion/benchmark_minimax_h3_fp8.py \
  --model /path/to/MiniMax-H3/FL2VA \
  --output-dir ./minimax-h3-fp8-benchmark
```

**vLLM Version:** 0.26.0

**vLLM-Omni Commit:** 6e99997dd94985aecc68fc611c58ab043879a121

**Software:** PyTorch 2.11.0+cu130, CUDA 13.0, driver 610.43.02

## Test Result

- MiniMax-H3 CPU suite: `29 passed`
- Focused quantization/parallel tests after rebase: `8 passed`
- All selected pre-commit hooks passed.

| Mode | Steady wall time | Denoise | Decode | Peak/GPU |
| BF16/FP32 | 79.08 s | 76.45 s | 1.61 s | 49,332 MB |
| Online FP8 | 58.42 s | 55.92 s | 1.28 s | 41,550 MB |

Online FP8 reduced steady latency by 26.1% (1.35x speedup) and sampled peak
memory by 7,782 MB/GPU (15.8%) for this workload.

### B300 text-encoder TP scaling

As a separate checkpoint-precision control, I profiled the Qwen3-VL encoder
with one B300 (`text_encoder_tp_size=1`) and four B300s
(`text_encoder_tp_size=4`). Both configurations were fully resident and used
the same T2VA input. Timings are CUDA-synchronized medians over five measured
requests after two warmups; each TP=4 sample uses the slowest rank as the
distributed wall time.

| Qwen presentation | Metric | 1x B300, TP=1 | 4x B300, TP=4 | TP=4 result |
| 25 tokens | Encoder (`encode_ids`) | 18.994 ms | 23.510 ms | 23.8% slower |
| 25 tokens | Full prompt-encode stage | 23.189 ms | 28.620 ms | 23.4% slower |
| 1,200 tokens | Encoder (`encode_ids`) | 65.494 ms | 43.400 ms | 1.51x faster |
| 1,200 tokens | Full prompt-encode stage | 71.267 ms | 49.719 ms | 1.43x faster |

For the normal short T2VA prompt, TP communication costs more than the
sharded encoder compute saves. At 1,200 tokens, TP=4 reduces encoder compute
latency by 33.7% and full prompt-encode latency by 30.2%. These are encoder TP
results, not FP8 gains; the encoder remains at checkpoint precision in the
BF16/FP8 comparison above.

Two FP8 requests completed. Each output decoded end-to-end as H.264 video plus
AAC stereo audio: 124 frames at 1344x768 and 24 FPS, with 32 kHz audio. Frame
sampling showed coherent motion without corrupt, black, or frozen output.
Because online FP8 changes the diffusion trajectory, same-seed BF16 and FP8
outputs are not expected to be pixel-identical.

https://github.com/user-attachments/assets/88e13614-efe4-4506-a8ad-f9a17f1a7d9d

https://github.com/user-attachments/assets/0b7c8f3c-c542-452a-871b-b0efd53dcea0

### DIFF
--- a/tests/diffusion/models/minimax_h3/test_minimax_h3_parallel.py
+++ b/tests/diffusion/models/minimax_h3/test_minimax_h3_parallel.py
@@ -1,6 +1,10 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
 
+import os
+from types import SimpleNamespace
+from unittest.mock import Mock
+
 import pytest
 import torch
 import torch.nn as nn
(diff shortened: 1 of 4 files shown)


vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
```diff
@@ -106,6 +106,10 @@ def from_mapping(cls, config: Mapping[str, Any]) -> MiniMaxH3DiTArchConfig:
 MINIMAX_H3_ADALN_MODALITY_NUM = 3
 
 
+def _join_prefix(prefix: str, name: str) -> str:
```
gcanlin, line 109: Could we use vLLM's `maybe_prefix`:

```
def maybe_prefix(prefix: str, name: str) -> str:
    return name if not prefix else f"{prefix}.{name}"
```

vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
```diff
@@ -881,14 +881,36 @@ def __init__(
             gather_output=True,
             params_dtype=_BF16_DTYPE,
             quant_config=quant_config,
+            prefix=_join_prefix(prefix, "condition_proj"),
```
hsliuustc0106, line 884: [P1] This local prefix does not match the repo's per-component quantization contract. `MiniMaxH3Pipeline` passes `ComponentQuantizationConfig` unchanged; for `{"transformer": {"method": "fp8"}, "vae": null}`, `resolve("condition_proj")` falls through to the default (`None`), so the explicitly requested DiT FP8 path silently runs BF16. Please resolve the `transformer` component before constructing this model (as other pipelines do), and add a routing test; the plain `quantization="fp8"` path can remain unchanged.

tests/diffusion/models/minimax_h3/test_minimax_h3_parallel.py
```diff
@@ -92,3 +96,121 @@ def test_tp_accepts_checkpoint_supported_sizes():
...
+        "final_adaln_out_features": 2 * 64,
+        "rope_inv_freq_len": 4,
+    }
+    od_config = SimpleNamespace(
+        tf_model_config=arch,
+        parallel_config=SimpleNamespace(ulysses_degree=1),
+    )
+    quant_config = Mock()
+    quant_config.weight_block_size = None
+    quant_config.get_quant_method.return_value = UnquantizedLinearMethod()
```
hsliuustc0106, line 163: [P1] This is named online quantization but uses `UnquantizedLinearMethod`, so it never calls `MiniMaxH3DiTModel.load_weights()`—the code changed in this PR to reorder fused QKV and split FC1 before vLLM's online wrapper. A broken loader or shard call can therefore pass this test. Please add a small synthetic regression using the actual online FP8 method (or a wrapper faithful to it) that loads one QKV and one FC1 tensor and verifies the transformed/sharded weights after processing.

lishunyang12:
Superseded by #5737

PR #5711, opened 2026-08-03
[Frontend][Doc] Warn on misplaced Omni VAE config

### LINKED ISSUE #5639: [Bug]: --additional-config silently fails for parallel_config.vae_patch_parallel_size — should warn or document --vae-patch-parallel-size instead
## Summary

Using `--additional-config` to set `vae_patch_parallel_size` (vLLM's standard config-passing mechanism) is **silently ignored** — no error, no warning, config appears in `non-default args` but never reaches `od_config.parallel_config`. Users waste hours debugging why VAE decode isn't sharding. The dedicated CLI flag `--vae-patch-parallel-size` works correctly.

## Environment
- Image: `quay.io/ascend/vllm-omni:v0.23.0rc1-a3` (also tested on `vllm-ascend:v0.22.1rc1-a3`)
- vllm: 0.23.0, vllm-omni: 0.23.0rc2.dev147, CANN: 9.0.0
- torch 2.10.0 / torch_npu 2.10.0
- NPU: Ascend 910C (SOC `ascend910_9391`, A3), 8 dies, TP=8
- Model: Cosmos3-Nano-16B (video VAE = Wan2.2 VAE / `AutoencoderKLWan`)

## Problem

```bash
# This DOESN'T work (silently ignored):
vllm serve ... --tensor-parallel-size 8 \
    --additional-config '{"vae_use_tiling":true,"parallel_config":{"vae_patch_parallel_size":8}}'

# This WORKS:
vllm serve ... --tensor-parallel-size 8 \
    --vae-patch-parallel-size 8 --vae-use-tiling
```

The `--additional-config` JSON is parsed (confirmed in `non-default args` log), but `vae_patch_parallel_size` never propagates to `od_config.parallel_config.vae_patch_parallel_size` — the registry reads the default value `1`, so `is_distributed_enabled()` returns `False` and VAE decode runs single-rank (23.71s instead of 4.49s).

**No error or warning is emitted.** The user has no indication that the config was ignored.

## Root Cause (confirmed via source-level diagnostics)

Added diagnostic prints to `registry.py` + `is_distributed_enabled()`:

```
# --additional-config parsed (non-default args confirms):
'additional_config': {'vae_use_tiling': True, 'parallel_config': {'vae_patch_parallel_size': 8, 'vae_parallel_mode': 'spatial_shard_height'}}

# but registry reads default 1:
[diag3] REGISTRY vae_patch: vae_pp_size=1 is_distributed_vae=True vae_type=DistributedAutoencoderKLWan vae_use_tiling=False

# so decode gets defaults:
[diag2] decode: use_tiling=False ps=1 mode=tile → is_distributed_enabled=False
```

`--additional-config` is intentionally forwarded to the platform plugin via `VllmConfig.additional_config`; it is not an overlay for Omni's `DiffusionParallelConfig`. The `stage_config.py` parallel_config extraction logic only covers the CLI flag path, so fields placed under `additional_config.parallel_config` are silently accepted by the parser but have no effect on Omni's diffusion runtime.

## Verification (clean container, ZERO source patches)

Recreated container from image `quay.io/ascend/vllm-omni:v0.23.0rc1-a3` (no patches), tested with CLI flag only:

```bash
vllm serve /efs_bj4_az1/Cosmos3-Nano-16B --omni --port 8000 \
    --enforce-eager --max-model-len 8192 --no-guardrails --tensor-parallel-size 8 \
    --vae-patch-parallel-size 8 --vae-use-tiling
```

Result (3× I2V, 1280×720, 240 frames, 50 steps):

| run | VAE decode | pipeline | E2E |
|---|---|---|---|
| warmup | 4.49s | 28.51s | — |
| run 1 | 8.53s | 232.39s | 241.17s |
| run 2 | 8.19s | 231.42s | 240.42s |
| run 3 | 8.00s | 231.09s | 240.09s |

vs `--additional-config` (broken): decode 23.71s, pipeline 254.75s, E2E 269.61s.

**2.9× speedup on VAE decode, 11% on end-to-end.** NPU E2E 240s vs A800 357.41s = **NPU 33% faster**.

## Ask

1. **Warn or error when `--additional-config` contains `parallel_config` sub-fields** that don't propagate to `od_config.parallel_config` — at minimum, log a warning so users know the config was ignored.
2. **Document** in the [VAE Patch Parallelism guide](https://docs.vllm.ai/projects/vllm-omni/en/latest/user_guide/diffusion/parallelism/vae_patch_parallel/): explicitly recommend `--vae-patch-parallel-size` CLI flag and note that `--additional-config` does not work for `parallel_config` fields.
3. ~~**Fix the propagation**~~ (retracted — see [comment](#issuecomment) below; PR #5711's warn approach is correct since `--additional-config` is intentionally a platform-plugin channel, not an Omni overlay).

### Suggest a potential alternative/fix

_No response_

qybaihe:
## Purpose

Fixes #5639.

### What broke

`--additional-config` accepts JSON such as:

```bash
vllm serve MODEL --omni \
  --additional-config '{"vae_use_tiling":true,"parallel_config":{"vae_patch_parallel_size":8}}'
```

The command parses successfully, but those VAE values do not configure Omni's `DiffusionParallelConfig`. Operators receive no indication that VAE patch parallelism remains disabled.

### Root cause

`--additional-config` is intentionally forwarded unchanged to the platform plugin through `VllmConfig.additional_config`; it is not a generic overlay for Omni diffusion configuration. Treating it as another overlay would create ambiguous configuration precedence.

### Fix

- Warn when `--additional-config` contains known Omni VAE fields at the top level or under `parallel_config`.
- Point each misplaced field to the dedicated flags:
  - `--vae-patch-parallel-size`
  - `--vae-parallel-mode`
  - `--vae-use-tiling`
- Preserve `additional_config` unchanged for the platform plugin.
- Document this boundary in the VAE parallelism guide.
- Add real CLI-parser regression coverage for both misplaced VAE fields and valid platform configuration.

## Test Plan

**vLLM Version:** repository CI environment

**vLLM-Omni Commit:** `4b83848f187aeab44414293d0dfbb0dda5197aaf`

```bash
pytest -q tests/entrypoints/test_serve.py -m 'core_model and cpu'
uvx pre-commit run --files \
  vllm_omni/entrypoints/cli/serve.py \
  tests/entrypoints/test_serve.py \
  docs/user_guide/diffusion/parallelism/vae_parallelism.md
```

## Test Result

- Python 3.12 syntax compilation: passed.
- Ruff check and format check: passed.
- All changed-file pre-commit hooks: passed.
- Repository pytest suite: delegated to project CI because the local macOS environment does not provide vLLM runtime packages.

### DIFF
--- a/vllm_omni/entrypoints/cli/serve.py
+++ b/vllm_omni/entrypoints/cli/serve.py
@@ -9,6 +9,7 @@ import argparse
 import json
 import os
 import signal
+from collections.abc import Mapping
 from types import FrameType
 
 import uvloop
@@ -41,6 +42,54 @@ Search by using: `--help=<ConfigGroup>` to explore options by section (e.g.,
 --help=OmniConfig)
   Use `--help=all` to show all available flags at once.
 """
+
+_OMNI_VAE_ADDITIONAL_CONFIG_PATHS: dict[tuple[str, ...], str] = {
+    ("vae_patch_parallel_size",): "--vae-patch-parallel-size",
+    ("vae_parallel_mode",): "--vae-parallel-mode",
+    ("vae_use_tiling",): "--vae-use-tiling",
+    ("parallel_config", "vae_patch_parallel_size"): "--vae-patch-parallel-size",
+    ("parallel_config", "vae_parallel_mode"): "--vae-parallel-mode",
+    ("parallel_config", "vae_use_tiling"): "--vae-use-tiling",
+}
+
+
+def _warn_on_misplaced_omni_vae_config(args: argparse.Namespace) -> None:
+    """Warn when platform config is mistaken for Omni VAE config.
+
+    ``--additional-config`` is forwarded unchanged to the platform plugin via
+    ``VllmConfig.additional_config``. It is not an overlay for Omni's
+    diffusion configuration, so VAE fields placed there would otherwise be
+    accepted by the parser but silently have no effect.
+    """
+    if not getattr(args, "omni", False):
+        return
+
+    additional_config = getattr(args, "additional_config", None)
+    if not isinstance(additional_config, Mapping):
+        return
+
+    misplaced: list[tuple[str, str]] = []
+    for path, flag in _OMNI_VAE_ADDITIONAL_CONFIG_PATHS.items():
+        value: object = additional_config
+        for key in path:
+            if not isinstance(value, Mapping) or key not in value:
+                break
+            value = value[key]
+        else:
+            misplaced.append((".".join(path), flag))
+
+    if not misplaced:
+        return
+
+    fields = ", ".join(path for path, _ in misplaced)
+    flags = ", ".join(dict.fromkeys(flag for _, flag in misplaced))
+    logger.warning(
+        "--additional-config is forwarded to the platform plugin and does not "
+        "configure Omni VAE options. These fields will not take effect here: "
+        "%s. Use the dedicated CLI flags instead: %s.",
+        fields,
+        flags,
+    )
 
 
 def _ensure_vllm_platform():
@@ -131,6 +180,8 @@ class OmniServeCommand(CLISubcommand):
         # vLLM equivalents on the command line would silently disagree with
         # those sources of truth, so reject them at parse time.
         if getattr(args, "omni", False):
+            _warn_on_misplaced_omni_vae_config(args)
+
             explicit_cli_keys: set[str] = getattr(args, "_cli_explicit_keys", set()) or set()
             prohibited_with_omni: dict[str, str] = {
                 "data_parallel_size": "--data-parallel-size",
--- a/docs/user_guide/diffusion/parallelism/vae_parallelism.md
+++ b/docs/user_guide/diffusion/parallelism/vae_parallelism.md
@@ -104,6 +104,15 @@ vllm serve Tongyi-MAI/Z-Image-Turbo --omni --port 8091 \
     --vae-use-tiling
 ```
 
+!!! warning "Use the dedicated VAE CLI options"
+    `--additional-config` is forwarded to the platform plugin through
+    `VllmConfig.additional_config`; it is not merged into Omni's
+    `DiffusionParallelConfig`. Putting `vae_patch_parallel_size` or
+    `vae_parallel_mode` under `additional_config.parallel_config`, or putting
+    `vae_use_tiling` in `additional_config`, therefore does not configure the
+    VAE. Use `--vae-patch-parallel-size`, `--vae-parallel-mode`, and
+    `--vae-use-tiling` as shown above.
+
 ---
 
 ## Configuration Parameters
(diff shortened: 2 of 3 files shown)


vllm_omni/entrypoints/cli/serve.py
```diff
@@ -42,6 +43,54 @@
...
+    ("vae_patch_parallel_size",): "--vae-patch-parallel-size",
+    ("vae_parallel_mode",): "--vae-parallel-mode",
+    ("vae_use_tiling",): "--vae-use-tiling",
+    ("parallel_config", "vae_patch_parallel_size"): "--vae-patch-parallel-size",
+    ("parallel_config", "vae_parallel_mode"): "--vae-parallel-mode",
+    ("parallel_config", "vae_use_tiling"): "--vae-use-tiling",
+}
+
+
+def _warn_on_misplaced_omni_vae_config(args: argparse.Namespace) -> None:
```
hsliuustc0106, line 56: I think we need to work from the checks in additional_config, this will also leave the possibility of misplacing other configs into additional config

gengwei-SlimAgileCommando:
Thanks for this — confirming the bug and the proposed direction.

I'm the author of #5639 and ran source-level diagnostics on the same
image (`quay.io/ascend/vllm-omni:v0.23.0rc1-a3`, Cosmos3-Nano-16B,
Ascend 910C, TP=8). The bug is real and the warn-only approach here
is the correct fix.

Evidence (clean container, zero source patches):

[diag2] is_distributed_enabled: parallel_size=1 dist_init=True use_tiling=False → False
[diag3] REGISTRY vae_patch: vae_pp_size=1 is_distributed_vae=True vae_type=DistributedAutoencoderKLWan vae_use_tiling=False
→ vae_pp_size=1 (NOT 8!) — od_config.parallel_config.vae_patch_parallel_size stayed at default

Impact: 23.71s decode (broken) vs 8.00s (with --vae-patch-parallel-size 8 --vae-use-tiling) — 2.9× slowdown. Verified across 3 runs on clean container.

A few notes on the PR:

1. **Coverage looks correct.** The 6 paths in `_OMNI_VAE_ADDDITIONAL_CONFIG_PATHS` cover all three VAE fields (`vae_patch_parallel_size`, `vae_parallel_mode`, `vae_use_tiling`) at both top level and under `parallel_config`. This matches what I observed in diagnostics — `vae_use_tiling` was also silently ignored, not just `vae_patch_parallel_size`.

2. **Preserving `additional_config` unchanged is the right call.** Your root cause explanation is correct: `--additional-config` is forwarded to the platform plugin on purpose, and merging it into `DiffusionParallelConfig` would create ambiguous precedence. I'll update #5639 to retract the "fix the propagation" ask (Option 3) — it was based on a misunderstanding of the config boundary.

3. **Re: hsliuustc0106's comment** about generalizing beyond VAE — agreed in principle, but suggest doing VAE first (this PR is ready) and following up with a separate PR for a broader misplaced-config check. Scope creep here would slow down a fix that unblocks a 2.9× regression.

I'll update #5639 to point at this PR as the fix.

PR #5712, opened 2026-08-03
[Model] Cosmos3: Reduce video output transfer overhead

rahul-steiger-nv:
## Purpose

Reduce Cosmos3 video response latency and peak GPU memory after VAE decode.

Previously, Cosmos3 returned decoded video as a bfloat16 tensor in
`[B, C, T, H, W]` layout. The diffusion IPC path then promoted bfloat16 to
float32 for NumPy/shared-memory compatibility, and postprocessing converted the
result to display range before the video encoder converted it again to uint8.
For a 1280x720, 189-frame video, this widened the shared-memory payload to about
2.09 GB even though the final encoder input is only about 0.52 GB of uint8
frames.

This change converts decoded video on the output rank directly to display-ready
uint8 `[B, T, H, W, C]` before IPC. The existing asynchronous pinned-memory D2H
path then transfers the narrower payload without an additional format
conversion.

Key details:

- A 64 MiB temporal chunk budget bounds conversion intermediates at the
  post-VAE memory peak.
- Historical bfloat16 arithmetic and final display bytes are preserved.
- Conversion runs only on the output rank and is skipped for T2I and CPU
  offload.
- Guardrails consume uint8 directly, avoiding a float round trip.
- `np`, `pt`, and `pil` output types remain supported.

## Test Plan

- CPU tests for output parity, chunking, rank/offload gates, output types, and
  guardrails.
- CUDA regression for sub-threshold device output reaching NumPy postprocess.
- Online-serving T2I/T2V/I2V/V2V smoke coverage.
- Same-node before/after serving benchmarks.

```bash
pytest -q \
  tests/diffusion/models/cosmos3/test_cosmos3_pipeline.py \
  tests/diffusion/models/cosmos3/test_cosmos3_guardrails.py
```

**vLLM Version:** 0.25.0

**vLLM-Omni Baseline Commit:** `a4ea67a21b`

**vLLM-Omni Optimized Commit:** `d80fa73a8a`

### Unit and E2E correctness

- CPU: `89 passed, 1 skipped`; CUDA regression: `1 passed`; online smoke:
  `2 passed`; Ruff: all checks passed.
- Every baseline and optimized benchmark completed 5/5 measured requests.

### Single-GPU benchmark

NVIDIA GB300; T2V 1280x720, 189 frames, 35 steps; no VAE tiling;
concurrency 1; 1 warmup + 5 measured requests; CUDNN attention; guardrails off.

| revision | mean latency | median latency | throughput | peak GPU memory | mean stage generation |
| baseline | 87.243 s | 87.295 s | 0.01146 req/s | 49.572 GB | 83.010 s |
| optimized | 84.131 s | 84.132 s | 0.01189 req/s | 47.560 GB | 82.144 s |
| change | **-3.112 s (-3.6%)** | **-3.163 s (-3.6%)** | **+0.00042 req/s (+3.7%)** | **-2.012 GB (-4.1%)** | **-0.865 s (-1.0%)** |

### Four-GPU benchmark

4x NVIDIA GB300; same workload; CFG parallel 2; Ulysses 2; VAE patch
parallel 4 with `spatial_shard_width`; concurrency 1; 1 warmup + 5 measured
requests; CUDNN attention; guardrails off.

| revision | mean latency | median latency | throughput | peak GPU memory | mean stage generation |
| baseline | 28.846 s | 28.712 s | 0.03467 req/s | 41.240 GB | 24.622 s |
| optimized | 27.294 s | 27.285 s | 0.03664 req/s | 39.244 GB | 23.631 s |
| change | **-1.552 s (-5.4%)** | **-1.427 s (-5.0%)** | **+0.00197 req/s (+5.7%)** | **-1.996 GB (-4.8%)** | **-0.991 s (-4.0%)** |

### Notes

Pipeline `forward` and VAE decode times were unchanged, confirming that the gain
comes from output handling. MP4 encoding improved by 0.544 s (24.9%) on one GPU
and 0.547 s (25.1%) on four GPUs. The four-GPU stage-generation reduction
(0.991 s) plus encoding reduction accounts for nearly all of its 1.552 s E2E
improvement. Byte-parity tests confirm chunking does not alter display output.

### DIFF
--- a/vllm_omni/diffusion/models/cosmos3/pipeline_cosmos3.py
+++ b/vllm_omni/diffusion/models/cosmos3/pipeline_cosmos3.py
@@ -567,6 +567,68 @@ def get_cosmos3_pre_process_func(od_config: OmniDiffusionConfig):
     return pre_process_func
 
 
+# Bound the temporary float32 conversion immediately after VAE decode.
+_DISPLAY_CHUNK_BYTES = 64 << 20
+
+
+def _display_chunk_frame_count(video: torch.Tensor, chunk_bytes: int | None = None) -> int:
+    """Return the number of frames whose float32 form fits the budget."""
+    if video.ndim != 5:
+        raise ValueError(f"Expected decoded video with shape [B, C, T, H, W], got {tuple(video.shape)}.")
+    if chunk_bytes is None:
+        chunk_bytes = _DISPLAY_CHUNK_BYTES
+    batch, channels, _, height, width = video.shape
+    # Non-float32 inputs briefly coexist with their float32 form during the
+    # cast that precedes scaling to display bytes.
+    bytes_per_value = 4 if video.dtype == torch.float32 else video.element_size() + 4
+    intermediate_bytes_per_frame = batch * height * width * channels * bytes_per_value
+    return max(1, chunk_bytes // intermediate_bytes_per_frame)
+
+
+def to_display_uint8(video: torch.Tensor) -> torch.Tensor:
+    """Convert a decoded video from VAE range to display-ready uint8 frames.
+
+    The VAE emits ``[B, C, T, H, W]`` in ``[-1, 1]``; every consumer ultimately
+    wants ``[B, T, H, W, C]`` uint8 in ``[0, 255]``. Narrowing before IPC
+    reduces the device-to-host and shared-memory payload.
+
+    The arithmetic runs in float32 and matches the old two-step path exactly —
+    ``postprocess_video`` produced ``(x / 2 + 0.5).clamp(0, 1)`` and the encoders
+    then scaled and rounded — so output is bit-identical to what callers saw, and
+    slicing does not change that because frames convert independently.
+    """
+    video = video.detach()
+    batch, channels, frames, height, width = video.shape
+    out = torch.empty(
+        (batch, frames, height, width, channels),
+        dtype=torch.uint8,
+        device=video.device,
+    )
+
+    step = _display_chunk_frame_count(video)
+    for start in range(0, frames, step):
+        stop = min(start + step, frames)
+        # Preserve the input dtype for denormalization to match the historical
+        # VideoProcessor path, which performed this arithmetic before casting
+        # to float32 for NumPy.
+        chunk = video[:, :, start:stop].permute(0, 2, 3, 4, 1).to(video.dtype, copy=True)
+        chunk = chunk.mul_(0.5).add_(0.5).clamp_(0, 1)
+        if chunk.dtype != torch.float32:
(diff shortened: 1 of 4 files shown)


rahul-steiger-nv:
Friendly ping for review when you have a chance 🙂 Would appreciate any feedback. Thanks!

rahul-steiger-nv:
I will leave this PR open until #6477 and #6541 have been resolved. However, I do not plan on merging this into the main branch

vllm_omni/diffusion/models/cosmos3/pipeline_cosmos3.py
```diff
@@ -688,7 +750,20 @@ def _postprocess_action(action: Any, metadata: dict[str, Any]) -> Any:
         guardrails_enabled = is_guardrails_enabled(od_config, sampling_params)
         if guardrails_enabled:
             video = check_video_safety(video)
-        processed_video = video_processor.postprocess_video(video, output_type=output_type)
+        if video.dtype == torch.uint8:
```
RuixiangMa, line 836: np and pt output types silently change dtype, layout, and value range

vllm_omni/diffusion/models/cosmos3/pipeline_cosmos3.py
```diff
@@ -688,7 +750,20 @@ def _postprocess_action(action: Any, metadata: dict[str, Any]) -> Any:
...
+            # Display-ready output intentionally remains uint8 for ``np`` and
+            # ``pt`` so serving does not widen it only to encode it as bytes.
+            if output_type == "pt":
+                processed_video = video
+            elif output_type == "np":
+                processed_video = video.detach().cpu().numpy()
+            elif output_type == "pil":
+                vae_video = video.detach().cpu().permute(0, 4, 1, 2, 3).float().div_(127.5).sub_(1.0)
```
RuixiangMa: This allocates a full float32 [B,C,T,H,W] (~4.18 GB for 1280×720×189 frames) on CPU, negating the memory win for the pil path. The affine is also done twice (once to make uint8, once to undo it). Consider a direct uint8→PIL path via Image.fromarray per frame.

vllm_omni/diffusion/models/cosmos3/pipeline_cosmos3.py
rahul-steiger-nv: updated

vllm_omni/diffusion/models/cosmos3/pipeline_cosmos3.py
rahul-steiger-nv: good point, I updated it

Issue #4473, opened 2026-06-16
[Bug] sleep/wake can silently corrupt generation or crash when wake tags are incomplete


hbhflw2000:
## Summary

We are using vLLM-Omni as an async rollout engine for Qwen3-Omni-30B-A3B-Instruct in a colocated RL post-training setup. Cold generation is healthy. However, standalone vLLM-Omni sleep/wake probes reproduce two failure modes without Megatron/PPO:

1. `sleep(level=2)` followed by `wake(tags=None/default)` returns successfully, but post-wake generation is semantically corrupted.
2. `sleep(level=1)` followed by `wake(tags=["weights"])` allows generation to proceed while KV/runtime state appears to remain asleep, then crashes with CUDA illegal memory access / EngineDead.

There is also a related lifecycle issue: duplicate `wake(tags=None)` can attempt to wake already-warm CuMem allocations. A local workaround that tracks warm/sleeping tags and skips already-warm tags makes the colocate-style probe pass.

## Environment

```text
Python: 3.12
vLLM: 0.20.2+cu126
vLLM-Omni: 0.1.dev1 lineage around c7178d89
GPU: 4 x A800/A100 80GB
Model: Qwen3-Omni-30B-A3B-Instruct
Runtime: vLLM-Omni async serving, thinker-only TP4
dtype: bfloat16
max_model_len: 1024
enable_sleep_mode: true
distributed_executor_backend: mp
```

Local vLLM-Omni checkout used for this repro:

```text
branch: align/vllm-omni-c7178d89
head: 05eeec970cc197241c9e937606592a2347f7b3a2
```

## Why this points at sleep/wake

The same model and prompts generate coherent GSM8K reasoning before sleep. A colocate-style fake update probe can also pass when duplicate full wake is skipped safely. The failures are triggered by sleep/wake state transitions:

```text
cold generate -> sleep -> wake -> generate
```

or by admitting generation after only a partial wake:

```text
cold generate -> sleep(level=1) -> wake(tags=["weights"]) -> generate
```

## Reproduction A: `sleep(level=2)` silently corrupts generation

Minimal sequence:

```text
cold generation
sleep(level=2)
wake(tags=None/default)
post-wake generation
```

Expected:

```text
Generation remains coherent after wake, or level-2 sleep is explicitly rejected as unsupported.
```

Observed:

```text
cold generation: coherent GSM8K reasoning
[...]

PR #5713, opened 2026-08-03
[CI/Build] Speed up omni sleep-mode entrypoint tests with shared engines

yenuo26:

## Purpose

Refactor and thin `tests/entrypoints/test_omni_sleep_mode.py` so merge **Entrypoint Test with H100** spends less wall time on repeated BAGEL cold starts.

Previously the module cold-started BAGEL many times (function-scoped engines + per-test GPU cleanup), which pushed the Entrypoint H100 job past the old 60m timeout after the vLLM 0.26 image startup regression.

This PR:

- Shares class-scoped `llm_engine` / `diffusion_engine` fixtures (with `_ensure_awake` between cases).
- Runs device cleanup once per module instead of per test.
- Merges overlapping diffusion lifecycle cases into `test_diffusion_sleep_wake_generate`.
- Keeps pure-diffusion coverage as `test_pure_diffusion_sleep_wake` on **L4 / TP=1** (`qwen_image_random`); diffusion **TP=2** sleep/wake remains covered by `TestOmniDiffusionSleepMode`.
- Moves `#4473` Repro A (`wake` after `sleep(level=2)`) off the BAGEL entrypoint path:
  - **Live HTTP e2e** on `Qwen/Qwen-Image` in `tests/dfx/reliability/invalid_param_test/test_invalid_server_control.py` (`test_wakeup_after_level2_sleep_fails`): shared module server with `--enable-sleep-mode`; case runs last because level=2 discards weights. Asserts OpenAI-style JSON with **HTTP 501** (`sleep(level=2)` / `not yet implemented`) via upstream `build_openai_app`'s `NotImplementedError` handler. No `vllm_omni` production error-mapping change.
  - **CPU route check** in `tests/entrypoints/openai_api/test_omni_sleep_wakeup.py` (`test_wakeup_route_propagates_not_implemented_error`): bare FastAPI `TestClient` only verifies `/v1/omni/wakeup` does not swallow `NotImplementedError` (mock authors the exception on purpose).
- Renames GPU-centric helpers to device-neutral names (`clean_device_envs`, `_module_device_cleanup`).

No production `vllm_omni/` code changes.

## Test Plan

**vLLM Version:** (CI image / local env as used by Buildkite CUDA merge / weekly)

**vLLM-Omni Commit:** `3ec7f548` (branch `entrypoints`)

### Ready / CPU (route propagation)

```bash
pytest -sv tests/entrypoints/openai_api/test_omni_sleep_wakeup.py -m "core_model and cpu" --run-level core_model
```

### Merge (Entrypoint H100)

```bash
pytest -s -v tests/entrypoints/ -m "advanced_model and cuda and H100" --run-level "advanced_model"
```

### Weekly (invalid-param / Qwen-Image level-2 HTTP)

```bash
pytest -s -v tests/dfx/reliability/invalid_param_test/test_invalid_server_control.py --run-level full_model
```

### Ready CI

8c61b651dd603cf7881e210a5a352aa7
5cc1204f-2344-4f49-8435-bd1ff60b6faa

### Merge CI (**2100.9s -> 901.24s**, **49min -> 35min**)

**Before**

89b29653-b4df-461a-ab16-630be01a029c

**After**

a16d6fc4-9f37-490a-acb2-4c593bc5c6b7
c473b0ea-7dd7-47d7-b191-c16b20ee51c5

### Weekly CI

d09003b8-21d9-423d-865f-f942ffc29c99

Live level-2 wakeup failure (when asserting `err_code: 500` by mistake) surfaced as **501** with message containing `sleep(level=2)` / `not yet implemented`, confirming the e2e expectation.


yenuo26:
@lishunyang12 @Gaohan123 please help to add ready and merge-test label

tests/dfx/reliability/invalid_param_test/test_invalid_server_control.py
```diff
@@ -93,3 +101,37 @@ def test_omni_sleep_wakeup_invalid_requests(
...
+    sleep_resp = sleep_resps[0]
+    assert sleep_resp.status_code == 200, sleep_resp
+    assert sleep_resp.json_body is not None
+    assert sleep_resp.json_body.get("status") == "SUCCESS", sleep_resp.json_body
+
+    openai_client.send_omni_wakeup_http_request(
+        {
+            "json": {"stage_ids": [0]},
+            "timeout": 120,
+            "err_code": 501,
```
Gaohan123, line 134: [P1] This expectation does not match the current HTTP error mapping. `omni_wakeup()` directly awaits `engine_client.wake_up()` and there is no exception handler that converts `NotImplementedError` into a 501 response. Under a live server, this exception is therefore surfaced as a generic HTTP 500 response, whose body will not contain `sleep(level=2)` or `not yet implemented`.

This also contradicts the PR description and the new CPU test's docstring, both of which state that the live endpoint returns a generic 500. If this PR intentionally avoids changing production error mapping, please expect HTTP 500 here and remove the `err_message` assertion. Alternatively, the production route must explicitly translate `NotImplementedError` into a structured 501 response.

tests/entrypoints/openai_api/test_omni_sleep_wakeup.py
```diff
@@ -175,6 +175,31 @@ def test_wakeup_engine_not_support(sleep_incapable_engine):
...
 
 
+def test_wakeup_after_level2_sleep_raises_not_implemented(sleep_capable_engine, mocker):
+    """#4473 Repro A message check without changing HTTP error mapping.
+
+    Live servers return a generic 500 body for unhandled ``NotImplementedError``.
+    ``TestClient(raise_server_exceptions=True)`` re-raises in-process so we can
+    still assert the engine contract text.
+    """
+    sleep_capable_engine.wake_up = mocker.AsyncMock(
```
Gaohan123: [P2] This test does not verify the real level-2 wake contract because both the exception type and its exact message are authored inside the mock. It never performs `sleep(level=2)` or invokes the production component responsible for detecting the discarded-weight state. Consequently, the test would still pass if the real engine stopped raising this exception or changed its message.

Please either rename this as a route exception-propagation test and add a unit test around the real sleep/wake state implementation, or configure a lightweight real component whose `sleep(level=2)` followed by `wake_up()` produces the production exception.

tests/dfx/reliability/invalid_param_test/test_invalid_server_control.py
yenuo26: Agreed that omni_wakeup() itself doesn't explicitly convert the exception locally. However, the live app registers an upstream NotImplementedError → 501 handler via build_openai_app, so the live endpoint returning 501 with an OpenAI-style JSON message is expected behavior (as verified when I set the error code check to 500, with the following actual response).
This PR does not change the production error mapping. The CPU test only covers route propagation (bare FastAPI), while the E2E test covers the full error mapping. I'll update the CPU test docstring to correct the claim that live returns 500, and also rename the test case accordingly.
```
AssertionError: (HttpResponse(status_code=501, success=False, error_message='{"error":{"message":"wake_up() after sleep(level=2) is not yet implemented: weights were discarded from GPU and reloading from disk is not yet supported. Use sleep(level=1) instead, which offloads weights to CPU RAM and supports fast DMA restore.","type":"NotImplementedError","param":null,"code":501}}', json_body={'error': {'message': 'wake_up() after sleep(level=2) is not yet implemented: weights were discarded from GPU and reloading from disk is not yet supported. Use sleep(level=1) instead, which offloads weights to CPU RAM and supports fast DMA restore.', 'type': 'NotImplementedError', 'param': None, 'code': 501}}), 500)
```

tests/entrypoints/openai_api/test_omni_sleep_wakeup.py
yenuo26: Good catch — agreed this case was overstating what it covers.
I’ll rename it to a route exception-propagation test and update the docstring accordingly: it only checks that /v1/omni/wakeup does not swallow a NotImplementedError raised by the engine client (mock authors the exception on purpose).
The real level-2 contract — sleep(level=2) then wake_up() hitting the production discarded-weight path and the live HTTP error mapping — is already covered by the HTTP e2e in tests/dfx/reliability/invalid_param_test/test_invalid_server_control.py(test_wakeup_after_level2_sleep_fails)

Gaohan123 (approved):
LGTM. Thanks

PR #4795, opened 2026-06-30
[Feature] timestamps for TTS via a forced-aligner pooling

wjinxu:
## Purpose

Adds **word-level timestamps** to TTS output by running a **forced aligner as a non-generative pooling stage** appended to the Omni pipeline. This is the first concrete user of RFC #4468 (auxiliary / non-generative stage support) and supersedes the in-process-sidecar approach from #4034.

The aligner is a pooling model (`runner="pooling"`, `pooling_task="token_classify"`) that consumes the synthesized audio + ground-truth text and emits per-word `[start_ms, end_ms]`. Instead of a separate in-process `vllm.LLM`, it runs as a real pipeline stage, so it shares the engine-core machinery (scheduling, IPC, worker) with the rest of the pipeline.

```
Talker (LLM_AR) ──▶ Code2Wav (LLM_GENERATION, audio) ──▶ ForcedAligner (pooling, timestamps)
                                       │                              │
                                       ▼                              ▼
                                  audio to client             word timestamps to client
```

**Usage** — launch with `--forced-aligner <model>` and set `word_timestamps` on the request:

```bash
vllm-omni serve Qwen/Qwen3-TTS-12Hz-1.7B-CustomVoice --omni \
    --deploy-config vllm_omni/deploy/qwen3_tts.yaml --trust-remote-code \
    --forced-aligner Qwen/Qwen3-ForcedAligner-0.6B \
    --forced-aligner-device 2
```

`--forced-aligner-device` places the aligner stage on its own GPU; omit it to let the stage co-locate on an existing stage's GPU (it is small).

- non-stream `POST /v1/audio/speech` → `X-Word-Timestamps` response header
- streaming `WS /v1/audio/speech/stream` → per-sentence `audio.chunk` frames with a trailing timestamps frame
- demo: `examples/online_serving/text_to_speech/qwen3_tts/word_timestamps_demo.py` (per-word highlighting + barge-in)

### Design — feature isolated, core stays generic

The flag injects the stage onto the tail of *any* TTS pipeline (`inject_forced_aligner_stage`), so users flip a flag instead of editing a deploy per base model. All the core changes are **generic pooling-stage support keyed off vLLM's own signals**, not forced-aligner specifics:

- `stage_init` selects `PoolingParams` when `runner=="pooling"`;
- the orchestrator advertises the stage's pooling tasks for `PoolingParams` requests;
- the AR model runner routes `is_pooling_model` output to `_pool()` (skipping the omni hidden-state prefix cache / mm extraction);
[...]

PR #3526, opened 2026-05-12
[Frontend][Model][Qwen3-Omni] Enable input audio transcription with inline SSE transcript

nicko170:
## Purpose

Qwen3-Omni's Thinker subcomponent implements `SupportsTranscription` (it has `get_speech_to_text_config`, `get_generation_prompt`, the supported-language map, etc.) but the omni wrapper that the model registry resolves does not, so:

1. **`/v1/audio/transcriptions` is unreachable** — `AsyncOmniEngine` hard-codes `supported_tasks` to `{generate, speech}` regardless of the loaded model's capability (the existing `# TODO(Peiqi): Hack here` comment flags it). Even if the route were registered, `OpenAIServingTranscription.__init__` crashes with `AttributeError: type object 'Qwen3OmniMoeForConditionalGeneration' has no attribute 'get_speech_to_text_config'`.

2. **`/v1/chat/completions` cannot return the user-side transcript** — only the assistant's text/audio modalities. Users who want a live conversation transcript have to make a separate ASR call per turn, doubling HTTP round-trips.

This PR addresses [RFC #1481](https://github.com/vllm-project/vllm-omni/issues/1481), which lists `{\"type\": \"transcript\", \"text\": ...}` SSE events as future work.

## Changes

**Three files, ~260 LoC.**

### 1. `vllm_omni/model_executor/models/qwen3_omni/qwen3_omni.py`
The omni wrapper now declares `SupportsTranscription` and delegates the protocol's classmethods (`get_speech_to_text_config`, `get_generation_prompt`, `validate_language`, `get_num_audio_tokens`, `post_process_output`) to the inner Thinker. No new behavior — just exposes the Thinker's existing ASR surface through the class the entrypoint resolves.

### 2. `vllm_omni/engine/async_omni_engine.py`
Replaces the hardcoded \`supported_tasks\` set with a **capability-based probe**: when the comprehension stage's model class passes \`supports_transcription(model_cls)\`, \`\"transcription\"\` is added to \`self.supported_tasks\`. Resolves via \`ModelRegistry.resolve_model_cls(architectures, model_config)\`. Removes the \`TODO(Peiqi): Hack here\` comment. Works generically — any future omni model whose comprehension stage implements the interface gets the route for free.

### 3. \`vllm_omni/entrypoints/openai/serving_chat.py\`
[...]

PR #5714, opened 2026-08-03
[Frontend] OpenAI-compliant /v1/audio/transcriptions with word-level alignment

TheCodeWrangler:
> **Design discussion: #5722.** This is over the LOC threshold in `docs/contributing/README.md`, so direction belongs there. Still draft pending answers on two questions in that thread.

## Purpose

Makes `/v1/audio/transcriptions` reachable under `--omni`, and adds OpenAI-compliant word-level timestamps to it.

If you only want a transcript you do not need this: `vllm serve Qwen/Qwen3-ASR-1.7B` already serves the endpoint. Word timestamps are the different problem. Qwen3-ASR cannot emit them (`supports_segment_timestamp = False`, and upstream rejects `verbose_json` for it), so getting them means running a second model, a forced aligner, over the same audio. Two models behind one endpoint is what vLLM-Omni is for.

Both models are already registered upstream in vLLM 0.26 (`qwen3_asr.py`, `qwen3_asr_forced_aligner.py`). #4034 already vendored the aligner here for streaming TTS. Upstream even defines `TranscriptionResponseVerbose.words` and `timestamp_granularities[]` and never populates them. So this is mostly composition, plus the work to make the composition fast.

## Changes

**Reach the endpoint.** `AsyncOmniEngine` hardcoded `supported_tasks` to `{generate, speech}`, so `api_server` never constructed `OpenAIServingTranscription` and the route was dead for every model. Replaced with a capability probe on vLLM's `supports_transcription()`.

The probe also requires the stage to be a *text final-output* stage. A request cannot name a stage: `output_modalities` resolves to the last stage emitting that modality. `aura_omni` forced this, since its stage 0 is Qwen3-ASR but stage 1 is the text terminal, so a transcription request there would return a conversational reply rather than a transcript.

Supersedes the engine half of #3526, rebased onto the `stage_pools` -> `stage_clients` refactor that PR predates, carried with `Co-authored-by:`. Its other half (Qwen3-Omni wrapper delegation) is orthogonal and still worth landing separately.

**A `qwen3_asr` pipeline.** One stage, audio in, text out. HF `model_type` is `qwen3_asr`, so `vllm-omni serve Qwen/Qwen3-ASR-1.7B --omni` finds it.

**Populate `words`.** Subclasses upstream transcription serving and fills in timestamps from the shared aligner when the request asks for word granularity. Off by default. Since upstream refuses `verbose_json` without segment-timestamp support, a word-only request runs the inner call as `json` and we build the verbose envelope.

## Results

1xH100, 30s clips, no prefix or mm caching, steady state, 3 runs, median. Harness included at `benchmarks/asr/benchmark_transcriptions.py`.

| | req/s | RTFx | GPU util |
|---|---:|---:|---:|
| ASR only | 57.1 | 1,714 | 45% |
| With word timestamps | 52.0 | 1,560 | 51% |

**Alignment costs 9%.** Getting there was mostly bug-fixing, and none of it was where I first looked:

- **Audio decode is GIL-bound.** ~75 ms for a 30s 48kHz stereo mp3, flat past ~4 threads, so `VLLM_MAX_AUDIO_PREPROCESS_WORKERS` does nothing. vLLM's answer is `--api-server-count`, which vllm-omni rejects under `--omni`. A process pool restores it (`audio_decode_procs`, off by default).
- **The aligner never batched.** `align()` took `_encode_lock` for its idempotent load check while the worker holds that lock for the whole encode, so nothing could enqueue mid-batch and batch size sat at 1. Now 8-12.
- **The ASR-to-aligner handoff missed 65% of the time.** Keyed by content hash with a single slot, so concurrent requests carrying identical audio clobbered each other and two thirds of them decoded the upload twice. Now 0%.
- **The aligner was sized as a generative model.** It inherited `gpu_memory_utilization: 0.30` from the TTS profile, but it is a pooling model (`token_classify`) with no KV cache, so it was reserving 17.9 GiB it cannot fill. At 0.10 with the ASR stage at 0.80, alignment goes from costing 16% to 9%.

Note the ceiling is not the aligner: ASR-only sits at **45% GPU utilisation** on its own, and turning alignment on *raises* utilisation to 51% because it fills idle GPU rather than competing for busy GPU. That is why the cost is 9% rather than proportional to the extra model.

## Test plan

- `tests/engine/test_transcription_capability.py` - 21 tests for the capability matrix against synthetic stage pools, no weights.
- `tests/utils/test_forced_aligner_batching.py` - 12 tests for `align()`'s batching contract. `forced_aligner` is shared with streaming TTS, whose own tests mock `forced_align` and so would pass even if it were broken outright.
- `tests/entrypoints/openai_api/test_serving_transcription.py` - 26 tests for the alignment paths, upstream serving stubbed.
- All three mutation-checked. Reverting the single-slot handoff, the `_encode_lock` fix, the int16 clip, the text-terminal constraint, or the job/result routing each fails a test.
- `tests/engine/ tests/utils/ tests/entrypoints/ -m cpu`: **1739 passed**. Two pre-existing failures (`test_resolve_dreamzero_config`, `test_uses_registry_default_deploy_...`) reproduce on unmodified `main`.
- End to end on H100 with `Qwen/Qwen3-ASR-1.7B` + `Qwen/Qwen3-ForcedAligner-0.6B`: 78 word timestamps spanning 0.0-30.0s on a 30s clip.
- Streaming TTS regression checked against a live Qwen3-TTS server, 1 and 6 concurrent WebSocket sessions, 17 monotonic word timestamps each, no cross-talk.

## Sizing

The shipped profile is H100-80GB shaped and says so. Verified failing on a `MIG 1g.10gb` slice, where 3.9 GiB of weights plus an encoder cache profiled for 84 audio items leaves nothing for KV. `gpu_memory_utilization` is also the wrong unit for the aligner specifically, since its requirement is absolute (~1.2 GiB of weights). Making that declarable is follow-up work, discussed in #5722.

## Known gaps

- [ ] The aligner is an in-process sidecar. #5722 proposes it become a pooling stage; that is blocked on two questions in that thread (terminal side-output, and coordination with #4795).
- [ ] `stream=true` does not emit word timestamps. Needs a per-sentence hook.
- [ ] `/v1/audio/translations` untouched.
- [ ] Memory schema: `kv_cache_memory_bytes` per stage, an absolute floor for non-generative stages, and validating that stage allocations fit the device. Follow-up PRs.

Closed since the first draft: `fork` with CUDA initialised in the API server (now `forkserver` with bounded worker lifetime), the aligner loading lazily and failing the first request instead of startup, and `serving_transcription.py` having no tests.


TheCodeWrangler:
`audio_decode_procs` is plumbed through five files (`stage_config.py`, `async_omni_engine.py`, `async_omni.py`, `api_server.py`, `serving_transcription.py`), which is a lot of surface for one knob. Posting the measurement that justifies it.

## The A/B

Same server, same input, same GPU, only the knob changed. Qwen3-ASR-1.7B on one H100, 30s 48 kHz stereo mp3, concurrency 384, no prefix or mm caching, 3 runs, median. Harness is `benchmarks/asr/benchmark_transcriptions.py`:

```
audio_decode_procs: 0     19.9 req/s    spread  2%
audio_decode_procs: 16    52.6 req/s    spread 39%
```

2.6x. The 2% spread on the unpooled arm (19.71 / 19.86 / 20.14) is what a hard bottleneck looks like: the server is doing exactly as much work as it can and nothing else varies.

Decode is GIL-bound. A 30s 48 kHz stereo mp3 costs ~75 ms to decode, downmix and resample, and it plateaus around 40 clips/s per process no matter how many *threads* upstream's preprocess pool is given (flat past ~4). Raising `VLLM_MAX_AUDIO_PREPROCESS_WORKERS` does nothing for the same reason.

Worth noting there is a second, independent route past the same bottleneck: a client sending 16 kHz mono skips the downmix and resample entirely, ~1.1 ms per clip against ~56 ms. The two do not compound. With the pool enabled, changing input format measures no further gain, because decode has already stopped being the constraint.

## Defaults

`audio_decode_procs` defaults to 0, so nothing changes for anyone who does not set it. The `qwen3_asr` deploy profile sets 16.

TheCodeWrangler:
Found a defect in this PR's own defaults while benchmarking the follow-up. Flagging it here because it affects the config this PR ships.

**`qwen3_asr.yaml` at `gpu_memory_utilization: 0.80` plus the aligner at `0.10` runs out of GPU memory on realistic audio.**

```
torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 1.43 GiB
  StageEngineCoreProc_stage0 -> gpu_ar_model_runner.execute_model
```

It does not reproduce on a benchmark that posts the same audio file repeatedly, which is what I had been running. Identical bytes hash the same, so vLLM's multimodal processor cache serves the encoder output and the encoder activations are never allocated. Send distinct clips, as any real workload does, and stage 0 has to encode every request; the activation peak no longer fits beside a second model in the same process.

Reproduced with 2600 distinct 30s clips (same speech, inaudible per-clip noise, so only the content hash differs). Dropping stage 0 to `0.70` and giving the aligner `0.15` runs clean.

Two things follow. The packaged default should probably move to something like `0.70` / `0.15`, since the current split only survives a workload with repeating audio. And the throughput numbers in this PR were measured with a repeated-file harness, so they are inflated for the same reason -- on ASR alone I measured 97 req/s with repeated audio against 58 req/s with distinct clips and the mm cache disabled. I would not trust the current figures without a re-run.

Happy to push the config change here if you want it in this PR rather than the follow-up.

PR #5255, opened 2026-07-21
[Bugfix] Support GQA/MQA in the ring attention SDPA path

linzhenpl07:
### Problem

The Ring attention SDPA path crashes when num_kv_heads != num_heads because the aten SDPA operators used for LSE do not expose enable_gqa.

### Fix

Expand K/V locally immediately before the aten SDPA call. Ring communication still exchanges compressed K/V tensors, so communication volume is unchanged.

### Tests

The explicit TORCH_SDPA Ring GQA/MQA regression passed on 4 GPUs with maximum absolute error 0.015625 (tolerance 0.05).

The Hybrid Ulysses+Ring positive case is intentionally configured with num_heads=4, ulysses_degree=2, and ring_degree=2; K/V heads divide evenly and no K/V-head padding is needed. It was re-run on 4 GPUs and passed (1 passed, 25.30 s).

num_heads=3 is not a valid positive case for this configuration: 3 K/V heads cannot divide evenly across 2 Ulysses ranks, so K/V head padding would be required. Hybrid advanced_uaa intentionally rejects that case with the padding guard. The padded-MQA guard test passes separately.

### Benchmark note

The exploratory microbenchmark covers Ring GQA/MQA timing cells across sequence lengths, SP degrees, and link modes. Timing results are supplementary; the numerical regression and explicit padding-guard behavior are the acceptance criteria.

PR #5716, opened 2026-08-03
[Bugfix][Diffusion] Keep the GQA ratio when padding Ulysses heads

Xenoryn:
## Purpose

**What broke.** When Ulysses shards KV heads across a world size that does not evenly divide them, the current path pads Q and KV independently, each rounded up to `world_size`. This breaks the GQA `Q // KV` ratio and the head-dim `all_to_all` reshape either crashes or silently produces wrong attention.

**Repro.** Boogu-Image (`28Q / 7KV`) at SP=2: KV rounds up to `8`, Q rounds up to `28`, giving `28Q / 8KV = 3.5` per rank — not a valid GQA shape.

**Root cause.** `ulysses.py` pads Q and KV in isolation. The `Q // KV` invariant is never restated post-padding.

**Fix.** Pad KV first, then derive the padded Q count as `padded_kv * (Q // KV)`, so the ratio survives the split (`28/7` → `32/8`, i.e. `16/4` per rank). Add a validation guard for `Q % KV == 0` and matching K/V head counts, since the whole scheme assumes a well-formed GQA shape upstream. MHA is unaffected — the derived Q padding degenerates to the previous ceil for `Q == KV`.

The same coordinated rule is applied to `AttentionMetadata.joint_*`. For hybrid Ulysses+Ring, the PyTorch Ring kernel expands per-rank KV heads to match Q before calling the low-level fused SDPA operators, which do not perform GQA expansion themselves.

## Series context

> Series: **1/3 (this PR)** → #5717 → #5718.

This series ends with BOOGU-Image sequence-parallel support. BOOGU is a `28Q / 7KV` GQA model, so without this fix it cannot enable Ulysses at `SP=2` at all — the head-dim reshape hits the `3.5`-heads-per-rank crash on the very first attention call. This PR lands on its own merits (any GQA model with `KV % world_size != 0` hits it), but flagging the dependency for reviewer context.

## Test Plan

- `test_uaa_gqa_head_padding_preserves_the_query_to_kv_ratio` — CPU unit, asserts `(padded_Q, padded_K, padded_V) == (32, 8, 8)` and that the per-rank shape is a valid GQA.
- `test_uaa_rejects_head_counts_that_are_not_a_gqa_shape` — rejects `10Q/4KV` with a clear `ValueError`.
- `test_ulysses_uaa_gqa_matches_baseline` — multi-GPU parity, parametrized over `(sp, Q, KV) ∈ {(2, 28, 7), (2, 6, 3), (4, 12, 3)}`.
- `test_uaa_joint_gqa_head_padding_preserves_the_query_to_kv_ratio` — covers the same coordinated padding for `AttentionMetadata.joint_*` tensors.
- `test_uaa_rejects_joint_head_counts_that_are_not_a_gqa_shape` — rejects malformed joint GQA shapes.
- `test_ulysses_uaa_joint_gqa_matches_baseline` — compares SP=2 joint attention with a single-GPU baseline for `28Q/7KV` and `6Q/3KV`.
- `test_ulysses_uaa_hybrid_ring_matches_baseline` — covers hybrid Ulysses+Ring for MHA, `28Q/7KV` GQA, and `28Q/7KV` GQA with a joint stream.
- Mutation-verified: reverting to independent Q/K ceil padding fails the corresponding main and joint regression tests.

Command:

```bash
pytest tests/diffusion/attention/test_ulysses_uaa.py \
  -k 'gqa_head_padding or rejects_head_counts or joint_head_counts or gqa_matches_baseline' -v
```

**vLLM Version:** 0.24.0
**vLLM-Omni Commit:** 2358911d

## Test Result

```
tests/diffusion/attention/test_ulysses_uaa.py::test_uaa_gqa_head_padding_preserves_the_query_to_kv_ratio PASSED
tests/diffusion/attention/test_ulysses_uaa.py::test_uaa_rejects_head_counts_that_are_not_a_gqa_shape PASSED
tests/diffusion/attention/test_ulysses_uaa.py::test_ulysses_uaa_gqa_matches_baseline[2-6-28-7] PASSED
tests/diffusion/attention/test_ulysses_uaa.py::test_ulysses_uaa_gqa_matches_baseline[2-5-6-3] PASSED
tests/diffusion/attention/test_ulysses_uaa.py::test_ulysses_uaa_gqa_matches_baseline[4-8-12-3] PASSED

===== 5 passed in 42.7s =====
```

### DIFF
--- a/tests/diffusion/attention/test_ulysses_uaa.py
+++ b/tests/diffusion/attention/test_ulysses_uaa.py
@@ -29,6 +29,86 @@ pytestmark = [pytest.mark.core_model, pytest.mark.diffusion, pytest.mark.cpu]
 
 @pytest.mark.core_model
 @pytest.mark.cpu
+def test_uaa_gqa_head_padding_preserves_the_query_to_kv_ratio(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """Q must be padded to a multiple of the *KV* padding, not of world_size.
+
+    Rounding each up independently is the obvious implementation and is wrong:
+    28Q/7KV at SP2 would become 28/8, i.e. 14/4 per rank, whose 3.5 ratio is not
+    a valid GQA shape. Deriving Q from the padded KV count gives 32/8 -> 16/4.
+    """
+    from vllm_omni.diffusion.attention.parallel import ulysses
+
+    class FakeGroup:
+        ulysses_group = object()
+        ulysses_world_size = 2
+        ulysses_rank = 0
+        ring_world_size = 1
+
+    observed = []
+
+    def fake_all_to_all(pg, tensor, **kwargs):
+        observed.append(kwargs["padded_head_cnt"])
+        return tensor, tensor.shape[2]
+
+    monkeypatch.setattr(ulysses, "_ulysses_all_to_all_any_qkv", fake_all_to_all)
+    monkeypatch.setattr(ulysses, "get_ulysses_mode", lambda **kwargs: "advanced_uaa")
+    monkeypatch.setattr(ulysses, "_all_gather_int", lambda *args, **kwargs: [3, 3])
+    strategy = ulysses.UlyssesParallelAttention(
+        FakeGroup(),
+        scatter_idx=2,
+        gather_idx=1,
+        use_sync=False,
+    )
+
+    with set_forward_context():
+        strategy.pre_attention(
+            torch.zeros(1, 3, 28, 4),
+            torch.zeros(1, 3, 7, 4),
+            torch.zeros(1, 3, 7, 4),
+            None,
+        )
+
+    q_padded, k_padded, v_padded = observed
+    assert (q_padded, k_padded, v_padded) == (32, 8, 8)
+    # The per-rank shape must still be a valid GQA shape.
+    assert (q_padded // 2) % (k_padded // 2) == 0
+
+
+@pytest.mark.core_model
+@pytest.mark.cpu
+def test_uaa_rejects_head_counts_that_are_not_a_gqa_shape(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    from vllm_omni.diffusion.attention.parallel import ulysses
+
+    class FakeGroup:
+        ulysses_group = object()
+        ulysses_world_size = 2
+        ulysses_rank = 0
+        ring_world_size = 1
+
+    monkeypatch.setattr(ulysses, "get_ulysses_mode", lambda **kwargs: "advanced_uaa")
+    monkeypatch.setattr(ulysses, "_all_gather_int", lambda *args, **kwargs: [3, 3])
+    strategy = ulysses.UlyssesParallelAttention(
+        FakeGroup(),
+        scatter_idx=2,
+        gather_idx=1,
+        use_sync=False,
+    )
+
+    with set_forward_context(), pytest.raises(ValueError, match="multiple of KV heads"):
+        strategy.pre_attention(
+            torch.zeros(1, 3, 10, 4),
+            torch.zeros(1, 3, 4, 4),
+            torch.zeros(1, 3, 4, 4),
+            None,
+        )
+
+
+@pytest.mark.core_model
+@pytest.mark.cpu
 def test_advanced_uaa_hybrid_rejects_padded_mqa_heads(monkeypatch) -> None:
     monkeypatch.setattr(
         "vllm_omni.diffusion.attention.parallel.ulysses.get_ulysses_mode",
@@ -67,6 +147,7 @@ def _run_attention_case(
     ring_degree: int = 1,
     split_sizes: list[int] | None = None,
     sdp_kernel_mode: str = "math",
+    num_kv_heads: int | None = None,
 ) -> None:
     device = torch.device(f"{current_omni_platform.device_type}:{local_rank}")
     current_omni_platform.set_device(device)
@@ -102,6 +183,7 @@ def _run_attention_case(
             head_size=head_size,
             causal=False,
             softmax_scale=1.0 / (head_size**0.5),
+            num_kv_heads=num_kv_heads,
         ).to(device=device, dtype=torch.float32)
 
         with np.load(input_file, allow_pickle=False) as payload:
@@ -327,6 +409,93 @@ def test_ulysses_uaa_hybrid_ring_matches_baseline() -> None:
         # the SDPA kernel path used by Ring attention. Use a looser tolerance to
         # keep the test stable across GPUs/kernels while still catching regressions.
         torch.testing.assert_close(sp_t, baseline_t, atol=5e-4, rtol=5e-4)
+    finally:
+        for path in (input_file, baseline_file, sp_file):
+            try:
+                os.remove(path)
+            except OSError:
+                pass
+
+
+@pytest.mark.parametrize(
+    "sp_world_size,seq_len,num_heads,num_kv_heads",
+    [
+        (2, 6, 28, 7),  # BOOGU-like GQA: neither Q nor KV divisible by P=2
+        (2, 5, 6, 3),  # KV divisible by P, Q is not
+        (4, 8, 12, 3),  # KV not divisible by P=4
+    ],
+)
+def test_ulysses_uaa_gqa_matches_baseline(
+    sp_world_size: int,
+    seq_len: int,
+    num_heads: int,
+    num_kv_heads: int,
+) -> None:
+    """End-to-end GQA under UAA: SP output must match the unsharded output."""
+    if current_omni_platform.get_device_count() < sp_world_size:
+        pytest.skip(f"Test requires {sp_world_size} GPUs")
+
+    batch_size = 2
+    head_size = 8
+
+    base_init_method = get_distributed_init_method()
+    sp_init_method = get_distributed_init_method()
+
+    with tempfile.NamedTemporaryFile(delete=False, suffix=".npz") as f_in:
+        input_file = f_in.name
+    with tempfile.NamedTemporaryFile(delete=False, suffix=".npy") as f_base:
+        baseline_file = f_base.name
+    with tempfile.NamedTemporaryFile(delete=False, suffix=".npy") as f_sp:
+        sp_file = f_sp.name
+
+    try:
+        torch.manual_seed(0)
+        q = torch.randn(batch_size, seq_len, num_heads, head_size, dtype=torch.float32)
+        k = torch.randn(batch_size, seq_len, num_kv_heads, head_size, dtype=torch.float32)
+        v = torch.randn(batch_size, seq_len, num_kv_heads, head_size, dtype=torch.float32)
+        np.savez(input_file, q=q.numpy(), k=k.numpy(), v=v.numpy())
+
+        torch.multiprocessing.spawn(
+            _run_attention_case,
+            args=(
+                1,
+                base_init_method,
+                input_file,
+                baseline_file,
+                num_heads,
+                head_size,
+                1,
+                "strict",
+                1,
+                None,
+                "math",
+                num_kv_heads,
+            ),
+            nprocs=1,
+        )
+        torch.multiprocessing.spawn(
+            _run_attention_case,
+            args=(
+                sp_world_size,
+                sp_init_method,
+                input_file,
+                sp_file,
+                num_heads,
+                head_size,
+                sp_world_size,
+                "advanced_uaa",
+                1,
+                None,
+                "math",
+                num_kv_heads,
+            ),
+            nprocs=sp_world_size,
+        )
+
+        baseline_t = torch.from_numpy(np.load(baseline_file, allow_pickle=False))
+        sp_t = torch.from_numpy(np.load(sp_file, allow_pickle=False))
+        assert baseline_t.shape == sp_t.shape
+        torch.testing.assert_close(sp_t, baseline_t, atol=1e-5, rtol=1e-5)
     finally:
         for path in (input_file, baseline_file, sp_file):
             try:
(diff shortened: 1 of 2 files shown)


vllm_omni/diffusion/attention/parallel/ulysses.py
```diff
@@ -311,11 +317,43 @@ def pre_attention(
...
+            # 14/4, which is not a valid GQA shape).
+            query_head_cnt = int(query.shape[2])
+            kv_head_cnt = int(key.shape[2])
+            if key.shape[2] != value.shape[2] or query_head_cnt % kv_head_cnt != 0:
+                raise ValueError(
+                    "Ulysses GQA requires equal K/V head counts and query heads "
+                    f"to be a multiple of KV heads, got Q={query_head_cnt}, "
+                    f"K={kv_head_cnt}, V={int(value.shape[2])}."
+                )
+            padded_kv_head_cnt = _ceil_div(kv_head_cnt, ulysses_world_size) * ulysses_world_size
```
hsliuustc0106, line 374: `AttentionMetadata.joint_*` still uses independent head padding. For 28Q/7KV at SP=2, the main tensors now become 16Q/4KV per rank, but the joint tensors remain 14Q/3KV, so the concat below fails with a head-size mismatch. Please apply the same coordinated GQA padding to joint Q/K/V and add a regression test, or explicitly reject/document joint GQA and narrow the “any GQA model” claim.

Xenoryn: Thanks for catching this!

I’ve pushed a follow-up that applies the same GQA-preserving padding rule to the `AttentionMetadata.joint_*` path: pad joint KV to a multiple of `world_size`, then derive joint Q as `padded_kv * (Q // KV)`. It also now raises a clear ValueError when joint Q isn’t a multiple of joint KV.

Validated on 2× H800:
- Two CPU unit tests covering `28Q/7KV` at SP=2 → `16Q/4KV` per rank, plus rejection of invalid joint GQA shapes.
- GPU parity against the single-GPU baseline for both `28Q/7KV` (BOOGU-like) and `6Q/3KV`.

vllm_omni/diffusion/attention/parallel/ulysses.py
```diff
@@ -311,11 +340,43 @@ def pre_attention(
...
+
             query, orig_head_cnt = _ulysses_all_to_all_any_qkv(
-                self._ulysses_pg, query, seq_lens=seq_lens, use_sync=self._use_sync
+                self._ulysses_pg,
+                query,
+                seq_lens=seq_lens,
+                use_sync=self._use_sync,
+                padded_head_cnt=padded_query_head_cnt,
+            )
+            key, _ = _ulysses_all_to_all_any_qkv(
```
hsliuustc0106, line 384: Hybrid Ulysses+Ring still breaks for GQA/MQA. For 28Q/7KV at U=2, this produces 16 Q / 4 KV heads per rank, but `ring_degree > 1` routes FP32/no-Flash runs to `ring_pytorch_attn`, whose low-level `_scaled_dot_product_efficient_attention` requires equal head counts and has no GQA expansion; I reproduced `Expected query.size(2) == key.size(2)`. The added GQA tests all use `ring_degree=1`. Please add GQA handling to the Ring kernel (or reject this topology) and cover the hybrid case.

Xenoryn: Fixed in the latest update — the Ring PyTorch path now expands GQA K/V heads before calling SDPA.
The hybrid Ulysses+Ring parity test now covers MHA, `28Q/7KV` GQA, and GQA with a joint stream.

Xenoryn:
Hi @hsliuustc0106, thanks again for the previous reviews. I’ve addressed both follow-up issues, including joint GQA padding and hybrid Ulysses+Ring coverage, and the current CI checks are green.
When you have time, could you please take another look and let me know whether #5716 can move forward? Since #5717 and #5718 are stacked on top of it, if you’d prefer, we can also start reviewing those follow-ups in order while #5716 awaits final approval. Thanks!

tests/diffusion/attention/test_ulysses_uaa.py
```diff
@@ -259,7 +520,17 @@ def test_ulysses_uaa_matches_baseline(sp_world_size: int, seq_len: int, num_head
                 pass
 
 
-def test_ulysses_uaa_hybrid_ring_matches_baseline() -> None:
+@pytest.mark.parametrize(
+    "num_heads,num_kv_heads,joint_len",
+    [
+        (3, None, None),  # MHA baseline (head_cnt not divisible by ulysses_degree=2)
+        (28, 7, None),  # GQA: exercises Ring pytorch_attn_forward KV expansion (Q!=K heads)
```
hsliuustc0106, line 597: [P1] The Ring KV-expansion fix referenced here no longer exists at the current head — the merge from main dropped it.

Commit 2358911 added the GQA KV expansion to `pytorch_attn_forward` (`ring_kernels.py`: expand K/V via repeat_interleave when `k.shape[1] != q.shape[1]` before the SDPA call). The final merge commit (ec3741d) resolved `ring_kernels.py` back to main's version: at this head the file is byte-identical to main, no expansion exists anywhere in the tree, and this PR's diff no longer touches that file — only this test (whose comment still claims it exercises the expansion) survived. On a 4-GPU machine the `(28, 7, None)` and `(28, 7, 5)` parity rows should fail against the code as-is.

Since the PR is also CONFLICTING with main and needs a rebase anyway: please re-apply the 2358911 `ring_kernels.py` hunk in the rebase, and double-check the resolution diff afterwards so the second fix doesn't get lost again.

hsliuustc0106 (changes requested):
The main-path and joint GQA-padding fixes are verified correct at head with good test coverage, but the P1 inline blocks: the merge from main silently dropped the Ring KV-expansion hunk from `ring_kernels.py` (commit 2358911's production change is gone — the file is byte-identical to main at this head, only the test survived). Since the PR is CONFLICTING with main and needs a rebase anyway, please re-apply that hunk in the rebase and double-check the resolution diff so it doesn't get lost again.

Xenoryn:
> The main-path and joint GQA-padding fixes are verified correct at head with good test coverage, but the P1 inline blocks: the merge from main silently dropped the Ring KV-expansion hunk from `ring_kernels.py` (commit [2358911](https://github.com/vllm-project/vllm-omni/commit/2358911d8256229294a0b68862b05b4be35b942d)'s production change is gone — the file is byte-identical to main at this head, only the test survived). Since the PR is CONFLICTING with main and needs a rebase anyway, please re-apply that hunk in the rebase and double-check the resolution diff so it doesn't get lost again.

@hsliuustc0106  Thanks! I've rebased the branch onto the latest main and reorganized it into 3 self-contained commits. To address your comments:

**The Ring-side GQA expansion is not dropped**
#5255 (24ede9e2) on main already folds the GQA/MQA `repeat_interleave` expansion into Ring's `pytorch_attn_forward` (`ring_kernels.py` lines 86-97), which is equivalent to the ring patch this branch used to carry. So it's correct that the rebased PR diff no longer touches `ring_kernels.py` — re-applying our old hunk would just be duplicate code.

**The real conflict is the guard that #5255 added to `ulysses.py`**
#5255 added an unconditional rejection under `advanced_uaa`: when `ring_world_size > 1`, any K/V head count not divisible by `ulysses_degree` raises an error ("does not support K/V head padding"). This PR removes the need for that guard. Non-GQA layouts (Q not a multiple of KV, K≠V) are still rejected with an explicit error. The joint stream follows the same rule.

**Tested on 4×H800, 17/17 green**, including the exact case the guard rejected: hybrid Ulysses×2 + Ring×2 with `num_heads=3`, plus 28Q/7KV GQA and 28Q/7KV + joint text stream, all matching the no-SP baseline.

Commits: (1) derived-Q padding for the main path, (2) the same rule for joint heads, (3) guard removal + hybrid parity tests, with the reasoning recorded in the commit message.

tests/diffusion/attention/test_ulysses_uaa.py
hsliuustc0106: Retracting this — I was wrong. Main has carried the GQA/MQA KV expansion in `pytorch_attn_forward` since #5255 (landed 08-07, `ring_kernels.py:87-97`); when I claimed the file was byte-identical to main with no expansion I compared only a line window that stopped just above it. Your merge resolving `ring_kernels.py` to main's version was correct — 2358911 was redundant with what main already had — and the rebased head makes that explicit: the hybrid-rejection guard is gone because the derived-Q padding rule plus main's expansion genuinely support hybrid GQA. Apologies for the churn; approving the rebase.

hsliuustc0106 (approved):
Rebase verified and my blocking P1 retracted (see inline — it was my error: main's ring path has had the GQA/MQA expansion since #5255). The rebased head is the right design: GQA-preserving padding on the main and joint paths, the obsolete hybrid-rejection guard removed, and the parity tests extended to MHA/GQA/joint-GQA hybrid shapes. Nothing outstanding from my side.

hsliuustc0106:
@Bounty-hunter PTAL whether it will affect other models?

vllm_omni/diffusion/attention/parallel/ulysses.py
```diff
@@ -345,11 +362,43 @@ def pre_attention(
...
+            # 14/4, which is not a valid GQA shape).
+            query_head_cnt = int(query.shape[2])
+            kv_head_cnt = int(key.shape[2])
+            if key.shape[2] != value.shape[2] or query_head_cnt % kv_head_cnt != 0:
+                raise ValueError(
+                    "Ulysses GQA requires equal K/V head counts and query heads "
+                    f"to be a multiple of KV heads, got Q={query_head_cnt}, "
+                    f"K={kv_head_cnt}, V={int(value.shape[2])}."
+                )
+            padded_kv_head_cnt = _ceil_div(kv_head_cnt, ulysses_world_size) * ulysses_world_size
```
hsliuustc0106, line 443: The ratio-preserving padding is correct, but it can materially increase work and temporary memory for uneven-KV GQA/MQA: for example, 28Q/7KV at U=2 runs 32 padded Q heads (+14%), while 32Q/1KV at U=8 runs 256. Please provide same-hardware baseline-vs-PR latency and peak-VRAM measurements for representative GQA/MQA cases, and either document this tradeoff or add a padding-ratio guard/fallback.

hsliuustc0106 (commented):
Please add the requested GQA/MQA performance evidence and document or guard the padding overhead.

Xenoryn:
> Please add the requested GQA/MQA performance evidence and document or guard the padding overhead.

@hsliuustc0106 Sure — added both a warn-guard and same-hardware main-vs-PR numbers.

## 1. Warn-guard

Padding is a correctness requirement (Ulysses splits along the head dim, so `padded_KV % world_size == 0` and `padded_Q / padded_KV == original_Q/KV` are non-negotiable), so no cheap fallback. 

Instead: `_maybe_warn_uaa_pad_ratio` emits a one-shot warning when Q blow-up crosses **1.5×**, naming the shape and world size, so MQA-like blow-ups surface with an actionable suggestion (pick a `ulysses_degree` that divides `kv_heads`). Wired into both the main-tensor and joint-tensor padding sites; dedup key is `(Q, KV, U, tensor_label)`. 

New CPU test covers `28/7@U=2` (no-warn), `32/1@U=8` (warn), `16/2@U=4` (warn) plus once-only dedup. Comment at the padding site now spells out the MQA @ U=N → N× worst case.

## 2. Same-hardware main vs PR

4× H800, real `UlyssesParallelAttention` under `torchrun`. Bench + driver: https://gist.github.com/Xenoryn/71d91f30b4dd9c458cf46f473fdfb3ea. `batch=2, seq=4096, head_dim=128, dtype=fp32, SDPA=math, advanced_uaa`, 20 timed iters (median), 5 warmup.

| Case | Q/KV | U | main (ms / MiB) | PR (ms / MiB) | Verdict |
|---|---|---:|---|---|---|
| MHA aligned | 16/16 | 2 | 6.51 / 2784 | 6.55 / 2784 | no regression (+0.6%) |
| GQA aligned | 32/8 | 2 | 12.05 / 5280 | 12.03 / 5280 | no regression (−0.1%) |
| GQA aligned | 32/8 | 4 | 6.48 / 2752 | 6.45 / 2752 | no regression (−0.5%) |
| **FLOPs-ref for 28/7 @ U=2** (per-rank 14Q/14KV) | 28/28 | 2 | 11.40 / 4848 | 11.34 / 4848 | no regression (−0.5%) |
| **GQA uneven (BOOGU-like)** | 28/7 | 2 | `ValueError` | **12.14 / 5236** | **+7.1% ms / +8.0% MiB vs FLOPs-ref** |
| GQA uneven | 28/7 | 4 | `ValueError` | 6.52 / 2718 | PR enables shape main rejects |
| **MQA worst-case** | 32/1 | 4 | 6.42 / 2674 | **24.04 / 9970** | **+275% ms / +273% MiB** |

Takeaways:

1. **Aligned shapes: zero regression** (rows 1–4, ≤ 0.6% noise).
2. **BOOGU-like 28Q/7KV @ U=2 costs +7% ms / +8% MiB** vs the FLOPs-equivalent aligned reference (28/28 @ U=2, same per-rank Q count of 14). Matches the theoretical 14 → 16 Q head inflation (+14.3%); actual is lower because launch overhead dilutes it. That's the real cost for the mainstream uneven-GQA case, and it is modest.
3. **MQA 32/1 @ U=4 is genuinely bad**: +275%. Note main **does not error** here — it silently zero-pads KV heads and produces incorrect output (3 of 4 padded KV heads are zeros still participating in the softmax). PR pays 3.75× to be correct. This is exactly the scenario the new warn-guard flags: `warning_once("... GQA padding inflates query heads 8.00x ...")` with a suggestion to pick a divisor-friendly `ulysses_degree`.

Aligned shapes untouched, uneven-GQA cost is ~7%, extreme MQA is surfaced by a warning rather than silently expensive. Happy to iterate on the threshold or wording.

hsliuustc0106 (approved):

PR #5717, opened 2026-08-03
[Perf][Diffusion] Skip the Ulysses length all-gather when shards are padded equal

Xenoryn:
## Purpose

`_ulysses_all_to_all_any_qkv` calls `_all_gather_int` on every attention invocation to discover each rank's sequence length. `_all_gather_int` is `@torch.compiler.disable`d, so every attention call in the compiled region pays a graph break, a host sync, and a small collective — per denoising step, per layer.

When shards come from the framework's `_shard_with_auto_pad` boundary, every rank is guaranteed to hold the same padded length, so the collective is redundant. This PR records that guarantee in `ForwardContext._sp_equal_pad_stack` (a stack, not a counter, so mixed auto_pad / manual regions still pop the right entry) and takes a fast path that reuses `query.shape[1]` as a `SymInt`, keeping `torch.compile(dynamic=True)` happy.

**Note on the missing `int(...)`.** On the fast path `query.shape[1]` is deliberately not materialized. Under `torch.compile(dynamic=True)` it is a SymInt; calling `int()` would (a) force a host sync, (b) break the graph, and (c) specialize the compiled artifact on this seq_len, triggering a recompile whenever the resolution changes. The slow path keeps `int(...)` only because `_all_gather_int` is `@torch.compiler.disable`d — the graph break is already paid, so there is nothing left to preserve.

## Series context

> Depends on #5716. Follow-up: #5718. Until #5716 merges, please review only commit `f5772b59`; this draft's cumulative diff also contains the prerequisite.

This is 2/3 of a series that ends with BOOGU-Image sequence-parallel support. BOOGU can only run in `advanced_uaa` mode (its three shard groups are not divisibility-friendly), so BOOGU's SP=2 latency carries this PR's full benefit — without it the length-gather graph break lands in every attention call on every step. This PR lands on its own merits (any UAA-mode model benefits), but flagging the dependency for reviewer context.

## Test Plan

- `test_uaa_equal_rank_contract_skips_length_gather` — asserts the fast path fires and no gather is issued.
- `test_uaa_without_contract_keeps_dynamic_length_gather` — the manual-shard path still gathers.
- `test_uaa_mixed_boundaries_keep_the_dynamic_length_gather` — a manual region inside an auto_pad region correctly preserves the slow path.
- `TestEqualPadContract` — verifies auto-pad/manual split contracts and that gather pops only its own boundary.
- Mutation-verified: `all(...)` → `any(...)` and `pop` → `clear` both fail their respective tests.

E2E measured with a custom driver (loads the pipeline once, 1 warmup + 3 timed iterations, per-iter mean reported):

```bash
python tests/diffusion/distributed/bench_qwen_image_sp.py \
       --model Qwen/Qwen-Image --ulysses-degree 2 \
       --ulysses-mode {strict,advanced_uaa} \
       --height 1024 --width 1024 --steps 20 --warmup 1 --iters 3
```

- **Hardware:** 2× H800.
- **Software:** torch 2.11.0+cu130, vLLM 0.24.0, driver 570.148.08.
- **Backend:** FLASH_ATTN + lazy regional `torch.compile(dynamic=True)` (production default; `DIFFUSION_ATTENTION_BACKEND` unset).

**vLLM Version:** 0.24.0
**vLLM-Omni Commit:** 885d68f9

## Test Result

Qwen-Image SP=2, `advanced_uaa`, 1024², 20 steps:

| config                         | mean/iter (3 runs) |
|--------------------------------|--------------------|
| `strict` (no UAA overhead)     | 1.971 s            |
| `advanced_uaa`, before this PR | 2.869 s            |
| `advanced_uaa`, after this PR  | **1.995 s**        |

**1.44× end-to-end speedup.** `advanced_uaa` now matches `strict` within noise (+1.2%), where before the length all-gather + graph break made it 45.6% slower. The whole overhead the UAA path was paying vs `strict` was the discovery collective, not anything intrinsic to the padding.

Focused unit-test command:

```bash
pytest \
  tests/diffusion/attention/test_ulysses_uaa.py::test_uaa_equal_rank_contract_skips_length_gather \
  tests/diffusion/attention/test_ulysses_uaa.py::test_uaa_without_contract_keeps_dynamic_length_gather \
  tests/diffusion/attention/test_ulysses_uaa.py::test_uaa_mixed_boundaries_keep_the_dynamic_length_gather \
  tests/diffusion/distributed/test_sp_plan_hooks.py::TestEqualPadContract -v
```

Previous focused unit-test result:

```
6 passed
```

### DIFF
--- a/tests/diffusion/attention/test_ulysses_uaa.py
+++ b/tests/diffusion/attention/test_ulysses_uaa.py
@@ -786,6 +786,187 @@ def test_ulysses_uaa_matches_baseline(
                 pass
 
 
+def _run_contract_skip_case(
+    local_rank: int,
+    world_size: int,
+    init_method: str,
+    input_file: str,
+    num_heads: int,
+    head_size: int,
+    num_kv_heads: int | None,
+) -> None:
+    """Per-rank worker: slow path vs contract fast path on real rank pairs.
+
+    Phase 1 leaves the equal-pad stack empty so the Ulysses length all-gather
+    really runs (counted); phase 2 pushes the split hook's auto_pad marker and
+    replaces _all_gather_int with a raising guard. With ring_degree=1 any call
+    to it is the Ulysses length gather, so the guard proves the skip, and the
+    per-rank output comparison proves the fast path is numerically equivalent
+    to the slow path it replaces.
+    """
+    from vllm_omni.diffusion.attention.parallel import ulysses as ulysses_mod
+
+    device = torch.device(f"{current_omni_platform.device_type}:{local_rank}")
+    current_omni_platform.set_device(device)
+
+    os.environ["DIFFUSION_ATTENTION_BACKEND"] = "TORCH_SDPA"
+
+    init_distributed_environment(world_size=world_size, rank=local_rank, distributed_init_method=init_method)
+    initialize_model_parallel(
+        data_parallel_size=1,
+        cfg_parallel_size=1,
+        sequence_parallel_size=world_size,
+        ulysses_degree=world_size,
+        ring_degree=1,
+        tensor_parallel_size=1,
+        pipeline_parallel_size=1,
+    )
+
+    parallel_config = DiffusionParallelConfig(
+        pipeline_parallel_size=1,
+        data_parallel_size=1,
+        tensor_parallel_size=1,
+        sequence_parallel_size=world_size,
+        ulysses_degree=world_size,
+        ring_degree=1,
+        cfg_parallel_size=1,
+        ulysses_mode="advanced_uaa",
+    )
+    od_config = OmniDiffusionConfig(model="test", dtype=torch.float32, parallel_config=parallel_config)
+
+    with set_forward_context(omni_diffusion_config=od_config), set_current_diffusion_config(od_config):
+        attn = Attention(
+            num_heads=num_heads,
+            head_size=head_size,
+            causal=False,
+            softmax_scale=1.0 / (head_size**0.5),
+            num_kv_heads=num_kv_heads,
+        ).to(device=device, dtype=torch.float32)
+
+        with np.load(input_file, allow_pickle=False) as payload:
+            q_full = torch.from_numpy(payload["q"]).to(device=device)
+            k_full = torch.from_numpy(payload["k"]).to(device=device)
+            v_full = torch.from_numpy(payload["v"]).to(device=device)
+
+        def make_shards() -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+            # Fresh contiguous shards for each phase; seq_len is divisible by
+            # world_size, so every rank gets the same local length.
+            return (
+                torch.tensor_split(q_full, world_size, dim=1)[local_rank].contiguous(),
+                torch.tensor_split(k_full, world_size, dim=1)[local_rank].contiguous(),
+                torch.tensor_split(v_full, world_size, dim=1)[local_rank].contiguous(),
+            )
+
+        # The Attention layer only enables SP communication when
+        # ForwardContext.sp_active is True; in production the split hook sets
+        # this (and pushes the auto_pad equal-length marker onto
+        # _sp_equal_pad_stack). Here we drive it directly.
+        ctx = get_forward_context()
+        ctx._sp_shard_depth = 1
+
+        def make_sdp_ctx() -> object:
+            # Fresh instance per phase: the @contextmanager instance releases
+            # its state on first exit and cannot be re-entered.
+            return torch.backends.cuda.sdp_kernel(enable_flash=False, enable_mem_efficient=False, enable_math=True)
+
+        # Phase 1: contract absent -> the length all-gather must really run.
+        real_gather = ulysses_mod._all_gather_int
+        gather_calls: list[object] = []
+
+        def counting_gather(pg: object, value: int, *, device: torch.device) -> list[int]:
+            gather_calls.append(pg)
+            return real_gather(pg, value, device=device)
+
+        ulysses_mod._all_gather_int = counting_gather
+        try:
+            q, k, v = make_shards()
+            with make_sdp_ctx():
+                out_slow = attn(q, k, v, attn_metadata=None).contiguous()
+        finally:
+            ulysses_mod._all_gather_int = real_gather
+        assert len(gather_calls) >= 1, "slow path must call the Ulysses length all-gather"
+
+        # Phase 2: replicate the split hook's auto_pad marker so the fast
+        # path fires; any _all_gather_int call in this phase is the Ulysses
+        # length gather and must not happen.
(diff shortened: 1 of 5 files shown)


Xenoryn:
> This PR appears to belong to: **docs/design/module/diffusion/parallelism.md**.
> 
> Module owners: @princepride @yuanheng-zhao @xuechendi @Bounty-hunter
> 
> @Xenoryn, please review your own changes and leave a short self-review comment describing what you checked. PRs without author self-review may not be assigned a reviewer.
> 

## Self-Review
**Change:** A single commit stacked on top of the final #5716. The core is recording the auto_pad equal-length guarantee on `ForwardContext`, letting `advanced_uaa` skip the Ulysses length all-gather.

**Correctness check:**
- The skip fires only under the contract, and only for the `Ulysses-group` gather
- The joint path never had a length gather, so no compatibility concern
- Mixed auto_pad/manual regions conservatively fall back to the dynamic gather via the `stack + all()`
- Callers without a `ForwardContext` fall back to the dynamic gather

Xenoryn:
Hi @david6666666 @hsliuustc0106, this is ready for review. Would appreciate your thoughts—thanks!

tests/diffusion/attention/test_ulysses_uaa.py
```diff
@@ -29,28 +29,413 @@
...
+        ulysses_rank = 0
+        ring_world_size = 1
+
+    observed = []
+
+    def fake_all_to_all(pg, tensor, **kwargs):
+        observed.append((kwargs["seq_lens"], kwargs["padded_head_cnt"]))
+        return tensor, tensor.shape[2]
+
+    monkeypatch.setattr(ulysses, "_ulysses_all_to_all_any_qkv", fake_all_to_all)
```
hsliuustc0106, line 49: This replaces the real UAA all-to-all, while the existing distributed parity test never enables `_sp_equal_pad_stack`; the optimized collective path therefore has no committed regression coverage. Please enable the contract in at least one multi-rank parity case and verify that `_all_gather_int` is skipped.

hsliuustc0106 (changes requested):
Please add reproducible performance/accuracy evidence before merge: the cited `tests/diffusion/distributed/bench_qwen_image_sp.py` is absent, the baseline SHA and run variance are missing, and there is no same-seed output or peak-VRAM comparison.

Validation performed:
- Commit `f5772b59` cherry-picked cleanly onto current #5716 head `4b6d949e`.
- 7 focused CPU tests passed on the restacked patch.
- A two-rank `torch.compile(dynamic=True)` Gloo probe completed two sequence lengths with one graph and no graph breaks (PyTorch 2.13).
- A real 2×L20X NCCL fast-path case matched the single-GPU strict baseline (`atol=rtol=1e-5`; 139.24 s).
- DCO, pre-commit, wheel builds, docs, and mergeability pass.

The claimed 1.44× H800 result remains unverified because the local hardware/software differ and the benchmark driver is unavailable. Please provide the driver, exact baseline/PR SHAs, mean ± stddev for at least 3 measured runs after warmup, same-seed output parity, and peak VRAM.

Xenoryn:
> Please add reproducible performance/accuracy evidence before merge: the cited `tests/diffusion/distributed/bench_qwen_image_sp.py` is absent, the baseline SHA and run variance are missing, and there is no same-seed output or peak-VRAM comparison.
> 
> Validation performed:
> 
> * Commit `f5772b59` cherry-picked cleanly onto current #5716 head `4b6d949e`.
> * 7 focused CPU tests passed on the restacked patch.
> * A two-rank `torch.compile(dynamic=True)` Gloo probe completed two sequence lengths with one graph and no graph breaks (PyTorch 2.13).
> * A real 2×L20X NCCL fast-path case matched the single-GPU strict baseline (`atol=rtol=1e-5`; 139.24 s).
> * DCO, pre-commit, wheel builds, docs, and mergeability pass.
> 
> The claimed 1.44× H800 result remains unverified because the local hardware/software differ and the benchmark driver is unavailable. Please provide the driver, exact baseline/PR SHAs, mean ± stddev for at least 3 measured runs after warmup, same-seed output parity, and peak VRAM.

@hsliuustc0106  Thanks for the detailed review — both points addressed below. 🚀

## 1. Multi-rank coverage (the inline comment)

Added `test_ulysses_uaa_contract_skips_length_gather_matches_slow_path` (74245c45): a real 2-rank NCCL case with the contract on, for MHA (3 heads) and GQA (28/7 → 32/8).

- Phase 1 (no contract): the real `_all_gather_int` runs on the Ulysses group — the call is counted, so the check can't pass vacuously.
- Phase 2 (contract on): the split hook's auto_pad marker goes on the stack, `_all_gather_int` becomes a raising guard, and the fast-path output is compared to phase 1 per rank.

Full file passes 23/23 on a 2×H800 pod (hybrid Ulysses+Ring parity included; the hybrid contract CPU test still pins that the ring check on `ring_group` runs).

## 2. Rerun perf + accuracy

> Baseline is origin/main `2052deef`; "after" = main + this PR. 2×H800, vllm 0.28 / torch 2.13 / FA3, regional `torch.compile(dynamic=True)`, SP=2, 1024², 20 steps, n=6 (2 sessions × 3 iters, fixed seeds, warmup excluded; mean = arithmetic mean, std = sample, n−1):

| config | mean ± std |
| --- | --- |
| strict (main) | 2.011 ± 0.024 s |
| advanced_uaa before (main) | 3.234 ± 0.030 s |
| advanced_uaa after (main + PR) | 2.012 ± 0.011 s |

- **1.608×** before→after, and after ≈ strict (1.001) — the length all-gather overhead is gone. ✅
- Same-seed outputs: strict ↔ after are **bitwise identical** (3/3 seeds). before ↔ after differ only in fine detail — visually equivalent, eyeball-confirmed. That's expected: the removed gather forced a host-sync graph break, and without it the compiled path matches strict again. 
- Peak VRAM: 61271 MiB (before/after) vs 60869 MiB (strict) — no regression.
- Driver: https://gist.github.com/Xenoryn/dd7a6ec91aba105f3dd05b2605b2d2d0 — I figured this one-off perf claim probably doesn't warrant a permanent in-repo harness, so I kept it as a gist for now; happy to commit it if you'd rather have it in the tree.

Visual check (seed 43, 1024²):

|strict (main) | advanced_uaa before (main)| advanced_uaa after (main + PR) |
| --- | --- | --- |
| v4-system-main--strict--seed43 | v4-system-main--before-advanced_uaa-gather--seed43 | v4-system-main5717--after-skip--seed43 |

hsliuustc0106:
@Bounty-hunter @wtomin PTAL

Xenoryn:
Rebased onto latest `main` now that #5716 has been merged; this PR now contains only the two remaining commits on top of it.

hsliuustc0106:
@Bounty-hunter PTAL

PR #5719, opened 2026-08-03
[Frontend] Add response.cancel to /v1/realtime for Qwen3-Omni

TheCodeWrangler:
> [!IMPORTANT]
> **Stacked on #5555 → #5565 → #5566 → #5655.** GitHub only allows `base: main` for a
>
> This PR's own change is **+170/-2 across 2 files** (1 commit).
> Best reviewed once #5655 has landed.

## Why

Closing the connection was the only way to stop server-side work, because
closing is what runs the engine-side abort in `_run_generation`'s `finally`.
That also discards everything the connection holds — tools, voice, instructions,
history — so a voice client whose user interrupts routinely paid a full
reconnect and reseed every time. For a barge-in-capable UI that is the *common*
case, not the exceptional one.

`response.cancel` (as in OpenAI's Realtime API, and already implemented on the
experimental duplex path) aborts the turn in flight and replies `response.done`
with status `"cancelled"`, keeping the session.

## How

**Order matters.** The engine request is aborted first via `AsyncOmni.abort`, so
the three stages stop even if the reader task takes a moment to unwind — the
request id is now retained on the connection for that. The reader task is *then*
cancelled too, because it may be parked in `_await_tool_results_and_continue`
waiting on a tool result rather than inside the result generator, where aborting
the engine alone would not reach it. Both paths are covered by tests.

A `_cancelling` flag keeps `_run_generation`'s terminal-event path quiet, so the
client sees one `response.done` instead of a spurious `response.audio.done` for a
turn that never finished. It is cleared in a `finally`: leaking it `True` would
silence the terminal event for every later turn on that connection, which a test
pins by making `engine.abort` raise.

Cancel also clears the turn's debris immediately — pending calls, stale tool
results, queued audio, the round budget. `start_generation` already does that,
but only on the *next* commit, so a cancel followed by a disconnect would strand
it.

Cancelling with nothing in flight returns an explicit `no_active_response` error
rather than silently succeeding, so a client bug is visible.

## No `conversation.item.truncate` needed

OpenAI needs truncate because its server owns the transcript, so after a barge-in
it must be told how much audio was actually heard. Here the **client** owns
history, so after a cancel it simply seeds the text it actually played. That
removes the need for item identity, `conversation.item.created` acks, and
truncation bookkeeping.

## Verification

Unit tests: 73 pass at this level of the stack (67 → 73).

Against a live Qwen3-Omni instance:

| | result |
|---|---|
| cancel with nothing in flight | `error: No active response to cancel` |
| cancel mid-turn | `response.done status=cancelled` in **~1 ms** |
| reuse the same socket afterwards | works; history seeded before the cancel still resolves |

That last row is the point of the change: previously the equivalent required
dropping the connection and reseeding.

### DIFF
--- a/tests/entrypoints/openai_api/test_qwen3_omni_realtime_websocket.py
+++ b/tests/entrypoints/openai_api/test_qwen3_omni_realtime_websocket.py
@@ -208,6 +208,37 @@ WEATHER_TOOL = {
 }
 
 
+# Two cities in one utterance, to make the model request two calls in one turn.
+TOOL_CALLING_TWO_CITY_PHRASE_TEXT = "What is the weather like in Boston and in Denver right now?"
+
+# Distinct results per city, so the final reply shows each result reached the call
+# that asked for it rather than both being collapsed into one <tool_response>.
+_TOOL_RESULT_BY_CITY = {
+    "boston": "sunny and 72 degrees",
+    "denver": "snowing and 28 degrees",
+}
+
+
+def _tool_result_for(arguments: str) -> str:
+    """Answer a get_weather call based on the city it asked about."""
+    city = (json.loads(arguments).get("city") or "").lower()
+    for key, result in _TOOL_RESULT_BY_CITY.items():
+        if key in city:
+            return result
+    raise AssertionError(f"Tool call asked about an unexpected city: {arguments!r}")
+
+
+def _two_city_tool_calling_pcm16_input() -> bytes:
+    syn = generate_synthetic_audio(
+        10,
+        1,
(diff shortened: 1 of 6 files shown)


Sy0307:
`response.cancel` emits `response.done` with `status=cancelled` but without the active `response_id`. With multiple in-flight responses, clients cannot correlate the terminal event or enforce exactly-one terminal per response. Please carry the cancelled response id through the cancel path and add cancel-after-created and cancel-while-tool-waiting correlation tests.

PR #5720, opened 2026-08-03
[Diffusion] Add Minimax-H3 modular pipeline support

Isotr0py:

## Purpose
- Actually, Minimax-H3's FL2VA and Ref2VA only differ from DiT's weight, so we can share other components like text_encoder and vae for them.
- This PR boardcast `task_type` from tts models to diffusion models, then we can use `task_type` to control pipeline initialization during server startup
- By default, Minimax-H3 will initialize both FL2VA and Ref2VA's DiTs now. User can still specify `--task-type fl2va` or `--task-type ref2va` to only initialize one DiT backbone to save VRAM in special case.

## Test Plan
- launch the server, then send the T2VA request and Ref2VA request one by one
```
vllm serve MiniMaxAI/MiniMax-H3 --omni --num-gpus 4 --usp 4 --ring 1 --port 8091
```
**T2VA request**
```
curl -sS \
  -o t2va.mp4 \
  -X POST http://127.0.0.1:8091/v1/videos/sync \
  -F 'prompt=A small sailboat gliding across a moonlit bay, with gentle waves and quiet night wind.' \
  -F width=512 \
  -F height=320 \
  -F fps=24 \
  -F num_inference_steps=20 \
  -F flow_shift=12 \
  -F seed=6610 \
  -F 'extra_params={"task":"t2va","duration":2.0,"audio_flow_shift":3.0}'
```
**Ref2VA request**
```
curl -sS \
  -o ref2va.mp4 \
  -X POST http://127.0.0.1:8091/v1/videos/sync \
  -F 'prompt=Continue the moonlit sailboat scene as it approaches glowing harbor lights, preserving the boat and ambient sound.' \
  -F width=512 \
  -F height=320 \
  -F fps=24 \
  -F num_inference_steps=20 \
  -F flow_shift=12 \
  -F seed=7711 \
  -F 'extra_params={"task":"ref2va","duration":2.0,"audio_flow_shift":3.0}' \
  -F 'input_references=@t2va.mp4;type=video/mp4'
```

## Test Result

**T2VA**

https://github.com/user-attachments/assets/5275f4bf-e80b-4d7a-8716-4c32077958af

**Ref2VA**

https://github.com/user-attachments/assets/f256df58-5c58-4554-a77e-9b48543182c6


vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
```diff
@@ -281,24 +336,64 @@ def __init__(
...
+            self.weights_sources.append(
+                DiffusersPipelineLoader.ComponentSource(
+                    model_or_path=str(ref2va_model_path),
+                    subfolder="transformer",
+                    revision=od_config.revision,
+                    prefix="transformers_ref.",
+                    fall_back_to_pt=False,
+                )
+            )
+            self._dit_modules.append("transformers_ref")
```
lishunyang12, line 609: [P1] This registers the second DiT, but Cache-DiT still enables and refreshes only `pipeline.transformer` (and its summary only knows `transformer_2`). With `cache_backend=cache_dit`, `task=ref2va` therefore runs `transformers_ref` without cache acceleration. Please extend the cache lifecycle to all `_dit_modules`, or explicitly reject this backend for combined H3, and add coverage for both task routes.

recipes/MiniMaxAI/MiniMax-H3.md
```diff
@@ -74,21 +81,26 @@ vllm serve "${MODEL}" \
...
-The best validated four-GPU configuration on four NVIDIA B300 GPUs is:
+For a combined service on four high-memory GPUs, use:
 
 - no CPU or layerwise offload;
 - Ulysses sequence parallelism degree 4;
 - native tiled VAE patch parallelism degree 4;
 - regional `torch.compile` for the repeated DiT blocks;
 - FlashAttention, with Ring and TP left at 1.
 
+Both DiTs remain resident in this no-offload configuration. If they do not fit,
```
lishunyang12, line 214: [P1] The new combined/no-offload command keeps both DiTs resident, but the only latency figures below were collected with one partition. Please add a combined-service measurement covering peak VRAM and warm latency/throughput for both `fl2va` and `ref2va` (plus output parity), or clearly mark this recipe path as unvalidated; otherwise the new default mode is not performance-qualified.

Isotr0py:
| Mode | Weights loading memory usage / GPU | Peak memory usage / GPU |
-- | -- | -- 
Combined | 148.27 GiB | 158.67 GiB |
FL2VA-only | 86.27 GiB | 91.41 GiB |
Ref2VA-only | 86.27 GiB | 97.40 GiB |

vllm_omni/diffusion/model_metadata.py
```diff
@@ -42,6 +42,10 @@ class DiffusionModelMetadata:
         max_multimodal_image_inputs=9,
         supports_mixed_reference_inputs=True,
     ),
+    "MiniMaxH3ModularPipeline": DiffusionModelMetadata(
+        supports_multimodal_inputs=True,
+        max_multimodal_image_inputs=1,
```
hsliuustc0106: [P1] Reuse the existing MiniMax-H3 capability metadata for the modular alias. `modular_model_index.json` resolves to `MiniMaxH3ModularPipeline`, which the registry maps to this same pipeline, but this entry advertises only one image and leaves mixed references disabled. The serving layer uses these fields before `forward`, so valid Ref2VA requests with multiple images or mixed image/video references are rejected before the pipeline's 9-image/mixed-input validation runs. Please set this alias to the same `max_multimodal_image_inputs=9` and `supports_mixed_reference_inputs=True` contract, and add an admission test.

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
```diff
@@ -579,13 +686,18 @@ def transformer_weights():
...
     def _resolve_task(
         self,
         requested: str | None,
         multi_modal_data: dict[str, Any],
     ) -> str:
         if requested is None:
-            if self.partition == "ref2va":
+            if multi_modal_data.get("video") is not None or multi_modal_data.get("audio") is not None:
                 requested = "ref2va"
             elif multi_modal_data.get("image") is not None:
```
hsliuustc0106, line 727: [P1] Preserve the Ref2VA partition's implicit task default. With `--task-type ref2va` (or a local Ref2VA checkpoint), `supported_tasks` is only `{ref2va}`; an image-only request that omits `extra_params.task` now infers `fl2va` and fails at the membership check below. The previous implementation defaulted any request on the Ref2VA partition to `ref2va`. Please retain that behavior for single-partition startup and add a regression test.

vllm_omni/entrypoints/cli/serve.py
```diff
@@ -213,9 +213,10 @@ def subparser_init(self, subparsers: argparse._SubParsersAction) -> TrackingArgu
             "--task-type",
             type=str,
             default=None,
-            choices=["CustomVoice", "VoiceDesign", "Base"],
-            help="Default task type for TTS models (CustomVoice, VoiceDesign, or Base). "
-            "If not specified, will be inferred from model path.",
+            help="Model-defined startup task type. The selected model validates "
```
hsliuustc0106, line 216: [P2] Restore model-specific validation for the existing TTS flag before removing `choices`. Qwen3-TTS consumes `task_type` from each request and does not validate this startup value, so `vllm serve <TTS-model> --task-type fl2va` now starts successfully but silently ignores the invalid value instead of failing at argument parsing. Please preserve the TTS choices via model-aware validation or validate the generic flag during model initialization.

hsliuustc0106 (commented):
Posting the three additional actionable findings from the review. The Cache-DiT and combined-service evidence findings are already covered by active maintainer threads.

vllm_omni/diffusion/model_metadata.py
lishunyang12: Fixed in c56eab27: the modular alias now advertises the same 9-image and mixed-reference admission contract as MiniMaxH3Pipeline, with regression coverage.

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
lishunyang12: Fixed in c56eab27: Ref2VA-only startup now defaults omitted tasks to ref2va before modality inference, with a regression test.

vllm_omni/entrypoints/cli/serve.py
lishunyang12: Fixed in c56eab27: Qwen3-TTS validates its startup --task-type during OmniModelConfig initialization, while H3 task selectors remain supported.

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
lishunyang12: Fixed in 639f6236: Cache-DiT now enables, refreshes, and summarizes every declared `_dit_modules` DiT, with combined two-DiT regression coverage.

lishunyang12 (approved):

recipes/MiniMaxAI/MiniMax-H3.md
lishunyang12: Updated in d0c183c7: the recipe now presents the combined service configuration without a separate performance qualification.

SamitHuang (dismissed):
a few code quality and convention issues that need to be addressed

tests/diffusion/models/minimax_h3/test_minimax_h3_contract.py
```diff
@@ -24,12 +25,325 @@ def test_pipeline_import_registry_and_component_discovery():
...
+        download_calls.append(kwargs)
+        return str(tmp_path)
+
+    monkeypatch.setattr(
+        pipeline_module,
+        "download_weights_from_hf_specific",
+        fake_download,
+    )
+
+    od_config = SimpleNamespace(
```
SamitHuang: Please avoid using `SimpleNamespace` to fake typed objects like `OmniDiffusionConfig`, `Qwen2TokenizerFast`, and `Qwen3VLProcessor`. Consider using `unittest.mock.Mock` or instantiating the real config objects to prevent misleading type hints and fragility.

vllm_omni/diffusion/cache/cachedit/backend.py
```diff
@@ -27,14 +27,18 @@
 CUSTOM_DIT_ENABLERS: dict[str, CacheDiTEnabler] = {}
 
 
+def _dit_module_names(pipeline: Any) -> tuple[str, ...]:
```
SamitHuang: Please avoid adding new `Any` type hints. Use more specific types or `typing.Protocol` if the exact type is dynamic.

vllm_omni/engine/async_omni_engine.py
```diff
@@ -1019,6 +1019,7 @@ def _create_default_diffusion_stage_cfg(kwargs: dict[str, Any]) -> list:
             "max_num_seqs": kwargs.get("max_num_seqs") or 1,
             "parallel_config": parallel_config,
             "model_class_name": kwargs.get("model_class_name", None),
+            "task_type": kwargs.get("task_type", None),
```
SamitHuang, line 1022: Please avoid adding new `kwargs.get` string lookups. If `task_type` is a known parameter, consider adding it explicitly to the method signature or a typed config object.

vllm_omni/diffusion/utils/hf_utils.py
```diff
@@ -7,6 +7,28 @@
...
+def get_diffusion_model_index(
+    model_name: str,
+    *,
+    revision: str | None = None,
+) -> dict | None:
+    """Read the first standard Diffusers pipeline index available."""
+    for filename in DIFFUSION_MODEL_INDEX_FILES:
+        try:
+            config = get_hf_file_to_dict(filename, model_name, revision=revision)
+        except Exception as exc:
```
SamitHuang: Please avoid broad `except Exception:` swallows. Catch specific exceptions like `OSError` or `json.JSONDecodeError` instead.

SamitHuang (commented):
suggested minor fixes

hsliuustc0106:
fix conflicts

vllm_omni/diffusion/cache/cachedit/backend.py
Isotr0py: Fix in https://github.com/vllm-project/vllm-omni/pull/5720/commits/b0131985c880fb6b0a74ce31545296932484c67c. Actually, I found `Any` is terribly widely used in this file to present diffusion pipeline. 🥵  

Just refresh all of them to use `SupportsComponentDiscovery`.

lishunyang12 (approved):

PR #5461, opened 2026-07-27
[Refactor][1/N Scheduler]Remove duplicated AR/generation scheduler plumbing and establish explicit shared lifecycle contracts.

R2-Y:
PLEASE FILL IN THE PR DESCRIPTION HERE.

Remove duplicated AR/generation scheduler plumbing and establish explicit shared lifecycle contracts.

## Purpose
second step of https://github.com/vllm-project/vllm-omni/issues/5279 

- Centralized Omni I/O initialization and finish cleanup.
- Added lossless upstream NewRequestData conversion.
- Preserved Omni payloads on generation fallback scheduling.
- Shared queue cleanup, KV stats/events, and scheduler stats handling.
- Consolidated repeated AR KV lifecycle mechanics.
- Centralized schedule preparation, queue restoration, and output postprocessing.
- Shared request-output, failed-KV-load, finished-set, stats, and event handling.
- Preserved AR-only synthetic abort behavior as an explicit policy.
- Introduced narrow full-payload coordinator and receive-handle names with compatibility aliases.
- Removed broad schedule-output exception suppression.
- Updated scheduler analysis and module ownership notes.

Fast-path Omni entries remain unchanged, while upstream entries retain every base dataclass field before Omni payloads are attached. Scheduler-specific scheduling and finished-ID semantics remain local.

## Test Plan

**vLLM Version:**
0.26.0
**vLLM-Omni Commit:**
0.26.0
## Test Result

- 83 focused tests passed.
- Qwen3-Omni three-stage async-chunk text/audio request: 1/1 succeeded.
- Generated 18.30 seconds of 24 kHz audio.
- Syntax, diff, test-marker, and IDE lint checks passed.
-Minicpm-o4.5 three-stage async-chunk text/audio request: 1/1 succeeded.
image

**BEFORE SUBMITTING:** read [CONTRIBUTING.md](https://github.com/vllm-project/vllm-omni/blob/main/CONTRIBUTING.md) and run the [precheck-pr skill](https://github.com/vllm-project/vllm-omni/blob/main/.claude/skills/precheck-pr/SKILL.md) with the code agent for a self-check against project conventions.
(anything written below this line will be removed by GitHub Actions)

PR #3774, opened 2026-05-20
[Bugfix] purge chunk-transfer zombies on every schedule tick to keep engine-core alive on aborts (fixes #3736)

abinggo:
## Summary

Fixes the `engine-core` crash described in #3736: under sustained client-abort pressure on a chunk-transfer pipeline (`OmniARScheduler` or `OmniGenerationScheduler` with `async_chunk` + `OmniChunkTransferAdapter`), an aborted request that was parked in `waiting_for_chunk_running_requests` becomes a zombie once `Scheduler._free_request` deletes its tracking entry. The adapter still holds a reference, and `restore_queues` blindly `extend`s it back onto `running_queue`. The next `schedule()` tick packs the zombie into `scheduled_cached_reqs`, and the worker's `_update_states` crashes with `KeyError` reading `self.requests[req_id]` — taking down every client on that stage.

## Fix

- Add `_purge_untracked_chunk_requests` on the adapter: in-place filter (`popleft` + append-live) over a deque, dropping entries whose `request_id` is not in `scheduler_requests` and reclaiming their receiver-side state via `cleanup_receiver`. Survivor order is preserved.
- Make `scheduler_requests` a **required, keyword-only** parameter on both `process_pending_chunks` and `restore_queues`. This is intentionally a hard break: `OmniGenerationScheduler:97-98` was already silently calling the unguarded form on `async_chunk` generation deployments, which would have left the bug live there too — required-kwarg means any future caller that forgets gets a `TypeError` at the call site, not a worker crash three ticks later.
- Run the purge **inside `restore_queues` as well**, not only in `process_pending_chunks`. `restore_queues` runs in the scheduler's `finally`-clause, so an abort that fires between the two calls (during `super().schedule()`) would otherwise reopen the same race window. The second purge closes it.
- Update all 5 production call sites (1 in `OmniARScheduler.schedule()`; in `OmniGenerationScheduler.schedule()`: 1 `process_pending_chunks` + 2 `restore_queues`) to pass `scheduler_requests=self.requests`.

The fix is conceptually the one-liner the issue suggests, but the issue body assumed the dormant zombie-purge guard already lived inside the adapter. It didn't on `main` — `process_pending_chunks` had a `(waiting_queue, running_queue)` signature with no purge logic. So the actual delta is the guard + the call-site rewires + the API hardening.

## Fixes

- #3736

## Why required keyword-only

[...]

PR #5721, opened 2026-08-03
[Bugfix][Core] Sweep skipped_waiting for aborted requests in OmniARScheduler

TheCodeWrangler:
## Purpose

`OmniARScheduler.schedule()` drops `FINISHED_ABORTED` requests before handing control to upstream, because upstream raises on that status instead of tolerating it. The sweep covered `waiting` and `running`. Upstream admits from **three** queues:

```python
# vllm/v1/core/sched/scheduler.py, Scheduler.schedule() — vLLM 0.26.0
669:  while (self.waiting or self.skipped_waiting) and token_budget > 0:
676:      request_queue = self._select_waiting_queue_for_scheduling()   # may return skipped_waiting
679:      request = request_queue.peek_request()
...
1011:     self.running.append(request)
1016:     if request.status == RequestStatus.WAITING:      ...
1018:     elif request.status == RequestStatus.PREEMPTED:  ...
1021:     else: raise RuntimeError(f"Invalid request status: {request.status}")
```

`skipped_waiting` is an admission source, not a holding area, and it persists across ticks (line 1057 merges each step's skipped set back into it). An aborted request parked there survives the sweep and raises on a later tick:

```
RuntimeError: Invalid request status: FINISHED_ABORTED
  vllm/v1/core/sched/scheduler.py in schedule
  vllm_omni/core/sched/omni_ar_scheduler.py in schedule
[Orchestrator] Orchestrator died unexpectedly. See logs above.
```

That kills the stage's engine core and takes the orchestrator with it — every in-flight session is lost. Observed under duplex load with 10 sessions closing together.

**The fix:** sweep `skipped_waiting` too.

### Why omni reaches this state and upstream does not

Upstream `finish_requests` does clean `skipped_waiting` — but only on its non-running branch:

```python
if request.status == RequestStatus.RUNNING:
    running_requests_to_remove.add(request)     # only self.running is filtered
else:
    waiting_requests_to_remove.append(request)  # waiting AND skipped_waiting
```

Omni can reach that branch with the wrong status. `OmniChunkTransferAdapter.finish_requests` runs first and restores `requests_origin_status[req_id]` onto `request.status`. That table is stamped `RUNNING` when a request is parked out of `self.running`, and `restore_queues` never clears it, so it goes stale across the admit transition — the same staleness already documented in `_realign_request_status_to_queues` (#3774).

If the request has since become `WAITING_FOR_STREAMING_REQ` (a duplex session awaiting its next input chunk) and been parked in `skipped_waiting`, the stale `RUNNING` sends `finish_requests` down the running branch, the waiting queues are never touched, and the request is stranded there with `FINISHED_ABORTED`. `_realign_request_status_to_queues` doesn't repair it either — it builds `waiting_ids` from `self.waiting` only. Stock vLLM never gets here because nothing rewrites `request.status` behind `finish_requests`.

Same shape as #5662 and #5672: omni keeps state alongside upstream's, and a path that updates one but not the other starves or crashes a stage.

### Also in this diff

Removal now goes through `RequestQueue.remove_requests()`. The old sweep called `queue.remove(req)`, which worked only because `FCFSRequestQueue` subclasses `deque` — `PriorityRequestQueue` has no `remove()`, so under `--scheduling-policy priority` the sweep raised `AttributeError` on the first abort. `remove_requests()` is implemented by both policies and is what upstream `finish_requests` and the rest of omni already use. Same statement as the primary fix, so it is fixed here rather than left as a landmine.

The sweep moves into `_drop_aborted_queued_requests()` so it is testable without a KV cache manager and a model config. Call-site behaviour is otherwise unchanged.

It stays on `OmniARScheduler` rather than moving to `OmniSchedulerMixin` alongside `_purge_finished_from_running`, because `OmniARScheduler.schedule()` is the only scheduler that sweeps today — hoisting it without a second caller would be an abstraction with one implementation, and giving `OmniGenerationScheduler` a caller is a behaviour change beyond this fix. Happy to move it if you'd rather it live with the shared lifecycle helpers from #5461.

### Out of scope

Each needs its own reproduction and PR:

- a retry loop around `super().schedule()` for the narrower race where a request is aborted *while* upstream is mid-`schedule()` (upstream appends to `self.running` before checking status, so no pre-sweep can prevent that one)
- teaching `_realign_request_status_to_queues` about `skipped_waiting`, which would fix the producer path rather than this symptom
- anything touching `active_stream_window` in shipped example configs

No functional overlap with #5662, which also adds to `schedule()` a few lines below this hunk.

## Test Plan

```bash
pytest -sv tests/core/sched/test_omni_ar_scheduler_aborted_queue_sweep.py -m "core_model and cpu"
pytest -q tests/core tests/engine tests/distributed tests/model_executor -m "core_model and cpu"
```

New tests, each parametrized over the FCFS and PRIORITY queue policies:

| Test | Pins |
|---|---|
| `..._drops_aborted_request_from_skipped_waiting` | the regression |
| `..._drops_aborted_requests_from_waiting_and_running` | the two queues already covered stay covered |
| `..._leaves_live_requests_intact` | no over-sweep: a live `WAITING_FOR_STREAMING_REQ` request in `skipped_waiting`, plus live `waiting` / `running` entries, survive |

`test_omni_ar_scheduler_unified_decode_deferral.py` needed `skipped_waiting` on its hand-built scheduler; the real `Scheduler.__init__` always creates it.

**vLLM Version:** 0.26.0 (torch 2.11.0+cu130, transformers 5.14.1)

**vLLM-Omni Commit:** 6fdc4ca62dc6080bc683552c6ec9e310e9c80867

## Test Result

New tests: **6 failed** with `upstream/main`'s version of `omni_ar_scheduler.py` in place, **6 passed** with the fix.

L1 (`core_model and cpu`) over `tests/core tests/engine tests/distributed tests/model_executor`, both sides on this rebase, same box and container:

```
6fdc4ca6 (base):  22 failed, 1627 passed, 8 skipped, 205 deselected in 270.88s
this branch:      22 failed, 1633 passed, 8 skipped, 205 deselected in 308.25s
```

Byte-identical failure sets — the 22 are pre-existing on the base commit in this environment (model-specific cosyvoice3 / step_audio2 / voxcpm2 tests plus one orchestrator test); none are in `tests/core` or `tests/distributed`. The 6 extra passes are the new tests.

For a quick look at the mechanism without running the suite, replaying the pre-fix sweep against the real `RequestQueue` types shows both defects directly (ad-hoc script, not checked in — the three tests above cover the same ground):

```
[fcfs]     after the sweep, skipped_waiting = [sess-aborted(FINISHED_ABORTED)]
[fcfs]     _select_waiting_queue_for_scheduling picked skipped_waiting, peek = sess-aborted(FINISHED_ABORTED)
[fcfs]     sweep of `waiting`: ok
[priority] after the sweep, skipped_waiting = [sess-aborted(FINISHED_ABORTED)]
[priority] _select_waiting_queue_for_scheduling picked skipped_waiting, peek = sess-aborted(FINISHED_ABORTED)
[priority] sweep of `waiting` raised AttributeError: 'PriorityRequestQueue' object has no attribute 'remove'
```

The aborted request survives the sweep and is the very request upstream picks next — one step from the `raise`.

### Not verified here

The original duplex crash reproduction bundled two other changes (a retry loop around `super().schedule()` and an `active_stream_window` config change), so its numbers don't isolate this one, and the live ramp has not been re-run with only this change. The mechanism and the queue-set gap are proven above; no end-to-end measurement is claimed.

Anyone reproducing live should know `active_stream_window: 1` **masks this crash** — it admits one stream at a time to the chunk-transfer adapter's active set, so aborts never overlap. Reproduce with `0`.

### DIFF
--- a/tests/core/sched/test_omni_ar_scheduler_aborted_queue_sweep.py
+++ b/tests/core/sched/test_omni_ar_scheduler_aborted_queue_sweep.py
@@ -1,0 +1,115 @@
+"""Tests for ``OmniARScheduler._drop_aborted_queued_requests``.
+
+``schedule()`` sweeps ``FINISHED_ABORTED`` requests out of the queues before
+handing control to upstream, because upstream ``Scheduler.schedule()`` raises
+``RuntimeError: Invalid request status: FINISHED_ABORTED`` for any request it
+admits in a finished state -- which kills the stage's engine core, not just
+the request.
+
+Regression for the sweep missing ``skipped_waiting``, the third queue
+upstream admits from. An aborted duplex session parked there was re-selected
+by ``_select_waiting_queue_for_scheduling`` on a later tick and crashed the
+stage.
+"""
+
+from __future__ import annotations
+
+import pytest
+from vllm.v1.core.sched.request_queue import SchedulingPolicy, create_request_queue
+from vllm.v1.request import RequestStatus
+
+from vllm_omni.core.sched.omni_ar_scheduler import OmniARScheduler
+
+pytestmark = [pytest.mark.core_model, pytest.mark.cpu]
+
+_POLICIES = [
+    pytest.param(SchedulingPolicy.FCFS, id="fcfs"),
+    pytest.param(SchedulingPolicy.PRIORITY, id="priority"),
+]
+
+
+class _StubRequest:
+    """Minimal ``Request`` stub with the surface the sweep exercises.
+
+    Carries the ``(priority, arrival_time)`` ordering ``PriorityRequestQueue``
+    relies on so the same stub works under both scheduling policies.
+    """
+
+    def __init__(self, request_id: str, status: RequestStatus, arrival_time: float = 0.0) -> None:
+        self.request_id = request_id
+        self.status = status
+        self.priority = 0
+        self.arrival_time = arrival_time
+
+    def __lt__(self, other: _StubRequest) -> bool:
+        return (self.priority, self.arrival_time) < (other.priority, other.arrival_time)
+
+
+def _make_scheduler(policy: SchedulingPolicy, *, waiting=(), skipped_waiting=(), running=()):
+    """Bare scheduler with just the surface the sweep reads."""
+    scheduler = OmniARScheduler.__new__(OmniARScheduler)
+    scheduler.waiting = create_request_queue(policy)
+    for req in waiting:
+        scheduler.waiting.add_request(req)
+    scheduler.skipped_waiting = create_request_queue(policy)
+    for req in skipped_waiting:
+        scheduler.skipped_waiting.add_request(req)
+    scheduler.running = list(running)
+    return scheduler
+
+
+@pytest.mark.parametrize("policy", _POLICIES)
+def test_sweep_drops_aborted_request_from_skipped_waiting(policy: SchedulingPolicy) -> None:
+    """The sweep must cover ``skipped_waiting``.
+
+    Upstream ``schedule()`` admits from ``skipped_waiting`` as readily as from
+    ``waiting``, so an aborted request left there is re-selected and raises.
+    """
+    aborted = _StubRequest("req-aborted", RequestStatus.FINISHED_ABORTED)
+    scheduler = _make_scheduler(policy, skipped_waiting=[aborted])
+
+    scheduler._drop_aborted_queued_requests()
+
+    assert list(scheduler.skipped_waiting) == []
+
+
+@pytest.mark.parametrize("policy", _POLICIES)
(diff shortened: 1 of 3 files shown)


Sy0307:
The sweep itself is correct, but the tests call `_drop_aborted_queued_requests()` directly. Please add one regression through `OmniARScheduler.schedule()` with a real `skipped_waiting` request, so the test proves the sweep runs before upstream queue selection and prevents `Invalid request status: FINISHED_ABORTED` for both FCFS and PRIORITY.

PR #5725, opened 2026-08-03
[Bugfix] Make diffusion output wait timeout configurable

Xunzhuo:
## Summary
- make the background diffusion D2H/shared-memory output wait configurable with `VLLM_OMNI_DIFFUSION_OUTPUT_TIMEOUT`
- fall back to the existing `VLLM_OMNI_VIDEO_SYNC_TIMEOUT` so long-video deployments need only one timeout unless they want separate control
- preserve the existing 30-second default and reject invalid, non-positive, or non-finite values at startup

The API-level video timeout currently does not cover the internal wait for an async output to finish D2H transfer and shared-memory packing. On a slower large-video path, model execution can finish successfully but the fixed 30-second internal wait cancels the result before the API receives it.

## Test plan
- [x] `18 passed` — `tests/diffusion/test_diffusion_engine_cleanup.py`
- [x] all pre-commit hooks pass for the changed files
- [x] tests cover dedicated override precedence, fallback to the video timeout, the unchanged default, and invalid timeout values
- [x] reproduced on a source-built ROCm environment with 209-frame 1344x768 MiniMax H3 output where model forward completed but output packing exceeded 30 seconds
- [x] patched path returned HTTP 200 for 209-frame 1344x768 and 2048x1088 video+audio outputs with `VLLM_OMNI_VIDEO_SYNC_TIMEOUT=1800/3600`

The default behavior is unchanged for deployments that do not set either environment variable. This is a platform-independent diffusion robustness fix.

### DIFF
--- a/tests/diffusion/test_diffusion_engine_cleanup.py
+++ b/tests/diffusion/test_diffusion_engine_cleanup.py
@@ -2,6 +2,7 @@
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
 
 import asyncio
+import concurrent.futures
 import queue
 import threading
 from types import SimpleNamespace
@@ -45,6 +46,15 @@ def _make_engine() -> DiffusionEngine:
     return engine
 
 
+def test_async_output_timeout_defaults_to_30_seconds(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    monkeypatch.delenv("VLLM_OMNI_DIFFUSION_OUTPUT_TIMEOUT", raising=False)
+    monkeypatch.delenv("VLLM_OMNI_VIDEO_SYNC_TIMEOUT", raising=False)
+
+    assert diffusion_engine_module._get_async_output_timeout() == 30.0
+
+
 def test_async_output_timeout_uses_video_sync_timeout_as_fallback(
     monkeypatch: pytest.MonkeyPatch,
 ) -> None:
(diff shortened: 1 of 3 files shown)


vllm_omni/diffusion/diffusion_engine.py
```diff
@@ -55,7 +57,31 @@
 
 logger = init_logger(__name__)
 
-_ASYNC_OUTPUT_TIMEOUT = 30.0  # seconds
+
+def _get_async_output_timeout() -> float:
+    """Resolve the timeout for background D2H/shared-memory output transfer."""
+    raw_timeout = os.environ.get(
```
hsliuustc0106, line 63: can you help update the env var docs as well?

tests/diffusion/test_diffusion_engine_cleanup.py
```diff
@@ -45,6 +45,35 @@ def _make_engine() -> DiffusionEngine:
     return engine
 
 
+def test_async_output_timeout_uses_video_sync_timeout_as_fallback(
```
hsliuustc0106, line 58: [P2] Add a regression test through the async-output consumer

These tests only call `_get_async_output_timeout()`, so they would still pass if `step_streaming()` kept using the old hard-coded 30-second value. Please add a focused test that exercises the `asyncio.wait_for(..., timeout=_ASYNC_OUTPUT_TIMEOUT)` path (or otherwise verifies the resolved value is consumed) so this bugfix is protected end to end.

Xunzhuo: Added consumer-path regressions in 2d8753ce. One drives `step_streaming()` through `asyncio.wait_for` and captures the configured 45.5-second timeout; another drives the synchronous future path and verifies 72 seconds is passed to `Future.result`. Timeout resolution now happens at engine initialization instead of module import. Full cleanup test file: 21 passed on MI300X.

vllm_omni/diffusion/diffusion_engine.py
```diff
@@ -55,7 +57,31 @@
...
+            "VLLM_OMNI_DIFFUSION_OUTPUT_TIMEOUT (or fallback "
+            "VLLM_OMNI_VIDEO_SYNC_TIMEOUT) must be a positive finite number, "
+            f"got {raw_timeout!r}."
+        ) from exc
+    if not math.isfinite(timeout) or timeout <= 0:
+        raise ValueError(
+            "VLLM_OMNI_DIFFUSION_OUTPUT_TIMEOUT (or fallback "
+            "VLLM_OMNI_VIDEO_SYNC_TIMEOUT) must be a positive finite number, "
+            f"got {raw_timeout!r}."
+        )
```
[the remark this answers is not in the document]
Xunzhuo: Fixed in 2d8753ce by constructing the validation error once and reusing it for parse and finite/positive validation failures.

vllm_omni/diffusion/diffusion_engine.py
Xunzhuo: Updated the MiniMax H3 recipe in 2d8753ce. It now documents `VLLM_OMNI_DIFFUSION_OUTPUT_TIMEOUT`, its 30-second default, fallback to `VLLM_OMNI_VIDEO_SYNC_TIMEOUT`, the separate responsibilities of both variables, and positive-finite validation.

SamitHuang (approved):
LGTM

PR #5730, opened 2026-08-03
[Bugfix][MiniCPM-o] Fix the audio_embeds input path

eval-dev:
## Summary

Two defects made the OpenAI `audio_embeds` content part unusable with MiniCPM-o 4.5. Either one alone is fatal, so the path could never have worked end to end.

**1. Wrong modality.** `MiniCPMOAudioEmbeddingItems` registered itself under `modality="image"`, copy-pasted from `MiniCPMVImageEmbeddingItems` (the sibling video class already uses `"video"`). Every request failed with:

```
Modality 'image' not found. Available modalities: {'audio'}
```

**2. Placeholder sized by `s * h` instead of `s`.** `_get_prompt_updates` computed the length from `sum(map(len, single_audio_embeds))`. `MiniCPMOAudioEmbeddingInputs` declares `TensorShape("bn", "s", "h")`, so a single item is `(s, h)` and that expression iterates rows and sums the hidden size. A one-second slice of shape `(10, 4096)` requested **40960** placeholder tokens instead of 10, so any real request exceeded `max_model_len`:

```
The decoder prompt (length 123112) is longer than the maximum model length of 32768
```

## Verification

Reproduced and fixed against a live server. With both fixes, an interleaved `image_url` + `audio_embeds` request returns a coherent response that demonstrably attends to the audio content (it describes a background melody present only in the audio track), where before it could not be served at all.

## Tests

Adds `tests/model_executor/models/minicpmo_4_5/test_audio_embeds_input.py` — CPU-only, `core_model`, no GPU or checkpoint required. All five cases fail on the unfixed code and pass after.

The placeholder cases run through the real `_get_prompt_updates` path rather than recomputing the length independently, so they genuinely exercise the regression; one case asserts the count does not scale with hidden size, which is the precise failure mode of the old expression.

```
5 passed
```

### DIFF
--- a/tests/model_executor/models/minicpmo_4_5/test_audio_embeds_input.py
+++ b/tests/model_executor/models/minicpmo_4_5/test_audio_embeds_input.py
@@ -1,0 +1,138 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+"""Regression tests for the MiniCPM-o 4.5 ``audio_embeds`` input path.
+
+Two defects made the OpenAI ``audio_embeds`` content part unusable with this model:
+
+1. ``MiniCPMOAudioEmbeddingItems`` registered itself under ``modality="image"``
+   (copy-pasted from ``MiniCPMVImageEmbeddingItems``), so every request failed with
+   ``Modality 'image' not found. Available modalities: {'audio'}``.
+2. ``_get_prompt_updates`` sized the placeholder from
+   ``sum(map(len, single_audio_embeds))``. ``MiniCPMOAudioEmbeddingInputs`` declares
+   ``TensorShape("bn", "s", "h")``, so one item is ``(s, h)`` and that expression
+   iterates rows and sums the hidden size -- ``s * h`` instead of ``s``. A one-second
+   slice of shape ``(10, 4096)`` asked for 40960 placeholder tokens instead of 10,
+   pushing every real request past ``max_model_len``.
+"""
+
+from __future__ import annotations
+
+import math
+from types import SimpleNamespace
+
+import pytest
+import torch
+from vllm.multimodal.parse import MultiModalDataItems
+
+from vllm_omni.model_executor.models.minicpmo_4_5.minicpmo_4_5_omni_llm import (
+    MiniCPMO45OmniLLMMultiModalProcessor,
+    MiniCPMO45OmniLLMProcessingInfo,
+    MiniCPMOAudioEmbeddingItems,
+    _minicpmo_field_config,
+)
+
+pytestmark = [pytest.mark.core_model, pytest.mark.cpu]
+
+POOL_STEP = 5
+HIDDEN_SIZE = 4096
+#: Audio embedding vectors produced per second: 100 mel -> 50 cnn -> (50-5)//5+1.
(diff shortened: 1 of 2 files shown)


linyueqian:
This is reviewed and ready from my side, and the only thing standing between it and an approval is a stuck CI build. A rebase onto current main will clear it.

The AMD lane has never reported at this head. Build 11401 was created at 19:05 and has sat at three steps finished and seventeen `waiting_for_dependencies` ever since, with `state: started` and no `finished_at`. It is not queued behind other work: AMD builds created hours later for other PRs have run to completion in under ten minutes. That build is wedged on its own dependency graph and will not recover on its own, and because it is still in a running state, cycling the `ready` label does not create a replacement, it is deduplicated against the one already in flight. A new commit is what produces a fresh build.

Everything else at this head is in good shape. The general lane and NPU are both green. The Intel lane is red, but it never executed a test body: both Intel steps hard-failed during collection with a `ModuleNotFoundError` on a test module unrelated to this PR, exiting status 2. PR #6137 hit the identical failure in build 11399 within two minutes of yours, on a different missing module and from a completely disjoint file set, which puts that break in the XPU image rather than in either branch.

So: please rebase onto current main. That gets you a fresh AMD build, and assuming it comes back consistent with the other lanes I will approve immediately. Your change is confined to the MiniCPM-o `audio_embeds` input path and its regression test, with no device-specific or kernel code, so I do not expect ROCm to have an opinion about it, but I would rather have the lane actually report than approve around a gap.

linyueqian (approved):
Both fixes match the declared `audio_embeds` schema. The item now registers under the audio modality, and placeholder sizing uses the embedding sequence length instead of multiplying it by hidden size. The regression tests exercise both failures through the real prompt-update path, and I did not find a material issue in the diff.

Validation: static review of the two-file diff at `d20f9549` against its current merge base. No fork code was executed locally.

CI at this head. The general lane, build 14648, and NPU, build 6398, are green. Intel build 7803 is red but never ran a test body: both `INTEL Tests` and `Intel: XPU Test` hard-failed at collection with `ImportError while importing test module tests/e2e/accuracy/minimax_h3/test_minimax_h3_t2va_similarity.py`, `ModuleNotFoundError: No module named 'minimax_h3.test_minimax_h3_t2va_similarity'`, exiting status 2. This PR changes `minicpmo_4_5_omni_llm.py` and its regression test, so it cannot affect how that accuracy package resolves. PR #6137's Intel build 7801 failed the same two steps the same way two minutes earlier on a different module, `examples.offline_inference`, from a disjoint file set, which places the break in the XPU image rather than in either branch.

AMD build 11401 has since gone terminal, and it reports as failed without ever producing a code signal. It started at 19:06 on 2026-09-04 and closed at 20:06 on 2026-09-05, twenty-five hours later, with 17 of its 20 steps in the `errored` outcome and 3 passed. Not one step hard-failed, so no test body ran and no assertion was evaluated on ROCm. That is the same infrastructure-expiry shape as AMD build 11272, where 14 jobs expired before starting, rather than the ROCm import failures that show up as genuine hard failures. So the ROCm evidence for this change is absent rather than negative, and I am approving on the green general and NPU lanes plus static review. The change is confined to the MiniCPM-o `audio_embeds` input path with no device-specific or kernel code, so the ROCm risk is low, but the gap belongs on the record rather than glossed as a passing lane.

PR #5753, opened 2026-08-04
[Frontend] Derive capability tasks for stages and pin speech-to-text serving to text output

ieaves:
## Purpose

Multi-stage pipelines never advertise the `transcription` task because the orchestrator derives
`supported_tasks` from stage shape alone (e.g. `generate` from a comprehension stage, `speech` from an
audio-output stage), so a pipeline whose comprehension model implements `SupportsTranscription` —
(like Qwen3-Omni's thinker) never mounts `/v1/audio/transcriptions` or `/v1/audio/translations`,
even though `omni_init_app_state` wires the handlers when the task is present.

Three changes:

1. **Capability-derived tasks.** `_MODEL_CAPABILITY_TASKS` maps `ModelRegistry` capability probes to tasks; a task is served when a stage passing the entry's `eligible` predicate runs a model passing its `probe` (`_CapabilityTask.check`). Probe failures degrade to "unsupported" rather than failing engine boot. 

2. **Text-only pin for speech-to-text serving.** vLLM's speech-to-text serving calls `generate()` through the `EngineClient` protocol, which has no `output_modalities` parameter such that a request on a thinker+talker pipeline would also synthesize audio of its own transcript.

3. **Qwen3-Omni composite class implements `SupportsTranscription`.** vLLM-Omni registers its own `Qwen3OmniMoeForConditionalGeneration` for the checkpoint architecture, shadowing vLLM's thinker mapping, so serving resolves a class without the speech-to-text interface. The composite class now delegates `supported_languages`, `get_speech_to_text_config`, and `get_generation_prompt` to vLLM's thinker implementation.

## Test Plan

Unit (CPU):

pytest tests/engine/test_derive_supported_tasks.py tests/entrypoints/openai_api/test_text_only_engine_client.py

Serving (2×H100, before/after against the same command):

```
vllm serve Qwen/Qwen3-Omni-30B-A3B-Instruct --omni --port 8091

curl -X POST localhost:8091/v1/audio/transcriptions -F model=Qwen/Qwen3-Omni-30B-A3B-Instruct -F file=@test.wav
```

regression: text chat, and chat with modalities ["text","audio"] still produces talker audio

**vLLM Version:** 0.26.0

**vLLM-Omni Commit:** 0a063c5b (branch head, based on main @ 900a7f08)

## Test Result

Currently on `main`: startup logs `Supported tasks: {'generate', 'speech'}`; transcriptions and translations both 404.

After this PR: startup logs `Supported tasks: {'transcription', 'generate', 'speech'}`; `/v1/audio/transcriptions` returns 200 with a transcript of the test clip

Unit tests: 16 new tests pass. Full `tests/engine -m cpu` suite passes (360 tests).


vllm_omni/model_executor/models/qwen3_omni/qwen3_omni.py
```diff
@@ -45,6 +54,9 @@
     Qwen3OmniMoeThinkerMultiModalProcessor,
     Qwen3OmniMoeThinkerProcessingInfo,
 )
+
+if TYPE_CHECKING:
+    from vllm.config import SpeechToTextConfig, SpeechToTextParams
```
alex-jw-brooks: nit: this shouldn't need to be in type checking since we always have vLLM

vllm_omni/model_executor/models/qwen3_omni/qwen3_omni.py
```diff
@@ -86,6 +98,7 @@ class Qwen3OmniMoeForConditionalGeneration(
     CustomProcessMixin,
     SupportsMRoPE,
     SupportsRealtime,
+    SupportsTranscription,
```
alex-jw-brooks: I think it is outside of the scope of this PR, but I wonder if maybe we are going to need a conditional way of specifying whether or not some of these things are supported in the future 🤔 For this one, it shouldn't matter, but given that we already have thinker only models with no audio out, I could also imagine a situation where some of these are only applicable if optional stages are present, which could cause weird behaviors. Curious if you have any thoughts?

vllm_omni/engine/async_omni_engine.py
```diff
@@ -83,6 +88,69 @@
...
+        supported.add("generate")
+
+    if any(getattr(client, "final_output_type", None) == "audio" for client in clients):
+        supported.add("speech")
+
+    for rule in _MODEL_CAPABILITY_TASKS:
+        if any(rule.check(pool) for pool in stage_pools):
+            supported.add(rule.task)
+
+    return tuple(supported) if supported else ("generate",)
```
alex-jw-brooks: Can you add an `INFO` level log for what `_derive_supported_tasks` returns? I think this is a bit nonintuitive, and may also help us catch if more things need to be added to `_MODEL_CAPABILITY_TASKS` that aren't there yet

vllm_omni/engine/async_omni_engine.py
```diff
@@ -83,6 +88,69 @@
 _REQUEST_QUEUE_MAXSIZE = 256
 
 
+# vLLM's closed task set plus ``speech``, the composition-level task an
+# audio-output stage adds; this is the task universe an omni pipeline serves.
+OmniSupportedTask: TypeAlias = Literal[SupportedTask, "speech"]
+
+
+@dataclasses.dataclass(frozen=True)
```
alex-jw-brooks: I think this import is missing

lishunyang12: Import dataclass and use @dataclass here so the engine can load.

ieaves: Thanks @lishunyang12 fixed.

PR #5758, opened 2026-08-04
[Bugfix] Drain async diffusion outputs before worker shutdown

l-wave:
## Summary

Ensure the diffusion worker drains pending asynchronous output work and stops its background D2H thread before worker and device-context teardown.

## Problem

In request mode, the worker publishes `COMPUTE_DONE` after queueing an output. The background thread later waits for the device event, performs D2H and shared-memory packing, and publishes `OUTPUT_READY`.

The previous shutdown path set `_running` to false and immediately called `worker.shutdown()` and `context.term()`. A thread blocked in `Queue.get` was not awakened or joined. A thread processing one output could also exit before processing later queued outputs, leaving a published `COMPUTE_DONE` without its corresponding `OUTPUT_READY`.

## Changes

- Add an explicit stop sentinel to the asynchronous output queue.
- Place the sentinel after queued work so pending outputs are drained in FIFO order.
- Synchronize the D2H stream during thread teardown.
- Join the asynchronous output thread before worker and device-context cleanup.
- Clear the thread and queue references after shutdown.
- Preserve asynchronous packing error reporting.

## Regression coverage

`tests/diffusion/test_async_output_worker.py` covers:

- publishing `OUTPUT_READY` after background packing
- deterministic draining of two queued outputs while the first packing operation is blocked
- waiting for the background thread before shutdown returns
- shutdown when asynchronous state was not initialized

The pending-output test uses threading events rather than timing sleeps.

## Observed shutdown case

A single-card HunyuanImage3 request on Ascend 950PR completed and saved its image and latent trajectory. During close, the worker then exceeded the five-second process shutdown timeout and was force-terminated. The process subsequently aborted with `corrupted size vs. prev_size` and exit code 134.

The allocator abort can also be reproduced by importing `vllm_omni` and exiting without a diffusion request. This PR therefore addresses the concrete asynchronous-output teardown race and does not claim to fix every source of exit code 134.

### DIFF
--- a/vllm_omni/diffusion/worker/diffusion_worker.py
+++ b/vllm_omni/diffusion/worker/diffusion_worker.py
@@ -789,6 +789,12 @@ class CustomPipelineWorkerExtension:
 class WorkerProc:
     """Wrapper that runs one Worker in a separate process."""
 
+    _ASYNC_OUTPUT_STOP = object()
+    # Keep this below the executor's graceful process join timeout so a stuck
+    # D2H operation takes the worker's force-exit path before the parent sends
+    # SIGTERM.
+    _ASYNC_OUTPUT_JOIN_TIMEOUT_S = 4.0
+
     def __init__(
         self,
         od_config: OmniDiffusionConfig,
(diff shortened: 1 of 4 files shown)


hsliuustc0106:
any exact broken cases?

l-wave:
> any exact broken cases?

Thanks for asking. The incident that led me to inspect this path was a single-card HunyuanImage3 request-mode run on Ascend 950PR with 8 denoising steps.

The request completed successfully and saved the generated image, metrics, and latent trajectory. The shutdown log then showed:

Worker 0: Received shutdown message
event loop terminated.
Worker 0: Shutdown complete.
Terminating diffusion worker DiffusionWorker-0 after timeout
corrupted size vs. prev_size

The process exited with code 134. We observed the same shutdown sequence in multiple MXFP8 and MXFP4 runs.

I want to clarify the causality. We later reproduced the same glibc allocator abort by importing vllm_omni and exiting without running a diffusion request. Therefore, exit code 134 is evidence of a real shutdown problem in this environment, but it is not a clean reproducer that attributes the crash solely to the async-output thread.

The pre-patch async-output race is still concrete. In request mode, the worker publishes COMPUTE_DONE immediately after placing the output in the async queue. The background thread then waits for the device event, performs D2H transfer and shared-memory packing, and publishes OUTPUT_READY.

During shutdown, the old code set _running to false and immediately called worker.shutdown and context.term without waking or joining the async-output thread. If the thread was blocked in Queue.get, it remained alive during device and context teardown. If one output was being processed while another was queued, the while self._running condition could terminate the loop after the current output, leaving the queued output without OUTPUT_READY even though COMPUTE_DONE had already been published. This could leave the client waiting for an output that would never become ready.

l-wave:
> any exact broken cases?

Yes. We have a concrete case from a single-card HunyuanImage3 request-mode run on Ascend 950PR.

The run used NPU 0, 8 denoising steps, guidance scale 2.5, seed 42, and one input image of size 1024x1024. The input image SHA256 was:

5ee3faa6febdbaa5864fe9432f6e5aae821530c25ba2230e4f65657d34cb2809

The request completed successfully in 4.983 seconds. The generated image and latent trajectory were both saved. Shutdown then produced the following log:

INFO [Omni] Shutting down
INFO [AsyncOmniEngine] Shutting down Orchestrator
INFO [Orchestrator] Received shutdown signal
INFO Worker 0: Received shutdown message
INFO event loop terminated.
INFO Worker 0: Shutdown complete.
WARNING Terminating diffusion worker DiffusionWorker-0 after timeout
INFO Stage 0 replica 0 shut down
corrupted size vs. prev_size
Aborted
benchmark_rc=134

The exact broken behavior in this run was that the diffusion worker did not exit within the five-second shutdown timeout and had to be force-terminated.

In the old implementation, the request-mode DiffusionAsyncOutput thread returned to Queue.get after publishing the output. Setting _running to false did not wake that blocked Queue.get call. Worker shutdown and context termination then proceeded without sending a stop item or joining the thread.

I should also disclose that the same glibc allocator error can be reproduced by importing vllm_omni and exiting without running a diffusion request. Therefore, this PR should not claim to fix every source of exit code 134. It specifically fixes the async-output shutdown ordering by waking the thread, draining pending outputs, synchronizing its stream, and joining it before worker and device-context teardown.

I can attach the exact input image, prompt, generated output, metrics, and full log to the PR for reproduction.

vllm_omni/diffusion/worker/diffusion_worker.py
```diff
@@ -895,35 +898,64 @@ def _async_output_loop(self):
...
+            return
+
+        output_queue.put(WorkerProc._ASYNC_OUTPUT_STOP)
+        output_thread.join(timeout=WorkerProc._ASYNC_OUTPUT_JOIN_TIMEOUT_S)
+        if output_thread.is_alive():
+            logger.warning(
+                "Async output thread did not stop within %.1f seconds; waiting before device teardown",
+                WorkerProc._ASYNC_OUTPUT_JOIN_TIMEOUT_S,
+            )
+            output_thread.join()
```
hsliuustc0106: [P1] The timeout above does not actually bound shutdown: if `pack_diffusion_output_shm` or `d2h_stream.synchronize()` is stuck, this unconditional `join()` hangs the worker indefinitely and can defeat the parent shutdown timeout. Please define a bounded cancellation/force-teardown path, or document why teardown must wait forever.

l-wave: > [P1] The timeout above does not actually bound shutdown: if `pack_diffusion_output_shm` or `d2h_stream.synchronize()` is stuck, this unconditional `join()` hangs the worker indefinitely and can defeat the parent shutdown timeout. Please define a bounded cancellation/force-teardown path, or document why teardown must wait forever.

Addressed in 332e9d5. The async output thread now receives a single bounded 4-second join, which is shorter than the parent process's graceful shutdown timeout. If the thread remains blocked in SHM packing or device synchronization, the worker skips device and ZMQ teardown and immediately exits the child process, avoiding concurrent teardown while the thread still owns an active device stream. The parent cleaner now also escalates from graceful join to terminate and finally kill, with bounded waits at every stage. Regression coverage was added for the bounded thread timeout and the terminate-to-kill process path.

PR #5759, opened 2026-08-04
[NPU] Add configurable MoE communication method

l-wave:
## Summary

Add a stage-local configuration override for selecting the NPU MoE communication method used by diffusion models.

## Motivation

The optimal MoE communication method can vary with model topology, tensor parallel configuration, expert parallel configuration, and NPU platform. Automatic selection makes performance experiments difficult to reproduce and prevents a stage from selecting a known suitable method for its own topology.

## Changes

Read `npu_moe_comm_method` from each worker or stage's `VllmConfig.additional_config`.

Supported values:

```text
ALLGATHER
ALLTOALL
```

Values are case-insensitive. Invalid values and non-string values fail during runtime preparation with a clear error. Explicit `ALLTOALL` selection also fails early unless expert parallelism is enabled with an EP world size greater than one.

When the key is absent, the existing platform-based automatic selection remains unchanged.

## Usage

```python
additional_config = {
    "npu_moe_comm_method": "ALLTOALL",
}
```

For a diffusion stage YAML configuration:

```yaml
stages:
  - stage_id: 0
    additional_config:
      npu_moe_comm_method: ALLTOALL
```

Each stage carries its own value, allowing multi-stage deployments to select methods independently for different EP or TP topologies.

## Validation

The NPU regression test is located at:

```text
tests/platforms/npu/test_fused_moe_comm_selection.py
```

Coverage includes:

- ALLGATHER and ALLTOALL selection through `additional_config`
- invalid strings, empty strings, and falsey non-string values
- rejection of ALLTOALL when EP is disabled or the EP group has one rank
- NPU-only collection with `vllm_ascend` guarded by `pytest.importorskip`

### DIFF
--- a/tests/platforms/npu/test_fused_moe_comm_selection.py
+++ b/tests/platforms/npu/test_fused_moe_comm_selection.py
@@ -1,0 +1,86 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+
+from types import SimpleNamespace
+
+import pytest
+
+pytest.importorskip("vllm_ascend")
+
+from vllm_ascend.ascend_forward_context import MoECommType
+
+from vllm_omni.platforms.npu.layers import fused_moe
+
+pytestmark = [pytest.mark.core_model, pytest.mark.npu]
+
+
+def _vllm_config(additional_config=None, *, enable_expert_parallel=True):
+    return SimpleNamespace(
+        additional_config=additional_config,
+        parallel_config=SimpleNamespace(enable_expert_parallel=enable_expert_parallel),
+    )
+
+
+def _set_ep_size(monkeypatch, ep_size):
+    monkeypatch.setattr(
+        fused_moe,
+        "get_ep_group",
+        lambda: SimpleNamespace(world_size=ep_size),
+    )
+
+
+@pytest.mark.parametrize(
+    ("configured", "expected"),
+    [
+        ("allgather", MoECommType.ALLGATHER),
+        ("ALLTOALL", MoECommType.ALLTOALL),
+    ],
+)
+def test_additional_config_override(monkeypatch, configured, expected):
+    _set_ep_size(monkeypatch, 2)
+
+    result = fused_moe._select_moe_comm_method(_vllm_config({"npu_moe_comm_method": configured}))
+
+    assert result is expected
+
+
+def test_invalid_override_fails_fast(monkeypatch):
+    _set_ep_size(monkeypatch, 2)
+
+    with pytest.raises(ValueError, match="Supported values"):
+        fused_moe._select_moe_comm_method(_vllm_config({"npu_moe_comm_method": "mc2"}))
+
+
+@pytest.mark.parametrize("configured", [False, 0, []])
+def test_falsey_non_string_override_fails_fast(monkeypatch, configured):
+    _set_ep_size(monkeypatch, 2)
+
+    with pytest.raises(TypeError, match="must be a string"):
+        fused_moe._select_moe_comm_method(_vllm_config({"npu_moe_comm_method": configured}))
+
+
+def test_empty_string_override_fails_fast(monkeypatch):
+    _set_ep_size(monkeypatch, 2)
+
+    with pytest.raises(ValueError, match="Supported values"):
+        fused_moe._select_moe_comm_method(_vllm_config({"npu_moe_comm_method": ""}))
+
+
+@pytest.mark.parametrize(
+    ("enable_expert_parallel", "ep_size"),
+    [(False, 1), (True, 1)],
+)
+def test_alltoall_override_requires_multi_rank_ep(
+    monkeypatch,
+    enable_expert_parallel,
+    ep_size,
+):
+    _set_ep_size(monkeypatch, ep_size)
+
+    with pytest.raises(ValueError, match="requires expert parallelism"):
+        fused_moe._select_moe_comm_method(
+            _vllm_config(
+                {"npu_moe_comm_method": "alltoall"},
+                enable_expert_parallel=enable_expert_parallel,
+            )
+        )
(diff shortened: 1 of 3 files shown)


vllm_omni/platforms/npu/layers/fused_moe.py
```diff
@@ -47,6 +48,24 @@ def _init_mc2_group_for_diffusion(
 
 
 def _select_moe_comm_method(vllm_config: VllmConfig) -> MoECommType | None:
+    override = os.getenv("VLLM_OMNI_NPU_MOE_COMM_METHOD")
```
hsliuustc0106: [P1] Please avoid making this a process-global precedence rule. additional_config is carried per diffusion worker or stage, but this inherited env value overrides it for every worker, so a multi-stage deployment cannot choose different methods for different EP or TP topologies once the variable is set. Please make stage-local config authoritative and use the env var only as a fallback, or resolve it in each worker launch context.

vllm_omni/platforms/npu/layers/fused_moe.py
```diff
@@ -47,6 +48,24 @@ def _init_mc2_group_for_diffusion(
...
+            raise TypeError(f"npu_moe_comm_method must be a string, got {type(override).__name__}")
+        normalized_override = override.strip().upper()
+        supported_overrides = {
+            "ALLGATHER": MoECommType.ALLGATHER,
+            "ALLTOALL": MoECommType.ALLTOALL,
+        }
+        if normalized_override not in supported_overrides:
+            supported = ", ".join(sorted(supported_overrides))
+            raise ValueError(f"Unsupported VLLM_OMNI_NPU_MOE_COMM_METHOD={override!r}. Supported values: {supported}")
+        return supported_overrides[normalized_override]
```
hsliuustc0106: [P1] Keep the EP topology guard in effect for explicit overrides. This return runs before the existing ep_size == 1 check, so ALLTOALL can be selected with EP disabled or a singleton EP group. In the vLLM-Ascend v0.26.0rc branch, ALLTOALL is registered only when ep_size > 1; otherwise only ALLGATHER is available, so this can fail later with a missing communication context. Please reject incompatible overrides or preserve the ALLGATHER fallback: https://github.com/vllm-project/vllm-ascend/blob/releases/v0.26.0rc/vllm_ascend/ops/fused_moe/moe_comm_method.py

vllm_omni/platforms/npu/layers/fused_moe.py
```diff
@@ -47,6 +48,24 @@ def _init_mc2_group_for_diffusion(
 
 
 def _select_moe_comm_method(vllm_config: VllmConfig) -> MoECommType | None:
+    override = os.getenv("VLLM_OMNI_NPU_MOE_COMM_METHOD")
+    if not override:
+        additional_config = getattr(vllm_config, "additional_config", None)
+        if isinstance(additional_config, dict):
+            override = additional_config.get("npu_moe_comm_method")
+    if override:
```
hsliuustc0106: [P2] Use presence-based validation here (override is not None) instead of truthiness. Falsey invalid values from additional_config such as False, 0, or [] silently fall back to automatic selection, contradicting the fail-fast contract. Please add regression coverage for these values.

tests/platforms/npu/test_fused_moe_comm_selection.py
```diff
@@ -0,0 +1,45 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+
+from types import SimpleNamespace
+
+import pytest
+from vllm_ascend.ascend_forward_context import MoECommType
```
hsliuustc0106, line 10: [P1] This test is marked cpu, but importing vllm_ascend here happens during collection. The CPU test environment does not install that package, so pytest -m 'core_model and cpu' fails before running the test. Please move this to an NPU-only marker or load the module with the repository fake-module pattern.

l-wave: > [P1] This test is marked cpu, but importing vllm_ascend here happens during collection. The CPU test environment does not install that package, so pytest -m 'core_model and cpu' fails before running the test. Please move this to an NPU-only marker or load the module with the repository fake-module pattern.

@hsliuustc0106 Thanks for the detailed review. I updated the PR so stage-local additional_config takes precedence, with VLLM_OMNI_NPU_MOE_COMM_METHOD used only as a fallback. Explicit ALLTOALL selection now fails early unless expert parallelism is enabled with ep_size > 1. Falsey invalid configuration values are validated by presence and covered by regression tests. I also moved the test to the NPU marker and added a collection-time importorskip so CPU environments without vllm_ascend do not fail during collection. The branch has been updated.

hsliuustc0106 (commented):
avoid adding env variable

l-wave:
> avoid adding env variable

Removed the environment variable entirely. The MoE communication method is now configured only through each stage's VllmConfig.additional_config.

The existing EP topology guard, invalid-value validation, and NPU-only test collection coverage are preserved. I also updated the PR description and usage example accordingly.

vllm_omni/platforms/npu/layers/fused_moe.py
```diff
@@ -47,8 +47,32 @@ def _init_mc2_group_for_diffusion(
 
 
 def _select_moe_comm_method(vllm_config: VllmConfig) -> MoECommType | None:
+    override = None
+    additional_config = getattr(vllm_config, "additional_config", None)
+    if isinstance(additional_config, dict):
+        override = additional_config.get("npu_moe_comm_method")
```
hsliuustc0106, line 53: [P2] Please document `npu_moe_comm_method` in the NPU deployment/configuration docs, including the accepted values and the `ALLTOALL` expert-parallelism requirement. This is a new user-facing override, and the current coverage is NPU-only.

l-wave: The NPU deployment guide now describes the stage-local npu_moe_comm_method setting, its YAML and CLI usage, the case-insensitive ALLGATHER and ALLTOALL values, and the requirement that ALLTOALL must run with expert parallelism enabled and an effective EP world size greater than one. The guide also documents the automatic selection behavior when the override is omitted.

PR #5760, opened 2026-08-04
[NPU] Add offline MXFP8 support for routed experts

l-wave:
## Summary

Add offline MXFP8 checkpoint support for NPU `RoutedExperts` layers.

This connects vLLM Omni diffusion MoE layers to the existing vLLM Ascend MXFP8 fused MoE implementation while preserving the standard vLLM quantization method interface.

## Motivation

The diffusion MXFP8 configuration supports NPU linear layers but previously returned no quantization method for `RoutedExperts`. Offline MXFP8 MoE checkpoints therefore could not select the Ascend MXFP8 fused expert implementation through the standard quantization path.

## Changes

- Detect `RoutedExperts` for serialized MXFP8 checkpoints on NPU.
- Create `AscendW8A8MXFP8DynamicFusedMoEMethod` for expert execution.
- Adapt the scheme with `AscendFusedMoEMethod`.
- Keep `get_quant_method` aligned with the two-argument vLLM `QuantizationConfig` interface.
- Provide the MXFP8 group size through `quant_description`.
- Preserve the unquantized fallback for ignored routed-expert layers.
- Extend ignored-layer matching so a parent block prefix excludes its child modules.

## Behavior

```text
RoutedExperts
  -> AscendW8A8MXFP8DynamicFusedMoEMethod
  -> AscendFusedMoEMethod
```

When the layer or a parent prefix appears in `ignored_layers`, routed experts retain the existing unquantized fallback. Child linear layers under the same prefix use `UnquantizedLinearMethod`.

Existing linear-layer and non-NPU behavior remains unchanged.

## Validation

`tests/diffusion/quantization/test_mxfp8_config.py` covers:

- existing offline NPU MXFP8 linear selection
- Ascend fused MoE method creation
- ignored routed-expert fallback
- child linear exclusion through a parent prefix

### DIFF
--- a/vllm_omni/quantization/mxfp8_config.py
+++ b/vllm_omni/quantization/mxfp8_config.py
@@ -34,6 +34,7 @@ Extending to a new MX precision with a different calling convention (e.g. dual-s
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
+from collections.abc import Mapping
 from typing import TYPE_CHECKING, Any
 
 import torch
@@ -62,6 +63,24 @@ if TYPE_CHECKING:
 
 logger = init_logger(__name__)
 
+
+def _is_layer_or_child_skipped(
+    prefix: str,
+    ignored_layers: list[str],
+    fused_mapping: Mapping[str, list[str]],
+) -> bool:
+    if is_layer_skipped(
+        prefix=prefix,
+        ignored_layers=ignored_layers,
+        fused_mapping=fused_mapping,
+    ):
+        return True
+    prefix_without_model = prefix.removeprefix("model.")
+    return any(
+        prefix_without_model.startswith(f"{ignored_layer.removeprefix('model.')}.") for ignored_layer in ignored_layers
+    )
+
+
 # ---------------------------------------------------------------------------
 # Config
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


vllm_omni/quantization/mxfp8_config.py
```diff
@@ -122,11 +142,33 @@ def get_quant_method(
         layer: torch.nn.Module,
         prefix: str,
     ) -> QuantizeMethodBase | None:
+        if current_omni_platform.is_npu() and self.is_checkpoint_mxfp8_serialized:
+            from vllm.model_executor.layers.fused_moe.routed_experts import RoutedExperts
```
hsliuustc0106, line 146: [P2] Please add Ascend/NPU regression coverage (or CI) that instantiates `RoutedExperts` and exercises the returned `AscendFusedMoEMethod` with a serialized expert checkpoint. This `is_npu` branch is skipped by the current CPU environment, so the core new path is not runtime-tested here.

l-wave: Addressed. I added an A3 NPU regression that constructs a real RoutedExperts layer with a serialized MXFP8 configuration and verifies that it selects the real AscendFusedMoEMethod and AscendW8A8MXFP8DynamicFusedMoEMethod.

The test loads synthetic gate, up, and down expert weights together with their MXFP8 group scales through the routed-expert weight loaders, verifies the loaded parameter slices, and exercises post-loading layout conversion. I also added this regression to the NPU ready pipeline.

PR #5761, opened 2026-08-04
[Bugfix] Fix MXFP8 scale sharding for row-parallel layers

l-wave:
## Summary

Fix tensor-parallel sharding of offline MXFP8 weight scales for row-parallel linear layers.

## Problem

Offline MXFP8 weights use one scale for every 32 input elements. The `weight_scale` tensor therefore has the following logical dimensions:

```text
output_size × input_groups
```

The parameter currently declares no input partition dimension:

```python
input_dim=None
```

For row-parallel layers, the weight tensor is sharded along its input dimension. The corresponding MXFP8 scale tensor must shard its input-group dimension in the same way.

Without this metadata, each TP rank may attempt to load the complete scale dimension, causing scale shape mismatches or incorrect scale alignment with the local weight shard.

## Changes

Set the MXFP8 `weight_scale` input partition dimension to dimension 1:

```python
ModelWeightParameter(
    data=...,
    input_dim=1,
    output_dim=0,
    weight_loader=weight_loader,
)
```

This allows vLLM's model weight loader to shard the scale tensor consistently with row-parallel MXFP8 weights.

Column-parallel output sharding continues to use dimension 0.

## Example

For an input size of 3072, group size 32, and tensor parallel size 2:

```text
Global input groups: 3072 / 32 = 96
Groups per TP rank: 96 / 2 = 48
Local weight_scale shape: 4096 × 48
```

## Validation

Validated with the latest vLLM Omni source tree, vLLM 0.26, and vLLM Ascend on Ascend 950PR.

```text
test_offline_weight_scale_tracks_input_partition_dimension
1 passed
```

The test verifies:

- The local scale shape is `4096 × 48`
- `input_dim` is set to 1
- `output_dim` remains 0

Ruff, formatting checks, Python syntax compilation, and `git diff --check` also pass.
```

### DIFF
--- a/vllm_omni/quantization/mxfp8_config.py
+++ b/vllm_omni/quantization/mxfp8_config.py
@@ -345,7 +345,7 @@ class NPUMxfp8LinearMethod(MXFPLinearMethodBase):
             "weight_scale",
             ModelWeightParameter(
                 data=torch.empty(output_size_per_partition, num_groups, dtype=torch.uint8),
-                input_dim=None,
+                input_dim=1,
                 output_dim=0,
                 weight_loader=weight_loader,
             ),
--- a/tests/diffusion/quantization/test_mxfp8_config.py
+++ b/tests/diffusion/quantization/test_mxfp8_config.py
@@ -204,6 +204,48 @@ def test_get_quant_method_non_linear_returns_none(monkeypatch: pytest.MonkeyPatc
     assert config.get_quant_method(norm_layer, "blocks.0.norm1") is None
 
 
+@pytest.mark.parametrize("tp_rank", [0, 1])
+def test_offline_weight_scale_loads_input_partition_columns(
+    mocker: MockerFixture,
+    tp_rank: int,
+):
+    """RowParallel weight scales must load the rank-local K-group columns."""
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: vllm_omni/quantization/mxfp8_config.py [hunk 1]
```diff
@@ -345,7 +345,7 @@ def create_weights(
             "weight_scale",
             ModelWeightParameter(
                 data=torch.empty(output_size_per_partition, num_groups, dtype=torch.uint8),
```
hsliuustc0106, line 347: [P2] Please extend the regression to load a known full scale tensor at TP rank 0 and rank 1 and assert the selected columns. Shape plus `input_dim`/`output_dim` metadata would still pass if the loader slices the wrong dimension.

l-wave: Addressed. The regression now runs at TP rank 0 and rank 1, loads a full scale tensor with distinct values for every input-group column through load_row_parallel_weight, and verifies the exact 48 columns stored on each rank. This covers the actual sharding behavior in addition to shape and parameter metadata.

PR #5762, opened 2026-08-04
[Bugfix] Fix HunyuanImage3 QKV scale splitting

l-wave:
## Summary

Fix HunyuanImage3 QKV splitting for quantization scale tensors whose input width differs from the model hidden size.

## Problem

HunyuanImage3 checkpoints store Q, K, and V tensors in an interleaved grouped-query attention layout. The loader uses `_split_qkv_weight` to reorder them into the layout expected by vLLM.

The existing implementation assumes that the second tensor dimension always equals `config.hidden_size`:

```python
hidden_size = self.config.hidden_size
```

This assumption is valid for regular weight tensors. It does not hold for grouped quantization scales.

For MXFP8, the second dimension represents input groups:

```text
input_size / group_size
```

Using the hidden size when reshaping an MXFP8 scale tensor results in an invalid reshape or incorrectly reordered scale data.

## Changes

Use the actual tensor width when splitting QKV tensors:

```python
input_width = qkv.shape[1]
```

The QKV reshape and final Q, K, and V flattening now preserve the input width of the supplied tensor.

This supports both regular weight tensors and grouped quantization scale tensors through the same loading path.

## Validation

Validated with the latest vLLM Omni source tree, vLLM 0.26, and vLLM Ascend on Ascend 950PR.

```text
tests/model_executor/models/test_hunyuan_image3_mxfp8_loading.py
1 passed
```

The test uses an MXFP8 group-scale tensor with shape:

```text
6144 × 128
```

It verifies:

- The output shape is preserved
- Q, K, and V rows are reordered correctly
- The implementation does not assume the second dimension equals the hidden size

Ruff, formatting checks, Python syntax compilation, and `git diff --check` also pass.
```

### CODE REVIEW

File: vllm_omni/model_executor/models/hunyuan_image3/hunyuan_image3.py [hunk 1]
```diff
@@ -121,11 +121,16 @@ def _split_qkv_weight(self, qkv: torch.Tensor):
         else:
             attention_head_dim = self.config.hidden_size // num_attention_heads
 
-        qkv = qkv.reshape(num_kv_heads, num_key_value_groups + 2, attention_head_dim, hidden_size)
+        qkv = qkv.reshape(
```
hsliuustc0106, line 124: [P2] Please add a loader-level regression that feeds a packed QKV scale tensor through the Hunyuan loader and verifies the q/k/v parameter slices. The current test validates `_split_qkv_weight` directly but not the checkpoint-to-parameter loading contract.

l-wave: Addressed. I added a loader-level regression that passes a packed qkv_proj.weight_scale tensor through HunyuanModel.load_weights. The test verifies the q, k, and v shard order and validates each resulting parameter slice independently.

PR #5763, opened 2026-08-04
[Bugfix] Materialize safetensors before device copies

l-wave:
## Summary

Add an optional stage-local weight-loading path that materializes memory-mapped safetensors on CPU before copying them to an accelerator device.

## Problem

Safetensors may expose CPU tensors backed directly by memory-mapped files. On some accelerator runtime and driver combinations, copying a large mmap-backed tensor directly to the device can stall during model loading. This was observed while loading quantized HunyuanImage3 weights on Ascend 950PR.

## Changes

Add `OmniDiffusionConfig.materialize_safetensors_weights`, which defaults to `false`. When enabled for a diffusion stage, its safetensors iterator clones each CPU tensor before passing it to the model weight loader.

```python
if enabled and tensor.device.type == "cpu":
    tensor = tensor.clone()
```

The option applies only to safetensors. Other checkpoint formats retain their current behavior.

## Usage

```yaml
stages:
  - stage_id: 0
    materialize_safetensors_weights: true
```

The setting is carried by each diffusion stage and does not introduce process-global environment state.

## Memory behavior

Only the current tensor is cloned. Additional peak host-memory usage is bounded by approximately the largest individual tensor being loaded.

## Validation

`tests/diffusion/model_loader/test_diffusers_loader.py` covers enabled and disabled iterator behavior, including storage independence for materialized CPU tensors.

### DIFF
--- a/vllm_omni/diffusion/data.py
+++ b/vllm_omni/diffusion/data.py
@@ -715,6 +715,7 @@ class OmniDiffusionConfig:
     # Parallel weight loading (for faster diffusion model startup)
     enable_multithread_weight_load: bool = True
     num_weight_load_threads: int = 4
+    materialize_safetensors_weights: bool = False
 
     # Enable sleep mode
     enable_sleep_mode: bool = False
--- a/docs/configuration/stage_configs.md
+++ b/docs/configuration/stage_configs.md
@@ -48,7 +48,22 @@ Each entry under `stages:` accepts any `StageDeployConfig` field directly (no ne
 | `output_connectors` | dict \| null | optional | `null` | Keyed by `to_stage_<n>`; values are names registered under top-level `connectors:`. |
 | `input_connectors` | dict \| null | optional | `null` | Keyed by `from_stage_<n>`; values are names registered under top-level `connectors:`. |
 | `default_sampling_params` | dict \| null | optional | `null` | Baseline sampling params. Deep-merged with pipeline `sampling_constraints` (pipeline wins). |
+| `materialize_safetensors_weights` | bool | optional | `false` | Diffusion stages only. Materialize CPU safetensors into owned host memory before device copies. |
 | `engine_extras` | dict | optional | `{}` | Catch-all for keys not listed above; deep-merged across overlays. Also carries per-stage overrides of pipeline-wide settings (e.g. stage-specific `dtype`). |
+
+#### Materializing safetensors before device copies
+
+For diffusion stages, enable `materialize_safetensors_weights` when direct device copies from memory-mapped safetensors stall on the target accelerator:
+
+```yaml
+stages:
+  - stage_id: 1
+    materialize_safetensors_weights: true
+```
+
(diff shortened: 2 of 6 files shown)


vllm_omni/diffusion/data.py
```diff
@@ -715,6 +715,7 @@ class OmniDiffusionConfig:
     # Parallel weight loading (for faster diffusion model startup)
     enable_multithread_weight_load: bool = True
     num_weight_load_threads: int = 4
+    materialize_safetensors_weights: bool = False
```
hsliuustc0106, line 718: [P1] This field is not included in `_DiffusionConfigProjection`, so it is absent from `_DIFFUSION_STAGE_ENGINE_FIELDS`. A stage YAML value therefore lands in `engine_extras` and is dropped by `_DiffusionEngineOverrides.from_engine`, leaving the loader's default `False` in structured stage startup. Please thread this field through the projection/known-field set and add a stage-YAML-to-loader round-trip test.

l-wave: Addressed. materialize_safetensors_weights is now included in _DiffusionConfigProjection, so it is preserved by _DIFFUSION_STAGE_ENGINE_FIELDS during structured stage startup. I also added a regression test that loads the option from stage YAML, constructs the runtime OmniDiffusionConfig, and verifies that the enabled value reaches DiffusersPipelineLoader.

lishunyang12 (approved):
Reviewed the re-requested fix and ran the focused config/loader tests (110 passed, 1 skipped). The stage-config projection now preserves the option and materialization remains scoped to CPU safetensors.

hsliuustc0106 (commented):
[P2] Document the new stage option — `vllm_omni/config/omni_config.py:470`

Assuming `materialize_safetensors_weights` is intended as a supported user-facing stage YAML option (as documented in this PR), please add it to `docs/configuration/stage_configs.md` or a linked diffusion-loading reference. The current docs only describe the generic `engine_extras` fallback, so users cannot discover the option or its important contract: default `false`, safetensors/CPU scope, and approximately one-largest-tensor of additional host-memory usage when enabled.

The config propagation fix and loader behavior otherwise look correct. Focused config/loader tests passed, and no P0/P1 correctness issue was found.

l-wave:
> [P2] Document the new stage option — `vllm_omni/config/omni_config.py:470`
> 
> Assuming `materialize_safetensors_weights` is intended as a supported user-facing stage YAML option (as documented in this PR), please add it to `docs/configuration/stage_configs.md` or a linked diffusion-loading reference. The current docs only describe the generic `engine_extras` fallback, so users cannot discover the option or its important contract: default `false`, safetensors/CPU scope, and approximately one-largest-tensor of additional host-memory usage when enabled.
> 
> The config propagation fix and loader behavior otherwise look correct. Focused config/loader tests passed, and no P0/P1 correctness issue was found.

Addressed. I documented `materialize_safetensors_weights` in the stage configuration reference, including its default value of `false`, its diffusion-stage and CPU-safetensors scope, a stage YAML example, and the approximate one-largest-tensor peak host-memory cost. Other checkpoint formats and tensors already on another device remain unaffected.

PR #5764, opened 2026-08-04
feat(minimax-h3): enable RTX 4090/5090 support with DLO

lishunyang12:
## Why

MiniMax-H3 combines a 66.3 GB DiT, a 51.5 GB Qwen3-VL encoder, and video/audio VAEs. The existing resident or pure-Ulysses paths cannot fit the released BF16/FP32 pipeline on two 24/32 GB workstation GPUs.

## What changed

- Add MiniMax-H3-aware distributed layerwise offload (DLO) with explicit module residency.
- Support TP-local streaming with `--dlo-no-use-allgather`, avoiding reconstruction of full DiT blocks on each rank.
- Stage the text encoder and executable VAE decoder blocks while keeping the required H3 encoders resident.
- Use VAE patch parallelism across both GPUs and cuDNN attention for the RTX consumer path.
- Add correctness, residency, offloader, cuDNN-attention, and multiprocess tests.
- Add an all-task runner for T2VA, FL2VA, image+audio Ref2VA, and two-video Ref2VA.

## Two-GPU profiles

| Profile | GPUs | Starting shape | Resident DiT blocks | Attention | Execution | Validation |
|---|---:|---:|---:|---|---|---|
| `rtx5090` | 2 x 32 GB | 1344x768 | 20 | `CUDNN_ATTN` | eager | Target-hardware validated |
| `rtx4090` | 2 x 24 GB | 1024x576 | 12 | `CUDNN_ATTN` | eager | Capacity-proxy starting point |

```bash
RUN_ROOT=/path/to/run-root \
MODEL_ROOT=/path/to/MiniMax-H3 \
GPU_IDS=0,1 \
PROFILE=rtx5090 \
bash recipes/MiniMaxAI/scripts/run_h3_2gpu_all_tasks.sh
```

Use `PROFILE=rtx4090` for the conservative 24 GB defaults. `DLO_RESIDENT_LAYERS=N` overrides either profile.

## RTX 5090 validation

At commit `ae6577ea`, one full 50-step T2VA request completed on 2 x RTX 5090 without OOM:

| Shape | Frames | Client E2E | Sampled peak/GPU | Output |
|---:|---:|---:|---:|---|
| 1344x768 | 124 at 24 FPS | 8 min 38 s | approximately 22.6 GiB | H.264 + 32 kHz stereo AAC; full `ffmpeg` decode passed |

Environment: vLLM 0.26.0, vLLM-Omni `0.26.1.dev14+gae6577ea`, PyTorch 2.11.0+cu130. This is a single end-to-end validation run, not a warmed multi-run benchmark; the memory value is a sampled `nvidia-smi` peak rather than a CUDA allocator high-water mark.

[Target-run record](https://github.com/lishunyang12/vllm-omni-rankings/blob/dcd06d7e83cb069842535918c0169ee9f3f29ba0/scripts/%E5%BE%AE%E4%BF%A1%E5%9B%BE%E7%89%87_20260805000034_86_237.png)


hsliuustc0106:
@yuanheng-zhao do we still need to keep layerwise offload? this looks like a special case of DLO

examples/offline_inference/minimax_h3/run_h3_2gpu_all_tasks.sh
```diff
@@ -0,0 +1,309 @@
+#!/usr/bin/env bash
+# Run every MiniMax-H3 task supported by vLLM-Omni on two workstation GPUs.
```
hsliuustc0106: we seldom put bash files under recipes, it;s better to move them under examples

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
```diff
@@ -219,6 +219,15 @@ def _shard_and_pin(
             weights_with_local = []
             for name, t in name2weights.items():
                 local_t = t.to_local() if hasattr(t, "to_local") else t
+                mmap_transform = getattr(t, "mmap_weight_transform", None)
```
hsliuustc0106, line 222: [P1] This transform path is never populated for MiniMax-H3. MiniMaxH3DiTModel.load_weights() applies _reorder_grouped_qkv_to_qkv() to a local tensor, but no production code assigns mmap_weight_transform to the parameter (the only assignments are in tests). Since _remap_ckpt_key() opts H3 into mmap and the mmap loader skips load_weights(), DLO+AllGather installs raw grouped-QKV weights. Please attach the transform, and cover the fused MLP layout, or keep H3 on regular loading.

vllm_omni/diffusion/executor/multiproc_executor.py
```diff
@@ -521,6 +521,18 @@ def execute_batch(self, scheduler_output: DiffusionSchedulerOutput) -> BaseRunne
         if len(scheduler_output.scheduled_new_reqs) <= 1:
             return self.execute_request(scheduler_output)
 
+        parallel_config = getattr(self.od_config, "parallel_config", None)
+        dp_size = getattr(parallel_config, "data_parallel_size", 1)
+        if (
+            dp_size > 1
+            and getattr(self.od_config, "enable_distributed_layerwise_offload", False)
+            and getattr(self.od_config, "dlo_use_allgather", True)
```
hsliuustc0106, line 526: [P1] The DP-concurrency guard also needs to cover dlo_use_allgather=False. The engine enables dp_concurrent for any DLO+DP configuration, but this branch sends no-AllGather multi-request waves to execute_model_batch(). MiniMax-H3 rejects those batches because forward() requires one prompt, so concurrent requests fail in the advertised no-AllGather mode. Please route these waves through the independent-request path or disable DP concurrency when AllGather is off.

hsliuustc0106, line 526: [P1] Once this new DLO+DP path routes a wave through execute_request(), the extra_args check above must read nr.req.sampling_params.extra_args. It currently reads nr.req.extra_args, which is always None for normal requests, so differing pipeline-specific schedules can pass the guard and make DP ranks enter different AllGather sequences and hang. Please compare the sampling_params field using a safe canonical signature.

vllm_omni/engine/arg_utils.py
```diff
@@ -482,6 +482,7 @@ class OrchestratorArgs:
     enable_layerwise_offload: bool = False
     enable_distributed_layerwise_offload: bool = False
     dlo_use_allgather: bool = True
+    dlo_resident_layers: int = 0
```
hsliuustc0106, line 498: [P1] This new CLI field is still treated as an orchestrator-only key in the registry/deploy path. build_stage_runtime_overrides() filters OrchestratorArgs fields unless they are in StageDeployConfig or _PIPELINE_WIDE_ENGINE_FIELDS; dlo_resident_layers and the companion DLO flags are absent from both, so these CLI values are silently dropped for registered/deploy-config pipelines. Please add the fields to the deploy override schema or allowlist and cover that resolution path.

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
```diff
@@ -688,7 +724,9 @@ def _encode_text_hidden(
             # swaps the resident DiT and encoder.
             return self.text_encoder(input_ids, **vision_kwargs)
 
-        if self.od_config.enable_layerwise_offload:
+        if self.od_config.enable_layerwise_offload or getattr(
```
hsliuustc0106, line 970: [P1] This DLO branch is unreachable when enable_cpu_offload=True because the earlier CPU-offload branch wins. However OffloadConfig gives DLO priority and the DLO backend leaves the H3 encoder offloaded to CPU, so self.text_encoder(...) here reaches encode_ids() without staging and fails every request. Please reject the incompatible combination or check the effective DLO/layerwise mode before the CPU-offload branch.

recipes/MiniMaxAI/MiniMax-H3.md
```diff
@@ -77,6 +79,98 @@ Use a GPU with enough memory for the active H3 component and enough system RAM
...
+decode across both GPUs. cuDNN attention is selected explicitly for the RTX
+consumer path; the server stays eager to avoid an unqualified compile path.
+
+The resident count changes placement and transfer frequency only; it does not
+quantize or change the BF16/FP32 denoise math. Re-measure peak memory before
+increasing it on a different request shape.
+
+### RTX 5090 target-hardware validation
+
+At vLLM-Omni commit `ae6577ea`, one full 50-step T2VA request completed on
```
hsliuustc0106: [P1] This is a [Perf] PR, but the evidence here is only one un-warmed run on an older commit plus a B300 proxy. There is no matched main-vs-PR latency, throughput, or peak-VRAM comparison or baseline accuracy/quality comparison, and there are no repeated runs. Please add same-hardware A/B results with fixed inputs and seeds, warmup excluded, mean +/- standard deviation, and output-quality evidence before approval.

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
```diff
@@ -955,17 +1117,18 @@ def _init_dp_group(self) -> None:
...
-        Moving the module to GPU permanently avoids this issue at the
-        cost of slightly higher HBM usage (VAE/encoders are small).
+        Components that expose an explicit stage lifecycle are initially
+        offloaded and loaded by their pipeline only around encode/decode.
+        Other models retain the conservative resident behavior because a
+        generic post-forward hook can disrupt the DiT prefetch streams.
         """
+        offload_to_cpu = getattr(module, "offload_to_cpu", None)
+        if callable(offload_to_cpu):
+            offload_to_cpu()
```
david6666666, line 1138: Any module exposing `offload_to_cpu` is now offloaded here and trusts its pipeline to stage it back, but nothing enforces that pairing. Only MiniMax-H3 wires this up today -- other DLO models would silently keep their encoders on CPU if they ever add this method.

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
```diff
@@ -1163,93 +1401,41 @@ def enable(self, pipeline: nn.Module) -> None:
                 dit_module.to(self.device)
                 continue
 
+            resident_count = 0
+            if plan is not None and dit_name in plan.resident_dit_paths:
+                resident_count = min(self.config.dlo_resident_layers, len(blocks))
```
david6666666, line 1480: Nonzero `--dlo-resident-layers` is silently ignored unless the model declares `resident_dit_paths` in its `_offload_plan`. A warning here would make that visible.

david6666666:
**Proposal: make encoder layerwise offload generic in the DLO backend**

The H3 encoder offload currently lives in the model (`MiniMaxH3Qwen3VLEncoder.enable_omni_layerwise_offload()` + the `_omni_layerwise_*` branches in `load_to_device`/`offload_to_cpu`). The hook wiring itself is encoder-agnostic, and the code already documents why it must stay rank-local (the encoder has its own TP group; reusing the DiT AllGather group would concatenate different TP shards and corrupt weights).

Since DLO is meant to be a framework, I'd suggest moving this into `distributed_layerwise_backend.py` and driving it declaratively, same as DiT blocks:

- `OffloadPlan` gains `encoder_block_attrs: dict[str, tuple[str, ...]]` (pipeline attribute path -> block-list paths).
- A small `_LayerwiseStreamedComponent` in the backend applies the sequential `LayerwiseOffloadHook` per block group (rank-local, no AllGather) and tracks the hooked param/buffer ids, so `load_to_device`/`offload_to_cpu` only move the non-block parts.
- `enable()`'s encoder loop becomes: try generic encoder staging, else fall back to `_register_on_demand_hook`.

Then MiniMax-H3 only declares:

```python
_offload_plan = OffloadPlan(
    offload_submodules={"token_refiner": "blocks"},
    resident_dit_paths=frozenset({"transformer"}),
    encoder_block_attrs={"text_encoder": ("vision.blocks", "text_model.layers")},
)
```

and `enable_omni_layerwise_offload` plus the `_omni_layerwise_*` branches go away (~70 lines). Any other model with a streamable encoder then needs only a one-line plan entry instead of model-specific code. Happy to share the diff if this direction works for you.

vllm_omni/diffusion/models/minimax_h3/encoder.py
```diff
@@ -1029,16 +1029,84 @@ def _load_weights(self, model_path: str) -> None:
...
+            for name, child in self.text_model.named_children():
+                if name != "layers":
+                    child.to("cpu")
+            torch.accelerator.empty_cache()
+            return
         self.vision.to("cpu")
         self.text_model.to("cpu")
         torch.accelerator.empty_cache()
 
+    def enable_omni_layerwise_offload(self, *, pin_memory: bool = True) -> None:
```
david6666666, line 1061: Could this move into the DLO backend? The wiring is encoder-agnostic (and already documents why it must stay rank-local), so a plan-driven `_LayerwiseStreamedComponent` in `distributed_layerwise_backend.py` would let other models reuse it with a one-line `OffloadPlan.encoder_block_attrs` entry instead of model-specific code.

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
```diff
@@ -1130,6 +1365,9 @@ def enable(self, pipeline: nn.Module) -> None:
         # during the DiT forward pass.  They are only needed briefly for
         # text-encoding (before DiT) and VAE-decode (after DiT).
         for enc in modules.encoders:
```
david6666666: With encoder block groups declared in the plan, this loop could try a generic `_try_layerwise_offload_encoder` first and only fall back to `_register_on_demand_hook` -- that removes the `enable_omni_layerwise_offload` special case at this call site too.

vllm_omni/diffusion/offloader/offload_plan.py
```diff
@@ -35,10 +35,15 @@ class OffloadPlan:
...
             e.g. ``{"context_encoder": "layers"}``
+        resident_dit_paths: DiT paths whose leading blocks may be kept on the
+            device when ``dlo_resident_layers`` is nonzero. Keeping this
+            model-declared avoids applying a consumer-GPU tuning knob to
+            auxiliary or dual DiTs unintentionally.
     """
 
     block_attrs: dict[str, tuple[str, ...]] = field(default_factory=dict)
     offload_submodules: dict[str, str] = field(default_factory=dict)
+    resident_dit_paths: frozenset[str] = field(default_factory=frozenset)
```
david6666666, line 51: An `encoder_block_attrs` field here (pipeline path -> block-list paths, e.g. `{"text_encoder": ("vision.blocks", "text_model.layers")}`) would let the backend stage any encoder's blocks generically. See the proposal comment above.

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
```diff
@@ -267,6 +273,30 @@ class MiniMaxH3Pipeline(
...
+    def _remap_ckpt_key(key: str) -> str | None:
+        """Map released transformer keys to pipeline parameter names.
+
+        The DLO mmap loader scans all component indexes recursively.  Restrict
+        the mapping to H3 transformer roots so Qwen encoder keys cannot be
+        mistaken for DiT parameters.
+        """
+        if key.partition(".")[0] not in MiniMaxH3Pipeline._MMAP_TRANSFORMER_ROOTS:
+            return None
+        return f"transformer.{key}"
```
yuanheng-zhao: Is this method necessary? dlo without allgather enforces `dp_size=1`, while `supports_mmap_loading` works in the branch `if self.config.dlo_use_allgather and self.dp_size > 1:`

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
```diff
@@ -1163,93 +1401,41 @@ def enable(self, pipeline: nn.Module) -> None:
...
+            self._prepare_dit_non_block_modules(
+                dit_module,
+                dit_name,
+                blocks_attr_names,
+                all_dit_modules,
+                plan,
+            )
+
             num_blocks = len(blocks)
+            if num_blocks == 0:
```
yuanheng-zhao, line 1501: I think all the models other than MiniMax-H3 would get `num_blocks=0` here

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
```diff
@@ -1163,93 +1401,41 @@ def enable(self, pipeline: nn.Module) -> None:
...
             if num_blocks <= 1:
                 logger.warning(
-                    "#Target layers (blocks) <= 1. Skipping offloading on %s (%s)",
+                    "#Streaming target layers (blocks) <= 1. Keeping the final block resident on %s (%s)",
                     dit_name,
                     dit_module.__class__.__name__,
                 )
-                dit_module.to(self.device)
+                self._resident_blocks.extend(blocks)
                 continue
```
yuanheng-zhao, line 1511: Same concern as above.

vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
```diff
@@ -376,18 +383,26 @@ def _run_packed_attention(
...
+        # per layer and .item() would serialize every attention launch.
+        if not 0 < max_seqlen <= packed_total:
+            raise ValueError(
+                f"max_seqlen must be within the packed sequence, got {max_seqlen} for length {packed_total}"
+            )
+        used = min(max_seqlen, packed_total)
         attn_mask = None
-        if used < packed_total:
+        prefix_slice = self.attention.attn_backend.supports_prefix_kv_slicing
+        if used < packed_total and not prefix_slice:
```
yuanheng-zhao: A concern here:`supports_prefix_kv_slicing` is retrieved from the cls type, rather than an instance. What if the dispatched attn is not the same as the configured attn backend, for example, ring attn running

https://github.com/vllm-project/vllm-omni/blob/556c5f88a0f4b1d2e7285c6c69feecf0e3e39505/vllm_omni/diffusion/attention/layer.py#L280-L282

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
yuanheng-zhao: There's no producer for `mmap_weight_transform`

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
```diff
@@ -477,6 +489,107 @@ def remove_distributed_block_hook(module: nn.Module) -> None:
         logger.debug("Removed distributed offload hook from %s", module.__class__.__name__)
 
 
+class _PinnedResidentLayerGroup:
```
yuanheng-zhao: Suggest to put with `PinnedModuleStager` together into the same module. And prevent private naming (`_PinnedResidentLayerGroup` -> `PinnedResidentLayerGroup`)

examples/offline_inference/minimax_h3/run_h3_2gpu_all_tasks.sh
lishunyang12: Moved the runner under examples and updated the documentation link.

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
lishunyang12: H3 now explicitly stays on the regular loader until mmap can apply the same QKV and fused-MLP transforms.

vllm_omni/diffusion/executor/multiproc_executor.py
lishunyang12: DP concurrency is now enabled only for the AllGather DLO path.

lishunyang12: Now compares a canonical signature of sampling_params.extra_args.

vllm_omni/engine/arg_utils.py
lishunyang12: Added all DLO fields to StageDeployConfig so deploy overrides retain them.

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
lishunyang12: The encoder now takes the DLO/layerwise path before CPU offload when both are configured.

recipes/MiniMaxAI/MiniMax-H3.md
lishunyang12: Retitled this as a feature change;

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
lishunyang12: On-demand CPU staging is now opt-in through OffloadPlan.on_demand_component_paths.

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
lishunyang12: Added a warning when resident layers are requested without declared resident DiT paths.

vllm_omni/diffusion/models/minimax_h3/encoder.py
lishunyang12: Moved encoder block streaming into the DLO backend and declared the H3 paths in OffloadPlan.

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
lishunyang12: The backend now tries generic plan-driven encoder streaming before on-demand staging.

vllm_omni/diffusion/offloader/offload_plan.py
lishunyang12: Added OffloadPlan.encoder_block_attrs for declarative rank-local encoder block groups.

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
lishunyang12: Kept the explicit H3 opt-out so loader and backend share one safety gate until transformed mmap loading is implemented.

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
lishunyang12: The existing empty-block guard keeps unsupported DiTs resident instead of registering hooks.

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
lishunyang12: Same guard applies here; models with no discovered block list remain resident.

vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
lishunyang12: Guarded the prefix-slicing fast path so Ring dispatch always receives an explicit mask.

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
lishunyang12: Agreed; rather than relying on a missing producer, H3 now explicitly disables mmap loading.

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
lishunyang12: Renamed it PinnedResidentLayerGroup; it remains with DLO for now because it directly uses its distributed shard-and-pin helper.

lishunyang12:
@hsliuustc0106 Kept the standalone layerwise strategy for non-distributed use cases; H3 now uses the generic DLO path.

lishunyang12:
@david6666666 Implemented the plan-driven generic encoder streaming path; H3 now only declares its encoder block paths.

lishunyang12:
Follow-up plan: add declarative checkpoint-to-runtime adapters (including H3 QKV/MLP), rank-local mmap staging for no-AllGather TP, mmap staging for encoder/VAEs, GPU-only resident storage, and TP/quality/RSS regression coverage. The current H3 mmap opt-out remains intentional until this framework path is validated.

lishunyang12:
Tracking #5720 modular H3 pipeline compatibility in 44da2441: consumer-GPU recipes remain single-task only (`--task-type fl2va`/`ref2va` after #5720), and the DLO plan has a TODO to derive staging from the selected partition so combined mode never stages both DiTs unintentionally.

hsliuustc0106 (approved):
it seems we have not updated the docs for cpu_offload and we can leave it after the H3 supports PR

PR #5768, opened 2026-08-04
[Perf] [MiniMax H3] Gate scheduler finite-tensor validation

Sy0307:
## Purpose

MiniMax H3's scheduler validates video and audio latents with `torch.isfinite(...).all().item()` on every denoise update. These checks do not participate in the scheduler math, but each one scans a full tensor and synchronizes the host with the device.

For a joint video/audio update, the previous path performs 12 full-tensor finite reductions across `xt`, velocity, denoised state, scheduler input, and scheduler output. A 49-update request therefore performs 588 large-tensor scans.

This PR:

- skips the large latent finite scans by default;
- keeps shape, dtype, scalar timestep, and sigma validation enabled;
- reads the validation policy once before the MiniMax H3 denoise loop and passes it to both video and audio scheduler calls;
- allows full finite validation to be restored with `VLLM_OMNI_H3_VALIDATE_FINITE_TENSORS=1`;
- adds exact-output and NaN/Inf injection tests for both modes.

For finite inputs, the scheduler operations and outputs are unchanged. The intentional behavior change is that non-finite values in large latent tensors fail immediately only when diagnostic validation is enabled; otherwise they propagate through the existing math.

### Scheduler microbenchmark

Measured on one NVIDIA H20 with PyTorch `2.11.0+cu128`, at final PR head `2c1f2e32`. The benchmark models one 49-update request with production FL2VA latent shapes: video `[58032, 96]` and stereo audio `[696, 32]`. Each row uses 5 warmups and 30 paired, alternating ON/OFF rounds. Outputs were bitwise equal in both dtypes.

| Dtype | Finite validation ON, CUDA median | Default OFF, CUDA median | Removed scheduler time | Scheduler-only reduction |
| FP32 | 51.786 ms | 18.321 ms | 33.465 ms | 64.6% (2.83x) |
| BF16 | 51.515 ms | 21.180 ms | 30.335 ms | 58.9% (2.43x) |

Command:

```bash
CUDA_VISIBLE_DEVICES=0 python3 /tmp/h3_scheduler_gate_microbench.py \
  --repo /path/to/vllm-omni \
  --device cuda:0 \
  --dtype bfloat16 \
  --video-rows 58032 \
  --audio-rows 696 \
  --steps 49 \
  --warmups 5 \
  --rounds 30
```

This is a scheduler-only measurement. No end-to-end percentage is claimed: the removed work is about 30-33 ms per 49-update request at this shape, while the full DiT dominates E2E latency.

### Production-shape serving parity

On 2x H20 with TP2, text-encoder TP2, FLASH_ATTN, regional compile, and FL2VA `1248x768`, 209 frames, seed 2101, a 2-step diagnostic ON/OFF run produced byte-identical MP4 output:

- SHA256: `9a8767f37514d9594814ae2d0a2b513940ca24935d930a4e1bf84756b48c0a0e`
- container: 209-frame H.264 video at `1248x768`, plus 32 kHz stereo AAC audio
- peak memory: 82,244 MB in both modes

This full serving run was performed at `ff7fd2d4`. The final rebase to `2c1f2e32` added only upstream MOSS-TTS changes; exact-head scheduler tests and microbenchmarks were rerun after that rebase. The serving timings (`82.466s` diagnostic ON and `83.453s` default OFF) are reported only as parity evidence and are not used as a performance claim.

## Test Plan

**Final head:** `2c1f2e32f7597bdbd506d044afe585e35940abfb`  
**Remote environment:** NVIDIA H20, PyTorch `2.11.0+cu128`, vLLM `0.26.0`

- [x] Focused finite-gate tests: `3 passed`
- [x] MiniMax-H3 non-full-model suite: `71 passed, 2 skipped, 1 deselected`
- [x] Production-shape FP32/BF16 CUDA microbenchmarks: bitwise equal
- [x] Production-shape serving ON/OFF: byte-identical H.264/AAC MP4
- [x] `ruff check` on all three changed files
- [x] `ruff format --check` on all three changed files
- [x] `python -m compileall` on all three changed files
- [x] `git diff --check origin/main...HEAD`

(anything written below this line will be removed by GitHub Actions


hsliuustc0106:
Verified locally on one NVIDIA L20X using the stated production latent shapes, 49 updates, 5 warmups, and 30 alternating rounds:

- FP32: 51.877 ms validation ON vs 19.021 ms OFF, saving 32.855 ms (63.3%, 2.73x).
- BF16: 52.972 ms validation ON vs 22.347 ms OFF, saving 30.624 ms (57.8%, 2.37x).
- Current `main`, PR diagnostic ON, and PR default OFF produced bitwise-identical scheduler outputs in both dtypes.
- The current MiniMax-H3 core suite passed: 96 passed, 3 skipped.

The scheduler performance and finite-input parity claims reproduce. Independent verification covered the scheduler-level A/B and unit suite; I did not rerun the 2xH20 209-frame serving parity test.

Follow-ups:

1. The PR description says NaN/Inf injection is covered in both modes, but `test_scheduler_finite_gate_rejects_nonfinite_inputs` only runs with `VLLM_OMNI_H3_VALIDATE_FINITE_TENSORS=1`. Please add the default-OFF case and assert that non-finite values propagate through the existing math.
2. Clarify whether this environment switch is supported user-facing behavior. If so, document the default and diagnostic rollback in `recipes/MiniMaxAI/MiniMax-H3.md`; if it is internal-only, keep the constant/parser out of `__all__`.
3. Please include or link `h3_scheduler_gate_microbench.py`; the command currently points to a non-reproducible `/tmp` path.

david6666666:
Independent verification on a B300 single GPU (vllm 0.26.0, torch 2.11.0+cu130, FLASH_ATTN, eager, t2va 5s @1344x768 50 steps, seed 1101) vs current main (a874b8e0), using the FL2VA partition:
- Accuracy: default OFF and ON (VLLM_OMNI_H3_VALIDATE_FINITE_TENSORS=1) outputs are both bit-exact vs main (max frame diff 0, audio identical); ON vs OFF is also byte-identical, confirming the parity claim.
- Scheduler microbench at production shapes (video [58032,96], audio [696,32], 49 updates, BF16): ON 38.4ms vs OFF 14.6ms -> ~23.8ms saved per request (~62% scheduler-only), consistent with the H20 numbers in the body.
- E2E: same-GPU back-to-back ON vs OFF delta ~+0.15s on a ~157s run - negligible, matching the 'no e2e percentage' caveat.
One note: the earlier parallel-wave ON run looked +4.4s slower, which was inter-GPU noise, not the validation cost.

PR #5779, opened 2026-08-04
[Attention] Refine TRTLLM attention support for MiniMax H3

### LINKED ISSUE #5771: [Attention] MiniMax-H3 packed padding is incompatible with TRTLLM_ATTN
## Summary

MiniMax-H3 cannot currently run with `TRTLLM_ATTN`, which prevents validating the acceleration from #5283 (TRTLLM diffusion attention with Skip-Softmax) and #5509 (typed SAGE for `TRTLLM_ATTN`).

This appears to be an integration/metadata contract issue in the TRTLLM attention backend rather than a missing or broken SAGE cubin.

## Reproduction

Environment:

- Model: MiniMax-H3 `FL2VA`
- Hardware: 4 x NVIDIA B300 SXM6 AC, compute capability 10.3
- PyTorch: 2.11.0+cu130
- FlashInfer: 0.6.16.post1
- Resolution: 1248 x 768
- Duration: 8.7 seconds, 209 frames
- Inference steps: 2 for the minimal smoke test
- Parallel config: TP1, Ulysses4, Ring1, text encoder TP1, VAE tile4
- Regional compile enabled

The following configurations all fail at the first packed attention call:

1. `TRTLLM_ATTN + skip_softmax`
2. `TRTLLM_ATTN + quant={dtype_qk: fp8_e4m3, q_block_size: 1, k_block_size: 16}`
3. The combined Skip-Softmax + SAGE configuration

Observed error:

```
ValueError: TRTLLM_ATTN does not support attn_mask (Wan sets one only under SP with mask_sp_padding=True and a non-divisible seq len). Either set parallel_config.mask_sp_padding=False, or select a mask-capable backend (e.g. CUDNN_ATTN / TORCH_SDPA).
```

## Root cause

For this H3 workload, the packed sequence contains:

- used tokens: 58758
- aligned sequence length: 58816
- tail padding: 58 tokens

The H3 packed attention path creates a prefix-valid mask:

```python
attn_mask = torch.arange(packed_total, device=q.device)[None] < used
```

and passes it through `AttentionMetadata`.

`TRTLLM_ATTN.forward_cuda()` rejects any non-None `attn_mask` before invoking the TRTLLM/SAGE kernel. In the SAGE test, the SM103 SAGE cubin was successfully downloaded and loaded, so this failure occurs before the SAGE kernel executes.

The Skip-Softmax path also logs that MiniMax-H3 does not publish a denoise timestep, so the configured timestep gate remains dense:

```
TRTLLM skip: disabled_until_timestep=0.94 set but this pipeline does not publish denoise_timestep; staying dense.
```

## Exact code locations

- Mask creation: `vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py`, `_run_packed_attention()`.
- Backend rejection: `vllm_omni/diffusion/attention/backends/trtllm_attn.py`, `TrtllmAttentionImpl.forward_cuda()`.
- Timestep integration: `vllm_omni/diffusion/forward_context.py` provides `DenoiseProgressMixin.record_denoise_step()`; the MiniMax-H3 denoise loop/pipeline does not currently call it. Wan pipelines provide the reference integration pattern.

## Suggested fix

The H3 mask is only structural suffix padding, so a low-level kernel change may not be necessary.

Possible backend/integration fix:

1. Detect a prefix-valid mask in `trtllm_attn.py`.
2. Convert it to a valid sequence length, or trim the padded Q/K/V tokens before invoking TRTLLM/SAGE.
3. Restore the original packed output shape after attention.
4. Add denoise timestep publication in the H3 denoise loop so Skip-Softmax can activate.

Simply dropping `attn_mask` is unsafe unless the padded tokens are excluded from attention by another equivalent mechanism.

## Baseline

The same recipe configuration with `FLASH_ATTN` completes successfully on this workload. The measured steady-state E2E latency is 79.65 seconds.

The expected acceptance test is a 50-step H3 recipe A/B comparison with identical outputs/media validation for:

- FLASH_ATTN baseline
- TRTLLM_ATTN dense
- TRTLLM_ATTN + Skip-Softmax
- TRTLLM_ATTN + typed SAGE
- TRTLLM_ATTN + Skip-Softmax + typed SAGE

bobboli:
## Summary

- reuse packed `cu_seqlens` metadata in the TRTLLM ragged attention call
- recognize only prefix-valid structural padding masks
- trim invalid suffix tokens before dense or SAGE attention and restore the physical tensor shape with zero padding
- use dense TRTLLM for auxiliary sequences shorter than one SAGE K quantization block
- retain rejection of arbitrary masks and incomplete or invalid packed metadata
- make TRTLLM the MiniMax H3 default

## Why

MiniMax-H3 packs 58,758 valid tokens into a 58,816-token aligned tensor. Passing the 58 padding tokens as another valid sequence works for dense attention but lets SAGE quantize invalid values and can produce non-finite output. Trimming the structural suffix keeps padding out of both quantization and attention while restoring the aligned shape required by Ulysses communication.

H3 also has a 14-token auxiliary attention site. FlashInfer SAGE with `k_block_size=16` returns non-finite output for a sequence shorter than one quantization block, so that call uses dense TRTLLM while the main DiT attention remains SAGE.

Fixes #5771

## Testing

B300 SXM6 AC, 4-way Ulysses, Ring1, 1248x768, 209 frames, 2 inference steps, regional compile enabled, FlashInfer 0.6.16.post1:

- focused B300 tests: `37 passed, 1 skipped`; the skipped INT8 SAGE kernel is unavailable on SM103
- TRTLLM dense: HTTP 200, valid 209-frame MP4 with AAC audio
- TRTLLM + Skip-Softmax: HTTP 200, valid 209-frame MP4 with AAC audio
- TRTLLM + FP8 SAGE: HTTP 200, valid 209-frame MP4 with AAC audio
- TRTLLM + Skip-Softmax + FP8 SAGE: HTTP 200, valid 209-frame MP4 with AAC audio
- `ruff check`
- `ruff format --check`
- `git diff --check`

FlashAttention baseline follow-up on the same B300 node:

- The official vLLM 0.26.0 image did not include `flash-attn-4`, so `FLASH_ATTN` fell back to Hopper-only `fa3-fwd 0.0.3` and failed on SM103 with `no kernel image is available`.
- After installing `flash-attn-4[cu13]==4.0.0b18`, vLLM-Omni selected `CuTe FlashAttention-4 on Blackwell`; a direct BF16 FA4 kernel call also returned finite output.
- 50-step FA4 baseline with the same prompt/seed and 1248x768, 209-frame shape: model stage 88.98 s, HTTP 200, valid H.264/AAC MP4, and coherent multi-frame visual output with a non-silent audio track.

Dense TRTLLM 50-step follow-up on the same B300 node:

- Same prompt, seed, 1248x768 resolution, 209-frame duration, and 50 steps as the FA4 baseline: HTTP 200, valid H.264/AAC MP4, coherent multi-frame output, and non-silent audio.
- Encoded-video comparison against FA4: average PSNR 27.10 dB and SSIM 0.8880. The outputs are not pixel-identical but preserve the same scene, motion, and composition.
- This was a cold TRTLLM service run and included regional/VAE compilation plus slow CPU MP4 encoding, so its end-to-end latency is not reported as steady-state backend performance.

### DIFF
--- a/vllm_omni/diffusion/attention/backends/trtllm_attn.py
+++ b/vllm_omni/diffusion/attention/backends/trtllm_attn.py
@@ -255,8 +255,16 @@ class TrtllmAttentionImpl(AttentionImpl):
         value: torch.Tensor,
         attn_metadata: AttentionMetadata | None = None,
     ) -> torch.Tensor:
+        extra = getattr(attn_metadata, "extra", {}) if attn_metadata is not None else {}
+        packed_keys = ("cu_seqlens_q", "cu_seqlens_k", "max_seqlen_q", "max_seqlen_k")
+        present_packed_keys = [key for key in packed_keys if key in extra]
+        if present_packed_keys and len(present_packed_keys) != len(packed_keys):
+            missing = sorted(set(packed_keys) - set(present_packed_keys))
+            raise ValueError(f"Incomplete packed TRTLLM attention metadata; missing {missing}")
+        has_packed_metadata = len(present_packed_keys) == len(packed_keys)
+
         attn_mask = getattr(attn_metadata, "attn_mask", None) if attn_metadata is not None else None
-        if attn_mask is not None:
+        if attn_mask is not None and not has_packed_metadata:
             raise ValueError(
                 "TRTLLM_ATTN does not support attn_mask (Wan sets one only under SP with "
                 "mask_sp_padding=True and a non-divisible seq len). Either set "
@@ -270,24 +278,41 @@ class TrtllmAttentionImpl(AttentionImpl):
                 "another backend via --diffusion-attention-backend."
             )
 
-        batch, q_len, num_q_heads, head_dim = query.shape
+        physical_batch, q_len, num_q_heads, head_dim = query.shape
         kv_len, num_kv_heads = key.shape[1], key.shape[2]
 
         device = query.device
 
-        q = query.reshape(batch * q_len, num_q_heads, head_dim).contiguous()
-        k = key.reshape(batch * kv_len, num_kv_heads, head_dim).contiguous()
-        v = value.reshape(batch * kv_len, num_kv_heads, head_dim).contiguous()
+        q = query.reshape(physical_batch * q_len, num_q_heads, head_dim).contiguous()
+        k = key.reshape(physical_batch * kv_len, num_kv_heads, head_dim).contiguous()
+        v = value.reshape(physical_batch * kv_len, num_kv_heads, head_dim).contiguous()
 
-        seq_lens = torch.full((batch,), kv_len, dtype=torch.int32, device=device)
-        cu_seq_lens_q = torch.arange(0, (batch + 1) * q_len, step=q_len, dtype=torch.int32, device=device)
-        cu_seq_lens_kv = torch.arange(0, (batch + 1) * kv_len, step=kv_len, dtype=torch.int32, device=device)
+        if has_packed_metadata:
+            cu_seq_lens_q = extra["cu_seqlens_q"]
+            cu_seq_lens_kv = extra["cu_seqlens_k"]
+            if cu_seq_lens_q.ndim != 1 or cu_seq_lens_kv.ndim != 1:
+                raise ValueError("Packed TRTLLM attention cu_seqlens tensors must be one-dimensional")
+            if cu_seq_lens_q.numel() != cu_seq_lens_kv.numel() or cu_seq_lens_q.numel() < 2:
+                raise ValueError("Packed TRTLLM attention requires matching non-empty Q and KV sequence batches")
+            if int(cu_seq_lens_q[-1].item()) != q.shape[0] or int(cu_seq_lens_kv[-1].item()) != k.shape[0]:
+                raise ValueError("Packed TRTLLM attention cu_seqlens must cover all Q/K/V tokens")
+            batch = cu_seq_lens_q.numel() - 1
+            seq_lens = (cu_seq_lens_kv[1:] - cu_seq_lens_kv[:-1]).to(dtype=torch.int32).contiguous()
+            max_q_len = int(extra["max_seqlen_q"])
+            max_kv_len = int(extra["max_seqlen_k"])
+        else:
+            batch = physical_batch
+            seq_lens = torch.full((batch,), kv_len, dtype=torch.int32, device=device)
+            cu_seq_lens_q = torch.arange(0, (batch + 1) * q_len, step=q_len, dtype=torch.int32, device=device)
+            cu_seq_lens_kv = torch.arange(0, (batch + 1) * kv_len, step=kv_len, dtype=torch.int32, device=device)
+            max_q_len = q_len
+            max_kv_len = kv_len
         workspace = self._get_workspace(device)
 
         bmm1_scale = self.softmax_scale
         bmm2_scale = 1.0
 
-        _skip_factor = self._resolve_skip_factor(kv_len)
+        _skip_factor = self._resolve_skip_factor(max_kv_len)
 
         # SAGE kwargs are only understood by newer FlashInfer builds; pass them exclusively when
         # SAGE quant is active (which already requires the kernel, checked at init) so the dense
@@ -304,8 +329,8 @@ class TrtllmAttentionImpl(AttentionImpl):
             value=v,
             workspace_buffer=workspace,
             seq_lens=seq_lens,
-            max_q_len=q_len,
-            max_kv_len=kv_len,
+            max_q_len=max_q_len,
+            max_kv_len=max_kv_len,
             bmm1_scale=bmm1_scale,
             bmm2_scale=bmm2_scale,
             o_sf_scale=-1.0,
@@ -319,4 +344,4 @@ class TrtllmAttentionImpl(AttentionImpl):
             skip_softmax_threshold_scale_factor=_skip_factor,
             **sage_kwargs,
         )
-        return out.reshape(batch, q_len, num_q_heads, head_dim)
+        return out.reshape(physical_batch, q_len, num_q_heads, head_dim)
--- a/vllm_omni/diffusion/attention/backends/trtllm_attn.py
+++ b/vllm_omni/diffusion/attention/backends/trtllm_attn.py
@@ -298,6 +298,16 @@ class TrtllmAttentionImpl(AttentionImpl):
             if int(cu_seq_lens_q[-1].item()) != q.shape[0] or int(cu_seq_lens_kv[-1].item()) != k.shape[0]:
                 raise ValueError("Packed TRTLLM attention cu_seqlens must cover all Q/K/V tokens")
 
+            # Drop trailing empty sequences retained by packed producers when
+            # alignment adds no padding tokens.
+            while (
+                cu_seq_lens_q.numel() > 2
+                and int(cu_seq_lens_q[-1].item()) == int(cu_seq_lens_q[-2].item())
+                and int(cu_seq_lens_kv[-1].item()) == int(cu_seq_lens_kv[-2].item())
+            ):
+                cu_seq_lens_q = cu_seq_lens_q[:-1]
+                cu_seq_lens_kv = cu_seq_lens_kv[:-1]
+
             # A packed prefix mask denotes structural suffix padding, not another
             # document. Exclude it before quantization and attention; in particular,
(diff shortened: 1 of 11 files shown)


lishunyang12:
LGTM, let's make trtllm as the default backend for SM100.

lishunyang12 (approved):

david6666666 (commented):
LGTM

lishunyang12 (commented):
Follow-up from `lishunyang12`:

I prepared [commit `20cc23ae`](https://github.com/lishunyang12/vllm-omni/commit/20cc23ae) on the `agent/minimax-h3-trtllm-default` branch.

It makes MiniMax H3 declare its packed-sequence contract, so the platform auto-selects dense BF16 `TRTLLM_ATTN` on supported datacenter Blackwell (SM100/SM103) when FlashInfer's trtllm-gen kernel is available. Workstation Blackwell (SM120/SM121), unsupported head dimensions, and mask-using paths retain their normal fallback. No Skip-Softmax or attention quantization is enabled by default.

Validation:
- `41 passed, 2 skipped`
- Ruff check/format and `git diff --check` passed
- Matched 4x B300 SM103 run: 1248x768, 209 frames, 50 steps, seed 1101, Ulysses4/Ring1/TP1, VAE tile4, regional compile
- One compile warmup excluded; steady-state results:
  - FA4: 83.854 s diffusion / 88.558 s wall
  - Dense BF16 TRTLLM: 71.990 s diffusion / 76.176 s wall
  - TRTLLM is 14.1% faster for diffusion and 14.0% faster end to end in this workload

The complete test data and reproduction scripts are in [vllm-omni-rankings/scripts/minimax_h3_trtllm_e2e](https://github.com/lishunyang12/vllm-omni-rankings/tree/main/scripts/minimax_h3_trtllm_e2e):

- [FA4 E2E video](https://github.com/lishunyang12/vllm-omni-rankings/blob/main/scripts/minimax_h3_trtllm_e2e/minimax_h3_t2va_fa4_4xb300_8p7s_50step.mp4)
- [TRTLLM dense BF16 E2E video](https://github.com/lishunyang12/vllm-omni-rankings/blob/main/scripts/minimax_h3_trtllm_e2e/minimax_h3_t2va_trtllm_dense_bf16_4xb300_8p7s_50step.mp4)
- [Serve script](https://github.com/lishunyang12/vllm-omni-rankings/blob/main/scripts/minimax_h3_trtllm_e2e/serve_b300.sh)
- [Curl T2VA request 1](https://github.com/lishunyang12/vllm-omni-rankings/blob/main/scripts/minimax_h3_trtllm_e2e/curl_t2va.sh)
- [Offline A/B benchmark script](https://github.com/lishunyang12/vllm-omni-rankings/blob/main/scripts/minimax_h3_trtllm_e2e/run_attention_ab.py)
- [README and exact environment/results](https://github.com/lishunyang12/vllm-omni-rankings/blob/main/scripts/minimax_h3_trtllm_e2e/README.md)

The uploaded videos are combined packed-H3 optimization evidence rather than an isolated run of #5779 alone; the follow-up commit is ready to cherry-pick if desired.

lishunyang12 (approved):

vllm_omni/diffusion/attention/backends/trtllm_attn.py
```diff
@@ -270,30 +278,79 @@ def forward_cuda(
...
+        output_tokens = q.shape[0]
+
+        if has_packed_metadata:
+            cu_seq_lens_q = extra["cu_seqlens_q"]
+            cu_seq_lens_kv = extra["cu_seqlens_k"]
+            if cu_seq_lens_q.ndim != 1 or cu_seq_lens_kv.ndim != 1:
+                raise ValueError("Packed TRTLLM attention cu_seqlens tensors must be one-dimensional")
+            if cu_seq_lens_q.numel() != cu_seq_lens_kv.numel() or cu_seq_lens_q.numel() < 2:
+                raise ValueError("Packed TRTLLM attention requires matching non-empty Q and KV sequence batches")
+            if int(cu_seq_lens_q[-1].item()) != q.shape[0] or int(cu_seq_lens_kv[-1].item()) != k.shape[0]:
```
hsliuustc0106, line 300: [P1] This validation breaks H3 with AllGather-KV SP. `AllGatherKVParallelAttention` gathers K/V but leaves `extra["cu_seqlens_q"]` global while Q remains rank-local. With `allgather_degree > 1`, this compares a global final offset against local Q tokens and raises; padded H3 inputs also retain a global 2-D mask. Please slice/rebase the packed Q metadata and mask for the local shard, or explicitly disable the H3 TRTLLM default for AllGather-KV.

vllm_omni/diffusion/attention/backends/trtllm_attn.py
```diff
@@ -270,30 +278,79 @@ def forward_cuda(
...
+        _skip_factor = self._resolve_skip_factor(max_kv_len)
 
         # SAGE kwargs are only understood by newer FlashInfer builds; pass them exclusively when
         # SAGE quant is active (which already requires the kernel, checked at init) so the dense
         # path stays compatible with older builds that lack these parameters.
         sage_kwargs: dict = {}
-        if self.quant.enabled:
+        # The SAGE kernel requires every KV sequence to contain at least one full
+        # quantization block. Small auxiliary attention sites use the dense kernel.
+        use_sage = self.quant.enabled and bool(torch.all(seq_lens >= self.quant.k_block_size).item())
```
hsliuustc0106, line 364: [P2] Exactly aligned packed inputs silently disable SAGE. When `used == packed_total`, H3 still supplies a duplicate terminal boundary (`[0, used, used]`), so the derived lengths include a zero-length sequence and this `torch.all(...)` condition is false. The old dense path quantized these tokens. Please collapse trailing empty segments before deriving `seq_lens`, or add an explicit aligned-path rule and regression test so packed metadata does not turn SAGE off.

vllm_omni/diffusion/attention/backends/trtllm_attn.py
bobboli: Thanks, confirmed. TRTLLM's packed path currently assumes that Q/KV metadata cover their tensors symmetrically, which is incompatible with AllGather-KV's rank-local Q and global K/V layout.

I pushed `fix(attention): reject TRTLLM AllGather-KV` to explicitly reject `TRTLLM_ATTN` with `allgather_degree > 1` during attention initialization. This covers both platform-default and explicitly selected TRTLLM backends. Full local-Q/global-KV packed metadata support can be handled separately.

While tracing this, I noticed that H3's packed FA4 path also appears to receive the global `cu_seqlens_q` after AllGather-KV, even though Q is rank-local, and it does not explicitly validate or slice that metadata. Is FA4 + AllGather-KV expected to support this H3 packed layout elsewhere, or should we guard that combination as well until local Q metadata is rebuilt?

vllm_omni/diffusion/attention/backends/trtllm_attn.py
bobboli: Fixed in `fix(attention): trim empty packed suffixes`: TRTLLM now normalizes `[0, N, N]` to `[0, N]`, so an empty alignment suffix no longer disables SAGE.

H3’s main DiT attention operates on the long packed multimodal sequence, while its token refiner attends only over the prompt tokens and can be shorter than SAGE’s 16-token K block requirement. Both previously used role `self`, causing the refiner to inherit the main DiT’s quantization and skip-softmax settings. The refiner now uses the dedicated role `minimax_h3.token_refiner`, so it can be configured independently without disabling these optimizations for the main DiT attention.

If SAGE quantization is configured but a runtime input is too short, TRTLLM falls back to unquantized attention and warns once from `forward`, including the affected role.

lishunyang12:
Readiness fixes are now pushed: corrected the H3 Skip-Softmax timestep to use `t = 1 - sigma`, restored the non-negative threshold contract, and rewrote the mismatched DCO author identity. Focused TRTLLM tests pass (34 passed, 4 skipped); required CI has rerun.

vllm_omni/diffusion/attention/backends/trtllm_attn.py
lishunyang12: Let's handle this compatiablility in the next pr, but i don't think all-gather kv is a optimized sol compared to ring-attn and ulysses as it require higher memory bandwidth and VRAM max usage, so we typically won't use all-gather kv.

PR #5709, opened 2026-08-03
[Model] Add MiniMax H3 T2VA accuracy test

david6666666:
## Summary

- add an end-to-end accuracy test for `MiniMaxAI/MiniMax-H3` FL2VA using the model repository's pinned `assets/t2va.mp4` reference
- reproduce the official 768p T2VA workload at 1344x768, 24 FPS, 243 frames, 50 denoising steps, and seed 0
- validate video/audio metadata and gate full-video similarity with SSIM >= 0.82 and PSNR >= 20 dB
- run the nightly test on 4x H100 with USP4, HSDP shard size 4, text-encoder TP4, and VAE patch parallelism 4

The model repository and golden video are pinned to revision `73372e6cf53e414edd3ab03e357717fb0602e758`.

## Local validation

Validated end to end on 4x NVIDIA B300 using the new pytest case and the HSDP4 server configuration.

| Configuration | Result | Model memory / GPU | Model-stage latency | SSIM | PSNR |
| --- | --- | ---: | ---: | ---: | ---: |
| Recipe topology, no HSDP | calibration run completed | 86.13 GiB | 157.51 s | 0.856172 | 22.307010 dB |
| HSDP shard size 4 | **PASS** | 39.85 GiB | 156.44 s | 0.843701 | 21.310656 dB |

```bash
CUDA_VISIBLE_DEVICES=0,1,2,3 \
VLLM_TEST_MINIMAX_H3_FL2VA_MODEL=/path/to/MiniMax-H3/FL2VA \
PYTHONPATH=$PWD \
/home/zjy/code/david/.venv/bin/python -m pytest -s -v \
  tests/e2e/accuracy/minimax_h3/test_minimax_h3_t2va_similarity.py \
  -m full_model --run-level full_model
```

Result: `1 passed in 361.09s`.

## Test plan

- [x] new accuracy test collected successfully
- [x] full 4x B300 end-to-end run with HSDP4
- [x] `pre-commit run --files` for all changed files
- [x] `git diff --check`

PR #5780, opened 2026-08-04
[Bugfix][XPU][Tests] Add tests/e2e/accuracy/__init__.py to fix pytest…

Joshna-Medisetty:
## Summary
Fixes a pytest collection crash on the XPU CI caused by two test packages sharing the top-level name `minimax_h3`.

## Problem
`tests/e2e/accuracy/minimax_h3/` (added in #5709) and `tests/diffusion/models/minimax_h3/` both resolve to a top-level `minimax_h3` package under pytest's default `prepend` import mode, because `tests/e2e/accuracy/` has no `__init__.py`. When the XPU CI runs `pytest -m "... and xpu and B60"` over the whole `tests/` tree, both are imported in one session and collide:

This aborts collection (exit 2) even though the selected tests live elsewhere. Other platforms (CUDA/AMD/NPU) don't hit it because they scope pytest to specific paths/files and never collect both dirs together.

## Fix
Add `tests/e2e/accuracy/__init__.py` so the package resolves as `tests.e2e.accuracy.minimax_h3`, distinct from `tests/diffusion/models/minimax_h3`. Consistent with the existing layout (`tests/` and `tests/e2e/` already have `__init__.py`) and the repo's `from tests.e2e.accuracy...` import convention.

## Before
Screenshot 2026-08-04 110153

## After

Screenshot 2026-08-04 110313

## Test
Full XPU CI (vLLM 0.26, Intel B60) passes: `core_model` 7 passed, `mxfp8_config` 24 passed, `advanced_model` 3 passed, `omni` 2 passed — 0 failures, 0 collection errors.

### DIFF
--- a/tests/e2e/accuracy/__init__.py
+++ b/tests/e2e/accuracy/__init__.py
@@ -1,0 +1,1 @@
+"""End-to-end accuracy tests."""


xuechendi (approved):
LGTM

Issue #5722, opened 2026-08-03
[RFC]: OpenAI-compliant /v1/audio/transcriptions with word-level alignment


TheCodeWrangler:
### Motivation.

If you just want a transcript, you don't need vLLM-Omni. `vllm serve Qwen/Qwen3-ASR-1.7B` already serves `/v1/audio/transcriptions` today.

Word timestamps are different. Qwen3-ASR can't produce them — it sets `supports_segment_timestamp = False`, and base vLLM rejects `verbose_json` for it. You need a second model, a forced aligner, running over the same audio.

Two models behind one endpoint is what vLLM-Omni is for. So I think ASR + aligner belongs here: one server that gives you a transcript with word timings, instead of two servers and glue code.

The pieces already exist. vLLM 0.26 registers both `Qwen3ASRForConditionalGeneration` and `Qwen3ASRForcedAlignerForTokenClassification`. #4034 already vendored the aligner here for TTS word timestamps. Upstream even defines `TranscriptionResponseVerbose.words` and `timestamp_granularities[]` — and never fills them in. Nobody has wired the two together and pointed them at input audio.

So most of this is composition. The interesting part is making it fast enough to be worth using.

Draft: #5714.

### Proposed Change.

**Reach the endpoint.** `AsyncOmniEngine` hardcodes `supported_tasks` to `{generate, speech}`, so `api_server` never builds `OpenAIServingTranscription` and the route is dead under `--omni` for every model. Replace with a capability probe on vLLM's `supports_transcription()`.

The probe also requires the stage to be a *text final-output* stage, which is worth review. A request has no way to name a stage — `output_modalities` resolves to the last stage emitting that modality. `aura_omni` forced this: stage 0 is Qwen3-ASR, but stage 1 is the text terminal, so a transcription request there returns a conversational reply, not a transcript.

Supersedes the engine half of #3526, rebased, `Co-authored-by:` carried. Its other half (Qwen3-Omni wrapper delegation) is orthogonal and still worth landing.

**A `qwen3_asr` pipeline.** One stage, audio in, text out. HF `model_type` is `qwen3_asr`, so `vllm-omni serve Qwen/Qwen3-ASR-1.7B --omni` finds it.

**Fill in `words`.** Subclass upstream transcription serving, call the aligner when the request asks for word granularity. Off by default. Upstream refuses `verbose_json` without segment-timestamp support, so a word-only request runs the inner call as `json` and we build the verbose envelope.

### Performance.

[...]

Issue #4468, opened 2026-06-16
[RFC]: Auxiliary (non-generative) stage support in the Omni pipeline


wjinxu:
### Motivation.

The Omni stage orchestrator today only models **generative, linear** stages (autoregressive LLM and diffusion). PR #4034 adds a forced aligner that is a **non-generative** model (`runner="pooling"`, `pooling_task="token_classify"`), runs **once per finished sentence**, and returns its result **directly to the client** rather than feeding a downstream stage. Because nothing in the pipeline expresses that shape, #4034 ships the aligner as a standalone in-process `vllm.LLM()` in the API serving layer, which implicitly grabs the API server's first visible card and shares it with the TTS stages.

This RFC proposes first-class support for **auxiliary / non-generative stages** so models like the forced aligner (and future embedding / classify / reranker stages) can be placed, scheduled, and routed by the orchestrator instead of living outside it. #4034 stays as-is (opt-in, working); this is the follow-up to integrate it properly.

### Current stage model is generative-only

`StageExecutionType` has exactly three values, all generative (`vllm_omni/config/stage_config.py`, "3 combinations today"):

```python
class StageExecutionType(str, Enum):
    LLM_AR = "llm_ar"
    LLM_GENERATION = "llm_generation"
    DIFFUSION = "diffusion"
```

The orchestration loop routes outputs through exactly two paths — diffusion vs. LLM — with no third shape (`vllm_omni/engine/orchestrator.py`, `_orchestration_loop`):

```python
if pool.stage_type == "diffusion":
    output = pool.poll_diffusion_output(replica_id)
    if output is None:
        continue
    pool.record_output_timestamps([output])
    await self._handle_processed_outputs(stage_id, replica_id, [output])
    idle = False
else:
    raw_outputs = await pool.poll_llm_raw_output(replica_id, timeout_s=0.001)
    ...
```

## Goals

- Introduce an **auxiliary / non-generative stage** category the orchestrator can place, schedule, and route.
- Give such stages **first-class per-stage device placement** (explicit `devices:` like generative stages), instead of an implicit card-0 grab.
- Define a **terminal side-output** routing path: a stage whose result is collected once per segment and returned to the client, not fed downstream.
- Keep the change **additive**: existing generative pipelines and #4034's in-process fallback are unaffected.

### CC List.

@linyueqian @lishunyang12

### Before submitting a new issue...

[...]

PR #5781, opened 2026-08-04
[Core][Model] Qwen3 forced aligner as a pooling pipeline stage, with word timestamps

TheCodeWrangler:
Draft. Reimplements the forced aligner as a **pipeline stage** rather than the in-process sidecar in #5714, following the discussion on RFC #5722.

Stacks on #5714 (`feat/transcription-capability-probe`), which contributes the capability probe, the `qwen3_asr` topology and the serving layer. GitHub cannot base a PR on a fork branch, so this targets `main` and the diff includes #5714; **review the last seven commits** against that branch.

## What it does

Adds a `qwen3_asr_align` topology: Qwen3-ASR at stage 0, the Qwen3 forced aligner as a token-classification **pooling** stage at stage 1. `/v1/audio/transcriptions` populates the OpenAI `words[]` array from that stage rather than a second inference pass.

## What it took

A pooling stage was not merely unconfigured, it was unrepresentable. Three places assumed every stage is generative, and each produced a runtime failure rather than a configuration error:

| Where | Assumption | Symptom |
|---|---|---|
| `orchestrator.py` | `supported_tasks=("generate",)` hard-coded for downstream requests | `ValueError: This model does not support pooling` before the stage saw the request |
| `stage_init_utils.py` | always builds `SamplingParams` | pooling params never constructed |
| `stage_config.py` | omni AR scheduler for every LLM stage | schedules 0 tokens for a request with no decode loop, then hands the executor an empty batch |

Every branch keys off `runner: "pooling"`, which nothing currently sets, so generative pipelines are untouched.

Two things specific to this model. The checkpoint declares the *generative* architecture it shares with Qwen3-ASR, so the token-classification head is selected via an `hf_overrides` arch override. And the aligner needs the input audio, which is no stage's output: a rendered transcription prompt keeps only processed features, and the framework's default stage input processor looks for raw `multi_modal_data` that this entrypoint never sets. The stage input processor therefore splices its prompt onto stage 0's token ids through the audio span, so the `mm_features` the orchestrator forwards stay valid and the audio is encoded once for the whole pipeline. The two checkpoints have byte-identical preprocessor configs, so that reuse is sound rather than coincidental.

## Correctness

Validated against the sidecar implementation on the same clip: **78 words both ways, mean delta 0.0005s, max 0.080s** (two timestamp bins, batching nondeterminism).

Two bugs found on the way, both of which would have shipped silently:

- The stage segmented stage 0's *raw* generation. Qwen3-ASR emits `"language {lang}<asr_text>{transcription}"`, so `language`, `English`, `asr` and `text` were aligned as spoken words and every later timestamp landed on the wrong word. The word count still looked plausible, which is why it now has a regression test.
- Omni's prefix cache cannot serve a token-classification stage: it sizes slots to `hidden_size` but the head emits `classify_num`. Disabled for that stage, where it has nothing to reuse anyway.

36 tests pass, 29 of them new.

## Throughput

One H100, both models resident, 30s clips, concurrency 256, 3 runs, matched memory budgets (0.70 stage 0 + 0.15 aligner) on every arm.

Methodology matters more than usual here. Earlier numbers I posted on this PR were inflated roughly 2x because the harness replayed one audio file, so identical content hashes let vLLM's multimodal processor cache serve the encoder output on nearly every request. These runs use **2600 distinct 30s clips** with prefix caching and the mm processor cache disabled.

```
                     req/s   spread   RTFx   words
ASR only              53.9     24%    1618     -
stage + alignment     45.4     20%    1363    78
in-process sidecar    34.3     32%    1029    78
```

Alignment as a stage costs about **16%**, and the stage is about **1.3x the sidecar**. The stage gets vLLM's continuous batching; the sidecar batches through a hand-rolled queue-and-flush worker.

Caveat on precision: run-to-run spread is 20-32% even at 3 repeats, so treat these as directional. The stage-versus-sidecar gap is comfortably outside that band; the 16% alignment cost is not much larger than the noise and should be read as "roughly 10-20%".

## Known gaps

- **Returning the stage's result relies on a workaround.** Both stages are final outputs and upstream's transcription loop reads `.text` off whichever arrives, so the entrypoint splits the stream itself. This is the terminal side-output gap from RFC #4468, and the cleanest evidence I have for it: `OutputModality` deliberately excludes pooling ("vLLM-Omni's layer does not participate"), so there is no natural place for this today.
- Streaming responses are not aligned; that needs a per-sentence hook rather than a whole-response pass.

## Unrelated bug found while benchmarking

`vllm_omni/core/prefix_cache.py:291` `_consume_pending_write` raises `IndexError: index_copy_(): index -1 is out of bounds` and kills the stage-0 engine. Reproducible at concurrency 64 with a second GPU consumer in the process; ASR alone survives the same path, so it looks timing dependent. Not caused by anything here, and it will take down a server with prefix caching enabled. Filing separately.


TheCodeWrangler:
Updated: the feature is now complete end to end, and the stage beats the sidecar it replaces.

## Word timestamps work

`/v1/audio/transcriptions` now populates the OpenAI `words[]` array from the pooling stage rather than a second inference pass. Validated against the sidecar implementation on the same clip:

```
stage = 78 words    sidecar = 78 words
mean delta 0.0005s, max 0.080s (two timestamp bins)
```

Two bugs found on the way, both of which would have shipped silently:

- The stage was segmenting stage 0's *raw* generation. Qwen3-ASR emits `"language {lang}<asr_text>{transcription}"`, so `language`, `English`, `asr` and `text` were being aligned as spoken words and every timestamp after them landed on the wrong word. Caught by logging the constructed aligner prompt.
- Omni's prefix cache cannot serve a token-classification stage: it sizes slots to `hidden_size` but the head emits `classify_num`. Disabled for that stage, where it has nothing to reuse anyway.

## Throughput

One H100, both models resident, 30s clips, **prefix caching off**, each arm warmed then measured, peak of a 256/384 sweep:

```
ASR only                        100.5 req/s   RTFx 3014
stage + reused audio encoding    89.1 req/s   RTFx 2674    ~11% for alignment
stage, re-encoding the audio     83.4 req/s   RTFx 2501    ~17%
in-process sidecar               70.4 req/s   RTFx 2113    ~30%
```

The stage is **27% faster than the sidecar**. It gets vLLM's continuous batching instead of the sidecar's hand-rolled queue-and-flush worker, and that more than pays for the process boundary.

Reusing stage 0's audio encoding is worth ~7% of throughput, but its larger effect is stability: run-to-run spread drops from 42% to 12%. The two checkpoints have byte-identical preprocessor configs, so the second mel pass was pure duplicated work; the orchestrator already forwarded `mm_features` downstream, gated on a hard-coded `model_stage == "thinker"`.

Caveat on the ratios: ASR-only is the noisy arm (38-50% spread at these concurrencies) while the alignment arms sit at 7-12%, so read the alignment cost as roughly 11-16% rather than a precise figure.

## Unrelated bug found while benchmarking

`vllm_omni/core/prefix_cache.py:291` `_consume_pending_write` raises `IndexError: index_copy_(): index -1 is out of bounds` and kills the stage-0 engine. Reproducible at concurrency 64 with a second GPU consumer in the process; ASR-only survives the same path, so it looks timing dependent. Not caused by anything here, and it will take down a server with prefix caching enabled. I can file it separately if that is useful.

TheCodeWrangler:
Correction to the throughput numbers in my previous comment: **they were inflated by caching and should not be relied on.** New numbers to follow once a clean run finishes.

I disabled `enable_prefix_caching` and described the result as "no cache". I did not disable vLLM's multimodal processor cache, and the benchmark harness reads one audio file's bytes once and posts the identical payload for every request. Identical bytes mean an identical mm hash, so the audio encoder result was served from cache on essentially every request. For a 30s clip that is a large fraction of the work.

Measured just now on one H100, ASR only, 30s clips, concurrency 384:

```
repeated audio, prefix cache off              97.3 req/s     <- what I reported
repeated audio, prefix + mm cache off         55.7 req/s
256 distinct clips, prefix + mm cache off     58.3 req/s
```

So the honest ASR-only figure is around 58 req/s, not ~100. The mm cache was worth about 1.9x on its own.

Two things worth noting. Once the mm cache is off, repeated versus distinct audio barely matters (55.7 vs 58.3), which confirms the cache was the whole effect rather than something else keyed on content. And ~58 req/s closely matches a baseline I measured in an earlier session and then wrongly dismissed as environmental drift; that measurement was correct and today's were not.

Every arm in the previous comment was measured the same way, so the ranking may well survive, but I am not going to assert that until it is re-measured. The re-run uses 256 distinct clips (same speech, inaudible per-clip noise, so compute is unchanged and only the content hash differs) with both caches disabled on every model in the process, including the sidecar's aligner.

The functional results in that comment are unaffected: word timestamps still match the sidecar to a mean of 0.0005s, and the two bugs described are real.

PR #5782, opened 2026-08-04
[Bugfix][XPU] Fix diffusion output corruption from async D2H copy

Joshna-Medisetty:
## Summary
Diffusion output on XPU shows a band of corrupted pixels along the bottom
edge of generated images. This forces the blocking D2H copy on XPU to fix it.

## Root cause
The async output path (#4981) records a GPU event and copies the result
device→host on a side stream gated by that event. On XPU this cross-stream
ordering is unreliable, so the copy can start before the compute stream
finishes writing the last rows, reading stale memory.

## Fix
Disable the async D2H path on XPU in `_tensor_to_shm`  the single choke
point every large diffusion output passes through, so it falls back to the
existing synchronous `.cpu()` copy. On XPU the async overlap saves nothing
(millisecond copy vs seconds of compute).

## Scope
Applies to all diffusion modalities (image, video, audio) since they share
the `_tensor_to_shm` path. Verified on images only; the corruption has not
been observed on video or audio, so no before/after is included for those.

## Before / After bottom strips
Screenshot 2026-08-04 141740

Screenshot 2026-08-04 141814

Screenshot 2026-08-04 141828

## Test
Built from upstream `main` + this change and ran each model's standard CLI on
Intel XPU. The bottom-edge band is eliminated (last-row pixel discontinuity
drops to the image median): Z-Image-Turbo 23.5→2.3, ERNIE-Image 25–29→4,
FLUX.1-dev 14–33→0.3.

### DIFF
--- a/vllm_omni/diffusion/ipc.py
+++ b/vllm_omni/diffusion/ipc.py
@@ -16,6 +16,7 @@ from typing import Any
 import torch
 
 from vllm_omni.diffusion.data import DiffusionOutput
+from vllm_omni.platforms import current_omni_platform
 
 _SHM_TENSOR_THRESHOLD = 1_000_000  # 1 MB
 DIFFUSION_RPC_RESULT_ENVELOPE = "diffusion_rpc_result"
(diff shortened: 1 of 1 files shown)


vllm_omni/diffusion/ipc.py
```diff
@@ -40,6 +41,10 @@ def _tensor_to_shm(
     import numpy as np
 
     original_dtype = tensor.dtype
+    # XPU's stream synchronization is unreliable (see sequential_backend.py:79-82).
+    # Force the blocking .cpu().contiguous() path to avoid corrupted D2H transfers.
+    if d2h_stream is not None and current_omni_platform.is_xpu():
```
hsliuustc0106, line 46: can you extract a tiny help function?

Joshna-Medisetty: #5571 is a better fix for this issue. Please help review @hsliuustc0106

PR #5783, opened 2026-08-04
[Perf] [MiniMax H3] Precompute AdaLN modulation schedules

Sy0307:
## Purpose

MiniMax H3 evaluates 50 block AdaLN projections plus one final-layer projection on every denoise forward. These projections contain about 13B parameters globally, while their outputs depend only on the finite timestep schedule and modality, not on prompt or latent contents.

This PR adds an opt-in exact-schedule cache that computes the modulation rows once per schedule and reuses them across denoise forwards. It also adds an optional mode that moves the now-redundant projection weights to CPU and releases their CUDA storage after the tables are installed.

Both modes are disabled by default:

```bash
export VLLM_OMNI_H3_ADALN_SCHEDULE_CACHE=1
export VLLM_OMNI_H3_ADALN_OFFLOAD_WEIGHTS=1  # optional; requires the cache
```

## Design and safety

- Build request-level timestep routing once, removing the per-step packed-row `torch.unique` work.
- Key schedules by exact float32 timestep bit patterns; same-schedule requests reuse the installed tables.
- Precompute all 50 block tables and the final-layer table before changing the live forward path.
- On a schedule change, restore the projection weights, build and atomically install replacement tables, then offload the weights again.
- Fall back to direct projection when tables would occupy at least as much device memory as the projection weights.
- Leave DTensor/HSDP parameters under their distributed owner rather than manually offloading them.
- When layerwise offload is active, let the exact-schedule cache own `adaln_proj.linear.*` so the block hooks do not transfer the same parameters on every step.

The direct-projection path remains available whenever the feature is disabled or unsupported.

## Validation environment

- **vLLM-Omni commit:** `635215a96d3eaf6b5bac66020feae51b34daa5c9`
- **Runtime:** Python 3.12, PyTorch 2.11.0+cu128, vLLM 0.26.0
- **Hardware:** 2x NVIDIA H20 98 GB, physical GPUs 0 and 3
- **Model:** MiniMax H3 FL2VA
- **Serving path:** eager TP2 + text-encoder TP2, FlashAttention, one request in flight
- **Workload:** T2VA, 448x256, 29 requested / 39 aligned frames, 50 sigma points / 49 denoise forwards, 24 FPS, seed 42, video/audio flow shifts 12/3
- **Measurement:** one warmup or schedule-miss request followed by five steady requests

### Test results

- `31 passed`: focused AdaLN cache, H3 parallel ownership, and layerwise-offloader tests.
- `45 passed, 2 skipped, 1 deselected`: complete `tests/diffusion/models/minimax_h3 -m 'not full_model'` suite.
- `ruff check .`, changed-file `ruff format --check`, `compileall`, and `git diff --check` passed.
- GitHub build, pre-commit, DCO, Buildkite release, and documentation checks passed.

## E2E performance

All rows use the same current-SHA server path and fixed request. The weight-offload row reports schedule hits after its one-time 50-step table build.

| Mode | Steady wall mean | Server inference mean | Encode mean | Diffuse mean | Decode mean | Wall change |
| Cache disabled | 15.2858 s | 15.2828 s | 0.04256 s | 14.49705 s | 0.43516 s | baseline |
| Schedule cache, weights retained | 14.9879 s | 14.9850 s | 0.05281 s | 14.19402 s | 0.43558 s | -1.95% |
| Schedule cache + weight offload | 15.0294 s | 15.0254 s | 0.05531 s | 14.21263 s | 0.43550 s | -1.68% |

The retained and offloaded steady paths differ by 0.28% wall time. The offload option is therefore a memory/capacity feature; it preserves the cache speedup rather than claiming an additional latency gain on the resident TP2 path.

### Device memory

For the 50-step schedule, on each TP rank:

- projection weights released: **12.144 GiB**;
- installed tables: **0.885 GiB**;
- net device-memory reduction from code accounting: **11.259 GiB/GPU**.

The 50-step schedule-miss request peaked at 68,996 MB in the CUDA allocator while weights and replacement tables coexisted. After release, all five hits reported 56,866 MB. The sampled `nvidia-smi` build maximum was 71,623 / 71,363 MiB on GPUs 0/3; post-offload idle residency was 59,493 / 59,233 MiB. The optimization lowers steady residency, not the first-build peak.

### Correctness

- Every one of the 18 primary 50-step outputs across cache-off, retained-weight, and weight-offload runs had the same MP4 SHA256: `294c585b404159dff9958635c44f98fa101dfe8110adda985856a0dc17a8e28f`.
- Outputs are valid H.264/AAC MP4s: 39 frames at 24 FPS, 448x256 video, and 32 kHz stereo audio.
- The eager workload is therefore byte-identical across all three modes for this fixed seed and input.

## Schedule lifecycle

The same live service was switched across schedules to exercise weight restore, table replacement, and re-offload:

| Transition | Table size | Net device reduction | Miss wall | Following hit |
| cold 2-step | 0.009 GiB | 12.135 GiB/GPU | 12.44 s server E2E | completed |
| 2 -> 50 steps | 0.885 GiB | 11.259 GiB/GPU | 25.294 s | 15.025 s five-run mean |
| 50 -> 10 steps | 0.163 GiB | 11.981 GiB/GPU | 13.211 s | 3.412 s |
| 10 -> 50 steps | 0.885 GiB | 11.259 GiB/GPU | 24.948 s | 15.001 s |

The returned 50-step outputs retained the baseline SHA256. The roughly 10-second miss cost is expected because a changed schedule must restore the projections, rebuild all tables, and release the weights again; same-schedule serving avoids it.

## Layerwise-offload interaction

A real TP2 layerwise-offload service confirmed that all 50 blocks remain layerwise managed while `adaln_proj.linear.*` stays under the exact-schedule cache. Two 2-step requests were compared pairwise against a cache-disabled layerwise control; both corresponding MP4s were byte-identical. The cache-disabled control already has a small audio-only difference between its first and second request, and the cache/offload path reproduced the same two outputs exactly, so this PR adds no layerwise output regression.

## Reproduction

```bash
CUDA_VISIBLE_DEVICES=0,3 \
VLLM_OMNI_H3_ADALN_SCHEDULE_CACHE=1 \
VLLM_OMNI_H3_ADALN_OFFLOAD_WEIGHTS=1 \
python3 -m vllm_omni.entrypoints.cli.main serve /path/to/MiniMax-H3/FL2VA \
  --omni --host 127.0.0.1 --port 18093 --trust-remote-code \
  --num-gpus 2 --tensor-parallel-size 2 --text-encoder-tp-size 2 \
  --diffusion-attention-backend FLASH_ATTN \
  --enable-diffusion-pipeline-profiler --enforce-eager
```

The measured request used `/v1/videos/sync` with prompt `A quiet cinematic night scene with matching ambient sound.`, width 448, height 256, 29 frames, 50 inference steps, seed 42, flow shift 12, and `extra_params={"task":"t2va","audio_flow_shift":3.0}`.

## Test plan

- [x] Focused AdaLN cache, MiniMax H3 parallel, and layerwise-offload tests on CUDA
- [x] Complete MiniMax H3 non-full-model test suite
- [x] Cache disabled versus enabled byte-identical output parity on the fixed H20 workload
- [x] Retained weights versus weight offload memory and latency A/B
- [x] Same-schedule hit and changed-schedule restore/rebuild/re-offload
- [x] Real layerwise-offload interaction against a cache-disabled control
- [x] DTensor/HSDP ownership fallback tests

### DIFF
--- a/tests/diffusion/models/minimax_h3/test_minimax_h3_adaln_cache.py
+++ b/tests/diffusion/models/minimax_h3/test_minimax_h3_adaln_cache.py
@@ -1,0 +1,353 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+
+import pytest
+import torch
+
+pytestmark = [pytest.mark.core_model, pytest.mark.cpu, pytest.mark.diffusion]
+
+
+def _make_branch(layout: str):
+    from vllm_omni.diffusion.models.minimax_h3.denoise_loop import (
+        MiniMaxH3DenoiseBranch,
+    )
+
+    update_mask = {
+        "t2va": [True, True],
+        "fl2va": [False, True],
+        "ref2va": [True, True],
+    }[layout]
+    audio_update_mask = {
+        "t2va": [True, True],
+        "fl2va": [True, True],
+        "ref2va": [False, True],
+    }[layout]
+    packed = {
+        "seq_len": 8,
+        "text_pos": torch.tensor([0, 1]),
+        "img_pos": torch.tensor([2, 3]),
+        "audio_pos": torch.tensor([4, 5]),
+        "update_mask": torch.tensor(update_mask),
+        "audio_update_mask": torch.tensor(audio_update_mask),
+        "img_position_ids": torch.zeros(8, 3, dtype=torch.long),
+        "cu_seqlens": torch.tensor([0, 6, 8], dtype=torch.int32),
+    }
+    return MiniMaxH3DenoiseBranch(
+        packed=packed,
+        text_embeddings=torch.zeros(2, 4),
+        token_tags=torch.zeros(8, dtype=torch.long),
+        device=torch.device("cpu"),
+    )
+
+
+def _reference_full_vector_unique(
+    branch,
+    *,
+    sigma_video: float,
+    sigma_audio: float,
+    imgvid_cond_timestep: float,
+    audio_ref_cond_timestep: float,
+) -> tuple[torch.Tensor, torch.Tensor]:
+    t_video = 1.0 - sigma_video
+    t_audio = 1.0 - sigma_audio
+    timesteps = torch.full((branch.seq_len,), t_video, dtype=torch.float32)
+    timesteps[branch.img_pos[branch.update_mask]] = t_video
+    timesteps[branch.img_pos[~branch.update_mask]] = max(t_video, imgvid_cond_timestep)
+    timesteps[branch.audio_pos[branch.audio_update_mask]] = t_audio
+    timesteps[branch.audio_pos[~branch.audio_update_mask]] = max(t_audio, audio_ref_cond_timestep)
+    return torch.unique(timesteps, sorted=True, return_inverse=True)
+
+
+@pytest.mark.parametrize("layout", ["t2va", "fl2va", "ref2va"])
+def test_timestep_metadata_matches_full_vector_unique(layout: str):
+    branch = _make_branch(layout)
+    sigmas_video = [1.0, 0.5, 0.0]
+    sigmas_audio = [1.0, 0.25, 0.0]
+
+    cached = branch.prepare_timestep_metadata(
+        sigmas_video=sigmas_video,
+        sigmas_audio=sigmas_audio,
+        imgvid_cond_noise_aug_for_inference=0.999,
+        audio_cond_noise_aug_for_inference=1.0,
+    )
+
+    assert len(cached) == 2
+    for step, metadata in enumerate(cached):
+        expected_unique, expected_inverse = _reference_full_vector_unique(
+            branch,
+            sigma_video=sigmas_video[step],
+            sigma_audio=sigmas_audio[step],
+            imgvid_cond_timestep=0.999,
+            audio_ref_cond_timestep=1.0,
+        )
+        torch.testing.assert_close(metadata.unique_timesteps.cpu(), expected_unique, rtol=0, atol=0)
+        torch.testing.assert_close(metadata.inverse_indices.cpu(), expected_inverse, rtol=0, atol=0)
+
+
+def test_adaln_schedule_key_uses_exact_float32_bits():
+    from vllm_omni.diffusion.models.minimax_h3.adaln_schedule_cache import (
+        minimax_h3_adaln_schedule_key,
+    )
+
+    branch = _make_branch("t2va")
+    first = branch.prepare_timestep_metadata(
+        sigmas_video=[1.0, 0.5, 0.0],
+        sigmas_audio=[1.0, 0.25, 0.0],
+        imgvid_cond_noise_aug_for_inference=0.999,
+        audio_cond_noise_aug_for_inference=1.0,
+    )
+    same_after_float32_rounding = branch.prepare_timestep_metadata(
+        sigmas_video=[1.0, 0.5 + 1e-12, 0.0],
+        sigmas_audio=[1.0, 0.25 + 1e-12, 0.0],
+        imgvid_cond_noise_aug_for_inference=0.999,
+        audio_cond_noise_aug_for_inference=1.0,
+    )
+    changed = branch.prepare_timestep_metadata(
+        sigmas_video=[1.0, 0.5, 0.0],
+        sigmas_audio=[1.0, 0.25000003, 0.0],
+        imgvid_cond_noise_aug_for_inference=0.999,
+        audio_cond_noise_aug_for_inference=1.0,
+    )
+
+    first_key = minimax_h3_adaln_schedule_key(first)
+    assert first_key == minimax_h3_adaln_schedule_key(same_after_float32_rounding)
+    assert first_key != minimax_h3_adaln_schedule_key(changed)
+
+
+class _FakeAdalnLinear(torch.nn.Module):
+    def __init__(self, out_features: int):
+        super().__init__()
+        self.weight = torch.nn.Parameter(torch.arange(out_features * 4, dtype=torch.bfloat16).reshape(out_features, 4))
+        self.bias = torch.nn.Parameter(torch.arange(out_features, dtype=torch.bfloat16))
+        self.out_features = out_features
+        self.calls = 0
+
+    def forward(self, x):
+        assert self.weight.numel() > 0
+        assert self.bias.numel() > 0
+        self.calls += 1
+        rows = torch.arange(
+            x.shape[0] * self.out_features,
+            device=x.device,
+            dtype=torch.float32,
+        ).reshape(x.shape[0], self.out_features)
+        rows = rows + x.float().sum(dim=-1, keepdim=True)
+        return rows.to(torch.bfloat16), None
+
+
+def _make_tiny_adaln_projection(*, expand_ratio: int = 2, modality_num: int = 3):
+    from vllm_omni.diffusion.models.minimax_h3.minimax_h3_transformer import (
+        MiniMaxH3AdalnProj,
+    )
+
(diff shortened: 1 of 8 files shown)


Sy0307:
## Additional production-shape validation

Validated PR #5783 at commit `635215a96d3eaf6b5bac66020feae51b34daa5c9` on NVIDIA H20 GPUs.

### Environment

- Model: MiniMax H3 FL2VA
- Workload: T2VA, 1248x768, 209 frames, 24 FPS, 50 steps (49 denoise forwards), seed 2101, video/audio flow shifts 12/3
- Runtime: Python 3.12, PyTorch 2.11.0+cu128, vLLM 0.26.0
- Attention: FlashAttention

### TP2 regional-compile A/B

| Mode | Request | DiT / diffuse | Server E2E | Client wall | NVML peak GPU0/1 | Post-hit GPU0/1 |
| Cache disabled | steady | 1271.280 s | 1292.307 s | 1298.497 s | 84.85 / 90.46 GiB | 82.51 / 84.74 GiB |
| Cache retained | cache miss / compile warmup | 1275.460 s | 1301.098 s | 1307.617 s | — | — |
| Cache retained | steady hit | 1270.902 s | 1292.014 s | 1298.456 s | 83.83 / 85.76 GiB | 83.51 / 85.74 GiB |
| Cache + weight offload | cache miss / compile warmup | 1289.850 s | 1315.189 s | 1321.707 s | — | — |
| Cache + weight offload | steady hit | 1277.686 s | 1298.676 s | 1305.133 s | 71.57 / 73.81 GiB | 71.25 / 73.49 GiB |

The cache-disabled compile warmup took 1276.119 s in DiT and 1301.684 s server E2E; it is excluded from the steady-state comparison.

For cache-retained versus cache-disabled, the steady differences were only -0.03% DiT, -0.02% server E2E, and -0.003% client wall: no distinguishable latency gain in this one-sample production run. The retained tables increased post-request NVML residency by 1020 MiB per GPU.

For cache+weight-offload versus cache-disabled, the steady run was 0.50% slower in DiT, 0.49% slower in server E2E, and 0.51% slower in client wall, while post-request NVML residency dropped by 11.256 GiB per GPU. The implementation log independently reported 12.144 GiB of AdaLN weights replaced by 0.885 GiB of tables, or 11.259 GiB net per GPU. This mode is therefore a capacity tradeoff on TP2, not a latency optimization in this workload.

The retained miss added 4.558 s of DiT time relative to its steady hit. The cache+offload miss added 12.164 s relative to its steady hit, including the one-time table build and weight transfer to CPU.

### Layerwise-offload A/B

This is a production-shape 2-step (one denoise forward) clean A/B in eager mode on one H20. It is intentionally short because a 50-step layerwise request would take about 40 minutes per sample on this machine.

| Mode | Request | DiT / diffuse | Server E2E | Client wall |
|---|---:|---:|---:|---:|
| Cache disabled | clean baseline | 53.156 s | 126.931 s | 133.276 s |
| Cache + weight offload | cache miss | 67.398 s | 135.705 s | 141.919 s |
| Cache + weight offload | steady hit | 52.774 s | 126.338 s | 132.576 s |

On the steady hit, DiT improved by 0.72%, server E2E by 0.47%, and client wall time by 0.53% versus the clean baseline. This is one short sample, so it is directional rather than a standalone throughput claim.

The runtime confirmed that the cache+offload path excluded `adaln_proj.linear.*` from layerwise movement and offloaded 24.288 GiB of weights after building 0.009 GiB of tables (24.279 GiB net reduction from the temporary pre-build residency). NVML observed an 85.83 GiB cold peak before the first offload, then 29.10 GiB after the steady hit; the baseline ended at 30.22 GiB. The cold peak matters: this optimization does not help a device that cannot fit the one-time table-build phase.

### Correctness artifacts

- Layerwise steady baseline and cache+offload hit are byte-identical MP4 files: `d209897836fedfe0d515533f59dcc6e7638ab7720d5d05acc0d4bf2b902d4b7d`.
- Their decoded video streams are also identical: `3853f04ad55b49576c6df1cdc5bd69fd73421974c0cdf0c8d92d64656ebf6d8e`.
- Their decoded audio streams are identical: `4c24d0fda05fcd25cb9a086ddc80e1300656b722e845aa9e0dbb7337e5f8849f`.
- The cold miss has the same decoded video hash but a different audio hash. The same cold-to-steady audio difference occurs with the cache disabled, so cold requests are not used as the parity pair.
- The three TP2 steady outputs (cache disabled, retained, and cache+offload) are byte-identical MP4 files: `ae19eaec29833f095aefd728d745d0e9721e4d67a5fc8d839bc2a970f15acb13`.
- TP2 output metadata: H.264 video, 1248x768, 209 frames at 24 FPS; AAC audio, 32 kHz stereo; 8.732 s container duration.

### Notes

- The TP2 production-shape run uses regional `torch.compile` and excludes the first compile/cache-miss request from steady-state timing.
- The layerwise-offload run uses eager mode and reports its exact step count explicitly.
- NVML values are process-level device residency samples, not PyTorch allocator-only accounting.
- Each production TP2 steady mode has one measured sample because each request takes about 21.6 minutes. Sub-1% latency deltas should be treated as directional; the bitwise parity and 11.259 GiB-per-GPU offload result are deterministic observations.
- All claims are limited to the hardware and workload above.

vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
```diff
@@ -935,6 +1058,126 @@ def post_load_weights(self) -> None:
...
+
+        for projection in self._adaln_projections:
+            projection.restore_projection()
+
+        t_embs = tuple(self.time_embedder(timestep.view(-1)) for timestep in unique_timestep_plan)
+        device = t_embs[0].device
+        if any(t_emb.device != device for t_emb in t_embs):
+            raise ValueError("AdaLN timestep embeddings must share one device")
+        step_cursor = torch.zeros((), dtype=torch.long, device=device)
+        tables = tuple(minimax_h3_build_adaln_table(projection, t_embs) for projection in self._adaln_projections)
```
lishunyang12, line 1099: Unshard each HSDP block before building its AdaLN table.

vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
```diff
@@ -528,14 +538,118 @@ def __init__(
...
+        }
+        for name, tensor in restored.items():
+            self.linear.get_parameter(name).data = tensor
+        self._offloaded_projection_state = None
+
+    def compute_flat(self, t_emb: torch.Tensor) -> torch.Tensor:
+        """Run the original projection and return `[M*modalities, ratio*H]`."""
         x = nn.functional.silu(t_emb)
-        x, _ = self.linear(x.to(_BF16_DTYPE))
+        x, _ = self.linear(x.to(self.linear.weight.dtype))
```
lishunyang12, line 607: Keep activations and cached tables in BF16 when projection weights are FP8.

tests/diffusion/models/minimax_h3/test_minimax_h3_adaln_cache.py
```diff
@@ -0,0 +1,409 @@
...
+    first_plan = _schedule(branch, middle_video=0.5, middle_audio=0.25)
+    changed_plan = _schedule(branch, middle_video=0.4, middle_audio=0.2)
+    model = _make_adaln_schedule_model()
+    expected_projection_bytes = sum(
+        parameter.numel() * parameter.element_size()
+        for projection in model._adaln_projections
+        for parameter in projection.linear.parameters()
+    )
+
+    first = model.prepare_adaln_schedule_cache(
```
lishunyang12, line 285: Mock empty_cache in this CPU test; CPUPlatform doesn't implement it.

PR #5784, opened 2026-08-04
[BugFix][OmniVoice] Add ASR for missing reference text and match audio preprocessing

### LINKED ISSUE #5659: [Bug]: k2-fsa/OmniVoice single word prompt produces white noise intermittently
### Your current environment

<details>

```text
Collecting environment information...
INFO 07-31 15:17:49 [patch.py:252] NVFP4 W4A4 weight_scale NaN-clamp: installed.
INFO 07-31 15:17:50 [patch.py:481] [cumem-cuda] CuMemAllocator._python_free_callback patched: asleep guard extended to all platforms.
        System Info
OS                           : Ubuntu 22.04.5 LTS (x86_64)
GCC version                  : (Ubuntu 11.4.0-1ubuntu1~22.04.3) 11.4.0
Clang version                : Could not collect
CMake version                : version 3.22.1
Libc version                 : glibc-2.35

       PyTorch Info
PyTorch version              : 2.11.0+cu130
CUDA used to build PyTorch   : 13.0

      Python Environment
Python version               : 3.12.13 (main, Jun 23 2026, 15:18:55) [Clang 22.1.3 ] (64-bit runtime)
Python platform              : Linux-6.8.0-1052-aws-x86_64-with-glibc2.35

       CUDA / GPU Info
Is CUDA available            : True
CUDA runtime version         : 12.9.86
CUDA_MODULE_LOADING set to   : 
GPU models and configuration : 
GPU 0: NVIDIA RTX PRO 4500 Blackwell Server Edition
GPU 1: NVIDIA RTX PRO 4500 Blackwell Server Edition

Nvidia driver version        : 580.126.09
cuDNN version                : Could not collect
HIP runtime version          : N/A
MIOpen runtime version       : N/A

          CPU Info
Architecture:                            x86_64
CPU op-mode(s):                          32-bit, 64-bit
Address sizes:                           48 bits physical, 48 bits virtual
Byte Order:                              Little Endian
CPU(s):                                  48
On-line CPU(s) list:                     0-47
Vendor ID:                               GenuineIntel
Model name:                              Intel(R) Xeon(R) 6975P-C
CPU family:                              6
Model:                                   173
Thread(s) per core:                      1
Core(s) per socket:                      48
Socket(s):                               1
Stepping:                                1
BogoMIPS:                                5400.00
Flags:                                   fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ss ht syscall nx pdpe1gb rdtscp lm constant_tsc arch_perfmon rep_good nopl xtopology nonstop_tsc cpuid aperfmperf tsc_known_freq pni pclmulqdq monitor ssse3 fma cx16 pdcm pcid sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand hypervisor lahf_lm abm 3dnowprefetch cpuid_fault ssbd ibrs ibpb stibp ibrs_enhanced fsgsbase tsc_adjust bmi1 avx2 smep bmi2 erms invpcid avx512f avx512dq rdseed adx smap avx512ifma clflushopt clwb avx512cd sha_ni avx512bw avx512vl xsaveopt xsavec xgetbv1 xsaves avx_vnni avx512_bf16 wbnoinvd ida arat avx512vbmi umip pku ospke waitpkg avx512_vbmi2 gfni vaes vpclmulqdq avx512_vnni avx512_bitalg tme avx512_vpopcntdq rdpid bus_lock_detect cldemote movdiri movdir64b md_clear serialize amx_bf16 avx512_fp16 amx_tile amx_int8 flush_l1d arch_capabilities
Hypervisor vendor:                       KVM
Virtualization type:                     full
L1d cache:                               2.3 MiB (48 instances)
L1i cache:                               3 MiB (48 instances)
L2 cache:                                96 MiB (48 instances)
L3 cache:                                480 MiB (1 instance)
NUMA node(s):                            1
NUMA node0 CPU(s):                       0-47
Vulnerability Gather data sampling:      Not affected
Vulnerability Indirect target selection: Not affected
Vulnerability Itlb multihit:             Not affected
Vulnerability L1tf:                      Not affected
Vulnerability Mds:                       Not affected
Vulnerability Meltdown:                  Not affected
Vulnerability Mmio stale data:           Not affected
Vulnerability Reg file data sampling:    Not affected
Vulnerability Retbleed:                  Not affected
Vulnerability Spec rstack overflow:      Not affected
Vulnerability Spec store bypass:         Mitigation; Speculative Store Bypass disabled via prctl
Vulnerability Spectre v1:                Mitigation; usercopy/swapgs barriers and __user pointer sanitization
Vulnerability Spectre v2:                Mitigation; Enhanced / Automatic IBRS; IBPB conditional; PBRSB-eIBRS Not affected; BHI BHI_DIS_S
Vulnerability Srbds:                     Not affected
Vulnerability Tsa:                       Not affected
Vulnerability Tsx async abort:           Not affected
Vulnerability Vmscape:                   Not affected

[pip3] flashinfer-python==0.6.14
[pip3] numpy==2.3.5
[pip3] nvidia-cublas==13.1.0.3
[pip3] nvidia-cuda-cccl==13.3.3.4.1
[pip3] nvidia-cuda-crt==13.3.73
[pip3] nvidia-cuda-cupti==13.0.85
[pip3] nvidia-cuda-nvcc==13.3.73
[pip3] nvidia-cuda-nvdisasm==13.3.73
[pip3] nvidia-cuda-nvrtc==13.0.88
[pip3] nvidia-cuda-runtime==13.0.96
[pip3] nvidia-cudnn-cu13==9.19.0.56
[pip3] nvidia-cudnn-frontend==1.26.0
[pip3] nvidia-cufft==12.0.0.61
[pip3] nvidia-cufile==1.15.1.6
[pip3] nvidia-curand==10.4.0.35
[pip3] nvidia-cusolver==12.0.4.66
[pip3] nvidia-cusparse==12.6.3.3
[pip3] nvidia-cusparselt-cu13==0.8.0
[pip3] nvidia-cutlass-dsl==4.6.0
[pip3] nvidia-cutlass-dsl-libs-base==4.6.0
[pip3] nvidia-cutlass-dsl-libs-core==4.6.0
[pip3] nvidia-cutlass-dsl-libs-cu12==4.6.0
[pip3] nvidia-cutlass-dsl-libs-cu13==4.6.0
[pip3] nvidia-ml-py==13.610.43
[pip3] nvidia-nccl-cu13==2.28.9
[pip3] nvidia-nvjitlink==13.0.88
[pip3] nvidia-nvshmem-cu13==3.4.5
[pip3] nvidia-nvtx==13.0.85
[pip3] nvidia-nvvm==13.3.73
[pip3] onnxruntime==1.28.0
[pip3] pyzmq==27.1.0
[pip3] tokenspeed-triton==3.8.10.post20260721
[pip3] torch==2.11.0+cu130
[pip3] torch_c_dlpack_ext==0.1.5
[pip3] torch-einops-utils==0.1.12
[pip3] torchaudio==2.11.0+cu130
[pip3] torchcodec==0.15.0+cu130
[pip3] torchsde==0.2.6
[pip3] torchvision==0.26.0+cu130
[pip3] transformers==5.14.1
[pip3] triton==3.6.0
[pip3] x-transformers==2.24.1
[conda] Could not collect

         vLLM Info
vLLM-Omni Version            : 0.26.0rc2.dev56+g4956a41f1 (git sha: 4956a41f1)
vLLM Build Flags:
  CUDA Archs: Not Set; ROCm: Disabled
GPU Topology:
  	GPU0	GPU1	CPU Affinity	NUMA Affinity	GPU NUMA ID
GPU0	 X 	NODE	0-47	0		N/A
GPU1	NODE	 X 	0-47	0		N/A

Legend:

  X    = Self
  SYS  = Connection traversing PCIe as well as the SMP interconnect between NUMA nodes (e.g., QPI/UPI)
  NODE = Connection traversing PCIe as well as the interconnect between PCIe Host Bridges within a NUMA node
  PHB  = Connection traversing PCIe as well as a PCIe Host Bridge (typically the CPU)
  PXB  = Connection traversing multiple PCIe bridges (without traversing the PCIe Host Bridge)
  PIX  = Connection traversing at most a single PCIe bridge
  NV#  = Connection traversing a bonded set of # NVLinks

     Environment Variables
LD_LIBRARY_PATH=/home/ubuntu/vllm-omni/.venv/lib/python3.12/site-packages/cv2/../../lib64:/usr/local/cuda-12.9/lib:/usr/local/cuda-12.9/lib64:/usr/local/cuda-12.9:/usr/local/cuda-12.9/targets/x86_64-linux/lib/:/usr/local/cuda-12.9/extras/CUPTI/lib64:/opt/amazon/openmpi/lib:/usr/local/lib:/usr/lib
PYTORCH_NVML_BASED_CUDA_CHECK=1
TORCHINDUCTOR_COMPILE_THREADS=1
TORCHINDUCTOR_CACHE_DIR=/tmp/torchinductor_ubuntu

```

</details>

### Your code version

<details>

```text
0.26.0
```
</details>
<details>

```text
Version = 0.26.0rc2.dev56+g4956a41f1
Commit id = 4956a41f1291923845b69be4eba4f842932697eb
```
</details>

### 🐛 Describe the bug

While testing with French set as the language and using short inputs eg merci or bonjour the output would occassinally be white noise. Using a longer input this didnt happen. To verify that short words are the issue I tested the model with English. The same behaviour is seen, short words like yes, no, bye sometimes produce an empty output. I tested the model using multiple longer form inputs in English and never encountered an issue with white noise outputs.

akshatvishu:
### Purpose

Resolves #5659
OmniVoice voice cloning failed when a request included reference audio without reference text. The pipeline encoded the reference audio but it did not create the transcript that OmniVoice uses with those audio tokens. The model therefore received audio tokens without matching text and the generated audio could be silence or white noise or very low volume.

The official OmniVoice implementation transcribes missing reference text with Whisper. It also prepares the reference waveform before transcription and audio tokenization. The preparation raises quiet audio to the target level, removes silence, aligns the sample count to the tokenizer hop size and keeps the original volume for the final output. The official implementation also adds missing punctuation and applies silence removal, volume adjustment, fades and padding to generated audio.

This branch adds the missing ASR path and the shared audio processing code. It uses the same prepared waveform for transcription and voice tokenization. It also keeps blank reference text absent from offline and online requests so the pipeline can detect that it must transcribe the reference audio.

### Test plan

ROCM Version                 : 7.2.53211-c2d9476115
vLLM-Omni Version            : 0.1.dev2415+gb9a51592f.rocm (git sha: b9a51592f, date: ocm)

Run the CPU tests for reference audio preparation and automatic transcription. The tests use synthetic audio and mocked model components, so they do not need model weights or a GPU.

```bash
pytest -q \
    tests/model_executor/models/omnivoice/test_asr.py \
    tests/model_executor/models/omnivoice/test_audio_processing.py
```

```bash
pytest \
    --run-level advanced_model \
    -q tests/e2e/offline_inference/test_omnivoice_expansion.py \
    -k ref_audio_without_ref_text
```

```bash
pytest \
    --run-level advanced_model \
    -q tests/e2e/online_serving/test_omnivoice_expansion.py \
    -k ref_audio_only
```

```bash
pytest \
    --run-level advanced_model \
    -q tests/e2e/online_serving/test_omnivoice_expansion.py \
    -k "not ref_audio_only"
```
1 failed, 3 passed, 1 deselected, 15 warnings in 1876.42s (0:31:16)

updated fail test pass:
```bash
pytest \
    --run-level advanced_model \
    -q -s \
    'tests/e2e/online_serving/test_omnivoice_expansion.py::TestOmniVoiceSeed::test_speech_auto_voice_seed_determinist[omni_server0]'
```
1 passed, 15 warnings in 650.18s (0:10:50)
```bash
pytest \
    --run-level advanced_model \
    -q tests/e2e/offline_inference/test_omnivoice_expansion.py \
    -k "not ref_audio_without_ref_text" \
```
1 passed, 1 deselected, 16 warnings in 23.77s

### Test result
```text
The CPU tests passed with 41 passed tests and 15 warnings.

The offline regression log records 1 passed test, 1 deselected test, and 16 warnings in 275.04 seconds.

The online regression log records 1 passed test, 4 deselected tests, and 15 warnings in 497.61 seconds.
```
The `test_speech_auto_voice_seed_deterministic` test was updated because this branch added  `pydub` dependency  to prepare and write the audio. Repeated requests can produce the same audio samples while the complete WAV files differ slightly because of sample rounding and file encoding. The old byte by byte check therefore failed even though generation was deterministic. The test now checks the sample rate and length and compares the decoded audio samples with a small tolerance.

### Performance

script:<https://gist.github.com/akshatvishu/e6f7d14105652216668bdb8b0582f743>
The script runs  zero-shot voice cloning using `tests/assets/qwen3_tts/clone_2.wav` as the reference audio, with one warm-up run and three measured runs. The main uses : `--ref-text "Okay. Yeah. I resent you. I love you. I respect you. But you know what? You blew it! And thanks to you."`

Benchmarked current changes (without `--ref-text` as we use whisper) v/s main (in `main` we need to provide `-ref-text` otherwise voice cloning returns silent `.wav` outputs)
| Benchmark dimension     |          Supplied `ref_text` |     Missing `ref_text`, cold | Missing `ref_text`, warm and uncached |  Repeated same audio |
| ----------------------- | ---------------------------: | ---------------------------: | ------------------------------------: | -------------------: |
| ASR behavior            |                  ASR skipped | Model load and transcription |   Transcription with the loaded model | Transcript cache hit |
| End to end latency      | 0.475 s mean over 3 requests |                      4.559 s |   0.671 s mean over 3 different clips |              0.474 s |
| Individual warm times   |               Not applicable |               Not applicable |             0.641, 0.652, and 0.722 s |       Not applicable |
| ASR GPU allocation      |                         None |                   +1.507 GiB |                      Already resident |     Already resident |
| ASR parameter placement |                   Not loaded |                     `cuda:0` |                              `cuda:0` |             `cuda:0` |

The warm benchmark used three different reference clips, so every measured request ran Whisper and created a new cache entry. No measured request used the transcript cache. The cold request was excluded from the 0.671s mean.

### DIFF
--- a/requirements/common.txt
+++ b/requirements/common.txt
@@ -16,9 +16,6 @@ diffusers==0.38.0
 safetensors>=0.8.0
 accelerate==1.12.0
 soundfile>=0.13.1
-pydub>=0.25.1
-# Python 3.13 removed audioop, which pydub uses for audio processing.
-audioop-lts; python_version >= "3.13"
 cache-dit==1.3.0
 tqdm>=4.66.0
 torchsde>=0.2.6
--- a/requirements/common.txt
+++ b/requirements/common.txt
@@ -16,6 +16,9 @@ diffusers==0.38.0
 safetensors>=0.8.0
 accelerate==1.12.0
 soundfile>=0.13.1
+pydub>=0.25.1
+# Python 3.13 removed audioop, which pydub uses for audio processing.
+audioop-lts; python_version >= "3.13"
 cache-dit==1.3.0
 tqdm>=4.66.0
 torchsde>=0.2.6
--- a/vllm_omni/deploy/omnivoice.yaml
+++ b/vllm_omni/deploy/omnivoice.yaml
@@ -11,3 +11,7 @@ stages:
     trust_remote_code: true
     distributed_executor_backend: "mp"
     dtype: "float32"
+    additional_config:
+      omnivoice_asr:
+        load_asr: false
+        asr_model_name: "openai/whisper-large-v3-turbo"
--- a/vllm_omni/deploy/omnivoice.yaml
+++ b/vllm_omni/deploy/omnivoice.yaml
@@ -13,6 +13,8 @@ stages:
     dtype: "float32"
     additional_config:
       omnivoice_asr:
-        load_asr: false
+        # false defers Whisper until ref_audio is sent without ref_text;
+        # it does not disable automatic transcription. Set true to load at startup.
+        load_asr_on_startup: false
         asr_model_name: "openai/whisper-large-v3-turbo"
         asr_device: null
--- a/vllm_omni/deploy/omnivoice.yaml
+++ b/vllm_omni/deploy/omnivoice.yaml
@@ -15,3 +15,4 @@ stages:
       omnivoice_asr:
         load_asr: false
         asr_model_name: "openai/whisper-large-v3-turbo"
+        asr_device: null
--- a/vllm_omni/diffusion/models/omnivoice/audio.py
+++ b/vllm_omni/diffusion/models/omnivoice/audio.py
@@ -1,0 +1,264 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+
+"""Audio preparation and output processing for OmniVoice."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+import numpy as np
+import torch
+import torchaudio
+from pydub import AudioSegment
+from pydub.silence import detect_leading_silence, detect_nonsilent, split_on_silence
+
+_END_PUNCTUATION = {
+    ";",
+    ":",
+    ",",
+    ".",
+    "!",
+    "?",
+    "…",
+    ")",
+    "]",
+    "}",
+    '"',
+    "'",
+    "“",
+    "”",
+    "‘",
+    "’",
+    "；",
+    "：",
+    "，",
+    "。",
+    "！",
+    "？",
+    "、",
+    "……",
+    "）",
+    "】",
+}
+
+
+@dataclass(frozen=True)
+class PreparedReferenceAudio:
+    """Reference waveform shared by ASR and the audio tokenizer."""
+
+    waveform: np.ndarray
+    sample_rate: int
+    original_rms: float
+
+
+def _numpy_to_audio_segment(audio: np.ndarray, sample_rate: int) -> AudioSegment:
+    """Convert a channel-first float waveform to an in-memory PCM segment."""
+    audio = np.asarray(audio, dtype=np.float32)
+    if audio.ndim == 1:
+        audio = audio[np.newaxis, :]
+    audio_int = np.clip(audio * 32768.0, -32768, 32767).astype(np.int16)
+    if audio_int.shape[0] > 1:
+        audio_int = audio_int.T.reshape(-1)
+    else:
+        audio_int = audio_int.reshape(-1)
+    return AudioSegment(
+        data=audio_int.tobytes(),
+        sample_width=2,
+        frame_rate=sample_rate,
+        channels=audio.shape[0],
+    )
+
+
+def _audio_segment_to_numpy(audio: AudioSegment) -> np.ndarray:
+    """Convert an in-memory PCM segment to a channel-first float waveform."""
+    data = np.asarray(audio.get_array_of_samples(), dtype=np.float32) / 32768.0
+    if audio.channels == 1:
+        return data[np.newaxis, :]
+    return data.reshape(-1, audio.channels).T
+
+
+def remove_silence(
+    audio: np.ndarray,
+    sample_rate: int,
+    *,
+    middle_silence_ms: int,
+    leading_silence_ms: int,
+    trailing_silence_ms: int,
+) -> np.ndarray:
+    """Remove long gaps and trim edge silence using official thresholds."""
+    wave = _numpy_to_audio_segment(audio, sample_rate)
+    if middle_silence_ms > 0:
+        non_silent_segments = split_on_silence(
+            wave,
+            min_silence_len=middle_silence_ms,
+            silence_thresh=-50,
+            keep_silence=middle_silence_ms,
+            seek_step=10,
+        )
+        wave = AudioSegment.silent(duration=0)
+        for segment in non_silent_segments:
+            wave += segment
+
+    start_idx = detect_leading_silence(wave, silence_threshold=-50)
+    start_idx = max(0, start_idx - leading_silence_ms)
+    wave = wave[start_idx:]
+
+    wave = wave.reverse()
+    start_idx = detect_leading_silence(wave, silence_threshold=-50)
+    start_idx = max(0, start_idx - trailing_silence_ms)
+    wave = wave[start_idx:].reverse()
+
+    return _audio_segment_to_numpy(wave)
+
+
+def trim_long_audio(
+    audio: np.ndarray,
+    sample_rate: int,
+    *,
+    max_duration_s: float = 15.0,
+    min_duration_s: float = 3.0,
+    trim_threshold_s: float = 20.0,
+) -> np.ndarray:
+    """Trim long reference audio at a suitable silence boundary."""
+    if audio.shape[-1] / sample_rate <= trim_threshold_s:
+        return audio
+
+    segment = _numpy_to_audio_segment(audio, sample_rate)
+    non_silent_ranges = detect_nonsilent(
+        segment,
+        min_silence_len=100,
+        silence_thresh=-40,
+        seek_step=10,
+    )
+    if not non_silent_ranges:
+        return audio
+
+    max_ms = int(max_duration_s * 1000)
+    min_ms = int(min_duration_s * 1000)
+    best_split = 0
+    for start, end in non_silent_ranges:
+        if start > best_split and start <= max_ms:
+            best_split = start
+        if end > max_ms:
+            break
+
+    if best_split < min_ms:
+        best_split = min(max_ms, len(segment))
+    return _audio_segment_to_numpy(segment[:best_split])
+
+
+def prepare_reference_audio(
+    waveform: np.ndarray | torch.Tensor,
+    sample_rate: int,
+    *,
+    target_sample_rate: int,
+    hop_length: int,
+    trim_long: bool,
+) -> PreparedReferenceAudio:
+    """Prepare reference audio before ASR and audio-tokenizer encoding."""
+    if isinstance(waveform, torch.Tensor):
+        waveform = waveform.detach().cpu().numpy()
+    waveform = np.array(waveform, dtype=np.float32, copy=True)
+    if waveform.ndim == 1:
+        waveform = waveform[np.newaxis, :]
+    elif waveform.ndim != 2:
+        raise ValueError(f"OmniVoice reference audio must be 1D or 2D, got {waveform.ndim} dimensions.")
+    if not waveform.size or not np.any(waveform):
+        raise ValueError("Reference audio is empty after silence removal.")
+    if waveform.shape[0] > 1:
+        waveform = waveform.mean(axis=0, keepdims=True)
+
+    if sample_rate != target_sample_rate:
+        waveform = torchaudio.functional.resample(
+            torch.from_numpy(waveform),
+            orig_freq=sample_rate,
+            new_freq=target_sample_rate,
+        ).numpy()
+
+    original_rms = float(np.sqrt(np.mean(waveform**2))) if waveform.size else 0.0
+    if 0 < original_rms < 0.1:
+        waveform = waveform * (0.1 / original_rms)
+
+    if trim_long:
+        waveform = trim_long_audio(waveform, target_sample_rate)
+    waveform = remove_silence(
+        waveform,
+        target_sample_rate,
+        middle_silence_ms=200,
+        leading_silence_ms=100,
+        trailing_silence_ms=200,
+    )
+    if waveform.shape[-1] == 0:
+        raise ValueError("Reference audio is empty after silence removal.")
+
+    remainder = waveform.shape[-1] % hop_length
+    if remainder:
+        waveform = waveform[:, :-remainder]
+    if waveform.shape[-1] == 0:
+        raise ValueError("Reference audio is shorter than one audio-tokenizer hop after preprocessing.")
+
+    return PreparedReferenceAudio(
+        waveform=np.ascontiguousarray(waveform, dtype=np.float32),
+        sample_rate=target_sample_rate,
+        original_rms=original_rms,
+    )
+
+
+def add_reference_punctuation(text: str) -> str:
+    """Add an English or Chinese sentence terminator when one is missing."""
+    text = text.strip()
+    if not text:
+        return text
+    if text[-1] in _END_PUNCTUATION:
+        return text
+    if any("\u4e00" <= char <= "\u9fff" for char in text):
+        return f"{text}。"
+    return f"{text}."
+
+
+def postprocess_generated_audio(
+    audio: np.ndarray,
+    *,
+    sample_rate: int,
+    reference_rms: float | None,
+) -> np.ndarray:
+    """Apply official OmniVoice output silence, volume, fade, and padding rules."""
+    audio = np.array(audio, dtype=np.float32, copy=True)
+    if audio.ndim == 1:
+        audio = audio[np.newaxis, :]
+    if audio.ndim != 2:
+        raise ValueError(f"OmniVoice generated audio must be 1D or 2D, got {audio.ndim} dimensions.")
+
+    audio = remove_silence(
+        audio,
+        sample_rate,
+        middle_silence_ms=500,
+        leading_silence_ms=100,
+        trailing_silence_ms=100,
+    )
+
+    if audio.shape[-1] == 0:
+        return np.ascontiguousarray(audio, dtype=np.float32)
+
+    if reference_rms is not None and reference_rms < 0.1:
+        audio = audio * (reference_rms / 0.1)
+    elif reference_rms is None:
+        peak = float(np.abs(audio).max())
+        if peak > 1e-6:
+            audio = audio / peak * 0.5
+
+    fade_samples = int(0.1 * sample_rate)
+    if fade_samples > 0:
+        fade_length = min(fade_samples, audio.shape[-1] // 2)
+        if fade_length > 0:
+            fade_in = np.linspace(0, 1, fade_length, dtype=np.float32)[np.newaxis, :]
+            fade_out = np.linspace(1, 0, fade_length, dtype=np.float32)[np.newaxis, :]
+            audio[:, :fade_length] *= fade_in
+            audio[:, -fade_length:] *= fade_out
+
+    pad_samples = int(0.1 * sample_rate)
+    if pad_samples > 0:
+        padding = np.zeros((audio.shape[0], pad_samples), dtype=np.float32)
+        audio = np.concatenate([padding, audio, padding], axis=-1)
+    return np.ascontiguousarray(audio, dtype=np.float32)
--- a/vllm_omni/diffusion/models/omnivoice/audio.py
+++ b/vllm_omni/diffusion/models/omnivoice/audio.py
@@ -10,8 +10,6 @@ from dataclasses import dataclass
 import numpy as np
 import torch
 import torchaudio
-from pydub import AudioSegment
-from pydub.silence import detect_leading_silence, detect_nonsilent, split_on_silence
 
 _END_PUNCTUATION = {
     ";",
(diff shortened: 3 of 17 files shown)


vllm_omni/diffusion/models/omnivoice/pipeline_omnivoice.py
```diff
@@ -38,10 +44,15 @@
...
 
-import torchaudio
+try:
+    from transformers import pipeline as hf_pipeline
+except ImportError:
+    hf_pipeline = None
 
 logger = init_logger(__name__)
 
+_ASR_MODEL_NAME = "openai/whisper-large-v3-turbo"
```
linyueqian, line 56: [High] Make ASR opt-in and configurable rather than always-on with a hardcoded model. Upstream landed the same knobs after the initial release: k2-fsa/OmniVoice#221 (`asr_model_name`) and k2-fsa/OmniVoice#224 (`asr_device`). As written, every OmniVoice deployment silently grows a second model on the same device as the diffusion worker plus an outbound HF download that only happens when the first ref-audio-without-ref-text request arrives, so an offline or air-gapped deployment fails at request time instead of at startup, and the engine's memory-utilization budget no longer reflects what is actually resident.

Suggest a deploy-yaml block (enable flag, model name, device) defaulting to off, and resolving it at init so a misconfiguration surfaces during startup.

vllm_omni/diffusion/models/omnivoice/pipeline_omnivoice.py
```diff
@@ -310,20 +403,53 @@ def forward(self, req: DiffusionRequestBatch) -> DiffusionOutput:
                 cached = self._speaker_cache.get(_cache_key)
                 if cached is not None:
                     ref_audio_tokens = cached["ref_audio_tokens"].to(device)
+                    reference_rms = cached["reference_rms"]
+                    if needs_asr and cached.get("ref_text"):
+                        ref_text = cached["ref_text"]
                     _cache_key = None  # hit → don't store again
                     logger.debug("Speaker cache HIT for OmniVoice speaker '%s'", voice_name)
 
+            if needs_asr and not ref_text:
```
linyueqian: [High] The transcript is only reused when `voice_name` is set, since `_speaker_cache` is not consulted otherwise. The repro in #5659 posts inline base64 `ref_audio` with no uploaded voice, so it takes this branch and re-transcribes on every single request, which is exactly the throughput cliff in your perf table.

Two options that avoid it: transcribe once at voice-upload time (`/v1/audio/voices` already carries a `ref_text` field, so the auto-filled transcript can be stored alongside the speaker entry), and/or key a small transcript cache on a hash of the prepared waveform so the inline path also gets reuse.

requirements/common.txt
```diff
@@ -16,6 +16,9 @@ diffusers==0.38.0
 safetensors>=0.8.0
 accelerate==1.12.0
 soundfile>=0.13.1
+pydub>=0.25.1
```
linyueqian: [Med] This adds a repo-wide dependency (plus `audioop-lts` on 3.13) for one model's audio postprocessing. Can the silence trim, RMS match, fade and padding be done with numpy and torchaudio, which are already hard deps? Worth noting `openai-whisper` is already in this file, so ASR itself is not the new-dependency concern here, pydub is.

vllm_omni/diffusion/models/omnivoice/pipeline_omnivoice.py
```diff
@@ -371,6 +497,13 @@ def forward(self, req: DiffusionRequestBatch) -> DiffusionOutput:
         )
         # Decode tokens to audio
         audio = self.decoder(tokens)  # [1, 1, samples]
+        audio_np = audio.squeeze(0).detach().cpu().float().numpy()
+        audio_np = postprocess_generated_audio(
```
linyueqian, line 605: [Med] This runs unconditionally, so it also changes output on the auto-voice path where there is no reference audio at all. That is a behavior change for users who are not doing voice cloning, and it is why `test_speech_auto_voice_seed_deterministic` had to be relaxed from a byte comparison to a tolerance comparison.

Preference would be to scope the postprocess to the cloning path so the existing auto-voice output stays bit-identical and the determinism test can stay strict. If it genuinely needs to apply everywhere to match upstream, please call that out explicitly in the PR description as an intentional output change.

vllm_omni/diffusion/models/omnivoice/pipeline_omnivoice.py
```diff
@@ -185,6 +197,86 @@ def __init__(self, *, od_config: OmniDiffusionConfig, prefix: str = ""):
...
+        return transcript
+
+    def _resolve_ref_text(self, ref_audio, ref_text: str | None) -> str | None:
+        """Resolve missing reference text only when reference audio is present."""
+        if ref_audio is not None and (ref_text is None or not ref_text.strip()):
+            logger.info("Automatically transcribing OmniVoice reference audio")
+            return self._transcribe_ref_audio(ref_audio)
+        return ref_text
+
+    def _estimate_target_len(
```
linyueqian, line 306: [Nit] Good catch. The old `estimate_duration(text, "Nice to meet you.", 25)` meant the cloning path never used the real reference length or transcript, so target duration was wrong for every voice-clone request regardless of whether `ref_text` was supplied. This part is a correctness fix and stands on its own, so if the ASR discussion above drags on it would be reasonable to split this hunk into its own PR and land it first.

linyueqian (commented):
Thanks for chasing this down. The duration hardcode is a real bug and I agree the cloning path was broken independently of ASR, so the `_estimate_target_len` part is worth landing on its own.

On the Whisper fallback, some context on why it was not in the original port: in vllm-omni every cloning TTS takes `ref_text` as an explicit request field (CosyVoice3, Fish Speech, GLM-TTS, higgs v2, IndexTTS2), and upstream's ASR fallback lives in the `OmniVoice.from_pretrained()` convenience wrapper rather than in the model. The port implemented the engine and left the transcript to the caller. That is a scope call, not a correctness claim, so I am not against adding it, but it needs to be opt-in and it needs to not sit on the request path unconditionally. Details inline.

[Med] On the perf table: the 16.7s is worth a second look before we accept 33x as the cost. Steady-state `whisper-large-v3-turbo` on a few-second clip should be a few hundred ms, and `self._asr_pipeline` is cached on the instance, so either the warm-up run is not actually warming the ASR pipeline or something is falling back badly on ROCm. Could you re-measure on CUDA and report warm-only numbers? The `0.0 MB` peak VRAM delta also looks off given a second model is loaded on `self.device`.

vllm_omni/diffusion/models/omnivoice/pipeline_omnivoice.py
akshatvishu: Added a bounded cache with eight entries for inline reference audio. The key includes the prepared waveform, its original RMS and whether the request used automatic or supplied text. Each entry stores the transcript and audio tokens, so repeated requests skip both Whisper and audio tokenization.

requirements/common.txt
akshatvishu: Removed both proposed dependencies. The port now uses NumPy and torchaudio  as suggested and follows the same silence window, range merge, edge trim, RMS, fade and padding rules as upstream.

vllm_omni/diffusion/models/omnivoice/pipeline_omnivoice.py
akshatvishu: I changed the logic so the new processing runs only for voice-cloning requests and not for text-only generation. Text-only generation now returns the decoder tensor before that processing runs and I added a CPU test that fails if the text-only path reaches the voice-cloning logic.

I also restored the exact WAV byte assertion as requested. On the available ROCm MI300X GPU, that assertion is flaky and fails on both the unchanged `main` branch (the original test) and this branch.

vllm_omni/diffusion/models/omnivoice/pipeline_omnivoice.py
akshatvishu: Added and validated `load_asr`, `asr_model_name` and `asr_device`. The default `load_asr: false` skips startup loading, while a request with reference audio and no `ref_text` still loads Whisper when needed. I kept this behavior because upstream uses `load_asr` to choose eager or lazy loading and  not to disable automatic transcription but in-case we wanna move away from this then I can remove it!

akshatvishu:
Thanks for the quick review @linyueqian. You were rightl; he original numbers came from an older branch where Whisper was still on CPU and the 0.0 MB metric did not measure ASR memory.

After fixing device placement, I repeated the test on the available AMD mi300x single GPU. PyTorch reports the device as `cuda:0`, and the log confirms that all Whisper parameters are on that device.

I removed the old 33.7x latency and 97% throughput claims. I also removed the generator latency claim because generator-only time was not measured separately. The output duration and RMS rows were removed because the reference clips and conditioning text differed and RMS alone does not measure output quality.

The observed difference between the separate explicit text and warm automatic transcription benchmarks is about 0.197s  but I would not present that number as isolated Whisper latency because the two benchmarks used different reference clips.

vllm_omni/deploy/omnivoice.yaml
```diff
@@ -11,3 +11,8 @@ stages:
     trust_remote_code: true
     distributed_executor_backend: "mp"
     dtype: "float32"
+    additional_config:
+      omnivoice_asr:
+        load_asr: false
```
linyueqian: [Nit] The name reads like a master switch but the semantics are startup-eager versus lazy: with `load_asr: false` a request carrying `ref_audio` and no `ref_text` still pulls `openai/whisper-large-v3-turbo` onto the worker device on first use, so there is no configuration that declines ASR outright.

I think the default behavior is defensible, since the only requests it fires on are the ones that produce broken audio today, so I am not asking to change it. But `load_asr` is easy to read as "ASR off" by an operator sizing GPU memory, and they would then be surprised by a lazy load. Either rename to something like `preload_asr` / `load_asr_on_startup`, or keep the name and add a third state where ASR is genuinely disabled and the request fails with a clear message telling the caller to supply `ref_text`. The docs you added describe the lazy behavior correctly, so this is only about the key name.

linyueqian (commented):
Thanks, this addressed everything and the perf correction was the right call. For the record on the numbers: withdrawing the 33.7x latency and -97% throughput rows rather than defending them, and dropping the output-duration and RMS comparisons because the two runs used different reference clips, is exactly the right handling. A ~0.197s warm delta that you decline to present as isolated ASR latency is more useful than a confident number that is not comparable.

The other four are resolved as far as I can tell from reading: pydub and audioop-lts are out of `requirements/common.txt` with the silence detection reimplemented on numpy and torchaudio, the postprocess is scoped to the cloning path so text-only output stays as it was, and the inline reference cache is content-addressed on a SHA-256 of the prepared waveform with an 8-entry LRU, which closes the re-transcribe-every-request path from #5659. One follow-up on the config block inline.

vllm_omni/deploy/omnivoice.yaml
akshatvishu: I agree, changed the variable name from load_asr to load_asr_on_startup + add a brief comment to describe the behavior!

akshatvishu:
Also @linyueqian ,  the restored raw byte equality test fails on ROCm, with the first mismatch at byte 4872. It fails the same way on both our branch and `main`, so this looks like an existing ROCm baseline failure rather than a regression from this PR. Thus, can you check if it's passing for you on NVIDIA device?  Because if it's passing for you and just failing for only ROCm /AMD devices then it demands a separate investigation on it's own!

log from when I ran the byte equality test on main :[omnivoice_main_seed_byte_parity.log](https://github.com/user-attachments/files/30787069/omnivoice_main_seed_byte_parity.log)

linyueqian (commented):
Re-reviewed the delta `a781e904..bc0b0941`. Every prior finding is resolved, I found no new code issue in the delta, and there is nothing inline this round. Holding at comment rather than approve for one reason that has nothing to do with the code: the PR is currently `mergeable=false`, `state=dirty`, so resolving those conflicts will produce a head that nobody has reviewed. Worth re-running once it merges cleanly.

The delta itself is the rename of the ambiguous ASR eager-loading switch to `load_asr_on_startup`, applied consistently across the implementation, the deploy YAML, tests, examples and documentation, and the deploy YAML now states explicitly that `false` defers ASR rather than disabling automatic transcription. That was the follow-up left open last round and it is done.

For the record, the seven items from the earlier rounds and how each stands now. ASR configuration is resolved under the lazy and eager design accepted previously: `_parse_asr_config` validates the startup flag, model name and device during initialization, and eager loading happens during worker construction when requested. The inline reference path reuses transcripts and encoded audio through a content-addressed SHA-256 cache keyed on waveform metadata and bounded to eight LRU entries. The dependency concern is gone: neither `pydub` nor `audioop-lts` appears in the dependency files, and audio processing now goes through numpy, torch and torchaudio. Voice-clone postprocessing is correctly scoped, with text-only generation returning the decoder tensor before postprocessing while cloned output receives silence processing, RMS adjustment, fades and padding, and the auto-voice deterministic test keeps strict byte equality. Target duration estimation uses the resolved reference transcript and the actual reference-token count for cloning while keeping the previous fallback inputs for text-only generation. The unsupported performance comparison is removed, with none of the previously questioned figures still present.

Validation, including what was not checked. Both reviewed heads share merge base `b9a51592`, `bc0b0941` directly follows `a781e904`, and each footprint was compared against that base rather than via a two-dot diff. The intervening commit and the relevant current-head code and tests were read statically, stale configuration keys and removed dependencies were searched for, and `git diff --check` is clean on both the full patch and the delta. No code from the head was executed, imported or run. Not verified: anything needing the network, GPU or model tests, ASR behaviour reproduction, and the author's NVIDIA and ROCm byte-parity claim. The CI visible on this head covers DCO, the 3.11 and 3.12 builds, docs and pre-commit, but no model or GPU lane has reported.

vllm_omni/diffusion/models/omnivoice/audio.py
```diff
@@ -0,0 +1,405 @@
...
+    if sample_rate != target_sample_rate:
+        waveform = torchaudio.functional.resample(
+            torch.from_numpy(waveform),
+            orig_freq=sample_rate,
+            new_freq=target_sample_rate,
+        ).numpy()
+
+    original_rms = float(np.sqrt(np.mean(waveform**2))) if waveform.size else 0.0
+    if 0 < original_rms < 0.1:
+        waveform = waveform * (0.1 / original_rms)
```
Gaohan123, line 326: [P2] RMS gain can exceed normalized waveform range

A quiet clip with an isolated peak can be amplified far beyond [-1, 1] here; a reproduced RMS=0.0119 input produced a peak of 8.4018 before both Whisper and HiggsAudio tokenization. Upstream's int16 AudioSegment conversion clips at this boundary. Clamp after the gain and add a peaky-quiet-waveform regression test.

akshatvishu: fixed in c4766e422 !

linyueqian (approved):
Re-reviewed the delta at c4766e42. Since the round 3 approval at 768a8f9e the only substantive change is the reference-audio clamp, and it is correct.

The gain-up branch scales a quiet reference by `0.1 / original_rms` to reach the target RMS, which is safe for a waveform whose energy is spread out but not for a peaky one. Your new test is the exact worst case: a single unit sample in 7056 zeros has an RMS of about 0.0119, so the gain is roughly 8.4 and the peak lands near 8.4 before this change, well outside the range the downstream stages assume. Clamping after the multiply bounds that at 1.0. Worth noting explicitly that clipping necessarily gives up the RMS target for such inputs, since you cannot both normalize energy and stay in range here; bounding the peak is the right trade for reference conditioning, and the alternative of peak-normalizing instead would change the loudness of every ordinary reference rather than just the pathological ones.

The remaining changes in the delta are the two SPDX headers moved to the vLLM-Omni project and one annotation on the fake ASR's `inputs` list.

Approving again at this head. DCO, both wheel builds, docs, and pre-commit are green, and GitHub reports the branch mergeable.

Validation: static review only. I did not execute the fork head or run the audio tests, so the clamp's numerical behaviour rests on your added test rather than my own run.

linyueqian:
This is approved and the only thing between it and merging is a rebase.

The branch went conflicting against main, and the conflict took CI with it: all four lanes are now red at `c4766e42`, general build 14652, AMD 11405, Intel 7807, and NPU 6402. That is not four separate problems, it is one merge state that no lane can build from, so there is nothing here worth triaging until the branch merges cleanly again.

Please rebase onto current main. That will clear the conflict and give all four lanes a tree they can actually build, and I will re-check the results at the new head. If anything real surfaces once they run properly I will raise it then, but I am not expecting it: the approval covered the ASR fallback and the audio preprocessing alignment, and nothing in those areas is implicated by a merge conflict.

Worth knowing while you are in there: the `ready` label is already on this PR, and pushing to a branch that already carries it does not re-fire per-PR CI on its own. The rebase itself will produce a fresh run, but if you ever need to re-run without a new commit, the label has to be cycled rather than simply left in place.

linyueqian:
Correction to my last note: the merge conflict does not explain all four CI failures. General 14652 and AMD 11405 both ran tests and caught `final_output.request_output` in the new regression test at `tests/e2e/offline_inference/test_omnivoice_expansion.py:105`. That accessor no longer exists, so the test cannot reach its audio assertions. Please update it and the nested-output fallback when rebasing.

Intel 7807 and the NPU Platforms job fail separately on the old `create_error_response` import. AMD also has two diffusion attention tolerance failures. A rebase is still needed, but we need to check the test fix and fresh CI before merging.

akshatvishu:
@linyueqian Thanks for the correction. I’ve rebased and pushed the accessor fix, including removing the redundant nested-output fallback. Local targeted tests pass! Can you please re-run the CI test again (The Intel/NPU import  errors and AMD attention tolerance failures remain separate concerns at previous CI run)

PR #5788, opened 2026-08-05
Fix local media paths for diffusion speech

jsntcheng:
Fix local media access for pure diffusion speech serving.

## Purpose

Pure diffusion speech models such as `k2-fsa/OmniVoice` do not have a regular `model_config` available during API server initialization. As a result, the diffusion speech path created `MediaConnector()` without propagating `--allowed-local-media-path`.

Although the flag was accepted by the CLI, local `file://` reference audio failed with:

```text
RuntimeError: Cannot load local files without `--allowed-local-media-path`.
```

This change:

- Passes `allowed_local_media_path` from `api_server.py` to `OmniOpenAIServingSpeech.for_diffusion`.
- Passes `allowed_media_domains` through the same path.
- Configures `MediaConnector` with both values for pure diffusion speech requests.
- Preserves the existing behavior for non-diffusion speech models.

The configured local media directory remains enforced by vLLM's `MediaConnector`, so files outside the allowlisted directory are still rejected.

## Test Plan

**vLLM Version:** `0.26.0`

**vLLM-Omni Commit:** `390c1a54`

- Ran `python3 -m compileall` on the modified modules.
- Ran `git diff --check`.
- Installed the repository in editable mode with:

```bash
python -m pip install --no-deps -e ~/vllm-omni
```

- Started OmniVoice with:

```bash
CUDA_VISIBLE_DEVICES=2 \
vllm serve k2-fsa/OmniVoice \
  --omni \
  --port 8108 \
  --trust-remote-code \
  --allowed-local-media-path /home/lcc/ref_audio
```

## Test Result

- OmniVoice model loaded successfully on physical GPU 2.
- API server started successfully on port 8108.
- Health check returned HTTP 200.
- `/v1/audio/speech` accepted local reference audio from the allowlisted directory.
- Speech generation returned HTTP 200 and generated 117,120 audio frames.
- The previous `Cannot load local files without --allowed-local-media-path` error was resolved.


vllm_omni/entrypoints/openai/api_server.py
```diff
@@ -793,6 +793,8 @@ async def omni_init_app_state(
             diffusion_engine=engine_client,
             model_name=model_name,
             stage_configs=diffusion_stage_configs,
+            allowed_local_media_path=getattr(args, "allowed_local_media_path", ""),
```
NickCao: We could instead pass MediaConnector directly here? So if it ever accepts more parameters, we don't have to thread them all though layer and layers.

PR #5506, opened 2026-07-28
[CI][MiniCPM-o] Run e2e tests on single-GPU deploy config

amy-why-3459:
Summary
Switch MiniCPM-o 4.5 offline/online e2e suites from minicpmo_4_5_batching.yaml (2-GPU) to minicpmo_4_5.yaml (single large-memory GPU layout).
Update @hardware_test from npu: A2, num_cards=2 to npu: A3, num_cards=1, matching the single-GPU deploy memory profile after #5447.
Align CI scheduling with the default single-GPU recipe so e2e coverage matches the documented deployment path.

PR #5790, opened 2026-08-05
[CI] MiniCPM-o4.5 CI concurrency=8 OOM bugfix

R2-Y:

## Purpose
Temporarily work around the CI concurrency=8 OOM issue.
https://buildkite.com/vllm/vllm-omni/builds/13167/list?sid=019fce2a-1512-492e-bc32-e6adbb09bc63&tab=output
& update documents

Seed-TTS concurrency 8 OOMs in Stage-2 Code2Wav. With max_num_seqs: 16 (#5506), exact-shape batching formed a 5-request HiFi-GAN/flow batch that tried to allocate 2.76 GiB with only 1.96 GiB free (Stage0/1 already held 43.67/11.99 GiB). Concurrency 1/4 passed; only the 8×128 sweep failed (51/128 completed). This PR cap Stage-2 (and colocated Stage-0/1) batch size so Code2Wav peak activations fit the single-H100 residual.

Call chain: perf client (c=8) → Stage0/1 → Code2Wav exact-shape batch → HiFi-GAN alloc; max_num_seqs=4 prevents bs≥5.

## Test Plan

**vLLM Version:**
0.25.0

**vLLM-Omni Commit:**
0.25.0

## Test Result


amy-why-3459:
Capping max_num_seqs to 4 works as a temporary fix, but this should be a short-term workaround. The underlying issue is that exact-shape batching in Stage 2 can form batches that exceed residual GPU memory even at concurrency 8. Reducing max_num_seqs limits throughput globally — we should find the root cause (e.g. unbounded Code2Wav activation peaks, missing memory budget enforcement at batch formation) and fix it properly.

amy-why-3459 (approved):

PR #4958, opened 2026-07-08
[Perf][Qwen3-TTS] Fuse QKV and gate_up projections in code predictor

l-wave:
**Before submitting:** run the precheck-pr skill with code agent for a self-check against project conventions — catches dead code, missing benchmarks, and title format issues.

## Purpose

Fuse the per-layer `q_proj` / `k_proj` / `v_proj` and `gate_proj` / `up_proj` projections in the Qwen3-TTS / Qwen3-Omni code predictor into a single `qkv_proj` and `gate_up_proj` respectively. The packing preserves the same linear algebra as the original implementation: concatenate the individual HF-style rows along `dim=0`, run one plain `nn.Linear`, then split the result on `dim=-1`. GPU executions may still show tiny bit-level drift when kernel selection or accumulation order changes, but the mathematical operation is unchanged.

The intended upside is reduced eager host-side `cudaLaunchKernel` count from 5 to 2 per code-predictor layer. This targets the host-side bottleneck visible in `talker_mtp` NVTX traces under high concurrency: the code predictor's 5-layer x 17-AR-step loop launches several thousand small kernels per second, and the q/k/v + gate/up projections are the dominant offenders because they are sibling `nn.Linear` calls executed back-to-back with no intermediate op to amortize launch overhead.

The HF-numerics-compatible constraint introduced by #2277 / #2291 is preserved. This change does **not** reintroduce vLLM's `QKVParallelLinear` or fused RMSNorm kernels; it only reorganizes plain `nn.Linear` calls and re-densifies packed k/v slices before SDPA. There are no new env vars and no intentional behavioral change beyond possible GPU bit-level drift.

Closes nothing; no existing issue tracks this exact change. See PR #3662 for prior high-concurrency optimization work on the same module.

### Why packed == separate at the linear-algebra level

For three independent `nn.Linear(H, D_i, bias=False)` modules with weights `W_q in R^{D_q x H}`, `W_k in R^{D_k x H}`, and `W_v in R^{D_v x H}`:

```text
[q | k | v] = [x * W_q^T | x * W_k^T | x * W_v^T]
            = x * [W_q^T | W_k^T | W_v^T]
            = x * cat([W_q, W_k, W_v], dim=0)^T
```

So `qkv_proj.weight = torch.cat([W_q, W_k, W_v], dim=0)` produces the same mathematical result as three separate linears, split on `dim=-1`. The same algebra applies to `gate_up_proj`.

## Review follow-up fixes

This revision also addresses the reviewer follow-up items:

[...]

PR #5791, opened 2026-08-05
[Perf][Qwen3-TTS] Fuse QKV and gate_up projections in code predictor

### LINKED ISSUE #5776: [Bug]: L3 test failed about entrypoint tests due to PR 4958
### Your current environment

<details>

```text
CI Env
```

</details>

### Your code version

<details>

```text
0.26.0
```
</details>
<details>

```text
6fdc4ca62dc6080bc683552c6ec9e310e9c80867
```
</details>

### 🐛 Describe the bug

For details, please check: https://buildkite.com/vllm/vllm-omni/builds/13145/canvas?sid=019fcca9-bd1e-4a4e-892e-f212dca0d859&tab=output

l-wave:
This is a re-submit after the CI revert, with the Qwen3-Omni loader issue fixed.

Root cause: the fused loader path only delegated `model.*` tensors from `CodePredictorWrapper.load_weights` into `CodePredictorBaseModel.load_weights`. The Qwen3-Omni realtime CI artifact can expose the code-predictor body tensors directly as `layers.*` / `codec_embedding.*` / `norm.*` rather than under `model.*`. Those q/k/v and gate/up shards therefore did not enter the fused repacking path, and the new missing-fused guard correctly failed with `missing fused parameters`.

Fix: `CodePredictorWrapper.load_weights` now normalizes both naming forms before delegating to the inner fused loader:
- current wrapper layout: `model.*`
- Qwen3-Omni body-direct layout: `layers.*`, `codec_embedding.*`, `norm.*`, `rotary_emb.*`

I also added `test_wrapper_load_weights_handles_legacy_unprefixed_model_shards`, which builds synthetic body-direct Qwen3-Omni-style shards and asserts they are packed into:
- `model.layers.*.self_attn.qkv_proj.weight == cat([q, k, v])`
- `model.layers.*.mlp.gate_up_proj.weight == cat([gate, up])`

Local validation:
- `ruff check`
- `ruff format --check`
- `py_compile`


vllm_omni/model_executor/models/common/qwen3_code_predictor.py
```diff
@@ -175,11 +177,23 @@ def __init__(self, config, *, prefix: str = "") -> None:
...
+        # is narrower: reduce eager host launch overhead while keeping the
+        # downstream q_norm / k_norm / SDPA math and HF checkpoint loading
+        # contract intact.
         bias = getattr(config, "attention_bias", False)
-        self.q_proj = nn.Linear(self.hidden_size, self.num_heads * self.head_dim, bias=bias)
-        self.k_proj = nn.Linear(self.hidden_size, self.num_kv_heads * self.head_dim, bias=bias)
-        self.v_proj = nn.Linear(self.hidden_size, self.num_kv_heads * self.head_dim, bias=bias)
+        self._q_size = self.num_heads * self.head_dim
+        self._kv_size = self.num_kv_heads * self.head_dim
+        self.qkv_proj = nn.Linear(
```
hsliuustc0106, line 192: Because this changes the autoregressive audio path and the comment above acknowledges GPU bit-level drift, please add a main-vs-PR A/B before merging: same checkpoint/inputs/hardware/dtype, at least 3 warmup-excluded latency/throughput/VRAM runs, plus an audio or code-quality comparison. The PR currently provides only static checks, so a quality regression could be silent.

tests/platforms/npu/test_310p_patches.py
```diff
@@ -414,6 +414,21 @@ def random_sample(_probs, generators):
...
 
+def test_310p_attention_forward_uses_fused_qkv(monkeypatch: pytest.MonkeyPatch) -> None:
+    """The fused code predictor dropped q_proj/k_proj/v_proj in favor of a single
+    ``qkv_proj`` (+ ``_split_qkv``). The 310P attention override must follow
+    suit, otherwise its NPU flash-attention path raises AttributeError on 310P.
+    The NPU path itself cannot run in CPU CI, so guard the contract statically."""
+    import inspect
+
+    module, _ = _load_qwen3_tts_patch(monkeypatch)
+    source = inspect.getsource(module._Qwen3CodePredictorAttention310P.forward)
```
hsliuustc0106: This test only inspects source text and never instantiates the patched attention or exercises `prepare_qkv_weights`, packed weight/bias layout, or `_npu_flash_attention`; the fake attention fixture does not even expose `qkv_proj`. Please add a runtime test with NPU ops mocked (or an NPU-marked test) that runs the fused path.

l-wave: > This test only inspects source text and never instantiates the patched attention or exercises `prepare_qkv_weights`, packed weight/bias layout, or `_npu_flash_attention`; the fake attention fixture does not even expose `qkv_proj`. Please add a runtime test with NPU ops mocked (or an NPU-marked test) that runs the fused path.

Thanks for the detailed review. I updated the 310P coverage by replacing the previous source-text check with a mocked runtime test that instantiates the patched attention, runs prepare_qkv_weights(), verifies the packed qkv weight/bias layout, and exercises the fused _npu_flash_attention path end to end with mocked NPU ops. For the main-vs-PR validation on the autoregressive audio path, that A/B was already completed and posted in the earlier PR #4958 ([Perf][Qwen3-TTS] Fuse QKV and gate_up projections in code predictor), using the same checkpoint / inputs / hardware / dtype setup with warmups excluded, plus throughput / latency / VRAM and output-quality comparison. So for this round, I kept the code change focused on closing the runtime-test gap you pointed out, rather than duplicating the already-posted benchmark work in the patch itself.

vllm_omni/model_executor/models/common/qwen3_code_predictor.py
```diff
@@ -997,16 +1145,27 @@ def _prepare_npu_weights(self) -> None:
...
+        some Qwen3-Omni artifacts expose the shared body directly as
+        ``layers.*`` / ``codec_embedding.*``.  Normalize both forms before
+        calling ``CodePredictorBaseModel.load_weights`` so its q/k/v and gate/up
+        repacking guards see the HF shards instead of treating them as unrelated
+        wrapper-level tensors.
+        """
         loaded: set[str] = set()
         model_weights: list[tuple[str, torch.Tensor]] = []
         other_weights: list[tuple[str, torch.Tensor]] = []
+        inner_model_prefixes = ("codec_embedding.", "layers.", "norm.", "rotary_emb.")
```
hsliuustc0106, line 1172: The Qwen3-Omni E2E that motivated this re-submit is still red on this head. Buildkite [build 13660](https://buildkite.com/vllm/vllm-omni/builds/13660), `Omni · Qwen3-Omni Test`: `test_video_to_audio[omni_runner0]` errors with `Orchestrator initialization failed: Engine core initialization failed` and `test_mix_to_text_audio_001[default]` with `Stage 1 replica 0 exited ... before API server became ready`; the stage-1 replica traceback ends inside `load_model`. `test_thinker_only_model_request` passes, so this is specific to the full-pipeline load. Main is green on this PR's exact base commit (build 13666, d1e230c9), so the failure is change-induced — the `model.*` + body-direct normalization isn't sufficient for the artifact the realtime CI uses. Could you pull the root-cause line from the stage-1 replica log (`/tmp/omni_stage_1_replica_0_50349.log`)? `Entrypoint Test with H100` in the same build fails at a similar point and may share the cause.

tests/platforms/npu/test_310p_patches.py
```diff
@@ -414,6 +414,99 @@ def random_sample(_probs, generators):
...
+            {key: value.detach().clone() if torch.is_tensor(value) else value for key, value in kwargs.items()}
+        )
+        kwargs["out"].copy_((kwargs["query"] + kwargs["key"] + kwargs["value"]).to(kwargs["out"].dtype))
+
+    monkeypatch.setattr(module, "maybe_trans_nz", maybe_trans_nz)
+    monkeypatch.setattr(module, "aligned_16", lambda tensor: tensor)
+    monkeypatch.setattr(module.torch_npu, "npu_rotary_mul", npu_rotary_mul, raising=False)
+    monkeypatch.setattr(module.torch_npu, "_npu_flash_attention", npu_flash_attention, raising=False)
+
+    attention = module._Qwen3CodePredictorAttention310P()
```
hsliuustc0106: `_Qwen3CodePredictorAttention310P.__init__` forwards `*args` to `CodePredictorAttention.__init__`, which requires a positional `config` — this zero-arg construction raises `TypeError` before the test exercises anything. Since this file is marked `core_model, cpu`, it is collected by the `Simple · Other Test` lane, which is red on both gpu_1_queue shards of [build 13660](https://buildkite.com/vllm/vllm-omni/builds/13660) and on [AMD build 10614](https://buildkite.com/vllm/vllm-omni-amd-ci/builds/10614). The runtime-coverage direction from the last round is right — it just needs a tiny config passed in (or bypassing the parent `__init__`) so the test actually runs.

vllm_omni/model_executor/models/common/qwen3_code_predictor.py
hsliuustc0106: Still outstanding on this re-submit — the PR body lists only ruff/py_compile. Before merge we still need the main-vs-PR A/B (same checkpoint/inputs/hardware/dtype, at least 3 warmup-excluded latency/VRAM runs) plus an audio/code-quality comparison, since the packing changes GPU accumulation order.

tests/platforms/npu/test_310p_patches.py
l-wave: @hsliuustc0106 Thanks for the review. I fixed the remaining 310P patch issue in the attention override.

The root cause was that `_Qwen3CodePredictorAttention310P.__init__()` assumed the base `CodePredictorAttention` had already populated `num_heads`, `num_kv_heads`, and `head_dim`. In the source-loaded 310P patch test, the base attention is replaced with a minimal fake class, so those fields were missing during initialization. I updated the 310P override to initialize them from `config` when they are not already set, while keeping the production path unchanged.

I also re-ran the relevant checks on H20:
- `ruff format --check` and `ruff check` on the touched files: passed
- `python -m pytest tests/model_executor/models/qwen3_tts/test_code_predictor_dtype.py -q`: 31 passed
- `python -m pytest tests/platforms/npu/test_310p_patches.py::test_310p_attention_forward_runs_fused_qkv_runtime -q`: passed
- `python -m pytest tests/platforms/npu/test_310p_patches.py -q`: 8 passed

The branch has been updated with the fix.

l-wave:
The remaining red RTD job is unrelated to this patch: it comes from the existing `.readthedocs.yml` pre-install command (`python -m pip install --index-url https://download.pytorch.org/whl/cpu torch`), which fails to resolve transitive deps like `flit_core`, and I reverted my docs-side workaround here to avoid introducing an unrelated merge conflict in this PR.

Gaohan123 (approved):
LGTM. Thanks

Gaohan123:
@l-wave Please resolve CI failure. Thanks

PR #5792, opened 2026-08-05
[Model] Batch MiniCPM-o Talker codec sampling

Zhou248:
## Summary

This PR removes per-request serialization from MiniCPM-o 4.5 Stage 1 Talker codec-logit preparation.

- Collects all sampling-eligible Talker rows before codec sampling.
- Runs the codec projection, request-local repetition penalties, top-k/top-p filtering, and softmax as one batch.
- Preserves one RNG generator per request, so request compaction, request ordering, or another request finishing does not perturb codec sequences.
- Replaces one `sampled.item()` device-to-host synchronization per request with one batched transfer.
- Keeps EOS, minimum-token, maximum-token, incomplete-prefill, request-routing, and native-duplex behavior request-local.

`torch.multinomial` intentionally remains row-wise because PyTorch accepts only one generator for an entire call. The optimization batches the expensive logit path and host synchronization without changing request-local RNG semantics.

## Motivation

Profiling concurrent MiniCPM-o requests showed `make_omni_output` invoking `_sample_audio_code` serially for each active request. Each invocation repeated the projection and logit-processing path and ended with its own host synchronization. This made Stage 1 output processing scale linearly with the number of active requests.

## Review follow-up

- Bounds repetition-penalty `torch.bincount` workspace to chunks of at most 16 requests instead of allocating an int64 `batch_size * vocab_size` tensor. The output is preallocated so chunking does not introduce an additional full-batch concatenation buffer.
- Removes the unused single-request `_sample_audio_code` compatibility wrapper.
- Adds request-order independence coverage for request-local RNG and documents why multinomial must remain per-row.
- Documents the default-continue semantics of `stop_flags` for nonterminal and ineligible rows.
- Rebases the branch onto the latest upstream `main`.

## Correctness and tests

Added CPU L1 coverage for:

- equivalence between batched and request-local repetition penalties;
- bounded repetition-penalty workspace across 16/16/1 request chunks;
- deterministic request-local RNG after batch compaction and request reordering;
- one batched sampling call for multiple active requests;
- request-aligned codec deltas and duplex metadata;
- incomplete-prefill, EOS, and maximum-token terminal behavior.

Completed locally:

```bash
PYTHONPATH=/tmp/codex-ruff-014 python3 -m ruff check \
  vllm_omni/model_executor/models/minicpmo_4_5/minicpmo_4_5_omni_tts.py \
  tests/model_executor/models/minicpmo_4_5/test_talker_batching.py
PYTHONPATH=/tmp/codex-ruff-014 python3 -m ruff format --check \
  vllm_omni/model_executor/models/minicpmo_4_5/minicpmo_4_5_omni_tts.py \
  tests/model_executor/models/minicpmo_4_5/test_talker_batching.py
python3 -m compileall -q \
  vllm_omni/model_executor/models/minicpmo_4_5/minicpmo_4_5_omni_tts.py \
  tests/model_executor/models/minicpmo_4_5/test_talker_batching.py
git diff --check upstream/main...HEAD
```

The focused pytest command is:

```bash
pytest -q tests/model_executor/models/minicpmo_4_5/test_talker_batching.py \
  -m "core_model and cpu" --run-level core_model
```

It was not run in the local authoring environment: the Homebrew Python lacks pytest, while the available Anaconda PyTorch installation cannot load `libtorch_cpu.dylib`. CI is expected to run the CPU regression suite.

## Performance validation

This PR deliberately makes no numerical performance claim yet. NPU concurrency 1 and 4 profiling should be attached before merge, including Stage 1 output latency, end-to-end TTFT/throughput, peak memory for codec sampling, and the `_batch_stop_logits` path before and after this change.

### DIFF
--- a/tests/model_executor/models/minicpmo_4_5/test_talker_batching.py
+++ b/tests/model_executor/models/minicpmo_4_5/test_talker_batching.py
@@ -192,13 +192,6 @@ def test_batched_repetition_penalty_matches_rows_across_chunks(mocker) -> None:
         for row in range(batch_size)
     ]
 
-    bincount = mocker.spy(torch, "bincount")
-    actual = _apply_batched_repetition_penalty(
-        logits,
-        histories,
-        penalty=1.2,
-        window_size=5,
-    )
     expected = torch.cat(
         [
             _reference_repetition_penalty(
@@ -210,6 +203,13 @@ def test_batched_repetition_penalty_matches_rows_across_chunks(mocker) -> None:
             for row, history in enumerate(histories)
         ],
         dim=0,
+    )
+    bincount = mocker.spy(torch, "bincount")
+    actual = _apply_batched_repetition_penalty(
+        logits,
+        histories,
+        penalty=1.2,
+        window_size=5,
     )
 
     assert torch.equal(actual, expected)
--- a/tests/model_executor/models/minicpmo_4_5/test_talker_batching.py
+++ b/tests/model_executor/models/minicpmo_4_5/test_talker_batching.py
@@ -12,6 +12,7 @@ from vllm_omni.model_executor.models.minicpmo_4_5.minicpmo_4_5_omni import (
     MiniCPMO45OmniForConditionalGeneration,
 )
 from vllm_omni.model_executor.models.minicpmo_4_5.minicpmo_4_5_omni_tts import (
+    _REPETITION_PENALTY_CHUNK_SIZE,
     MiniCPMO45OmniTTSForConditionalGeneration,
     _apply_batched_repetition_penalty,
     _max_audio_tokens,
@@ -89,6 +90,35 @@ def _reference_repetition_penalty(
     return torch.where(logits < 0, logits * alpha, logits / alpha)
 
 
+def _make_sampling_talker() -> MiniCPMO45OmniTTSForConditionalGeneration:
+    talker = _make_talker()
+    talker._codec_temperature = 0.8
+    talker._codec_top_k = 5
+    talker._codec_top_p = 0.85
+    talker._codec_repetition_penalty = 1.05
+    talker._request_audio_states = {
+        "req-a": {"min_tokens": 2},
+        "req-b": {"min_tokens": 0},
+    }
+    talker.head_code = nn.ModuleList([nn.Linear(2, 8, bias=False)])
+    with torch.no_grad():
+        talker.head_code[0].weight.copy_(
+            torch.tensor(
+                [
+                    [0.1, -0.3],
+                    [0.2, 0.4],
+                    [-0.5, 0.2],
+                    [0.7, -0.1],
+                    [-0.2, 0.8],
+                    [0.6, 0.3],
+                    [-0.4, -0.7],
+                    [0.3, 0.5],
+                ]
+            )
+        )
+    return talker
+
+
 @pytest.mark.parametrize(
     ("condition_tokens", "expected"),
     [(3, 64), (100, 1000), (1000, 2048)],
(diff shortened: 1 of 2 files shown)


amy-why-3459:
## Summary

This PR batches the codec-sampling path in MiniCPM-o 4.5's Stage 1 Talker (`make_omni_output` → `_sample_audio_codes`): it collects all sampling-eligible rows first, runs the projection / repetition-penalty / top-k / top-p / softmax once as a batch, and replaces one `.item()` host sync per request with a single batched transfer. Request-local RNG streams are preserved by keeping the per-row `torch.multinomial` loop.

The core idea and refactoring are sound, and the tests cover the key request-alignment scenarios. A few points need attention before merging.

---

## 1. (P1) `bincount` in `_apply_batched_repetition_penalty` can blow up memory at larger batch sizes

```python
encoded = encoded_rows[0] if len(encoded_rows) == 1 else torch.cat(encoded_rows)
frequencies = torch.bincount(encoded, minlength=batch_size * vocab_size).reshape(batch_size, vocab_size)
```

`vocab_size` is `_num_audio_tokens` (default **6562**). With a larger concurrent batch (e.g. multi-card / NPU with high concurrency), the `batch_size * vocab_size` intermediate counting tensor is allocated in one shot. The old code ran one `bincount(minlength=vocab_size)` per request, so peak allocation was bounded by a single row.

Consider:
- If the batch is typically small (2–8), the current approach is fine.
- If concurrency can reach tens of rows on NPU, use a chunked / per-row `bincount` (e.g. loop over rows or process in chunks, then `torch.stack`) to avoid the `batch_size * vocab_size` intermediate.

Also note that each history is truncated to `[-_REPETITION_WINDOW:]`, so `torch.cat(encoded_rows)` peak is proportional to `batch_size * window_size` — per-row bincounting then stacking would keep memory more bounded.

---

## 2. (P2) The `_sample_audio_code` compatibility wrapper has no remaining callers

The single-request wrapper is kept:

```python
def _sample_audio_code(self, hidden_state, history, request_id, step):
    """Compatibility wrapper for one request."""
    return self._sample_audio_codes(hidden_state, (history,), (request_id,), (step,)).reshape(())
```

`git grep` shows no production callers inside vLLM-Omni anymore — only the tests use it. If it is not part of a public/external interface:
- Remove the wrapper and keep only `_sample_audio_codes`, or
- Add a comment documenting who consumes it (e.g. external plugins / deployment code).

---

## 3. (P2) RNG determinism test coverage is a bit thin

`test_batched_codec_sampling_preserves_request_rng_after_compaction` verifies:
- Batched vs standalone sampling produce identical per-request results (`first_batch[1] == standalone_b_first`).
- Request RNG does not drift after compaction (`compacted_b[0] == standalone_b_second`).

It does **not** cover:
- Whether reordering requests in a batch changes per-request output (it should not, but it isn't pinned).
- The constraint that `torch.multinomial` must remain per-row because PyTorch accepts only one generator per call. If someone later replaces the loop with a single `torch.multinomial(prob, generator=shared)`, request-local RNG semantics silently break.

Suggest adding a brief comment and/or a test asserting per-row multinomial is mandatory for request-local RNG.

---

## 4. (P2) `stop_flags` default-`False` semantics should be documented

```python
stop_flags = [False] * len(infos)
```

Every branch assigns the right value (finished → `True`, everything else stays `False`), so the logic is correct. But because the `continue` branches no longer explicitly append a placeholder row (unlike the old `stop_rows` append style), the intent of "default False, set True only when finished" should be commented to prevent a future refactor from silently leaving a wrong flag.

---

## 5. (P2) Performance claim not yet validated (author acknowledges)

The PR description explicitly says: "This draft deliberately makes no numerical performance claim yet" and plans NPU concurrency 1 and 4 profiling. As a reviewer:

- Consider approving only after NPU measurements are attached (Stage 1 output latency, e2e TTFT/throughput before/after).
- At minimum, confirm the `_batch_stop_logits` change from `torch.stack(stop_rows)` to `hidden.new_tensor(list)` does not regress the (small) finished-row path.

---

## Checklist

| Item | Priority | Status |
|---|---|---|
| `bincount` batch memory amplification | P1 | Needs attention at larger batches |
| Dead `_sample_audio_code` wrapper | P2 | Remove or document |
| RNG determinism test coverage | P2 | Suggestion |
| `stop_flags` default-`False` semantics comment | P2 | Suggestion |
| NPU performance data | P2 | Awaiting author's profiling |

The batching direction and request-local semantics are correct, and tests cover the key scenarios. Recommend addressing the P1 item (and ideally attaching NPU perf numbers) before merge.

amy-why-3459:
@vllm-omni-review-bot

amy-why-3459 (approved):

PR #5794, opened 2026-08-05
[Feature] Add topology-aware load balancer for KV-transfer-aware routing

Claisenn:
## Summary

Add a new `topology-aware` load balancing policy that prefers replicas sharing the same `topology_domain` (NVLink/NUMA/IB domain) as the downstream consumer stage.

## Motivation

In disaggregated multi-stage serving (e.g. Thinker to Talker to Code2Wav, or prefill to decode PD), the KV cache / hidden states transferred between stages can be gigabytes per request. Inter-GPU bandwidth varies by ~72x depending on physical topology:

- NVLink within a domain: ~900 GB/s
- InfiniBand across nodes: ~50 GB/s
- TCP across data centers: ~12.5 GB/s

(Source: arXiv:2607.28633, "Topology-Aware Data Movement for Disaggregated GPU Inference")

The existing balancers (random, round-robin, least-queue-length) ignore this, so a request can be routed to a replica on a different NVLink domain than its consumer, forcing KV transfer over a slow cross-domain link.

## What this PR does

- `load_balancer.py`: new `TopologyAwareBalancer` class + `TOPOLOGY_AWARE = "topology-aware"` enum value. When a task carries a `topology_domain` hint, it restricts candidates to replicas in that domain, then picks the one with the smallest `queue_length`. Falls back to global `least-queue-length` when no hint is given or no replica matches.
- `messages.py`: `ReplicaInfo` gains an optional `topology_domain: str | None = None` field (placed last, fully backward compatible).
- `__init__.py`: export `TopologyAwareBalancer`.
- `stage_runtime.py`: register the new policy in `_build_load_balancer_factory`.
- `serve.py`: add `"topology-aware"` to `--omni-lb-policy` choices.

## Why this is safe

- No torch/CUDA dependency - pure Python, testable on CPU.
- Backward compatible - `topology_domain` defaults to `None`; existing `ReplicaInfo` constructors and tests are unchanged.
- Additive - new policy, no change to existing `random` / `round-robin` / `least-queue-length` behavior.
- Graceful fallback - if no replica shares the requested domain, behaves identically to `least-queue-length`.

## Testing

Verified locally with standalone Python (no GPU):
- Hint present -> always picks least-loaded replica in the matching domain
- No hint / unknown domain -> falls back to global least-queue-length
- Empty replicas / negative `queue_length` -> raises `ValueError` (consistent with `LeastQueueLengthBalancer`)
- Existing `RandomBalancer` / `RoundRobinBalancer` / `LeastQueueLengthBalancer` still work

## Follow-up (not in this PR)

- Populate `topology_domain` from actual NVLink/NUMA topology during replica registration (`OmniCoordClientForStage`).
- Propagate the consumer stage's chosen replica domain as the `topology_domain` hint in `StagePool.pick()`.

## References

- arXiv:2607.28633 - Topology-Aware Data Movement for Disaggregated GPU Inference

### DIFF
--- a/vllm_omni/distributed/omni_coordinator/messages.py
+++ b/vllm_omni/distributed/omni_coordinator/messages.py
@@ -29,6 +29,7 @@ class ReplicaEvent:
     event_type: str  # "update" | "heartbeat"
     status: ReplicaStatus  # Current status
     queue_length: int  # Current queue length
+    topology_domain: str | None = None  # NVLink/NUMA/IB domain for topology-aware routing
 
 
 @dataclass
--- a/vllm_omni/distributed/omni_coordinator/messages.py
+++ b/vllm_omni/distributed/omni_coordinator/messages.py
@@ -46,6 +46,7 @@ class ReplicaInfo:
     queue_length: int  # Current queue length of this replica
     last_heartbeat: float  # Timestamp of the last heartbeat received (seconds)
     registered_at: float  # Timestamp when the replica was registered (seconds)
+    topology_domain: str | None = None  # NVLink/NUMA/IB domain for topology-aware routing
 
 
 @dataclass
--- a/vllm_omni/engine/orchestrator.py
+++ b/vllm_omni/engine/orchestrator.py
@@ -311,6 +311,7 @@ class _OrchestratorDuplexStagePort:
                     context.request_id,
                     request,
                     request_state,
+                    stage0_replica_id=replica_id,
                 )
         request_state.duplex_stage_fences[context.stage_id] = context.fence
         request_state.stage_submit_ts[context.stage_id] = _time.time()
@@ -682,7 +683,7 @@ class Orchestrator:
         preprocess_ms = msg.preprocess_ms
         if preprocess_ms > 0:
             req_state.pipeline_timings["preprocess_ms"] = preprocess_ms
-        await self.stage_pools[stage_id].submit_initial(
+        stage0_replica_id = await self.stage_pools[stage_id].submit_initial(
             request_id,
             req_state,
             prompt,
@@ -690,7 +691,9 @@ class Orchestrator:
         )
 
         if self.async_chunk and stage_id == 0 and final_stage_id > 0:
-            await self._prewarm_async_chunk_stages(request_id, prompt, req_state)
+            await self._prewarm_async_chunk_stages(
+                request_id, prompt, req_state, stage0_replica_id=stage0_replica_id
+            )
 
     async def _handle_streaming_update(self, msg: StageSubmissionMessage) -> None:
         """Handle a streaming_update message for an existing request."""
@@ -725,7 +728,10 @@ class Orchestrator:
         )
 
         if self.async_chunk and stage_id == 0 and final_stage_id > 0:
-            await self._prewarm_async_chunk_stages(request_id, request, req_state)
+            stage0_replica_id = self.stage_pools[stage_id].get_bound_replica_id(request_id)
+            await self._prewarm_async_chunk_stages(
+                request_id, request, req_state, stage0_replica_id=stage0_replica_id
+            )
 
     async def _handle_add_companion(self, msg: AddCompanionRequestMessage) -> None:
         """Handle an add_companion_request message: submit companion to stage 0."""
@@ -1602,11 +1608,13 @@ class Orchestrator:
         if (stage_id + 1) in parent_state.stage_submit_ts:
             return
 
+        src_replica_id = self.stage_pools[stage_id].get_bound_replica_id(parent_id)
         await self._forward_to_next_stage(
             parent_id,
             stage_id,
             deferred["engine_outputs"],
             parent_state,
+            src_replica_id=src_replica_id,
         )
 
     async def _handle_kv_ready_raw_outputs(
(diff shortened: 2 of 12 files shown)


Claisenn:
@xuechendi @alex-jw-brooks @Sy0307 @princepride @yuanheng-zhao @tzhouam gentle ping for review when you have a chance 🙏

This PR adds a new `topology-aware` load balancing policy that routes requests to replicas sharing the same `topology_domain` (NVLink/NUMA/IB domain) as the downstream consumer, which can significantly reduce KV-cache / hidden-state transfer latency in disaggregated multi-stage serving.

The change is additive and backward compatible (new enum value, optional `topology_domain` field defaulting to `None`, graceful fallback to `least-queue-length` when no hint/match). It's pure Python with no torch/CUDA dependency, so it's testable on CPU.

vllm_omni/distributed/omni_coordinator/messages.py
```diff
@@ -46,6 +46,7 @@ class ReplicaInfo:
     queue_length: int  # Current queue length of this replica
     last_heartbeat: float  # Timestamp of the last heartbeat received (seconds)
     registered_at: float  # Timestamp when the replica was registered (seconds)
+    topology_domain: str | None = None  # NVLink/NUMA/IB domain for topology-aware routing
```
yuanheng-zhao, line 50: Where is this `topology_domain` set or assigned?

yuanheng-zhao (commented):
Thanks for contributing. The purpose is good while the current code changes seem not completed, as there's no caller ever puts `topology_domain` into a Task

Claisenn:
@yuanheng-zhao Thanks for the review! I've completed both follow-ups to make the feature fully functional. Could you please take another look?

**What was added to address your feedback:**

1. **`topology_domain` is now populated during replica registration**
   - `ReplicaEvent` gains a `topology_domain` field, sent by `OmniCoordClientForStage` in every event (update + heartbeat).
   - New `detect_topology_domain()` helper returns `OMNI_TOPOLOGY_DOMAIN` env var if set, otherwise the hostname (a reasonable first proxy for NVLink/NUMA/IB domain).
   - `OmniCoordinator` parses it and stores it in `ReplicaInfo`; `OmniCoordClientForHub` decodes it from the published `ReplicaList`.
   - Both LLM (`stage_engine_core_proc.py`) and diffusion (`stage_diffusion_proc.py`) stage procs pass `detect_topology_domain()` when creating their coord client.

2. **`topology_domain` is now put into the `Task` for the consumer stage**
   - `StagePool.get_replica_topology_domain(replica_id)` looks up the domain from the hub's cached replica list.
   - `pick` / `preselect_replica_id` / `submit_initial` / `submit_update` accept a `topology_domain` kwarg and inject it into `Task["topology_domain"]` before calling the balancer.
   - In `Orchestrator._forward_to_next_stage`, the source (producer) replica's domain is read and passed to the downstream (consumer) stage's submit call — so the consumer is placed in the same NVLink/NUMA/IB domain as its producer, keeping KV-cache transfer on a fast local link.
   - The async-chunk prewarm path also chains the hint (seeded from stage-0's chosen replica, updated each iteration from the just-picked replica) so prewarm doesn't pre-bind a cross-domain replica that the real forward would then be stuck with.

**Backward compatibility / safety:**
- All new fields/params default to `None`; missing `topology_domain` in wire payloads is handled via `.get()`.
- When no hint is present or no replica shares the requested domain, `TopologyAwareBalancer` falls back to global `least-queue-length`.
- Already-bound requests (streaming updates) keep their existing replica; the hint only applies to the first pick.

The full data flow is now: stage proc detects domain → coord client sends it → coordinator stores it → hub decodes it → orchestrator reads the producer's domain → passes it as hint to the consumer stage's balancer → balancer prefers same-domain replicas.

vllm_omni/distributed/omni_coordinator/messages.py
Claisenn: Good catch. `topology_domain` is now set at two layers:

1. **Replica registration time** — the stage proc calls `detect_topology_domain()` (in `omni_coord_client_for_stage.py`), which returns the `OMNI_TOPOLOGY_DOMAIN` env var if set, otherwise the hostname. That value is passed into `OmniCoordClientForStage`, stored as `self._topology_domain`, and included in every `ReplicaEvent` sent to the coordinator. The coordinator stores it on `ReplicaInfo`, and the hub decodes it from the published `ReplicaList`.

2. **Cross-stage routing time** — in `Orchestrator._forward_to_next_stage`, the producer replica's domain is read via `StagePool.get_replica_topology_domain(src_replica_id)` and passed as the `topology_domain` kwarg to the consumer stage's `submit_initial`/`submit_update`. That flows into `pick` → `Task["topology_domain"]`, which `TopologyAwareBalancer` reads to prefer same-domain replicas.

So the field is populated at registration and consumed at routing time. See the latest push (`ead92463`) for the full wiring.

yuanheng-zhao:
PTAL @SamitHuang @Gaohan123 @alex-jw-brooks , do we need this feature for now?

vllm_omni/engine/orchestrator.py
```diff
@@ -2145,6 +2194,7 @@ async def _prewarm_async_chunk_stages(
                     req_state,
                     request,
                     prompt_text=None,
+                    topology_domain=topology_domain,
```
lishunyang12, line 2195: Resolve the actual producer before prewarming receivers across sender-only stages.

vllm_omni/engine/orchestrator.py
```diff
@@ -311,6 +311,7 @@ async def submit(self, submission: DuplexStageSubmission) -> DuplexStageSubmissi
                     context.request_id,
                     request,
                     request_state,
+                    stage0_replica_id=replica_id,
```
lishunyang12, line 314: Update the prewarm test doubles and assertions for both new routing arguments.

vllm_omni/engine/orchestrator.py
Claisenn: Fixed in ebf6b1255. Prewarm now resolves each receiver’s actual producer from its `engine_input_source`, then looks up that producer’s request-bound replica and topology domain. This avoids incorrectly chaining through skipped sender-only stages; stage 0 uses the replica ID returned by the initial submission.

vllm_omni/engine/orchestrator.py
Claisenn: Fixed in ebf6b1255. The duplex prewarm test now asserts `stage0_replica_id=10`, and the sender-only bridge test records submit kwargs and verifies that the resolved producer topology is passed as `topology_domain` to the receiver.

PR #5804, opened 2026-08-05
Fix to preserve parallel_config in stage overrides

mengker33:
A parallel_config passed in offline diffusion examples was classified as an internal orchestrator argument and dropped during registry/deploy stage resolution. Preserve the override and merge it into the diffusion stage's nested engine_args.parallel_config, so TP, CFG, SP, and etc. settings reach the diffusion worker.

## Test Plan

**vLLM Version:**
0.26.0
**vLLM-Omni Commit:**
6510de344ebff729e9e066bef1a23ff120ce422c
## Test Result
<pre>
python examples/offline_inference/text_to_video/text_to_video.py --model Wan-AI/Wan2.2-T2V-A14B-Diffusers --prompt "Two anthropomorphic cats in comfy boxing gear and bright gloves fight intensely on a spotlighted stage." --height 720 --width 1280 --num-frames 81 --num-inference-steps 40 --boundary-ratio 0.875 --flow-shift 5.0 --fps 16 --tensor-parallel-size 4  --enable-cpu-offload  --vae-use-slicing --vae-use-tiling --enable-diffusion-pipeline-profiler
</pre>
We will see tp=4 does not take effect, the model is running on single card.
This PR fix this issue and ensure parallelism config is passed correctly. 

### DIFF
--- a/vllm_omni/config/stage_config.py
+++ b/vllm_omni/config/stage_config.py
@@ -64,12 +64,16 @@ def build_stage_runtime_overrides(
     if internal_keys is None:
         from vllm_omni.engine.arg_utils import SHARED_FIELDS, internal_blacklist_keys
 
-        # Some fields are modeled as orchestrator-owned for top-level CLI
-        # parsing, but are also legitimate deploy-time stage overrides. Keep
-        # the default blacklist for true orchestrator/shared fields while
-        # allowing any field explicitly represented by the deploy schema to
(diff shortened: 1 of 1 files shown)


mengker33:
The issue is gone with merged commit: 8dc3d504d8bae73bd6a627b4b759a4dbe6cd35fb
Close this PR.

PR #5808, opened 2026-08-05
[Perf][DreamZero] Build the RoPE freqs table once per forward, and rotate in real float32 arithmetic

ZJLi2013:
## Purpose

Three inefficiencies on the same RoPE path in `causal_wan_model.py`, one commit each, separable.

**The table was rebuilt per call.** `causal_rope_action_apply()` appended the step's action/state rows to
the video table inside itself, so its two `torch.cat` ran 40 layers × q and k = 80 times per forward for a
table that depends only on the step index. It is now split out as `causal_rope_action_freqs()` and called
once in `_forward_blocks`; `freqs_action`, `freqs_state` and `current_start_frame` retire from two
module-internal signatures.

**The rotation broke inductor's fusion.** Inductor cannot generate code for complex operators, so the
complex multiply falls back to eager and, being opaque to the scheduler, also splits the surrounding
qk-norm and layout work into separate fusion groups — three kernels where one would do. Written out in real
arithmetic the whole segment fuses. Eager keeps the complex form, selected by
`torch.compiler.is_compiling()`: unfused, the real form is about 4x slower.

**The rotation ran in float64 to produce a bfloat16 result.** Only the angles need the width — the table
reaches 1e4 radians, where float32 would resolve about 1e-3 per radian — and they are spent once
`torch.polar` has run. So `rope_params()` still accumulates in float64 and hands the table over as
complex64, and the rotation runs in float32. Dropping only one of the two would not help: a complex128
table promotes a complex64 activation straight back.

## Numerical behaviour

**The first two commits are bitwise identical on the bf16 output; the third moves it by up to one bfloat16
ulp.**

The hoist is a seam split — the build moves out verbatim and the rotation that followed it is already
`rope_apply()` line for line — and the `cat` feeds no reduction, so the packed table does not depend on how
many times it is evaluated. The real form evaluates the same products as the complex multiply; their
intermediates can differ by under 1 ulp where a complex multiply contracts into an FMA, but that does not
survive the cast back to bfloat16: `torch.equal` holds over 1.83e8 elements.

Float32 rotation moves 4595 of those 1.83e8 elements (2.5e-5), by at most 1.562e-2 = 2⁻⁶, one bfloat16 ulp
for values in [4, 8). Measured over 20 seeds at the rollout's shapes, after the `.type_as(v)` that feeds
attention.

This also matches what the rest of the repo does. `x.to(torch.float64)` appears in two places — this file
and `diffusion/layers/rope.py`, both carried over from the Wan2.1 reference — while `qwen_image`,
`omnigen2`, `boogu_image`, `lingbot_video` and `z_image` build the table in float64 and rotate in float32.
Three transformers already gate on `current_omni_platform.supports_float64()`, so this also lets DreamZero
run where float64 is unavailable.

## Performance

**DreamZero-DROID rollout on MI350X (gfx950), TP=1 + CFG-parallel=2, 15 chunks / 16 steps**, mean over the
steady-state steps. Each commit is measured against the tree carrying the ones before it, comparing per
step position so the 4-step window-refresh cycle cancels:

| commit | before | after | paired per-step diff |
|---|---|---|---|
| rotate in real arithmetic | 502.8 ms | **497.2 ms** | **−5.6 ms, 30/30 positions faster** |
| rotate in float32 | 499.8 ms | **492.2 ms** | **−7.6 ms, sd 1.3, 15/15 positions faster** |

The two rows are separate sessions; the absolute numbers are not additive.

The compiled segment (qk-norm + RoPE + layout) at q of `[1, 1785, 40, 128]` goes **96.4 → 78.2 → 65.7 µs**
across the second and third commits. The hoist takes the table build from 1333 µs per forward to 22 µs,
worth 469–556 µs (1.08–1.10x) over all 80 projections.

Both mechanisms are TorchInductor and bandwidth properties rather than ROCm ones, so they should carry to
NVIDIA; the magnitude has not been measured there.

## Test Plan

```bash
pytest -sv tests/diffusion/models/dreamzero/
```

MI350X (gfx950), ROCm 7.2.3, torch 2.11.0+gitd0c8b1f, image
`vllm/vllm-openai-rocm:nightly-49f31d7cee425a6d38f8c5bc76877986daf832ed` plus this repo's runtime deps.

**vLLM Version:** 0.23.1rc1.dev1533+g49f31d7ce.rocm723

**vLLM-Omni Commit:** 2eada79b99ae64acd835e5ca3794eb6ff60f4424

## Test Result

```
49 passed in 1.13s
```

Same 49 on the base with the change reverted.

### DIFF
--- a/vllm_omni/diffusion/models/dreamzero/causal_wan_model.py
+++ b/vllm_omni/diffusion/models/dreamzero/causal_wan_model.py
@@ -70,6 +70,12 @@ def sinusoidal_embedding_1d(dim: int, position: torch.Tensor) -> torch.Tensor:
 
 def rope_params(max_seq_len: int, dim: int) -> torch.Tensor:
     """Precompute complex-valued RoPE frequencies (polar form).
+
+    The angles are accumulated in float64 -- at the far end of the table they
+    reach 1e4 radians, where float32 would resolve them to only ~1e-3 -- and the
+    table is handed out as complex64, since cos and sin land in [-1, 1] and the
+    rotation they feed ends in bfloat16.
+
     Returns: complex tensor [max_seq_len, dim // 2]
     """
     if dim % 2 != 0:
@@ -79,7 +85,7 @@ def rope_params(max_seq_len: int, dim: int) -> torch.Tensor:
         1.0 / torch.pow(10000, torch.arange(0, dim, 2).to(torch.float64).div(dim)),
     )
     freqs = torch.polar(torch.ones_like(freqs), freqs)
-    return freqs
+    return freqs.to(torch.complex64)
 
 
 def rope_apply(x: torch.Tensor, freqs: torch.Tensor) -> torch.Tensor:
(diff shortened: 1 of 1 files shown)


vllm_omni/diffusion/models/dreamzero/causal_wan_model.py
```diff
@@ -79,16 +85,40 @@ def rope_params(max_seq_len: int, dim: int) -> torch.Tensor:
...
+    pairs = x.to(torch.float32).reshape(B, seq_len, n, -1, 2)
+    if not torch.compiler.is_compiling():
+        rotated = torch.view_as_complex(pairs) * torch.view_as_complex(freqs).unsqueeze(0)
+        return torch.view_as_real(rotated).flatten(3)
+    x_re, x_im = pairs[..., 0], pairs[..., 1]
+    cos, sin = freqs[..., 0].unsqueeze(0), freqs[..., 1].unsqueeze(0)
+    return torch.stack((x_re * cos - x_im * sin, x_re * sin + x_im * cos), dim=-1).flatten(3)
 
 
 def rope_action_apply(
```
hsliuustc0106, line 124: `rope_action_apply` seems to have no callers left anywhere in the repo (the multi-step path was already unused before this PR). Since you're touching it anyway, delete it?

Issue #5700, opened 2026-08-03
[Roadmap] MiniMax-H3 follow-up: accuracy/performance CI, quantization, and optimizations


david6666666:
## MiniMax-H3 progress <-> Issue/PR mapping (updated 2026-09-11)

Legend: :white_check_mark: merged/done - :hourglass_flowing_sand: open/in-progress - :warning: partial/limited or closed without delivery - :x: unimplemented/unvalidated follow-up

Verified against `vllm-omni/main@3d952d133` and live GitHub states on 2026-09-11. Merged implementation, measured performance, and hardware/feature-combination qualification are recorded separately; benchmark figures retain the cited revision and workload scope.

#### 1.1 API alignment

| Status | Ref | Note |
|---|---|---|
| :warning: | #5751 (issue) + #5752 + #6391 (issue) + #6497 | The official FL2VA/Ref2VA input matrix is merged; follow-ups track an Ascend first/last-frame control regression and restore aspect-preserving letterbox preprocessing for reference images |
| :white_check_mark: | #7281 | Mixed Ref2VA requests now keep video-soundtrack and standalone-audio budgets separate. The formerly rejected image + video + 15-second WAV request completes on 4xB300; individual input limits remain enforced. |
| :warning: | #7254 (closed issue) | The mixed-reference content mismatch was also reproduced on CUDA; explicit reference roles and narration improved the tested request. The issue is closed after investigation, but #7281 fixes input acceptance only. Hosted-API prompt/preprocessing parity and general quality parity are not established. |
| :white_check_mark: | #5720 | Modular pipeline: FL2VA/Ref2VA share text_encoder/vae; startup task selection via `--task-type` |
| :hourglass_flowing_sand: | #5756 + #5905 + #6782 | ComfyUI r2v frontend support is merged; #5905 exposes `generate_sound`, while #6782 separately fixes the PyAV single-pass demux bug that drops H3 audio when converting MP4 responses to ComfyUI `VIDEO` objects and remains in review |

#### 1.2 Feature completeness and compatibility

| Status | Ref | Note |
|---|---|---|
| :warning: | #5709 + #5978 | The original T2VA reference-accuracy case was merged, then removed by #5978 because it did not meet the retained strict gate. Current official-asset accuracy coverage is I2VA/Ref2VA; T2VA smoke/performance coverage is not pixel-level reference alignment. |
[...]

PR #5810, opened 2026-08-05
[Feature][MiniMax-H3] Support diffusion continuous batching

princepride:
## Purpose

Closes the Diffusion Continuous Batching item of #5700.

MiniMax-H3 ran its whole denoise loop inside one `forward()`, so it could not join the step-wise scheduler: one request occupied the engine end to end. This implements the step-execution contract (`prepare_encode` / `denoise_step` / `step_scheduler` / `post_decode`).

H3's DiT is already a variable-length packed model, so co-batched requests are concatenated into **one sequence that keeps one attention document per request** (plus that request's 64-row padding tail). Attention never crosses a request boundary and a batch costs one DiT forward.

Notes for reviewers:

- Request mode and step mode share `_prepare_request_inputs()`, `_build_denoise_inputs()`, `minimax_h3_prepare_denoise_rows()` and `_unpack_denoised_rows()`, so the two paths cannot drift.
- Video rows are the batched tensor the runner slices per request; audio rows have a different width, so they travel through request-private state along with the audio sigma schedule.
- Multi-request packing needs a backend that consumes `cu_seqlens` (`--diffusion-attention-backend FLASH_ATTN`). Other backends stay correct by falling back to one forward per request.
- `_run_packed_attention` now skips the KV prefix length and the 1-D pad mask when `cu_seqlens` describes more than one document, because neither can express a block-diagonal batch.
- Step mode rejects two request-mode-only features with explicit errors: `num_outputs_per_prompt > 1` (a request state holds one latent tensor) and distributed layerwise offload (its resident-layer window spans a whole denoise loop).

**Batching does not speed H3 up.** It is implemented for scheduler-level control, and the measurements below are recorded in the recipe so users are not misled.

## Test Plan

**vLLM Version:** 0.26.0
**vLLM-Omni Commit:** `ce17696a` (rebased on `9235b0ae`)
**Hardware:** 2 x H100 80GB

### 1. Correctness

```bash
pytest -sv tests/diffusion/models/minimax_h3 -m 'core_model and cpu'
```

Plus, on real weights, the same seeded prompt through both paths, and two concurrent requests co-batched:

```bash
vllm serve $MODEL_ROOT/FL2VA --omni --port 8091 --trust-remote-code \
  --tensor-parallel-size 2 --text-encoder-tp-size 2 --vae-use-tiling --enforce-eager \
  --diffusion-attention-backend FLASH_ATTN --step-execution --max-num-seqs 2
```

### 2. Performance

Same workload against three server configs (`request` = no `--step-execution`, `step1` = `--max-num-seqs 1`, `step4` = `--max-num-seqs 4`):

```bash
python3 benchmarks/diffusion/diffusion_benchmark_serving.py \
  --base-url http://127.0.0.1:8091 --endpoint /v1/videos --task t2v --dataset random \
  --model $MODEL_ROOT/FL2VA --num-prompts 4 --max-concurrency 4 --warmup-requests 1 \
  --width 672 --height 384 --fps 24 --num-inference-steps 30 --seed 1101 --disable-tqdm
```

### 3. Operator comparison

Loads only the DiT and profiles 3 separate single-request forwards against 1 fused three-request forward, using the same code paths `denoise_step()` takes.

<details><summary>profile_batching.py</summary>

```python
import json, os
from collections import defaultdict
from pathlib import Path
from types import SimpleNamespace

os.environ.setdefault("DIFFUSION_ATTENTION_BACKEND", "FLASH_ATTN")
os.environ.setdefault("MASTER_ADDR", "127.0.0.1")
os.environ.setdefault("MASTER_PORT", "29591")

import torch
from safetensors.torch import safe_open
from vllm.config import VllmConfig, set_current_vllm_config
from vllm.distributed import init_distributed_environment, initialize_model_parallel

from vllm_omni.diffusion.models.minimax_h3.denoise_loop import MiniMaxH3DenoiseBranch
from vllm_omni.diffusion.models.minimax_h3.minimax_h3_transformer import MiniMaxH3DiTModel
from vllm_omni.diffusion.models.minimax_h3.packed_sequence import minimax_h3_packed_sequence
from vllm_omni.diffusion.models.minimax_h3.step_batch import minimax_h3_batched_forward_kwargs

MODEL = os.environ["MODEL_ROOT"] + "/FL2VA"
REQUESTS, ITERS, WARMUP = 3, 3, 2
# 209 frames at 672x384 -> 16384 packed rows per request.
SHAPE = dict(text_len=9, latent_t=62, latent_h=24, latent_w=42, audio_t=348)

init_distributed_environment(world_size=1, rank=0, local_rank=0, distributed_init_method="env://", backend="nccl")
config = json.loads((Path(MODEL) / "transformer" / "config.json").read_text())
od_config = SimpleNamespace(tf_model_config=config, parallel_config=SimpleNamespace(ulysses_degree=1))
vllm_config = VllmConfig()
with set_current_vllm_config(vllm_config):
    initialize_model_parallel(tensor_model_parallel_size=1)
    torch.set_default_device("cuda")
    model = MiniMaxH3DiTModel(od_config, quant_config=None)

    def weights():
        for shard in sorted((Path(MODEL) / "transformer").glob("*.safetensors")):
            with safe_open(shard, framework="pt", device="cpu") as f:
                for name in f.keys():
                    yield name, f.get_tensor(name)

    model.load_weights(weights())
    model.post_load_weights()
model.eval()
torch.set_default_device("cpu")

branches, videos, audios = [], [], []
for index in range(REQUESTS):
    packed = minimax_h3_packed_sequence(**SHAPE, include_keyframe_cond=False)
    gen = torch.Generator().manual_seed(index)
    branch = MiniMaxH3DenoiseBranch(
        packed=packed,
        text_embeddings=torch.randn(SHAPE["text_len"], 5120, generator=gen, dtype=torch.bfloat16),
        token_tags=packed["token_tags"],
        device=torch.device("cuda"),
    )
    branches.append(branch)
    videos.append(torch.randn(int(branch.img_pos.shape[0]), 96, generator=gen).cuda())
    audios.append(torch.randn(int(branch.audio_pos.shape[0]), 32, generator=gen).cuda())


def separate():
    for i, b in enumerate(branches):
        model(**b.forward_kwargs(video_rows=videos[i], audio_rows=audios[i], t_video=0.6, t_audio=0.5,
                                 imgvid_cond_timestep=0.999, audio_ref_cond_timestep=1.0))


def fused():
    n = len(branches)
    model(**minimax_h3_batched_forward_kwargs(
        branches=branches, video_rows=videos, audio_rows=audios,
        t_video=[0.6] * n, t_audio=[0.5] * n,
        imgvid_cond_timesteps=[0.999] * n, audio_ref_cond_timesteps=[1.0] * n))


def run(fn, label):
    from torch.profiler import ProfilerActivity, profile
    with torch.inference_mode():
        for _ in range(WARMUP):
            fn()
        torch.cuda.synchronize()
        start, end = torch.cuda.Event(enable_timing=True), torch.cuda.Event(enable_timing=True)
        start.record()
        for _ in range(ITERS):
            fn()
        end.record()
        torch.cuda.synchronize()
        with profile(activities=[ProfilerActivity.CPU, ProfilerActivity.CUDA]) as prof:
            for _ in range(ITERS):
                fn()
            torch.cuda.synchronize()
    ops, counts = defaultdict(float), defaultdict(int)
    for e in prof.key_averages():
        us = getattr(e, "self_device_time_total", 0.0) or 0.0
        if us > 0 and not e.key.startswith("aten::") and e.key != "Command Buffer Full":
            ops[e.key] += us / ITERS / 1000.0
            counts[e.key] += e.count // ITERS
    print(f"[{label}] wall {start.elapsed_time(end) / ITERS:.1f} ms, kernels {sum(ops.values()):.1f} ms")
    return ops, counts


sep_ops, sep_n = run(separate, "separate")
fus_ops, fus_n = run(fused, "fused")
for key in sorted(set(sep_ops) | set(fus_ops), key=lambda k: fus_ops.get(k, 0) - sep_ops.get(k, 0)):
    a, b = sep_ops.get(key, 0.0), fus_ops.get(key, 0.0)
    if max(a, b) >= 1.0:
        print(f"{b - a:9.1f}{a:10.1f}{b:10.1f}  {sep_n.get(key, 0):>5}->{fus_n.get(key, 0):<5} {key[:56]}")
```

</details>

### 1. Correctness

| Check | Result |
|---|---|
| `tests/diffusion/models/minimax_h3 -m 'core_model and cpu'` | 86 passed |
| Request mode vs step mode, same seed, real weights | **bitwise identical** video and audio (max abs diff `0.000e+00`) |
| Two concurrent requests | co-batched into one 65280-row forward; each output identical to running it alone |

### 2. Performance (BF16, TP2, 672x384, 209 frames, 30 steps, 4 requests at concurrency 4)

| Config | Wall time | Mean latency | Peak memory |
|---|---|---|---|
| request mode | **174.8 s** | 111.5 s | 72.4 GB |
| `--step-execution --max-num-seqs 1` | 179.0 s | 113.8 s | 72.4 GB |
| `--step-execution --max-num-seqs 4` | 182.1 s | 175.7 s | 78.3 GB |

Same picture with online `int8` (153.3 s / 158.4 s / 161.0 s), so quantization does not change the verdict.

Where a request's time goes (pipeline profiler, per request):

| Stage | Time | Share |
|---|---|---|
| `encode_prompt` | 0.37 s | 0.9% |
| `diffuse` (29 steps) | 38.34 s | 90.1% |
| `decode` (VAE) | 3.83 s | 9.0% |

### 3. Operator comparison (1 GPU, 3 requests x 16384 rows, one denoise step)

Two runs, to show the run-to-run spread:

| | 3 separate forwards | 1 fused forward | Ratio |
|---|---|---|---|
| GPU kernel time | 10510.7 / 10517.1 ms | 10333.4 / 10456.1 ms | 0.983 / 0.994 |
| Wall time | 7115.6 / 7122.7 ms | 7057.3 / 7113.1 ms | 0.992 / 0.999 |

The fused forward is never slower, but the margin is inside run-to-run noise: **fusing the forward is roughly time-neutral**, so the end-to-end regression comes from outside it.

The per-kernel split is stable across both runs:

| Kernel | Separate | Fused | Ratio | Launches |
|---|---|---|---|---|
| FlashAttention (varlen) | 3354.5 ms | 3232.3 ms | **0.96** | 156 -> 52 |
| Token-refiner GEMM (9 text rows) | 27.4 ms | 9.1 ms | **0.33** | 150 -> 50 |
| Main DiT GEMM | 2407.6 ms | 2443.5 ms | 1.01 - 1.03 | 600 -> 200 |
| RMSNorm | 244.7 ms | 266.2 ms | 1.09 - 1.10 | 630 -> 210 |
| `torch.cat` | 238.6 ms | 259.6 ms | 1.09 - 1.11 | — |

Fusing wins on kernel-launch amortization (attention, and the tiny `M=9` token-refiner GEMM) and loses on cache locality: the main GEMM is already compute-bound at 16384 rows, and the bandwidth-bound norm/cat/elementwise kernels get 5-11% worse on the larger working set. The two roughly cancel.

Per-step cost is therefore close to linear in batch size, which is why merging requests does not reduce total time:

| Requests per step | Time per step | vs batch=1 |
|---|---|---|
| 1 | 1.323 s | 1.00x |
| 3 | 3.945 s | 2.98x |
| 4 | 5.164 s | 3.90x |

Unlike LLM decoding, which is memory-bandwidth bound and batches almost for free, one H3 denoise step already has ~16k rows of dense math per request, so N requests cost N times the FLOPs.

### 4. Open-loop arrival experiment (4× H100)

**Configuration:** 4× H100 80GB, TP=2, USP=2, text-encoder TP=4, VAE patch parallelism=4, BF16, FlashAttention.  
**Workload:** 10 requests at a fixed 5-second arrival interval; 672×384, 4 seconds / 24 FPS, 20 inference steps.

| Configuration | Wall time | Throughput | Mean latency | Peak memory |
|---|---:|---:|---:|---:|
| request mode | 60.8 s | 9.88 req/min | 11.7 s | 56.9 GB |
| `--step-execution --max-num-seqs 4` | 67.9 s | 8.84 req/min | 23.6 s | 57.9 GB |

The step-mode run reached `peak compute overlap = 4`, confirming packed continuous batching. It does not improve H3 throughput: mean `diffuse` time grew from 5.27 s to 16.46 s, so throughput fell 10.5% and mean latency increased. H3's dense DiT cost scales close to linearly with packed request count; step execution is useful for scheduler-level control, not throughput.

```
=== comparison
request  (0 - 31s, each cell 0.5s;  '-' waiting, '=' computing)
#0  |>--==============                                                   |    7.5s
#1  |          >-----==============                                      |    8.4s
#2  |                     >-------=============                          |    9.2s
#3  |                                >--------==============             |   10.1s
#4  |                                           >-----------=============|   11.3s

step4  (0 - 35s, each cell 0.5s;  '-' waiting, '=' computing)
#0  |>-===================                                               |   10.8s
#1  |         >----=============================                         |   16.9s
#2  |                   >-----===========================                |   16.8s
#3  |                             >-------===========================    |   17.7s
#4  |                                      >-------======================|   15.1s

metric               |        request |          step4
------------------------------------------------------
completed / failed   |          5 / 0 |          5 / 0
wall time (s)        |           31.3 |           35.1
throughput (req/min) |           9.59 |           8.55
latency mean (s)     |            9.3 |           15.5
latency p50 (s)      |            9.2 |           16.8
latency max (s)      |           11.3 |           17.7
queue delay mean (s) |            3.5 |            3.0
peak outstanding     |              2 |              4
peak compute overlap |              2 |              3
peak memory (MB)     |          56866 |          57210

mean pipeline stage time (s), server-reported
decode               |           0.52 |           0.52
diffuse              |           5.24 |          11.15
encode_prompt        |           0.10 |           0.08
post_decode          |           0.00 |           0.52
prepare_encode       |           0.00 |           0.15
```


princepride:
@lishunyang12 @david6666666 PTAL

lishunyang12:
Local validation on PR head `f3c87809a9074b7bdc91880dcb41428119c81596`:

Environment: Python 3.12.3, PyTorch 2.11.0+cu130, vLLM 0.26.0.

| Scope | Command | Result |
|---|---|---:|
| MiniMax-H3 contracts + step execution | `pytest -sv tests/diffusion/models/minimax_h3 -m 'core_model and cpu'` | **125 passed, 5 deselected** |
| Shared diffusion scheduler + multiprocess concurrency | `pytest -q tests/diffusion/test_diffusion_scheduler.py tests/diffusion/test_multiproc_engine_concurrency.py` | **98 passed** |
| Async output + streaming-output integration | `pytest -q tests/diffusion/test_async_output_worker.py tests/diffusion/test_diffusion_streaming_output.py` | **15 passed** |

Total: **238 passed**. This covers request/step parity, packed request isolation, batched vs independent execution, progress propagation, abort-after-inflight-step, scheduler admission/retirement, worker-death handling, and async output behavior.

GPU note: I prepared a 2×B300 real-weight packed-batch parity run, but all 8 B300s on this machine are occupied by a pre-existing long-running DLO soak. I did not interrupt that experiment, so I am not claiming a fresh GPU E2E result in this comment.

princepride:
@lishunyang12 Seems I can enable DLO and CB in the same time.

lishunyang12:
Post-fix B300 E2E is now complete on commit `308bdb81` (1x B300, Ref2VA, BF16, `FLASH_ATTN`, `--step-execution --max-num-seqs 2`).

1. Unsupported multi-output:
   - before: HTTP 500
   - after: HTTP 400 in 55 ms
   - body: `MiniMax H3 step execution produces one output per request, got num_outputs_per_prompt=2`

2. Original invalid+valid concurrent reproducer:
   - invalid singular video-reference request: HTTP 400, `ref2va accepts at most 3 video references`
   - healthy image-reference sibling: HTTP 200
   - valid output: H.264 448x256 @ 24 fps + AAC stereo 32 kHz, duration 4.482 s
   - output SHA256: `f8e31dfbf3550061db4f908a086291af734136db74c5f00b9dbaeeb47bf8569c`
   - server log contains the isolated preparation error and the healthy request's subsequent denoise steps; there is no batch-level `Execution failed for diffusion requests` event.

The test server and exact child PIDs were stopped after validation; GPU memory returned to zero. Together with the green CI and CPU regressions above, the request-isolation issue is closed.

hsliuustc0106 (changes requested):
## Changes requested

I found three correctness issues in the continuous-batching path:

1. **[P1] Disable packed co-batching when Ring attention is active**

   [pipeline_minimax_h3.py#L2020](https://github.com/vllm-project/vllm-omni/blob/d5db537fd4b553340205c2d31a3ea6183eea0eec/vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py#L2020) gates `_packed_batch_supported()` only on `attn_backend.get_name()`. With `ring_degree > 1`, this can still be `FLASH_ATTN`, so `denoise_step()` packs multiple requests. Attention then dispatches through `RingParallelAttention`, whose ring kernels do not consume the packed `cu_seqlens` metadata. Requests can therefore attend across document boundaries and produce incorrect outputs. Please gate this path on the actual parallel strategy (at minimum reject `ring_degree > 1`) or add segmented Ring support.

2. **[P1] Synchronize preparation failures across DiT ranks**

   [diffusion_model_runner.py#L680](https://github.com/vllm-project/vllm-omni/blob/d5db537fd4b553340205c2d31a3ea6183eea0eec/vllm_omni/diffusion/worker/diffusion_model_runner.py#L680) catches `prepare_encode()` failures independently on each rank. H3 reference-video preparation is rank-0-only, so an invalid reference can make rank 0 catch and skip the request while other ranks proceed into encoder/VAE collectives. That can leave the process group out of sync and deadlock the worker. Please broadcast a per-request preparation success/error status and make every rank skip the same request before continuing.

3. **[P2] Reject request-scoped Cache-DiT in step execution**

   [pipeline_minimax_h3.py#L2046](https://github.com/vllm-project/vllm-omni/blob/d5db537fd4b553340205c2d31a3ea6183eea0eec/vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py#L2046) calls the request preparation path, which installs the H3 Cache-DiT profile whenever a request asks for `quality=high`, even when no startup cache backend is configured. Cache-DiT context and buffers live on the transformer rather than in `StepRequestState`, so interleaved or co-batched requests can reuse incompatible cache state. Please reject or disable Cache-DiT for step mode (or make its state request-local) and document the restriction for `quality=high`.

### Validation

- MiniMax-H3 core CPU suite: 133 passed, 5 deselected.
- Shared diffusion scheduler/step-pipeline tests: 92 passed.
- Changed-file diff check and Python compilation passed.
- I did not run a fresh full-weight distributed GPU E2E locally; the rank-synchronization issue above needs explicit multi-rank negative-path coverage.

princepride:
@lishunyang12 @hsliuustc0106 PTAL

docs/user_guide/diffusion/execution_modes.md
```diff
@@ -101,8 +101,16 @@ set `DIFFUSION_ATTENTION_BACKEND=TORCH_SDPA` or configure
 `--max-num-seqs >1`. See the
 [HunyuanImage-3.0 recipe](https://github.com/vllm-project/vllm-omni/blob/main/recipes/Tencent/HunyuanImage-3.0-Instruct.md)
 for its validated configuration. Helios supports single-request step execution only: use
-`--step-execution --max-num-seqs 1` for Helios. Consult the selected pipeline's
-documentation and source for the latest support status.
+`--step-execution --max-num-seqs 1` for Helios. MiniMax H3 supports step-wise
```
hsliuustc0106, line 104: why we need to add model specific desc here? I suggest to remove them and updates the docs

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
```diff
@@ -419,6 +499,58 @@ def _dit_rank_world() -> tuple[Any, int, int]:
     return group, dist.get_rank(group), dist.get_world_size(group)
 
 
+def _broadcast_rank0_exception(exc: Exception | None) -> None:
```
hsliuustc0106, line 510: @fhfuih PTAL, do we have a reference to tell the rules of adding this feature?

recipes/MiniMaxAI/MiniMax-H3.md
```diff
@@ -819,7 +859,12 @@ vllm serve "${MODEL_ROOT}/FL2VA" \
 - TeaCache is calibrated for FL2VA only; Ref2VA requests run uncached.
 - Combined serving requires sibling `FL2VA` and `Ref2VA` directories, loads
   both task-specific DiTs, and loads shared components once from `FL2VA`.
-- H3 currently executes one generation request per diffusion batch.
+- Request mode executes one generation request per diffusion batch. Use
```
hsliuustc0106, line 915: can we have a test case which DCB can help?

docs/user_guide/diffusion/execution_modes.md
princepride: Makes sense. That section put too much model-specific information into the general doc. We should keep execution_modes.md focused strictly on APIs, CLI, and core concepts.

recipes/MiniMaxAI/MiniMax-H3.md
princepride: I hope we can see gains on Blackwell or MI355, but I don't have them right now.

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
princepride: This is a hang bugfix, not a new feature.

H3's reference-video/audio prep runs only on rank 0. If rank 0 raises there, it skips the downstream dist.broadcast and other ranks hang forever. The helper syncs the exception right after the rank-0-only work so all ranks raise together.

H3 is currently the only pipeline with this pattern, so I kept it local. Happy to move it to vllm_omni/diffusion/distributed/ as a shared util if you prefer. @fhfuih please point me at any existing precedent.

fhfuih: Sounds good to keep it local. AFAIK indeed I am not aware of a common pattern to broadcast rank0 error during inference

vllm_omni/diffusion/forward_context.py
```diff
@@ -215,6 +215,17 @@ def set_forward_context_denoise_step_idx(step_idx: int | None) -> None:
...
+    """Set the normalized (descending, 1 -> 0) denoise timestep.
+
+    Timestep-gated attention features read this; pipelines that drive their own
+    denoise loop can publish it directly instead of going through
+    :meth:`DenoiseProgressMixin.record_denoise_step`.
+    """
+    if _forward_context is not None:
+        _forward_context.denoise_timestep = None if timestep is None else float(timestep)
+
+
```
fhfuih, line 228: Where is this `denoise_timestep` read from the forward context? I searched the diff part of this PR and could not find one.

In addition, this DI is primary for higher level objects like configs. Sounds a bit weird to record denoise timestep here. Any better place for it? (Like StepRequestState)?

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
hsliuustc0106: this is not an optimal choice? what's the rootcause of possible hang bug?

fhfuih: If to move it to `vllm_omni/diffusion/distributed/` module though, I think it can be moved to a function in `GroupCoordinator`. This is besides other similar interfaces like `broadcast_object / broadcast_object_list`

SamitHuang (commented):
good feature for h3! how about the generation precision diff vs non-batching or sample-wise batching?

princepride:
> good feature for h3! how about the generation precision diff vs non-batching or sample-wise batching?

Sure, I will do it later.

princepride:
```
### Precision A/B (4x H100, 672x384, 96 frames, 30 steps, 4 concurrent t2va requests, same seed)
| pair                                | video SSIM | video PSNR | audio cos | audio RMS |
|:--                                  |    ---:    |    ---:    |    ---:   |    ---:   |
| request-mode vs itself (rerun)      | 1.0000     | inf        | 1.000     | 1.000     |
| step-mode vs itself (rerun)         | 0.985      | 34+ dB     | 0.998     | 1.000     |
| **request-mode vs `--step-execution --max-num-seqs 4`** | **0.905** | **30.8 dB** | **0.976** | **0.994** |
Audio passes existing H3 thresholds with margin (≥ 0.80 cos, ∈ [0.5, 2.0] RMS).
Request mode is bit-identical on rerun; step mode is not, because per-step
packing composition depends on arrival timing — same character as LLM
continuous batching. Content is equivalent, bytes are not.
```

@SamitHuang

vllm_omni/diffusion/forward_context.py
princepride: > Where is this denoise_timestep read from the forward context?

TRTLLM Skip-Softmax gate. Consumer is pre-existing; this PR only adds the H3 producer so the "this pipeline does not publish denoise_timestep; staying dense" warning stops firing in step mode.

> In addition, this DI is primary for higher level objects like configs. Sounds a bit weird to record denoise timestep here. Any better place for it? (Like StepRequestState)?

it's per-DiT-forward, not per-request — N packed requests share one timestep, so putting it in `StepRequestState` means N duplicate copies. Reader is inside a fused attention kernel where `StepRequestState` isn't visible. Same pattern already exists in `forward_context.py` for `denoise_step_idx` (Hunyuan-Image-3) and `ref_latent` (Ming); this PR reuses it, doesn't introduce it.

princepride:
@SamitHuang @hsliuustc0106 PTAL

lishunyang12:
@hsliuustc0106 PTAL

yuanheng-zhao:
@princepride Did you encounter the error

```text
ImportError: cannot import name 'MistralToolCall' from 'vllm.tool_parsers.mistral_tool_parser'
```

when running 
```bash
vllm serve $MODEL_ROOT/FL2VA --omni --port 8091 --trust-remote-code \
  --tensor-parallel-size 2 --text-encoder-tp-size 2 --vae-use-tiling --enforce-eager \
  --diffusion-attention-backend FLASH_ATTN --step-execution --max-num-seqs 2
```

My env:
vllm: 0.27.0
vllm-omni: 1e4b7c01

hsliuustc0106:
```

[2026-08-14T10:33:53Z] FAILED tests/diffusion/cache/test_teacache_extractors.py::TestMiniMaxH3Extractor::test_modulated_input_shape - TypeError: MiniMaxH3DiTModel._embed() missing 1 required keyword-only argument: 'num_requests'
--
  | [2026-08-14T10:33:53Z] FAILED tests/diffusion/cache/test_teacache_extractors.py::TestMiniMaxH3Extractor::test_run_transformer_blocks_callable - TypeError: MiniMaxH3DiTModel._embed() missing 1 required keyword-only argument: 'num_requests'
  | [2026-08-14T10:33:53Z] FAILED tests/diffusion/cache/test_teacache_extractors.py::TestMiniMaxH3Extractor::test_postprocess_callable - TypeError: MiniMaxH3DiTModel._embed() missing 1 required keyword-only argument: 'num_requests'
  | [2026-08-14T10:33:53Z] FAILED tests/diffusion/cache/test_teacache_extractors.py::TestMiniMaxH3Extractor::test_postprocess_output_shape - TypeError: MiniMaxH3DiTModel._embed() missing 1 required keyword-only argument: 'num_requests'
  | [2026-08-14T10:33:53Z] FAILED tests/diffusion/cache/test_teacache_extractors.py::TestMiniMaxH3Extractor::test_extractor_matches_native_forward - TypeError: MiniMaxH3DiTModel._embed() missing 1 required keyword-only argument: 'num_requests'
  | [2026-08-14T10:33:53Z] FAILED tests/diffusion/cache/test_teacache_extractors.py::TestMiniMaxH3Extractor::test_run_transformer_blocks_forwards_packed_layout - TypeError: MiniMaxH3DiTModel._embed() missing 1 required keyword-only argument: 'num_requests'
  | [2026-08-14T10:33:53Z] FAILED tests/diffusion/cache/test_teacache_extractors.py::TestMiniMaxH3Extractor::test_teacache_reuses_h3_block_residual - TypeError: MiniMaxH3DiTModel._embed() missing 1 required keyword-only argument: 'num_requests'
  | [2026-08-14T10:33:53Z] = 7 failed, 2790 passed, 30 skipped, 367 deselected, 31 warnings in 468.59s (0:07:48) =
  | [2026-08-14T10:33:57Z] /usr/lib/python3.12/multiprocessing/resource_tracker.py:279: UserWarning: resource_tracker: There appear to be 1 leaked shared_memory objects to clean up at shutdown
  | [2026-08-14T10:33:57Z]   warnings.warn('resource_tracker: There appear to be %d '
  | [2026-08-14T10:34:00Z] 🚨 Error: The command exited with status 1


```

hsliuustc0106:
@Semmer2 PTAL

vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
```diff
@@ -1068,6 +1102,7 @@ def _embed(
         text_pos: torch.Tensor,
         refiner_cu_seqlens: torch.Tensor,
         refiner_max_seqlen: int,
+        num_requests: int,
```
hsliuustc0106: This required kwarg breaks the TeaCache H3 extractor: `extract_minimax_h3_context` calls `module._embed(...)` without `num_requests` (`vllm_omni/diffusion/cache/teacache/extractors.py:1323`), so request-mode H3 with `--cache-backend teacache` now raises `TypeError: _embed() missing 1 required keyword-only argument` at runtime. It's also what's failing the main Buildkite pipeline (7 failures in `TestMiniMaxH3Extractor`). Either read it in the extractor via `_psp_optional(psp, "num_requests", 1)` or give it a default of 1 here.

vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
```diff
@@ -45,6 +45,12 @@
 
 logger = init_logger(__name__)
 
+# Backends that attend by ``cu_seqlens`` and can therefore keep one document per
+# request in a co-batched packed sequence. Every other backend must run one
+# request per forward; the pipeline gates on this set before packing, and
+# ``_run_packed_attention`` enforces it.
+MINIMAX_H3_PACKED_BATCH_BACKENDS = frozenset({"FLASH_ATTN"})
```
hsliuustc0106: This name gate passes on NPU, where FLASH_ATTN is the platform default — but the NPU packed path only accepts the `[real, pad]` two-document contract: `_resolve_packed_seq_npu` declines when `cu_seqlens` has more than 3 bounds (flash_attn.py:513). A co-batched layout emits 2 bounds per request (>=5 total), so varlen declines, and since `num_requests > 1` sets `attn_mask=None` with `valid_kv_length=packed_total`, the mask-rebuild fallback produces an all-true mask and runs dense attention across request boundaries — silently. NPU runs H3 (the distilled 4-step schedule was just added for it), so I'd gate on whether the resolved backend actually supports multi-document varlen rather than the name alone.

vllm_omni/diffusion/worker/diffusion_model_runner.py
```diff
@@ -700,25 +725,65 @@ def _update_states(self, scheduler_output: DiffusionSchedulerOutput) -> tuple[li
...
+        self,
+        states: list[StepRequestState],
+        new_request_ids: list[str],
+    ) -> tuple[list[StepRequestState], InputBatch | None, list[RunnerOutput]]:
         # process new reqs
+        prepared_states: list[StepRequestState] = []
+        error_outputs: list[RunnerOutput] = []
         for state in states:
             if state.request_id in new_request_ids:
                 self._initialize_generator(state.sampling)
```
hsliuustc0106: An exception here (or in `clear_pipeline_stage_durations`) skips this rank's `_dit_any_rank_failed` while the peers enter the all-reduce and hang until the NCCL timeout. Move both inside the try?

recipes/MiniMaxAI/MiniMax-H3.md
```diff
@@ -355,6 +355,46 @@ No restart is needed: `task=fl2va` routes to `FL2VA/transformer`, while
...
+--step-execution --max-num-seqs 4
+```
+
+Co-batched requests are packed into a single sequence that keeps one attention
+document per request, so a batch costs one DiT forward. That packing requires
+`--diffusion-attention-backend FLASH_ATTN`; other backends fall back to one
+forward per request. `--max-num-seqs 1` keeps the conservative single-request
+step path. Cache acceleration (`--cache-backend`) is not available in step mode.
+
+!!! warning "Batching does not currently improve H3 throughput"
```
hsliuustc0106: All four prompts here are submitted at time 0 (`--max-concurrency 4`, `--request-rate` inf), which is the one arrival pattern where co-batching cannot win: a single 16k-row request already saturates the GPUs, and simultaneous submission just bunches completions. The two settings where step execution should show a benefit are unmeasured:

1. **Staggered arrivals.** The user-visible benefit of this feature is admission latency: in request mode a new request queues behind a whole in-flight generation (~45s at this size), while step mode admits at the next denoise-step boundary (~1.3s). `diffusion_benchmark_serving.py` already supports `--request-rate` (Poisson arrivals) — please run request mode vs `--step-execution --max-num-seqs 4` with e.g. one request every 10-20s and report mean/p95 latency, which then includes queueing. That is the experiment that justifies shipping the feature.

2. **Small requests.** "Batching does not improve H3 throughput" is drawn from 672x384/209-frame requests (16,384 packed rows). A short clip at lower resolution packs a few thousand rows and may not saturate the hardware — if co-batching helps there, scope this warning (and the execution-modes doc) to large requests rather than stating it unconditionally.

As it stands, every measured effect of enabling the feature is negative (5% slower, +6 GB peak memory), so a queueing benchmark is needed to show what the scheduler-level control actually buys.

hsliuustc0106 (commented):
@Semmer2 do we have to add a standalone step_batch.py to support this feature?

Semmer2:
> @Semmer2 do we have to add a standalone step_batch.py to support this feature?

Hi, @hsliuustc0106  the code itself is reasonable. I'm fine with it staying in a standalone file or being merged into `denoise_loop.py` — either works for me.
Actually, my concern is the name: `step_batch.py` looks like an implementation related to the engine's STEP_BATCH execution mode, while it actually packs N requests' layouts into one forward. The name like `batched_packing.py` would be clearer and avoid any possible confusion.

princepride:
> Actually, my concern is the name: step_batch.py looks like an implementation related to the engine's STEP_BATCH execution mode, while it actually packs N requests' layouts into one forward. The name like batched_packing.py would be clearer and avoid any possible confusion.

Sure, I can change it to `batched_packing.py` to avoid ambiguity.

princepride:
@Semmer2 @hsliuustc0106 PTAL

Semmer2:
> @Semmer2 @hsliuustc0106 PTAL

LGTM

hsliuustc0106:
CI failure not related and fixed already

Issue #5739, opened 2026-08-04
[Bug]: Weekly CI, tests/e2e/offline_inference/test_cosyvoice3_expansion.py::test_cosyvoice3_offline_reference_zero_shot[sync] , vllm_omni.entrypoints.omni_base.OmniEngineDeadError: Got unsupported ScalarType BFloat16


yenuo26:
### Your current environment

<details>
<summary>The output of <code>python collect_env.py</code></summary>

```text
CI env
```

</details>

### Your code version

<details>
<summary>The commit id or version of vllm</summary>

```text
0.26.0
```
</details>
<details>
<summary>The commit id or version of vllm-omni</summary>

```text
67c54777bb22e9e7e08fdf7c47a64f06b566fc47
```
</details>

### 🐛 Describe the bug

https://buildkite.com/vllm/vllm-omni/builds/13066/canvas?sid=019fbf5a-afe8-4301-9169-52b0d9ba4a7c&tab=output
```
[2026-08-01T22:27:10Z] WARNING 08-01 22:27:10 [input_processor.py:282] Passing raw prompts to InputProcessor is deprecated and will be removed in v0.18. You should instead pass the outputs of Renderer.render_cmpl() or Renderer.render_chat().
Actions for L1148
1148
[2026-08-01T22:27:12Z] ERROR 08-01 22:27:12 [orchestrator.py:551] [Orchestrator] Fatal error in orchestrator tasks
Actions for L1149
1149
[2026-08-01T22:27:12Z] ERROR 08-01 22:27:12 [orchestrator.py:551] Traceback (most recent call last):
Actions for L1150
1150
[2026-08-01T22:27:12Z] ERROR 08-01 22:27:12 [orchestrator.py:551]   File "/workdir/vllm_omni/engine/orchestrator.py", line 531, in run
Actions for L1151
1151
[2026-08-01T22:27:12Z] ERROR 08-01 22:27:12 [orchestrator.py:551]     await asyncio.gather(*tasks)
Actions for L1152
1152
[2026-08-01T22:27:12Z] ERROR 08-01 22:27:12 [orchestrator.py:551]   File "/workdir/vllm_omni/engine/orchestrator.py", line 891, in _orchestration_output_handler
Actions for L1153
1153
[2026-08-01T22:27:12Z] ERROR 08-01 22:27:12 [orchestrator.py:551]     await self._orchestration_loop()
Actions for L1154
1154
[...]

PR #5814, opened 2026-08-05
[Bugfix] Replace hardcoded Vivian speaker default with config-ordered default

khairulkabir1661:
## Summary

CustomVoice TTS requests without a voice parameter hardcoded "Vivian" as the default speaker. Models without Vivian in their config (e.g., Qwen3-Omni) crash with ValueError: Unsupported speaker: vivian. This PR selects the first speaker from the model's spk_id config as the default, with warnings when no default speaker is available.

- `load_supported_speakers()` returns `list[str]` to preserve config order
- Add `default_speaker` field to `TTSCapabilities`, set from first speaker at adapter init with INFO log
- Replace hardcoded "Vivian" in Qwen3-TTS adapter's `_build_tts_params` with `default_speaker`
- Add early validation: CustomVoice + no voice + no default_speaker returns 400 instead of downstream error
- Update all adapter `_load_supported_speakers()` overrides to return `list[str]`
- Fix `test_qwen3_tts_build_constructs_prepared_request` to set capabilities with a configured speaker

## Test plan

- [x] `test_default_speaker_is_first_in_config` — first speaker from config resolves as default
- [x] `test_qwen3_custom_voice_uses_default_speaker` — calls Qwen3 adapter's `_build_tts_params` with/without default speaker
- [x] `test_qwen3_tts_build_constructs_prepared_request` — updated to use config-ordered default
- [x] Existing `test_no_tts_detection` updated for list return type

### DIFF
--- a/vllm_omni/entrypoints/openai/serving_speech.py
+++ b/vllm_omni/entrypoints/openai/serving_speech.py
@@ -2495,7 +2495,13 @@ class OmniOpenAIServingSpeech(OpenAIServing, AudioMixin):
                 logger.info("Using precomputed Qwen3-TTS custom voice profile: %s (mode=%s)", voice_lower, mode)
 
         elif params["task_type"][0] == "CustomVoice":
-            params["speaker"] = ["Vivian"]  # Default for CustomVoice
+            if self._adapter is not None and self._adapter.capabilities.default_speaker:
+                params["speaker"] = [self._adapter.capabilities.default_speaker]
+            else:
+                raise ValueError(
+                    "CustomVoice task requires a speaker but no speakers are "
+                    "configured in this model's spk_id config."
+                )
 
         # Instructions for style/emotion control
         if request.instructions is not None:
--- a/vllm_omni/entrypoints/openai/tts_adapters/base.py
+++ b/vllm_omni/entrypoints/openai/tts_adapters/base.py
@@ -124,6 +124,7 @@ class TTSCapabilities:
     supported_speakers: frozenset[str] = frozenset()
     supported_languages: frozenset[str] = DEFAULT_TTS_LANGUAGES
     codec_frame_rate: float | None = None
+    default_speaker: str | None = None
 
 
 class TTSModelAdapter(ABC):
@@ -240,19 +241,25 @@ class TTSModelAdapter(ABC):
         return None
 
     def load_capabilities(self) -> TTSCapabilities:
+        speakers_list = self._load_supported_speakers()
+        default_speaker = speakers_list[0] if speakers_list else None
+        if default_speaker:
+            logger.info("Default speaker for CustomVoice task: %s", default_speaker)
         self.capabilities = TTSCapabilities(
             precomputed_speakers=self._load_precomputed_speakers(),
-            supported_speakers=frozenset(self._load_supported_speakers()),
+            supported_speakers=frozenset(speakers_list),
             supported_languages=self._load_supported_languages(),
             codec_frame_rate=self._load_codec_frame_rate(),
+            default_speaker=default_speaker,
         )
         return self.capabilities
 
     def _load_precomputed_speakers(self) -> dict[str, dict[str, Any]]:
         return {}
 
-    def _load_supported_speakers(self) -> set[str]:
+    def _load_supported_speakers(self) -> list[str]:
         # Preserve the legacy default path, which reads talker_config.
+        # Returns a list to preserve config order.
         return load_supported_speakers(self.ctx.engine_client)
 
     def _load_supported_languages(self) -> frozenset[str]:
--- a/vllm_omni/entrypoints/openai/tts_adapters/capabilities.py
+++ b/vllm_omni/entrypoints/openai/tts_adapters/capabilities.py
@@ -16,23 +16,26 @@ from vllm_omni.utils.speaker_cache import iter_custom_voice_profiles, load_valid
 logger = init_logger(__name__)
 
 
-def load_supported_speakers(engine_client: Any, config: Any = None) -> set[str]:
-    """Extract supported speaker names from a model configuration."""
+def load_supported_speakers(engine_client: Any, config: Any = None) -> list[str]:
+    """Extract supported speaker names from a model configuration.
+
+    Returns a list to preserve the order from the config file.
+    """
     try:
         if config is None:
             config = getattr(engine_client.model_config.hf_config, "talker_config", None)
             if config is None:
-                return set()
+                return []
 
         for attr_name in ("spk_id", "speaker_id"):
             speakers = config.get(attr_name) if isinstance(config, dict) else getattr(config, attr_name, None)
             if speakers and isinstance(speakers, dict):
-                return {speaker.lower() for speaker in speakers}
+                return [speaker.lower() for speaker in speakers]
 
         logger.warning("No speakers found in config (checked spk_id and speaker_id)")
     except Exception as exc:
         logger.warning("Could not load speakers from model config: %s", exc)
-    return set()
+    return []
 
 
 def load_codec_frame_rate(engine_client: Any) -> float | None:
--- a/vllm_omni/entrypoints/openai/tts_adapters/glm_tts.py
+++ b/vllm_omni/entrypoints/openai/tts_adapters/glm_tts.py
@@ -49,8 +49,8 @@ class GlmTTSAdapter(ARTTSAdapter):
         prompt = await self.ctx.server._build_glm_tts_prompt(request, has_inline_ref_audio=has_inline_ref_audio)
         return PreparedRequest(prompt=prompt, tts_params={}, model_type="glm_tts")
 
-    def _load_supported_speakers(self) -> set[str]:
-        return set()
+    def _load_supported_speakers(self) -> list[str]:
+        return []
 
     def apply_sampling_overrides(
         self,
--- a/vllm_omni/entrypoints/openai/tts_adapters/ming_flash_omni_tts.py
+++ b/vllm_omni/entrypoints/openai/tts_adapters/ming_flash_omni_tts.py
@@ -65,6 +65,6 @@ class MingFlashOmniTTSAdapter(ARTTSAdapter):
         prompt = server._build_ming_flash_omni_prompt(request)
         return PreparedRequest(prompt=prompt, tts_params={}, model_type="ming_flash_omni_tts")
 
-    def _load_supported_speakers(self) -> set[str]:
+    def _load_supported_speakers(self) -> list[str]:
         # Speaker selection is driven by caption JSON rather than a static table.
-        return set()
+        return []
--- a/vllm_omni/entrypoints/openai/tts_adapters/ming_tts.py
+++ b/vllm_omni/entrypoints/openai/tts_adapters/ming_tts.py
@@ -217,8 +217,8 @@ class MingTTSAdapter(ARTTSAdapter):
         except Exception as e:
             logger.warning(f"Failed to derive Ming codec frame rate from hf_config: {e}")
 
-    def _load_supported_speakers(self) -> set[str]:
-        return set()
+    def _load_supported_speakers(self) -> list[str]:
+        return []
 
     def _load_codec_frame_rate(self) -> float | None:
         codec_frame_rate = self._load_ming_tts_codec_frame_rate()
--- a/vllm_omni/entrypoints/openai/tts_adapters/voxcpm2.py
+++ b/vllm_omni/entrypoints/openai/tts_adapters/voxcpm2.py
@@ -115,8 +115,8 @@ class VoxCPM2Adapter(ARTTSAdapter):
             validate_profile=validate_voxcpm2_profile,
         )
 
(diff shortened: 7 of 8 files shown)


vllm_omni/entrypoints/openai/serving_speech.py
```diff
@@ -2608,7 +2606,13 @@ def _build_tts_params(self, request: OpenAICreateSpeechRequest) -> dict[str, Any
                 logger.info("Using precomputed Qwen3-TTS custom voice profile: %s (mode=%s)", voice_lower, mode)
 
         elif params["task_type"][0] == "CustomVoice":
-            params["speaker"] = ["Vivian"]  # Default for CustomVoice
+            if self.supported_speakers:
+                params["speaker"] = [next(iter(sorted(self.supported_speakers)))]
```
alex-jw-brooks, line 2610: Could you change this to set the `default_supported_speaker` as an attribute at init time with an `info` level log? I don't think there is a very obvious default for models that have a speaker bank, so logging it would be helpful I think. 

I think using the first speaker in the config for the speaker bank as the default would also be better, rather than the first alphabetically, which also probably makes more sense since this is a set

khairulkabir1661: Fixed

vllm_omni/entrypoints/openai/tts_adapters/base.py
```diff
@@ -240,19 +241,25 @@ def validate_tts_embedding_dim(self, emb_dim: int) -> str | None:
         return None
 
     def load_capabilities(self) -> TTSCapabilities:
+        speakers_list = self._load_supported_speakers()
+        default_speaker = speakers_list[0] if speakers_list else None
+        if default_speaker:
+            logger.info("Default speaker for CustomVoice task: %s", default_speaker)
```
alex-jw-brooks, line 250: Can you fix the logger initialization?

vllm_omni/entrypoints/openai/serving_speech.py
```diff
@@ -2495,7 +2495,13 @@ def _build_tts_params(self, request: OpenAICreateSpeechRequest) -> dict[str, Any
                 logger.info("Using precomputed Qwen3-TTS custom voice profile: %s (mode=%s)", voice_lower, mode)
 
         elif params["task_type"][0] == "CustomVoice":
-            params["speaker"] = ["Vivian"]  # Default for CustomVoice
+            if self._adapter is not None and self._adapter.capabilities.default_speaker:
```
alex-jw-brooks: After looking a bit closer, I think this is actually used on the qwen3 tts path and just a (very) badly named helper in the entrypoint, so we should always have an adapter in this case, but I think we should be careful to ensure this is checked correctly and not accidentally broken in future refactoring.

Can you refactor this to instead:

- `warn` if we have no adapter, since it should not happen in this case
- `warn` (different message) if we DO have an adapter but no `default_speaker`
- Check to see the behavior if we just set this to `None`?

alex-jw-brooks: I think  the no speaker case is also handled in a few places (i.e., here, in the qwen3tts adapter, and also in the model itself for creating the prompt embeds). Can you please add some unit tests for ensuring:

- That if we have a config with different `spk_ids`, the first value resolves to the default correctly
- If we have a config with no `spk_ids`, this actually catches correctly?

vllm_omni/entrypoints/openai/tts_adapters/base.py
khairulkabir1661: Fixed

vllm_omni/entrypoints/openai/serving_speech.py
khairulkabir1661: Done as suggested

khairulkabir1661: yes, did

linyueqian (changes requested):
Reviewed at `7ee5af48`. Selecting an ordered model speaker instead of hardcoding Vivian is the right direction, but this head breaks an existing return-type test and the new serving regression only copies the implementation into the test. The current pre-commit and DCO checks also fail. I am requesting changes before the behavior can be validated.

The PR metadata needs correction too. Commit `c8c6a815` says `Fixes #5739`, but #5739 is an unrelated CosyVoice3 BF16 serialization failure, and the PR description still promises a missing-voice 400 response while this head selects a default speaker or defers failure downstream.

Validation: static review of the ten-file patch, all speaker-capability callers, existing speech tests, the current-main merge tree, live checks, and the linked issue. The branch merges cleanly. No fork code was executed locally.

khairulkabir1661:
Hi @linyueqian, thanks for the review, can you please have a look again? thanks a lot.

vllm_omni/entrypoints/openai/tts_adapters/qwen3_tts.py
```diff
@@ -332,7 +332,17 @@ def _build_tts_params(self, request: "OpenAICreateSpeechRequest") -> dict[str, A
                 logger.info("Using precomputed Qwen3-TTS custom voice profile: %s (mode=%s)", voice_lower, mode)
 
         elif params["task_type"][0] == "CustomVoice":
-            params["speaker"] = ["Vivian"]  # Default for CustomVoice
+            if self.capabilities.default_speaker is not None:
```
linyueqian, line 340: [important] This branch changes the answer for an adapter whose capabilities were never loaded, and one existing test relies on the old answer: `test_qwen3_tts_build_constructs_prepared_request` (same test file, around line 253) constructs `Qwen3TTSAdapter` over a bare `SimpleNamespace` engine client, never calls `load_capabilities()`, so `self.capabilities` is the default `TTSCapabilities()` with `default_speaker=None`, and then asserts `prepared.tts_params["speaker"] == ["Vivian"]`. With this change the branch below runs instead and the assertion fails. Please update that test to load or set capabilities with a configured speaker and assert the config-ordered default (note it will be lower-cased, so `["vivian"]`, which is what the prompt builder expects), or assert `None` if you want it to pin the no-config behaviour. The lane I fired at this head will show the failure in the Entrypoints step.

vllm_omni/entrypoints/openai/tts_adapters/qwen3_tts.py
```diff
@@ -332,7 +332,17 @@ def _build_tts_params(self, request: "OpenAICreateSpeechRequest") -> dict[str, A
...
+                params["speaker"] = [self.capabilities.default_speaker]
+            else:
+                available = sorted(self.capabilities.supported_speakers) if self.capabilities.supported_speakers else []
+                logger.warning(
+                    "CustomVoice task requires a speaker but no default speaker is "
+                    "configured in the model's spk_id config. Available speakers: %s. "
+                    "Request will fail downstream with error from model.",
+                    ", ".join(available) if available else "none",
+                )
+                params["speaker"] = None
```
linyueqian, line 350: [suggestion] Sending `speaker=None` here means a CustomVoice request with no `voice` on a model without an ordered default reaches the talker and fails there, which the client sees as a server error rather than the request error it is. `validate()` already has the CustomVoice speaker block (it rejects when no speakers are available at all and when `voice` is unknown); adding `if request.voice is None and self.capabilities.default_speaker is None: return "CustomVoice requires 'voice' on this model"` there turns this into a 400 and lets you drop the warning. The case is rare in practice (every shipped Qwen3-TTS CustomVoice checkpoint has `spk_id`), so not blocking.

linyueqian (changes requested):
Re-reviewed at `5bf410f4`. First, an apology: my 09-15 review was posted by a tooling bug with its four inline items missing, so you only saw the summary. Most of what I asked for is nonetheless in this head. The import grouping is fixed and pre-commit and DCO are green; the two new tests drive production code (`TTSModelAdapter.load_capabilities` for the ordered default and `Qwen3TTSAdapter._build_tts_params` for the CustomVoice branch) instead of copying it; the misattributed `Fixes #5739` and the promised 400 are gone from the description; and the branch is rebased onto yesterday's `main` with no conflicts. The ordered default is right too: `load_supported_speakers` lower-cases names and `prompt_embeds_builder.py` lower-cases both the request speaker and the `spk_id` keys before the lookup, so `vivian` resolves exactly as `Vivian` did.

Two items remain inline. The `[important]` one is a deterministic test failure this head introduces: `test_qwen3_tts_build_constructs_prepared_request` in the same test file still asserts `tts_params["speaker"] == ["Vivian"]`, and with an adapter whose capabilities were never loaded the new branch now yields `None`, so the Entrypoints test step will go red at this head; I added `ready` so the lane shows it, and the fix is a two-line change to that test. The `[suggestion]` is the request-error path I asked for last time: when a CustomVoice request has no `voice` and the model has no ordered default, the adapter now sends `speaker=None` and logs that the request will fail downstream, which surfaces as a talker error rather than a 400; `validate()` already has the CustomVoice speaker block, so rejecting that case there is small.

Validation: static read of the PR-only diff at this head against `origin/main` and of the tests it touches; no PR code executed. Once the existing test is updated and the lane is green this is ready to approve.

vllm_omni/entrypoints/openai/tts_adapters/qwen3_tts.py
khairulkabir1661: addressed

vllm_omni/entrypoints/openai/tts_adapters/qwen3_tts.py
khairulkabir1661: addressed

khairulkabir1661:
@linyueqian, please review now.

khairulkabir1661:
@linyueqian can you please approve if everything looks good?

vllm_omni/entrypoints/openai/tts_adapters/base.py
```diff
@@ -315,19 +320,25 @@ def validate_tts_embedding_dim(self, emb_dim: int) -> str | None:
         return None
 
     def load_capabilities(self) -> TTSCapabilities:
+        speakers_list = self._load_supported_speakers()
+        default_speaker = speakers_list[0] if speakers_list else None
```
linyueqian, line 324: [blocking] This indexes whatever the subclass returns, and `GepardAdapter._load_supported_speakers` (and every override on current main) returns a `set`, so Gepard deployments fail at server construction with `TypeError: 'set' object is not subscriptable` (lane 15915, five `test_gepard_tts_adapter.py` errors). Either convert the overrides to ordered lists, or keep this robust to any iterable: normalise once with `speakers_list = list(self._load_supported_speakers())` and document that only list-returning overrides carry a meaningful default order.

linyueqian (changes requested):
The delta fixes both earlier findings: the omitted-voice default now comes from the checkpoint's own speaker order instead of assuming Vivian, and the validation rejects a missing default before generation. What blocks is a regression the general lane caught at `e3db6bf6` (build 15915): `load_capabilities` now indexes `speakers_list[0]`, and `_load_supported_speakers` is retyped to return a list, but the subclass overrides were not updated. `GepardAdapter._load_supported_speakers` still returns `{_DEFAULT_VOICE}`, a set, so constructing the speech server for a Gepard deployment raises `TypeError: 'set' object is not subscriptable`, and five existing tests in `tests/entrypoints/openai_api/test_gepard_tts_adapter.py` error out (3004 pass around them). Breeze also returns a set but an empty one, so the guard skips the index there; every override on current main returns a set, so anything added after a rebase will hit the same path. The fix is inline. Two smaller notes from this pass: please add a regression test for an omitted voice with `default_speaker=None` and a non-empty speaker set (for example uploaded speakers only), asserting rejection before generation and acceptance once a default is configured; and the description names Qwen3-Omni, which `/v1/audio/speech` rejects rather than routing through this adapter, so a Qwen3-TTS CustomVoice checkpoint without Vivian is the right example. Static read at `e3db6bf6` against merge-base `0570f7a0`, plus the lane log; fork head, no PR code executed locally. This branch predates the vLLM 0.30 rebase on main, so please rebase with the fix.

vllm_omni/entrypoints/openai/tts_adapters/base.py
khairulkabir1661: done, thanks.

khairulkabir1661:
@linyueqian thanks a lot, now have a look.

linyueqian (approved):
Both items from the last round are fixed at `9c05b4e4`. `load_capabilities` now normalises the override result with `list(...)` before taking the first entry, and the overrides that return a fixed speaker set were moved to ordered lists, so the Gepard server no longer fails at construction. The new `test_qwen3_validate_rejects_no_voice_no_default_then_accepts` pins the missing-default path: a request with no voice and no configured default is rejected before generation and accepted once a default exists. `test_default_speaker_is_first_in_config` covers the config-order default. The branch is rebased onto `3f473a35`, which includes the vLLM 0.30 rebase. Approving. The description still names Qwen3-Omni as an example model, and `/v1/audio/speech` rejects that model; a Qwen3-TTS CustomVoice checkpoint without Vivian would be the accurate example. Static read of the ten files at `9c05b4e4` against merge-base `3f473a35`; fork head, no PR code executed.

PR #5817, opened 2026-08-05
[Model] MiniMax-H3 reuse packed attention masks across layers

LynnColeArt:
## Purpose

MiniMax-H3's direct-attention path can construct the same packed padding mask inside every attention layer. A standard forward traverses 50 DiT blocks plus 2 token-refiner blocks, so those immutable masks are needlessly reallocated across each block stack.

This change derives each request mask once and passes it through the existing attention call chain. It is rebased onto the current prefix-slicing, Ring, and video-layout contracts from `main`:

- prefix-slicing backends continue to receive no padding mask;
- direct attention retains a local fallback for standalone callers;
- `valid_kv_length`, packed totals, `VideoTokenLayout`, Ring/Ulysses behavior, and per-call `AttentionMetadata` remain intact.

The attention math, packed-sequence boundaries, and mask values are unchanged. No end-to-end latency claim is made.

## Test plan

- exercise precomputed-mask propagation and the standalone fallback;
- exercise prefix slicing, Ring layout, video-token layout, and fused-RoPE coverage from current `main`;

**vLLM Version:** 0.26.0

## Test result

```text
81 passed, 1 skipped, 2 deselected, 15 warnings
Ruff 0.14.10 check: passed
Ruff 0.14.10 format: passed
```

### DIFF
--- a/vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
+++ b/vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
@@ -108,6 +108,18 @@ MINIMAX_H3_FP32_BUFFER_NAMES = frozenset({"rope.inv_freq"})
 MINIMAX_H3_ADALN_MODALITY_NUM = 3
 
 
+def _build_packed_attention_mask(
+    *,
+    used_seqlen: int,
+    packed_seqlen: int,
+    device: torch.device,
+) -> torch.Tensor | None:
+    """Build H3's padding mask from host-side packed-layout lengths."""
+    if used_seqlen >= packed_seqlen:
+        return None
+    return torch.arange(packed_seqlen, device=device)[None] < used_seqlen
+
+
 def _required_kwarg(kwargs: dict[str, Any], key: str) -> Any:
     if key not in kwargs or kwargs[key] is None:
         raise ValueError(f"MiniMaxH3DiTModel.forward requires kwarg {key!r}")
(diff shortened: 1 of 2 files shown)


vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
```diff
@@ -399,13 +412,16 @@ def _run_packed_attention(
...
         used = min(max_seqlen, packed_total)
-        attn_mask = None
         # Ring attention can dispatch to a different implementation from the
         # configured backend, so this no-mask fast path is local-only.
         prefix_slice = (
             not getattr(self.attention, "use_ring", False) and self.attention.attn_backend.supports_prefix_kv_slicing
         )
-        if used < packed_total and not prefix_slice:
+        if prefix_slice:
+            attn_mask = None
```
david6666666, line 421: Good - dropping the precomputed mask for prefix-slicing backends keeps their behavior unchanged, and the fallback below still covers direct callers. Mask values match the old per-layer build (`arange(packed) < used`), so reusing one request mask across all blocks is safe.

david6666666:
Verified on a B300 single GPU (vllm 0.26.0, torch 2.11.0+cu130, FLASH_ATTN, eager, t2va 5s @1344x768 50 steps, seed 1101) vs the d219d93b baseline: output is bit-exact (max frame diff 0, audio identical). Perf: ~159.9s vs baseline ~157.3s (+1.6%, within run-to-run noise) - no measurable e2e gain, which matches the PR making no latency claim; the mask build is a small tensor per request so the win is mainly fewer kernel launches in compiled mode.

PR #5818, opened 2026-08-05
[Model] MiniMax-H3 use fused SwiGLU activation

LynnColeArt:
## Purpose

MiniMax-H3's MLP already produces a merged `[gate, up]` projection, but its forward path chunks that tensor and executes `silu(gate) * up` as separate eager operations. That materializes an additional full-size activation and adds avoidable memory traffic in every MLP layer.

This change uses vLLM's existing `SiluAndMul` operator. The checkpoint mapping and `[gate, up]` shard order are unchanged.

## Test plan

- verify the H3 MLP constructs vLLM's fused SwiGLU operator;
- retain current-main fused-RoPE coverage;

**vLLM Version:** 0.26.0

## Test result

```text
80 passed, 1 skipped, 2 deselected, 15 warnings
Ruff 0.14.10 check: passed
Ruff 0.14.10 format: passed
```

## Isolated DGX Spark microbenchmark

BF16 production activation shape `[6016, 28672]`, 10 warmups, then 15 samples of 20 CUDA-event iterations:

| Activation | Median | Range |
| --- | ---: | ---: |
| Eager `silu(gate) * up` | 4.6124 ms | 3.7433–4.7752 ms |
| vLLM `SiluAndMul` | 2.4649 ms | 2.1885–2.5184 ms |

Outputs were bit-exact across BF16 values spanning `[-4, 4]`. The fused operator was 1.871x faster in this isolated kernel benchmark; no end-to-end generation claim is made.

### DIFF
--- a/tests/diffusion/models/minimax_h3/test_minimax_h3_parallel.py
+++ b/tests/diffusion/models/minimax_h3/test_minimax_h3_parallel.py
@@ -88,6 +88,27 @@ def test_h3_fused_rope_matches_reference_and_preserves_unrotated_dims():
     torch.testing.assert_close(actual[..., 96:], x[..., 96:], atol=0, rtol=0)
 
 
+def test_mlp_uses_fused_swiglu(init_fake_tp_group):
+    del init_fake_tp_group
+    from vllm.model_executor.layers.activation import SiluAndMul
+
+    from vllm_omni.diffusion.models.minimax_h3.minimax_h3_transformer import (
+        MiniMaxH3DiTArchConfig,
+        MiniMaxH3MLP,
+    )
+
+    mlp = MiniMaxH3MLP(
+        MiniMaxH3DiTArchConfig(
+            hidden_size=4,
+            ffn_hidden_size=8,
+        ),
+        quant_config=None,
+        prefix="blocks.0.mlp",
+    )
+
+    assert isinstance(mlp.act_fn, SiluAndMul)
+
+
 @pytest.mark.parametrize(
     ("tp_size", "message"),
     [
--- a/vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
+++ b/vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
@@ -18,6 +18,7 @@ import torch.nn as nn
 from cache_dit import ForwardPattern
 from vllm.distributed import get_tensor_model_parallel_world_size
 from vllm.logger import init_logger
+from vllm.model_executor.layers.activation import SiluAndMul
 from vllm.model_executor.layers.linear import (
     ColumnParallelLinear,
     MergedColumnParallelLinear,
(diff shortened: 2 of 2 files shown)


vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
```diff
@@ -513,8 +513,7 @@ def __init__(
 
     def forward(self, x: torch.Tensor) -> torch.Tensor:
         hidden, _ = self.fc1(x)
-        gate, up = hidden.chunk(2, dim=-1)
-        hidden = nn.functional.silu(gate) * up
+        hidden = self.act_fn(hidden)
```
david6666666, line 516: `SiluAndMul` computes `silu(x[..., :d]) * x[..., d:]`, which is exactly the old `chunk(2)` + `silu(gate) * up`; since `MergedColumnParallelLinear` still emits the `[gate, up]` shard order, TP sharding and numerics are preserved. The bit-exact isolated bench looks right.

david6666666, line 516: Heads-up on the bit-exact claim: on a B300 single-GPU run (vllm 0.26.0, torch 2.11.0+cu130) the fused op is not bit-exact with the old eager path. Direct op check with inputs clamped to [-4, 4]: ~26.6% of bf16 elements differ, max_abs 0.0625 (1 bf16 ulp); the fused result equals fp32-computed-then-rounded-to-bf16 exactly, while eager `F.silu(gate) * up` rounds silu to bf16 first. E2E this amplifies through the 50-step denoise - same seed produces a visibly different video vs the `d219d93b` baseline (2-step run already differs). Math is correct (verified not reversed, closer to fp32), so this is a precision change, not a formula bug - but the `Outputs were bit-exact` claim doesn't hold here, and any seed-reproducibility test against the eager baseline will fail. Worth either qualifying the claim (platform/kernel dependent) or keeping the eager path if bit-exact output matters.

david6666666:
Verified on a B300 single GPU (same config as above) vs the d219d93b baseline: output is NOT bit-exact (max frame diff 255, mean ~9.3; a 2-step run already differs) - see the inline comment for the op-level numbers. Perf: ~155.1s vs ~157.3s (-1.5%, diffuse -3.4s). Math is correct and the fused op is closer to fp32, but the 'bit-exact' claim in the body doesn't hold on this platform; if seed-reproducible output vs the eager baseline matters, this needs qualification or the eager path kept.

tests/diffusion/models/minimax_h3/test_minimax_h3_parallel.py
```diff
@@ -88,6 +88,27 @@ def test_h3_fused_rope_matches_reference_and_preserves_unrotated_dims():
...
+    mlp = MiniMaxH3MLP(
+        MiniMaxH3DiTArchConfig(
+            hidden_size=4,
+            ffn_hidden_size=8,
+        ),
+        quant_config=None,
+        prefix="blocks.0.mlp",
+    )
+
+    assert isinstance(mlp.act_fn, SiluAndMul)
```
hsliuustc0106, line 109: Non-blocking: please add a numerical assertion here against `F.silu(gate) * up` with an explicit BF16 tolerance. This type-only check would still pass if the fused slicing/order or math changed, while the observed platform-dependent rounding difference means exact equality would be too strict.

hsliuustc0106:
fix conflicts

PR #5819, opened 2026-08-05
[Model] MiniMax-H3 precompute static conditioning

LynnColeArt:
## Purpose

MiniMax-H3's prompt projection, two-layer token refiner, and 3D RoPE inputs are invariant across a generation request, but `main` recomputes them inside every denoise forward.

This change prepares those tensors once after the packed layout is known and carries them through the denoise branch in a frozen, request-scoped `MiniMaxH3StaticConditioning` value. There is no model-global cache, so concurrent requests cannot share or overwrite conditioning state. The raw-input fallback remains available for direct transformer and denoise-branch callers.

Preparation calls the model through its normal forward path, preserving CPU-offload hooks. The implementation is rebased onto current `VideoTokenLayout` handling.

## Test plan

- verify prepared and fallback conditioning are numerically equivalent;
- verify model forward hooks run during preparation;
- verify repeated denoise forwards reuse request-scoped projection, refiner, and RoPE tensors;
- retain current-main video-layout and fused-RoPE coverage;

**vLLM Version:** 0.26.0

## Test result

```text
81 passed, 1 skipped, 2 deselected, 15 warnings
Ruff 0.14.10 check: passed
Ruff 0.14.10 format: passed
```

For a 19-forward short schedule, prompt projection, token refinement, and RoPE generation each change from 19 calls to 1 (94.74% fewer). Tests enforce the call counts and numerical equivalence. No full-generation latency claim is made.

### DIFF
--- a/vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
+++ b/vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
@@ -106,6 +106,16 @@ MINIMAX_H3_FP32_BUFFER_NAMES = frozenset({"rope.inv_freq"})
 # video/text/audio tokens (padding is clamped to 0 before the embedding
 # lookup and masked out afterwards).
 MINIMAX_H3_ADALN_MODALITY_NUM = 3
+MINIMAX_H3_STATIC_CONDITIONING_KWARG = "static_conditioning"
+_MINIMAX_H3_PREPARE_STATIC_CONDITIONING_KWARG = "prepare_static_conditioning"
+
+
+@dataclass(frozen=True)
+class MiniMaxH3StaticConditioning:
+    """Timestep-invariant conditioning prepared for one generation request."""
+
+    refined_prompt_embeds: torch.Tensor
+    rope_freqs: torch.Tensor
 
 
 def _required_kwarg(kwargs: dict[str, Any], key: str) -> Any:
@@ -135,6 +145,8 @@ _FORWARD_SUPPORTED_KWARGS = frozenset(
         "packed_seq_params",
         "refiner_packed_seq_params",
         "video_token_layout",
+        MINIMAX_H3_STATIC_CONDITIONING_KWARG,
+        _MINIMAX_H3_PREPARE_STATIC_CONDITIONING_KWARG,
     }
 )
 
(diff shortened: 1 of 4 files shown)


vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
```diff
@@ -1032,18 +1044,117 @@ def _psp_field(psp: Any, key: str, field: str) -> Any:
...
+    def prepare_static_conditioning(
+        self,
+        *,
+        prompt_embeds: torch.Tensor,
+        img_position_ids: torch.Tensor,
+        refiner_cu_seqlens: torch.Tensor,
+        refiner_max_seqlen: int,
+    ) -> MiniMaxH3StaticConditioning:
+        """Prepare request conditioning through the model's forward hooks."""
+        result = self(
```
david6666666, line 1092: Routing prepare through `forward` is a neat way to keep CPU-offload/forward hooks firing once per request - the frozen tensors then let the denoise loop skip projection, refiner, and RoPE entirely. The raw fallback in `_resolve_static_conditioning` keeps direct transformer callers working too.

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
```diff
@@ -1300,11 +1300,23 @@ def diffuse(
 
         tags = packed["token_tags"].clone()
         tags[packed["text_pos"]] = text_tags.cpu()
+        text_len = int(packed["text_pos"].view(-1).shape[0])
+        static_conditioning = self.transformer.prepare_static_conditioning(
```
david6666666, line 1304: Worth checking when this gets rebased onto current main: upstream main has since added combined-mode ref2va (`transformers_ref` + `_transformer_for_task` in `pipeline_minimax_h3.py`) and the denoise loop picks the weights it actually executes via `self._transformer_for_task(task)`. This call site hardcodes `self.transformer`, so after a rebase the static conditioning would be computed with the FL2VA model but consumed by the Ref2VA model in combined mode - a silent producer/consumer mismatch for ref2va. Suggest routing prepare through the same task-resolved model the loop runs.

david6666666:
Verified on a B300 single GPU (same config) vs the d219d93b baseline: output is bit-exact (max frame diff 0, audio identical). Perf: ~155.6s vs ~157.3s (-1.1%, diffuse -2.9s) - a small but consistent e2e gain from skipping projection/refiner/RoPE per step; the call-count reduction claim (19->1) holds by construction. Also left an inline note about the combined-mode ref2va producer/consumer mismatch to watch on rebase to current main.

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
hsliuustc0106, line 1304: `prepare_static_conditioning()` now runs before any `record_denoise_step()`, while the token refiner `self` attention path can consume the current step through TRTLLM Skip-Softmax gating or `diffusion_kv_cache_skip_steps`. Previously the refiner ran inside the loop after the step was recorded; caching this pre-loop result for every iteration changes outputs under those supported configs. Please make the refiner explicitly step-independent or opt it out of those features, or retain per-step refinement for affected configs, and add a step-aware regression test.

PR #5820, opened 2026-08-05
[Model] MiniMax-H3 use direct embedding scatters

LynnColeArt:
## Purpose

MiniMax-H3 assembles its packed multimodal embedding with three `index_add_` calls, one each for text, video, and audio. The packed-layout contract assigns every used row to exactly one modality, so these operations never perform a real reduction.

This change uses `index_copy_` for the three disjoint row sets. The regression test covers text-to-video, first/last-frame conditioning, and mixed image/audio/video reference layouts, verifying that text, image, and audio positions partition every used row exactly once.

Padding remains zero, packed row order is unchanged, and the resulting embedding tensor is bit-exact.

## Test plan

- verify modality positions are unique, disjoint, and exhaustive across supported layout families;
- retain current-main fused-RoPE coverage;

**vLLM Version:** 0.26.0

## Test result

```text
80 passed, 1 skipped, 2 deselected, 15 warnings
Ruff 0.14.10 check: passed
Ruff 0.14.10 format: passed
```

## Isolated DGX Spark microbenchmark

For BF16 output `[6016, 5376]` with 5962 unique used rows:

| Packed assembly | Median | Range |
| --- | ---: | ---: |
| Three `index_add_` scatters | 1.1891 ms | 1.1858–1.2649 ms |
| Three `index_copy_` scatters | 0.8788 ms | 0.8776–0.9287 ms |

Outputs were bit-exact. Direct assignment was 1.353x faster in this isolated packed-assembly benchmark; no end-to-end generation claim is made.

### DIFF
--- a/vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
+++ b/vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
@@ -1067,9 +1067,11 @@ class MiniMaxH3DiTModel(nn.Module):
         )
 
         embeddings = torch.zeros((seq_len, self.hidden_size), device=device, dtype=_BF16_DTYPE)
-        embeddings.index_add_(0, text_pos, text_embed.to(_BF16_DTYPE)[: text_pos.shape[0]])
-        embeddings.index_add_(0, img_pos, video_embed.to(_BF16_DTYPE)[: img_pos.shape[0]])
-        embeddings.index_add_(0, audio_pos, audio_embed.to(_BF16_DTYPE)[: audio_pos.shape[0]])
+        # Every used packed row belongs to exactly one modality, so direct
+        # assignment avoids the atomic addition required by index_add_.
+        embeddings.index_copy_(0, text_pos, text_embed.to(_BF16_DTYPE)[: text_pos.shape[0]])
+        embeddings.index_copy_(0, img_pos, video_embed.to(_BF16_DTYPE)[: img_pos.shape[0]])
+        embeddings.index_copy_(0, audio_pos, audio_embed.to(_BF16_DTYPE)[: audio_pos.shape[0]])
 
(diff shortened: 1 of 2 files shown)


vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
```diff
@@ -1067,9 +1067,11 @@ def _embed(
         )
 
         embeddings = torch.zeros((seq_len, self.hidden_size), device=device, dtype=_BF16_DTYPE)
-        embeddings.index_add_(0, text_pos, text_embed.to(_BF16_DTYPE)[: text_pos.shape[0]])
-        embeddings.index_add_(0, img_pos, video_embed.to(_BF16_DTYPE)[: img_pos.shape[0]])
-        embeddings.index_add_(0, audio_pos, audio_embed.to(_BF16_DTYPE)[: audio_pos.shape[0]])
+        # Every used packed row belongs to exactly one modality, so direct
+        # assignment avoids the atomic addition required by index_add_.
+        embeddings.index_copy_(0, text_pos, text_embed.to(_BF16_DTYPE)[: text_pos.shape[0]])
```
david6666666, line 1072: `index_copy_` is only equivalent to `index_add_` because the packed layout partitions used rows across the three modalities - the partition test pins that for t2va/fl2va/ref2va. If a future layout ever shares a row between modalities, this silently becomes last-write-wins, so that test is doing the real safety work here.

david6666666:
Verified on a B300 single GPU (same config) vs the d219d93b baseline: output is bit-exact (max frame diff 0, audio identical). Perf: ~158.4s vs ~157.3s (+0.7%, within noise) - no measurable e2e change, consistent with the PR making no latency claim; the packed-assembly scatter is a small fraction of the step cost.

PR #5691, opened 2026-08-02
[Model] Add MiniMax H3 diffusion support

Isotr0py:
PLEASE FILL IN THE PR DESCRIPTION HERE.

## Purpose
- Support `MiniMaxAI/MiniMax-H3` diffusion pipeline

## Test Plan

**vLLM Version:**

**vLLM-Omni Commit:**

## Test Result

**BEFORE SUBMITTING:** read [CONTRIBUTING.md](https://github.com/vllm-project/vllm-omni/blob/main/CONTRIBUTING.md) and run the [precheck-pr skill](https://github.com/vllm-project/vllm-omni/blob/main/.claude/skills/precheck-pr/SKILL.md) with the code agent for a self-check against project conventions.
(anything written below this line will be removed by GitHub Actions)

PR #5824, opened 2026-08-05
[Model] Fail MiniMax H3 encoder load when a weight or fused shard is missing

ShengleiFu:
## Purpose

Addresses review feedback on #5691 that landed unaddressed when that PR merged:
https://github.com/vllm-project/vllm-omni/pull/5691#discussion_r3699593302
(roadmap #5700, §1.2 feature completeness).

`MiniMaxH3Qwen3VLEncoder._load_weights` recorded only the fused *target*
parameter name, so a fused QKV or gate/up parameter counted as loaded once any
single source shard arrived. Those parameters start as `torch.empty` and
unloaded parameters only produced a warning, so a checkpoint missing `q`/`k`/`v`
or `gate`/`up` left uninitialized rows in the text encoder. The pipeline then
reported every eagerly loaded text-encoder parameter to the runner's strict
missing-parameter check, hiding the gap.

This PR:

- Tracks the loaded source shard ids per fused parameter and raises when any is
  missing, naming the fused target and the missing sources
  (`...qkv_proj.weight: ['v']`, `...gate_up_proj.weight: ['gate']`).
- Seeds that bookkeeping from the modules that own a multi-shard `weight_loader`,
  so a fused weight that receives *no* source shard is caught too, not only a
  partially filled one.
- Raises instead of warning when any retained parameter was not loaded. Every
  retained parameter has a checkpoint source: the text model builds no final
  norm and caps its layers at `MINIMAX_H3_QWEN3VL_SELECTED_LM_LAYER`, which is
  exactly what `_map_weight_name` skips, and the rotary tables are plain
  attributes rather than parameters. Both raises use `RuntimeError`, matching the
  strict weight loading other models here already do (e.g. `higgs_audio_v3`).
- Notes in `MiniMaxH3Pipeline.load_weights` that reporting the text encoder's
  full parameter set is only exact because of that raise.

A complete checkpoint loads exactly as before.

## Test Plan

- [x] `pytest -q tests/diffusion/models/minimax_h3/test_minimax_h3_contract.py -k encoder_load`
- [x] `pytest -q tests/diffusion/models/minimax_h3 -m "core_model and cpu" --run-level core_model`
- [x] Same tests with `encoder.py` reverted to pre-fix state, to confirm the new
      cases actually fail without the change.
- [x] Name-coverage check against the real `MiniMaxAI/MiniMax-H3` checkpoint, to
      confirm the stricter raise cannot fire on a valid checkpoint.
- [x] `pre-commit run --files <changed files>` (ruff-check, ruff-format, typos,
      debug-statements, end-of-file-fixer, mixed-line-ending, trailing-whitespace,
      check-pickle-imports, check-test-ci-coverage, suggestion — all pass)

## Test Result

Run on Linux / Python 3.12 / torch 2.x / vLLM 0.24.0, CPU only — no GPU and no
MiniMax-H3 weights required. The fixtures build synthetic one-layer
Qwen3-VL-named safetensors from the real TP-aware fused layer classes.

- Focused: **8 passed, 51 deselected**. Whole model dir at `core_model and cpu`:
  **74 passed, 5 deselected**.
- With only `encoder.py` reverted: **7 failed, 1 passed** — every new missing-shard
  case fails with `Failed: DID NOT RAISE RuntimeError`, while
  `test_encoder_load_places_every_source_in_its_own_rows` still passes, since the
  old loader filled a complete checkpoint correctly and only failed to *detect* an
  incomplete one.
- New coverage: `q`, `k`, `v`, `gate`, `up` each missing individually; all of
  `q`/`k`/`v` missing at once (zero-shard fused weight); a plain parameter
  missing; and a positive case that gives every checkpoint tensor a distinct fill
  value and asserts each one landed in its own row range, so it pins placement as
  well as "no uninitialized `torch.empty` rows survive".

Because the stricter raise could regress a valid load, it was also checked
against the real checkpoint: the encoder was built on the `meta` device from the
real `text_encoder/config.json` (no weights, no GPU) and every key in the real
`model.safetensors.index.json` was mapped through `_map_weight_name`:

```
constructed params : 752
checkpoint keys    : 1058  (skipped by mapping: 156)
covered params     : 752
UNCOVERED params (would raise): 0
checkpoint keys mapped to a non-existent param: 0
fused targets: 100 -> complete
```

The 156 skipped keys are `lm_head.weight`, the final text norm, and decoder
layers at or beyond `MINIMAX_H3_QWEN3VL_SELECTED_LM_LAYER`, none of which the
encoder constructs.

Not run: a full weight-loading pass over the real checkpoint (its text encoder
alone is 66.7 GB across 14 shards). The check above is the substitute and covers
the mapping question completely — it compares the full constructed parameter set
against the full real key set. Shape and dtype handling in the per-layer
`weight_loader`s is unchanged by this PR.

## Notes

- Not a duplicate: no open PR adds source-shard validation to the text encoder
  loader. Happy to rebase if something adjacent lands first.
- Unknown checkpoint keys stay a warning while missing parameters become fatal.
  That asymmetry is deliberate: an extra key is a checkpoint the model does not
  fully consume, not a model with uninitialized weights. The coverage check above
  reports zero such keys today.
- AI assistance was used for this change (Claude Code). I reviewed every changed
  line.

### DIFF
--- a/tests/diffusion/models/minimax_h3/test_minimax_h3_contract.py
+++ b/tests/diffusion/models/minimax_h3/test_minimax_h3_contract.py
@@ -8,6 +8,7 @@ from unittest.mock import Mock
 import numpy as np
 import pytest
 import torch
+import torch.nn as nn
 
 pytestmark = [pytest.mark.core_model, pytest.mark.cpu, pytest.mark.diffusion]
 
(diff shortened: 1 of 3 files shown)


hsliuustc0106:
The root cause was that weight-loading completeness was tracked only by the fused target name. Since `q`/`k`/`v` all map to `qkv_proj.weight` (and `gate`/`up` to `gate_up_proj.weight`), loading any one source marked the whole target as loaded. Missing shards therefore left uninitialized `torch.empty` rows, while the existing warning and pipeline bookkeeping hid them from the runner’s strict check. This PR fixes that by tracking each expected source shard and failing fast when any shard or plain parameter is missing.

ShengleiFu:
@hsliuustc0106 Thanks for adding ready label, it seems the only pending CI check is the stuck amd one, which should be unrelated to my change. Are we good to merge this?

PR #5825, opened 2026-08-06
 [Bugfix] Read hidden layer key from model config instead of hardcoding 24

khairulkabir1661:
## Summary

  The stage input processor (`qwen3_omni.py`) hardcoded `_HIDDEN_LAYER_KEY = "24"` for the thinker→talker hidden state handoff. This matches Qwen3-Omni-30B but silently breaks any model variant with a different `accept_hidden_layer`. Replaced with a dynamic lookup that reads from model config.

  ### Changes
  - Add `_get_embedding_layer_index()` and `_get_accept_hidden_layer_index()` for semantic access to layer indices                                                                                   
  - `_get_accept_hidden_layer_index()` supports both model runner mixin (`_get_model_config`) and chunk transfer adapters (`config` field)
  - Use explicit `is None` checks instead of `or` for tensor dict lookups to avoid `RuntimeError` on multi-element tensors
  - Remap numeric layer indices to semantic keys (`"embedding"`, `"accept_hidden"`) in thinker2talker functions
  - Remove dead P/D disagg helper functions (~85 lines)
  - Update `_layer_tensor()` to support semantic string keys

  ## Test plan

  - [x] Qwen3-Omni-30B inference unaffected (config reads `accept_hidden_layer = 24`, same as before)                                                                                                
  - [x] Existing `test_thinker2talker_full_payload_packs_complete_tensors` updated with mock transfer_manager

### DIFF
--- a/vllm_omni/model_executor/stage_input_processors/qwen3_omni.py
+++ b/vllm_omni/model_executor/stage_input_processors/qwen3_omni.py
@@ -32,9 +32,9 @@ from vllm_omni.model_executor.stage_input_processors.tts_utils import (
 
 logger = logging.getLogger(__name__)
 
-# Pooling output layer keys: "0" = word embedding, "24" = accept_hidden_layer
+# Pooling output layer keys: "0" = word embedding; hidden layer read from config
 _EMBED_LAYER_KEY = "0"
-_HIDDEN_LAYER_KEY = "24"
+_DEFAULT_HIDDEN_LAYER_KEY = "24"
 # Per-model REPLACE-keys for the full-payload accumulator.  Keys in this
 # set use REPLACE semantics (subsequent emissions discard prior chunks)
 # instead of CONCAT.  qwen3-omni currently has none — model_outputs is
@@ -56,6 +56,16 @@ def _layer_tensor(layers: dict[Any, Any], key: str) -> torch.Tensor | None:
     if val is None:
         val = layers.get(key)
     return val if isinstance(val, torch.Tensor) else None
+
+
+def _get_hidden_layer_key(transfer_manager: Any) -> str:
+    """Read ``accept_hidden_layer`` from model config via *transfer_manager*."""
+    try:
+        mc = transfer_manager._get_model_config()
+        val = mc.hf_config.talker_config.accept_hidden_layer
+        return str(int(val))
+    except (AttributeError, TypeError, ValueError):
+        return _DEFAULT_HIDDEN_LAYER_KEY
 
 
 def _compute_talker_prompt_ids_length(info: OmniPayload, device: torch.device | str = "cuda") -> int:
@@ -199,6 +209,7 @@ def _merge_pd_embeddings(
     prefill_mm: dict[str, Any],
     device: torch.device,
     expected_total: int | None = None,
+    hidden_layer_key: str = _DEFAULT_HIDDEN_LAYER_KEY,
 ) -> tuple[torch.Tensor, torch.Tensor]:
     """Merge prefill prompt embeddings with decode generated embeddings.
 
@@ -213,7 +224,7 @@ def _merge_pd_embeddings(
     try:
         p_layers = prefill_mm.get("hidden_states", {}).get("layers", {})
         p_emb = p_layers[int(_EMBED_LAYER_KEY)].detach().to(device=device, dtype=torch.float)
-        p_hid = p_layers[int(_HIDDEN_LAYER_KEY)].detach().to(device=device, dtype=torch.float)
+        p_hid = p_layers[int(hidden_layer_key)].detach().to(device=device, dtype=torch.float)
     except (KeyError, AttributeError, TypeError) as exc:
         available_keys = list(prefill_mm.keys()) if isinstance(prefill_mm, Mapping) else type(prefill_mm).__name__
         logger.error(
@@ -222,7 +233,7 @@ def _merge_pd_embeddings(
             "Falling back to decode-only embeddings – talker user-segment will be degraded.",
             exc,
             _EMBED_LAYER_KEY,
-            _HIDDEN_LAYER_KEY,
+            hidden_layer_key,
             available_keys,
         )
         return decode_emb, decode_hid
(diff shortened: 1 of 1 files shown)


vllm_omni/model_executor/stage_input_processors/qwen3_omni.py
```diff
@@ -32,9 +32,9 @@
 
 logger = logging.getLogger(__name__)
 
-# Pooling output layer keys: "0" = word embedding, "24" = accept_hidden_layer
+# Pooling output layer keys: "0" = word embedding; hidden layer read from config
 _EMBED_LAYER_KEY = "0"
-_HIDDEN_LAYER_KEY = "24"
+_DEFAULT_HIDDEN_LAYER_KEY = "24"
```
alex-jw-brooks, line 37: Rather than using the layer indices as keys, can you actually revise this to use semantic keys? E.g., rather than `0` and `24` for embed / hidden respectively, something like `embedding` and `accept_hidden` would be more understandable, because the layer index is really just an implementation detail of the producer

vllm_omni/model_executor/stage_input_processors/qwen3_omni.py
```diff
@@ -58,6 +58,16 @@ def _layer_tensor(layers: dict[Any, Any], key: str) -> torch.Tensor | None:
...
 
 
+def _get_hidden_layer_key(transfer_manager: Any) -> str:
+    """Read ``accept_hidden_layer`` from model config via *transfer_manager*."""
+    try:
+        mc = transfer_manager._get_model_config()
+        val = mc.hf_config.talker_config.accept_hidden_layer
+        return str(int(val))
+    except (AttributeError, TypeError, ValueError):
+        return _DEFAULT_HIDDEN_LAYER_KEY
```
alex-jw-brooks, line 68: Nit: I think we should avoid fallbacks like this and hardcoding it as a default kwarg. Since it probably would be incorrect for that model, it would be better to explode and fix the lookup for the layer indices

alex-jw-brooks (changes requested):
Hi @khairulkabir1661, thanks for fixing this! Some suggestions, I think we should avoid using the indices as keys here since it's not super understandable :)

vllm_omni/model_executor/stage_input_processors/qwen3_omni.py
khairulkabir1661: addressed

vllm_omni/model_executor/stage_input_processors/qwen3_omni.py
khairulkabir1661: Fixed

vllm_omni/model_executor/stage_input_processors/qwen3_omni.py
alex-jw-brooks: Actual, I think this P/D disagg helpers might be dead code..?

vllm_omni/model_executor/stage_input_processors/qwen3_omni.py
```diff
@@ -58,6 +57,22 @@ def _layer_tensor(layers: dict[Any, Any], key: str) -> torch.Tensor | None:
...
+    return 0
+
+
+def _get_accept_hidden_layer_index(transfer_manager: Any) -> int:
+    """Read accept_hidden_layer index from model config.
+
+    No fallback - raises if config is missing or invalid, ensuring
+    incorrect layer indices are caught rather than silently using
+    the wrong layer.
+    """
```
alex-jw-brooks: Can you remove this:

```
    No fallback - raises if config is missing or invalid, ensuring
    incorrect layer indices are caught rather than silently using
    the wrong layer.
```

from the docstring, since this is pretty self explanatory and also doesn't have any explicit error handling?

vllm_omni/model_executor/stage_input_processors/qwen3_omni.py
khairulkabir1661: yes, removed

vllm_omni/model_executor/stage_input_processors/qwen3_omni.py
khairulkabir1661: removed.

vllm_omni/model_executor/stage_input_processors/qwen3_omni.py
```diff
@@ -48,16 +47,32 @@
...
 
 
+def _get_embedding_layer_index() -> int:
+    """Return the layer index for word embeddings (always layer 0)."""
+    return 0
+
+
+def _get_accept_hidden_layer_index(transfer_manager: Any) -> int:
+    """Read accept_hidden_layer index from model config."""
+    mc = transfer_manager._get_model_config()
```
lishunyang12: Handle chunk adapters via their config field; they don't implement _get_model_config.

vllm_omni/model_executor/stage_input_processors/qwen3_omni.py
```diff
@@ -454,13 +387,25 @@ def thinker2talker_async_chunk(
...
     thinker_embed = thinker_embed_raw if isinstance(thinker_embed_raw, dict) else {}
-    thinker_emb = _layer_tensor(thinker_layers, _EMBED_LAYER_KEY)
-    thinker_hid = _layer_tensor(thinker_layers, _HIDDEN_LAYER_KEY)
+
+    embedding_idx = _get_embedding_layer_index()
+    accept_hidden_idx = _get_accept_hidden_layer_index(transfer_manager)
+
+    # Remap numeric layer indices to semantic keys
+    semantic_layers = {
+        "embedding": thinker_layers.get(embedding_idx) or thinker_layers.get(str(embedding_idx)),
```
lishunyang12: Replace these or expressions with None checks; multi-element tensors raise on boolean evaluation.

vllm_omni/model_executor/stage_input_processors/qwen3_omni.py
khairulkabir1661: Did

vllm_omni/model_executor/stage_input_processors/qwen3_omni.py
khairulkabir1661: Thanks, addressed.

khairulkabir1661:
Hi @lishunyang12, thanks a lot, I addressed your comment, can you please have a look? Thanks.

alex-jw-brooks (commented):
Thanks @khairulkabir1661! lgtm. Before merging, can you please add a minimal test ensuring that this works properly for the case where Hidden layer key is patched to something that isn't 24 to ensure that the correct tensor is picked up?

E.g., `accept_hidden_layer=6` should match up to `hidden_states.layer_6)`

khairulkabir1661:
Hi @alex-jw-brooks, I have added the test, please have a look.

alex-jw-brooks (approved):
thanks :)

PR #5752, opened 2026-08-04
[Feature] Align MiniMax H3 official input matrix

david6666666:
Fixes #5751

## Summary

- align FL2VA `[0]`, `[-1]`, and `[0, -1]` keyframes with the official MiniMax H3 interface
- support the official Ref2VA image-only, multi-image, multi-video, external-audio, and mixed reference matrix
- preserve source video containers for HTTP references, add `start_time_seconds` and long-reference handling, and expose official duration, shape, and output-count parameters
- update recipes and model metadata and add API/contract regression coverage

## Validation

- `/home/zjy/code/david/.venv/bin/pytest -q tests/entrypoints/openai_api/test_video_server.py tests/diffusion/models/minimax_h3` — `104 passed, 1 skipped, 19 warnings`
- targeted pre-commit over all changed files — passed
- Ruff, `compileall`, and `git diff --check` — passed
- local-only `MiniMax/MiniMax-H3` two-video reference preparation — `10.041667s + 4.958333s`, `1344x768`, embedded audio detected on the second clip; the public recipe and supported-model entry use `MiniMaxAI/MiniMax-H3`
- current-branch FA4 smoke on 4 GPUs — successful on commit `d5e40ece`; `flash_attn.cute.interface.flash_attn_func`, Ulysses=4, Ring=1, TP=1, VAE tile parallelism=4, 2-step FL2VA in `46.262s`
- output media check — H.264 `1248x768` at 24 fps plus AAC stereo 32 kHz, approximately `8.7s`

This is a smoke validation only; no new performance claim is made by this PR.

PR #5838, opened 2026-08-06
[Bugfix] Fix image num check error

Bounty-hunter:

## Purpose
run `pytest -s -v tests/e2e/accuracy/test_hunyuan_image3.py` failed with:
```
"Received multiple input images. Only a single image is supported by this model."
```
**Reason**:
PR https://github.com/vllm-project/vllm-omni/pull/5752 exposed a pre-existing bug: the architecture name reported in HF config.json differs from the vLLM-Omni pipeline class name, so the metadata lookup failed to match and silently fell back to the wrong defaults.

**Propose**: An existing map already links HF architecture names to Omni pipeline class names — use it to resolve the correct pipeline class name.

## Test Plan
run `pytest -s -v tests/e2e/accuracy/test_hunyuan_image3.py`

**vLLM Version:**

**vLLM-Omni Commit:**

## Test Result
image

### DIFF
--- a/tests/diffusion/test_diffusion_config_propagation.py
+++ b/tests/diffusion/test_diffusion_config_propagation.py
@@ -15,7 +15,10 @@ from vllm_omni.diffusion.data import (
     DiffusionParallelConfig,
     OmniDiffusionConfig,
 )
-from vllm_omni.diffusion.model_metadata import QWEN_IMAGE_EDIT_PLUS_MAX_INPUT_IMAGES
+from vllm_omni.diffusion.model_metadata import (
+    HUNYUAN_IMAGE3_MAX_INPUT_IMAGES,
+    QWEN_IMAGE_EDIT_PLUS_MAX_INPUT_IMAGES,
+)
 
 pytestmark = [pytest.mark.core_model, pytest.mark.cpu]
 
@@ -134,6 +137,30 @@ def test_qwen_image_edit_plus_sets_generic_multimodal_limit():
 def test_task_type_roundtrip():
     od = _roundtrip_diffusion_config(model="x", task_type="model-defined-task")
     assert od.task_type == "model-defined-task"
+
+
+def test_architecture_name_resolves_via_pipeline_class_fallback():
+    # Direct hit: the architecture name itself is the metadata table key.
+    qwen_od_config = OmniDiffusionConfig(
+        model="Qwen/Qwen-Image-Edit-2511",
+        model_class_name="QwenImageEditPlusPipeline",
+    )
+    qwen_od_config.update_multimodal_support()
+
+    assert qwen_od_config.supports_multimodal_inputs is True
+    assert qwen_od_config.max_multimodal_image_inputs == QWEN_IMAGE_EDIT_PLUS_MAX_INPUT_IMAGES
+
+    # Fallback: the HF architecture name differs from the internal pipeline class
+    # name, so ``get_diffusion_model_metadata`` must consult ``_DIFFUSION_MODELS``
+    # and resolve the metadata via the pipeline class name.
+    hunyuan_od_config = OmniDiffusionConfig(
+        model="tencent/HunyuanImage-3.0-Instruct",
+        model_class_name="HunyuanImage3ForCausalMM",
+    )
+    hunyuan_od_config.update_multimodal_support()
+
+    assert hunyuan_od_config.supports_multimodal_inputs is True
+    assert hunyuan_od_config.max_multimodal_image_inputs == HUNYUAN_IMAGE3_MAX_INPUT_IMAGES
 
 
 def test_additional_config_roundtrip():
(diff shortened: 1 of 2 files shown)


Gaohan123 (commented):
Please supplement a UT for it. Thanks

Bounty-hunter:
> Please supplement a UT for it. Thanks

Added

vllm_omni/diffusion/model_metadata.py
```diff
@@ -64,4 +64,14 @@ def get_diffusion_model_metadata(model_class_name: str | None) -> DiffusionModel
...
         return DiffusionModelMetadata()
-    return _DIFFUSION_MODEL_METADATA.get(model_class_name, DiffusionModelMetadata())
+    metadata = _DIFFUSION_MODEL_METADATA.get(model_class_name)
+    if metadata is not None:
+        return metadata
+    # Some checkpoints report the HF architecture name diff from internal pipeline class name
+    # (e.g. HunyuanImage3ForCausalMM, WanVACEPipeline, OmniVoice ...).
+    from vllm_omni.diffusion.registry import _DIFFUSION_MODELS
+
+    entry = _DIFFUSION_MODELS.get(model_class_name)
```
fhfuih, line 74: I think if `_DIFFUSION_MODELS` is to be used externally, we can rename it to `DIFFUSION_MODELS`

fhfuih, line 74: Also, to prevent future changes to `DIFFUSION_MODELS` (for example, adding a field before the current [2]) we can use unpacking `_, _, hf_cls_name = DIFFUSION_MODELS.get` so future update naturally raises error.

fhfuih (commented):
LGTM, some minor comment

vllm_omni/diffusion/model_metadata.py
```diff
@@ -64,4 +64,14 @@ def get_diffusion_model_metadata(model_class_name: str | None) -> DiffusionModel
...
+    if entry is not None:
+        return _DIFFUSION_MODEL_METADATA.get(entry[2], DiffusionModelMetadata())
```
david6666666: Works for the reported case since `HunyuanImage3Pipeline` happens to be a metadata key. Just noting this is narrower than it looks: the registry's `cls_name` and the metadata keys are two separate key spaces — e.g. `WanVACEPipeline` resolves to `Wan22VACEPipeline` and `OmniVoice` to `OmniVoicePipeline`, neither of which is in `_DIFFUSION_MODEL_METADATA`, so those aliases still silently fall back to the empty defaults (no regression, just no coverage). Not blocking, but maybe worth keeping the two key spaces in sync or adding a comment.

tests/diffusion/test_diffusion_config_propagation.py
```diff
@@ -136,6 +140,20 @@ def test_task_type_roundtrip():
...
+def test_architecture_name_resolves_via_pipeline_class_fallback():
+    # Direct hit: the architecture name itself is the metadata table key.
+    qwen_metadata = get_diffusion_model_metadata("QwenImageEditPlusPipeline")
+    assert qwen_metadata.supports_multimodal_inputs is True
+    assert qwen_metadata.max_multimodal_image_inputs == QWEN_IMAGE_EDIT_PLUS_MAX_INPUT_IMAGES
+
+    # Fallback: the HF architecture name differs from the internal pipeline class
+    # name, so ``get_diffusion_model_metadata`` must consult ``_DIFFUSION_MODELS``
+    # and resolve the metadata via the pipeline class name.
+    hunyuan_metadata = get_diffusion_model_metadata("HunyuanImage3ForCausalMM")
```
david6666666: Nice, thanks for the UT. Small optional suggestion: the actual failure surfaced through `_get_max_edit_input_images` reading `od_config.supports_multimodal_inputs`, so asserting via `OmniDiffusionConfig(model_class_name=...).update_multimodal_support()` like the Qwen test above would lock the config plumbing path too, not just the lookup helper. Not blocking.

david6666666 (commented):
Thanks for the fix — the fallback resolves the reported HunyuanImage3 case and CI is green. Left two small non-blocking notes below.

david6666666 (approved):
LGTM. Thanks for the quick follow-up — the UT now exercises the real `update_multimodal_support()` path, the registry tuple is unpacked instead of the magic `[2]`, and the separate key-space note is documented. CI is green and the fallback fixes the HunyuanImage3 multi-image admission path. Nice and minimal.

david6666666 (approved):

vllm_omni/diffusion/model_metadata.py
Bounty-hunter: fixed

tests/diffusion/test_diffusion_config_propagation.py
Bounty-hunter: fixed

PR #5840, opened 2026-08-06
[MiniMax-H3] TeaCache support and Cache-DiT validation

dpeng123:
#5700 
## Purpose
Add TeaCache support for MiniMax-H3 (MiniMaxH3DiTModel / FL2VA partition)

### Environment

- GPU: 1x NVIDIA H100 PCIe, 81,559 MiB
- Container: `vastai/vllm-omni:v0.26.0-cuda-12.9`
- Container digest: `sha256:d9414bdd23480c4882d41a71d3f8f7e494444345e183d4a708e15468151f7660`
- Model: MiniMax-H3 `FL2VA`, revision `bfc8ed0353f5a9733be73e6b2c98ec0948195b86`
- Runtime: BF16, eager mode, FlashAttention, model-level CPU offload

### TeaCache coefficient calibration

Recalibrated the fourth-order polynomial for the valid 4-second/107-frame
workload. The fit used 70 diverse T2VA prompts, seeds 42 through 111,
256x448, 24 FPS, and 50 inference steps. Every DiT evaluation was forced to
run during collection; no cached residuals were used. The resulting data set
contains 3,360 consecutive-denoising-step pairs.

```text
[2283.7040658527776,
 -777.5977277886368,
 94.0841474135949,
 -4.232669906169421,
 0.2173782527946167]
```

Calibration statistics:

- Input relative-L1: min 0.010979, max 0.159318, mean 0.024522
- Output relative-L1: min 0.066069, max 0.362911, mean 0.172595
- Polynomial RMSE: 0.039703
- Polynomial R-squared: 0.094873

### Focused extractor tests

Run inside the one-H100 container. These are CPU-marked unit tests and do not
consume the GPU themselves.

```bash
python -m pytest -q \
  tests/diffusion/cache/test_teacache_extractors.py \
  -k TestMiniMaxH3Extractor
```

```text
........                                                                 [100%]
8 passed, 18 deselected in 2.52s
```

The current-main one-GPU E2E smoke also completed successfully:

```bash
python -m pytest -q \
  tests/diffusion/models/minimax_h3/test_minimax_h3_e2e.py::test_minimax_h3_t2va_teacache_single_gpu_smoke
```

```text
Exit code: 0
```

### Online smoke test

Server:

```bash
vllm serve /path/to/MiniMax-H3/FL2VA \
  --omni \
  --host 0.0.0.0 \
  --port 8091 \
  --trust-remote-code \
  --num-gpus 1 \
  --enable-cpu-offload \
  --enforce-eager \
  --diffusion-attention-backend FLASH_ATTN \
  --cache-backend tea_cache \
  --cache-config '{"rel_l1_thresh":0.2}'
```

Request:

```bash
curl -sS -X POST http://127.0.0.1:8091/v1/videos/sync \
  -F 'prompt=A quiet cinematic night scene in a rain-soaked city, with distant traffic and soft thunder.' \
  -F 'width=448' \
  -F 'height=256' \
  -F 'fps=24' \
  -F 'num_inference_steps=50' \
  -F 'flow_shift=12' \
  -F 'seed=42' \
  -F 'extra_params={"task":"t2va","duration":4.0,"aspect_ratio":"16:9","audio_flow_shift":3.0}' \
  -o online-teacache-newcoeff.mp4
```

Server evidence:

```text
TeaCache applied with rel_l1_thresh=0.2, transformer_class=MiniMaxH3DiTModel
POST /v1/videos/sync HTTP/1.1 200 OK
```

Result:

```text
HTTP status:          200
Client wall time:     62.904 s
Inference header:     62.897 s
Stage 0 generation:  62.606 s
Peak memory header:  75,310 MB
Output size:          789,597 bytes
SHA-256:              b28f07f646af83325046af3475d71b5a7f1427e73b4159bb200f709fa9d9babb
Container:            MP4
Video:                H.264, 448x256, 24 FPS
Audio:                AAC, stereo, 32,000 Hz
Duration:             4.482 s
```

### Offline baseline/TeaCache A/B

Both arms used the same prompt as the online test, seed 42, 256x448, 24 FPS,
4-second duration, 107 output frames, 50 inference steps, video flow shift 12,
and audio flow shift 3. Model initialization was excluded, but each arm's first
generation was otherwise unwarmed and ran in a separate container.

| Mode | Exit | Joint output | Wall time | Diffusion stage | Mean DiT evaluation time | Peak memory | LPIPS vs baseline | PSNR | MAE | Audio spectral cosine | Audio RMS ratio |
|---|---:|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| Baseline, no cache | 0 | 107-frame video + stereo audio | 75.1939 s | 72.9009 s | 1487.773 ms | 75,116 MB | 0.0000 | inf | 0.000000 | 1.0000 | 1.0000 |
| TeaCache, `rel_l1_thresh=0.2` | 0 | 107-frame video + stereo audio | 62.6917 s | 60.3697 s | 1232.035 ms | 75,310 MB | 0.3162 | 17.8896 dB | 0.060748 | 0.9897 | 0.6684 |

Performance result:

- End-to-end speedup: 1.20x; time reduction: 16.6%
- Diffusion-stage speedup: 1.21x; time reduction: 17.2%
- Peak-memory difference: +194 MB (+0.26%)
- Output contracts matched: video shape, audio shape, FPS, and audio sample rate

Quality result:

- LPIPS: 0.3161838450
- PSNR: 17.8895946571 dB
- MAE: 0.0607480220
- Audio spectral cosine: 0.9896786809
- Audio RMS ratio: 0.6684345316

For this prompt at threshold 0.2, the recalibrated coefficients selected the
same cache pattern as the previous coefficients, and the resulting TeaCache
NPZ was byte-identical. Recalibration therefore did not change this prompt's
LPIPS or other quality metrics. The test validates functionality and measurable
speedup, but LPIPS 0.3162 is still a material visual difference; it should not
be described as a close baseline match or a universal quality guarantee.

### Lower-threshold sweep

The same A/B was repeated at lower thresholds while reusing the verified,
deterministic baseline artifact.

| Threshold | Wall time | Wall reduction | Diffusion stage | Diffusion reduction | LPIPS | PSNR | MAE | Audio spectral cosine | Audio RMS ratio |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| Baseline | 75.1939 s | - | 72.9009 s | - | 0.000000 | inf | 0.000000 | 1.000000 | 1.000000 |
| 0.10 | 76.2015 s | -1.34% | 73.8988 s | -1.37% | 0.000000 | inf | 0.000000 | 1.000000 | 1.000000 |
| 0.15 | 78.5028 s | -4.40% | 76.1661 s | -4.48% | 0.000000 | inf | 0.000000 | 1.000096 | 1.000000 |
| 0.16 | 75.4619 s | -0.36% | 72.7927 s | 0.15% | 0.006252 | 40.6673 dB | 0.004040 | 0.999720 | 0.974102 |
| 0.17 | 72.9487 s | 2.99% | 70.6379 s | 3.10% | 0.013441 | 36.4038 dB | 0.005268 | 0.999300 | 0.950384 |
| 0.18 | 62.7435 s | 16.56% | 60.4265 s | 17.11% | 0.120304 | 23.8650 dB | 0.027227 | 0.993802 | 0.853479 |
| 0.19 | 62.9564 s | 16.27% | 60.6592 s | 16.79% | 0.316184 | 17.8896 dB | 0.060748 | 0.989679 | 0.668435 |
| 0.20 | 62.6917 s | 16.63% | 60.3697 s | 17.19% | 0.316184 | 17.8896 dB | 0.060748 | 0.989679 | 0.668435 |

At `0.10` and `0.15`, no cache reuse occurred: the TeaCache NPZ was
byte-identical to the baseline, and hook overhead made the runs slower.
Threshold `0.16` introduced a small quality delta without a useful wall-time
gain. Threshold `0.17` was the best conservative point in this sweep, with
LPIPS 0.0134 and a 3.0% wall-time reduction. At `0.18`, acceleration increased
sharply, but LPIPS rose to 0.1203; `0.19` and `0.20` produced the same larger
quality loss. This shows that threshold selection controls the observed
quality/performance tradeoff for this workload.

### Conclusion

MiniMax-H3 TeaCache passes its focused extractor tests, the one-H100 E2E smoke,
the online HTTP generation path, and offline 107-frame baseline/TeaCache A/B
tests with the recalibrated FL2VA coefficients. At `rel_l1_thresh=0.2`, this
workload shows about 1.20x end-to-end speedup but a material quality delta.
Threshold `0.17` is the best conservative point tested with the final
coefficients, with LPIPS 0.0134 and a 3.0% wall-time reduction on this prompt.
No threshold should be treated as a universal guarantee without representative
workload validation.

## Cache-DiT validation and tuning

No Cache-DiT implementation change was required. MiniMax-H3 already registers a
`CacheDiTAdapterConfig` for its `blocks` module using
`ForwardPattern.Pattern_3`, with joint conditional/unconditional execution
(`has_separate_cfg=False`). The validation below confirms that the generic
Cache-DiT backend recognizes that adapter and executes the joint video/audio
pipeline.

### Unit tests

```bash
python -m pytest -q \
  tests/diffusion/cache/test_cache_backends.py \
  tests/diffusion/cache/test_cache_dit.py
```

The container needed the test-only `pytest-asyncio` and `pytest-mock`
dependencies installed before the checkout's pytest configuration could run.

```text
.............................................                            [100%]
45 passed, 15 warnings in 6.83s
```

### Offline baseline/configuration comparison

The same prompt, seed, dimensions, frame count, step count, T2VA task, runtime,
and uncached baseline from the TeaCache A/B test were used. Two Cache-DiT
profiles were evaluated:

- Default: `Fn=1`, `Bn=0`, warmup 4, residual threshold 0.24, maximum 3
  continuous cached steps.
- Conservative: `Fn=1`, `Bn=0`, warmup 8, residual threshold 0.12, maximum 2
  continuous cached steps.

In both runs, Cache-DiT reported a successful `Pattern_3` adapter match and
refreshed its cache context for 50 inference steps.

| Mode | Exit | Joint output | Wall time | Diffusion stage | Peak memory | LPIPS vs baseline | PSNR | MAE | Audio spectral cosine | Audio RMS ratio |
|---|---:|---|---:|---:|---:|---:|---:|---:|---:|---:|
| Baseline, no cache | 0 | 39-frame video + stereo audio | 54.7057 s | 52.9084 s | 74,172 MB | 0.0000 | inf | 0.000000 | 1.0000 | 1.0000 |
| Cache-DiT default | 0 | 39-frame video + stereo audio | 51.8101 s | 49.9889 s | 74,172 MB | 0.1499 | 25.4682 dB | 0.025861 | 0.9385 | 0.7523 |
| Cache-DiT conservative | 0 | 39-frame video + stereo audio | 53.6149 s | 51.7763 s | 74,530 MB | 0.0449 | 31.3471 dB | 0.014183 | 0.9692 | 0.8668 |

Default-profile result:

- End-to-end speedup: 1.06x; generation-time reduction: 5.3%
- Diffusion-stage speedup: 1.06x; diffusion-time reduction: 5.5%
- Peak memory unchanged
- The output contract matched, but video and audio drift was materially larger
  than with TeaCache at its tested threshold.

Conservative-profile result:

- End-to-end speedup: 1.02x; generation-time reduction: 2.0%
- Diffusion-stage speedup: 1.02x; diffusion-time reduction: 2.1%
- Peak-memory difference: +358 MB (+0.48%)
- Substantially closer baseline match than the default Cache-DiT profile

This single-H100 CPU-offload profile is dominated by the fixed first-step DiT
transfer, as described above, so the timing results understate the benefit of
skipping later block computation. For quality-sensitive MiniMax-H3 T2VA usage,
the conservative profile is the safer starting point; the default profile
trades considerably more output fidelity for roughly three additional
percentage points of end-to-end time reduction in this workload.

## Conclusion
The existing MiniMax-H3 Cache-DiT adapter passes its focused unit suites,
activates without a model-specific runtime change, and preserves the joint
output contract. Its default configuration works but has a visible quality
tradeoff; the tested conservative configuration improves the match at the cost
of most of the measured speed gain. Broader prompts, resolutions, durations,
tasks, and resident multi-GPU configurations should be benchmarked before
claiming a universal quality or speed guarantee for either cache backend.


vllm_omni/diffusion/cache/teacache/extractors.py
```diff
@@ -1249,6 +1249,149 @@ def postprocess(h: torch.Tensor) -> SenseNovaU1CausalLMOutput:
...
+    )
+
+    def run_transformer_blocks() -> tuple[torch.Tensor, ...]:
+        hidden, block_rope, block_combined = module.sp_prepare(
+            decoder_input,
+            rope_freqs,
+            combined_indices,
+        )
+        for block in module.blocks:
+            hidden = block(
```
hsliuustc0106, line 1358: Current `main` added the required `packed_total` argument and forwards `video_token_layout` as `video_layout`, but this duplicate block call passes neither. I reproduced `TypeError: MiniMaxH3DiTBlock.forward() missing 1 required keyword-only argument: 'packed_total'` locally after applying this PR onto current `main`. Please mirror or share the native block-call contract here.

tests/diffusion/cache/test_teacache_extractors.py
```diff
@@ -423,3 +427,249 @@ def test_invalid_module_raises_error(self):
...
+        super().__init__(
+            hidden_size,
+            qkv_features,
+            total_num_heads=total_num_heads,
+            total_num_kv_heads=total_num_kv_heads,
+            **kwargs,
+        )
+
+
+class _MiniMaxH3FakeAttention(nn.Module):
```
hsliuustc0106, line 487: These new tests fail on current `main` before exercising TeaCache: `_run_packed_attention` reads `self.attention.attn_backend.supports_prefix_kv_slicing`, but this fake has no `attn_backend` (6 failed, 1 passed locally). Please make the fake contract-complete or reuse a shared attention test helper.

vllm_omni/diffusion/cache/teacache/extractors.py
dpeng123: @hsliuustc0106 Thanks for reviewing the PR! Just addressed the review comment, and updated PR description with the test result, could you check?

tests/diffusion/cache/test_teacache_extractors.py
dpeng123: Just addressed the review comment and run test again, test passed, could you check?

hsliuustc0106:
could you update with the recipe as well?

dpeng123:
> could you update with the recipe as well?

Updated, could you check?

lishunyang12 (changes requested):
Thanks for adding MiniMax-H3 TeaCache support. The extractor implementation and focused unit tests look solid, but I found two behavioral blockers and one recipe issue that should be addressed before merge.

1. **TeaCache is registered for all `MiniMaxH3DiTModel` instances, while the coefficients are explicitly FL2VA-only.** The generic enabler only hooks `pipeline.transformer`; the combined MiniMax-H3 pipeline stores Ref2VA separately in `transformers_ref`. As a result, combined mode silently accelerates only FL2VA, while a Ref2VA-only pipeline can be hooked with the FL2VA coefficients because both partitions use the same model class. Please add a MiniMax-H3-specific enabler that checks `pipeline.partition`, rejects or clearly disables unsupported Ref2VA caching, and handles every supported task transformer explicitly. The feature table and recipe should also state that support is FL2VA-only.

2. **The default threshold is known to produce material quality loss.** Enabling `tea_cache` without a cache config still uses `rel_l1_thresh=0.2`. This PR's own evaluation at 0.2 reports LPIPS 0.3162, PSNR 17.89 dB, audio RMS ratio 0.6684, and a material visual difference. The recipe's explicit 0.17 setting does not change the runtime default. Please provide a MiniMax-H3-specific safe default, or require an explicit threshold rather than treating 0.2 as minimal quality loss.

3. **The offline Python example is not runnable as written.** `model=f"{MODEL_ROOT}/FL2VA"` references a Python variable that is never defined. Please use `os.environ["MODEL_ROOT"]` with an export instruction, or provide a directly defined path.

Validation performed:

- Exact PR head: 8 MiniMax-H3 extractor tests passed; the full TeaCache extractor file passed 26/26.
- Clean merge onto current upstream main: TeaCache extractors 26/26, cache backend tests 32/32, the config-field test that failed in Buildkite 1/1, pre-commit checks, and `git diff --check` all passed.
- The main Buildkite failure is unrelated and passes on current main. The AMD build expired before its jobs started.

I could not independently rerun the model-backed GPU E2E test because the required MiniMax-H3 model path was not available. The committed two-step E2E smoke test checks the output contract but does not assert that real denoising produces a cache hit.

lishunyang12 (approved):

dpeng123:
> Thanks for adding MiniMax-H3 TeaCache support. The extractor implementation and focused unit tests look solid, but I found two behavioral blockers and one recipe issue that should be addressed before merge.
> 
> 1. **TeaCache is registered for all `MiniMaxH3DiTModel` instances, while the coefficients are explicitly FL2VA-only.** The generic enabler only hooks `pipeline.transformer`; the combined MiniMax-H3 pipeline stores Ref2VA separately in `transformers_ref`. As a result, combined mode silently accelerates only FL2VA, while a Ref2VA-only pipeline can be hooked with the FL2VA coefficients because both partitions use the same model class. Please add a MiniMax-H3-specific enabler that checks `pipeline.partition`, rejects or clearly disables unsupported Ref2VA caching, and handles every supported task transformer explicitly. The feature table and recipe should also state that support is FL2VA-only.
> 2. **The default threshold is known to produce material quality loss.** Enabling `tea_cache` without a cache config still uses `rel_l1_thresh=0.2`. This PR's own evaluation at 0.2 reports LPIPS 0.3162, PSNR 17.89 dB, audio RMS ratio 0.6684, and a material visual difference. The recipe's explicit 0.17 setting does not change the runtime default. Please provide a MiniMax-H3-specific safe default, or require an explicit threshold rather than treating 0.2 as minimal quality loss.
> 3. **The offline Python example is not runnable as written.** `model=f"{MODEL_ROOT}/FL2VA"` references a Python variable that is never defined. Please use `os.environ["MODEL_ROOT"]` with an export instruction, or provide a directly defined path.
> 
> Validation performed:
> 
> * Exact PR head: 8 MiniMax-H3 extractor tests passed; the full TeaCache extractor file passed 26/26.
> * Clean merge onto current upstream main: TeaCache extractors 26/26, cache backend tests 32/32, the config-field test that failed in Buildkite 1/1, pre-commit checks, and `git diff --check` all passed.
> * The main Buildkite failure is unrelated and passes on current main. The AMD build expired before its jobs started.
> 
> I could not independently rerun the model-backed GPU E2E test because the required MiniMax-H3 model path was not available. The committed two-step E2E smoke test checks the output contract but does not assert that real denoising produces a cache hit.

@lishunyang12 Thanks for reviewing the PR and pushed commit to fix the PR, really appreciate the help!

PR #5842, opened 2026-08-06
[Model] Support NVIDIA-NemotronLabs-VoiceChat-11B offline speech-to-speech

yuekaizhang:
## Purpose

Add offline speech-in / speech-out support for
[nvidia/NVIDIA-NemotronLabs-VoiceChat-11B](https://huggingface.co/nvidia/NVIDIA-NemotronLabs-VoiceChat-11B),
a full-duplex voice-chat model with a frame-locked 12.5 Hz timeline. This PR
covers single-turn offline inference (batch=1); online/duplex serving,
function-calling, and batch>1 are documented follow-ups.

The model is mapped onto a qwen3-omni-style 3-stage pipeline:

| stage | arch | role |
|---|---|---|
| 0 thinker | `NemotronVoiceChatThinkerForConditionalGeneration` (LLM_AR) | Conformer perception + NemotronH hybrid-Mamba backbone; consumes the user WAV + system prompt, emits the frame-locked text-token timeline (plus a function-call channel) |
| 1 talker | `NemotronVoiceChatTalker` (LLM_AR) | Gemma3-1B EAR-TTS; consumes the text timeline, emits 31-quantizer RVQ code stacks per frame |
| 2 code2wav | `NemotronVoiceChatCode2Wav` (LLM_GENERATION) | RVQ-VAE codec decode to 22.05 kHz PCM |

Key design points:

- **Vendored NeMo modules** (`nemo_vendored/`, ~9.7k lines): the perception
  Conformer, EAR-TTS talker, and RVQ-VAE codec do not exist in
  transformers/vLLM, and the reference implementation lives on an unreleased
  NeMo branch (`nemotron-labs-voicechat`), so a `nemo_toolkit` dependency is
  not pinnable. Files are dependency-stripped copies with upstream paths
  annotated in each header; a transitive dead-code pass removed
  training/export/streaming-only code, and the vendored librosa mel math was
  replaced with `torchaudio.functional.melscale_fbanks` (verified token- and
  WAV-identical).
- **Fused prefill contract**: the vLLM prompt is `logical_prompt_len + 1`
  tokens — the extra position is acoustic frame 0, so decode steps map 1:1 to
  12.5 Hz frames. The frame lock is enforced with `max_tokens=n_frames` +
  `ignore_eos=True` (the tokenizer's EOS is also the PAD/silence token, so
  `min_tokens` must never be set on the thinker).
- **Tokenizer**: the checkpoint ships no HF tokenizer; it resolves from the
  `nvidia/NVIDIA-Nemotron-Nano-9B-v2` HF id, overridable via
  `NEMOTRON_VOICECHAT_LLM_PATH` for air-gapped runs.
- Stage 0/1 run fp32 by default for exact reference parity; the deploy yaml
  documents a ~2x-faster bf16 thinker option (output within one word of the
  reference on the fixture).


yuekaizhang:
@linyueqian  I was wondering if you could help review the PR. Thanks!

vllm_omni/model_executor/models/nemotron_voicechat/nemotron_voicechat_thinker.py
```diff
@@ -0,0 +1,566 @@
...
+    n_fft = int(pre.get("n_fft") or 2 ** int(np.ceil(np.log2(window_size * sample_rate))))
+    hop = int(window_stride * sample_rate)
+    pad_amount = n_fft // 2 * 2  # torch.stft center padding on both sides
+    mel_len = (num_samples + pad_amount - n_fft) // hop + 1
+    subsampling_factor = int(encoder.get("subsampling", "dw_striding") and encoder.get("subsampling_factor", 8))
+    conv_kernel = int(encoder.get("subsampling_conv_kernel_size", 3) or 3)
+    repeat = int(np.log2(subsampling_factor))
+    out = calc_length(
+        torch.tensor([mel_len]),
+        all_paddings=2 * (conv_kernel // 2),
```
linyueqian: [blocking] This is the non-causal padding formula, but the checkpoint sets `model.stt.model.perception.encoder.causal_downsampling: true`. In the vendored `ConvSubsampling`, causal mode sets `_left_padding = kernel-1 = 2` and `_right_padding = stride-1 = 1`, and `forward()` calls `calc_length(all_paddings=self._left_padding + self._right_padding)`, so the real value is 3, not `2 * (conv_kernel // 2) = 2`.

I built the vendored `AudioPerceptionModule` from the real config and compared:

```
ConvSubsampling is_causal = True
  left_padding = 2  right_padding = 1  ->  all_paddings used by forward = 3

  samples     sec  helper   real  verdict
   250760   15.67     196    197  MISMATCH
    16000    1.00      13     14  MISMATCH
   100000    6.25      79     80  MISMATCH
   128000    8.00     101    101  OK
...
9/12 mismatched, always short by 1
```

Since `end2end.py` passes the result as `nvc_expected_frames`, the mismatch is fatal. A 4.45 s question (71271 samples) fails at prefill:

```
ValueError: NemotronVoiceChat frame-count mismatch: the request was sized for
56 acoustic frames but perception produced 57.
```

Suggested fix: read `causal_downsampling` from the encoder config and use `(kernel - 1) + (stride - 1)` when it is set, keeping `2 * (kernel // 2)` otherwise.

tests/model_executor/models/test_nemotron_voicechat_shapes.py
```diff
@@ -0,0 +1,151 @@
...
+)
+
+pytestmark = [pytest.mark.core_model, pytest.mark.cpu]
+
+# Perception config subset matching the checkpoint (mel 10 ms hop, n_fft 512,
+# dw-striding 8x subsampling, conv kernel 3).
+_STT_CFG = {
+    "perception": {
+        "preprocessor": {"sample_rate": 16000, "window_stride": 0.01, "window_size": 0.025, "n_fft": 512},
+        "encoder": {"subsampling": "dw_striding", "subsampling_factor": 8, "subsampling_conv_kernel_size": 3},
```
linyueqian: [important] This encoder subset omits `causal_downsampling`, which is what lets the blocking bug through. `compute_acoustic_frame_count()` reads only the keys listed here, so the test exercises the helper against a config that does not describe this checkpoint, and it passes for the same reason the helper is wrong.

Also worth reconciling: the docstring attributes `(250760, 196)` on line 33 to a NeMo reference run, but the vendored perception in this PR returns 197 for that input. One of the two is wrong, and it would be good to know which before pinning the golden.

Suggestion: add `"causal_downsampling": True` here, and add a case that asserts the helper against the real geometry, for example `perception.encoder.pre_encode._left_padding + _right_padding`, so the arithmetic cannot drift from the module it is meant to mirror.

vllm_omni/model_executor/models/nemotron_voicechat/configuration_nemotron_voicechat.py
```diff
@@ -0,0 +1,208 @@
...
+            try:
+                return AutoConfig.for_model(**dict(provided))
+            except (KeyError, ValueError):
+                return PretrainedConfig(**provided)
+        if not self.stt_cfg:
+            # Skeleton config (bare construction); resolved lazily by the stage.
+            return PretrainedConfig()
+        llm_ref = os.environ.get(_LLM_PATH_ENV) or self.stt_cfg.get("pretrained_llm") or _DEFAULT_LLM_ID
+        try:
+            return AutoConfig.from_pretrained(llm_ref, trust_remote_code=True)
```
linyueqian: [important] `trust_remote_code=True` makes the documented air-gapped path impossible. `nvidia/NVIDIA-Nemotron-Nano-9B-v2` ships `auto_map: {AutoConfig: configuration_nemotron_h.NemotronHConfig}`, so transformers takes the remote-code branch and requires `configuration_nemotron_h.py` to be present. A directory containing exactly what the error message below asks for (its `config.json`) therefore fails.

Tested against such a directory:

```
trust_remote_code=True:  FAIL OSError: ... does not appear to have a file named configuration_nemotron_h.py
trust_remote_code=False: OK   cls=NemotronHConfig hidden=4480 layers=56
```

`nemotron_h` is a native transformers model type and the thinker builds `NemotronHForCausalLM` through `init_vllm_registered_model` anyway, so dropping the flag both fixes the air-gap story and removes a hub round trip for the `.py` modules. The same argument applies to the `AutoTokenizer` calls at thinker line 257 and `end2end.py` line 76.

examples/offline_inference/nemotron_voicechat/end2end.py
```diff
@@ -0,0 +1,193 @@
...
+    )
+    tokenizer = AutoTokenizer.from_pretrained(tok_ref, trust_remote_code=True)
+    bos_id = tokenizer.convert_tokens_to_ids(stt_cfg.get("bos_token", "<s>"))
+    eos_id = tokenizer.convert_tokens_to_ids(stt_cfg.get("eos_token", "</s>"))
+    pad_id = tokenizer.convert_tokens_to_ids(stt_cfg.get("pad_token", "<SPECIAL_12>"))
+    # NeMo encode_system_prompt: [bos] + tokens + [eos].
+    prompt_ids = [bos_id] + tokenizer.encode(args.system_prompt, add_special_tokens=False) + [eos_id]
+
+    wav = load_wav_16k_mono(args.wav)
+    n_frames = compute_acoustic_frame_count(stt_cfg, int(wav.shape[0]))
```
linyueqian, line 106: [important] Two things worth surfacing to whoever runs this.

First, this call is the one affected by the causal padding bug, so the example currently fails on most input durations.

Second, even once that is fixed, the frame lock means the reply budget is the input length. The acoustic channel trails the text channel, so the spoken answer is truncated if the WAV does not carry enough trailing silence for the tail to finish, and the truncation is silent. Concretely, with a question ending at 4.45 s in an 8 s WAV the reply stopped mid-sentence and ASR gave WER 0.52 against the emitted text; padding the identical question to 16 s gave WER 0.0. The frame timelines show it clearly:

```
 8 s (101 frames):  PAD x9, TOK x1, PAD x50, TOK x31, PAD x10
16 s (201 frames):  PAD x9, TOK x1, PAD x50, TOK x20, PAD x121
```

A note in the example docstring and the recipe would save people a confusing debugging session. A cheap guard would be to warn when the text channel emits non-PAD tokens within the last N frames.

vllm_omni/model_executor/models/nemotron_voicechat/nemotron_voicechat_thinker.py
```diff
@@ -0,0 +1,566 @@
...
+        # NeMo's gen_function[prompt_len]. The cumulative debug timeline
+        # (nvc_function_tokens) is owned by postprocess, which also covers the
+        # final acoustic frame (no later preprocess exists for it).
+        if self._use_function_head:
+            prev_func = info.get("nvc_prev_function_token")
+            if prev_func is not None:
+                token = int(prev_func.reshape(-1)[0]) if isinstance(prev_func, torch.Tensor) else int(prev_func)
+                session["func_token"] = token
+
+        session["decode_step"] = int(session["decode_step"]) + 1
```
linyueqian: [suggestion] The frame index is derived from how many times `preprocess` has been called rather than from the request's position, so it is not idempotent. If vLLM ever re-enters the same decode position (preemption and recompute), the audio pointer advances an extra frame and the whole timeline silently misaligns against the text channel, with no error.

`_omni_num_computed_tokens` is already available here (the prefill branch above uses it), so `frame_idx` could be derived from it instead and the state would become position addressed. The shipped deploy avoids the issue today with `max_num_seqs: 1` and `enable_prefix_caching: false`, so this is about robustness rather than a live bug, but the failure mode is silent corruption which makes it worth hardening.

vllm_omni/model_executor/models/nemotron_voicechat/nemotron_voicechat_thinker.py
```diff
@@ -0,0 +1,566 @@
...
+            computed = int(kwargs.get("_omni_num_computed_tokens", 0) or 0)
+            prompt_len = int(kwargs.get("_omni_prompt_len", 0) or 0)
+            if prompt_len and computed + int(hidden_states.shape[0]) < prompt_len:
+                return {}  # intermediate prefill chunk: frame 0 not reached yet
+        with torch.inference_mode():
+            func_logits = self.function_head(hidden_states[-1, :].reshape(1, -1).to(self._dtype))
+        token = func_logits.argmax(dim=-1).reshape(()).detach()
+        existing = kwargs.get("nvc_function_tokens")
+        if isinstance(existing, torch.Tensor) and existing.numel() > 0:
+            timeline = torch.cat([existing.reshape(-1).cpu(), token.reshape(1).cpu()])
```
linyueqian: [suggestion] This cumulative debug timeline costs a device to host sync plus a growing `torch.cat` on every decode step, which is quadratic over the request, and in practice it never reaches the user. Neither of my two successful runs produced a `*_function_tokens.json`, because the best-effort dump in `end2end.py` only fires if the engine happens to surface the request info on the final output, and it does not.

To be clear about what is load bearing: `nvc_prev_function_token` genuinely is, since it feeds the fusion at weight 2.0, so the `function_head` matmul has to stay. It is only the cumulative `nvc_function_tokens` accumulation that is paid for and then discarded. Options are to keep it on device and materialize once at request end, gate it behind an env var, or drop it until the function-call channel is actually consumed.

vllm_omni/model_executor/models/nemotron_voicechat/nemotron_voicechat_code2wav.py
```diff
@@ -0,0 +1,227 @@
...
+
+        for i in range(num_req):
+            runtime_info = infos[i] if i < len(infos) else None
+            meta = runtime_info.get("meta") if isinstance(runtime_info, Mapping) else None
+            if isinstance(meta, Mapping) and meta.get("nemotron_voicechat_dummy_profile") is True:
+                continue
+            codes = self._codes_from_runtime_info(runtime_info)
+            if codes is None:
+                if input_ids is not None and input_ids.numel() > 0 and num_req == 1:
+                    codes = input_ids.reshape(-1)
```
linyueqian: [suggestion] This fallback treats `input_ids` as codec codes, but for this pipeline `talker2code2wav_token_only` builds `prompt_token_ids = [0] * n_frames` as pure placeholders, which the module docstring also states. So if the connector payload is ever missing, this decodes an all-zero code stack into audio instead of failing. `validate_code_stack` catches most cases through the divisibility check, but when `n_frames % 31 == 0` it passes and you get plausible-sounding garbage.

The inherited shape makes sense for `PersonaPlexCode2Wav`, where `input_ids` really do carry codebook-major codes. Here it looks safer to drop the fallback and let the missing payload be an explicit error.

recipes/NVIDIA/NemotronLabs-VoiceChat.md
```diff
@@ -0,0 +1,104 @@
...
+```
+
+The reply text should read as a coherent spoken-style answer to the question in
+the input WAV, and the WAV should transcribe to (approximately) the same text
+with any ASR model.
+
+#### Notes
+
+- Memory usage: the shipped yaml runs all three stages on one GPU
+  (`gpu_memory_utilization` 0.62 / 0.15 / 0.10); peak usage is dominated by the
```
linyueqian: [suggestion] These numbers have drifted from the shipped yaml, which uses 0.62 / 0.12 / 0.06 (lines 17, 41, 67 of `vllm_omni/deploy/nemotron_labs_voicechat.yaml`).

While here, it may be worth stating the fp32 memory floor explicitly. The fp32 thinker holds roughly 43 GB of weights (9B backbone, plus a 587M `embed_tokens`, plus a 587M `function_head`, plus the 0.6B conformer), so 48 GB cards cannot run the default configuration at all and need the documented bf16 option. That is useful to know before someone sizes a box from the '1x H100 80GB' heading.

recipes/NVIDIA/NemotronLabs-VoiceChat.md
```diff
@@ -0,0 +1,104 @@
...
+  fp32 thinker.
+- Key flags: sampling is greedy end to end. The thinker is frame-locked —
+  `max_tokens` equals the acoustic frame count with `ignore_eos=True`. Do NOT
+  set `min_tokens` on the thinker: the tokenizer's EOS token is also the
+  frame-locked PAD/silence token, so masking it forces the model to babble
+  instead of pausing.
+- The talker's `max_tokens` is 16383 (its stage prompt is one placeholder
+  token, and the stage context is 16384).
+- Known limitations: batch=1 only; offline single turn (no online/duplex
```
linyueqian, line 112: [question] On the duplex follow-up, is the plan to wire this into `vllm_omni/experimental/fullduplex/` the way PersonaPlex does, with `duplex_control_enabled` plus `duplex_runtime_extension` plus `duplex_serving_adapter` on the pipeline?

Asking because `fullduplex/DESIGN.md` says under 'Remaining Architecture Work' that adding a second native model should first introduce the versioned plugin descriptor and reject serving/engine plugin mismatches explicitly. This model would be the third native duplex consumer after minicpmo45 and personaplex, so that descriptor may be worth landing as part of the same effort rather than adding a third ad hoc pair of fields.

Two things in the current offline design will need unwinding first, and both look cheaper to plan for now than to retrofit. `_init_session` runs the whole Conformer once and caches `audio_embeds[N, H]`, so streaming needs the `cache_last_channel` / `cache_last_time` path (the vendored encoder still accepts those arguments, so the capability survived the dead-code pass). And the talker's `max_num_seqs == 1` guard is load bearing rather than defensive: `set_init_inputs()` writes `self._init_input_cache` on the module, so the warmup buffers are not per request, which blocks the multi-session admission the duplex control plane assumes.

linyueqian (changes requested):
Ran this end to end on the real checkpoint and it works: the reply is correct and on topic, and ASR of the generated WAV matches the emitted text channel exactly (WER 0.0 / CER 0.0 with whisper medium.en on a 16 s fixture). The output was byte identical across an H200 NVL and an RTX A6000 at seed 0, so the pipeline is deterministic across architectures. The 29 shipped unit tests pass. The staging work is clean: arch-keyed entries in the shared plumbing follow the existing convention, and the vendored tree carries per-file upstream paths plus a no-nemo-import guard test.

One blocking issue. `compute_acoustic_frame_count()` computes the subsampling output length with the non-causal padding formula, but this checkpoint sets `encoder.causal_downsampling: true`, so the real `ConvSubsampling` uses `all_paddings = (kernel-1) + (stride-1) = 3` while the helper assumes 2. I compared the helper against the vendored perception module over 12 durations and 9 of 12 disagree, always short by one frame, including the 250760-sample acceptance fixture (helper 196, perception 197). Because the example passes `nvc_expected_frames`, this is a hard failure rather than a silent one: an ordinary 4.45 s question (71271 samples) dies at prefill with `NemotronVoiceChat frame-count mismatch: the request was sized for 56 acoustic frames but perception produced 57`. The golden shape test cannot catch it because its `_STT_CFG` omits `causal_downsampling`, so it compares the helper against itself.

One note on scope, not a change request. Running with the frame-locked timeline made the offline mode's main sharp edge visible: the acoustic channel trails the text channel, so a reply is silently cut off if the input WAV does not carry enough trailing silence for the tail to finish. On an 8 s fixture the answer stopped mid-sentence (WER 0.52) with no error; padding the same question to 16 s gave WER 0.0. Worth a sentence in the recipe.

Everything below is inline. Caveat on my numbers: no 80 GB card was available to me, so I ran the bf16 thinker that the yaml documents rather than the fp32 default. The fp32 parity claim is untested on my side.

vllm_omni/model_executor/models/nemotron_voicechat/nemotron_voicechat_thinker.py
yuekaizhang: Confirmed and fixed in 8b54ebd1. I reproduced your table against the vendored `AudioPerceptionModule` built from the real config (7/12 of my durations mismatched, always short by one), and the helper now reads `encoder.causal_downsampling` and uses `(kernel-1)+(stride-1)` in causal mode — after the fix all 16 probe durations match the real module, and your 71271-sample repro runs end to end (sized 57, perception 57).

Root cause of why this survived our parity gate: the acceptance fixture is actually **249,734** samples — the 250,760 in the NeMo reference dump is the loader's padded batch width (`input_signal_lens` is what perception consumes). 249,734 happens to be a duration where both formulas agree, so the fixture never tripped the check.

tests/model_executor/models/test_nemotron_voicechat_shapes.py
yuekaizhang: Fixed in 8b54ebd1: `_STT_CFG` now sets `causal_downsampling: True` like the checkpoint, and the goldens were re-verified against the real perception module.

On the (250760, 196) reconciliation: both numbers were real but mispaired. The NeMo dump's 250,760 is the loader's padded batch width; perception consumed the true 249,734-sample length, which yields 196. Under the true geometry 250,760 genuinely yields 197. The golden table now pins (249734, 196) as the fixture row, keeps (250760, 197) as a regression row, and includes your 71271 repro. I also added the geometry cross-check you suggested: a test that builds the vendored `ConvSubsampling` (causal and non-causal) and asserts the helper against its actual `_left_padding + _right_padding` via `calc_length`, so the arithmetic can no longer drift from the module it mirrors.

vllm_omni/model_executor/models/nemotron_voicechat/configuration_nemotron_voicechat.py
yuekaizhang: Fixed in 8b54ebd1 — dropped `trust_remote_code` at all three sites (backbone AutoConfig, the thinker's AutoTokenizer, and the example's). Verified with transformers in the shipped venv: hub config, hub tokenizer, and a bare directory containing only `config.json` all load with `trust_remote_code=False` (`NemotronHConfig`, hidden 4480 / 56 layers).

examples/offline_inference/nemotron_voicechat/end2end.py
yuekaizhang: Both surfaced in 8b54ebd1. The frame-count failure is fixed at the helper (thread above). For the silent truncation: the example docstring and the recipe now document that the reply budget is the input duration and recommend generous trailing silence, and the script implements the cheap guard you suggested — it warns when the text channel emits non-PAD tokens within the last 12 frames (~1 s). Thanks for the paired 8 s / 16 s WER measurement; it made the failure mode easy to describe concretely.

vllm_omni/model_executor/models/nemotron_voicechat/nemotron_voicechat_thinker.py
yuekaizhang: Done in 8b54ebd1: the decode frame index is now derived from `_omni_num_computed_tokens - prompt_len` (position-addressed, idempotent under preemption/recompute), with the call counter kept only as a fallback when the runner does not inject the key, plus a guard rejecting a derived index < 1. Fixture parity is unchanged (196/196, bit-identical WAV).

vllm_omni/model_executor/models/nemotron_voicechat/nemotron_voicechat_thinker.py
yuekaizhang: Done in 8b54ebd1: the cumulative `nvc_function_tokens` accumulation is now gated behind `NEMOTRON_VOICECHAT_DEBUG_FUNCTION_TIMELINE=1` (default off — no per-step host sync or growing concat); the load-bearing `nvc_prev_function_token` path is untouched. You are right that the best-effort dump rarely fires; the env-gated timeline is a debug aid for parity work against the NeMo dumps rather than a user-facing output.

vllm_omni/model_executor/models/nemotron_voicechat/nemotron_voicechat_code2wav.py
yuekaizhang: Done in 8b54ebd1: the fallback is removed — a missing connector payload now raises with an explicit message instead of decoding the all-zero placeholder prompt. Agreed on the failure mode: with `n_frames % 31 == 0` the divisibility check passes and the result is plausible-sounding garbage, which is the worst kind.

recipes/NVIDIA/NemotronLabs-VoiceChat.md
yuekaizhang: Fixed in 8b54ebd1: the recipe now quotes the yaml's actual 0.62 / 0.12 / 0.06 and states the fp32 floor explicitly (~43 GB of thinker weights: 9B backbone + 587M embed_tokens + 587M function_head + 0.6B Conformer), pointing 48 GB cards at the documented bf16 option.

recipes/NVIDIA/NemotronLabs-VoiceChat.md
yuekaizhang: Yes — the plan is to conform to the `fullduplex/` framework rather than add a third ad-hoc pair of pipeline fields, and I agree the versioned plugin descriptor from DESIGN.md's "Remaining Architecture Work" should land with (or before) that effort instead of after a third integration.

Both unwinding points you identified are accurate, and thanks for checking that the streaming capability survived the dead-code pass:
- streaming perception will move from the one-shot `audio_embeds[N, H]` cache to the `cache_last_channel` / `cache_last_time` incremental path (deliberately preserved in the vendored encoder);
- the talker's `set_init_inputs()` writing `_init_input_cache` on the module makes the `max_num_seqs == 1` guard load-bearing, as you say — that warmup state needs to become per-session before the duplex control plane's multi-session admission can work.

Both are tracked under the online/duplex follow-up and out of scope for this offline PR.

yuekaizhang:
Thanks @linyueqian for the thorough review — every finding was verified and addressed in 8b54ebd1 (individual threads replied inline).

Highlights:
- The blocking causal-padding bug is confirmed and fixed; the helper now matches the real vendored perception on 16/16 probe durations, and your 71271-sample repro completes end to end. The reason our parity gate never caught it: the acceptance fixture is really 249,734 samples (the NeMo dump's 250,760 is the loader's padded batch width), a duration where both formulas happen to agree. The golden test now sets `causal_downsampling`, pins re-verified pairs including (250760, 197), and cross-checks the helper against the vendored `ConvSubsampling` geometry directly.
- `trust_remote_code` dropped at all three sites (verified against a bare config.json-only directory), the silent-truncation edge is documented in the example + recipe with a runtime warning over the last 12 frames, the decode frame index is now position-addressed, the debug function timeline is env-gated, the code2wav placeholder fallback is a hard error, and the recipe memory numbers are synced with the yaml plus the fp32 ~43 GB floor.

Re-verified after the changes: 33 CPU tests pass; fp32 GPU e2e on the fixture still matches the NeMo reference 196/196 tokens with a bit-identical WAV (691,532 B).

Sy0307:
I found non-causal padding bug before and now it is fixed. No blocker bug now. But can you support streaming handler ability for thinker stage? Or we support it in follow-up PR? 

Plz share your thought. @linyueqian

linyueqian:
i am ok with both options. depends on yuekai's bandwidth

yuekaizhang:
@Sy0307 Good timing — we'll do it in this PR rather than deferring.

Streaming handoff is a natural fit here and the framework already has the machinery: `qwen3_omni` ships an AR→AR `thinker2talker_async_chunk` processor, and step_audio2 / audex / higgs_audio_v3 / glm_tts all use the same `async_chunk_process_next_stage_input_func` dispatch. The plan:

1. **thinker → talker**: an async-chunk processor that streams the frame-locked text timeline incrementally, plus talker-session support for appending to the timeline buffer instead of receiving it whole at request creation.
2. **talker → code2wav**: chunked RVQ-VAE decode with left-context overlap. The codec's receptive field is tiny (n_fft=16, hop=4, near-causal) and the NeMo streaming pipeline already decodes it incrementally, so seam quality should be manageable; acceptance for the streamed path will be ASR/WER + seam checks rather than bit-exactness (the offline batch path keeps its bit-exact parity gate).
3. TTFP measurement to quantify the win over the batch handoff.

The offline batch path stays the default and keeps its 196/196 parity acceptance; streaming will be opt-in via the deploy yaml. Will push and re-request review when it's ready.

yuekaizhang:
@Sy0307 Streaming handler support for the thinker stage is in — landed as abebff4e, delivered in this PR as promised.

**What it does**: with the new opt-in deploy yaml (`vllm_omni/deploy/nemotron_labs_voicechat_streaming.yaml`, `async_chunk: true`), the three stages overlap instead of running serially. The thinker streams its cumulative frame-locked text timeline to the talker per decode step; the talker streams prompt-trimmed code stacks to code2wav every `codec_chunk_frames` (default 13 frames ≈ 1 s of audio); code2wav follows the NeMo `decode_one_audio_step` recipe — re-decode the received prefix, emit only the new samples — so the stream is seam-free by construction, no crossfade needed.

**Measured on 1x H100 (fp32, the parity fixture)**:
- First audio chunk at **~5 s** vs **~13 s** for the serial pipeline (≈60% TTFP reduction).
- Text channel: 196/196 tokens, identical to the batch path.
- Streamed audio: same sample count (345,744), **99.98% of samples bit-equal** to the one-shot batch decode; boundary diff max 3.3e-3 / mean 3.4e-8 (the codec's slight non-causality at chunk edges — same property as NeMo's own incremental decoding). Whisper transcription is verbatim-identical to the batch WAV.
- The non-streaming yaml is untouched and remains the bit-exact acceptance path: 36 CPU tests + 196/196 + bit-identical WAV re-verified after the change.

Run it with:
```bash
python examples/offline_inference/nemotron_voicechat/end2end.py \
    --checkpoint <ckpt> --wav <wav> \
    --stage-configs-path vllm_omni/deploy/nemotron_labs_voicechat_streaming.yaml
```

vllm_omni/model_executor/stage_input_processors/nemotron_voicechat.py
```diff
@@ -0,0 +1,355 @@
...
+    timeline is a few hundred ints, so cumulative re-ship is cheap and keeps
+    the payload self-contained. ``meta.num_processed_tokens`` carries the
+    logical prompt length for the code2wav producer's prompt-region trim.
+    """
+    del multimodal_output, kwargs
+    request_id = request.external_req_id
+    prompt_ids = list(getattr(request, "prompt_token_ids", None) or [])
+    # vLLM prompt = logical prompt + 1 placeholder (acoustic frame 0).
+    logical_prompt_len = max(len(prompt_ids) - 1, 0)
+    generated = list(getattr(request, "output_token_ids", None) or [])
```
lishunyang12, line 126: [P1] Snapshot the token history before queueing this producer. `save_async()` stores the live mutable `Request` and invokes this callback later on its background thread. If that thread falls behind by two thinker steps, both queued callbacks read the same newest `output_token_ids`; the first advances `nvc_timeline_len` to the newest length and the rest return `None`. The talker then receives one wake for multiple new positions and is parked again after consuming one, leaving the remaining timeline positions stranded. Please snapshot the per-step history into `multimodal_output`/the queued task, or make one payload explicitly wake and drain every newly added position. A delayed-save-thread regression test should cover two thinker outputs queued before either callback runs.

vllm_omni/model_executor/stage_input_processors/nemotron_voicechat.py
```diff
@@ -0,0 +1,355 @@
...
+        return None
+    transfer_manager.request_payload[request_id] = {"nvc_timeline_len": len(timeline)}
+
+    return OmniPayloadStruct(
+        ids=IdsStruct(all=timeline),
+        meta=MetaStruct(
+            finished=torch.tensor(bool(is_finished), dtype=torch.bool),
+            num_processed_tokens=logical_prompt_len,
+            # The talker's vLLM prompt stays a single placeholder token.
+            next_stage_prompt_len=1,
```
lishunyang12, line 146: [P2] Bootstrap the logical prompt region in the first chunk. The async receiver gives a downstream request one scheduler execution per received chunk before parking it again; this one-token prefill only calls `_init_session` and performs no EAR-TTS timeline step. Each later thinker chunk therefore advances the talker by only one of the `logical_prompt_len - 1` PAD steps whose code rows are trimmed away. First audio arrives after roughly `logical_prompt_len + codec_chunk_frames` thinker chunks, not the documented ~13, so long inputs may see little or no overlap. Please prime/drain the prompt-region steps during initial prefill (or otherwise schedule them without requiring one upstream chunk per PAD), and add a scheduler-level TTFA test with a nontrivial prompt length.

lishunyang12 (changes requested):
The new async-chunk path has two independent scheduling issues that are not covered by the current producer unit tests.

vllm_omni/model_executor/stage_input_processors/nemotron_voicechat.py
yuekaizhang: Fixed in 5571f717 via your second option — one wake now explicitly drains every newly added position. The talker's engine tokens are pure CONTINUE/STOP placeholders and the codes ride the cumulative payload, so running multiple EAR-TTS steps inside one engine step is safe: whichever wake observes a coalesced chunk consumes ALL of its positions, so a lagging save thread can no longer strand timeline entries (the producer's dedupe-to-None stays as a payload optimization, not a correctness dependency). A zero-progress wake emits a plain CONTINUE and never fabricates a frame.

Your delayed-save-thread scenario is pinned as a unit test (`test_talker_drains_all_received_positions_per_wake`, FakeTTS-backed session): a chunk carrying TWO new positions is fully consumed by a single wake, followed by a zero-progress wake asserting no step runs and no codes change. The sync path is gated behind a session flag and keeps its verified one-step-per-wake cadence — re-verified bit-exact after the change (196/196 tokens, bit-identical WAV).

vllm_omni/model_executor/stage_input_processors/nemotron_voicechat.py
yuekaizhang: Fixed by the same drain mechanism in 5571f717: chunk 0 already carries the whole PAD prompt prefix, and the first decode wake now drains all `logical_prompt_len - 1` prompt-region steps in one go — they no longer cost one upstream chunk each. The unit test asserts exactly this (session advances t: 1 -> P on a single wake with only chunk 0's PAD timeline received).

One honest note on the TTFA number: on the acceptance fixture the measured first-audio time stays ~5 s, because the binding constraint has moved from chunk scheduling to the talker's own serial compute — it still has to run ~68 EAR-TTS steps (55 PAD + 13 frames, each ~70 ms in fp32 with CFG's two forwards) before the first codec chunk exists. Pre-fix that scheduling artifact happened to hide behind a numerically similar thinker-paced bound, which is why the regression wasn't visible on this fixture. The structural fix matters for long prompts and for any config where the thinker is slower than the talker; driving TTFA further down is the bf16/CUDA-graph performance follow-up already queued. A full scheduler-level TTFA test needs a live multi-stage engine, so the scheduling contract is pinned at the session level (drain semantics) rather than end-to-end; happy to add an e2e timing gate under tests/dfx/perf if you'd prefer.

linyueqian (approved):
All findings from my previous review are addressed at `701095b5a`. Approving.

What I verified this round:

- The blocking causal frame-count fix is correct. I re-derived all five golden pairs by hand, including the `(249734, 196)` fixture, `(250760, 197)`, and the `71271 -> 57` repro, and the new geometry test pins the helper against the vendored `ConvSubsampling`'s actual `_left_padding + _right_padding`, which is exactly the drift that caused the original bug. The padded-batch-width explanation for the old `(250760, 196)` mispairing is convincing.
- The position-derived decode frame index checks out: `_omni_num_computed_tokens` at the first decode equals the vLLM prompt length (`logical_prompt + 1`), so `frame_idx` starts at 1 as the timeline contract requires, and the path is now idempotent under recompute.
- `trust_remote_code` removal at all three sites, the truncation doc plus 12-frame tail warning, the env-gated debug timeline, the code2wav placeholder-fallback removal, and the recipe sync are all in place.

I also reviewed the async-chunk streaming addition and the drain fix as new code:

- The drain-per-wake design is robust to coalesced, duplicated, and reordered chunks: timeline adoption is monotonic, `upstream_finished` never clears, and a zero-progress wake can only occur with the received timeline fully consumed, so it can never fabricate or strand a frame.
- No double-shipping in async mode: `should_accumulate_full_payload_output` structurally disables the sync full-payload flush when `async_chunk` is set, and the config layer swaps in the async producers, consistent with the cosyvoice3/glm_tts pattern.
- The streaming yaml's statement that prefix decodes are not bit-exact matches the reported 99.98% sample-equal result, and the unit coverage (drain semantics, cadence, left-context, replace keys) pins the parts that can regress silently.

[Low] One stale docstring, fine as a follow-up: `talker2code2wav_full_payload` in `vllm_omni/model_executor/stage_input_processors/nemotron_voicechat.py` still says the prompt-region rows "were never emitted by the talker ... so no extra trim is needed here", but the code trims at `audio[prompt_len - 1:]` and `test_full_payload_trims_prompt_region_rows` asserts that trim. Worth correcting so the row layout is not misdescribed.

For the record: the fork head runs build/pre-commit/DCO only, so the 38 CPU tests and the two GPU e2e results are as reported in the commit messages; I did not re-execute them this round. The TTFA note (talker serial compute now the bound, bf16/CUDA-graph queued) is a reasonable scope cut for this PR.

yuekaizhang:
@lishunyang12  Could you help review it again? Thanks!

linyueqian:
please resolve conflicts

PR #5844, opened 2026-08-06
[ROCm] [CI] Migrate READY test to MI300 CI Cluster

tjtanaa:

## Purpose

This is a part of the migration pipeline.

## Bugfixes

<details>
    <summary> bugfix </summary>
    
## Summary

Fix AR-Diffusion paged attention on ROCm by selecting FlashAttention backends in this order:

1. AITER `flash_attn_varlen_func`
2. Upstream `flash_attn` as a fallback when AITER is unavailable

The ROCm path gathers paged KV blocks into packed, contiguous tensors before invoking the selected varlen FlashAttention implementation.

## Problem

The existing implementation imports `vllm.vllm_flash_attn` unconditionally for GPU tensors. That package requires vLLM's CUDA FlashAttention extensions and fails to import on ROCm:

```text
ImportError: vllm.vllm_flash_attn requires the CUDA flash attention extensions
(_vllm_fa2_C or _vllm_fa3_C). On ROCm, use upstream flash_attn.
```

ROCm provides both AITER FlashAttention and upstream `flash_attn`, but their paged-attention behavior introduces two additional constraints:

- The installed AITER version exposes a `block_table` argument, but its paged-kernel generator fails internally with an `UnboundLocalError`.
- Upstream ROCm `flash_attn` requires paged KV cache block sizes divisible by 128, while AR-Diffusion intentionally uses frame-aligned 16-token blocks.

Therefore, neither ROCm implementation can directly consume the existing 16-token paged KV layout in this environment.

## Solution

The ROCm compatibility path resolves the block table before calling FlashAttention:

```text
Paged K/V cache + block_table
              |
              v
Gather referenced blocks in logical sequence order
              |
              v
Packed contiguous K/V tensors
              |
              v
AITER FlashAttention, or upstream flash_attn fallback
```

This preserves AITER as the preferred ROCm backend while avoiding its broken `block_table` kernel. Because the selected FlashAttention function receives ordinary packed K/V tensors without a block table, the upstream implementation's 128-token paged-block restriction also does not apply.

CUDA continues to use `vllm.vllm_flash_attn` and its native paged-attention API.

## Tradeoffs

The compatibility path adds an on-device gather and temporary packed K/V tensors. It may therefore use more temporary memory and be slower than a working native paged-attention kernel. It nevertheless keeps attention computation in FlashAttention and provides correct behavior for AR-Diffusion's 16-token cache blocks on ROCm.

## Testing

The originally failing ROCm case passes:

```bash
pytest -svvvv tests/diffusion/ar_diffusion/test_paged_attention.py::test_paged_attention_matches_dense_reference_gpu[False-0-1]
```

The complete paged-attention test module also passes:

```text
28 passed
```

## ROCm BF16 numerical tolerance

While validating the ROCm FlashAttention path, the following comparison initially failed:

```bash
pytest -svvv tests/diffusion/attention/test_flash_attn.py::test_fa_vs_sdpa
```

The original test used absolute-only thresholds:

```python
assert max_diff < 0.01
assert mean_diff < 0.001
```

On ROCm with the AITER BF16 kernel, the observed values were:

```text
Maximum absolute difference: 0.03125
Mean absolute difference:    approximately 0.00334
```

### Triage

This is deterministic BF16 numerical variance between different attention implementations, not a random ROCm bit flip or an attention/masking bug:

- Repeated AITER executions were bit-for-bit identical.
- The largest observed test mismatch was `-2.65625` versus `-2.6875`, which is approximately 1.2% relative error despite its `0.03125` absolute difference.
- Compared with an FP32 dense reference, AITER had approximately `0.00334` mean absolute error.
- Upstream ROCm `flash_attn` also exceeded the old absolute-only `0.01` maximum threshold, although it was closer to SDPA than AITER.
- A 20-seed comparison against the FP32 reference found a worst relative error of approximately 1.32% for reference values whose magnitude exceeded `0.1`.
- All 20 seeds satisfied the selected combined absolute and relative tolerance.

The old maximum-error assertion treated the same absolute difference equally near zero and at output magnitudes above 2.0. A combined tolerance better reflects BF16 rounding behavior by allowing a small relative difference for larger values while retaining an absolute bound near zero.

### Resolution

The comparison now uses the following tolerance on ROCm only:

```python
torch.testing.assert_close(
    output_fa_valid,
    output_sdpa_valid,
    rtol=1e-2,
    atol=1.5e-2,
)
```

Other platforms retain the existing stricter thresholds. This keeps the relaxation narrowly scoped to the deterministic ROCm BF16 behavior observed during testing.

The focused regression passes with the updated assertion. The related piecewise and packed-attention unit tests also pass. A full-module run was attempted but stopped during an unrelated multi-minute, first-time AITER dense-kernel JIT compilation; it had not reported a test failure before being interrupted.

## Duplex import-boundary failures in the combined ROCm suite

The combined entrypoints and engine test suite initially reported four failures:

```text
FAILED tests/engine/test_duplex_import_boundary.py::test_stable_engine_imports_do_not_load_experimental_duplex
FAILED tests/engine/test_duplex_import_boundary.py::test_stable_engine_does_not_expose_duplex_contract_modules
FAILED tests/engine/test_duplex_import_boundary.py::test_stable_outputs_do_not_declare_duplex_decision_field
FAILED tests/engine/test_duplex_import_boundary.py::test_stable_model_executor_does_not_expose_duplex_helper_module
```

These failures appeared to indicate that experimental full-duplex modules or contracts had leaked into stable packages. However, all four tests passed when run independently, and the prohibited stable files, exports, and dataclass field were absent.

### Root cause

The four boundary tests execute their assertions in isolated child Python processes. During the combined suite, an earlier test left conflicting GPU visibility variables in the parent pytest environment:

```text
HIP_VISIBLE_DEVICES=0
CUDA_VISIBLE_DEVICES=0,1
```

The child processes inherited both values. Importing `vllm_omni` caused vLLM to initialize its ROCm platform support, which rejected the inconsistent device configuration:

```text
ValueError: Inconsistent GPU visibility env vars:
HIP_VISIBLE_DEVICES='0' vs CUDA_VISIBLE_DEVICES='0,1'.
Please set only one, or ensure they match.
```

Consequently, each child process exited while importing vLLM and never reached its duplex-boundary assertion. The failure names were therefore misleading: this was environment contamination between tests, not a stable/experimental import-boundary violation.

### Solution

The import-boundary subprocess helper now copies the parent environment and removes both GPU visibility variables before starting the child process:

```python
env = os.environ.copy()
env.pop("CUDA_VISIBLE_DEVICES", None)
env.pop("HIP_VISIBLE_DEVICES", None)

subprocess.run(
    [sys.executable, "-c", script],
    cwd=REPO_ROOT,
    env=env,
    ...,
)
```

These tests inspect Python package structure and imports and do not execute GPU operations, so GPU selection is irrelevant to the behavior under test. Removing these variables makes the subprocess isolation explicit and prevents unrelated GPU configuration from masking the actual boundary result.

A regression test deliberately sets the conflicting values in the pytest process and verifies that an isolated `vllm_omni` import still succeeds.

### Verification

The boundary module passes, including the new regression:

```text
7 passed
```

The complete command that originally exposed the ordering-dependent failure also passes:

```bash
pytest -q tests/entrypoints tests/engine \
  -m 'core_model and cpu' \
  --cov=vllm_omni \
  --cov-report=xml
```

```text
1560 passed, 2 skipped, 125 deselected, 20 warnings
```

## MiniCPM-o realtime duplex driver `--help` timeout

The duplex-driver test suite initially failed while checking that the multi-session script was directly executable:

```text
FAILED tests/e2e/online_serving/test_minicpmo_realtime_duplex_drivers.py::test_realtime_duplex_multi_session_script_is_directly_executable
subprocess.TimeoutExpired: Command '['/usr/bin/python3.12',
'tests/e2e/online_serving/run_minicpmo_realtime_duplex_multi_session.py',
'--help']' timed out after 20 seconds
```

### Root cause

The multi-session script imported the full scenario harness at module initialization:

```python
from minicpmo_realtime_duplex_scenarios import (
    _ref_audio_data_url,
    _url_with_model,
    run_demo,
)
```

The scenario harness imports the experimental `vllm_omni` realtime client. Importing `vllm_omni` initializes vLLM and the ROCm platform stack, which can take more than the test's 20-second timeout on a busy or cold ROCm host.

Python performs module imports before entering `main()` and before `argparse` processes `--help`. As a result, a request that only needed static CLI help paid the full runtime initialization cost and sometimes timed out without reaching argument parsing.

This was a startup-order issue rather than a problem with the multi-session runtime or its CLI definition.

### Solution

The scenario harness is now loaded through a cached lazy-import helper:

```python
@lru_cache(maxsize=1)
def _scenario_module() -> ModuleType:
    import minicpmo_realtime_duplex_scenarios

    return minicpmo_realtime_duplex_scenarios
```

Small wrappers preserve the existing module-level interfaces while deferring the expensive import until runtime functionality is actually requested:

```python
def _ref_audio_data_url(path: str | None) -> str | None:
    return _scenario_module()._ref_audio_data_url(path)


def _url_with_model(*args, **kwargs) -> str:
    return _scenario_module()._url_with_model(*args, **kwargs)


async def run_demo(*args, **kwargs) -> dict[str, object]:
    return await _scenario_module().run_demo(*args, **kwargs)
```

Normal multi-session execution still loads and uses the same scenario implementation. The only behavioral change is that lightweight CLI operations such as `--help` no longer initialize vLLM, ROCm, or the realtime client.

### Verification

Direct help startup decreased from more than 20 seconds to approximately 0.09 seconds:

```bash
python tests/e2e/online_serving/run_minicpmo_realtime_duplex_multi_session.py --help
```

The previously failing focused test passes, and the complete duplex-driver module also passes:

```text
63 passed, 15 warnings
```

</details>    
    

## Test Plan

**vLLM Version:**

**vLLM-Omni Commit:**

## Test Result

`test-ready`: https://buildkite.com/vllm/vllm-omni-amd-ci/builds/10263/list


hsliuustc0106:
ready to test now?

tjtanaa:
> ready to test now?

yes. Ready to test the Ready test.yaml now.

tjtanaa:
Closing this PR in favor of  https://github.com/vllm-project/vllm-omni/pull/5886

PR #5845, opened 2026-08-06
[CI/Build] Resolve hardware-nested perf baselines per concurrency sweep

yenuo26:

## Purpose

Perf result JSON previously persisted the entire test-config `baseline` blob as-is. After migrating baselines to a hardware-nested shape (`{"H100": {...}, "B200": {...}}`), each concurrency / request-rate sweep step still wrote **all** sweep values (lists / multi-step maps) into that step's result file.

This PR:

1. Migrates DFX perf JSON baselines to the hardware-nested format.
2. Resolves baselines **per sweep step** when writing results: keep **every hardware bucket**, but reduce each metric to the value for the current concurrency / request-rate (list index or scalar).
3. Requires hardware-nested baselines only (no flat top-level metric map).
4. Aligns Boogu configs to sweep-aligned arrays (same style as other models).
5. Adds unit coverage in `tests/dfx/perf/tests/test_runner_metadata.py`.

Example: for `max-concurrency=[1,4,8,16,32]`, the concurrency-32 result stores:

```json
"baseline": {
  "H100": {"mean_ttft_ms": <c32>, ...},
  "B200": {"mean_ttft_ms": <c32>, ...}
}
```

not the full five-element arrays.

## Test Plan

**vLLM Version:** CI image / env pin (same as Buildkite CUDA nightly when this lands)

**vLLM-Omni Commit:** `0fd5ee8f` (branch `benchmark-baseline`)

### UT
```bash
# L1: baseline resolution + runner metadata
pytest -sv tests/dfx/perf/tests/ -m "core_model and cpu"
```

### Perf
```
pytest -s -v tests/dfx/perf/scripts/run_benchmark.py --test-config-file tests/dfx/perf/tests/test_higgs_audio_v3.json
pytest -s -v tests/dfx/perf/scripts/run_diffusion_benchmark.py --test-config-file tests/dfx/perf/tests/ test_qwen_image_layered_vllm_omni.json
```

### UT
9df29a80-13f0-48ef-9c26-27014a186193

### Perf
be1c9d6cf8411a616ed79083f68cf347
5d0802717935162600c2494b3918887e
464b6c3bc0addfb1fa083a47035c9ced
ea0ed2f656628f3e6683e335051d1fe1


amy-why-3459:
## Summary

This PR builds on the hardware-nested baseline shape (`{"H100": {...}, "B200": {...}}`) by resolving each metric to the current concurrency / request-rate sweep step when writing result JSON. That avoids storing full sweep lists or multi-step maps inside every step's result file. Boogu configs are converted to sweep-aligned arrays, and unit coverage in `test_runner_metadata.py` looks solid.

**Verdict: Approve with nits.** The design is correct and more complete than “nest only, resolve never.” A few robustness / documentation points are worth addressing before or right after merge.

---

## 1. Core design — LGTM

`resolve_baseline_for_sweep`:

- Keeps **every** hardware bucket (`H100` / `B200` / …).
- Reduces each metric to the list element at `sweep_index`, or leaves scalars as-is.
- Rejects flat top-level metric maps and per-metric `{concurrency: value}` dicts.

Wiring into `run_benchmark` and diffusion `_build_run_params` / `_iter_sweep_runs` matches the PR description.

---

## 2. (P2) `is_hardware_nested_baseline` relies on a brittle metric-name heuristic

```python
_BASELINE_METRIC_HINTS = ("throughput", "latency", "mean_", "peak_", "p50", ...)
return any(any(hint in key.lower() for hint in _BASELINE_METRIC_HINTS) for key in value)
```

If a hardware bucket only contains custom metric names that do **not** match these substrings, the structure is still hardware-nested but will be rejected with `ValueError`.

Suggestions:

- Document the metric naming convention clearly; and/or
- Detect nesting by structure (all top-level values are dicts) plus a rule that top-level keys must **not** look like metrics, or an explicit hardware-label allowlist / denylist.

---

## 3. (P2) Boogu dict → list conversion must stay aligned with the concurrency sweep order

```diff
- "throughput_qps": {"1": 0.0303, "2": 0.0302, "4": 0.0302},
+ "throughput_qps": [0.0303, 0.0302, 0.0302],
```

This is only equivalent if `max-concurrency` is exactly `[1, 2, 4]` in the same order. Please confirm in the PR (or with a short JSON comment) that the array order is pinned to the concurrency list so a future edit to one does not silently desync the other.

---

## 4. (P2) `sweep_index=None` + list baselines will hard-fail

Omni `run_benchmark` only passes `sweep_index` inside the QPS / concurrency loops. A call with hardware-nested list metrics and `sweep_index=None` raises:

`ValueError: list baseline requires sweep_index`

Diffusion’s non-sweep path defaults to `sweep_index=0`, which is safer. Please confirm there is no omni single-point entry that still passes list baselines without an index; if there is, default to `0` or require an index at the call site.

---

## 5. Nits

| Item | Note |
|---|---|
| PR description | Remove leftover `PLEASE FILL IN THE PR DESCRIPTION HERE.` |
| Docs examples | Docs use `B200`; many existing configs use `A3` — call out that hardware keys must match the runtime resource label |
| Breaking change | Flat baselines are no longer accepted — good, but a short migration note / CHANGELOG blurb would help |
| Tests | `test_runner_metadata.py` coverage for resolve / reject / diffusion sweep is sufficient |

yenuo26:
> ## Summary
> This PR builds on the hardware-nested baseline shape (`{"H100": {...}, "B200": {...}}`) by resolving each metric to the current concurrency / request-rate sweep step when writing result JSON. That avoids storing full sweep lists or multi-step maps inside every step's result file. Boogu configs are converted to sweep-aligned arrays, and unit coverage in `test_runner_metadata.py` looks solid.
> 
> **Verdict: Approve with nits.** The design is correct and more complete than “nest only, resolve never.” A few robustness / documentation points are worth addressing before or right after merge.
> 
> ## 1. Core design — LGTM
> `resolve_baseline_for_sweep`:
> 
> * Keeps **every** hardware bucket (`H100` / `B200` / …).
> * Reduces each metric to the list element at `sweep_index`, or leaves scalars as-is.
> * Rejects flat top-level metric maps and per-metric `{concurrency: value}` dicts.
> 
> Wiring into `run_benchmark` and diffusion `_build_run_params` / `_iter_sweep_runs` matches the PR description.
> 
> ## 2. (P2) `is_hardware_nested_baseline` relies on a brittle metric-name heuristic
> ```python
> _BASELINE_METRIC_HINTS = ("throughput", "latency", "mean_", "peak_", "p50", ...)
> return any(any(hint in key.lower() for hint in _BASELINE_METRIC_HINTS) for key in value)
> ```
> 
> If a hardware bucket only contains custom metric names that do **not** match these substrings, the structure is still hardware-nested but will be rejected with `ValueError`.
> 
> Suggestions:
> 
> * Document the metric naming convention clearly; and/or
> * Detect nesting by structure (all top-level values are dicts) plus a rule that top-level keys must **not** look like metrics, or an explicit hardware-label allowlist / denylist.
> 
> ## 3. (P2) Boogu dict → list conversion must stay aligned with the concurrency sweep order
> ```diff
> - "throughput_qps": {"1": 0.0303, "2": 0.0302, "4": 0.0302},
> + "throughput_qps": [0.0303, 0.0302, 0.0302],
> ```
> 
> This is only equivalent if `max-concurrency` is exactly `[1, 2, 4]` in the same order. Please confirm in the PR (or with a short JSON comment) that the array order is pinned to the concurrency list so a future edit to one does not silently desync the other.
> 
> ## 4. (P2) `sweep_index=None` + list baselines will hard-fail
> Omni `run_benchmark` only passes `sweep_index` inside the QPS / concurrency loops. A call with hardware-nested list metrics and `sweep_index=None` raises:
> 
> `ValueError: list baseline requires sweep_index`
> 
> Diffusion’s non-sweep path defaults to `sweep_index=0`, which is safer. Please confirm there is no omni single-point entry that still passes list baselines without an index; if there is, default to `0` or require an index at the call site.
> 
> ## 5. Nits
> Item	Note
> PR description	Remove leftover `PLEASE FILL IN THE PR DESCRIPTION HERE.`
> Docs examples	Docs use `B200`; many existing configs use `A3` — call out that hardware keys must match the runtime resource label
> Breaking change	Flat baselines are no longer accepted — good, but a short migration note / CHANGELOG blurb would help
> Tests	`test_runner_metadata.py` coverage for resolve / reject / diffusion sweep is sufficient

will fix 2

amy-why-3459 (approved):

PR #5846, opened 2026-08-06
[Bugfix] npu omni runner alignment

psv666:
Two NPU model-runner divergences from their GPU counterparts break MiniCPM-o 4.5 on the --no-async-chunk (full-payload) path. Both are omissions from the original port; the GPU runners have had the missing logic all along.

1. Connector init gated only on a hardcoded arch allowlist
npu_ar_model_runner.py / npu_generation_model_runner.py decide whether to call init_omni_connectors() purely from _OMNI_CONNECTOR_INIT_ARCHS. The GPU runners additionally accept any stage the deploy config has given a sender/receiver role:

if (model_arch in _OMNI_CONNECTOR_INIT_ARCHS
        or get_stage_connector_role(self.model_config) is not None):
MiniCPMO45OmniForConditionalGeneration and MiniCPMO45Code2Wav are not in the allowlist but are wired stage 1 → stage 2 in minicpmo_4_5.yaml. On NPU init_omni_connectors therefore never ran: _omni_connector was absent, the hasattr guard in execute_model skipped the whole recv/flush block, _should_accumulate_full_payload_output() returned False, and stage 2 starved forever. Streaming audio requests hung until the client timed out, with no server-side error.

2. requires_request_ids not forwarded
MiniCPMO45Code2Wav sets requires_request_ids = True. gpu_generation_model_runner honours it; the NPU runner never sets model_kwargs["request_ids"]. Code2Wav then falls back to scraping ids out of runtime_additional_information, where tts2code2wav_full_payload writes the literal string "None" when the producer request carries no id. Every concurrent request collapses onto the same state key:

RuntimeError: MiniCPMO45Code2WavBatchError
  {"reason": "duplicate_request_in_forward", "request_ids": ["None", "None"]}
→ EngineDeadError → API server dies
Single-request runs pass only because ["None"] has no duplicate and "None" is a non-empty string — the per-request streaming state was already keyed on garbage.

Verified on A3 NPU, tests/e2e/online_serving/test_minicpmo_4_5_expansion.py, both [default] (--no-async-chunk) and [async_chunk] groups green. Previously 3 of the [default] cases hung for the full 600 s client timeout.

Follow-up (not in this PR): tts2code2wav_full_payload should fall back to req_id the way the mixin's _resolve_external_req_id does, instead of emitting str(None).

3. Update baseline metrics for test_minicpmo_4_5_duplex_seed_tts

### DIFF
--- a/tests/dfx/perf/tests/test_minicpmo_4_5_duplex_seed_tts.json
+++ b/tests/dfx/perf/tests/test_minicpmo_4_5_duplex_seed_tts.json
@@ -45,19 +45,19 @@
         "percentile-metrics": "ttft,e2el,audio_ttfp,audio_rtf,audio_duration",
         "baseline": {
           "request_throughput": [
-            0.2415
+            0.1594
           ],
           "mean_ttft_ms": [
-            122.6731
+            274.164
           ],
           "mean_e2el_ms": [
-            4073.804
+            6250.7696
           ],
           "mean_audio_ttfp_ms": [
-            523.4768
+            994.422
           ],
           "mean_audio_rtf": [
-            0.28295
+            0.46163
           ]
         }
       }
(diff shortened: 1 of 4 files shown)


psv666:
@R2-Y @amy-why-3459 PTAL

R2-Y:
LGTM, Thanks

PR #5851, opened 2026-08-06
[Feature] Add Sol-Attn sparse attention backend for MiniMax H3

david6666666:
## Summary

This PR adds `SOL_ATTN` as a diffusion attention backend for MiniMax-H3. It
integrates NVLabs' training-free, on-the-fly block-sparse attention kernel,
keeps configurable early denoise steps and DiT layers dense, and supports the
packed-varlen H3 layout.

This is the Sol-Attn kernel integration only, not the full Sol-Engine stack.

- Project: [Sol-Attn](https://nvlabs.github.io/Sana/Sol-Attn/)
- Paper: [Sol-Attn: Accelerating Video Generation Inference via On-the-Fly Attention Sparsification](https://arxiv.org/abs/2607.24027)
- Upstream implementation: [NVlabs/Sana `sol-engine` branch](https://github.com/NVlabs/Sana/tree/sol-engine)

## What this PR adds

- `SolAttnBackend` / `SolAttnImpl` and `SOL_ATTN` registry/platform routing.
- Packed-varlen handling that excludes alignment-padding rows before invoking
  the sparse kernel.
- Typed `SolAttnSpec` configuration for `tau`, `thresh_type`, `sink_tokens`,
  `sink_start`, `dense_steps`, `dense_layers`, and `kv_splits`.
- Exact-KV sink-range clamping for short packed sequences.
- Platform-specific dense guards:
  - SM120 uses cuDNN for early dense steps and always-dense layers.
  - Other supported CUDA architectures use the FlashAttention varlen fallback.
- A resumable, isolated-process offline speed/quality sweep with SSIM, PSNR,
  and LPIPS reporting.
- Unit coverage for configuration parsing, dense-step/layer routing, packed
  lengths, SM120 fallback selection, and sweep case construction.

## Requirements

- CUDA
- BF16 activations
- `head_dim=128`
- The upstream `sol_attn` package (not bundled in vLLM-Omni)
- Ring sequence parallelism is not supported; use Ulysses with `ring_degree=1`

Install Sol-Attn:

```bash
pip install git+https://github.com/NVlabs/Sana.git@sol-engine#subdirectory=techniques/sparse_backends
```

### Recommended quality-first server configuration

```bash
export MODEL=/path/to/MiniMax-H3/FL2VA

CUDA_VISIBLE_DEVICES=0,1,2,3 \
VLLM_WORKER_MULTIPROC_METHOD=spawn \
VLLM_OMNI_VIDEO_SYNC_TIMEOUT=1800 \
vllm serve "$MODEL" \
  --omni \
  --trust-remote-code \
  --num-gpus 4 \
  --usp 4 \
  --ring 1 \
  --vae-patch-parallel-size 4 \
  --vae-parallel-mode tile \
  --vae-use-tiling \
  --diffusion-attention-config '{"default": {"backend": "SOL_ATTN", "sol_attn": {"tau": 1.0, "thresh_type": "diag", "sink_tokens": 951, "sink_start": 0, "dense_steps": 10, "dense_layers": "0,1", "kv_splits": 1}}}'
```

On SM120, no extra fallback flag is required: dense guard work automatically
uses cuDNN and the remaining eligible work uses the Sol-Attn sparse kernel.

The equivalent configuration dictionary for offline/programmatic use is:

```python
diffusion_attention_config = {
    "default": {
        "backend": "SOL_ATTN",
        "sol_attn": {
            "tau": 1.0,
            "thresh_type": "diag",
            "sink_tokens": 951,
            "sink_start": 0,
            "dense_steps": 10,
            "dense_layers": "0,1",
            "kv_splits": 1,
        },
    },
}
```

### Tuning guidance from the measured sweep

All presets below keep `dense_layers="0,1"`, `sink_tokens=951`,
`thresh_type=diag`, and `kv_splits=1`.

| Goal | `tau` | Dense steps | Median latency | Speedup | Visual quality |
| Quality-first default | 1.0 | 10 | 129.91 s | 1.097× | PASS |
| Balanced | 1.5 | 8 | 124.95 s | 1.140× | PASS |
| Fastest measured PASS | 1.0 | 5 | 122.75 s | 1.161× | PASS |
| Aggressive | 2.0 | 5 | 118.29 s | 1.205× | FAIL (LPIPS) |

`kv_splits="auto"` also passed with a 1.098× speedup. Fixed
`kv_splits=2` and `kv_splits=4` failed in this run and are not recommended on
this tested configuration.

### Test configuration

- Hardware: 4× NVIDIA RTX PRO 5000 Blackwell (SM120), 73,415 MiB per GPU
- Physical GPU order: `4,6,5,7`
- Driver / CUDA reported by `nvidia-smi`: 580.95.05 / 13.0
- Model: MiniMax-H3 `FL2VA`, BF16
- Parallelism: diffusion TP4, text-encoder TP4
- Output: 1344×768, 24 FPS, 5 seconds, 16:9
- Denoise: 20 requested steps (19 progress-bar updates)
- Flow shifts: video 12.0, audio 3.0
- Seed: 1101
- Timing: one warmup followed by three measured runs; reported latency is the
  median
- Visual quality: measured run 1 compared with the same-seed dense cuDNN output
- SSIM / PSNR: every decoded frame at native resolution
- LPIPS: up to 16 uniformly sampled frames resized to 256×256

Prompt:

> At night, while their owner sleeps, three cats march into a bedroom playing
> tiny brass instruments, freeze, and quietly march out.

### Main speed/quality result

| Configuration | Median (s) | Speedup | Latency reduction | SSIM | PSNR (dB) | LPIPS | Quality |
|---|---:|---:|---:|---:|---:|---:|---:|
| Dense cuDNN | 142.49 | 1.000× | 0.00% | 1.00000 | inf | 0.00000 | PASS |
| Sol-Attn recommended | 129.91 | 1.097× | 8.83% | 0.93344 | 33.28 | 0.08563 | PASS |
| Sol-Attn balanced | 124.95 | 1.140× | 12.31% | 0.91780 | 31.49 | 0.11771 | PASS |
| Sol-Attn fastest passing | 122.75 | 1.161× | 13.86% | 0.90985 | 30.58 | 0.11525 | PASS |
| Sol-Attn aggressive | 118.29 | 1.205× | 16.99% | 0.87586 | 27.00 | 0.22556 | FAIL |

Quality gates:

- SSIM >= 0.82
- PSNR >= 20.0 dB
- LPIPS <= 0.20

### Dense-guard speed/quality curve

Reducing `dense_steps` increases the fraction of work handled by the sparse
kernel. Five dense steps was the fastest point that passed all configured
quality gates.

| Dense steps | Median (s) | Speedup | SSIM | PSNR (dB) | LPIPS | Quality |
|---:|---:|---:|---:|---:|---:|---:|
| 20 | 142.73 | 0.998× | 0.95825 | 35.60 | 0.05441 | PASS |
| 15 | 136.73 | 1.042× | 0.95129 | 35.10 | 0.05834 | PASS |
| 8 | 127.13 | 1.121× | 0.93170 | 32.66 | 0.08395 | PASS |
| 5 | 122.75 | 1.161× | 0.90985 | 30.58 | 0.11525 | PASS |
| 0 | 115.35 | 1.235× | 0.75299 | 20.90 | 0.54201 | FAIL |

The full 21-case result table, methodology, and limitations are available in:

- [Rankings report](https://github.com/lishunyang12/vllm-omni-rankings/tree/main/scripts/minimax_h3_sol_attn_sm120)
- [Report attached to this PR branch](https://github.com/david6666666/vllm-omni/blob/feat/sol-attn-minimax-h3/benchmarks/diffusion/sol_attn_sm120_results.md)

## Reproduce the offline sweep

Install quality dependencies and make `ffmpeg` / `ffprobe` visible in `PATH`:

```bash
python -m pip install --no-deps lpips==0.1.4
export PATH=/path/to/ffmpeg/bin:$PATH
```

Run the full isolated-process sweep:

```bash
python benchmarks/diffusion/sol_attn_offline_sweep.py \
  --model /path/to/MiniMax-H3/FL2VA \
  --gpus 4,6,5,7 \
  --num-gpus 4 \
  --tensor-parallel-size 4 \
  --text-encoder-tp-size 4 \
  --suite full \
  --output-dir results/sol-sm120-offline-sweep \
  --lpips-device cpu
```

The command is resumable. To recompute only SSIM, PSNR, LPIPS, and the report
from existing videos:

```bash
python benchmarks/diffusion/sol_attn_offline_sweep.py \
  --output-dir results/sol-sm120-offline-sweep \
  --phase quality \
  --lpips-device cpu
```

Generated artifacts include `manifest.json`, per-case `timings.json` and MP4s,
`summary.csv`, `quality.csv`, `failures.json`, and `report.md`.

## Validation

- [x] End-to-end SM120 sweep completed for the dense baseline and 20 Sol-Attn
  configurations.
- [x] Same-seed full-video SSIM and PSNR completed.
- [x] LPIPS completed with `lpips==0.1.4` and AlexNet weights.
- [x] DCO passed for the pushed commit range.
- [x] Python 3.11 and Python 3.12 wheel builds passed.
- [x] Pre-commit passed.

## Limitations

- Visual quality currently covers one prompt and one seed. Timing uses three
  measured repeats, but changing the recommended defaults should receive a
  multi-seed confirmation.
- SSIM, PSNR, and LPIPS evaluate video frames only; audio quality is not
  measured by this sweep.
- `sol_kv_splits_2` and `sol_kv_splits_4` failed and are excluded from the
  reported table. `sol_kv_splits_auto` completed successfully.
- These numbers describe this exact MiniMax-H3 workload and 4× RTX PRO 5000
  SM120 system; they are not a cross-model or cross-hardware claim.

### CI States

Latest PR Test (Base): TBD
Latest PR Test (Extra): TBD


david6666666:
## 4× B300 verification (MiniMax-H3)

Verified on **4× NVIDIA B300 (sm_103)** with the four-GPU B300 profile from `recipes/MiniMaxAI/MiniMax-H3.md`: Ulysses-4, VPP4 native tile, no offload, regional compile.

**Setup:** 1344×768, 124 frames (5.17 s @ 24 FPS), 20 denoise steps, seed 1101, `lossless`. Server mode; one discard warmup, then a timed run. Dense baseline = `FLASH_ATTN` (CuTe FlashAttention-4, confirmed in the server log). Denoise-stage time is measured inside the H3 denoise loop; E2E is the HTTP client latency of `/v1/videos/sync`.

### Case 1: H3 t2va

**Prompt:** *At night, while their owner sleeps in a bedroom, three cats march in loudly playing tiny brass instruments, then abruptly file out.*

| Backend | Config | Denoise (timed) | Speedup | E2E | PSNR | SSIM |
|---------|--------|-----------------|---------|-----|------|------|
| `FLASH_ATTN` | — | 14.55 s | — | 19.71 s | — | — |
| `SOL_ATTN` | τ=1.0, dense_steps=10, layers=0,1 (default) | 14.25 s | **1.02×** | 18.91 s | **35.35 dB** | **0.956** |

PSNR/SSIM are computed with `ffmpeg`'s `psnr`/`ssim` filters against the `FLASH_ATTN` output at the same prompt/seed/shape.

### Parameter sweep (H3 t2va)

| Config | Parameters | Denoise | Speedup vs FA | PSNR | SSIM |
|--------|------------|---------|----------------|------|------|
| **Default (recommended)** | τ=1.0, dense_steps=10, layers=0,1 | 14.25 s | **1.02×** | **35.35 dB** | **0.956** |
| Fewer dense steps | τ=1.0, dense_steps=5, layers=0,1 | 14.13 s | 1.03× | 28.94 dB | 0.906 |
| Aggressive | τ=2.0, dense_steps=5, layers=0 | 12.22 s | 1.19× | 26.32 dB | 0.863 |

**Takeaway:** the **default config** is the right production starting point — it keeps fidelity well above the 31 dB bar (PSNR 35.35 dB / SSIM 0.956). Raising `tau` or cutting `dense_steps` buys speed fast (up to 1.19× denoise) but PSNR drops several dB per step of `tau`, so those knobs are for latency-bound runs with a lower quality bar.

### Notes on this hardware

- On B300 (`sm_103`) the upstream kernel routes to the **Triton** backend — the CuTe kernel paths in the upstream package currently cover `sm_90` / `sm_100` / `sm_120`. The default-config speedup on B300 is therefore modest (~1.02×). The CuTe path on Hopper / B200-class hardware is expected to show larger gains at the default config, consistent with the speedups reported on the [Sol-Attn project page](https://nvlabs.github.io/Sana/Sol-Attn/).
- The packed-varlen path slices the sequence to its real length before the kernel (the alignment-padding document is excluded, so attention never crosses document boundaries). Dense steps/layers run the identical packed-varlen FlashAttention path, so the sparse kernel is measured against a bit-identical dense baseline.
- `kv_splits` auto-resolves to `1` on B300 (the 2/4 split path is SM90-only upstream).

### Robustness fixes found during verification

- With `dense_layers="0"`, the text-refiner attention (which runs on the short text rows only) can be shorter than the configured `sink_tokens=951` and previously hit the kernel's `sink_start + sink_tokens <= T` check. The sink range is now clamped to the actual sequence length (`_clamp_sink_range`), covered by a unit test.

### Output contract

Both backends produce h264 1344×768 @ 24 FPS, 124 frames with synchronized 32 kHz stereo audio; full `ffmpeg` decode passes on both.

vllm_omni/diffusion/attention/backends/sol_attn.py
```diff
@@ -0,0 +1,314 @@
...
+        )
+        q = query[:, :used].contiguous()
+        k = key[:, :used].contiguous()
+        v = value[:, :used].contiguous()
+        if q.dtype != torch.bfloat16:
+            raise TypeError(
+                f"Sol-Attn requires bfloat16 activations, got {q.dtype}. "
+                "Select a backend that supports the model dtype instead."
+            )
+        out = sol_attn(
```
hsliuustc0106, line 361: The sink makes the configured prefix exact only as K/V; it does not make those rows dense as queries. The 951-row MiniMax-H3 prefix includes generated audio, and the upstream H3 integration explicitly overwrites these query rows with dense attention: https://github.com/NVlabs/Sana/blob/8a26fb0ec9e353125ead798cb2e312d5ce48cded/models/minimax_h3/optimized/sol_attn_h3.py#L277-L290. Please reuse the dense helper to recompute `out[:, sink_start:sink_start + sink_tokens]` and add a forward-path regression test.

vllm_omni/diffusion/attention/backends/sol_attn.py
```diff
@@ -0,0 +1,314 @@
...
+        head_size: int,
+        softmax_scale: float,
+        causal: bool = False,
+        num_kv_heads: int | None = None,
+        prefix: str = "",
+        qkv_layout: str | None = None,
+        backend_kwargs: dict | None = None,
+        **extra_impl_args,
+    ) -> None:
+        del num_heads, num_kv_heads, qkv_layout, extra_impl_args
```
hsliuustc0106: Please do not drop `num_heads` and `num_kv_heads` here. Upstream Sol-Attn is noncausal and requires Q/K/V to share shape `[B, T, H, 128]`; as written, causal or GQA configurations work during dense steps and then silently change semantics or fail on the first sparse step. Reject `causal=True` and unequal Q/KV head counts at construction, and pass `scale=self.softmax_scale` to the sparse call so both paths use the same scale.

vllm_omni/diffusion/attention/backends/sol_attn.py
```diff
@@ -0,0 +1,314 @@
...
+            thresh_type=str(raw.get("thresh_type", "diag")),
+            kv_splits=raw.get("kv_splits", "auto"),
+            sink_tokens=int(raw.get("sink_tokens", 0)),
+            sink_start=None if raw.get("sink_start") is None else int(raw["sink_start"]),
+            dense_steps=int(raw.get("dense_steps", 10)),
+            dense_layers=_parse_layer_ranges(raw.get("dense_layers", "0,1")),
+        )
+
+
+class SolAttnBackend(AttentionBackend):
```
hsliuustc0106, line 107: Set `supported_platforms = ("cuda",)` here. Without it, explicit `SOL_ATTN` selection passes the shared backend validation on ROCm, XPU, and NPU and fails only after model construction or forward dispatch.

vllm_omni/diffusion/data.py
```diff
@@ -1504,6 +1548,7 @@ def __post_init__(self) -> None:
         self.skip_softmax = self._coerce(self.skip_softmax, SkipSoftmaxSpec, "skip_softmax")
         self.quant = self._coerce(self.quant, AttnQuantSpec, "quant")
         self.block_sparse = self._coerce(self.block_sparse, BlockSparseSpec, "block_sparse")
+        self.sol_attn = self._coerce(self.sol_attn, SolAttnSpec, "sol_attn")
```
hsliuustc0106, line 1551: Please also add `sol_attn` to `AttentionConfig._flatten_per_role_entry()` / `spec_keys`. The new block works under `default`, but `AttentionConfig(per_role={"self": {"backend": "SOL_ATTN", "sol_attn": {"tau": 1.0}}})` currently raises `ValueError: cannot mix backend/skip_softmax with nested role keys`. Add this construction to the config tests.

vllm_omni/diffusion/data.py
```diff
@@ -1488,6 +1488,49 @@ def __post_init__(self) -> None:
...
+            raise ValueError(
+                f"sol_attn.thresh_type={self.thresh_type!r} unsupported; use one of {sorted(self._VALID_THRESH_TYPES)}."
+            )
+        if self.sink_tokens < 0:
+            raise ValueError(f"sol_attn.sink_tokens must be >= 0; got {self.sink_tokens!r}.")
+        if self.sink_start is not None and self.sink_start < 0:
+            raise ValueError(f"sol_attn.sink_start must be >= 0 or None; got {self.sink_start!r}.")
+        if self.dense_steps < 0:
+            raise ValueError(f"sol_attn.dense_steps must be >= 0; got {self.dense_steps!r}.")
+        if isinstance(self.kv_splits, int):
```
hsliuustc0106, line 1525: Validate integer `kv_splits` against `{1, 2, 4}`. The current `>= 1` contract accepts values such as `3`, which upstream rejects only when the first sparse step runs; `2/4` are additionally SM90-only. The docs should describe the same constraints and unsupported architectures should fail before inference.

hsliuustc0106:
Local verification against frozen head `798dd1e2147982f0a18fbc5b7d88addc71899400` on 4× NVIDIA L20X (SM90), with the PR commits applied to latest main `a874b8e09bb5251bb16e3de8bc22edf3880eec09`: MiniMax-H3, BF16, 1344×768, 124 frames, 20 steps, seed 1101, regional compile, HSDP4, Ulysses4/ring1.

| Measurement | FLASH_ATTN | SOL_ATTN | Result |
|---|---:|---:|---:|
| Raw 3-request diffuse mean | 33.589 s | 34.168 s | 0.983× |
| Steady state, first full request discarded from both | 32.250 ± 0.394 s | 28.076 ± 0.204 s | **1.149×** |
| Peak GPU memory | 54,098 MB | 54,038 MB | unchanged |

The harness warmup runs only two dense steps, so the first Sol request compiled the sparse branch at step 11 and took 46.35 s. The raw short-run aggregate therefore includes cold compilation; the steady-state comparison shows a 12.9% denoise reduction on SM90. This validates a local CuTe-path gain, not the B300/Triton number in the PR body.

For one fixed local prompt/seed, decoded video comparison against FLASH_ATTN gave PSNR 40.04 dB and SSIM 0.9762. Decoded audio waveform SNR was approximately 14.6 dB; this is a diagnostic, not a perceptual metric. All three outputs within each backend were byte-identical because the seed was fixed, so quality evidence represents one unique sample.

Targeted tests passed (`18 passed, 73 deselected`), but the added tests never execute either dense or sparse forward logic. Before marking this ready, please address the inline correctness/configuration issues and add forward-path coverage plus at least three repeated, full-path-warmed A/B runs with timing variance, VRAM, video metrics, and an audio-quality check. The PR body headline currently says 1.1–1.2× for the default B300 config while the table reports 1.02× from one timed run; those claims should be reconciled.

vllm_omni/diffusion/attention/backends/sol_attn.py
```diff
@@ -0,0 +1,314 @@
...
+        qkv_layout: str | None = None,
+        backend_kwargs: dict | None = None,
+        **extra_impl_args,
+    ) -> None:
+        del num_heads, num_kv_heads, qkv_layout, extra_impl_args
+        if head_size != _SOL_ATTN_HEAD_DIM:
+            raise ValueError(f"Sol-Attn requires head_size={_SOL_ATTN_HEAD_DIM}, got {head_size}")
+        self.causal = causal
+        self.softmax_scale = softmax_scale
+        self.config = SolAttnConfig.from_backend_kwargs(backend_kwargs)
```
hsliuustc0106, line 156: [P2] Reject ring SP before it bypasses Sol-Attn

With `ring_degree > 1`, `Attention._forward_impl()` calls `_run_ring_attention()` instead of `_run_local_attention()`, so `SolAttnImpl.forward_cuda()` and every `sol_attn` setting are skipped. The service still returns dense ring output, silently losing the requested sparse speedup even though the new docs say this combination is unsupported. Please mirror RainFusion's constructor-time parallel-config validation and add a `ring_degree=2` rejection test.

vllm_omni/platforms/cuda/platform.py
```diff
@@ -198,6 +198,25 @@ def get_diffusion_attn_backend_cls(
                     raise ValueError(
                         "TRTLLM_ATTN diffusion attention backend requires flashinfer, which is not installed."
                     )
+            if backend_upper == "SOL_ATTN":
```
hsliuustc0106, line 201: [P2] Reject pre-SM80 GPUs when selecting SOL_ATTN

`compute_supported` is already available above, but this branch checks only `head_size` and whether the package imports. Upstream Sol-Attn rejects compute capability below 8.0 only when the sparse kernel is first invoked; with the default `dense_steps=10`, that failure can be delayed until inference is underway. Please fail here with a clear SM80+ error, as the other architecture-specific backends do, and cover a mocked 7.x capability.

hsliuustc0106:
Repeat review against the same frozen head `798dd1e2147982f0a18fbc5b7d88addc71899400`:

- No commits have been added since the first review, and all five earlier inline threads remain unanswered/unresolved.
- CI is still green; the PR remains draft, mergeable, blocked on review.
- A deeper static pass found two additional compatibility gaps: ring sequence parallelism silently bypasses `SOL_ATTN`, and the CUDA selector accepts pre-SM80 GPUs that upstream rejects only at sparse execution time.
- I did not rerun the local benchmark because the code and test surface are byte-for-byte unchanged. The prior same-SHA result remains 1.149× steady-state denoise speedup on 4× L20X, with unchanged peak VRAM; the cold-inclusive three-run mean was 0.983× because the first timed request compiled the sparse branch.

Verdict remains **not ready**. Please resolve the seven inline threads, add actual dense/sparse forward-path tests, and reconcile the PR performance headline with repeated fully warmed measurements.

PR #5853, opened 2026-08-06
[Model][Feat] Support Minimax-H3 quality grading requests through dynamic loading/unloading with Cache-DiT

mglyn:
## Purpose

This PR addresses section 2.2 of #5700 by adding an optional diffusion request field:

```python
quality: str | None = None
```

The common request layer accepts `lossless` and `high`, routes the field through online and offline entrypoints, and isolates different quality values into separate worker batches.

## MiniMax H3 behavior

| Request | Behavior |
|---|---|
| omitted | Keep or restore the server-configured Cache-DiT profile |
| `lossless` | Uninstall Cache-DiT hooks and run the reference path |
| `high` | Keep or install the H3 high-quality Cache-DiT profile |

Without startup Cache-DiT capability, omitted and `lossless` are no-ops, `high` returns an error.

The model policy chooses the desired state and the H3 pipeline applies it immediately before denoising. Repeated requests avoid redundant transitions: the same profile only refreshes request context, while a changed profile first uninstalls the old hooks.

The H3 `high` profile uses one front compute block, four warmup steps, a 0.04 residual-difference threshold, and at most one consecutive cached step. Cache-DiT still makes dynamic per-step hit/miss decisions, so the resulting speed/quality tradeoff is deployment-dependent. `lossless` remains the reference path.

Other models do not opt into this lifecycle and retain the existing runner-owned Cache-DiT behavior. TeaCache, MagCache, and StepCache are unchanged.

## End-to-end result

The online server was exercised on 4x NVIDIA L20X with SP4, text-encoder TP4, 1344x768, 124 frames, 24 FPS, and 50 steps. One full `lossless` request (92.48 s) was used as warmup and excluded. Three prompt/seed pairs were then run in the balanced order `L->H`, `H->L`, `L->H`.

| Pair | lossless | high | Latency reduction | SSIM | PSNR |
| 1 | 85.49 s | 63.94 s | 25.2% | 0.9783 | 41.28 dB |
| 2 | 85.58 s | 63.36 s | 26.0% | 0.9694 | 33.04 dB |
| 3 | 85.34 s | 63.24 s | 25.9% | 0.9709 | 34.98 dB |
| Median | 85.49 s | 63.36 s | 25.9% | 0.9709 | 34.98 dB |

`high` delivered about 1.35x throughput. All seven requests returned valid 1344x768 H.264/AAC output; audio PSNR exceeded 163 dB per channel. The sequence covered startup removal, profile installation, reuse, and both switch directions without restarting. These measurements apply to this deployment and are not universal performance or quality guarantees.

## Test result

```text
Affected unit/API tests: 327 passed, 1 unrelated test deselected
Ruff: PASS
git diff --check: PASS
4-GPU online E2E: PASS
```

The tests cover request routing, batch isolation, startup backend adoption, lossless uninstall, repeated-request no-op/refresh, profile switching, and the no-Cache-DiT error path for `high`.

https://github.com/user-attachments/assets/b6789818-b5fe-4d02-9db5-1f60bb8a5cef

### DIFF
--- a/recipes/MiniMaxAI/MiniMax-H3.md
+++ b/recipes/MiniMaxAI/MiniMax-H3.md
@@ -574,6 +574,42 @@ default. `short_edge` controls the 768-pixel canvas and must be `768`.
 outputs; the synchronous raw-MP4 endpoint returns the first output when more
 than one is requested.
 
+## Request-scoped quality
+
+Start the H3 server with Cache-DiT capability by adding this option to one of
+the launch commands above:
+
+```bash
+--cache-backend cache_dit
+```
+
+Then add one of these fields to any HTTP request above. Omitting `quality`
+keeps or restores the startup Cache-DiT profile.
+
+```bash
+-F 'quality=lossless'  # Native reference path for this request.
+-F 'quality=high'      # H3's conservative Cache-DiT profile.
+```
+
+Without startup Cache-DiT, omitted and `lossless` requests use the reference
+path, while `high` returns an error.
+
+The following result was measured on 4× NVIDIA H200 with SP4, text-encoder
+TP4, 1344×768, 124 frames, 24 FPS, and 50 inference steps. One full
+`lossless` warmup was excluded, followed by three fixed prompt/seed pairs in
+balanced switch order.
+
+| `quality` | Median inference latency | Speedup | SSIM vs `lossless` | PSNR vs `lossless` | Expected trade-off |
+|---|---:|---:|---:|---:|---|
+| `lossless` | 85.49 s | 1.00× | 1.0000 | exact | Native reference path |
+| `high` | 63.36 s | 1.35× | 0.9709 | 34.98 dB | Faster with measured same-seed deviation |
+
+> [!NOTE]
+> `high` selects a fixed Cache-DiT profile, but its cache hit rate and
+> resulting latency/quality trade-off may vary by hardware, topology, and
+> workload. The values above apply to this deployment and are not universal
+> guarantees. `lossless` remains the exact reference path.
+
 ## Key parameters
 
 | Parameter | Recommended value | Notes |
--- a/recipes/MiniMaxAI/MiniMax-H3.md
+++ b/recipes/MiniMaxAI/MiniMax-H3.md
@@ -578,6 +578,8 @@ than one is requested.
 
 | Parameter | Recommended value | Notes |
 |-----------|-------------------|-------|
+| `quality` | omitted or `lossless` | Request-level quality intent; `high` requires startup `--cache-backend cache_dit` and selects H3's conservative Cache-DiT profile |
+| `extra_params.force_refresh` | omitted | Optional positive 1-based denoising-step hint for a Cache-DiT context refresh; pair with `extra_params.force_refresh_policy`=`once` or `repeat` |
 | `task` | `t2va`, `fl2va`, or `ref2va` | Passed in `extra_params`; selects the task-specific DiT |
 | `duration` | Workload-specific | Decimal seconds in `extra_params`; converted to H3-compatible frame count |
 | `fps` | `24` | H3 output FPS is fixed |
--- a/recipes/MiniMaxAI/MiniMax-H3.md
+++ b/recipes/MiniMaxAI/MiniMax-H3.md
@@ -576,23 +576,18 @@ than one is requested.
 
 ## Request-scoped quality
 
-Start the H3 server with Cache-DiT capability by adding this option to one of
-the launch commands above:
-
-```bash
---cache-backend cache_dit
-```
-
-Then add one of these fields to any HTTP request above. Omitting `quality`
-keeps or restores the startup Cache-DiT profile.
+Add one of these fields to any HTTP request above. No Cache-DiT startup option
+is required; H3 installs its conservative profile when a `high` request
+arrives and removes it for a `lossless` request.
 
 ```bash
 -F 'quality=lossless'  # Native reference path for this request.
 -F 'quality=high'      # H3's conservative Cache-DiT profile.
 ```
 
-Without startup Cache-DiT, omitted and `lossless` requests use the reference
-path, while `high` returns an error.
+Omitting `quality` preserves the startup default: it uses the reference path
+normally, or the server-configured profile when the server was started with
+`--cache-backend cache_dit`.
 
 The following result was measured on 4× NVIDIA H200 with SP4, text-encoder
 TP4, 1344×768, 124 frames, 24 FPS, and 50 inference steps. One full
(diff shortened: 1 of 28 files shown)


mglyn:
I added the request-scoped Cache-DiT refresh hint in this PR (extra_params.force_refresh_step_hint, with an optional once/repeat policy). Keeping it together with the quality switch is intentional: both features control the same per-request Cache-DiT lifecycle and share the same policy/runtime path. The change remains H3-scoped, while other models retain their existing behavior by default.

vllm_omni/diffusion/models/minimax_h3/quality_policy.py
```diff
@@ -0,0 +1,95 @@
...
+                )
+            return MiniMaxH3QualityPlan(
+                cache_dit=CacheDiTRequestSpec(
+                    installation_key=MINIMAX_H3_HIGH_CACHE_KEY,
+                    cache_config=_high_quality_cache_config(),
+                    num_inference_steps=num_inference_steps,
+                ),
+            )
+
+        generic_requested = self._configured_backend == "cache_dit" and quality is None
```
david6666666: For a request like quality="fast" this falls through to cache_dit=None, i.e. Cache-DiT disabled -- the slowest H3 path. Since ast is a valid level in the shared protocol (DIFFUSION_QUALITY_LEVELS), a caller asking for the fastest tier silently gets the reference path instead. Suggest rejecting unsupported levels like high does without cache_dit, or mapping ast to an actual accelerated profile; at minimum log a warning so this isn't silent.

vllm_omni/entrypoints/openai/serving_video.py
```diff
@@ -200,6 +200,7 @@ async def _run_and_extract(
 
         if "num_inference_steps" in provided_fields and request.num_inference_steps is not None:
             gen_params.num_inference_steps = request.num_inference_steps
+        gen_params.quality = request.quality
```
david6666666: Small consistency nit: every sibling field above only overrides when explicitly provided (provided_fields), but this always assigns. Harmless today since the default is None, but it would silently clobber a model-set default quality if one is ever introduced. Mind gating it like 
um_inference_steps for symmetry.

vllm_omni/entrypoints/openai/serving_video_output_stream.py
```diff
@@ -535,6 +535,7 @@ async def _build_prompt_and_sampling_params(
         provided_fields = request.model_fields_set
         if "num_inference_steps" in provided_fields and request.num_inference_steps is not None:
             gen_params.num_inference_steps = request.num_inference_steps
+        gen_params.quality = request.quality
```
david6666666: Same as the sync path: unconditional assignment while sibling fields are gated on provided_fields. Fine today (None default), but a future model default quality would be silently overwritten. Consider gating for consistency.

david6666666:
Single-GPU sanity check (B300 x1, 1344x768, 5s/124 frames, 50 steps, seed 1101, FLASH_ATTN, eager, FL2VA t2va): main @ a874b8e0 vs PR head 7a94a444.

| config | elapsed | vs lossless |
|---|---|---|
| main (reference, no cache) | 157.4s | -- |
| quality=lossless (no cache) | 153.4s | -- |
| quality=lossless (server started with cache_dit) | 158.8s | -- |
| quality=omitted (cache_dit startup, generic profile) | 100.7s | -34% |
| quality=high | 96.3s | -37% |

Precision:
- lossless vs main baseline: frames + audio bit-identical (max diff 0) -- reference path is exactly aligned with main.
- lossless on a cache_dit server vs plain lossless: bit-identical -- the uninstall path leaves no hooks behind.
- high vs main: SSIM 0.778 / PSNR 21.7 dB on this single-GPU workload, lower than the 0.97 / 33-41 dB reported for 4xL20X SP4. The cache hits more aggressively here (~1.59x vs the claimed ~1.35x), so the quality/throughput tradeoff shifts with deployment.

The mechanism (routing + batch isolation + request-scoped lifecycle) checks out; the lossless path is a faithful reference and the perf gain is real. The high quality numbers look deployment-dependent rather than universal.

vllm_omni/diffusion/models/minimax_h3/quality_policy.py
mglyn: Resolved. I removed fast from the shared diffusion quality levels.

vllm_omni/entrypoints/openai/serving_video.py
mglyn: Resolved. quality is now applied only when explicitly present in request.model_fields_set, preserving the engine/model default when omitted.

vllm_omni/entrypoints/openai/serving_video_output_stream.py
mglyn: Resolved. The streaming path now uses the same provided_fields guard and preserves the default quality when the request omits quality.

mglyn:
> Single-GPU sanity check (B300 x1, 1344x768, 5s/124 frames, 50 steps, seed 1101, FLASH_ATTN, eager, FL2VA t2va): main @ [a874b8e](https://github.com/vllm-project/vllm-omni/commit/a874b8e09bb5251bb16e3de8bc22edf3880eec09) vs PR head [7a94a44](https://github.com/vllm-project/vllm-omni/commit/7a94a44452b0e1e9311688efc31f3f36ae818662).
> 
> config	elapsed	vs lossless
> main (reference, no cache)	157.4s	--
> quality=lossless (no cache)	153.4s	--
> quality=lossless (server started with cache_dit)	158.8s	--
> quality=omitted (cache_dit startup, generic profile)	100.7s	-34%
> quality=high	96.3s	-37%
> Precision:
> 
> * lossless vs main baseline: frames + audio bit-identical (max diff 0) -- reference path is exactly aligned with main.
> * lossless on a cache_dit server vs plain lossless: bit-identical -- the uninstall path leaves no hooks behind.
> * high vs main: SSIM 0.778 / PSNR 21.7 dB on this single-GPU workload, lower than the 0.97 / 33-41 dB reported for 4xL20X SP4. The cache hits more aggressively here (~1.59x vs the claimed ~1.35x), so the quality/throughput tradeoff shifts with deployment.
> 
> The mechanism (routing + batch isolation + request-scoped lifecycle) checks out; the lossless path is a faithful reference and the perf gain is real. The high quality numbers look deployment-dependent rather than universal.

Our high request selects a fixed Cache-DiT profile, but its actual hit rate and speed/quality tradeoff are deployment-dependent. We intentionally do not gate it by GPU, topology, or request shape; lossless remains the exact reference path. I updated the PR text to scope the reported metrics to our 4×L20X SP4 experiment and avoid a universal guarantee. Thanks for the additional validation.

recipes/MiniMaxAI/MiniMax-H3.md
david6666666, line 576: you can add quality use example here and paste the measured trade-off like table
```
quality | Meaninferencelatency | Speedup | SSIM vslossless | PSNR vslossless | Expectedtrade-off
-- | -- | -- | -- | -- | --
```

david6666666 (commented):
you can add some usage on docs/user_guide/diffusion/cache_acceleration/cache_dit.md

recipes/MiniMaxAI/MiniMax-H3.md
mglyn: sure

david6666666 (approved):
LGTM now.

PR #5508, opened 2026-07-28
[Model] Add native SANA-Video 2B T2V and I2V support

liuyao0322:
# Summary

Adds single-GPU SANA-Video 2B support to vLLM-Omni for text-to-video
(T2V) and image-to-video (I2V) generation.

Supported checkpoints:

- `Efficient-Large-Model/SANA-Video_2B_480p_diffusers`
- `Efficient-Large-Model/SANA-Video_2B_720p_diffusers`

This PR introduces:

- Native vLLM-Omni SANA-Video Transformer.
- Native 480p/720p T2V and I2V pipelines.
- Checkpoint-compatible SANA ReLU-kernel linear self-attention.
- vLLM-Omni Attention for prompt cross-attention.
- Distributed VAE wrappers for the Wan VAE and LTX-2 VAE variants.
- A Diffusers-adapter compatibility/reference backend.
- Offline and OpenAI-compatible online-serving examples.
- L1 component regressions, on-demand correctness/alignment tools, and an
  eight-case L4 nightly serving matrix.
- A model recipe covering commands, hardware observations, validation, and
  current limitations.

This PR completes the single-GPU SANA-Video 2B milestone. SANA-Video 2.0
5B/14B and its gated/hybrid architecture are not included.

---

# Supported Pipelines

- `SanaVideoPipeline`: native text-to-video.
- `SanaImageToVideoPipeline`: native image-to-video.
- `--diffusion-load-format diffusers`: Diffusers-adapter backend.

Supported checkpoint profiles:

| Checkpoint | Resolution |
|---|---|
| 480p | 832 x 480 |
| 720p | 1280 x 704 |

---

# Implementation

## Native Transformer

Implemented:

- SANA-Video Transformer blocks.
- ReLU-kernel linear self-attention. SANA-Video 2B does not use the gated
  linear attention introduced by SANA-Video 2.0.
- Prompt cross-attention using vLLM-Omni Attention.
- RMSNorm and rotary embeddings.
- Timestep embeddings and AdaLayerNormSingle.
- PixArt text projection.
- Temporal GLUMB convolution.
- Strict checkpoint loading.

Checkpoint compatibility:

| Variant | Matched | Missing | Unexpected |
|---|---:|---:|---:|
| 480p | 496/496 | 0 | 0 |
| 720p | 496/496 | 0 | 0 |

## T2V / I2V Pipelines

T2V supports prompt encoding, latent initialization, CFG denoising,
scheduler execution, VAE decoding, and video postprocessing.

[...]

PR #5861, opened 2026-08-06
[Model] Add TP and CFG parallelism to SANA-Video 2B

cr-gao:
> **Update 2026-08-25**: #5508 merged; this branch is rebased onto current `main`
> (10 commits, conflicts resolved). The upstream final #5508 absorbed this PR's
> earlier cross-attention boolean-mask fix, so that commit was dropped, and the
> backend restriction it implied is gone (upstream now routes masked
> cross-attention through the SDPA fallback in-code). Post-rebase sanity on the
> same 4×A800 box: tp1 is **bitwise identical** to merged `main`'s single-GPU
> output (1-step, real weights), tp2/cfg2 sit at rel_l2 2.4e-2/2.7e-2 (same
> bf16 reduction-reorder magnitude as the table below), and the tiny multi-GPU
> L2 matrix passes (7 cases).

## What this does

Adds native tensor parallelism (TP, up to 2 GPUs) and CFG parallelism to the
SANA-Video 2B native pipelines (T2V and I2V), on top of the single-GPU support
from #5508. The serial (single-GPU) path is unchanged; the parallel paths only
shard work and re-order reductions.

## Changes (18 files, +1647/-114)

- **CFG parallelism** — both pipelines inherit `CFGParallelMixin`; the denoise
  loop is split so CFG1 keeps the batched path and CFG2 runs one branch per
  rank and gathers. (`pipeline_sana_video.py`, `pipeline_sana_video_i2v.py`)
- **Tensor parallelism** — self / cross attention use Column/Row parallel
  linears (separate q/k/v, checkpoint keys preserved); SANA-local distributed
  RMSNorm with fused Q/K all-reduce; GLUMB temporal conv is weight-sharded with
  a hand-written merged loader. (`transformer_sana_video.py`)
- **Topology validation** — `validate_sana_video_parallel_config` rejects
  unsupported topologies (TP != 1/2, CFG != 1/2, SP, text-encoder TP) before
  weights load; the Diffusers adapter path explicitly rejects TP.
- **Single-process usability** — `diffuse()` tolerates uninitialized parallel
  groups (falls back to the serial path), so the pipelines keep working in
  single-process runs and unit tests.
- **Tests** — L1 CFG/TP unit tests (attention, distributed RMSNorm, GLUMB
  shard, topology rejection, CFG combine incl. tuple predictions); tiny
  multi-GPU L2 wiring (T2V TP2/CFG2/TP2×CFG2, I2V CFG2/TP2×CFG2).

## Correctness

The serial path is untouched and the parallel paths are mathematically
equivalent up to reduction order (bf16). Verified against #5508 head
(`220379011`) on the same machine and inputs (4×A800), comparing the decoded
video tensor per topology × backend.

The reference is #5508's own output produced **on this machine** (its source
swapped in at tp1), not a diffusers golden. A diffusers golden is regenerated
per machine and is hardware-sensitive — e.g. this 480p I2V scores ~0.82 SSIM
against an off-machine diffusers golden, *identically* on #5508 and on this
branch — so it cannot separate a parallel regression from a hardware
difference. Same-machine parity against #5508 isolates exactly this PR's change.

These numbers come from a one-off local harness (swap in #5508 as the baseline,
compare each topology); it is **not** part of this PR — a throwaway A/B check
with no lasting value. The regression coverage that ships is the tiny
multi-GPU L2 suite below.

- **tp1 is bitwise-identical to `220379011`** on all four cells (480p/720p ×
  T2V/I2V) and both backends: `max_abs = 0`, PSNR = inf, SSIM ≈ 1.0. The
  parallel refactor did not change the serial code path.
- **Parallel topologies (tp2 / cfg2 / tp2×cfg2)** reproduce the baseline within
  the expected bf16 reduction-reorder drift, amplified across the 50-step CFG
  trajectory. Full grid — decoded-tensor drift and SSIM/PSNR vs the `220379011`
  baseline (50 steps, guidance 6.0), all 24 parallel cases passed:

  | cell | topology | backend | rel_l2 | SSIM | PSNR (dB) |
  |------|----------|---------|--------|------|-----------|
  | 480p T2V | tp2 | CUDNN | 0.128 | 0.955 | 27.9 |
  | 480p T2V | tp2 | SDPA | 0.111 | 0.960 | 29.1 |
  | 480p T2V | cfg2 | CUDNN | 0.162 | 0.939 | 25.9 |
  | 480p T2V | cfg2 | SDPA | 0.123 | 0.953 | 28.3 |
  | 480p T2V | tp2×cfg2 | CUDNN | 0.129 | 0.955 | 27.8 |
  | 480p T2V | tp2×cfg2 | SDPA | 0.102 | 0.964 | 29.9 |
  | 480p I2V | tp2 | CUDNN | 0.130 | 0.948 | 29.0 |
  | 480p I2V | tp2 | SDPA | 0.122 | 0.952 | 29.6 |
  | 480p I2V | cfg2 | CUDNN | 0.060 | 0.982 | 35.7 |
  | 480p I2V | cfg2 | SDPA | 0.043 | 0.990 | 38.6 |
  | 480p I2V | tp2×cfg2 | CUDNN | 0.126 | 0.950 | 29.3 |
  | 480p I2V | tp2×cfg2 | SDPA | 0.126 | 0.948 | 29.3 |
  | 720p T2V | tp2 | CUDNN | 0.036 | 0.988 | 38.5 |
  | 720p T2V | tp2 | SDPA | 0.039 | 0.987 | 37.8 |
  | 720p T2V | cfg2 | CUDNN | 0.084 | 0.967 | 31.3 |
  | 720p T2V | cfg2 | SDPA | 0.072 | 0.973 | 32.5 |
  | 720p T2V | tp2×cfg2 | CUDNN | 0.066 | 0.975 | 33.3 |
  | 720p T2V | tp2×cfg2 | SDPA | 0.087 | 0.965 | 30.9 |
  | 720p I2V | tp2 | CUDNN | 0.047 | 0.993 | 38.8 |
  | 720p I2V | tp2 | SDPA | 0.197 | 0.951 | 26.4 |
  | 720p I2V | cfg2 | CUDNN | 0.034 | 0.994 | 41.6 |
  | 720p I2V | cfg2 | SDPA | 0.089 | 0.986 | 33.3 |
  | 720p I2V | tp2×cfg2 | CUDNN | 0.169 | 0.958 | 27.7 |
  | 720p I2V | tp2×cfg2 | SDPA | 0.236 | 0.940 | 24.8 |

  SSIM stays 0.94–0.99 across every cell (visually close to serial); the drift
  is the chaotic CFG trajectory magnifying per-step bf16 differences and varies
  run to run, the same way outputs differ across GPUs/backends.
- Backends: measured on `CUDNN_ATTN` and `TORCH_SDPA`. (At measurement time
  FLASH mishandled the cross-attention mask; upstream #5508 now routes masked
  cross-attention through the SDPA fallback in-code, so no backend flag is
  required anymore.)
- Tiny multi-GPU L2 suite passes (all ranks complete, weights load, CFG-rank
  latent checksums match, tolerance vs serial reference).

Note on drift: TP shards each matmul and adds an all-reduce; in bf16 the
reordered summation differs slightly from the single-GPU reduction, and
diffusion + CFG amplify per-step differences across steps. This is expected
(the same reason outputs differ across GPUs/backends), not a regression — tp1
bitwise + the tiny L2 numerics are the correctness guarantees.

## Performance (4×A800, NVLink, 480p T2V)

Environment: 4× A800-SXM4-80GB, all-pairs NV8 NVLink; this PR at `7a29e66df`,
baseline `main` at `f0808c979` (this PR's merge-base), measured in the same
session on the same box; torch 2.11.0+cu130, vLLM 0.26.0; 480p checkpoint
revision `fed3bce4`. Protocol: 1 warmup + 3 runs at 10 steps + 3 runs at 20
steps per topology, `TORCH_SDPA`, guidance 6.0, seed 42; DiT-loop latency is
the 10→20-step slope reported at 50 steps. All raw per-run times are listed so
min/max spread is visible (≤1% per cell).

| topology | cards | 10-step runs (s) | 20-step runs (s) | DiT-loop (50) | vs main tp1 | comm kernels, % of GPU busy |
|---|---|---|---|---|---|---|
| main tp1 (`f0808c979`) | 1 | 28.23 / 28.25 / 28.25 | 50.90 / 51.02 / 51.07 | 113.9 s | 1.00× | 0% |
| tp1 (this PR) | 1 | 27.76 / 27.79 / 27.83 | 50.07 / 50.26 / 50.57 | 112.3 s | no regression¹ | 0% |
| tp2 | 2 | 21.03 / 21.05 / 21.09 | 35.89 / 35.97 / 35.99 | 74.6 s | 1.53× | 9.6% |
| cfg2 | 2 | 17.48 / 17.55 / 17.67 | 28.56 / 28.60 / 28.62 | 55.3 s | 2.06× | 0.4% |
| tp2×cfg2 | 4 | 14.02 / 14.05 / 14.09 | 21.74 / 21.74 / 21.77 | 38.4 s | 2.96× | 9.6% |
| tp1, 720p ckpt² | 1 | 8.06 / 8.08 / 8.14 | 14.65 / 14.67 / 14.67 | 32.9 s | — (see ²) | 0% |
| tp2, 720p ckpt² | 2 | 7.90 / 7.91 / 7.95 | 12.32 / 12.42 / 12.67 | 22.5 s | 1.46× vs 720p tp1 | 10.1% |

¹ Serial-path check for the baseline question: an interleaved A/B (3
alternating engine rounds on the same GPU, 4×10-step runs each) gives main
28.22–28.61 s vs this PR 27.80–28.24 s per 10 steps — the PR's serial path is
consistently ~1.4% faster, never slower, so the speedup denominators are not
inflated by a regressed tp1. A per-kernel diff of the two traces attributes
the delta to the replacement RMSNorm's cheaper eager kernel mix (fp32
sum-based path vs the previous mean + mixed-dtype-cast chain); the vLLM
linear layers' GEMM kernel change is roughly neutral.

² The 720p checkpoint is officially shipped with a different VAE
(`AutoencoderKLLTX2Video`, 32× spatial / 8× temporal compression, vs
`AutoencoderKLWan` 8×/4× for 480p — verified against the HF repo at the pinned
revision). Its DiT sequence is therefore ~13.5× shorter (measured latents:
480p `(16, 21, 60, 104)` → 32 760 tokens vs 720p `(128, 11, 22, 40)` → 2 420
tokens), which is why 720p steps are faster than 480p; the tp2 speedup for
this checkpoint is computed against its own tp1 row.

**Compute vs comm decomposition** (per-rank `torch.profiler` trace of a
10-step request, kernels classified by name; NCCL kernel time = comm):

- tp2: 60 large bf16 all-reduces per step of 294 MB each (row-parallel
  outputs; dims `[2, 32760, 2240]` read from a record_shapes trace, ~2.3 ms
  each) + 80 tiny fp32 all-reduces (distributed-RMSNorm sum-of-squares,
  3.5 ms/step total) → 143 ms/step of collective kernels, 9.6% of GPU busy.
- cfg2: one small gather per step → 0.4% of GPU busy; compute-bound.
- GPU idle: every gap ≥1 ms sits in the first ~0.35 s of the profiled
  request — CPU-side trace shows them inside `c10d::allgather_` /
  communicator lazy init, profiler start, and pre-loop syncs — one-time
  setup, zero such gaps inside the denoise loop. Steady-state loop idle is
  micro launch gaps only (tp1: 25 840 gaps <50 µs averaging 3.6 µs plus ~900
  of 50 µs–1 ms, totalling ~19 ms/step = 0.8%; tp2 ~21 ms/step = 1.4%), so
  CUDA graphs would recover at most ~1–2% here. The non-loop prefix
  (text-encode etc.) is ≈1% of GPU busy time in these windows.

Bench harness: the runner constructs `OmniRunner(model,
model_class_name="SanaVideoPipeline", parallel_config=…,
diffusion_attention_backend="TORCH_SDPA")` and times `generate()`; exact
script attached in the PR conversation (measurement-only tool, not part of
the diff).

A TP4 read-only spike (validator temporarily relaxed for that measurement,
then restored) taken earlier at `3e69ed0e3` reached only 2.15× — worse than
tp2×cfg2 on the same 4 GPUs — so TP4 is **not** promoted into the support
contract.

Honest caveat unchanged: TP speedup is interconnect-bound (NVLink here); on
PCIe-only systems TP can be slower than a single GPU.

## Limitations

- TP is capped at 2 GPUs (support contract 1/2); CFG at 2.
- Sequence parallelism and Cache-DiT are not supported (follow-ups: #5940, #5882).
- The Diffusers adapter path does not offer native parallelism.

## Test commands

```bash
# L1 (CPU)
uv run --no-sync pytest -q tests/diffusion/models/sana_video/
# tiny multi-GPU L2 (2–4 GPUs)
uv run --no-sync pytest -q tests/model_tests/diffusion/test_common_offline.py \
  -k "SanaVideoPipeline or SanaImageToVideoPipeline"
```


cr-gao:
@hsliuustc0106 @congw729 Now that #5508 is merged, this PR is rebased onto current `main` (clean 10-commit diff, no conflicts) and marked ready for review. Post-rebase sanity on 4×A800: tp1 bitwise-identical to merged `main`, tp2/cfg2 within the expected bf16 drift, tiny multi-GPU L2 matrix green — details in the updated description.

Review order suggestion for the SANA-Video 2B series: this PR first, then #5940 (SP, stacked on this one); #5882 (Cache-DiT/offload) is independent and can go in parallel. We'd appreciate a timely first pass — the follow-up SANA 2.0 integration work is queued behind this series. Thanks!

cr-gao:
Self-review — what I checked:

- **Serial path unchanged**: tp1 output is bitwise-identical to the merged #5508 single-GPU output (real weights, decoded tensor, 4 cells × 2 backends pre-rebase; re-verified 1-step against current `main` after the rebase).
- **Parallel correctness**: tp2 / cfg2 / tp2×cfg2 reproduce the serial baseline within bf16 reduction-reorder drift (SSIM 0.94–0.99 across the 24-case grid); CFG1 keeps the batched single-forward path (dispatch asserted in unit tests); GLUMB TP keeps nn.Conv2d kernels and only shards weights, with the merged-checkpoint loader covered by unit tests.
- **Topology validation** rejects TP>2, CFG>2, SP>1 and text-encoder TP before weight allocation; the Diffusers adapter rejects TP explicitly.
- **Rebase diff audit**: the cross-attn boolean-mask commit was dropped because upstream #5508's final revision subsumed it (equivalent tests exist upstream); the new single-process guard has a red→green regression test (removing the guard fails `test_diffuse_runs_without_initialized_parallel_groups`).
- **Regression**: L1 155 passed locally; tiny multi-GPU L2 matrix (T2V TP2/CFG2/TP2×CFG2, I2V CFG2/TP2×CFG2) passed on 4×A800; ruff/typos/signoff clean.

vllm_omni/diffusion/models/sana_video/pipeline_sana_video.py
```diff
@@ -724,6 +730,108 @@ def num_timesteps(self):
...
+        negative_prompt_attention_mask: torch.Tensor | None,
+        guidance_scale: float,
+        extra_step_kwargs: dict,
+        dtype: torch.dtype,
+        output_slice: int | None,
+    ) -> torch.Tensor:
+        do_true_cfg = self.do_classifier_free_guidance
+        try:
+            cfg_parallel = get_classifier_free_guidance_world_size() > 1
+        except AssertionError:
```
hsliuustc0106: nit: catching `AssertionError` (also in `pipeline_sana_video_i2v.py:246`) works but silently flips to the serial path if the CFG getter ever raises for a non-single-process reason, e.g. partial init. `model_parallel_is_initialized()` (`parallel_state.py`) distinguishes the same two cases without depending on the assert.

recipes/NVIDIA/SANA-Video-2B.md
```diff
@@ -119,6 +119,30 @@ MODEL=Efficient-Large-Model/SANA-Video_2B_480p_diffusers \
...
+  --dtype bfloat16 \
+  --port 8091
+```
+
+For CFG parallelism on 2 GPUs, replace `--tensor-parallel-size 2` with
+`--cfg-parallel-size 2`; it splits the guided and unguided branches across the
+two GPUs and only helps when `guidance_scale` is above 1. Passing both flags
+combines the two on 4 GPUs.
+
+Whether tensor parallelism lowers latency depends on the interconnect. Each
```
hsliuustc0106, line 141: Nice perf section. Two things would make the tables actionable instead of just convincing:

1. **Compute vs comm split of the DiT-loop per topology** (e.g. NCCL/custom-AR kernel time from `torch.profiler`). Right now the ratios already imply it — tp2 1.52x ≈ ~24% of the loop in collectives, cfg2 2.04x ≈ ~0, sp2 1.80x in between — but a measured split would pin down "interconnect-bound" and the "SP communicates ~50x less than TP" claim in #5940, which is exactly what a user deciding between tp2/sp2/cfg2 on slower links needs.
2. **Reproducibility:** the exact bench commands (the throwaway harness isn't attached) and min/max over the 3 runs — your own tp1 (120.5s) vs #5940's sp1 baseline (112.7s) on the "same box family" shows ~7% environmental drift, so the ratios carry noise the reader can't see.

Nice-to-have, not blocking: one 720p cell (parity grid covers it, speedups don't), and a note that timings were taken at `3e69ed0e3` before the rebase.

vllm_omni/diffusion/models/sana_video/transformer_sana_video.py
```diff
@@ -109,6 +141,115 @@ def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
...
+        x_float = x.float()
+        local_sum_sq = x_float.pow(2).sum(dim=-1, keepdim=True)
+        return x_float, local_sum_sq, x.shape[-1]
+
+    def _scale(
+        self,
+        x_float: torch.Tensor,
+        global_sum_sq: torch.Tensor,
+        global_count: int,
+        x: torch.Tensor,
```
nagisa-kunhah: The `x` parameter is unused in `_scale()`. Could we remove it to avoid confusion?

vllm_omni/diffusion/models/sana_video/transformer_sana_video.py
```diff
@@ -19,12 +19,44 @@
...
+from vllm.model_executor.layers.linear import ColumnParallelLinear, RowParallelLinear
 from vllm.model_executor.models.utils import AutoWeightsLoader
+from vllm.model_executor.utils import set_weight_attrs
 
 from vllm_omni.diffusion.attention.backends.abstract import AttentionMetadata
 from vllm_omni.diffusion.attention.layer import Attention as OmniAttention
 
 
+def validate_sana_video_parallel_config(parallel_config) -> None:
+    """Reject unsupported SANA-Video parallel topologies before weights load."""
```
nagisa-kunhah, line 36: Since this PR does not support pipeline parallelism (PP) or HSDP, could we also validate these settings and raise an error during initialization when either of them is enabled?

vllm_omni/diffusion/models/sana_video/transformer_sana_video.py
```diff
@@ -109,6 +141,115 @@ def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
...
+
+        return self._scale(x_float, global_sum_sq, global_count, x)
+
+
+def fused_qk_rms_norm(
+    norm_q: nn.Module,
+    norm_k: nn.Module,
+    q: torch.Tensor,
+    k: torch.Tensor,
+) -> tuple[torch.Tensor, torch.Tensor]:
```
nagisa-kunhah: This optimization may be outside the scope of this PR, and the PR currently does not provide any evidence that it is effective. Could you provide benchmark or profiling results demonstrating its benefit?

vllm_omni/diffusion/models/sana_video/transformer_sana_video.py
```diff
@@ -109,6 +141,115 @@ def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
...
+
+    NUMERICALLY IDENTICAL to ``norm_q(q), norm_k(k)``: all-reduce is elementwise,
+    so packing along the last dim reduces each slice independently with the same
+    fp32 accumulation. Requires q and k to share the same shape (true for
+    self-attention — both come from the same hidden states). Falls back to
+    independent application when either norm is not a SanaDistributedRMSNorm
+    (e.g. nn.Identity when qk_norm=False).
+    """
+    if not (isinstance(norm_q, SanaDistributedRMSNorm) and isinstance(norm_k, SanaDistributedRMSNorm)):
+        return norm_q(q), norm_k(k)
```
nagisa-kunhah: This fallback appears to be dead code. `fused_qk_rms_norm()` is only called from `vllm_omni/diffusion/models/sana_video/transformer_sana_video.py:639`.

When `qk_norm` is `None`, both `self.norm_q` and `self.norm_k` are `None`, and `fused_qk_rms_norm()` is not called. Otherwise, both are instances of `SanaDistributedRMSNorm`, so the fallback branch for other norm types is unreachable.

Could we remove the fallback and explicitly type `norm_q` and `norm_k` as `SanaDistributedRMSNorm`?

tests/model_tests/diffusion/test_common_online.py
```diff
@@ -31,7 +31,7 @@
 
 
 @pytest.mark.parametrize(
-    "model_name,accelerations,supported_tasks,check_multioutput,check_determinism",
+    "model_name,accelerations,supported_tasks,check_multioutput,check_determinism,check_i2v_t2v_divergence",
```
nagisa-kunhah: `check_i2v_t2v_divergence` is unused in this online test. It is included in the parametrization and accepted by `test_online_on_supported_tasks()`, but it is never read. Could we remove it from the online parametrization, or implement the corresponding online I2V/T2V divergence check?

nagisa-kunhah (commented):
The performance results do not include a comparable baseline from before this PR, so it is difficult to determine the actual speedup or identify possible regressions. Could you provide baseline results under the same hardware and workload?

vllm_omni/diffusion/models/sana_video/pipeline_sana_video.py
cr-gao: Switched both pipelines to a `model_parallel_is_initialized()` check (`b34207f88`) — a failing group lookup inside an initialized run now propagates instead of silently going serial, and a unit test pins that.

vllm_omni/diffusion/models/sana_video/transformer_sana_video.py
cr-gao: Removed in `6ab78a07d`; `7a29e66df` later folded the helper into `forward` altogether.

vllm_omni/diffusion/models/sana_video/transformer_sana_video.py
cr-gao: Added in `903292a09`. `pipeline_parallel_size > 1` and `use_hsdp` now raise `NotImplementedError` before weights load, with tests.

tests/model_tests/diffusion/test_common_online.py
cr-gao: Removed from the online parametrization in `5da57ff9f`, so the flag now only reaches the offline test, where the comparison actually runs.

vllm_omni/diffusion/models/sana_video/transformer_sana_video.py
cr-gao: Fair challenge — I measured it. A 2-rank microbenchmark on the real tp2 shapes shows no per-call saving (fused 3703.5 µs vs separate 3703.2 µs), and end-to-end tp2 is within noise. The trace explains why: the packed all-reduce moves exactly the same bytes as two separate ones, and these tiny fp32 all-reduces are only ~0.3% of a tp2 step to begin with. So `7a29e66df` drops the fusion and keeps plain per-tensor norms.

vllm_omni/diffusion/models/sana_video/transformer_sana_video.py
cr-gao: Removed in `6ab78a07d`; the whole `fused_qk_rms_norm` is gone as of `7a29e66df` (see the thread above), so the norms are now called directly.

recipes/NVIDIA/SANA-Video-2B.md
cr-gao: Reworked the performance section along these lines: a same-session `main` baseline row, per-topology comm share from profiler traces, raw per-run times plus the exact protocol, and a 720p cell. One observation worth the footnote: the 720p checkpoint ships a 32×/8× LTX-2 VAE, so its DiT sequence is 13.5× shorter than 480p — that's why its steps are faster. Everything re-measured in one session at the current head; the old `3e69ed0e3` timings are superseded.

cr-gao:
@nagisa-kunhah re: the missing baseline is added. `main` at the merge-base `f0808c979`, measured in the same session on the same box: 113.9 s vs 112.3 s for this PR's tp1, and an interleaved A/B shows the PR consistently ~1.4% faster, never slower. A per-kernel trace diff attributes that to the new RMSNorm's cheaper kernel mix, not to anything suspicious. All speedup columns in the description are now computed against the main baseline, with raw per-run times in the table.

cr-gao:
Bench harness used for the performance table (measurement-only tool, runs against a checkout via `PYTHONPATH`; not part of the diff):

<details>
<summary>rr1_bench.py</summary>

```python
"""Review-round-1 bench: one topology per invocation. Local tool, not for PR.

Same methodology as the original bench_sana.py (1 warmup + N repeats at two
step counts, DiT-loop latency from the slope) but additionally records the raw
per-run times so min/median/max can be published, and prints provenance
(vllm_omni.__file__ + tree SHA marker) so every number is attributable.

Run from inside the target tree with PYTHONPATH set to it.
"""

import argparse
import json
import os
import statistics
import time

import vllm_omni
from tests.helpers.runtime import OmniRunner
from vllm_omni.diffusion.data import DiffusionParallelConfig
from vllm_omni.inputs.data import OmniDiffusionSamplingParams, OmniTextPrompt

PROMPT = "A cat walking on the grass, facing the camera. motion score: 30."
NEG = "blurry, low quality"
FPS = 16


def _sampling(steps: int, height: int, width: int, frames: int) -> OmniDiffusionSamplingParams:
    return OmniDiffusionSamplingParams(
        height=height, width=width, num_frames=frames, fps=FPS,
        num_inference_steps=steps, guidance_scale=6.0, seed=42,
    )


def _prompt() -> OmniTextPrompt:
    p = OmniTextPrompt(prompt=PROMPT)
    p["negative_prompt"] = NEG
    return p


def _time_runs(runner: OmniRunner, steps: int, repeats: int, height: int, width: int, frames: int) -> list[float]:
    times = []
    for _ in range(repeats):
        t0 = time.perf_counter()
        runner.generate([_prompt()], [_sampling(steps, height, width, frames)])
        times.append(time.perf_counter() - t0)
    return times


def main() -> None:
    ap = argparse.ArgumentParser()
    ap.add_argument("--label", required=True)
    ap.add_argument("--tp", type=int, default=1)
    ap.add_argument("--sp", type=int, default=1)
    ap.add_argument("--cfg", type=int, default=1)
    ap.add_argument("--backend", default="TORCH_SDPA")
    ap.add_argument("--steps-lo", type=int, default=10)
    ap.add_argument("--steps-hi", type=int, default=20)
    ap.add_argument("--repeats", type=int, default=3)
    ap.add_argument("--height", type=int, default=480)
    ap.add_argument("--width", type=int, default=832)
    ap.add_argument("--frames", type=int, default=81)
    ap.add_argument("--model-env", default="SANA_VIDEO_480P_MODEL")
    ap.add_argument("--out", required=True)
    args = ap.parse_args()

    sha_file = os.path.join(os.getcwd(), "TREE_SHA")
    sha = open(sha_file).read().strip() if os.path.exists(sha_file) else "unknown"
    cards = max(args.tp, 1) * max(args.sp, 1) * max(args.cfg, 1)
    print(f"PROVENANCE vllm_omni={vllm_omni.__file__} tree_sha={sha} cwd={os.getcwd()}", flush=True)

    parallel = DiffusionParallelConfig(
        tensor_parallel_size=args.tp,
        cfg_parallel_size=args.cfg,
        sequence_parallel_size=args.sp,
        ulysses_degree=args.sp if args.sp > 1 else 1,
    )
    with OmniRunner(
        os.environ[args.model_env],
        model_class_name="SanaVideoPipeline",
        parallel_config=parallel,
        diffusion_attention_backend=args.backend,
        enable_cpu_offload=False,
    ) as runner:
        dims = (args.height, args.width, args.frames)
        _time_runs(runner, args.steps_lo, 1, *dims)  # warmup (lazy compile)
        lo = _time_runs(runner, args.steps_lo, args.repeats, *dims)
        hi = _time_runs(runner, args.steps_hi, args.repeats, *dims)

    med_lo, med_hi = statistics.median(lo), statistics.median(hi)
    step_span = args.steps_hi - args.steps_lo
    per_step = (med_hi - med_lo) / step_span if step_span else float("nan")
    rec = {
        "label": args.label,
        "hwf": f"{args.height}x{args.width}x{args.frames}",
        "backend": args.backend,
        "cards": cards,
        "tree_sha": sha,
        "vllm_omni_file": vllm_omni.__file__,
        "lo_steps": args.steps_lo,
        "hi_steps": args.steps_hi,
        "lo_runs_s": [round(t, 3) for t in lo],
        "hi_runs_s": [round(t, 3) for t in hi],
        "per_step_s": round(per_step, 4),
        "dit50_s": round(per_step * 50, 3),
        "e2e50_s": round(med_lo + per_step * (50 - args.steps_lo), 3),
    }
    with open(args.out, "a") as f:
        f.write(json.dumps(rec) + "\n")
    print("BENCH " + json.dumps(rec), flush=True)


if __name__ == "__main__":
    main()
```

</details>

Per-topology invocation: `CUDA_VISIBLE_DEVICES=<gpus> PYTHONPATH=<tree> python rr1_bench.py --label <name> --tp N --cfg N --sp N --out results.jsonl` (plus `--height 704 --width 1280 --model-env SANA_VIDEO_720P_MODEL` for the 720p rows).

vllm_omni/diffusion/models/sana_video/pipeline_sana_video.py
hsliuustc0106: Verified at 7a29e66df — guard switched in both pipelines and the new propagation test pins the behavior. Thanks!

recipes/NVIDIA/SANA-Video-2B.md
hsliuustc0106: Verified — the profiler comm decomposition, raw per-run spreads, and same-session main baseline answer both asks; the 720p VAE footnote is a nice touch. Nothing further from me.

hsliuustc0106 (approved):
All threads resolved and verified at 7a29e66df — init-guard fix with propagation test, fused-qk packing removed after measurement, PP/HSDP fail-closed rejections, and the reworked perf section (comm decomposition + raw runs + 720p cell). LGTM.

hsliuustc0106:
@cr-gao The label-triggered Buildkite run ([build #14269](https://buildkite.com/vllm/vllm-omni/builds/14269)) has one failing job, "Simple · Other Test" (pytest -m 'core_model and cpu'); the other 22 jobs are green.

First error — `tests/model_tests/diffusion/test_alignment.py:105`, `test_no_excluded_models_have_test_settings`:

```
AssertionError: assert 1 == 0
 +  where 1 = len({'SanaImageToVideoPipeline'})
```

Root cause: #5508 put `SanaImageToVideoPipeline` in `EXCLUDED_MODELS` because "the common harness does not yet define an image-to-video task" — the exact limitation this PR lifts by adding `tiny_sana_video_i2v_builder` and a `DIFFUSION_TEST_SETTINGS` entry. The stale exclusion just needs to go: delete the `"SanaImageToVideoPipeline"` entry and its comment block in `EXCLUDED_MODELS` (test_alignment.py:65-70). Local check: `pytest tests/model_tests/diffusion/test_alignment.py -m 'core_model and cpu'` — everything else in the matrix (incl. the multi-GPU lanes) is already green.

cr-gao:
Good catch, the exclusion was exactly the limitation this PR lifts. Dropped it in `6cda9f0eb` (local `pytest tests/model_tests/diffusion/test_alignment.py -m 'core_model and cpu'` red→green, full alignment + sana_video suites pass); #5940 rebased on top.

cr-gao:
@hsliuustc0106 The alignment failure you flagged is fixed in 6cda9f0eb and all checks are green now. Ready to merge from my side. #5940 and #5882 are stacked on this branch, so this one merges first.

hsliuustc0106 (approved):

PR #5864, opened 2026-08-06
[Diffusion] Fix DLO DP concurrent request execution

lishunyang12:
## Why

Follow-up to #5764. DLO + AllGather can schedule one request per DP replica, but the multiprocess result path assumed one rank-0 result queue. An invalid or control-flow-incompatible request could fail on one DP replica while another replica entered AllGather, leaving the request wave hung.

This PR is intentionally scoped to DLO/DP request correctness, result routing, IPC, and shared worker lifecycle behavior.

## Architecture

```mermaid
flowchart TB
    U[API / Offline request] --> E[AsyncOmni / DiffusionEngine]
    E --> S[RequestScheduler<br/>Collect up to DP-size requests]
    S --> G{DLO + AllGather<br/>DP wave?}

    G -- No --> N[Normal single-request path]
    G -- Yes --> V[Wave preflight<br/>Shape / CFG / Steps / LoRA<br/>extra_args / non-empty prompt]
    V -- Incompatible --> REJECT[Reject the whole wave<br/>before worker dispatch]
    V -- Compatible --> EX[MultiprocDiffusionExecutor<br/>collective_rpc]
    N --> EX
    EX --> B[Broadcast MessageQueue<br/>same wave_id to every rank]

    subgraph R0[DP replica 0: request A]
        W0[Rank 0<br/>TP primary]
        W1[Rank 1<br/>TP worker]
    end

    subgraph R1[DP replica 1: request B]
        W2[Rank 2<br/>TP primary]
        W3[Rank 3<br/>TP worker]
    end

    B --> W0
    B --> W1
    B --> W2
    B --> W3

    W0 <-->|Layer weight AllGather| W2
    W1 <-->|Layer weight AllGather| W3

    W0 --> Q0[Rank 0 result queue]
    W2 --> Q2[Rank 2 result queue]
    W1 -. Non-primary: no reply .-> NR[Execute only]
    W3 -. Non-primary: no reply .-> NR

    Q0 --> P0[Result pump 0]
    Q2 --> P2[Result pump 2]
    P0 --> C[Future / sync result buffer]
    P2 --> C
    C --> SORT[Sort by dp_rank]
    SORT --> OUT[Return request A and B independently]

    EX -- Wave timeout --> F[_is_failed = True]
    F --> SD[shutdown<br/>single termination owner]
    SD --> CLEAN[Send shutdown<br/>global 15s join]
    CLEAN -->|Survivors| TERM[Terminate survivors<br/>global 5s join]
```

## Out of scope

Per-replica SP communication groups and rank-local mmap sharing are a larger topology redesign and belong in a follow-up. This PR keeps the existing cross-DP AllGather topology, admits only compatible request waves, and fails closed on unforeseen asymmetry. It does not claim the pre-existing DLO memory reduction as a new benefit.

### DIFF
--- a/tests/diffusion/test_ipc_async.py
+++ b/tests/diffusion/test_ipc_async.py
@@ -137,6 +137,29 @@ class TestPackDiffusionOutputShmWithD2hStream:
             d2h_stream=d2h,
         )
 
+    def test_dp_tagged_output_path_with_stream(self, mocker):
+        mock_pack_fields = mocker.patch(
+            "vllm_omni.diffusion.ipc._pack_diffusion_fields",
+        )
+        d2h = mocker.MagicMock()
+        result = DiffusionOutput()
+        tagged = {"dp_rank": 1, "output": result}
+        pack_diffusion_output_shm(tagged, d2h_stream=d2h)
+        mock_pack_fields.assert_called_once_with(result, d2h_stream=d2h)
+
+    def test_rpc_envelope_dp_tagged_output_path_with_stream(self, mocker):
+        mock_pack_fields = mocker.patch(
+            "vllm_omni.diffusion.ipc._pack_diffusion_fields",
+        )
+        d2h = mocker.MagicMock()
+        result = DiffusionOutput()
+        envelope = {
+            "type": "diffusion_rpc_result",
+            "result": {"dp_rank": 1, "output": result},
+        }
+        pack_diffusion_output_shm(envelope, d2h_stream=d2h)
+        mock_pack_fields.assert_called_once_with(result, d2h_stream=d2h)
+
     def test_runner_output_path_with_stream(self, mocker):
         mock_pack_fields = mocker.patch(
             "vllm_omni.diffusion.ipc._pack_diffusion_fields",
--- a/tests/diffusion/test_ipc_async.py
+++ b/tests/diffusion/test_ipc_async.py
@@ -137,27 +137,16 @@ class TestPackDiffusionOutputShmWithD2hStream:
             d2h_stream=d2h,
         )
 
-    def test_dp_tagged_output_path_with_stream(self, mocker):
+    @pytest.mark.parametrize("wrapped_in_rpc_envelope", [False, True])
+    def test_dp_tagged_output_path_with_stream(self, mocker, wrapped_in_rpc_envelope):
         mock_pack_fields = mocker.patch(
             "vllm_omni.diffusion.ipc._pack_diffusion_fields",
         )
         d2h = mocker.MagicMock()
         result = DiffusionOutput()
         tagged = {"dp_rank": 1, "output": result}
-        pack_diffusion_output_shm(tagged, d2h_stream=d2h)
-        mock_pack_fields.assert_called_once_with(result, d2h_stream=d2h)
-
-    def test_rpc_envelope_dp_tagged_output_path_with_stream(self, mocker):
-        mock_pack_fields = mocker.patch(
-            "vllm_omni.diffusion.ipc._pack_diffusion_fields",
-        )
-        d2h = mocker.MagicMock()
-        result = DiffusionOutput()
-        envelope = {
-            "type": "diffusion_rpc_result",
-            "result": {"dp_rank": 1, "output": result},
-        }
-        pack_diffusion_output_shm(envelope, d2h_stream=d2h)
+        value = {"type": "diffusion_rpc_result", "result": tagged} if wrapped_in_rpc_envelope else tagged
+        pack_diffusion_output_shm(value, d2h_stream=d2h)
         mock_pack_fields.assert_called_once_with(result, d2h_stream=d2h)
 
     def test_runner_output_path_with_stream(self, mocker):
--- a/tests/diffusion/test_multiproc_engine_concurrency.py
+++ b/tests/diffusion/test_multiproc_engine_concurrency.py
@@ -276,6 +276,8 @@ class TestRequestModeDispatch:
         def worker():
             request = req_q.get(timeout=2)
             assert "rpc_id" not in request
+            assert request["output_rank"] is None
+            assert request["collect_rank_status"] is False
             wave_id = request["wave_id"]
             for dp_rank in range(2):
                 res_q.put(
@@ -296,6 +298,83 @@ class TestRequestModeDispatch:
         thread.join(timeout=2)
 
         assert [result.error for result in results] == ["0", "1"]
+
+    def test_step_execution_reads_each_dp_primary_result_queue(self):
+        executor, _, _ = _make_executor(num_gpus=4)
+        executor.od_config = SimpleNamespace(
+            step_execution=True,
+            parallel_config=SimpleNamespace(
+                data_parallel_size=2,
+                tensor_parallel_size=2,
+                sequence_parallel_size=1,
+                pipeline_parallel_size=1,
+                cfg_parallel_size=1,
+            ),
+        )
+        request: dict = {}
+        executor._broadcast_mq = SimpleNamespace(enqueue=lambda value: request.update(value))
+
+        def result_queue(dp_rank):
+            return SimpleNamespace(
+                dequeue=lambda timeout=None: {
+                    "dp_rank": dp_rank,
+                    "output": _tagged_output(str(dp_rank)),
+                    "wave_id": request["wave_id"],
+                }
+            )
+
+        non_primary_1 = MagicMock()
+        non_primary_3 = MagicMock()
+        executor._result_mqs = [result_queue(0), non_primary_1, result_queue(1), non_primary_3]
+        executor._result_mq = executor._result_mqs[0]
+
+        results = executor.collective_rpc(
+            "execute_model",
+            unique_reply_rank=None,
+            exec_all_ranks=True,
+        )
+
+        assert [result.error for result in results] == ["0", "1"]
+        non_primary_1.dequeue.assert_not_called()
+        non_primary_3.dequeue.assert_not_called()
+
+    @pytest.mark.parametrize("empty_prompt", ["", {"prompt": ""}])
+    def test_dlo_dp_rejects_empty_prompt_before_worker_dispatch(self, empty_prompt):
+        executor, _, _ = _make_executor(num_gpus=2)
+        executor.od_config = SimpleNamespace(
+            step_execution=False,
+            parallel_config=SimpleNamespace(data_parallel_size=2),
+            enable_distributed_layerwise_offload=True,
+            dlo_use_allgather=True,
+        )
+        executor.collective_rpc = Mock()
+        scheduler_output = _make_sched_output("invalid", "valid")
+        scheduler_output.scheduled_new_reqs[0].req.prompt = empty_prompt
+
+        with pytest.raises(ValueError, match="non-empty prompt"):
+            executor.execute_request(scheduler_output)
+
+        executor.collective_rpc.assert_not_called()
+
+    def test_dlo_dp_valid_wave_still_runs_after_rejected_wave(self):
+        executor, _, _ = _make_executor(num_gpus=2)
+        executor.od_config = SimpleNamespace(
+            step_execution=False,
+            parallel_config=SimpleNamespace(data_parallel_size=2),
+            enable_distributed_layerwise_offload=True,
+            dlo_use_allgather=True,
+        )
+        executor.collective_rpc = Mock(return_value=[_tagged_output("A"), _tagged_output("B")])
+        invalid_wave = _make_sched_output("invalid", "valid")
+        invalid_wave.scheduled_new_reqs[0].req.prompt = ""
+
+        with pytest.raises(ValueError, match="non-empty prompt"):
+            executor.execute_request(invalid_wave)
+
+        result = executor.execute_request(_make_sched_output("A", "B"))
+
+        assert [output.result.error for output in result.runner_outputs] == ["A", "B"]
+        executor.collective_rpc.assert_called_once()
 
 
 # ───────────────── concurrent collective RPC ─────────────────
--- a/tests/diffusion/test_multiproc_engine_concurrency.py
+++ b/tests/diffusion/test_multiproc_engine_concurrency.py
@@ -376,6 +376,86 @@ class TestRequestModeDispatch:
         assert [output.result.error for output in result.runner_outputs] == ["A", "B"]
         executor.collective_rpc.assert_called_once()
 
+    def test_dlo_dp_allows_shared_default_denoise_steps(self):
+        executor, _, _ = _make_executor(num_gpus=2)
+        executor.od_config = SimpleNamespace(
+            step_execution=False,
+            parallel_config=SimpleNamespace(data_parallel_size=2),
+            enable_distributed_layerwise_offload=True,
+            dlo_use_allgather=True,
+        )
+        executor.collective_rpc = Mock(return_value=[_tagged_output("A"), _tagged_output("B")])
+        scheduler_output = _make_sched_output("A", "B")
+        for new_req in scheduler_output.scheduled_new_reqs:
+            new_req.req.sampling_params.num_inference_steps = None
+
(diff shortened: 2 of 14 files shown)


hsliuustc0106:
@yuanheng-zhao @asukaqaq-s PTAL

vllm_omni/diffusion/worker/diffusion_worker.py
```diff
@@ -925,6 +928,40 @@ def _async_output_loop(self):
                     )
                 )
 
+    def shutdown(self) -> None:
```
yuanheng-zhao, line 933: Is this aiming for DLO scenarios or common cases

vllm_omni/diffusion/ipc.py
```diff
@@ -109,6 +106,45 @@ def _pack_tensor_if_large(
     return val
 
 
+def _ndarray_to_shm(array: np.ndarray) -> dict[str, Any]:
```
yuanheng-zhao, line 130: This func quite similar to `_tensor_to_shm`. Better merge or adapt

vllm_omni/diffusion/ipc.py
```diff
@@ -109,6 +106,45 @@ def _pack_tensor_if_large(
...
+        "name": shm.name,
+        "shape": list(array.shape),
+        "numpy_dtype": str(array.dtype),
+        "nbytes": array.nbytes,
+    }
+    shm.close()
+    return handle
+
+
+def _ndarray_from_shm(handle: dict[str, Any]) -> np.ndarray:
```
yuanheng-zhao, line 137: Please check the impl and usage of `_tensor_from_shm`

vllm_omni/diffusion/worker/diffusion_worker.py
```diff
@@ -925,6 +928,40 @@ def _async_output_loop(self):
...
+        """Stop background work and release worker-owned IPC resources."""
+        self._running = False
+
+        if self._async_output_queue is not None:
+            self._async_output_queue.put(None)
+        if self._async_output_thread is not None:
+            self._async_output_thread.join(timeout=10)
```
yuanheng-zhao: nit: Better find a place to hold configs such like `XX_JOIN_TIMEOUT_S=XXX` rather than magic #`timeout=10`

hsliuustc0106 (changes requested):
I reproduced the main DLO/DP fix on head `e656fc3a` with MiniMax-H3 FL2VA, vLLM 0.26.0, PyTorch 2.11.0+cu130, and 2x L20X (the PR reports 2x B300). I used the PR topology and `enforce_eager=True` because the execution mode is not stated.

The functional results are good:
- the empty+valid wave rejected both requests in 5.8–6.1 ms and named the empty request ID;
- the next DP2 wave in the same engine completed in 31.6–34.5 s;
- both outputs matched `[124, 768, 1344, 3]` video and `[1, 2, 165600]` audio, with different frame hashes;
- the seven changed test modules passed (`138 passed`).

However, the clean-shutdown claim did not reproduce. Both full E2E runs logged:

```text
[AsyncOmniEngine] Orchestrator did not stop within 30.0 seconds; continuing cleanup
[AsyncOmniEngine] Deferring StageRuntime shutdown until the Orchestrator exits
resource_tracker: There appear to be 1 leaked shared_memory objects to clean up at shutdown
```

The second run closed the asyncio loop and released output references before synchronous shutdown, with the same result.

Please address this before marking the PR ready:
1. Make the real MiniMax-H3 DP2 lifecycle exit without the orchestrator timeout, deferred StageRuntime shutdown, or leaked shared memory.
2. Add a multiprocess lifecycle regression that transfers a large nested NumPy result through the per-worker queues and asserts result pumps, worker processes, and shared-memory resources are gone after shutdown. The current mock-only shutdown tests do not cover this failure.
3. Publish the exact two-GPU E2E script/command, environment, and execution mode, and rerun it with complete shutdown logs. Since `WorkerProc.shutdown()` is shared code, please also validate an ordinary non-DLO request-mode case.
4. Clarify the “before” revision: the literal parent `48298030` rejects this DP2 configuration during initialization, while the archived force-pushed head `64b19970` does reproduce the >115 s hang.

The deadlock prevention works, but the lifecycle behavior claimed by this PR is not clean yet.

vllm_omni/diffusion/executor/multiproc_executor.py
```diff
@@ -424,6 +456,12 @@ def execute_request(self, scheduler_output: DiffusionSchedulerOutput) -> BaseRun
                     "share identical extra_args. Different extra_args can change "
                     "the forward schedule and cause AllGather deadlock."
                 )
+            empty_prompt_ids = [nr.request_id for nr in new_reqs if _is_empty_dp_prompt(nr.req.prompt)]
```
asukaqaq-s, line 486: Currently, we only check step and extra_args. Do we also need to check CFG (= 1 or ≠ 1), which may cause the number of execution steps to differ? Another potential issue is that different shapes or token lengths may cause the batch to wait for the slowest sample._

asukaqaq-s:
Should DLO be restricted to a single replica? For multi-GPU SP:

```text
DP replica 0:
  SP ranks [0,1,2,3] → request A
  Weight AllGather within the SP group

DP replica 1:
  SP ranks [4,5,6,7] → request B
  Weight AllGather within the SP group
```

We would disable AllGather across DP replicas and only allow co-located replicas to share CPU memory via `mmap` to the same directory. Otherwise, all DP replicas on the node must process identical token requests, making asynchronous request handling difficult.

The communication topology must be configured accordingly. For eight GPUs with `DP=2` and `SP=4`, create two communication groups, each with a group size of four.

The original approach is more suitable for offline batch inference:

```text
SP/TP groups (model parallelism for one request):
  [0,1,2,3] → request A
  [4,5,6,7] → request B

DP peer groups (same SP/TP coordinate across replicas):
  [0,4] → AllGather
  [1,5] → AllGather
  [2,6] → AllGather
  [3,7] → AllGather
```

Requests A and B may require kernel-level padding to ensure identical shapes; otherwise, the cross-replica AllGather introduces many synchronization points.

david6666666:
The request batching mechanism within the request queue should ideally be configurable to accommodate various request scenarios.

image

david6666666:
before:
image

after:
image

vllm_omni/diffusion/worker/diffusion_worker.py
lishunyang12: Addressed in e55738af: the magic value is now the named module constant `_ASYNC_OUTPUT_THREAD_JOIN_TIMEOUT_S = 10.0`.

vllm_omni/diffusion/ipc.py
lishunyang12: Addressed in e55738af. Tensor and NumPy serialization now share `_array_to_shm`; the tensor wrapper only normalizes the dtype marker. This removes the duplicated shared-memory allocation/copy/metadata logic.

vllm_omni/diffusion/worker/diffusion_worker.py
lishunyang12: This is common request-mode lifecycle cleanup, not DLO-only. In e55738af I named the join timeout (`_ASYNC_OUTPUT_THREAD_JOIN_TIMEOUT_S`) and added a real spawn-based per-worker result-queue regression. DLO/DP makes the lifecycle issue easier to trigger, but the shutdown path remains shared by all request-mode workers.

vllm_omni/diffusion/ipc.py
lishunyang12: Addressed in e55738af. Both deserializers now use `_array_from_shm`, which attaches, copies, closes, and unlinks the segment in one place; `_tensor_from_shm` additionally restores the torch dtype. The new spawn-based test round-trips nested NumPy outputs through two worker-owned queues and verifies every SHM name is gone afterward.

vllm_omni/diffusion/executor/multiproc_executor.py
lishunyang12: Addressed in e55738af by reusing the complete scheduler sampling-compatibility key before DP dispatch. It covers output shape, CFG mode/value, denoise steps/sigmas, output count, and LoRA, while `extra_args` is still checked separately. Different prompt token lengths remain valid: ranks execute independent requests and only need an identical collective schedule, not identical compute time. A bounded DP-wave timeout now fails closed if an unforeseen asymmetric path still occurs. Tests cover mismatched steps, CFG, and width being rejected before dispatch, plus valid shared defaults (`num_inference_steps=None`).

lishunyang12:
Addressed the review feedback in `e55738af` and `84a05548`.

@hsliuustc0106:

1. Lifecycle: shutdown now closes the diffusion engine before waiting for dispatch tasks. Worker joins use one global 15-second deadline, followed by termination of all survivors and one global 5-second grace deadline. The previous code could spend the full timeout once per worker.
2. Multiprocess regression: `test_spawned_per_worker_queues_release_nested_numpy_shm` starts two real spawn children, creates one production queue and result pump per worker, transfers nested video/audio NumPy arrays through SHM, shuts down pumps, queues, and workers, then asserts every child exited and every SHM name is gone. The seven focused modules now report `145 passed` with no `resource_tracker` warning.
3. Reproduction: `examples/offline_inference/minimax_h3/dlo_dp2_lifecycle.py` now provides the exact two-GPU `dlo-dp2` and ordinary non-DLO TP2 `request` modes. The PR body contains the complete environment and commands, and explicitly states `enforce_eager=True`.
4. Baseline: the PR body now distinguishes literal parent `48298030`, which rejects the configuration, from archived head `64b19970`, which reproduces the greater-than-115-second hang.

@david6666666 request admission waiting is now configurable as requested. `request_batch_max_wait_ms=0` remains zero and disables waiting; a nonzero value such as the runner value of 500 ms explicitly enables coalescing.

@asukaqaq-s I agree that per-replica SP groups plus rank-local mmap sharing are the cleaner long-term topology. That redesign is intentionally out of this DP correctness PR. The current cross-DP AllGather path now accepts only requests with a complete matching scheduler compatibility key and fails closed on a partial-wave timeout, so an uncertain process group is never reused.

lishunyang12:
DCO follow-up: the two review-fix commits were rewritten only to add the required `Signed-off-by` trailers. Code content is unchanged.

- `e55738af` is now `b7c59afa`
- `84a05548` is now `74d35ffa`

The PR description has been updated to the signed hashes.

lishunyang12:
Target-GPU rerun is now complete on head `2d190bf6` using 2x B300, vLLM 0.26.0, PyTorch 2.11.0+cu130, `CUDNN_ATTN`, and `enforce_eager=True`.

The first current-head run exposed a real lifecycle regression: both ordinary TP2 and DLO/DP2 generated successfully, but `engine.close()` returned with both workers alive. The root cause was the interaction between two states: the orchestrator pre-marks `client._shutting_down = True` to suppress false death reports, while the new idempotence guard returned immediately when it saw that pre-mark. Consequently, the actual `DiffusionEngine.close()` was skipped.

Commit `2d190bf6` separates shutdown-in-progress from shutdown-completed and serializes teardown. Added regression coverage verifies that an orchestrator-pre-marked client still closes the engine and that a second shutdown remains idempotent.

Post-fix GPU results:

- Ordinary non-DLO TP2: generation 19.296 s, shutdown 4.848 s, `active_children=[]`.
- DLO AllGather DP2: empty+valid wave rejected both requests in 0.005 s; the same engine completed two recovery requests in 26.937 s; shutdown 14.204 s; `active_children=[]`.
- Both modes produced video `[124, 768, 1344, 3]` and audio `[1, 2, 165600]`.
- Both logs contain worker `Shutdown complete`; neither contains orchestrator timeout, deferred StageRuntime shutdown, unexpected worker death, `resource_tracker`, or leaked shared memory.
- The DLO log tracebacks are the intentionally injected empty-prompt wave and are followed by successful same-engine recovery plus exit status 0.

Focused CPU regression: `108 passed`; inline-client module: `11 passed`; Ruff and `git diff --check` pass.

@hsliuustc0106 PTAL when convenient.

vllm_omni/diffusion/executor/multiproc_executor.py
```diff
@@ -369,6 +411,29 @@ def _monitor() -> None:
...
+        logger.error(
+            "DLO DP collective wave timed out after %.1fs; terminating all workers: %s",
+            _DLO_DP_WAVE_TIMEOUT_S,
+            exc,
+        )
+        self._is_failed = True
+        for process in self._processes:
+            if process.is_alive():
+                process.terminate()
+        self.shutdown()
```
yuanheng-zhao: nit: redundant terminates (another one in `self.shutdown()`)

Consider letting `shutdown()` own termination and only setting `_is_failed` here.

tests/diffusion/test_ipc_async.py
```diff
@@ -137,6 +137,29 @@ def test_rpc_envelope_path_with_stream(self, mocker):
...
+            "vllm_omni.diffusion.ipc._pack_diffusion_fields",
+        )
+        d2h = mocker.MagicMock()
+        result = DiffusionOutput()
+        envelope = {
+            "type": "diffusion_rpc_result",
+            "result": {"dp_rank": 1, "output": result},
+        }
+        pack_diffusion_output_shm(envelope, d2h_stream=d2h)
+        mock_pack_fields.assert_called_once_with(result, d2h_stream=d2h)
```
yuanheng-zhao: These two tests are nearly duplicated

tests/diffusion/test_multiproc_engine_concurrency.py
```diff
@@ -297,6 +299,163 @@ def worker():
...
+                    "dp_rank": dp_rank,
+                    "output": _tagged_output(str(dp_rank)),
+                    "wave_id": request["wave_id"],
+                }
+            )
+
+        non_primary_1 = MagicMock()
+        non_primary_3 = MagicMock()
+        executor._result_mqs = [result_queue(0), non_primary_1, result_queue(1), non_primary_3]
+        executor._result_mq = executor._result_mqs[0]
```
yuanheng-zhao: I think this test is weak, for it already assumes and constructs the dp primary ranks are 0 and 2.

yuanheng-zhao (commented):
Generally LGTM

tests/diffusion/test_ipc_async.py
lishunyang12: Yep, folded them into one parametrized test in 07fc831e.

tests/diffusion/test_multiproc_engine_concurrency.py
lishunyang12: Good point — the test now makes every rank queue viable and checks multiple TP/SP/PP/CFG layouts in 07fc831e.

vllm_omni/diffusion/executor/multiproc_executor.py
lishunyang12: Good catch — fixed in 07fc831e, shutdown() owns worker termination now.

lishunyang12:
Addressed all latest inline review feedback.

vllm_omni/diffusion/executor/multiproc_executor.py
lishunyang12: Good point. For a DLO+DP AllGather wave, the preflight now reuses `RequestBatchSamplingParamsKey`, so shape (height/width/frames), CFG/guidance values, denoise schedule, output count, LoRA, and `extra_args` must match exactly; prompt text/token length and seed/generator may differ. H3 is CFG-distilled and always runs one transformer forward per step, so CFG does not alter its collective count, while different token lengths are allowed and only add a straggler/perf cost because they do not change the DLO AllGather sequence.

hsliuustc0106:
LGTM

hsliuustc0106 (approved):

nannaer:
The blog says DLO uses weight sharding + AllGather to avoid the N × model_size host-memory overhead across ranks. If I enable TP, each TP rank already owns only a shard of the model weights. Does TP already solve this host-memory issue? If so, when would DLO still be preferable to simply using TP + regular layerwise offload? @lishunyang12

PR #5866, opened 2026-08-06
[Bugfix][Diffusion] Fix FLASH_ATTN cross-attention key-padding unpad

### LINKED ISSUE #5865: [Bug] FLASH_ATTN mishandles cross-attention key-padding masks when query_len != key_len
## Summary

On the `FLASH_ATTN` diffusion attention backend, a cross-attention call with a
2D key-padding mask and `query_len != key_len` zeroes out most query positions
in the output. The same inputs are handled correctly by `TORCH_SDPA` and
`CUDNN_ATTN`. This lives in shared attention infrastructure, so it affects any
diffusion model that runs prompt cross-attention with a padded key mask on
FLASH_ATTN, not just one model.

## Affected component

The FLASH varlen-masked path (`FlashAttentionImpl._forward_varlen_masked` ->
HF-derived `_upad_input` in `backends/utils/fa.py`) is written for
self-attention, where `query_len == key_len`. In cross-attention the image
query (thousands of tokens) is much longer than the text key (a few hundred).
The `_upad_input` else-branch slices the mask with `attention_mask[:, -query_length:]`
(assuming left padding on a self-attention sequence) and then unpads the
*image query* with the *text* key mask. Only the query rows whose index happens
to fall on a True mask position survive; `_pad_input` scatters everything else
back as zeros, so the prompt conditioning is effectively dropped.

With `batch_size >= 2` it is worse: the kept indices are computed on the
flattened `[batch, key_len]` mask but applied to the flattened
`[batch, query_len]` query, so the surviving rows are not even the leading
rows of each batch element.

## Impact

Discovered on SANA-Video 2B (native pipelines) on A800 (Ampere), where the
platform default backend is FLASH_ATTN. A single transformer forward on real
480p weights (tp=1), compared against a frozen golden output tensor:

| backend    | cosine  | rel_l2 | max_abs | vs thresholds (0.999 / 0.017 / 0.2) |
|------------|---------|--------|---------|-------------------------------------|
| FLASH_ATTN | 0.9157  | 0.412  | 8.64    | FAIL                                |
| TORCH_SDPA | 0.99993 | 0.0122 | 0.086   | PASS                                |
| CUDNN_ATTN | 0.99991 | 0.0141 | 0.125   | PASS                                |

Over 50 CFG denoising steps this compounds into unusable video (480p T2V lands
at ~0.29 SSIM on FLASH_ATTN vs ~0.94 on TORCH_SDPA / CUDNN_ATTN).

## Steps to reproduce (no model weights required)

Run a single cross-attention forward with `query_len != key_len` and a padded
key mask through FLASH_ATTN, and compare against `scaled_dot_product_attention`:

```python
import torch
import torch.nn.functional as F

torch.manual_seed(0)
device = "cuda"
b, heads, dim = 1, 8, 64
q_len, k_len = 4096, 300          # image query vs text key: q_len != k_len
q = torch.randn(b, heads, q_len, dim, device=device, dtype=torch.bfloat16)
k = torch.randn(b, heads, k_len, dim, device=device, dtype=torch.bfloat16)
v = torch.randn(b, heads, k_len, dim, device=device, dtype=torch.bfloat16)

# key-padding: only the first 120 text tokens are valid.
keep = torch.zeros(b, k_len, dtype=torch.bool, device=device)
keep[:, :120] = True
attn_mask = keep[:, None, None, :]            # broadcast over heads and query

ref = F.scaled_dot_product_attention(q, k, v, attn_mask=attn_mask)

# Route the identical q/k/v/keep-mask through the vLLM-Omni FLASH_ATTN
# cross-attention path (OmniAttention, role='cross', backend forced to
# FLASH_ATTN) and diff against `ref`.
```

## Expected vs actual

- **Expected:** the FLASH_ATTN output matches the SDPA/CUDNN output; every query
  row carries a real value.
- **Actual:** `max_abs` diff ~2.08; in the FLASH_ATTN output most query rows
  are ~0, because the key mask is (incorrectly) used to unpad the query.

## Environment

- GPU: NVIDIA A800-SXM4-80GB (Ampere), also reproduced on RTX 4070 (Ada)
- Platform default diffusion attention backend on these GPUs: FLASH_ATTN
- Model where it surfaced: SANA-Video 2B native T2V / I2V pipelines
- Correct backends: TORCH_SDPA, CUDNN_ATTN

## Suggested fix

In this scenario the query has no padding at all — the mask describes keys
only. So when `query_len != key_len`, keep the query dense and unpad only
K/V with the key mask, instead of reusing the key mask for the query.
Alternatively, route cross-attention with a key-only padding mask onto the
dense masked kernel that TORCH_SDPA / CUDNN_ATTN use.

## Workaround

Run the affected model on `TORCH_SDPA` or `CUDNN_ATTN` via
`--diffusion-attention-backend`. On Blackwell the platform default already
selects `CUDNN_ATTN`, so the issue does not surface there; it appears on GPUs
whose default resolves to `FLASH_ATTN` (e.g. Ampere).

cr-gao:
## Purpose

Fixes #5865.

`FlashAttentionImpl._forward_varlen_masked` reuses the 2D key-padding mask to
unpad the query. That is only valid for self-attention (`query_len == key_len`).
In cross-attention (e.g. image query of thousands of tokens vs a padded text
key of a few hundred), the `_upad_input` else-branch drops every query row
whose index does not land on a True key-mask position, and `_pad_input`
scatters those rows back as zeros — the prompt conditioning is effectively
lost. With `batch_size >= 2` the kept rows are additionally misaligned,
because indices computed on the `[batch, key_len]` mask are applied to the
`[batch, query_len]` query.

Observed on SANA-Video 2B on A800, where the platform default backend is
FLASH_ATTN: a single 480p transformer forward drops to cosine 0.9157 vs the
golden output (TORCH_SDPA / CUDNN_ATTN both give 0.9999), and 50 CFG steps
turn that into ~0.29 SSIM video. Details and a no-weights repro in #5865.

## Modification

In `_forward_varlen_masked`, when `query_len != key_len` the mask describes
keys only, so keep the query dense (all rows valid) and unpad only K/V:

- self-attention (`query_len == key_len`): unchanged, still goes through
  `_upad_input`.
- cross-attention: `_unpad_input` on K/V with the key mask, dense
  `cu_seqlens_q`, and a plain reshape on the output instead of the
  zero-scatter `_pad_input`.

## Test Plan

New regression test `test_cross_attn_key_padding_vs_sdpa` in
`tests/diffusion/attention/test_flash_attn.py`: batch 2, `q_len=256`,
`k_len=40`, 17 valid keys, FLASH_ATTN vs SDPA on identical inputs.

```bash
pytest -s -v tests/diffusion/attention/test_flash_attn.py
```

## Test Result

On RTX 4070 (FA via `vllm.vllm_flash_attn` varlen):

- Before the fix the new test fails with max diff 2.66 (most query rows zero).
- After the fix: max diff 0.000488, mean diff ~0 — within the same
  tolerances the existing self-attention FA-vs-SDPA case uses.
- `tests/diffusion/attention/` suite: 158 passed, 19 skipped, 1 pre-existing
  environmental failure (`test_kernels_hub_execution`, fails identically on
  clean `main`, needs kernels-hub network access).


cr-gao:
Verified on A800 (Ampere, the platform whose default backend resolves to FLASH_ATTN) with real SANA-Video 2B 480p weights — single transformer forward vs frozen golden, `FLASH_ATTN` forced for cross-attention, FA served by the `vllm.vllm_flash_attn` varlen fallback:

| | cosine | rel_l2 | max_abs | thresholds (0.999 / 0.017 / 0.2) |
|---|---|---|---|---|
| before fix | 0.9157 | 0.412 | 8.64 | FAIL |
| after fix | 0.99996 | 0.0096 | 0.0625 | PASS |

Reference backends on the same case: TORCH_SDPA 0.99993 / CUDNN_ATTN 0.99991.

`tests/diffusion/attention/test_flash_attn.py` also passes on the A800 (7 passed; that box additionally exercises the dense `flash_attn_func` path).

vllm_omni/diffusion/attention/backends/flash_attn.py
```diff
@@ -113,17 +113,29 @@ def _forward_varlen_masked(
...
             flash_attn_varlen_func,
         )
 
         assert attention_mask.ndim == 2, "attention_mask must be 2D, (batch_size, seq_len)"
-        query_length = query.size(1)
-        q, k, v, indices_q, (cu_seq_lens_q, cu_seq_lens_k), (max_length_q, max_length_k) = _upad_input(
-            query, key, value, attention_mask, query_length, _unpad_input
-        )
+        batch_size, query_length = query.shape[:2]
+        if query_length == key.size(1):
```
Gaohan123: Don’t infer self-attention from equal sequence lengths

`query_length == key.size(1)` does not imply self-attention. A cross-attention call can legitimately have equal Q/K lengths while its 2D mask still describes keys only. In that case this branch passes the key mask to `_upad_input`, which removes valid query rows and `_pad_input` restores them as zeros—the same silent output corruption this PR fixes for unequal lengths.

For example, with batch 2, Q/K length 4, and two valid keys per sample, the current branch keeps query rows `[0, 1, 4, 5]` and drops `[2, 3, 6, 7]`.

Use the existing semantic attention role instead of sequence-length equality to select the self-attention path, and add an equal-length cross-attention regression case.

cr-gao: Fixed.

Gaohan123:
@cr-gao Please resolve conflicts. Thanks

cr-gao:
> @cr-gao Please resolve conflicts. Thanks

Resolved, new code reads fa_deterministic in __init__, which leads to a conflict. Simply kept both. All tests passed.

tests/diffusion/attention/test_flash_attn.py
```diff
@@ -308,6 +308,54 @@ def test_flash_attn_func_preferred_over_varlen():
...
+    The mask covers only the key sequence; every query position must attend
+    to the valid keys. FA and SDPA outputs should be very close. k_len=256
+    covers equal Q/K lengths, which must not be mistaken for self-attention.
+    """
+    device = torch.device(current_omni_platform.device_type)
+    dtype = torch.bfloat16
+
+    batch_size = 2
+    q_len = 256
+    valid_len = 17
```
RuixiangMa: The test uses a single valid_len for all batch elements. If different batch elements had different valid key lengths, cu_seq_lens_k from _unpad_input would need to handle variable lengths correctly. The logic should be correct (since _unpad_input handles variable lengths), but it's untested.

tests/diffusion/attention/test_flash_attn.py
```diff
@@ -308,6 +308,54 @@ def test_flash_attn_func_preferred_over_varlen():
     print("✓ flash_attn_func forward works correctly!")
 
 
+@pytest.mark.skipif(not is_gpu, reason="FlashAttention requires CUDA or XPU")
```
RuixiangMa, line 311: NIT: If CI has no GPU, the cross-attention path is never tested. Suggest adding a CPU test that verifies the routing logic (is_cross_attn flag, cu_seq_lens_q construction, indices_q = None) even without FlashAttention.

tests/diffusion/attention/test_flash_attn.py
cr-gao: Variable key length has been added to test_cross_attn_key_padding_vs_sdpa.

tests/diffusion/attention/test_flash_attn.py
cr-gao: Added test_varlen_masked_routing_by_role

Gaohan123 (approved):
LGTM. Thanks

PR #5869, opened 2026-08-06
[Perf][MiniCPM-O-4.5]Add CUDA Graph For HiFTGenerator

sphinxkkkbc:
## Purpose

Reuse the in-tree `HiFTGenerator` and `mel_spectrogram` implementations introduced for CosyVoice3 in the Step-Audio2 and MiniCPM-o Code2Wav paths.

This PR also adds optional CUDA Graph support for MiniCPM-o HiFT inference. The wrapper:

- Pre-captures the fixed initial-chunk and cached intermediate-chunk shapes derived from the connector configuration.
- Uses batch size 1 by default.
- Lazily captures the variable-length final chunk during normal operation, because the preceding fixed-size chunks are covered by the pre-captured graphs.
- Bounds the number of lazily captured final-chunk graphs to prevent unbounded graph memory growth.
- Falls back to eager execution when the batch size is unsupported or the lazy-capture limit has been reached.

## Notes

 - The current PyTorch implementation of `torch.istft()` performs NOLA validation that introduces CPU-GPU synchronization, which is incompatible with CUDA Graph capture. This PR therefore captures the HiFT computation through magnitude and phase generation, while leaving `torch.istft()` and waveform clamping in eager mode.

- The pre-captured bucket sizes are derived from `codec_chunk_frames` and `codec_left_context_frames`, together with the Flow pre-lookahead length and token-to-mel upsampling ratio.

- `initial_codec_chunk_frames` is not currently enabled for MiniCPM-o 4.5, so this PR does not add a dedicated pre-captured bucket for it. If it is enabled in the future without updating the HiFT bucket derivation, the first chunk will be handled through lazy capture. Comments documenting this behavior have been added to the MiniCPM-o deployment YAML files.

- Because `torch.istft()` cannot be captured, the performance results below measure a partially graphed HiFT path: `HiFTGenerator._inference_pre_istft()` runs through CUDA Graph replay, while `HiFTGenerator._istft()` and waveform clamping remain eager.

- HiFT CUDA Graph execution is independent of Stage 2's `enforce_eager` setting. It remains enabled unless `enable_hift_graph` is set to `false`.

The primary purpose of this PR is to reuse the in-tree HiFT implementation. CUDA Graph support is included as an initial optimization of the HiFT path.

## Profile A/B

Profiled on a single NVIDIA A800 using the following commands:

```bash
vllm serve openbmb/MiniCPM-o-4_5 --omni \
    --trust-remote-code \
    --host 0.0.0.0 \
    --port 8091
```

```bash
curl http://localhost:8091/v1/chat/completions \
    -d '{
        "model": "openbmb/MiniCPM-o-4_5",
        "messages": [{
            "role": "user",
            "content": "Say hello, then introduce vLLM in one sentence."
        }],
        "modalities": ["text", "audio"],
        "chat_template_kwargs": {
            "use_tts_template": true
        }
    }'
```

### Before

HiFT profile before CUDA Graph optimization

### After

HiFT profile after CUDA Graph optimization

**HiFT latency per chunk:** approximately 30 ms → 5 ms
**Aggregate HiFT latency for the profiled request:** approximately 250 ms to 43 ms.

## Test Plan

- Compare the outputs of the in-tree `HiFTGenerator` and `mel_spectrogram` implementations against `flashcosyvoice`.
- Verify eager-versus-graph output parity for uncached and cached HiFT inputs.
- Verify lazy capture for previously unseen shapes and eager fallback after reaching the maximum number of lazy-captured graphs.
- Run the MiniCPM-o Code2Wav batching and pipeline configuration tests.

**vLLM Version:**  
0.26.0

**vLLM-Omni Commit:**  
8fa28d88e05ee767734adbefd51819db16d97665

## Test Result
```text
52 passed, 15 warnings
```

Command:

```bash
pytest -v \
    tests/model_executor/models/step_audio2/test_hift_parity.py \
    tests/model_executor/models/minicpmo_4_5/test_cuda_graph_wrapper.py
```

**BEFORE SUBMITTING:** read [[CONTRIBUTING.md](https://github.com/vllm-project/vllm-omni/blob/main/CONTRIBUTING.md)](https://github.com/vllm-project/vllm-omni/blob/main/CONTRIBUTING.md) and run the [[precheck-pr skill](https://github.com/vllm-project/vllm-omni/blob/main/.claude/skills/precheck-pr/SKILL.md)](https://github.com/vllm-project/vllm-omni/blob/main/.claude/skills/precheck-pr/SKILL.md) with the code agent for a self-check against project conventions.


amy-why-3459:
Hi @sphinxkkkbc — thanks for the HiFT vendor + CUDA Graph work. This looks useful for the MiniCPM-o Code2Wav path.

Just checking: is [#5869](https://github.com/vllm-project/vllm-omni/pull/5869) ready for review now (description / tests / profiling settled), or are there still planned follow-ups before we dig in?

If it’s ready, I’ll start a full pass.

sphinxkkkbc:
> Hi @sphinxkkkbc — thanks for the HiFT vendor + CUDA Graph work. This looks useful for the MiniCPM-o Code2Wav path.
> 
> Just checking: is [#5869](https://github.com/vllm-project/vllm-omni/pull/5869) ready for review now (description / tests / profiling settled), or are there still planned follow-ups before we dig in?
> 
> If it’s ready, I’ll start a full pass.

Hi @amy-why-3459,  thanks for checking. The HiFT path is ready for review, and the description has been updated accordingly.

I’m also working on vendoring the entire flow component and adding some vLLM-native operators, since the implementation in the external library leaves room for optimization. However, that would expand the scope significantly beyond HiFT. If you’d prefer HiFT and the full flow component to be vendored in a single PR, there is still more work to do. Otherwise, I think the current HiFT-focused PR is ready for a full review.

amy-why-3459:
Thanks for confirming HiFT-only is ready — that scope split looks right; please keep full flow vendoring as a follow-up PR.

P1 – device guard: enable_hift_graph: true unconditionally calls HiFTGraphWrapper.capture() (torch.cuda.*). On non-CUDA devices this will fail backend init instead of falling back to eager. Please gate on device.type == "cuda" (and log + disable otherwise).

P1 – tests vs description: The PR text mentions CUDA Graph wrapper unit tests, but the diff mostly adds flashcosyvoice parity + a fake inference shim. Can we add at least one CUDA-skipable parity test (replay vs hift.inference for uncached + cached shapes) and coverage for lazy-capture / max-lazy fallback?

P1 – pre_lookahead_len: int(getattr(lookahead_layer, "pre_lookahead_len")) is not None-safe when pre_lookahead_layer is missing. Please mirror BatchedToken2Wav._pre_lookahead_len().

P2 – drift risk: HiFTGraphWrapper.replay re-implements istft/clamp/cache finalize outside HiFTGenerator.inference. A shared finalize helper would make future changes safer.

P2 – perf / config notes: Please clarify that the 30→5 ms numbers are HiFT-scoped (istft still eager), independent of stage enforce_eager. Also document that capture buckets are derived from codec_chunk_frames / left context — if we later add initial_codec_chunk_frames, the first chunk will miss the pre-captured graphs and rely on lazy capture unless that shape is included.

Nits: SPDX header on cuda_graph_wrapper.py, rename capture_batch_size → capture_batch_sizes, and add logs for enable/capture/fallback.

sphinxkkkbc:
@amy-why-3459 Thanks for your feedback. All comments are resolved and descriptions updated. PTAL, thanks!

lishunyang12 (approved):
Reviewed the current head and ran the focused CUDA-graph, pipeline, batching, and HiFT parity tests (50 passed, 1 skipped). I found no additional blocker beyond the earlier review feedback.

PR #5459, opened 2026-07-27
[WIP] Omni Tiny Model Tests

alex-jw-brooks:
## Purpose
WIP integration for adding non-diffusion models to tiny model testing framework. / common tests + will add tests for Qwen3Omni.

Still refactoring a bit, and tests are not actually added yet, since I still need to integrate it to the Omni obj creation + case filtering, but opening as a draft in case anyone has comments, since I think it's easy to see the direction it is going in for code structure.

Current state since there is a lot moving around atm:

For the builders in this PR:
- [X] Tiny model builder implemented for Qwen3Omni
    - [X] Also support thinker only
- [X] Tiny model builder implemented for Bagel
- [x] Tiny model builder implemented for Qwen3TTS
     - Builder is implemented and loads without exploding, but need some tweaks for the tests 

Other things to do before this is out of wip (hopefully tomorrow):
- [ ] Ensure tests are deterministic (e.g., audio gen tests can fail for qwen3omni atm due to random weights creating empty audio)
- [ ] Add non-hacky fix for < 24 layer qwen3omni models
- [x] Add support for running text -> text and text -> audio request
- [ ] Clean up mark / device / sampling param / output handling to correctly handle multistage models
- [ ] Fail instead of skip when device conditions aren't met
- [ ] Add checks for multi output / determinism for all modalities instead of just image
- [X] Optimize startup time settings
   - This is mostly disabling prefix cache (to skip encoder cache profiling) and turning down GPU usage, because reducing it reduces the time for kv cache init in AR stages by a lot, and we don't need high utilization for AR models

CC @NickCao @yenuo26 @congw729

PR #5875, opened 2026-08-06
[Bugfix] Apply function module marker per-model instead of hardcoding diffusion

khairulkabir1661:
## Summary

The test files hardcoded `pytestmark = [pytest.mark.diffusion]`, so non-diffusion models (omni, TTS) added to the framework would get the wrong marker. This PR adds a `ModelTypeMarker` enum and a required `model_type_marker` field on `DiffusionModelTestOpts`, then applies it per-test-case via `pytest.param(..., marks=...)`.

## Test plan
- [x] All 54 tests collect successfully
- [x] `-m diffusion` still selects all current tests


tests/model_tests/diffusion/case_filtering.py
```diff
@@ -29,14 +29,20 @@
 MAX_CI_DEVICES = 4
 
 
-def get_test_group_marks(model_name: str, test_group: list[DiffusionAccs] | None, model_marks: list | None) -> list:
+def get_test_group_marks(
+    model_name: str,
+    test_group: list[DiffusionAccs] | None,
+    model_marks: list | None,
+    model_type_marker: str = "diffusion",
```
alex-jw-brooks, line 36: Can you make the model_type_marker an enum, use that as the type here to restrict the type, and also remove the default since it will always be passed?

tests/model_tests/diffusion/test_common_offline.py
```diff
@@ -8,18 +8,13 @@
...
 from tests.model_tests.diffusion.task_runners import (
-    run_and_validate_determinism,
     run_and_validate_image_to_image_request,
-    run_and_validate_image_to_video_request,
-    run_and_validate_multi_output,
+    run_and_validate_text_to_image_determinism,
+    run_and_validate_text_to_image_multi_output,
     run_and_validate_text_to_image_request,
     run_and_validate_text_to_video_request,
 )
```
alex-jw-brooks, line 18: Can you keep the comment and just update it to something like `# NOTE: Hardware and model type marks are added dynamically based on test requirements and model type`?

alex-jw-brooks (changes requested):
Thanks, I think this looks good to me, and will be helpful for https://github.com/vllm-project/vllm-omni/pull/5459 :) some small suggestions

tests/model_tests/diffusion/case_filtering.py
khairulkabir1661: did

tests/model_tests/diffusion/test_common_offline.py
khairulkabir1661: Did

tests/model_tests/diffusion/test_common_online.py
```diff
@@ -26,10 +26,7 @@
     run_and_validate_online_text_to_video_request,
 )
 
-# NOTE : Hardware marks are added dynamically based on test requirements
-pytestmark = [pytest.mark.diffusion, pytest.mark.xdist]
```
alex-jw-brooks, line 30: Can you make sure not to remove the `xdist` mark? This is needed for running tests concurrently, which speeds things up a lot in the CI

alex-jw-brooks (changes requested):
I think the `xdist` mark was removed. Can you please add it back and also fix DCO?

tests/model_tests/diffusion/test_common_online.py
khairulkabir1661: yes, did it.

tests/model_tests/diffusion/test_common_offline.py
```diff
@@ -19,8 +19,8 @@
...
 
-# NOTE : Hardware marks are added dynamically based on test requirements
-pytestmark = [pytest.mark.diffusion, pytest.mark.xdist]
+# NOTE: Hardware and model type marks are added dynamically based on test requirements and model type
+pytestmark = [pytest.mark.xdist, pytest.mark.xdist_group(name="offline_diffusion")]
```
alex-jw-brooks: I think this should just be `pytestmark = [pytest.mark.xdist]` for both this file and the online file 🙂  Adding `xdist_group`s would constrain all of the tests in the online / offline modules to single respective workers, and we need xdist to be able to distribute the tests, since they are already clumped for setup per model since each task runs as a subtest

alex-jw-brooks (changes requested):

tests/model_tests/diffusion/test_common_offline.py
khairulkabir1661: Done

khairulkabir1661:
@alex-jw-brooks, thanks, now please review.

alex-jw-brooks (approved):
Thanks!

PR #5876, opened 2026-08-06
[Bugfix] Unify model settings and tiny model resolution for all model types

khairulkabir1661:
## Summary

Tiny model test framework was hardcoded to diffusion models in three places, blocking omni/TTS support. This PR adds a unified `MODEL_SETTINGS` dict, makes `resolve_tiny_model_path()` fall back to HF model ID matching, and broadens the runtime marker check to include `omni` and `tts`.

## Test plan

- [x] `pytest --collect-only -m diffusion tests/model_tests/` still collects 12 tests (Flux2Klein + LTX2)
- [x] `test_alignment.py` still uses `DIFFUSION_TEST_SETTINGS` directly
- [x] `resolve_tiny_model_path()` resolves omni models by HF model ID (no more `ValueError`)
- [x] Runtime fixtures substitute tiny models for `omni`/`tts` markers, not just `diffusion`

### DIFF
--- a/tests/helpers/runtime.py
+++ b/tests/helpers/runtime.py
@@ -56,6 +56,16 @@ from vllm_omni.platforms import current_omni_platform
 
 logger = init_logger(__name__)
 
+_MODEL_TYPE_MARKERS = ("diffusion", "omni", "tts")
+
+
+def _has_model_type_marker(request) -> bool:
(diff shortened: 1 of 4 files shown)


NickCao:
See also https://github.com/vllm-project/vllm-omni/pull/5459

PR #5877, opened 2026-08-06
[Bugfix | Model] Fix SenseNova & Use Well-defined Model Configs

alex-jw-brooks:
## Purpose
While working on the TeaCache refactor, I noticed SenseNova was broken by the 0.26 rebase because it can't initialize its encapsulated LogitsProcessor for the AR component.

Example:
```python
from vllm_omni.entrypoints import Omni

if __name__ == "__main__":
    MODEL_NAME = "sensenova/SenseNova-U1-8B-MoT"
    model = Omni(MODEL_NAME, enforce_eager=True)
    outputs = model.generate("A white cat sitting on a windowsill")
    img = outputs[0].images[0]
    img.save("result.png")
```

ultimately breaks with:
```
  File "/home/alex-jw-brooks/vllm-omni/vllm_omni/diffusion/registry.py", line 379, in initialize_model
    model = model_class(od_config=od_config)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alex-jw-brooks/vllm-omni/vllm_omni/diffusion/models/sensenova_u1/pipeline_sensenova_u1.py", line 559, in __init__
    self.language_model = SenseNovaU1ForCausalLM(

                          ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alex-jw-brooks/vllm-omni/vllm_omni/diffusion/models/sensenova_u1/sensenova_u1_transformer.py", line 740, in __init__
    self.logits_processor = LogitsProcessor(config.vocab_size)
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alex-jw-brooks/venvs/omni/lib64/python3.12/site-packages/vllm/model_executor/layers/logits_processor.py", line 62, in __init__
    self.head_dtype = model_config.head_dtype if model_config is not None else None
```

To be able to initialize the logits processor, we need to define the attribute on the Diffusion config.

I also noticed while writing a quick test for this that the configs for SenseNova are `SimpleNamespace` objects which defaults set backed on `.get`s from a JSON loaded dict, so went ahead and added well-defined config classes in the transformers utils, and migrated to use the hf pretrained config subclass, which are then used in the test to create the small LLM to validate that we can initialize without issues.

## Test Plan
Added unit test to ensure the LLM can be initialized.

## Test Result
Test passes (i.e., LogitsProcessor is able to initialize without crashing) and the example runs without crashing.

result

### DIFF
--- a/tests/diffusion/models/sensenova_u1/test_sensenova_u1_logits_processor.py
+++ b/tests/diffusion/models/sensenova_u1/test_sensenova_u1_logits_processor.py
@@ -18,9 +18,7 @@ from vllm.distributed.parallel_state import (
 from vllm_omni.diffusion.models.sensenova_u1.sensenova_u1_transformer import (
     SenseNovaU1ForCausalLM,
 )
-from vllm_omni.diffusion.worker.diffusion_worker import (
-    _DiffusionVllmModelConfig,
-)
+from vllm_omni.diffusion.vllm_config import _DiffusionVllmModelConfig
 from vllm_omni.transformers_utils.configs.sensenova_u1 import (
     SenseNovaU1Config,
 )
@@ -55,6 +53,8 @@ def test_logits_processor_head_dtype_under_diffusion_shim():
     fake_diff_config = _DiffusionVllmModelConfig(
         model="sensenova-u1-test",
         dtype=torch.bfloat16,
+        max_model_len=8192,
+        original_max_model_len=8192,
     )
     vllm_config = VllmConfig(
         device_config=DeviceConfig(device="cpu"),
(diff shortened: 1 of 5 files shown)


vllm_omni/diffusion/worker/diffusion_worker.py
```diff
@@ -77,6 +77,8 @@ class _DiffusionVllmModelConfig:
     disable_cascade_attn: bool = False
     enable_return_routed_experts: bool = False
     is_moe: bool = False
+    # Needed for models that bundle things like LogitsProcessors, e.g., SenseNova
+    head_dtype: torch.dtype | None = None
```
NickCao: This class has been moved to `vllm_omni/diffusion/vllm_config.py`, please rebase.

NickCao (approved):
Otherwise LGTM

PR #5878, opened 2026-08-06
[Frontend] Loosen Validation for TTS Models with No Uploaded Speakers

alex-jw-brooks:
## Purpose
The OpenAI spec has `voice` as a required string parameter that should not be optional (i.e., see [here](https://github.com/openai/openai-python/blob/main/src/openai/resources/audio/speech.py#L55); voice is not optional, and the `Voice` type alias does not include None type.

```python
Voice: TypeAlias = Union[
    str, Literal["alloy", "ash", "ballad", "coral", "echo", "sage", "shimmer", "verse", "marin", "cedar"], VoiceID
]
```

This means that if you start a model that doesn't have any uploaded speakers, e.g., `OmniVoice`, you end up with the following point of awkwardness:
```python
from openai import OpenAI

client = OpenAI(base_url="http://localhost:9999/v1", api_key="unused")
response = client.audio.speech.create(
    model="k2-fsa/OmniVoice",
    input="Hello, world.",
)
```

raises on the client side:
```
TypeError: Speech.create() missing 1 required keyword-only argument: 'voice'
```

while passing any string for voice:
```python
from openai import OpenAI

client = OpenAI(base_url="http://localhost:9999/v1", api_key="unused")
response = client.audio.speech.create(
    model="k2-fsa/OmniVoice",
    input="Hello, world.",
    voice="vivian",
)
```

gives a 400 on the omni side
```
openai.BadRequestError: Error code: 400 - {'error': {'message': "Invalid voice 'vivian'. Supported: none", 'type': 'BadRequestError', 'param': None, 'code': 400}}
```
including if you pass `voice="none"`, which is how most users probably interpret this. I.e., in that case you'll get this very confusing response:
```
openai.BadRequestError: Error code: 400 - {'error': {'message': "Invalid voice 'none'. Supported: none", 'type':'BadRequestError', 'param': None, 'code': 400}}
```

It is worth noting that you technically _can_ pass `voice=None` currently, and it will pass it and succeed, but this is awkward, and we should not rely on it since `None` is not a valid type for that kwarg.

This PR cleans up voice validation a bit and allows `voice` to be passed as a string for models that do not have any uploaded speakers, so that the request will still work, but they'll just get a warning that voice has no effect. So we only raise if the model actually has supported speakers, and the one that was passed is wrong.

## Test Plan
Add a unit test to omni voice expansion tests to ensure we can succeed when a voice is passed.

## Test Result
unit test passes, and
```python
from openai import OpenAI

client = OpenAI(base_url="http://localhost:9999/v1", api_key="unused")
response = client.audio.speech.create(
    model="k2-fsa/OmniVoice",
    input="Hello, world.",
    voice="vivian",
)
```

runs without issues, with the server side log
```
(APIServer pid=1256471) WARNING 08-06 23:58:21 [serving_speech.py:3627] This model does not have any uploaded speakers, but one was provided; its value will be ignored.
```

CC @NickCao @clodaghwalsh17

### DIFF
--- a/vllm_omni/entrypoints/openai/serving_speech.py
+++ b/vllm_omni/entrypoints/openai/serving_speech.py
@@ -135,6 +135,12 @@ _QWEN3_TTS_REF_AUDIO_CACHE_KEY = "_qwen3_tts_ref_audio_cache_key"
 _TTS_MAX_INSTRUCTIONS_LENGTH = 500
 _TTS_MAX_NEW_TOKENS_MAX = 4096
 _MING_DEFAULT_PROMPT = MING_DEFAULT_PROMPT
+_DEFAULT_VOICE_NAME = "default"
+
+def _is_default_voice(voice, supported_speakers):
+    """Check if a lowercased voice name is the placeholder default and not
+    an actual registered/built-in speaker."""
+    return voice == _DEFAULT_VOICE_NAME and voice not in supported_speakers
 
 
 def _create_wav_header(sample_rate: int, num_channels: int = 1, bits_per_sample: int = 16) -> bytes:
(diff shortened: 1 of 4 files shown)


vllm_omni/entrypoints/openai/serving_speech.py
```diff
@@ -3608,6 +3608,28 @@ async def _generate_audio_bytes(
...
+        the voice is:
+            - lowercased if it's a valid supported speaker
+            - None if the the model doesn't have any uploaded speakers
+
+        If a voice is passed and we do have uploaded speakers, but it's still
+        not supported, we raise an error.
+        """
+        if voice:
+            voice = voice.lower()
+            if self.uploaded_speakers:
```
NickCao: In the openai API spec, custom voices are not passed as strings, but as VoiceID objects with an `id` field: https://developers.openai.com/api/reference/resources/audio/subresources/speech/methods/create, we may wish to add support for that too.

NickCao (approved):

PR #5801, opened 2026-08-05
[Perf] Minimax-H3 support fused RMSNorm and RoPE opt

fan2956:
PLEASE FILL IN THE PR DESCRIPTION HERE.

## Purpose
Enable MiniMax-H3 to use the fused RMSNorm and RoPE implementations provided by the Omni diffusion layers.
- Add configurable weight dtype support to Omni diffusion `RMSNorm`, while preserving fp32 accumulation in the native fallback.
- Extend Omni diffusion `RotaryEmbedding` to support full-head-dimension `cos`/`sin` inputs required by MiniMax-H3.

## Test Plan
```
MODEL="/data/weights/h3/MiniMax-H3/FL2VA"
vllm serve "${MODEL}" \
  --omni \
  --host 0.0.0.0 \
  --port "${PORT}" \
  --trust-remote-code \
  --num-gpus 8 \
  --usp 8 \
  --ring 1 \
  --text-encoder-tp-size 8 \
  --enable-layerwise-offload \
  --vae-parallel-mode tile \
  --vae-use-tiling \
  --vae-patch-parallel-size 8 \
  --diffusion-attention-backend FLASH_ATTN 
```
prompt："A little girl grows up."
image: 
first_frame

**vLLM Version:**
0.26.0

**vLLM-Omni Commit:**
5215e03a91adecbb5ffece29aa74360a7569d0c5

## Test Result
**Performance:**

E2E Latency: 491014ms -> 475070ms.      **-3.35%**

ori: 
Pasted Graphic 1

opt:
Pasted Graphic

**Precision:**
ori (left) vs opt (right)

https://github.com/user-attachments/assets/3cbd06eb-8e77-4106-a42a-eba346f71873

**BEFORE SUBMITTING:** read [CONTRIBUTING.md](https://github.com/vllm-project/vllm-omni/blob/main/CONTRIBUTING.md) and run the [precheck-pr skill](https://github.com/vllm-project/vllm-omni/blob/main/.claude/skills/precheck-pr/SKILL.md) with the code agent for a self-check against project conventions.
(anything written below this line will be removed by GitHub Actions)

PR #5881, opened 2026-08-07
[Perf][MUSA] Restore MiniMax-H3 dynamic RoPE fusion

yeahdongcn:
## Summary

- add a MUSA `RMSNorm.forward_musa` path that preserves `aten::rms_norm` for dynamic Inductor fusion;
- implement full-dimension NeoX RoPE in the shared `RotaryEmbedding.forward_musa` backend while retaining `forward_native` for other MUSA layouts;
- keep the MiniMax-H3 caller platform-agnostic and derive its static rotary width from the model config.

## Motivation

This is a MUSA follow-up to #5801. After #5801, MiniMax-H3 routes Q/K normalization and RoPE through the shared RMSNorm/RotaryEmbedding implementation. On MUSA with the production regional compile mode (`torch.compile(fullgraph=True, dynamic=True)`), the actual H3 `_apply_rope` path no longer forms the previous fused schedule.

`RotaryEmbedding.forward_native` is functionally correct, but its generic frequency normalization and expansion do not preserve the fast H3 schedule on MUSA. This PR keeps the model caller on the shared custom-op interface and puts the verified full-dimension NeoX implementation in `RotaryEmbedding.forward_musa`. Half-dimension and interleaved MUSA layouts continue to use `forward_native`; non-MUSA backends are unchanged.

## Performance

Measured from `main` at `a874b8e09bb5251bb16e3de8bc22edf3880eec09`, which contains #5801 merge commit `f2e3d2e25bebc0c7110ccb945d7dcf4be81bf6db`.

Environment and shape:

- 1x MTT S5000, MUSA driver `3.3.5-server`;
- image `registry.mthreads.com/mcconline/inference/vllm-omni@sha256:bdf15f76a86a736cb64bea41a46eeb05be4382f9efe5a520034e7b2f07ce04a5`;
- BF16 packed QKV, sequence length 96,000, 7 local heads, head dimension 128, rotary dimension 96;
- `torch.compile(fullgraph=True, dynamic=True)`;
- 5 warmups, 2 discarded pilot cycles, 20 interleaved cold-cache samples with a 512 MiB cache flush;
- MUSA event timing cross-checked with synchronized wall time.

| exact H3 Q/K norm + RoPE path | MUSA event median | synchronized wall median |
| --- | ---: | ---: |
| current main | 30.0062 ms | 30.0695 ms |
| this PR through `RotaryEmbedding.forward_musa` | 1.6323 ms | 1.6893 ms |
| change | **-94.56% (18.38x)** | **-94.38% (17.80x)** |

In the final interface A/B, the platform-dispatched path was within noise of the direct inline reference (`1.6323` vs `1.6316` ms by MUSA events). A forced shared-native control measured `3.3009` ms, confirming that calling `forward_native` directly does not retain the MUSA fast path. Output is finite, shape-identical, and passes the BF16 correctness gate.

These numbers cover the exact Q/K RMSNorm + H3 RoPE compiled region; they are not presented as TP8 end-to-end request latency.

## Tests

```text
ruff check vllm_omni/diffusion/layers/norm.py \
  vllm_omni/diffusion/layers/rope.py \
  vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py \
  tests/diffusion/layers/test_norm.py \
  tests/diffusion/layers/test_rope_broadcast.py \
  tests/diffusion/models/minimax_h3/test_minimax_h3_parallel.py
# All checks passed

pytest -q -p no:cacheprovider \
  tests/diffusion/layers/test_norm.py \
  tests/diffusion/layers/test_rope_broadcast.py \
  tests/diffusion/models/minimax_h3/test_minimax_h3_parallel.py
# The 64 remaining cases were all included in the passing 69-case superset run.
```

### DIFF
--- a/vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
+++ b/vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
@@ -333,6 +333,7 @@ class MiniMaxH3Attention(nn.Module):
         )
         self.num_heads = self.qkv_proj.num_heads
         self.num_kv_heads = self.qkv_proj.num_kv_heads
+        self.rot_dim = 6 * arch.rope_inv_freq_len
         self.q_norm = _norm(arch.attention_head_dim, eps=arch.qk_norm_eps)
         self.k_norm = _norm(arch.attention_head_dim, eps=arch.qk_norm_eps)
         self.rope = RotaryEmbedding(is_neox_style=True, half_head_dim=False)
@@ -365,7 +366,7 @@ class MiniMaxH3Attention(nn.Module):
         x: [T, heads, head_dim]; freqs: [T, rot_dim]. In the unfused path, cos/sin
         are cast to the activation dtype before the elementwise math.
         """
-        rot_dim = freqs.shape[-1]
+        rot_dim = self.rot_dim
         x_rot, x_pass = x[..., :rot_dim], x[..., rot_dim:]
         cos = torch.cos(freqs).to(x.dtype)  # [T, rot_dim]
         sin = torch.sin(freqs).to(x.dtype)
--- a/tests/diffusion/models/minimax_h3/test_minimax_h3_parallel.py
+++ b/tests/diffusion/models/minimax_h3/test_minimax_h3_parallel.py
@@ -69,6 +69,7 @@ def test_h3_fused_rope_matches_reference_and_preserves_unrotated_dims():
 
     attention = object.__new__(MiniMaxH3Attention)
     nn.Module.__init__(attention)
+    attention.rot_dim = 96
     attention.rope = RotaryEmbedding(is_neox_style=True, half_head_dim=False)
     attention.rope._forward_method = attention.rope.forward_native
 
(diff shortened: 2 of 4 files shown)


yeahdongcn:
@fan2956 Please also take a look. Thanks!

vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
```diff
@@ -365,13 +367,25 @@ def _apply_rope(self, x: torch.Tensor, freqs: torch.Tensor) -> torch.Tensor:
...
+        if current_omni_platform.is_musa():
+            return self._apply_rope_musa(x, freqs)
         rot_dim = freqs.shape[-1]
         x_rot, x_pass = x[..., :rot_dim], x[..., rot_dim:]
         cos = torch.cos(freqs).to(x.dtype)  # [T, rot_dim]
         sin = torch.sin(freqs).to(x.dtype)
         x_rot = self.rope(x_rot, cos, sin)
         return torch.cat((x_rot, x_pass), dim=-1)
 
+    def _apply_rope_musa(self, x: torch.Tensor, freqs: torch.Tensor) -> torch.Tensor:
```
hsliuustc0106: @gcanlin do we have unified interface for important ops like rope? each hardware can have their own implementation

yeahdongcn: Alternatively, we could add an `_apply_rope_native` implementation as a fallback. Or I can rename `_apply_rope_musa` to it.

hsliuustc0106: check vllm_omni/diffusion/layers/rope.py

yeahdongcn: Moved the `MUSA`-specific handling into this file.

PR #5882, opened 2026-08-07
[Model] Add Cache-DiT and CPU offload support for SANA-Video 2B

cr-gao:
> **Update 2026-08-25**: #5508 merged; this branch is rebased onto current
> `main` (16 commits, conflicts resolved — the Cache-DiT backend refactor and
> the multi-DiT enable loop that landed on `main` are used as-is). Two changes
> came out of the post-rebase validation on 4×A800 with `cache-dit==1.5.0`
> (the version `main` now pins):
>
> - tp1 output is **bitwise identical** to merged `main`'s single-GPU output
>   (1-step, real weights), and the tiny cache/offload L2 matrix passes
>   (6 cases).
> - One new commit fixes a latent shared-infra issue this PR's tests are the
>   first to expose: `cache_dit` 1.5.0 marks the adapter wrapper pipe's
>   *class* as cached, so the second single-GPU Cache-DiT engine in one
>   process (exactly what the serial CI worker runs) skipped cache-context
>   installation and crashed. The backend now clears the stale class marker
>   before enabling. Existing declarative-adapter models (e.g. Flux2Klein)
>   are unaffected — their cache tests run in spawned multi-GPU workers,
>   which is why this never surfaced before.

### What this adds

Cache-DiT and CPU offload for the native SANA-Video 2B pipelines (T2V and I2V):

- Cache-DiT via `--cache-backend cache_dit`, declared with an explicit
  `ForwardPattern.Pattern_3` adapter config on `SanaVideoTransformer3DModel`.
- Model-level CPU offload (`--enable-cpu-offload`), layerwise CPU offload
  (`--enable-layerwise-offload`), and distributed layerwise offload
  (`--enable-distributed-layerwise-offload`).
- Cache-DiT combined with either model-level or layerwise CPU offload.
- A `default_num_inference_steps` pipeline contract (50 for SANA) so cache
  refresh works when a request omits `num_inference_steps`.

Component loading now honours the loader's default-device context instead of
forcing the runtime device, which is what lets offload keep weights on the host
without an extra CPU → GPU → CPU round trip.

Unsupported combinations fail before any checkpoint component is loaded:
Cache-DiT or offload with TP/CFG/SP > 1, Cache-DiT with distributed layerwise
offload (per-rank cache skips would desynchronize the weight AllGather), and
cache backends other than `cache_dit`.

### Test plan

- 300 CPU/mock tests pass (`tests/diffusion/models/sana_video/`,
  `tests/diffusion/cache/`, `tests/diffusion/test_diffusion_model_runner.py`,
  `tests/model_tests/diffusion/test_alignment.py`).
- New `tests/diffusion/models/sana_video/test_cache_offload.py` covers cache
  metadata and pattern check, cache refresh across requests, component load
  device, offloader hook lifecycle (including cleanup on exception), and the
  rejection paths above.
- The tiny common-harness model gains five acceleration groups: Cache-DiT,
  model offload, layerwise offload, and the two Cache-DiT + offload combos.
- GPU runs below reproduce through the public offline example (the local
  parity/benchmark driver wraps the same `Omni` call and stays out of the PR):

```bash
cd examples/offline_inference/text_to_video
python text_to_video.py \
    --model Efficient-Large-Model/SANA-Video_2B_480p_diffusers \
    --prompt "A cat walking on the grass, facing the camera. motion score: 30." \
    --height 480 --width 832 --num-frames 81 \
    --num-inference-steps 50 --guidance-scale 6.0 --seed 42 \
    --cache-backend cache_dit          # or: --enable-cpu-offload / --enable-layerwise-offload
```

### GPU evidence

One A800-SXM4-80GB, driver 580.126.09, CUDA 13.0, PyTorch 2.11.0+cu130,
Diffusers 0.38.0, vLLM 0.26.0. Weights verified file-by-file (LFS sha256)
against the revisions pinned by #5508: 480p `fed3bce4…`, 720p `8bda5e62…`.
Attention backend pinned to `CUDNN_ATTN` (plus a `TORCH_SDPA` cross-check)
because this box is Ampere and the platform default FLASH backend has a known
cross-attention key-padding bug (#5865) that would pollute the comparison.
19 configurations, each in its own process, 81 frames, 50 steps, seed 42.

**Functional matrix** — Cache-DiT, model offload, layerwise offload and both
combinations all produce video on 480p/720p × T2V/I2V; 19/19 runs succeeded.
Single-rank distributed layerwise offload runs; `cache_dit` + distributed
layerwise offload raises before load, as designed.

**Peak GPU memory** (480p T2V, generation phase):

| Configuration | startup peak | generation peak | vs baseline |
|---|---:|---:|---:|
| Baseline | 14.28 GiB | 24.07 GiB | — |
| Model CPU offload | 6.14 GiB | 16.81 GiB | −7.26 GiB |
| Layerwise offload | 6.80 GiB | 20.75 GiB | −3.32 GiB |

**Cache-DiT speedup** (median of 3 measured runs after 1 warmup):
480p T2V 120.19 s → 77.11 s (**1.56×**); 720p I2V 37.0 s → 24.1 s (**1.54×**).
Offload latency is report-only: 121.15 s (model-level) and 118.63 s (layerwise)
against a 120.19 s baseline.

**Note for reviewers on similarity metrics.** This pipeline is not run-to-run
deterministic on the test hardware (identical code and seed differ by SSIM
0.9712 at 50 steps, unchanged by the PyTorch/cuBLAS determinism switches), so
SSIM against a reference measures trajectory distance rather than quality and
is not used as a gate. At 1 step the uncached and offload paths match a run of
the #5508 HEAD at the measured noise floor (rel_l2 ≈ 1e-2); Cache-DiT output
was verified frame by frame against the uncached run instead — see below.

### Frame comparison

Frames 0/26/53/80, 50 steps, seed 42.

**480p T2V — no Cache-DiT**

480p T2V baseline

**480p T2V — Cache-DiT (1.56×)**

480p T2V Cache-DiT

**720p I2V — no Cache-DiT**

720p I2V baseline

**720p I2V — Cache-DiT (1.54×)**

720p I2V Cache-DiT

The frame-26 transition in the 720p I2V cell looks the same with and without
Cache-DiT; it comes from the baseline.

### Not covered

- Multi-rank distributed layerwise offload (single-GPU box; only the gate and
  the single-rank path were exercised).
- Any parallel × cache/offload combination — rejected by design in this PR.
- Non-A800 hardware; all numbers above are single-GPU A800.
- TeaCache and step execution for the native pipelines.


vllm_omni/diffusion/worker/diffusion_model_runner.py
```diff
@@ -425,6 +422,8 @@ def _refresh_cache_for_requests(
         # RequestBatchSamplingParamsKey, so the first request's num_inference_steps applies
         # to the whole runner batch.
         num_inference_steps = first_req.sampling_params.num_inference_steps
+        if num_inference_steps is None:
+            num_inference_steps = getattr(self.pipeline, "default_num_inference_steps", None)
```
congw729, line 451: Cache-DiT uses the default 50 (need to be confirmed) steps when num_inference_steps is omitted, ignoring custom timesteps or sigmas. This can misconfigure the cache and produce incorrect results.

congw729:
The fallback itself is necessary — without the refresh, cache state leaks across requests. That said, it would be better to refresh with the actual executed step count: prefer the effective steps from a custom schedule (len(timesteps) or len(sigmas)), then num_inference_steps, and only fall back to default_num_inference_steps last.

As for a custom schedule overriding an explicitly passed num_inference_steps — that's a pre-existing issue in the shared cache-refresh logic on main, not something introduced here. I expect it'll be handled uniformly later, so no need to address it in this PR.

cr-gao:
> The fallback itself is necessary — without the refresh, cache state leaks across requests. That said, it would be better to refresh with the actual executed step count: prefer the effective steps from a custom schedule (len(timesteps) or len(sigmas)), then num_inference_steps, and only fall back to default_num_inference_steps last.
> 
> As for a custom schedule overriding an explicitly passed num_inference_steps — that's a pre-existing issue in the shared cache-refresh logic on main, not something introduced here. I expect it'll be handled uniformly later, so no need to address it in this PR.

Revised and updated relevant tests.

cr-gao:
@congw729 @hsliuustc0106 Rebased onto current `main` after the #5508 merge (16 commits, conflicts resolved, now on `cache-dit==1.5.0`) and marked ready. The earlier review feedback on the cache-refresh step resolution is addressed (`Prefer custom schedule length for cache refresh steps`), and post-rebase validation surfaced one latent shared-infra issue that a new small commit fixes (stale Cache-DiT pipe-class marker between engines in one process — details in the updated description). This PR is independent of #5861/#5940. A timely pass would be much appreciated — SANA 2.0 work is queued behind this series. Thanks!

cr-gao:
Self-review — what I checked:

- **No-regression baseline**: with cache and offload off, tp1 output is bitwise-identical to merged `main` (1-step, real weights, post-rebase).
- **Supported combinations**: Cache-DiT alone, model-level offload, layerwise offload, and Cache-DiT × either offload all pass the tiny L2 matrix (single GPU, real engine chain); unsupported combinations (Cache-DiT × distributed layerwise offload, any cache/offload with TP/CFG/SP > 1, non-cache_dit backends) fail closed before checkpoint components load.
- **Cache refresh**: per-request refresh resolves steps from explicit `num_inference_steps`, then `timesteps`/`sigmas` length, then the pipeline's 50-step default — the fallback chain congw729 asked for, covered by runner unit tests.
- **cache-dit 1.5.0 migration**: re-validated everything on the version `main` now pins; this surfaced the stale pipe-class marker issue fixed in the last commit, which has a red→green regression test (removing the fix fails `test_second_engine_enables_cache_dit_after_undisabled_first`) and leaves existing declarative-adapter models unaffected (Flux2Klein control re-run).
- **Regression**: 311 CPU tests (SANA + cache suites) and the tiny L2 matrix pass; measurements in the description (offload −7.26/−3.32 GiB, Cache-DiT 1.56×/1.54×) are from real weights on A800.

congw729 (approved):
Look good to me.

hsliuustc0106:
short question: do we need to offload for this 2B model? what's the use case?

cr-gao:
Good question. The peak is not driven by the 2B DiT weights: a 480p T2V generation peaks at 24.07 GiB with everything resident (Gemma2 text encoder, VAE, activations), which does not fit a 24 GB consumer card. Model CPU offload brings that to 16.81 GiB at essentially no latency cost (121.15 s vs 120.19 s baseline, numbers in the description). So the use case is running 480p generation on 24 GB and smaller consumer GPUs.

cr-gao:
@hsliuustc0106 @congw729 CI is green and all review comments are resolved. Ready to merge from my side, after #5861 since this branch stacks on it.

hsliuustc0106 (approved):
lgtm

cr-gao:
Resolved the conflict with main after #5861 merged (union of the test groups and recipe notes). One real interaction surfaced: the transformer from #5861 needs an initialized parallel group at construction, so `test_cache_offload.py` now reuses the `_init_distributed` fixture from the transformer tests. Full local SANA suite on the merged tree: 180 passed on vllm 0.28.

Issue #5879, opened 2026-08-07
[Bug]: Nightly CI,  tests/e2e/online_serving/test_bagel_expansion.py::test_bagel[single_card_cache_dit], ValueError: Pipeline BagelPipeline has no declared DiT modules for Cache-DiT


yenuo26:
### Your current environment

<details>
<summary>The output of <code>python collect_env.py</code></summary>

```text
CI env
```

</details>

### Your code version

<details>
<summary>The commit id or version of vllm</summary>

```text
0.26.0
```
</details>
<details>
<summary>The commit id or version of vllm-omni</summary>

```text
a874b8e09bb5251bb16e3de8bc22edf3880eec09
```
</details>

### 🐛 Describe the bug

https://buildkite.com/vllm/vllm-omni/builds/13258/canvas?sid=019fd83f-98f7-4be2-95cc-f97f26a7d6d2&tab=output
```
Process DiffusionWorker-0:
Traceback (most recent call last):
  File "/usr/lib/python3.12/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/usr/lib/python3.12/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/workspace/build/buildkite/vllm_omni/diffusion/worker/diffusion_worker.py", line 1177, in worker_main
    worker_proc = WorkerProc(
                  ^^^^^^^^^^^
  File "/workspace/build/buildkite/vllm_omni/diffusion/worker/diffusion_worker.py", line 824, in __init__
    self.worker = self._create_worker(gpu_id, od_config, worker_extension_cls, custom_pipeline_args)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/build/buildkite/vllm_omni/diffusion/worker/diffusion_worker.py", line 852, in _create_worker
    wrapper = WorkerWrapperBase(
              ^^^^^^^^^^^^^^^^^^
  File "/workspace/build/buildkite/vllm_omni/diffusion/worker/diffusion_worker.py", line 1249, in __init__
    self.worker = worker_class(
                  ^^^^^^^^^^^^^
  File "/workspace/build/buildkite/vllm_omni/diffusion/worker/diffusion_worker.py", line 256, in __init__
[...]

PR #5884, opened 2026-08-07
[bugfix][CI] Fix Cache-DiT nested module discovery [issue 5879]

natureofnature:

## Purpose
Fix issue for issue https://github.com/vllm-project/vllm-omni/issues/5879

### Root Cause

Bagel declares its DiT module using the nested path language_model.model. Cache-DiT used getattr(), which only resolves direct attributes, so the nested module was ignored and the pipeline was reported as having no declared DiT modules.

### Solution
Use operator.attrgetter consistently to resolve both direct and nested DiT module paths during discovery, enable/refresh, and summary. Cache summaries are also limited to modules with an active Cache-DiT context, avoiding summaries for uncached nested modules such as Cosmos3’s UND language model.

 ## Test Plan

  - Reproduce the original failure on unmodified upstream main.
  - Run the Cache-DiT unit test suite.
  - Run adjacent Bagel and module-discovery tests.
  - Run the exact Bagel Cache-DiT full-model E2E case.
  - Run pre-commit checks on the modified files.

  vLLM Version: 0.26.0

  vLLM-Omni Commit: e8b1fd6e8ff613f6b5eaf94048a5d663614aed48
  Based on upstream main at 6ec90330655b929175ecd7d3b8ea26bcaf87a0a2.

  ## Test Result

  - Cache-DiT unit tests: 30 passed.
  - Adjacent Bagel and module-discovery tests: 31 passed.
  - Bagel full-model E2E: 1 passed in 212.18s, including valid 512x512 response validation.
  - Pre-commit and commit hooks: all passed.

@yenuo26 

### DIFF
--- a/vllm_omni/diffusion/cache/cachedit/backend.py
+++ b/vllm_omni/diffusion/cache/cachedit/backend.py
@@ -6,6 +6,7 @@
 from __future__ import annotations
 
 from collections.abc import Callable
+from operator import attrgetter
 from typing import TYPE_CHECKING, Any, TypeAlias, cast
 
 import cache_dit
@@ -38,14 +39,27 @@ def _dit_module_names(pipeline: SupportsComponentDiscovery) -> tuple[str, ...]:
     names = getattr(pipeline, "_dit_modules", None)
     if not isinstance(names, (list, tuple)):
         names = ("transformer",)
-    return tuple(name for name in names if isinstance(name, str) and getattr(pipeline, name, None) is not None)
+
+    resolved_names = []
+    for name in names:
+        if not isinstance(name, str):
+            continue
+        try:
+            module = attrgetter(name)(pipeline)
+        except AttributeError:
+            continue
+        if module is not None:
+            resolved_names.append(name)
+    return tuple(resolved_names)
 
 
 def cache_summary(pipeline: SupportsComponentDiscovery, details: bool = True) -> None:
     """Log Cache-DiT statistics for every transformer on the pipeline."""
 
-    transformers = [getattr(pipeline, name) for name in _dit_module_names(pipeline)]
+    transformers = [attrgetter(name)(pipeline) for name in _dit_module_names(pipeline)]
     for transformer in transformers:
+        if not BlockAdapter.is_cached(transformer):
+            continue
         cache_dit.summary(transformer, details=details)
 
     if not transformers:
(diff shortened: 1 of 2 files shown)


natureofnature:
@codex review

natureofnature:
@codex review

natureofnature:
@Gaohan123  PTAL

vllm_omni/diffusion/cache/cachedit/backend.py
fhfuih, line 1: If I understand it correctly, the main edit is the `_dit_module_names()`'s return clause. The rest `attrgetter` parts seems unnecessary. `attrgetter(attr_name)(obj)` is the same as `getattr(obj, attr_name)` unless you define one getter with one fixed name and use it multiple times. (But in your case, name is dynamic and for-looped)

Correct me if I am wrong, or I suggest removing the usage of `attrgetter` 🤔

david6666666: For flat names they're equivalent, but this PR is really about the dotted case: `getattr(pipeline, "language_model.model")` raises AttributeError and is treated as missing, while `attrgetter` walks the chain — that's why Bagel's nested DiT was silently dropped. The getter factory and `cache_summary` keep the same resolution so enable/summary stay consistent for multi-level names.

vllm_omni/diffusion/cache/cachedit/backend.py
```diff
@@ -38,14 +39,27 @@ def _dit_module_names(pipeline: SupportsComponentDiscovery) -> tuple[str, ...]:
...
     if not isinstance(names, (list, tuple)):
         names = ("transformer",)
-    return tuple(name for name in names if isinstance(name, str) and getattr(pipeline, name, None) is not None)
+
+    resolved_names = []
+    for name in names:
+        if not isinstance(name, str):
+            continue
+        try:
+            module = attrgetter(name)(pipeline)
```
david6666666, line 48: Nice — this mirrors `vllm_omni/diffusion/offloader/module_collector.py`, which already resolves dotted `_dit_modules` paths via `attrgetter` (same try/except AttributeError, skip missing). If you ever want to dedupe, a tiny shared "resolve module path or None" helper would keep cache-dit and the offloader from drifting.

fhfuih (approved):

vllm_omni/diffusion/cache/cachedit/backend.py
fhfuih: Ohhh 

ohh

david6666666 (approved):

PR #4613, opened 2026-06-22
[CI] Lower Qwen-Image e2e SSIM threshold from 0.97 to 0.94

NumberWan:
PLEASE FILL IN THE PR DESCRIPTION HERE.

## Purpose
Related to [#4029](https://github.com/vllm-project/vllm-omni/issues/4029).

The nightly CI accuracy test test_qwen_image_matches_diffusers has been failing intermittently with SSIM scores slightly below the current threshold (e.g. 0.958 vs >= 0.9700). The root cause of this accuracy fluctuation is not yet understood. 

As a temporary workaround, lower the Qwen-Image e2e SSIM threshold from 0.97 to 0.94 in tests/e2e/accuracy/test_qwen_image.py to reduce false CI failures. This does not address the underlying issue; the threshold should be restored to 0.97 once the root cause is identified and fixed.

Related CI result: https://buildkite.com/vllm/vllm-omni/builds/11630/canvas?sid=019eeb61-d3d5-4a47-9ef1-366bd9f1909e&tab=output
## Test Plan

**vLLM Version:**

**vLLM-Omni Commit:**

## Test Result

**BEFORE SUBMITTING:** read [CONTRIBUTING.md](https://github.com/vllm-project/vllm-omni/blob/main/CONTRIBUTING.md) and run the [precheck-pr skill](https://github.com/vllm-project/vllm-omni/blob/main/.claude/skills/precheck-pr/SKILL.md) with the code agent for a self-check against project conventions.
(anything written below this line will be removed by GitHub Actions)

PR #5887, opened 2026-08-07
[BugFix][Nightly CI] Opt in FA deterministic for Qwen-Image accuracy

### LINKED ISSUE #5734: [Bug]: Nightly / CI failed Occasional- tests/e2e/accuracy/test_qwen_image.py::test_qwen_image_matches_diffusers, AssertionError: PSNR below threshold for Qwen/Qwen-Image: got 27.800467, expected >= 30.000000
### Your current environment

<details>

```text
ci env
```

</details>

### Your code version

<details>

```text
(not found in Buildkite step log)
```
</details>
<details>

```text
900a7f0813d0
```
</details>

### 🐛 Describe the bug

**Buildkite build:** https://buildkite.com/vllm/vllm-omni/builds/13108
**Buildkite step:** [:full_moon: Diffusion X2I(&A&T) · Accuracy Test](https://buildkite.com/vllm/vllm-omni/builds/13108#019fc8cb-55ed-478e-bb9f-8ad4729f1fa0)

**Failure kind:** pytest FAILED
**Test node:** `tests/e2e/accuracy/test_qwen_image.py::test_qwen_image_matches_diffusers`

**Log reason:**
AssertionError: PSNR below threshold for Qwen/Qwen-Image: got 27.800467, expected >= 30.000000.

**Analysis:**
Category: assertion failure. Expected vs actual mismatch — check test logic, edge cases, and data setup.

**Error log excerpt:**

```text
accuracy_artifact_root = PosixPath('/workspace/build/buildkite/tests/e2e/accuracy/artifacts')



    @pytest.mark.benchmark

    @hardware_test(res={"cuda": "H100"}, num_cards=1)

    def test_qwen_image_matches_diffusers(accuracy_artifact_root: Path) -> None:

        model = _model_name()

        output_dir = model_output_dir(accuracy_artifact_root, MODEL_ID)

    

        vllm_output = _run_vllm_omni_qwen_image(model=model, output_path=output_dir / "vllm_omni.png")

        diffusers_output = _run_diffusers_qwen_image(model=model, output_path=output_dir / "diffusers.png")

    

>       assert_similarity(

            model_name=MODEL_ID,

            vllm_image=vllm_output,

            diffusers_image=diffusers_output,

            width=WIDTH,

            height=HEIGHT,

            ssim_threshold=SSIM_THRESHOLD,

            psnr_threshold=PSNR_THRESHOLD,

        )



tests/e2e/accuracy/test_qwen_image.py:204: 

_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 



model_name = 'Qwen/Qwen-Image', vllm_image = <PIL.Image.Image image mode=RGB size=512x512 at 0x73434F3C5F10>

diffusers_image = <PIL.Image.Image image mode=RGB size=512x512 at 0x7341DC70EAB0>, ssim_threshold = 0.94, psnr_threshold = 30.0, width = 512, height = 512

compare_mode = 'RGB'



    def assert_similarity(

        *,

        model_name: str,

        vllm_image: Image.Image,

        diffusers_image: Image.Image,

        ssim_threshold: float,

        psnr_threshold: float,

        width: int | None = None,

        height: int | None = None,

        compare_mode: str = "RGB",

    ) -> None:

        requested_size = (width, height) if width is not None and height is not None else None

        if requested_size is not None and diffusers_image.size != requested_size:

            pytest.skip(

                "Skipping as diffusers baseline output is corrupt and not comparable: "

                f"dimensions do not match requested size; requested={requested_size}, got={diffusers_image.size}."

            )

    

        assert vllm_image.size == diffusers_image.size, (

            f"Online and diffusers output sizes mismatch: online={vllm_image.size}, diffusers={diffusers_image.size}"

        )

    

        ssim_score, psnr_score = compute_image_ssim_psnr(

            prediction=vllm_image,

            reference=diffusers_image,

            compare_mode=compare_mode,

        )

        print(f"{model_name} similarity metrics:")

        print(f"  SSIM: value={ssim_score:.6f}, threshold>={ssim_threshold:.6f}, range=[-1, 1], higher_is_better=True")

        print(

            f"  PSNR: value={psnr_score:.6f} dB, threshold>={psnr_threshold:.6f} dB, range=[0, +inf), higher_is_better=True"

        )

    

        assert ssim_score >= ssim_threshold, (

            f"SSIM below threshold for {model_name}: got {ssim_score:.6f}, expected >= {ssim_threshold:.6f

...(truncated for GitHub URL length; see HTML report for full log)
```

---
*Generated from a nightly HTML report. Redact secrets before submitting; complete the checkboxes on GitHub.*

NumberWan:
FlashAttention defaults to deterministic=False, which allows non-unique results under regional compile (Omni↔Omni bistable first-denoise outputs in #5734). Add OmniDiffusionConfig.fa_deterministic and CLI --fa-deterministic; honor the flag on the local FLASH_ATTN dense path. The Qwen-Image accuracy test enables the CLI via OmniServer and restores SSIM_THRESHOLD to 0.97; serving defaults stay unchanged.

## Purpose
Mitigates intermittent failures in the Qwen-Image accuracy test tracked by [#5734](https://github.com/vllm-project/vllm-omni/issues/5734).

Serving defaults are **unchanged**: regional `torch.compile` + platform-default local `FLASH_ATTN` still use FlashAttention’s library default `deterministic=False` (faster; does not guarantee unique/identical outputs across runs—`deterministic=True` is the slower tradeoff). This PR adds an **opt-in CLI** `--fa-deterministic` so accuracy CI can request `deterministic=True` on the dense local FA call (no env var).

Also restores `SSIM_THRESHOLD` from 0.94 → **0.97**, reversing the temporary workaround in [#4613](https://github.com/vllm-project/vllm-omni/pull/4613). That lower gate was needed because bistable runs could land on a low mode (~SSIM 0.958 on [#13108](https://buildkite.com/vllm/vllm-omni/builds/13108)); with CI pinning the stable FA mode, the original 0.97 gate is appropriate again.

## Investigation

Omni↔Omni divergence with identical inputs is **bistable** (two fixed modes). Pre-loop tensors stay bit-identical; first diverge at **`pos_noise_pred`** after the first denoise.

| Config | Stable? |
|--------|---------|
| `enforce_eager` + FA | Yes (≥15) |
| Regional compile + FA (in Inductor) | No — `pos_noise_pred` |
| Regional compile + `TORCH_SDPA` | Yes (≥15) |
| Regional compile + FA outside Inductor | Yes (≥15) |
| Regional compile + FA in Inductor + `deterministic=True` | Yes (≥15) |

Root cause: FA’s default nondeterministic dense kernel (`deterministic=False`). FA can stay in the compiled graph once `deterministic=True`.

Nightly [#13108](https://buildkite.com/vllm/vllm-omni/builds/13108) (same commit): fail SSIM **0.958** / PSNR **27.80**, manual retry SSIM **0.986** / PSNR **33.67** — classic bistable jump, not a hard regression. Omni↔Diffusers systematic bias on some boxes (~PSNR 28) is separate from this flake.

## Changes

1. `OmniDiffusionConfig.fa_deterministic` (default `False`) + CLI `--fa-deterministic`
2. Local `FLASH_ATTN` dense path passes `deterministic=True` when the flag is set (hub FA untouched)
3. `warning_once` if the flag is set but the call takes a non-dense (piecewise/varlen/masked) path
4. `test_qwen_image_matches_diffusers`: `OmniServer` `server_args` includes `--fa-deterministic`; restore `SSIM_THRESHOLD=0.97`

## Test Result

- [x] Config / dense FA path: default → omit `deterministic`; `fa_deterministic=True` → pass `deterministic=True`
- [x] `test_qwen_image_matches_diffusers` (`--run-level full_model`) on L20X with deterministic opt-in: SSIM **0.960** / PSNR **28.188** (L20X systematic Omni↔Diffusers bias; not the H100 bistable jump). H100 nightly is the real gate for PSNR≥30 and SSIM≥0.97.

## Supplement - Options considered

**A. Test-only `--enforce-eager` / `TORCH_SDPA`**  
Zero shared code, but CI leaves default FA+compile uncovered; eager misses Inductor×FA regressions; SDPA also mismatches Diffusers’ FA reference. Rejected.

**B. Qwen-only `@torch.compiler.disable` on joint attention**  
Keeps FA, no shared FA API change, CI≈new Qwen default—but changes Qwen serving compile semantics (product/perf). CI-only boundary rejected (greens a path users don’t run).

**C. Shared FA `deterministic` opt-in via CLI (this PR)**  
Keeps FA+compile in CI; serving default unchanged; no env var; matches the hunt that fixed bistability. Cost: CI ≠ stock serving bit-for-bit (`True` vs `False`); small shared FA / CLI surface.

**Why C:** we want CI on FA + regional compile, stop Omni↔Omni bistability, and not change serving defaults. A drops coverage; B changes Qwen production behavior; C is the smallest shared change that uses the proven knob and stays discoverable via `--help`.

Scope: CI runs FA + regional compile + `deterministic=True`—default architecture plus an accuracy-only kernel flag, not a claim that stock `deterministic=False` serving is bit-stable.

### DIFF
--- a/vllm_omni/diffusion/attention/backends/flash_attn.py
+++ b/vllm_omni/diffusion/attention/backends/flash_attn.py
@@ -15,6 +15,7 @@ from vllm_omni.diffusion.attention.backends.sdpa import _maybe_reshape_attn_mask
 from vllm_omni.diffusion.attention.backends.utils.piecewise_attn import (
     piecewise_attn,
 )
+from vllm_omni.diffusion.config import get_current_diffusion_config_or_none
 
 logger = init_logger(__name__)
 
@@ -71,8 +72,19 @@ class FlashAttentionImpl(AttentionImpl):
         self.causal = causal
         self.softmax_scale = softmax_scale
         self.qkv_layout = qkv_layout
+        cfg = get_current_diffusion_config_or_none()
+        self.fa_deterministic = bool(getattr(cfg, "fa_deterministic", False)) if cfg is not None else False
         if backend_kwargs:
             logger.warning("FlashAttentionImpl ignoring backend_kwargs: %s", list(backend_kwargs.keys()))
+
+    def _warn_fa_deterministic_non_dense(self, path: str) -> None:
+        if not self.fa_deterministic:
+            return
+        logger.warning_once(
+            "fa_deterministic=True is ignored on the %s FlashAttention path; "
+            "only the dense flash_attn_func path passes deterministic=True.",
+            path,
+        )
 
     @staticmethod
     def _unwrap_flash_output(out: torch.Tensor | tuple[torch.Tensor, ...]) -> torch.Tensor:
@@ -250,6 +262,7 @@ class FlashAttentionImpl(AttentionImpl):
 
         # Try piecewise attention
         if full_attn_spans is not None:
+            self._warn_fa_deterministic_non_dense("piecewise")
             logger.debug("Using piecewise Flash Attention for mixed causal/full mask")
             if flash_attn_func is not None:
                 attn_func = partial(
@@ -280,6 +293,7 @@ class FlashAttentionImpl(AttentionImpl):
             if len(present_packed_keys) != len(packed_keys):
                 missing = sorted(set(packed_keys) - set(present_packed_keys))
                 raise ValueError(f"Incomplete packed FlashAttention metadata; missing {missing}")
+            self._warn_fa_deterministic_non_dense("packed-varlen")
             return self._forward_varlen_packed(
                 query,
                 key,
@@ -291,6 +305,7 @@ class FlashAttentionImpl(AttentionImpl):
             )
 
         if attention_mask is not None and torch.any(~attention_mask):
+            self._warn_fa_deterministic_non_dense("masked-varlen")
             return self._forward_varlen_masked(
                 query,
                 key,
@@ -299,15 +314,16 @@ class FlashAttentionImpl(AttentionImpl):
             )
 
         if flash_attn_func is not None:
-            out = flash_attn_func(
-                query,
-                key,
-                value,
-                causal=self.causal,
-                softmax_scale=self.softmax_scale,
-            )
+            fa_kwargs = {
+                "causal": self.causal,
+                "softmax_scale": self.softmax_scale,
+            }
+            if self.fa_deterministic:
+                fa_kwargs["deterministic"] = True
+            out = flash_attn_func(query, key, value, **fa_kwargs)
             return self._unwrap_flash_output(out)
 
+        self._warn_fa_deterministic_non_dense("dense-varlen-fallback")
         return self._forward_varlen_dense(
             query,
             key,
--- a/vllm_omni/diffusion/data.py
+++ b/vllm_omni/diffusion/data.py
@@ -613,6 +613,9 @@ class OmniDiffusionConfig:
 
     # Attention
     diffusion_attention_config: "AttentionConfig" = field(default_factory=lambda: AttentionConfig())
+    # Opt-in FlashAttention deterministic kernel (slower). Default False keeps
+    # serving on the fast library path; enable via --fa-deterministic.
+    fa_deterministic: bool = False
 
     # Running mode
     # mode: ExecutionMode = ExecutionMode.INFERENCE
--- a/vllm_omni/diffusion/data.py
+++ b/vllm_omni/diffusion/data.py
@@ -613,8 +613,6 @@ class OmniDiffusionConfig:
 
     # Attention
     diffusion_attention_config: "AttentionConfig" = field(default_factory=lambda: AttentionConfig())
-    # Opt-in FlashAttention deterministic kernel (slower). Default False keeps
-    # serving on the fast library path; enable via --fa-deterministic.
     fa_deterministic: bool = False
 
     # Running mode
--- a/tests/e2e/accuracy/test_qwen_image.py
+++ b/tests/e2e/accuracy/test_qwen_image.py
@@ -29,7 +29,7 @@ HEIGHT = 512
 NUM_INFERENCE_STEPS = 20
 TRUE_CFG_SCALE = 4.0
 SEED = 42
-SSIM_THRESHOLD = 0.94
+SSIM_THRESHOLD = 0.97
 PSNR_THRESHOLD = 30.0
 
 MODEL_2512_ID = "Qwen/Qwen-Image-2512"
@@ -69,7 +69,15 @@ def _local_files_only(model: str) -> bool:
 
 
 def _run_vllm_omni_qwen_image(*, model: str, output_path: Path) -> Image.Image:
-    server_args = ["--num-gpus", "1", "--stage-init-timeout", "300", "--init-timeout", "900"]
+    server_args = [
+        "--num-gpus",
+        "1",
+        "--stage-init-timeout",
+        "300",
+        "--init-timeout",
+        "900",
+        "--fa-deterministic",
+    ]
     with OmniServer(model, server_args, use_omni=True) as omni_server:
         response = requests.post(
             f"http://{omni_server.host}:{omni_server.port}/v1/images/generations",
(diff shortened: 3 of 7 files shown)


NumberWan:
@yenuo26 PTAL

hsliuustc0106 (commented):
can we avoid env variable

david6666666:
please avoid ENV

vllm_omni/diffusion/attention/backends/flash_attn.py
```diff
@@ -299,13 +302,13 @@ def forward_cuda(
...
-                causal=self.causal,
-                softmax_scale=self.softmax_scale,
-            )
+            fa_kwargs = {
+                "causal": self.causal,
+                "softmax_scale": self.softmax_scale,
+            }
+            if self.fa_deterministic:
+                fa_kwargs["deterministic"] = True
+            out = flash_attn_func(query, key, value, **fa_kwargs)
```
david6666666, line 323: Small heads-up: this flag only reaches the dense `flash_attn_func` branch here — the varlen/packed/masked paths (and `forward_xpu`/`forward_npu`) never see it. Fine for Qwen-Image (dense, unmasked), but anyone enabling `VLLM_OMNI_FA_DETERMINISTIC=1` on a model that takes a varlen/masked path would get no determinism, silently. A debug log when the flag is set on a non-dense path would make that visible.

vllm_omni/diffusion/data.py
```diff
@@ -1027,6 +1030,10 @@ def __post_init__(self):
         self.diffusion_kv_cache_skip_step_indices = parse_kv_cache_skip_selector(self.diffusion_kv_cache_skip_steps)
         self.diffusion_kv_cache_skip_layer_indices = parse_kv_cache_skip_selector(self.diffusion_kv_cache_skip_layers)
 
+        # Env override for FA determinism (accuracy CI). Explicit config True wins.
```
david6666666: Nit: the comment says "Explicit config True wins", but as written `VLLM_OMNI_FA_DETERMINISTIC=1` always forces `True` over a config `False` (and can't disable an explicit `True` either). Maybe "env can only enable, never disable" reads closer to what it does.

vllm_omni/diffusion/attention/backends/flash_attn.py
fhfuih, line 1: If this is after all an env var injection, what do you think if we directly read env var from attention backend's `__init__`, instead of maintaining a second variable in od_config?

If you choose to present the field in od_config, I would assume is it intended to be a user-aware toggle (instead of hidden and secretly used by CI). In this case, I think you should let CI pass an extra server_arg (to be forwarded to od_config), not setting an env var.

NumberWan:
> can we avoid env variable

Thanks — Dropped the env var and switched to a CLI flag --fa-deterministic that forwards into OmniDiffusionConfig.fa_deterministic. The accuracy test enables it via OmniServer server_args.

NumberWan:
> please avoid ENV

Thanks — Dropped the env var and switched to a CLI flag --fa-deterministic that forwards into OmniDiffusionConfig.fa_deterministic. The accuracy test enables it via OmniServer server_args.

vllm_omni/diffusion/attention/backends/flash_attn.py
NumberWan: Qwen-Image uses the dense unmasked path, so the accuracy test is covered. Added a warning_once when fa_deterministic=True but the call takes piecewise / packed / masked / varlen-fallback paths, so enabling the flag on those models isn’t silent.

vllm_omni/diffusion/data.py
NumberWan: That comment is gone with the env avoided.

vllm_omni/diffusion/attention/backends/flash_attn.py
NumberWan: Applied with the second option. Kept fa_deterministic on OmniDiffusionConfig as a user-facing toggle and wired CLI --fa-deterministic → config; the accuracy test passes it as a server_arg. No env injection in the attention backend.

david6666666 (commented):
LGTM now

yenuo26:
https://buildkite.com/vllm/vllm-omni/builds/13321/canvas?sid=019fe28b-f393-49ea-bf70-635e913ef58d&tab=output
However, in practice, the SSIM can still drop to 0.958. @NumberWan

PR #5888, opened 2026-08-07
[Model] Add SkyReels V3 A2V (audio-to-video / talking avatar)

NancyFyong:
# [Model] Add SkyReels-V3 A2V (audio-to-video / talking avatar)

## Summary

Adds `SkyReelsV3A2VPipeline` to vLLM-Omni, integrating Skywork's `SkyReels-V3-A2V-19B` talking-avatar model (portrait image + driving audio + text prompt → video). Wires L1 CPU unit tests, L4 Function + Perf nightly cases, and extends `send_video_diffusion_request` so future audio-conditioned X2V pipelines share one client helper.

- Commit `Add SkyReels V3 A2V diffusion support` — pipeline + vendored components (transformer, VAE, T5, CLIP, wav2vec2, XLM-Roberta, tokenizers, RoPE, avatar utils), model routing, CPU unit tests, `--audio` example CLI, docs, and one VAE-decode dtype fix.
- Commit `[Test][CI] Wire SkyReels V3 A2V into L4 nightly (function + perf + audio_reference helper)` — L4 expansion test, perf JSON, `audio_reference` handling in `send_video_diffusion_request`, and nightly YAML entries.

Depends on the R2V PR [#5856](https://github.com/vllm-project/vllm-omni/pull/5856) (`Add SkyReels V3 R2V diffusion support` + `Use SkyReels-V3 training resolution buckets for R2V output size` + `[Doc] Add SkyReels V3 R2V preview image`). While #5856 is unmerged this PR shows all 5 commits against `main`; once #5856 lands, rebasing this branch collapses the diff to the top 2 A2V commits (~4908 net lines).

AI assistance disclosure: this branch was co-authored with Claude Code. Every changed line was reviewed by NancyFyong and validated by the tests below.

## What A2V is

A2V is one of the three SkyReels-V3 pipelines (R2V / V2V / A2V). Given a single portrait image plus a driving audio clip (up to ~200 s in the reference implementation; this PR emits one 81-frame ≈ 3.24 s clip per request), it generates lip-synced talking avatar video. Audio conditioning goes through Chinese-wav2vec2 → audio projection MLP → per-frame audio context tokens injected into the DiT via `SingleStreamMutiAttention` blocks. Text conditioning is UMT5-XXL. Image conditioning is CLIP visual (fp16 autocast) + VAE-encoded first frame. Three-branch CFG (`cond`, `drop_text`, `uncond`) with independent `text_guide_scale` and `audio_guide_scale`.

Default preset (see `pipeline_skyreels_v3_a2v.py`):

| Field | Default |
|---|---|
| `num_inference_steps` | 40 |
| `num_frames` | 81 |
| `fps` | 25 |
| Resolution bucket | 480P (720P available) |
| `text_guide_scale` | 1.0 |
| `audio_guide_scale` | 1.0 |
| `flow_shift` | 11.0 |

At the shipped default (`guide_scale = 1.0`) three-branch CFG is disabled — the pipeline runs a single conditional forward per step. Setting either guide scale > 1 activates full CFG and roughly triples per-step compute.

## Correctness

**Dtype fix** — `pipeline_skyreels_v3_a2v.py:742`. The initial port cast the latent to `param_dtype` (BF16) before `WanVAE.decode`, but the vendored `Wan2.1_VAE.pth` is fp32 and its `Conv3d`/`RMS_norm` layers reject a BF16 activation with:

```
Input type (c10::BFloat16) and bias type (float) should be the same
```

Which surfaced as `Dummy run failed: ...` at engine warmup. Fix restores the fp32 input path (VAE encode already ingests fp32; decode now matches).

**Unit tests (L1, CPU)** — `tests/diffusion/models/skyreels_v3/test_skyreels_v3_a2v.py`, 7 cases:

```text
$ pytest -s -v tests/diffusion/models/skyreels_v3/test_skyreels_v3_a2v.py -m "core_model and cpu" --run-level=core_model
7 passed in 22.49s
```

Coverage: model class routing (config-based + path-based), stage-config inference short-circuit, dummy-run frame count declaration, preprocess audio+image reshape into 480P bucket, preprocess rejection of missing/short audio, extra-body param exposure. `core_model` marker was missing on the initial file; now added so `test-ready.yml → Simple Unit Test` picks it up.

**Runtime correctness on 1 × H20** — end-to-end offline inference against `Skywork/SkyReels-V3-A2V-19B`, via the standard `examples/offline_inference/image_to_video/image_to_video.py` CLI:

| Run | Frames | Steps | Guidance | GPUs | Peak GPU mem | Model init | Generation time | Notes |
|---|---:|---:|---:|---:|---:|---:|---:|---|
| Smoke | 25 | 4 | 1.0 | 1×H20 (96 GB) | ~50 GB | 253 s | **19.97 s** | 480P, sanity check |
| Full defaults | 81 | 40 | 1.0 | 1×H20 | 60.9 GB | 223 s | **721.50 s** | production preset (no CFG active at default guide=1) |
| Baseline (perf JSON case 1) | 81 | 4 | 1.0 | 1×H20 | ~55 GB | 225 s | **84.24 s** | 4-step preset, no CFG |
| CFG (single-GPU) | 81 | 4 | 5.0 | 1×H20 | ~65 GB | 298 s | **225.27 s** | three-branch CFG active (~56 s/step, 3× baseline per step) |
| **CFG-Parallel-2 (perf JSON case 2)** | 81 | 4 | 5.0 | 2×H20 | 60.9 GB / card | 243 s | **156.89 s** | **1.44× speedup vs 1-GPU CFG** — three CFG branches split rank 0: {cond, drop_text}, rank 1: {uncond} |
| CFG-Parallel-2 at default guide=1 | 81 | 40 | 1.0 | 2×H20 | 60.9 GB / card | 253 s | 722.30 s | byte-identical output to 1-GPU baseline (CFG code path skipped) |

The `guide_scale=1.0` cross-check (1-GPU vs 2-GPU CFG-Parallel producing identical 510828-byte MP4s) proves the CFG-parallel path is a no-op when three-branch CFG is not triggered. Speedup only materializes when a caller sets `text_guide_scale > 1` or `audio_guide_scale > 1`.

Two additional dtype/offload behaviors surfaced during validation and are called out separately below:

- **Fixed in this PR**: `pipeline_skyreels_v3_a2v.py:742` — VAE decode was passing a BF16 latent to fp32 VAE weights, causing "Input BF16, bias Float" during dummy_run. Now casts to fp32.
- **Follow-up dtype bug**: `--enable-cpu-offload` triggers "Input cuda.HalfTensor / weight HalfTensor" inside `CLIPModel.visual` — the offloader parks `clip.model` on CPU but no hook reloads it before its forward call. `SupportsComponentDiscovery` declares `_encoder_modules=["text_encoder", "clip.model", "audio_encoder"]` but the CLIPModel wrapper isn't `nn.Module` so its `.visual()` doesn't participate in the offloader's reload discipline. Not blocking for baseline / CFG-parallel deploys — deferred to a companion PR. This case is excluded from the L4 expansion matrix until fixed.

Sample media hosted on the [`skyreels-v3-a2v-assets`](https://github.com/NancyFyong/vllm-omni/tree/skyreels-v3-a2v-assets/assets/skyreels-v3-a2v) fork branch (same convention as `skyreels-v3-r2v-assets` / `lumina-image2-assets`), not committed to the tree. Config, timing, per-file description, and reproduction commands are in that branch's README.

### Side-by-side: Skywork upstream ‖ vLLM-Omni (seed=42, 4-step, 480P, `guide=1`)

<video src="https://github.com/NancyFyong/vllm-omni/raw/skyreels-v3-a2v-assets/assets/skyreels-v3-a2v/side_by_side_official_vs_omni_4step.mp4" controls muted playsinline width="800"></video>

Left panel is Skywork's `generate_video.py --task_type talking_avatar` (first 81 of its 125 frames — the overlap window). Right panel is vLLM-Omni's `image_to_video.py` at the same recipe. Same subject / studio / lighting; small per-frame lip-sync micro-motion differences are expected from two independent samplers off the same seed.

### 5-row static frame panel with PSNR/SSIM labels

Reviewers who want to reproduce: `--model Skywork/SkyReels-V3-A2V-19B --image <portrait> --audio <audio> --seed 42 --extra-body '{"resolution": "480P"}'` and match the num_frames / num_inference_steps / guidance-scale columns above.

**Quality parity with the Skywork upstream implementation** — direct 4-step 480P run with seed=42 on the same portrait + 5 s driving audio. Skywork ran in a dedicated `uv` venv (`torch==2.8.0+cu128`, `flash-attn==2.8.3.post1`, `diffusers==0.34.0`, `xfuser==0.4.3`) since Skywork pins conflict with vLLM-Omni's `torch==2.11` venv.

| | Skywork official | vLLM-Omni |
|---|---|---|
| Output frames | 125 (stitches 3 × 81-frame chunks to cover the full 5 s audio) | 81 (single chunk; audio > 3.24 s truncated) |
| Per-step time | **17.74–17.83 s/step** | **17.74 s/step** (byte-for-byte identical kernel path) |
| Total generation time | ~230 s (3 × 4 sampling steps + audio muxing) | **84.24 s** (4 sampling steps) |
| End-to-end init | 60 s weight load + 11.29 s audio preprocess | 225 s (spawn multiproc + weight load; one-time) |

Per-frame quality on the 81 overlapping frames (`metrics.json`):

| Metric | Mean | Median | Min | Max |
|---|---:|---:|---:|---:|
| PSNR | 17.996 dB | 17.447 dB | 15.787 dB | **40.889 dB** (frame 0, the shared reference portrait) |
| SSIM | 0.590 | 0.539 | 0.470 | **0.977** (frame 0) |

The `compare.png` side-by-side at frames {0, 20, 40, 60, 80} shows both trajectories converge to the same person, background, lighting, and studio setup with small per-frame lip/head micro-motion differences. Frame 0 (the input portrait re-emitted after VAE round-trip) matches to 40.9 dB / 0.977 SSIM. Numerical divergence on frames 1+ is expected and comes from four independent sources documented in `results/README.md`:

1. Different RNG entry (`torch.manual_seed` + `torch.randn` in Skywork vs `randn_tensor(generator=...)` in vLLM-Omni),
2. Different flow-matching step implementation (Wan2.2 shared `unipc` here vs Skywork's own),
3. Skywork's 3-chunk stitching adjusts `audio_start_idx` (0, 64, …) even for the first chunk,
4. Slightly different fp16/BF16 cast boundaries in the CLIP → transformer handoff.

Bit-exact parity was not a goal for this PR (same seed producing byte-identical output between two independent samplers isn't guaranteed for diffusion pipelines). The stronger claim we can support with these numbers is that **vLLM-Omni runs the same weights, same architecture, same 17.74 s/step transformer path, and produces a visually equivalent talking-avatar trajectory** — which is what matters for the model integration.

Comparison artifacts (mp4s + per-frame frames + `metrics.json` + `compare.png` + logs + `README.md` explaining the setup) live on the fork's [`skyreels-v3-a2v-assets`](https://github.com/NancyFyong/vllm-omni/tree/skyreels-v3-a2v-assets/assets/skyreels-v3-a2v) branch. Direct URLs are inlined in the sample-media block above.

Related follow-up implied by the frame-count mismatch:

- vLLM-Omni's A2V should adopt **multi-chunk stitching** (Skywork's motion-frame + drop-frame protocol) so audio > 3.24 s is not silently truncated. Non-blocking for this PR because the current one-chunk output is correct within its scope; blocking for feature parity.

## Parallelism / acceleration coverage in this PR

| Feature | A2V status | Notes |
|---|---|---|
| CPU offload (`--enable-cpu-offload`, `--enable-layerwise-offload`) | **partial** | `SupportsComponentDiscovery` declares `_encoder_modules=["text_encoder", "clip.model", "audio_encoder"]`, but `CLIPModel` is a wrapper class (not `nn.Module`); the offloader parks `clip.model` on CPU but `self.clip.visual()` reaches the CPU-side weight and errors on dtype mismatch. Fix deferred to a follow-up |
| CFG-Parallel (2 GPUs) | supported | Value only at `guide_scale > 1.0`; three-branch A2V CFG splits `cond/drop_text/uncond` across ranks via `CFGParallelMixin.combine_multi_branch_cfg_noise`. Measured 1.44× speedup on 4-step 480P @ g=5 |
| USP / Sequence Parallel | **not** supported | The vendored `WanModel` uses per-item variable-length `rope_apply` (iterates over `grid_sizes` per batch item) and `flash_attention(k_lens=seq_lens, window_size=...)`. Adding `_sp_plan` requires rewriting the forward path to unified `[B, seq, dim]` semantics + a shardable RoPE — several hundred lines of risky code that would need per-layer numerical alignment vs Skywork upstream. Follow-up PR. |
| HSDP | **not** supported | Same reason (would compose with USP). |
| Tensor-Parallel | **not** supported | Vendored attention uses raw `nn.Linear` (not `QKVParallelLinear`/`RowParallelLinear`). Follow-up. |
| Cache-DiT | **not** wired | No `CUSTOM_DIT_ENABLERS` entry yet; `_repeated_blocks` also unset (`Regional compilation skipped because the model does not define _repeated_blocks.` warning at init). Follow-up. |

Follow-up work is intentionally kept out of this PR so the initial correctness change is reviewable in isolation.

## Tests wired

| Level | File | Marker | Runs where |
|---|---|---|---|
| **L1** | `tests/diffusion/models/skyreels_v3/test_skyreels_v3_a2v.py` | `core_model + diffusion + cpu` | `test-ready.yml` Simple Unit Test |
| **L4 Function** | `tests/e2e/online_serving/test_skyreels_v3_a2v_expansion.py` (2 cases: `cuda_a2v_baseline`, `cuda_a2v_cfg_parallel`) | `full_model + diffusion + H100` | Nightly `Diffusion X2V · Other Function Test` (`h100_2`) |
| **L4 Perf** | `tests/dfx/perf/tests/test_skyreels_v3_a2v_vllm_omni.json` (2 cases matching L4 Function deploys) | `full_model + diffusion + H100` | Nightly `Diffusion X2V · Perf Test` (`h100_2`) |

Baselines in the perf JSON are calibrated from measured 1×H20 numbers extrapolated to H100 (baseline 84.24 s at 4 steps; CFG-parallel `2` @ g=5 156.89 s = 1.44× speedup vs 1-GPU CFG). The **`cpu_offload` case is not included** in this PR — the offloader-CLIP wrapper mismatch mentioned above needs to be fixed first.

The perf JSON is **committed as intent** — the actual harness client (`benchmarks/diffusion/diffusion_benchmark_serving.py`) does not yet know about `--task a2v`, so `run_diffusion_benchmark.py` bails when it tries to build a text-only request against a pipeline that requires audio conditioning. The follow-up PR listed below adds `--task a2v` support (audio path selection + `audio_reference` form field via `RequestFuncInput.audio_paths` → `async_request_v1_videos`).

`send_video_diffusion_request` gained one code path: when `request_config["audio_reference"]` is present it is forwarded as the `audio_reference` form field carrying `{"audio_url": "..."}`. Both `data:audio/...;base64,...` and remote `http(s)://...` URLs flow through this string — the /v1/videos endpoint has no dedicated audio file-upload slot. This is fully backwards-compatible for callers that only pass image/video references.

## Test plan

- [x] L1 CPU unit tests locally (7/7 pass, `pytest ... -m "core_model and cpu" --run-level=core_model`).
- [x] Single-GPU offline inference at smoke config + production default preset + CFG-parallel-2 + guidance=5 four-step (matrix above).
- [x] Output MP4 sanity — 832×480 @ 25 fps × 81 frames + AAC audio track carried through; mid-frame PNG mean/std within expected range (not black, not saturated).
- [ ] Nightly `Diffusion X2V · Other Function Test` on `h100_2` — must run once (all three deploy variants).
- [ ] Nightly `Diffusion X2V · Perf Test` on `h100_2` — first run will calibrate the H100 baselines in `test_skyreels_v3_a2v_vllm_omni.json`.
- [ ] Cross-model quality parity — inherited from R2V PR's `official_vs_omni_compare.png` + `metrics.json`; A2V-specific numerical diff deferred to a follow-up once Skywork's inference env is available in CI.

## Not in this PR (intentional follow-ups)

- **Fix A2V `--enable-cpu-offload` dtype mismatch** — CLIPModel wrapper is not an `nn.Module`, so the offloader's reload hook doesn't fire on its `.visual()` forward path. Small fix; excluded from this PR to keep scope tight.
- **Extend `benchmarks/diffusion/diffusion_benchmark_serving.py`** with `--task a2v` support so the perf JSON's audio-conditioned cases actually run through `run_diffusion_benchmark.py`. Adds `RequestFuncInput.audio_paths`, wires `_load_a2v_data`, and appends an `audio_reference` form field in `async_request_v1_videos`.
- **Multi-chunk stitching** in the A2V pipeline so audio > 3.24 s isn't silently truncated (Skywork upstream chunks with motion-ref overlap; vLLM-Omni currently emits one 81-frame clip).
- USP / HSDP / TP for A2V (requires rewriting the vendored `WanModel` forward for uniform `[B, seq, dim]` semantics).
- Cache-DiT `CUSTOM_DIT_ENABLERS` entry + `_repeated_blocks` on `WanModel`.
- Refactoring `a2v_avatar_utils.py::ASPECT_RATIO_627/960` onto `skyreels_v3/aspect_ratio.py` (introduced in the R2V bucket fix commit) — needs to happen after R2V lands so the diff is small.


hsliuustc0106:
This PR touches **vllm_omni/diffusion/, examples/offline_inference/, tests/diffusion/, .buildkite/cuda/, docs/assets/** (30 files). Based on CODEOWNERS coverage of the changed files, the most-related reviewers appear to be:

> @david6666666 @SamitHuang @RuixiangMa

Issue #2201, opened 2026-03-25
[RFC] Streaming Video Input for Omni-Modal Real-Time Interaction


lishunyang12:
## Motivation

Real-time video understanding — where a model continuously processes a live camera or video stream and responds to user queries about what it sees — is rapidly becoming a core capability of frontier AI platforms (Gemini Live, Doubao, Apple Visual Intelligence). vLLM-OMNI is uniquely positioned: it already has Qwen3-Omni's 3-stage pipeline, async chunk streaming, and audio-in-video interleaving. This RFC proposes adding streaming video input to complete the "Doubao experience": point camera, speak question, get spoken answer.

**No open-source serving engine currently offers streaming video in with text+speech out.** This RFC fills that gap.

### Market Context

```mermaid
pie title AI Video Market Segments (2025 est., $B)
    "Visual Inspection — $30B" : 30
    "Video Analytics — $13B" : 13
    "Video Surveillance — $6B" : 6
    "Robotics / Embodied AI — $5B" : 5
```

Sources:
- Visual Inspection: [Mordor Intelligence, AI Visual Inspection Market 2025](https://www.mordorintelligence.com/industry-reports/ai-powered-visual-inspection-market)
- Video Analytics: [Mordor Intelligence, AI Video Analytics Market 2025](https://www.mordorintelligence.com/industry-reports/global-ai-video-analytics-market)
- Video Surveillance: [SNS Insider, AI in Video Surveillance Market 2025](https://www.globenewswire.com/news-release/2026/03/23/3260344/0/en/AI-in-Video-Surveillance-Market-Size-to-Worth-USD-49-02-Billion-by-2035-Research-by-SNS-Insider.html)
- AI video company funding $3.08B in 2025 (+94.6% YoY): [Crunchbase, AI Funding Trends 2025](https://news.crunchbase.com/ai/big-funding-trends-charts-eoy-2025/)

### Benchmarks

```mermaid
xychart-beta
    title "StreamingBench Scores — higher is better (human = 91.66)"
    x-axis ["Human", "Seed1.5-VL", "Gemini 1.5 Pro", "MiniCPM-o 2.6", "GPT-4o", "Qwen2-VL 7B"]
    y-axis "Score (%)" 50 --> 95
    bar [91.66, 82.80, 67.07, 66.01, 60.15, 54.14]
```

[...]

PR #5894, opened 2026-08-07
[FEAT][Fronted] Support incremental prefill for realtime video streaming

ZhengWG:
## Purpose

Implements the **incremental prefill** path called out in [RFC #2201](https://github.com/vllm-project/vllm-omni/issues/2201)(*Streaming Video Input for Omni-Modal Real-Time Interaction*):

> start encoding and prefilling video frames into KV cache as they arrive,
> so by the time the query comes in, most of the prefill is done and TTFT drops.

This PR lands that missing piece on `/v1/video/chat/stream` for **text-only output** sessions when stage-0 prefix caching is enabled (earlier phases already shipped the WS endpoint / EVS / audio-in-video / async-chunk plumbing from the same RFC):

- Eager warmup (`max_tokens=1`) on frame arrival → vision encode + prefill into the prefix cache before `video.query`.
- Stable per-frame `uuid` (arrival-time md5) shared by warmup and query prompts so mm hashes match and warmed KV is not invalidated.
- Query consumes buffered frames into append-only multimodal `message_history` (no query-time `num_frames` subsampling); commit on success, restore (+ thin) on interrupt/failure.
- Audio-output sessions stay on the legacy full-reprefill path (talker needs thinker hidden states over the whole prompt).

## Test Plan

E2E A/B verification on `/v1/video/chat/stream` with Qwen3-Omni-30B-A3B (3-stage: thinker on GPU0, talker + code2wav on GPU1, `async_chunk: true`, 2× H20 96GB). Two server configs differing **only** in stage-0 `enable_prefix_caching`:

- **APC-on** — feature active (incremental prefill + eager warmup)
- **APC-off** — feature dormant, legacy full-rebuild path (baseline)

Probe client: WebSocket multi-turn sessions with synthetic 320×240 JPEG frames (unique moving square per frame, `enable_frame_filter: false` so no frames are dropped), measuring per turn: query → `response.start` (ack), → first
`response.text.delta` (TTFT), → text/audio done; 120 s timeout guard to catch hangs.

Checks:

1. Text session (3 turns × 8 frames, APC-on): TTFT, `video-warmup-*` requests in server log, stage-0 prefix-cache hit rate from `/metrics`.
2. Same session on APC-off → TTFT baseline.
3. Audio session (2 turns × 6 frames, APC-on): no warmup requests, legacy full prefill, audio produced.
4. Identical audio session replayed (worst case: full prefix-cache hit on the whole prompt) → talker must not stall.
5. APC-off dormancy: zero warmup requests, zero prefix-cache queries.
6. Restore APC-on, smoke turn.

**vLLM Version:** 0.27.1

**vLLM-Omni Commit:** ff2f8d905a1a2e7fef11ceee7ae0ef9adadc3775

## Test Result

All checks passed.

**Text-turn TTFT (query → first text token):**

| Turn | APC-off (legacy) | APC-on (incremental) | Δ |
|---|---|---|---|
| 1 | 352.9 ms | 62.7 ms | **−82%** |
| 2 | 155.7 ms | 61.1 ms | −61% |
| 3 | 134.6 ms | 62.1 ms | −54% |

TTFT stays **flat (~62 ms) as context grows** while retaining the full multimodal history; the legacy path keeps only 8 sampled frames + 2 text-only history turns


ZhengWG:
@lishunyang12 @hsliuustc0106  PTAL~

vllm_omni/entrypoints/openai/video_stream_base.py
```diff
@@ -401,7 +637,8 @@ async def _processor() -> None:
                             dropped = frame_buffer.pop(0)
                             dropped_metadata = frame_metadata.pop(0)
                             dropped_frame_id = dropped_metadata.get("frame_id")
-                            frame_pil_cache.pop(dropped, None)
+                            if dropped not in pinned_frame_refs:
+                                _drop_frame_cache(dropped)
```
lishunyang12, line 655: Keep cached frames while duplicates remain buffered; this eviction silently drops a valid frame.

ZhengWG: Good catch — Fixed by  keeping the entry while any copy remains buffered (`dropped not in frame_buffer`).

vllm_omni/entrypoints/openai/video_stream_base.py
```diff
@@ -377,7 +613,7 @@ async def _processor() -> None:
                             frame_buffer[:] = [frame_buffer[index] for index in retained_indices]
                             frame_metadata[:] = [frame_metadata[index] for index in retained_indices]
                         if frame_pil_cache.get(frame_data) is _BAD_FRAME:
-                            frame_pil_cache.pop(frame_data, None)
+                            _drop_frame_cache(frame_data)
```
amy-why-3459, line 620: [P2] Preserve failed-decode status while a query snapshot is pinned

`_internal.frame_decode_failed` calls `_drop_frame_cache()` even when an in-flight query has pinned this frame and is waiting for decoding. The failure handler can run before that query resumes and snapshots `frame_pil_cache`, removing the `_BAD_FRAME` marker. `_select_frame_indices()` then treats the missing entry as usable (`None is not _BAD_FRAME`), while `_build_frame_image_parts()` skips it. `selected_metadata` is built from the former selection, so `video.frames.consumed` can report a failed frame that was never in the prompt, including an incorrect `latest_pts_ms`.

Reproduced with one delayed invalid frame followed by one valid frame: start the query before releasing the failing decoder; the query selects 2 frames but builds only 1 image part. Please retain the BAD marker/readiness state until pinned snapshots release it, or compute selection and metadata from the same successfully decoded image set. Add an asynchronous regression covering decode failure after the query snapshot is taken.

vllm_omni/entrypoints/openai/video_stream_base.py
```diff
@@ -304,35 +447,122 @@ async def _cancel_active_query(*, abort_now: bool = False) -> None:
...
 
+            # --- Incremental prefill: keep the prefix cache warmed to the latest
+            # buffered frame so a query only pays for its own text suffix. ---
+            warmup_task: asyncio.Task[Any] | None = None
+            warmup_request_id: str | None = None
+            warmed_signature: tuple[Any, ...] | None = None
+
+            def _context_signature() -> tuple[Any, ...]:
+                history_len = len(message_history) if isinstance(message_history, list) else None
+                return (history_len, len(frame_buffer), frame_buffer[-1] if frame_buffer else None)
```
amy-why-3459: [P2] Use a context revision that changes when any buffered frame changes

The signature only records history length, buffer length, and the last frame. Different windows can therefore compare equal. With `max_frames=2` and frame filtering disabled, start a delayed warmup for `[A,B]`, then receive `C` and `B` while it is running. The live buffer is now `[C,B]`, but both signatures are `(0, 2, B)`. When the first warmup finishes, `_maybe_start_warmup()` skips the new context. I reproduced exactly one warmup request where a second is needed; the next query cannot reuse the old vision prefix and pays the new window's prefill cost.

Use a monotonically increasing context revision, updated for every prompt-affecting buffer/history change, or include the complete ordered frame identity sequence. Please add a delayed-warmup/FIFO regression with a repeated final frame.

amy-why-3459 (changes requested):
Reviewed head `c0e40a2d96380fd6efb340628a670a2e7e5b9907` against the current upstream `main`, `23f41264456684c793283502f811aab7dcda2c88` (also this PR's base). I found two reproducible edge cases, detailed inline.

Direction relative to main / RFC #2201:

The overall direction is consistent: this extends the existing shared WebSocket handler and uses the engine's existing prefix cache to move vision prefill ahead of the query. There is no need to reject it merely because it uses short warmup requests rather than a persistent engine request.

However, the implementation is **request-level APC warmup for text-only sessions**, not the full persistent, append-only multimodal session KV described in the PR body/RFC. `_history_prefix_messages()` keeps only the latest two text messages in the prompt, the live frame buffer remains bounded and is not consumed on success, and each warmup creates a separate `max_tokens=1` request. Changing the history prefix or FIFO-evicting the first frame invalidates reuse of subsequent prefix blocks, so a long-running stream can require full-window prefill again. These are reasonable scope limitations, but they should be explicit in the PR purpose and evidence.

Please update the body, which still claims append-only multimodal history with commit/restore semantics and TTFT staying flat while retaining full multimodal history. Re-run or clearly qualify the A/B results for the reviewed head: the report cites `ff2f8d90`, and APC-on/off also use different frame-selection semantics, so the comparison does not isolate caching. Include a case beyond `max_frames` and more than a few turns, reporting cache-hit behavior and total prefill work as well as query TTFT. The new documentation correctly warns that audio-output isolation is not implemented; this should remain an explicit text-only deployment limitation, not a claim that the RFC's video-to-speech goal is complete.

Validation: two minimal asynchronous scenarios exercised the actual handler, frame decoding, frame selection, and prompt construction with a fake warmup engine. They reproduced (1) two selected frames but only one prompt image after a pinned decode failure and (2) no follow-up warmup when `[A,B]` changes to `[C,B]`. The harness bypassed the unrelated API-server package initializer and did not exercise real model inference. The repository test file could not be collected locally because the installed vLLM lacks `vllm.entrypoints.generate.base.protocol`; this is an environment limitation, not an attributed PR regression. No end-to-end latency or audio-output validation was performed.

vllm_omni/entrypoints/openai/video_stream_base.py
ZhengWG: Fixed

vllm_omni/entrypoints/openai/video_stream_base.py
ZhengWG: Fixed.

amy-why-3459:
Reviewed head `b52523f00123dc0bb0944ba8316c8f0ecc6e5b01`, including its relationship to #7585. I recommend addressing the following latency regression before merging.

**[P2] Do not make legacy queries wait for frames they will not select.**

At [`video_stream_base.py:565–569`](https://github.com/vllm-project/vllm-omni/blob/b52523f00123dc0bb0944ba8316c8f0ecc6e5b01/vllm_omni/entrypoints/openai/video_stream_base.py#L565-L569), `_await_frames_ready(query_frames)` unconditionally waits for the entire buffered window before frame selection. This also affects the legacy path, including sessions with APC disabled or audio output enabled.

I reproduced this with APC disabled and `num_frames=1`: decoding old frame A was delayed, while the latest frame B was already decoded. The query eventually selected only B, but could not start until A finished decoding. An unselected frame therefore adds latency to a query that does not need it.

Please determine the legacy sampling positions before waiting, await only those candidate frames, then derive the prompt and consumed metadata from the same successfully decoded selection. The incremental path can continue waiting for the full window. Add a regression where an unselected old frame remains blocked while the selected latest frame is ready. This reproduction used the actual session handler with unrelated engine-output imports isolated and query execution overridden to observe when it started; it was not a full-model E2E test.

**Relationship to #7585**

- #7585 merged on September 17, and its merge commit `882bf8bc95b1a9a08af667a7edd006ebd21e4b57` is already an ancestor of this head. There is no outstanding merge dependency, and the PRs do not directly overlap in changed files.
- This PR optimizes `/v1/video/chat/stream`, using `video.frame` plus `video.query` and separate APC warmup requests.
- #7585's shared UI uses `/v1/realtime`. Its Qwen VAD camera path creates image conversation items and submits generation through the duplex session/plugin lifecycle. Merging this PR will not automatically accelerate that path.
- Duplex initialization currently sets `openai_streaming_video` to `None`, so the legacy video endpoint is unavailable under that deployment. Neither this PR's request-level APC warmup nor Qwen's commit-based duplex integration provides persistent append-only multimodal KV continuation.

Please document this endpoint/deployment distinction explicitly. If the intended follow-up is to accelerate the shared Qwen camera UI, warmup should integrate with the duplex session/plugin lifecycle and its cancellation/history/request ownership, rather than simply redirecting the frontend to the legacy endpoint. Also, the documented lack of isolation between audio-output requests and enabled stage-0 APC means this PR's enablement configuration should not be applied unchanged to the shared voice-call deployment.

**Performance validation suggestion**

The current A/B submits the full buffered window with APC enabled but stride-samples with APC disabled, so it does not isolate the warmup benefit. The revised PR description correctly acknowledges this. Please compare identical submitted frame sequences, history, and sampling parameters while switching warmup, and include FIFO eviction, continuous frame arrival, and concurrent sessions when reporting TTFT and throughput.

The earlier findings about duplicate-frame eviction, failed-decode markers in pinned snapshots, and incomplete window signatures appear fixed in this revision and should not be raised again.

Validation: the latency regression above was reproduced in isolation. The full handler pytest suite could not be collected because the local vLLM installation lacks `vllm.entrypoints.generate.base.protocol`; no full-model E2E run was performed. The inspected head's pre-commit, build, and DCO checks were successful.

PR #5895, opened 2026-08-07
[Bugfix][MiniCPM-o] Wire the duplex sampling hook into the NPU AR model runner

psv666:
## Purpose

`test_duplex_single_session_response_required` passes on H100 but fails on A3 NPU. The MiniCPM-o 4.5 native duplex model answers turn 1 normally, then goes silent: every subsequent unit decides `listen` and `response.created` never arrives.

**What broke**

```
TimeoutError: Timed out waiting for 继续 response.created after 2 committed input turns;
model_listen=2 buffering_listen=0 errors=0
```

No server-side error, no crash, no OOM — the model simply never chooses to speak again.

**Root cause**

`GPUARModelRunner` and `NPUARModelRunner` are siblings, not parent and child: the former derives from `OmniGPUModelRunner`, the latter from `OmniNPUModelRunner`, so neither inherits the other's `_sample`. The NPU runner carried a copy of the GPU `_sample` with the `prepare_duplex_sampling` call dropped, and had no counterpart for the other three call sites (`__init__`, `load_model`, `_update_states`). Nothing under `vllm_omni/platforms/npu/` mentioned duplex at all.

`prepare_duplex_sampling` is the only writer of `_minicpmo45_duplex_row_sessions`, the batch-row -> session map. With that map empty on Ascend:

1. `_minicpmo45_duplex_state_for_row()` returns `None` for every row.
2. `_finalize_minicpmo45_native_duplex_sample()` can therefore never tell that a turn is still in progress, so it never rewrites a mid-turn `<|listen|>` into `<|tts_bos|>` — the sampled `<|listen|>` is emitted verbatim. **This is the `model_listen=2` in the failure.**
3. Two further consequences were silent: `force_listen` logit masking never fired, and the repetition penalty fell back to `recent_tokens` instead of the session-scoped `state.generated_tokens`.

Turn 1 still spoke because `_minicpmo45_native_duplex_prompt_rows()` has a fallback that identifies duplex rows by counting `<unit>` tokens in the prompt. So the native duplex sampler itself ran — it just ran blind to session state, and session state only matters for the continue-speaking decision.

**Fix**

Extract the four call sites into a platform-neutral `DuplexSamplingRunnerMixin` beside `DuplexSamplingHelper`, switch the GPU runner to it (pure move, no behavior change), and wire the NPU runner to the same mixin.

| Mixin method | Called from |
|---|---|
| `_init_duplex_sampling_state()` | `__init__` |
| `_resolve_duplex_sampling_hook(force=True)` | `load_model` |
| `_update_duplex_sampling_states()` | `_update_states` |
| `_apply_duplex_sampling()` | `_sample`, before the sampler |
| `_clear_duplex_sampling()` | teardown |

**Two things reviewers should weigh in on**

1. *The GPU-side refactor is not strictly required to fix the NPU bug.* A minimal patch would add four blocks to `npu_ar_model_runner.py` and leave `gpu_ar_model_runner.py` alone. I chose the shared mixin because copy-paste between these two siblings is exactly what produced this bug. Happy to shrink it to an NPU-only patch if maintainers prefer the smaller diff.
2. *The NPU AR runner does not gain a `shutdown()` override.* The GPU `shutdown()` calls `_clear_duplex_sampling()`; the NPU runner has no `shutdown()` today and this PR does not add one. That method only clears a set of request-id strings, holds no device memory, and matters only at process teardown.

## Test Plan

Run on A3 NPU with real weights. `--run-level` matters: it defaults to `core_model`, which patches every stage to `load_format: dummy` (random weights), so the speak/listen decision would be meaningless.

```bash
export VLLM_ALLOW_LONG_MAX_MODEL_LEN=1

# 1. The failing case on its own
pytest -s -v \
  "tests/e2e/online_serving/test_minicpmo_4_5_duplex.py::test_duplex_single_session_response_required" \
  --run-level advanced_model

# 2. The nightly duplex selection verbatim (.buildkite/npu/test-npu-nightly.yml:73-77)
pytest -s -v \
  tests/e2e/online_serving/test_minicpmo_4_5_duplex.py \
  tests/e2e/online_serving/test_minicpmo_4_5_duplex_expansion.py \
  -m "(full_model or advanced_model) and npu and A3 and omni" \
  --run-level full_model

# 3. Unit tests, including the new drift guards
pytest -q tests/worker/test_native_duplex_hooks.py
```

**vLLM Version:** 0.1.dev1+ge5588e49b

**vLLM-Omni Commit:** 0.25.0rc2.dev151+g8989a0732

## Test Result

Hardware: A3 NPU, 1 card, MiniCPM-o 4_5, three-stage-single-gpu deployment.

| Test | Before | After |
|---|---|---|
| `test_duplex_single_session_response_required` | `TimeoutError` (`model_listen=2 errors=0`) | **1 passed** in 351.51s |
| NPU nightly duplex selection | `1 failed, 2 passed, 2 deselected` | **3 passed, 2 deselected** in 685.90s |
| `tests/worker/test_native_duplex_hooks.py` | 57 passed | **61 passed** in 2.40s |

The "before" column for the first two rows is Buildkite `vllm-omni-npu-ci` build 4866.

### Regression coverage

The new tests parse `gpu_ar_model_runner.py` and `npu_ar_model_runner.py` with `ast` and assert both runners inherit the mixin and call it from all four methods. Source parsing rather than imports, because `NPUARModelRunner` needs `vllm_ascend`, which is absent on the CPU images that run this test file — the precedent is `tests/core/sched/test_sched_package_imports.py`.

Verified the guard actually bites: deleting the `_apply_duplex_sampling` line from the NPU `_sample` produces

```
AssertionError: NPUARModelRunner._sample must call self._apply_duplex_sampling
```

The 57 pre-existing tests passing unchanged is what establishes that the GPU-side move is behavior-preserving.

### DIFF
--- a/tests/worker/test_native_duplex_hooks.py
+++ b/tests/worker/test_native_duplex_hooks.py
@@ -2466,3 +2466,71 @@ def test_minicpmo_stage0_session_context_includes_resolved_ref_audio():
 
     assert state.context_token_ids == [1, 2, 3, 151683, 151683, 4, 5]
     assert len(state.context_embeds) == 6
+
+
+# ---------------------------------------------------------------------------
+# Cross-platform wiring: every AR model runner must reach the duplex hook.
+#
+# ``GPUARModelRunner`` and ``NPUARModelRunner`` are siblings -- neither inherits
+# the other's ``_sample`` -- so the NPU runner once carried a copy of the GPU
+# ``_sample`` with the ``prepare_duplex_sampling`` call dropped.  That left the
+# MiniCPM-o 4.5 thinker without a row -> session map on Ascend, so every mid-turn
+# ``<|listen|>`` was emitted verbatim and the model stopped speaking after turn 1.
+#
+# These checks parse the source instead of importing it: the NPU runner needs
+# ``vllm_ascend``, which is absent on the CPU test images that run this file.
+# ---------------------------------------------------------------------------
+
+_AR_RUNNERS = [
+    ("vllm_omni/worker/gpu_ar_model_runner.py", "GPUARModelRunner"),
+    ("vllm_omni/platforms/npu/worker/npu_ar_model_runner.py", "NPUARModelRunner"),
+]
+
+# method that must call it -> mixin method it must call
+_REQUIRED_DUPLEX_CALLS = {
+    "__init__": "_init_duplex_sampling_state",
+    "load_model": "_resolve_duplex_sampling_hook",
+    "_update_states": "_update_duplex_sampling_states",
+    "_sample": "_apply_duplex_sampling",
+}
+
+
+def _ar_runner_classdef(relative_path: str, class_name: str):
+    import ast
+    from pathlib import Path
+
+    source = (Path(__file__).parents[2] / relative_path).read_text()
+    for node in ast.walk(ast.parse(source)):
+        if isinstance(node, ast.ClassDef) and node.name == class_name:
+            return node
+    raise AssertionError(f"{class_name} not found in {relative_path}")
+
+
+@pytest.mark.parametrize(("relative_path", "class_name"), _AR_RUNNERS)
+def test_ar_runner_inherits_duplex_sampling_mixin(relative_path: str, class_name: str):
+    import ast
+
+    classdef = _ar_runner_classdef(relative_path, class_name)
+    bases = {base.id for base in classdef.bases if isinstance(base, ast.Name)}
+    assert "DuplexSamplingRunnerMixin" in bases, (
+        f"{class_name} must inherit DuplexSamplingRunnerMixin so the duplex sampling "
+        f"hook is wired the same way on every platform"
+    )
+
+
+@pytest.mark.parametrize(("relative_path", "class_name"), _AR_RUNNERS)
+def test_ar_runner_calls_every_duplex_sampling_hook_site(relative_path: str, class_name: str):
+    import ast
+
+    classdef = _ar_runner_classdef(relative_path, class_name)
+    methods = {node.name: node for node in classdef.body if isinstance(node, ast.FunctionDef)}
+
+    for method_name, required_call in _REQUIRED_DUPLEX_CALLS.items():
+        method = methods.get(method_name)
+        assert method is not None, f"{class_name}.{method_name} is missing; it must call {required_call}"
+        called = {
+            node.func.attr
+            for node in ast.walk(method)
+            if isinstance(node, ast.Call) and isinstance(node.func, ast.Attribute)
+        }
+        assert required_call in called, f"{class_name}.{method_name} must call self.{required_call}"
--- a/tests/worker/test_native_duplex_hooks.py
+++ b/tests/worker/test_native_duplex_hooks.py
@@ -2481,11 +2481,6 @@ def test_minicpmo_stage0_session_context_includes_resolved_ref_audio():
 # ``vllm_ascend``, which is absent on the CPU test images that run this file.
 # ---------------------------------------------------------------------------
 
-_AR_RUNNERS = [
-    ("vllm_omni/worker/gpu_ar_model_runner.py", "GPUARModelRunner"),
-    ("vllm_omni/platforms/npu/worker/npu_ar_model_runner.py", "NPUARModelRunner"),
-]
-
 # method that must call it -> mixin method it must call
 _REQUIRED_DUPLEX_CALLS = {
     "__init__": "_init_duplex_sampling_state",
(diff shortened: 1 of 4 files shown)


amy-why-3459:
## Summary

The root cause analysis is accurate and the fix direction is correct. This is a classic copy-drift bug: `GPUARModelRunner` and `NPUARModelRunner` are siblings (the former derives from `OmniGPUModelRunner`, the latter from `OmniNPUModelRunner`), so the NPU runner carried a copy of the GPU `_sample` with the `prepare_duplex_sampling` call dropped. With no row → session map on Ascend, `_finalize_minicpmo45_native_duplex_sample()` could never rewrite a mid-turn `<|listen|>` into `<|tts_bos|>`, and the MiniCPM-o 4.5 thinker went silent after the first turn (`model_listen=2 errors=0`).

Extracting the four call sites into a platform-neutral `DuplexSamplingRunnerMixin` and guarding the seam with AST-based tests is the right call — copy-paste is exactly what produced this bug.

**Verdict: Approve with nits.** The functionality is correct and the fix is effective. A few robustness points are worth addressing before or right after merge.

---

## 1. Core design — LGTM

The four mixin call sites map cleanly onto the runner lifecycle:

| Mixin method | Called from |
|---|---|
| `_init_duplex_sampling_state()` | `__init__` |
| `_resolve_duplex_sampling_hook(force=True)` | `load_model` |
| `_update_duplex_sampling_states()` | `_update_states` |
| `_apply_duplex_sampling()` | `_sample`, before the sampler |
| `_clear_duplex_sampling()` | teardown |

The MRO (`NPUARModelRunner → OmniNPUModelRunner → … → OmniConnectorModelRunnerMixin → DuplexSamplingRunnerMixin`) is valid: `OmniConnectorModelRunnerMixin` does not define `load_model` / `_update_states`, so `super()` chains resolve correctly. The `TYPE_CHECKING` imports are safe under `from __future__ import annotations`.

---

## 2. (P1) The AST test only asserts the call *exists*, not that it runs *before* the sampler

`test_ar_runner_calls_every_duplex_sampling_hook_site` walks `ast.Call` nodes and passes as soon as `self._apply_duplex_sampling(...)` appears in `_sample`. That is a weak guard: if someone moves `_apply_duplex_sampling` *after* `model_sample(...)` in the NPU `_sample`, the hook silently stops firing and the test still passes — the exact regression this PR is meant to prevent.

```python
prepared_sampling_metadata = self._sampling_metadata_for_model_sampler(sampling_metadata)
self._apply_duplex_sampling(logits, prepared_sampling_metadata)
sampler_output = model_sample(logits, prepared_sampling_metadata)
```

Suggestions:

- Assert `_apply_duplex_sampling` appears *before* the `model_sample` call (compare `lineno` / `col_offset` in source order); and/or
- Assert the sampler call comes last within the `prefer_model_sampler` branch.

This is the highest-value regression guard in the PR and should land before merge.

---

## 3. (P1) `_init_duplex_sampling_state()` runs at the *end* of NPU `__init__`

On the NPU side the call is appended at the end of a long `__init__`:

```python
self._downstream_payload_cache: dict[str, bool] = {}
self._init_duplex_sampling_state()
```

`_update_states` tolerates the missing attributes via `getattr(..., False)` / `getattr(..., None)` defaults, so nothing crashes — but if `super().__init__()` ever triggers `_update_states` (e.g. a warmup path), the duplex attributes are silently uninitialized. Move the call to the top of `__init__` (before `super().__init__()`) to match the GPU side and remove the implicit "missing attribute means uninitialized" dependency.

---

## 4. (P2) The AST runner list needs manual maintenance for future platforms

`_AR_RUNNERS` hard-codes the two current runners:

```python
_AR_RUNNERS = [
    ("vllm_omni/worker/gpu_ar_model_runner.py", "GPUARModelRunner"),
    ("vllm_omni/platforms/npu/worker/npu_ar_model_runner.py", "NPUARModelRunner"),
]
```

A third platform runner (x86, XPU, etc.) would silently miss the duplex wiring unless the list is hand-edited. Consider scanning `vllm_omni/platforms/*/worker/*_ar_model_runner.py` (or at least adding a comment telling future contributors to append here).

---

## 5. (P2) No symmetric `shutdown()` cleanup point on the NPU runner

The PR explicitly declares this: the NPU AR runner has no `shutdown()` override, so `_clear_duplex_sampling()` never runs on Ascend. The helper only holds a set of request-id strings and matters only at process teardown, so the impact is small — but the GPU side cleans up in `shutdown()` while the NPU side has no counterpart. Options:

- Document it (already done in the PR description); and/or
- Make `_apply_duplex_sampling` self-clear when `helper.active_request_ids` is empty, which is more robust than depending on a teardown path that may not exist.

---

## 6. (P2) `_apply_duplex_sampling` behavior when `helper is None`

```python
prepare_duplex_sampling = self._resolve_duplex_sampling_hook()
if prepare_duplex_sampling is None:
    return
helper = getattr(self, "_duplex_sampling_helper", None)
rows = helper.rows(self) if helper is not None and helper.active_request_ids else ()
if rows or (helper is not None and helper.hook_active):
    prepare_duplex_sampling(logits, prepared_sampling_metadata, rows)
if helper is not None:
    helper.hook_active = bool(rows)
```

This is a verbatim move of the GPU logic — no behavior change. Flagging only because if `_duplex_sampling_hook` is resolved but `_duplex_sampling_helper` was never created (a state the current control flow cannot reach, since they are created together), rows would silently be empty. Not a bug today; worth a comment.

psv666:
Thanks for the review — points 2 and 4 are real and are addressed in c2d85781. Pushing back on 3, 5 and 6 below with the specifics.

## Adopted

### 2. Ordering guard — agreed, this was the weak spot

You're right that the check passed as soon as the call appeared anywhere in `_sample`. Ordering is load-bearing: `_apply_duplex_sampling` masks the logits (`force_listen`) *and* publishes the row → session map that the model's own sampler reads, so running it after `model_sample()` is exactly as silent as not running it at all.

Added `test_ar_runner_applies_duplex_sampling_before_the_model_sampler`, which compares source line numbers. It also fails loudly if `model_sample` is ever renamed, so the check can't rot into a vacuous pass.

Verified by injection — moving the NPU call after the sampler now gives:

```
AssertionError: vllm_omni/platforms/npu/worker/npu_ar_model_runner.py::NPUARModelRunner._sample
calls self._apply_duplex_sampling at line 891, after model_sample() at line 890.
```

### 4. Runner discovery — adopted, and it's less hypothetical than it looks

A third AR runner already exists: `vllm_omni/platforms/xpu/worker/xpu_ar_model_runner.py`. No actual gap, because `XPUARModelRunner(GPUARModelRunner)` inherits the wiring for free — but it does mean the plain glob you suggested would have **failed** on it, since its class body defines none of the four methods.

So the discovery is inheritance-aware: scan `vllm_omni/worker/gpu_ar_model_runner.py` plus `vllm_omni/platforms/*/worker/*_ar_model_runner.py`, and assert each `*ARModelRunner` either wires the hook itself **or** derives from one that does.

Verified both directions: a newly added unwired platform runner fails with an actionable message, and the real `XPUARModelRunner` stays green.

## Not adopting

### 3. `_init_duplex_sampling_state()` placement

> Move the call to the top of `__init__` (before `super().__init__()`) to match the GPU side

The GPU side already has it at the end, after `super().__init__()` — the NPU placement matches it exactly:

```
gpu_ar_model_runner.py:304    super().__init__(*args, **kwargs)
gpu_ar_model_runner.py:337    self._downstream_payload_cache: dict[str, bool] = {}
gpu_ar_model_runner.py:338    self._init_duplex_sampling_state()

npu_ar_model_runner.py:110    super().__init__(*args, **kwargs)
npu_ar_model_runner.py:143    self._downstream_payload_cache: dict[str, bool] = {}
npu_ar_model_runner.py:144    self._init_duplex_sampling_state()
```

Moving only the NPU call would *introduce* the GPU/NPU divergence this PR exists to remove. On the `super().__init__()` concern: `_update_states` takes a `SchedulerOutput` and runs per scheduler step; there is no construction-time path that reaches it. Happy to move **both** call sites if you'd still prefer it, but they should stay in lockstep.

### 5. Self-clearing when `active_request_ids` is empty

> Make `_apply_duplex_sampling` self-clear when `helper.active_request_ids` is empty

That's what `hook_active` already does:

```python
rows = helper.rows(self) if helper is not None and helper.active_request_ids else ()
if rows or (helper is not None and helper.hook_active):
    prepare_duplex_sampling(logits, prepared_sampling_metadata, rows)   # one final call with rows == ()
if helper is not None:
    helper.hook_active = bool(rows)
```

When the last duplex row leaves the batch, the hook fires once more with empty rows so the model drops its stale row → session mapping, then `hook_active` goes False and the helper stops scanning request state. The mixin docstring on `_apply_duplex_sampling` spells this out, and `tests/worker/test_native_duplex_hooks.py::test_minicpmo_duplex_sample_clears_stale_rows_once_without_scanning` covers it.

`shutdown()`'s `helper.clear()` is belt-and-braces on top of that: `_update_states` already drops finished request ids every step, so the set is empty long before teardown. That's why the missing NPU `shutdown()` is a documentation note rather than a leak.

### 6. Comment on the `helper is None` branch

As you note, the hook and the helper are created together in `_resolve_duplex_sampling_hook`, so the state is unreachable. The `getattr(..., None)` reads are a verbatim move of the existing GPU code, not something this PR introduces. I'd rather not annotate an unreachable branch — a comment there reads as evidence that it *is* reachable and invites someone to write handling for it. If the goal is to make the invariant explicit, the cleaner follow-up is deleting the `helper is not None` guards entirely, which is a separate change to unmoved code.

---

`tests/worker/test_native_duplex_hooks.py`: 62 passed. No production code changed since the NPU validation in the PR description, so those A3 results still stand.

PR #5899, opened 2026-08-07
[Perf][MiniMax H3] Skip redundant per-step token-refiner recompute

ChoHee15:
## Purpose

MiniMax H3's token-refiner output (`condition_proj` → `token_refiner`) depends only on the request's text embeddings and refiner packed-sequence layout — it has **no timestep dependency** — yet it is recomputed inside every denoise forward. A 50-step request runs the projection + two refiner blocks 50 times when once is enough.

This PR removes that redundancy in an **offload-mode-aware** way:

- extract `MiniMaxH3DiTModel.compute_refined_text(...)`, a request-static helper outside the regionally compiled DiT blocks (no torch.compile / cudagraph capture boundary crossed);
- **layer-wise / no offload** (refiner weights resident for the whole request): `MiniMaxH3DenoiseBranch.prepare_refined_text(model)` runs it **once** before the loop; every `forward` consumes the cached `refined_text_embed`;
- **model (CPU) offload**: the model-level hook only onloads the transformer *inside* `forward`. Hoisting would run the refiner while its weights are still on CPU, forcing an extra CPU↔GPU round-trip a two-block refiner can't pay back — so the loop does **not** hoist; `forward` refines internally each step, exactly matching the pre-optimization path.

For valid inputs this does not change the denoise algorithm or outputs.

### Correctness

Generated video + audio SHA256 match baseline **bit-for-bit**, verified on **two independent H20 machines** with separately downloaded weights, in **both** offload modes:
- video `9d76478d…`, audio `a7f6bb80…` — identical across both machines and both versions.

### Performance

Single H20, t2va 448×256, `duration=4.0`, seed 42, layer-wise offload. The token-refiner is a 2-block, text-only module (well under 1% of a forward), so the saving is at the level of run-to-run timing noise. Wall-clock jitter is one-directional, so the **minimum over repeated runs** best reflects true compute time; across 15 interleaved runs each:

| Mode | metric | Baseline | This PR | Δ |
| layer-wise | min of 15 runs | 46.517 s | 46.418 s | **-0.21% (faster)** |
| model (CPU) offload | — | (byte-identical fallback: same code path) | | 0 |

This is a redundancy-removal / clarity cleanup with a **no-regression** contract: the change does strictly less work in the layer-wise path and takes the unchanged code path under model offload. It is not a meaningful speedup on its own.

## Test Plan

**vLLM Version:** 0.26.0

**vLLM-Omni Commit:** 44e38d7b (rebased)

- New unit test `test_minimax_h3_token_refiner_cache.py`: drives the real denoise loop with a call-counting model; asserts the refiner runs **once** under layer-wise/no-offload (cached embed consumed) and **once per step** under model offload (no hoist, nothing cached).
- Existing H3 CPU suite.
- `pre-commit run` (ruff check / ruff format / typos / pytestmark checks).
- H20 end-to-end A/B: bit-identical media hashes in both offload modes, cross-verified on two machines.

## Test Result

- [x] `test_minimax_h3_token_refiner_cache.py`: 2 passed
- [x] H3 CPU suite: 111 passed, 5 deselected
- [x] `pre-commit run` clean on changed files
- [x] E2E A/B bit-identical (layer-wise + model offload), two-machine cross-verified

### DIFF
--- a/tests/diffusion/models/minimax_h3/test_minimax_h3_token_refiner_cache.py
+++ b/tests/diffusion/models/minimax_h3/test_minimax_h3_token_refiner_cache.py
@@ -1,0 +1,148 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+"""The token-refiner output is request-static, so the denoise loop avoids
+recomputing it every step. How it does so depends on the offload mode:
+
+* layerwise / no offload -- the refiner weights are resident on GPU for the
+  whole request, so the loop precomputes the refined text once above the step
+  loop and every ``forward`` receives it via ``refined_text_embed``.
+* model (cpu) offload -- the model-level hook only onloads the transformer
+  inside ``forward``; hoisting the refiner above the loop would run it while its
+  weights are still on CPU, forcing an extra CPU<->GPU round-trip that a
+  two-layer refiner can't pay back. So the loop does *not* hoist: it passes the
+  raw text inputs and ``forward`` refines internally every step (weights already
+  onloaded, no extra transfer), exactly matching the pre-optimization path and
+  guaranteeing parity rather than a regression.
+"""
+
+from types import SimpleNamespace
+
+import pytest
+import torch
+
+pytestmark = [pytest.mark.core_model, pytest.mark.cpu, pytest.mark.diffusion]
+
+
+class _RefinerCountingModel:
+    """Minimal stand-in for MiniMaxH3DiTModel used by the denoise loop.
+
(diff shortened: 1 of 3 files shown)


hsliuustc0106:
This PR touches **vllm_omni/diffusion/, tests/diffusion/** (3 files). Based on CODEOWNERS coverage of the changed files, the most-related reviewers appear to be:

> @SamitHuang @david6666666 @RuixiangMa

ChoHee15:
This PR seems to be largely covered by #5819 ("precompute static conditioning"), which also caches the RoPE frequencies and handles CPU offload in one path. Is it still worth pursuing?

Issue #5809, opened 2026-08-05
[RFC] Unified pluggable storage backend for video artifacts & metadata


wangrzneu:
## Summary

Introduce a single pluggable `StorageBackend` abstraction covering both video **artifacts** (blob files) and job **metadata** (records), implemented on **Apache OpenDAL** (adopted per @NickCao's suggestion), with `file` (default, unchanged behavior) and `s3` backends.

## Motivation

- **#2531** (merged) added artifact storage (`StorageBaseManager`) — file-only, local fs.
- **#2244** (open, for **#2227**) adds metadata persistence (`VideoMetadataStore`) — records-only.
- **#2361** (open) wants generated videos uploaded to S3-compatible object storage.

Today these are three incompatible interfaces; #2227/#2361 block on each other's review cycles.

## Design (adopted after evaluating OpenDAL 0.47.5 — details in [comment](https://github.com/vllm-project/vllm-omni/issues/5809#issuecomment-5207546141))

**1. Plain layout — `video_id` unchanged**

```
storage root / bucket
├── artifacts/{video_id}        # video payload (large object)
└── meta/{video_id}.json        # full job record {status, created_at, expires_at, ...}
```

**2. `StorageBackend` = thin wrapper over one OpenDAL operator**

```python
class StorageBackend(ABC):
    # lifecycle
    async def start(self) -> None: ...
    async def stop(self) -> None: ...
    # artifact
    async def save(self, key, data: bytes, *, ttl: int | None = None) -> SaveContext
    async def open(self, key) -> StorageHandle | None      # path | presigned URL | stream
    async def delete(self, key) -> bool                    # composite: artifact + meta
    async def iter_chunks(self, key, chunk_size=1 << 20) -> AsyncIterator[bytes]
    # record (raw bytes; serialization lives in the caller)
    async def record_get(self, key) -> bytes | None
    async def record_set(self, key, data: bytes) -> None
    async def record_pop(self, key) -> bytes | None
    async def record_keys(self) -> list[str]               # requires list capability (fs/s3)
```

**3. TTL — `expires_at` is authoritative, dual-path sweep**

`meta/{id}.json` stores `expires_at` (written as `now + ttl` at save time). Sweep strategy probed once at startup:

- fs/s3 → `list.last_modified + file_ttl` (identical semantics to current `st_mtime`-based TTL, zero reads);
- any backend with `list` but no `last_modified` → read `meta/{id}.json` `expires_at`;
[...]

Issue #2361, opened 2026-03-31
[Feature]: Generate the video and then save it to S3 object storage


lengrongfu:
### 🚀 The feature, motivation and pitch

Current video generate after will save to local dir, this is unacceptable in a production environment.

### Alternatives

I have two suggestions:
1. Support uploading to S3-compatible object storage as the default implementation.
2. Add a plugin registration mechanism to allow other users to implement custom storage operations (optional).

Awaiting Feedback.

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://vllm-omni.readthedocs.io), which can answer lots of frequently asked questions.

Issue #2227, opened 2026-03-26
[Feature]: Persist the metadata generated from the video.


lengrongfu:
### 🚀 The feature, motivation and pitch

Current video metadata is save in memory key-value store, this will result in a failure to download the video file after a restart when using the previous video_id.

### Alternatives

The system provides users with an option to utilize the `diskcache` library, enabling the persistence of video metadata to disk so that it remains queryable even after a system restart.

It might also be possible to introduce a distributed key-value store like Redis; however, this would introduce excessive complexity regarding system dependencies. Therefore, I recommend using `diskcache`—or alternatively, if you have other recommendations.

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://vllm-omni.readthedocs.io), which can answer lots of frequently asked questions.

lengrongfu:
@SamitHuang @hsliuustc0106 Do you have any suggestions? thanks ~

hsliuustc0106:
i prefer not to introduce redis

lishunyang12:
> i prefer not to introduce redis

This feature is a bit niche..

lengrongfu:
> > i prefer not to introduce redis
> 
> This feature is a bit niche..

To ensure availability in a production environment, I believe it is necessary to persist the metadata.

lengrongfu:
@ieaves what do you think? because this pr is you commit https://github.com/vllm-project/vllm-omni/pull/1665

ieaves:
`diskcache` is a great tool. It gives us persistence across restarts without introducing `redis` as a hard dependency. The tradeoff is being SQLite-based it's better suited to local single-node storage than shared or distributed deployments, and it can run into locking issues on networked storage.

I would prefer to add this behind a pluggable metadata-store interface rather than couple the implementation directly to diskcache. That keeps the default in-memory behavior, enables a `diskcache` backend for local persistence, and makes it straightforward to add `redis` later if we want a distributed option.

lengrongfu:
https://github.com/vllm-project/vllm-omni/issues/2361 @ieaves Storing videos presents a similar challenge and may also require support for distributed persistence.

ieaves:
S3 seems like the right solution for that as well. I'd be happy to help out with any of these as well if there was interest from the core team.

ywang96:
[...]

PR #5902, opened 2026-08-07
[Refactor] Introduce StorageBackend abstraction for artifact storage

wangrzneu:
## Purpose

Introduce a `StorageBackend` abstraction for artifact storage, the first step of the unified storage design in RFC #5809.

Split `storage.py` into a package (base/local/factory/__init__.py), replacing the generic `StorageBaseManager` with a concrete ABC. The next step implements the abstraction on OpenDAL, which provides S3 and other storage schemes. Zero behavior change: local manager logic kept verbatim, public API (STORAGE_MANAGER, handles, managers, get_storage_manager) unchanged.

## Test Plan

- ruff check / format on the new package
- mypy (strict) on the package files
- pytest tests/entrypoints/openai_api/test_storage_manager.py -> 5 passed

## Test Result

Existing storage tests pass; no behavior change.

Related: #5809, #2227, #2361

### DIFF
--- a/vllm_omni/entrypoints/openai/storage/__init__.py
+++ b/vllm_omni/entrypoints/openai/storage/__init__.py
@@ -1,0 +1,38 @@
+"""Pluggable storage backend for generated artifacts (e.g. video files).
+
+Public surface: the ``StorageBackend`` contract, local filesystem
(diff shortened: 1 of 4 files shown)


hsliuustc0106:
@natureofnature PTAL

PR #5903, opened 2026-08-07
[Model] Reuse frame-local KV in MOSS-TTS-Local-v1.5

YuDeng0102:
## Summary

- replace repeated full-prefix re-prefill in `MossTTSLocalDepthTransformer.generate_frame()` with incremental, within-frame K/V reuse
- preallocate and freeze a fixed-address per-layer KV workspace before runner-owned CUDA graph capture
- preserve MOSS Local's GPT-J/interleaved RoPE and per-frame cache-reset semantics
- gate the fast path with the checkpoint-aligned `use_static_local_kv_cache` field
- retain the growing-prefix path as a compatibility/debug path and test oracle
- add focused eager, compiled, sampling, and CUDA graph regression tests

Related to #4676.

## Configuration and compatibility

`MossTTSLocalConfig` declares:

```python
use_static_local_kv_cache: bool = False
```

Configs that omit the field keep vLLM's existing growing-prefix behavior. The official
`OpenMOSS-Team/MOSS-TTS-Local-Transformer-v1.5` checkpoint explicitly stores
`"use_static_local_kv_cache": true`, so it automatically opts into the fixed-address incremental path.

HF uses dynamic `past_key_values` when its static-cache flag is disabled, while vLLM retains its existing
growing-prefix fallback. The field name and static-cache opt-in intent are aligned.

Only the enabled branch calculates cache capacity, allocates/freezes the workspace, and compiles
`_forward_token_static`. The disabled branch allocates no local KV workspace and compiles `_forward_prefix`.
The runner remains the sole CUDA graph owner; this module does not add a nested graph manager.

## Correctness checks

The focused tests cover:

- config fallback and explicit opt-in
- incremental hidden states/logits vs. full prefix
- greedy and fixed-seed sampled stop/codes parity
- no cross-frame or cross-batch cache leakage
- disabled mode remains allocation-free
- fixed cache capacity/address and overflow handling
- compile target selection for both flag values
- CUDA graph capture/replay of the compiled fixed-address path

Current local verification:

```text
tests/model_executor/models/moss_tts: 53 passed
ruff check: passed
ruff format --check: passed
git diff --check: passed
```

The cached implementation was also checked with the real v1.5 checkpoint in BF16 at batches 1 and 2:

- eager cached vs. eager full-prefix: continue/stop and greedy codes exact
- compiled cached vs. compiled full-prefix: continue/stop and greedy codes exact
- hidden states/logits close at `rtol=0.02, atol=0.04`
- max absolute error: hidden `0.03125`, logits `0.25`

## Warmed end-to-end A/B

Hardware: 1x A100 PCIe 40 GB. Baseline: `44e38d7b`. Cached implementation: `221f4874` before the
history-only rebase/squash.

Each version ran in an independently started server with the same settings. The workload repeats one fixed
prompt/reference pair to avoid cold reference-audio encoding dominating the run. Each server first received an
uncounted `C=8, N=8` prewarm; formal runs used two client warmups, `hf_output_len=80`, and `request-rate=inf`.
All 40 formal requests succeeded with no OOM.

Change is `cached / baseline - 1`; negative is faster for latency metrics. Audio-duration change describes
generated output length, not performance.

| Concurrency / requests | Metric | Baseline | Cached | Change |
|---:|:---|---:|---:|---:|
| 1 / 8 | Mean RTF | 0.2134 | 0.2060 | -3.5% |
| 1 / 8 | Mean TTFP | 91.1 ms | 92.3 ms | +1.3% |
| 1 / 8 | Mean E2EL | 1206.5 ms | 1102.6 ms | -8.6% |
| 1 / 8 | Mean audio duration | 5.660 s | 5.360 s | -5.3% |
| 8 / 32 | Mean RTF | 0.3450 | 0.3291 | -4.6% |
| 8 / 32 | Mean TTFP | 169.2 ms | 170.8 ms | +0.9% |
| 8 / 32 | Mean E2EL | 1781.7 ms | 1773.1 ms | -0.5% |
| 8 / 32 | Mean audio duration | 5.165 s | 5.413 s | +4.8% |

This is a single warmed short run, not a claim of a statistically stable service-level speedup. Mean RTF,
which normalizes latency by generated audio duration, improved by 3.5% at C1 and 4.6% at C8. Mean TTFP was
approximately unchanged (+1.3% and +0.9%), as expected for an optimization of within-frame incremental work.

MOSS Local's internal codebook sampler retains checkpoint sampling semantics, so final audio duration still
varies between requests even with the same prompt. E2EL and throughput therefore cannot be interpreted without
the corresponding output duration; in particular, the C1 E2EL change partly reflects its shorter generated
audio. All 8/8 C1 and 32/32 C8 requests completed successfully in both runs.

## Status

- [x] real-checkpoint BF16 eager/compiled parity
- [x] fixed-address CUDA graph replay coverage
- [x] checkpoint-aligned configuration gate
- [x] warmed C1/C8 end-to-end A/B
- [x] rebased onto current upstream main and locally verified


hsliuustc0106:
This PR touches **vllm_omni/model_executor/, tests/model_executor/** (3 files). Based on CODEOWNERS coverage of the changed files, the most-related reviewers appear to be:

> @NickCao @Bounty-hunter @amy-why-3459

YuDeng0102:
Hi @NickCao @Bounty-hunter @amy-why-3459, just a gentle follow-up on this PR. It implements the frame-local KV-cache item from #4676, includes eager/compiled correctness coverage and warmed end-to-end results, and all CI checks are passing. I would appreciate a review when you have time. Thanks!

PR #5905, opened 2026-08-07
[Bugfix][Frontend] Preserve generated audio in ComfyUI video output

Devindan:
## Summary

Fix audio handling in the `ComfyUI-vLLM-Omni` video generation workflow.

This PR addresses two related issues:

1. Exposes the `/v1/videos` `generate_sound` option in the ComfyUI `Generate Video` node and forwards it to the vLLM-Omni API.
2. Fixes audio tracks being dropped when converting generated MP4 responses into ComfyUI `VIDEO` objects.

## Problem

Video models such as MiniMax-H3 can generate native audio together with video when `generate_sound=true`.

However, the current ComfyUI integration has two problems:

- `Generate Video` does not expose or forward the `generate_sound` request parameter.
- Even when the vLLM-Omni server returns an MP4 containing a valid audio stream, `bytes_to_video()` may drop the audio during decoding.

The second issue is caused by decoding the streams sequentially from the same PyAV container:

```python
container.decode(video_stream)
````

followed by:

```python
container.decode(audio_stream)
```

After the video stream has been fully decoded, the container has already been consumed to EOF, so the subsequent audio decode may return no frames.

As a result, the generated ComfyUI `VIDEO` object contains the video frames but has no audio.

### Add `generate_sound` support

Add a `generate_sound` boolean option to the `Generate Video` node and forward it through `VLLMOmniClient.generate_video()` as a top-level multipart field for `/v1/videos`.

The default remains `false`, matching the API's existing behavior.

### Preserve audio during video decoding

Update `bytes_to_video()` to decode the selected video and audio streams together in a single pass:

```python
for frame in container.decode(*streams):
    ...
```

Video and audio frames are then handled according to their frame type.

This avoids consuming the container before the audio stream is decoded.

## Testing

Tested end-to-end with:

* MiniMax-H3 FL2VA
* vLLM-Omni video serving API
* ComfyUI-vLLM-Omni
* Python 3.12
* 2x NVIDIA H800 80GB

Verified the following cases:

* `generate_sound=true` produces a ComfyUI `VIDEO` object with working generated audio.
* The MP4 returned directly by the vLLM-Omni server contains both video and audio streams.
* Audio is preserved after conversion through `bytes_to_video()`.
* `generate_sound=false` continues to generate video without audio normally.
* Video-only responses continue to decode correctly.
* Ruff and the relevant pre-commit checks pass for the modified Python files.

## Scope

The decoding fix is not specific to MiniMax-H3. It should apply to any video generation backend that returns an MP4 containing both video and audio streams through the vLLM-Omni video API.

### DIFF
--- a/apps/ComfyUI-vLLM-Omni/comfyui_vllm_omni/nodes.py
+++ b/apps/ComfyUI-vLLM-Omni/comfyui_vllm_omni/nodes.py
@@ -163,6 +163,7 @@ class VLLMOmniGenerateVideo(_VLLMOmniGenerateBase):
                 "height": ("INT", {"default": 480, "min": 1}),
                 "fps": ("INT", {"default": 16, "min": 1}),
                 "num_frames": ("INT", {"default": 41, "min": 1}),
+                "generate_sound": ("BOOLEAN", {"default": False}),
             },
             "optional": {
                 "frame": ("IMAGE",),
@@ -195,6 +196,7 @@ class VLLMOmniGenerateVideo(_VLLMOmniGenerateBase):
         height: int,
         fps: int,
         num_frames: int,
+        generate_sound: bool,
         negative_prompt: str | None = None,
         frame: torch.Tensor | None = None,
         references: dict | None = None,
@@ -234,6 +236,7 @@ class VLLMOmniGenerateVideo(_VLLMOmniGenerateBase):
             height=height,
             num_frames=num_frames,
             fps=fps,
+            generate_sound=generate_sound,
             negative_prompt=negative_prompt,
             sampling_params=sampling_params,
             lora=lora,
--- a/apps/ComfyUI-vLLM-Omni/comfyui_vllm_omni/utils/api_client.py
+++ b/apps/ComfyUI-vLLM-Omni/comfyui_vllm_omni/utils/api_client.py
@@ -265,6 +265,7 @@ class VLLMOmniClient:
         height: int,
         num_frames: int,
         fps: int,
+        generate_sound: bool = False,
         negative_prompt: str | None = None,
         frame: torch.Tensor | None = None,
         references: dict | None = None,
@@ -284,6 +285,10 @@ class VLLMOmniClient:
         form.add_field("height", str(height))
         form.add_field("num_frames", str(num_frames))
         form.add_field("fps", str(fps))
+        form.add_field(
+            "generate_sound",
+            "true" if generate_sound else "false",
+        )
         if negative_prompt:
             form.add_field("negative_prompt", negative_prompt)
         if sampling_params is not None:
--- a/apps/ComfyUI-vLLM-Omni/comfyui_vllm_omni/utils/format.py
+++ b/apps/ComfyUI-vLLM-Omni/comfyui_vllm_omni/utils/format.py
@@ -152,12 +152,29 @@ def bytes_to_video(video_bytes: bytes) -> VideoInput:
             if video_stream is None:
                 raise ValueError("No video stream found in decoded payload.")
 
+            audio_stream = container.streams.audio[-1] if len(container.streams.audio) else None
+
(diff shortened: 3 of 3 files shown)


hsliuustc0106:
@fhfuih PTAL

fhfuih (commented):
The implementation looks good to me. But I want to double check with your claim 

> "Video models such as MiniMax-H3 can generate native audio together with video when generate_sound=true."

I cannot find where minimax h3 pipeline recognize this param (that's probably why I didn't pay attention to this param. ) When I tried H3 with ConfyUI, it always generates with sound even without this param.

Now that seeing this PR, I do notice Cosmos recognize this param though

And as that being said, I didn't yet encounter the 2nd problem you mentioned: _"`bytes_to_video()` may drop the audio during decoding."_. 

So could you help test 2x2=4 cases of generation using (H3, Cosmos) model and with this toggle (on or off), and attach screen recordings---or other test cases that demonstrate both the `generate_sound` problem and the audio-dropping problem before and after the fix

Devindan:
@fhfuih 

Thanks for the careful review. You're right about MiniMax-H3 — I re-checked the pipeline and my original wording was inaccurate.

MiniMax-H3 currently generates video and audio jointly and does not appear to use `generate_sound` as a switch. So for H3, I would expect audio to be generated regardless of whether the toggle is on or off.

Cosmos is different: its pipeline does recognize the sound-generation flag, so the expected behavior should be:

| Model      | `generate_sound=false` | `generate_sound=true` |
| ---------- | ---------------------- | --------------------- |
| MiniMax-H3 | audio                  | audio                 |
| Cosmos     | no audio               | audio                 |

I'll update the PR description to avoid implying that `generate_sound` controls audio generation for H3 specifically. The reason for exposing this option in the ComfyUI node is that it is part of the generic `/v1/videos` API and is meaningful for models such as Cosmos.

For the second issue, the behavior I observed was reproducible locally with H3: the MP4 returned by the vLLM-Omni server contained a valid audio stream, but the ComfyUI `VIDEO` object produced by the original `bytes_to_video()` implementation had no audio. After changing the decoder to consume the video and audio streams together in a single pass, the audio was preserved correctly.

The key difference is:

```python
# before
container.decode(video_stream)
container.decode(audio_stream)

# after
container.decode(video_stream, audio_stream)
```

My understanding is that after the first decode loop consumes the container to EOF, the second decode may not yield the interleaved audio packets anymore. This explains why the problem may not reproduce for every media layout, but it did reproduce consistently with the H3-generated MP4s in my environment.

I won't be able to provide screen recordings, but I can include the exact configurations and whether an audio stream was present in each generated result.

fhfuih (commented):
Thanks for the double check. I think it now it makes sense. One last thing: AFAIK if you change a node implementation in ComfyUI, when you import a workflow that references to that node, it still shows the old set of toggles. So could you also

- copy/rsync your updated integration source code to your local comfyui installation
- import the example custom workflows that uses video generation
- delete the main video generation node, re-grag it from the node menu, and set all io connections & fields again like before
- same the current state of the workflow
- find the saved workflow JSON in comfyui directory and copy (overwrite) it here, and commit that to this PR


And if you cannot offer screen recording, but still at least have the final video from the comfyui output node, could you upload them here just for everyone's record? (the no sound old video and the new video). I think you can find a download button from the output video preview control, which should be what's outputted by the decoder

apps/ComfyUI-vLLM-Omni/comfyui_vllm_omni/nodes.py
```diff
@@ -195,6 +196,7 @@ async def generate(
         height: int,
         fps: int,
         num_frames: int,
+        generate_sound: bool,
```
lishunyang12, line 199: Update the video tests to pass generate_sound; all six existing cases now raise TypeError.

PR #5911, opened 2026-08-08
[Diffusion] Enable independent requests for rank-local DLO DP

lishunyang12:
## Summary

Follow-up to #5864. This PR enables independent request dispatch when distributed layerwise offload runs with `data_parallel_size > 1` and `dlo_use_allgather=False`.

- dispatch one request assignment per DP replica
- pad partial waves with an explicit idle assignment instead of duplicating work
- allow heterogeneous prompts, shapes, CFG values, denoise steps, and `extra_args` in rank-local mode
- keep the existing compatibility checks for DLO AllGather waves
- scope request-local DiT and KV-transfer collectives to one DP replica

This PR does **not** claim the existing DLO GPU-memory reduction as a new benefit. mmap sharing, SP-local weight AllGather, and component-selective offload remain separate follow-ups.

## Problem

The scheduler previously enabled DP request concurrency only for DLO AllGather. Simply enabling it for rank-local DLO was not enough: request-local collectives still used a process group containing every DP rank.

For a partial DP wave, rank 0 executed request A while rank 1 was idle. Rank 0 then blocked in:

```text
_prepare_request_for_forward
  -> receive_multi_kv_cache_distributed
  -> _broadcast_kv_payload(world)
```

because `world` still included the idle DP rank.

## Topology

```mermaid
flowchart LR
    A[request A] --> R0[DP replica 0]
    B[request B or idle] --> R1[DP replica 1]

    subgraph Request-local groups
      R0 --> D0[DiT / KV group 0]
      R1 --> D1[DiT / KV group 1]
    end

    subgraph DLO weight topology
      W0[DP rank 0 weight shard] <-->|AllGather mode only| W1[DP rank 1 weight shard]
    end

    D0 -. no request payload collective .- D1
```

For `DP=2, SP=4`, request-local groups are `[0,1,2,3]` and `[4,5,6,7]`. Orthogonal DP peer groups remain `[0,4]`, `[1,5]`, `[2,6]`, and `[3,7]` for DLO weight AllGather.

## Behavior

- `dlo_use_allgather=True`: every DP rank remains in a compatible collective wave; existing shape/CFG/step/`extra_args` checks stay enabled.
- `dlo_use_allgather=False`: each DP replica owns complete rank-local CPU weights and may execute a different request. Missing assignments are `None`, so an idle replica returns without running the model.
- `get_dit_group()` still exposes the underlying device `ProcessGroup` for existing VAE and model consumers; KV transfer uses the new coordinator to access the replica-local CPU/device groups.

## Tests

CPU regression suite:

```text
172 passed
ruff check: passed
ruff format --check: 11 files already formatted
```

2x B300 MiniMax-H3 FL2VA rank-local DLO E2E (`DP=2`, `TP=1`, `dlo_use_allgather=False`, 768x1344, 124 frames):

| Case | Before | After |
| --- | --- | --- |
| Partial wave (one request, one idle replica) | hangs in global KV broadcast | completes in 15.399s |
| Heterogeneous pair (2 vs 3 denoise steps, different prompts) | unsupported | both outputs complete in 25.853s |
| Output validation | N/A | video `[124,768,1344,3]`, audio `[1,2,165600]` for all 7 requests |
| Shutdown | N/A | `active_children=[]` |

Serial heterogeneous pair: 26.272s. Concurrent heterogeneous pair: 25.853s (`1.016x` on this host). The primary benefit here is correctness and independent scheduling; rank-local DLO streams contend for host-to-device bandwidth, so this test does not claim linear throughput scaling.

2x B300 DLO AllGather regression (`dlo_use_allgather=True`):

- incompatible asymmetric wave rejected for both requests in 6.3ms
- compatible pair completed in 31.808s
- both video/audio outputs validated
- clean shutdown with no active child processes

### DIFF
--- a/docs/design/module/diffusion/offloader.md
+++ b/docs/design/module/diffusion/offloader.md
@@ -35,8 +35,8 @@ Distributed layerwise offload has two scheduling contracts:
   weight collective in the same order.
 - **Rank-local mode** streams the tensors produced by the standard loader on
   each rank. DP replicas may execute heterogeneous requests independently, idle
-  replicas skip model execution, and request-scoped process groups keep runtime
-  collectives local to the assigned replica.
+  replicas skip model execution, and replica-local DiT groups keep runtime
+  collectives inside the assigned replica.
 
 These modes trade host-memory efficiency for scheduling independence. Changing
 the mode MUST update admission, dispatch, result routing, abort handling, and
--- a/docs/design/module/diffusion/offloader.md
+++ b/docs/design/module/diffusion/offloader.md
@@ -18,13 +18,29 @@ validation_paths:
   - tests/diffusion/offloader/**
 upstream_refs:
   - torch.nn.Module.to
-last_reviewed: 2026-07-16
+last_reviewed: 2026-08-09
 ---
 
 # Diffusion offloader
 
 The offloader owns component residency, transfer scheduling, memory accounting,
 prefetch, and teardown for diffusion execution.
+
+## Distributed layerwise execution modes
+
+Distributed layerwise offload has two scheduling contracts:
+
+- **AllGather mode** shards each layer across the participating ranks. Requests
+  execute in compatible collective waves because every rank must enter the same
+  weight collective in the same order.
+- **Rank-local mode** streams the tensors produced by the standard loader on
+  each rank. DP replicas may execute heterogeneous requests independently, idle
+  replicas skip model execution, and request-scoped process groups keep runtime
+  collectives local to the assigned replica.
+
+These modes trade host-memory efficiency for scheduling independence. Changing
+the mode MUST update admission, dispatch, result routing, abort handling, and
+teardown as one contract.
 
 ## Candidate invariants
 
--- a/vllm_omni/diffusion/diffusion_engine.py
+++ b/vllm_omni/diffusion/diffusion_engine.py
@@ -254,7 +254,12 @@ class DiffusionEngine:
         # dispatch heterogeneous requests independently. Ordinary DP with a
         # non-batch pipeline still does not schedule multiple requests.
         dp_size = getattr(getattr(self.od_config, "parallel_config", None), "data_parallel_size", 1)
-        if _uses_dlo_dp_concurrency(self.od_config):
+        uses_dlo_dp_concurrency = _uses_dlo_dp_concurrency(self.od_config)
+        if uses_dlo_dp_concurrency and self.step_execution:
+            self.scheduler.max_num_running_reqs = 1
+            self.dp_concurrent = False
+            logger.info("DLO DP concurrency is disabled for step execution until step dispatch is DP-aware.")
+        elif uses_dlo_dp_concurrency:
             self.scheduler.max_num_running_reqs = dp_size
             self.dp_concurrent = True
             logger.info(
--- a/vllm_omni/diffusion/diffusion_engine.py
+++ b/vllm_omni/diffusion/diffusion_engine.py
@@ -119,11 +119,7 @@ def _max_num_seqs(od_config: OmniDiffusionConfig) -> int:
 def _uses_dlo_dp_concurrency(od_config: OmniDiffusionConfig) -> bool:
     parallel_config = getattr(od_config, "parallel_config", None)
     dp_size = getattr(parallel_config, "data_parallel_size", 1)
-    return (
-        dp_size > 1
-        and getattr(od_config, "enable_distributed_layerwise_offload", False)
-        and getattr(od_config, "dlo_use_allgather", True)
-    )
+    return dp_size > 1 and getattr(od_config, "enable_distributed_layerwise_offload", False)
 
 
 def _move_tensor_tree_to_cpu(value: object) -> object:
@@ -253,10 +249,10 @@ class DiffusionEngine:
 
     def _init_runtime_state(self) -> None:
         # DP multi-concurrency: allow batching dp_size requests so each
-        # worker processes a different request in parallel.  Only enabled
-        # for distributed layerwise offload (which shards weights and
-        # needs all ranks active simultaneously).  Ordinary DP with a
-        # non-batch pipeline should not schedule multiple requests.
+        # DP replica processes one request in parallel. AllGather-backed DLO
+        # keeps replicas in a compatible collective wave; rank-local DLO can
+        # dispatch heterogeneous requests independently. Ordinary DP with a
+        # non-batch pipeline still does not schedule multiple requests.
         dp_size = getattr(getattr(self.od_config, "parallel_config", None), "data_parallel_size", 1)
         if _uses_dlo_dp_concurrency(self.od_config):
             self.scheduler.max_num_running_reqs = dp_size
--- a/vllm_omni/diffusion/distributed/parallel_state.py
+++ b/vllm_omni/diffusion/distributed/parallel_state.py
@@ -547,14 +547,24 @@ def init_vllm_model_parallel_group(
 
 
 def init_dit_group(
-    dit_parallel_size: int,
+    group_ranks: list[list[int]],
+    local_rank: int,
     backend: str,
 ):
     global _DIT
-    _DIT = torch.distributed.new_group(ranks=list(range(dit_parallel_size)), backend=backend)
+    assert _DIT is None, "DIT group is already initialized"
+    _DIT = GroupCoordinator(
+        group_ranks=group_ranks,
+        local_rank=local_rank,
+        torch_distributed_backend=backend,
+    )
 
 
 def get_dit_group():
+    return get_dit_group_coordinator().device_group
+
+
+def get_dit_group_coordinator() -> GroupCoordinator:
     assert _DIT is not None, "DIT group is not initialized"
     return _DIT
 
@@ -969,12 +979,20 @@ def initialize_model_parallel(
             use_all2all=True,
         )
 
-    init_dit_group(dit_parallel_size, backend)
+    # A DiT group contains the model-parallel ranks for one DP replica. DP
+    # replicas must not share request-local collectives (for example condition
+    # broadcasts or VAE patch parallelism), because they may execute different
+    # requests. DLO weight sharing uses the orthogonal DP groups above.
+    init_dit_group(
+        rank_generator.get_ranks("tp-sp-pp-cfg"),
+        get_world_group().local_rank,
+        backend,
+    )
 
 
 def destroy_model_parallel():
     """Set the groups to none and destroy them."""
-    global _DP, _CFG, _SP, _PP, _FS, _EXPERT_PARALLEL_GROUP_RANKS
+    global _DP, _CFG, _SP, _PP, _FS, _DIT, _EXPERT_PARALLEL_GROUP_RANKS
 
     if vllm_parallel_state._DP and vllm_parallel_state._DP is not _DP:
         vllm_parallel_state._DP.destroy()
@@ -1012,6 +1030,10 @@ def destroy_model_parallel():
     if _FS:
         _FS.destroy()
     _FS = None
+
+    if _DIT:
+        _DIT.destroy()
+    _DIT = None
 
 
 def destroy_distributed_environment():
--- a/vllm_omni/diffusion/executor/multiproc_executor.py
+++ b/vllm_omni/diffusion/executor/multiproc_executor.py
@@ -461,24 +461,30 @@ class MultiprocDiffusionExecutor(DiffusionExecutor):
 
         has_diffusion_kv_metadata = any(new_req.diffusion_kv_metadata is not None for new_req in new_reqs)
 
-        # DP multi-concurrency: when DLO+AllGather is active and multiple
-        # requests are scheduled, send ALL requests in one broadcast RPC.
-        # Each rank picks req[rank % len(reqs)] and computes independently.
-        # All ranks reply (unique_reply_rank=None) so we collect dp_size
-        # responses and match by dp_rank.
-        if (
-            len(new_reqs) > 1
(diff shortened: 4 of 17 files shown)


hsliuustc0106:
update the feature design docs

hsliuustc0106:
Could reviewers confirm that this animation matches the intended rank-local DLO DP behavior?

DLO rank-local dispatch

The animation shows:

1. Before #5911: with `dlo_use_allgather=False`, DP=2, and one request, the active rank can enter a global request-local KV/DiT collective that includes the idle rank and hang.
2. After #5911: the scheduler dispatches `[Request A, None]`, each DP replica uses a replica-local request group, and the idle replica skips model execution.
3. Heterogeneous requests can run independently in rank-local mode.

My understanding is that this is primarily a correctness and scheduling fix; it does not add a new memory-reduction claim. The request-local groups are separate from the orthogonal DLO weight AllGather groups used by the AllGather path.

vllm_omni/diffusion/diffusion_engine.py
```diff
@@ -119,11 +119,7 @@ def _max_num_seqs(od_config: OmniDiffusionConfig) -> int:
 def _uses_dlo_dp_concurrency(od_config: OmniDiffusionConfig) -> bool:
     parallel_config = getattr(od_config, "parallel_config", None)
     dp_size = getattr(parallel_config, "data_parallel_size", 1)
-    return (
-        dp_size > 1
-        and getattr(od_config, "enable_distributed_layerwise_offload", False)
-        and getattr(od_config, "dlo_use_allgather", True)
-    )
+    return dp_size > 1 and getattr(od_config, "enable_distributed_layerwise_offload", False)
```
hsliuustc0106, line 122: [P1] Rank-local DLO still cannot admit heterogeneous requests. This enables DP concurrency for rank-local DLO, but RequestScheduler still uses BaseScheduler._can_schedule_waiting(), which requires the next request's sampling-parameter key to equal the running request's key. I reproduced this in the PR environment with max_num_running_reqs=2: two requests with 2 and 3 inference steps produced SCHEDULED ['a'] and WAITING ['b']. The real two-GPU MiniMax-H3 run likewise logged max_num_running_reqs=2 but serialized the 2-step/3-step pair. Please add a rank-local admission policy that skips this compatibility check only for dlo_use_allgather=False, and add an engine-level regression test that submits heterogeneous requests through RequestScheduler.schedule(); AllGather should retain compatible-wave admission.

hsliuustc0106:
Local verification against PR head `9e73ee1a50ce247c638052011914d8027d717f28`:

- Environment: `/home/hsliu/tmp/venvs/vllm-omni-pr-5864-vllm-0.26.0`, with `PYTHONPATH` forced to the PR worktree; 2x NVIDIA L20X used.
- Targeted CPU/DLO/KV/scheduler suite: **271 passed**, 18 warnings.
- MiniMax-H3 rank-local DLO (`dlo_use_allgather=False`) initialized and ran on two GPUs. With valid settings (`duration=4`, `steps=2`, `512x512`), all seven generation requests completed and the output-shape checks passed for the validated waves.
- The lifecycle example exited 1 only at its final cleanup assertion because the worker processes remained alive past the shutdown wait; it also reported one leaked shared-memory object. The workers were subsequently terminated and no GPU/process residue remained.
- The scheduler reproduction and the GPU logs both show the heterogeneous 2-step/3-step pair being serialized despite `max_num_running_reqs=2`; this is the blocking issue called out inline.

The CI gates currently pass: DCO, pre-commit, and both build jobs.

vllm_omni/diffusion/diffusion_engine.py
lishunyang12: Fixed in 7e4ba8e3. RequestScheduler now bypasses sampling-compatibility admission only for DP greater than 1 rank-local DLO with dlo_use_allgather=False. Added an engine-level regression with heterogeneous 2-step and 3-step requests: rank-local admits both, while AllGather admits only the compatible wave. Relevant engine/scheduler/concurrency/RPC suites pass (158 tests).

lishunyang12:
Addressed the design-doc request in 419dd6e5. The diffusion offloader design now distinguishes compatible AllGather collective waves from heterogeneous rank-local execution, and the continuous-batching design defines the mode-specific admission contract. The P1 scheduler fix is in 7e4ba8e3 with 158 relevant tests passing. Please re-review when convenient.

docs/design/module/diffusion/offloader.md
```diff
@@ -17,14 +17,30 @@ validation_paths:
   - tests/diffusion/offloader/**
```
hsliuustc0106, line 18: update the feature design and user guidance as well

lishunyang12: Addressed in 7f1e35b2. The feature design now specifies heterogeneous rank-local admission, partial-wave idle assignments, DP-rank result routing, and replica-local DiT/KV collective boundaries. The user guide adds the AllGather versus no-AllGather request matrix, automatic scheduling behavior, shared-default denoise-step rule, host-memory tradeoff, and deployment recommendations. I also corrected the module wording from request-scoped to replica-local process groups. Documentation pre-commit checks pass.

hsliuustc0106 (changes requested):
Re-review at `7f1e35b2452d1bbc5bc5814b5dad8028c167e62d` against base `ebe93f5d1e794202598e652a45fa74a29a7f698b`.

Findings:

1. **[P1] Existing async-prefetch tests fail after the topology API change.** `vllm_omni/distributed/omni_connectors/kv_transfer_manager.py:596` now uses `get_dit_group_coordinator()`, but `tests/distributed/omni_connectors/test_kv_async_prefetch.py:245` and `:322` still patch `get_world_group()`. On the PR head, that suite is **5 failed, 19 passed**; on the base it is **24 passed**. Update those fixtures to patch the new coordinator API, including the uninitialized-group test.

2. **[P1] Rank-local DLO DP concurrency is enabled for step/streaming execution without DP-aware step dispatch.** `vllm_omni/diffusion/diffusion_engine.py:122` enables `dp_concurrent` for rank-local DLO even when `step_execution=True` (also reached from `streaming_output=True`). The scheduler admits multiple requests, but `vllm_omni/diffusion/executor/multiproc_executor.py:640-645` broadcasts one shared `execute_stepwise` batch to every replica; no per-DP assignment vector is used. Each replica duplicates all requests while only rank 0's results are collected. Either implement DP-aware step dispatch/result routing or disable and document rank-local DP concurrency for step/streaming mode.

The previous scheduler-admission finding is fixed. The focused regression suite passes (271 tests), and a fresh heterogeneous rank-local GPU pair completed on the PR head.

hsliuustc0106:
resolve conflicts

lishunyang12:
Addressed the requested changes in 7f927e1e0:

- Async prefetch topology tests now patch get_dit_group_coordinator, matching the implementation.
- Rank-local DLO DP concurrency remains enabled for request-batch execution, but step execution, including streaming output, is forced to one running request until dispatch/result routing is DP-aware.
- Added a regression and documented the limitation.

Validation: 121 passed; ruff and diff checks passed.

@hsliuustc0106 could you please double-check the update?

vllm_omni/diffusion/distributed/parallel_state.py
```diff
@@ -969,12 +979,20 @@ def initialize_model_parallel(
...
+        rank_generator.get_ranks("tp-sp-pp-cfg"),
+        get_world_group().local_rank,
+        backend,
+    )
 
 
 def destroy_model_parallel():
     """Set the groups to none and destroy them."""
-    global _DP, _CFG, _SP, _PP, _FS, _EXPERT_PARALLEL_GROUP_RANKS
+    global _DP, _CFG, _SP, _PP, _FS, _DIT, _EXPERT_PARALLEL_GROUP_RANKS
```
hsliuustc0106, line 995: @Bounty-hunter PTAL, do we need to introduce the DIT paralell group?

vllm_omni/diffusion/executor/multiproc_executor.py
```diff
@@ -488,20 +494,24 @@ def execute_request(self, scheduler_output: DiffusionSchedulerOutput) -> BaseRun
...
                 raise ValueError(
                     "DP multi-concurrency requires a non-empty prompt for every request; "
                     f"empty prompt request IDs: {empty_prompt_ids}."
                 )
 
             reqs_list = [nr.req for nr in new_reqs]
+            if not dlo_use_allgather:
+                if len(reqs_list) > dp_size:
+                    raise RuntimeError(f"Scheduled {len(reqs_list)} requests for {dp_size} DP replicas.")
+                reqs_list.extend([None] * (dp_size - len(reqs_list)))
```
hsliuustc0106, line 514: [P1] Route the prefetch job with the replica assignment

With DP=2, async prefetch, and A/B/C queued, `BaseScheduler` creates one prefetch job for C, but this rank-local RPC forwards the same job to every replica. Since `get_dit_group_coordinator()` makes each replica's primary a non-follower, multiple owners race the consumptive pull; the wrong replica can take C's KV before the replica that will run C. Pass a per-replica prefetch assignment (or disable async prefetch for DP-replica dispatch) and cover this with a DP=2 one-shot-connector test.

vllm_omni/diffusion/executor/multiproc_executor.py
```diff
@@ -488,20 +494,24 @@ def execute_request(self, scheduler_output: DiffusionSchedulerOutput) -> BaseRun
...
             empty_prompt_ids = [nr.request_id for nr in new_reqs if _is_empty_dp_prompt(nr.req.prompt)]
-            if empty_prompt_ids:
+            if dlo_use_allgather and empty_prompt_ids:
                 raise ValueError(
                     "DP multi-concurrency requires a non-empty prompt for every request; "
                     f"empty prompt request IDs: {empty_prompt_ids}."
                 )
 
             reqs_list = [nr.req for nr in new_reqs]
+            if not dlo_use_allgather:
```
hsliuustc0106, line 511: [P1] Keep rank-local failures request-scoped

This branch treats requests as independent, but `collective_rpc` raises on any tagged replica error and the catch below assigns that error to every scheduled request. With A succeeding on DP0 and B failing on DP1, A is incorrectly returned as failed. Preserve ordered per-`dp_rank` result/error outcomes and map only B's error; keep wave-wide failure for AllGather or transport failures.

vllm_omni/diffusion/worker/diffusion_worker.py
```diff
@@ -463,8 +463,17 @@ def execute_model(
...
+                raise ValueError("DP request assignments must not be empty.")
+            if getattr(od_config, "dlo_use_allgather", True):
+                idx = dp_rank % len(req)
+            else:
+                if dp_rank >= len(req):
+                    raise RuntimeError(f"Missing request assignment for DP rank {dp_rank}; got {len(req)} assignments.")
+                idx = dp_rank
             req = req[idx]
+            if req is None:
+                return {"dp_rank": dp_rank, "output": None}
```
hsliuustc0106, line 476: [P1] Aggregate non-primary failures into the replica reply

The new rank-local path returns one reply per DP primary, while `_execute_rpc` disables rank-status collection for this mode and only the primary may report errors. In TP/SP/CFG/PP replicas, a non-primary can fail while the primary returns success, so the executor reports a false success. Gather rank statuses within each replica before emitting the tagged primary reply, and add a non-primary failure regression.

hsliuustc0106 (commented):
Static review at `7f927e1e0f69414053eda56daf0a80b05d70cc55`; three blocking findings are inline. I did not execute the fork head locally.

vllm_omni/diffusion/distributed/parallel_state.py
Bounty-hunter: yes, we need it. A typical scenario is that VAE parallelism occurs across `tp-sp-pp-cfg`, similar to EP

hsliuustc0106:
Could we step back and agree on the ownership boundary before adding this path? With `dlo_use_allgather=False`, the offloader forces its effective DLO DP size to 1 and every replica streams a complete standard-loader weight set. The new per-rank request/idle dispatch therefore makes one diffusion engine orchestrate several independent layerwise-offloaded replicas. I think horizontal request routing should remain in the deployment layer (vLLM Router/Dynamo); vLLM-Omni should own only ranks with an execution or weight-collective dependency.

The more useful no-AllGather direction is node-local shared host weights. Independently routed replicas on the same host could map one read-only checkpoint/weight store and retain only private layer-sized pinned staging buffers. That preserves independent schedules and removes DP AllGather without keeping N full host copies. This is orthogonal to request routing, and the current code already leaves rank-local mmap as a TODO.

Unless there is a capability that cannot be provided by Router/Dynamo plus separately launched replicas, I suggest rejecting `DP > 1 && !dlo_use_allgather` for now and re-scoping this work toward a shared rank-local weight backend. Can we settle that design before merging the scheduler/RPC/failure semantics introduced here?

lishunyang12:
Status update (2026-08-23): moving this back to draft. The current blocker is the ownership boundary raised in review, plus three unresolved P1 threads: replica-aware prefetch assignment, request-scoped rank-local failures, and aggregation of non-primary failures. Next step is to agree whether horizontal routing belongs in Router/Dynamo; then either rescope toward a shared host-weight backend or address the RPC semantics in this PR.

hsliuustc0106:
I think this PR could be closed since HWR has supported this while more tests and validation will be carried out

nannaer:
The blog says DLO uses weight sharding + AllGather to avoid the N × model_size host-memory overhead across ranks. If I enable TP, each TP rank already owns only a shard of the model weights. Does TP already solve this host-memory issue? If so, when would DLO still be preferable to simply using TP + regular layerwise offload?

PR #5917, opened 2026-08-08
[Bugfix][StepAudio2] Fix async-chunk metadata mismatch and boundary glitches

### LINKED ISSUE #5916: [Bug]: Periodic audio glitches at chunk boundaries with Step-Audio-2-mini
### Your current environment

<details>

```text
Your output of `python collect_env.py` here
```

</details>

### Your code version

<details>

```text
0.26.0
```
</details>
<details>

```text
81b48e83ecb228712890d48658b212c90441d7d7
```
</details>

### 🐛 Describe the bug

When serving `stepfun-ai/Step-Audio-2-mini` with `async_chunk: true`:

```bash
vllm serve stepfun-ai/Step-Audio-2-mini --omni --port 8091 \
    --deploy-config vllm_omni/deploy/step_audio_2_async_chunk.yaml \
    --trust-remote-code --enforce-eager
```

and sending the following request:

```bash
curl -X POST http://localhost:8091/v1/audio/speech \
    -d '{"model":"stepfun-ai/Step-Audio-2-mini","input":"Please introduce vLLM in a few words.","voice":"default"}' \
    --output output.wav
```

the generated audio contains audible glitches at chunk boundaries. The glitches occur periodically, approximately once per async chunk, this issue is not observed when `async_chunk` is disabled.

sphinxkkkbc:
## Purpose
Fix #5916: audible chunk-boundary glitches when serving Step-Audio-2-mini with `async_chunk: true`.

**Root Cause:**

The stage input processor sends `left_context_size` as a nested metadata field:

```python
OmniPayloadStruct(
    codes=CodesStruct(
        audio=torch.tensor(remaining_tokens, dtype=torch.long),
    ),
    meta=MetaStruct(
        left_context_size=1,  # 1 = last chunk
        finished=torch.tensor(True, dtype=torch.bool),
    ),
)
```
The Step-Audio2 Token2Wav model previously looked for `left_context_size` at the top level:
```python
"left_context_size" in info
```
As a result, the async-chunk path was not selected, even when `async_chunk` was enabled. Each overlapping `chunk_size + pre_lookahead_len` token window was instead decoded independently through the full-sequence path, the generated waveform contained audible glitches at chunk boundaries.

The existing tests used the same incorrect flat metadata structure as the model, so they did not detect the mismatch with the actual transported payload.

## This PR:

- Adds a helper to read `left_context_size` from the nested `meta` payload.
- Updates the Token2Wav tests to use the actual transported metadata structure.

## Test Plan

```bash
pytest -v tests/model_executor/models/step_audio2/test_step_audio2_token2wav_async_chunk.py
```

**vLLM Version:**  
0.26.0

**vLLM-Omni Commit:**  
f967cfbbf5d9104266dc691c120d49d17794ddc2

## Test Result

```text
4 passed
```

**BEFORE SUBMITTING:** Read [CONTRIBUTING.md](https://github.com/vllm-project/vllm-omni/blob/main/CONTRIBUTING.md) and run the [precheck-pr skill](https://github.com/vllm-project/vllm-omni/blob/main/.claude/skills/precheck-pr/SKILL.md) with the code agent for a self-check against project conventions.


tests/model_executor/models/step_audio2/test_step_audio2_token2wav_async_chunk.py
```diff
@@ -116,7 +116,7 @@ def _reset_stream_for(state):
     out = model.forward(
         input_ids=torch.tensor([10, 11, 12]),
         positions=torch.tensor([0, 1, 2]),
-        runtime_additional_information=[{"left_context_size": 1}],
+        runtime_additional_information=[{"meta": {"left_context_size": 1}}],
```
hsliuustc0106, line 120: `left_context_size=0` is the normal non-final marker, but it is only covered through a direct `_forward_async_chunk` call above. Please add a `forward()` case with `meta.left_context_size=0` and assert `stream_chunk_for` is used, so a regression in the dispatch condition cannot silently route non-final chunks through the full-sequence path.

sphinxkkkbc: done

Gaohan123 (approved):
LGTM. Thanks

PR #5921, opened 2026-08-08
[Diffusion] Fuse Qwen-Image-Edit select01 modulation for CUDA

dongbo910220:
## Purpose

This PR optimizes the Qwen-Image-Edit select01 modulation path by fusing image-stream LayerNorm, select01 modulation, residual update, and gate extraction into Triton kernels.

The fused kernels replace the original decomposition of select/select, LayerNorm, scale, shift, residual add, and gate application.

## Implementation

- Add fused select01 modulation kernels in `vllm_omni/diffusion/layers/qwen_select01_modulation.py`.
- Route Qwen-Image-Edit image-stream norm1 and norm2 select01 modulation through the fused helpers.
- Keep non-CUDA, no-Triton, and `torch.compile` paths on the native PyTorch implementation.
- Add focused correctness tests for native/fused parity and fallback behavior.
- Add a local kernel benchmark for the select01 modulation path.

## Correctness

```bash
python -m pytest -q tests/diffusion/layers/test_qwen_select01_modulation.py
```

Result:

```text
all passed
```

### Methodology

| Item | Configuration |
| --- | --- |
| GPU | NVIDIA RTX PRO 6000 |
| Baseline | `main` |
| PR | this branch |

### Operator microbenchmark

Compared with the original decomposition, the fused CUDA path improves `layernorm_select01` by up to 5.82x and `residual_layernorm_select01` by up to 3.62x on the measured shapes.

| Seq len | layernorm main ms | layernorm PR ms | Speedup | residual main ms | residual PR ms | Speedup |
| ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| 128 | 0.086032 | 0.055200 | 1.56x | 0.095808 | 0.073296 | 1.31x |
| 1024 | 0.085248 | 0.055696 | 1.53x | 0.095312 | 0.072944 | 1.31x |
| 4096 | 0.244144 | 0.064560 | 3.78x | 0.304512 | 0.119360 | 2.55x |
| 8192 | 0.599712 | 0.125264 | 4.79x | 0.687888 | 0.251216 | 2.74x |
| 16384 | 1.453280 | 0.249616 | 5.82x | 1.804752 | 0.498032 | 3.62x |

### Qwen-Image-Edit E2E

Measured with `--enforce-eager`, 1024x1024, 50 steps, seed 42, same input image and prompt.

| Main median ms | PR median ms | Main / PR |
| ---: | ---: | ---: |
| 60378.15 | 54936.06 | 1.099x |

### Output quality

Compared `main` and this branch with the same E2E workload. Pixel deltas are normalized to `[0, 1]`.

| SSIM | PSNR | Mean absolute pixel delta | P99 absolute pixel delta |
| ---: | ---: | ---: | ---: |
| 0.96154 | 36.09 dB | 0.00883 | 0.06275 |

Both sides were deterministic across three measured runs and produced visually valid images. This is `main`-versus-PR output-stability evidence, not the repository's 1190x2032, 40-step Diffusers-parity accuracy test.

### DIFF
--- a/benchmarks/kernels/bench_qwen_select01_modulation.py
+++ b/benchmarks/kernels/bench_qwen_select01_modulation.py
@@ -1,0 +1,364 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+"""Benchmark the Qwen-Image-Edit select01 modulation hot path.
+
+This script intentionally starts with the current PyTorch implementation. It is
+used as the before/after baseline for SGLang PR #20395/#21318 style fusion:
+
+* layernorm + scale/shift + binary select01 gate
+* residual + gated update + layernorm + scale/shift + binary select01 gate
+
+Run on a CUDA machine for actionable numbers, for example:
+
+    python benchmarks/kernels/bench_qwen_select01_modulation.py \
+        --device cuda --dtype bf16 --seq-lens 4096 8192 16384 \
+        --output-json /tmp/qwen_select01_baseline.json
+"""
+
+from __future__ import annotations
+
+import argparse
+import importlib.util
+import json
+import statistics
+import sys
+import time
+from collections.abc import Callable
+from pathlib import Path
+
+import torch
+import torch.nn.functional as F
+
+for candidate in Path(__file__).resolve().parents:
+    if (candidate / "vllm_omni").is_dir():
+        if str(candidate) not in sys.path:
+            sys.path.insert(0, str(candidate))
+        break
+
+
+_FUSED_FNS: tuple[Callable[..., object], Callable[..., object]] | None = None
+
+
+def _dtype(name: str) -> torch.dtype:
+    return {
+        "fp32": torch.float32,
+        "float32": torch.float32,
+        "fp16": torch.float16,
+        "float16": torch.float16,
+        "bf16": torch.bfloat16,
+        "bfloat16": torch.bfloat16,
+    }[name]
+
+
+def _select01_modulation(
+    mod_params: torch.Tensor,
+    index: torch.Tensor | None,
+) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+    shift, scale, gate = mod_params.chunk(3, dim=-1)
+
+    if index is None:
+        return scale.unsqueeze(1), shift.unsqueeze(1), gate.unsqueeze(1)
+
+    actual_batch = shift.size(0) // 2
+    shift_0, shift_1 = shift[:actual_batch], shift[actual_batch:]
+    scale_0, scale_1 = scale[:actual_batch], scale[actual_batch:]
+    gate_0, gate_1 = gate[:actual_batch], gate[actual_batch:]
+
+    idx = index.unsqueeze(-1) == 0
+    shift = torch.where(idx, shift_0.unsqueeze(1), shift_1.unsqueeze(1))
+    scale = torch.where(idx, scale_0.unsqueeze(1), scale_1.unsqueeze(1))
+    gate = torch.where(idx, gate_0.unsqueeze(1), gate_1.unsqueeze(1))
+    return scale, shift, gate
+
+
+def _native_layernorm_select01(
+    x: torch.Tensor,
+    mod_params: torch.Tensor,
+    index: torch.Tensor,
+    eps: float,
+) -> tuple[torch.Tensor, torch.Tensor]:
+    scale, shift, gate = _select01_modulation(mod_params, index)
+    out = F.layer_norm(x.float(), (x.shape[-1],), eps=eps).to(x.dtype)
+    return out * (1 + scale) + shift, gate
+
+
+def _native_residual_layernorm_select01(
+    x: torch.Tensor,
+    residual: torch.Tensor,
+    residual_gate: torch.Tensor,
+    mod_params: torch.Tensor,
+    index: torch.Tensor,
+    eps: float,
+) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+    scale, shift, gate = _select01_modulation(mod_params, index)
+    residual_out = residual + residual_gate * x
+    out = F.layer_norm(residual_out.float(), (residual_out.shape[-1],), eps=eps).to(residual_out.dtype)
+    return out * (1 + scale) + shift, residual_out, gate
+
+
+def _load_fused_fns() -> tuple[Callable[..., object], Callable[..., object]]:
+    global _FUSED_FNS
+    if _FUSED_FNS is None:
+        module_paths = [
+            *(
+                candidate / "vllm_omni/diffusion/layers/qwen_select01_modulation.py"
+                for candidate in Path(__file__).resolve().parents
+            ),
+            Path(__file__).with_name("qwen_select01_modulation.py"),
+        ]
+        module_path = next((path for path in module_paths if path.exists()), None)
+        if module_path is None:
+            from vllm_omni.diffusion.layers.qwen_select01_modulation import (
+                fused_layernorm_select01,
+                fused_residual_layernorm_select01,
+            )
+        else:
+            spec = importlib.util.spec_from_file_location("qwen_select01_modulation", module_path)
+            if spec is None or spec.loader is None:
+                raise ImportError(f"Cannot load fused module from {module_path}")
+            module = importlib.util.module_from_spec(spec)
+            sys.modules[spec.name] = module
+            spec.loader.exec_module(module)
+            fused_layernorm_select01 = module.fused_layernorm_select01
+            fused_residual_layernorm_select01 = module.fused_residual_layernorm_select01
+
+        _FUSED_FNS = (fused_layernorm_select01, fused_residual_layernorm_select01)
+    return _FUSED_FNS
+
+
+def _time_cuda(fn: Callable[[], object], warmup: int, iters: int) -> list[float]:
+    for _ in range(warmup):
+        fn()
+    torch.accelerator.synchronize()
+
+    times: list[float] = []
+    for _ in range(iters):
+        start = torch.cuda.Event(enable_timing=True)
+        end = torch.cuda.Event(enable_timing=True)
+        start.record()
+        fn()
+        end.record()
+        torch.accelerator.synchronize()
+        times.append(float(start.elapsed_time(end)))
+    return times
+
+
+def _reset_peak_memory_stats(device: torch.device) -> None:
+    reset_peak_memory_stats = getattr(torch.accelerator, "reset_peak_memory_stats", None)
+    if reset_peak_memory_stats is None:
+        reset_peak_memory_stats = getattr(getattr(torch, device.type), "reset_peak_memory_stats")
+    reset_peak_memory_stats(device)
+
+
+def _max_memory_allocated(device: torch.device) -> int:
+    max_memory_allocated = getattr(torch.accelerator, "max_memory_allocated", None)
+    if max_memory_allocated is None:
+        max_memory_allocated = getattr(getattr(torch, device.type), "max_memory_allocated")
+    return int(max_memory_allocated(device))
+
+
+def _time_cpu(fn: Callable[[], object], warmup: int, iters: int) -> list[float]:
+    for _ in range(warmup):
+        fn()
+
+    times: list[float] = []
+    for _ in range(iters):
+        t0 = time.perf_counter()
+        fn()
+        times.append((time.perf_counter() - t0) * 1000.0)
+    return times
+
+
+def _summarize(times: list[float]) -> dict[str, float]:
+    ordered = sorted(times)
+    return {
+        "median_ms": statistics.median(ordered),
+        "min_ms": ordered[0],
+        "p20_ms": ordered[max(0, int(len(ordered) * 0.20) - 1)],
+        "p80_ms": ordered[min(len(ordered) - 1, int(len(ordered) * 0.80))],
+    }
+
+
+def _make_inputs(
+    batch: int,
+    seq_len: int,
+    hidden_size: int,
+    dtype: torch.dtype,
+    device: torch.device,
+) -> dict[str, torch.Tensor]:
+    torch.manual_seed(0)
+    return {
+        "x": torch.randn(batch, seq_len, hidden_size, device=device, dtype=dtype),
+        "residual": torch.randn(batch, seq_len, hidden_size, device=device, dtype=dtype),
+        "residual_gate": torch.randn(batch, 1, hidden_size, device=device, dtype=dtype),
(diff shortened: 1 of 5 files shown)


Dong1017:
## Review summary

Reviewed head `56f55b35f1a8` vs base `5e4baea6`.

Math restructuring is equivalent to base (select01 semantics, gate broadcasting, `hidden_states_before_attn` residual split). SP-padded `modulate_index` matches the kernel's row→batch mapping; shape mismatches fail fast in `_validate_select01_inputs`.

**Blocking:** fused path not gated on `HAS_TRITON` (inline). Plus dead branch in `_modulate`, unrelated SPDX change, benchmark loader simplification — all inline.

**Validation gaps (non-blocking):**
- No CI runs the Triton path: CUDA CI uses `-m 'core_model and cpu'` (`.buildkite/cuda/test-merge.yml:22`), so parity tests only ran on the author's GPU.
- No quality evidence: fp32 fused accumulation vs base's bf16 rounding → outputs diverge slightly over 50 steps. `tests/e2e/accuracy/test_qwen_image_edit.py` (pixelwise, `zero_cond_t` path) is the right check. PR reports E2E latency (1.098x) but no quality comparison.

**Observation:** `minimax_h3_modulation.py` already fuses gated residual + norm + `(1+scale)+shift` (gather variant). Not reusable (RMS vs LN), just noting for a future shared modulation-kernel layer.

vllm_omni/diffusion/layers/qwen_select01_modulation.py
```diff
@@ -0,0 +1,491 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM-Omni project
+
+from __future__ import annotations
+
+import torch
+import torch.nn.functional as F
+
+try:
+    from vllm.triton_utils import tl, triton
```
Dong1017: Not gated on Triton availability. With triton installed but driver check failed, `vllm.triton_utils` (pinned vLLM v0.28.0, `importing.py`) exports placeholders — this import still succeeds, then `triton.next_power_of_2` at L273/363 raises `AttributeError` instead of falling back. Contradicts "keep unsupported paths on the native implementation".

```python
from vllm.triton_utils import HAS_TRITON, tl, triton
...
if x.is_cuda and HAS_TRITON and not is_compiling:
```

Also drop the `try/except ImportError` — `vllm.triton_utils` imports fine without triton, so the except branch is unreachable (see `fused_group_norm_silu.py:21` for the pattern).

vllm_omni/diffusion/models/qwen_image/qwen_image_transformer.py
```diff
@@ -1,5 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
-# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM-Omni project
```
Dong1017, line 2: Unrelated churn: rewrites vLLM copyright to vLLM-Omni. Siblings (`pipeline_qwen_image_edit.py` etc.) keep the vLLM line — please revert.

vllm_omni/diffusion/models/qwen_image/qwen_image_transformer.py
```diff
@@ -848,8 +852,18 @@ def forward(
         txt_mod1, txt_mod2 = txt_mod_params.chunk(2, dim=-1)  # Each [B, 3*dim]
 
         # Process image stream - norm1 + modulation
-        img_scale1, img_shift1, img_gate1 = self._modulate(img_mod1, modulate_index)
-        img_modulated = self.img_norm1(hidden_states, img_scale1, img_shift1)
+        if modulate_index is not None:
+            img_modulated, img_gate1 = fused_layernorm_select01(
```
Dong1017, line 829: FYI: `minimax_h3_modulation.py` has the same fused pattern with a gather. Non-interchangeable (RMS vs LN), just flagging in case a shared modulation-kernel layer is worth extracting later.

vllm_omni/diffusion/models/qwen_image/qwen_image_transformer.py
```diff
@@ -848,8 +852,18 @@ def forward(
         txt_mod1, txt_mod2 = txt_mod_params.chunk(2, dim=-1)  # Each [B, 3*dim]
 
         # Process image stream - norm1 + modulation
-        img_scale1, img_shift1, img_gate1 = self._modulate(img_mod1, modulate_index)
-        img_modulated = self.img_norm1(hidden_states, img_scale1, img_shift1)
+        if modulate_index is not None:
```
Dong1017: Note: the `if index is not None:` branch in `_modulate` (L800-829) is now dead code — after this PR `_modulate` is only called with `index=None` (here and the norm2 else) or with no index (txt streams). Delete the branch and parameter, or delegate to `select01_modulation_native` — the select01 logic shouldn't exist twice.

benchmarks/kernels/bench_qwen_select01_modulation.py
```diff
@@ -0,0 +1,364 @@
...
+    index: torch.Tensor,
+    eps: float,
+) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+    scale, shift, gate = _select01_modulation(mod_params, index)
+    residual_out = residual + residual_gate * x
+    out = F.layer_norm(residual_out.float(), (residual_out.shape[-1],), eps=eps).to(residual_out.dtype)
+    return out * (1 + scale) + shift, residual_out, gate
+
+
+def _load_fused_fns() -> tuple[Callable[..., object], Callable[..., object]]:
```
Dong1017: Over-engineered: L22-27 already put the repo root on `sys.path`, so a plain `from vllm_omni.diffusion.layers.qwen_select01_modulation import ...` replaces this ~30-line loader (and avoids shadowing `sys.modules`).

vllm_omni/diffusion/layers/qwen_select01_modulation.py
dongbo910220: Fixed in `810dafe1`. I imported `HAS_TRITON` from `vllm.triton_utils`, removed the unreachable `ImportError` fallback, and gated both CUDA launch paths on `HAS_TRITON`. I also added a CUDA regression test that forces `HAS_TRITON=False` and verifies that both public functions use the native implementation without invoking either Triton launcher. The focused tests all passed.

vllm_omni/diffusion/models/qwen_image/qwen_image_transformer.py
dongbo910220: I checked this against the current required pre-commit hook before reverting it. `tools/pre_commit/check_spdx_header.py` explicitly treats the vLLM copyright line as `LEGACY_COPYRIGHT_TEXT` and rewrites modified files to `Copyright contributors to the vLLM-Omni project`. Reverting this line causes `check-spdx-header` to modify the file and fail, while the Omni line passes. The sibling files still carry legacy headers because they have not yet been rewritten as changed files by this hook, so I kept the header required by current `main`.

vllm_omni/diffusion/models/qwen_image/qwen_image_transformer.py
dongbo910220: There is conceptual overlap, but the current contracts differ in normalization (`LayerNorm` versus `RMSNorm`), parameter/index layout, and returned values. I kept this PR scoped to the Qwen-specific implementation and left a shared modulation-kernel layer for a separate follow-up rather than expanding this change.

vllm_omni/diffusion/models/qwen_image/qwen_image_transformer.py
dongbo910220: Removed the now-dead `index` parameter and select01 branch from `_modulate`, and updated the two image fallback call sites to use the remaining non-indexed modulation path directly. The select01 implementation now lives only in `qwen_select01_modulation.py`. I also verified the remaining `_modulate` behavior against the chunk-and-unsqueeze reference, and the focused tests all passed.

benchmarks/kernels/bench_qwen_select01_modulation.py
dongbo910220: Removed the file-location loader, its `sys.modules` mutation, and the function cache. The benchmark now imports both fused functions directly from `vllm_omni.diffusion.layers.qwen_select01_modulation` after adding the repository root to `sys.path`. The simplified benchmark's CUDA correctness checks all passed.

dongbo910220:
Thanks @Dong1017 for the detailed review. I addressed the actionable code issues in `810dafe1` and updated the PR description.

On CI coverage: the current merge pipeline selects `core_model and cpu`, so it does not execute this Triton path. I added a CUDA regression test for the no-Triton fallback and ran the full focused test module locally; all passed. I did not expand this kernel PR into a repository-wide CI topology change or claim that public CI exercised the Triton kernel.

On output quality: I added the same-machine, same-model, same-input, same-prompt, same-seed 1024x1024/50-step `main`-versus-PR results to the description. The measured values are SSIM `0.96154`, PSNR `36.09 dB`, normalized mean absolute pixel delta `0.00883`, and normalized p99 delta `0.06275`; each side was deterministic across three measured runs and all images were visually valid. These mean/p99 values are below the numeric `0.03/0.4` thresholds used by the current 2511 pixelwise test.

Dong1017:
## Review update (head `9bf3359b4`)

Previous four inline findings are resolved in `810dafe1` (HAS_TRITON gate + regression test, dead `_modulate` branch, benchmark loader). The SPDX change is fine — `tools/pre_commit/check_spdx_header.py` rewrites modified files to the vLLM-Omni line, so no revert needed.

Two follow-ups inline: an NPU regression (eager fallback bypasses the mindiesd fused path, measured ~1.1 ms/norm on Ascend 910B1) and a note that ROCm takes the Triton path unvalidated. Prior validation gaps still stand (no CI runs the Triton path; no base-vs-head quality evidence).

vllm_omni/diffusion/models/qwen_image/qwen_image_transformer.py
```diff
@@ -848,8 +822,18 @@ def forward(
...
```
Dong1017: [NPU regression] This routes all non-CUDA select01 paths through the eager fallback, bypassing `AdaLayerNorm.forward_npu`'s mindiesd fused kernel. Measured on Ascend 910B1 (v0.28.0 image, B=2, L=4096, C=3072, bf16): BASE (select + `mindiesd.layernorm_scale_shift(fused=True)`) ~1.86 ms vs this PR (select + eager `F.layer_norm`) ~2.94 ms → ~1.1 ms/norm × 60 layers × 2 ≈ ~130 ms/step, ~6.5s over 50 steps. Numerics fine (max abs diff 0.125); purely speed.

Fix: keep select01 selection on non-CUDA but flow through the existing AdaLayerNorm dispatch:

```python
if modulate_index is not None and hidden_states.is_cuda and HAS_TRITON:
    img_modulated, img_gate1 = fused_layernorm_select01(...)
else:
    scale, shift, gate = select01_modulation_native(img_mod1, modulate_index)
    img_modulated = self.img_norm1(hidden_states, scale, shift)
```

Verified on-device that `mindiesd.layernorm_scale_shift` accepts `[B, L, C]` scale/shift (incl. non-contiguous `torch.where` output), so the native path works as-is. Same at L867.

vllm_omni/diffusion/layers/qwen_select01_modulation.py
```diff
@@ -0,0 +1,486 @@
...
+    x: torch.Tensor,
+    mod_params: torch.Tensor,
+    index: torch.Tensor,
+    eps: float,
+    weight: torch.Tensor | None = None,
+    bias: torch.Tensor | None = None,
+) -> tuple[torch.Tensor, torch.Tensor]:
+    scale0, shift0, gate0, scale1, shift1, gate1 = split_select01_mod_params(mod_params)
+    is_compiling = torch.compiler.is_compiling()
+    if x.is_cuda and HAS_TRITON and not is_compiling:
```
Dong1017: FYI: ROCm also satisfies `x.is_cuda`, so ROCm takes the Triton path, but perf/parity evidence is CUDA-only. Plain `tl` ops so likely fine — flagging it as unvalidated.

vllm_omni/diffusion/models/qwen_image/qwen_image_transformer.py
dongbo910220: Fixed in `becceb94`. For select01 inputs, the fused helpers are now used only when the NVIDIA CUDA/Triton predicate passes. Other backends use `select01_modulation_native` for scale/shift/gate selection and then call the existing `self.img_norm1` and `self.img_norm2`, preserving `AdaLayerNorm.forward_npu` and its MindIE-SD dispatch. The norm2 fallback also applies the gated residual before that dispatch, matching the original path. I added a model-level regression test that fails if either fused helper is called on the fallback route and verifies that both image norms are invoked. The focused tests, targeted pre-commit, and CUDA compile smoke all passed. I did not remeasure this on Ascend locally.

vllm_omni/diffusion/layers/qwen_select01_modulation.py
dongbo910220: Fixed in `becceb94` by restricting the support predicate to `HAS_TRITON and current_platform.is_cuda() and x.is_cuda`. A ROCm tensor no longer enters the unvalidated Triton route and instead follows native select01 parameter selection plus the existing `AdaLayerNorm.forward_hip` dispatch. NVIDIA CUDA eager still uses Triton, while `torch.compile` keeps the native/Inductor fallback. I added predicate coverage for the ROCm-like case where `x.is_cuda` is true but the platform is not CUDA. The focused tests and CUDA compile smoke all passed.

princepride (approved):
LGTM

dongbo910220:
@princepride The three failed CI lanes have been investigated and the PR is updated.

The CUDA regression was fixed by supporting a broadcast `[1, L]` select index when `num_outputs_per_prompt > 1`. 
The AMD MAGI failure is already fixed on current `main`, which has now been merged.
The Intel failure was a checkout-time `No space left on device` infrastructure error before PR code ran. Relevant CUDA, Qwen-Image, Qwen-Image-Edit-Plus, and MAGI validation passed on the updated head.

dongbo910220:
@NickCao Could you help remove and re-add the `ready` label to retrigger accelerator CI on head `0d40dad8`? The author account is denied label permission with HTTP 403, and the existing CUDA, AMD, and Intel builds are still attached to the old head `becceb94`.

dongbo910220:
@princepride Quick update on this PR: The only remaining red check is Intel CI #8314, which failed during repository checkout after six `No space left on device` errors on `aicf-jf-cicd-15-xpu-1`; no PR code or tests ran. This is the same runner issue tracked in #7587.

Issue #5925, opened 2026-08-08
`VLLM_TORCH_PROFILER_DIR` produces no `/start_profile` route and no warning on the omni diffusion path


stewtong:
Setting `VLLM_TORCH_PROFILER_DIR` when serving a diffusion model via `--omni` has no effect: no `/start_profile` endpoint registers, no warning is logged, and no error is raised. Observed serving `nvidia/Cosmos3-Super` on the `vllm/vllm-omni:cosmos3` image (digest `sha256:6d2630c7d637...`, vllm-omni `0.25.0rc2.dev62+g9c1b7504b`); the code references below are to that build. The gate is `_should_enable_profiler_endpoints()` in `vllm_omni/entrypoints/openai/api_server.py`, which requires `engine_args.profiler_config.profiler` to be set from a stage config file. No stage config ships for Cosmos3-Super (and, as far as I can tell, no other diffusion model in this repo), and no CLI flag exposes `profiler_config` directly (`vllm serve --help` has no `profiler` match on the omni path).

The practical effect: there is no way to decompose the fixed per-request cost outside the denoise loop (roughly 14s, or 12% of latency at 35 steps, on Cosmos3-Super) into its text-encode / VAE-decode / MP4-encode / scheduling components through the documented serve path. `worker/base.py` logs a message explaining the stage-config requirement, but only when profiling is attempted through a path that reaches that log line; setting the env var alone produces nothing.

**Suggested fix:** either honor `VLLM_TORCH_PROFILER_DIR` directly on the omni diffusion path (matching the documented behavior of vLLM's LLM serving path), or, at minimum, log a warning at startup when the env var is set but the profiler endpoints do not register, so the silent case is distinguishable from "profiler ran and found nothing." The current behavior is worse than an error, because it looks identical to success.

**Suggested fix (alternative, lower effort):** ship a minimal stage config for at least one diffusion model as a worked example, so operators have something to copy rather than needing to reverse-engineer the gate.

hsliuustc0106:
@ZacheryAU PTAL

moisf56:
@ZacheryAU are you on this? If not I'd like to take it.

Confirmed on current main: `_should_enable_profiler_endpoints()` in
`vllm_omni/entrypoints/openai/api_server.py:221` only registers `profiler_router`
if some stage sets `engine_args.profiler_config.profiler`. There is no else branch,
[...]

Issue #5923, opened 2026-08-08
Hardcoded 600s client timeout in `backends.py` silently kills long requests and understates throughput


stewtong:
`benchmarks/diffusion/backends.py` hardcodes `timeout_seconds = 600.0` with no CLI flag to override it.

Serving `nvidia/Cosmos3-Super` at a production shape (t2v, 1280x720, 189 frames, 35 steps, guardrails on) and running `diffusion_benchmark_serving.py` with `--dataset random --request-rate inf --max-concurrency 8` over 16 prompts, mean latency measures 729s once the timeout is raised far enough for every request to complete. Under the stock 600s value the client's own timeout fires before most requests finish, those requests are recorded as failures, and reported throughput is computed against the requests that did finish within 600s.

Measured on 8x H200 SXM, image `vllm/vllm-omni:cosmos3` (digest `sha256:6d2630c7d637...`), vllm-omni `0.25.0rc2.dev62+g9c1b7504b`:

| Client timeout | Completed | Reported throughput |
|---|---|---|
| 600s (stock) | 5 of 16 | 14.96 clips/hr |
| 5400s (patched) | 16 of 16 | 30.96 clips/hr |

The stock harness understates throughput by 52% on this shape and presents it as a collapse under load rather than as a client-side timeout. Any concurrency result for a long-running generation model taken with the unmodified harness should be re-checked against a longer timeout.

**Suggested fix:** expose `timeout_seconds` as a CLI flag (e.g. `--client-timeout`), defaulting to the current 600s for backward compatibility. Happy to open a PR if that's useful.

HaningZS:
I can take this. I reproduced the hard-coded timeout on current `main` (`b4b907e7`) with a deterministic L1 CPU regression test.

I plan to keep the change focused: expose `--client-timeout` with the existing 600-second default, propagate it to `/v1/videos` job polling (including warmups), add timeout/cleanup and CLI coverage, and document the flag. I will open a PR once the focused and related benchmark tests pass.

PR #5918, opened 2026-08-08
[Model] Add native MAGI-2 Preview diffusion support

hsliuustc0106:
## Summary

Adds a native vLLM-Omni implementation of
[`sand-ai/MAGI-2-preview`](https://huggingface.co/sand-ai/MAGI-2-preview).
The runtime does not import SandAI's inference package or require a reference
checkout.

- Implements the complete native Preview path: 40-layer DiT, mHC, learned-sink
  attention, MagiMoE, packed CFG sampling, checkpoint loading, TurboVAE video
  decode, and Oobleck audio decode. The CUDA path uses the shared versioned
  vLLM FlashAttention utility: FA3 on Hopper and as the Blackwell fallback, FA4 on supported Blackwell,
  and FA2 on earlier devices.
- Supports 272p/540p T2VA and I2VA with 10-second, 125-frame, 12.5-fps video and
  synchronized 44.1-kHz stereo audio. Released-checkpoint qualification covers
  full 100-step T2VA/I2VA and reference/distributed parity.
- Keeps resident SP4 as the recommended four-device default; TP4 and TP2SP2 are
  supported, with DLO DP4/DP2SP2 and rank-local SP4 available for memory or
  throughput tradeoffs. The measured Pareto frontiers are included below.
- Adds HSDP over SP, two-way CFG parallelism, exact TurboVAE tile parallelism,
  and Cache-DiT composition on the repeated transformer-layer stack.
- Enables one-device ordinary layerwise DiT offload while retaining MAGI's
  native pinned-CPU staging for Qwen, image VAE, TurboVAE, and Oobleck. The
  released four-step run used 49.29 GiB peak HBM and 371.97 GiB host PSS;
  standalone whole-model CPU offload remains intentionally rejected.
- Integrates Hub/local detection, OpenAI video serving, the deployment recipe,
  and the existing shared T2V/I2V scripts. It also captures the native-model
  integration lessons in the repository's diffusion-model and recipe skills.
  No benchmark or MAGI-specific example script is added.

Architecture and rollout rationale: [#5949](https://github.com/vllm-project/vllm-omni/issues/5949).

### Code-change summary

Counts are relative to the current `main` merge base.

| Component | Files | Added | Removed | Scope |
|---|---:|---:|---:|---|
| Native MAGI-2 model | 13 | 6,114 | 0 | Transformer, attention, MagiMoE, parallelism, conditioning, sampler, data proxy, and native decoders |
[...]

Issue #5799, opened 2026-08-05
The parameter dlo-resident-layer is not supported.


wuxiaohui0:
### Your current environment

```text
The output of `python collect_env.py`
```

### How you are installing vllm-omni

(omni) root@3e06e081e50d:~/xinglin-data# CUDA_VISIBLE_DEVICES=0,1 \
VLLM_WORKER_MULTIPROC_METHOD=spawn \
VLLM_OMNI_VIDEO_SYNC_TIMEOUT=14400 \
vllm serve /root/xinglin-data/MiniMax-H3/FL2VA \
  --omni \
  --host 0.0.0.0 \
  --port 8000 \
  --trust-remote-code \
  --num-gpus 2 \
  --tensor-parallel-size 2 \
  --usp 1 \
  --ring 1 \
  --text-encoder-tp-size 2 \
  --vae-patch-parallel-size 2 \
  --vae-parallel-mode tile \
  --vae-use-tiling \
  --enable-distributed-layerwise-offload \
  --dlo-no-use-allgather \
  --dlo-resident-layers "${DLO_RESIDENT_LAYERS}" \
  --enforce-eager \
  --diffusion-attention-backend CUDNN_ATTN
INFO 08-05 15:12:04 [patch.py:252] NVFP4 W4A4 weight_scale NaN-clamp: installed.
INFO 08-05 15:12:04 [patch.py:481] [cumem-cuda] CuMemAllocator._python_free_callback patched: asleep guard extended to all platforms.
INFO 08-05 15:12:06 [main.py:54] Delegating entrypoint handling to vllm-omni
usage: vllm [-h] [-v] {serve,bench} ...
vllm: error: unrecognized arguments: --dlo-resident-layers 12

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://vllm-omni.readthedocs.io), which can answer lots of frequently asked questions.

harleywang:
I also encountered this problem.

wuxiaohui0:
> I also encountered this problem.

Have you resolved this issue?

Gaohan123:
Please change the title of issue for clearance. Thanks

cadslab:
I got vllm: 
error: unrecognized arguments: ~~--enable-distributed-layerwise-offload --dlo-no-use-allgather3~~ --dlo-resident-layers 12
[...]

PR #5926, opened 2026-08-08
[Doc] Add Cosmos3-Super serving performance dashboard entry (8x H200 and 8x B200)

stewtong:
## Purpose

Adds a serving performance dashboard entry for nvidia/Cosmos3-Super on 8x H200 and 8x B200, alongside the existing Qwen-Image and Wan 2.2 entries. It covers service launch configuration across five parallel strategies, benchmark methodology on the first-party diffusion serving harness, a latency decomposition, a production-shape concurrency envelope with three controls, attention-backend results, guardrail cost scoping, a mixed-workload trace, and a reproducibility checklist including a determinism caveat that affects output comparison. The primary serving tables use the pinned `vllm/vllm-omni:cosmos3` image; other image and batch-path comparisons identify their versions separately. Its digest and versions are in the entry; the single-request cells match the numbers in the Cosmos3-Super recipe PR (#5909).

Section 9 adds a B200 addendum measured at the same anchor shape on 8x B200 SXM: the parallel-strategy pair, a scaling ladder, a concurrency spot-check with peak-memory readings, restart-drift measurements, and batch-path results.

Section 10 answers the DLO review request with a separate B200 sweep on pinned v0.28.0. Both failures reproduced on v0.26.0 (DP2 request-batch rejection and SP4 worker crash); v0.28.0 completes one cold plus three valid warm waves for each. This is a release-level comparison because the images also change dependencies. The resident frontier comprises hybrid CFG2 x Ulysses4 x HSDP8 (74.9 s, 6.01 videos/h/GPU), TP4 (125.9 s, 7.15 videos/h/GPU), and single GPU (381.8 s, 9.43 videos/h/GPU). Times are mean warm-wave completion times, measured from the first request submission to the last response-body completion in each wave. Tested DLO layouts use less sampled GPU memory (19-23 GiB/GPU versus 36-133 GiB), but every DLO point is dominated on the two measured speed axes. Equal output quality is unestablished, and slow DP-family requests have substantial scheduler queue wait of unresolved cause.

The addition includes the plot, all three warm-wave spans for each configuration, image pins, launch flags and prompt files. Historical H200 results remain scoped to their original image.

## Test Plan

Documentation-only change: adds one new entry under `benchmarks/diffusion/performance_dashboard/` following the structure of the existing Qwen-Image and Wan 2.2 entries. Methodology for every measured figure is described in the entry itself.

**Primary serving image:** vLLM 0.25.0 / vLLM-Omni 0.25.0rc2.dev62+g9c1b7504b. Sections 2 and 9 identify their separate v0.26.0 controls; section 9 also identifies the `npa-cosmos3:1.2.2-cu130` batch path.

**DLO comparison in section 10:** digest-pinned vLLM/vLLM-Omni v0.26.0 failure controls and v0.28.0 frontier.

## Test Result

Sections 1 through 8 report H200 measurements; section 9 reports B200 measurements. Image versions, benchmark arguments and reproducibility guidance accompany the results. Section 10 was measured on B200 with v0.28.0, guardrails off, using direct `/v1/videos/sync` requests. Cold waves are excluded from its rate calculation. Outputs pass full decoding and format checks, but differ across topologies. Measurements were independently recomputed from primary records because the benchmark event log has invalid hashes; excluded attempts, variable telemetry intervals and UTC wave timing are disclosed in the section. One link points to the profiler-routing issue filed alongside this PR (#5925).

Validation for this documentation update: repository pre-commit checks, recalculation of all 24 warm waves, linked-file checks and visual plot inspection. No new GPU run or runtime-code change.

## Updates

September 10, 2026: [B200 DLO rerun and section 10](https://github.com/vllm-project/vllm-omni/pull/5926#issuecomment-5629234586).

[Correction to section links, version scope and metric labels](https://github.com/vllm-project/vllm-omni/pull/5926#issuecomment-5629751413). Measurement values are unchanged.


hsliuustc0106:
@bjf-frz PTAL

stewtong:
Follow-up commit adds a B200 addendum (section 9): the same anchor-shape methodology on 8x B200 SXM — parallel-strategy pair (recommended over TP-8 holds, 10.7%/12.5%), scaling ladder vs the published grid, a concurrency spot-check with byte-identical peak memory, restart-drift measurements, and batch-path results. Same measurement discipline as the H200 entry; per-step scaling derived with the entry's own section 7.2 step-delta method.

hsliuustc0106:
can you try to use DLO with different parallelism and draw the pareto frontier? 
image

stewtong:
Sure. One catch: this entry's pinned image (`vllm/vllm-omni:cosmos3`, pushed 2026-07-20) predates DLO (#5397, merged 2026-07-30), so the sweep has to run on v0.26.0 instead. That image ships without the cosmos-guardrail package (#5936), so points will be guardrails-off, labeled with image and posture so they aren't read against the tables above.

Plan (updated after reading the v0.26.0 DLO source): DLO-off baselines (recommended CFG2 x Ulysses4 x HSDP8, TP-8, TP-4, single GPU) plus the Cosmos3-DistOffload recipe layouts (DP4 AllGather, DP2+SP2 AllGather, SP4 rank-local), each DLO cell driven by a concurrent same-step wave sized to its DP degree. Frontier in your MAGI-2 format: warm user latency vs warm throughput per device, with per-GPU peak memory and host PSS as columns. Running now; results will land as a follow-up section here.

stewtong:
Ran the sweep on v0.26.0 (the pinned image predates #5397), guardrails off, anchor shape (720p/189f/35steps, seed 17), on 8x H200. Full run evidence is reproducible from the artifacts; method per #5918 (warm waves = dp-size concurrent identical-step requests; x = warm wave completion ceiling, y = outputs/warm-wave-seconds/N devices).

Short version: DLO cannot serve Cosmos3-Super traffic at v0.26.0 on our stack, so there is no warm DLO frontier to plot. Every recipe layout in Cosmos3-DistOffload.md was tried with the wave protocol:

1. --enable-distributed-layerwise-offload --data-parallel-size 4 (4 GPU, wave of 4): server ready in 110 s, mmap weight path engaged (~8,100 MiB HBM per card, model resident in host page cache), but every request is rejected instantly: "RuntimeError: Cosmos3OmniDiffusersPipeline does not support request-batch forward" (HTTP 500). Same for --data-parallel-size 2 --usp 2 and the recipe's own --data-parallel-size 2 for Super. The DP multi-concurrency admission path needs a batched forward this pipeline does not implement.

2. --enable-distributed-layerwise-offload --usp 4 --dlo-no-use-allgather (4 GPU, single request): denoise started (step 1 at 10.97 s/it), then DiffusionWorker-0 segfaults and the engine stays dead for subsequent requests ("Executor shut down", 500).

3. --dlo-resident-layers is not a v0.26.0 flag at all (argparse: "vllm: error: unrecognized arguments: --dlo-resident-layers 21"). That flag appears in MiniMax-H3 community recipes; upstream has --enable-distributed-layerwise-offload, --dlo-use-allgather/--dlo-no-use-allgather, --data-parallel-size, --usp only. This matches the open #5799 ("The parameter dlo-resident-layer is not supported").

So the honest per-config numbers are the baseline cells at the anchor shape (wall / stage_gen_time_ms / peak MiB per GPU): recommended CFG2xU4xHSDP8 8x 124.0 s wall (120.1 stage, 45,977 MiB rank-0 peak), TP-8 8x 133.3 s (129.0 stage, 43,881), TP-4 4x 227.0 s (223.7, 57,573), 1 GPU 757.0 s (753.7, 141,665 MiB = 98.5% of the 143,771 MiB card at this shape). Per-device throughput clips/hr: 4.76 / 3.97 / 3.63 / 3.38 for 1/4/8 GPUs (B4/B3/B1/B2).

Notes worth having upstream regardless: (a) /v1/videos/sync has a 600 s server-side default (VLLM_OMNI_VIDEO_SYNC_TIMEOUT, api_server.py:2955); a 1-GPU 720p/189f request necessarily exceeds it and gets aborted mid-generation with a 504 (related to but distinct from the 600 s client timeout in #5923). (b) The DLO cells' memory picture looked exactly as advertised (2-block double-buffer, weights in host page cache) until the first request; the diagnosis gaps are in the request path, and #5397's "GPUs: NVIDIA B300" prerequisite line may overstate current GPU coverage.

Happy to supply the full per-cell logs, mem series, and error traces; the SP4 segfault was observed once (the engine stays dead after it), and we can rerun with debug env on the same hardware at your convenience. Filed the two DLO defects separately: #5953 (DP admission vs missing request-batch forward) and #5954 (SP4 rank-local worker segfault).

diddidhruv:
Woah!!! This is 🔥 🔥 🔥

stewtong:
Updated the title and description to match the branch: section 9 adds the 8x B200 addendum (commit 1411dbf7), and the DLO result from the review request is now summarized in the description with links to #5953 and #5954. Also corrected the #5909 reference, since that recipe PR is open rather than merged. No measurements changed.

stewtong:
Added [section 10](https://github.com/stewtong/vllm-omni/blob/19ccfc95e43bb77bcdf483548d6f15fb3f285f9b/benchmarks/diffusion/performance_dashboard/cosmos3_super_serving_performance.md#10-dlo-on-b200-with-v0280) with the B200 DLO results and frontier plot. v0.26.0 reproduces the DP2 request-batch rejection and SP4 worker crash. With pinned v0.28.0, both complete one cold plus three valid warm waves; the tested SP4 configuration no longer needs the eager workaround. This compares releases: Torch, Diffusers and Triton also change between images.

The resident layouts form the measured latency/throughput frontier. DLO uses less sampled GPU memory (19-23 GiB/GPU versus 36-133 GiB), but every tested DLO layout is slower and produces fewer videos per GPU-hour than at least one resident layout. Outputs decode correctly but differ across topologies; equal output quality remains unestablished. Slow DP-family requests also show substantial scheduler queue wait of unresolved cause.

The section documents excluded attempts and timing limitations. Measurements were recomputed from the primary records because the benchmark event log has invalid hashes.

stewtong:
Corrected the description's image/version scope, labeled the three frontier configurations and their timing metric, and restored the missing section 10 heading and link. The measured values and conclusions are unchanged.

PR #5928, opened 2026-08-08
[BugFix][Test] Initialize weights in MoT und-mode linear tests

### LINKED ISSUE #5927: [Bug]: MoT und-mode linear tests rely on uninitialized weights and fail spuriously when CUDA allocations are zero-filled
### Your current environment

Observed on an 8x H200 host (vLLM 0.26.0, vLLM-Omni main @ cfa6d66e, Python 3.12, CUDA 12.x). The failure is host-dependent (see root cause); other machines may not reproduce it.

### 🐛 Describe the bug

`tests/diffusion/kernels/mot/test_mot_linear.py::test_mot_qkv_und_mode` and `::test_mot_row_und_mode` fail deterministically on some hosts:

```
FAILED tests/diffusion/kernels/mot/test_mot_linear.py::test_mot_qkv_und_mode[QKV_K3584_N4608_bf16-True]
...
AssertionError: Cosine similarity too low: 0.000000
  [QKV und-mode M=1026 K=3584]  max_abs=0.0000e+00  max_rel=0.0000e+00  min_cos_sim=0.000000
```

Note `max_abs=0`: the MoT output and the reference output are actually **identical** — both are all-zero. The und-mode fallback path itself is correct.

**Root cause (test bug, not a kernel bug):** unlike the gen-mode tests, which randomize weights via `_sync_weights`, the two und-mode tests construct the reference layer and copy its weights into the MoT layer **without ever initializing them**. vLLM linear layers allocate parameters via `torch.empty`, so the test outcome depends on whatever bytes the allocator returns:

- On hosts where fresh CUDA allocations are zero-filled (e.g. fresh pages on this machine), weights are all-zero → both outputs are all-zero → `cosine_similarity` of zero vectors returns 0 → the `cos_sim > 0.98` assert fails spuriously.
- On hosts where the memory contains recycled garbage, the tests happen to pass. This also makes the failure flaky within a session (a case can pass when the caching allocator reuses non-zero memory from earlier tests).

Verified on the same machine: after properly initializing the weights (`normal_(0, 0.02)`) the und-mode outputs match the reference exactly (`min_cos_sim = 1.0`, `max_abs_diff = 0.0`).

Two related observations:

1. `tests/diffusion/kernels/` is currently not included in any Buildkite CUDA pipeline (`test-ready/merge/nightly/weekly.yml` select other directories, and these tests carry the `gpu` marker while the pipelines filter on `cuda`), so this never surfaced in CI.
2. `test_adalayernorm.py` errors when run in the same pytest invocation after the MoT tests (process-group state interference) but passes standalone; also pre-existing and unrelated.

Fix: initialize the weights in the two und-mode tests before copying them into the MoT layer, mirroring what the gen-mode tests already do.

- [x] Make sure you already searched for relevant issues, and checked the documentation and examples for answers to frequently asked questions.

congw729:
## Purpose

Fixes #5927.

`test_mot_qkv_und_mode` and `test_mot_row_und_mode` copied the reference layer's weights into the MoT layer without ever initializing them. vLLM linear layers allocate parameters via `torch.empty`, so on hosts where fresh CUDA allocations are zero-filled both the reference and the MoT outputs are all-zero and the cosine-similarity gate fails spuriously (`cosine_similarity` of zero vectors is 0), even though the two outputs are identical.

Minimal fix: initialize the weights (`normal_(0, 0.02)`, matching what `_sync_weights` does for the gen-mode tests) before copying them into the MoT layer. No production code is touched.

## Test Plan

**vLLM Version:** 0.26.0 (H200 dev machine)

**vLLM-Omni Commit:** based on f663d120

Run the full MoT linear kernel test file on a machine that reproduces the spurious failure:

```
pytest tests/diffusion/kernels/mot/test_mot_linear.py -q
```

## Test Result

- Before (unmodified main on the same machine): `8 failed, 40 passed` — all failures in the two und-mode tests.
- After this fix: `48 passed in 8.97s`.


congw729:
@hsliuustc0106 This is ready for merge

congw729:
@Gaohan123  This can also be merged.
The issue appeared after the machine reboot.

congw729:
The fix is already on main through #6606, merged on August 27. Both test_mot_qkv_und_mode and test_mot_row_und_mode now initialize the reference weights and bias before copying them to the MoT layer, covering the changes proposed here. The overlapping edits also explain the current merge conflict. Closing this PR as superseded by #6606; the original issue #5927 is resolved by that change.

Issue #5512, opened 2026-07-28
[Feature] [Design] RL Integration: Batch-Invariant Diffusion Rollouts


Ronnie-Rui:
## Problem

> Roadmap item: [vLLM-Omni #4901](https://github.com/vllm-project/vllm-omni/issues/4901)
> Design context: [#4864](https://github.com/vllm-project/vllm-omni/issues/4864), [#3435](https://github.com/vllm-project/vllm-omni/issues/3435)

RL rollout workers submit many requests concurrently, so the same logical request — same prompt, same seed — can land at a different batch index or next to different neighbors on each run. Because GPU floating-point addition is not associative, changing batch composition changes how reductions are partitioned, which can change the last few bits of the output.

Ordinary inference never notices. RL does, because rollout output is reused as training evidence: a last-bit difference in a logprob feeds into reward ranking, then advantage, then the gradient. What starts as a rounding difference ends up as a different training run.

So what we need is:

```text
output(A alone at BS1) == output(A inside a real BSN batch)
```

Bitwise, and holding when A changes batch index, when its neighbors change, and when the batch splits differently.

Two weaker properties are easy to mistake for this. Same-seed reruns matching is necessary but says nothing about batching. And running one request at a time (`max_num_seqs=1`) does give stable output, but by avoiding co-batching rather than surviving it — useful as a control baseline, nothing more. The practical risk is a system that quietly falls back to serialized execution and reports success, delivering the weaker property while claiming the stronger one.

## Dependency direction

```mermaid
flowchart LR
    A[VeRL-Omni<br/>explicit seeds, fixed-recipe E2E]
    B[vLLM-Omni<br/>request contract + capability gate]
    C[native vLLM<br/>batch-invariant kernels and runtime]
    A --> B --> C
```

vLLM-Omni reuses native vLLM batch-invariance facilities rather than reimplementing kernels, and must never depend on the training framework. The training side consumes the capability; it does not define it.

## Design: three enforcement layers

Three separate things have to be true at once, and they happen at different moments in a request's life, so they cannot be collapsed into a single switch.

```mermaid
flowchart TD
    A[Request submitted] --> B{BIC mode enabled?}
    B -- No --> C[Existing default behavior, unchanged]
    B -- Yes --> D[Layer 1 · RNG identity<br/>request boundary]
    D --> E[Layer 2 · Capability gate<br/>engine construction + each request]
    E -- Unsupported --> F[Fail closed, list every offending field]
[...]

PR #5935, opened 2026-08-09
[Diffusion] [batch invariance] Install batch invariance in the diffusion worker and document its verified scope [2/2]

Ronnie-Rui:
### Purpose

Design issue #5512 splits batch-invariance enforcement into three layers. **This PR is Layer 2, and it stacks on #5815**.

#5815 makes diffusion randomness reproducible: a request must carry one explicit integer seed. What it does not do is install vLLM's batch-invariant operator overrides on the diffusion side, so the same seed still runs through kernels whose reduction partitioning depends on batch composition, and the low bits still move. This PR closes that half.

Scope is communicated through documentation rather than refusal. determinism holds for the operators vLLM actually replaces, and anything else is *unverified* rather than *unsupported*. This follows vLLM's own approach of listing validated models while noting that others may also work.

### What this PR changes

Production:

- `vllm_omni/diffusion/worker/diffusion_worker.py` — bootstrap batch invariance during worker startup, after the CUDA device is selected and before `init_distributed_environment`, so vLLM's NCCL determinism overrides land before the communicator exists. Returns silently on ROCm and non-CUDA.
- `vllm_omni/diffusion/batch_invariance.py` — add `validate_batch_invariant_diffusion_request`, which additionally rejects `generator_device`, `latents` and `sigmas`. These are not recipe constraints: caller-supplied `latents` mean the seed no longer decides the initial noise, and `sigmas` rewrite the schedule the seeded trajectory follows.
- `vllm_omni/diffusion/request.py` — record whether the seed was caller-supplied, and run the request contract from `__post_init__`, which sees every construction path. Warmup requests stay exempt via `is_dummy_run()`.

Docs:

- `docs/features/batch_invariance.md` (new) and `docs/.nav.yml` — enabling, the seed contract, the measured evidence table, why only SD3 has evidence, the multi-GPU position, and known limitations.

### Test Plan

`pytest tests/diffusion/test_batch_invariance_seed_contract.py tests/diffusion/test_diffusion_request.py`, plus the Layer 1 suites from #5815 (`tests/engine/test_stage_pool_diffusion_seed.py`).

### Measured batch-invariance evidence

Shared recipe: `StableDiffusion3Pipeline` (stable-diffusion-3.5-medium), 512x512, 8 steps, `torch.bfloat16`, `TORCH_SDPA`, `enforce_eager=True`, `num_gpus=1`. Four requests A/B/C/D each pin their own prompt and seed (A = `"a red ceramic cup on a plain white table"`, seed 101); the BS4 batch is submitted as `[B, A, D, C]` and only the A path is compared. Every hash is a per-sample tensor sha256 over `shape=[1, 3, 512, 512]`, `nbytes=1572864`, bf16 — not a whole-batch tensor.

**RTX 4090 / SM 8.9 — invariance holds**

| | A alone (BS1) | A inside BS4 | Verdict |
|---|---|---|---|
| gate on | `a731ab61eac277b30a2fcbaf547e812251fd184b605360d8d453603c6f79b478` | `a731ab61eac277b30a2fcbaf547e812251fd184b605360d8d453603c6f79b478` | **bitwise identical** |
| gate off | `daedf31688ccd792404b956873aa82441be673efab8ccd54b7a5511d0be72db6` | `b3a971fe2ba4be0f7d6cc659600560a9236f40fc0ebdfe1b098a899dfcf4ff40` | differs, as expected |

Rendered side by side, with the per-pixel absolute difference on the right. Both heatmaps use the **same color scale** (`0` → `0.1934`), so their shading is directly comparable.

Gate on — the difference image is empty because every element is exactly equal:

| A alone (BS1) | A inside BS4 `[B, A, D, C]` | \|diff\| heatmap |
|---|---|---|
| gate on, A alone | gate on, A inside BS4 | gate on, absolute difference |

Gate off — same prompt, same seed, same recipe, batch composition the only variable:

| A alone (BS1) | A inside BS4 `[B, A, D, C]` | \|diff\| heatmap |
|---|---|---|
| gate off, A alone | gate off, A inside BS4 | gate off, absolute difference |

| | gate on | gate off |
|---|---|---|
| differing elements | 0 / 786,432 (0.0000%) | 646,539 / 786,432 (82.2117%) |
| max abs diff | 0.000000 | 0.193359 |
| mean abs diff | 0.00000000 | 0.00775459 |
| max pixel diff (uint8) | 0 / 255 | 25 / 255 |
| visibly different pixels | 0.00% | 61.33% |

The gate-on tensor is not degenerate: `std = 0.4988` across 1,974 distinct sampled values, so "identical" is not two constant images agreeing.

Sampling self-check: both groups enumerated all four paths of their BS4 run and got `distinct=4/4`, with the A path matching the hashes above character for character, so the comparison is not reading the wrong sample. At the operator level, inside the same worker process, one row multiplied alone versus the same row multiplied inside a batch gave `max_abs_diff=1.0` (not equal) before `init_batch_invariance()` and `0.0` (equal) after. Environment: RTX 4090 24564 MiB, `get_device_capability(0) = [8, 9]`, `is_device_capability_family(80) = True` (both printed at runtime, not inferred), driver 580.105.08, torch 2.11.0+cu130, vllm 0.25.0.

### Not covered yet: multi-GPU and video

No experiment was run for either. This is a scope statement, not a measured negative result.

**Multi-GPU** — every run above used `num_gpus=1`, so no collective communication was exercised. Upstream's own batch-invariance CI also runs at TP=1 only, and batch invariance at TP>1 is still an open upstream bug (vllm#51290). A run today would most likely report "not invariant" without being able to attribute it, so multi-GPU stays unverified in both directions.

**Video** — upstream's overrides cover matmul-shaped operators plus softmax and `mean.dim`; there is no batch-invariant convolution or `group_norm` override at all. Video pipelines are built on `Conv3d`, where the time dimension enlarges the reduction and pushes implementations toward shape-dependent tiling. A convolution is not disqualifying by itself — SD3's verified path already contains a `Conv2d` patch embed whose reduction extent is independent of batch composition — but the 3D case can only be settled once upstream has batch-invariant convolution kernels.


hsliuustc0106:
This PR touches **vllm_omni/diffusion/, tests/diffusion/, docs/, docs/features/, tests/engine/** (9 files). Based on CODEOWNERS coverage of the changed files, the most-related reviewers appear to be:

> @david6666666 @yenuo26 @NickCao

vllm_omni/diffusion/request.py
```diff
@@ -84,6 +86,15 @@ def __post_init__(self):
...
             raise ValueError("OmniDiffusionRequest.request_id must be a non-empty string.")
 
+        self.seed_was_explicit = self.sampling_params.seed is not None
+
+        # Must run after seed_was_explicit above, which it reads, and before the
+        # fallback below, which would otherwise hide a missing seed behind a random
+        # one. Internal warmup requests are exempt: they are constructed by the
+        # engine itself and legitimately rely on that fallback.
+        if not self.is_dummy_run():
+            validate_batch_invariant_diffusion_request(self)
```
lishunyang12, line 96: Validate external requests only; Helios rebuilds requests after the runner initializes their generators.

Ronnie-Rui: Thanks @lishunyang12! Moved validation to request admission so internal requests can be rebuilt after generator initialization. Also preserved explicit-seed checks.

PR #5904, opened 2026-08-07
[Model] Optimize MiniCPM-o 4.5 initial codec chunk

0xtoward:
## Purpose

Similar to Qwen3-omni, use a smaller codec chunk for the first audio payload while keeping the steady-state chunk size unchanged. 

## Implement and Test
I performed multiple paired experiments, fresh-service for 1)candidate frame-size and 2)their adjacent 25-frame baselines. Each arm contains 2 warmups and 32 measured Seed-TTS requests at C=1 as the bench.

| Initial frames | Mean TTFT, control → candidate (ms) | Mean audio TTFP, control → candidate (ms) | TTFP change | Mean RTF, control → candidate | RTF change | Mean E2E, control → candidate (ms) | Mean audio duration/request | Total audio/arm | Output tokens/arm | Success |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 5  | 326.71 → 324.65 | 1122.01 → 877.01 | **-21.84%** | 0.58067 → 0.58069 | +0.00% | 2483.55 → 2492.88 | 4.34375 s | 139.0 s | 481 | 64/64 |
| 8  | 319.99 → 320.70 | 1057.04 → 869.61 | **-17.73%** | 0.47955 → 0.47488 | -0.97% | 2036.71 → 2019.92 | 4.34375 s | 139.0 s | 481 | 64/64 |
| 12 | 320.75 → 322.51 | 1044.33 → 922.84 | **-11.63%** | 0.47126 → 0.48651 | +3.24% | 2000.29 → 2066.53 | 4.34375 s | 139.0 s | 481 | 64/64 |
| 16 | 305.26 → 317.91 | 1023.01 → 948.99 | **-7.24%** | 0.46791 → 0.47176 | +0.82% | 1980.13 → 2001.86 | 4.34375 s | 139.0 s | 481 | 64/64 |

The previous perf shaking is caused by CPU bound, which can be resolved after setting CPU_AFFINITY_CONF=1. 

@lishunyang12 @amy-why-3459 Now there is only 1 line changed for the default size from 12 to 5 (5 is the smallest frame size where the HiFT will keep 160ms PCM, making 4-frame output nothing)

## Future
Add this size to cuda graph batch size.

PR #5938, opened 2026-08-09
[Perf][MiniCPM-o] Add configurable first codec chunk threshold to cut TTFP

worryzyy:
## Purpose

Reduce time-to-first-packet (TTFP) for MiniCPM-o 4.5 streaming speech without touching steady-state throughput.

The Talker→Code2Wav async chunking currently waits for a full `codec_chunk_frames` (default 25 frames = 1 s of audio) before emitting the first chunk, so the first audio packet cannot ship until the Talker has produced 25 codec tokens **and** the vocoder has processed all of them.

This PR introduces an optional `initial_codec_chunk_frames` key in the shared-memory connector `extra`: only the **first** emission of a request may use a smaller threshold; all subsequent chunks keep the steady-state size. The effective value is `min(initial_codec_chunk_frames, codec_chunk_frames)`, and the **default equals `codec_chunk_frames`, so behavior is unchanged unless explicitly configured**. This mirrors the existing `initial_codec_chunk_frames` mechanism in fish_speech.

The first-chunk threshold and the steady-state chunk size become independently tunable: deployments can ship the first packet after e.g. 8 frames (0.32 s of audio) while keeping — or even enlarging — the steady-state chunk for RTF.

## Test Plan

**vLLM Version:** 0.1.dev1+ge5588e49b (as shipped in `quay.io/ascend/vllm-omni:v0.25.0-a3`)

**vLLM-Omni Commit:** benchmarked on `minicpm-challenge` @ `1be47550`; patch re-applies cleanly on current HEAD `009b80d`

Setup: Ascend 910C (Atlas A3), single card, exclusive; deploy config `vllm_omni/deploy/minicpmo_4_5.yaml` with only `initial_codec_chunk_frames: 8` added (A/B on the same machine, same restart procedure, 3 warmup requests, single variable isolated).

Benchmark: `vllm bench serve --omni --backend openai-chat-omni --dataset-name seed-tts --seed-tts-locale en --num-prompts 32 --max-concurrency 1 --no-oversample` with `{"modalities":["text","audio"],"chat_template_kwargs":{"enable_thinking":false,"use_tts_template":true}}`, 3 rounds each.

Quality regression: seed-tts zh full set (2000 utterances), seed-tts-eval protocol (funasr paraformer-zh CER), same server/config.

## Test Result

Concurrency 1 (3-round means, same machine A/B):

| Metric | baseline (unmodified yaml) | `initial_codec_chunk_frames: 8` | Δ |
|---|---|---|---|
| Mean AUDIO_TTFP | 1049.04 ms | **861.86 ms** | **-17.8%** |
| Mean AUDIO_RTF | 0.4567 | 0.4600 | +0.7% (noise) |
| Mean TTFT | 316.27 ms | 315.99 ms | -0.1% |
| Streaming continuity | 100% | 100% | — |
| Requests | 96/96 ok | 96/96 ok | — |

Quality (seed-tts zh, 2000 utterances, CER): **1.377%** with the feature enabled (F16 reference 1.414%); speaker-similarity proxy (WavLM) 0.8475; zero failed requests.

Known trade-off (documented for completeness): with a small first chunk **and** the default 25-frame steady-state chunk, concurrency-8 RTF degrades ~13% because the vocoder receives more calls under saturation; pairing the feature with a larger steady-state chunk (e.g. 75) removes this and improves RTF at all tested concurrencies. The feature itself is off by default.

---

**参赛队伍：味蕾**（MiniCPM × 昇腾挑战赛 · 赛道一 · 子赛道 B / vLLM-Omni）

### DIFF
--- a/vllm_omni/model_executor/stage_input_processors/minicpmo_4_5_omni.py
+++ b/vllm_omni/model_executor/stage_input_processors/minicpmo_4_5_omni.py
@@ -125,20 +125,25 @@ def _coerce_int(value):
         return None
 
 
-def _codec_config(transfer_manager: Any) -> tuple[int, int]:
+def _codec_config(transfer_manager: Any) -> tuple[int, int, int]:
     connector = getattr(transfer_manager, "connector", None)
     raw_config = getattr(connector, "config", {}) or {}
     config = raw_config.get("extra", raw_config) if isinstance(raw_config, dict) else {}
     config = config if isinstance(config, dict) else {}
     chunk_frames = int(config.get("codec_chunk_frames", 25))
     left_context_frames = int(config.get("codec_left_context_frames", 3))
-    if chunk_frames <= 0 or left_context_frames < 0:
+    # Optional smaller threshold for the very first emitted chunk of a request:
+    # lowers time-to-first-packet without touching steady-state chunk size
+    # (same idea as fish_speech ``initial_codec_chunk_frames``).
+    first_chunk_frames = int(config.get("codec_first_chunk_frames", chunk_frames))
+    if chunk_frames <= 0 or left_context_frames < 0 or first_chunk_frames <= 0:
         raise ValueError(
             "Invalid MiniCPM-o codec chunk config: "
             f"codec_chunk_frames={chunk_frames}, "
+            f"codec_first_chunk_frames={first_chunk_frames}, "
             f"codec_left_context_frames={left_context_frames}"
         )
-    return chunk_frames, left_context_frames
+    return chunk_frames, left_context_frames, min(first_chunk_frames, chunk_frames)
 
 
 def _codec_scalars(value: Any) -> list[int]:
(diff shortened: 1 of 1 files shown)


vllm_omni/model_executor/stage_input_processors/minicpmo_4_5_omni.py
```diff
@@ -125,20 +125,25 @@ def _coerce_int(value):
...
     connector = getattr(transfer_manager, "connector", None)
     raw_config = getattr(connector, "config", {}) or {}
     config = raw_config.get("extra", raw_config) if isinstance(raw_config, dict) else {}
     config = config if isinstance(config, dict) else {}
     chunk_frames = int(config.get("codec_chunk_frames", 25))
     left_context_frames = int(config.get("codec_left_context_frames", 3))
-    if chunk_frames <= 0 or left_context_frames < 0:
+    # Optional smaller threshold for the very first emitted chunk of a request:
+    # lowers time-to-first-packet without touching steady-state chunk size
+    # (same idea as fish_speech ``initial_codec_chunk_frames``).
```
NickCao, line 137: Then shall we not also name it `initial_codec_chunk_frames` for consistency?

worryzyy: Good point — renamed to `initial_codec_chunk_frames` in 266da17 (and the stacked config PR #5939 is updated to match). Thanks for the review!

Junbo2002:
#5904 and #5938 implement the same idea, so I ran them **head to head**: same
machine, same day, same protocol, same control arm, same value
(`initial_codec_chunk_frames: 5` — #5904 ships that default, #5938 ships the
code without a default, so I set the same value for it). Both arms are
`ecd9d99da` differing only by the PR; two arms at once, one per die, each
round repeated with the dies swapped.

| PR | round | ΔRTF | ctl TTFP | PR TTFP | **ΔTTFP** |
| --- | --- | ---: | ---: | ---: | ---: |
| #5904 | 1 | −3.73% | 1019.3 | 801.5 | **−217.8 ms** |
| #5904 | 2 | −2.11% | 1031.2 | 829.3 | **−201.9 ms** |
| #5938 | 1 | −2.22% | 1023.1 | 798.0 | **−225.1 ms** |
| #5938 | 2 | −4.89% | 1025.5 | 816.3 | **−209.2 ms** |

32/32, 435 output tokens, 135.64 s of audio in all eight arms.

1. **On performance the two are indistinguishable** — all four ΔTTFP land in a
   −202…−225 ms band and the ranges overlap completely.
2. **The difference is duplex.** #5904 counts with `segment_chunks_emitted`
   and resets it at the end of a segment, so every turn of a duplex session
   gets the smaller first chunk (its
   `test_initial_chunk_resets_at_duplex_segment_boundary` covers this).
   #5938 tests `int(state["codec_end"]) == 0`, a request-lifetime cursor, so
   after the first segment it never applies again.
3. #5904 ships the default in the deploy YAML; #5938 ships the mechanism with
   none, so nothing turns it on unless the operator knows the key exists.

PR #5940, opened 2026-08-09
[Model] Add sequence parallelism to SANA-Video 2B

cr-gao:
> **Built on #5861** (TP/CFG parallel, merged 2026-09-02). This branch is
> updated with `main` after that merge, so the diff below shows only the SP
> change itself (`transformer_sana_video.py` + one pipeline class attribute +
> tests + docs).
>
> **Update 2026-09-02**: #5861 merged and this branch is merged with `main`
> (`a36b8dae2`); the full local SANA suite passes on the merged tree
> (190 tests, vllm 0.28).
>
> **Update 2026-08-25**: #5508 merged; the stack is rebased onto current `main`
> with conflicts resolved. Post-rebase sanity on the same 4×A800 box: the
> in-PR fp32 multi-process numeric gate passes (10 tests), sp1 is **bitwise
> identical** to merged `main`'s single-GPU output (1-step, real weights),
> sp2 / sp2×cfg2 sit at rel_l2 2.3e-2 / 2.7e-2 (the expected bf16 magnitude
> below), and the tiny SP/CFG L2 groups pass (11 cases). Upstream's new
> dummy-run frame handling keeps working with the `dummy_run_num_frames = 25`
> warmup contract.

## Purpose

Add sequence parallelism (degrees 1/2/4) to the native SANA-Video 2B
transformer, for both T2V and I2V, combinable with CFG parallel (SP2 x CFG2 on
4 GPUs).

SANA-Video's backbone is not a softmax DiT, so the generic SP paths don't fit:

- Self-attention is ReLU **linear attention**: the only sequence coupling is
  two sums over tokens (`k_sum` and the d x d KV state). Ulysses all-to-all
  would give up the linear attention's O(N) memory advantage (and has no mount
  point — this attention intentionally bypasses the unified attention layer);
  Ring's online-softmax LSE merge does not exist for linear attention.
- The generic `_sp_plan` hooks split on even token boundaries; SANA needs
  frame-aligned shards for its per-frame spatial convs and the GLUMB temporal
  conv, and the default frame counts (F'=21 at 480p, F'=11 at 720p) are odd,
  so shards are necessarily uneven. Hook `auto_pad` would feed pad tokens into
  the linear attention's global sums with no mask to hide them.

The implementation is therefore **manual SP inside the transformer** (the
framework's documented manual-SP path, as in hunyuan_video_15 / cosmos3 /
wan2_2_s2v): balanced frame-aligned sharding, one small packed all-reduce per
block for the attention state (95 µs measured at sp2), boundary-frame halo
exchange for the temporal conv (a neighbor-only batched isend/irecv since
`725696608`), and a single gather after `proj_out`. Measured with
torch.profiler on 4×A800: 5.8 ms/step of collective kernels (0.5% of GPU
busy) at `725696608`, 8.5 ms/step with the earlier all_gather halo, vs
143 ms/step (9.6%) for TP2, roughly 25× less collective time per step.
(This replaces an earlier ~50× byte estimate that did not survive
measurement.)

`--usp` carries the degree; the docs state explicitly that the implementation
is frame sharding + state reduction, not Ulysses all-to-all.

## Scope

- `transformer_sana_video.py` only (+ tests/docs): `_sp_frame_split_sizes`,
  `_sp_gather_frames`, attn1 packed state all-reduce, conv_temp halo,
  forward shard/gather, I2V 5D per-token timestep frame slicing,
  validation allowlist, `_sp_plan = {}` registry wiring,
  `dummy_run_num_frames = 25` (engine warmup must satisfy F' >= degree).
- Pipelines, scheduler, CFG mixin, sp_sharding, hooks, registry, vllm layers:
  untouched.
- Fail-closed: ring/AllGather-KV SP, SP8. TP x SP combines since `5ed4aa75f`
  (validation and numbers below).
- Checkpoint keys unchanged; SP1 is bit-identical by construction (no new
  branch entered at world size 1).

## Correctness evidence

1. **L1 (CPU, mocked collectives)**: 27 unit tests — balanced split sizes
   (chunk's ceil semantics produce <P chunks for F'=5/6/9 at P=4, hence the
   local helper), attn1 state-reduction parity + single packed collective,
   halo off-by-one teeth, uneven gather roundtrip with NaN pad canaries, I2V
   timestep slice parity, full-model lockstep-thread SP2 parity (t2v + i2v).
2. **Real multi-process numeric gate (in this PR,
   `tests/diffusion/distributed/test_sana_video_sp_numeric.py`)**: tiny
   transformer, real NCCL groups, fp32: `rel_l2(SP=N vs SP=1) <= 1e-5`
   asserted for sp2/sp4 x {t2v, i2v} x F' in {21, 5}, plus sp2 x cfg2 on 4
   processes; 10 passed on 4x A800. Self-contained (no external baseline) and
   collected by the existing nightly `tests/diffusion/distributed/` sweep —
   no CI wiring change needed.
3. **L2 tiny e2e (smoke)**: new `[sequence_parallel]` and
   `[sequence_parallel+cfg_parallel]` groups for T2V and I2V — full engine
   chain, real weight loading, no hang; 11/11 SANA cases pass on 4x A800.
4. **Real weights (4x A800, 1-step comparison vs pre-implementation HEAD
   baseline regenerated on the same machine, CUDNN_ATTN and TORCH_SDPA)**:
   - sp1: **bitwise identical** in all 4 cells x 2 backends (max_abs = 0).
   - sp2 / sp4 / sp2 x cfg2 (480p-t2v, 720p-i2v): rel_l2 1.7e-2 – 2.6e-2,
     SSIM 0.98 – 0.995 — same order as the TP2 drift accepted in #5861
     (bf16 reduction reordering; 20 passed).
   - 50-step, informational only (this box's same-code same-seed reruns
     already differ at SSIM ~0.971 over 50 steps, so no threshold is
     claimed): sp2 vs pre-impl sp1 480p t2v = SSIM 0.9754 / PSNR 32.3dB,
     i.e. at the machine's own reproducibility noise floor; frames visually
     identical with no shard-boundary artifacts.

## Benchmark (4×A800 NVLink, 480p T2V, TORCH_SDPA)

DiT-loop(50) from the 10/20-step slope, protocol identical to #5861.
Measured at the current head `5ed4aa75f` (post merge with `main`), all rows
in one session on the same box; torch 2.13.0+cu130, vLLM 0.28.0, checkpoint
revision `fed3bce4`, 81 frames, guidance 6.0, seed 42. Raw per-run times
listed so the spread is visible (≤1% per cell). The earlier session at
`22aac925f` (vLLM 0.26, with a `main` tp1 row) agreed within 2% per row:
sp2 1.75×, sp4 3.24×, sp2×cfg2 3.56×.

| topology | cards | 10-step runs (s) | 20-step runs (s) | DiT-loop (50) | speedup vs sp1 |
|---|---|---|---|---|---|
| sp1 | 1 | 27.92 / 27.93 / 27.95 | 50.32 / 50.35 / 50.33 | 112.0 s | 1.00×¹ |
| sp2 | 2 | 18.47 / 18.46 / 18.54 | 31.04 / 31.05 / 31.07 | 62.9 s | **1.78×** |
| tp2×sp2 | 4 | 14.61 / 14.64 / 14.66 | 23.25 / 23.29 / 23.25 | 43.1 s | **2.60×** |
| sp4 | 4 | 12.80 / 12.80 / 12.82 | 19.53 / 19.55 / 19.56 | 33.8 s | **3.32×** |
| sp2×cfg2 | 4 | 12.31 / 12.24 / 12.33 | 18.57 / 18.57 / 18.55 | 31.3 s | **3.58×** |

¹ No serial regression: in the `22aac925f` session sp1 was ~1.4% faster than
the `main` tp1 baseline measured alongside it (112.2 vs 113.9 s), and at the
merged head the serial 1-step output is bitwise identical to the pre-merge
`main` baseline. The serial path is shared with #5861, where an interleaved
A/B and a per-kernel trace diff cover this in detail.

Reference (#5861, measured in the `22aac925f` session): tp2 1.53×, cfg2
2.06×, tp2×cfg2 2.96×; sp2 beats tp2 at 2 cards, and sp2×cfg2 is the best
4-card configuration measured.

- Numbers are NVLink; PCIe was not measured. Profiled collective time at
  `22aac925f` is 8.5 ms/step for sp2 (0.7% of GPU busy) vs 143 ms/step for
  tp2 (9.6%). Re-measured at `725696608` after the neighbor P2P halo change:
  the sp2 collective time drops to 5.8 ms/step (0.5%), sp2×cfg2 from
  4.7 ms/step to 3.6 ms/step.
- TP × SP combines since `5ed4aa75f`. Validation on 4×A800: the fp32
  multi-process gate gained a tp2×sp2 case (11 tests green), a 1-step real-weights run sits at rel_l2 2.5e-2
  vs serial (the same bf16 magnitude as sp2), and the tiny e2e matrix gained a
  TP×SP group. Its performance is a row in the table above (2.60×, below sp4
  and sp2×cfg2), so the pure configurations remain the recommendation when
  four GPUs serve one request; tp2×sp4 is the 8-GPU option when
  guidance_scale is 1.

## Test commands

```bash
# L1 (CPU, no GPU needed)
pytest -sv tests/diffusion/models/sana_video/ -m "core_model and cpu"

# Multi-GPU numeric gate (2/4 GPUs; collected by the existing nightly
# tests/diffusion/distributed sweep)
pytest -sv tests/diffusion/distributed/test_sana_video_sp_numeric.py

# L2 tiny e2e SP groups (4 GPUs)
pytest -sv tests/model_tests/diffusion/test_common_offline.py \
  -k "(SanaVideoPipeline or SanaImageToVideoPipeline) and sequence"
```


hsliuustc0106:
This PR touches **tests/diffusion/, vllm_omni/diffusion/, tests/model_tests/, examples/online_serving/, tests/e2e/** (56 files). Based on CODEOWNERS coverage of the changed files, the most-related reviewers appear to be:

> @yenuo26 @NickCao @hsliuustc0106

cr-gao:
@hsliuustc0106 @yenuo26 @NickCao Rebased onto current `main` after the #5508 merge and marked ready. The diff still includes #5861's 10 commits until it merges — the SP change itself is the last 5 commits. Post-rebase sanity on 4×A800: the in-PR fp32 multi-process numeric gate passes (10 tests), sp1 bitwise-identical to merged `main`, sp2/sp2×cfg2 within expected bf16 drift; benchmark evidence unchanged (sp2 1.80×, sp2×cfg2 3.57×). Best reviewed after #5861. Thanks!

cr-gao:
Self-review — what I checked:

- **SP1 is bit-identical by construction**: no new branch is entered at world size 1 (asserted by a no-collective unit test), and sp1 real-weight output is bitwise-equal to the serial baseline (re-verified 1-step against current `main` post-rebase).
- **Numeric equivalence**: the in-PR multi-process fp32 gate (`test_sana_video_sp_numeric.py`) asserts rel_l2 ≤ 1e-5 for sp2/sp4 × {t2v, i2v} × F' ∈ {21, 5} plus sp2×cfg2 — 10 passed on real NCCL groups, 4×A800.
- **Sharding invariants**: balanced frame split (torch.chunk's ceil semantics would under-produce chunks for F'=5/6/9 at P=4 — covered), attn1 single packed fp32 all-reduce (shape/dtype asserted), conv_temp halo off-by-one teeth, uneven gather roundtrip with NaN-pad canaries, I2V 5D timestep frame slicing parity.
- **Engine warmup**: `dummy_run_num_frames = 25` keeps F' ≥ degree for both VAE variants; verified this still composes with upstream's new `resolve_video_num_frames` dummy-run handling after the rebase.
- **Regression**: L1 189 passed; tiny SP/CFG L2 groups 11 passed on 4×A800; benchmark evidence in the description is from real weights on NVLink (sp2 1.80×, sp2×cfg2 3.57×).

docs/user_guide/diffusion_features.md
```diff
@@ -162,7 +162,7 @@ The following tables show which models support each feature:
 | **LTX-2**                    |     ❌     |     ✅      |           ✅           |       ✅        |         ✅         |         ❌         |   ✅    |             ✅             |      ✅ (decode)      |       ❌        |        ❌         |
 | **LTX-2.3**                  |     ❌     |     ✅      |           ✅           |       ✅        |         ✅         |         ❌         |   ✅    |             ✅             |      ✅ (decode)      |       ❌        |        ❌         |
 | **LTX-2.5**                  |     ❌     | ❓ (one-stage) | ❓ (Ulysses only) | ✅ (Full only) |         ❓         |         ❌         |   ✅    |             ✅             |      ✅ (decode)      |    ❓ (FP8)     |        ❌         |
-| **SANA-Video-2B T2V I2V**   |     ❓     |     ❓      |           ❓           |       ❓        |         ❓         |         ❓         |   ❓    |             ❓             |          ❓           |       ❓        |        ❌         |
+|  **SANA-Video-2B T2V I2V**   |     ❌     |     ❌      |   ✅ (Ulysses only)   |       ✅       |  ✅ (TP=2 only)   |         ❌         |   ❌   |            ❌             |          ❌          |       ❌       |        ❌        |
```
hsliuustc0106: nit: "✅ (Ulysses only)" reads as Ulysses all-to-all support, but this PR deliberately implements frame sharding + state reduction instead (the recipe is explicit about it). Maybe "✅ (`--usp`, frame-sharded)" so the matrix doesn't contradict the serving docs?

hsliuustc0106:
can you post the compute and comm time decomposition? in addution, is there any host launch free time?

cr-gao:
@hsliuustc0106 Measured with per-rank torch.profiler traces (10-step 480p request):

| topo | comm kernels /step | % of GPU busy |
|---|---|---|
| sp2 | 8.5 ms | 0.7% |
| sp2×cfg2 | 4.7 ms | 0.7% |
| tp2 (#5861) | 143 ms | 9.6% |

The gap is structural: TP all-reduces every block's output (60 × 294 MB per step, sizes read from a record_shapes trace), while SP only does one small state all-reduce plus a boundary-frame halo per block — ~17× less collective time. I've also corrected the PR description: the earlier "~0.6 GB vs ~29 GB (~50×)" was an estimate that doesn't survive measurement, and it now quotes these measured numbers instead.

On host launch free time: inside the steady denoise loop the GPU only idles in micro launch gaps (0.8% of wall serial, ~1.4% at two ranks); every gap ≥1 ms sits in the first ~0.35 s of a request and the CPU timeline shows it's one-time NCCL communicator setup, not per-step. So CUDA graphs would buy ~1–2% at most and I'd leave that out of this PR.

docs/user_guide/diffusion_features.md
cr-gao: Reworded to `✅ (--usp, frame-sharded)` in the doc commit.

hsliuustc0106: Verified, reads clearly now. Thanks!

vllm_omni/diffusion/models/sana_video/transformer_sana_video.py
```diff
@@ -404,12 +553,30 @@ def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
...
+            # Every rank contributes its (first, last) frame pair; the global
+            # first/last rank keeps the dense zero padding on its outer side.
+            x = hidden_states_temporal
+            boundary = torch.cat([x[:, :, :1], x[:, :, -1:]], dim=2)
+            parts = get_sp_group().all_gather(boundary, dim=0, separate_tensors=True)
+            rank = get_sequence_parallel_rank()
+            world_size = get_sequence_parallel_world_size()
+            zero = torch.zeros_like(boundary[:, :, :1])
+            left = parts[rank - 1][:, :, 1:2] if rank > 0 else zero
+            right = parts[rank + 1][:, :, 0:1] if rank < world_size - 1 else zero
```
nagisa-kunhah: Each rank only needs the last frame from rank - 1 and the first frame from rank + 1, but the current implementation uses all_gather(), causing every rank to receive boundary frames from all ranks. This introduces unnecessary communication and temporary tensor materialization, especially as the SP degree increases. 

Could we replace this with a neighbor-only P2P halo exchange, similar to the existing Wan spatial-shard implementation?
https://github.com/vllm-project/vllm-omni/blob/be335a86f9af5bde108e11fb3e01d729553a3532/vllm_omni/diffusion/distributed/autoencoders/wan_spatial_shard.py#L259-L268

tests/diffusion/models/sana_video/test_transformer_sana_video_sp.py
```diff
@@ -0,0 +1,486 @@
...
+        outs = list(pool.map(run_rank, range(world_size)))
+
+    num_layers = _TINY_SP_CONFIG["num_layers"]
+    # One attn1 state reduce per block, its inputs distinct per rank (the ranks
+    # really processed different frames), one halo gather per block plus the
+    # final frame gather.
+    assert len(group.reduce_inputs) == num_layers
+    for parts in group.reduce_inputs:
+        assert not torch.equal(parts[0], parts[1])
+    assert group.gather_calls == num_layers + 1
```
nagisa-kunhah: Following the all-gather issue mentioned above, this test should also be updated.

vllm_omni/diffusion/models/sana_video/pipeline_sana_video.py
```diff
@@ -294,6 +300,9 @@ class SanaVideoPipeline(
     _encoder_modules = ["text_encoder"]
     _vae_modules = ["vae"]
     supports_step_execution = False
+    # Warmup must yield >= 4 latent frames on both VAE variants (temporal /4
+    # and /8) so every allowed SP degree has one frame per rank.
+    dummy_run_num_frames = 25
```
nagisa-kunhah, line 338: [Minor] The warmup currently always uses 25 dummy frames. However, the minimum number of frames required depends on the configured SP degree and VAE temporal compression ratio. Could we derive this value from the current configuration instead of always using the SP4 worst-case value?

vllm_omni/diffusion/models/sana_video/transformer_sana_video.py
cr-gao: Agreed. Replaced the all_gather with a neighbor-only halo exchange using batched isend/irecv pairs, following the Wan spatial-shard pattern (725696608). Each rank now sends one boundary frame to each neighbor and receives only the two frames it uses, instead of materializing all boundary frames from every rank. The real multi-process fp32 equivalence gate passes unchanged at sp2 and sp4.

tests/diffusion/models/sana_video/test_transformer_sana_video_sp.py
cr-gao: Updated together with the change: the mocks now stand in for the batched P2P exchange (mailbox plus barrier in the lockstep test, two-pass replay in the unit test) and assert the halo talks to the two neighbors only, with the final frame gather left as the single full collective.

vllm_omni/diffusion/models/sana_video/pipeline_sana_video.py
cr-gao: The constraint is that dummy_run_num_frames is read as a class attribute by the shared warmup helper before any parallel config reaches the pipeline, so deriving it per configuration needs a signature change in shared code, which I have kept out of this PR. Warmup runs once per engine start, so I kept the SP4 worst case (25 pixel frames, the minimum would be 13 at sp4 with the /4 VAE). Happy to plumb the config through the shared helper here or in a follow-up if you think it is worth it.

vllm_omni/diffusion/models/sana_video/transformer_sana_video.py
cr-gao: Measured the effect on the same 2-rank 10-step trace protocol: collective kernel time per step drops from 8.5 ms to 5.8 ms (0.5% of GPU busy). The halo is now one batched SendRecv per block, and the only remaining all_gather is the final frame gather. End-to-end times are unchanged within noise, as expected at this comm share.

cr-gao:
@nagisa-kunhah All three of your comments are addressed: the halo exchange is now a neighbor-only batched isend/irecv (725696608), the mock test drives the real P2P path, and the warmup constraint is explained in the thread. The fp32 equivalence gate and tiny e2e are green on 4x A800, and CI is green. Please take another look when you have time.

cr-gao:
Updated this branch with main as well. Note for the merge commit: the auto merge silently duplicated the pre-SP parallel-config validator next to the SP-enabled one (the merge base predates #5861, so both sides added the function), and the duplicate shadowed the SP version. The merge keeps the SP validator only; docs and test lists are union merged. Full local SANA suite on the merged tree: 190 passed on vllm 0.28.

cr-gao:
Resolved the conflict with main after #5882 merged. All hunks were union style: imports and the Cache-DiT adapter attributes in the transformer, `dummy_run_num_frames` next to `default_num_inference_steps` in the pipeline, test groups, and the recipe limitation notes. Full local SANA suite on the merged tree: 214 passed on vllm 0.28.

cr-gao:
Re-validated the merged head `e41a7e2bb` on the 4x A800 box with vllm 0.28.0 (torch 2.13): the in-PR fp32 multi-process numeric gate passes (10 tests), the 1-step sp2 output sits at rel_l2 2.3e-2 against the same-tree serial run (the same value as before the merges), the serial 1-step output is bitwise identical to the pre-merge main baseline, and the tiny e2e SANA groups pass (16 cases plus 18 subtests, now including the cache and offload groups from #5882). The full local SANA suite also passes on the box (214 tests).

cr-gao:
@hsliuustc0106 @nagisa-kunhah @congw729 Re-validated the merged head on 4x A800 with vllm 0.28.0 (torch 2.13): the in-PR fp32 multi-process numeric gate passes, the 1-step sp2 output sits at rel_l2 2.3e-2 against the same-tree serial run (the same value as before the merges), the serial 1-step output is bitwise identical to the pre-merge main baseline, and the tiny e2e SANA groups pass. The full local SANA suite also passes on the GPUs (214 tests). Ready for merge!

vllm_omni/diffusion/models/sana_video/transformer_sana_video.py
```diff
@@ -521,7 +563,33 @@ def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
         hidden_states_temporal = hidden_states.view(batch_size, num_frames, num_channels, height * width).permute(
             0, 2, 1, 3
         )
-        hidden_states = hidden_states_temporal + self.conv_temp(hidden_states_temporal)
+        if get_sequence_parallel_world_size() > 1:
+            # The temporal conv reads one frame across each shard boundary, so
```
wtomin, line 562: In the SP path of `GLUMBTempConv`, each rank exchanges only its first/last local frame with a neighbor. That is correct only when `conv_temp` has a temporal receptive field of exactly three frames with zero padding of one on each side. 

I understand it is correct as it's hard-coded:
```python
        self.conv_inverted = nn.Conv2d(in_channels, hidden_channels * 2, 1, 1, 0)
        self.conv_depth = nn.Conv2d(hidden_channels * 2, hidden_channels * 2, 3, 1, 1, groups=hidden_channels * 2)
        self.conv_point = nn.Conv2d(hidden_channels, out_channels, 1, 1, 0, bias=False)
```
But I also consider future extensions. Is this kernel fixed by the architecture, or should there be an assertion/config guard so a future change to the temporal kernel width cannot silently break SP parity?

vllm_omni/diffusion/models/sana_video/transformer_sana_video.py
```diff
@@ -621,9 +689,15 @@ def apply_rotary_emb(
...
         # intentionally matches Diffusers 0.38.0's SanaLinearAttnProcessor3_0;
         # only the rotated numerator path is accumulated in FP32.
-        z = 1 / (key.sum(dim=-1, keepdim=True).transpose(-2, -1) @ query + 1e-15)
-
+        k_sum = key.sum(dim=-1, keepdim=True)
         scores = torch.matmul(value, key_rotate.transpose(-1, -2))
+        if get_sequence_parallel_world_size() > 1:
+            # Token sums are the only sequence coupling; one packed all-reduce
+            # turns the per-rank partial state into the global one.
+            packed = torch.cat([scores, k_sum.float()], dim=-1)
```
wtomin, line 696: In the attention state reduction around new lines ~695–710, `k_sum.float()` causes the packed `[scores, k_sum]` all-reduce to be FP32 even when the model runs in BF16. Is that intentionally for numerical stability?

vllm_omni/diffusion/models/sana_video/transformer_sana_video.py
```diff
@@ -46,12 +52,22 @@ def validate_sana_video_parallel_config(parallel_config) -> None:
...
+            "log-sum-exp state to merge across ring stages. Use --usp for sequence parallel."
+        )
+    if parallel_config.allgather_degree > 1:
+        raise NotImplementedError("SANA-Video does not support AllGather-KV sequence parallel. Use --usp instead.")
+    if tp_size > 1 and sp_size > 1:
         raise NotImplementedError(
-            "Sequence parallel is not supported for SANA-Video: its linear attention and "
-            "GLUMB temporal conv have no SANA-specific SP implementation. Use tensor and/or "
-            "CFG parallel instead."
+            "SANA-Video does not support combining tensor parallel with sequence parallel. "
```
wtomin: Why TP + SP combination is not supported?
> A tp2×sp2 spike (validator temporarily relaxed for the measurement, then
restored) taken earlier at the pre-rebase head reached 2.55× — clearly
worse than pure-SP 4-card configurations, so the combination stays
fail-closed.

In the PR body, I think you might tested `tp2×sp2` once. How is the performance?

vllm_omni/diffusion/models/sana_video/transformer_sana_video.py
cr-gao: The kernel is fixed by the checkpoint architecture. `conv_temp` is created with `kernel_size=(3, 1)` (in GLUMBTempConv in a few lines above), but you are right that nothing stops a future edit from breaking the single-frame halo silently. Added an assertion in the SP branch in 95c8f45f8.

vllm_omni/diffusion/models/sana_video/transformer_sana_video.py
cr-gao: Yes, it's intentional, and is actually a packing requirement. `scores` is already FP32 on the serial path (the rotated numerator is accumulated in FP32 to match Diffusers 0.38.0's SanaLinearAttnProcessor3_0), and the single fused all-reduce needs one dtype, so `k_sum` is upcast for the collective and cast back to the input dtype right after.

vllm_omni/diffusion/models/sana_video/transformer_sana_video.py
cr-gao: Yes I once measured and found that a tp2 x sp2 spike on the same 4x A800 box reached 2.55x, which I thought was not that good. But it makes sense to support it for any potential needs. The combination is now open in 5ed4aa75f: the fp32 gate and a 1-step real-weights check both pass, and a fresh same-session benchmark puts tp2 x sp2 at 2.60x. It is now a row in the benchmark table in the updated PR body.

cr-gao:
@hsliuustc0106 @wtomin Thanks for the review comments, I have addressed the concerns and updated the latest parity test and benchmark results.

PR #4428, opened 2026-06-15
[Bugfix] Fix Fish Speech kvcache attention KV cache unpacking for vLLM 0.23.0

nagelanping:
## Purpose

The Fish Speech Models decode-only kvcache fast path unpacked the paged KV cache with kv_cache.unbind(0), assuming the pre-0.23 layout where key/value live on dim 0. After the rebase to vLLM 0.23.0, the layout changed to (num_blocks, 2, block_size, num_kv_heads, head_size), placing key/value on dim 1. unbind(0) now splits along num_blocks and raises "ValueError: too many values to unpack (expected 2)" during CUDA graph capture (_dummy_run), crashing Stage 0 (FishSpeechSlowAR) engine init.

Switch to kv_cache.unbind(1) to match FlashAttentionImpl. Each unpacked cache is then 4-D (num_blocks, block_size, num_kv_heads, head_size), consistent with can_use_fish_kvcache_attn's shape guards.

### Tests

Added unit tests in `tests/attention/test_fish_kvcache_attn.py`:

- Updated the existing backend forward tests to build the KV cache with the vLLM 0.23.0 layout (a shared `_decode_kv_cache()` helper). They previously used the pre-0.23 layout and would not exercise the new `unbind(1)` path.
- `test_fish_kvcache_backend_unbinds_kv_on_vllm_cache_layout`: sources the cache shape from vLLM's own `FlashAttentionBackend.get_kv_cache_shape` and asserts `unbind(1)` yields the 4-D `(num_blocks, block_size, num_kv_heads, head_size)` tensors the decode kernel and guard expect — the dimensions are compatible with 0.23.0.
- `test_fish_kvcache_vllm_cache_unbinds_then_falls_back_until_contiguous`: end-to-end with the real (unmocked) guard, asserts `unbind(1)` no longer crashes and the request completes via the correct fallback.

### Known limitation (out of scope, tracked separately)

`unbind(1)` on the interleaved `(num_blocks, 2, ...)` layout returns **non-contiguous** key/value views. `can_use_fish_kvcache_attn` rejects them on its `is_contiguous()` check, so on 0.23.0 the request falls back to the original attention and the Fish kvcache fast path stays dormant. Output is correct (verified below), but the optimization does not engage. The dimensions line up — only contiguity blocks the fast path. Re-enabling it would require a `.contiguous()` copy per decode step with its own performance trade-off, so it is left for a follow-up issue. The second test above documents this behavior (asserts `guard_miss` fallback).

## Test Plan

[...]

PR #5943, opened 2026-08-09
[Kernel] Fix Fish attention fallback on strided KV cache

### LINKED ISSUE #5942: [Bug]: Fish kvcache fast path always falls back for packed strided vLLM cache
## Your current environment

- GPU: NVIDIA GeForce RTX 4090
- vLLM: 0.26.0
- vLLM-Omni commit: `f201b717`
- The issue is independent of loading the full Fish Speech checkpoint and can be reproduced with synthetic tensors.

## Code version

- vLLM-Omni: `f201b717` (`main`)
- vLLM: 0.26.0

## Describe the bug

The Fish Speech decode-only KV-cache attention backend correctly unpacks the current vLLM FlashAttention cache layout with:

```python
key_cache, value_cache = kv_cache.transpose(1, 2).split(head_size, dim=-1)
```

The packed cache has shape `(num_blocks, num_kv_heads, block_size, 2 * head_size)`. After `transpose(1, 2)` and `split`, K and V have the expected shape `(num_blocks, block_size, num_kv_heads, head_size)`, but both are non-contiguous strided views.

`can_use_fish_kvcache_attn()` currently requires K and V to be contiguous. Therefore the guard rejects the actual vLLM cache layout, every otherwise-compatible Fish decode request records a `guard_miss`, and the Fish Triton fast path always falls back to the original attention implementation.

This is the known limitation documented in #4428: that PR fixed cache unpacking after the vLLM layout change, but intentionally left support for the resulting non-contiguous views to follow-up work.

### Minimal reproduction

```python
import torch
from vllm.v1.attention.backends.flash_attn import FlashAttentionBackend

num_blocks = 8
block_size = 16
num_kv_heads = 8
head_size = 128

cache_shape = FlashAttentionBackend.get_kv_cache_shape(
    num_blocks, block_size, num_kv_heads, head_size
)
kv_cache = torch.empty(cache_shape, device="cuda", dtype=torch.float16)
key_cache, value_cache = kv_cache.transpose(1, 2).split(head_size, dim=-1)

print(cache_shape)
print(key_cache.shape, key_cache.stride(), key_cache.is_contiguous())
print(value_cache.shape, value_cache.stride(), value_cache.is_contiguous())
```

Observed result:

```text
(8, 8, 16, 256)
torch.Size([8, 16, 8, 128]) (...) False
torch.Size([8, 16, 8, 128]) (...) False
```

The shapes and dtypes are supported by the Fish kernel, but the contiguity check prevents dispatch.

## Expected behavior

The Fish attention fast path should consume the packed cache views without materializing per-decode `.contiguous()` copies. The guard should accept safe strided K/V tensors when the head dimension is dense, and the Triton kernels should calculate K/V addresses from the tensors' actual strides.

## Suggested fix and validation

- Replace the full K/V contiguity requirement with explicit stride safety checks.
- Pass all K/V strides to the short- and long-context Triton kernels.
- Add a backend regression test proving the real vLLM packed layout hits the fast path instead of falling back.
- Validate FP16/BF16 strided-cache outputs against a PyTorch reference for short and long sequence lengths, including CUDA Graph capture.

pxljs:
## Purpose

Fixes #5942.

The Fish Speech decode-only KV-cache backend unpacks vLLM's current packed cache layout with `transpose(1, 2) + split`. This produces correctly shaped K/V tensors, but they are non-contiguous strided views. The fast-path guard required both tensors to be fully contiguous, so every otherwise-compatible decode request fell back to the original attention implementation.

This is the known limitation left by #4428 after the cache-unpacking fix.

This PR:

- allows safe strided K/V views when their head dimension is dense;
- passes the actual K/V strides into both the short- and long-context Triton kernels;
- calculates K and V offsets independently without materializing per-decode contiguous copies;
- adds a backend regression test proving the real vLLM packed cache layout hits the Fish fast path instead of recording `guard_miss`;
- validates strided-cache kernel outputs against a PyTorch reference.

No public API, model configuration, or dependency changes are included.

## Test Plan

The bug is reproducible with synthetic tensors using the cache shape returned by vLLM's `FlashAttentionBackend.get_kv_cache_shape`; loading the full Fish Speech checkpoint is not required.

Coverage in `tests/attention/test_fish_kvcache_attn.py` includes:

- real vLLM packed/strided KV-cache layout and backend fast-path dispatch;
- rejection of layouts whose head dimension is not dense;
- FP16 and BF16 numerical comparison against a PyTorch reference;
- short and long sequence lengths;
- mixed short/long and zero-length rows;
- short- and long-path CUDA Graph capture.

**vLLM Version:** 0.26.0

**vLLM-Omni Commit:** `90102075`

**Test hardware:** NVIDIA GeForce RTX 4090

## Test Result

```bash
pytest -q tests/attention/test_fish_kvcache_attn.py
```

All tests in the module passed locally, including the CUDA-gated FP16/BF16 strided-cache correctness and CUDA Graph cases.

```bash
pre-commit run --files \
  vllm_omni/attention/fish_kvcache_attn.py \
  vllm_omni/attention/fish_kvcache_triton.py \
  tests/attention/test_fish_kvcache_attn.py
```

All hooks passed.

### DIFF
--- a/tests/attention/test_fish_kvcache_attn.py
+++ b/tests/attention/test_fish_kvcache_attn.py
@@ -169,6 +169,28 @@ def test_fish_kvcache_attn_guard_rejects_unsupported_block_size(monkeypatch):
     )
 
 
+def test_fish_kvcache_attn_guard_rejects_non_dense_head_dimension(monkeypatch):
+    monkeypatch.setenv("VLLM_OMNI_FISH_KVCACHE_ATTN", "1")
+    monkeypatch.setattr(fish_kvcache_attn, "is_available", lambda: True)
+    key_cache = torch.empty((8, 16, 8, 256), dtype=torch.float16)[..., ::2]
+    value_cache = torch.empty((8, 16, 8, 256), dtype=torch.float16)[..., ::2]
+
+    assert key_cache.stride(-1) == 2
+    assert not fish_kvcache_attn.can_use_fish_kvcache_attn(
+        query=torch.empty((4, 32, 128), dtype=torch.float16),
+        key_cache=key_cache,
+        value_cache=value_cache,
+        block_table=torch.zeros((4, 8), dtype=torch.int32),
+        seq_lens=torch.ones((4,), dtype=torch.int32),
+        max_query_len=1,
+        max_seq_len=128,
+        dcp_world_size=1,
+        use_cascade=False,
+        alibi_slopes=None,
+        sliding_window=None,
+    )
+
+
 def test_fish_kvcache_backend_wraps_only_model_instance(monkeypatch):
     fish_kvcache_backend.reset_fish_kvcache_attn_stats()
     monkeypatch.setenv("VLLM_OMNI_FISH_KVCACHE_ATTN", "1")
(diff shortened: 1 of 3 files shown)


BeatSeat:
I think this PR could fix the issue after a quick test and code review, and can be merged for kernel optimization.

PR #5945, opened 2026-08-09
[Perf][MiniCPM-o] Cut Talker per-step host overhead in the sampling path

worryzyy:
## Purpose

Cut avoidable per-step host overhead in the MiniCPM-o 4.5 Talker decode path. On Ascend 910C the Talker step is host-bound; py-spy profiling (200 Hz, steady-state decode) shows the sampling path spends ~1 ms/step on work that produces no value:

| Hot spot | Cost | Fix |
|---|---|---|
| `sample()` constructs `Sampler()` every decode step | module init per token | module-level singleton (the vLLM sampler is stateless) |
| Full sampler pipeline over 2-wide one-hot stop rows (`[0,-inf]` / `[-inf,0]`) | ~10 host-dispatched kernels for a deterministic pick | argmax fast-path, guarded on 2-wide logits and no logprobs request |
| `make_omni_output()` rebuilds constant tensors (stop rows, terminal flags, empty delta) via `new_tensor`/`torch.tensor` each step | per-step H2D copies | cache per (device, dtype) |
| `_apply_top_k_top_p()` clones the full-vocab row each step | one 6.5k-float copy per token | in-place masking behind an explicit `inplace=True` opt-in (caller's tensor is freshly materialized; default unchanged) |
| AR runners rebuild the full decoded-token history (D2H sync + list copies per step) for `prefer_model_sampler` models | grows with sequence length | models may declare `model_sampler_needs_output_token_ids = False`; skipped only when penalties don't need the history either. The Talker keeps its own codec history; only the duplex LLM sampler reads vLLM's. Applied to both the NPU and GPU AR runners. |

All changes are **bit-exact by construction**: no sampling distribution, RNG draw order, or output value changes. The argmax fast-path emits **int32** to match the standard sampler's output dtype — the runner `scatter_`s the ids into int32 input buffers on the async-scheduling path (int64 crashes there at concurrency > 1).

## Test Plan

**vLLM Version:** 0.1.dev1+ge5588e49b (as shipped in `quay.io/ascend/vllm-omni:v0.25.0-a3`)

**vLLM-Omni Commit:** benchmarked on `minicpm-challenge` (base of this branch)

Setup: Ascend 910C (Atlas A3), single card, exclusive; same-machine A/B with identical restart procedure and 3 warmup requests. Deploy config: the A3 low-latency variant from #5939 (also reproduces on the default `minicpmo_4_5.yaml` — the levers are config-independent).

- Per-step timing: env-gated instrumentation around the sampling path, plus stage ITL from server stats logs.
- E2E: `vllm bench serve --omni --backend openai-chat-omni --dataset-name seed-tts --seed-tts-locale en --no-oversample`, concurrency 1 = 3 rounds × 32 prompts, concurrency 4/8 = 64/128 prompts.
- Quality: seed-tts zh, seed-tts-eval protocol (funasr paraformer-zh CER), 400 utterances, same server.

## Test Result

Per-step (stage1 Talker, single stream):

| Metric | before | after |
|---|---|---|
| sampling path (chain + sync) | 1.50 ms/step | 1.35 ms/step |
| output assembly total | 1.75 ms/step | 1.48 ms/step |
| stage1 ITL | 8.7–9.6 ms/token | **7.6–8.4 ms/token** |

E2E (3-round means, concurrency 1): AUDIO_RTF **0.38 → 0.37**, TTFT/TTFP unchanged (295.8 ms / 811.8 ms), streaming continuity 100%, 96/96 requests OK.

Concurrency: no regression — conc 4 RTF 1.31 (64/64 OK), conc 8 RTF 2.01 (128/128 OK), continuity 100%.

Quality: seed-tts zh 400-utterance check CER **1.409%** (median 0.0, speaker-sim proxy 0.8485, zero failures) — within run-to-run noise of the F16 reference (1.414%) and of pre-change runs (1.377%/1.386%), as expected for a bit-exact change (the Talker's per-request RNG is not request-seed controlled, so run-to-run CER jitters regardless).

---

**参赛队伍：味蕾**（MiniCPM × 昇腾挑战赛 · 赛道一 · 子赛道 B / vLLM-Omni）

### DIFF
--- a/vllm_omni/model_executor/models/minicpmo_4_5/minicpmo_4_5_omni_tts.py
+++ b/vllm_omni/model_executor/models/minicpmo_4_5/minicpmo_4_5_omni_tts.py
@@ -23,6 +23,7 @@ from vllm.logger import init_logger
 from vllm.model_executor.models.interfaces import SupportsPP
 from vllm.model_executor.models.llama import LlamaModel
 from vllm.model_executor.models.utils import maybe_prefix
+from vllm.v1.outputs import SamplerOutput
 from vllm.v1.sample.sampler import Sampler
 
 from vllm_omni.experimental.fullduplex.engine.intermediate import get_tts_handoff
@@ -65,6 +66,17 @@ def _restore_weight_norm_weight(weight_g: torch.Tensor, weight_v: torch.Tensor) 
     return torch._weight_norm(weight_v, weight_g, dim=0)
 
 
+_SHARED_VLLM_SAMPLER: "Sampler | None" = None
+
+
+def _shared_vllm_sampler() -> "Sampler":
+    """The vLLM sampler is stateless; constructing it per decode step is pure overhead."""
+    global _SHARED_VLLM_SAMPLER
+    if _SHARED_VLLM_SAMPLER is None:
+        _SHARED_VLLM_SAMPLER = Sampler()
+    return _SHARED_VLLM_SAMPLER
+
+
 def _apply_repetition_penalty(
     logits: torch.Tensor,
     history: torch.Tensor,
@@ -87,9 +99,10 @@ def _apply_top_k_top_p(
     top_k: int | None,
     top_p: float | None,
     min_tokens_to_keep: int = 3,
+    inplace: bool = False,
 ) -> torch.Tensor:
     """Apply the same candidate floors as the upstream Transformers warpers."""
-    filtered = logits.clone()
+    filtered = logits if inplace else logits.clone()
     vocab_size = filtered.shape[-1]
     # MiniCPM-o's gen_logits() appends TopPLogitsWarper before
     # TopKLogitsWarper. The order is observable for fixed-seed sampling.
@@ -392,11 +405,14 @@ class MiniCPMO45OmniTTSForConditionalGeneration(nn.Module, SupportsPP):
         )
         if step < min_tokens:
             logits[..., eos_id] = float("-inf")
+        # ``logits`` is freshly materialized above (head_code -> float -> penalty),
+        # so the warpers may mask it in place.
         logits = _apply_top_k_top_p(
             logits,
             top_k=self._codec_top_k,
             top_p=self._codec_top_p,
             min_tokens_to_keep=3,
+            inplace=True,
         )
         probabilities = torch.softmax(logits, dim=-1)
         return torch.multinomial(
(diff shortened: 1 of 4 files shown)


Junbo2002:
Reproduced on two Atlas A3 (910C) machines. Both arms are `ecd9d99da`
differing only by this PR; the repo's own harness (Seed-TTS zh,
`openai-chat-omni`, 32 prompts, concurrency 1); both arms at once, one per
die, each round repeated with the dies swapped.

**End-to-end, four paired rounds:** −6.04% / +3.61% on one machine,
+4.26% / +5.61% on the other. The sign follows the die, not the arm, so these
numbers argue neither for nor against the PR. 32/32, 435 output tokens,
135.64 s of audio in all eight arms.

**In place** — same wrapper in both arms, two die-swapped rounds:

| method | calls / 32 req | round 1 | round 2 |
| --- | ---: | --- | --- |
| `_sample_audio_code` | 3557 | 1079.97 → 1060.74 µs | 1099.14 → **1151.26** µs |
| `sample` | 3591 | 228.18 → **56.15** µs | 228.69 → **57.10** µs |

1. **The `sample()` change is real and large**: −172 µs per call, −75%, nearly
   identical across the die swap; at ~112 calls per request that is
   −19.3 ms/request. Constructing a `Sampler()` per decode step, for logits
   that are two columns wide, costs ~228 µs of host time per step.
2. **The other two changes are not measurable here** — `_sample_audio_code`
   reads −19.2 µs in one round and +52.1 µs in the other.
3. Total ≈ 17 ms against a ~1900 ms request (0.9%), while two identically
   configured arms drift ±4–6% on these dies. That is why the end-to-end table
   cannot show this change, in either direction.

vllm_omni/model_executor/models/minicpmo_4_5/minicpmo_4_5_omni_tts.py
```diff
@@ -635,7 +674,22 @@ def compute_logits(self, hidden_states, *args, **kwargs):
...
+        # ``compute_logits`` emits one-hot stop rows ([0, -inf] / [-inf, 0]),
+        # so a full sampler pass is ~10 host-dispatched kernels for a
+        # deterministic pick; argmax is bit-identical on these rows. int32
+        # matches the standard sampler's output dtype -- the runner scatters
+        # these ids into int32 input buffers on the async path.
+        if (
+            isinstance(logits, torch.Tensor)
+            and logits.ndim == 2
+            and logits.shape[-1] == 2
+            and not getattr(sampling_metadata, "max_num_logprobs", None)
```
lishunyang12, line 686: Check max_num_logprobs is None; zero still requests the sampled token's logprob.

PR #5952, opened 2026-08-09
[MiniMax-H3] Enable calibrated TeaCache for Ref2VA

lishunyang12:
## Summary

Follow-up to #5840. This removes the FL2VA-only TeaCache limitation for MiniMax-H3.

- add an independently fitted Ref2VA TeaCache coefficient profile and model default threshold
- enable Ref2VA-only and combined serving, with isolated hooks and request-scoped state for the FL2VA and Ref2VA DiTs
- use a conservative Ref2VA policy: 35 full-compute calls followed by at most five cache hits on the calibrated 50-step workload
- fail closed to full computation for non-50-step Ref2VA requests
- let online serving defer to model-specific TeaCache defaults instead of forcing `rel_l1_thresh=0.2`
- document the ready-to-run Ref2VA serve command and measured trade-off

## B300 validation

The profile was fitted from 15 BF16 full-compute Ref2VA cases on one B300: 6 image-only, 6 image+audio, and 3 video+audio cases (720 consecutive full-compute pairs).

Same-seed A/B, CUDNN attention, 448x256, 107 output frames, 50 steps:

| Case | Cache hits | Baseline | TeaCache | Video quality | Audio quality |
| Image-only | 0/49 | 53.34 s | 53.26 s | SSIM 1.000 | sample-identical |
| Video+audio | 5/49 | 244.19 s | 221.24 s | SSIM 0.9795, PSNR 41.60 dB | PSNR 168.8 dB |

Both video+audio runs peaked at 140,972 MiB. TeaCache reduces compute latency; it is not a weight-memory optimization.

## User-facing usage

```bash
vllm serve "${MODEL_ROOT}/Ref2VA" \
  --omni \
  --trust-remote-code \
  --cache-backend tea_cache
```

Omitting `cache_config` selects the task-specific calibrated default. Custom thresholds keep the Ref2VA warmup and cache-hit budget.

## Tests

- `ruff check` on all changed Python files
- `pytest -q tests/diffusion/cache/test_cache_backends.py tests/diffusion/cache/test_teacache.py tests/diffusion/cache/test_teacache_extractors.py tests/engine/test_async_omni_engine_input.py`
- 82 passed
- `git diff --check`

### DIFF
--- a/vllm_omni/diffusion/cache/teacache/config.py
+++ b/vllm_omni/diffusion/cache/teacache/config.py
@@ -83,29 +83,9 @@ _MODEL_COEFFICIENTS = {
     ],
     # LongCat Image transformer coefficients
     "LongCatImageTransformer2DModel": [652.5980, -424.1615, 84.5526, -4.5923, 0.1694],
-    # MiniMax-H3 FL2VA coefficients.
-    "MiniMaxH3DiTModel": [
-        2.283704065852778e03,
-        -7.775977277886368e02,
-        9.408414741359490e01,
-        -4.232669906169421e00,
-        2.173782527946167e-01,
-    ],
-    # MiniMax-H3 Ref2VA coefficients.
-    "MiniMaxH3DiTModel:ref2va": [
-        2.1282372412524368e04,
-        -2.2712379331933703e03,
-        -5.8373015511205919e00,
-        6.2854270826260406e00,
-        2.5209538122600766e-01,
-    ],
 }
 
 _DEFAULT_REL_L1_THRESH = 0.2
-_MODEL_DEFAULT_REL_L1_THRESH = {
-    "MiniMaxH3DiTModel": 0.17,
-    "MiniMaxH3DiTModel:ref2va": 0.295,
-}
 
 
 @dataclass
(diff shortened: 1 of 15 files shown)


vllm_omni/diffusion/cache/teacache/config.py
```diff
@@ -143,3 +139,23 @@ def __post_init__(self) -> None:
...
+            raise ValueError(
+                f"calibrated_num_inference_steps must be positive or None, got {self.calibrated_num_inference_steps}"
+            )
+
+    def resolve_min_compute_steps(self, num_inference_steps: int) -> int:
+        """Return the request-local warmup, failing closed outside calibration."""
+        if (
+            self.calibrated_num_inference_steps is not None
+            and num_inference_steps != self.calibrated_num_inference_steps
+        ):
```
david6666666, line 159: [P1] When num_inference_steps is omitted, the runner passes 0 because MiniMaxH3Pipeline has no num_inference_steps attribute, while __call__ defaults to 50. Returning 0 here disables the 35-step warmup, so the default Ref2VA request can use an uncalibrated cache. Please resolve the pipeline effective default before refresh (or treat non-positive values as uncached).

lishunyang12: Fixed in 0d9594f21. MiniMaxH3Pipeline now exposes the effective 50-step default, and forward() uses the same attribute, so runner cache refresh and sampling agree when num_inference_steps is omitted. Added runner and pipeline contract regressions; 111 focused tests pass.

lishunyang12:
Addressed the latest review in 0d9594f21 and resolved the thread. The effective 50-step default is now shared by cache refresh and pipeline sampling, with runner and contract regressions. Validation: 111 passed; ruff and diff checks passed.

@david6666666 could you please double-check the update?

lishunyang12:
Follow-up: I rebased the branch onto current upstream/main and resolved the feature-matrix/pipeline conflicts. The reviewed default-step fix is now f86bd6ad7; the resolution preserves both the shared 50-step uniform default and the new partition-specific distilled sigma schedules. The H3 contract, TeaCache, and runner suites pass on the rebased head.

@david6666666 please double-check the rebased head.

PR #5955, opened 2026-08-09
[Diffusion] Add bounded and pipelined host staging for DLO

lishunyang12:
## H3 capacity benefit

The checked MiniMax-H3 FL2VA BF16 safetensors headers contain 50 DiT blocks plus two token-refiner blocks:

| Payload | GiB |
| --- | ---: |
| Persistent pinned shards without bounded staging | 61.559 |
| Largest layer | 1.202 |
| Two bounded pinned staging buffers | 2.405 |
| Pinned tensor payload avoided | **59.154 (96.09%)** |

This is the single-rank checkpoint-layout capacity model, not process RSS. TP sharding, quantization, allocator overhead, and model revisions can change per-rank values. The implementation therefore prints the actual reduction for the selected runtime topology.

The host-memory budget covers the private staging pool only. OS page cache, the initial checkpoint loader, encoders, VAEs, request state, and accelerator allocations are outside the bound.

## Host staging pipeline result

Host-only microbenchmark methodology:

- four file-backed BF16 shards at the H3 maximum layer size: 1.202 GiB each
- two pinned staging buffers
- 250 ms simulated per-layer compute window
- local ext4 storage
- 16 PyTorch CPU threads
- median of three alternating synchronous/async runs

| Bounded path | Exposed staging wait | Four-layer pipeline time |
| --- | ---: | ---: |
| Synchronous file-backed to pinned copy | 1.8149 s | 2.8159 s |
| Background one-layer-ahead copy | **1.0182 s (-43.90%)** | **2.0196 s (1.394x)** |

This measures the bounded host staging pipeline. It is not an end-to-end H3 GPU latency or production throughput claim.

## Correctness and lifecycle

- At most `buffer_count - 1` background tasks reserve slots, so an unexpected layer can always make progress.
- A staging slot is unavailable between `stage` and `mark_in_flight`.
- Reuse waits for the recorded H2D event on the background worker, not the model caller thread.
- An unexpected miss drains stale prefetches before rebuilding the correct lookahead sequence.
- Shutdown stops pending workers before hook dictionaries and mmap mappings are released.
- The bounded mode currently fails closed with DLO AllGather; it is scoped to rank-local no-AllGather DLO.

## Configuration

```bash
--dlo-no-use-allgather \
--dlo-host-memory-budget-gib 8 \
--dlo-host-cache-dir /local-nvme/vllm-omni-dlo \
--dlo-pinned-staging-buffer-count 2 \
--dlo-prefetch-depth 2
```
=
## Limitations and follow-ups

- The regular H3 loader still constructs the model before the file-backed cache is finalized, so this PR reduces steady private pinned payload but does not lower the documented 200 GiB first-start RAM requirement.
- Each worker needs additional local cache storage. Files are ephemeral and removed on clean shutdown.
- Cold requests remain dependent on local storage and page-cache behavior.
- A persistent preprocessed rank-local cache plus a block-streaming loader is the follow-up needed to reduce first-start peak RAM.
- Full H3 end-to-end capacity and latency A/B remains pending on an unoccupied GPU host, so this PR stays Draft.

## Documentation

The RTX 5090 recipe documents flags, the exact H3 payload model, host pipeline benchmark, storage requirements, and first-start boundary. The GB10 recipe continues to reject DLO because pinned staging shares the same unified-memory pool.

### DIFF
--- a/recipes/MiniMaxAI/MiniMax-H3-5090.md
+++ b/recipes/MiniMaxAI/MiniMax-H3-5090.md
@@ -39,6 +39,35 @@ private transfer memory is approximately:
 staging buffer count × sum(max layer-shard elements per dtype)
 ```
 
+For the checked FL2VA BF16 checkpoint, safetensors metadata gives the following
+single-rank capacity model for the 50 DiT blocks and two token-refiner blocks:
+
+| DLO tensor payload | GiB |
+| --- | ---: |
+| Persistent pinned shards without bounded staging | 61.559 |
+| Largest layer | 1.202 |
+| Two bounded staging buffers | 2.405 |
+| Pinned payload avoided | 59.154 (96.09%) |
+
+These values describe tensor payload, not process RSS. Runtime TP sharding,
+quantization, allocator overhead, and model revisions can change the per-rank
+values, so the startup log reports the actual file-backed bytes,
+private staging bytes, and pinned-payload reduction for the selected topology.
+
+The background copy also prevents the new file-backed-to-pinned transfer from
+remaining entirely on the layer critical path. In a host-only microbenchmark
+using four 1.202 GiB file-backed BF16 shards, two pinned buffers, a 250 ms
+per-layer compute window, local ext4 storage, and 16 PyTorch CPU threads, the
+median of three alternating runs was:
+
+| Bounded staging path | Exposed staging wait | Four-layer pipeline time |
+| --- | ---: | ---: |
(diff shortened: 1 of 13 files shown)


hsliuustc0106:
This PR touches **vllm_omni/diffusion/, recipes/MiniMaxAI/, vllm_omni/config/, vllm_omni/engine/, tests/diffusion/** (13 files). Based on CODEOWNERS coverage of the changed files, the most-related reviewers appear to be:

> @Isotr0py @princepride @SamitHuang

Issue #5281, opened 2026-07-21
[RFC]: PiD: Pixel Diffusion Decoder — Low-Cost High-Resolution Generation via Miniature DiT


Semmer2:
### 1.1 Problem: Native High-Resolution DiT Generation Is Prohibitively Expensive

Generating high-resolution images (2K/4K) natively with large DiT backbones is extremely costly. For Qwen-Image at 2048², each denoising step on a single A100 takes ~1.9s. A full 50-step run costs ~95s just for the DiT, plus another ~2.9s for VAE decoding — roughly **98s per image**.

The root cause is that DiT inference cost scales with spatial resolution. Generating at 512² latent space is cheap (~0.1s/step), while generating at 2048² latent space is an order of magnitude more expensive. Users who want high-resolution outputs pay a steep latency tax.

**PiD solves this** by decoupling resolution from the main DiT: run the large model at low resolution (e.g., 512² latent) for semantic composition, then use a small distilled DiT (1.3B, 4 steps) to upscale directly to high-resolution pixels. This replaces the conventional approach of running the main DiT at native high resolution.

### 1.2 Why PiD Works

PiD is built on three simple insights:

1. **Semantics vs. texture are separable**. The main DiT (e.g., FLUX.2 32B, Qwen-Image) handles global layout and semantics — this work can be done at low resolution. PiD (1.3B PixDiT) only handles pixel-level detail synthesis, which does not need 32B-parameter reasoning. A compact distilled model suffices.

2. **Compute asymmetry**. The main DiT runs ~50 forward passes with a large backbone. PiD runs only **4 steps** with a 1.3B backbone. The compute gap is drastic: FLUX.2 native 2K generation takes 102.2s, while FLUX.2(512²) + PiD takes 7.1s — a 14.3× speedup (from the PiD paper, Figure 9).

3. **Distillation does not degrade quality**. DMD2 distillation compresses multi-step teacher behavior into a 4-step student. The student actually **outperforms** the 50-step teacher on IQA metrics (MUSIQ, NIQE, MANIQA) as shown in the paper's Table 2.

### 1.3 Benchmarked Performance (Qwen-Image + PiD, Single A100 80GB)

| Pipeline | Denoise Latency | Decode Latency | E2E | Speedup |
|----------|----------------|----------------|-----|---------|
| Qwen-Image native 2048² | 1893ms × 50 = 94.7s | VAE decode: 2.86s | **~98s** | 1× |
[...]

PR #5965, opened 2026-08-10
[Feature] PiD: Pixel Diffusion Decoder

Jay-one-q:
#5281 
## Summary
This PR introduces the Pixel Diffusion Decoder (PiD) as a model-agnostic high-resolution decoding backend for the latent diffusion pipeline in vLLM-Omni. Instead of using the traditional VAE decoding, the x_0 latent produced by the upstream LDM is directly decoded into high-resolution RGB images with scale times pixel resolution (default: 4x) using the distilled PixDiT-T2I student (PidNet + Gemma-2-2b-it text encoder).

PiD reconstructs the decoding process from "latent → pixel" into a conditional pixel-space diffusion model, unifying decoding and super-resolution. It can directly generate 4K-level outputs without requiring retraining of the upstream generative model. As long as the generated latent belongs to the latent space supported by PiD, the same set of weights can be reused.

The core changes are located in the newly added [vllm_omni/diffusion/pid/](file:///c:/Users/jayQ/PycharmProjects/vllm-omni/vllm_omni/diffusion/pid) module, which provides two levels of entry points: CLI and OpenAI API, implementing both startup-level switches and per-request control.
## Motivation
At high resolution, VAE decoding becomes a bottleneck, and details appear blurry.
This enables the pipeline of text-to-image models like  Qwen-Image to directly obtain the "latent → high-definition/super-resolution" capability, eliminating the need for cascading external super-resolution models.
Maintains model independence: Any LDM that outputs latents can be reused.
### Offline inference (text_to_image example, A100 single card)
```bash
  --model Qwen/Qwen-Image \
  --prompt "a cat on a beach" \
  --enable-pid \
  --pid-checkpoint /path/to/pid_sr4x_v1pt5.pth \
```
### Online Services (OpenAI Compatible)
```bash
vllm-omni serve Qwen/Qwen-Image \
  --enable-pid \
  --pid-checkpoint /path/to/pid_sr4x_v1pt5.pth \
```
request `/v1/images/generations`：
```json
{
  "prompt": "a cat on a beach",
  "size": "512x512",
  "pid_decode": {
    "enabled": true,
    "num_steps": 4
  }
}
```
> `pid_decode` supports the following keys: `enabled`, `scale`, `num_steps`, `seed`, and `degrade_sigma`. If `enabled` is set to `false`, the VAE path is used. Similarly, `/v1/chat/completions` can be accessed through `extra_body.pid_decode`.
### Adapted Models
Five backbone net-configs are registered in omni ( qwenimage , flux , sd3 , sdxl , flux2 ), but currently only one pipeline actually inherits PidDecodeMixin :
- QwenImagePipeline ( PID_BACKBONE="qwenimage" )
The remaining four backbones have ready configs but their pipelines have not yet been wired up (integration = inherit the mixin + declare PID_BACKBONE ).
### Platform Support
| Platform | Status | Notes |
|----------|--------|-------|
| **CUDA (NVIDIA)** | ✅ Verified OK | Tested on A100 |
| **NPU (Ascend)** | ✅ Verified OK | Tested on A3;The current performance is worse than that of the GPU. The fault is being located and analyzed. `torch.compile` is skipped (runs eager) since OMNI's NPU platform reports `supports_torch_inductor()=False` |
| **ROCm / XPU / MUSA / others** | 🔶 Theoretically supported (to be verified) | All ops are native torch ops; device is derived via the OMNI platform abstraction 

Summary : PiD natively supports 10+ model families and 9 classes of open-source weights; omni currently adapts 1 model (Qwen-Image), with 4 more backbone configs ready to wire up. CUDA and NPU are both verified, other platforms are theoretically feasible.
## Adapting a new model (no pipeline code changes)
1. **Remote analysis** — check the upstream PiD repo for a released checkpoint
   and confirm the target model's latent space is compatible (latent channels,
   spatial compression).
2. **omni support check** — verify vllm-omni has a pipeline for the model and
   that its `output_type="latent"` branch returns the raw latent (original
   behavior is preserved).
3. **Latent form** — if the pipeline latent differs from the checkpoint's
   expected form (e.g. Flux2 returns a VAE-ready 32ch grid while PiD expects a
   BN-normalized 128ch patchified grid), implement a pure `to_x0` function and
   register the family in `LATENT_FORMS` (`pid/latent_forms.py`).
4. **Config registration** — add the backbone's `PidNet` config (latent
   channels, down factor, `lq_latent_unpatchify_factor`, etc.) to
   `pid/config.py` and wire the checkpoint path.
5. **Validate** — `_validate_x0` fails loudly on shape mismatch; contract:
   `C == lq_latent_channels` and `zH * latent_spatial_down_factor == pid_h`
   (`lq_latent_unpatchify_factor` is a channel-level op inside
   `LQProjection2D`, not a spatial term).
> The PiD configuration must be consistent with the original `PID_SR4X` settings (this repository has built-in configurations, so do not explicitly modify `pixel_attn_hidden_size`/`pixel_num_groups`, and ensure `repa_encoder_index=6`). Otherwise, the checkpoint weights cannot be loaded correctly. 
## reference link
- **PiD paper**: [PiD: Fast and High-Resolution Latent Decoding with Pixel Diffusion](https://arxiv.org/abs/2605.23902)
- **PiD project**: [NVIDIA Spatial Intelligence Lab — PiD](https://research.nvidia.com/labs/sil/projects/pid/)
- **PiD official repository**: [github.com/nv-tlabs/PiD](https://github.com/nv-tlabs/PiD)
- **PiD weights /checkpoints (HF)**: [huggingface.co/nvidia/PiD](https://huggingface.co/nvidia/PiD)
- Qwen-Image backbone: `checkpoints/PiD_v1pt5_res2kto4k_sr4x_official_qwenimage_distill_4step` (This PR uses the v1pt5 4-step distilled checkpoint in `.pth` format.)
## Test Result
vllm commit：2cf0a6915c

A100 single-GPU
Pipeline | Denoise Latency | Decode Latency | E2E | Speedup
-- | -- | -- | -- | --
Qwen-Image native 2048² | 1893ms × 50 = 94.7s | VAE decode: 0.36s | ~95s | 1×
Qwen-Image 512² + PiD → 2048² | 115ms × 50 = 5.75s | PiD: 1.6s | ~7.4s | 12.8×
Qwen-Image 512² + PiD (compile) → 2048² | 115ms × 50 = 5.75s | PiD: 1.2s | ~7s | 13.5×
Qwen-Image 1024² + PiD(compile) → 4096² | 16.8s |PiD: 10.8s | 28s | —

### DIFF
--- a/examples/offline_inference/text_to_image/text_to_image.py
+++ b/examples/offline_inference/text_to_image/text_to_image.py
@@ -377,6 +377,45 @@ def parse_args() -> argparse.Namespace:
         default=None,
         help="Supplementary auxiliary text encoder parameters model name or path (especially for Hidream-l1-full).",
     )
+    # PiD (Pixel Diffusion) super-resolution decode flags. When --pid-enable is
+    # set, the matching keys are packed into a ``pid_decode`` dict by
+    # AsyncOmniEngine and forwarded into OmniDiffusionConfig.pid_decode.
+    parser.add_argument(
+        "--pid-enable",
+        action="store_true",
+        default=False,
+        help="Enable PiD super-resolution decode (replaces VAE decode). ",
+    )
+    parser.add_argument(
+        "--pid-checkpoint",
+        type=str,
+        default=None,
+        help="Path to the PiD decoder checkpoint (.pth).",
+    )
+    parser.add_argument(
+        "--pid-gemma",
+        type=str,
+        default="Efficient-Large-Model/gemma-2-2b-it",
+        help="Gemma text encoder used by PiD.",
+    )
+    parser.add_argument(
+        "--pid-scale",
+        type=int,
+        default=4,
+        help="Super-resolution factor (e.g. 4 for 4x).",
+    )
+    parser.add_argument(
+        "--pid-num-steps",
+        type=int,
+        default=4,
+        help="Number of distilled SDE sampling steps (4 for the distilled checkpoint).",
+    )
+    parser.add_argument(
+        "--pid-seed",
+        type=int,
+        default=0,
+        help="Base RNG seed for PiD sampling.",
+    )
     current_omni_platform.pre_register_and_update(parser)
     return parser.parse_args()
 
@@ -478,6 +517,21 @@ def main():
     # gate is an engine-level config (offline analog of the server's --no-guardrails).
     if args.extra_body and "guardrails" in args.extra_body:
         omni_kwargs["model_config"] = {"guardrails": bool(args.extra_body["guardrails"])}
+    # PiD: forward CLI keys to AsyncOmniEngine, which packs them into a
+    # ``pid_decode`` dict consumed by OmniDiffusionConfig. Only inject when
+    # --pid-enable is set so non-PiD runs are unaffected.
+    if args.pid_enable:
+        omni_kwargs["pid_enable"] = True
+        if args.pid_checkpoint is not None:
+            omni_kwargs["pid_checkpoint"] = args.pid_checkpoint
+        if args.pid_gemma_path is not None:
+            omni_kwargs["pid_gemma"] = args.pid_gemma_path
+        if args.pid_scale is not None:
+            omni_kwargs["pid_scale"] = args.pid_scale
+        if args.pid_num_steps is not None:
+            omni_kwargs["pid_num_steps"] = args.pid_num_steps
+        if args.pid_seed is not None:
+            omni_kwargs["pid_seed"] = args.pid_seed
     omni = Omni(**omni_kwargs)
     model_class_name = get_model_class_name(omni)
     declared_extra_body_params = get_extra_body_params(model_class_name)
--- a/tests/e2e/accuracy/test_qwen_image_pid_accuracy.py
+++ b/tests/e2e/accuracy/test_qwen_image_pid_accuracy.py
@@ -1,0 +1,102 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+"""L4 accuracy for PiD 4x super-resolution on Qwen-Image (nightly).
+
+Two quality guard lines:
+1. Structure preservation: PiD 4x output downsampled to 512 vs the VAE 512
+   baseline from the same server (content must not drift or hallucinate).
+2. Distill determinism (golden): PiD 4x output vs the nvidia/PiD official
+   reference for the same prompt/seed (guards regression of the distilled
+   student). Skipped when the golden reference image is absent.
+
+Thresholds are initial suggestions; calibrate on H100 before enabling.
+"""
+
+from __future__ import annotations
+
+import base64
+import io
+import os
+from pathlib import Path
+
+import pytest
+import requests
+from PIL import Image
+
+from tests.e2e.accuracy.helpers import assert_similarity, model_output_dir
+from tests.helpers.runtime import OmniServer
+
+pytestmark = [pytest.mark.full_model, pytest.mark.diffusion]
+
+MODEL_ID = os.environ.get("QWEN_IMAGE_MODEL", "Qwen/Qwen-Image")
+PID_CKPT = os.environ.get("PID_CKPT", "")
+PID_GEMMA = os.environ.get("PID_GEMMA", "Efficient-Large-Model/gemma-2-2b-it")
+PROMPT = "A photo of a cat sitting on a laptop keyboard, digital art style."
+SIZE = "512x512"
+SSIM_STRUCTURE, PSNR_STRUCTURE = 0.90, 28.0
+SSIM_GOLDEN, PSNR_GOLDEN = 0.94, 30.0
+
+_pid_args = ["--pid-enable", "--pid-gemma", PID_GEMMA]
+if PID_CKPT:
+    _pid_args += ["--pid-checkpoint", PID_CKPT]
+
+
+def _omni_pid(*, enabled: bool) -> Image.Image:
+    server_args = ["--num-gpus", "1", "--stage-init-timeout", "300", "--init-timeout", "900", *_pid_args]
+    with OmniServer(MODEL_ID, server_args, use_omni=True) as omni_server:
+        resp = requests.post(
+            f"http://{omni_server.host}:{omni_server.port}/v1/images/generations",
+            json={
+                "model": omni_server.model,
+                "prompt": PROMPT,
+                "size": SIZE,
+                "n": 1,
+                "response_format": "b64_json",
+                "num_inference_steps": 20,
+                "true_cfg_scale": 4.0,
+                "seed": 42,
+                "pid_decode": {"enabled": enabled, "scale": 4, "num_steps": 4, "seed": 42},
+            },
+            timeout=600,
+        )
+        resp.raise_for_status()
+        data = resp.json()["data"][0]["b64_json"]
+        img = Image.open(io.BytesIO(base64.b64decode(data))).convert("RGB")
+        img.load()
+        return img
+
+
+def _golden_reference() -> Image.Image | None:
+    # Produced once with nvidia/PiD scripts/pipeline_demo.py for the same prompt/seed.
+    golden = model_output_dir(Path(__file__).parent, "qwen_image_pid") / "golden_4x.png"
+    return Image.open(golden).convert("RGB") if golden.exists() else None
+
+
+def test_pid_4x_structure_preserved() -> None:
+    """PiD 4x downsampled to 512 matches the VAE 512 baseline (no content drift)."""
+    pid_4x = _omni_pid(enabled=True)
+    assert pid_4x.size == (2048, 2048)
+    baseline = _omni_pid(enabled=False)
+    pid_down = pid_4x.resize(baseline.size, Image.LANCZOS)
+    assert_similarity(
+        model_name="qwen_image_pid",
+        vllm_image=pid_down,
+        diffusers_image=baseline,
+        ssim_threshold=SSIM_STRUCTURE,
+        psnr_threshold=PSNR_STRUCTURE,
+    )
+
+
+def test_pid_4x_golden_match() -> None:
+    """PiD 4x output matches the nvidia/PiD official reference (distill determinism)."""
+    golden = _golden_reference()
+    if golden is None:
+        pytest.skip("missing nvidia/PiD official reference output; skipping golden comparison")
+    pid_4x = _omni_pid(enabled=True)
+    assert_similarity(
+        model_name="qwen_image_pid",
+        vllm_image=pid_4x,
+        diffusers_image=golden,
+        ssim_threshold=SSIM_GOLDEN,
+        psnr_threshold=PSNR_GOLDEN,
+    )
--- a/vllm_omni/diffusion/pid/config.py
+++ b/vllm_omni/diffusion/pid/config.py
@@ -1,0 +1,179 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+"""PiD network & sampling config registry for all supported backbones.
+
+All backbones share the same PixDiT_T2I architecture; only the LQ-related
+constructor args differ per VAE.
+
+Backbone -> VAE characteristics:
+    Qwen-Image:  16ch latent, 8x spatial compression
+    Flux1:       16ch latent, 8x spatial compression
+    SD3:         16ch latent, 8x spatial compression
+    SDXL:         4ch latent, 8x spatial compression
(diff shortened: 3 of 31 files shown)


Semmer2:
Hi everyone, @hsliuustc0106  @Gaohan123 @princepride  @lishunyang12 kindly ping. This PR contains a large amount of model structure code,  but the overall modification are quite clear.

Jay-one-q:
Hi everyone,@Semmer2 @hsliuustc0106 @Gaohan123 @princepride @lishunyang12 kindly ping. 
This PR integrates the PiD super-resolution feature, which has been successfully tested on multiple models. It requires no invasive modifications to the model-side code, and subsequent adaptation to other models only requires modifications to the PiD-related configuration code

examples/offline_inference/text_to_image/text_to_image.py
```diff
@@ -478,6 +517,21 @@ def main():
...
     if args.extra_body and "guardrails" in args.extra_body:
         omni_kwargs["model_config"] = {"guardrails": bool(args.extra_body["guardrails"])}
+    # PiD: forward CLI keys to AsyncOmniEngine, which packs them into a
+    # ``pid_decode`` dict consumed by OmniDiffusionConfig. Only inject when
+    # --pid-enable is set so non-PiD runs are unaffected.
+    if args.pid_enable:
+        omni_kwargs["pid_enable"] = True
+        if args.pid_checkpoint is not None:
+            omni_kwargs["pid_checkpoint"] = args.pid_checkpoint
+        if args.pid_gemma_path is not None:
```
hsliuustc0106: **[P1]** `args.pid_gemma_path` doesn't exist — the flag defined above is `--pid-gemma` (dest `pid_gemma`), so any `--pid-enable` run of this example (including the quickstart in the PR description) crashes with `AttributeError` before the engine starts. Fix: `args.pid_gemma`.

Also: `pid_scale`/`pid_num_steps`/`pid_seed` are forwarded into `omni_kwargs` but `AsyncOmniEngine.__init__` only pops `pid_enable`/`pid_checkpoint`/`pid_gemma` — the other three leak into the engine kwargs and are (at best) silently ignored, and `serve.py` has no `--pid-scale/--pid-num-steps/--pid-seed` equivalents. Either pop them in AsyncOmniEngine and pack them into the `pid_decode` dict (scale/num_steps/seed keys), or drop the three flags here.

vllm_omni/diffusion/pid/pixeldit.py
```diff
@@ -0,0 +1,998 @@
...
+                joint_value=vy,
+                joint_strategy="front",
+            )
+            # The Ulysses/Ring strategies re-concatenate the joint text to the
+            # front of the output ([Ny + Nx, H, Hc]), so drop the text part and
+            # keep only the (sharded) image tokens.
+            out_x = self.vattn(qx, kx, vx, md)  # [B, Ny + Nx, H, Hc]
+            out_x = out_x[:, Ny:, :, :]
+            out_x = out_x.reshape(B, Nx, C)
+            out_y = torch.zeros_like(y)
```
hsliuustc0106: **[P2]** Under SP this drops the gathered joint output's text part and returns zeros for the text stream, so the MMDiT text residual loses its joint-attention update at every patch block (and that stream is the KV source for later blocks) — SP decode diverges numerically from the verified single-card path. The comment "same as Qwen-Image under SP" is not what Qwen-Image does: `qwen_image_transformer.py:703-723` uses the gathered text output (`txt_attn_output = joint_hidden_states[:, :seq_len_txt, :]`), and per your own comment above the gathered tensor here already contains the true text part in `out_x[:, :Ny]`. If the zeroing isn't required by a backend limitation, `out_y` can come from the same gathered tensor at no extra communication; otherwise please correct the comment, document the divergence in `pid_decode.md`, and add an SP-vs-single-card quality check so SP quality isn't unverified.

vllm_omni/diffusion/pid/pid_model.py
```diff
@@ -0,0 +1,325 @@
...
+        seed: int = 0,
+    ) -> torch.Tensor:
+        """Run PiD decode.
+
+        Returns:
+            (B, 3, H, W) tensor in [-1, 1].
+        """
+        if isinstance(caption, str):
+            caption = [caption]
+        B = len(caption)
```
hsliuustc0106: **[P2]** `B = len(caption)` breaks for `n > 1`: a request with `n >= 2` reaches `_decode_one` as one latent tensor `[n, seq, C]` (QwenImagePipeline.forward splits per request, not per image), but the caption here is a single string, so `noise` is built with `B=1` while `lq_latent` has batch `n` → shape mismatch crash inside the net → 500 on every n≥2 request on a PiD-enabled server. `maybe_pid_passthrough` already falls back to VAE for img2img/latents; either broadcast the caption list to `lq_latent.shape[0]` (splitting per image) or add the same VAE fallback + test for `num_outputs_per_prompt > 1`.

vllm_omni/diffusion/pid/runner_integration.py
```diff
@@ -0,0 +1,375 @@
...
+    caption: str | None,
+    pid_override: dict[str, Any] | None = None,
+) -> torch.Tensor | None:
+    """Run one PiD super-resolution decode (override / rank gating here).
+
+    Returns the super-resolved RGB tensor ``[B, 3, pid_h*scale, pid_w*scale]``
+    in [-1, 1], or None on non-SP non-main ranks (results are not gathered).
+    """
+    if pid_override is not None:
+        overrides = {k: pid_override[k] for k in ("scale", "num_steps", "seed", "degrade_sigma") if k in pid_override}
```
hsliuustc0106, line 138: **[P2]** Per-request `pid_decode` dicts are never validated: unknown keys are silently filtered here (e.g. `"scales": 4` → ignored), and out-of-range values (`scale: -1`, `num_steps: "4"`) flow into tensor ops. This contradicts the new weekly test `tests/dfx/reliability/invalid_param_test/test_invalid_pid_params.py:52`, which expects `pid_decode={"extra": 1}` to return 4xx/5xx with `"extra"` in the message — against this code it returns 200, so that test cannot pass when its H100 lane runs. Smallest fix: validate the dict (strict keys + types) once — in the protocol layer or here in `maybe_pid_passthrough` — and raise a 400-able error on unknown keys. Related nit: `enabled=true` without `--pid-enable` raises loudly in the batch path but is silently ignored by `stepwise_pid_active` in the streaming path — worth converging on one behavior.

vllm_omni/diffusion/pid/config.py
```diff
@@ -0,0 +1,186 @@
...
+        4,
+    ),
+}
+
+# ---------------------------------------------------------------------------
+# Typed config wrappers
+# ---------------------------------------------------------------------------
+
+
+@dataclass
```
hsliuustc0106: Nit: `PidNetConfig`, `PidSamplingConfig`, and `get_pid_sampling_config()` (line 184) have no consumer anywhere — production code, tests, and the docs all go through `get_pid_net_config()` / `PID_SAMPLING_CONFIG` directly. Since this is brand-new code, suggest deleting the two dataclasses and the getter (plus their `__init__.py` exports and the `adaptation-recipe.md:160` mention) instead of shipping dead public API.

tests/e2e/accuracy/test_qwen_image_pid_accuracy.py
```diff
@@ -0,0 +1,102 @@
...
+"""L4 accuracy for PiD 4x super-resolution on Qwen-Image (nightly).
+
+Two quality guard lines:
+1. Structure preservation: PiD 4x output downsampled to 512 vs the VAE 512
+   baseline from the same server (content must not drift or hallucinate).
+2. Distill determinism (golden): PiD 4x output vs the nvidia/PiD official
+   reference for the same prompt/seed (guards regression of the distilled
+   student). Skipped when the golden reference image is absent.
+
+Thresholds are initial suggestions; calibrate on H100 before enabling.
```
hsliuustc0106: Evidence gap for the headline numbers: the PR body's 12.2× table has no exact commands, warmup/repetition counts, run-to-run variability, or memory figures, the "20–30% additional reduction with torch.compile" claim has no supporting run, and the 1024²→4096² row has no numbers at all — per the repo's perf contract these toggle-comparison columns should come from the same head with pinned invariants (steps/pixels/seed). Also this file calibrates thresholds "on H100" while the perf table is A100 — please name the hardware (and the resolved SSIM/PSNR values from that calibration run) next to each number so the claimed operating point is reproducible.

vllm_omni/diffusion/pid/pixeldit.py
Jay-one-q: Thank you for pointing out that the backend does not need to set the text to zero, modify out_y to retrieve from the same collected tensor, and verify that the sp quality is normal.

examples/offline_inference/text_to_image/text_to_image.py
Jay-one-q: These three parameters have been deleted, and the parameter name pid_gemma has been unified.

vllm_omni/diffusion/pid/pid_model.py
Jay-one-q: Fixed the n > 1 case in pid_model.py#L269-L274 : B is now taken from lq_latent.shape[0] (which is n for a multi-image request), and the single caption is broadcast to match it, so noise is built with the correct batch size and the shape mismatch no longer occurs.And for the case where sp is enabled and n > 1, there is no issue with quality

vllm_omni/diffusion/pid/runner_integration.py
Jay-one-q: Validation has been added, and the test case in tests/dfx/reliability/invalid_param_test/test_invalid_pid_params.py is functioning normally

vllm_omni/diffusion/pid/config.py
Jay-one-q: Deleted

tests/e2e/accuracy/test_qwen_image_pid_accuracy.py
Jay-one-q: The comment has been revised and the test results have been updated. For detailed results, please refer to #5281

Jay-one-q:
@hsliuustc0106 Thank you for identifying multiple issues, all of which have been fixed. Please review again

hsliuustc0106:
fix conflicts

Jay-one-q:
> fix conflicts

@hsliuustc0106 fixed

hsliuustc0106:
I suggest to change CLI to "--enable-pid"

Jay-one-q:
> I suggest to change CLI to "--enable-pid"
Alright, completed

Jay-one-q:
@hsliuustc0106  Thank you. All comments have been revised.

vllm_omni/diffusion/pid/pid_model.py
```diff
@@ -0,0 +1,326 @@
...
+        effective_steps = num_steps or self._cfg.student_sample_steps
+
+        if effective_steps == 1:
+            t_student = torch.full(
+                (B,),
+                self._cfg.student_t_list[0],
+                **tensor_kwargs,
+            )
+            t_scaled = t_student * self._cfg.fm_timescale
+            v = net(
```
lishunyang12: Wrap single-step decoding in autocast too; BF16 weights currently receive FP32 inputs.

vllm_omni/diffusion/pid/pixeldit.py
```diff
@@ -0,0 +1,993 @@
...
+        if height == width:
+            pos = get_2d_sincos_pos_embed(self.hidden_size_output, height).to(device=device, dtype=dtype)  # [H*W, D]
+        else:
+            # Build a non-square grid (H x W) and compute 2D sin/cos embedding.
+            grid_h = torch.arange(height, dtype=torch.float32)
+            grid_w = torch.arange(width, dtype=torch.float32)
+            grid = torch.meshgrid(grid_w, grid_h, indexing="xy")  # w first to match existing convention
+            grid = torch.stack(grid, dim=0).reshape(2, 1, height, width)
+            pos = get_2d_sincos_pos_embed_from_grid(self.hidden_size_output, grid).to(device=device, dtype=dtype)
+        self._pos_cache[key] = pos
```
lishunyang12: Bound the positional caches; each new resolution permanently retains another GPU tensor.

vllm_omni/diffusion/pid/pid_model.py
Jay-one-q: Fixed

vllm_omni/diffusion/pid/pixeldit.py
Jay-one-q: You are right. Currently, the maximum cache size is limited to 8

PR #5898, opened 2026-08-07
[CI] Re-enable bagel shared-memory connector test (#5475)

yenuo26:
## Purpose

Re-enable `tests/distributed/omni_connectors/test_bagel_shared_memory_connector.py::test_bagel_img2img_shared_memory_connector` by removing the temporary `@pytest.mark.skip(reason="#issue 5475")`.

Net change vs `main`: only this unskip. An earlier draft skip for #5893 on the diffusion config classification test was added and then reverted on this branch, so it is not part of the final diff.

Related: https://github.com/vllm-project/vllm-omni/issues/5475

## Test Plan

**vLLM Version:** 0.26.0 (aligned with current main CI)

**vLLM-Omni Commit:** this PR head
pytest tests/distributed/omni_connectors/test_bagel_shared_memory_connector.py --run-level advanced_model

## Test Result
60e0f022-6b3c-4949-9a4b-6330ac899395

Local: skip marker removed; no production code change.

PR #5524, opened 2026-07-29
[CI][MiniCPM-o] Add MiniCPM-o 4.5 accuracy and performance coverage

R2-Y:
PLEASE FILL IN THE PR DESCRIPTION HERE.

## Purpose
Added continuous monitoring for accuracy and performance regression for MiniCPM-o 4.5.
- Added accuracy threshold tests for Daily-Omni.
- Added a performance baseline configuration for a two-card setup.
- Added two nightly CI jobs.

## Test Plan

**vLLM Version:**
0.26.0
**vLLM-Omni Commit:**
0.26.0

```
#!/usr/bin/env bash
set -uo pipefail

REPO="${REPO:-/data/yrr/vllm-omni}"
VENV="${VENV:-/data/yrr/.venv}"
RESULT_ROOT="${RESULT_ROOT:-/data/yrr/rein_test/minicpm_ci_results}"
LOGDIR="${RESULT_ROOT}/logs_pr5524"
SUMMARY="${LOGDIR}/summary.txt"
PYTEST="${VENV}/bin/pytest"

export PATH="${VENV}/bin:${PATH}"
export PYTHONPATH="${REPO}${PYTHONPATH:+:${PYTHONPATH}}"
CU13_LIB="${VENV}/lib/python3.12/site-packages/nvidia/cu13/lib"
if [[ -d "${CU13_LIB}" ]]; then
  export LD_LIBRARY_PATH="${CU13_LIB}${LD_LIBRARY_PATH:+:${LD_LIBRARY_PATH}}"
fi

unset HF_HUB_OFFLINE HF_DATASETS_OFFLINE TRANSFORMERS_OFFLINE || true
export HF_HOME="${HF_HOME:-/data/models}"
export VLLM_WORKER_MULTIPROC_METHOD=spawn
export TOKENIZERS_PARALLELISM=false
export RAYON_NUM_THREADS=1
export OPENBLAS_NUM_THREADS=1
export OMP_NUM_THREADS=1
export MKL_NUM_THREADS=1
export NUMEXPR_NUM_THREADS=1
export VLLM_DAILY_OMNI_QA_JSON="${VLLM_DAILY_OMNI_QA_JSON:-/data/models/datasets/Daily-Omni/qa.json}"
export VLLM_DAILY_OMNI_VIDEO_DIR="${VLLM_DAILY_OMNI_VIDEO_DIR:-/data/models/datasets/Daily-Omni/Videos}"
[...]

PR #5976, opened 2026-08-10
[Rebase] Rebase to vllm 0.27.0

tzhouam:
## Purpose

Rebase vLLM-Omni onto the released **vLLM v0.27.0** (base image `vllm/vllm-openai:v0.27.0`, torch 2.13.0). Beyond the mechanical rebase, this PR fixes every regression the alignment surfaced across nightly builds 2952–2957, grouped below by root cause.

### A. vLLM 0.27 API / behavior drift

| Bug | Root cause | Fix |
|---|---|---|
| Every AR/generation stage died at warmup with `TypeError: _dummy_run() got an unexpected keyword argument 'randomize_inputs'` (~20 jobs in build 2952) | vLLM 0.27 added `randomize_inputs` to `GPUModelRunner._dummy_run` and FlashInfer autotune passes it; omni's overrides redeclare the signature | 701bcc68 accepts and forwards the kwarg |
| Code2wav stages died with CUDA device-side asserts (11 jobs in build 2953) | Follow-up to the fix above: forwarding randomization into **generation** stages fills dummy ids with vocab-uniform randoms, which code2wav treats as packed RVQ codec codes → out-of-range embedding lookup at `code_embedding(codes + code_offset)` | dedda28f — generation runner accepts but never forwards `randomize_inputs`; the thinker (a real MoE) keeps randomizing |
| Plain `vllm serve Qwen/Qwen3-Omni-30B-A3B-Instruct` crashed in `profile_run` with `TypeError: only integer tensors of a single element can be converted to an index` | Since omni archs unconditionally override upstream registry entries, non-staged vLLM runs construct omni classes; the thinker `forward` always returned a `(hidden_states, captured_hidden_states)` tuple that stock vLLM's runner cannot index | e3efeaaf — bare tensor when no capture is requested; capture only in staged runs (only a talker consumes it, and each capture costs a clone per layer per step); `make_omni_output` accepts both shapes. Verified by a dummy-weight serve counterfactual: fails before, serves after |
| HunyuanImage3 (MoE DiT, TP=4 + EP) failed weight loading with `assert all2all_manager is not None` | vLLM 0.27's MoE oracle requires the EP group's device communicator to carry an all2all manager, which vLLM only builds for groups created with `use_all2all=True`; the diffusion stack's EP group never passed it | c382e801 — pass `use_all2all=True`, version-guarded for older vLLM |
| seed-TTS WER regressed 0.2695 -> ~0.38 (nightly 2946 -> 2949+; median 0.0 — a tail of corrupted utterances) and intermittent `audio content is not same as the text` e2e failures | Two candidate defects landed in the same window. (1) The async-discard reimplementation (59b9e719) seeded `num_stale_output_tokens` in placeholder units while the drain subtracts scheduled-token units; on divergence (spec drafts, in-flight prefill) the leftover counter silently swallowed the new segment's first valid talker frames (hole in the codec stream) or underflowed the drain assert. CPU-proven by scheduler-level tests (e21bffaf). (2) The same window switched CI to the v0.27.0 image (torch 2.13). | c2b65233 fixes (1) — seed from `num_in_flight_tokens` (same units as the drain) at all three discard sites; pre-discard frames drain exactly to zero by construction. **Nightly confirmation is negative for the WER metric**: build 2957 (first nightly with the fix) scored 0.3916, within the pre-fix range (0.3756 / 0.3826 / 0.3936 / 0.5114), so (1) alone does not restore the 0.27 baseline. The WER regression stays open — tracked under Known remaining issues. |
| Multi-replica stage startup died with `Cubin file saved by TritonBundler not found` | vLLM 0.27's AOT compile: replicas of the same stage compile the same model concurrently and race on the shared `VLLM_CACHE_ROOT` | 9c319283 — spawn-time per-replica cache isolation; replica 0 keeps the warm shared cache |

### B. torch 2.13 toolchain fallout

| Bug | Root cause | Fix |
|---|---|---|
| fp8 FLUX died at first compiled forward: `torch._inductor.exc.InductorError: CantSplit: 15360*s31 + 15360*s87 not divisible by s31 + s87` | torch ≤ 2.11 proved divisibility via python `%` (sympy `Mod`, whose eval factors common terms); torch 2.13 switched to torch's own unevaluated `Mod` for compile-time cost, losing provability of factorable sums — exactly FLUX's two-dynamic-dim fp8 graph | acf245b2 — patch `SizeVarAllocator.statically_known_multiple_of` with an exact-polynomial-division proof (`sympy.cancel(n/d)` yielding an integer-coefficient polynomial ⇒ divisibility holds for every symbol assignment; unconditionally sound). Polynomial-only + ≤20-symbol guards; self-extinguishes on torch versions that already prove the canonical case |
| (Interim) engine crashed on lazy compiler bugs generally | torch.compile errors surface at the first forward, past the setup-time try/except | e3b42850 added a compiler-error-only eager fallback; 30603e43 made it signature-transparent after it broke cache-dit block matching; **de3502cf removed the fallback** once the root cause above was fixed — future lazy compiler bugs now fail fast and loud. The cache-dit signature contract keeps a regression test |
| Qwen-Image accuracy tests crashed fetching FA3: `Cannot find a build variant for this system in kernels-community/flash-attn3` | The hub kernel ships no prebuilt variant for torch 2.13 | 213c1497 — probe once; when unavailable, run the omni server with `TORCH_SDPA` and the diffusers reference with `native` so both comparison sides keep matched attention math; b6956a93 applies the same idea to `test_diffusers_backend_similarity` by walking the diffusers adapter's own backend fallback chain (exported as a shared constant). 6638b5d2 gates the SDPA path at SSIM 0.96 (measured 0.9687; the strict 0.97 re-engages automatically once the hub ships a torch-2.13 build) |

### C. Feature bugs from the merge window

| Bug | Root cause | Fix |
|---|---|---|
| LingBot TI2V ignored `extra_params={"size": "320x192"}` and generated 480×480 | The #5311 request resolver read size/resolution/ratio/width/height from the offline prompt dict only, never from `sampling.extra_args` where `/v1/videos` merges `extra_params` | 2ebda68a — dimensions consult extra_args first, same precedence fps/num_frames already use |
| LingBot T2I doc example crashed: `'numpy.ndarray' object has no attribute 'save'` | The pipeline emits NumPy into `OmniRequestOutput.images` (declared PIL) for the serving path; the example saved it directly | Fixed independently on main by 73a312d5 (#6031, `_normalize_images_for_save` in the example); our equivalent fix 06e36ebd was reverted (c59d1fcf) in favor of main's when merging main (6965a5a2) |
| All four vLLM-text perf cases failed with `Request failures exist` after every request had already succeeded (build 2956) | The duplex-metrics aggregation merged from main (#5524) reads `duplex_request_metrics` / `duplex_session_metrics` on every output, but plain-vLLM backends (`backend='vllm'`) return upstream `RequestFuncOutput` objects without the Mix fields; the AttributeError at aggregation discarded the whole benchmark result (`completed=0`) | 2dc5e513 — tolerant `getattr` reads with defaults |

### D. Test-infrastructure corrections

| Issue | Fix |
|---|---|
| Bagel SHM connector golden pixels (imported from main via #5898) deterministically mismatch this branch's output — bit-identical actuals across builds 2953/2954 on two different torch stacks; only probe (512, 336) drifted beyond ±10 | 64e8cd38 makes the assert report all ten probes as a paste-ready table; 5c1fe35e re-goldens from the reported values |
| `fp8_ltx2` LPIPS unstable near its 0.1 gate (0.1104 / pass / 0.1291 across three nightlies, including compiled-vs-compiled runs) | e39f2604 — both quality-test arms run `enforce_eager` so LPIPS measures quantization only (mirrors the in-tree trajectory-similarity tool). Under observation; if it still exceeds the gate this is a genuine borderline-quality question |
| MiniCPM-o 4.5 merge job brushed its 50-minute timeout — build 2956 spent 28+ min inside a single test, vs ~15–19 min for the whole job on main | 322600bf had downgraded the job to h100_1, which silently moved Whisper transcript grading to CPU (~30 s → 8–10 min per clip; the helper only borrows an accelerator when a spare GPU exists). 86bee6d3 restores the spare h100_2 so Whisper grades on GPU again |

## Review follow-ups & housekeeping

- a1e88349 addresses @Gaohan123's review: `PureDiffusionLauncherAdapter` moved out of `api_server.py` into `vllm_omni/entrypoints/utils.py`; mammoth_moda2's `_is_interleaved` promoted to a shared `is_interleaved` in `model_executor/models/utils.py` (other models hit by upstream removal 26d725c334 can reuse it); documented that `benchmarks/patch/patch.py` tracks the `probe_request_rate` loop vLLM v0.27.0 added to `vllm/benchmarks/serve.py`.
- 81aecaa3 adds DCO sign-offs for all prior branch commits; 25b80bb9 applies ruff formatting.

## Known remaining issues (not addressed here)

- **FLUX.2-dev gated repo 403** in Tiny Model Tests — the CI HF token needs access; not a code issue, and **it reproduces on main as well** (pre-existing, not introduced by this branch).
- **Qwen-Image-Edit-2511 perf**: one run completed 9/10 requests with a clean shutdown signal mid-benchmark — suspected harness race, observing.
- **Qwen-Image SDPA-fallback PSNR fluctuates near its FA3-calibrated 30.0 gate** (2954: SSIM 0.9687 misses while PSNR passes; 2956: SSIM passes while PSNR lands 28.66). A looser fallback-only PSNR gate was tried (9b4eccc4) and reverted (b26b9085) — the strict gate stands while the FA3 hub lacks a torch-2.13 build.
- **Qwen3-Omni seed-TTS WER stuck at ~0.39 (baseline 0.27, gate 0.35)** — confirmed not resolved by c2b65233 (build 2957: 0.3916, inside the pre-fix range). Evidence points at an image/torch-2.13 numerics effect in the audio path rather than content changes: the regression window (2946->2949) is exactly the switch to the v0.27.0 image; daily-omni MCQ accuracy shifted deterministically 71.60% -> 70.68% at the same boundary; per-request output length is unchanged (~19 tokens, so no reasoning/content insertion); median WER stays 0.0000 — a tail of corrupted utterances grew. Next step: run one nightly with `seed_tts_wer_save_items=True` to capture per-item WER and transcripts in the artifact JSON, then A/B the worst utterances on the 0.26 vs 0.27 images to isolate talker sampling vs code2wav synthesis.

Resolved since the last update: the **MiniCPM-o 4.5 duplex soft-interrupt** (session ran but the model emitted 0 responses / 13 `listen` decisions) no longer reproduces on recent runs and is considered resolved.

## Test Plan

**vLLM Version:** v0.27.0 (released tag; base image `vllm/vllm-openai:v0.27.0`, torch 2.13.0)

**vLLM-Omni Commit:** 25b80bb9 (includes the merge of main through f8fa86fb)

- New L1 regression tests per fix: thinker forward contract (8), LingBot dimension resolution (4), EP `use_all2all` plumbing (2), replica cache-env scoping (4), inductor divisibility patch (12, incl. a simulated-torch-2.13 install path), cache-dit signature contract (1), stale-drain unit correctness (4, promoted from strict-xfail).
- Nightly rebase pipeline (`vllm-omni-rebase`) runs as the end-to-end gate: builds 2953 → 2954 → 2955 → 2956 → 2957 at successive fix states.

## Test Result

- Build 2952 (pre-fix): ~25 failed jobs. Build 2953: 86 passed / 18 failed, every failure root-caused. Build 2954 (canceled by the follow-up nightly, near-complete): 85 passed / 11 failed — the entire 11-job code2wav-assert cluster went green, along with vLLM-text serving, LingBot, doc examples, and TTS/Omni perf jobs.
- Builds 2955/2956 validated the inductor patch, fallback removal, cache isolation, bagel re-golden, and gate adjustments. The residual 2956 failures are each fixed above: the duplex-metrics aggregation crash on plain-vLLM perf backends (2dc5e513) and the MiniCPM merge-job timeout from CPU Whisper grading (86bee6d3).
- Build 2957 (first nightly with the stale-drain fix) confirmed the drain fix does not move the seed-TTS WER metric (0.3916, in the pre-fix range) — that regression remains open under Known remaining issues.


vllm_omni/benchmarks/patch/patch.py
```diff
@@ -1687,6 +1688,28 @@ async def limited_request_func(request_func_input, session, pbar):
         async with semaphore:
             return await request_func(request_func_input=request_func_input, session=session, pbar=pbar)
 
+    probe_outputs: list[MixRequestFuncOutput] = []
```
Gaohan123, line 1694: It seems the modification in the file is unrelated to rebase

vllm_omni/diffusion/distributed/parallel_state.py
```diff
@@ -524,13 +526,23 @@ def init_vllm_model_parallel_group(
...
+    kwargs = {}
+    if use_all2all and "use_all2all" in inspect.signature(vllm_parallel_state.init_model_parallel_group).parameters:
+        kwargs["use_all2all"] = True
     return vllm_parallel_state.init_model_parallel_group(
         group_ranks=group_ranks,
         local_rank=local_rank,
         backend=backend,
         group_name=group_name,
         use_device_communicator=True,
+        **kwargs,
```
Gaohan123, line 545: I think we can set all2all explicitly here rather than wrap it in kwargs

vllm_omni/entrypoints/openai/api_server.py
```diff
@@ -274,6 +275,34 @@ async def _get_vllm_config(engine_client: EngineClient) -> Any:
     return getattr(engine_client, "vllm_config", None)
 
 
+class _PureDiffusionLauncherAdapter:
```
Gaohan123: Please move it to utils

vllm_omni/model_executor/models/mammoth_moda2/mammoth_moda2.py
```diff
@@ -59,6 +58,19 @@
 from vllm_omni.transformers_utils.configs.mammoth_moda2 import Mammothmoda2Config
 
 
+def _is_interleaved(config) -> bool:
```
Gaohan123: Please move it to utils

vllm_omni/benchmarks/patch/patch.py
tzhouam: This hunk is rebase alignment, although it doesn't look like it at first glance: `vllm_omni/benchmarks/patch/patch.py` is our patched copy of upstream `vllm/benchmarks/serve.py`, and upstream **v0.27.0 added the `probe_request_rate` probe loop to that file** (`serve.py:798,987,995` at the v0.27.0 tag; absent in 0.26). The change ports the new upstream feature into the patched copy so it stays in sync with its 0.27 base — dropping it would make the patched bench diverge from the file it patches. Added a provenance comment on the hunk in 24dbef7e.

vllm_omni/diffusion/distributed/parallel_state.py
tzhouam: The kwargs indirection is the version guard rather than a style choice: pre-0.27 vLLM's `init_model_parallel_group` doesn't accept `use_all2all`, so passing it as an explicit argument raises `TypeError` there. The `inspect.signature` check forwards it only when the running vLLM supports it (0.27's MoE oracle requires it; older vLLM has neither the kwarg nor the requirement). If you'd prefer it explicit, I can split into two literal calls — `init_model_parallel_group(..., use_all2all=True)` in the supported branch and the bare call otherwise — at the cost of duplicating the call site; or, if this branch is fine requiring vLLM ≥ 0.27, drop the guard and pass it explicitly everywhere. Which do you prefer?

vllm_omni/entrypoints/openai/api_server.py
tzhouam: Done in 24dbef7e — moved to `vllm_omni/entrypoints/utils.py` as `PureDiffusionLauncherAdapter`, keeping its docstring about the vLLM 0.27 launcher-shutdown path.

vllm_omni/model_executor/models/mammoth_moda2/mammoth_moda2.py
tzhouam: Done in 24dbef7e — moved to `vllm_omni/model_executor/models/utils.py` as a public `is_interleaved`, keeping the note that it replicates the helper upstream removed in 26d725c334 so other models hitting the same removal can reuse it.

vllm_omni/diffusion/distributed/parallel_state.py
Gaohan123: just align with upstream

Gaohan123 (approved):
LGTM. Thanks

PR #5983, opened 2026-08-10
[Bugfix] Prevent DiffusionResultPump crash on cancelled futures (#5793)

### LINKED ISSUE #5793: [Bug]: DiffusionResultPump thread dies with InvalidStateError when a request is aborted — engine becomes a zombie (accepts jobs, never runs them)
## Summary

When an in-flight diffusion request is aborted for any reason, the worker's
next (late) result for that request reaches `_result_pump`, which calls
`fut.set_result()` on a future that is already **CANCELLED**. That raises
`concurrent.futures._base.InvalidStateError` *inside the pump thread*, and the
thread dies.

The pump is a singleton, so from that moment the engine is a zombie: the API
server stays healthy (`/health` 200), `POST /v1/videos` accepts jobs, jobs move
to `in_progress` — and nothing ever executes them (GPU idle, progress stays 0).
There is no error surfaced to clients, no recovery, and no way to see the
difference between "queued behind a long render" and "queue is dead" from the
API. Only a server restart recovers.

Reproduced twice on the same deployment via two different abort triggers, with
identical crash signatures.

## Environment

- `vllm/vllm-omni:v0.26.0` docker image (arm64), vLLM 0.26.0 / vLLM-Omni 0.26.0
- plus the two MiniMax-H3 files from merged PR #5737 (H3 online FP8) overlaid —
  i.e. effectively `main`'s H3 model code; `multiproc_executor.py` is stock
- Model: MiniMaxAI/MiniMax-H3 `FL2VA`, `--quantization fp8`, single GPU
- Hardware: NVIDIA GB10 (DGX Spark, sm_121, aarch64, 128 GB unified memory)
- Serve: `vllm serve <FL2VA> --omni --num-gpus 1 --quantization fp8 --vae-use-tiling`

## Crash signature (identical in both reproductions)

```
Exception in thread DiffusionResultPump:
Traceback (most recent call last):
  File "/usr/lib/python3.12/threading.py", line 1075, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.12/threading.py", line 1012, in run
    self._target(*self._args, **self._kwargs)
  File ".../vllm_omni/diffusion/executor/multiproc_executor.py", line 766, in _result_pump
    fut.set_result(msg.output)
  File "/usr/lib/python3.12/concurrent/futures/_base.py", line 544, in set_result
    raise InvalidStateError('{}: {!r}'.format(self._state, self))
concurrent.futures._base.InvalidStateError: CANCELLED: <Future at 0x... state=cancelled>
```

## Reproduction 1 — worker-side generation error triggers the abort

A `t2va` request at `duration=15`, 672×384, 50 steps on a single 128 GB GB10
(a shape that plausibly exhausts activation memory at ~360 frames). Generation
failed at denoise step 0; the engine aborted the request; the pump died on the
worker's next message:

```
ERROR ... async_omni.py:825 ... _raise_nonfatal_error_message(result)
ERROR ... errors.py:81 ... raise fallback(message)
RuntimeError
INFO  ... [Orchestrator] Aborted request(s) ['video_gen_...-...']
Exception in thread DiffusionResultPump:  <signature above>
```

## Reproduction 2 — client-side inter-output timeout triggers the abort

A larger shape ran at ~38 s per denoise step on GB10. `diffusion_engine.py`
waits for streamed outputs with a hardcoded 30-second timeout:

```python
_ASYNC_OUTPUT_TIMEOUT = 30.0  # seconds
...
output = await asyncio.wait_for(asyncio.wrap_future(fut), timeout=_ASYNC_OUTPUT_TIMEOUT)
```

The 49-step denoise **completed** worker-side (tqdm reached 49/49 after 31:12),
but the client-side wait timed out during the run, aborted the finished
request... and the pump died the same way:

```
TimeoutError            (asyncio.wait_for, timeout=_ASYNC_OUTPUT_TIMEOUT)
ERROR ... [AsyncOmni] Orchestrator error for req=video_gen_...-... stage-0:
INFO  ... [AsyncOmni] Aborted request(s) video_gen_...-...
INFO  ... [Orchestrator] Aborted request(s) ['video_gen_...-...']
Exception in thread DiffusionResultPump:  <signature above>
```

## Suggested fixes

1. **Make the pump abort-safe** (the actual bug). In `_result_pump`, a late
   result for a cancelled/aborted request is an expected race, not a fatal
   condition:

   ```python
   if not fut.cancelled():
       fut.set_result(msg.output)
   ```

   (or `try/except InvalidStateError` around `set_result`/`set_exception`,
   plus a debug log). A wrapping `try/except` at the pump loop level would
   also prevent any single bad message from killing the thread permanently.

2. **Make `_ASYNC_OUTPUT_TIMEOUT` configurable** (the trigger for repro 2).
   30 s between streamed outputs assumes datacenter-class step times; a
   single consumer-Blackwell GPU legitimately runs large H3 shapes at
   38+ s/step with a minutes-long VAE tail. An env var
   (e.g. `VLLM_OMNI_ASYNC_OUTPUT_TIMEOUT`) would avoid killing renders that
   are making progress — in repro 2 the engine discarded 31 minutes of
   completed denoise work.

3. Optionally: surface pump liveness in `/health` (or fail jobs when the pump
   is dead) so a zombie engine is detectable by orchestration.

## Workaround we're running

Overlay-mounting a patched `diffusion_engine.py` with the timeout raised to
600 s avoids repro 2's trigger; nothing avoids repro 1's class of trigger
(any worker error), so operationally every failed render requires a server
restart before the queue works again.

anurag12-webster:
## What broke

`MultiprocDiffusionExecutor._result_pump` is the sole reader of the diffusion worker result queue, running as a singleton background thread. Every dispatch branch resolves a shared future with:

```python
if fut is not None and not fut.done():
    fut.set_result(result)  # or fut.set_exception(...)
```

The `not fut.done()` check and the `set_result()`/`set_exception()` call are not atomic. A consumer awaiting the future via `asyncio.wrap_future()` under `asyncio.wait_for()` (an inter-output timeout, or a request abort) can cancel it in the gap between the two. When that happens, `set_result()`/`set_exception()` raises `concurrent.futures.InvalidStateError`, which nothing catches — killing the pump thread permanently. Since the pump is a singleton, the engine then accepts and "runs" jobs that silently never complete (`/health` stays 200, GPU sits idle).

## Repro

```
concurrent.futures._base.InvalidStateError: CANCELLED: <Future at 0x... state=cancelled>
  File "vllm_omni/diffusion/executor/multiproc_executor.py", line 883, in _result_pump
    fut.set_result(msg.output)
  File ".../concurrent/futures/_base.py", line 544, in set_result
    raise InvalidStateError(...)
```

Reported originally in #5793 via two real triggers: a worker-side generation error aborting a request, and a client-side inter-output timeout (`_ASYNC_OUTPUT_TIMEOUT`) firing while the worker was still running. Both land on the same race.

## Root cause

Check-then-act race on `concurrent.futures.Future` state across threads: the pump's `done()` check and its resolve call aren't atomic, so a concurrent `cancel()` from the consumer side can land in between.

## Fix

Added `_try_set_result`/`_try_set_exception` helpers that wrap `set_result`/`set_exception` in a narrow `except concurrent.futures.InvalidStateError`, dropping the late resolution instead of crashing the thread. Applied at every call site in `_result_pump` that resolves a future pulled from the shared `_rpc_futures`/`_output_futures` dicts, plus the same pattern in `shutdown()`'s cleanup loop for consistency. Left untouched the two `set_result` calls on freshly-created local `Future()` objects that aren't shared with any other thread yet — no race possible there.

## Testing

Added `tests/diffusion/test_multiproc_result_pump.py` (CPU-only, no GPU needed): a `_RacyFuture` subclass whose `done()` deterministically lies (`False`) while genuinely cancelled underneath, reproducing the exact race window without relying on real thread timing.

Verified on an RTX 3090 (vLLM 0.26.0, vllm-omni installed per docs):
- With the fix: `tests/diffusion/test_multiproc_result_pump.py` — 3 passed.
- With the fix reverted: all 3 fail with the exact `InvalidStateError` traceback from the original report, at the same `fut.set_result(msg.output)` call site.
- Neighboring suite `tests/diffusion/test_multiproc_engine_concurrency.py` — 60 passed, no regressions.
- `ruff check`, `ruff format --check`, and the full `pre-commit` hook suite pass on both changed files.

Fixes #5793

### DIFF
--- a/vllm_omni/diffusion/executor/multiproc_executor.py
+++ b/vllm_omni/diffusion/executor/multiproc_executor.py
@@ -12,7 +12,7 @@ import weakref
 from collections.abc import Callable
 from dataclasses import dataclass
 from multiprocessing.synchronize import Event
-from typing import TYPE_CHECKING, Any, TypeVar, cast
+from typing import TYPE_CHECKING, Any, cast
 
 import zmq
 from vllm.distributed.device_communicators.shm_broadcast import Handle, MessageQueue
@@ -24,6 +24,7 @@ from vllm_omni.diffusion.data import SHUTDOWN_MESSAGE, AsyncDiffusionOutput, Asy
 from vllm_omni.diffusion.executor.abstract import DiffusionExecutor
 from vllm_omni.diffusion.ipc import DIFFUSION_RPC_RESULT_ENVELOPE, unpack_diffusion_output_shm
 from vllm_omni.diffusion.sched.request_scheduler import build_request_batch_sampling_params_key
+from vllm_omni.diffusion.utils.future_utils import try_set_exception, try_set_result
 from vllm_omni.diffusion.worker import WorkerProc
 
 if TYPE_CHECKING:
@@ -48,26 +49,6 @@ def _is_empty_dp_prompt(prompt: object) -> bool:
     if isinstance(prompt, dict) and "prompt" in prompt:
         return not prompt["prompt"]
     return False
-
-
-_ResultT = TypeVar("_ResultT")
-
-
-def _try_set_result(fut: concurrent.futures.Future[_ResultT], result: _ResultT) -> None:
-    # fut may be cancelled concurrently (e.g. asyncio.wait_for timeout) between
-    # the caller's fut.done() check and this call; drop the late result instead
-    # of crashing the pump thread.
-    try:
-        fut.set_result(result)
-    except concurrent.futures.InvalidStateError:
-        logger.debug("Dropping late result for already-resolved/cancelled future")
-
-
-def _try_set_exception(fut: concurrent.futures.Future[Any], exc: BaseException) -> None:
-    try:
-        fut.set_exception(exc)
-    except concurrent.futures.InvalidStateError:
-        logger.debug("Dropping late exception for already-resolved/cancelled future")
 
 
 @dataclass
@@ -890,9 +871,9 @@ class MultiprocDiffusionExecutor(DiffusionExecutor):
                     fut = self._rpc_futures.pop(msg.rpc_id, None) if msg.rpc_id else None
                 if fut is not None and not fut.done():
                     if msg.error:
-                        _try_set_exception(fut, RuntimeError(msg.error))
+                        try_set_exception(fut, RuntimeError(msg.error))
                     else:
-                        _try_set_result(fut, msg)
+                        try_set_result(fut, msg)
             elif msg.kind == AsyncOutputKind.OUTPUT_READY:
                 batch_id = msg.async_output_id
                 with self._futures_lock:
@@ -923,9 +904,9 @@ class MultiprocDiffusionExecutor(DiffusionExecutor):
                             pending = self._output_futures.pop(batch_id, None)
                             if pending is not None and not pending.done():
                                 if exc is not None:
-                                    _try_set_exception(pending, exc)
+                                    try_set_exception(pending, exc)
                                 else:
-                                    _try_set_result(pending, output_result)
+                                    try_set_result(pending, output_result)
                             else:
                                 fut = concurrent.futures.Future()
                                 if exc is not None:
@@ -953,7 +934,7 @@ class MultiprocDiffusionExecutor(DiffusionExecutor):
             with self._futures_lock:
                 pending = self._output_futures.pop(per_req_id, None)
                 if pending is not None and not pending.done():
-                    _try_set_result(pending, per_req_result)
+                    try_set_result(pending, per_req_result)
                 else:
                     fut: concurrent.futures.Future = concurrent.futures.Future()
                     fut.set_result(per_req_result)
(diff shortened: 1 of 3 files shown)


anurag12-webster:
@vllm-omni-review-bot

anurag12-webster:
@Isotr0py @princepride PTAL

anurag12-webster:
hi @Isotr0py @princepride can you take a look at this

anurag12-webster:
rebased onto main. #6023 rewrote `_result_pump`/`_deliver_batch_split` and merged before this pr, so the `InvalidStateError` fix here got dropped along with the old code shape it patched. the original #5793 crash was reproducible on main again.

reapplied the same fix on top of #6023's new code, and added the regression test to `tests/diffusion/test_result_pump.py` (the test file #6023 introduced) instead of a separate file.

verified on a real gpu: with the fix reverted, all 3 new tests fail with the exact `InvalidStateError` from the original report, at the 3 call sites in the new code. with the fix applied, all 21 tests in that file pass (3 new + 18 existing, no regressions).

vllm_omni/diffusion/executor/multiproc_executor.py
```diff
@@ -50,6 +50,26 @@ def _is_empty_dp_prompt(prompt: object) -> bool:
     return False
 
 
+_ResultT = TypeVar("_ResultT")
+
+
+def _try_set_result(fut: concurrent.futures.Future[_ResultT], result: _ResultT) -> None:
```
Gaohan123: I suggest move these 2 tool functions into vllm_omni/diffusion/utils

anurag12-webster: okay sure

anurag12-webster: done, moved to vllm_omni/diffusion/utils/future_utils.py

Gaohan123 (approved):
LGTM. Thanks

anurag12-webster:
thankyouu!! @Gaohan123

PR #5991, opened 2026-08-10
[NPU][Model] Add distilled 4-step sigma schedule support for MiniMax H3 t2va

Huangzjun:
## Purpose

This PR adds **4-step distilled checkpoint inference support** for MiniMax H3 FL2VA models trained with DMD2.

The existing pipeline generates sigma schedules from a uniform rectified-flow schedule. For few-step distilled models, this does not match the schedule used during training.

Example distilled schedule:

```text
base_schedule = [1.0, 0.7, 0.4, 0.15, 0.0]
```

This PR adds checkpoint-level `base_schedule` support so inference uses the same sigma trajectory as training.

### Changes

* **`time_request.py`**

  * Add optional `base_schedule` support to `_time_shift_sigmas()`.
  * Apply the existing H3 time-shift mapping to explicit schedule positions.
  * Validate schedule order and endpoints.

* **`pipeline_minimax_h3.py`**

  * Read `_minimax_h3.base_schedule` from `FL2VA/model_index.json`.
  * Use it for both video and audio sigma schedules.
  * Infer the number of denoising steps from the schedule.
  * Reject mismatched explicit `num_inference_steps`.

* **Tests**

  * Verify distilled video/audio sigma values.
  * Verify malformed schedules are rejected.

No changes are required to DiT weight loading or the denoising solver.

### Checkpoint metadata

```json
{
  "_minimax_h3": {
    "partition": "fl2va",
    "tasks": ["t2va", "fl2va"],
    "sigma_shift_scales": {
      "video": 12.0,
      "audio": 3.0
    },
    "base_schedule": [1.0, 0.7, 0.4, 0.15, 0.0]
  }
}
```

## Usage

```bash
vllm-omni serve <path-to-merged-fl2va-root> \
  --task-type t2va
```

Notes:

* Omit `num_inference_steps`; the checkpoint metadata selects the 4-step schedule automatically.
* Use `--task-type t2va` to load the distilled FL2VA partition.
* Official H3 checkpoints without `base_schedule` keep the existing 50-step behavior.

## Test Plan

**vLLM Version:** `0.26.0`
**vLLM-Omni Commit:** `<branch-head>`

1. Serve a merged DMD2 FL2VA checkpoint with `base_schedule`.
2. Verify inference runs with the expected 4 denoising steps.
3. Verify mismatched `num_inference_steps` returns an error.
4. Verify the official 50-step H3 checkpoint is unaffected.
5. Run:

```bash
pytest tests/diffusion/models/minimax_h3/test_minimax_h3_contract.py \
  -k "base_schedule or shifted_sigma" -v
```

## Test Result

**Hardware:** `<hardware>`

**Configuration:**

* T2VA
* 1344×768
* 124 frames
* `base_schedule=[1.0,0.7,0.4,0.15,0.0]`
* `quality=lossless`
* `cfg_parallel_size=1`

### Distilled 4-step inference

https://github.com/user-attachments/assets/cd9e4994-9215-4d7b-aad5-b416f0b4f509

### 50-step teacher baseline

https://github.com/user-attachments/assets/2bdd4169-075d-4c93-ad02-942b796929fd

## Related PR

* #5706 — RainFusion attention and INT8 online quantization for MiniMax H3
* MiniMax H3 base serving integration


hsliuustc0106:
I remember we have a general DMD2 pipelinemixin, please check

Huangzjun:
Thanks for the pointer — I did check DMD2PipelineMixin. The main blocker is that it works by swapping in a DMD2EulerScheduler, but H3 doesn't hold a self.scheduler at all (sigmas are built functionally and stepped by minimax_h3_rf_v_to_x0 / minimax_h3_euler_eta0_step), so the swap wouldn't take effect. H3 also needs two sigma schedules from one shared base u (video shift=12, audio shift=3), which the single-modality DMD2Config doesn't express. So I kept the H3-local path and just added the missing input: base positions from _minimax_h3.base_schedule. @hsliuustc0106

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
```diff
@@ -1755,7 +1764,17 @@ def forward(self, request: DiffusionRequestBatch) -> DiffusionOutput:
                     ref_audio_t = audio_lengths[0]
 
         seed = int(sampling.seed if sampling.seed is not None else 42)
-        num_steps = int(sampling.num_inference_steps or 50)
+        base_schedule = self.default_base_schedule
```
david6666666: In combined mode this schedule comes from FL2VA, but it is also used for Ref2VA requests. A distilled FL2VA paired with a regular Ref2VA would therefore force the Ref2VA transformer onto four steps. Could we select the schedule from the active task/partition instead?

vllm_omni/diffusion/models/minimax_h3/time_request.py
```diff
@@ -31,18 +33,46 @@ def _audio_latent_t(duration_seconds: float) -> int:
     return int(round(float(duration_seconds) * 40.0))
 
 
+def _validate_base_schedule(base_schedule: Sequence[float]) -> list[float]:
+    values = [float(value) for value in base_schedule]
+    if len(values) < 2:
+        raise ValueError("MiniMax H3 base_schedule needs at least 2 entries")
+    if values[0] != 1.0 or values[-1] != 0.0:
+        raise ValueError("MiniMax H3 base_schedule must start at 1.0 and end at 0.0")
+    if any(curr <= nxt for curr, nxt in zip(values, values[1:], strict=True)):
```
david6666666: An interior NaN passes this ordering check because comparisons with NaN are false, then propagates into every denoising step. Please reject non-finite schedule entries before checking the order.

hsliuustc0106:
> Thanks for the pointer — I did check DMD2PipelineMixin. The main blocker is that it works by swapping in a DMD2EulerScheduler, but H3 doesn't hold a self.scheduler at all (sigmas are built functionally and stepped by minimax_h3_rf_v_to_x0 / minimax_h3_euler_eta0_step), so the swap wouldn't take effect. H3 also needs two sigma schedules from one shared base u (video shift=12, audio shift=3), which the single-modality DMD2Config doesn't express. So I kept the H3-local path and just added the missing input: base positions from _minimax_h3.base_schedule. @hsliuustc0106

but at least we should consider how to integrate with the mixin. If it does not satisfy our need, you can propose how to modify it

vllm_omni/diffusion/models/minimax_h3/time_request.py
hsliuustc0106: [P1] This rejects every valid schedule. `values` has length n while `values[1:]` has length n-1, so `strict=True` raises `ValueError: zip() argument 2 is shorter than argument 1` after the comparisons finish. The new valid-schedule test reproduces this. Please use non-strict pairwise iteration (and keep the malformed-order check).

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
hsliuustc0106: [P1] This unconditional attribute read breaks the existing contract test, which constructs `MiniMaxH3Pipeline` with `object.__new__` and does not initialize `default_base_schedule`; the full contract file now fails with `AttributeError`. Please update that fixture (and add distilled-forward coverage), or provide a safe class-level/default fallback.

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
```diff
@@ -1755,7 +1764,17 @@ def forward(self, request: DiffusionRequestBatch) -> DiffusionOutput:
...
+        if base_schedule is None:
+            num_steps = int(sampling.num_inference_steps or 50)
+        else:
+            num_steps = len(base_schedule)
```
hsliuustc0106: [P1] This counts sigma boundaries rather than denoising steps. For the advertised `[1.0, 0.7, 0.4, 0.15, 0.0]` schedule, the denoise loop runs `len(sigmas) - 1 == 4` iterations, but this sets `num_steps` to 5, rejects an explicit `num_inference_steps=4`, and passes 5 to Cache-DiT. Please use the actual denoising-step count for request/cache semantics while retaining the full boundary schedule for the solver.

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
```diff
@@ -605,6 +606,11 @@ def __init__(
         shifts = release.get("sigma_shift_scales") or {}
         self.default_video_shift = float(shifts.get("video", 12.0))
         self.default_audio_shift = float(shifts.get("audio", 3.0))
+        # Distilled releases pin their own few-step rectified-flow positions;
+        # the uniform schedule derived from num_inference_steps does not match
+        # what such a checkpoint was trained on.
+        base_schedule = release.get("base_schedule")
+        self.default_base_schedule = minimax_h3_validate_base_schedule(base_schedule) if base_schedule else None
```
hsliuustc0106: [P2] An explicitly present empty schedule bypasses validation because `[]` is falsey, so the checkpoint silently falls back to the legacy uniform 50-step schedule. Please distinguish an absent key from `base_schedule=[]` (for example, validate when `base_schedule is not None`).

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
```diff
@@ -605,6 +606,11 @@ def __init__(
         shifts = release.get("sigma_shift_scales") or {}
         self.default_video_shift = float(shifts.get("video", 12.0))
         self.default_audio_shift = float(shifts.get("audio", 3.0))
+        # Distilled releases pin their own few-step rectified-flow positions;
+        # the uniform schedule derived from num_inference_steps does not match
+        # what such a checkpoint was trained on.
+        base_schedule = release.get("base_schedule")
```
hsliuustc0106: [P1] In combined mode, `release` was loaded from the FL2VA `model_index.json`, but this value is later used for every task, including Ref2VA. If the two partitions have different schedule metadata, Ref2VA will run with the FL2VA schedule. Please retain schedule metadata per partition/task or reject combined checkpoints whose schedule contracts differ.

hsliuustc0106:
I recommend integrating at the schedule layer rather than making MiniMax H3 inherit `DMD2PipelineMixin` as-is.

The current mixin is tied to scheduler-backed pipelines: `__init_dmd2__` installs `DMD2EulerScheduler`, and the mixin’s request/forward path rewrites `num_inference_steps`, guidance parameters, and `flow_shift`. H3 has no `self.scheduler`; it constructs coupled video/audio sigma arrays and uses separate `flow_shift` and `audio_flow_shift` values. Direct inheritance would therefore be either ineffective or likely to break H3-specific behavior.

Suggested refactor:

1. Extract a reusable `DMD2SigmaSchedule` (or equivalent) from `DMD2Config). It should validate the continuous RF `base_schedule`, expose `num_inference_steps = len(base_schedule) - 1`, and generate shifted sigmas. This should remain distinct from `denoising_timesteps`, which represents integer scheduler timesteps.
2. Keep `DMD2PipelineMixin` as the scheduler-backed adapter for Flux/Wan/LTX pipelines, but make it consume the shared schedule/config utility.
3. In H3, load and retain one schedule per partition/task (FL2VA and Ref2VA), select the active partition’s schedule during request resolution, and use that same base schedule to derive the video and audio sigmas with their respective shifts.
4. Keep H3’s custom denoise loop and request sanitization. Add tests for valid/invalid/empty schedules, combined-mode per-partition selection, and the advertised 4-step schedule with 5 sigma boundaries.

This shares the DMD2 metadata and validation contract without forcing H3 into the existing scheduler/MRO contract. I would avoid expanding the full pipeline mixin in this PR unless we first define explicit hooks for schedule resolution and request sanitization.

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
Huangzjun: Good catch — fixed. Schedule metadata is now stored per partition (fl2va / ref2va), and forward() selects it via _base_schedule_for_task(task), so a distilled FL2VA no longer forces Ref2VA onto 4 steps.

vllm_omni/diffusion/models/minimax_h3/time_request.py
Huangzjun: Fixed — _validate_base_schedule() now rejects non-finite entries before the monotonicity check, so interior NaN/inf no longer slip through.

Huangzjun: Fixed — switched to non-strict pairwise iteration. The old strict=True raised on every valid schedule because values[1:] is one element shorter; the malformed-order tests were also masking this because any() short-circuited before the zip finished.

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
Huangzjun: replaced the instance-only attribute with a class-level _base_schedule_by_partition = {} default, so partially constructed pipelines (including the object.__new__ fixtures) fall back to the legacy uniform path without an AttributeError.

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
Huangzjun: Fixed — when a distilled schedule is present, num_steps is now len(base_schedule) - 1 (4 denoising hops for the 5-point schedule). The full boundary list is still passed to the solver, and Cache-DiT / request validation use the denoising-step count. Added forward coverage for this.

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
Huangzjun: _read_base_schedule() now distinguishes a missing key (None → legacy uniform) from an explicit empty list, which is validated and rejected.

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
Huangzjun: same change as the combined-mode comment above: FL2VA and Ref2VA each load their own base_schedule from their respective model_index.json, and the active task picks the matching partition.

Huangzjun:
> I recommend integrating at the schedule layer rather than making MiniMax H3 inherit `DMD2PipelineMixin` as-is.

Thanks, this framing is much better than what I had — I've gone with the schedule-layer approach.

(1) Shared schedule utility. Added DMD2SigmaSchedule in vllm_omni/diffusion/models/dmd2/schedule.py. It validates the continuous RF base_schedule (length, finiteness, 1.0 → 0.0 endpoints, strict monotonicity), exposes num_inference_steps = len(base_schedule) - 1, and generates shifted sigmas via shifted_sigmas(shift_scale). It's kept deliberately separate from DMD2Config.denoising_timesteps and is torch-free, since these are continuous positions rather than integer scheduler timesteps — that's exactly what lets one schedule serve several coupled modalities.

(3) H3 consumes it. H3 no longer carries its own validator. _read_base_schedule is now just DMD2SigmaSchedule.from_metadata(release), one schedule is retained per partition (FL2VA / Ref2VA), the active one is selected during request resolution, and that same base derives the video and audio sigmas at shift 12 and 3 respectively.

(4) H3-specific behavior untouched. The custom denoise loop and request sanitization are unchanged. Tests cover valid/invalid/empty schedules, per-partition selection in combined mode, and the 4-step schedule with its 5 sigma boundaries; the shared class also gets its own test module.

(2) Deferred, per your last paragraph. I haven't touched DMD2PipelineMixin behavior in this PR. Wiring the scheduler-backed adapter onto the shared utility makes sense, but it needs the explicit schedule-resolution and request-sanitization hooks you mentioned, so I'd rather do it as a focused follow-up than bundle it here. Happy to open that one if you'd like.

vllm_omni/diffusion/sched/sigma_schedule.py
```diff
@@ -0,0 +1,82 @@
+# SPDX-License-Identifier: Apache-2.0
```
david6666666, line 1: I think we shoud move to vllm_omni/diffusion/sched

david6666666 (changes requested):

vllm_omni/diffusion/sched/sigma_schedule.py
Huangzjun: Done — moved DMD2SigmaSchedule to vllm_omni/diffusion/sched/sigma_schedule.py and updated all imports/tests. No behavior change.

Gaohan123 (approved):
LGTM. Thanks

Gaohan123:
The CI error also happened in the latest nightly CI: https://buildkite.com/vllm/vllm-omni/builds/13407/canvas?sid=019fecd9-14c2-476d-be9c-1f55634370e3&tab=output

PR #6002, opened 2026-08-10
[Refactor] Init and use per_request_initial_chunk_size_override helper 

NickCao:
## Purpose

So that the per request override for initial_chunk_size is honored.

## Test Plan

**vLLM Version:** 0.26.0

**vLLM-Omni Commit:** 0dbf8fac3564829455738f6b729752eab1c20203

`pytest tests/model_executor/stage_input_processors/test_tts_utils.py`

## Test Result

`8 passed, 15 warnings in 0.02s`

### DIFF
--- a/tests/model_executor/stage_input_processors/test_tts_utils.py
+++ b/tests/model_executor/stage_input_processors/test_tts_utils.py
@@ -1,0 +1,63 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+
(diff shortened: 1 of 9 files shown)


NickCao:
The remaining models:

Does not support initial chunk frames at all:
- cosyvoice3.py
- glm_tts.py
- higgs_audio_v2.py
- mimo_audio.py
- minicpmo_4_5_omni.py

Support initial chunk frames, but in a custom way.
- voxtral_tts.py — uses codec_chunk_frames_at_begin
- ming_tts.py — uses initial_latent_chunk_size

PR #6005, opened 2026-08-10
[Feat][MiniCPM-O-4.5][StepAudio2]Vendor Flow Component

sphinxkkkbc:

## Purpose
This PR integrates the Vendor Flow component from the external `step-audio2` library to improve inference performance. 
Replace separate Q/K/V linear layers with fused `QKVParallel`. Remove sync in `timestep_embedding` by registering frequency embeddings. This is the first step of a broader optimization effort. 

Most of the diff in this PR comes from vendoring the existing implementation. The performance-related changes are primarily contained in `dit.py` and `step_audio2_token2wav.py`.

Profiling shows that LayerNorm and AdaLN (modulate) are also hot paths. Further optimizations are planned but not yet ready. The current LayerNorm/AdaLN implementation in diffusion.layers is naive, so it is not included in this PR.

## Note
The external `step-audio2` library includes a CUDA Graph implementation for `DiT.blocks_forward_chunk()`. This PR intentionally does not include it, as there is already a separate PR working on the CUDA Graph integration.

## Test Plan

Compare three versions:
- **Baseline (main)**: vLLM-Omni main branch without vendor optimizations
- **Vendor Commit**: First commit integrating `step-audio2` library (pure vendor, no additional changes)
- **Optimized Commit (latest)**: Vendor + QKVParallel + remove sync in timestep_embedding
 
**Benchmark: Performance and Accuracy**

Serve:
```
vllm serve openbmb/MiniCPM-o-4_5 --omni \
    --trust-remote-code \
    --stage-init-timeout 600 \
    --host 0.0.0.0 --port 8091
```
Benchmark command:

```
vllm bench serve \
  --omni \
  --port 8091 \
  --trust-remote-code \
  --max-concurrency 1 \
  --num-warmup 3 \
  --dataset-name seed-tts \
  --dataset-path /workspace/seed-tts-eval \
  --num-prompts 32 \
  --no-oversample \
  --seed-tts-wer-eval \
  --seed-tts-wer-save-items \
  --endpoint /v1/chat/completions \
  --backend openai-chat-omni \
  --percentile-metrics ttft,tpot,itl,e2el,audio_ttfp,audio_rtf \
  --extra_body '{"modalities": ["text", "audio"], "chat_template_kwargs": {"enable_thinking": false, "use_tts_template": true}}'
```

**vLLM Version:**
0.26.0
**vLLM-Omni Commit:**
65a607d859a9a8888f4f451349b0900797c5fbea
## Test Result

| Metric | Baseline (main) | Vendor Commit | Vendor Delta | Optimized Commit | Optimized Delta |
| --- | --- | --- | --- | --- | --- |
| **Mean E2E Latency (ms)** | 1341.41 | 1338.55 | -0.21% | 1287.64 | -4.01% |
| **Mean TTFT (ms)** | 217.12 | 217.58 | +0.21% | 216.99 | -0.06% |
| **Mean AUDIO_TTFP (ms)** | 666.20 | 676.62 | +1.56% | 644.38 | -3.28% |
| **Mean AUDIO_RTF** | 0.31 | 0.31 | +0.00% | 0.30 | -3.23% |
| **Mean WER** | 0.0179 | 0.0179 | +0.00% | 0.0226 | +26.26% |

A/B testing indicates that the WER regression mainly comes from the parallel linear changes: 0.0179 at baseline, 0.0191 with QKVParallelLinear, and 0.0226 with both QKVParallelLinear and RowParallelLinear. Registering the timestep frequencies as a buffer showed no consistent WER impact; the standalone difference appears to be evaluation variance.


hsliuustc0106:
This PR touches **vllm_omni/model_executor/** (9 files). Based on CODEOWNERS coverage of the changed files, the most-related reviewers appear to be:

> @linyueqian @ZeldaHuang @princepride

sphinxkkkbc:
@amy-why-3459 PTAL,thanks!

amy-why-3459:
Thanks for the clear A/B — it already answers the WER question:

- vendor-only commit: WER unchanged (0.0179)
- +QKVParallelLinear: 0.0191 (+6.7%)
- +QKVParallelLinear + RowParallelLinear: 0.0226 (+26%)

So the regression is not from vendoring Step-Audio2 Flow; it comes from the
second commit's parallel-linear rewrites, with DiTAttention.proj
(RowParallelLinear) as the larger jump and QKV fusion as a real but smaller
hit. freqs-as-buffer is fine (your note matches that).

Request changes:

1. Please do not merge a "perf" change that trades ~4% E2E for +26% WER.
   Either (a) fix the parallel-linear path until WER is within a tight bound of
   baseline, or (b) land vendor-only first and split QKV/RowParallel to a
   follow-up.

2. Add parity tests before claiming the rewrite is equivalent:
   - separate Linear q/k/v vs QKVParallelLinear (same weights/inputs)
   - nn.Linear proj vs RowParallelLinear (same weights/inputs)
   - loader coverage: fused weights must match concat(q,k,v); restore
     strict-style missing/unexpected checks (old path used
     load_state_dict(..., strict=True))

3. Align TP flags: Conformer QKV uses disable_tp=True, but DiT to_qkv/proj
   do not. Flow is not meant to be TP-sharded — please disable_tp on DiT
   as well (or hard-fail if tp>1).

Happy to re-review once WER is back in line with the vendor-only baseline
and the parity tests are in.

sphinxkkkbc:
Added an end-to-end Flow parity test in `tests/model_executor/models/step_audio2/test_flow_parity.py`; all 4 tests pass. @amy-why-3459, PTAL. Thanks!

One correction regarding the WER results: I previously attributed the 0.0179 → 0.0209 change observed with the timestep-frequency `register_buffer` optimization to benchmark variance. However, the result remained reproducible across multiple runs and further investigation, so that conclusion was not sufficiently supported. I have not identified the root cause yet.

Similar to the QKVParallel and RowParallel changes, this optimization should be investigated separately with dedicated numerical-parity and WER validation. For this PR, I propose landing the vendor-only implementation first and addressing these performance optimizations in follow-up PRs.

sphinxkkkbc:
Marked as WIP. I will reopen this PR after investigating the performance–accuracy trade-off. If the accuracy regression cannot be avoided, I will provide detailed A/B results and document the trade-off explicitly.

The following optimization paths will be evaluated independently:
1. Precomputing and caching timestep frequencies.
2. Fused QKV projection versus separate Q/K/V projections.
3. The standalone LayerNorm inside `CausalConvBlock`.
4. DiT normalization fusion, including Q/K LayerNorm, AdaLN, and gated-residual AdaLN.

sphinxkkkbc:
@amy-why-3459 Marked as Ready.PTAL, thanks!

I need to correct my earlier interpretation of the WER results.

The previous WER comparisons were not reliable because Stage 2 still contained uncontrolled randomness. Across repeated service restarts, I observed different WER values for the same configuration. In some runs, the fused-QKV version even achieved a lower WER than the naive implementation.

The relevant random sources were:

- the initial noise used by CFM;
- the runtime phase and noise generation inside HiFT.

For a controlled A/B test, I fixed both the CFM and HiFT random states to seed 42. With the request seed, dataset order, concurrency, and Stage 2 noise fixed, repeated service restarts produced stable and reproducible results.

Under these controlled conditions, all four configurations produced exactly the same mean WER:

| Configuration | Mean WER |
|---|---:|
| Naive vendor implementation | 0.0226 |
| Timestep frequency buffer only | 0.0226 |
| Frequency buffer + fused QKV in DiT | 0.0226 |
| Frequency buffer + fused QKV in both DiT and Conformer | 0.0226 |

The per-utterance WER results were also unchanged. Therefore, the previously reported WER regression was caused by an uncontrolled benchmark setup and should not be attributed to QKV fusion or the timestep-frequency buffer.

The performance improvement remains reproducible, with no observed accuracy regression.

I also experimented with fused LayerNorm, AdaLN, and gated-residual kernels using CuTe and Triton which mentioned before. The current implementations did not outperform the existing PyTorch path for the representative tensor shapes, mainly because the native operations are already fast and the custom kernels introduced additional launch and dispatch overhead. Further optimization opportunities remain in the Flow module, including larger-scope fusion and more specialized kernels.


Also, the comments you left previously have all been addressed. `disable_tp` is now set, and the parity/weight_load tests are added.

```
pytest -v \
  tests/model_executor/models/step_audio2/test_qkv_fusion.py \
  tests/model_executor/models/step_audio2/test_flow_parity.py
```

with `10 passed`

PR #5843, opened 2026-08-06
[Core][Refactor][Diffusion] refactor request scheduler's admission wait policy

fhfuih:
## Purpose

The responsibility boundary of request batching is not clear. How to decide the waiting in written in diffusion engine, whereas it is clearer to reside in scheduler (same as step batching). This refactoring add such an abstraction.

## Test Plan

**vLLM Version:** 0.26

**vLLM-Omni Commit:**

```
commit f2e3d2e25bebc0c7110ccb945d7dcf4be81bf6db (upstream/main, upstream/HEAD, origin/main, origin/HEAD, main)
Author: fan2956 <zhoufan53@huawei.com>
Date:   Thu Aug 6 11:39:27 2026 +0800

    [Perf] Minimax-H3 support fused RMSNorm and RoPE opt (#5801)

    Signed-off-by: Fan <fan@FandeMacBook-Air.local>
    Co-authored-by: Fan <fan@FandeMacBook-Air.local>

```

Added unit tests in test_diffusion_scheduler to test the abstracted admission waiting policy.

## Test Result

All passed locally

**BEFORE SUBMITTING:** read [CONTRIBUTING.md](https://github.com/vllm-project/vllm-omni/blob/main/CONTRIBUTING.md) and run the [precheck-pr skill](https://github.com/vllm-project/vllm-omni/blob/main/.claude/skills/precheck-pr/SKILL.md) with the code agent for a self-check against project conventions.
(anything written below this line will be removed by GitHub Actions)

PR #5746, opened 2026-08-04
[Refactor] Add Ming Flash Omni TTS adapter

sphinxkkkbc:
PLEASE FILL IN THE PR DESCRIPTION HERE.

## Purpose
Register a serving adapter for `ming_flash_omni_tts` and remove its legacy validation and request-building branches from `serving_speech.py`.Preserves the existing request validation, prompt construction, and generation behavior without changing the adapter contract.

This PR moves request validation and routing into the adapter while continuing to reuse the existing `_build_ming_flash_omni_prompt()` helper. Consistent with the other migrated adapters, relocating the model-specific prompt builder is left to a follow-up PR, but it can be included here if preferred.

Removed `test_ming_flash_omni_not_migrated()` as it is no longer applicable

## Test Plan
`pytest -v tests/entrypoints/openai_api/test_serving_speech.py`
`pytest -v tests/entrypoints/openai_api/test_tts_adapter.py`
Covers basic request validation and prompt-building delegation through the adapter.

**vLLM Version:**
0.26.0
**vLLM-Omni Commit:**
9bd11895f7948bc0630e55c65a2057f23fad8d71
## Test Result
207 passed
9 passed

cc @linyueqian for review

**BEFORE SUBMITTING:** read [CONTRIBUTING.md](https://github.com/vllm-project/vllm-omni/blob/main/CONTRIBUTING.md) and run the [precheck-pr skill](https://github.com/vllm-project/vllm-omni/blob/main/.claude/skills/precheck-pr/SKILL.md) with the code agent for a self-check against project conventions.
(anything written below this line will be removed by GitHub Actions)

PR #6008, opened 2026-08-10
[BugFix] Don't fail the TTS ratchet when the branch count goes down

linyueqian:
## Purpose

The TTS ratchet I added in #5682 fails when the branch count goes **down**. That is backwards, and it broke `main`.

`check_tts_adapter.py` counts `self._tts_model_type ==` comparisons and compares them to a hardcoded budget. It errored if the count was over budget — correct — and also if it was under, with a message demanding someone hand-edit the constant. So removing a branch, which is the entire point of the tool, failed the build.

That is what happened. #5746 migrated the last legacy detector and the counts dropped to 27/0. It was green on its own base, which predated the ratchet, so nothing in its pipeline could have caught it. `main` went red on merge and stayed red **5h19m across 5 commits**:

| time (+0800) | commit | pre-commit |
|---|---|---|
| 08-09 21:30 | `ebe93f5d1` ratchet lands, budgets 29/1 | pass |
| 08-10 09:47 | `11f633aa8` #5746 migrates the last detector | **fail** |
| 08-10 10:02 | `36135ef09` LingBot-Video | fail |
| 08-10 10:04 | `e1aa6eae7` WeChat QR docs | fail |
| 08-10 11:26 | `e28bf8ff8` bagel CI | fail |
| 08-10 11:27 | `c588208cc` lora support | fail |
| 08-10 15:06 | `33dd9a5a3` #5843 carries the fix | pass |

Four unrelated PRs merged onto a red main, and #5843 — a diffusion scheduler refactor — had to bundle a change to my tool to get its own CI green. Sorry about that, @fhfuih.

## What changed

Only an increase fails now. A decrease prints a reminder and exits 0.

Two more problems turned up while testing the fix:

**The reminder was invisible.** pre-commit hides output from hooks that pass unless the hook is `verbose`. Verified on pre-commit 4.3.0 — a passing hook's stderr is shown only with `verbose: true`. Without it the ratchet would have kept its ceiling but lost all tightening pressure, which is worse than the behaviour I was trying to preserve.

**The checker could not check itself.** The hook's `files:` pattern listed only `serving_speech.py` and `tts_adapters/__init__.py`, so a commit that edited nothing but a budget constant ran no hook at all — `(no files to check) Skipped`. A typo'd budget would merge and land on whoever touched `serving_speech.py` next. Same shape as the bug above: the person who pays is not the person who caused it.

## Test plan

| budget vs actual | expected | exit |
|---|---|---|
| 40 > 27 (count went down) | pass with a reminder | 0 |
| 27 = 27 | pass silently | 0 |
| 5 < 27 (count went up) | **fail**, list the offenders | 1 |

Checked against `main` as-is (27/0) — exits 0.

## What this does not fix

The budget is still a hand-edited constant, so the check's verdict still depends on repo-wide counts and therefore on merge order. Two individually-green PRs can still collide. Both reviewers I consulted converged on the same better answer: derive the baseline from the merge base and require `current <= merge_base`, so the invariant is local to the PR and merge order stops mattering. That is a bigger change than a hotfix should carry, so I would rather land this and do it separately — happy to be told otherwise.

cc @sphinxkkkbc @Sy0307

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -66,7 +66,10 @@ repos:
       entry: python tools/pre_commit/check_tts_adapter.py
       language: python
       pass_filenames: false
-      files: ^vllm_omni/entrypoints/openai/(serving_speech\.py|tts_adapters/__init__\.py)$
+      files: ^(vllm_omni/entrypoints/openai/(serving_speech\.py|tts_adapters/__init__\.py)|tools/pre_commit/check_tts_adapter\.py)$
+      # The budget-is-now-slack reminder is printed on a PASSING run, and
+      # pre-commit hides output from passing hooks unless the hook is verbose.
+      verbose: true
 
     # Keep `suggestion` last
     - id: suggestion
--- a/tools/pre_commit/check_tts_adapter.py
+++ b/tools/pre_commit/check_tts_adapter.py
@@ -123,18 +123,29 @@ def _legacy_detectors(tree: ast.AST) -> list[tuple[int, str]]:
     return found
 
 
-def _check(label: str, path: Path, actual: int, budget: int, sample: list[str], budget_name: str) -> list[str]:
+def _check(
+    label: str,
+    path: Path,
+    actual: int,
+    budget: int,
+    sample: list[str],
+    budget_name: str,
+    notices: list[str],
+) -> list[str]:
     if actual > budget:
         listing = "\n".join(f"      {line}" for line in sample)
         return [
             f"{path}: {actual} {label} exceeds the budget of {budget}.\n{listing}\n\n{_GUIDANCE}\n",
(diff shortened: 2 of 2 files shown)


linyueqian:
@vllm-omni-review-bot

hsliuustc0106 (approved):

fhfuih (commented):
> Four unrelated PRs merged onto a red main, and https://github.com/vllm-project/vllm-omni/pull/5843 — a diffusion scheduler refactor — had to bundle a change to my tool to get its own CI green. Sorry about that, @fhfuih.

No problem. Thanks for a follow-up fix

Though I wonder why there needs a pre-commit check for production code. It should be primarily on file format and existence, etc. 😂

PR #6010, opened 2026-08-10
[Perf][Diffusion] Compile and prewarm VAE decode

hsliuustc0106:
## Summary

- compile VAE `decode` callables discovered through a pipeline's declared VAE components when generic diffusion `torch.compile` is enabled
- permanently fall back to eager decode after either synchronous setup failure or the first lazy compile/runtime failure
- run a second, best-effort startup request through the ordinary request path so production-only branches are compiled before traffic
- add `--[no-]diffusion-compile-vae` and propagate it through YAML, CLI, orchestrator, and diffusion configuration

## Why

This follows a dummy warmup may take different pipeline branches from a normal request, leaving VAE compilation on the first user request.

Pipelines that expose their own `setup_compile()` retain compile ownership; the generic VAE path is not layered on top of custom compilation.

## Behavior and trade-offs

- VAE compilation is enabled when diffusion compilation is enabled and can be disabled with `--no-diffusion-compile-vae`.
- The existing dummy warmup remains required. The added normal-path warmup is best effort and never prevents serving.
- Warmup uses the existing bounded 512x512 startup shape and the model-specific dummy frame count. Other shape/frame buckets can still compile lazily on first use.
- Compilation moves latency into startup. The eager fallback latches permanently after a compile failure.

## Validation

- `279 passed, 1 skipped` across the affected diffusion/config/entrypoint tests
- `285 passed` in `tests/config/test_config_factory.py`
- Ruff check and format check pass for all 15 touched files
- DCO sign-off present

### L20X microbenchmark

Actual Diffusers tiny Helios VAE decode, 100 steady-state iterations:

- FP32 eager: 3.7454 ms
- FP32 compiled: 1.3055 ms
- speedup: 2.87x
- FP32 max absolute difference: 5.2e-8
- first compilation: 6.39 s

BF16 compilation also completed and was repeat-stable. Relative to eager BF16 on the random tiny model, max/mean absolute differences were 0.0452/0.000449, consistent with compile-time numerical reordering; generated-quality validation is left to model-specific CI/benchmarking.

### DIFF
--- a/tests/config/test_omni_config.py
+++ b/tests/config/test_omni_config.py
@@ -751,6 +751,7 @@ def test_from_pipeline_config_routes_regional_compile_dynamic(tmp_path):
                 "  - stage_id: 0",
                 "    diffusion_compile_granularity: regional",
                 "    diffusion_compile_dynamic: false",
+                "    diffusion_compile_vae: false",
             ]
         )
     )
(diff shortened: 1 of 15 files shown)


hsliuustc0106:
This PR touches **tests/diffusion/, vllm_omni/diffusion/, vllm_omni/config/, vllm_omni/engine/, tests/config/** (15 files). Based on CODEOWNERS coverage of the changed files, the most-related reviewers appear to be:

> @NickCao @yenuo26 @Isotr0py

PR #6011, opened 2026-08-10
[Perf][Diffusion] Add opt-in LTX2 channels-last VAE path

hsliuustc0106:
## Summary

- add an LTX-2 video VAE causal-padding path that allocates and fills temporal padding directly in `channels_last_3d`
- convert the LTX video VAE's Conv3d weights and install the layout-preserving forwards only when `--vae-use-channels-last-3d` is explicitly enabled
- preserve the original Diffusers repeat/concatenate path byte-for-byte when the option is disabled
- propagate the strict opt-in through CLI, orchestrator, deploy YAML, stage projection, and `OmniDiffusionConfig`

## Why

Repeat plus concatenate produces NCDHW storage and can erase the benefit of channels-last Conv3d weights. The new path performs edge replication and layout conversion in one allocation/copy.

local L20X measurements show that the benefit is shape-dependent and that enabling CL3D unconditionally would regress an important large decode.

## Behavior

Use:

```bash
vllm-omni serve Lightricks/LTX-2 --omni --vae-use-channels-last-3d
```

The option is also accepted as `vae_use_channels_last_3d: true` in stage deploy YAML. It currently applies only to the vLLM-Omni LTX video VAE subclass; audio and unrelated VAEs are unchanged.

## Correctness

- temporal edge replication is exact against the original repeat/concatenate construction
- causal, non-causal, temporal-only, and pointwise Conv3d cases match the original FP32 path
- contiguous weights retain the original padding branch
- an actual BF16 LTX-2 VAE decode had max/mean absolute differences of 0.03125/0.000943, consistent with the Conv3d algorithm change
- the patched causal forward also executes successfully under `torch.compile`

## L20X measurements

Actual `Lightricks/LTX-2` video VAE, BF16, warm steady-state decode:

| Latent -> output shape | NCDHW | layout-preserving CL3D | Result |
|---|---:|---:|---:|
| `[1,128,2,8,8]` -> `[1,3,9,256,256]` | 8.93 ms | 8.21 ms | 1.09x |
| `[1,128,16,16,24]` -> `[1,3,121,512,768]` | 225.81 ms | 388.71 ms | 0.58x |

For the 121-frame case, peak reserved memory fell from 8.58 GiB to 8.04 GiB despite the latency regression. These results are why this PR exposes capability without selecting it automatically.

## Validation

- `602 passed, 1 skipped` across all LTX, config-factory, config projection, entrypoint, and affected engine tests
- Ruff check and format check pass for all 12 touched files
- `git diff --check` passes
- DCO sign-off present

### DIFF
--- a/vllm_omni/config/omni_config.py
+++ b/vllm_omni/config/omni_config.py
@@ -573,6 +573,7 @@ class _DiffusionConfigProjection:
     fa_deterministic: bool = False
     vae_use_slicing: bool = False
     vae_use_tiling: bool = False
+    vae_use_channels_last_3d: bool = Field(default=False, strict=True)
     mask_strategy_file_path: str | None = None
     skip_time_steps: int = 15
     VSA_sparsity: float = 0.0
--- a/vllm_omni/config/omni_config.py
+++ b/vllm_omni/config/omni_config.py
@@ -573,7 +573,6 @@ class _DiffusionConfigProjection:
     fa_deterministic: bool = False
     vae_use_slicing: bool = False
     vae_use_tiling: bool = False
-    vae_use_channels_last_3d: bool = Field(default=False, strict=True)
     mask_strategy_file_path: str | None = None
     skip_time_steps: int = 15
     VSA_sparsity: float = 0.0
--- a/vllm_omni/config/stage_config.py
+++ b/vllm_omni/config/stage_config.py
@@ -401,7 +401,6 @@ class StageDeployConfig:
     step_execution: bool | None = None
     vae_use_slicing: bool | None = None
     vae_use_tiling: bool | None = None
-    vae_use_channels_last_3d: bool | None = None
     boundary_ratio: float | None = None
     flow_shift: float | None = None
     diffusion_kv_cache_dtype: str | None = None
--- a/vllm_omni/config/stage_config.py
+++ b/vllm_omni/config/stage_config.py
@@ -401,6 +401,7 @@ class StageDeployConfig:
     step_execution: bool | None = None
     vae_use_slicing: bool | None = None
     vae_use_tiling: bool | None = None
+    vae_use_channels_last_3d: bool | None = None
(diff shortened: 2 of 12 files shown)


hsliuustc0106:
This PR touches **vllm_omni/diffusion/, tests/diffusion/, vllm_omni/config/, vllm_omni/engine/, tests/config/** (12 files). Based on CODEOWNERS coverage of the changed files, the most-related reviewers appear to be:

> @NickCao @yenuo26 @alex-jw-brooks

hsliuustc0106:
@mglyn PTAL

vllm_omni/config/omni_config.py
```diff
@@ -573,6 +573,7 @@ class keeps the diffusion-only knobs from ``OmniDiffusionConfig`` without
     fa_deterministic: bool = False
     vae_use_slicing: bool = False
     vae_use_tiling: bool = False
+    vae_use_channels_last_3d: bool = Field(default=False, strict=True)
```
hsliuustc0106: can we make it as a LTX2 only change and do not add to the overall configuration?

vllm_omni/config/stage_config.py
```diff
@@ -401,6 +401,7 @@ class StageDeployConfig:
     step_execution: bool | None = None
     vae_use_slicing: bool | None = None
     vae_use_tiling: bool | None = None
+    vae_use_channels_last_3d: bool | None = None
```
hsliuustc0106: same here

hsliuustc0106 (commented):
make this only work for ltx pipeline and do not change to the general config interface

mglyn:
Which real deployment scenario is expected to benefit from this path? The representative 121-frame decode has 72% higher latency, while the memory saving is only about 0.54 GiB.

Considering the large-shape regression and numerical differences, it doesn’t seem necessary to introduce this yet unless there are more results showing practical benefits on other shapes or GPUs.

PR #2783, opened 2026-04-14
[Feature]: Adds distilled LoRA support for diffusion models

Songrui625:
PLEASE FILL IN THE PR DESCRIPTION HERE ENSURING ALL CHECKLIST ITEMS (AT THE BOTTOM) HAVE BEEN CONSIDERED.

## Purpose
https://github.com/vllm-project/vllm-omni/issues/2782

## Test Plan
- Unitest 
```python
pytest -v tests/diffusion/lora/test_loader.py
```

- E2E test for Qwen-Image [lightx2v/Qwen-Image-2512-Lightning](https://huggingface.co/lightx2v/Qwen-Image-2512-Lightning/tree/main)
```python
# w/o lora
python3 text_to_image.py --model /data00/models/Qwen-Image-2512 \
    --prompt "A Chinese female college student, around 20 years old, with a very short haircut that conveys a gentle, artistic vibe. Her hair naturally falls to partially cover her cheeks, projecting a tomboyish yet charming demeanor. She has cool-toned fair skin and delicate features, with a slightly shy yet subtly confident expression—her mouth crooked in a playful, youthful smirk. She wears an off-shoulder top, revealing one shoulder, with a well-proportioned figure. The image is framed as a close-up selfie: she dominates the foreground, while the background clearly shows her dormitory—a neatly made bed with white linens on the top bunk, a tidy study desk with organized stationery, and wooden cabinets and drawers. The photo is captured on a smartphone under soft, even ambient lighting, with natural tones, high clarity, and a bright, lively atmosphere full of youthful, everyday energy." \
    --negative-prompt " " \
    --cfg-scale 4.0 \
    --num-inference-steps 4 \
    --seed 42 \

# 4 steps lora
python3 text_to_image.py --model /data00/models/Qwen-Image-2512 \
[...]

PR #758, opened 2026-01-13
[Feature] Diffusion LoRA Adapter Support (PEFT compatible) for vLLM alignment

AndyZhou952:
This is a joint work by  @AndyZhou952 and @dongbo910220.

Design doc [here](https://docs.google.com/document/d/1XdueGleiv9v8_dOVB4apYiTcmscN61035ggui1zyHa8).

## Purpose:

Following Issue #281  and PR #657, this PR adds diffusion LoRA Adapter Support (PEFT compatible) for vLLM alignment. 

This PR reuses the LoRA RPC logic from #657 (thanks to @dongbo910220's implementation), while utilizing the vLLM self-defined layer for LoRA support and PEFT format to be incorporated in the reinforcement training pipeline.

### How vLLM adds LoRA support:
Three steps: (1) initialization; (2) per-request; (3) inference (via vLLM self-defined LoRA layers, calculations in `forward`).

image

image

 ### vLLM-Omni PEFT LoRA integration Design:

image

Besides `add_lora` and `remove_lora`, we also support `pin_lora` and `list_lora` as public APIs to be consistent with the vLLM base behavior.

### Design principles:
- Preserve vLLM variable and function naming convention; try to keep the functionalities consistent with vLLM
- Reuse helper functions from vLLM whenever possible, add minimum new code for diffusion-specific wrappers
- Extensibility: easy to add support for more layer adapters and for multiple LoRA loading

### Design choices:
- Support PEFT LoRA format and reuse vLLM LoRA layers (to be consistent with/ vLLM behavior). We reuse `PEFTHelper` from vLLM, which will look for the file `adapter_config.json` when loading LoRAs.
- `DiffusionLoRAManager` does not inherit from `LoRAManager`, since (1) `LoRAManager` is LLM-centric with redundant variables in `__init__` for diffusion models; (2) diffusion component-based nature requires separate treatment;
- Incorporate LRU cache management within `DiffusionLoRAManager` to keep things compact, also no need for a separate `WorkerLoRAManager` as vllm-omni has `gpu_worker` that does the job. Also, LRU cache management is kept within `DiffusionLoRAManager`.
[...]

PR #6017, opened 2026-08-11
[Diffusion] Generalize LoRA loading and composition with MiniMax-H3 Turbo support

mglyn:
## Scenario and motivation

Diffusion deployments now combine acceleration, style, and identity LoRAs published in model-specific formats. They need weighted multi-adapter composition, startup fusion, and request-level switching without moving model checkpoint knowledge into the common backend. MiniMax-H3 Turbo makes this an immediate requirement.

This PR adds the following common surface:

| Scope | Field or interface | Meaning |
|---|---|---|
| Startup CLI/config | `--prefused-lora PATH[=SCALE]` / `prefused_lora` | Repeatable adapters materialized into the base weights during startup. |
| Startup CLI/config | `--dynamic-lora PATH[=SCALE]` / `dynamic_lora` | Repeatable adapters installed for runtime activation and reweighting. |
| Startup CLI/config | `--max-cpu-loras` | Sizes the existing CPU adapter cache used by request-time loading. |
| Image/video request | `lora: object \| object[] \| null` | Selects one or more adapters and their scales for a request. An empty list disables dynamic adapters. |
| Model extension | `get_lora_load_plan(...)` | Describes checkpoint mapping and conversion into low-rank tensors plus typed auxiliary updates. |
| Model extension | `get_lora_apply_plan()` | Describes target components, target modules, and packed-module mappings. |

This surface controls adapter loading and composition only; sampling remains under deployment/request control.

## MiniMax-H3 implementation and validation

Formal H3 support is limited to the Diffusers-format FL2VA adapters in `lightx2v/Minimax-h3-Turbo`. Its model-side loader maps the published PEFT-like keys to native H3 modules, restores the native `[gate; up]` FFN order, and reads alpha from checkpoint metadata (`128 / 128 = 1` for v1.0; legacy v0.1 without metadata falls back to its official alpha 8). It targets only the FL2VA transformer; Ref2VA rejects this adapter.

Example deployment with dynamic Turbo and four Transformer evaluations by default:

```bash
TURBO_LORA=$(hf download lightx2v/Minimax-h3-Turbo \
  minimax_h3_fl2v_turbo_4step_v1.0_768p_bf16.safetensors)

vllm serve MiniMaxAI/MiniMax-H3 \
  --omni \
  --trust-remote-code \
  --task-type fl2va \
  --dynamic-lora "$TURBO_LORA=1.0" \
  --default-sampling-params \
    '{"0":{"num_inference_steps":5,"extra_args":{"flow_shift":6.0,"audio_flow_shift":3.0}}}'
```

Requests can disable dynamic adapters with `lora=[]` or select and reweight Turbo explicitly:

```bash
-F 'flow_shift=6' \
-F 'extra_params={"audio_flow_shift":3.0}' \
-F "lora={\"name\":\"turbo\",\"path\":\"$TURBO_LORA\",\"scale\":1.0}"
```

LoRA selection never changes sampling parameters. For H3, five sigma grid points produce the four Transformer evaluations used by Turbo.

Benchmark: corrected Turbo v1.0 checkpoint with its published `alpha=128`, native H3 FFN layout conversion, and video/audio flow shifts 6/3; 4× H200, USP4/Ring1, VAE TP4, text-encoder TP1, regional `torch.compile`, and default `FLASH_ATTN`. All cases use the same 768×1344 FL2VA input, 107 frames/24 FPS, prompt, and seed. Times are median `stage_0_gen_ms` from five runs after two full-shape warmups.

| Case | LoRA execution | NFE | Stage 0 p50 | Five-run range |
|---|---|---:|---:|---:|
| Base reference | None | 49 | 69.466 s | 69.407–69.590 s |
| No-LoRA control | None | 4 | 9.178 s | 9.097–9.325 s |
| Turbo | Prefused | 4 | 8.954 s | 8.941–8.973 s |
| Turbo | Dynamic | 4 | 9.505 s | 9.481–9.543 s |
| Turbo + test-only Anime | Prefused Turbo + dynamic Anime | 4 | 9.599 s | 9.525–9.622 s |
| Turbo + test-only Anime | Dynamic Turbo + dynamic Anime | 4 | 9.736 s | 9.569–9.832 s |

At four NFE, dynamic Turbo is **7.31× faster** than the 49-NFE reference. Dynamic execution costs **6.15%** versus prefusion for Turbo alone; full dynamic costs **1.42%** versus prefused Turbo plus dynamic Anime in the two-adapter run. Prefusion is faster in this run, while dynamic execution retains request-level selection and reweighting.

Visual comparisons use the corrected Turbo v1.0 checkpoint, identical input/prompt/seed, and video/audio flow shifts 6/3. Audio comes from the rightmost case.

**Turbo lifecycle:** no LoRA at 4 NFE / no LoRA at 49 NFE / prefused Turbo at 4 NFE / dynamic Turbo at 4 NFE

https://github.com/user-attachments/assets/26a3e8e2-1ed5-45fd-b060-f3ebe6721240

**Dynamic composition:** no LoRA at 49 NFE / dynamic Anime at 49 NFE / prefused Turbo + dynamic Anime at 4 NFE / dynamic Turbo + dynamic Anime at 4 NFE

https://github.com/user-attachments/assets/fb5b3d3a-ba56-47a0-8335-6baeb23d3d7a

### Execution modes and numerical behavior

For $\Delta W_i = B_i A_i$ and optional additive bias $d_i$, prefused adapters $P$ and dynamic adapters $D$ share one additive definition:

$$
y = \left(W + \sum_{i \in P} s_i B_i A_i\right)x + \left(b + \sum_{i \in P} s_i d_i\right) + \sum_{j \in D} s_j\left(B_j A_j x + d_j\right)
$$

This supports prefused and dynamic adapters together. Repeating an adapter in both sets intentionally applies it twice. The modes are equal in exact arithmetic, but BF16 operation and rounding order differ; iterative denoising can amplify that difference, so they use separate regression floors.

Prefusion is rejected for quantized weights because a dense delta cannot be added directly to an arbitrary serialized/runtime quantized representation. Dynamic LoRA leaves the quantized base unchanged. First-request loading and rank expansion are supported in eager mode; compiled or offloaded deployments must preload compatible adapters before the graph is fixed.

### Model-described plans and ownership

Checkpoint naming, packed projections, and alpha normalization are model-specific, so models supply two declarative plans:

| Plan | Model describes | Common backend performs |
|---|---|---|
| Load plan | PEFT configuration, key mapping, and conversion into `ConvertedLoRAState` | Downloading, deserialization, validation, caching, and canonicalization |
| Apply plan | Adapted components, target modules, and packed-module layout | Module installation, activation, scale binding, fusion, and lifecycle management |

`ConvertedLoRAState` carries A/B plus typed auxiliary updates. This PR implements `AdditiveBiasUpdate` and rejects unknown types. Models own conversion; the backend owns download, validation, caching, composition, activation, and fusion. Sampling and recipes remain model/request-owned.

### Multiple weighted adapters through one core algorithm

Both lifecycles consume the same ordered canonical composition, `[(adapter_i, scale_i)]`. Multiple updates form one wider low-rank update; semicolons concatenate vertically and commas horizontally along rank:

$$
A_* = [A_1; A_2; \ldots; A_n]
$$

$$
B_* = [s_1B_1, s_2B_2, \ldots, s_nB_n]
$$

Therefore:

$$
B_* A_* x = \sum_i s_i B_i A_i x
$$

The backend applies this construction independently to packed-projection slices such as H3 QKV and MLP modules. Ordering is deterministic, and repeated references are combined by scale before application.

<details>
<summary><strong>Rebase and #2783 reproduction</strong></summary>

After [#2783](https://github.com/vllm-project/vllm-omni/pull/2783) landed, this branch was rebased onto `origin/main` at `9b0df390`. Qwen-Image and Wan checkpoint handling now lives in model-owned plans; the shared backend owns loading, validation, caching, composition, dynamic execution, and prefusion.

Latest hardening moves active dynamic A/B and bias sidecars to the input device during forward, preserving offload correctness at the transfer cost. Qwen alpha is folded once; non-zero unsupported Wan `.diff` is rejected; typed updates must bind uniquely and match TP/packed-aware output shapes. It also removes write-only manager state and unifies scheduler LoRA batch-key canonicalization. Existing CPU/fake-device suite: **124 passed**; corrected H3 loader targeted tests: **10 passed**; pre-commit: passed.

The complete #2783 visual comparison was reproduced from its published checkpoints and settings with eager execution and startup prefusion.

**Qwen-Image-2512, 1024×1024, seed 0**

| Base model with 4 steps, CFG 4.0 | Base model with 50 steps, CFG 4.0 | Base model + distilled LoRA with 4 steps, CFG 1.0 |
|---|---|---|
| qwen_base_4step_cfg4_seed0|qwen_base_50step_cfg4_seed0|qwen_lora_4step_cfg1_seed0|
| 1.65 s | 22.15 s | 0.87 s |

**Wan2.1/2.2 Text-To-Video, 832×480, 121 frames, 24 FPS, seed 42**

Wan2.2 base model, 4 steps, CFG 4.0 — 85.91 s:

https://github.com/user-attachments/assets/ee521bf6-ad21-41da-bf1b-fc30a46397a2

Wan2.2 base model + high/low-noise distilled LoRAs, 4 steps, CFG 1.0 — 25.26 s:

https://github.com/user-attachments/assets/347c5072-66ac-4ba2-bbc9-4652c95122d6

Wan2.1 base model + distilled LoRA, 4 steps, CFG 1.0 — 24.08 s:

https://github.com/user-attachments/assets/019d12a4-555e-45fc-9428-d11f4c5371fc

All videos are H.264, 832×480, 121-frame, 24 FPS artifacts. The displayed Wan2.1 file is the original pre-typed #2783 prefusion repro; `wan21_t2v_distill_lora_4step.mp4` is the new typed-backend validation. Typed bias updates compose or prefuse at the A/B scale; two reserved-GPU typed runs produced identical files and decoded-frame SHA256 values.

</details>

[#5700](https://github.com/vllm-project/vllm-omni/issues/5700)


david6666666:
@dongbo910220 @SamitHuang  ptal thx

hsliuustc0106:
review the agent code by yourself first with explicit comments

docs/user_guide/diffusion/lora.md
```diff
@@ -1,157 +1,207 @@
-# LoRA (Low-Rank Adaptation) Guide
+# Diffusion LoRA
```
hsliuustc0106, line 1: how about the feature design doc update and related skills if needed

vllm_omni/diffusion/lora/types.py
```diff
@@ -0,0 +1,169 @@
+# SPDX-License-Identifier: Apache-2.0
```
hsliuustc0106, line 1: why we put it here? does it work for diffusion lora only?

docs/user_guide/diffusion/lora.md
```diff
@@ -1,157 +1,207 @@
...
+  --omni \
+  --dynamic-lora /path/to/adapter=1.0 \
+  --max-cpu-loras 1
+```
 
-lora_path = "/path/to/lora_adapter"
+Repeat `--dynamic-lora` to install a weighted default composition. The cache
+must have room for every adapter that must remain resident:
 
-omni = Omni(
```
hsliuustc0106, line 42: is there any offline lora usage?

hsliuustc0106 (commented):
is there any perf data?

mglyn:
I manually reviewed the shared backend, model-owned plans, request/batching propagation, TP and packed-module binding, and compile/offload/quantization boundaries. This led to the offload, typed-update validation, Wan dense-delta rejection, Qwen alpha, and batching-key fixes now included in the PR.

> review the agent code by yourself first with explicit comments

docs/user_guide/diffusion/lora.md
mglyn: Added docs/design/feature/diffusion_lora.md covering backend/model ownership, load/apply plans, dynamic and prefused execution, runtime boundaries, three model examples, and performance data. The add-diffusion-model skill now points model integrations to these extension contracts.

vllm_omni/diffusion/lora/types.py
mglyn: Good point. These composition, parsing, and batching types are currently diffusion-specific, so I moved them to vllm_omni/diffusion/lora/types.py and updated all callers.

docs/user_guide/diffusion/lora.md
mglyn: Yes. The offline Python API and generic T2I/T2V/I2V scripts support startup dynamic_lora / prefused_lora and request-level lora_request / lora_scale. The user guide now includes a dedicated offline example.

mglyn:
> is there any perf data?

H3 performance data was already included in the PR description. I have now also added a concise performance table and interpretation to the design document.

hufangjian2017:
Thanks for the Turbo LoRA support. While validating FL2VA + lightx2v PEFT Turbo
(`minimax_h3_fl2v_turbo_4step_v1.0_768p_bf16.safetensors`) on NVIDIA and Iluvatar,
we hit two correctness issues in `vllm_omni/diffusion/models/minimax_h3/lora.py`:

### 1. `lora_alpha=8` under-scales Turbo

PR hardcodes `lora_alpha=8` for the rank-128 checkpoint → effective scale
`alpha/rank = 8/128 = 1/16`.

LightX2V training + published Turbo inference configs use **alpha=128** (scale=1.0), e.g.:
- `LightX2V/configs/minimax_h3/minimax_h3_t2av_dmd_lora_4step.json` → `"alpha": 128`
- `lightx2v_train/.../minimax_h3_t2av_dmd_lora.yaml` → `alpha: 128`
- ComfyUI twin checkpoints embed per-layer alpha tensors of 128

With `alpha=8` (and no FFN swap below), output is only “somewhat structured”
noise/blur. With `alpha=128` quality matches the Turbo recipe much more closely.

(The converter note that says `--lora_alpha 8` for merge looks like a different
path and should not drive the PEFT dynamic-load scale.)

### 2. Missing Diffusers → native FFN layout swap on `lora_B`

Key remapping alone is not enough for `ff.net.0.proj` → `mlp.fc1`:

| Checkpoint | fused fc1 layout | forward |
|---|---|---|
| Omni / native FL2VA | `[gate; up]` | `silu(gate) * up` |
| Diffusers / lightx2v PEFT | `[value; gate]` | `value * silu(gate)` |

Hard check: `FL2VA fc1 == swap_halves(Diffusers ff.net.0.proj)`.

So before `WeightsMapper`, PEFT keys matching `.ff.net.0.proj.lora_B.` need:

```python
value, gate = tensor.chunk(2, dim=0)
tensor = torch.cat((gate, value), dim=0)

**Comment by mglyn** [comment]
Thanks for catching and validating these two correctness issues. I have confirmed both: the published LightX2V Turbo configuration uses `alpha=128`, and the Diffusers FFN `lora_B` requires a half swap when mapping back to the native `[gate; up]` layout. I will fix the implementation and update the validation/demo results. @hufangjian2017

File: vllm_omni/entrypoints/openai/utils.py [hunk 4]
```diff
@@ -55,6 +53,27 @@ def parse_lora_request(lora_body: Any) -> tuple[LoRARequest | None, float | None
...
+    if not bodies:
+        return (), ()
+    parsed = tuple(_parse_single_lora_request(body) for body in bodies)
+    if len(parsed) == 1:
+        return parsed[0]
+    composition = normalize_lora_composition(
+        tuple(request for request, _ in parsed),
+        tuple(1.0 if scale is None else scale for _, scale in parsed),
+    )
+    return split_lora_composition(composition)
```
hsliuustc0106, line 90: [P3] Multi-adapter scale cancellation restores the deployment default instead of disabling dynamic adapters.

A request like lora=[{path: p, scale: 1.0}, {path: p, scale: -1.0}] normalizes to an empty composition, and split_lora_composition(()) returns (None, 1.0), so the worker treats the request as if lora were omitted and activates the deployment default. That contradicts the explicit lora=[] case (which returns ((), ())) and the single-adapter zero-scale case (which stays non-None). Returning ((), ()) when a non-empty list collapses to an empty composition would keep the explicit-vs-omitted distinction.

vllm_omni/engine/async_omni_engine.py
```diff
@@ -1205,16 +1205,19 @@ def _resolve_stage_configs(
                 if kwargs.get("lora_path") is not None:
```
hsliuustc0106: [P3] Leftovers from removing the legacy lora_path/lora_backend surface.

These engine_args.lora_path / lora_scale writes are dead (OrchestratorArgs no longer declares the fields and nothing reads them back; the od_config kwargs path is the live one), a list lora_path — the old worker unwrapped single-element lists — now fails late in stable_lora_int_id with 'list' object has no attribute 'encode', and stage-level lora_path/lora_backend/lora_scale in deploy YAML silently fall into engine_extras now that the StageDeployConfig fields are gone. A deprecation warning or type check for the removed names would make the migration loud.

docs/design/feature/diffusion_lora.md
```diff
@@ -0,0 +1,91 @@
...
+| Turbo dynamic | 4 | 9.520 s |
+
+Dynamic Turbo is 7.42× faster than the 49-NFE reference because the Turbo
+recipe reduces Transformer evaluations. At the same four NFE, dynamic execution
+cost 2.97% relative to prefusion and remained close to the no-LoRA control.
+The result supports dynamic LoRA as the serving default; it is not a universal
+performance guarantee for other models or adapter ranks.
+
+## Runtime boundaries
+
```
hsliuustc0106: [P3] Design question on runtime boundaries: freeze() deliberately permits cache_backend together with dynamic adapters, but the diffusion-KV prefix block-hash contract (diffusion_kv/request.py) has no adapter-composition term — once prefix publication lands, requests sharing a prompt but with different compositions would reuse prefix K/V computed under the wrong adapter. Worth a sentence here so the KV data-plane work folds the composition into the cache identity.

hsliuustc0106 (commented):
No blockers. The shared composition backend looks sound: the rank-concatenated composition ordering, TP rank-local slicing and prefusion, rank-0-only row-parallel bias, lora=[] vs omitted batch-key separation, and freeze-before-compile/offload ordering all check out, and the H200 A/B is properly controlled (NFE-matched controls, five-run medians, dynamic-under-compile measured).

Three minor threads posted: the multi-adapter scale-cancellation edge in parse_lora_request, leftovers from the legacy lora_path/lora_backend removal (dead engine_args writes, list lora_path crash, silently dropped stage-YAML keys), and a design note that the diffusion-KV prefix block hash should eventually carry the adapter composition. Please also fix the pre-commit F811 (duplicate LoRAConfig import in diffusion/lora/manager.py) so the lane goes green.

vllm_omni/diffusion/lora/manager.py
```diff
@@ -17,12 +15,24 @@
 from vllm.lora.peft_helper import PEFTHelper
 from vllm.lora.request import LoRARequest
 from vllm.lora.utils import (
-    get_adapter_absolute_path,
     get_supported_lora_modules,
     replace_submodule,
 )
 from vllm.model_executor.layers.linear import MergedColumnParallelLinear, QKVParallelLinear
-
+from vllm_omni.config.lora import LoRAConfig
```
hsliuustc0106: [P1] Please fix the duplicate `LoRAConfig` import. Ruff reports I001/F811 here and the required pre-commit check is red; keep a single import so the gate can pass.

vllm_omni/diffusion/worker/diffusion_model_runner.py
```diff
@@ -301,6 +332,8 @@ def get_memory_context() -> AbstractContextManager[Any]:
                 f"{self.od_config.model_class_name} does not support that contract."
             )
 
+        self.init_lora_manager()
```
hsliuustc0106: [P1] This ordering breaks DLO with startup LoRA. `take_host_weight_plan()` is produced from the unwrapped pipeline, but `init_lora_manager()` replaces targets with `*.base_layer` before DLO consumes it; DLO then cannot find the planned `...weight` tensors. Please make the host-weight plan and LoRA wrapping use the same runtime names, and add a DLO+LoRA startup test.

vllm_omni/diffusion/models/wan2_2/lora.py
```diff
@@ -0,0 +1,149 @@
...
+            key = f"{component_name}.{key}"
+        if key.endswith(".lora_B.bias"):
+            auxiliary_updates.append(
+                AdditiveBiasUpdate(
+                    module_name=key.removesuffix(".lora_B.bias"),
+                    tensor=value,
+                )
+            )
+        else:
+            converted[key] = value
```
hsliuustc0106, line 112: [P1] Diffusers Wan checkpoints can contain per-module `.alpha` tensors, but this branch copies every non-bias key into `lora_tensors`. `LoRAModel.from_lora_tensors()` then rejects `...alpha` as an unsupported LoRA weight. Please fold or remove alpha before passing the state to the shared loader, as the Qwen path does, and add a regression test.

vllm_omni/diffusion/worker/diffusion_model_runner.py
```diff
@@ -220,6 +223,34 @@ def _compile_transformer(self, attr_name: str) -> None:
...
+        )
+        graph_is_fixed = (
+            not self.od_config.enforce_eager
+            or self.od_config.enable_cpu_offload
+            or self.od_config.enable_layerwise_offload
+            or getattr(self.od_config, "enable_distributed_layerwise_offload", False)
+            or getattr(self.od_config, "cache_backend", None) not in (None, "none")
+        )
+        if graph_is_fixed:
+            self.lora_manager.freeze()
```
hsliuustc0106: [P1] This makes first-request dynamic loading fail on the default compiled path: with no startup adapter, `_max_lora_rank` is 0, so the first request hits the frozen-capacity error. The existing per-request LoRA e2e test does not preload adapters or enable eager mode. Please either support first-request loading here or make the preload/eager requirement explicit and enforce it before the request reaches the worker.

vllm_omni/entrypoints/openai/protocol/images.py
```diff
@@ -152,11 +152,11 @@ def validate_use_system_prompt(cls, v):
 
     # vllm-omni extension for per-request LoRA.
     # This mirrors the `extra_body.lora` convention in /v1/chat/completions.
-    lora: dict[str, Any] | None = Field(
+    lora: dict[str, Any] | list[dict[str, Any]] | None = Field(
```
hsliuustc0106, line 155: [P1] This widens the Images API contract to arrays, but the multi-stage image path forwards `lora` into chat code that still only parses a truthy dictionary. Lists are silently ignored, and `lora=[]` cannot clear a deployment default. Please wire the list and empty forms through the chat path and add an endpoint test.

vllm_omni/diffusion/lora/manager.py
```diff
@@ -215,117 +291,60 @@ def _get_packed_sublayer_suffixes(self, packed_module_suffix: str, n_slices: int
...
+            else normalize_lora_composition(lora_request, lora_scale)
         )
-        if adapter_id not in self._registered_adapters:
-            logger.info("Loading new adapter: id=%d, name=%s", adapter_id, lora_request.lora_name)
-            # Add the adapter + add to the cache
-            self.add_adapter(lora_request)
-        else:
-            # Just touch the cache access order
-            self._touch_adapter_info(adapter_id)
+        if lora_composition_key(composition) == lora_composition_key(self._active_composition):
```
hsliuustc0106, line 329: [P2] The fast-path key contains only `(adapter_id, scale)`, so reusing an explicit ID for another path returns here and keeps the old cached adapter. For example, ID 7 `/a` followed by ID 7 `/b` silently uses `/a`, bypassing the path check in `add_adapter()`. Please validate the request path before this fast path or include it in the cache identity.

vllm_omni/entrypoints/openai/utils.py
mglyn: Fixed in 9a9422a9: a non-empty request that normalizes to an empty composition now remains the explicit empty-composition disable sentinel, with a regression test.

vllm_omni/engine/async_omni_engine.py
mglyn: Fixed in 9a9422a9: removed the dead engine writes, reject the removed CLI/deploy-YAML fields with migration guidance, and reject legacy list lora_path early.

docs/design/feature/diffusion_lora.md
mglyn: Documented in 9a9422a9: future publishable diffusion prefix-KV identities must include the canonical LoRA composition; requests without canonical block hashes remain request-local.

vllm_omni/diffusion/lora/manager.py
mglyn: Resolved by the latest main rebase: manager.py now has a single LoRAConfig import and pre-commit is green.

vllm_omni/diffusion/models/wan2_2/lora.py
mglyn: Fixed in 73f3bd12: Diffusers per-module alpha is folded into LoRA B before shared loading. Wan and Qwen now reuse the same helper, with a Wan regression test.

vllm_omni/entrypoints/openai/protocol/images.py
mglyn: Fixed in 73f3bd12: all multistage image paths now propagate weighted lists and explicit empty LoRA lists; endpoint tests cover both forms.

vllm_omni/diffusion/lora/manager.py
mglyn: Fixed in 73f3bd12: adapter ID-to-path identity is validated before the active-composition fast path, with a regression test.

vllm_omni/diffusion/worker/diffusion_model_runner.py
mglyn: Fixed in 64ca3de6: common engine admission now rejects a non-empty request LoRA before scheduler/worker execution when compile, offload, or cache fixed the graph and no startup adapter reserved capacity. Eager mode, startup-preloaded capacity, and explicit empty compositions remain allowed; runner and engine share the same graph-fixed predicate. Docs and regression tests were added.

vllm_omni/diffusion/worker/diffusion_model_runner.py
mglyn: Fixed in e9544118 with a deliberately smaller compatibility boundary. DLO now supports startup dynamic LoRA through the ordinary loader path: base weights and post-load validation complete before wrappers are installed, LoRA A/B runtime slots stay device-resident, and DLO streams only the runtime base weights. This avoids creating a stale mmap plan. Prefused LoRA with DLO is explicitly rejected for now. Added config, loader, placement tests, and docs.

princepride (commented):
Sorry, I am confused. Why we must refactor the Lora loading if we only want support H3 Turbo? Can we try to split to more sub-pr?

docs/user_guide/diffusion/lora.md
```diff
@@ -1,157 +1,281 @@
...
+optimization. Prefused LoRA with DLO is rejected; use `--dynamic-lora`.
+
+## Sampling remains model-owned
+
+Loading an acceleration or distilled LoRA does not change timesteps, guidance,
+the scheduler, or the number of denoising steps. Set those independently
+through deployment defaults or request sampling parameters according to the
+adapter's published usage instructions.
+
+## Model integration and extension points
```
hsliuustc0106, line 181: [P2] The design page added earlier in this PR (docs/design/feature/diffusion_lora.md) was removed again in the lifecycle rework and folded into this guide.

The >500-production-line policy calls for a design RFC/issue for a change of this size, and get_lora_load_plan/get_lora_apply_plan are a cross-model public extension surface; #5700 is a roadmap tracker rather than a design RFC. Could you restore the page (the earlier version is in the branch history) or link the accepted design issue? Note the prefix-KV identity sentence cited in the earlier review reply lived there and did not survive the fold.

docs/user_guide/diffusion/lora.md
```diff
@@ -1,157 +1,281 @@
...
+accept a scale; clients set composition scales per request. Every selectable
+adapter must be listed before the server starts.
 
-outputs = omni.generate(
-    prompt="A piece of cheesecake",
-    lora_request=lora_request,
-    lora_scale=2.0, # optional arg, default 1.0
-)
-```
+### Request selection
```
hsliuustc0106, line 86: [P3] The PR body's request example still passes a path (-F lora={name: turbo, path: $TURBO_LORA, scale: 1.0}), but _parse_single_lora_request now rejects any path field with a client error. Worth updating the body example to name-only selection so the documented command actually works against the final design.

vllm_omni/diffusion/models/wan2_2/lora.py
```diff
@@ -0,0 +1,155 @@
...
+    nonzero_dense_deltas = [key for key, value in unsupported_dense_deltas.items() if torch.count_nonzero(value).item()]
+    if nonzero_dense_deltas:
+        raise ValueError(f"This Wan adapter contains unsupported non-zero dense deltas: {nonzero_dense_deltas[:3]}")
+    if unsupported_dense_deltas:
+        state_dict = {key: value for key, value in state_dict.items() if key not in unsupported_dense_deltas}
+
+    if any(key.startswith("diffusion_model.") for key in state_dict):
+        state_dict = _convert_non_diffusers_wan_lora_to_diffusers(dict(state_dict))
+
+    unsupported = [key for key in state_dict if key.endswith((".diff", ".diff_b"))]
```
hsliuustc0106, line 90: [P3] A checkpoint with lora_unet_/unprefixed lora_down naming plus .alpha tensors is claimed by the plan gate but never converted: the alpha fold requires .lora_A/.lora_B.weight suffixes and the non-diffusers converter requires a diffusion_model. prefix, so .alpha keys survive to this final check (it only matches .diff/.diff_b) and surface later as an opaque unsupported-weight error from from_lora_tensors. Adding .alpha to this rejection, or routing lora_unet_ keys like Qwen does, would make the failure self-explanatory.

vllm_omni/diffusion/worker/diffusion_worker.py
```diff
@@ -443,34 +444,8 @@ def set_kv_cache_configs(self, kv_cache_configs: list[KVCacheConfig]) -> None:
...
-            if hasattr(pipeline, "load_lora_weights"):
-                if self.od_config.lora_scale > 1.0:
-                    logger.warning("lora_scale > 1.0 may not take any effect when using distilled LoRA backend.")
-                pipeline.load_lora_weights(lora_path)
-            else:
-                logger.warning("Pipeline does not support loading distilled LoRA weights for now.")
-        else:
-            raise ValueError(f"Unknown LoRA backend: {lora_backend}. Available choices: {LoRABackend.__members__}")
+        """Alias the model runner's pre-offload/pre-compile LoRA manager."""
+        self.lora_manager = self.model_runner.lora_manager
```
hsliuustc0106, line 448: [P2] Lifecycle gap: re_init_pipeline (~line 846) can no longer free the old pipeline. The manager keeps a strong self.pipeline reference (plus wrapped modules and A/B buffers), so its del model_runner.pipeline + gc.collect() leaves the old checkpoint resident while the replacement loads — a transient ~2x VRAM peak, and with sleep mode enabled the next weights-pool entry trips the one-instance assert in _maybe_get_memory_pool_context. Dropping model_runner.lora_manager before that gc.collect() (load_model rebuilds it) would restore the pre-LoRA teardown behavior.

vllm_omni/diffusion/lora/manager.py
hsliuustc0106: Verified at d86d8583 — single import restored and pre-commit green.

vllm_omni/diffusion/worker/diffusion_model_runner.py
hsliuustc0106: Verified at d86d8583 — the mmap plan is suppressed whenever dynamic_lora is set, wrappers install only after ordinary loading plus the unloaded-weights check, A/B buffers stay device-resident outside DLO's parameter sweep, and prefused+DLO is rejected at config level with tests.

vllm_omni/diffusion/models/wan2_2/lora.py
hsliuustc0106: Verified — fold_diffusers_lora_alpha is shared with Qwen and the Wan regression test asserts the exact key set. One residual edge posted inline: lora_unet_-style .alpha keys skip both the fold and the converter and fail later with an opaque error.

vllm_omni/diffusion/worker/diffusion_model_runner.py
hsliuustc0106: Verified — superseded by the stronger registry model: all loading is startup-only, the engine rejects unregistered request adapters before scheduler admission with a per-request OmniClientError, and the manager drops its loader after construction.

vllm_omni/entrypoints/openai/protocol/images.py
hsliuustc0106: Verified — all three chat sites parse lists and the empty form and raise OmniClientError instead of swallowing; image and video endpoint tests cover both forms.

vllm_omni/diffusion/lora/manager.py
hsliuustc0106: Verified — _validate_adapter_identity runs before the fast path, and name-only request selection removes the path-spoofing surface entirely.

vllm_omni/entrypoints/openai/utils.py
hsliuustc0106: Verified at d86d8583 — the empty-composition sentinel now preserves the explicit disable. Thanks.

vllm_omni/engine/async_omni_engine.py
hsliuustc0106: Verified — the removed fields now raise with migration guidance in stage YAML, engine_extras, platform overrides, and engine kwargs, and the dead writes are gone.

docs/design/feature/diffusion_lora.md
hsliuustc0106: The sentence cited here was removed again along with the design page in the later lifecycle rework — see my new comment on the user guide about restoring it (or restating that boundary where the contract now lives).

mglyn:
@david6666666 @hsliuustc0106 @hufangjian2017 @princepride

You are right that the current PR has become too broad, and I will split it.

The original goal was to introduce a shared diffusion LoRA backend that could support model-specific checkpoint loading, model-specific application semantics, dynamic and prefused execution, and weighted multi-LoRA composition, with MiniMax-H3 Turbo as the first integration. After #2783 was merged, however, this PR also started migrating the existing #758 PEFT backend and the new #2783 distilled-LoRA path. As a result, #6017 now mixes three concerns and contains unnecessary breaking changes for an H3-focused PR.

The existing paths currently overlap but provide different and incomplete capabilities:

- #758 is a PEFT-oriented dynamic backend with LRU management and runtime `add/remove/pin/list` APIs. A request carries adapter path/ID, may trigger loading, supports only one active LoRA per batch, and has no model-owned loading or application contract.
- #2783 adds a separate startup-time `distill` path for Qwen-Image and Wan. It directly merges model-specific updates, but does not provide request-level selection, weighted composition, or a reusable model extension interface.
- #6017 attempted to unify both paths while adding H3, which made the change destructive and obscured the minimum dependency required by H3.

H3 still demonstrates why a new extension point is needed: its published Turbo adapter requires model-specific key conversion, exact alpha handling, and a Diffusers-to-native FFN layout transformation. These semantics cannot be described safely by the existing generic PEFT loader or by hard-coding another model branch into the #2783 loader.

I will therefore rework this into a parallel, minimal implementation containing only the shared contract required by H3 Turbo, without replacing #758 or migrating the #2783 models in the same PR. Once that path is reviewed and stable, the legacy PEFT and distilled-LoRA implementations can be migrated separately, with compatibility and removal handled explicitly.

vllm_omni/diffusion/worker/diffusion_worker.py
SamitHuang, line 448: Given the discussion to split this PR and focus on a minimal shared contract for H3 Turbo, we should ensure the new `lora_manager` lifecycle doesn't introduce memory regressions when integrated. Currently, assigning this alias keeps a strong reference to the manager (which holds `self.pipeline` and wrapped module buffers). This prevents the old pipeline from being garbage-collected during `re_init_pipeline` teardown, leading to a transient ~2x VRAM spike before the new pipeline loads. In the reworked PR, consider explicitly dropping `model_runner.lora_manager` before `gc.collect()` to preserve the clean pre-LoRA teardown behavior.

SamitHuang (commented):
PTAL. Agree on the PR split: 

1) support MiniMax-H3 with the current lora interfaces, but address the parameter name conversion issue with generation quality evaluation. 
2) refactor the current LoRA module to better support pre-fuse and dynamic lora. It should be with a RFC design discussion. Note that for RL, verl-omni uses `hijack DiffusionLoRAManager._load_adapter` to sync the lora weight from trainer to rollout side, it binds the old LoRA Mananger, so your refactor can affect the weight sync.

vllm_omni/diffusion/worker/diffusion_worker.py
mglyn: Thanks for pointing this out. This is fixed in the split PR #6473 (6cfc04c17): `re_init_pipeline()` now drops the legacy manager and both worker/model-runner runtime references before deleting the old pipeline and running GC. A weak-reference regression test verifies that the old pipeline is reclaimable before reload.

PR #6021, opened 2026-08-11
[Feature][MiniCPM-o] Support configurable concurrent duplex sessions

natureofnature:
## Purpose

Support configurable concurrent MiniCPM-o 4.5 native-duplex sessions while
preserving the shipped duplex defaults.

This PR addresses correctness, batching, and memory issues exposed by
multi-session streaming:

- preserve the confirmed-token watermark across segment transitions and rotate
  the sender watermark when a boundary is queued, so a new segment is not
  mistaken for replayed data;
- snapshot the mutable request fields consumed by an asynchronous custom stage
  processor at queue admission time, while leaving the normal no-hook path
  zero-copy;
- batch compatible Code2Wav requests while keeping request-owned state, output
  order, and session isolation;
- combine equal-history requests with different current chunk lengths or
  final-chunk flags in one masked ragged diffusion pass;
- write estimator attention output directly into destination cache storage;
- provide an opt-in BF16 attention-cache representation while keeping diffusion
  compute in FP32. NPU rejects this opt-in and retains its FP32 path.

The checked-in `minicpmo_4_5_duplex.yaml` behavior is unchanged: two admitted
sessions, one active stream, two sequences per stage, FP32 cache, and the base
HiFT/CFM CUDA Graph settings. The generic generation scheduler contains no
MiniCPM/Code2Wav-specific wait policy. Higher concurrency is selected through
deployment configuration; BF16 cache storage and CFM CUDA Graph remain
independent options.

N=8 below is capability evidence from an external run-scoped configuration on
one H800 80GB with CFM eager. It is not a new shipped default or a claim that
the saturated eight-session workload has per-session RTF below 1.

### Changed-area tests

```bash
CUDA_VISIBLE_DEVICES='' PYTHONPATH=$PWD python -m pytest -q \
  tests/core/sched/test_omni_ar_scheduler_streaming.py \
  tests/config/test_config_factory.py::TestDeployConfigLoading::test_load_minicpmo_duplex_deploy_config \
  tests/e2e/online_serving/test_minicpmo_realtime_duplex_drivers.py \
  tests/platforms/npu/test_graph_tools.py \
  tests/platforms/npu/test_minicpmo_code2wav_npugraph.py
```

### Configurable-concurrency E2E

The service used an external run-scoped deploy YAML with `max_sessions`,
`active_stream_window`, and the three stage capacities set to 8, CFM eager,
FP32 cache, and no scheduler wait/coalescing settings.

```bash
python tests/e2e/online_serving/run_minicpmo_realtime_duplex_multi_session.py \
  --url 'ws://127.0.0.1:38461/v1/realtime?duplex=1' \
  --sessions 8 \
  --turns 3 \
  --input-wav tests/assets/minicpmo_4_5/response_required_16k.wav \
  --ref-audio <MiniCPM-o-4_5>/assets/HT_ref_audio.wav \
  --realtime-input \
  --chunk-ms 200 \
  --first-turn-ms 1400 \
  --turn-duration-ms 1400 \
  --turn-duration-ms 1200 \
  --turn-duration-ms 1200 \
  --response-required \
  --temperature 0 \
  --verify-admission-limit 8 \
  --timeout-s 480
```

### Full Seed-TTS FP32/BF16 accuracy and HBM A/B

The same main-based ragged/BF16 Code2Wav kernel was tested once with FP32 cache
and once with only `code2wav_bfloat16_attention_cache: true` added:

```bash
CUDA_VISIBLE_DEVICES=<evaluator-gpu> PATH=<vllm-0.27-venv>/bin:$PATH \
python tests/e2e/accuracy/qwen3_omni/run_qwen_omni_acc_benchmark.py \
  --host 127.0.0.1 \
  --port <service-port> \
  --skip-daily-omni \
  --num-prompts 272 \
  --max-concurrency 8 \
  --temperature 0 \
  --max-seed-tts-mean-wer 0.05 \
  --seed-extra-body-json '{"save_duplex_request_metrics":true}' \
  --seed-backend openai-realtime-tts \
  --seed-endpoint /v1/realtime \
  --seed-tts-turns-per-session 4 \
  --seed-tts-wer-save-items \
  --seed-tts-eval-device cuda:0 \
  --trust-remote-code
```

**vLLM Version:**

0.27.0

**vLLM-Omni Commit:**

- Base used for the final GPU validation: `0e4b28f491721af618525f4ba9f679f0eb5a7157`
- PR: `4fa02f957f100edbe9bca574a21135c9fdcad9f2`

## Test Result

- Changed-area tests: **238 passed, 1 skipped, 14 warnings**.
- Async-request snapshot tests: **87 passed, 14 warnings** for the complete
  connector module. A 16,384-token snapshot smoke measured 0.743 ms mean and
  0.767 ms p95 over 128 snapshots; tensor leaves remained shared.
- External N=8 E2E: **8/8 sessions** and **24/24 audio responses** completed.
  Eight active sessions were admitted, the ninth was rejected with
  `resource_exhausted`, and a replacement was admitted after one session was
  released. Request identity/isolation, response ordering, audio presence, and
  admission/release checks passed.
- No CUDA OOM, `EngineDead`, traceback, illegal-memory-access, stale-watermark,
  request error, cancellation, or Code2Wav batch error was observed in the
  final N=8 run.
- YAML, Ruff, SPDX, forbidden-import, changed-file, and diff checks passed.

Full Seed-TTS result for the unchanged ragged/BF16 numeric kernel:

| Metric | FP32 cache | BF16 cache | BF16 change |
| --- | ---: | ---: | ---: |
| Sessions completed | 272/272 | 272/272 | - |
| Utterances evaluated | 1,088/1,088 | 1,088/1,088 | - |
| Mean WER | 0.024600 | 0.022736 | -0.001864 |
| Median WER | 0 | 0 | 0 |
| Mean audio TTFP | 1,939.50 ms | 1,938.67 ms | -0.83 ms |
| Mean audio RTF | 1.1417 | 1.1227 | -1.67% |
| Aggregate audio throughput | 7.0063x realtime | 7.1253x realtime | +1.70% |
| Benchmark duration | 723.15 s | 711.22 s | -1.65% |
| Peak service HBM | 80,648 MiB | 70,036 MiB | **-10,612 MiB (-13.16%)** |

Both arms evaluated all 1,088 utterances with zero request, missing-PCM, ASR,
WER, or audio-save failures and remained below the required mean WER threshold
of 0.05. The paired BF16-minus-FP32 mean WER delta was -0.001864; a 10,000-sample
paired bootstrap with seed 20260824 gave a 95% CI of
[-0.003838, -0.000183]. This validates generated-speech content WER, not speaker
similarity, MOS/UTMOS, or waveform identity. The performance values are
descriptive single-run measurements, not a claim of statistically significant
speedup.


natureofnature:
@codex review

natureofnature:
@codex review

natureofnature:
@amy-why-3459 PTAL, current due to the large usage of HBM memory from Code2Wav, max batch size is reduced to 2 for code2wav stage only, to keep 8 active sessions.

amy-why-3459:
Thanks for the focused fix — the segment-boundary watermark race and Code2Wav
tiling are the right levers for 8-way duplex.

Approve with comments:

1. Please sync the PR description with `code2wav_max_batch_size: 2` (text still
   says tiles of 4 / 4+4), and drop `[WIP]` from the title if this is ready.

2. `a59f01e` removed the duplex config assert on `code2wav_max_batch_size`.
   Either assert `> 0` (without pinning 2) or document that unset max can OOM
   at 8 sessions.

3. Harden min/max interaction: reject `0 < max < min`, or partition so every
   microbatch still satisfies `code2wav_min_batch_size` (Codex P2).

4. Confirm whether `code2wav_max_batch_size: 2` belongs on the base YAML
   (affects non-duplex too) vs the duplex overlay only.

5. Call out H800-only validation / H100 expectations, and if possible share
   TTFA or chunk-latency at concurrency 8 vs 2 under the 2-tile decode path.

The scheduler/connector watermark tests look solid; happy to approve once the
description + config-guard nits above are addressed.
```

**Comment by natureofnature** [comment]
@codex review

**Comment by amy-why-3459** [comment]
Thanks — the segment-boundary watermark race and a real Code2Wav batch of 8
are the right levers, and the H800 E2E / WER numbers make the direction
credible. Request changes on scope and two correctness holes.

1. Split 4a28e6ad (ReadTheDocs pip 23.1) into its own PR. It is unrelated
   to MiniCPM duplex and pulls docs/CI owners onto this feature review.

2. decode_ragged_batch must not filter Nones:
     [value for value in audios if value is not None]
   A missing vocoder row would shrink the lists and mis-zip caches onto
   the wrong sessions in forward(). Require every row to be filled, same
   order as the input bucket, and raise otherwise. Add a test that a
   hole in the middle fails closed.

3. --synchronized-start is broken on the repo's Python 3.10 (asyncio.Barrier
   is 3.11+) and can hang forever if one session exits before the barrier
   (Codex P2). Bound or abort the barrier; the advertised E2E command does
   not use this flag, but the PR introduces it.

Should-fix:
- Don't hardcode stage_id == 2 for gathering; key off connector extra or
  the Code2Wav model.
- One ready row with chunk_seq in {0,1} forces the whole wave onto the
  initial 50ms wait (no quiet window). Mixed first-chunk + steady-state
  will delay already-hot sessions — test it or document the trade.
- H800 6.4 GiB free + Talker gpu_memory_utilization 0.10 is not an H100
  8-session claim. Smoke H100 or mark the YAML as ≥80GB only.
- Seed-TTS WER used openai-realtime-tts, not native 200ms PCM. Say so.
- cache_signature still partitions by history shape, so the “different
  cache lengths in one masked pass” wording overclaims. Fix the text or
  pad history.

The watermark tests and the DiT short-row / BF16 cache tests look solid.
Happy to re-review once the RTD commit is gone and the ragged output
contract is strict.

**Comment by natureofnature** [comment]
@codex review

File: vllm_omni/model_executor/models/minicpmo_4_5/batched_token2wav.py [hunk 1]
```diff
@@ -727,3 +932,162 @@ def _decode_batch_once(
...
+
+        hidden_rows: list[torch.Tensor | None] = [None] * batch_size
+        conformer_cnn_rows: list[torch.Tensor | None] = [None] * batch_size
+        conformer_att_rows: list[torch.Tensor | None] = [None] * batch_size
+        for (_, last_chunk), rows in encoder_groups.items():
+            group_states = [states[row] for row in rows]
+            group_cache = self._stack_flow_cache(group_states)
+            group_tokens = torch.stack([tokens[row] for row in rows], dim=0)
+            with self._autocast(group_tokens.device):
+                hidden, conformer_cnn, conformer_att = self._encode_chunk(
```
Gaohan123, line 1008: **[P1] Preserve long-input slicing in the ragged decode path**

`decode_ragged_batch()` sends each length group directly to `_encode_chunk()`, bypassing the `plan_token2wav_encode_slices()` protection used by `decode_batch()`. Since `_bucket_key()` no longer includes the current token length, requests with compatible cache shapes but different chunk lengths now enter this path. If one row contains a non-async Talker dump with thousands of codec tokens, it can exceed the CosyVoice RelPos/1024-frame limit, cause excessive allocation or a shape failure, and fail the entire batch.

Please split oversized ragged rows into bounded encoder windows, carry each row’s state across its slices, and apply `last_chunk=True` only to that row’s final slice. A regression test should cover two requests with compatible history caches and different lengths where one exceeds the encoder frame limit, verifying parity with per-request `decode_batch()`.

Gaohan123 (commented):
The experiment result shows it is validated still based on vLLM 0.25. Please update it. And currently, the RTF is still > 1. Does it support realtime interaction?

Sy0307:
> The experiment result shows it is validated still based on vLLM 0.25. Please update it. And currently, the RTF is still > 1. Does it support realtime interaction?

Yes. We should limit RTF < 1 as a hard strict rule. But this PR can just keep the current situation. Dynamic batch scheduling can be added in the following PR.

vllm_omni/model_executor/models/minicpmo_4_5/batched_token2wav.py
natureofnature: Resolved on the current head. Oversized rows in `decode_ragged_batch()` are
delegated to the RelPos-safe exact decode path, which uses bounded
`plan_token2wav_encode_slices()` windows, carries request-owned state across
slices, and applies `last_chunk=True` only to that row's final slice.

The regression now mixes a 10-token final row with a legal 4-token non-final
row under `pre_lookahead_len=3` and a 4-frame encoder limit. It compares each
row's audio plus every Flow and HiFT cache against an independent per-request
`decode_batch()` reference. The Code2Wav file passes 46 tests, and the complete
changed-area suite passes 552 tests.

natureofnature:
Thanks. The PR description is now updated for `main` and vLLM 0.27.0, including
fresh H800 concurrency-8 measurements for Code2Wav gather targets 2, 4, and 8.

I also made the realtime qualification explicit: the measured audio RTF is
commit-to-last-audio time divided by emitted audio duration, and RTF > 1 means
the saturated eight-session workload does not meet a strict per-session
realtime-latency target. The ~7.1x audio throughput in the Seed-TTS table is
aggregate eight-session service throughput, not single-session realtime speed.
Dynamic scheduling remains follow-up work, as suggested.

natureofnature:
@codex review

tests/e2e/online_serving/run_minicpmo_realtime_duplex_multi_session.py
```diff
@@ -101,14 +165,24 @@ def _validate_semantic_isolation(
...
             else []
         )
-        if expected_token not in "".join(transcripts):
+        joined_transcripts = "".join(transcripts).casefold()
+        if expected_token not in joined_transcripts:
+            return False
+        if any(
+            other_token in joined_transcripts
+            for other_index, other_token in enumerate(normalized_expected_tokens)
+            if other_index != result_index
```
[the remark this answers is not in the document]
natureofnature: Resolved in 1e66f9f2cb. The CLI now rejects overlapping --session-expected-token values after the same strip().casefold() normalization used by semantic isolation, covering either ordering and equal markers. Added the cat/cater regression. Final verification: driver 69 passed; related 7-file suite 553 passed; changed-file pre-commit (including Python 3.10 mypy) passed.

natureofnature:
@codex review

natureofnature:
@codex review

vllm_omni/deploy/minicpmo_4_5_duplex.yaml
```diff
@@ -2,11 +2,37 @@
...
 # Thinker -> Talker -> Code2Wav topology. This overlay changes session
-# lifetime and bounds all three stages to two live duplex sessions.
+# lifetime and bounds all three stages to eight live duplex sessions.
+# The eight-session settings require an accelerator with at least 80 GB HBM.
+# They were validated on H800 80GB; H100 has not been validated.
 base_config: minicpmo_4_5.yaml
 pipeline: minicpmo_4_5
 session_mode: duplex
-active_stream_window: 1
+active_stream_window: 8
```
Gaohan123: Before we truly achieve rtf < 1, I don't suggest we modify the default config

vllm_omni/core/sched/omni_generation_scheduler.py
```diff
@@ -35,6 +38,16 @@ def __init__(self, *args, **kwargs):
         model_config = self.vllm_config.model_config
         self._init_omni_io_scheduling_state()
         self._retains_state_across_chunks = bool(getattr(model_config, "retains_state_across_chunks", False))
+        connector_config = getattr(model_config, "stage_connector_config", None)
+        connector_extra = connector_config.get("extra", {}) if isinstance(connector_config, dict) else {}
+        self._code2wav_batch_enabled = "code2wav_batch_target_size" in connector_extra
```
Gaohan123: Could we avoid introducing Code2Wav-specific configuration and state directly into the generic generation scheduler?

Fields such as `code2wav_batch_target_size`, `code2wav_initial_batch_size`, and the associated wait-window state make the core scheduling layer depend on one model backend's batching policy. It would be cleaner to expose this through a generic batch-coalescing policy or connector hook, while keeping the Code2Wav configuration parsing and model-specific semantics in the MiniCPM connector or model adapter.

vllm_omni/core/sched/omni_generation_scheduler.py
```diff
@@ -618,3 +634,76 @@ def _update_request_as_session(self, session: Request, update: StreamingUpdate)
...
+            request.request_id
+            for request in self.running
+            if request.request_id in self.requests and len(request.prompt_token_ids) > request.num_computed_tokens
+        }
+        ready_request_ids.update(
+            request.request_id
+            for request in self.waiting
+            if request.request_id in self.requests and request.prompt_token_ids
+        )
+        ready_count = len(ready_request_ids)
```
Gaohan123: This target check counts all ready requests together, but the Code2Wav execution layer subsequently partitions them by compatibility properties such as prompt identity, cache signature, cache epoch, and initial/steady wave.

As a result, reaching `target_size` here does not necessarily mean that the backend can form a physical batch of that size. For example, eight ready requests split across two prompt/cache buckets may still execute as two batches of four. Conversely, the scheduler may wait for requests that can never join the same backend batch.

Could the waiting decision be based on compatibility groups that use the same grouping semantics as `MiniCPMO45Code2Wav._bucket_key()`?

vllm_omni/core/sched/omni_generation_scheduler.py
```diff
@@ -99,6 +112,9 @@ def schedule(self, throttle_prefills: bool = False) -> SchedulerOutput:
         skipped_waiting_requests = create_request_queue(self.policy)
         req_index = 0
         self._process_pending_omni_inputs(model_mode="generation")
+        if self.chunk_transfer_adapter:
+            if self._should_wait_for_code2wav_batch(scheduled_timestamp):
+                token_budget = 0
```
Gaohan123: Setting the global `token_budget` to zero blocks every running and waiting request in this scheduler instance, not only the Code2Wav requests that are being held for batch coalescing.

Could we defer only the requests belonging to the relevant coalescing group instead? If this scheduler is intentionally guaranteed to serve only one homogeneous Code2Wav workload, that constraint should at least be represented explicitly and validated rather than being assumed here.

vllm_omni/core/sched/omni_generation_scheduler.py
```diff
@@ -618,3 +634,76 @@ def _update_request_as_session(self, session: Request, update: StreamingUpdate)
...
+        if ready_count == 0 or ready_count >= target_size:
+            self._reset_code2wav_batch_wait()
+            return False
+        if self._code2wav_batch_wait_started is None:
+            self._code2wav_batch_wait_started = now
+            self._code2wav_batch_last_ready_count = ready_count
+            self._code2wav_batch_last_progress = now
+            return True
+
+        if ready_count != self._code2wav_batch_last_ready_count:
```
Gaohan123: The quiet-window progress check only compares the number of ready requests. If one ready request disappears while another becomes ready, the membership changes but the count remains the same, so `_code2wav_batch_last_progress` is not refreshed.

If the quiet window is intended to measure time since the latest batch membership change, could we track the ready request ID set—or a compatibility-group generation/version—instead of only `ready_count`? A regression test covering request replacement with an unchanged count would also be useful.

natureofnature:
Updated the PR based on the latest review:

- restored the shipped MiniCPM-o duplex defaults and inherited HiFT/CFM Graph behavior;
- removed Code2Wav-specific wait/coalescing policy from the generic scheduler;
- kept BF16 cache storage and CFM Graph as independent explicit options;
- added enqueue-time request snapshots only for asynchronous custom stage processors;
- moved N=8 to external H800 validation evidence instead of a checked-in/default profile.

Fresh validation on the final candidate: 238 passed, 1 skipped; connector module 87 passed; external N=8 x 3-turn E2E completed 8/8 sessions and 24/24 audio responses with no fatal runtime signatures.

@codex review

Gaohan123 (approved):
LGTM. Thanks

amy-why-3459:
Thanks for restoring the shipped duplex defaults and taking Code2Wav
wait/coalescing back out of the generic scheduler. The watermark rotation,
enqueue-time snapshot, and RelPos-safe ragged fallback look like the right
fixes, and the DiT parity test is the kind of data this change needs.

Approve with comments — none of these need to block the squash if you
want to land CUDA + default-2-session first, but #1 and #2 are leftover
holes in code this PR already touched:

1. NPU never wraps `decode_ragged_batch`. After `_bucket_key()` dropped
   token length / `last_chunk`, the default duplex Code2Wav stage
   (`max_num_seqs: 2`) will enter ragged whenever the two in-flight
   chunks differ. Please apply the same `_flow_execution_context` wrap
   used by `decode_batch` / `setup_batch`, and add a regression that the
   ragged entry actually enters that context. The existing NPU test only
   checks that exact-shape graphs are skipped.

2. The NPU BF16-cache guard uses `bool(extra.get(...))` instead of the
   `_config_bool` helper sitting in the same file. `bool("false")` is
   True, and `test_code2wav_patch_rejects_bfloat16_attention_cache`
   currently parametrizes `"false"` / `"0"` / `"off"` as reject cases,
   so the test is locking in the bug. Please treat those as disabled.

3. Codex P2 is still open for unlimited initial-marker buckets: when
   `code2wav_initial_batch_size == 0` (the default), a singleton
   `tts_is_last_chunk` marker bypasses `_min_batch_size`. Default min=1
   is fine; reject or tile if min>1.

4. There is no steady-state Code2Wav batch cap anymore — only
   `code2wav_initial_batch_size`, and later waves stay unsplit
   (`test_initial_batch_limit_allows_later_full_batch` asserts HiFT 8).
   That matches “no scheduler policy / no default change”, but an H100
   or CFM-Graph deploy that raises `max_num_seqs` has no execution-layer
   HBM guard. Fine as follow-up; please don’t put it back in the
   generic scheduler.

Nits: expose `confirmed_num_computed_tokens` instead of calling the
adapter `_` helper (the scheduler already has a slightly different
`_get_confirmed_num_computed_tokens`); snapshot `prompt_token_ids` even
when it is a ConstantList; `importorskip("cosyvoice2")` on the real DiT
parity test; update the “exact-shape buckets execute independently”
comment now that mixed lengths share one ragged batch.

The 2-session default + H800 N=8-as-evidence framing is the right
product call given RTF>1. Happy to keep the LGTM.

PR #6033, opened 2026-08-11
[BugFix][TTS] R1.1-R1.4: turn silent async-chunk failures into visible ones

### LINKED ISSUE #3833: [Bug]: Async chunk requests can stay in WAITING_FOR_CHUNK indefinitely if upstream chunk never arrives
### Your current environment

<details>

OS: Ubuntu 22.04.5 LTS (x86_64)
Python: 3.12.11
PyTorch: 2.11.0+cu130
CUDA available: True
GPU: NVIDIA GeForce RTX 4090

</details>

### Your code version

<details>
vLLM Version: 0.20.0
</details>

###  Describe the bug

Image

In async chunk mode, downstream requests can remain in `RequestStatus.WAITING_FOR_CHUNK` indefinitely if the expected upstream chunk is never produced or never becomes visible through the connector.

The relevant path is `OmniChunkTransferAdapter.process_pending_chunks()`:

- downstream stage calls `load_async(request)`;
- request status is changed to `WAITING_FOR_CHUNK`;
- request is temporarily removed from the scheduler queue;
- `restore_queues()` puts it back after scheduling;
- if `_poll_single_request()` never observes the expected chunk, the request keeps cycling in `WAITING_FOR_CHUNK`.

Currently I could not find a timeout or cleanup path for this state. This means failures in the upstream stage, connector write failures, dropped terminal chunks, or a custom payload extraction path returning no payload can turn into a hanging streaming request instead of a visible request failure.

Expected behavior:

If a request waits too long for an async chunk, vllm-omni should eventually fail/cleanup the request instead of leaving it in `WAITING_FOR_CHUNK` forever.

linyueqian:
## Purpose

Parts R1.1–R1.4 of #4855. Closes #3833.

Four small changes with one theme: the async-chunk path has several ways to fail that produce a log line and nothing else. A request that will never complete keeps its slot, the client keeps waiting, and no error is ever emitted. Each item below turns one of those into a visible failure.

They ship together because they are individually too small to review in isolation and touch the same two files. They are not otherwise coupled: R1.1 and R1.2 are independent drains, and R1.4 guards the one knob R1.1 reads.

## The four items

**R1.1 — a chunk that never arrives now has a deadline.**
A request parks in `WAITING_FOR_CHUNK` and the receiver re-queues it on every failed poll with no attempt counter. The full-payload path has had a net since `OmniSchedulingCoordinator.collect_timed_out_request_ids`; this is its async-chunk counterpart, driven by the same `VLLM_OMNI_INPUT_WAIT_TIMEOUT_S`.

**R1.2 — a send that gives up now says which request it dropped.**
`save_loop` pops a task with `popleft()` and never re-queues it, and `connector.put` can report failure *without raising*: `SharedMemoryConnector.put` catches everything and returns `False`, reachable when `/dev/shm` is full. The only trace was a warning that printed `None` for the request id, because the task dict has no `request_id` key — a pre-existing bug this fixes.

This reports; it does not fail the request. Failing from the producer does not work: stage 0 is not the `final_output` stage, so an abort synthesized by `finish_requests` is not turned into a client-visible termination by `Orchestrator._route_output`. I had it failing the request until review showed that; making a producer-side drop client-visible needs an orchestrator-side path like `_fail_request_dead_stage`, which is a separate change. The consumer's R1.1 deadline is what ends the request today.

**R1.3 — an unprewarmable request now fails instead of stalling.**
An async-chunk request with no stage-0 `prompt_token_ids` logged a warning and returned, leaving every downstream stage unsubmitted. The request then produced neither output nor error. It now fails immediately with a message naming the cause.

**R1.4 — a typo'd timeout no longer disables the net.**
`VLLM_OMNI_INPUT_WAIT_TIMEOUT_S <= 0` meant "disabled", so `-600` silently turned the deadline off while looking like a real value. Negative and non-finite values are now rejected at startup; `0` remains a supported opt-out but logs which nets it disabled. `nan` and `inf` mattered too: `nan` compares false against every deadline check and `inf` is unreachable, so both disabled the net while looking like numbers.

Rejecting rather than substituting is deliberate — `-1` is a common "no limit" idiom, and quietly turning it into 600 s would give an operator who meant "never time out" mysterious failures ten minutes in.

## Why R2.1 is not here

R2.1 (replace the blocking `sync_q.put` with `put_nowait` + backpressure) is the remaining R item and is deliberately excluded. The four above change error paths only; R2.1 changes orchestration behaviour under normal load and needs its own testing. Bundling it would make this unreviewable and would put the other four at risk.

## On the evidence

**R1.1's reachability is argued from code, not reproduced.** The chain is: 30 of the 80 shipped deploy configs use `connector_of_shared_memory` (the rest use mooncake/mori/yuanrong, which have their own `return False` paths) → `shm_connector.py` catches every exception in `put` and returns `False`, reachable when `/dev/shm` fills → `base.py` pops the task without re-queueing → `chunk_transfer_adapter.py` had no `else` on `if success:` → `chunk_transfer_adapter.py` on `main` has no wait deadline at all (the existing `_waiting_since` lives in `omni_scheduling_coordinator.py` and covers only the full-payload path). Each link is verified. I did **not** trigger it at runtime; that needs a way to exhaust `/dev/shm` mid-run, and I did not want to add a fault-injection hook to production code to test it.

**The producer-crash case is already covered by #4583 and is not the point of this PR.** I measured it: SIGKILL the stage-0 engine mid-stream on current `main` and the request fails in **359.89 ms** with `Stage-0 has no live replica`. What #4583 cannot see is a producer that stays alive while a chunk is dropped — that is what R1.2 addresses.

## Test plan

```bash
pytest tests/core/sched/test_omni_scheduler_mixin_timeouts.py \
       tests/distributed/omni_connectors/test_chunk_transfer_adapter.py
```

**85 passed.** Covers the three triggers named in `collect_timed_out_request_ids`'s docstring (producer crashed, gave up after retries, never emitted a terminal chunk), plus R1.4's parsing table:

| `VLLM_OMNI_INPUT_WAIT_TIMEOUT_S` | result |
|---|---|
| unset / `30` / `0.5` | parsed as-is |
| `0` | stays `0.0`, warning names both transports |
| `-1` / `-600` / `-0.5` | **rejected at startup** |
| `nan` / `inf` / `-inf` / `1e999` | **rejected at startup** |
| `not-a-number` | falls back to `600.0` |

Regression on `tests/core/sched/` + `tests/distributed/omni_connectors/` + `tests/engine/`: the failures in `tests/engine/test_stage_engine_args.py` are pre-existing — current `main` and this branch both report **10 failed / 53 passed** on that file, run side by side in the same venv.

No e2e, for the reason given above.

- **vLLM version:** 0.26.0

## Review history

Three external reviewers went over this before it was updated, and found five real defects, all mine:

- the send-failure hook read `task.get("request_id")`, a key the task dict does not have, so every failure recorded `None` and was dropped
- failures were keyed on `external_req_id` while `self.requests` uses the internal UUID — they never match
- `_process_failed_chunk_sends` was missing the `present_ids` guard that the neighbouring `_process_pending_input_timeouts` already had
- failing the request from the producer does not reach the client, which is why R1.2 is now report-only
- `nan` / `inf` slipped past R1.4's validation

The description was fact-checked too, and three of its claims were wrong: a deploy-config count (I counted string occurrences, not files), a claim that `_waiting_since` appears zero times on `main` (it exists in `omni_scheduling_coordinator.py`), and a claim that R1.1 and R1.2 share a drain (they do not). Those are corrected above.


linyueqian:
@Sy0307

vllm_omni/engine/orchestrator.py
```diff
@@ -2216,10 +2216,31 @@ async def _prewarm_async_chunk_stages(
...
+                "and none were provided; failing the request",
                 request_id,
             )
+            await self.output_async_queue.put(
+                ErrorMessage(
+                    error=(
+                        "async_chunk requires prompt_token_ids on the stage-0 request; "
+                        "an embeds-only prompt cannot prewarm downstream stages"
+                    ),
+                    fatal=True,
```
hsliuustc0106: `fatal=True` means engine death, not request failure: `_handle_output_message` raises `OmniEngineDeadError` on fatal messages, the final-output loop then broadcasts the error to every in-flight request, and the API server treats it as a dead engine (`terminate_if_errored`). One client sending an embeds-only prompt would take down all concurrent requests. This should be a non-fatal `ErrorMessage` with a `status_code`, like `_handle_stage_error`.

vllm_omni/engine/orchestrator.py
```diff
@@ -2216,10 +2216,31 @@ async def _prewarm_async_chunk_stages(
...
+                    request_id=request_id,
+                    stage_id=0,
+                )
+            )
+            await self._cleanup_request_ids(
```
hsliuustc0106, line 2276: The duplex `submit` caller (line ~310) keeps going after this early return: it writes `duplex_stage_fences` / `stage_submit_ts` onto the req_state this cleanup just popped from `request_states`, re-registers the running counter, and returns a success `DuplexStageSubmissionResult` for a request that was just failed and aborted. The prewarm returning `None` gives it no way to know. Returning a bool (or raising) would let the caller bail out. Also missing `close_duplex_sessions=True`, which `_handle_stage_error` passes on its abort path.

vllm_omni/core/sched/omni_scheduler_mixin.py
```diff
@@ -198,6 +232,73 @@ def _process_pending_input_timeouts(self) -> None:
...
+        Making a producer-side drop client-visible needs an orchestrator-side
+        path like ``_fail_request_dead_stage``; the consumer's R1.1 deadline is
+        what actually ends the request today.
+        """
+        adapter = getattr(self, "chunk_transfer_adapter", None)
+        if adapter is None:
+            return
+        failures = adapter.collect_failed_send_request_ids()
+        for request_id, reason in failures.items():
+            if request_id in self.requests:
```
hsliuustc0106, line 293: A failure whose id is no longer in `self.requests` is drained here and never logged anywhere — a silent drop, in the path whose point is visibility. Consider logging outside the `in self.requests` guard.

tests/core/sched/test_omni_scheduler_mixin_timeouts.py
```diff
@@ -102,5 +107,160 @@ def test_process_pending_input_timeouts_disabled_when_timeout_zero(monkeypatch):
...
+def _reload_with(monkeypatch, raw):
+    import importlib
+
+    import vllm_omni.core.sched.omni_scheduler_mixin as mod
+
+    if raw is None:
+        monkeypatch.delenv("VLLM_OMNI_INPUT_WAIT_TIMEOUT_S", raising=False)
+    else:
+        monkeypatch.setenv("VLLM_OMNI_INPUT_WAIT_TIMEOUT_S", raw)
+    return importlib.reload(mod)
```
hsliuustc0106: The reload leaves the module holding whatever was parsed last — e.g. running only `-k zero_warns` leaves `DEFAULT_INPUT_WAIT_TIMEOUT_S = 0.0` for any later test in the same session that imports the mixin. Worth adding a finalizer that reloads with the env unset.

vllm_omni/engine/orchestrator.py
linyueqian: Fixed in 2966bd0cb. It is now a non-fatal `ErrorMessage` with `status_code=400` and `error_type="BadRequestError"`, the same shape `_handle_stage_error` uses for a request-scoped client error.

On the blast radius: in the async server the fatal message never reached `_handle_output_message`, because `final_output_loop` routes any `ErrorMessage` with a `request_id` to that request's queue first. So it was one request there, but `_read_orchestrator_results` still raised `OmniEngineDeadError`, which `serving_speech` / `serving_chat` catch as `EngineDeadError` and answer with `terminate_if_errored`. Offline is where it really bites: `OmniLLM.generate` calls `_handle_output_message` directly, so a fatal message raises out of the `while active_reqs` loop and fails every other request in the batch.

New test: `test_async_chunk_prewarm_without_prompt_token_ids_fails_only_that_request`.

vllm_omni/engine/orchestrator.py
linyueqian: Fixed in 2966bd0cb. `_prewarm_async_chunk_stages` now returns a bool, False on both of its failure exits (this one and the `if not submitted` return further down, which had the same problem), and `submit` raises when it gets False. `DuplexControlPlane.handle_append` wraps the whole append in `try/except Exception`, so the raise comes back to the client as an error result for the append rather than a bogus success.

Worth spelling out what the fall-through cost: `_cleanup_request_ids` sets `running_counter_registered = False` after decrementing, so the `if not request_state.running_counter_registered` block below would have incremented a counter that nothing decrements again.

`close_duplex_sessions=True` added. It is guarded by `self.duplex_control_plane is not None` inside `_cleanup_request_ids`, so it stays a no-op on the two non-duplex call sites.

New test: `test_duplex_submit_bails_out_when_prewarm_failed_the_request`, asserting no fences, no submit timestamp, and no counter increment.

vllm_omni/core/sched/omni_scheduler_mixin.py
linyueqian: Fixed in 2966bd0cb. The `else` now logs at warning, naming why the id is gone (`after the request left this scheduler (finished or aborted)`). Warning rather than error because a late failure for a request that already completed is not the same event as one for a live request, but it should not be invisible either.

New tests cover both branches: `test_log_failed_chunk_sends_reports_a_live_request` and `test_log_failed_chunk_sends_still_reports_a_departed_request`.

tests/core/sched/test_omni_scheduler_mixin_timeouts.py
linyueqian: Fixed in 2966bd0cb. `_reload_with` is now a `reload_with` fixture that calls `monkeypatch.undo()` and reloads once more on teardown, so the module goes back to the ambient value.

The rejection cases were worse than the `zero_warns` one you named: `float()` binds `DEFAULT_INPUT_WAIT_TIMEOUT_S` before the `math.isfinite` check raises, so `test_non_finite_timeout_is_rejected` left `nan` in the module, and `nan` compares false against every deadline check. Running that test alone would have silently disabled both nets for the rest of the session. Added `test_a_rejected_value_does_not_outlive_its_test`, which reproduces that leak and then shows the restore.

linyueqian:
Pushed 2966bd0cb with all four review items addressed; replies are in the threads. Summary:

| Item | Change |
|---|---|
| `fatal=True` on the R1.3 failure | Non-fatal `ErrorMessage` with `status_code=400` / `error_type="BadRequestError"`, matching `_handle_stage_error` |
| Duplex `submit` kept going after a failed prewarm | `_prewarm_async_chunk_stages` returns a bool; `submit` raises on False so `handle_append` reports it. Cleanup also passes `close_duplex_sessions=True` |
| Failure drained without a log when the id left `self.requests` | Logged at warning, naming why the id is gone |
| Module reload leaked into later tests | `reload_with` fixture undoes the env and reloads on teardown |

While verifying I found a test this PR had already broken and fixed it here: `_process_pending_omni_inputs` grew two calls (`_process_pending_chunk_timeouts`, `_log_failed_chunk_sends`), and `test_schedule_lifecycle_helpers_process_and_restore_both_input_paths` asserts that exact call sequence against a stub adapter that had neither method, so it raised `AttributeError`. The stub and the expected sequence are updated.

Verification on 4x L20X (CPU-only collection, vLLM 0.26.0):

- `tests/core/sched/` + `tests/engine/test_orchestrator_error_handling.py` + `tests/engine/test_orchestrator_stage_input_bridge.py`: **144 passed**
- `tests/engine/` + `tests/entrypoints/test_omni_entrypoints.py`: **495 passed, 1 skipped**
- `tests/distributed/omni_connectors/`: one unrelated pre-existing failure, `test_bagel_text2img_mooncake_connector` (`mooncake_master failed to start` on this host); confirmed failing on the pre-change head too

New coverage: the R1.3 non-fatal path end to end through the orchestrator harness, the duplex bail-out (no fences, no submit timestamp, no counter increment), both branches of the send-failure report, and the module-reload restore.

vllm_omni/engine/orchestrator.py
hsliuustc0106: Makes sense — non-fatal with the 4xx is the right shape, and thanks for tracing where the raise actually lands in each entrypoint; the offline batch case is indeed the worst one.

vllm_omni/engine/orchestrator.py
hsliuustc0106: Looks good — and good catch on the second `if not submitted` exit having the same problem.

vllm_omni/core/sched/omni_scheduler_mixin.py
hsliuustc0106: Makes sense.

tests/core/sched/test_omni_scheduler_mixin_timeouts.py
hsliuustc0106: Fair — the nan leak is worse than the zero one, good catch.

linyueqian:
@hsliuustc0106 all four of your points are addressed in 2966bd0cb and you signed off on each thread, so I think this one only needs your review state flipped from `COMMENTED` to `APPROVE` — `reviewDecision` is still `REVIEW_REQUIRED` and the base branch policy is holding it.

CI is green on the merged head `7dd31d47e`. It had no Buildkite build at all until yesterday, because a fork PR needs a label event after the push and the last one predated both my commit and your merge of main; re-toggling `ready` fired it:

| Pipeline | Result |
|---|---|
| `buildkite/vllm-omni` | success, 20/20 jobs |
| `buildkite/vllm-omni-intel-ci` | success |
| `buildkite/vllm-omni-npu-ci` | success |
| `buildkite/vllm-omni-amd-ci` | failure, pre-existing on main |

The amd red is not this PR: all 19 jobs ran, and the only failure is `Simple · Diffusion Test` with three `tests/diffusion/models/ltx2/test_ltx2_phase_adapter.py` cases hitting `NotImplementedError: Could not run 'vllm::rocm_unquantized_gemm'`. Build 10622 on `main` (`68bf8c2c4`) fails the identical three across its two shards, and amd-ci is red on 7 of the last 8 main commits.

I checked the green is real rather than a skipped-collection pass — the new tests show up by name in the logs. `Simple · Engine&Entrypoints` (1884 passed) ran `test_async_chunk_prewarm_without_prompt_token_ids_fails_only_that_request` and `test_duplex_submit_bails_out_when_prewarm_failed_the_request`; `Simple · Other` (2303 passed) ran all three `test_log_failed_chunk_sends_*`, `test_a_rejected_value_does_not_outlive_its_test`, and `test_schedule_lifecycle_helpers_process_and_restore_both_input_paths`.

Auto-merge (squash) is armed, so it will land on its own once the approval is in.

PR #6034, opened 2026-08-11
[Bugfix] Use request-scoped RNG for MOSS-TTS-Realtime audio codes

webcodes-cz:
## Purpose

`MossTTSRealtimeTalkerForGeneration` samples audio codes through
`MossTTSRealtimeLocalTransformer.generate_frame()`, which never receives a
`torch.Generator`. The leaf helper already supports one:

```python
def _sample_token(logits, temperature, top_k, top_p, do_sample,
                  generator: torch.Generator | None = None) -> torch.Tensor:
    ...
    sampled = torch.multinomial(flat, num_samples=1, generator=generator)
```

but `generate_frame()` neither accepts nor forwards it, so every Realtime
request samples from the process-global RNG. Two consequences:

1. **`seed` has no effect on audio codes.** Identical seeded requests produce
   different audio. Measured on `v0.24.0`: eight identical requests returned
   outputs from 1.84 s to 7.68 s, with peak per-second RMS varying about 12x.
2. **Concurrent requests contaminate each other.** The global RNG state advances
   across requests, so one request's sampling depends on how many others are
   interleaved with it.

The `talker_mtp` path already solves this for MOSS-TTS-Local in
`OmniGPUModelRunner`: `_explicit_talker_seed()` reads
`sampling_params.extra_args["tts_local_seed"]`, `_row_generator()` caches a
per-request `torch.Generator`, finished requests are evicted, and
`talker_mtp_accepts_per_row_generators` guards the batched path. Realtime has no
equivalent because it does not go through `talker_mtp`.

This PR wires Realtime into the same convention:

- `generate_frame()` accepts `generator` and forwards it to `_sample_token()`.
- `MossTTSRealtimeTalkerForGeneration` declares
  `uses_request_scoped_omni_generators` and consumes a `request_generators`
  sequence, one per in-flight request, failing loudly on a count mismatch rather
  than silently reusing one stream across rows.
- `OmniGPUModelRunner` builds those generators per request, reuses the same
  `extra_args["tts_local_seed"]` key as the Local path, and drops them in the
  existing `finished_req_ids` cleanup.

## Test Plan

Rebased onto current `main` and verified that the three changed modules import
cleanly (`py_compile`). Behaviour was exercised in a downstream vLLM-Omni
deployment of `OpenMOSS-Team/MOSS-TTS-Realtime` on a single dedicated RTX 5090,
using one fixed reference clip and one fixed prompt:

- repeated requests with the same explicit seed;
- requests with two different reference clips at the same seed;
- concurrent requests carrying distinct seeds.

**vLLM Version:** as pinned by vllm-omni `main` at the base commit below

**vLLM-Omni Commit:** `12a5f6fbf3ac`

## Test Result

- Identical seeded requests became bit-identical, where before they varied in
  both duration and amplitude.
- Concurrent requests with distinct seeds each stayed individually
  reproducible.
- No change in throughput was observed.

## Open question for reviewers

`_row_generator()` on the Local path returns `None` when no explicit seed is
given, so unseeded requests keep sharing the global RNG and can still
contaminate each other under concurrency. This PR allocates a randomly seeded
generator in that case, so isolation does not depend on the caller passing a
seed.

That is deliberate but inconsistent with the Local path. Happy to make Realtime
match Local instead, or to align both — whichever you prefer.


yinpeiqi:
Thanks for the fix! Please fix the pre-commit. The model-side wiring looks good.

One question: `_omni_request_generators` shouldn't live in `OmniGPUModelRunner`. Only MOSS-TTS-Realtime uses `uses_request_scoped_omni_generators` today, so this is model-specific plumbing in the base runner.

Could you extract the cache, finished-request cleanup, and kwargs assembly into a mixin (similar to `OmniConnectorModelRunnerMixin`) and mix it into the runner that actually needs it, e.g. `GPUGenerationModelRunner`?

vllm_omni/model_executor/models/moss_tts/modeling_moss_tts_talker.py
```diff
@@ -1101,6 +1108,7 @@ def make_omni_output(
                     do_sample=True,
                     repetition_penalty=1.1,
                     history_per_codebook=hist_per_cb,
+                    generator=request_generators[i],
```
lishunyang12, line 1111: Add CPU tests for seeded replay and interleaved Realtime requests.

PR #6035, opened 2026-08-11
[Bugfix] Wire MOSS-TTS-Realtime reference conditioning online

webcodes-cz:
## Purpose

The online MOSS-TTS-Realtime path accepts and decodes `ref_audio`, but sends it
to the engine as `prompt_audio_array`. The Realtime talker does not consume that
field: it expects the processor-built `(L, 17)` prompt in `prompt_token_ids` and
`codes.ref`, with remaining text tokens in `ids.all`.

As a result `/v1/audio/speech` *requires* a reference clip and then generates
without conditioning on it. The current code says as much:

> `AutoProcessor.from_pretrained` doesn't auto-discover `MossTTSRealtimeProcessor`
> […] full Realtime support needs a separate `processor.from_module` path which
> we don't wire here.

This PR wires that path:

- Locate `processing_mossttsrealtime.py` in the exact local or Hugging Face model
  snapshot and instantiate `MossTTSRealtimeProcessor` explicitly.
- Encode the reference at 24 kHz with the standalone 16-codebook
  `OpenMOSS-Team/MOSS-Audio-Tokenizer` codec. `MOSS_TTS_CODEC_PATH` remains an
  optional local-model override.
- Build the same system/reference, assistant-header and 12-token prefill grid the
  upstream Realtime inference path uses, then forward remaining text in `ids.all`.
- Cache CPU reference codes by decoded-audio identity and codec contract, with a
  single-flight guard so concurrent requests encode a given reference once.
  Encoding runs outside the asyncio event loop.

## Test Plan

Added regressions for the 16-codebook orientation and for reference codes
reaching the online Realtime prompt. Verified `py_compile` on both changed
modules and that the branch merges cleanly with current `main`.

Behaviour was exercised in a downstream deployment of
`OpenMOSS-Team/MOSS-TTS-Realtime` on a single dedicated RTX 5090: two different
official reference clips, identical prompt text and identical seed, each
repeated.

A matched control run of the upstream Transformers implementation, same
checkpoint revision and sampling defaults, was used to check that the
constructed prompt does not introduce behaviour of its own.

**vLLM Version:** as pinned by vllm-omni `main` at the base commit below

**vLLM-Omni Commit:** `12a5f6fbf3ac`

## Test Result

- Two different reference voices produced byte-distinct PCM, where before the
  output was byte-identical regardless of the reference supplied.
- Each reference repeated bit-for-bit at a fixed seed. The companion
  request-scoped RNG change (#6034) was applied in the same canary, so the
  fixed-seed comparison isolates reference identity rather than process-global
  sampling state.
- The Transformers control reproduced the model's quiet-output and
  long-output tail at least as strongly as this path, so the constructed prompt
  is not the source of that behaviour.

A full pytest run was not available in the scratch environment used; CI coverage
would be welcome.

## Operational note

The processor and CPU reference codec are loaded lazily, and the first encode in
a fresh process is substantially slower than subsequent ones. Deployments that
care about first-request latency should issue a reference-conditioned warmup
before marking the endpoint ready; later references reuse the loaded codec and
the decoded-reference cache.

### DIFF
--- a/vllm_omni/entrypoints/openai/serving_speech.py
+++ b/vllm_omni/entrypoints/openai/serving_speech.py
@@ -1,12 +1,14 @@
 import asyncio
 import base64
 import hashlib
+import importlib.util
 import io
 import json
 import math
 import os
 import re
 import struct
+import sys
 import time
 from collections import OrderedDict
 from collections.abc import Mapping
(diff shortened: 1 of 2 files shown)


vllm_omni/entrypoints/openai/serving_speech.py
```diff
@@ -1537,7 +1539,172 @@ def _get_moss_processor(self):
...
+                        wav_list,
+                        sr,
+                    )
+                    self._speaker_cache.put(cache_key, {"codes": codes})
+                    return codes.clone()
+
+            task = asyncio.create_task(_encode_and_cache())
+            inflight[cache_key] = task
+        try:
+            return tokenizer, processor, (await task).clone()
```
lishunyang12, line 1664: Shield the shared encoding task and remove it only after it finishes.

vllm_omni/entrypoints/openai/serving_speech.py
```diff
@@ -1537,7 +1539,172 @@ def _get_moss_processor(self):
...
+            module = importlib.util.module_from_spec(spec)
+            sys.modules[module_name] = module
+            spec.loader.exec_module(module)
+
+        tokenizer = AutoTokenizer.from_pretrained(model_path, trust_remote_code=True)
+        processor = module.MossTTSRealtimeProcessor(tokenizer=tokenizer)
+        codec_id = os.environ.get(
+            "MOSS_TTS_CODEC_PATH", "OpenMOSS-Team/MOSS-Audio-Tokenizer"
+        )
+        codec = AutoModel.from_pretrained(codec_id, trust_remote_code=True)
```
lishunyang12, line 1592: Load the processor and codec off the event loop under a shared initialization task.

PR #5891, opened 2026-08-07
[Perf][Diffusion] MiniMax-H3: opt into packed varlen attention on NPU to eliminate quadratic mask materialization

brandneway:
### Purpose

On NPU, every MiniMax-H3 attention call materializes the `[1, S]` padding mask into a
`[1, 1, Sq, Skv]` tensor via `_maybe_reshape_attn_mask(..., mask_mode="full_qk")`
(`expand(...).contiguous()`). For a typical t2va request (packed sequence S = 63,232) a
memory snapshot shows:

- **45.6 TiB of allocation churn per request** at this single line
  (`sdpa.py:46`, 4900 calls = ~100 per denoise step × 50 steps), two sizes observed:
  3.72 GiB (= 63,232² bool) and 14.90 GiB;
- the 14.90 GiB transient masks leave **two 14.90 GiB allocator segments permanently
  reserved but idle** — 33.7 GiB (59%) of the 57 GiB reserved HBM is unreclaimable
  fragmentation;
- the mask content is identical across all 4900 materializations (it only encodes the
  ≤63-row tail padding), so the work is pure waste.

The same model on CUDA never pays this: `forward_cuda` sees the packed
`cu_seqlens_q/k` + `max_seqlen_q/k` keys in `AttentionMetadata.extra` and goes straight
to flash-attn varlen, never reading `attn_mask`. The NPU path simply lacked the
equivalent.

### What this PR does

Adds two mask-free paths to the NPU flash backend, behind a **per-forward opt-in**
so no other model is affected:

- **Default: packed varlen** — `FlashAttentionImpl._forward_varlen_packed_npu` runs
  mindiesd `attention_forward_varlen` (TND layout) with document boundaries derived
  from the packed cu_seqlens metadata, mirroring what the CUDA path already does with
  the same keys. The padding document is excluded by the kernel itself; `attn_mask`
  is never read nor materialized.
- **When `MINDIE_SD_FA_TYPE=ascend_laser_attention`: prefix K/V slicing** —
  `_forward_prefix_kv_slice_npu` slices K/V to the real document prefix (zero-copy
  views) and calls mindiesd `attention_forward` with `attn_mask=None`, so the
  env-selected op type is honored (the TND varlen op cannot select op types).
- Both paths share one contract check (`_resolve_packed_seq_npu`): document lengths
  are built from the Python ints already in `max_seqlen_q/k`, so **no device sync**
  is introduced. Any metadata that doesn't match the `[real, pad]` two-document
[...]

PR #6037, opened 2026-08-11
[NPU][Diffusion] MiniMax-H3 RainFusion end_step tail fallback

HAAZZZEEEE:
## Purpose

Add two knobs to the RAINFUSION_ATTN block-sparse path for MiniMax-H3 on
Ascend NPU:

1. **`end_step` tail fallback** — keep the last N denoise steps dense so the
   tail of the video is never sparsified. Together with `start_step`, the
   config now supports `{start_step, end_step, sparsity}` to express a
   dense → sparse → dense window across the denoise schedule.

2. **`precision` knob for the rf_v3 sparse-attention kernel** — select the
   execution precision per `block_sparse` config:
   - `mix` (default): EagleQBSA mixed precision (Q/K per-block INT8 + V
     per-channel FP8)
   - `fp8`: BSA FP8 path (Hadamard rotation + full FP8 block quantization of
     Q/K/V)
   - `bf16`: no quantization, pure BF16 sparse attention (previous behavior)

`BlockSparseSpec.precision` is validated against `RainFusionPrecision` and
forwarded via `backend_kwargs → RainFusionConfig → mindiesd.sparse_attention`.

## Test Plan

**vLLM Version:** 0.28.0+empty

**vLLM-Omni Commit:** this PR (846ad825 + 6d7b000c)

- Hardware: Ascend 950PR (A5) node, 4x NPU, 128 GB HBM per card
- Software: vllm-ascend latest main + MindIE-SD dev (rf_v3 EagleQBSA)
- Serve: MiniMax-H3 FL2VA, `RAINFUSION_ATTN`, layerwise offload,
  sparsity=0.8
- Request (standard recipe): T2VA, 1344x768, 50 steps, 24 fps,
  duration 8.7 s
- Verification probes (gated by `EQBSA_PROBE=1`):
  - `rainfusion_attn.py::_resolve_plan` prints per-step
    `dense(start)|sparse|dense(end)` decisions
  - `sparse_flash_attn_rf_v3.py::rain_fusion_attention_v3` prints
    `precision/sparsity/seqlen` before the `eagle_quant_block_sparse_attention`
    call

### A/B performance

| config | fallback window | E2E time | output |
|--------|----------------|----------|--------|
| dense (FLASH_ATTN) | — | ~561 s | 9.7 MB |
| bf16, no fallback | start=0 / end=0 | ~272 s | 9.7 MB |
| bf16 | start=8 / end=12 | ~390 s | 9.9 MB |
| mix (EagleQBSA), no fallback | start=0 / end=0 | ~241 s | 9.6 MB |
| mix (EagleQBSA) | start=8 / end=12 | ~375 s | 9.8 MB |

Component-wise breakdown:
- Sparse vs dense (FLASH_ATTN): **~51% faster** (561 s -> 272 s, bf16 no
  fallback)
- EagleQBSA quantization vs bf16 at the same no-fallback window: **~11%
  faster** (272 s -> 241 s)
- 8/12 fallback window protects quality at ~130 s cost vs no fallback
  (mix: 241 s -> 375 s; bf16: 272 s -> 390 s)
- Best practice (mix 8/12) vs dense baseline: **~33% faster**
  (561 s -> 375 s)
- Maximum gain (mix, no fallback) vs dense baseline: **~57% faster**
  (561 s -> 241 s)

Kernel-level confirmation:
- `[EQBSA-PROBE]` fired with `precision='mix'`, sparsity 0.8 in the sparse
  window (5,800 invocations for 8/12; all 50 steps for no-fallback) — the
  EagleQBSA kernel (`eagle_quant_block_sparse_attention`, Q/K per-block
  INT8 + V per-channel FP8) is actually dispatched on NPU, not silently
  falling back to dense
- `[PLAN-PROBE]` confirmed the dense -> sparse -> dense step pattern for
  the 8/12 window: steps 0-7 dense(start), 8-36 sparse, 37-48 dense(end)

### Video output comparison

[//]: # (Upload videos here)

- **A: dense (FLASH_ATTN)** — 

https://github.com/user-attachments/assets/d80047dd-5ba9-4e6c-b29c-a531bba724f1

- **B: bf16, no fallback** —

https://github.com/user-attachments/assets/e808d88f-6192-4b16-8548-8c4364680bcd

- **C: bf16, 8/12** —

https://github.com/user-attachments/assets/4a747a68-e7f6-4b33-9b6c-c141e4bf62d9

- **D: mix (EagleQBSA), no fallback** — 

Uploading out_768p_E_mix_nofb.mp4…

- **E: mix (EagleQBSA), 8/12** — [upload: video file]

https://github.com/user-attachments/assets/97740b0e-87db-4f3c-b294-4fb08eb1b8af

### DIFF
--- a/vllm_omni/__init__.py
+++ b/vllm_omni/__init__.py
@@ -18,6 +18,17 @@ Architecture:
 # throw in patch if the versions differ.
 from .version import __version__, __version_tuple__  # isort:skip # noqa: F401
 
+# Early-register MindIE-SD custom ops (QBSA etc.) before CANN snapshots the
+# custom-op registry at the first regInfo lookup (vllm-ascend model
+# load/warmup). Without this, lazily-imported mindiesd ops (e.g.
+# eagle_quant_block_sparse_attention via RAINFUSION_ATTN) fail with EZ1014
+# "inferShape function does not exist" at first call. Mirrors PR #5891.
+try:
+    import mindiesd  # noqa: F401
+except Exception:  # pragma: no cover - mindiesd optional
+    pass
+
+
 try:
     from . import patch  # noqa: F401
 except ModuleNotFoundError as exc:  # pragma: no cover - optional dependency
--- a/vllm_omni/diffusion/attention/backends/rainfusion_attn.py
+++ b/vllm_omni/diffusion/attention/backends/rainfusion_attn.py
@@ -1,8 +1,9 @@
 # SPDX-License-Identifier: Apache-2.0
-# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM-Omni project
 
 from __future__ import annotations
 
+import functools
 import inspect
 import math
 from dataclasses import dataclass
@@ -57,6 +58,22 @@ _INCOMPATIBLE_MINDIESD = (
 )
 
 
+# Whether the installed mindiesd ``sparse_attention`` accepts ``precision=``.
+# Releases without it accept the kwarg through ``**kwargs`` but silently ignore
+# it, so a requested mix/fp8 mode would silently run the BF16 path. Cached
+# because forwards run per layer per denoise step.
+@functools.cache
+def _mindiesd_supports_precision() -> bool:
+    try:
+        from inspect import signature
+
+        from mindiesd import sparse_attention
+
+        return "precision" in signature(sparse_attention).parameters
+    except Exception:
+        return False
+
+
 def _try_extract_layer_index(prefix: str) -> int | None:
     if not prefix:
         return None
@@ -86,6 +103,7 @@ class RainFusionConfig:
     sparsity: float = 0.0
     start_step: int = 0
     end_step: int = 0
+    precision: str = "bf16"
     skip_layers: frozenset[int] = frozenset()
 
     @classmethod
@@ -95,6 +113,7 @@ class RainFusionConfig:
             sparsity=float(bk.get("sparsity", 0.0)),
             start_step=int(bk.get("start_step", 0)),
             end_step=int(bk.get("end_step", 0)),
+            precision=str(bk.get("precision", "bf16")),
             skip_layers=frozenset(bk.get("skip_layers") or ()),
         )
 
@@ -445,6 +464,13 @@ class RainFusionAttentionImpl(AttentionImpl):
             from mindiesd import sparse_attention
         except ImportError:
             raise ImportError(_MISSING_MINDIESD)
+        if self.rainfusion.precision != "bf16" and not _mindiesd_supports_precision():
+            raise RuntimeError(
+                f"block_sparse.precision={self.rainfusion.precision!r} requires MindIE-SD "
+                "with sparse_attention(precision=...) support; the installed mindiesd "
+                "silently ignores it and would run the BF16 path. Install a compatible "
+                "MindIE-SD release or use precision='bf16'."
+            )
 
         used = plan.used_len
         q, k, v = (tensor[:, :used] for tensor in (query, key, value))
@@ -457,6 +483,7 @@ class RainFusionAttentionImpl(AttentionImpl):
             "inner_precise": _INNER_PRECISE,
             "block_size": _BLOCK_SIZE,
             "sparsity": self.rainfusion.sparsity,
+            "precision": self.rainfusion.precision,
         }
         if plan.video_spans is not None:
             out = sparse_attention(
--- a/vllm_omni/diffusion/attention/backends/rainfusion_attn.py
+++ b/vllm_omni/diffusion/attention/backends/rainfusion_attn.py
@@ -72,6 +72,7 @@ class RainFusionConfig:
 
     sparsity: float = 0.0
     start_step: int = 0
+    end_step: int = 0
     skip_layers: frozenset[int] = frozenset()
 
     @classmethod
@@ -80,6 +81,7 @@ class RainFusionConfig:
         return cls(
             sparsity=float(bk.get("sparsity", 0.0)),
             start_step=int(bk.get("start_step", 0)),
+            end_step=int(bk.get("end_step", 0)),
             skip_layers=frozenset(bk.get("skip_layers") or ()),
         )
 
@@ -232,7 +234,16 @@ class RainFusionAttentionImpl(AttentionImpl):
             return None
         if is_forward_context_available():
             step_idx = get_forward_context().denoise_step_idx
+            total_steps = get_forward_context().total_denoise_steps
             if step_idx is not None and step_idx < rf.start_step:
+                return None
+            # Tail fallback: keep the last ``end_step`` denoise steps dense.
+            if (
+                rf.end_step > 0
+                and step_idx is not None
+                and total_steps is not None
+                and step_idx >= total_steps - rf.end_step
+            ):
                 return None
         if self.qkv_layout is None:
             # The sparse path reads the sequence off dim 1, which the tensors alone
--- a/vllm_omni/diffusion/attention/backends/rainfusion_attn.py
+++ b/vllm_omni/diffusion/attention/backends/rainfusion_attn.py
@@ -85,6 +85,7 @@ class RainFusionConfig:
 
     sparsity: float = 0.0
     start_step: int = 0
+    end_step: int = 0
     skip_layers: frozenset[int] = frozenset()
 
     @classmethod
@@ -93,6 +94,7 @@ class RainFusionConfig:
         return cls(
             sparsity=float(bk.get("sparsity", 0.0)),
             start_step=int(bk.get("start_step", 0)),
+            end_step=int(bk.get("end_step", 0)),
             skip_layers=frozenset(bk.get("skip_layers") or ()),
         )
 
@@ -252,7 +254,16 @@ class RainFusionAttentionImpl(AttentionImpl):
             return None
         if is_forward_context_available():
             step_idx = get_forward_context().denoise_step_idx
+            total_steps = get_forward_context().total_denoise_steps
             if step_idx is not None and step_idx < rf.start_step:
+                return None
+            # Tail fallback: keep the last ``end_step`` denoise steps dense.
+            if (
+                rf.end_step > 0
+                and step_idx is not None
+                and total_steps is not None
+                and step_idx >= total_steps - rf.end_step
+            ):
                 return None
         if self.qkv_layout is None:
             # The sparse path reads the sequence off dim 1, which the tensors alone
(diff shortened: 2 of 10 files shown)


vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
```diff
@@ -43,6 +43,136 @@
 
     from vllm_omni.diffusion.data import OmniDiffusionConfig
 
+
+@torch.no_grad()
```
hsliuustc0106: do not change model file for a hardware attn backend

vllm_omni/diffusion/attention/backends/rainfusion_attn.py
```diff
@@ -232,8 +236,17 @@ def _resolve_plan(self, attn_metadata: AttentionMetadata | None) -> RainFusionPl
...
         if is_forward_context_available():
             step_idx = get_forward_context().denoise_step_idx
+            total_steps = get_forward_context().total_denoise_steps
             if step_idx is not None and step_idx < rf.start_step:
                 return None
+            # Tail fallback: keep the last ``end_step`` denoise steps dense.
+            if (
+                rf.end_step > 0
+                and step_idx is not None
+                and total_steps is not None
```
david6666666, line 283: MiniMax-H3 never publishes total_denoise_steps: its loop only sets the step and timestep, so this condition is always false and end_step has no effect. Please plumb num_steps through that producer and cover the first tail step with a regression test.

vllm_omni/diffusion/attention/backends/rainfusion_attn.py
```diff
@@ -339,6 +352,7 @@ def _forward_sparse_npu(
             latent_shape_q=plan.latent_shape,
             latent_shape_k=plan.latent_shape,
             sparsity=self.rainfusion.sparsity,
+            precision=self.rainfusion.precision,
```
david6666666: Older MindIE-SD releases accept this through **kwargs but silently ignore it, so selecting mix or fp8 can still run the old BF16 path. Please gate this on a MindIE-SD capability/version, or pin and document the required revision.

vllm_omni/__init__.py
```diff
@@ -18,6 +18,17 @@
 # throw in patch if the versions differ.
 from .version import __version__, __version_tuple__  # isort:skip # noqa: F401
 
+# Early-register MindIE-SD custom ops (QBSA etc.) before CANN snapshots the
+# custom-op registry at the first regInfo lookup (vllm-ascend model
+# load/warmup). Without this, lazily-imported mindiesd ops (e.g.
+# eagle_quant_block_sparse_attention via RAINFUSION_ATTN) fail with EZ1014
+# "inferShape function does not exist" at first call. Mirrors PR #5891.
+try:
+    import mindiesd  # noqa: F401
```
david6666666: This imports an NPU-only package for every vllm_omni import, including CUDA processes. If MindIE-SD is present, its vendor/custom-op side effects now run before platform selection. Could we keep the early registration behind an NPU-only bootstrap so CUDA startup stays isolated?

vllm_omni/diffusion/data.py
```diff
@@ -1771,6 +1795,8 @@ def backend_kwargs(self) -> dict[str, Any] | None:
             bs = self.block_sparse
             kw["sparsity"] = bs.sparsity
             kw["start_step"] = bs.start_step
+            kw["end_step"] = bs.end_step
+            kw["precision"] = bs.precision
```
david6666666, line 1804: The existing block-sparse config tests still assert the old kwargs dictionaries, so this head will fail those tests. Please update them and add validation/forwarding coverage for both new knobs.

vllm_omni/diffusion/data.py
```diff
@@ -1678,17 +1692,27 @@ class BlockSparseSpec:
...
     skip_layers: str | list[int] | None = None
     skip_layer_indices: set[int] | None = field(default=None, repr=False)
 
     def __post_init__(self) -> None:
         self.sparsity = _in_range(self.sparsity, "block_sparse.sparsity", 0.0, 1.0) or 0.0
         if self.start_step < 0:
             raise ValueError(f"block_sparse.start_step must be >= 0; got {self.start_step!r}.")
+        if self.end_step < 0:
+            raise ValueError(f"block_sparse.end_step must be >= 0; got {self.end_step!r}.")
+        if self.precision not in {p.value for p in RainFusionPrecision}:
```
david6666666, line 1716: RainFusionPrecision members pass this check because they subclass str, but RainFusionConfig later calls str(value), producing RainFusionPrecision.MIX instead of mix. Please normalize enum inputs to .value here (or use StrEnum).

vllm_omni/diffusion/attention/backends/rainfusion_attn.py
HAAZZZEEEE: Fixed. You're right — we plumbed num_steps through: the publish function now takes a total_steps argument and writes it via the new set_forward_context_denoise_total_steps setter; the H3 denoise loop passes num_steps, and the batched publish site in the pipeline passes len(extra[_STEP_SIGMAS_VIDEO]) - 1; regression tests cover the first tail step (step_idx == total_steps - end_step), the step just before the tail window (still sparse), and end_step=0 (never falls back) — three cases.

vllm_omni/diffusion/attention/backends/rainfusion_attn.py
HAAZZZEEEE: Gated on MindIE-SD capability. _mindiesd_supports_precision() inspects the installed sparse_attention signature for a precision parameter (cached once per process). When a non-bf16 precision is requested but the installed mindiesd lacks it, _forward_sparse_npu now raises RuntimeError instead of silently running the BF16 path. The requirement is also documented in docs/user_guide/diffusion/attention_backends/rainfusion.md.

vllm_omni/__init__.py
HAAZZZEEEE: Plz take a look, does this approach work? 

Kept the early registration but made it NPU-only. 

vllm_omni/__init__.py now wraps `import mindiesd` with hasattr(torch, "npu") and torch.npu.is_available() — CUDA processes short-circuit on importing vllm_omni and never touch the NPU-only package; NPU processes keep the pre-snapshot registration to prevent EZ1014. 

We intentionally kept the package-level location: an earlier attempt to move it to the NPU platform hook (PR #5891 style) was measured to run too late in spawn workers (after the CANN custom-op registry snapshot), which would regress EZ1014. The platform.py hook is kept as a harmless no-op backstop.

vllm_omni/diffusion/data.py
HAAZZZEEEE: Switched RainFusionPrecision from (str, Enum) to StrEnum, so str() returns the member value directly (RainFusionPrecision.MIX -> "mix") and the value survives the backend_kwargs() -> RainFusionConfig.from_backend_kwargs() round trip. Added test_block_sparse_precision_enum_input_normalized as a regression guard.

vllm_omni/diffusion/data.py
HAAZZZEEEE: Updated. 

The two existing assertions now expect the new keys: test_block_sparse_defaults_applied_when_backend_selected and test_block_sparse_skip_layers_selector_expanded include end_step/precision. test_block_sparse_invalid_values adds end_step=-1 and precision="bogus" for validation coverage, and test_block_sparse_end_step_precision_forwarded covers forwarding of both new knobs.

vllm_omni/__init__.py
```diff
@@ -18,6 +21,21 @@
 # throw in patch if the versions differ.
 from .version import __version__, __version_tuple__  # isort:skip # noqa: F401
 
+# Early-register MindIE-SD custom ops (QBSA etc.) before CANN snapshots the
```
david6666666: please don't modify here

david6666666 (changes requested):

vllm_omni/__init__.py
HAAZZZEEEE: OK

vllm_omni/diffusion/attention/backends/rainfusion_attn.py
```diff
@@ -50,6 +50,26 @@
...
 
+# Whether the installed mindiesd ``sparse_attention`` accepts ``precision=``.
+# Releases without it accept the kwarg through ``**kwargs`` but silently ignore
+# it, so a requested mix/fp8 mode would silently run the BF16 path. Cached
+# because forwards run per layer per denoise step.
+_MINDIESD_HAS_PRECISION: bool | None = None
+
+
+def _mindiesd_supports_precision() -> bool:
+    global _MINDIESD_HAS_PRECISION
```
david6666666: why use global var, please change

david6666666 (dismissed):
please add A/B test to show performance gain, and paste your video output comparasion

vllm_omni/diffusion/attention/backends/rainfusion_attn.py
HAAZZZEEEE: Fixed: replaced the module-level `_MINDIESD_HAS_PRECISION` + `global` memoization with `functools.cache` — the cache is now encapsulated in the function object instead of the module namespace.

Rationale: the probe has a single call site (the gate in `_forward_sparse_npu`) that only reads the return value, so there is no cross-function sharing need that would justify a `global`; but it runs per layer per denoise step, so caching is still required. `functools.cache` keeps the function pure and the namespace clean.

Behavior unchanged: first call probes, later calls hit the cache. Covered by `test_precision_non_bf16_requires_mindiesd_support` and `test_precision_non_bf16_passes_gate_when_supported` (both pass).

HAAZZZEEEE:
> please add A/B test to show performance gain, and paste your video output comparasion
OK, already in test result

vllm_omni/diffusion/attention/backends/rainfusion_attn.py
```diff
@@ -446,6 +483,7 @@ def _forward_sparse_npu(
             "inner_precise": _INNER_PRECISE,
             "block_size": _BLOCK_SIZE,
             "sparsity": self.rainfusion.sparsity,
+            "precision": self.rainfusion.precision,
```
david6666666, line 486: Ran tests/e2e/accuracy/minimax_h3/test_minimax_h3_i2va_ref2va_similarity.py on 4x B300 with seed 0, 50 steps, USP4/HSDP4, and FLASH_ATTN: both I2VA and Ref2VA passed the SSIM/PSNR gates. The I2VA output is byte-identical to base (SSIM 0.972621, PSNR 35.074173 dB). Since this precision argument is only used by _forward_sparse_npu, no GPU accuracy regression was observed.

david6666666 (approved):

PR #6038, opened 2026-08-11
[BugFix][TTS] Stop a typo'd wait timeout from disabling the deadline

linyueqian:
## Purpose

Part R1.4 of #4855.

`VLLM_OMNI_INPUT_WAIT_TIMEOUT_S` treats any value `<= 0` as "disabled". That makes `-600` — a plausible typo for `600` — silently turn the safety net off while looking like a perfectly ordinary setting. Nothing in the logs says it happened.

This matters more than it did when the RFC row was written. #6033 adds the async-chunk deadline and reads **the same constant**, so one bad value now disables the deadline on *both* transports rather than just the full-payload one.

## What changed

`vllm_omni/core/sched/omni_scheduler_mixin.py`, one block at import time:

- **Negative → fall back to 600s, with a warning** that says to use exactly `0` if disabling was the intent. There is no sensible reading of a negative deadline, so treating it as "off" only ever hides a mistake.
- **Zero → still a supported opt-out**, but it now logs *which* nets it just turned off, naming both `WAITING_FOR_INPUT` and `WAITING_FOR_CHUNK`. The RFC offered "reject `<=0`" or "document `0` as an explicit opt-in"; rejecting it would break anyone deliberately running without a deadline, so this takes the second option and makes it loud.
- Unparseable values keep their existing fallback; the literal `600.0` is now a named constant instead of being repeated.

No behaviour change for any valid positive value, and no new env var or flag.

## Test plan

```bash
pytest tests/core/sched/test_omni_scheduler_mixin_timeouts.py
```

**18 passed** (9 new). The constant is parsed at import, so each new test reloads the module under a different environment:

| case | expected |
|---|---|
| unset / `30` / `0.5` | parsed as-is |
| `0` | stays `0.0`, warning names both transports |
| `-1` / `-600` / `-0.5` | falls back to `600.0` |
| `not-a-number` | falls back to `600.0` |

Regression on `tests/core/sched/` + `tests/distributed/omni_connectors/`: **337 passed, 3 failed** — the three are `test_bagel_*_connector`, which fail identically on current `main` (run side by side in the same venv).

## Note

Built on the R1.1 branch, so this shows #6033's diff too until that lands. The R1.4 change is the single commit `97045f50`.

- **vLLM version:** 0.26.0


linyueqian:
Folding this into #6033 — R1.4 is one of four R-series items there now, and keeping it separate meant this PR's diff also carried R1.1's, which read as a duplicate.

The change itself survived, with one correction from review: negative and non-finite values are now **rejected at startup** rather than falling back to 600 s. `-1` is a common "no limit" idiom, so quietly substituting a ten-minute deadline would give an operator who meant "never time out" mysterious failures. `nan` and `inf` also slipped past the original check.

Closing in favour of #6033.

PR #6042, opened 2026-08-11
[CI][MiniCPM-o] Native duplex Seed-TTS PERF with TPOT and global session metrics

R2-Y:

## Purpose
Upgrade MiniCPM-o 4.5 duplex Seed-TTS performance CI from explicit Realtime TTS to paper/demo-aligned native chunked PCM streaming, while keeping Stage-0 TPOT and adding demo-consistent global session metrics.
- Duplex PERF now uses `openai-realtime-duplex` with paced PCM append + commit.
- Chunk size is configurable (`--realtime-duplex-chunk-ms`, default `1000`; `200` supported).
- Stage-0 TPOT is recorded per turn / session and required by duplex CI.
- Keep `openai-realtime-tts` for Seed-TTS WER accuracy only.
- Add session-global TTFT / TTFP / RTF, shared with `realtime_duplex_demo.py`.
- Preserve existing per-turn metrics and their means.

## Test Plan

**vLLM Version:**
0.26.0

**vLLM-Omni Commit:**
0.26.0 + latest main

Functional
```
export CUDA_VISIBLE_DEVICES=4
pytest -s -v tests/e2e/online_serving/test_minicpmo_4_5_duplex.py \
  -m '(core_model or advanced_model) and cuda and H100 and omni' \
  --run-level advanced_model
```
Perf
```
pytest -s -v tests/dfx/perf/scripts/run_benchmark.py \
  --test-config-file /data/yrr/rein_test/test_minicpmo_4_5_duplex_seed_tts.local.json \
  --run-level full_model -m 'full_model and omni'
```
## Test Result
perf 
```
completed: 1
failed: 0
request_throughput: 0.02157090655608359
mean_ttft_ms: 369.499
mean_tpot_ms: 19.079
mean_e2el_ms: 46335.60341503471
mean_audio_ttfp_ms: 369.365
mean_audio_rtf: 1.342597
mean_duplex_global_ttft_ms: 356.317
mean_duplex_global_ttfp_ms: 356.211
mean_duplex_global_rtf: 1.577829
duplex_session_metrics[0]:
  session_id: seed-tts-native-bench-432b01af-0
  audio_turn_count: 4
  mean_ttft_ms: 369.499
  mean_ttfp_ms: 369.365
  mean_rtf: 1.342597
  mean_tpot_ms: 19.079
  input_chunk_ms: 200
  global_ttft_ms: 356.317
  global_ttfp_ms: 356.211
  global_rtf: 1.577829
  global_audio_generation_ms: 42916.947
  global_audio_duration_ms: 27200.0
  global_measurement_origin: {'ttft': 'first input_audio_buffer.commit client send to first non-empty text delta', 'ttfp': 'first input_audio_buffer.commit client send to first audio packet', 'rtf': 'first commit-to-last-audio receive time divided by total emitted audio duration'}
session_global_and_tpot_ok: true
duplex_request_metrics_count: 4
first_turn_tpot_ms: 17.453858348102692
first_turn_ttft_ms: 356.317
first_turn_ttfp_ms: 356.211
```


amy-why-3459:
Thanks — splitting openai-realtime-tts (WER) from openai-realtime-duplex
(native PCM) is the right cut, and the client-side TPOT / global-session
summaries are structured correctly. Request changes on the CI gate.

1. temperature: 0.0 in the PERF JSON never reaches configure().
   Native duplex reads extra_body["temperature"]; preprocess_serve_args
   only copies chunk_ms / pacing. The unit test passes because it sets
   extra_body.temperature; Nightly will configure(temperature=None) and
   fall back to the server default. Please forward --temperature into
   extra_body for this backend (and add a CLI test).

2. The new baseline asserts are not a gate.
   - float(result.get(metric) or 0.0) treats a missing metric as 0, so
     a latency/RTF ceiling passes when the field is absent.
   - Thresholds are 2–50× the numbers in the PR body (global TTFT
     19000 vs 356; TTFT/TTFP 6500 vs 369). Only mean_tpot_ms (22 vs
     19) is in range. Please rebuild from this successful run with
     ~20–30% slack and fail if a gated metric is missing.
   - request_throughput was dropped (0.159 → ~0.022 is expected after
     the protocol change). Put a new floor back or say why throughput
     is no longer gated.
   - Mark lists A3; baseline only has H100, so NPU runs skip the block.

3. This CI is not paper-style 1.0s native duplex.
   chunk_ms=200 is the demo cadence; max_input_ms=1400 is then padded
   to chunk_period_ms (usually 1000) → 2000 ms effective audio / turn.
   Turns are still serial (stream → commit → response.done → ack).
   Please say that in the JSON/PR, or add a separate chunk_ms=1000
   config if you want paper-aligned numbers.

Should-fix: assert the text_latency reconstruction
(ttft + mean_tpot * (output_tokens-1), itl empty) so calculate_metrics
cannot silently start using E2EL; and don't treat global_rtf as model
RTF — it includes inter-turn paced input + ack.

Happy to re-review once the gate matches the run you already posted.

Sy0307:
The hardware-baseline loop still coerces a missing metric to zero with `float(result.get(metric_name) or 0.0)`. For an upper-bound latency or RTF baseline, a missing metric therefore passes as 0 instead of failing closed. Please distinguish missing/non-finite metrics from a real zero and add a missing-metric regression.

vllm_omni/benchmarks/patch/patch.py
```diff
@@ -1735,19 +1759,293 @@ async def async_request_openai_realtime_duplex(
...
+
+
+async def async_request_openai_realtime_duplex(
+    request_func_input: RequestFuncInput,
+    session: aiohttp.ClientSession,
+    pbar: tqdm | None = None,
+) -> MixRequestFuncOutput:
+    """Stream Seed-TTS prompt WAVs through one native Realtime duplex session."""
+    del session
+    output = MixRequestFuncOutput()
```
lishunyang12, line 1832: Keep the OmniInteract dispatch in the native duplex backend.

vllm_omni/benchmarks/patch/patch.py
```diff
@@ -1735,19 +1759,293 @@ async def async_request_openai_realtime_duplex(
...
+                    wait_for_history_committed=True,
+                )
+
+            request_finished_at = time.perf_counter()
+            session_metrics = summarize_session_request_metrics(turn_metrics, session_id=session_id)
+            session_metrics["input_chunk_ms"] = chunk_ms
+            if first_commit_at_s is None:
+                raise RuntimeError("Seed-TTS native duplex session did not commit input audio")
+            global_metrics = client.events.global_timing_summary(
+                after_s=first_commit_at_s,
```
lishunyang12, line 2002: Start timing at the first append so pre-commit responses aren't discarded.

PR #6044, opened 2026-08-11
[Bugfix][MiniCPM-o] Preserve reference audio through Talker and Code2Wav

Kelsey050218:
## Purpose

MiniCPM-o 4.5 Seed-TTS requests carry reference audio for voice cloning, but vLLM-Omni did not preserve that audio through the multistage request path. The audio was either exposed as raw Talker multimodal input, although the Talker model does not consume it, or dropped before the first Code2Wav payload.

The request could still finish and return audio, but Code2Wav could fall back to the checkpoint's default prompt voice instead of conditioning on the request's reference speaker.

### What broke

The expected path is:

```text
request audio
  -> request-side input
  -> Thinker-to-Talker input bridge
  -> model_intermediate_buffer
  -> first Code2Wav payload
```

Before this fix:

- `requires_multimodal_data` represented both media consumed by a stage model and original request media needed only by an input bridge;
- enabling it exposed raw audio to Talker processing, while disabling it prevented the bridge-only audio from being materialized;
- the full-payload runner stored the handoff in `additional_information_cpu`, but the Code2Wav bridge only read the streaming/legacy locations;
- non-audio handoff metadata, including the reference waveform and sample rate, could be discarded when the next stage extracted acoustic codes.

### Root cause

Model multimodal input and request-scoped bridge input are different contracts. Reusing one flag for both made it impossible to say: "materialize this audio for the stage input bridge, but do not pass it to the Talker model."

The streaming and full-payload paths also expose their handoff state through different request attributes, so fixing only one path still loses reference conditioning on the other.

### Fix

- add `request_side_input_modalities` to the stage pipeline/runtime contract;
- materialize declared side inputs once without forwarding raw `multi_modal_data` to the Talker model;
- declare MiniCPM-o Talker audio as a bridge-side input;
- serialize the Seed-TTS reference audio into the Chat Completions request and preserve `ref_audio` / `ref_text` request fields;
- write the reference waveform and sample rate into `model_intermediate_buffer`;
- preserve non-audio metadata while extracting Talker acoustic codes;
- read streaming, full-payload CPU-side, and legacy handoff buffers in priority order;
- attach the reference waveform and sample rate to the first Code2Wav payload on both streaming and full-payload paths.

The benchmark adapter change only transports the existing reference audio through the API request. This PR does **not** replace prompt features or change speaker-feature extraction, acoustic-token sampling, or Code2Wav conditioning math.

### Out of scope

StepAudio2/FlashCosyVoice package-layout compatibility is independent of request-side routing and is intentionally excluded from this PR.

## Test Plan

```bash
PYTHONPATH=/tmp/vllm-omni-test-stubs:$PWD python -m pytest -q \
  tests/benchmarks/patch/test_patch.py \
  tests/engine/test_stage_engine_core_proc.py \
  tests/test_config_factory.py \
  tests/model_executor/models/minicpmo_4_5/test_llm2tts.py \
  tests/entrypoints/openai_api/test_serving_chat_multistage_generation.py
```

**vLLM Version:** `0.1.dev1+ge5588e49b`

**vLLM-Omni base:** `minicpm-challenge@009b80d686febcf683fdbc2bcdf3ad752884641e`

## Test Result

- Unit/regression suite: **517 passed, 2 skipped, 18 warnings** in 13.86 s.
- Diff-scoped pre-commit passed. `actionlint` was skipped because no workflow files changed and the host Go 1.17 cannot install the hook's Go 1.19+ dependencies.

### Fixed 32-request Seed-TTS comparison

This is a regression-focused before/after comparison, not a claim about an external aggregate baseline.

Conditions held constant:

- hardware: one Ascend 910B1 for all three serving stages;
- model: `openbmb/MiniCPM-o-4_5`;
- runtime image: `openbmb/minicpm-challenge-tts-seed:acba6622-wer`;
- dataset: Seed-TTS English, first 32 loader rows in fixed order;
- selection: `--num-prompts 32 --disable-shuffle --seed 42 --no-oversample`;
- serving: one turn per session, 3 warmups, max concurrency 1, request rate `inf`;
- request: text + inline reference audio, TTS template enabled, thinking disabled;
- scorer: BytedanceSpeech/seed-tts-eval `cal_sim`, fine-tuned WavLM-large, CPU;
- the before and after artifacts contain the same 32 utterance IDs in the same order.

| Metric | Before routing fix | After routing fix |
|---|---:|---:|
| Mean ASV similarity | 0.0710411 | **0.5445324** |
| Evaluated / failed | 32 / 0 | **32 / 0** |
| Generated audio duration | 142.28 s | **142.28 s** |
| Samples improved | — | **32 / 32** |

The mean ASV increase is **+0.4734912**, or **7.66x** the pre-fix value. Every one of the 32 matched samples improved; per-sample deltas ranged from +0.1745862 to +0.7539873.

The captured E2E routing head was `53b6c51da1b5213e1bf4eca63124da2e43710af1`. That test image requires an orthogonal import-only compatibility patch for its packaged FlashCosyVoice layout so Code2Wav can start. The compatibility patch does not touch request payloads, prompt features, acoustic tokens, generated outputs, or scoring, and has been removed from this PR to keep the bugfix focused. The preserved pre-fix artifact also reported a dirty source checkout whose diff was not retained, so this table is presented as matched regression evidence rather than a clean upstream performance benchmark.


Kelsey050218:
Superseded by #6052 after renaming the fork branch from `agent/minicpmo-request-side-input-routing` to `bugfix/minicpmo-request-side-input-routing`. The commit SHAs and patch contents are unchanged.

PR #6050, opened 2026-08-11
[Config] Reuse vLLM configs in VllmOmniConfig

Acerak01-fy:
This follows RFC #4021 . It reduces duplicated configuration definitions in the structured AR and generation stage configs by reusing upstream vLLM config contracts where the ownership and lifecycle boundaries match.oo

The existing `VllmOmniConfig` interface remains unchanged. Callers continue to use paths such as `stage.load_config`, `stage.cache_config`, `stage.model_config.compilation_config`, and `stage.runtime_config.profiler_config`. This PR changes how those values are defined and validated, not where callers read or write them.

## Purpose

The structured configs introduced by RFC #4021 originally repeated a subset of fields that already belong to upstream vLLM configs. Keeping independent copies of those declarations means their types, defaults, validation, and newly added upstream fields can drift over time.

This PR uses two forms of upstream reuse:

1. Inherit the complete upstream config class when the Omni sub-config and the upstream config represent the same concern.
2. Accept the exact upstream config object when Omni already exposes that config as a nested value.

Legacy mapping inputs remain supported. Prebuilt upstream config objects are transported without converting them back into dictionaries, while final `VllmConfig` materialization remains in the engine-owning process.

## Reuse Criteria

The selection criterion is not merely **"the field names match."** A config is reused only when all of the following conditions are satisfied:

| Criterion | Required condition | Why it matters |
| --- | --- | --- |
| Semantic equivalence | The Omni and upstream fields have the same meaning, accepted values, and backend effect. | Same-name fields can still represent different lifecycle stages or different products. |
| Ownership alignment | The fields belong to the same config concern, such as loading, cache, scheduling, or parallelism. | Whole-class reuse is stable only when the class boundaries also match. |
| Lifecycle safety | Constructing the structured config in the head process must not load HF metadata, resolve devices/ranks, probe ports, or otherwise perform terminal initialization. | Structured stage configs are transport objects; the owning engine process performs final materialization. |
| Public API stability | Existing Omni field paths and constructor inputs must remain valid. | RFC #4021 consumers should not require a coordinated migration for this refactor. |
| Projection compatibility | Reuse must preserve the current flat `EngineArgs` projection and must not leak newly inherited upstream defaults into backend arguments. | Inheritance should not silently change effective engine behavior. |

Upstream `EngineArgs` is used as a schema-audit index, not as the sole reason to inherit a config class. For the current AR/generation stage ownership surface, 43 stage-owned engine-input names are present in upstream `EngineArgs` and 20 remain explicitly Omni-only. A matching name identifies a reuse candidate; the criteria above determine whether reuse is safe.

### Whole-config reuse

| Existing Omni boundary | Upstream contract | Reuse form | Basis for reuse | Omni compatibility retained |
| --- | --- | --- | --- | --- |
| `OmniStageLoadConfig` | `vllm.config.LoadConfig` | Class inheritance | Both own checkpoint loading behavior, including `download_dir`, `load_format`, and upstream load validation. | Omni tokenizer-related stage inputs remain on the same class. Only the existing load projection allowlist is emitted to backend args. |
| `OmniStageCacheConfig` | `vllm.config.CacheConfig` | Class inheritance | Both own KV-cache sizing and memory behavior, including `kv_cache_memory_bytes`, `gpu_memory_utilization`, and prefix caching. | Existing nullable Omni defaults remain deferred to the backend. Omni compatibility fields such as `disable_hybrid_kv_cache_manager` and `mm_processor_cache_gb` retain their current paths. |
| `OmniStageSchedulerConfig` | `vllm.config.SchedulerConfig` | Class inheritance | Both own batching and request-scheduling behavior, including sequence/token limits and chunked prefill. | `max_model_len` remains an unresolved stage input. The Omni post-init keeps transport-safe validation without running terminal scheduler initialization. |
| `OmniStageParallelConfig` | `vllm.config.ParallelConfig` | Class inheritance | Pipeline, data, tensor, and expert-parallel dimensions have the same upstream ownership and backend meaning. | Omni overrides post-init so rank/backend/port/platform resolution stays in the owning process, and preserves the existing `world_size` semantics that include DP. |
| `OmniStageDiffusionParallelConfig` | `vllm.config.ParallelConfig` through `OmniStageParallelConfig` | Indirect class inheritance | Common distributed dimensions are shared with AR/generation stages. | Diffusion-only axes (`ulysses_degree`, `ring_degree`, CFG, VAE, HSDP, and related validation) remain defined by Omni. |

Inherited upstream classes can add fields in future vLLM versions, but those fields are not automatically projected into engine arguments. Explicit load, cache, scheduler, and parallel projection sets preserve the current behavioral surface until a new field is deliberately reviewed and adopted.

### Exact upstream object reuse

| Existing Omni field path | Accepted upstream type | Existing input retained | Basis for reuse |
| --- | --- | --- | --- |
| `model_config.compilation_config` | `vllm.config.CompilationConfig` | `Mapping[str, Any]` and `None` | Compilation is already represented as one nested terminal config. Accepting the upstream object reuses its complete schema without making `OmniStageModelConfig` inherit the terminal upstream `ModelConfig`. |
| `runtime_config.profiler_config` | `vllm.config.ProfilerConfig` | `Mapping[str, Any]` and `None` | Profiler settings form a complete upstream value object and do not require a separate Omni duplicate. |
| `stage.quantization_config` | `vllm.config.quantization.QuantizationConfigArgs` | Existing mapping, string, backend `QuantizationConfig`, and `None` forms | This matches the user-facing quantization object accepted by upstream `EngineArgs` while preserving existing Omni quantization inputs. |

`Mapping` appears first in these unions intentionally. Pydantic therefore keeps existing YAML/dict inputs as mappings instead of eagerly converting them into upstream config objects in the head process. If a caller already provides an upstream config object, its type is preserved through structured stage construction and typed engine-argument projection.

## Intentionally Not Reused

| Candidate | Decision | Reason |
| --- | --- | --- |
| `OmniStageModelConfig` -> upstream `ModelConfig` | Do not inherit in this PR. | Upstream `ModelConfig` is a terminal model config that reads HF metadata and resolves model capabilities. The Omni class stores unresolved per-stage inputs and Omni-only values such as `model_arch`, model/tokenizer subdirectories, streaming fields, and stage sampling data. Inheritance would move initialization into the head process or require overriding most of the upstream lifecycle. |
| `BaseVllmOmniStageConfig` -> upstream `VllmConfig` | Do not inherit. | `VllmConfig` aggregates fully materialized model, cache, scheduler, parallel, compilation, and profiler configs and performs cross-config initialization. A structured stage config is a transport representation, not that terminal aggregate. |
| `OmniStageRuntimeConfig` -> upstream `ParallelConfig`/`VllmConfig` | Keep the current Omni boundary. | Runtime combines Omni process placement (`devices`, `env`, and replicas) with backend runtime inputs. There is no one-to-one upstream class, and moving individual fields would change established paths. The nested `ProfilerConfig` can still be reused independently. |
| `OmniStageConnectorConfig` | Keep Omni-owned. | Stage connectors and `omni_kv_config` describe Omni pipeline topology and have no equivalent upstream config boundary. |
| New `AttentionConfig`, `KernelConfig`, or `MultiModalConfig` subtrees | Defer to a separate migration. | Upstream types exist, but the current OmniConfig API does not expose matching sub-config paths. Adding them here would require consumer changes rather than a reuse-only refactor. |

## Compatibility Boundaries

| Boundary | Preserved behavior |
| --- | --- |
| Public config API | No existing OmniConfig field or top-level path is renamed or moved. |
| YAML and dictionary inputs | Existing mappings remain mappings and continue through the current construction path. |
| Upstream object inputs | `CompilationConfig`, `ProfilerConfig`, and `QuantizationConfigArgs` retain their types and are deep-copied across the projection boundary. |
| Initialization ownership | Terminal config construction and platform/process resolution remain in the engine-owning process. |
| Backend defaults | Newly inherited upstream fields are excluded unless they belong to an explicit existing projection set. |
| Existing engine behavior | AR, generation, and diffusion effective backend arguments continue to match the pre-refactor structured projection. |

This also updates the related unit tests to cover:

- Upstream class inheritance and inherited field-schema tracking for load, cache, scheduler, and parallel configs.
- Existing mapping inputs remaining dictionaries rather than being eagerly materialized as upstream config objects.
- Prebuilt `CompilationConfig`, `ProfilerConfig`, and `QuantizationConfigArgs` values retaining their types.
- Upstream config objects surviving typed engine-argument projection for both AR and generation stages.
- Every current AR/generation owned engine field remaining either an upstream `EngineArgs` field or an explicitly listed Omni-only field.
- Newly inherited upstream defaults not leaking into projected backend arguments.
- Existing config-factory, registered-pipeline, legacy-parity, and diffusion behavior remaining unchanged.

## Test Plan

vLLM Version: 0.25.0

vLLM-Omni Base Commit: ae811043

vLLM-Omni PR Commit: ab5c9f05

### Test Result

Static checks:

```bash
../.venv-vllm-omni-main/bin/python -m ruff check \
  tests/engine/test_stage_engine_args.py \
  vllm_omni/config/omni_config.py \
  vllm_omni/engine/stage_init_utils.py

../.venv-vllm-omni-main/bin/python -m ruff format --check \
  tests/engine/test_stage_engine_args.py \
  vllm_omni/config/omni_config.py \
  vllm_omni/engine/stage_init_utils.py
```

Result:

```text
All checks passed!
4 files already formatted
```

Unit tests:

```bash
../.venv-vllm-omni-main/bin/python -m pytest -q \
  tests/config/test_config_factory.py \
  tests/engine/test_stage_engine_args.py
```

Result:

```text
464 passed, 2 skipped, 21 warnings in 5.88s
```

Pre-commit checks:

```bash
pre-commit run --files \
  vllm_omni/config/omni_config.py \
  vllm_omni/engine/stage_init_utils.py \
  tests/engine/test_stage_engine_args.py
```

Result:

```text
All applicable hooks passed.
```


hsliuustc0106:
please update the design under #4021

vllm_omni/config/omni_config.py
```diff
@@ -346,20 +357,18 @@ class OmniStageModelConfig:
 
 
 @config
-class OmniStageLoadConfig:
-    """Per-stage loading behavior and resolved tokenizer input."""
+class OmniStageLoadConfig(VllmLoadConfig):
```
hsliuustc0106: Direct inheritance changes the positional signatures of these exported dataclasses because upstream fields are prepended and `@config` is not keyword-only. For example, `OmniStageLoadConfig("/tok")` now binds `"/tok"` to `load_format` instead of `tokenizer`; the cache and parallel classes similarly remap their first parameters, with the latter silently changing DP/TP topology. Please preserve the legacy positional contract (or introduce an explicit keyword-only/deprecation migration) and add positional-construction coverage.

vllm_omni/config/omni_config.py
```diff
@@ -375,16 +384,19 @@ class OmniStageCacheConfig:
...
+    # Omni retains it as unresolved stage input until the owning engine process.
     max_num_seqs: int | None = Field(default=None, ge=1)
     max_num_batched_tokens: int | None = Field(default=None, ge=1)
     max_model_len: int | None = Field(default=None, ge=-1)
+    is_encoder_decoder: InitVar[bool] = False
     enable_chunked_prefill: bool | None = None
     async_scheduling: bool | None = None
 
-    def __post_init__(self) -> None:
+    def __post_init__(self, is_encoder_decoder: bool = False) -> None:
```
hsliuustc0106, line 452: `SchedulerConfig.__post_init__` is the only initializer for inherited `max_num_encoder_input_tokens` and `encoder_cache_size` (`init=False`, with no defaults). This override skips those assignments, so `_serialize_stage_config()` reaches `dataclasses.asdict()` with unset fields and structured stage transport can raise `AttributeError`. Please initialize them with transport-safe values here, without terminal validation, and add an `asdict`/registration test.

vllm_omni/config/omni_config.py
```diff
@@ -664,15 +677,19 @@ def __post_init__(self) -> None:
 
         self._propagate_quantization_from_tf_config(self.tf_model_config)
         if self.quantization_config is not None:
-            if isinstance(self.quantization_config, QuantizationConfig):
+            if isinstance(
+                self.quantization_config,
+                (QuantizationConfig, QuantizationConfigArgs),
```
hsliuustc0106: `QuantizationConfigArgs` is preserved here for diffusion, but `_DiffusionConfigProjection.enrich_config()` constructs `OmniDiffusionConfig`, whose terminal validation accepts only diffusion `QuantizationConfig`, `str`, or `Mapping`; the future typed startup path has the same mismatch. Please either reject this type for diffusion during ownership validation or normalize it to a supported diffusion representation, and add a terminal diffusion test.

hsliuustc0106 (commented):
Inline findings on constructor compatibility and structured-config transport/materialization.

Acerak01-fy:
> please update the design under #4021

Done.https://github.com/vllm-project/vllm-omni/issues/4021#issuecomment-5261131242

vllm_omni/config/omni_config.py
```diff
@@ -70,11 +79,78 @@
     }
 )
 
-_QuantizationConfigType: TypeAlias = QuantizationConfig | str | Mapping[str, Any] | None
+# Keep Mapping first so Pydantic preserves existing YAML dictionaries instead
+# of eagerly materializing terminal configs in the head process.
+_CompilationConfigType: TypeAlias = Mapping[str, Any] | VllmCompilationConfig | None
+_ProfilerConfigType: TypeAlias = Mapping[str, Any] | VllmProfilerConfig | None
+_LLMQuantizationConfigType: TypeAlias = Mapping[str, Any] | QuantizationConfigArgs | QuantizationConfig | str | None
+_DiffusionQuantizationConfigType: TypeAlias = Mapping[str, Any] | QuantizationConfig | str | None
```
hsliuustc0106: is there any different btw LLM/Diffusion in quantization?

do we have LLM/Diffusion parallel/profiler config independently?

docs/design/module/vllm_omni_config.md
```diff
@@ -18,8 +18,8 @@ validation_paths:
   - tests/config/**
 upstream_refs:
   - vllm.config
-last_reviewed: 2026-08-07
-last_verified_commit: 3d7fc3b9ba3cac88d579d4dc35b78b0b641675fc
+last_reviewed: 2026-08-11
+last_verified_commit: e356708da4cd39992405686dacfc32b18fcccc7f
 ---
 
 # vLLM-Omni configuration
```
hsliuustc0106, line 25: I think we need to update the module design in a follow-up PR

vllm_omni/config/omni_config.py
Acerak01-fy: > is there any different btw LLM/Diffusion in quantization?
> 
> do we have LLM/Diffusion parallel/profiler config independently?

Yes, quantization has different terminal schemas for LLM and diffusion.

  For AR/generation, `QuantizationConfigArgs` is vLLM's user-facing online quantization schema (`linear`, `moe`, and `ignore`) and is resolved by the vLLM engine config flow.

  Diffusion materializes into `OmniDiffusionConfig`, whose quantization schema is `str | Mapping | QuantizationConfig | None`. Its mappings use  `method`/`quant_method` and may contain per-component entries such as `transformer` and `vae`. There is no lossless conversion from `QuantizationConfigArgs`, so the types are separated and diffusion rejects that LLM-only object explicitly.

  Parallel config is also stage-type-specific:
  `OmniStageParallelConfig` reuses vLLM `ParallelConfig`, while `OmniStageDiffusionParallelConfig` adds diffusion-only Ulysses/Ring/CFG/VAE/HSDP  dimensions and is later materialized as `DiffusionParallelConfig`.

  Profiler config is intentionally shared because both terminal paths consume the same upstream vLLM `ProfilerConfig` contract.

docs/design/module/vllm_omni_config.md
Acerak01-fy: OK! I'll do it.

hsliuustc0106:
fix docs

Acerak01-fy:
> fix docs

fixed.

hsliuustc0106:
ci has passed https://buildkite.com/vllm/vllm-omni/builds/13530/list @alex-jw-brooks @NickCao PTAL

docs/design/module/vllm_omni_config.md
```diff
@@ -18,8 +18,8 @@ validation_paths:
   - tests/config/**
 upstream_refs:
   - vllm.config
-last_reviewed: 2026-08-07
-last_verified_commit: 3d7fc3b9ba3cac88d579d4dc35b78b0b641675fc
+last_reviewed: 2026-08-11
+last_verified_commit: e356708da4cd39992405686dacfc32b18fcccc7f
```
alex-jw-brooks, line 22: Nit: this will be squashed when we merge this PR.

I wonder if instead of last verified commit we should have `last_modified by PR:` or something similar, so that they don't point to commits off of `main`? @Acerak01-fy @hsliuustc0106

hsliuustc0106: will refresh all of the module design md in v0.28.0 stable release

vllm_omni/config/omni_config.py
alex-jw-brooks: In my opinion, the quantization config handling is actually mostly tech debt. I don't think cleanup should be in this PR since the problem is more messy handling of quant config throughout the diffusion code, but I think the main case that is different is things like the component config. Since quantization is primarily applied at the layer level, it should really not be engine specific

vllm_omni/config/omni_config.py
```diff
@@ -70,11 +79,78 @@
     }
 )
 
-_QuantizationConfigType: TypeAlias = QuantizationConfig | str | Mapping[str, Any] | None
+# Keep Mapping first so Pydantic preserves existing YAML dictionaries instead
+# of eagerly materializing terminal configs in the head process.
+_CompilationConfigType: TypeAlias = Mapping[str, Any] | VllmCompilationConfig | None
```
alex-jw-brooks: IMO these mappings are a code smell, and should be minimized. Since the whole intention with Omni config is to add better typing and centralize everything, we should not type alias dicts like this, because it will be easy to run into a lot of the same problems we currently have around quantization config

docs/design/module/vllm_omni_config.md
```diff
@@ -39,11 +39,52 @@ The PR review proposal to capture environment-derived values in configuration
...
+
+The current RFC #4021 implementation uses upstream vLLM config classes as the
+schema source for AR and generation stages:
+
+- `OmniStageLoadConfig` inherits `vllm.config.LoadConfig`.
+- `OmniStageCacheConfig` inherits `vllm.config.CacheConfig`.
+- `OmniStageSchedulerConfig` inherits `vllm.config.SchedulerConfig`.
+- `OmniStageParallelConfig` inherits `vllm.config.ParallelConfig`.
+
+This inheritance preserves the existing `VllmOmniConfig` paths while reducing
```
alex-jw-brooks: I think inheriting for many of these config types does make sense, but is the intention of this to reduce the scope of this PR, or is there another reason?


IMO we should not preserve the existing paths. I.e., `VllmOmniConfig` has not been flipped on in production yet, and is also a largely internal object, so it would be more ideal to just update callers to use the proper interfaces where applicable than to have legacy codepaths before we have even migrated

vllm_omni/engine/stage_init_utils.py
```diff
@@ -781,13 +786,14 @@ def stage_runtime_env(stage_id: int, runtime_cfg: Any) -> Generator[None, None,
...
     *,
+    include: frozenset[str] | None = None,
     exclude: frozenset[str] = frozenset(),
 ) -> dict[str, Any]:
     """Copy defined typed config fields into backend adapter kwargs."""
     projected: dict[str, Any] = {}
     for config_field in fields(config):
         name = config_field.name
-        if name in exclude:
+        if (include is not None and name not in include) or name in exclude:
```
alex-jw-brooks: A bit worried that this part of the design is brittle and that we will silently drop things here

alex-jw-brooks (commented):
Thanks for looking at this! Some thoughts - I think inheritance does make sense for most of these config types, but have a few concerns. More specifically, my main worries are

- I think we should avoid typealiasing `dict` into too many types and just ensure things get converted to the right well-defined types as early as possible, otherwise we'd have to redundantly validate etc throughout the code

- The deprecation handling for kwargs feels premature, and I think we should avoid it 😅 Omni config still isn't active in production, so it would be better to just update the callers and avoid having legacy paths before the code is even active

- Also a bit worried that it will be easy to drop stuff / change behavior silently, since keeping expected kwargs etc in sync with things tied to upstream vLLM is hard to catch

vllm_omni/config/omni_config.py
Acerak01-fy: Agreed. First, regarding the quantization configuration: 
> I removed the QuantizationConfigArgs reuse and the temporary LLM/diffusion quantization split from this PR. Quantization now remains on the existing Omni-owned transport contract (str, Mapping, QuantizationConfig, or None), and this PR no longer treats it as an upstream engine-config reuse boundary. A broader quantization redesign should be handled separately. 

The second mapping issue:
> I removed the mapping-based public aliases for CompilationConfig and ProfilerConfig. They are now concrete structured-stage fields typed as VllmCompilationConfig | None and VllmProfilerConfig | None. YAML/dictionary inputs are validated and materialized into the corresponding upstream types during structured config construction, so downstream consumers only see one resolved representation. Quantization remains the existing exception and is not expanded by this PR.

vllm_omni/engine/stage_init_utils.py
Acerak01-fy: Good point. I replaced the static AR/generation projection allowlists with field maps generated from the intersection of each upstream config dataclass and upstream EngineArgs. Known naming differences such as cache_dtype -> kv_cache_dtype, policy -> scheduling_policy, and data_parallel_master_ip -> data_parallel_address are handled explicitly. The structured configs also record constructor-explicit fields, so inherited upstream defaults are not leaked into engine arguments. If a caller explicitly supplies an inherited field that has no valid projection, projection now raises an error instead of silently dropping it. Diffusion keeps its existing fixed ownership and projection surface.

Acerak01-fy:
> Thanks for looking at this! Some thoughts - I think inheritance does make sense for most of these config types, but have a few concerns. More specifically, my main worries are
> 
> * I think we should avoid typealiasing `dict` into too many types and just ensure things get converted to the right well-defined types as early as possible, otherwise we'd have to redundantly validate etc throughout the code
> * The deprecation handling for kwargs feels premature, and I think we should avoid it 😅 Omni config still isn't active in production, so it would be better to just update the callers and avoid having legacy paths before the code is even active
> * Also a bit worried that it will be easy to drop stuff / change behavior silently, since keeping expected kwargs etc in sync with things tied to upstream vLLM is hard to catch

Thank you very much for your suggestion. Previously, the reuse of the upstream config design did cause me some headaches. In terms of reuse, parameter validation and initialization, as well as the issue of ownership and code simplicity, it was quite difficult for me to make a balanced decision.

hsliuustc0106:
@DarkLight1337 @ywang96 @ZJY0516 @Isotr0py PTAL whether the reuse of vllm config mechanisim works

vllm_omni/config/omni_config.py
```diff
@@ -353,20 +384,18 @@ class OmniStageModelConfig:
 
 
 @config
-class OmniStageLoadConfig:
-    """Per-stage loading behavior and resolved tokenizer input."""
+class OmniStageLoadConfig(_TrackExplicitConfigFields, VllmLoadConfig):
```
SamitHuang, line 409: Inheriting directly from `VllmLoadConfig` prepends upstream fields to the constructor, changing the positional signature (e.g., `OmniStageLoadConfig("/tok")` now binds to `load_format`). Please enforce keyword-only initialization (e.g., `@config(kw_only=True)`) or restore the positional contract, and add tests for positional construction as requested in the previous review.

vllm_omni/config/omni_config.py
```diff
@@ -432,22 +469,69 @@ class OmniStageRuntimeConfig:
     num_gpus: int = Field(default=1, ge=1)
     log_level: str = "info"
     log_stats: bool = False
-    profiler_config: dict[str, Any] | None = None
 
 
 @config
-class OmniStageParallelConfig:
+class OmniStageParallelConfig(_TrackExplicitConfigFields, VllmParallelConfig):
```
SamitHuang, line 500: This inheritance implicitly reorders `pipeline_parallel_size`, `tensor_parallel_size`, and `data_parallel_size` based on upstream `VllmParallelConfig`'s field order. This can silently change DP/TP topology for callers using positional arguments. This class must be keyword-only or maintain the original positional argument order.

SamitHuang (changes requested):
This PR breaks the positional argument contract for structured config constructors by inheriting upstream dataclasses without `kw_only=True` or an explicit migration. This must be fixed to prevent silent configuration corruption.

DarkLight1337:
It looks reasonable to me. The `kw_only=True` restriction suggested by @SamitHuang also makes sense, we don't expect the configs to be constructed from positionsl arguments anyway.

Acerak01-fy:
> This PR breaks the positional argument contract for structured config constructors by inheriting upstream dataclasses without `kw_only=True` or an explicit migration. This must be fixed to prevent silent configuration corruption.

Thanks for catching this. I updated all structured stage config classes that inherit an upstream vLLM dataclass to use keyword-only initialization: 
> OmniStageLoadConfig, OmniStageCacheConfig, OmniStageSchedulerConfig, OmniStageParallelConfig, and OmniStageDiffusionParallelConfig.

Acerak01-fy:
> It looks reasonable to me. The `kw_only=True` restriction suggested by @SamitHuang also makes sense, we don't expect the configs to be constructed from positionsl arguments anyway.

yeah, I fixed it.

hsliuustc0106:
```

FAILED tests/engine/test_stage_engine_args.py::test_typed_llm_projection_discovers_explicit_upstream_config_fields - pydantic_core._pydantic_core.ValidationError: 1 validation error for OmniStageSchedulerConfig
--
  | [2026-08-15T13:33:49Z] max_num_partial_prefills
  | [2026-08-15T13:33:49Z]   Unexpected keyword argument [type=unexpected_keyword_argument, input_value=2, input_type=int]
  | [2026-08-15T13:33:49Z]     For further information visit https://errors.pydantic.dev/2.13/v/unexpected_keyword_argument
  | [2026-08-15T13:33:49Z] = 1 failed, 1893 passed, 3 skipped, 72 deselected, 18 warnings in 228.69s (0:03:48) =
  | [2026-08-15T13:33:56Z] 🚨 Error: The command exited with status 1


```

Acerak01-fy:
> ```
> 
> FAILED tests/engine/test_stage_engine_args.py::test_typed_llm_projection_discovers_explicit_upstream_config_fields - pydantic_core._pydantic_core.ValidationError: 1 validation error for OmniStageSchedulerConfig
> --
>   | [2026-08-15T13:33:49Z] max_num_partial_prefills
>   | [2026-08-15T13:33:49Z]   Unexpected keyword argument [type=unexpected_keyword_argument, input_value=2, input_type=int]
>   | [2026-08-15T13:33:49Z]     For further information visit https://errors.pydantic.dev/2.13/v/unexpected_keyword_argument
>   | [2026-08-15T13:33:49Z] = 1 failed, 1893 passed, 3 skipped, 72 deselected, 18 warnings in 228.69s (0:03:48) =
>   | [2026-08-15T13:33:56Z] 🚨 Error: The command exited with status 1
> ```

fixed.

hsliuustc0106 (approved):

PR #5934, opened 2026-08-09
[CI]drop DreamZero upstream parity e2e; keep OpenPI serving contract

yenuo26:
PLEASE FILL IN THE PR DESCRIPTION HERE.

## Purpose

Weekly DreamZero OpenPI e2e was failing in CI ([#5749](https://github.com/vllm-project/vllm-omni/issues/5749)):

- accuracy parity (`test_dreamzero_openpi_upstream_parity`) hard-failed without `DREAMZERO_REPO` / upstream checkout
- online serving (`test_dreamzero_openpi_online`) hard-failed without PyPI `openpi-client`

Running the **upstream numerical parity** path is not practical in the vLLM-Omni CI / shared test env:

- official DreamZero upstream install (`pip install -e .` / its dependency tree) conflicts with the current vLLM stack
- PyPI `openpi-client` pins `numpy<2`, which is incompatible with the rest of the workspace
- provisioning a separate upstream server + checkpoint tree for weekly jobs is not maintainable as a hard CI gate

Because that environment cannot support a reliable upstream numerical comparison, this PR **deletes** the accuracy parity test and the upstream mock server helpers instead of skipping or papering over missing env vars.

Coverage that remains is the online OpenPI serving **contract** test only:

- keep `tests/e2e/online_serving/test_dreamzero_expansion.py`
- assert fixed metadata / action tensor shape+dtype+finiteness / reset status
- provide a local `openpi_client.msgpack_numpy` stand-in (reusing vLLM-Omni pack/unpack) so CI does not need PyPI `openpi-client`
- move DreamZero-only observation helpers out of `tests/helpers/runtime.py` into the single online test that uses them

Fixes #5749

## Test Plan

**vLLM Version:** N/A (test-only)

**vLLM-Omni Commit:** tip of this PR

```bash
pytest -sv tests/e2e/online_serving/test_dreamzero_expansion.py --run-level full_model
```

## Test Result
f79b6bfd-fdb0-4712-aeb1-15abbd2ec21d

**BEFORE SUBMITTING:** read [CONTRIBUTING.md](https://github.com/vllm-project/vllm-omni/blob/main/CONTRIBUTING.md) and run the [precheck-pr skill](https://github.com/vllm-project/vllm-omni/blob/main/.claude/skills/precheck-pr/SKILL.md) with the code agent for a self-check against project conventions.

Issue #4506, opened 2026-06-17
[Bug]: OpenPI robot serving broken with real openpi-client — numpy msgpack wire-format mismatch


loveysuby:
### Your current environment

<details>
<summary>The output of <code>python collect_env.py</code></summary>

```text
This is an environment-independent serialization (msgpack wire-format) bug in the vllm-omni entrypoint layer. It reproduces deterministically on CPU with no GPU and no model load, so a full collect_env.py dump from the affected GPU host is not meaningful here. The only relevant component is the client library:

  openpi-client : 0.1.2   (pip install openpi-client==0.1.2)
  numpy         : 1.26.4  (pinned by openpi-client; behaviour is np-version independent)
  msgspec       : 0.21.1

The bug is in `vllm_omni/entrypoints/openpi/connection.py` and does not depend on
torch/CUDA/GPU, vLLM version, or the served model.
```

</details>

### Your code version

<details>
<summary>The commit id or version of vllm</summary>

```text
Not relevant — the bug is entirely in the vllm-omni entrypoint (msgpack wire
format) and is independent of the bundled vLLM version.
```
</details>
<details>
<summary>The commit id or version of vllm-omni</summary>

```text
Reproduces on current main: 64a5bb92de727e8c6af60585affa3bdbab6a38c8
Introduced by:              e2e79177db29f55a5e6e94493b518f5fb5d28a69  (#4282)
```
</details>

## Summary
I think this bug is need to compare with #4282 
The OpenPI robot serving path (`/v1/realtime/robot/openpi`) is broken for real
`openpi-client` usage on `main`. A numpy **msgpack wire-format mismatch**
introduced by #4282 makes the server unable to decode arrays packed by
`openpi-client`, and also makes `openpi-client` unable to decode the server's
responses. A DROID/OpenPI policy client gets `"Internal inference error"` on
every inference.

This is a regression: #3673 (the original endpoint) worked because both sides
used `openpi_client.msgpack_numpy` directly.

## Root cause

`vllm_omni/entrypoints/openpi/connection.py` `_pack_numpy` / `_unpack_numpy`
(added in #4282) use the standard `msgpack-numpy` (lebedov) marker format:

```python
ndarray -> {nd: True,  type, kind, shape, data}
scalar  -> {nd: False, type, kind, data}
```

But the `openpi-client` package (v0.1.2) bundles its **own**
`openpi_client/msgpack_numpy.py`, which uses a **different** marker set:

```python
[...]

PR #4282, opened 2026-06-09
Support DROID policy server for Cosmos3 OpenPI

yuzhudong:
## Purpose

- Support the DROID policy server/OpenPI action-serving flow for Cosmos3.
- Add parity bridge updates for Cosmos3 action outputs.
- Align OpenPI connection/serving tests with the review feedback.

## Test Plan

- Syntax-check the touched Python files.
- Existing OpenPI connection/serving tests cover the updated behavior.

## Test Result

```bash
python -m py_compile vllm_omni/diffusion/models/cosmos3/pipeline_cosmos3.py vllm_omni/diffusion/models/cosmos3/transformer_cosmos3.py vllm_omni/diffusion/diffusion_engine.py vllm_omni/entrypoints/openpi/connection.py tests/entrypoints/openai_api/test_openpi_connection.py tests/entrypoints/openai_api/test_openpi_serving.py
```

Passed.

Issue #3554, opened 2026-05-12
[RFC]: Robotics Evaluation Interface Integrations


matchyc:
### Motivation.

# vLLM-Omni Robotics Integrations 

---

## 1. What this RFC is about

vLLM-Omni plan to host robot policies since #1987 . Each hosted policy is **useful only insofar as a real client**, such as an evaluation harness, a teleop tool, a downstream deployment, can talk to it in realtime over the wire. This RFC tracks that surface:

- **Which robot policies are hosted today.**
- **Which realtime-interaction communities each hosted policy can already talk to.**
- **Which communities we'd like to connect with next.**

The unit of value is the **cell** in a matrix: *policy × community*.

---

### 2.1. Progress
- [x] OpenPI entry point #3673 

#### 2.1.1 OpenPI entry point
When a community client speaks the OpenPI-style realtime wire (e.g. molmospace over msgpack WebSocket at `/v1/realtime/robot/openpi`), vLLM-Omni needs a dedicated server entrypoint—not only a policy pipeline under `vllm_omni/diffusion/models/`. The work on the serving side is:

1. **`api_server.py`** Register the OpenPI route(s).
2. **`entrypoints/openpi/`** Isolate OpenPI-specific serving from other entrypoints (OpenAI-compatible HTTP, media diffusion, etc.):
   - **`connection.py`** Session and transport: accept WebSocket (or HTTP) connections, track sessions, forward inbound observations to serving and outbound actions back to the client.
   - **`serving.py`** Inference bridge: map OpenPI request shapes to the hosted policy pipeline and vLLM-Omni engine; batching, streaming, and response serialization live here.
   - **`__init__.py`**  Public exports for `api_server` registration.

Layout (OpenPI slice only):

```text
└── entrypoints/
    └── openpi/
        ├── __init__.py
        ├── connection.py         # connection & session
        └── serving.py            # inference & engine bridge
```
### 2.1 Hosted policies

| Policy | Source | Pipeline path | Status |
|---|---|---|---|
| **DreamZero-DROID** #2162 | GEAR-Dreams | `vllm_omni/diffusion/models/dreamzero/` | Hosted, eval-validated |
| **InternVLA-A1** #1948  | Shanghai AI Lab | `vllm_omni/diffusion/models/internvla/` |Hosted |
| **GR00T** #3553  | Nvidia | `vllm_omni/diffusion/models/gr00t/` | Planned |

[...]

PR #6051, opened 2026-08-11
[Bugfix][Frontend] Accept the msgpack-numpy package's numpy markers on the OpenPI endpoint

ZJLi2013:
## Purpose

`/v1/realtime/robot/openpi` rejects every array packed by the [`msgpack-numpy`](https://pypi.org/project/msgpack-numpy/) PyPI package — the observation fails to decode and the client gets `Invalid request payload` instead of an action chunk:

```
ValueError: NumPy dtype marker kind mismatch: 'f' != ''
```

This breaks a client we ship: MolmoSpaces' `DreamZeroWebsocketClient` packs with that package, so `examples/online_serving/dreamzero/molmospace_dreamzero_eval_demo.py` cannot reach the endpoint today.

**Root cause.** Three dialects share the `nd` marker shape but disagree on `kind`:

| Dialect | plain ndarray | `kind` |
|---|---|---|
| `openpi-client` | `{__ndarray__, dtype, shape, data}` | field absent |
| vLLM-native, from #4282 | `{nd, type, kind, shape, data}` | dtype kind character, e.g. `'f'` |
| `msgpack-numpy` package | `{nd, type, kind, shape, data}` | `b''` plain, `b'V'` structured |

`_decode_vllm_numpy_marker` required `kind == dtype.kind`, which no plain array from the package can satisfy. #4506 fixed the outbound half of this mismatch; the inbound half was listed in its suggested fix but never implemented.

**Fix.** Accept either spelling of `kind` — `""` or the dtype kind character. Reject the structured form before `np.dtype()` runs: with `kind == "V"` the package puts a dtype *descriptor list* in `type`, which would otherwise surface as an opaque `TypeError`.

`kind` stays **required** — it is what separates an array marker from a plain user mapping (`test_unpack_leaves_user_dict_without_numpy_kind_marker_unchanged` pins this). Outbound packing is unchanged: replies keep the `openpi-client` markers.

## Test Plan

**vLLM Version:** N/A — the fix is in the vLLM-Omni entrypoint and does not depend on the bundled vLLM version.

**vLLM-Omni Commit:** `03a56644`, base `4ab1b4ff`

CPU only, no GPU and no model load:

```bash
pytest -q tests/entrypoints/openai_api/test_openpi_connection.py
```

New and changed cases:

- `test_unpack_msgpack_numpy_packed_observation` — round-trip through the **real package** under `importorskip`. This is what would have caught the bug.
- `test_unpack_accepts_msgpack_numpy_marker_dicts` — hand-built `kind=b""`, unconditional.
- `test_unpack_rejects_msgpack_numpy_structured_array_markers` — structured form raises.
- The previous holder of the second name asserted the vLLM-native dialect, so it is renamed to `test_unpack_accepts_vllm_native_marker_dicts`. That naming is how the two dialects got conflated.

Direct reproduction of the reported failure:

```bash
pip install msgpack-numpy
python -c "
import msgpack_numpy, numpy as np
from vllm_omni.entrypoints.openpi.connection import _unpack
print(_unpack(msgpack_numpy.packb({'observation/joint_position': np.zeros(7, dtype=np.float32)})))
"
```

## Test Result

```
# base 4ab1b4ff
3 failed, 12 passed

# this branch
15 passed
```

The repro snippet raises the `ValueError` above on base and prints the decoded `float32` array on this branch. Environment: `msgpack-numpy 0.4.8`, `msgpack 1.2.1`, `numpy 2.4.6`, Python 3.12.

The real-package round-trip also covers what #5934 gives up by swapping `openpi-client` for a stand-in built from our own pack/unpack, and `msgpack-numpy` has no `numpy<2` pin so it fits that CI. The two PRs are independent.

## Not in scope

A compatibility fix, not a redesign of the marker handling:

- **Scalars.** The package's scalar branch emits `{nd: False, type, data}` with no `kind`, so it still decodes as a mapping. Fixing that means moving the discriminator off `kind` — e.g. onto the presence of the `nd` key — which reclassifies every inbound payload. Nothing in-tree hits it: DreamZero observations are arrays.
- **One inbound dialect.** The `nd` branch exists only for clients written against servers from the #4282 era. If there are none, dropping it and keeping `__ndarray__` only is less surface than maintaining two, which also matches the principle in #3554 that each policy takes its wire from its upstream community. Worth confirming with the author of #4282.

### DIFF
--- a/tests/entrypoints/openai_api/test_openpi_connection.py
+++ b/tests/entrypoints/openai_api/test_openpi_connection.py
@@ -68,7 +68,7 @@ def test_pack_and_unpack_round_trip_numpy_values():
     assert decoded["nested"][0]["done"] == np.bool_(True)
 
 
-def test_unpack_accepts_msgpack_numpy_marker_dicts():
+def test_unpack_accepts_vllm_native_marker_dicts():
     action = np.asarray([[1.0, 2.0]], dtype=np.float32)
     payload = {
         b"actions": {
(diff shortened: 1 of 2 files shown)


linyueqian (approved):
Reviewed at `03a56644`. This correctly accepts the empty `kind` emitted for plain msgpack-numpy arrays without broadening the accepted dtype set: structured, object, and complex dtypes still fail through the same dtype gates. Package scalars and decoding the outbound action markers with msgpack-numpy remain outside this change.

The tests distinguish the vLLM and msgpack-numpy marker shapes, although the real-package case is `importorskip`-gated when that optional dependency is absent.

CI at this head. General build 14496, Intel 7676, and NPU 6571 are green. AMD build 11272 reports failed but produced no code signal: 17 of its 20 steps ended in the `errored` outcome with 3 passed and not one hard failure, across eleven hours from 05:08 to 16:05 on 2026-09-01. No test body ran, so nothing was asserted on ROCm. That is the same infrastructure-expiry shape as AMD build 11401, and cycling the `ready` label cannot replace it, because Buildkite deduplicates per commit and a finished build still blocks its own replacement.

So the ROCm evidence here is absent rather than negative, and I am approving on the three green lanes plus static review rather than waiting on a lane that cannot re-run without a new commit. This change is msgpack decoder dtype acceptance in the frontend with no device-specific or kernel code, so ROCm has no particular opinion about it, but the gap belongs on the record.

Validation: static review of the two-file diff, the decoder failure paths, the tests, and public Buildkite step metadata for every lane. Claude Opus independently confirmed no blocking or important code defect. No fork code was executed locally.

PR #6052, opened 2026-08-11
[Bugfix][MiniCPM-o] Preserve reference audio through Talker and Code2Wav

Kelsey050218:
## Purpose

MiniCPM-o 4.5 Seed-TTS requests carry reference audio for voice cloning, but vLLM-Omni did not preserve that audio through the multistage request path. The audio was either exposed as raw Talker multimodal input, although the Talker model does not consume it, or dropped before the first Code2Wav payload.

The request could still finish and return audio, but Code2Wav could fall back to the checkpoint's default prompt voice instead of conditioning on the request's reference speaker.

### What broke

The expected path is:

```text
request audio
  -> request-side input
  -> Thinker-to-Talker input bridge
  -> model_intermediate_buffer
  -> first Code2Wav payload
```

Before this fix:

- `requires_multimodal_data` represented both media consumed by a stage model and original request media needed only by an input bridge;
- enabling it exposed raw audio to Talker processing, while disabling it prevented the bridge-only audio from being materialized;
- the full-payload runner stored the handoff in `additional_information_cpu`, but the Code2Wav bridge only read the streaming/legacy locations;
- non-audio handoff metadata, including the reference waveform and sample rate, could be discarded when the next stage extracted acoustic codes.

### Root cause

Model multimodal input and request-scoped bridge input are different contracts. Reusing one flag for both made it impossible to say: "materialize this audio for the stage input bridge, but do not pass it to the Talker model."

The streaming and full-payload paths also expose their handoff state through different request attributes, so fixing only one path still loses reference conditioning on the other.

### Fix

- add `request_side_input_modalities` to the stage pipeline/runtime contract;
- materialize declared side inputs once without forwarding raw `multi_modal_data` to the Talker model;
- declare MiniCPM-o Talker audio as a bridge-side input;
- serialize the Seed-TTS reference audio into the Chat Completions request and preserve `ref_audio` / `ref_text` request fields;
- write the reference waveform and sample rate into `model_intermediate_buffer`;
- preserve non-audio metadata while extracting Talker acoustic codes;
- read streaming, full-payload CPU-side, and legacy handoff buffers in priority order;
- attach the reference waveform and sample rate to the first Code2Wav payload on both streaming and full-payload paths.

The benchmark adapter keeps the configured system-prompt text unchanged and only adds the existing reference audio as an `audio_url` content part. This PR does **not** add or replace prompt wording, prompt features, speaker-feature extraction, acoustic-token sampling, or Code2Wav conditioning math.

### Out of scope

StepAudio2/FlashCosyVoice package-layout compatibility is independent of request-side routing and is intentionally excluded from this PR.

## Test Plan

```bash
PYTHONPATH=/tmp/vllm-omni-test-stubs:$PWD python -m pytest -q \
  tests/benchmarks/patch/test_patch.py \
  tests/engine/test_stage_engine_core_proc.py \
  tests/test_config_factory.py \
  tests/model_executor/models/minicpmo_4_5/test_llm2tts.py \
  tests/entrypoints/openai_api/test_serving_chat_multistage_generation.py
```

**vLLM Version:** `0.1.dev1+ge5588e49b`

**vLLM-Omni base:** `minicpm-challenge@009b80d686febcf683fdbc2bcdf3ad752884641e`

## Test Result

- Unit/regression suite after scope cleanup: **511 passed, 2 skipped**. The six removed cases covered the out-of-scope prompt wording and redundant deploy assertions.
- Diff-scoped pre-commit passed. `actionlint` was skipped because no workflow files changed and the host Go 1.17 cannot install the hook's Go 1.19+ dependencies.

### Fixed 32-request Seed-TTS comparison

This is a regression-focused before/after comparison, not a claim about an external aggregate baseline.

Conditions held constant:

- hardware: one Ascend 910B1 for all three serving stages;
- model: `openbmb/MiniCPM-o-4_5`;
- runtime image: `openbmb/minicpm-challenge-tts-seed:acba6622-wer`;
- dataset: Seed-TTS English, first 32 loader rows in fixed order;
- selection: `--num-prompts 32 --disable-shuffle --seed 42 --no-oversample`;
- serving: one turn per session, 3 warmups, max concurrency 1, request rate `inf`;
- request: text + inline reference audio, TTS template enabled, thinking disabled;
- scorer: BytedanceSpeech/seed-tts-eval `cal_sim`, fine-tuned WavLM-large, CPU;
- the before and after artifacts contain the same 32 utterance IDs in the same order.

| Metric | Before routing fix | After routing fix |
|---|---:|---:|
| Mean ASV similarity | 0.0710411 | **0.5445324** |
| Evaluated / failed | 32 / 0 | **32 / 0** |
| Generated audio duration | 142.28 s | **142.28 s** |
| Samples improved | — | **32 / 32** |

The mean ASV increase is **+0.4734912**, or **7.66x** the pre-fix value. Every one of the 32 matched samples improved; per-sample deltas ranged from +0.1745862 to +0.7539873.

The captured E2E routing head was `53b6c51da1b5213e1bf4eca63124da2e43710af1`. Both matched arms used the same locale-specific benchmark prompt wrapper, so prompt wording was held constant. That wrapper is not included in this scoped PR; consequently, the table is historical matched routing evidence rather than a fresh accuracy run of the final scoped diff. That test image requires an orthogonal import-only compatibility patch for its packaged FlashCosyVoice layout so Code2Wav can start. The compatibility patch does not touch request payloads, prompt features, acoustic tokens, generated outputs, or scoring, and has been removed from this PR to keep the bugfix focused. The preserved pre-fix artifact also reported a dirty source checkout whose diff was not retained, so this table is presented as matched regression evidence rather than a clean upstream performance benchmark.


vllm_omni/entrypoints/openai/serving_chat.py
```diff
@@ -922,12 +955,13 @@ async def _prepare_multistage_multimodal_inputs(
             stripped_content: list[Any] = []
             changed = False
             for part in content:
-                modality, payload = self._deferred_multimodal_part(part, deferred_modalities)
+                modality, payload = self._deferred_multimodal_part(part, downstream_modalities)
```
linyueqian, line 958: [important] Media that is a downstream modality but is ALSO consumed by stage 0 gets materialized twice, and for MiniCPM-o the reference audio is exactly that case.

The loop collects the part into `downstream_parts` unconditionally, but only removes it from `stripped_content` when the modality is in `deferred_modalities`. MiniCPM-o stage 0 declares `requires_multimodal_data=True` and stage 1 declares `request_side_input_modalities=("audio",)`, so audio is downstream but not deferred: it is recorded here AND left in the message. `_materialize_deferred_multimodal_parts` then fetches and decodes it through a fresh `MediaConnector`, and the retained content part is later handed to `render_chat_async` for ordinary stage-0 media processing, which fetches and decodes it again.

The cost is not only latency. Two independent loads of the same reference clip means an HTTP source is requested twice, so a presigned, expiring or single-use URL can succeed on the first pass and fail on the second, and the failure surfaces in the path this PR exists to make reliable.

The fix wants shared media to have exactly one fetch and decode. Either reuse the stage-0 materialized audio for request-side inputs instead of re-materializing here, or cache by source within the request so both consumers hit the same decoded buffer. Worth a test that asserts a single fetch for a shared reference source, since nothing in the current eight test files would catch a regression here.

linyueqian (changes requested):
Reviewed at `12f17f60`, against this PR's own base branch `minicpm-challenge` at `009b80d6` rather than main, so the footprint here is the two commits you actually wrote: 16 files, +482/-55.

The direction is right and the root cause is addressed. Separating request-side audio from Talker model input, and preserving the reference waveform through both the streaming and full-payload buffer locations, is the correct shape for this bug, and the structured handoff reads consistently across the two paths.

One item, inline, and it lands squarely in the flow this PR exists to fix: for MiniCPM-o the reference audio ends up materialized twice, once by the downstream bridge and once again by stage 0's normal renderer path, because it is a downstream modality that stage 0 also consumes and therefore never gets stripped. That is extra latency on every request, and for an HTTP reference source it means two fetches, which breaks outright on expiring or single-use URLs.

Two things about the surrounding context rather than the code, neither of which is a criticism of the change.

CI on this head is unusually thin. The only checks reporting are DCO and readthedocs: no build, no pre-commit, no test job. So nothing here has been exercised beyond what I could read, and the eight test files in this change have not run anywhere I can see.

This PR targets `minicpm-challenge`, not `main`. It is `mergeable=true` and `state=clean` against that base, so there is nothing to fix, but it does mean the change reaches main only when the base branch does, and a reviewer coming from the main branch will see a very different diff than the one under review here.

Validation: static reads and CI evidence only, no code from this head executed or imported. I computed the footprint against `merge-base(origin/minicpm-challenge, HEAD)` rather than against main, which matters on this PR: comparing to main pulls the entire base branch in and reports 153 files of unrelated CI, Docker and benchmark work that you did not write. I traced the modality contract through `_downstream_multimodal_modalities` and `_deferred_multimodal_modalities` into the stripping loop, confirmed the MiniCPM-o stage flags in `pipeline.py`, and checked the anchor line falls inside a real diff hunk. I did not run the tests, reproduce the audio path, or verify anything on a GPU.

vllm_omni/entrypoints/openai/serving_chat.py
lishunyang12, line 958: Wrap input_audio.data in a data URL using its format before fetching.

PR #6054, opened 2026-08-11
[Bugfix][NPU] Eager-import mindiesd for every diffusion attention backend

brandneway:
The eager mindiesd import in get_diffusion_attn_backend_cls was gated on
backend_upper == "FLASH_ATTN", so selecting RAINFUSION_ATTN skipped it.
RAINFUSION_ATTN still reaches mindiesd through its dense FlashAttention
fallback (used before start_step and on layers without a sparsifiable
video segment). Without the eager import, CANN snapshots the custom-op
registry at the first regInfo lookup before mindiesd's env.py prepends
its vendor dirs to ASCEND_CUSTOM_OPP_PATH, so aclnnLaserAttention /
FusedAttentionScore fail with EZ1001 "does not support opType" for the
rest of the process.

Drop the backend gate so the import runs for every selected backend.


vllm_omni/platforms/npu/platform.py
```diff
@@ -139,10 +139,13 @@ def get_diffusion_attn_backend_cls(
                 )
                 backend_upper = "FLASH_ATTN"
 
-            if backend_upper == "FLASH_ATTN" and find_spec("mindiesd"):
-                # The NPU FLASH_ATTN backend imports mindiesd lazily at first
-                # forward, but CANN snapshots the custom-op registry at the
-                # first custom-op regInfo lookup in the process (e.g. a
+            if find_spec("mindiesd"):
```
david6666666: This also imports MindIE-SD for explicit `TORCH_SDPA` (and every other backend), so a broken optional install can now block a path that never uses it. Could we limit this to `FLASH_ATTN`/`RAINFUSION_ATTN` and add a small selector regression test?

david6666666:
@fan2956 ptal thx

david6666666 (approved):
LGTM

PR #6058, opened 2026-08-11
[Hardware][MUSA] Complete Omni platform interfaces

yeahdongcn:
## Summary

- implement MUSA device visibility environment hooks with MUSA_VISIBLE_DEVICES
- add an inductor-aware default IR op priority for diffusion workers
- query device capability through torch.musa and align common platform method ordering with ROCm

## Testing

- python -m compileall -q vllm_omni/platforms/musa/platform.py
- git diff --check
- pre-commit run --files vllm_omni/platforms/musa/platform.py

MUSA runtime tests were not run because this local checkout does not have torch, vllm, or vllm_musa installed.

### DIFF
--- a/vllm_omni/platforms/musa/platform.py
+++ b/vllm_omni/platforms/musa/platform.py
@@ -3,6 +3,8 @@
 
 
 import torch
+from vllm.config import VllmConfig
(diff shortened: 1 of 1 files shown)


gcanlin (approved):
LGTM

PR #6013, opened 2026-08-11
[Model][Wan] Add automatic spatial-shard VAE decode

hsliuustc0106:
## Summary

- keep `vae_parallel_mode="tile"` as the global default
- add `"auto"` as an explicit mode for Wan VAE decode
- select Wan spatial-shard decode only for 5D tiled-decode requests using the full DiT process group
- shard the longer eligible latent spatial axis, with width winning ties
- fall back to tile decode for partial groups, unsupported shapes, direct-decode requests, and unshardable axes
- keep selection, topology checks, and fallback policy in `DistributedAutoencoderKLWan`
- retain the direct Wan preparation path without adding a shared policy registry or generalized VAE protocol
- make Wan spatial wrappers request-scoped and reversible so one worker can alternate tile, height, and width paths
- install structural wrappers inside the tagged model-weight allocation pool and release halo scratch buffers after every request
- retain `tile`, `spatial_shard_height`, and `spatial_shard_width` as explicit escape hatches

## Scope

This PR is intentionally Wan-only. It does not change MiniMax H3 behavior or introduce a cross-model spatial-decode policy. Other VAE families retain their existing mode handling and may reject explicit Wan spatial modes.

## Why

This adapts the request-shape selection idea from sgl-project/sglang#28071 to vLLM-Omni's parity-tested Wan backend.

It intentionally does **not** copy SGLang's 4096-pixel threshold. vLLM-Omni reaches the selector only after Diffusers has already selected `tiled_decode`; applying that threshold here suppressed a measured speedup for the smallest tiled Wan shape.

## Performance

Wan2.1 VAE, BF16, L20X, maximum-rank decode wall time, 2 warmups + 7 measured iterations (median).

Fresh post-rebase 2-GPU results:

| latent (T,H,W) | tile | auto | speedup |
|---:|---:|---:|---:|
| (1,60,104) | 94.283 ms | 22.000 ms | 4.29x |
| (2,60,104) | 179.884 ms | 75.010 ms | 2.40x |
| (2,104,60) | 179.796 ms | 74.109 ms | 2.43x |
| (5,60,104) | 469.596 ms | 236.133 ms | 1.99x |

The same Wan implementation previously produced these 4-GPU results before the rebase:

| latent (T,H,W) | tile | auto | speedup |
|---:|---:|---:|---:|
| (1,60,104) | 72.085 ms | 17.509 ms | 4.12x |
| (2,60,104) | 117.321 ms | 49.981 ms | 2.35x |
[...]

PR #6062, opened 2026-08-11
[Bugfix] Fix Wan spatial reshard boundary

rahul-steiger-nv:
## Purpose

Fix Wan spatial-shard VAE decoding when padding creates an empty trailing shard.

After gathering and trimming the global tensor, the last rank could compute a start offset beyond the trimmed extent. `torch.narrow` rejects this offset even when the requested length is zero.

Clamp the start and valid extent to the actual tensor size, then preserve equal shard shapes with zero-padding. Add diagnostics for rank and attention extent mismatches.

## Test Plan

- Run CPU spatial-shard unit tests.
- Compare height/width sharded decode against the reference on multiple GPUs.
- Run a 4-GPU Cosmos3 video smoke test.

**vLLM Version:** 0.25.0

**vLLM-Omni Commit:** 688143b199e4ebde1fce6e8d29cabb10d5c3e94a

## Test Result

- CPU: `25 passed, 2 deselected`
- Multi-GPU reference comparison: `2 passed`
- 4-GPU Cosmos3 requests:
  - 80x64 boundary regression: HTTP 200
  - 320x192, 189 frames: HTTP 200

### DIFF
--- a/tests/diffusion/distributed/test_wan_spatial_shard.py
+++ b/tests/diffusion/distributed/test_wan_spatial_shard.py
@@ -233,6 +233,128 @@ def test_reshard_from_trimmed_width_pads_invalid_columns(monkeypatch: pytest.Mon
 
 @pytest.mark.core_model
 @pytest.mark.cpu
+@pytest.mark.parametrize("split_dim", ["height", "width"])
+@pytest.mark.parametrize(
+    ("full_extent", "world_size"),
+    [
+        pytest.param(5, 4, id="four-ranks-empty-tail"),
+        pytest.param(20, 8, id="production-eight-ranks-empty-tail"),
+        pytest.param(39, 2, id="two-ranks-partial-tail"),
+    ],
+)
+def test_reshard_from_trimmed_extent_clamps_and_reconstructs(
+    monkeypatch: pytest.MonkeyPatch,
+    split_dim: str,
+    full_extent: int,
+    world_size: int,
+):
+    local_extent = (full_extent + world_size - 1) // world_size
+    shape = (1, 1, 1, full_extent, 1) if split_dim == "height" else (1, 1, 1, 1, full_extent)
+    x = torch.arange(full_extent, dtype=torch.float32).reshape(shape)
+    outputs = []
+
+    for rank in range(world_size):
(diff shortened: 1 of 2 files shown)


vllm_omni/diffusion/distributed/autoencoders/wan_spatial_shard.py
```diff
@@ -203,8 +203,17 @@ def reshard_from_trimmed_extent(
...
+    if ctx is not None and ctx.rank != rank:
+        raise RuntimeError(
+            "Wan VAE spatial-shard rank mismatch while resharding: "
+            f"group_rank={rank}, context_rank={ctx.rank}, split_dim={split_dim!r}."
+        )
+
     valid_extent = _local_valid_extent(local_extent)
-    start = rank * local_extent
+    actual_extent = x.shape[dim]
+    start = min(rank * local_extent, actual_extent)
```
hsliuustc0106, line 215: This handles the empty-tail case: when `rank * local_extent` exceeds the trimmed extent, clamping to `actual_extent` makes the zero-length `narrow` legal while the existing padding restores the fixed shard size.

tests/diffusion/distributed/test_wan_spatial_shard.py
```diff
@@ -231,6 +231,128 @@ def test_reshard_from_trimmed_width_pads_invalid_columns(monkeypatch: pytest.Mon
     assert torch.equal(out.flatten(), torch.tensor([4.0, 0.0]))
 
 
+@pytest.mark.core_model
+@pytest.mark.cpu
+@pytest.mark.parametrize("split_dim", ["height", "width"])
+@pytest.mark.parametrize(
+    ("full_extent", "world_size"),
+    [
+        pytest.param(5, 4, id="four-ranks-empty-tail"),
```
hsliuustc0106, line 240: This 4-rank case reproduces the failure on #6013's current head; parametrizing both spatial axes also protects its dynamic height/width selection when that PR rebases.

hsliuustc0106 (approved):
LGTM. Verified the focused CPU suite and 2-GPU height/width parity tests.

PR #6063, opened 2026-08-11
[New Model] Add Falcon Perception support

avirajBevli:
## Purpose
Add vLLM-Omni support for Falcon Perception, which is a vision-language model for open-vocabulary instance segmentation
https://github.com/tiiuae/Falcon-Perception
https://huggingface.co/tiiuae/Falcon-Perception

This model follows a **2 stage pipeline**:
- **Stage0:** thinker stage (AR LLM): generates the `<coord>`, `<size>`, `<seg>` tokens autoregressively for each object instance. Object coordinates and sizes are decoded from the hidden states corresponding to the the sampled `<coord>`,` <size>` tokens. Their embeddings are then injected back into the next autoregressive decoding step.
- **Stage1:** segmentation stage : takes in the patch features, hidden states of <seg> tokens of each object instance as input. And as output, generates one segmentation mask per detected object instance

## Test Plan
**vLLM Version:** `0.26.0`

## Test Result
Examples of output segmentation masks overlayed on top of the original image.

- Query = birds
- Input image
image

- Output segmentation mask overlayed on top of the original image
image

### L2 test : pass
```
python -m pytest -s -v \
  tests/e2e/offline_inference/test_falcon_perception.py \
  -m "core_model and cuda" \
  --run-level core_model
```
image

### L3 test result: pass
```
VLLM_USE_FLASHINFER_SAMPLER=0 \
python -m pytest -s -v \
  tests/e2e/offline_inference/test_falcon_perception.py \
  -m "advanced_model and cuda" \
  --run-level advanced_model
```
image

## Out of scope
Support for Falcon perception detection-only model is not added here. That is left for a future PR

### DIFF
--- a/examples/offline_inference/falcon_perception/batch_inference.py
+++ b/examples/offline_inference/falcon_perception/batch_inference.py
@@ -1,0 +1,437 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+"""Falcon Perception: multi-request offline inference over a whole request set.
+
+``end2end.py`` is the usage demo — one image, one query. This script is the
+throughput form: every request is handed to a **single** ``omni.generate()`` call
+so vLLM's scheduler can keep up to ``max_num_seqs`` stage-0 sequences in flight
+(continuous batching on a paged KV cache). Calling ``generate()`` once per
+request in a Python loop admits only one sequence at a time and measures serial
+latency, not overlapping prefill/decode. Stage 1 still runs once per finished
+thinker request; map outputs by request id, not finish order.
+
+Usage:
+    # explicit pairs
+    python batch_inference.py --model tiiuae/Falcon-Perception \\
+        --image a.jpg --query "the red pepper" \\
+        --image b.jpg --query "every person" --max-num-seqs 4
+
+    # a JSONL manifest of {"image": ..., "query": ...} lines
+    python batch_inference.py --manifest requests.jsonl \
+        --deploy-config falcon_perception.yaml
+
+    # a HuggingFace dataset with `image` and `expression` columns
+    python batch_inference.py --dataset tiiuae/PBench --split level_1 \\
+        --limit 50 --deploy-config falcon_perception.yaml
+
+Reading the throughput numbers
+------------------------------
+Four things decide whether the reported ``images/s`` means anything:
+
+* **``max_num_seqs``.** ``falcon_perception.yaml`` uses the measured A100-80GB
+  optimum of 4 on both stages. Larger is not faster for this model. Use
+  ``--max-num-seqs`` to retune on other hardware; this script warns if an
+  override serialises the workload at 1.
+* **``enforce_eager``.** The shipped YAML sets it to ``false`` on both stages.
+  Stage 0 also uses ``compilation_config: {cudagraph_mode: FULL_DECODE_ONLY}``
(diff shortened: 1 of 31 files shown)


avirajBevli:
@hsliuustc0106  
Request to have a look at the PR when convenient. 
Request adding ready label so the 1 pending check (buildkite) can run? 

Note that the other 5 checks ran successfully.

vllm_omni/model_executor/models/falcon_perception/falcon_perception_segmentation.py
```diff
@@ -0,0 +1,600 @@
...
+        raise ValueError(f"binary masks must have shape (N, H, W), got {tuple(binary_masks.shape)}")
+
+    count, height, width = binary_masks.shape
+    if count <= 1:
+        return torch.arange(count, device=binary_masks.device, dtype=torch.long)
+
+    scale = min(1.0, nms_max_side / max(height, width))
+    target_h = max(1, int(round(height * scale)))
+    target_w = max(1, int(round(width * scale)))
+    masks = binary_masks.float()
```
lishunyang12: Downsample masks in chunks; this full-resolution float copy defeats the memory bound.

vllm_omni/model_executor/models/falcon_perception/falcon_perception_thinker.py
```diff
@@ -0,0 +1,1030 @@
...
+
+        Same contract as the in-tree talker stages: the runner writes the
+        returned dict into its per-request intermediate buffer, so nothing is
+        kept on the module and continuous batching stays safe.
+        """
+        # Runner-called, once per decode step, outside the compiled model.
+        with nvtx_range("fp/s0/postprocess"):
+            if hidden_states.numel() == 0:
+                return {}
+            return {"hidden_states": {"last": hidden_states[-1].detach().contiguous()}}
```
lishunyang12, line 717: Mark hidden_states.last GPU-resident; the runner currently copies it to CPU every token.

examples/offline_inference/falcon_perception/batch_inference.py
```diff
@@ -0,0 +1,437 @@
...
+    (48, 209, 88),
+]
+
+
+def build_prompt(query: str) -> str:
+    """The exact string the model expects. Deviating here silently degrades output."""
+    return f"<|image|>Segment these expressions in the image:<|start_of_query|>{query}<|REF_SEG|>"
+
+
+def overlay(image: Image.Image, masks: np.ndarray) -> Image.Image:
```
lishunyang12: Reuse end2end.py's prompt and overlay helpers instead of duplicating them here.

vllm_omni/model_executor/models/falcon_perception/falcon_perception_segmentation.py
avirajBevli: Makes sense. Fixed this

vllm_omni/model_executor/models/falcon_perception/falcon_perception_thinker.py
avirajBevli: Makes sense to avoid useless gpu -> cpu -> gpu transfer from postprocess of current step to preprocess of the next step. Fixed this

examples/offline_inference/falcon_perception/batch_inference.py
avirajBevli: Created a _common file from which both import. hence avoid duplicate

Issue #5584, opened 2026-07-30
[Bug]: Performance drop - Higgs


congw729:
### Your current environment

<details>
<summary>The output of <code>python collect_env.py</code></summary>

```text
ci env
```

</details>

### Your code version

<details>
<summary>The commit id or version of vllm</summary>

```text
0.26.0
```
</details>
<details>
<summary>The commit id or version of vllm-omni</summary>

```text

```
</details>

### 🐛 Describe the bug

Series | 2026-06-12 | 2026-06-13 | 2026-06-14 | 2026-06-15 | 2026-06-16 | 2026-06-17 | 2026-06-18 | 2026-06-19 | 2026-06-20 | 2026-06-21 | 2026-06-22 | 2026-06-23 | 2026-06-24 | 2026-06-25 | 2026-06-26 | 2026-06-27 | 2026-06-28 | 2026-06-29 | 2026-06-30 | 2026-07-01 | 2026-07-02 | 2026-07-03 | 2026-07-04 | 2026-07-05 | 2026-07-06 | 2026-07-07 | 2026-07-08 | 2026-07-09 | 2026-07-10 | 2026-07-11 | 2026-07-12 | 2026-07-13 | 2026-07-14 | 2026-07-15 | 2026-07-16 | 2026-07-17 | 2026-07-18 | 2026-07-19 | 2026-07-20 | 2026-07-21 | 2026-07-22 | 2026-07-24 | 2026-07-25 | 2026-07-26 | 2026-07-27 | 2026-07-28
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
[...]

PR #6068, opened 2026-08-11
[Bugfix][Higgs-Audio-V3] Disable XQA decode

Sy0307:
### What broke

Higgs Audio V3 Stage 0 explicitly selects `FLASHINFER`, but `FLASHINFER` is the overall vLLM backend rather than a guarantee that its native decode kernel will run. Starting with vLLM 0.25, the backend can route Hopper decode to TRT-LLM XQA when `attention_config.use_trtllm_attention` is unset. For the Higgs BF16 workload, that route regresses TTFP and RTF on H800. The H100 nightly regression is tracked in #5584.

### Reproduction

Run the existing Higgs performance test on Hopper:

```bash
export VLLM_USE_DEEP_GEMM=0 VLLM_MOE_USE_DEEP_GEMM=0
pytest -s -v tests/dfx/perf/scripts/run_benchmark.py \
  --test-config-file tests/dfx/perf/tests/test_higgs_audio_v3.json
```

With the current Stage 0 config, vLLM 0.25+ logs `decode_backend=xqa`. Adding `attention_config.use_trtllm_attention: false` changes it to `decode_backend=flashinfer-native` and restores the pre-v0.25 latency range.

### Why XQA is selected

The selection chain is:

```text
attention_backend: FLASHINFER
  -> use_trtllm_attention is unset (None)
  -> Hopper/SM90 has an available XQA cubin
  -> Higgs query/KV heads are divisible (32 / 8)
  -> FlashInfer decode builder statically selects backend=xqa
```

The important distinction is capability versus performance. `can_use_trtllm_attention()` checks whether TRT-LLM attention is disabled, supported on the platform, and compatible with the head layout. Higgs passes those correctness checks. The decode builder then resolves the kernel statically so CUDA graph capture does not switch backends at runtime.

That static decision does **not** compare model dtype, KV-cache dtype, context length, batch shape, or measured kernel time. In other words, `32 % 8 == 0` proves that XQA can execute; it does not prove that XQA is the fastest kernel for Higgs.

### Why this shape regresses

Higgs' eight audio codebooks are easy to misread as eight attention queries. They are not. One autoregressive audio step produces one hidden row per request, and the fused modality head projects that row into logits with shape `[1, 8, codebook_size]`. The codebook dimension is created **after** attention, so at concurrency one the attention kernel still sees one BF16 query row (`q_len=1`), not `q_len=8`.

This makes the relevant workload a repeated small BF16 decode shape. XQA's benefit is shape- and dtype-dependent: the upstream XQA integration reported a clear FP8 gain, but its BF16 benchmark had 21.16 ms mean TPOT for XQA versus 19.47 ms for FA3 ([upstream data](https://github.com/vllm-project/vllm/pull/43232#issuecomment-4501176335)). That comparison uses FA3 rather than native FlashInfer, so it is not the direct evidence for this fix; it does show why an SM90 capability gate is insufficient as a BF16 performance policy.

The direct evidence is the same-stack H800 A/B below: changing only `use_trtllm_attention` moves Higgs from XQA to native FlashInfer and recovers essentially the entire v0.25 regression. Since attention runs in every transformer layer and every generated audio step, the per-call disadvantage accumulates into visible TTFP and RTF loss.

What is proven here is the routing-level root cause: automatic XQA selection is slower for this end-to-end Higgs workload. The current data does not isolate one defective XQA instruction, launch, or memory transaction. Pinning that deeper mechanism down requires an attention-only benchmark and GPU profiler trace across the exact Higgs batch/context distribution; the local deployment note states this boundary explicitly.

### Fix

- Pin `attention_config.use_trtllm_attention: false` for Stage 0 in the default, high-throughput, and low-latency Higgs deploy profiles.
- Register vLLM 0.27 `attention_config` in the structured per-stage model scope so typed config projection preserves the pin instead of rejecting it as unowned.
- Extend the deploy-config unit test to require both `FLASHINFER` and the native FlashInfer pin in all three profiles.
- Document the selector mismatch, the actual one-row attention shape, the benchmark evidence, and the remaining kernel-level uncertainty in `vllm_omni/deploy/README_higgs_audio_v3.md`.

The change only disables the optional XQA sub-backend for these Higgs profiles; it keeps the existing FlashInfer backend and does not change Stage 1 or other models. A possible tradeoff is that a future XQA version could outperform native FlashInfer for some high-concurrency BF16 shapes while this model-level pin remains conservative. H100 c1/c16 A/B validation and a future workload-aware upstream selector are therefore follow-up items.

Related to #5584.

## Test Plan

**vLLM Version:** v0.25.0 for the H800 A/B; v0.27.0 on the current vLLM-Omni base

**vLLM-Omni Commit:** 5d2ecf15

- [x] `ruff check tests/model_executor/models/higgs_audio_v3/test_higgs_audio_v3.py`
- [x] `ruff format --check tests/model_executor/models/higgs_audio_v3/test_higgs_audio_v3.py`
- [x] Parse all three deploy YAMLs and verify Stage 0 selects native FlashInfer
- [x] Diff-scoped `precheck-pr` quick review (DCO, title, code quality, scope, and regression test)
- [ ] `pytest -q tests/model_executor/models/higgs_audio_v3/test_higgs_audio_v3.py::TestRegistry::test_deploy_yaml_uses_native_flashinfer` (local environment has no vLLM/Torch test dependencies)
- [x] Main CUDA Buildkite #13512: Engine&Entrypoints (`1849 passed`) and Other (`2204 passed`)
- [x] `TTS · Higgs-Audio-V3 Test`: `9 passed, 1 skipped`; Stage 0 logged `decode_backend=flashinfer-native`
- [x] H100 nightly c1/c16 benchmark: `2 passed`; c16 mean TTFP `291.03 ms`, mean RTF `0.55`

## Test Result

H800, `seed-tts-text`, c1/p20, same vLLM 0.25 stack unless noted ([full investigation](https://github.com/vllm-project/vllm-omni/issues/5584#issuecomment-5188700218)):

| Mode | Decode backend | TTFP (ms) | RTF |
|---|---|---:|---:|
| vLLM 0.25 default | XQA | 175.76 | 0.3405 |
| vLLM 0.25 with this pin | FlashInfer native | 166.39 | 0.3233 |
| vLLM 0.24 reference | FlashInfer native | 166.68 | 0.3209 |

### DIFF
--- a/tests/config/test_omni_config.py
+++ b/tests/config/test_omni_config.py
@@ -471,6 +471,7 @@ def test_sub_config_fields_match_structured_scopes():
         "trust_remote_code",
         "dtype",
         "attention_backend",
+        "attention_config",
         "moe_backend",
         "hf_overrides",
         "limit_mm_per_prompt",
--- a/tests/engine/test_stage_engine_args.py
+++ b/tests/engine/test_stage_engine_args.py
@@ -139,6 +139,7 @@ def _engine_arg_inputs(tmp_path: Path) -> tuple[PipelineConfig, DeployConfig, st
                 engine_extras={
                     "model": str(stage_model),
                     "attention_backend": "FLASHINFER",
+                    "attention_config": {"use_trtllm_attention": False},
                     "hf_overrides": {"rope_scaling": {"factor": 2}},
                     "limit_mm_per_prompt": {"audio": 1},
                     "logits_processors": ["test.custom.LogitsProcessor"],
@@ -229,6 +230,7 @@ def test_typed_llm_engine_args_preserve_legacy_adapter_behavior(tmp_path):
     assert thinker_args["omni_kv_config"] == {"need_send_cache": True}
     assert thinker_args["worker_cls"] == "test.custom.Worker"
     assert thinker_args["attention_backend"] == "FLASHINFER"
+    assert thinker_args["attention_config"] == {"use_trtllm_attention": False}
     assert thinker_args["trust_remote_code"] is True
     assert thinker_args["dtype"] == "float16"
     assert thinker_args["distributed_executor_backend"] == "mp"
--- a/tests/model_executor/models/higgs_audio_v3/test_higgs_audio_v3.py
+++ b/tests/model_executor/models/higgs_audio_v3/test_higgs_audio_v3.py
@@ -1326,7 +1326,9 @@ class TestRegistry:
 
         assert "higgs_multimodal_qwen3" in OMNI_PIPELINES
 
-    def test_deploy_yaml_exists(self):
+    def test_deploy_yaml_uses_native_flashinfer(self):
+        from vllm_omni.config.stage_config import load_deploy_config
+
         for name in (
             "higgs_multimodal_qwen3.yaml",
             "higgs_multimodal_qwen3_high_throughput.yaml",
@@ -1334,6 +1336,10 @@ class TestRegistry:
         ):
             yaml_path = get_deploy_config_path(name)
             assert os.path.isfile(yaml_path), f"Deploy YAML not found at {yaml_path}"
+            deploy_config = load_deploy_config(yaml_path)
+            stage_0 = next(stage for stage in deploy_config.stages if stage.stage_id == 0)
+            assert stage_0.engine_extras["attention_backend"] == "FLASHINFER"
+            assert stage_0.engine_extras["attention_config"] == {"use_trtllm_attention": False}
 
 
 # ---- AC-3: Prompt Builder ----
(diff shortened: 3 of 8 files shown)


herotai214:
Hi, thanks for this PR. I'd like to confirm my understanding:

`use_trtllm_attention` is not a good option for Higgs because:
- The `use_trtllm_attention` flag is intended for multi‑head attention models (where head counts satisfy XQA's divisibility conditions).
- Higgs, however, doesn't really behave like multiple attention heads in practice (it has 8 codebooks but only a single BF16 query during decoding)   → **meaningless and WRONG to use**.
- The benchmarks show that enabling it actually slows down inference (TTFP: 175ms → 166ms, RTF: 0.34 → 0.32 when disabled), so it brings no speedup  → **remove**.

Then yes I think this can be a permenant fix instead of a hot fix I been doubting in the issue page.
LGTM for this PR

linyueqian (approved):
lgtm

PR #5500, opened 2026-07-28
[Feat] LTX Standard Two-Stage Pipeline

mglyn:
## Purpose

This PR completes ordinary and full-distilled two-stage T2V/I2V for LTX-2 and LTX-2.3. Recipes define phase schedules, guidance, transitions, weights, and output routing while reusing the shared runtime. This work builds on #5148; follow-up progress is tracked in #4985.

## Pipelines and weights

| Pipeline | Stage 1 | Stage 2 | Required repositories |
|---|---|---|---|
| `LTX2TwoStagePipeline` | Base Transformer at half resolution; requested Euler steps with official guidance | Fixed three-step positive-only Euler refinement with distilled LoRA `1.0` | LTX-2: `Lightricks/LTX-2`; LTX-2.3: `diffusers/LTX-2.3-Diffusers` + `Lightricks/LTX-2.3` |
| `LTX2DistilledPipeline` | Merged distilled Transformer; fixed eight-step positive-only schedule | Same Transformer with fixed three-step positive-only refinement | LTX-2: `rootonchair/LTX-2-19b-distilled`; LTX-2.3: `diffusers/LTX-2.3-Distilled-Diffusers` + `Lightricks/LTX-2.3` upsampler |

Ordinary two-stage loads the matching distilled LoRA and x2 upsampler; LTX-2.3 uses v1.1 sidecars. Generic YAML `model_paths` provides authoritative local overrides; otherwise artifacts resolve from the model root and Lightricks repository.

## Phase LoRA modes

Dynamic evaluates `base(x) + lora_b(lora_a(x))` on one Transformer and supports a quantized base. Layer-fused evaluates `x · BF16(W + BF16(B @ A))`, matching official arithmetic without a second Transformer; it requires unquantized BF16. 

## Correctness and limits

- I2V conditioned tokens survive the transition; SP audio padding does not consume request RNG or affect sampler state/rescale. Rank-local guidance branches share one A/V RoPE layout.
- Masked LTX cross-attention uses SDPA after SP redistribution. The generic Ulysses fix gathers unequal Q/K/V shard lengths independently and validates 2D key-padding masks against Key length.
- TP, pure Ulysses SP, HSDP, distributed VAE, CPU/layerwise offload, compile, and VAE patch parallelism reuse shared paths. Ring SP rejects requests requiring an audio key-padding mask.
- Batching, custom sigmas, and request latents are rejected. Full-distilled also rejects negative prompts/guidance overrides; Cache-DiT waits for phase-aware state.
[...]

Issue #4983, opened 2026-07-09
[LTX2.3]: Align implementation and precision guard with the official pipeline


mglyn:
## Problem

Current LTX2.3 behavior is still primarily aligned with the HuggingFace Diffusers implementation, not the official LTX2.3 pipeline semantics.

The official LTX repository presents the native pipeline as the full-featured reference path, while Diffusers is positioned as a familiar/simple API.

The implementation differences are not limited to test plumbing. Local investigation found semantic differences in the actual generation path, including:

- text embedding preprocessing;
- initial latent sampling / packing behavior;
- guidance branch organization;
- CFG denoise forward organization;
- full multimodal guidance support: CFG, STG, modality guidance, rescale.

So keeping Diffusers as the reference risks locking Omni to the wrong behavior.

## Evidence

Manual video evidence, generated with a shared CFG-only config on NVIDIA L20:
```
`1024x576`, `81f`, `20` steps, `24 fps`, seed `42`
prompt: `A cinematic close-up of ocean waves at golden hour.`
negative prompt: `worst quality, inconsistent motion, blurry, jittery, distorted`
CFG-only guidance scale: `4.0`
```
- Omni main: 

https://github.com/user-attachments/assets/aa39061a-6295-42ad-81d3-f2659dc09876

- Diffusers: 

https://github.com/user-attachments/assets/d7bf21e9-8bf3-4268-a610-cc6e2fe68f6d

- Official: 

https://github.com/user-attachments/assets/63d9daae-a075-40f8-805c-fdec39d26e32

Pairwise decoded-video metrics:

| Pair | Frames | SSIM mean | SSIM min | PSNR mean (dB) | PSNR min (dB) | Mean abs diff |
|---|---:|---:|---:|---:|---:|---:|
| Omni main vs Diffusers | 81 | 0.937111 | 0.918324 | 35.026 | 33.928 | 0.011473 |
| Omni main vs Official | 81 | 0.536844 | 0.513602 | 15.197 | 14.162 | 0.133854 |
| Diffusers vs Official | 81 | 0.542993 | 0.517845 | 15.234 | 14.214 | 0.133204 |

The current Omni main output is much closer to Diffusers than to the official
pipeline output under the same request-level config.

## Proposed Direction

Both the implementation and CI coverage should be updated:

- update Omni LTX2.3 semantics to match the official native pipeline;
- replace the Diffusers-based precision guard with an official-pipeline-based
  guard.

[...]

PR #2527, opened 2026-04-06
[Cache Refactor 1/N] Simplify CacheDiT Integration

alex-jw-brooks:
## Purpose
See RFC https://github.com/vllm-project/vllm-omni/issues/2535

Currently, integration with Cache DiT involves copying a lot of code, but the underlying library that implements Cache DiT is very generic. We should not have to write much code for supporting new models in vLLM omni unless they need to handle special cases, e.g., Wan2's multi transformer architecture.

This PR is one of a few intended to simplify transformer caching in vLLM Omni so that we can hopefully have more generic integrations and have wider free model support, or at least make progress towards adding support simply being adding a new block adapter object.

For this PR:
- Creates a generic `build_cache_context_refresh` that we can use to build the cache refresh closure for different models; currently we have ~10 identical copy pastes of this function
- Moves common parts like creating the calibration config out
- Adds a common `enable_cache_for_dit` that takes as input a `BlockAdapter`, since this is the main thing that changes across models.
- Adds a `CacheDiTAdapterConfig` class, which we can set as an attribute on the DiT to enable CacheDiT, which should generally not require additional code changes / custom enablers except in some special cases. E.g.,
```
    _cache_dit_adapter_config = CacheDiTAdapterConfig(
        block_forward_patterns={
            "layers": ForwardPattern.Pattern_3,
        },
        has_separate_cfg=True,
        cached_adapter_cls=SensenovaCachedAdapter,
    )
```

I have verified that all outputs look reasonable and match `main` as much as possible for all models that aren't broken; the only thing that has changed is that I removed a hardcoded `Fn_compute_blocks = 2` for Flux2Klein, because it was inconsistent with what we had for Flux2dev, and also ignored user values.

CC @lishunyang12 @wtomin, could you please take a look when you get the chance?

PR #6070, opened 2026-08-11
[Model] Add LTX-2.5 support

lishunyang12:
- [x] Four LTX-2.5 pipeline classes with T2V and first-frame I2V.
  - [x] Offline examples and OpenAI-compatible online serving.
  - [x] Full/SFT and distilled one-/two-stage schedules, CRF-18 conditioning, x2 latent upsampling, LoRA 450 refinement, and FP32 re-noise.
  - [x] BF16 cuDNN official parity 
  - [x] HSDP2, DLO DP2, whole-model offload, and VAE slicing qualified as memory/capacity paths.
  - [x] Native DP2 scheduling, TP2/SP2, FP8, regional compile, VAE tiling/patch parallel, and H100 FA3 validation
  - [ ] Cache-DiT, TRTLLM attention, Ring SP, and TeaCache are not supported for LTX-2.5 in this PR.


vllm_omni/engine/async_omni_engine.py
```diff
@@ -309,16 +309,26 @@ def __init__(
...
+        not specify a class. ``model_class_name`` may be ``None``.
         """
         if self._diffusion_od_config_view is None:
             from types import SimpleNamespace
 
             from vllm_omni.diffusion.data import resolve_model_class_name
             from vllm_omni.diffusion.model_metadata import get_diffusion_model_metadata
 
-            model_class_name = resolve_model_class_name(self.model)
+            model_class_name = None
```
hsliuustc0106: why we need to make changes here?

hsliuustc0106:
I compared frozen head `80936f7` with an independent LTX-2.5 implementation built from the same target base and with the official Diffusers implementation. This PR should be the implementation base: it includes the LTX-2.5-specific CRF-18 I2V path, unpacked FP32 latent RNG/state, FP32 RoPE, static prompt AdaLN tables, cross-modality timestep semantics, the one-token cuDNN fallback, online pipeline-class resolution, and `transformer_full/` support. Those are material correctness/coverage improvements, not just refactoring.

Before merge:

- Tighten `requirements/common.txt` from `transformers >= 5.5.3` to `transformers >= 5.10.1, < 5.15`. The LTX-2.5 profile requires `Gemma4UnifiedForConditionalGeneration`; 5.8/5.9 do not provide it, 5.10.0 is yanked, and 5.15 has the known Gemma4 construction regression. The ImportError in `ltx2_components.py` should recommend the same range instead of `>=5.8,<5.15`.
- The CPU-safe affected suite currently reports `262 passed, 4 failed`. All four failures are the existing CFG-parallel cases in `test_ltx2_parallel.py`: the new FP32 paths directly access `self.model_version` on legacy test doubles. Using `getattr(self, "model_version", "2")` in the runtime dtype selection preserves the legacy contract and makes the suite green.
- Add LTX-2.5 to `docs/user_guide/diffusion_features.md`; the table and interpolation note still stop at LTX-2/2.3.
- Add or wire an online e2e test. The checked-in [B300 validation](https://github.com/vllm-project/vllm-omni/blob/80936f7f782e473def438f56cf2cab54f4e822df/recipes/LTX/LTX-2.5.md#L205-L251) is useful, but no model/GPU test check is visible in CI and there is no attached Diffusers quality-parity artifact.

A few useful choices from the alternative implementation are worth retaining:

- Use `AutoModelForImageTextToText` or otherwise keep version detection tolerant of Gemma4/transformer-signature conversions.
- State explicitly that the supported checkpoint is `Lightricks/LTX-2.5-Diffusers`; the raw `Lightricks/LTX-2.5` split-artifact repository is not directly loadable by this path.
- Decide explicitly whether request-provided `sigmas=` should remain supported. This PR rejects custom distilled schedules, while the official Diffusers API accepts them.
- For I2V, CRF-18 should remain the correct default; an explicit `image_crf=0` opt-out would be more flexible than categorically rejecting tensor inputs.

Recommendation: keep this PR's transformer/runtime/full-pipeline implementation, apply the dependency and test fixes above, and carry over only the focused loader/detection/API clarifications rather than combining both patches wholesale.

examples/offline_inference/text_to_video/text_to_video.py
```diff
@@ -104,12 +104,43 @@
...
+        "fps": 24,
+        "output": "ltx25_full_output.mp4",
+    },
 }
 
 
 def _detect_preset(model: str, model_class_name: str | None = None) -> dict:
     model_lower = model.lower()
     class_lower = (model_class_name or "").lower()
+    is_ltx25 = "ltx-2.5" in model_lower or "ltx_2.5" in model_lower or "ltx25" in model_lower
```
hsliuustc0106: `--model` accepts local paths, so `/workspace/model --model-class-name LTX2Pipeline` lands on the legacy LTX-2 preset (512x768, 40 steps) and LTX-2.5 later rejects the fixed schedule. Please resolve the version from checkpoint metadata via `detect_ltx_model_version()` and cover an arbitrary local path.

examples/offline_inference/image_to_video/image_to_video.py
```diff
@@ -350,12 +350,25 @@ def calculate_dimensions(
...
+    return image.resize((width, height), PIL.Image.Resampling.LANCZOS)
+
+
 def main():
     args = parse_args()
     generator = torch.Generator(device=current_omni_platform.device_type).manual_seed(args.seed)
     model_name = str(args.model).lower() if args.model is not None else ""
     model_class_name = args.model_class_name
     model_class_name_lower = (model_class_name or "").lower()
+    is_ltx25 = "ltx-2.5" in model_name or "ltx_2.5" in model_name or "ltx25" in model_name
```
hsliuustc0106: The same path-name check breaks local LTX-2.5 I2V checkpoints: `is_ltx25` remains false, so the image is pre-resized before the required CRF-18 round trip and the legacy defaults are selected. Use the checkpoint metadata detector here too.

vllm_omni/diffusion/registry.py
```diff
@@ -71,6 +71,11 @@
         "pipeline_ltx2",
         "LTX2Pipeline",
     ),
+    "LTX2FullPipeline": (
```
hsliuustc0106: Please add `LTX2FullPipeline` to the existing registry and post-process resolution parameterizations in `test_ltx2_pipeline.py`. This new public entry is currently omitted from those contract tests.

david6666666:
@mglyn ptal thx

mglyn:
This looks reasonable to merge before #5500, since it does not implement the Full guided two-stage path. 
#5500 could then rebase and add it using the LTX-2.5 upsampler and distilled LoRA 450.

vllm_omni/engine/async_omni_engine.py
lishunyang12: Removed the  change and its dedicated tests. Pipeline-class selection remains in the stage worker configuration, and the online smoke now verifies serving without this scope expansion.

examples/offline_inference/text_to_video/text_to_video.py
lishunyang12: Fixed. The generic T2V example now resolves the model class and calls  against checkpoint metadata, including arbitrary local paths. Regression coverage is in  and the LTX pipeline tests.

examples/offline_inference/image_to_video/image_to_video.py
lishunyang12: Fixed. The I2V example now resolves LTX version from checkpoint metadata rather than the path name. Online serving also preserves the source image geometry for , so CRF preprocessing remains pipeline-owned and occurs before resize.

vllm_omni/diffusion/registry.py
lishunyang12: Fixed.  is now covered by the existing registry, post-process, public-entry, and model-extras parameterizations in  and .

tests/diffusion/models/ltx2/test_ltx2_transformer.py
```diff
@@ -2,16 +2,62 @@
...
+    LTX2VideoTransformer3DModel,
+    LTX2VideoTransformerBlock,
+    _make_rms_norm,
+    apply_interleaved_rotary_emb,
+    apply_split_rotary_emb,
+)
 
 pytestmark = [pytest.mark.core_model, pytest.mark.cpu]
 
 pytestmark = [pytest.mark.core_model, pytest.mark.diffusion, pytest.mark.cpu]
```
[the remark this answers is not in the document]
lishunyang12: Fixed in 98af1ec73. The redundant marker assignment was removed; the file now has one  declaration containing , , and .

vllm_omni/diffusion/models/ltx2/ltx2_components.py
```diff
@@ -342,9 +414,11 @@ def initialize_pipeline_components(pipeline: Any, od_config: Any) -> None:
         subfolder="tokenizer",
         local_files_only=local_files_only,
     )
+    if profile.text_encoder_cls is None:
+        raise ImportError("LTX-2.5 requires Gemma4UnifiedForConditionalGeneration. Install transformers>=5.8,<5.15.")
```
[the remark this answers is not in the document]
lishunyang12: Keeping the explicit LTX-2.5 message and range intentionally.  is  only for the LTX-2.5 profiles when Gemma4 Unified is unavailable, and  is the verified runtime constraint requested in review and now declared in . The exception mirrors that installable constraint instead of suggesting a broader incompatible range.

vllm_omni/diffusion/models/ltx2/ltx2_transformer.py
```diff
@@ -1966,13 +1986,14 @@ def batch_step(step: torch.Tensor) -> torch.Tensor:
         audio_embedded_timestep = audio_embedded_timestep.view(batch_size, -1, audio_embedded_timestep.size(-1))
 
         # 3.2. Prepare global modality cross attention modulation parameters
+        video_ca_timestep = audio_sigma.flatten() if use_cross_timestep else timestep.flatten()
```
linyueqian: P2: This re-keys the cross-attn modulation sources for the released LTX-2 and LTX-2.3 paths, not just 2.5. On main, `av_cross_attn_video_scale_shift` took the video timestep and the a2v gate took the audio sigma. Neither branch of `use_cross_timestep` reproduces that combination: with the flag on (2.3 and 2.5) the video scale_shift input switches from the masked per-token video timestep to the uniform audio sigma; with it off (LTX-2) the a2v gate switches from the audio sigma to the masked video timestep. T2V is unaffected since all four sources are numerically equal there, but under I2V the video timestep is zeroed on conditioning-frame tokens, so LTX-2 and LTX-2.3 I2V outputs change. These embeddings are consumed whenever AV cross attention runs (line 1115), independent of the adaln flags.

Can you confirm both re-keyings match the pinned official implementations for 2 and 2.3, and share a run of the `ltx2_3_i2v` accuracy case on this head? Note there is no LTX-2 I2V accuracy case, so that side currently has no guard.

vllm_omni/diffusion/model_metadata.py
```diff
@@ -52,6 +53,7 @@ class DiffusionModelMetadata:
         max_multimodal_image_inputs=9,
         supports_mixed_reference_inputs=True,
     ),
+    "LTX2Pipeline": DiffusionModelMetadata(preserves_reference_image_size=True),
```
linyueqian: P2: `preserves_reference_image_size` is keyed on the `LTX2Pipeline` class, which also serves LTX-2 and LTX-2.3. For those versions the server previously LANCZOS-resized the reference image to the requested WxH, which made the pipeline's aspect-preserving resize plus center crop a no-op. With the server resize skipped, off-aspect inputs are now center-cropped instead of squashed, so legacy online I2V framing changes. The offline example still applies the LANCZOS squash for non-2.5 inputs (`prepare_primary_image`), so offline and online now disagree for 2/2.3.

If restoring the official preprocessing for legacy online I2V is intended, this deserves a release-note line; otherwise the flag needs to be version-conditional rather than class-wide.

vllm_omni/diffusion/models/ltx2/ltx2_transformer.py
```diff
@@ -1488,7 +1497,10 @@ class LTX2VideoTransformer3DModel(nn.Module):
         block_forward_patterns={
             "transformer_blocks": ForwardPattern.Pattern_0,
         },
-        has_separate_cfg=True,
+        # LTX fuses all guidance passes into the batch dimension and invokes
+        # the transformer once per denoise step. Cache-DiT must therefore
+        # advance its step counter for every transformer forward.
+        has_separate_cfg=False,
```
linyueqian, line 1536: P2: The flip itself looks correct. The guidance executor fuses every pass into the batch dimension and calls the transformer once per denoise step, so `has_separate_cfg=True` (from #2527) made Cache-DiT advance its step counter at the wrong rate for LTX. But Cache-DiT is documented as supported for LTX-2/2.3, so thresholds tuned under the old counting are stale after this change. A quick re-validation of the legacy Cache-DiT path, or at least a release-note callout, would cover it.

tests/e2e/accuracy/ltx/test_ltx_official_similarity.py
```diff
@@ -296,6 +361,28 @@ def _request(case: LTXAccuracyCase, image: Path | None) -> dict[str, object]:
...
+    return {
+        "prompt": LTX25_PROMPT,
+        "negative_prompt": NEGATIVE_PROMPT,
+        "width": 960,
+        "height": 544,
+        "num_frames": 121,
+        "fps": 24,
+        "num_inference_steps": 8,
+        "seed": 42,
+        "sigmas": LTX25_DISTILLED_SIGMAS,
```
linyueqian: Nit: `_run_diffusers` requires `request["sigmas"]`, but `_run_omni` forwards only the `video_*`/`audio_*` keys, so the Omni side silently ignores this key and falls back to the recipe schedule. Parity holds today because `LTX_DISTILLED_SIGMAS` equals these 8 values plus the terminal 0. If either side drifts, the test would compare different schedules without failing loudly. Consider forwarding `sigmas` through the extra args on the omni backend too, or asserting the recipe schedule matches the request.

linyueqian (commented):
Reviewed at 98af1ec73. The 2.5-specific work (CRF-18 I2V, unpacked FP32 latent RNG, dynamic sigma shift, version detection, profiles/recipes) looks well built and well tested. My comments are all about side effects on the released LTX-2/2.3 paths that share this code. Separately: the intel-ci failure is pre-existing on main HEAD, not from this PR.

mglyn:
One accuracy concern: in our previous LTX-2/2.3 work, we found clear numerical differences between the Diffusers and Lightricks official implementations. Since the LTX-2.5 accuracy results here use Diffusers as the reference, they demonstrate Diffusers parity rather than parity with the official pipeline.

mglyn:
#4983

vllm_omni/diffusion/models/ltx2/ltx2_transformer.py
lishunyang12: Confirmed and fixed on the current head. The version branch was removed: video scale/shift consumes the per-token video timestep, the A2V gate consumes audio sigma, audio scale/shift consumes audio timestep, and the V2A gate consumes video sigma. `test_ltx_cross_attention_modulations_use_official_timesteps` explicitly exercises the I2V-relevant masked video timestep `[0, 2]` against audio `[5]`. The PR already carries `diffusion-x2v-test`; its Other Accuracy Test runs the `ltx2_3_i2v` case on this head. I am leaving this thread open until that H100 result reports.

vllm_omni/diffusion/model_metadata.py
lishunyang12: Fixed on the current head without changing async engine plumbing. `OmniOpenAIServingVideo.preserves_reference_image_size` now detects the checkpoint version for the shared `LTX2Pipeline`: only 2.5 preserves the source geometry, while 2/2.3 retain the existing server resize. `test_ltx_i2v_resize_policy_is_version_conditional` covers 2.5, 2.3, and `LTX2FullPipeline`.

vllm_omni/diffusion/models/ltx2/ltx2_transformer.py
lishunyang12: Covered with the requested release-note callout in `recipes/LTX/LTX-2.md`: it states that LTX fused guidance is one Transformer call per denoise step and that residual thresholds from separate-CFG step accounting must be recalibrated. The adapter contract is also pinned by `test_ltx_transformer_has_fused_cfg_cache_dit_config`.

tests/e2e/accuracy/ltx/test_ltx_official_similarity.py
lishunyang12: Fixed. The Omni reference runner now forwards request `sigmas` through `apply_declared_extra_args`, so the tested schedule is explicit on both sides rather than relying on equality with the recipe default. The offline extra-body registry and its regression test also include the supported schedule keys.

hsliuustc0106:
fix docs build

recipes/LTX/LTX-2.5.md
```diff
@@ -0,0 +1,306 @@
...
+LTX-2.5 generates video and synchronized 48 kHz stereo audio. vLLM-Omni
+supports the following generation paths through three pipeline classes:
+
+> **Checkpoint layout:** This integration directly supports only
+> `Lightricks/LTX-2.5-Diffusers`. The raw `Lightricks/LTX-2.5` repository uses
+> split artifacts and cannot be passed directly to `--model` on this path.
+
+| Mode | `--model-class-name` | Output | Steps | Transformer |
+|---|---|---:|---:|---|
+| One-stage T2V | `LTX2Pipeline` | 960x544, 121 frames at 24 FPS | 8 | `transformer/` |
```
mglyn: I noticed a concrete request-contract issue: both LTX-2.5 distilled recipes inherit the default negative prompt and allow_negative_prompt=True, although their guidance is positive-only. As a result, negative prompts or embeddings are accepted but never used. They should use negative_prompt="" and allow_negative_prompt=False, consistent with the existing LTX-2 distilled recipe.

Separately, I suggest naming the transformer/ one-stage path LTX2DistilledOneStagePipeline. Rejection behavior remains recipe-driven, but the dedicated name would keep class semantics clear: LTX2Pipeline means guided Full/Dev one-stage execution for LTX-2/2.3, whereas the LTX-2.5 transformer/ path is distilled, positive-only, and eight-step. This should help prevent similar default and capability confusion later.

lishunyang12: Fixed in 237f775c7. Both LTX-2.5 positive-only recipes now set `negative_prompt=""` and `allow_negative_prompt=False`, with rejection tests for prompt and embedding inputs. After the pipeline API discussion, the canonical names are `LTX2Pipeline` for one-stage and `LTX2TwoStagePipeline` for two-stage; `LTX2DistilledPipeline` remains only as the existing compatibility alias. The recipe and public registries use the canonical two-stage name.

vllm_omni/diffusion/models/ltx2/ltx2_recipes.py
mglyn, line 11: LTX_DEFAULT_NEGATIVE_PROMPT is behind the latest native Lightricks default. The official prompt now starts with has_subtitles, has_blurbox, transition from black, transition to black, speech_ending_short.

tests/e2e/online_serving/test_ltx25.py
```diff
@@ -0,0 +1,70 @@
...
+
+pytestmark = [pytest.mark.diffusion, pytest.mark.full_model]
+SINGLE_CARD_MARKS = hardware_marks(res={"cuda": "H100"})
+
+
+def _cases():
+    return [
+        pytest.param(
+            OmniServerParams(
+                model=MODEL,
```
mglyn: REVIEW(blocking): This is not currently true for the LTX
production loader. ``--revision`` reaches config enrichment,
but version detection and component/weight loads still use
repository HEAD. Add a propagation test before keeping this
pinned-runtime contract.

vllm_omni/diffusion/models/ltx2/ltx2_latents.py
```diff
@@ -184,7 +193,10 @@ def prepare_video_latents(
             )
         if latents.ndim != 3:
             raise ValueError(f"Provided `latents` has shape {latents.shape}, expected [batch, seq, features].")
-        return create_noised_state(latents, noise_scale, generator).to(device=device, dtype=dtype)
+        latents = create_noised_state(latents, noise_scale, generator).to(device=device, dtype=dtype)
```
mglyn, line 218: REVIEW(high): #6146 found that native LTX-2/2.3 also obtains this
token-major storage layout by sampling 5D latents before patchifying.
Restricting it to 2.5 leaves the established 2/2.3 precision fix out
of this shared path. Rebase on #6146 or generalize the layout here.

vllm_omni/diffusion/models/ltx2/ltx2_runtime.py
```diff
@@ -137,14 +137,26 @@ class LTXRuntime(
 
     def __init__(self, *, od_config: OmniDiffusionConfig, prefix: str = "") -> None:
         del prefix
+        parallel_config = getattr(od_config, "parallel_config", None)
+        if getattr(parallel_config, "ulysses_mode", "strict") == "advanced_uaa":
+            raise ValueError(
+                f"{self.__class__.__name__} does not support ulysses_mode='advanced_uaa'. "
+                "Use the default ulysses_mode='strict' for LTX sequence parallelism."
+            )
```
mglyn, line 146: REVIEW(blocking): ``od_config.revision`` is dropped here and by the
downstream LTX component loaders.  A pinned request can therefore
resolve its config at one revision while detecting the version and
loading weights/tokenizer/schedulers from repository HEAD.  Thread
the standard revision through the complete LTX Hub-loading chain.

mglyn (commented):
b4497b67

vllm_omni/diffusion/models/ltx2/ltx2_runtime.py
```diff
@@ -610,7 +650,14 @@ def _denoise_timestep_kwargs(
         audio_token_count: int,
     ) -> dict[str, torch.Tensor]:
         del forward_ctx, denoise_ctx
-        return self.guidance_executor.timestep_kwargs(ts, video_token_count, audio_token_count)
+        parallel_config = getattr(getattr(self, "od_config", None), "parallel_config", None)
+        sp_size = int(getattr(parallel_config, "sequence_parallel_size", 1) or 1)
```
mglyn: REVIEW(blocking): main and native LTX pass per-token timesteps for
every model version.  Returning batch scalars for non-SP LTX-2/2.3
changes their numerical path and makes SP/non-SP semantics differ.
Expand for every version, and add a non-SP LTX-2/2.3 regression test.

vllm_omni/diffusion/models/ltx2/ltx2_latents.py
```diff
@@ -205,7 +217,10 @@ def prepare_video_latents(
             f"You have passed a list of generators of length {len(generator)}, but requested an effective batch"
             f" size of {batch_size}. Make sure the batch size matches the length of the generators."
         )
-    return randn_tensor(shape, generator=generator, device=device, dtype=dtype)
+    latents = randn_tensor(shape, generator=generator, device=device, dtype=dtype)
```
mglyn, line 240: REVIEW(high): Keep this consistent with the provided-latents branch
above; the native token-major layout is not specific to LTX-2.5.

vllm_omni/diffusion/models/ltx2/ltx2_runtime.py
lishunyang12: Fixed in b30693b4a. od_config.revision now reaches version detection, Hub prefetch, tokenizer, each component load, transformer config and weight source, scheduler, raw snapshot materialization, raw component sources, and output sample-rate detection. The production-chain regression covers both converted and official split layouts.

vllm_omni/diffusion/models/ltx2/ltx2_latents.py
lishunyang12: Fixed in b30693b4a. prepare_video_latents now materializes the official token-major storage layout for LTX-2, 2.3, and 2.5. The seeded RNG regression is parameterized across all three versions.

vllm_omni/diffusion/models/ltx2/ltx2_runtime.py
lishunyang12: Fixed in b30693b4a. Timestep kwargs are expanded per token for every LTX version regardless of SP, with a non-SP parameterized regression for 2, 2.3, and 2.5.

vllm_omni/diffusion/models/ltx2/ltx2_transformer.py
lishunyang12: Current head c4ee9ab51 retains the official modulation routing fix and test_ltx_cross_attention_modulations_use_official_timesteps; the post-merge affected suite is 406 passed. The diffusion-x2v-test label remains attached. I am leaving this thread open until the external H100 ltx2_3_i2v result is reported, as requested.

vllm_omni/diffusion/models/ltx2/ltx2_latents.py
lishunyang12: Fixed in b30693b4a. The provided-latents branch now applies the same official token-major layout for 2, 2.3, and 2.5, with a version-parameterized stride/value regression.

tests/e2e/online_serving/test_ltx25.py
lishunyang12: Fixed in b30693b4a. The pinned online case now uses the official split checkpoint, and revision propagation is covered for both paths: raw materialization pins the exact selected files and every raw weight source; the converted-loader regression checks prefetch, tokenizer, all six components, transformer config, scheduler, and transformer weight source. The post-process sample-rate lookup is pinned too.

vllm_omni/diffusion/models/ltx2/ltx2_recipes.py
lishunyang12: Fixed in b30693b4a. LTX_DEFAULT_NEGATIVE_PROMPT now begins with the five current native defaults: has_subtitles, has_blurbox, transition from black, transition to black, and speech_ending_short, before the existing list.

docs/user_guide/diffusion_features.md
```diff
@@ -168,10 +169,17 @@ The following tables show which models support each feature:
 > **Step execution note:** Helios supports single-request step execution only;
 > use `max_num_seqs=1`.
 
+> **LTX-2.5 qualification note:** the official raw split-checkpoint path is
```
hsliuustc0106: do not add the model specific notes here

recipes/LTX/LTX-2.5.md
```diff
@@ -0,0 +1,269 @@
+# LTX-2.5
```
hsliuustc0106, line 1: please use the latest add-recipe skill

vllm_omni/entrypoints/openai/serving_video.py
```diff
@@ -100,6 +100,34 @@ def set_stage_configs_if_missing(self, stage_configs: list[Any] | None) -> None:
         if self._stage_configs is None and stage_configs is not None:
             self._stage_configs = stage_configs
 
+    @property
+    def preserves_reference_image_size(self) -> bool:
```
hsliuustc0106, line 106: do not add model specific patch here

vllm_omni/entrypoints/utils.py
```diff
@@ -284,6 +285,11 @@ def resolve_model_config_path(model: str) -> str | None:
     Raises:
         ValueError: If model_type cannot be determined
     """
+    # Official LTX-2.5 uses a split-artifact repository with no root config.
```
hsliuustc0106: model spefic patch, can we move it into models folder?

docs/user_guide/diffusion_features.md
lishunyang12: Fixed in dad5acb87. Removed the model-specific LTX-2.5 qualification block from the shared diffusion-features page; the table now keeps only the compact capability row.

recipes/LTX/LTX-2.5.md
lishunyang12: Fixed in dad5acb87 using the current recipe template: added when-to-use guidance, the exact B300 environment and command, verification, operational notes, installation, and key-parameter sections while retaining all three official pipeline selectors.

vllm_omni/entrypoints/openai/serving_video.py
lishunyang12: Fixed in dad5acb87. serving_video now calls a generic metadata resolver and contains no LTX class/version branch. The LTX-2.5 CRF-18 resize policy lives in models/ltx2; the complete serving suite is 88 passed.

vllm_omni/entrypoints/utils.py
lishunyang12: Fixed in dad5acb87. Removed the direct LTX predicate from entrypoints/utils.py. Unindexed diffusion repositories now use a model-agnostic is_diffusion_model fallback, covered by a generic split-checkpoint regression; the complete utility suite is 34 passed.

lishunyang12:
## Official LTX-2.5 BF16 cuDNN parity and runnable pipelines

Updated at `8ce11ef77`. The comparison pins native Lightricks source [`7954dcb`](https://github.com/Lightricks/LTX-2/commit/7954dcb0d986bdc36ef272564a9789ade07fcc65), raw model revision `8a4ff96`, connector revision `a6de4b5`, seed 42, identical schedules and conditioning, and explicit BF16 cuDNN attention in both runtimes. MATH is used only for the degenerate one-token fallback.

| vLLM-Omni pipeline | Native reference | Task | SSIM mean | SSIM min | PSNR | LPIPS |
|---|---|---|---:|---:|---:|---:|
| `LTX2Pipeline` | `TI2VidOneStagePipeline` Full/SFT | T2V | 0.999652 | 0.999564 | 58.80 dB | 0.0000367 |
| `LTX2Pipeline` | `TI2VidOneStagePipeline` Full/SFT | I2V | 0.999589 | 0.999480 | 62.37 dB | 0.0000576 |
| `LTX2TwoStagePipeline` | `TI2VidTwoStagesPipeline` Full + LoRA 450 | T2V | 0.999588 | 0.999485 | 56.40 dB | 0.0000687 |
| `LTX2TwoStagePipeline` | `TI2VidTwoStagesPipeline` Full + LoRA 450 | I2V | 0.999585 | 0.999475 | 62.10 dB | 0.0000627 |
| `LTX2DistilledTwoStagePipeline` | `DistilledPipeline` | T2V | 0.997186 | 0.996164 | 46.04 dB | 0.003450 |
| `LTX2DistilledTwoStagePipeline` | `DistilledPipeline` | I2V | 0.996489 | 0.995810 | 46.29 dB | 0.003787 |

The Full two-stage I2V diagnosis is stronger than decoded similarity: native and Omni are bitwise exact at the Stage 1 packed output, Stage 2 input, and Stage 2 final packed latent. The fixed issue was premature BF16 rounding of the official FP32 Stage 2 re-noise scalar. All public paths now pass SSIM mean/min >= 0.99 and PSNR >= 40 dB.

The native project has no public distilled one-stage entrypoint. `LTX2DistilledOneStagePipeline` remains a supported Omni mode and is covered by unit/runnable checks, but is intentionally not presented as an independent public native E2E row.

Permanent guard:

- [`test_ltx25_official_similarity.py`](https://github.com/lishunyang12/vllm-omni/blob/feat/ltx-2-5-support/tests/e2e/accuracy/ltx/test_ltx25_official_similarity.py)
- [`run_ltx25_reference.py`](https://github.com/lishunyang12/vllm-omni/blob/feat/ltx-2-5-support/tests/e2e/accuracy/ltx/run_ltx25_reference.py)

### Full/SFT one-stage

```bash
python examples/offline_inference/text_to_video/text_to_video.py \
  --model Lightricks/LTX-2.5-Diffusers --model-class-name LTX2Pipeline \
  --prompt "A cinematic red fox walking through a snowy forest at dawn, tracked alongside." \
  --height 544 --width 960 --num-frames 121 --num-inference-steps 30 \
  --frame-rate 24 --fps 24 --enforce-eager --output ltx25-full-one-stage.mp4
```

### Full/SFT two-stage

```bash
python examples/offline_inference/text_to_video/text_to_video.py \
  --model Lightricks/LTX-2.5-Diffusers --model-class-name LTX2TwoStagePipeline \
  --prompt "A cinematic red fox walking through a snowy forest at dawn, tracked alongside." \
  --height 1088 --width 1920 --num-frames 121 --num-inference-steps 30 \
  --frame-rate 24 --fps 24 --enforce-eager --output ltx25-full-two-stage.mp4
```

### Distilled one-stage

```bash
python examples/offline_inference/text_to_video/text_to_video.py \
  --model Lightricks/LTX-2.5-Diffusers --model-class-name LTX2DistilledOneStagePipeline \
  --prompt "A cinematic red fox walking through a snowy forest at dawn, tracked alongside." \
  --height 544 --width 960 --num-frames 121 --num-inference-steps 8 \
  --frame-rate 24 --fps 24 --enforce-eager --output ltx25-distilled-one-stage.mp4
```

### Distilled two-stage

```bash
python examples/offline_inference/text_to_video/text_to_video.py \
  --model Lightricks/LTX-2.5-Diffusers --model-class-name LTX2DistilledTwoStagePipeline \
  --prompt "A cinematic red fox walking through a snowy forest at dawn, tracked alongside." \
  --height 1088 --width 1920 --num-frames 121 --num-inference-steps 8 \
  --frame-rate 24 --fps 24 --enforce-eager --output ltx25-distilled-two-stage.mp4
```

For I2V, use `examples/offline_inference/image_to_video/image_to_video.py`, preserve the same class-specific dimensions and steps, and add `--image /absolute/path/to/first-frame.png`. The complete offline and online commands are in [`recipes/LTX/LTX-2.5.md`](https://github.com/lishunyang12/vllm-omni/blob/feat/ltx-2-5-support/recipes/LTX/LTX-2.5.md). Hosted long-form outputs remain in the [B300 gallery](https://lishunyang12.github.io/vllm-omni-rankings/scripts/ltx25_official_b300_1080p20s/).

tests/e2e/accuracy/ltx/test_ltx25_official_similarity.py
```diff
@@ -0,0 +1,887 @@
...
+        "omni_model_revision": omni_model_revision,
+        "resolved_omni_model_path": str(omni_model),
+        "video": video_metrics,
+        "audio": audio_metrics,
+    }
+    (output_root / "metrics.json").write_text(json.dumps(result, indent=2) + "\n")
+    print(json.dumps(result, indent=2))
+
+    ssim_mean_threshold = LTX25_VIDEO_SSIM_MEAN_THRESHOLD if task == "t2v" else LTX25_I2V_VIDEO_SSIM_MEAN_THRESHOLD
+    assert video_metrics["ssim_mean"] >= ssim_mean_threshold
```
hsliuustc0106, line 586: [P1] Official LTX-2.5 parity fails at the pinned revisions

I ran this exact added test at head `e2721dc97d48a412128327039b2a80bb07c5b0ea` on one NVIDIA L20X (Python 3.12.13, vLLM 0.27.0, vLLM-Omni `0.27.0rc2.dev60+ge2721dc97`, PyTorch 2.13.0+cu129):

```bash
CUDA_VISIBLE_DEVICES=2 HF_HOME=/workspace/model \
python -m pytest -q -s \
  'tests/e2e/accuracy/ltx/test_ltx25_official_similarity.py::test_ltx25_distilled_two_stage_matches_official[t2v]'
```

The harness resolved official source `7954dcb0d986bdc36ef272564a9789ade07fcc65`, raw model `8a4ff96f581e72bedc1b44367581c49d544a05f1`, and Diffusers model `a6de4b5354f078db24d9cf4778c14846788aea3d`; both sides used `torch_sdpa`, seed 42, 1024x576, 25 frames, and the same 8+3 schedules. It nevertheless failed in 322.35 s with video SSIM mean/min `0.66285/0.65434` (required `0.995/0.99`), PSNR `17.812 dB` (required `40`), LPIPS `0.36954` (required `<=0.01`), audio cosine `0.14073` (required `>=0.95`), and relative L2 `2.59623` (required `<=0.3`). Because the two runtimes run sequentially on the same device/backend and the mismatch is this large, this does not look like a minor L20X-vs-H100 tolerance difference. Please reproduce and fix the official/Omni trajectory mismatch before treating this parity coverage as passing.

For comparison, the public pipelines themselves are runnable locally. These recipe commands both exited 0 and produced valid H.264 + 48 kHz stereo AAC MP4s:

```bash
CUDA_VISIBLE_DEVICES=2 HF_HOME=/workspace/model python \
  examples/offline_inference/text_to_video/text_to_video.py \
  --model Lightricks/LTX-2.5-Diffusers \
  --model-class-name LTX2DistilledOneStagePipeline \
  --prompt 'A cinematic shot of a red fox walking through a snowy forest at dawn, the camera tracking alongside, snow crunching underfoot.' \
  --height 544 --width 960 --num-frames 121 --num-inference-steps 8 \
  --frame-rate 24 --fps 24 --enforce-eager --output ltx25-distilled-one-stage.mp4

CUDA_VISIBLE_DEVICES=2 HF_HOME=/workspace/model python \
  examples/offline_inference/text_to_video/text_to_video.py \
  --model Lightricks/LTX-2.5-Diffusers \
  --model-class-name LTX2DistilledTwoStagePipeline \
  --prompt 'A cinematic shot of a red fox walking through a snowy forest at dawn, the camera tracking alongside, snow crunching underfoot.' \
  --height 1088 --width 1920 --num-frames 121 --num-inference-steps 8 \
  --frame-rate 24 --fps 24 --enforce-eager --output ltx25-distilled-two-stage.mp4
```

Observed one-stage: 9.42 s generation, 78,666 MiB peak, 960x544/121 frames/5.042 s. Observed two-stage: 23.58 s generation, 111,716 MiB peak, 1920x1088/121 frames/5.042 s. The focused CPU-safe pipeline file also passed: `36 passed in 2.70s`.

hsliuustc0106:
## Independent LTX-2.5 parity and runnable pipeline check

Locally tested at PR head `e2721dc97d48a412128327039b2a80bb07c5b0ea` on NVIDIA L20X GPUs with Python 3.12.13, vLLM 0.27.0, vLLM-Omni `0.27.0rc2.dev60+ge2721dc97`, PyTorch 2.13.0+cu129, Diffusers 0.38.0, and PyAV 18.1.0. GPU 1/2 were isolated for the final runs and were released afterward.

### Official parity reproduction

The PR's added harness pinned native Lightricks source [`7954dcb`](https://github.com/Lightricks/LTX-2/commit/7954dcb0d986bdc36ef272564a9789ade07fcc65), raw model revision `8a4ff96f581e72bedc1b44367581c49d544a05f1`, and Diffusers/connector revision `a6de4b5354f078db24d9cf4778c14846788aea3d`. Native and Omni ran sequentially on the same local GPU with PyTorch SDPA, seed 42, 24 FPS, 25 decoded frames, identical connector inputs, and the checked-in schedules.

```bash
CUDA_VISIBLE_DEVICES=2 HF_HOME=/workspace/model python -m pytest -q -s \
  tests/e2e/accuracy/ltx/test_ltx25_official_similarity.py \
  -k 'ltx25_distilled_two_stage_matches_official or ltx25_full_matches_official'
```

All six public native comparisons completed but failed their checked-in similarity gates locally:

| vLLM-Omni pipeline | Native reference | Task | SSIM mean | SSIM min | PSNR | LPIPS | Audio rel. L2 | Audio cosine | Result |
|---|---|---|---:|---:|---:|---:|---:|---:|---|
| `LTX2Pipeline` | `TI2VidOneStagePipeline` Full/SFT | T2V | 0.931155 | 0.915337 | 24.65 dB | 0.044330 | 0.55994 | 0.848014 | Failed |
| `LTX2Pipeline` | `TI2VidOneStagePipeline` Full/SFT | I2V | 0.990583 | 0.986151 | 38.09 dB | 0.005834 | 0.89176 | 0.605688 | Failed (audio) |
| `LTX2TwoStagePipeline` | `TI2VidTwoStagesPipeline` Full + LoRA 450 | T2V | 0.730983 | 0.658280 | 19.28 dB | 0.124694 | 1.18078 | 0.411082 | Failed |
| `LTX2TwoStagePipeline` | `TI2VidTwoStagesPipeline` Full + LoRA 450 | I2V | 0.954715 | 0.934198 | 28.58 dB | 0.036060 | 0.53330 | 0.856538 | Failed |
| `LTX2DistilledTwoStagePipeline` | `DistilledPipeline` | T2V | 0.662850 | 0.654343 | 17.81 dB | 0.369542 | 2.59623 | 0.140727 | Failed |
| `LTX2DistilledTwoStagePipeline` | `DistilledPipeline` | I2V | 0.926517 | 0.876693 | 27.38 dB | 0.043394 | 1.14267 | 0.418109 | Failed |

The full gates are SSIM mean/min >= 0.95/0.90, PSNR >= 30 dB, LPIPS <= 0.01 (T2V) or 0.02 (I2V), audio relative L2 <= 0.3, and cosine >= 0.95. The distilled gates are stricter: mean SSIM >= 0.995 (T2V) or 0.99 (I2V), min SSIM >= 0.99, PSNR >= 40 dB, LPIPS <= 0.01, with the same audio gates.

These L20X/cu129 results do not reproduce the [author's B300/cu130 parity matrix](https://github.com/vllm-project/vllm-omni/pull/6070#issuecomment-5285202605). Hardware/software differ from that report, but each local native/Omni pair used the same device, stack, weights, seed, backend, and conditioning; the gaps are too large to classify as ordinary tolerance noise. The checked-in harness should be reconciled or the hardware-specific numerical contract documented before using it as permanent CI evidence.

Permanent paths at the tested head:

- [`test_ltx25_official_similarity.py`](https://github.com/vllm-project/vllm-omni/blob/e2721dc97d48a412128327039b2a80bb07c5b0ea/tests/e2e/accuracy/ltx/test_ltx25_official_similarity.py)
- [`run_ltx25_reference.py`](https://github.com/vllm-project/vllm-omni/blob/e2721dc97d48a412128327039b2a80bb07c5b0ea/tests/e2e/accuracy/ltx/run_ltx25_reference.py)

### Focused tests and runnable public examples

```bash
CUDA_VISIBLE_DEVICES='' HF_HOME=/workspace/model python -m pytest -q \
  tests/diffusion/models/ltx2/test_ltx25_pipeline.py
```

Result: **36 passed in 2.70 s**.

I also ran the four documented T2V pipeline classes at their class-specific dimensions/steps and one first-frame I2V command. Each isolated command exited 0 and encoded H.264 video at 24 FPS plus 48 kHz stereo AAC:

| Public pipeline | Task | Dimensions / frames / steps | Generation | Peak reserved GPU | Encoded output |
|---|---|---|---:|---:|---|
| `LTX2Pipeline` | T2V | 960x544 / 121 / 30 | 77.00 s | 82,168 MiB | 5.042 s, 4,364,209 bytes |
| `LTX2TwoStagePipeline` | T2V | 1920x1088 / 121 / 30 | 92.24 s | 120,618 MiB | 5.042 s, 11,494,223 bytes |
| `LTX2DistilledOneStagePipeline` | T2V | 960x544 / 121 / 8 | 9.42 s | 78,666 MiB | 5.042 s, 2,940,282 bytes |
| `LTX2DistilledTwoStagePipeline` | T2V | 1920x1088 / 121 / 8 | 23.58 s | 111,716 MiB | 5.042 s, 9,125,194 bytes |
| `LTX2DistilledTwoStagePipeline` | I2V | 1920x1088 / 121 / 8 | 24.76 s | — | 5.042 s, 1,799,839 bytes |

Exact T2V commands used (substitute an available GPU):

```bash
python examples/offline_inference/text_to_video/text_to_video.py \
  --model Lightricks/LTX-2.5-Diffusers --model-class-name LTX2Pipeline \
  --prompt "A cinematic red fox walking through a snowy forest at dawn, tracked alongside." \
  --height 544 --width 960 --num-frames 121 --num-inference-steps 30 \
  --frame-rate 24 --fps 24 --enforce-eager --output ltx25-full-one-stage.mp4

python examples/offline_inference/text_to_video/text_to_video.py \
  --model Lightricks/LTX-2.5-Diffusers --model-class-name LTX2TwoStagePipeline \
  --prompt "A cinematic red fox walking through a snowy forest at dawn, tracked alongside." \
  --height 1088 --width 1920 --num-frames 121 --num-inference-steps 30 \
  --frame-rate 24 --fps 24 --enforce-eager --output ltx25-full-two-stage.mp4

python examples/offline_inference/text_to_video/text_to_video.py \
  --model Lightricks/LTX-2.5-Diffusers --model-class-name LTX2DistilledOneStagePipeline \
  --prompt "A cinematic red fox walking through a snowy forest at dawn, tracked alongside." \
  --height 544 --width 960 --num-frames 121 --num-inference-steps 8 \
  --frame-rate 24 --fps 24 --enforce-eager --output ltx25-distilled-one-stage.mp4

python examples/offline_inference/text_to_video/text_to_video.py \
  --model Lightricks/LTX-2.5-Diffusers --model-class-name LTX2DistilledTwoStagePipeline \
  --prompt "A cinematic red fox walking through a snowy forest at dawn, tracked alongside." \
  --height 1088 --width 1920 --num-frames 121 --num-inference-steps 8 \
  --frame-rate 24 --fps 24 --enforce-eager --output ltx25-distilled-two-stage.mp4
```

For I2V I used `examples/offline_inference/image_to_video/image_to_video.py` with the same class-specific dimensions/steps and `--image /absolute/path/to/first-frame.png`. The full command set is in [`recipes/LTX/LTX-2.5.md`](https://github.com/vllm-project/vllm-omni/blob/e2721dc97d48a412128327039b2a80bb07c5b0ea/recipes/LTX/LTX-2.5.md).

Bottom line: the public pipelines are runnable locally on L20X and produce valid AV files, while the official-reference numerical parity claimed by the new E2E harness is not reproduced in this environment.

tests/e2e/accuracy/ltx/test_ltx25_official_similarity.py
lishunyang12: Resolved at current head `8ce11ef77`.

The `e2721dc97` harness selected generic PyTorch SDPA independently in the two runtimes; the `torch_sdpa` metadata did not pin the same kernel. Kernel selection differed by shape/runtime on L20X, which explains the trajectory-scale mismatch. The permanent guard now selects BF16 cuDNN explicitly on both sides: native uses `PytorchAttention(priority=[CUDNN_ATTENTION, MATH])`, while Omni uses `CUDNN_ATTN`; MATH remains only the one-token fallback.

A second issue affected Full two-stage I2V: multiplying Python `noise_scale=0.909375` by a BF16 conditioning mask rounded it early to `0.91015625`. The mask arithmetic now stays FP32. After the fix, native and Omni are bitwise exact at Stage 1 output, Stage 2 input, and Stage 2 final packed latent.

Current B300/cu130 BF16-cuDNN decoded parity for all six public native T2V/I2V paths is:

| Path | Task | SSIM mean/min | PSNR | LPIPS |
|---|---|---:|---:|---:|
| Full one-stage | T2V | 0.999652 / 0.999564 | 58.80 dB | 0.0000367 |
| Full one-stage | I2V | 0.999589 / 0.999480 | 62.37 dB | 0.0000576 |
| Full two-stage | T2V | 0.999588 / 0.999485 | 56.40 dB | 0.0000687 |
| Full two-stage | I2V | 0.999585 / 0.999475 | 62.10 dB | 0.0000627 |
| Distilled two-stage | T2V | 0.997186 / 0.996164 | 46.04 dB | 0.003450 |
| Distilled two-stage | I2V | 0.996489 / 0.995810 | 46.29 dB | 0.003787 |

The thresholds now require SSIM mean/min >= 0.99 and PSNR >= 40 dB on every public path. The focused affected CPU suite is `111 passed`. Please rerun the pinned test on current head; it now exercises an explicit cross-runtime cuDNN contract instead of architecture-dependent generic SDPA dispatch.

vllm_omni/diffusion/models/ltx2/ltx2_components.py
```diff
@@ -73,10 +81,15 @@ class LTXComponentProfile:
...
     vocoder_cls: type = LTX2Vocoder
+    text_encoder_cls: type | None = Gemma3ForConditionalGeneration
     vocoder_fallback_cls: type | None = None
     artifact_repo_id: str | None = None
     latent_upsampler_filename: str | None = None
     distilled_lora_filename: str | None = None
+    transformer_subfolder: str = "transformer"
+    scheduler_use_dynamic_shifting: bool = False
+    scheduler_shift_terminal: float | None = None
+    checkpoint_kind: LTXCheckpointKind | None = None
```
mglyn: Could we derive the expected checkpoint kind from `pipeline_kind` instead of storing it in every component profile? The base LTX-2/LTX-2.3 one-stage profiles currently leave this as `None`, so `validate_ltx_checkpoint()` returns early and `LTX2Pipeline` can accept a merged-distilled checkpoint while executing the regular CFG/STG recipe. A single mapping (`one_stage`/`two_stage` -> regular, distilled entries -> distilled, DMD2 -> unchecked) removes duplicate state and closes that gap.

vllm_omni/diffusion/data.py
```diff
@@ -1210,9 +1213,9 @@ def enrich_config(self) -> None:
                             exc,
                         )
                 else:
-                    tf_config_dict = get_hf_file_to_dict("transformer/config.json", self.model)
+                    tf_config_dict = get_hf_file_to_dict("transformer/config.json", self.model, revision=self.revision)
```
mglyn: LTX-2.5 Full constructs its actual DiT from `transformer_full/config.json`, but the worker-level `tf_model_config` is still initialized from `transformer/config.json` and consumed before pipeline construction. The current BF16 checkpoints may share the same structure, but serialized-quant detection and any feature reading `hf_config` can bind to the distilled config. Can the selected transformer subfolder be reflected in config discovery, or can unsupported combinations be rejected explicitly?

vllm_omni/diffusion/models/ltx2/ltx2_runtime.py
```diff
@@ -700,6 +733,9 @@ def _predict_noise_for_step(
             state,
             forward_ctx,
             denoise_ctx,
+            preserve_positive_velocity=(
+                getattr(forward_ctx.video_audio_step_adapter, "_sampler", "euler") == "euler_ancestral"
```
mglyn: Can we avoid reaching into the adapter's private `_sampler` field here? This makes guidance behavior depend on a concrete scheduler adapter implementation. Carrying a public `sampler` or `preserve_positive_velocity` semantic in `LTXForwardContext`, populated from the phase recipe, would keep the sampler/adapter boundary explicit.

tests/e2e/accuracy/ltx/test_ltx25_official_similarity.py
```diff
@@ -0,0 +1,888 @@
...
+
+    frames, actual_audio, sample_rate = _unwrap_omni_output(output)
+
+    assert frames is frame
+    assert actual_audio is audio
+    assert sample_rate == 48_000
+
+
+@dataclass(frozen=True)
+class LTXAccuracyCase:
```
mglyn: This `LTXAccuracyCase` block and its LTX-2/2.3 helper chain (`CASES`, `_resolve_model`, `_resolve_checkpoint`, `_resolve_image`, `_request`, and the old `_official_source`) appear unused in this file. The active 2/2.3 coverage already lives in `test_ltx_official_similarity.py`. Removing the copied dead harness saves roughly 180 lines without changing coverage.

vllm_omni/diffusion/models/ltx2/ltx2_conditioning.py
```diff
@@ -82,6 +82,45 @@ def _preprocess_i2v_pil_images(
...
+    try:
+        import av
+    except ImportError as exc:
+        raise ImportError("PyAV with a libx264 encoder is required for LTX-2.5 image conditioning (CRF 18).") from exc
+
+    with BytesIO() as output_file:
+        container = av.open(output_file, "w", format="mp4")
+        try:
+            stream = container.add_stream("libx264", rate=1, options={"crf": str(crf), "preset": "veryfast"})
+            height = image.shape[0] // 2 * 2
```
mglyn, line 100: The pinned native preprocessing returns the original image when either dimension is below 2. Here a 1-pixel dimension is rounded down to zero before configuring libx264. Please mirror the native `min(height, width) < 2` guard before the H.264 round trip.

mglyn (commented):
small correctness and complexity findings

tests/e2e/accuracy/ltx/run_ltx25_reference.py
```diff
@@ -0,0 +1,559 @@
...
+            sys.path.insert(0, path)
+
+
+def _configure_official_sdpa(pipeline: Any) -> None:
+    """Pin official connector and denoiser attention to cuDNN, with MATH fallback."""
+    from ltx_core.loader.attention_ops import set_attention_module_op
+    from ltx_core.model.transformer.attention import PytorchAttention
+    from torch.nn.attention import SDPBackend
+
+    attention = PytorchAttention(priority=[SDPBackend.CUDNN_ATTENTION, SDPBackend.MATH])
```
mglyn, line 123: Could we pin the process-wide SDPA state here in the same way as the existing LTX-2.3 accuracy guard (`torch.backends.cuda.enable_cudnn_sdp(False)` before model construction)? This module-op override covers the connector and DiT paths, but not Gemma/the text encoder. Since importing vLLM changes the global CUDA SDP state while the standalone official process may retain a different state, H100/H200 may route Gemma attention through different implementations, creating a reproducibility risk for these strict parity thresholds. Would mirroring the LTX-2.3 global pin so both runners use the same process-wide state make the test conditions more reproducible?

.buildkite/cuda/test-nightly.yml
```diff
@@ -487,6 +493,13 @@ steps:
           - pytest -s -v 'tests/e2e/accuracy/ltx/test_ltx_official_similarity.py::test_ltx_matches_official[ltx23_two_stage_layer_fused_i2v]' -m full_model --run-level full_model
         mirror_hardwares: h100_1
 
+      - label: ":full_moon: Diffusion X2V · LTX-2.5 Official Parity Test"
+        timeout_in_minutes: 180
+        commands:
+          - pytest -s -v tests/e2e/accuracy/ltx/test_ltx25_official_similarity.py::test_ltx25_distilled_two_stage_matches_official -m full_model --run-level full_model
+          - pytest -s -v tests/e2e/accuracy/ltx/test_ltx25_official_similarity.py::test_ltx25_full_matches_official -m full_model --run-level full_model
+        mirror_hardwares: h100_1
```
mglyn, line 501: Could we make a reduced CI resolution together with layerwise offload the preferred configuration for these two-stage cases, similar to the existing LTX-2.3 accuracy guard? Both variants load the x2 upsampler, and the Full path additionally materializes the large refinement LoRA, so their representative 1080p configurations exceed a single 80 GB H100.
The current tests appear to already use reduced shapes and pass --enable-layerwise-offload; could we verify and record the actual peak memory under this exact CI configuration to ensure sufficient H100 headroom?
If that is still insufficient, moving the cases to a higher-memory runner or removing them from this job would be preferable. TP or VAE tiling could be fallback options, but both materially change numerical behavior, so they would likely require rebaselined and relaxed accuracy thresholds.

mglyn (commented):
Some reasons that can cause accuracy to be unreproducible.

vllm_omni/diffusion/models/ltx2/ltx2_transformer.py
lishunyang12: Confirmed on pushed head `f7e2d6836`. The routing now matches the pinned native implementation: video scale/shift uses the per-token video timestep, A2V gate uses audio sigma, audio scale/shift uses audio timestep, and V2A gate uses video sigma. The requested current-head B300 `ltx2_3_i2v` run completed against official source `9377758131b1ffde4b7f766804590a6617bf2ab9`: video SSIM mean/min `0.992836/0.989994` and PSNR `40.66 dB`, passing its `0.95/0.90/30 dB` video gates. The overall legacy case still reports its independent pre-existing audio mismatch (`cosine=0.8306`); the I2V video/modulation portion requested here passes. `test_ltx_cross_attention_modulations_use_official_timesteps` guards the same official routing for LTX-2, which has no public I2V accuracy case.

tests/e2e/accuracy/ltx/test_ltx25_official_similarity.py
lishunyang12: Reproduced after the fixes on pushed head `f7e2d6836`. The exact B300 BF16-cuDNN distilled two-stage T2V test passed (`1 passed in 239.28s`): SSIM mean/min `0.997186/0.996164`, PSNR `46.04 dB`, LPIPS `0.003452`, audio relative L2/cosine `0.307041/0.954879`. The runner now also pins process-wide Gemma SDPA state before both official and Omni construction while keeping DiT explicitly on cuDNN with the one-token MATH fallback.

vllm_omni/diffusion/models/ltx2/ltx2_components.py
lishunyang12: Fixed in `cfee8a9e9`. `checkpoint_kind` was removed from component profiles and `resolve_ltx_checkpoint_kind()` now derives the contract from `pipeline_kind`: regular for one/two-stage, distilled for both distilled modes, and unchecked only for DMD2. This also closes the LTX-2/LTX-2.3 base-profile early-return gap; unit coverage includes every mapping and unknown-kind rejection.

vllm_omni/diffusion/data.py
lishunyang12: Fixed in `cfee8a9e9` without adding LTX-specific branching to common config code. A generic model-extras transformer-config resolver now selects the declared subfolder; only Full `LTX2Pipeline`/`LTX2TwoStagePipeline` register the LTX resolver, which returns `transformer_full` for 2.5 and `transformer` for 2/2.3. Distilled classes retain the standard `transformer` default. The enrichment regression writes distinct markers into both configs and verifies Full binds to `transformer_full`.

vllm_omni/diffusion/models/ltx2/ltx2_runtime.py
lishunyang12: Fixed in `cfee8a9e9`. `LTXForwardContext` now exposes public `sampler`, populated directly from the phase recipe, and runtime guidance uses `forward_ctx.sampler == "euler_ancestral"`. No runtime code reaches into the adapter private `_sampler` field.

tests/e2e/accuracy/ltx/test_ltx25_official_similarity.py
lishunyang12: Fixed in `cfee8a9e9`. Removed the unused copied LTX-2/2.3 case dataclass, cases, source/model/checkpoint/image resolvers, and request builder from the LTX-2.5 accuracy file (196 deletions total in the review commit). Active LTX-2/2.3 coverage remains only in `test_ltx_official_similarity.py`.

vllm_omni/diffusion/models/ltx2/ltx2_conditioning.py
lishunyang12: Fixed in `cfee8a9e9`. `_apply_image_conditioning_crf` now returns the original uint8 RGB array before importing/configuring PyAV when either dimension is below 2. The regression uses a 1x4 image, makes `av` unavailable, and verifies identity preservation.

tests/e2e/accuracy/ltx/run_ltx25_reference.py
lishunyang12: Fixed in `cfee8a9e9`. Both `_run_official` and `_run_omni` call `torch.backends.cuda.enable_cudnn_sdp(False)` before model construction, so Gemma uses the same process-wide dispatch. Connector/DiT remain independently pinned to explicit BF16 cuDNN, with MATH only for degenerate one-token attention. The current distilled two-stage B300 parity run passes the strict gates.

.buildkite/cuda/test-nightly.yml
lishunyang12: Addressed in `cfee8a9e9`. The nightly cases already use reduced exact-CI shapes (distilled 1024x576x25, Full 512x384x25) plus `--enable-layerwise-offload`. The harness now records in-process official allocated/reserved peaks and the existing diffusion-worker-reported Omni peak in every metrics artifact (rather than incorrectly sampling the lightweight parent). On B300, the exact Full two-stage I2V CI request reports `39,034 MiB` Omni worker peak, versus the previous 225+ GiB no-offload stress configuration. The H100 nightly run will now emit its own actual worker value in the permanent artifact.

vllm_omni/diffusion/models/ltx2/ltx2_components.py
```diff
@@ -206,15 +294,15 @@ def resolve_ltx_component_profile(pipeline_kind: str, model_version: str) -> LTX
...
     if os.path.isdir(model):
         path = os.path.join(model, filename)
         if not os.path.isfile(path):
             return {}
     else:
         try:
-            path = hf_hub_download(repo_id=model, filename=filename)
+            path = hf_hub_download(repo_id=model, filename=filename, revision=revision)
         except Exception:
             return {}
```
hsliuustc0106, line 307: The blanket `except Exception` here silently degrades version detection to `"2"` on a transient hub error (network blip, partial cache). That is not just cosmetic: detection feeds `should_preserve_reference_image_size` and the DiT config subfolder (`transformer_full` vs `transformer`), so a one-off download failure at startup can flip the resize semantics and load the wrong transformer config for the lifetime of the process — and `serving_video.py` caches the detection in a `cached_property`, so it never recovers. Could we at least `logger.debug/warn` the swallowed exception so this is diagnosable in the wild? Narrowing to `(OSError, requests exceptions, EntryNotFoundError)` would also keep genuinely-missing metadata on the fast path.

hsliuustc0106 (approved):

Issue #1867, opened 2026-03-13
[RFC]: Multi-Stage KV Cache Management Roadmap


Vivo50E:
### Motivation.

This document serves as a progress summary and tracking hub for [#1150 - Support KV Cache CPU Offloading](https://github.com/vllm-project/vllm-omni/issues/1150). Issue #1150 proposed integrating vLLM v1's native `MultiConnector` (`OffloadingConnector` + `LMCacheConnectorV1`) into each AR stage for KV cache CPU offloading, and outlined three Future Work items. This document consolidates the current status of all work items, associated PRs/Issues, and serves as the unified entry point for ongoing development.

### Relationship to Existing Roadmaps

This tracking document complements — but does not duplicate — the following broader roadmaps:

- [#677 - vLLM-Omni 2026 Q1 Roadmap](https://github.com/vllm-project/vllm-omni/issues/677) — mentions KV offloading (#1150) and P/D disaggregation (#1303) as line items without sub-task breakdown
- [#1192 - Omni Connector Full Disaggregation Q1 Roadmap](https://github.com/vllm-project/vllm-omni/issues/1192) — covers P/D split and Mooncake transport at high level (1/11 sub-issues done), but does not address MultiConnector, OffloadingConnector, LMCache, HBM coordination, or diffusion KV reuse

Neither the vllm-omni community nor upstream vLLM has a dedicated KV cache management tracking issue. This document fills that gap by providing cross-cutting dependency tracking, sub-task breakdown, and blocking bug visibility across Phase 1 and all three Future Work streams.

---

## Architecture Overview

```
┌─────────────────────────────────────────────────────────┐
│                 Per-Stage Layer (vLLM Native)            │
│  MultiConnector                                         │
│   ├── OffloadingConnector  (GPU ↔ CPU swap)             │
│   ├── LMCacheConnectorV1   (prefix cache / compression) │
│   └── MooncakeConnector    (RDMA cross-node, future)    │
├─────────────────────────────────────────────────────────┤
│              Cross-Stage Layer (vllm-omni Custom)        │
│  OmniConnector                                          │
│   ├── ChunkTransferAdapter       (hidden states)        │
│   └── OmniKVTransferManager      (cross-stage KV)      │
│        Backends: SharedMemory / Mooncake / Yuanrong     │
└─────────────────────────────────────────────────────────┘
```

---

## Phase 1: Per-Stage KV Offloading (Core Capability)

| Status | Work Item | Description |
[...]

PR #6071, opened 2026-08-11
[Feature] Add static cross-stage HBM admission control

Haofan144:
## Purpose
This PR implements the static-budget sub-phase of cross-stage HBM coordination proposed in #1867: a per-stage HBM ceiling that bounds vLLM's KV-cache allocation, plus a scheduler-side guard that defers admission when that ceiling is exhausted rather than letting requests queue indefinitely or the allocator overshoot.

The changes include:
- Per-stage `hbm_limit_gb` / `hbm_reserved_gb` configuration in deployment YAML, propagated through Omni config and engine-argument layers.
- A pre-launch, per-device admission ledger (`stage_admission.py`) that sums each replica's total HBM envelope per physical GPU and rejects any layout that overcommits a device before any stage launches — covering single-GPU shared stages, multi-GPU separate stages, TP replicas, and mixed layouts.
- A worker-side KV budget derivation: `KV bytes = total budget − profiled non-KV bytes − reserve`, capped at any explicit `kv_cache_memory_bytes` when both are set.
- Extension of the same budget contract to paged-scheduler diffusion workers (dense diffusion and non-AR generation workers are not supported yet).
- Unit tests covering the budget arithmetic, the device ledger, and worker profiling.

This PR intentionally introduces **no new scheduler-side admission gate**. Native `Scheduler.schedule()` and `KVCacheManager.allocate_slots()` keep their existing watermark, preemption, and prefix-reuse behavior unchanged — this PR only constrains how much KV capacity they are handed. Dynamic memory monitoring, runtime budget rebalancing, and a scheduler-side pause/resume admission guard are out of scope for this PR and remain tracked in #1867.

### Component matrix

`experiments/static_hbm/run_experiments.py --gpu` runs the full validation matrix through production code paths: the real YAML loader, `_build_engine_args`, `check_admission`, and (for GPU cases) the real `OmniGPUWorkerBase.determine_available_memory` on a live model.

Configuration:
- Model: Qwen3-0.6B
- GPU: NVIDIA RTX A6000 48 GiB
- 21 existing unit tests, 4 layout ledgers (pass + over-budget reject each), a budget scan, and multi-rank KV cache config generation
- 5 GPU probes at concurrency 1, 8, 32 (511-token prompts, 64-token outputs, eager, FP16)

```bash
python3 experiments/static_hbm/run_experiments.py --gpu
```

### Same-GPU co-location

`experiments/static_hbm/run_shared.py` runs two independently budgeted Qwen3-0.6B engines on one GPU, synchronized to start concurrently.

```bash
python3 experiments/static_hbm/run_shared.py
```

Full logs and results: `experiments/static_hbm/results/`.

**vLLM Version:** 0.28.0

**vLLM-Omni Commit:** 6192fe71f3e66d6e88cec3e69207132f8314f332

### Component matrix: all cases passed

21 unit tests, 8 layout cases (4 pass + 4 correctly rejected as over-budget), the budget scan, and multi-rank block normalization all passed. All 5 GPU probes completed with zero failed requests and no OOM.

**Budget sizing (vs. unlimited baseline `native_original`):**

| Group | Budget GiB | KV GiB | Blocks | Alloc shortfalls (cumulative) | Preemptions | 32-req time (s) | Peak memory GiB |
|---|---:|---:|---:|---:|---:|---:|---:|
| static_3g | 3.0 | 1.19 | 696 | 63 | 2 | 2.487 | 2.94 |
| static_4g | 4.0 | 2.19 | 1282 | 0 | 0 | 1.407 | 3.97 |
| static_pressure | 3.0 (0.25 GiB cap) | 0.25 | 146 | 520 | 0 | 9.228 | 1.95 |
| native_equal | — (0.25 GiB cap) | 0.25 | 146 | 520 | 0 | 8.979 | 1.96 |
| native_original | — (auto) | auto | 3336 | 0 | 0 | 1.442 | 7.47 |

Results demonstrate:
- Measured KV bytes match the budget formula exactly: `KV = total − reserve − non_KV`, capped at any explicit KV limit.
- `static_pressure` and `native_equal` (same 256 MiB cap, one via `hbm_limit_gb`, one via native `kv_cache_memory_bytes`) match on blocks, output hashes, allocation shortfalls, and preemptions — confirming this PR changes only the capacity handed to the scheduler, not its behavior.
- Every case drained its queue with all requests completing their full output, including under pressure (520 shortfalls, 2 preemptions) — KV exhaustion is absorbed by the existing scheduler retry/preemption path, not by dropped requests.

### Same-GPU co-location: two budgeted engines coexist

Two engines, each budgeted at 3 GiB with a 256 MiB KV cap, ran concurrently with overlapping inference windows (26.6 s overlap). Both hit KV capacity pressure and both drained their queues successfully.

### Scope

Single-GPU only; no physical multi-GPU/TP, full Omni multi-stage startup, CUDA graph/offload, or repeated-load performance testing in this round. GPU cases are small-model, single-run mechanism checks (41 requests per case), not a production-scale benchmark.


hsliuustc0106:
is this PR still active? what's the use case of this feature?

Haofan144:
> is this PR still active? what's the use case of this feature?

Yes, still active.

The use case is what has been mentioned in https://github.com/vllm-project/vllm-omni/issues/1867,  vLLM-Omni often co-locates multiple stages (e.g. AR + diffusion) on the same GPU, and each of them profiles memory independently. There's currently no cross-stage awareness, so it's easy to overcommit a GPU and hit OOM at runtime instead of at launch.

This PR adds per-stage `hbm_limit_gb/hbm_reserved_gb` config and a pre-launch admission check that sums declared budgets per physical device and rejects overcommitted deployments before any process starts. It also makes KV-cache sizing respect the declared total instead of greedily grabbing free memory. Scope is intentionally narrow for this milestone (paged-scheduler AR/diffusion only, no new scheduler-side gate), and I'm also working on adding the dynamic monitoring in https://github.com/vllm-project/vllm-omni/issues/1867

PR #6064, opened 2026-08-11
[Perf][Diffusion] Avoid redundant MiniMax-H3 reference video scans

yeahdongcn:
## Purpose

Speed up MiniMax-H3 Ref2VA reference-video decoding while preserving the
lossless RGB preprocessing introduced by #5978.

#5978 correctly decodes the prepared video once instead of starting one ffmpeg
process per sampled frame. However, its PyAV fallback and unconditional
`ffprobe -count_frames` probe both scan the complete high-bitrate
`libx264rgb` stream for Qwen sampling and again for the video VAE.
The codec context is CPU-only (`is_hwaccel=False`, `hwaccel=None`) in the tested
environment.

This change:

- decodes only Qwen's requested frame indices through one ffmpeg raw-video
  pipe;
- uses the same ffmpeg raw-video path for the full-frame fallback when decord
  is unavailable;
- uses container-provided `nb_frames` metadata without a full scan, with a
  count-frames fallback only when that metadata is absent;
- validates the exact expected RGB byte count before reshaping the result.

The implementation is platform independent and does not add a MUSA-specific
execution path.

## Correctness

On the real #5978 prepared reference video (97 frames, 1344x768), both the
selective and full ffmpeg outputs are pixel-exact against PyAV's RGB24 output.

- selective indices: `0, 12, ..., 96` (9 frames)
- full comparison: all 97 frames
- upstream-main targeted tests: 12 passed
- pre-commit on both changed files: passed
- Ref2VA TP8 E2E: two HTTP-200 requests, 28.74s and 24.67s

## Performance

CPU decode microbenchmark on the same prepared video:

| path | latency |
|---|---:|
| #5978 PyAV/metadata path | 8.84s + 9.06s |
| ffmpeg selective decode (9 frames) | 0.85s |
| ffmpeg full decode (97 frames) | 1.02s |

The PyAV/metadata measurements were collected on the 32.88 host; the direct
ffmpeg pixel-equivalence measurements and the E2E table below were collected
on the 34.79 host. E2E comparisons always use the same host and request
matrix for both candidates.

Ref2VA E2E on 8 x MTT S5000, 768x432, 4 seconds, 24 FPS, seed 0,
4 denoising steps:

| candidate | request latency |
|---|---:|
| #5978 + #5937 stack before this PR | 51.59s, 47.98s |
| this PR + #6058 v0.24 port | 28.74s, 24.67s |

On the same 34.79 TP8 host, this is a 46.4% reduction versus the unmodified
[...]

PR #5978, opened 2026-08-10
[CI] Align MiniMax H3 Ref2VA inputs and add I2VA/Ref2VA accuracy coverage

david6666666:
## Purpose

Align MiniMax H3 Ref2VA model-card input handling and keep reproducible official-asset accuracy coverage for I2VA and Ref2VA.

- Bound reference video and audio conditions to the generated temporal window.
- Preserve RGB reference pixels losslessly through preprocessing.
- Normalize audio timestamps so muxed AAC does not expose encoder priming as leading silence.
- Expose standalone audio encoding in the MiniMax H3 profiler.
- Add SHA-pinned model-card assets and shared metadata/audio/video similarity checks for I2VA and Ref2VA.
- Use one accuracy gate for both retained cases: SSIM >= 0.97 and PSNR >= 34 dB.
- Remove the obsolete T2VA accuracy case and update the CUDA nightly path.

## Test Plan

**vLLM Version:** 0.26.0

**Hardware:** 4x NVIDIA B300, BF16, Ulysses 4, ring 1, HSDP shard size 4, VAE patch parallel 4, tiled VAE, FlashAttention, 50 steps, seed 0.

image

- `pre-commit run --files .buildkite/cuda/test-nightly.yml tests/e2e/accuracy/minimax_h3/test_minimax_h3_i2va_ref2va_similarity.py`
- `pytest --collect-only -q tests/e2e/accuracy/minimax_h3/test_minimax_h3_i2va_ref2va_similarity.py` (2 tests collected)
- Contract/API validation: 167 passed.
- I2VA: SSIM `0.975190`, PSNR `35.981286 dB`, passing the common gate.
- Ref2VA: SSIM `0.980577`, PSNR `42.650204 dB`, passing the common gate.
- Both retained cases match the official 1344x768, 24 FPS, AAC 32 kHz stereo contract; frame counts are 192 and 124 respectively.

The T2VA case was removed from the retained accuracy/nightly coverage because the current strict gate was not met; this PR does not claim T2VA pixel-level alignment.

PR #6072, opened 2026-08-12
[Bugfix][Diffusion] Fix MiniMax-H3 model-level CPU offload residency

yeahdongcn:
## Summary

- Formalize the custom-pipeline model CPU-offload hook as a vendor-neutral protocol.
- Discover MiniMax-H3 DiT, encoder, and VAE components through `ModuleDiscovery`.
- Keep DiT off the accelerator before direct VAE calls and preserve exception-safe cleanup.
- Keep image-only references in one Video-VAE residency window.
- Keep mixed image + video references in one Video-VAE residency window, avoiding an extra full Video-VAE H2D/D2H cycle.
- Keep embedded-video audio and standalone audio references in one Audio-VAE residency window.

## Why

MiniMax-H3 invokes VAE stages directly from the pipeline. With
`--enable-cpu-offload`, those calls can bypass normal module traversal. This
change makes the direct-call offload lifecycle explicit and keeps each VAE
resident across all consecutive direct calls belonging to the same request.

This is complementary to #5764 and does not change the DLO plan. #5978 and
#6064 are orthogonal reference-video decoding changes.

### Focused checks

- Candidate MiniMax-H3 offload tests on the exact image: **11 passed in 0.25 s**.
- Added profiler-target regression coverage for the aggregate visual/audio
  reference helpers.
- `py_compile`, Ruff, format, and `git diff --check`: passed.

### MUSA transfer-level A/B

Hardware/software:

- 1 x MTT S5000, driver `3.3.5-server`.
- Image: `registry.mthreads.com/mcconline/inference/vllm-omni:minimax-h3-20260815`.
- Image digest: `sha256:23ae27867cd19ce848a27688dba262d0569c67ff3c3b599cc2a429f8ab184a8b`.
- PyTorch `2.11.0.post1+musa5.2.0`, one visible MUSA device.
- A/B baseline source before this review commit:
  `811d08ce7369e6b856f6bbe111b87c2f1edba58a`.

The current head and candidate used the same container and synthetic 32 MiB
Video-VAE module. Encoder output was stubbed so this measurement isolates
residency and parameter movement.

| Input | Path | VAE scopes | H2D | D2H |
| 1 image | current head | 1 | 32 MiB | 32 MiB |
| 1 image | candidate | 1 | 32 MiB | 32 MiB |
| 2 images | current head | 1 | 32 MiB | 32 MiB |
| 2 images | candidate | 1 | 32 MiB | 32 MiB |
| 2 images + 1 video | current head | 2 | 64 MiB | 64 MiB |
| 2 images + 1 video | candidate | 1 | 32 MiB | 32 MiB |

The candidate removes one complete Video-VAE residency lifecycle from the
mixed image + video path. Image-only single- and multi-reference inputs retain
one lifecycle. This is transfer-level evidence; it is not a claim of a full
model peak-VRAM reduction or a full MiniMax-H3 E2E quality benchmark.

## Scope

Single-image and image-only multi-reference requests retain one VAE residency
window. The shared visual scope applies when image and video references are
combined. The shared audio scope applies when embedded-video and standalone
audio references occur in the same request. No vendor-specific device logic is
introduced.

### DIFF
--- a/vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
+++ b/vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
@@ -8,7 +8,7 @@ import math
 import os
 import tempfile
 from collections.abc import Iterable, Mapping, Sequence
-from contextlib import contextmanager
+from contextlib import contextmanager, nullcontext
 from itertools import groupby
 from pathlib import Path
 from typing import Any, ClassVar
@@ -50,6 +50,7 @@ from vllm_omni.diffusion.offloader import (
     remove_sequential_offload,
     sequential_offload_component,
 )
+from vllm_omni.diffusion.offloader.module_collector import ModuleDiscovery
 from vllm_omni.diffusion.profiler.diffusion_pipeline_profiler import (
     DiffusionPipelineProfilerMixin,
 )
@@ -548,9 +549,8 @@ class MiniMaxH3Pipeline(
     _PROFILER_TARGETS: ClassVar[list[str]] = [
         "_prepare_reference_videos",
         "encode_prompt",
-        "_encode_video_conditions",
-        "_encode_video_audio_conditions",
-        "_encode_audio_conditions",
+        "_encode_visual_conditions",
+        "_encode_reference_audio_conditions",
         "diffuse",
         "decode",
     ]
@@ -1163,12 +1163,6 @@ class MiniMaxH3Pipeline(
             or getattr(od_config, "enable_distributed_layerwise_offload", False)
         )
 
-    def _model_cpu_offload_dits(self) -> list[nn.Module]:
-        return [module for module in (self.transformer, getattr(self, "transformers_ref", None)) if module is not None]
-
-    def _model_cpu_offload_stages(self) -> list[nn.Module]:
-        return [self.text_encoder, self.video_vae, self.audio_vae]
-
     def enable_omni_model_cpu_offload(
         self,
         *,
@@ -1179,8 +1173,9 @@ class MiniMaxH3Pipeline(
         if getattr(self, "_model_cpu_offload_modules", None):
             return
 
-        dits = self._model_cpu_offload_dits()
-        stages = self._model_cpu_offload_stages()
+        components = ModuleDiscovery.discover(self)
+        dits = components.dits
+        stages = [*components.encoders, *components.vaes]
         modules = [*dits, *stages]
         apply_sequential_offload(
             dit_modules=dits,
@@ -1228,27 +1223,31 @@ class MiniMaxH3Pipeline(
     ) -> tuple[torch.Tensor | None, list[tuple[int, int, int]]]:
         rows: list[torch.Tensor] = []
         shapes: list[tuple[int, int, int]] = []
-        if images:
-            _, rank, _ = _dit_rank_world()
-            image_rows = None
-            if rank == 0:
-                with self._component_on_device(self.video_vae):
+        _, rank, _ = _dit_rank_world()
+        # Keep image and video references in one residency window when both
+        # appear in a request; otherwise the video branch would reload the VAE.
+        needs_video_vae = video_count > 0 or (rank == 0 and bool(images))
+        video_vae_context = self._component_on_device(self.video_vae) if needs_video_vae else nullcontext()
+        with video_vae_context:
+            if images:
+                image_rows = None
+                if rank == 0:
                     image_rows = torch.cat([self.video_vae.encode_image(image) for image in images])
-            rows.append(
-                _broadcast_tensor(
-                    image_rows,
-                    dtype=torch.float32,
-                    device=self.device,
+                rows.append(
+                    _broadcast_tensor(
+                        image_rows,
+                        dtype=torch.float32,
+                        device=self.device,
+                    )
                 )
-            )
-            shapes.extend((1, image.height // 16, image.width // 16) for image in images)
-        if video_count:
-            video_rows, video_shapes = self._encode_video_conditions(
-                prepared_videos,
-                count=video_count,
-            )
-            rows.append(video_rows)
-            shapes.extend(video_shapes)
+                shapes.extend((1, image.height // 16, image.width // 16) for image in images)
+            if video_count:
+                video_rows, video_shapes = self._encode_video_conditions_resident(
+                    prepared_videos,
+                    count=video_count,
+                )
+                rows.append(video_rows)
+                shapes.extend(video_shapes)
         return (torch.cat(rows) if rows else None), shapes
 
     def _encode_audio_condition(
@@ -1277,6 +1276,20 @@ class MiniMaxH3Pipeline(
         return rows, int(audio_t_tensor.item())
 
     def _encode_audio_conditions(
+        self,
+        audios: list[tuple[torch.Tensor, int]],
+        *,
+        max_duration_seconds: float | None = None,
+    ) -> tuple[torch.Tensor | None, list[int]]:
+        if not audios:
+            return None, []
+        with self._component_on_device(self.audio_vae):
+            return self._encode_audio_conditions_resident(
+                audios,
+                max_duration_seconds=max_duration_seconds,
+            )
+
+    def _encode_audio_conditions_resident(
         self,
         audios: list[tuple[torch.Tensor, int]],
         *,
@@ -1419,6 +1432,34 @@ class MiniMaxH3Pipeline(
         return (
             _broadcast_tensor(rows, dtype=torch.float32, device=self.device),
             [int(value) for value in lengths.tolist()],
+        )
+
+    def _encode_reference_audio_conditions(
+        self,
+        prepared_videos: list[dict[str, Any]] | None,
+        *,
+        has_audio: list[bool],
+        standalone_audios: list[tuple[torch.Tensor, int]],
+        max_duration_seconds: float,
+    ) -> tuple[torch.Tensor | None, list[int], torch.Tensor | None, list[int]]:
+        # Embedded and standalone audio are consecutive direct Audio-VAE
+        # calls. Keep the component resident across both paths.
+        needs_audio_vae = any(has_audio) or bool(standalone_audios)
+        audio_vae_context = self._component_on_device(self.audio_vae) if needs_audio_vae else nullcontext()
+        with audio_vae_context:
+            embedded_condition, embedded_lengths = self._encode_video_audio_conditions_resident(
+                prepared_videos,
+                has_audio=has_audio,
+            )
+            external_condition, external_lengths = self._encode_audio_conditions_resident(
+                standalone_audios,
+                max_duration_seconds=max_duration_seconds,
+            )
+        return (
+            embedded_condition,
+            embedded_lengths,
+            external_condition,
+            external_lengths,
         )
 
     def _initial_noise(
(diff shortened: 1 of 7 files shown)


hsliuustc0106:
Recommended verdict: request changes for validation evidence. The implementation itself looks functionally sound.

The remaining gap is performance evidence: this `[Perf]` PR reports the patched MUSA smoke, but does not include a baseline-vs-PR comparison for latency, peak VRAM, or output quality. Please add an A/B table using identical model, inputs, resolution, steps, hardware, and warmup methodology—ideally mean ± standard deviation over at least three runs. Alternatively, narrow the PR title/claims to a functional offload fix.

Local validation on the PR head (`a2c7826adbaebe5f4c890a62c9e68ef1db41e401`):

- Focused MiniMax/offloader tests: **118 passed**.
- Broader MiniMax/offloader tests: **146 passed, 3 skipped**; one H100-marked FP8 quality test could not start on the local L20X hardware and is classified as an environment mismatch.
- Real MiniMax-H3 CPU-offload smoke on 1× NVIDIA L20X: valid finite video `(107, 256, 448, 3)` and audio `(1, 2, 142400)`.
- Nested regional-compile + top-level offload CUDA smoke: passed.

The real smoke’s shutdown exceeded the existing 30-second orchestrator timeout, but the worker exited and GPU memory returned to zero; I did not treat this as a PR correctness failure.

yeahdongcn:
@hsliuustc0106 I ran a comparison today:

1. This optimization is only effective when multiple input images are provided. For a single image, both implementations use the same single Video-VAE offload scope, so no meaningful gain is expected.
2. Measured on the same Ref2VA setup (TP8/TE8/VAE8) with two 1280×704 input images:

Metric | Pre-#6072 | #6072 candidate | Reduction
-- | -- | -- | --
Engine time | 83.577 s | 82.235 s | 1.342 s (1.61%)
Client time | 83.676 s | 82.336 s | 1.341 s (1.60%)

The gain comes from avoiding repeated Video-VAE H2D/D2H transfers for each input image.

hsliuustc0106:
Thanks for adding the two-image comparison—the ~1.6% latency reduction supports the intended multi-image VAE residency optimization.

To close the remaining [Perf] evidence gap, please add:

- exact hardware/software/model/config and commands for both the pre-#6072 and candidate runs;
- warmup and repeat counts, with mean ± standard deviation over at least three runs (or clarify if the posted values are single runs);
- peak VRAM for both paths; and
- output-quality/equivalence validation using identical inputs/seeds.

If the intended scope is a functional CPU-offload lifecycle fix rather than a performance optimization, please narrow the PR title and description to match that scope.

yeahdongcn:
> If the intended scope is a functional CPU-offload lifecycle fix rather than a performance optimization, please narrow the PR title and description to match that scope.

Done.

hsliuustc0106:
@yuanheng-zhao PTAL

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
```diff
@@ -1160,8 +1163,53 @@ def _uses_manual_component_offload(self) -> bool:
...
+            return
+        remove_sequential_offload(modules)
+        self._model_cpu_offload_modules = []
+
     @contextmanager
     def _component_on_device(self, component: nn.Module):
+        if getattr(self, "_model_cpu_offload_modules", None):
+            with sequential_offload_component(component):
+                yield
+            return
```
yuanheng-zhao, line 1207: With `sequential_offload_component` which triggers hook call outside a forward, I think the audio vae is not resident on GPU any more from the code:

https://github.com/vllm-project/vllm-omni/blob/4cd667875f983819414ca6a19cf612b58e6c138e/vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py#L1758-L1765

_encode_video_audio_conditions already has the `_component_on_device` and moved audio vae to CPU, and then _encode_audio_conditions actually make the audio vae encoder run on CPU, correct?

yuanheng-zhao (changes requested):

yuanheng-zhao:
Thanks for contributing @yeahdongcn . I'm a bit confused reading the following PR desc section:

> The control uses the same image and restores only the pre-PR per-image VAE residency scope. All other source, weights, dependencies, inputs, and serving arguments are identical.

I'm assuming the "per-image VAE residency scope" is the batching processing of all videos inside the context `_component_on_device`, which seems a partial trail results, rather than the major change / feature of your PR.

And the peak memory of control groups in the two tables (columns Omni peak and Peak memory, respectively) are basically the same. Could you update PR desc with control groups comparing the peak memory before and after this whole PR?

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
```diff
@@ -1160,8 +1163,53 @@ def _uses_manual_component_offload(self) -> bool:
             or getattr(od_config, "enable_distributed_layerwise_offload", False)
         )
 
+    def _model_cpu_offload_dits(self) -> list[nn.Module]:
+        return [module for module in (self.transformer, getattr(self, "transformers_ref", None)) if module is not None]
+
+    def _model_cpu_offload_stages(self) -> list[nn.Module]:
+        return [self.text_encoder, self.video_vae, self.audio_vae]
```
yuanheng-zhao: These two are redundant, and might drift from SupportsComponentDiscovery attributes

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
```diff
@@ -1160,8 +1163,53 @@ def _uses_manual_component_offload(self) -> bool:
...
+        device: torch.device,
+        pin_memory: bool,
+        use_hsdp: bool,
+    ) -> None:
+        if getattr(self, "_model_cpu_offload_modules", None):
+            return
+
+        dits = self._model_cpu_offload_dits()
+        stages = self._model_cpu_offload_stages()
+        modules = [*dits, *stages]
```
yuanheng-zhao: Can try
```suggestion
        modules = ModuleDiscovery.discover(self)
        dits = modules.dits
        stages = [*modules.encoders, *modules.vaes]
```

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
```diff
@@ -1195,9 +1228,20 @@ def _encode_visual_conditions(
...
-        for image in images:
-            rows.append(self._encode_visual_condition(image))
-            shapes.append((1, image.height // 16, image.width // 16))
+        if images:
+            _, rank, _ = _dit_rank_world()
+            image_rows = None
+            if rank == 0:
+                with self._component_on_device(self.video_vae):
+                    image_rows = torch.cat([self.video_vae.encode_image(image) for image in images])
+            rows.append(
```
yuanheng-zhao: Wonder what is the reason to add `_component_on_device` context here. This part moves video VAE to CPU, and the following handling of video (video_count branch) would require video VAE on GPU again. A whole D2H + H2D?

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
yeahdongcn: Removed.

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
yeahdongcn: Updated that way. Thanks!

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
yeahdongcn: Thanks for catching this. I updated the request path so embedded and standalone audio references share one Audio-VAE residency window.

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
yeahdongcn: The scope was originally added so direct VAE calls such as `encode_image()` outside `forward` correctly activate Video VAE under model-level offload, while allowing multiple images to share one residency window.

I widened the residency window to cover all image and video references in the request and call the resident video encoder inside that scope. The test now verifies a single activate/offload pair.

I also reran a current-head/candidate A/B on one S5000. For two images plus one video, current head uses two residency lifecycles while the candidate uses one. With a 32 MiB test VAE, H2D/D2H drop from 64/64 MiB to 32/32 MiB (I also updated the PR description). The single-image and image-only multi-reference paths remain at one lifecycle, so the existing consolidation does not regress.

yeahdongcn:
@yuanheng-zhao Could you take another look when you have a chance? Thanks!

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
```diff
@@ -1755,12 +1840,15 @@ def forward(self, request: DiffusionRequestBatch) -> DiffusionOutput:
                     prepared_videos,
                     video_count=video_count,
                 )
-                embedded_audio_condition, embedded_audio_lengths = self._encode_video_audio_conditions(
```
yuanheng-zhao, line 1758: This deletes the only caller of the method `_encode_video_audio_conditions` (I'm viewing with main branch fc163cd9)

yuanheng-zhao: If we want to deprecate and have ensured on callers of the private method, we should remove it, otherwise dead code left

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
```diff
@@ -1755,12 +1840,15 @@ def forward(self, request: DiffusionRequestBatch) -> DiffusionOutput:
...
+                (
+                    embedded_audio_condition,
+                    embedded_audio_lengths,
+                    external_audio_condition,
+                    external_audio_lengths,
+                ) = self._encode_reference_audio_conditions(
                     prepared_videos,
                     has_audio=has_audio,
-                )
-                external_audio_condition, external_audio_lengths = self._encode_audio_conditions(
```
yuanheng-zhao, line 1762: Same dead code concern at here. This is the only caller of `_encode_audio_conditions` and it has been deleted. Remove dead code in its impl and in tests `tests/diffusion/models/minimax_h3/test_minimax_h3_contract.py`

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
```diff
@@ -1195,16 +1223,31 @@ def _encode_visual_conditions(
     ) -> tuple[torch.Tensor | None, list[tuple[int, int, int]]]:
         rows: list[torch.Tensor] = []
         shapes: list[tuple[int, int, int]] = []
-        for image in images:
-            rows.append(self._encode_visual_condition(image))
-            shapes.append((1, image.height // 16, image.width // 16))
-        if video_count:
-            video_rows, video_shapes = self._encode_video_conditions(
```
yuanheng-zhao, line 1202: Same here. the only caller of `_encode_video_conditions` deleted - plz check its impl and tests `tests/diffusion/models/minimax_h3/test_minimax_h3_contract.py`. Let's remove dead code

yeahdongcn: Removed

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
yeahdongcn: Removed

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
yeahdongcn: Removed

yuanheng-zhao (approved):
Generally LGTM

PR #6073, opened 2026-08-12
fix: keep DLO hooks outside regional compilation

### LINKED ISSUE #5954: [Bug]: DLO rank-local (--usp 4 --dlo-no-use-allgather) segfaults DiffusionWorker-0 at denoise step 1 serving Cosmos3-Super on H200 at v0.26.0
### Environment

`vllm/vllm-omni:v0.26.0` @ `sha256:5cba1538c6f8ee81e8bea6708c24e68d7b2640f466a9fbf2ef15e68f2168b48b`, 8x NVIDIA H200 SXM (143,771 MiB/GPU), driver 580.126.09, CUDA 13.0 host image, x86_64. Model `nvidia/Cosmos3-Super`, guardrails off (see #5936).

### What happened

    vllm serve nvidia/Cosmos3-Super --omni --enable-distributed-layerwise-offload --usp 4 --dlo-no-use-allgather --no-guardrails --vae-use-tiling --vae-patch-parallel-size 1

Startup is clean: ready in ~230 s, 14,933-14,953 MiB per GPU across 4 GPUs, full weights per rank in host page cache per the rank-local design. A single t2v request (720p/189 frames/35 steps, seed 17) enters denoise and completes step 1 (10.97 s/it), then the worker crashes:

      3%|.         | 1/35 [00:10<06:13, 10.97s/it]!!!!!!! Segfault encountered !!!!!!!
      File "/usr/local/go/src/runtime/sys_linux_amd64.s", line 330, in runtime.sigfwd
    ERROR [multiproc_executor.py:355] Diffusion worker(s) died unexpectedly: ['DiffusionWorker-0(exitcode=None)']
    ERROR [inline_stage_diffusion_client.py:155] Diffusion request video_sync-... failed: Executor shut down

The engine does not recover; all subsequent requests fail with 500 until the server is restarted. Observed once in this run (the DP AllGather layouts fail differently, filed separately as #5953); we can rerun with debug env or flags on the same hardware if useful.

### Related

#5397 (DLO feature; test plan on Ascend 910B3), the `Cosmos3-DistOffload.md` rank-local section, measurement context in #5926.

hsliuustc0106:
## Summary

- Keep DLO/layerwise HookRegistry dispatch outside the regional `torch.compile` graph.
- Compile the original repeated-block forward instead of the offload dispatcher.
- Update hook references (for example, MagCache) to the compiled callable.
- Add a regression test covering the wrapper/compiled-forward interaction.

## Root cause

DLO installs a HookRegistry wrapper around each repeated block. The wrapper owns stream, event, storage, and layerwise offload coordination. Regional compilation previously compiled that wrapper, which pulled the offload control flow into Dynamo/Inductor and could cause a native crash on the first lazy compiled denoise call.

This change preserves the dispatcher as the eager `forward` path and compiles only the block's original compute callable.

## Validation

- `pytest -q tests/diffusion/test_compile.py tests/diffusion/hooks/test_hook_registry.py tests/diffusion/offloader/test_distributed_layerwise_backend.py tests/diffusion/model_loader/test_diffusers_loader.py tests/diffusion/test_diffusion_model_runner.py`
  - **90 passed**
- Synthetic 4-GPU NVIDIA L20X DLO regional-compile smoke test:
  - **20 iterations passed with finite outputs**
- LingBot Video on one NVIDIA L20X with rank-local DLO and regional compile:
  - `origin/main` reproduced a worker segfault on the first generation call
  - this patch generated a valid 9-frame H.264 MP4
- MiniMax-H3 FL2VA on 2 NVIDIA L20X GPUs with TP2, rank-local DLO, 20 resident layers, and eager execution:
  - generated valid H.264 video plus AAC audio

## Scope and limitations

- Direct 5090/H200 hardware reproduction was not available.
- The MiniMax consumer-profile run uses `enforce_eager`, so it validates the DLO runtime path but not regional compilation. The 5090 SIGBUS variant therefore remains hardware-specific and is not claimed as independently reproduced here.

Fixes #5954

### DIFF
--- a/vllm_omni/diffusion/compile.py
+++ b/vllm_omni/diffusion/compile.py
@@ -56,26 +56,50 @@ def regionally_compile(
     # Build all compiled callables before mutating the model. This keeps setup
     # failures atomic: callers can safely continue with the uncompiled model if
     # torch.compile raises synchronously for any repeated block.
-    compiled_forwards: list[tuple[nn.Module, Any]] = []
+    # Keep the pre-hook callable separately.  DLO and the ordinary layerwise
+    # offloader install a HookRegistry wrapper in ``module.forward``.  That
+    # wrapper performs stream/event synchronization and storage rebinding, so
+    # compiling it pulls offload control flow into the graph.  Compile the
+    # original block compute instead and leave the wrapper outside the graph.
+    compiled_forwards: list[tuple[nn.Module, Any | None, Any]] = []
     for name, submod in model.named_modules():
         if _matches_repeated_block(name, submod, repeated_blocks, repeated_block_attrs):
             # Compile the block compute while keeping nn.Module.__call__ hooks
-            # outside the compiled graph. NOTE: anything that wraps this
-            # callable must stay signature-transparent — cache-dit's
-            # BlockAdapter matches blocks by inspecting forward's parameter
-            # names and return annotation, which torch.compile preserves.
+            # outside the compiled graph. If a HookRegistry is already
+            # installed, ``submod.forward`` is the hook dispatcher and the
+            # original callable is stored in ``_omni_original_forward``.
+            # NOTE: anything that wraps this callable must stay
+            # signature-transparent — cache-dit's BlockAdapter matches blocks
+            # by inspecting forward's parameter names and return annotation,
+            # which torch.compile preserves.
+            original_forward = getattr(submod, "_omni_original_forward", None)
+            forward_to_compile = original_forward if original_forward is not None else submod.forward
             compiled_forwards.append(
                 (
                     submod,
-                    torch.compile(submod.forward, *compile_args, **compile_kwargs),
+                    original_forward,
+                    torch.compile(forward_to_compile, *compile_args, **compile_kwargs),
                 )
             )
 
     if not compiled_forwards:
         logger.warning(f"Regional compilation skipped because {repeated_blocks} classes are not found in the model.")
     else:
-        for submod, compiled_forward in compiled_forwards:
-            submod.forward = compiled_forward
+        for submod, original_forward, compiled_forward in compiled_forwards:
+            if original_forward is None:
+                submod.forward = compiled_forward
+                continue
+
+            # Keep HookRegistry's dispatcher as ``forward`` and replace only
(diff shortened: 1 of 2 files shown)


hsliuustc0106:
@lishun PTAL

vllm_omni/diffusion/compile.py
```diff
@@ -56,26 +56,50 @@ def regionally_compile(
...
+                submod.forward = compiled_forward
+                continue
+
+            # Keep HookRegistry's dispatcher as ``forward`` and replace only
+            # the callable it dispatches to.  Hooks such as MagCache retain a
+            # reference to the original callable as well, so update those
+            # references when they still point at the pre-compile function.
+            submod._omni_original_forward = compiled_forward
+            registry = getattr(submod, "_hook_registry", None)
+            for hook in getattr(registry, "_hooks", {}).values():
```
hsliuustc0106, line 99: Would it be cleaner to put this rebind behind a `HookRegistry` method (e.g. `rebind_original_forward(old, new)`)? As stringly `getattr`s on `_hook_registry` / `_hooks` / `fn_ref`, a rename in `hooks/base.py` would make this loop silently no-op, and pre-compile MagCache hooks would quietly fall back to the eager original with no signal. A typed method keeps the coupling in one place and fails loudly.

hsliuustc0106:
Same-pattern check on the other per-block compile installer: DreamZero's `setup_compile` (`vllm_omni/diffusion/models/dreamzero/pipeline_dreamzero.py:702`) still compiles `block.forward` directly, and it runs after `offload_backend.enable` in `load_model` (`worker/diffusion_model_runner.py:302` → `:307-314`). `CausalWanModel` declares `_layerwise_offload_blocks_attrs = ["blocks"]` (`models/dreamzero/causal_wan_model.py:746`), so DLO + compile on DreamZero would compile the `_WrappedForward` dispatcher — with `fullgraph=True`, no less — i.e. the same root cause as #5954 on a path not covered here.

Not a regression from this PR (pre-existing), but it might be cheap to close now: the resolve-`_omni_original_forward`-and-rebind logic added in `compile.py` is exactly what that loop needs, so extracting it into a shared helper both installers call would fix both. Alternatively a follow-up issue is fine.

PR #6076, opened 2026-08-12
[Misc] Consolidate LingBot text-to-video runner

hsliuustc0106:
## Purpose

Consolidate LingBot-Video offline inference into the shared `text_to_video.py` runner and remove the dedicated `text_to_video_lingbot.py` script.

- add LingBot model/class detection and preserve the dedicated example's defaults
- build the canonical video request envelope in the shared runner
- route LingBot-only options through `--extra-body`
- preserve already-normalized `[0, 1]` frame tensors during export
- scope distilled preset detection to the LTX family and support class-based preset detection
- update the LingBot recipe, user docs, and dense parity benchmark reference

## Test Plan

- Run focused CPU coverage for shared T2V prompts/presets/export normalization, LingBot request normalization, and model-extra routing.
- Run the shared runner on one NVIDIA L20X with the cached LingBot dense checkpoint.
- Compare the default shared-runner artifact against the retired dedicated runner using the same model, seed, and sampling settings.
- Exercise migrated LingBot `batch_cfg` and `output_type` options through `--extra-body`.
- Run a non-LingBot shared-runner smoke with the tiny Helios pipeline.

**vLLM Version:** 0.26.0

**vLLM-Omni Commit:** 0a593bddedcc38a7586ccc0c7a07e950681e28ec

## Test Result

- `127 passed, 14 warnings in 1.81s`
- pre-commit hooks passed on all changed files
- LingBot default: valid H.264, 320x192, 24 FPS, 9 frames; byte-identical MP4 and decoded frame hashes versus the dedicated runner
- LingBot model extras: `--extra-body '{"batch_cfg": true, "output_type": "np"}'` produced a valid H.264, 320x192, 24 FPS, 9-frame artifact
- Helios non-LingBot guardrail: valid H.264, 64x64, 16 FPS artifact using the tiny modular pipeline with SDPA


examples/offline_inference/text_to_video/text_to_video.md
```diff
@@ -8,6 +8,7 @@ A unified script for text-to-video generation. Supports multiple models with mod
 |---|---|---|---|---|---|
```
hsliuustc0106, line 8: do we need to change online serving as well?

hsliuustc0106: No. This PR only consolidates the offline examples. Online serving already uses the shared `OmniOpenAIServingVideo` `/v1/videos` path, which builds the canonical `modalities=["video"]` envelope and forwards the same sampling and model-specific parameters. There is no LingBot-specific online script to consolidate. I also verified the shared serving contract with the focused T2V prompt, sampling-parameter, and model-extra tests (`3 passed`).

benchmarks/lingbot_video/dense_pipeline_parity.py
```diff
@@ -167,7 +167,7 @@ def main() -> None:
     ]
     native_cmd = [
         sys.executable,
-        "examples/offline_inference/text_to_video/text_to_video_lingbot.py",
+        "examples/offline_inference/text_to_video/text_to_video.py",
```
wtz2333, line 170: changing the file to text_to_video.py, should pass --model-class-name LingBotVideoPipeline

examples/offline_inference/text_to_video/text_to_video.py
```diff
@@ -110,27 +121,49 @@
...
+    return result
+
+
+def _normalize_float_tensor(tensor: torch.Tensor) -> torch.Tensor:
+    """Normalize decoded frames without remapping tensors already in [0, 1]."""
+    if not tensor.is_floating_point():
+        return tensor
+    if float(tensor.min()) < 0.0:
+        return tensor.clamp(-1, 1) * 0.5 + 0.5
+    return tensor.clamp(0, 1)
```
wtz2333: Here is a review from AI:This infers the semantic range from the payload’s current minimum. List outputs are normalized frame by frame, so an all-positive [-1, 1] frame takes the clamp-only branch while an adjacent frame containing a negative value is remapped, causing brightness jumps. Please select the transform from the pipeline/output contract once per video and add an ambiguous-range test case.

PR #5994, opened 2026-08-10
[CI] Expand E2E source_file_dependencies for shared model code

yenuo26:
PLEASE FILL IN THE PR DESCRIPTION HERE.

## Purpose

Improve CUDA L2/L3 E2E path-based scheduling and CI cost:

1. **Expand `source_file_dependencies`** so shared implementation changes still schedule the right E2E jobs (prefix match only covers listed paths).
   - Qwen3-Omni / Qwen3-TTS: `common/qwen3_code_predictor.py`, `snake_activation.py`, `alias_free_activation.py` (TTS), plus Omni cross-deps (`qwen2_5_omni_thinker.py`, `qwen3_tts/cuda_graph_decoder_wrapper.py`).
   - MiniCPM-o: `minicpmo_4_5/`, `step_audio2/`, Stage-2 `cosyvoice3/code2wav_core/` + `cosyvoice3/utils.py`, stage processor, deploy YAMLs.
   - MiniCPM Duplex additionally: `engine/orchestrator.py`, `entrypoints/openai/api_server.py`, `duplex_capability.py`, `config/stage_config.py`.
   - Qwen-Image / Wan2.2: `diffusion/models/dmd2/` (and Qwen-Image AE path where applicable).

2. **Move heavy / lower-priority E2E jobs out of ready/merge** (Higgs-Audio-V3, VoxCPM2, Bagel, Qwen Image Edit/Layered, Z-Image, FLUX.2 Klein as applicable) and retarget the corresponding pytest markers from `core_model` / `advanced_model` to `slow`, so they no longer occupy L2/L3 queues by default.

Deploy YAML is listed only when the test actually uses `stage_config_path` / `get_deploy_config_path`; pure diffusion jobs that only pass `model=` intentionally omit deploy paths.

## Test Plan

**vLLM Version:** N/A (CI YAML + pytest markers)

**vLLM-Omni Commit:** `b7afce43`

- Diff review of `.buildkite/cuda/test-ready.yml` / `test-merge.yml` against each remaining E2E job's pytest targets and imported shared modules.
- Optional local dry-run:
```bash
python3 .buildkite/common/scripts/upload_pipeline.py .buildkite/cuda/test-ready.yml
python3 .buildkite/common/scripts/upload_pipeline.py .buildkite/cuda/test-merge.yml
```
- Marker-only changes in the listed e2e modules; no new functional test cases.

## Test Result

[...]

PR #6079, opened 2026-08-12
[CI][MiniCPM-o] Add MiniCPM-o 4.5 perf coverage to the ready gate

amy-why-3459:
## Purpose

MiniCPM-o 4.5 performance is currently only measured nightly, so a per-PR
performance regression is not visible until the next nightly run. This adds a
trimmed-down version of the nightly perf case to the ready gate on both CUDA and
NPU.

The new `tests/dfx/perf/tests/test_minicpmo_4_5_ready.json` mirrors the nightly
`test_minicpmo_4_5.json` (same deploy config, same seed-tts dataset, same request
body and percentile metrics) with two differences:

- Marked `core_model` instead of `full_model` so it runs in the ready gate.
- Concurrency sweep narrowed from `[1, 4, 8]` to `[1, 4]`, with `num_prompts`
  `[10, 40]`, to keep the step within the ready-gate time budget.

CI wiring:

- `.buildkite/cuda/test-ready.yml`: new `Omni · MiniCPM-o 4.5 Perf Test` step on
  `h100_1`, gated by `source_file_dependencies` on the perf config, the runner,
  and the MiniCPM-o deploy/model sources.
- `.buildkite/npu/test-npu-ready.yml`: matching step on `a3_npu_2`, using the
  same env and mark convention as the other MiniCPM-o steps in that group.

Both steps run the benchmark directly without `--assert-baseline`, so they gate
on request failures (`assert_result`) rather than on baseline deltas — same as
the existing nightly perf jobs.

## Test Plan

**vLLM Version:** 0.26.0
**vLLM-Omni Commit:** (fill after push)

- Config/pipeline validation:
  - `python -c "import json,yaml; json.load(open('tests/dfx/perf/tests/test_minicpmo_4_5_ready.json'))"`
  - YAML parse of both modified pipeline files
  - `pytest tests/dfx/perf/scripts/run_benchmark.py --test-config-file tests/dfx/perf/tests/test_minicpmo_4_5_ready.json -m 'core_model and cuda' --collect-only`
  - `pytest tests/dfx/perf/scripts/run_benchmark.py --test-config-file tests/dfx/perf/tests/test_minicpmo_4_5_ready.json -m 'core_model and npu and A3 and omni' --collect-only`
- [ ] Full run on H100 (1x)
- [ ] Full run on A3 (1x)

## Test Result

- JSON and both pipeline YAMLs parse cleanly.
- Mark selection resolves to exactly 1 case under both the CUDA and the NPU
  mark expression.
- (fill in wall-clock time and measured metrics from the first H100 / A3 runs)

### DIFF
--- a/.buildkite/cuda/test-ready.yml
+++ b/.buildkite/cuda/test-ready.yml
@@ -227,6 +227,18 @@ steps:
             "
         mirror_hardwares: h100_1
 
+      - label: "Omni · MiniCPM-o 4.5 Perf Test"
+        timeout_in_minutes: 60
+        source_file_dependencies:
+          - tests/dfx/perf/tests/test_minicpmo_4_5_ready.json
+          - vllm_omni/deploy/minicpmo_4_5.yaml
+          - vllm_omni/model_executor/models/minicpmo_4_5/
+          - vllm_omni/model_executor/stage_input_processors/minicpmo_4_5_omni.py
+        commands:
+          - export BENCHMARK_DIR=tests/dfx/perf/results
+          - pytest -s -v tests/dfx/perf/scripts/run_benchmark.py --test-config-file tests/dfx/perf/tests/test_minicpmo_4_5_ready.json -m 'core_model and cuda'
+        mirror_hardwares: h100_1
+
       - label: "TTS · VoxCPM2 Test"
         source_file_dependencies:
           - tests/e2e/online_serving/test_voxcpm2_tts.py
--- a/.buildkite/npu/test-npu-ready.yml
+++ b/.buildkite/npu/test-npu-ready.yml
@@ -17,6 +17,15 @@ steps:
           VLLM_ALLOW_LONG_MAX_MODEL_LEN: "1"
         commands:
           - pytest -s -v tests/e2e/online_serving/test_minicpmo_4_5_duplex.py -m 'core_model and npu and A3 and omni' --run-level 'core_model'
+      - label: "Omni · MiniCPM-o 4.5 Perf Test"
+        timeout_in_minutes: 60
+        mirror_hardwares: a3_npu_2
+        env:
+          VLLM_WORKER_MULTIPROC_METHOD: spawn
+          VLLM_ALLOW_LONG_MAX_MODEL_LEN: "1"
+        commands:
+          - export BENCHMARK_DIR=tests/dfx/perf/results
+          - pytest -s -v tests/dfx/perf/scripts/run_benchmark.py --test-config-file tests/dfx/perf/tests/test_minicpmo_4_5_ready.json -m 'core_model and npu and A3 and omni'
 
   - group: ":card_index_dividers: TTS Model Test"
     key: ready-tts-group
--- a/tests/dfx/perf/tests/test_minicpmo_4_5_ready.json
+++ b/tests/dfx/perf/tests/test_minicpmo_4_5_ready.json
@@ -1,0 +1,80 @@
+[
+  {
+    "test_name": "test_minicpmo_4_5_ready",
+    "mark": [
+      {
+        "hardware_marks": {
+          "res": {
+            "cuda": "H100",
+            "npu": "A3"
+          },
+          "num_cards": 1
+        }
+      },
+      "core_model",
+      "omni"
+    ],
+    "server_params": {
+      "model": "openbmb/MiniCPM-o-4_5",
+      "extra_cli_args": [
+        "--deploy-config",
+        "vllm_omni/deploy/minicpmo_4_5.yaml",
+        "--trust-remote-code"
+      ]
+    },
+    "benchmark_params": [
+      {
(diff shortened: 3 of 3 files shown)


amy-why-3459:
@yenuo26 @R2-Y PTAL

.buildkite/cuda/test-ready.yml
```diff
@@ -227,6 +227,20 @@ steps:
             "
         mirror_hardwares: h100_1
 
+      - label: "Omni · MiniCPM-o 4.5 Perf Test"
+        timeout_in_minutes: 60
+        source_file_dependencies:
+          - tests/dfx/perf/tests/test_minicpmo_4_5_ready.json
+          - tests/dfx/perf/scripts/run_benchmark.py
```
yenuo26: Actually, I'd suggest that changes to these test utilities shouldn't trigger this job. Even though they might cause this job to fail, nightly would already catch that. My understanding is that this job is mainly meant to guard against regressions in MiniCPM changes.

.buildkite/cuda/test-ready.yml
```diff
@@ -227,6 +227,20 @@ steps:
...
+          - tests/dfx/conftest.py
+          - vllm_omni/deploy/minicpmo_4_5.yaml
+          - vllm_omni/model_executor/models/minicpmo_4_5/
+          - vllm_omni/model_executor/stage_input_processors/minicpmo_4_5_omni.py
+        commands:
```
yenuo26, line 237: Also, could you evaluate whether we still need to keep the original functional test cases in the 'ready' pipeline, since we already have these performance test cases? At the 'ready' level, those functional tests are essentially just simple validations like checking if the request call succeeds.

.buildkite/cuda/test-ready.yml
```diff
@@ -227,6 +227,20 @@ steps:
             "
         mirror_hardwares: h100_1
 
+      - label: "Omni · MiniCPM-o 4.5 Perf Test"
+        timeout_in_minutes: 60
+        source_file_dependencies:
```
yenuo26, line 232: Why does this performance test trigger much less frequently compared to the functional test? Was that intentionally designed?

.buildkite/cuda/test-ready.yml
amy-why-3459: Removed. run_benchmark.py and tests/dfx/conftest.py are shared perf infrastructure exercised by every perf config in nightly, so gating a MiniCPM-specific job on them only adds noise. The trigger set is now MiniCPM-owned only: the perf JSON, minicpmo_4_5.yaml, model_executor/models/minicpmo_4_5/, and stage_input_processors/minicpmo_4_5_omni.py.

.buildkite/cuda/test-ready.yml
amy-why-3459: > Also, could you evaluate whether we still need to keep the original functional test cases in the 'ready' pipeline, since we already have these performance test cases? At the 'ready' level, those functional tests are essentially just simple validations like checking if the request call succeeds.

The perf case isn't a superset of the functional ones. At -m core_model the online functional job runs test_text_to_text_001 (stream=False, asserts "Beijing" in the text) and test_mix_to_text_audio_001 (text+audio+video+image input), and the offline job covers the offline inference path. The perf case only sends seed-tts text prompts through /v1/chat/completions and asserts result["completed"] == num_prompts — no multimodal ingestion, no non-stream path, no offline path, and no output-content assertion. So dropping the functional jobs would lose the image/video/audio input paths and the only correctness signal we have at ready level. If the goal is to cut ready cost, trimming the functional set looks like the better lever than removing it.

.buildkite/cuda/test-ready.yml
amy-why-3459: > Why does this performance test trigger much less frequently compared to the functional test? Was that intentionally designed?

After dropping the two test-infra entries, the perf job triggers on a superset of the online functional job, which lists only test_minicpmo_4_5.py + minicpmo_4_5.yaml and notably not model_executor/models/minicpmo_4_5/. It is narrower than the duplex job, which additionally guards shared paths (core/sched/, engine/async_omni_engine.py, engine/stage_init_utils.py, entrypoints/cli/serve.py, outputs/multimodal_accumulation.py). That part was intentional: this is a 60-minute H100 job, and I'd rather not run it on every scheduler/engine touch when nightly already covers the cross-cutting surface. Happy to add those shared paths if you'd prefer perf to gate them too.

Separately, the online functional job not listing the MiniCPM model directory looks like a pre-existing gap — a change to the model code doesn't currently retrigger it. I can align that here or in a follow-up, whichever you prefer.

yenuo26: Regarding the trigger condition changes, I've already updated them in PR #5994 . If there are still any omissions, please feel free to point them out there.

yenuo26 (approved):
LGTM

PR #6082, opened 2026-08-12
[Perf][MiniCPM-O-4.5]Add CUDA Graph For CFM DiT estimator

stringl1l1l1l:
## Purpose

Add CUDA Graph acceleration for the MiniCPM-o 4.5 CFM DiT estimator. During the 10-step Euler ODE solver loop, each step calls `estimator.blocks_forward_chunk`, which contains the full computation: `transpose → in_proj → 16 DiT blocks → final_layer → transpose`. By capturing `blocks_forward_chunk` directly as the graph target with per-shape capture and replay, we eliminate Python dispatch and kernel launch overhead.

This PR follows the design pattern of [#5869](https://github.com/vllm-project/vllm-omni/pull/5869) (HiFTGenerator CUDA Graph), reusing the same wrapper structure. The upstream cosyvoice2 DiT's `_init_cuda_graph_chunk` also captures `blocks_forward_chunk` directly, validating this approach.

### Graph Target Selection

The single-step computation of the CFM DiT estimator consists of two parts:

1. **Input preparation** (in `batched_token2wav.py:_estimator_step`): `torch.cat` to concatenate `x/mu/speaker_features/cond`, `expand` to broadcast speaker embedding, allocate `cnn_out`/`att_out` buffers
2. **Core computation** (`estimator.blocks_forward_chunk`): `transpose → in_proj → 16×DiT block → final_layer → transpose`, with in-place writes to cnn/att cache buffers

This PR directly references `estimator.blocks_forward_chunk` as `graph_fn`. The `t_embedder` call and input preparation in `_estimator_step` execute outside the graph; their outputs are passed as inputs to `wrapper.replay()`. The graph captures only the pure tensor computation and cache writes of `blocks_forward_chunk`.

### Hyperparameters

```yaml
# vllm_omni/deploy/minicpmo_4_5.yaml
code2wav:
  enable_cfm_graph: true
  cfm_max_graphs: 32
```

- `enable_cfm_graph`: Whether to enable CFM CUDA Graph (default `false`, set to `true` in this PR)
- `cfm_max_graphs`: Maximum number of per-shape graph cache entries (default `32`)

#### max_graphs Sweep

Test conditions: H100 PCIe, 2×GPU, seed-tts-eval EN, 1088 requests, concurrency=4

| max_graphs | WER | duration (s) | throughput (req/s) | E2E latency (ms) | TTFP (ms) | RTF | peak GPU0 (GiB) |
|---|---|---|---|---|---|---|---|
| 4 | 3.13% | 1164 | 0.934 | 4274 | 1679 | 0.951 | 48.2 |
| 8 | 3.04% | 1099 | 0.990 | 4034 | 1636 | 0.901 | 46.5 |
| 16 | 3.10% | 1004 | 1.084 | 3686 | 1486 | 0.818 | 46.9 |
| **32** | **3.10%** | **976** | **1.115** | **3580** | **1441** | **0.791** | **47.1** |
| 48 | 3.12% | 979 | 1.112 | 3592 | 1441 | 0.792 | 50.4 |
| 64 | 3.15% | 974 | 1.117 | 3574 | 1430 | 0.789 | 52.3 |

**Conclusion**: `max_graphs=32` is the optimal cost-effectiveness point. Beyond 32, performance saturates (throughput difference <0.2%) while GPU memory continues to grow (47.1→52.3 GiB). `max_graphs=16` covers most common chunk sizes; 32 provides better coverage; 64 yields negligible improvement.

### Environment

- **GPU**: 2× NVIDIA H100 PCIe (80 GiB)
- **PyTorch**: 2.11.0+cu130
- **CUDA**: 13.0
- **vLLM**: 0.26.0
- **Model**: MiniCPM-o-4.5
- **Dataset**: seed-tts-eval EN (`zhaochenyang20/seed-tts-eval`)
- **Prompts**: 1088
- **HiFT Graph**: enabled

### Baseline vs CFM Graph

**Concurrency=4** (low concurrency, latency-oriented):
Baseline: CFM graph disabled

| Metric | Baseline | CFM Graph | Improvement |
|------|-----------------|-----------------|------|
| WER | 3.30% | 3.38% | — |
| E2E latency (ms) | 4335 | 3661 | −15.5% |
| Audio TTFP (ms) | 1885 | 1450 | −23.1% |
| RTF | 0.971 | 0.812 | −16.4% |
| Throughput (req/s) | 0.922 | 1.091 | +18.3% |
| Duration (s) | 1180 | 997 | −15.5% |

**Concurrency=10** (high concurrency, throughput-oriented):

| Metric | Baseline | CFM Graph | Improvement |
|------|-----------------|-----------------|------|
| WER | 3.26% | 3.29% | — |
| E2E latency (ms) | 8305 | 6315 | −24.0% |
| Audio TTFP (ms) | 3859 | 3477 | −9.9% |
| RTF | 1.867 | 1.430 | −23.4% |
| Throughput (req/s) | 1.202 | 1.581 | +31.5% |
| Duration (s) | 905 | 688 | −24.0% |

Under both concurrency levels, WER remains in the 3.2%–3.4% range, well below the 5% threshold, with no accuracy regression. The higher the concurrency, the more significant the throughput improvement from CFM Graph (+18.3% → +31.5%), because the elimination of kernel launch overhead is more pronounced when the GPU is continuously saturated.

### 1. Run unit tests

```bash
python -m pytest -v \
    tests/model_executor/models/minicpmo_4_5/test_cuda_graph_wrapper.py \
    tests/model_executor/models/minicpmo_4_5/test_pipeline.py
```

### 2. Run performance benchmark

Start the server:

```bash
vllm serve <model_path> --omni --trust-remote-code \
    --host 0.0.0.0 --port 8091
```

Run seed-tts-eval WER + latency benchmark:

```bash
vllm bench serve --omni --host 127.0.0.1 --port 8091 \
    --model <model_path> --endpoint /v1/chat/completions \
    --backend openai-chat-omni --request-rate inf \
    --num-prompts 2000 --max-concurrency 10 --no-oversample \
    --num-warmups 0 --percentile-metrics ttft,tpot,itl,e2el,audio_ttfp,audio_rtf \
    --save-result --result-dir /tmp/cfm_bench --trust-remote-code \
    --dataset-name seed-tts --dataset-path zhaochenyang20/seed-tts-eval \
    --seed-tts-locale en --seed-tts-wer-eval \
    --extra-body '{"modalities":["text","audio"],"chat_template_kwargs":{"enable_thinking":false,"use_tts_template":true}}'
```

## Profile
graph disable:
image

graph enable：
image

`_decode_cfm`:  150 ms →  63 ms
every `_estimator_step`: 15 ms →  6 ms

## Alternative Approach: Custom `_estimator_step_graph_fn`

Before adopting direct `blocks_forward_chunk` capture, we explored an alternative: defining a custom `_estimator_step_graph_fn` in `CFMGraphWrapper` as the graph target. This approach requires decomposing and rewriting the internal logic of `blocks_forward_chunk`:

```python
def _estimator_step_graph_fn(
    self,
    x: torch.Tensor,
    mu: torch.Tensor,
    time_embedding: torch.Tensor,
    speakers: torch.Tensor,
    cond: torch.Tensor,
    cnn_cache: torch.Tensor,
    att_cache: torch.Tensor,
    cnn_out: torch.Tensor,
    att_out: torch.Tensor,
) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
    width = int(x.shape[-1])
    time_emb = time_embedding.unsqueeze(1)
    speaker_features = speakers.unsqueeze(-1).expand(-1, -1, width)
    estimator_input = torch.cat((x, mu, speaker_features, cond), dim=1)

    h = estimator_input.transpose(1, 2)
    h = self.estimator.in_proj(h)
    for b_idx in range(len(self.estimator.blocks)):
        h, cnn_out[b_idx], att_out[b_idx] = self.estimator.blocks[b_idx].forward_chunk(
            h, time_emb, cnn_cache[b_idx], att_cache[b_idx], None
        )
    h = self.estimator.final_layer(h, time_emb)
    h = h.transpose(1, 2)
    return h, cnn_out, att_out
```

`blocks_forward_chunk` takes the preprocessed `estimator_input` (with `x/mu/speaker_features/cond` already concatenated), whereas `_estimator_step_graph_fn` incorporates input preparation (`cat`, `expand`, `unsqueeze`) into the graph, covering more operators. Additionally, `_estimator_step_graph_fn` directly returns a `(h, cnn_out, att_out)` triple, while `blocks_forward_chunk` writes `cnn_out`/`att_out` in-place into passed buffers (not via return values), requiring extra cloning of static buffers after graph replay.

### Performance Comparison

The following data was measured under concurrency=4:

| Metric | `blocks_forward_chunk` (this PR) | `_estimator_step_graph_fn` (reference) | Difference |
|------|----------------------------------|--------------------------------------|------|
| WER | 3.38% | 3.42% | — |
| E2E latency (ms) | 3661 | 3520 | −3.9% |
| Audio TTFP (ms) | 1450 | 1386 | −4.4% |
| RTF | 0.812 | 0.778 | −4.2% |
| Throughput (req/s) | 1.091 | 1.135 | +4.0% |
| Duration (s) | 997 | 959 | −3.8% |

`_estimator_step_graph_fn` is ~4% faster than `blocks_forward_chunk`, mainly because it avoids the `transpose` operations and in-place buffer writes inside the graph, and returns the triple directly without extra cloning.

### Why Not Adopted

`_estimator_step_graph_fn` is essentially a decomposed rewrite of `blocks_forward_chunk`'s internal logic (`in_proj → for block in blocks → final_layer`). When the upstream `blocks_forward_chunk` implementation changes (e.g., block invocation, cache write logic, new preprocessing steps), `_estimator_step_graph_fn` must be synchronously maintained, introducing drift risk. In contrast, directly referencing `estimator.blocks_forward_chunk` as `graph_fn` requires no rewritten compute logic and stays consistent with the upstream cosyvoice2 DiT's `_init_cuda_graph_chunk`. The 4% performance gap does not justify the maintenance cost.

**vLLM Version:** 0.26.0


amy-why-3459:
Thanks for the focused perf work — capturing `blocks_forward_chunk` directly
(instead of a hand-rewritten `_estimator_step_graph_fn`) is the right call, and
the maintenance-cost argument against the 4%-faster alternative is convincing.

Request changes on four points:

1. Please drop the unrelated `sys.modules["cosyvoice2"] = stepaudio2.cosyvoice2`
   injection in `_build_backend` (or split it into its own PR). It adds a hard
   `stepaudio2` import and rewrites module resolution process-wide, and it is
   not mentioned in the PR description.

2. The graph path replaces `cnn_cache=None` with zeros and `att_cache=None` with
   a zero-length tensor. `setup_batch` hits this on every request's prompt pass,
   so it can change first-chunk numerics if upstream `blocks_forward_chunk`
   branches on `cache is None`. The new tests cannot catch it: `_MiniDiT`
   never reads `cnn_cache`/`att_cache`, so the cache assertions are vacuous.
   Please add a parity test through `_estimator_step`.

3. The cache never evicts (`move_to_end` without `popitem`), so "LRU" in the
   commit message is inaccurate, and once 32 shapes are pinned everything else
   falls back forever. That fallback also clones all 6 inputs before running
   eager, making it slower than baseline. Please (a) call eager on `inputs`
   directly, (b) bucket the att length (like HiFT's `derive_capture_bucket_size`)
   or implement real eviction, and (c) report graph hit rate at c=4 / c=10 —
   without it the `max_graphs=32` conclusion cannot be evaluated.

stringl1l1l1l:
### 1. `sys.modules["cosyvoice2"]` injection — removed

**Root cause**: `flow.yaml` uses `hyperpyyaml` syntax `!new:cosyvoice2.flow.flow.CausalMaskedDiffWithXvec`, which calls `importlib.import_module("cosyvoice2.flow.flow")`. But the installed pip package is `stepaudio2`,`cosyvoice2` is only a sub-package. The import fails with `ModuleNotFoundError: No module named 'cosyvoice2'`. This is a pre-existing issue in the base branch, unrelated to CFM CUDA Graph.

**Solution (no code change)**: Create a `.pth` file in `site-packages`:

```bash
echo "/path/to/site-packages/stepaudio2" > site-packages/cosyvoice2_alias.pth
```

Or a symlink:

```bash
ln -s site-packages/stepaudio2/cosyvoice2 site-packages/cosyvoice2
```

Both make `cosyvoice2` importable as a top-level package. The ideal fix is in `stepaudio2`'s packaging (register `cosyvoice2` as a top-level package in `setup.py`). 

---

### 2. `cnn_cache=None` → zeros changes first-chunk numerics — addressed

**Issue**: The graph path replaces `cnn_cache=None` with `torch.zeros_like(cnn_out)` and `att_cache=None` with a zero-length tensor. `setup_batch` hits this on every request's prompt pass. The original tests could not catch regressions because `_MiniDiT` never read `cnn_cache`/`att_cache`.

**Current behavior**: Upstream `CausalConv1d.forward_chunk` creates zeros internally when `cnn_cache is None` — identical to passing zeros. `Attention.forward_chunk` skips `torch.cat` when `att_cache is None` — identical to passing a zero-length tensor. Both paths produce the same output, confirmed by e2e WER (3.26% eager vs 3.29% graph, no regression).

**Fix**:

1. **Modified `_MiniDiT`**: The original implementation only wrote to buffers and never read the incoming `cnn_cache`/`att_cache`, making cache assertions vacuous. Modified to actually read cache, simulating upstream `CausalConv1d` (causal padding with cache) and `Attention` (concatenating historical KV with cache) semantics.

2. **Added `test_cfm_none_cache_parity_between_graph_and_eager`**: Constructs `cnn_cache=None` scenario (simulating `setup_batch` prompt pass), runs through both eager path (`[None]*depth`) and graph path (`torch.zeros_like`), and asserts output/cnn_out/att_out are identical. If upstream changes to branch differently on `None`, this test will fail.

---

### 3. LRU eviction, fallback performance, and hit rate — fixed

**(a) No real eviction — fixed**: Replaced manual `OrderedDict` + `move_to_end`  with `functools.lru_cache(maxsize=max_graphs)`.

**(b) Fallback cloned all 6 inputs — fixed**: The eager fallback path now calls `_call_graph_fn(inputs)` directly on the original inputs, without cloning.

**(c) Hit rate measurement**: Logged `lru_cache.cache_info()` on every capture:

```
concurrency=10, 1088 requests, max_graphs=32
hits=53,557, misses=144, total=53,701
hit rate=99.73%
```

144 misses = 32 initial captures + 112 LRU evictions and re-captures.

The cache key is composed of `(shape, dtype, device)` of all 6 input tensors. Four dimensions determine shape variation: `batch_size`, `chunk_size`, `old_att_len`, and `new_att_len` (= `old_att_len + chunk_size`). In practice, `old_att_len` has only 4 values (0/302/352/402, corresponding to prompt pass and chunk 1/2/3), but `chunk_size` has 26 distinct values (6–54, step 2) and `batch_size` has 4 values (2/4/6/8). This yields ~150 unique shapes total, far exceeding the 32-slot capacity.

Evictions are concentrated on `batch=2, old_att_len=402` with its 26 chunk_size variants — the largest group, frequently accessed but too many for 32 slots. When an evicted shape reappears, it triggers re-capture. Despite this, hit rate still remains **99.73%**.

**Bucket padding — explored, not adopted**:

**Design**: `old_att_len` grows with chunk accumulation (0 → 302 → 352 → 402 → ...), each value producing a distinct graph shape. The bucket approach pads `old_att_len` up to the nearest multiple of 64 (e.g., 352→384, 402→448), collapsing different `old_att_len` values into the same bucket to reduce unique shape count.

Implementation:
1. `_bucket_pad`: pad att_cache along dim=3 to `ceil(old_att_len / 64) * 64` with zeros
2. Generate attention mask: `mask[:, :, valid_len:] = False` to block padding positions in SDPA
3. Capture graph with padded att_cache (shape fixed at bucket boundary), truncate back to original length after replay

**e2e results** (concurrency=4, 1088 requests):

| Approach | WER | RTF | Peak GPU |
|---|---|---|---|
| No bucket, max32 (this PR) | 3.10% | 0.791 | 47.1G |
| Bucket64, max32 | 3.04% | 0.820 | 50.5G |
| Bucket64, max8 | 3.16% | 0.972 | 47.3G |

**Why bucket padding is slower**:
1. **Increased attention computation**: Padded att_cache is longer (e.g., 402→448), so `torch.cat([k, k_cache])` and `scaled_dot_product_attention` process longer sequences. This overhead applies to every block and every Euler step — 10 steps × 16 blocks = 160 attention calls — outweighing the benefit of fewer cache misses.
2. **Increased memory**: Padded buffers are larger, +3.4G at max32 (47.1→50.5G), OOM at max16.
3. **Implementation complexity and bug risk**: mask must correctly block padding (initial `mask=None` caused crash).

**Conclusion**: The LRU approach with 99.73% hit rate is sufficient. 112 evictions across 53,701 calls (0.27%) cost only one re-capture each (3 warmups + 1 graph capture), far less than the per-replay attention overhead of bucket padding. No bucket + LRU is simpler, faster, and more memory-efficient.

amy-why-3459:
@stringl1l1l1l CI is currently red on this PR — please fix the failing checks
(and push a fix, not only re-run) so we can continue review. Thanks!

stringl1l1l1l:
> @stringl1l1l1l CI is currently red on this PR — please fix the failing checks (and push a fix, not only re-run) so we can continue review. Thanks!

@amy-why-3459  Sorry, I accidentally force-pushed with new changes, which reset the review status back to pending and prevented me from re-running the CI checks. I have recovered it. Could you please re-approve / trigger the workflow? Thanks!

amy-why-3459 (approved):

amy-why-3459:
Thanks, LGTM.

PR #6084, opened 2026-08-12
[Feature]: Support pause / resume and sleep / wake for AR stages in AsyncOmni

### LINKED ISSUE #6083: [Feature]: Support pause / resume and sleep / wake for AR stages in AsyncOmni

AsyncOmni should support **pause / resume** and **sleep / wake** for **AR (autoregressive / LLM) stages** so serving can safely stop scheduling, free GPU memory (weights / KV), and restore generation afterward — with the same operational meaning as vLLM’s `AsyncLLM` APIs.

Today this works more reliably for diffusion worker paths than for AR stages. AR serving needs first-class generation control for RLHF / weight sync, multi-model colocated serving, and idle memory reclaim.

When serving multimodal / multi-stage models (e.g. Qwen3-Omni thinker → talker → code2wav):

1. **Pause** is needed to stop accepting / progressing AR work without tearing down the engine (e.g. before weight updates or maintenance).
2. **Sleep** is needed to **release GPU memory** (weights and/or KV) while the process stays up.
3. **Wake + resume** must restore a state where **new generation succeeds** again.

Without correct AR behavior, operators see misleading “paused/slept” state while AR stages continue scheduling or while memory is not actually reclaimable / restorable.

### Alternatives

_No response_

### Current behavior

- Frontend-level pause can block new admissions, but AR stages may still keep scheduling / progressing work.
- Sleep / wake for AR does not consistently match vLLM `AsyncLLM` semantics (scheduler pause + memory offload/restore + resume).
- Diffusion stages already have a worker sleep/wake path; AR stages need equivalent **generation-control** guarantees at the AR runtime boundary.
- After sleep, generate may fail or be unsafe if wake/resume is incomplete or tags are only partially restored.

### Desired behavior

For AR stages exposed through `AsyncOmni`:

| API | Expected behavior |
| --- | --- |
| `pause_generation(...)` | Stop AR generation progress according to pause mode (`abort` / `wait` / `keep`); optional cache clear; no new AR work progresses while paused. |
| `resume_generation()` | Restore AR generation so new requests can run. |
| `sleep(level=1, ...)` | Pause AR work as needed, then free/offload GPU resources (wake-able). |
| `wake_up(...)` | Restore previously slept resources. |
| `generate(...)` after full wake + resume | Succeeds like a cold/warm engine that is awake. |

Diffusion stages should keep working as they do today (no regression).

Multi-stage pipelines (mixed AR + diffusion) should allow controlling AR stages without breaking the overall Omni lifecycle.

Related: https://github.com/vllm-project/vllm-omni/issues/5800

knlnguyen1802:
## Purpose

Fix: https://github.com/vllm-project/vllm-omni/issues/6083

AsyncOmni should support **pause / resume** and **sleep / wake** for **AR stages** with the same operational meaning as vLLM `AsyncLLM`, so multimodal serving can stop AR progress, reclaim GPU memory, and restore generation for RLHF / weight sync, colocated serving, and idle reclaim.

**Before this change:**

- `pause_generation` mostly flipped a frontend flag; AR EngineCore could keep scheduling.
- AR `sleep` used worker sleep while EngineCore kept running.
- Pipelined `generate()` could race into stages during sleep drain/offload.
- Abort cleared frontend request state before orchestrator cleanup finished.

**After this change:**

1. **AR EngineCore control** — `pause_generation` / `resume_generation` / AR `sleep` / `wake_up` go through existing Orchestrator `collective_rpc`. For AR stages, `StagePool.collective_rpc` prefers vLLM `AsyncMPClient` helpers (`{method}_async`). Diffusion stages keep worker `handle_sleep_task` / `handle_wake_task` (no regression).
2. **Sleep admission gate** — `sleep()` sets frontend `_paused = True` before any sleep RPC so `generate()` waits on `_pause_cond` during drain/offload. Does not double-call `pause_scheduler`. `wake_up()` does not auto-`resume_generation`; callers must resume explicitly.
3. **Acknowledged abort** — optional `rpc_id` on `AbortRequestMessage` + `AbortResultMessage`. `AsyncOmniEngine.abort_async()` waits for orchestrator cleanup; sync `abort()` stays fire-and-forget. `AsyncOmni._abort` awaits ack before popping frontend `request_states`.

Typical trainer / serving order:

```text
pause → abort → sleep(level=1) → train/idle → wake → resume → generate
```

## Test Plan

**vLLM Version:** (fill before submit)

**vLLM-Omni Commit:** (fill after push)

```bash
pytest tests/entrypoints/test_async_omni_pause_sleep_routing.py -q
pytest tests/engine/test_async_omni_engine_abort_ack.py -q
pytest tests/engine/test_orchestrator.py -q -k abort
pytest tests/engine/test_rpc_result_router.py -q -k abort
pytest tests/entrypoints/test_async_omni.py -q -k abort
```
**vLLM Version:**

**vLLM-Omni Commit:**

## Test Result


hsliuustc0106 (commented):
add the docs first

zhtmike:
ping @hbhflw2000 @AndyZhou952 for review, since it is quite important for async training. 

cc @SamitHuang

vllm_omni/entrypoints/async_omni.py
```diff
@@ -1179,41 +1257,96 @@ async def reset_prefix_cache(
...
 
     async def wake_up(self, stage_ids: list[int] | None = None, tags: list[str] | None = None) -> list[OmniACK]:
+        """Wake stages after sleep.
+
+        AR/LLM stages use EngineCore.wake_up (restore memory, auto-resume
+        scheduler). Diffusion stages keep the worker-level wake RPC.
+
+        Does **not** clear the frontend ``_paused`` admission gate set by
+        :meth:`sleep` / :meth:`pause_generation`. Call :meth:`resume_generation`
+        when the trainer is ready to admit new requests.
```
SamitHuang: In sleep(), both WEIGHTS and KV_CACHE are added to _sleeping_tags even when level=1. When wake_up() is subsequently called without explicit tags, lines 1395-1398 default to waking only [CuMemTag.KV_CACHE.value] because _level2_sleeping is False. This leaves WEIGHTS in _sleeping_tags, causing subsequent generate() calls to fail with a RuntimeError ("Engine is partially or fully asleep...").

vllm_omni/entrypoints/async_omni.py
```diff
@@ -1090,38 +1119,87 @@ async def _abort(self, request_ids: list[str]) -> None:
...
+           Orchestrator loop (abort/wait/keep + optional cache clear).
+        3. Diffusion stages have no EngineCore scheduler — only frontend
+           admission is paused for them.
+
+        Note: ``sleep()`` already pauses the AR scheduler internally (same as
+        vLLM EngineCore.sleep). Call this API when you need pause *without*
+        freeing GPU memory (e.g. weight sync).
+        """
+        if wait_for_inflight_requests:
+            mode = "wait"
```
SamitHuang, line 1170: If self._paused was set prior to pause_generation() (for instance during sleep() or an earlier pause_generation call), pause_generation() returns immediately on line 1157. As a result, _engine_core_rpc("pause_scheduler", ...) and cache clearing (reset_prefix_cache, reset_mm_cache, reset_encoder_cache) are skipped even if the caller requests a cache clear or targets a different stage_ids subset. The method should execute pause_scheduler and clear caches regardless of whether frontend admission was already marked paused.

SamitHuang (commented):
This PR adds support for `pause_generation`, `resume_generation`, `sleep`, and `wake_up` for Autoregressive (AR) stages via `EngineCore` control paths, alongside an acknowledged abort mechanism (`abort_async`).

#### Key Findings & Action Items
1. **Sleep / Wake Tag Tracking Mismatch**: In `sleep()`, both `WEIGHTS` and `KV_CACHE` tags are recorded when `level=1`. However, `wake_up()` defaults to waking only `KV_CACHE` when `tags` is not provided and `_level2_sleeping` is `False`. This leaves `WEIGHTS` in `_sleeping_tags`, blocking subsequent generation requests.
2. **Early Return in `pause_generation()`**: `pause_generation()` returns immediately if `_paused` is `True`. If `sleep()` ran prior or if a second call targets a different `stage_ids` subset or requests `clear_cache=True`, the `pause_scheduler` RPC and cache resetting are skipped.
3. **`StagePool.collective_rpc` Dispatch Edge Cases**: When `client_method` is invoked directly, `timeout` is dropped, and passing `args=None` will raise a `TypeError` during unpacking.
4. **Out-of-Bounds Stage ID Check**: `_split_stage_ids_by_type` directly indexes `self.engine.stage_clients[sid]` without validating whether `sid` is within valid bounds.

vllm_omni/engine/stage_pool.py
```diff
@@ -1222,6 +1222,15 @@ async def collective_rpc(
                 "error": f"stage {self.stage_id} replica {replica_id} is not attached",
             }
         try:
+            if self.stage_type != "diffusion" and method != "collective_rpc":
+                client_method = getattr(client, f"{method}_async", None)
+                if callable(client_method):
+                    if args and kwargs:
```
SamitHuang: If args is None, unpacking *args on line 1231 will raise a TypeError because args is not normalized to () like kwargs is normalized to {}. Additionally, the timeout parameter passed into StagePool.collective_rpc is omitted when invoking client_method. Consider normalizing args via args = tuple(args or ()) and simplifying the dispatch to return await client_method(*args, **kwargs).

vllm_omni/entrypoints/async_omni.py
```diff
@@ -1090,38 +1119,87 @@ async def _abort(self, request_ids: list[str]) -> None:
         if self.log_stats:
             logger.info("[AsyncOmni] Aborted request(s) %s", ",".join(request_ids))
 
+    def _split_stage_ids_by_type(self, stage_ids: list[int] | None = None) -> tuple[list[int], list[int]]:
+        """Split stage ids into AR/LLM (EngineCore) vs diffusion (worker RPC)."""
+        if stage_ids is None:
+            stage_ids = list(range(len(self.engine.stage_clients)))
+        ar_stage_ids: list[int] = []
+        diffusion_stage_ids: list[int] = []
+        for sid in stage_ids:
```
SamitHuang, line 1141: _split_stage_ids_by_type accesses self.engine.stage_clients[sid] directly without validating whether sid is within valid bounds (0 <= sid < len(self.engine.stage_clients)). Passing an out-of-range stage ID will raise an unhandled IndexError instead of a clear validation error. Consider adding a boundary check or raising a ValueError for unknown stage IDs.

vllm_omni/entrypoints/async_omni.py
knlnguyen1802: Fix in 5523e0ca6c35116c7f8ba0a031f2b4313f59541f

vllm_omni/entrypoints/async_omni.py
knlnguyen1802: Fix in 5523e0ca6c35116c7f8ba0a031f2b4313f59541f

vllm_omni/entrypoints/async_omni.py
knlnguyen1802: Fix in 5523e0ca6c35116c7f8ba0a031f2b4313f59541f

SamitHuang (commented):
Do we have an end2end test for the pause / resume and sleep / wake interfaces?

knlnguyen1802:
> Do we have an end2end test for the pause / resume and sleep / wake interfaces?

I can add e2e test, need to check time duration of the test

SamitHuang:
> > Do we have an end2end test for the pause / resume and sleep / wake interfaces?
> 
> I can add e2e test, need to check time duration of the test

How is the test result

knlnguyen1802:
> > > Do we have an end2end test for the pause / resume and sleep / wake interfaces?
> > 
> > 
> > I can add e2e test, need to check time duration of the test
> 
> How is the test result

currently I put the test on https://github.com/vllm-project/vllm-omni/pull/6367 

Test duration is about 2 mins

hsliuustc0106:
ci failed

knlnguyen1802:
@SamitHuang Fix ci please help trigger ci again

SamitHuang (approved):
LGTM

PR #6089, opened 2026-08-12
[Model] Add native full-duplex Nemotron VoiceChat serving

Sy0307:
## Purpose

Add model-native full-duplex Realtime serving for
`nvidia/NVIDIA-NemotronLabs-VoiceChat-11B` on top of the merged three-stage
offline pipeline.

The native path consumes one 1280-sample, 16 kHz PCM frame every 80 ms and
keeps a resumable Stage-0 request across wakes. It preserves Thinker, Talker,
and Code2Wav state, projects listen/speak plus Stage-2 audio events, and
supports the NVIDIA `AVAILABLE_TOOLS` / `TOOLCALL` / `TOOL_RESPONSE` contract.
The server validates and accepts client-owned `function_call_output` events,
queues concurrent tool results, and reinjects them into the live model request;
it does not execute arbitrary tools itself.

The capability contract reports `supports_core_resumable_request=true` and
`supports_core_kv_lease=false`: vLLM core retains request state across
resumable updates, while this adapter does not claim a public reusable KV lease.

## Validation added in this update

- Rebased onto current `main` (`678aa258`); the PR is `behind 0`.
- Preserved the offline async-stream completion contract while keeping native
  duplex segment completion, with regression coverage for both modes.
- Made queued connector sends abort-safe when an external request id is reused.
- Queued overlapping function results instead of raising a stage-fatal error,
  and made a failed runtime delivery retryable by recording the result only
  after `conversation.item.created` is emitted.
- Strengthened the generic full-model Nemotron gate to require 48 fixed 80 ms
  PCM16 packets, RMS at least `1e-3`, and output duration proportional to the
  input. No model-specific nightly pipeline was added.
- Kept total PR test additions at `+998/-27` relative to `origin/main`.

## Test plan

H200 targeted regression suite:

```bash
python -m pytest -o addopts= -q \
  tests/e2e/features/fullduplex/nemotron_voicechat/test_runtime_contract.py \
  tests/model_executor/stage_input_processors/test_nemotron_voicechat.py \
  tests/entrypoints/openai_api/test_duplex_handler.py
```

Result: `306 passed`.

H200 real-checkpoint validation used
`/root/models/NVIDIA-NemotronLabs-VoiceChat-11B`:

- Offline streaming: 41.20 s output, RMS `0.008171`; byte-identical to the
  current-main baseline.
- Native Realtime duplex: 190 input frames, 48 audio packets, 169344 PCM16
  bytes, RMS `0.007909`, 2 speak events, and 2 completed responses.

Static checks:

```bash
ruff check <changed Python files>
ruff format --check <changed Python files>
python tools/pre_commit/check_test_marks.py <changed test files>
python tools/pre_commit/check_spdx_header.py <changed Python files>
python tools/pre_commit/check_forbidden_imports.py <changed Python files>
python tools/pre_commit/check_torch_cuda.py <changed Python files>
git diff --check
```

Manual model-level command (the existing generic full-model selector collects
this test; this PR does not add a model-specific nightly job):

```bash
pytest -s -v tests/e2e/online_serving/test_nemotron_voicechat_duplex.py \
  -m 'full_model and H100 and omni' --run-level full_model
```

Manual tool-call round trip:

```bash
python tests/e2e/online_serving/nemotron_voicechat_realtime_duplex.py \
  --model nemotron-voicechat \
  --input-wav /path/to/NVIDIA-NemotronLabs-VoiceChat-11B/tool_call.wav \
  --expect-function-call \
  --expected-function-name generate_random_number \
  --expected-function-arguments '{"min":1,"max":50}' \
  --function-output 20 \
  --expected-post-tool-text "random number"
```

Known deployment limitation: the shipped eager fp32 profile is functionally
streaming but not wall-clock realtime on the validation hardware. The default
native deployment allows one active session and does not provide deterministic
barge-in.

**vLLM Version:** 0.27.0

**vLLM-Omni base:** `678aa25864334286d778fb3853b3c34f4b2ef0a6`


hsliuustc0106:
cleanup it first

Sy0307:
PTAK @yuekaizhang @linyueqian

linyueqian (commented):
Reviewed `03703356`. I ran the PR's own test surface on an H200 (vLLM 0.26.0, torch 2.11+cu130):

```
tests/e2e/features/fullduplex/nemotron_voicechat/
tests/model_executor/models/test_nemotron_voicechat_shapes.py
tests/model_executor/stage_input_processors/test_nemotron_voicechat.py
tests/entrypoints/openai_api/test_duplex_handler.py
tests/core/sched/test_omni_ar_scheduler_streaming.py
tests/worker/test_gpu_ar_model_runner.py
-> 316 passed
```

No findings from me. The split between `experimental/fullduplex/nemotron_voicechat/` and the shared `experimental/fullduplex/openai/` layer looks like the right shape, and the shared serving code stays free of model-named branching, which is what we want as more duplex models land.

Two things before this can merge, neither a code comment:

The branch is `CONFLICTING` against main and needs a merge from it. Worth doing soon given how much has landed in the last two weeks.

Buildkite has not reported yet. Since this is a fork PR, a build only fires on a label event **after** the most recent push, so if the `ready` label predates your last commit it will sit pending indefinitely. Toggling the label off and on again is enough to kick it. I have hit this on three PRs this week, including my own, where everything looked green while no test had actually run.

I have not done a deep read of the duplex runtime and data-plane logic beyond the test run, so treat this as coverage-and-integration feedback rather than a full correctness review of the new runtime. Happy to go deeper on `data_plane.py` and `runtime.py` if useful.

Sy0307:
The current pre-commit run is still red: `check-forbidden-imports` rejects `import base64` in `vllm_omni/experimental/fullduplex/nemotron_voicechat/input.py:8`; use the repository-approved `pybase64` import before merge. Separately, the added model E2E proves speech output but not the advertised tool/function-call round trip.

vllm_omni/model_executor/models/nemotron_voicechat/nemotron_voicechat_talker.py
```diff
@@ -333,25 +360,24 @@ def _step_session(self, session: dict[str, Any]) -> tuple[torch.Tensor, bool]:
...
         session["past_key_values"] = past_key_values
         session["step"] = t + 1
-        # Async mode: exhausting the received timeline only ends the request
-        # once the upstream marked its final chunk; otherwise more entries are
-        # in flight and the scheduler parks us until they arrive.
-        finished = (t + 1) >= int(session["timeline"].numel()) and bool(session["upstream_finished"])
+        # Exhausting the currently received timeline ends this scheduler
+        # segment. A resumable async-chunk request is then parked with its
+        # model-side TTS state intact until the next thinker frame arrives.
+        finished = (t + 1) >= int(session["timeline"].numel())
```
linyueqian: [blocking] Dropping `and bool(session["upstream_finished"])` here is not gated on duplex, so it also changes the pre-existing async-chunk offline streaming path. In that mode the thinker streams a growing cumulative timeline, so the talker now reports finished at the first timeline exhaustion instead of waiting for the upstream terminal marker. `embeds[:, 0] = 1.0` then emits the stop token and the request ends after its first chunk.

Measured on 1x H200 with `nemotron_labs_voicechat_streaming.yaml` and `turn_taking.wav`: main produces 41.20 s of speech (RMS 8.171e-03), this head produces 0.16 s at RMS 2.477e-05. Emitted text is byte-identical, so only the audio path is lost.

Gating it keeps the behavior you need for native duplex and restores the old contract everywhere else. With this change the offline run reproduces main byte for byte:

```python
finished = (t + 1) >= int(session["timeline"].numel()) and (
    bool(session["codec_streaming"]) or bool(session["upstream_finished"])
)
```

Sync mode already has `upstream_finished=True`, so it is unaffected.

vllm_omni/model_executor/models/nemotron_voicechat/nemotron_voicechat_talker.py
```diff
@@ -388,21 +414,32 @@ def preprocess(
...
         if steps_run == 0:
             # Zero-progress wake (duplicate/terminal-marker chunk): no TTS step
-            # was run, so emit a plain CONTINUE placeholder — never a
-            # fabricated frame. The request finishes only when the upstream is
-            # exhausted AND the timeline is fully consumed.
-            finished = bool(session["upstream_finished"])
+            # was run, so never fabricate a frame. Stop the current scheduler
+            # segment when its timeline is already consumed; a resumable
+            # request will wake again when the connector provides more input.
+            finished = int(session["step"]) >= int(session["timeline"].numel())
```
linyueqian: [blocking] Same issue as the zero-progress counterpart of line 380, and it needs the same gate. Previously a zero-progress wake only finished when the upstream was exhausted; now a duplicate or terminal-marker chunk arriving with the timeline already consumed ends the request outright on the offline streaming path.

```python
finished = int(session["step"]) >= int(session["timeline"].numel()) and (
    bool(session["codec_streaming"]) or bool(session["upstream_finished"])
)
```

vllm_omni/model_executor/models/nemotron_voicechat/nemotron_voicechat_code2wav.py
```diff
@@ -176,19 +190,42 @@ def forward(
...
             lens = torch.tensor([frames], device=device, dtype=torch.long)
+            codec_cache = None
+            if codec_streaming:
+                if not isinstance(cache_request_id, str) or not cache_request_id:
+                    raise ValueError("Nemotron VoiceChat incremental codec input lacks meta.request_id")
+                from vllm_omni.model_executor.models.nemotron_voicechat.nemo_vendored.ear_tts_vae_codec import (
+                    CausalConv1dCache,
+                )
+
+                codec_cache = self._duplex_codec_caches.setdefault(cache_request_id, CausalConv1dCache())
```
linyueqian, line 201: [blocking] The native duplex audio channel produces digital silence. For 15.2 s of input I get 0.80 s of audio at RMS 2.509e-05, and the entire WAV contains only three distinct PCM16 values (`{-1: 11921, 0: 5716, 1: 3}`), so peak amplitude is one LSB. The text channel in the same session is correct. Reproduced with this PR's own probe and with an independent browser client.

To save you time: this cache is not the cause. I decoded real codes frame by frame through a shared `CausalConv1dCache`, exactly as this path does, and compared against a one-shot full decode. They are numerically identical (both rms 9.2514e-03, peak 1.5129e-01), so the streaming decode contract itself is sound.

The problem is what reaches it. Debug traces show this stage receiving 12 inputs of shape `(1, 31)`, each correctly emitting 1764 samples, while the talker's cumulative history only ever reaches 31 rows (`prompt_len=19`) across 190 input frames. The text channel covered about 117 frames over the same period. So the audio channel covers under 10% of the session, and those few frames are the leading listen region, which is silent by nature. That is the same early-stop signature as the talker finding, so please check whether the gate there also fixes this.

One more difference worth reconciling with the reference: `nemo_vendored/duplex_ear_tts.py:778-792` calls `replace_control_speech_codes(...)` on every incremental decode, mapping control codes to valid silence codes before the codec sees them. That helper is never called anywhere in the production path.

tests/e2e/online_serving/test_nemotron_voicechat_duplex.py
```diff
@@ -0,0 +1,56 @@
...
+)
+def test_native_duplex_turn_taking_streams_model_audio(omni_server, model_prefix: str, tmp_path: Path) -> None:
+    root = Path(model_prefix) / MODEL if model_prefix else Path(MODEL)
+    if not root.is_dir():
+        root = Path(snapshot_download(MODEL, local_files_only=True))
+    args = parse_args(
+        (
+            f"--url ws://{omni_server.host}:{omni_server.port}/v1/realtime --model {omni_server.model} "
+            f"--input-wav {root / 'turn_taking.wav'} --input-channel 0 --max-frames 190 "
+            f"--minimum-audio-chunks 10 --minimum-audio-rms 0.00001 --no-realtime --timeout-s 300 "
```
linyueqian: [blocking] This threshold cannot detect the silence described above. One PCM16 LSB is about 3.05e-05, and `--minimum-audio-rms 0.00001` sits below it, so a digitally silent stream passes. It also overrides the probe's own `1e-4` default in the looser direction.

I ran this test's exact probe invocation against the broken output and all six assertions pass: `ok is True`, `input_frames == 190`, `audio_bytes > 0`, `response.speak > 0`, `response.audio.delta >= 10`, `response.done > 0`. The measured RMS was 2.509e-05, and `response.audio.delta` came in at exactly 10 against an `>= 10` assert, so there is no margin there either.

Since this is the only full-model gate for the feature, it would be worth an RMS floor around 1e-3 plus an assertion that output duration tracks the input frame count. Both would have caught this.

Separately, I could not find any `nemotron_voicechat` reference in `.buildkite`, so this test is only picked up by the generic `tests/e2e -m "full_model and H100 and omni"` nightly selector, and there is no offline or streaming e2e test for the model at all.

vllm_omni/model_executor/models/nemotron_voicechat/nemotron_voicechat_thinker.py
```diff
@@ -324,6 +434,235 @@ def _fuse(
...
+        try:
+            generation = int(runtime_config.get("nvc_function_response_generation", 0))
+        except (TypeError, ValueError) as exc:
+            raise ValueError("Nemotron VoiceChat function-response generation must be an integer") from exc
+        seen = int(session.get("function_response_generation", 0))
+        if generation < seen:
+            raise ValueError("Nemotron VoiceChat function-response generation moved backwards")
+        if generation > seen:
+            if session.get("forced_function_token") is not None or session.get("forced_function_tokens"):
+                raise ValueError("Nemotron VoiceChat received overlapping function responses")
```
linyueqian: [blocking] This raise is reachable through the tool flow the PR itself advertises, and it takes the server down rather than failing one request.

`_render_tool_prompt` explicitly instructs the model to emit multiple calls in one block (`<TOOLCALL>[{"name": "tool_name1", ...}, {"name": "tool_name2", ...}]</TOOLCALL>`), and `data_plane._project_function_token` emits one `function_call` event per parsed call. A client answering both sends two `function_call_output` items, each incrementing `nvc_function_response_generation`. The first response drains one token per 80 ms frame, and a `<TOOL_RESPONSE>[...]` serialization is roughly 15 to 30 tokens, so the second lands inside the 1.2 to 2.4 s replay window and hits this branch. `_render_tool_response` wraps a single output as `[value]`, so a client cannot batch two results into one item to avoid it either.

The failure is not request-local: `stage_engine_core_proc.py` wraps `run_busy_loop()` in `except Exception`, logs a fatal error, calls `engine_core._send_engine_dead()` and re-raises, so the stage process exits. With the shipped single replica and `max_num_seqs: 1` every later request fails too.

Queueing the second response behind the first, or rejecting it at the serving layer with a proper Realtime error, would both keep this in-process.

vllm_omni/distributed/omni_connectors/transfer_adapter/chunk_transfer_adapter.py
```diff
@@ -931,7 +968,18 @@ def finish_requests(
...
+        # later request reusing the same external id starts with empty state.
+        for req_id in request_ids:
+            request = requests.get(req_id) if requests else None
+            external_req_id = (
+                getattr(request, "external_req_id", None)
+                if request is not None
+                else self.request_ids_mapping.get(req_id)
+            ) or req_id
+            self.cleanup_receiver(req_id)
+            self.cleanup_sender(external_req_id)
```
linyueqian, line 1121: [important] This new unconditional `cleanup_sender()` on the abort path races the background save thread, and it affects every async-chunk model, not just this one.

`save_async` has already queued tasks into `_pending_save_reqs`, and that queue is deliberately preserved across cleanup: `test_cleanup_preserves_pending_save` asserts it explicitly, with the comment "Cleanup must NOT remove `_pending_save_reqs` to avoid losing unsent chunks". So a queued task can run after this line. `put_req_chunk` is a `defaultdict(int)`, so `chunk_id` at line 367 comes back as 0 and the task re-populates the state the abort just cleared, writing under `{external_req_id}_{stage_id}_0`. If a later request reuses that external id it can pick up the aborted stream's stale chunk 0.

The processor side compounds it: with `request_payload` popped, `talker2code2wav_async_chunk` sees `state is None`, resets `frames_seen` to 0, and re-ships the full accumulated history into the stateful codec cache instead of the unseen suffix.

The docstring on `cleanup_sender` at lines 438-439 states the invariant this breaks. Either drop or await the queued tasks for that external id before reclaiming sender state.

linyueqian (changes requested):
Reviewed `1c57ccc6` and ran it end to end on 1x H200 (vLLM 0.27.0, torch 2.13.0+cu130) with the real checkpoint, plus a separate browser WebSocket client against the live Realtime endpoint.

Two blocking results, both measured rather than read off the diff.

**This PR regresses the existing offline streaming path.** Same fixture, same deploy yaml (`nemotron_labs_voicechat_streaming.yaml`), same environment, only the tree differs:

| tree | samples | duration | RMS | peak |
| --- | --- | --- | --- | --- |
| main `3d035bfa` | 908,460 | 41.20 s | 8.171e-03 (-41.8 dBFS) | 2.076e-01 |
| this PR `1c57ccc6` | 3,528 | 0.16 s | 2.477e-05 (-92.1 dBFS) | 9.155e-05 |

The emitted text is byte-identical between the two, so the thinker is fine and the entire loss is in the talker to code2wav path. 3,528 samples is exactly 2 x `wav_per_frame` (1764), i.e. the talker stops after about two timeline steps. I bisected it to the `upstream_finished` removal: restoring that guard, gated so native duplex keeps its new per-segment behavior, produces a **byte-identical WAV to main** (`cmp -s` passes). Details inline.

**The native duplex audio channel emits digital silence.** The text channel is correct in the same session ("Hello! How can I help you today?", "Yes, I can share a simple chocolate chip recipe if you'd like."), but the audio is 0.80 s for 15.2 s of input at RMS 2.509e-05. The whole WAV contains three distinct PCM16 values, `{-1: 11921, 0: 5716, 1: 3}`, so peak amplitude is one LSB. Reproduced independently by this PR's own probe and by a hand-written browser client.

I want to rule one thing out for you, because it is the natural first suspect: the codec streaming API is **not** at fault. Decoding frame by frame through a shared `CausalConv1dCache`, exactly as the new code2wav path does, is numerically identical to a one-shot full decode on real codes (both rms 9.2514e-03, peak 1.5129e-01). The real signal is upstream: server debug traces show the talker's cumulative history reaching only 31 rows (`prompt_len=19`, so about 12 acoustic rows) across 190 input frames, while the text channel covered roughly 117 frames. The audio channel covers under 10% of the session, and those frames are the leading listen region, which is naturally silent. That is the same starvation as the offline regression above, so it is worth checking whether the first fix also resolves this one.

Worth noting separately: the vendored reference calls `replace_control_speech_codes(...)` before every incremental decode (`nemo_vendored/duplex_ear_tts.py:778-792`), and that function is never called anywhere in the production path.

For validation context: the 396 CPU tests on the changed surface pass at this head, and I could not find any `nemotron_voicechat` reference in `.buildkite`, so the only CI coverage is the generic `tests/e2e -m "full_model and H100 and omni"` nightly selector. There is no offline or streaming e2e test for this model at all, which is why the regression above was invisible.

I also checked and found nothing wrong in `_resume_downstream_chunk_receiver`'s counter accounting (the decrement balances upstream's increment and there is no double enqueue), and the `meta.codec_streaming` gate in the connector is inert for other models today since no other processor sets that field on a returned payload.

vllm_omni/experimental/fullduplex/openai/realtime_input.py
```diff
@@ -1052,6 +1052,45 @@ async def _conversation_item_to_duplex(self, event: dict[str, object]) -> dict[s
...
+                        ),
+                        event_id=event.get("event_id"),
+                        param="item.output",
+                    )
+                )
+                return None
+            # ``function_call_output`` is a non-message item and returns below
+            # before the normal history path; retain it here so the same
+            # call_id cannot be submitted twice.
+            self._conversation_items[item_id] = item
```
congw729: **[P2] Record the function result only after runtime delivery succeeds**

This inserts the `function_call_output` into `_conversation_items` before `session_runner` has delivered its runtime update. If `wait_for_native_append_tail()` or `_signal_runtime_session()` fails, no `conversation.item.created` event is emitted and the result never reaches the model, but this provisional entry remains. Retrying the same `call_id` is then rejected by `duplicate_output`, so a transient delivery failure permanently prevents the live request from receiving that tool result. Please defer recording the output until the runtime update succeeds, or roll this entry back on every failure path.

congw729 (commented):
One additional tool-output lifecycle issue.

tests/e2e/online_serving/test_nemotron_voicechat_duplex.py
```diff
@@ -0,0 +1,58 @@
...
+        ).split()
+    )
+    result = asyncio.run(run(args))
+    assert result["ok"] is True
+    assert result["input_frames"] == 190
+    assert result["audio_bytes"] > 0
+    assert result["event_counts"]["response.speak"] > 0
+    assert result["event_counts"]["response.audio.delta"] >= 10
+    assert result["event_counts"]["response.done"] > 0
+    assert result["audio_bytes"] >= result["input_frames"] * 1764 * 2 // 4
```
linyueqian, line 56: [suggestion] This assertion does the right thing and it is the check that would have caught the silence, so this is only about where the constant sits.

On my run it passes by 1.05%: `audio_bytes` came in at 169,344 against a required 167,580. The `// 4` puts the bar at 25% of frame-locked coverage, which is almost exactly what the implementation currently emits (25.3%), so the threshold is calibrated to today's behavior rather than to a target. Any small drop in coverage turns this red without a real regression.

Related question rather than a request: across the same run I counted 48 audio frames against 102 spoken transcript deltas. If the intent is that the acoustic channel is frame-locked to the text channel, that ratio looks worth a second look. If instead the gap is just the model listening for most of the fixture, then this is expected and only the margin comment above applies.

vllm_omni/distributed/omni_connectors/transfer_adapter/chunk_transfer_adapter.py
```diff
@@ -334,6 +364,16 @@ def _poll_single_request(self, request: Request):
...
 
     def _send_single_request(self, task: dict):
+        request = task["request"]
+        external_req_id = request.external_req_id
+        with self._sender_state_lock:
+            sender_token = task.get("sender_token")
+            if sender_token is not None and self._sender_tokens.get(external_req_id) is not sender_token:
+                logger.debug("Discarding stale queued chunk for aborted request %s", external_req_id)
+                return
+            self._send_single_request_locked(task)
```
linyueqian: [question] The token check that invalidates stale queued chunks is exactly right, and it closes the abort race cleanly.

What I would like your read on is the lock scope. `_send_single_request_locked` runs the whole payload build and `self.connector.put(...)` while still holding `_sender_state_lock`, and `save_async` takes that same lock on the scheduler thread on every step. For this model that is harmless, since native duplex runs a single session at `max_num_seqs: 1`.

The concern is that this is shared plumbing. On a high concurrency async-chunk deployment such as qwen3_tts, the scheduler would block behind a full payload build plus connector write on every save. Holding the lock only for the token check and the counter updates, and dropping it before the put, would keep the same guarantee without putting the transport inside the critical section. Fine as a follow-up if you would rather not touch it here.

linyueqian (approved):
Re-reviewed at `5370fda93` and re-ran the same end to end checks on 1x H200. All five findings from my previous review are fixed, verified by measurement rather than by reading the diff. Lifting my `CHANGES_REQUESTED`.

| finding | status | evidence at `5370fda93` |
| --- | --- | --- |
| offline streaming regression | fixed | offline WAV is now **byte-identical to main** (`cmp -s` passes), 41.20 s at RMS 8.1711e-03 |
| duplex audio was digital silence | fixed | RMS 2.509e-05 to **8.4835e-03** (-41.4 dBFS), duration 0.80 s to 3.84 s, distinct PCM16 values 3 to **3091**, `response.audio.delta` 10 to 48 |
| nightly gate could not detect it | fixed | `--minimum-audio-rms 0.001 --minimum-audio-chunks 48` plus the new duration assertion |
| multi tool call killed the stage process | fixed | the raise is replaced by the `nvc_function_response_batches` queue with contiguity checks |
| `cleanup_sender` raced the save thread | fixed | `_sender_state_lock` plus the `_sender_tokens` generation token, stale queued chunks are discarded |

The CPU surface is 399 passed, up from 396.

One correction to my own earlier reasoning, since it is worth putting on the record. I argued the duplex silence could not be fixed by the talker gate, because duplex sets `codec_streaming=True` and the gate is a no-op on that path. The talker path is indeed unchanged for duplex, and the audio is fixed anyway, so the cause was somewhere in the connector and engine changes rather than in the talker. I never identified the mechanism, only confirmed the symptom is gone. If you know which change actually fixed it, it would be worth a line in the commit message, because that is the part a future regression test should pin.

Two non-blocking notes inline, neither of which should hold up the merge.

The one thing still outstanding is not a review item: the branch is currently `CONFLICTING` against main and needs a merge before it can go in.

PR #5885, opened 2026-08-07
[Feature] Support text encoder Disaggregation for H3

gcanlin:
## Purpose

Close #5707.

This PR disaggregates the MiniMax H3 text encoder from the diffusion pipeline and runs it with the native vLLM model runner. The diffusion stage consumes the encoded hidden states and token-role tags instead of loading and maintaining a second Transformers-based Qwen3-VL implementation.

## Overview

```mermaid
flowchart LR
    Request["Ref2VA request<br/>text + image/video + audio"]

    subgraph Stage0["Stage 0: vLLM Text Encoder"]
        Processor["Qwen3-VL processor"]
        Runner["MiniMaxH3TextEncoder<br/>vLLM model runner · TP"]
        Conditioning["Layer-50 residual stream<br/>hidden states + token-role tags"]
        Processor --> Runner --> Conditioning
    end

    Bridge["Conditioning bridge<br/>preserve original media inputs"]

    subgraph Stage1["Stage 1: MiniMax H3 Diffusion"]
        Inline["Inline DiffusionEngine<br/>USP"]
        DiT["Video/audio DiT denoising"]
        VAE["Video VAE + Audio VAE"]
        Inline --> DiT --> VAE
    end

    Request --> Processor
    Conditioning --> Bridge --> Inline
    Request -. "original media" .-> Bridge
    VAE --> Response["MP4 response"]
```

```mermaid
sequenceDiagram
    participant API as OpenAI-compatible API
    participant TE as Stage 0: vLLM Text Encoder
    participant OR as Orchestrator
    participant DIFF as Stage 1: Inline Diffusion

    API->>TE: Canonical H3 presentation + visual inputs
    TE-->>OR: BF16 hidden states + token-role tags
    OR->>DIFF: Original request + external conditioning
    DIFF->>DIFF: DiT denoising and VAE decode
    DIFF-->>API: In-process result, then MP4 encoding
```

Key changes:

- Adds `MiniMaxH3TextEncoder`, built on the upstream vLLM `Qwen3VLForConditionalGeneration`, and returns the unnormalized residual stream after layer 50.
- Reuses the vLLM Qwen3-VL processor, TP execution, scheduler, paged attention, prefix cache, multimodal kernels, and model implementation.
- Adds the H3 presentation adapter and a typed conditioning contract containing `[tokens, 5120]` hidden states and token-role tags.
- Allows the diffusion pipeline to consume external text conditioning without loading its local text encoder.
[...]

PR #6093, opened 2026-08-12
[Core] Add NIXL omni connector

yuanwu2017:
## Purpose

Adds `NixlConnector`, an optional `OmniConnectorBase` implementation for moving arbitrary stage payloads across nodes through [NIXL](https://github.com/ai-dynamo/nixl).

RFC: #6160

## When to Choose It

Use `NixlConnector` for intra-node or inter-node stage transfer when a deployment already uses NIXL for vLLM P/D, needs UCX transport coverage (RDMA, TCP, and accelerator-native paths), or targets Intel XPU through UCX Level Zero.

| Connector | Best fit | Constraint compared with NIXL |
| --- | --- | --- |
| `SharedMemoryConnector` | Colocated stages | POSIX shared memory only; not cross-node |
| Mooncake connectors | Deployments standardised on Mooncake | Requires Mooncake runtime and its memory model |
| Mori connector | Deployments standardised on Mori | Requires Mori IOEngine |
| Yuanrong connectors | Ascend/Yuanrong deployments | NPU- and Datasystem-specific |
| `NixlConnector` | Cross-node, NIXL/UCX-enabled deployments | Requires NIXL plus a usable UCX backend and a small ZMQ control plane |

NIXL is not a replacement for existing connectors. It remains an optional lazy import and is only constructed when a deployment YAML names it. It implements Omni `put`/`get` for arbitrary payloads; it cannot reuse vLLM P/D's NIXL connector because that interface operates on block-table, layer-wise KV semantics.

## Design and Deployment Constraints

- **Control plane:** producer ZMQ `ROUTER`, consumer `REQ`. A consumer obtains metadata with `GET_META{key}` and reports `XFER_DONE` after successful consumption. Callers that forward `put()` metadata directly do not need a handshake socket. Ordinary cross-stage payload requests carry the producer endpoint through the scheduler request path.
- **Data plane:** consumer-initiated NIXL `READ`. The producer registers source memory; the consumer allocates destination tensors on its own device and pulls data.
- **Ports:** `zmq_port` is a base port. The centralized resolver derives a unique endpoint from purpose (`request_forwarding`, `kv_transfer`, or `orchestrator`), producer stage, replica, and TP local rank. This prevents collisions between TP workers and connector purposes.
- **Payloads:** tensors, structured Python objects with tensor leaves, and serialized objects. Metadata schema v1 groups descriptors by NIXL memory type and retains global tensor indices, so structured payloads may safely combine CPU and accelerator tensors.
- **Ownership:** source registrations have a configurable lease (`lease_seconds`, default 3600; `VLLM_OMNI_NIXL_LEASE_S`). A reaper releases abandoned registrations even on an idle producer.
- **Timeout safety:** NIXL 1.3 has no supported cancellation API. On a receive timeout, the connector retains buffers, registrations, descriptor lists, transfer handles, and the remote agent until NIXL reaches a terminal state; `close()` waits before releasing them. This prevents native DMA into deregistered memory.
- **Installation:** CUDA deployments can install the published `nixl` package. `docker/Dockerfile.xpu` uses the upstream vLLM XPU image, which already ships a matched torch-xpu, UCX Level Zero, and NIXL runtime; this PR adds no custom UCX/NIXL wheel build.

## Validation Matrix

| Scenario | Coverage | Result |
| --- | --- | --- |
| ZMQ metadata handshake and completion cleanup | Stubbed NIXL unit tests | Passed |
| Tensor, object, and structured payload normalization | Stubbed NIXL unit tests | Passed |
| Mixed CPU/accelerator descriptor grouping | `test_structured_payload_groups_descriptors_by_memory_type` | Passed |
| Producer lease expiry while idle | `test_idle_producer_lease_expires_without_another_put` | Passed |
| Timeout, deferred native-resource ownership, and shutdown | NIXL timeout lifecycle tests | Passed |
| TP=1 / TP>1 port isolation | adapter and worker resolver tests | Passed |
| Heterogeneous TP source-rank endpoint selection | TP rank-aware tests | Passed |
| Ordinary payload producer endpoint propagation | coordinator and worker endpoint tests | Passed |
| Intra-node real NIXL transfer | Two-process native NIXL/UCX, GPU 0 to GPU 1 | Passed |
| Inter-node RDMA/TCP transfer | Not run across hosts | Not validated |
| Real CUDA GPU transfer | Two CUDA devices with native NIXL/UCX | Passed |
| Real Intel XPU transfer | Not run on target hardware | Not validated |
| Representative connector integration | Actual OmniPayloadStruct with mixed CPU/GPU leaves | Passed |

Validation run in the CUDA test image:

```text
235 passed, 3 subtests passed
```

Native smoke: `1 passed`; focused NIXL and affected transport suites included above. Ruff checks, Ruff format checks, and `git diff --check` passed.

## Non-goals

- Broadcast from one producer to multiple consumers.
- Replacing existing Omni connectors.
- Making NIXL a required dependency.


hsliuustc0106:
open an RFC first

yuanwu2017:
> open an RFC first

Ok. Let me finish the integration with https://github.com/vllm-project/vllm-omni/pull/5885.

yuanwu2017:
> open an RFC first

Done — filed as #6160.

I've also filled in the PR description. Two things worth flagging for review:

- This PR is the connector only. The first consumer (MiniMax-H3 text-encoder → diffusion conditioning, on top of #5885) is kept in a separate branch so this one stays reviewable on its own.
- Validate the Wan mode with #6285

yuanwu2017:
@hsliuustc0106 @xuechendi Please help to review the PR.

vllm_omni/distributed/omni_connectors/connectors/nixl_connector.py
```diff
@@ -0,0 +1,683 @@
...
+
+    def _notify_transfer_done(self, get_key: str, metadata: dict[str, Any]) -> None:
+        """Tell the producer its buffer is drained so it can deregister now.
+
+        Without this the producer would hold the registration until the lease
+        expires, which for long-lived stages means unbounded growth.
+        """
+        host = metadata.get("sender_host")
+        port = metadata.get("sender_zmq_port")
+        if not host or not port or self._zmq_ctx is None:
```
hsliuustc0106: [P1] In the documented mode without `zmq_port`, `put()` returns metadata without a sender endpoint. This early return skips the completion notification; because the handshake listener is also disabled, `_cleanup_expired_pending()` runs only during later `put()` calls. A successfully consumed or abandoned buffer can therefore remain registered indefinitely on an idle producer. Please add an explicit completion/cleanup path or an independent lease reaper.

vllm_omni/distributed/omni_connectors/connectors/nixl_connector.py
```diff
@@ -0,0 +1,683 @@
...
+        """Return complete NIXL transfer metadata for ``get_key``.
+
+        Callers that forwarded the ``put`` metadata are served directly; the
+        remaining cases fall back to the handshake, either at the endpoint
+        named in a partial metadata dict or at the configured default producer.
+        """
+        if isinstance(metadata, dict) and metadata.get("schema_version") == _SCHEMA_VERSION:
+            return metadata
+
+        if isinstance(metadata, dict) and metadata.get("sender_host") and metadata.get("sender_zmq_port"):
```
hsliuustc0106: [P1] `OmniKVTransferManager.receive_kv_cache_for_request()` passes partial metadata as `source_host/source_port` for each source rank. This resolver ignores those keys and falls back to one `_sender_*` endpoint, so TP4→TP2 queries source-rank 1/3 through the local target-rank port and cannot receive those shards. Please accept the `source_*` form or preserve per-rank endpoints.

vllm_omni/distributed/omni_connectors/connectors/nixl_connector.py
```diff
@@ -0,0 +1,683 @@
...
+            endpoint = (str(metadata["sender_host"]), int(metadata["sender_zmq_port"]))
+        elif self._sender_host and self._sender_zmq_port:
```
hsliuustc0106: [P1] This worker receive path cannot resolve the producer endpoint. `OmniConnectorModelRunnerMixin._send_single_request()` drops the metadata returned by `put()`, and its receive path calls `get(..., metadata=None)`, while `get_stage_connector_config()` only adds `role="receiver"` (no `sender_host`/`sender_zmq_port`). Consequently the documented config has neither metadata nor a handshake endpoint and every ordinary payload stalls. Please propagate the returned metadata or initialize the receiver with the producer endpoint.

vllm_omni/distributed/omni_connectors/connectors/nixl_connector.py
```diff
@@ -0,0 +1,683 @@
...
+    def _init_handshake(self) -> None:
+        """Set up the ZMQ control plane used when ``get`` receives no metadata.
+
+        The handshake stays off unless ``zmq_port`` is configured, so
+        deployments that forward the ``put`` metadata themselves keep their
+        current behaviour and open no extra sockets.
+        """
+        role = self.config.get("role")
+        self._role = str(role).lower() if role else None
+        self._zmq_port = self.config.get("zmq_port")
```
hsliuustc0106: [P1] The async-chunk scheduler constructs the connector through `OmniChunkTransferAdapter.create_connector()` (which passes this config directly), so it bypasses `_rank_aware_extra()`. Each TP sender therefore binds this raw YAML port; `NixlConnector` also does not expose the `stage_id` that the adapter reads. TP1 can fail with `AttributeError`, and TP>1 with `EADDRINUSE`; please centralize connector creation/port resolution. If payload and KV use the same connector, their two sender instances also bind the same computed port.

vllm_omni/distributed/omni_connectors/connectors/nixl_connector.py
```diff
@@ -0,0 +1,683 @@
...
+                header_tensor = torch.frombuffer(header, dtype=torch.uint8).clone().contiguous()
+                tensors, tensor_specs = self._normalize_tensor_payload([header_tensor, *payload_tensors])
+                kind = _KIND_STRUCTURED
+            else:
+                payload = self.serialize_obj(data)
+                tensor = torch.frombuffer(payload, dtype=torch.uint8).clone().contiguous()
+                tensors, tensor_specs = self._normalize_tensor_payload(tensor)
+                kind = _KIND_OBJECT
+
+            memory_type = self._resolve_memory_type(tensors[0])
```
hsliuustc0106: [P1] Structured payloads put the serialized CPU skeleton at `tensors[0]`, so this selects `DRAM` even when subsequent tensor leaves are on an accelerator. That single memory type is then used for every NIXL registration/descriptor, making GPU leaves invalid. Please group registrations by memory type or reject mixed-memory structured payloads explicitly.

vllm_omni/distributed/omni_connectors/connectors/nixl_connector.py
```diff
@@ -0,0 +1,683 @@
...
+            self._metrics["bytes_transferred"] += size
+            logger.debug("NixlConnector get %s->%s key=%s size=%d", from_stage, to_stage, get_key, size)
+            return payload, size
+        except Exception:
+            self._metrics["errors"] += 1
+            logger.error("NixlConnector get failed for %s", get_key, exc_info=True)
+            return None
+        finally:
+            if xfer_handle is not None:
+                self._safe_call(self._agent.release_xfer_handle, xfer_handle)
```
hsliuustc0106, line 441: [P1] On a timeout, `release_xfer_handle()` can fail while NIXL still owns an active READ. Suppressing that failure and continuing to release descriptor lists/deregister the destination lets native DMA write into deregistered or freed memory. Please retain the transfer resources until abort succeeds or the handle reaches `DONE`.

hsliuustc0106:
Could you expand the PR description with the decision criteria for choosing `NixlConnector` over the existing connectors (SharedMemory, Mooncake, Mori, and Yuanrong), including the concrete use cases this enables? It would also help to summarize the pros/cons and deployment constraints: supported hardware and topologies, the extra ZMQ control plane, memory ownership/lease behavior, and the optional NIXL/UCX installation burden. Finally, please provide a validation matrix covering intra-/inter-node transfers, CPU/GPU/XPU paths, tensor/object/structured payloads, TP=1, TP>1, heterogeneous TP, timeout/cancellation/retry/cleanup, and at least one representative end-to-end case. This will make it clear when this connector is the right choice and what evidence supports the advertised behavior.

xuechendi:
@yuanwu2017 , we now move on to directly using vLLM-xpu docker, the nixl installation can be done by default docker image, may you check

vllm_omni/engine/stage_client.py
```diff
@@ -94,6 +94,8 @@ def get_kv_sender_info(
         kv_transfer_port_offset: int = ...,
     ) -> dict[str, Any] | None: ...
 
+    def get_payload_sender_info(self) -> dict[str, Any] | None: ...
```
hsliuustc0106, line 97: what's the function usage?

vllm_omni/engine/orchestrator.py
```diff
@@ -2151,6 +2151,7 @@ def _build_next_stage_request(
         *,
         mm_features: list | None = None,
         resumable: bool = False,
+        payload_sender_info: dict[str, Any] | None = None,
```
hsliuustc0106, line 2151: do we have to add this ?

vllm_omni/distributed/omni_connectors/utils/initialization.py
```diff
@@ -38,6 +39,81 @@
 # comfortably sized block per replica for TP-rank and stage offsets.
 KV_REPLICA_PORT_STRIDE = 1024
 
+REQUEST_FORWARDING_PORT_OFFSET = 0
+ORCHESTRATOR_PORT_OFFSET = 200
+
+ConnectorPurpose = Literal["request_forwarding", "kv_transfer", "orchestrator"]
+
+
+def connector_zmq_port(
```
hsliuustc0106: do we have similar function before or is this an abstraction?

hsliuustc0106:
```suggestion
def compute_connector_zmq_port(
```

hsliuustc0106: or resolve_connector_zmq_port

vllm_omni/engine/__init__.py
```diff
@@ -79,6 +79,7 @@ class OmniEngineCoreRequest(EngineCoreRequest):
     # GPUModelRunner.model_intermediate_buffer instead of using the deprecated
     # additional_information request transport.
     model_intermediate_buffer: dict[str, Any] | None = None
+    payload_sender_info: dict[str, Any] | None = None
```
hsliuustc0106, line 82: is this a have-to change?

vllm_omni/distributed/omni_connectors/transfer_adapter/chunk_transfer_adapter.py
```diff
@@ -254,11 +256,20 @@ def create_connector(cls, model_config: Any):
                 "extra": getattr(connector_config, "extra", {}),
             }
 
-        connector_specs = ConnectorSpec(
+        connector_spec = ConnectorSpec(
             name=connector_config.get("name", "SharedMemoryConnector"),
             extra=connector_config.get("extra", {}),
         )
-        return OmniConnectorFactory.create_connector(connector_specs)
+        stage_id = int(getattr(model_config, "stage_id", connector_spec.extra.get("stage_id", 0)))
```
hsliuustc0106, line 289: I thought chunk transfer is used for async_chunk, why we need to change it here?

lishunyang12 (commented):
I re-reviewed the current head (`dd63e8d7`). The earlier feedback fixes are mostly present, but I still think this needs changes before merge:

1. **[P0] Stage client construction currently fails universally.** [`stage_engine_core_client.py:156`](https://github.com/vllm-project/vllm-omni/blob/dd63e8d78f32b300d18b26a6346244632f339164/vllm_omni/engine/stage_engine_core_client.py#L156) calls `_build_payload_sender_info()` before `super().__init__()`, while that helper reads `self.vllm_config` and vLLM's `MPClient.__init__` is what assigns it. This raises `AttributeError` for every `StageEngineCoreClient`/`DPLBStageEngineCoreClient`, even without NIXL configured. Please use the constructor argument directly or move this after `super()`, and add a constructor regression test.

2. **[P1] Producer endpoints need to remain request-scoped.** [`register_chunk_recv`](https://github.com/vllm-project/vllm-omni/blob/dd63e8d78f32b300d18b26a6346244632f339164/vllm_omni/worker/omni_connector_model_runner_mixin.py#L1094-L1100) receives endpoint data selected from `get_bound_client(request_id)`, but `update_sender_info()` overwrites one connector-wide host/port. With concurrent requests from different producer replicas, registering B redirects A's later metadata query to B. Please retain the endpoint per request and pass it to `get()` as partial source metadata.

3. **[P1] The async-chunk adapter drops that endpoint entirely.** [`_poll_single_request`](https://github.com/vllm-project/vllm-omni/blob/dd63e8d78f32b300d18b26a6346244632f339164/vllm_omni/distributed/omni_connectors/transfer_adapter/chunk_transfer_adapter.py#L444-L450) invokes `get()` without `entry.request.payload_sender_info`. With the documented `host: auto` receiver config, this can attempt `tcp://auto:<port>`; it also cannot route multi-replica requests. Please pass `{source_host, source_port}` from the request.

4. **[P1] The NIXL agent is used concurrently without enabling NIXL synchronization.** The [`nixl_agent_config`](https://github.com/vllm-project/vllm-omni/blob/dd63e8d78f32b300d18b26a6346244632f339164/vllm_omni/distributed/omni_connectors/connectors/nixl_connector.py#L103-L108) leaves `enable_listen_thread=False` and `sync_mode=None`; NIXL 1.3 resolves that to `NIXL_THREAD_SYNC_NONE`. The connector nevertheless calls the same native agent from put/get, lease, handshake, and transfer-reaper threads. The Python state lock does not cover those native calls. Please configure STRICT/RW synchronization or serialize all agent operations. See NIXL's [configuration default](https://github.com/ai-dynamo/nixl/blob/5949ccf9f55ad60f6d7f79c28200bde152cbb203/src/api/python/_api.py#L149-L216) and [`SYNC_NONE` implementation](https://github.com/ai-dynamo/nixl/blob/5949ccf9f55ad60f6d7f79c28200bde152cbb203/src/core/sync.h#L25-L53).

5. **[P1] Non-timeout failures can still release resources owned by active DMA.** All descriptor-group transfers are submitted before the waits, but [`get()` only defers resources for `TimeoutError`](https://github.com/vllm-project/vllm-omni/blob/dd63e8d78f32b300d18b26a6346244632f339164/vllm_omni/distributed/omni_connectors/connectors/nixl_connector.py#L354-L372). If one handle reports `ERR` while a sibling remains `PROC`, or later group setup fails after an earlier submission, `finally` deregisters destination buffers/dlists while DMA may remain active. Please defer owned resources whenever any submitted handle is nonterminal.

6. **[P1] Metadata lookup blocks the shared receive loop.** [`_query_metadata_at`](https://github.com/vllm-project/vllm-omni/blob/dd63e8d78f32b300d18b26a6346244632f339164/vllm_omni/distributed/omni_connectors/connectors/nixl_connector.py#L542-L563) retries for 60 seconds by default, while both consumers use one background thread to iterate pending requests and describe each poll as non-blocking. One late/abandoned key can therefore starve every ready request behind it. A single short query should return `None` and let the outer receive loop retry fairly.

7. **[P1] Normal `OmniPayloadStruct` chunks bypass tensor extraction.** [`_contains_tensor`](https://github.com/vllm-project/vllm-omni/blob/dd63e8d78f32b300d18b26a6346244632f339164/vllm_omni/distributed/omni_connectors/connectors/nixl_connector.py#L813-L821) only traverses dict/list/tuple, but the async adapter sends a `msgspec.Struct`. These payloads take the OBJECT path: tensor fields are copied through CPU serialization, transferred as one DRAM blob, and reconstructed as CPU tensors in a dict. Please normalize supported structs to mappings before scanning so tensor leaves use the accelerator-native path and preserve device semantics.

8. **[P2] Existing connector-spec tests are now red.** The unconditional [`from_stage`/`to_stage` injection](https://github.com/vllm-project/vllm-omni/blob/dd63e8d78f32b300d18b26a6346244632f339164/vllm_omni/engine/stage_init_utils.py#L1783-L1786) changes the `SharedMemoryConnector` config contract. The two parameterizations of `test_outgoing_stage_uses_sender_connector_spec` and `test_full_payload_consumer_uses_receiver_connector_spec` fail. If only transfer-engine connectors need these fields, please gate the injection; otherwise update the contract/tests here.

Validation on this head:

- Focused NIXL/coordinator/worker tests: **109 passed**.
- Existing targeted adapter-flow tests: **3 failed** as described above.
- `git diff --check`: passed.
- A real NIXL transfer and representative E2E remain untested. Given the native resource-lifetime paths, I recommend at least an intra-node real-NIXL smoke test before approval.

Scope assumption: I treated ordinary payload, async-chunk, and multi-replica paths as supported because this PR explicitly changes/tests them. If the intended initial scope is only the RFC's two-stage direct-metadata consumer, the generic paths should be guarded or clearly documented as unsupported; findings 1, 4, and 5 still apply either way.

vllm_omni/engine/stage_engine_core_client.py
```diff
@@ -154,6 +154,7 @@ def __init__(
         self._kv_sender_host = self._resolve_contact_host()
         self._kv_sender_info: dict[str, Any] | None = None
         self._kv_sender_initialized = False
+        self._payload_sender_info = self._build_payload_sender_info()
```
david6666666: [P0] This invokes _build_payload_sender_info before MPClient.__init__ assigns self.vllm_config, and the helper dereferences that attribute immediately. I reproduced AttributeError on the current head; every StageEngineCoreClient and DPLBStageEngineCoreClient construction takes this path, even when NIXL is not configured. Please use the vllm_config argument directly or move this initialization after super(), and add a constructor regression test.

vllm_omni/worker/omni_connector_model_runner_mixin.py
```diff
@@ -1092,6 +1091,13 @@ def register_chunk_recv(self, request: Any) -> None:
         # across requests.
         ext = getattr(request, "external_req_id", None)
         self._request_ids_mapping[request_id] = ext if ext is not None else request_id
+        sender_info = getattr(request, "payload_sender_info", None)
+        if isinstance(sender_info, dict) and self._omni_connector is not None:
+            host = sender_info.get("host")
+            port = sender_info.get("zmq_port")
+            update_sender_info = getattr(self._omni_connector, "update_sender_info", None)
+            if host and port and callable(update_sender_info):
+                update_sender_info(str(host), int(port))
```
david6666666: [P1] The orchestrator chooses payload_sender_info from get_bound_client(request_id), so simultaneous requests can originate from different producer replicas. update_sender_info stores only one connector-wide endpoint; registering request B therefore redirects request A's later metadata lookup to B. Please retain the endpoint per request and pass it to get() as partial source metadata.

vllm_omni/distributed/omni_connectors/transfer_adapter/chunk_transfer_adapter.py
```diff
@@ -254,11 +256,20 @@ def create_connector(cls, model_config: Any):
...
+        role = connector_spec.extra.get("role")
+        resolved_spec = resolve_connector_spec(
```
david6666666, line 291: [P1] Resolving this static receiver spec is insufficient for async-chunk routing. load_async retains request.payload_sender_info, but _poll_single_request later calls connector.get without passing it or updating the connector. With the documented host: auto configuration, the receiver can try tcp://auto:<port>; multi-replica requests also cannot select their bound producer. Please thread the request endpoint through _LoadEntry into get() as source metadata.

vllm_omni/distributed/omni_connectors/connectors/nixl_connector.py
```diff
@@ -0,0 +1,909 @@
...
+        self._backends = list(backends) if isinstance(backends, (list, tuple)) else [str(backends)]
+        if nixl_agent_config is None:
+            agent_config = None
+        else:
+            non_ucx_backends = [backend for backend in self._backends if backend != "UCX"]
+            if non_ucx_backends:
+                agent_config = nixl_agent_config(backends=self._backends, capture_telemetry=False)
+            else:
+                num_threads = int(self.config.get("num_threads", 4))
+                agent_config = nixl_agent_config(num_threads=num_threads, capture_telemetry=False)
```
david6666666: [P1] nixl_agent_config leaves enable_listen_thread false and sync_mode unset, which NIXL 1.3 resolves to NIXL_THREAD_SYNC_NONE. This connector nevertheless invokes the same native agent from put/get, handshake/lease cleanup, and the transfer reaper. The Python state lock does not cover those calls. Please select STRICT/RW synchronization or serialize all agent operations with a dedicated lock.

vllm_omni/distributed/omni_connectors/connectors/nixl_connector.py
```diff
@@ -0,0 +1,909 @@
...
+                        remote_dlist,
+                        desc_ids,
+                    )
+                    xfer_handles.append(xfer_handle)
+                    self._agent.transfer(xfer_handle)
+
+            for xfer_handle in xfer_handles:
+                try:
+                    self._wait_for_transfer(xfer_handle, get_key)
+                except TimeoutError:
```
david6666666: [P1] All descriptor-group transfers are submitted before these waits, but resources are deferred only for TimeoutError. If one handle reports ERR while a sibling remains PROC, or later group setup fails after an earlier submission, the general exception path reaches finally and deregisters buffers/dlists that active DMA may still own. Please defer the complete ownership bundle whenever any submitted handle is nonterminal, not only on timeout.

vllm_omni/distributed/omni_connectors/connectors/nixl_connector.py
```diff
@@ -0,0 +1,909 @@
...
+    def _query_metadata_at(self, get_key: str, host: str, port: int) -> dict[str, Any] | None:
+        """Fetch transfer metadata for ``get_key`` from a producer's ROUTER socket.
+
+        A consumer can outrun its producer, so a missing key is retried until
+        ``handshake_max_wait_s`` instead of failing the request immediately.
+        """
+        zmq_addr = f"tcp://{host}:{port}"
+        deadline = time.monotonic() + self._handshake_max_wait_s
+        request = _GET_META_MSG + msgspec.msgpack.encode({"key": get_key})
+        while not self._stop_event.is_set():
```
david6666666: [P1] The connector consumers use one background thread to iterate pending requests and explicitly describe each poll as non-blocking, but this loop can hold that thread for handshake_max_wait_s (60 seconds by default). One late or abandoned key therefore starves ready requests behind it. Please perform one short metadata query and return None so the outer receive loop can retry requests fairly.

vllm_omni/distributed/omni_connectors/connectors/nixl_connector.py
```diff
@@ -0,0 +1,909 @@
...
+    @staticmethod
+    def _is_tensor_payload(payload: Any) -> bool:
+        return isinstance(payload, torch.Tensor) or (
+            isinstance(payload, (list, tuple))
+            and bool(payload)
+            and all(isinstance(item, torch.Tensor) for item in payload)
+        )
+
+    @classmethod
+    def _contains_tensor(cls, payload: Any) -> bool:
```
david6666666, line 945: [P1] Normal async payloads are OmniPayloadStruct instances (msgspec.Struct), but this traversal only recognizes dict/list/tuple. Those payloads consequently take the OBJECT path: their tensors are serialized through CPU, transferred as one DRAM blob, and reconstructed as CPU tensors in a dict. Please normalize supported structs to mappings before scanning so tensor leaves use the direct accelerator path and preserve device semantics.

vllm_omni/engine/stage_init_utils.py
```diff
@@ -1778,10 +1778,12 @@ def get_stage_connector_spec(
     # sender-only so the scheduler does not park orchestrator-provided inputs
     # waiting for an upstream payload.
     target_stage = str(stage_id)
-    for (from_stage, _to_stage), spec in getattr(omni_transfer_config, "connectors", {}).items():
+    for (from_stage, to_stage), spec in getattr(omni_transfer_config, "connectors", {}).items():
         if from_stage == target_stage:
             extra = dict(spec.extra or {})
             extra.setdefault("role", "sender")
+            extra["from_stage"] = int(from_stage)
```
david6666666: [P2] These routing fields are injected for every connector, changing the established SharedMemoryConnector spec contract. On the current head, both parameterizations of test_outgoing_stage_uses_sender_connector_spec and test_full_payload_consumer_uses_receiver_connector_spec fail because of these extra keys. If only transfer-engine connectors need them, gate the injection by connector type; otherwise update the contract and tests in this PR.

david6666666 (changes requested):
Requesting changes on the current head (5c3d75f). The inline comments identify correctness, routing, native-resource-lifetime, and regression issues that should be resolved before merge.

Please also update the PR description with test commands and results from a real NIXL installation, scoped to the connector and generic stage-transport behavior introduced here:

- A native two-process sender/receiver transfer without the fake NIXL agent, covering the ZMQ metadata handshake and XFER_DONE cleanup.
- Tensor and structured payloads, including an actual OmniPayloadStruct and a mixed CPU/accelerator payload; verify values, dtype, shape, destination device, and released producer registrations.
- TP=1 and TP>1 rank-aware endpoints. If multi-replica routing remains claimed, include concurrent requests bound to different producer replicas.
- Timeout/error cleanup with a real agent, demonstrating that no memory, descriptors, or handles are released while a transfer is still active.
- An inter-node UCX transfer if cross-node operation remains a supported claim; otherwise narrow the documented scope.
- The affected local unit suites, including tests/distributed/omni_connectors/test_adapter_and_flow.py, which currently has three failures on this head.

A full MiniMax-H3 or other model E2E is outside the required scope for this connector-only PR. A minimal ordinary stage-payload integration test is sufficient, but stub-only coverage does not validate the core NIXL data plane.

vllm_omni/engine/stage_engine_core_client.py
yuanwu2017: Fixed in 07fa639d5. `_payload_sender_info` is now built only after `super().__init__()` has initialized the base client state. Added a constructor regression covering the no-NIXL/default path as well as sender-info construction.

vllm_omni/worker/omni_connector_model_runner_mixin.py
yuanwu2017: Fixed in 07fa639d5. The receive registration now stores `payload_sender_info` per request instead of mutating connector-global routing state, and the request endpoint is passed into `connector.get()` as `source_host`/`source_port`. Regression coverage interleaves requests from distinct producer endpoints and verifies each lookup keeps its own route.

vllm_omni/distributed/omni_connectors/transfer_adapter/chunk_transfer_adapter.py
yuanwu2017: Fixed in 07fa639d5. `_LoadEntry` retains the request-scoped producer metadata, and `_poll_single_request` forwards it to `get()` as `source_host`/`source_port`, so `host: auto` and multi-replica async-chunk requests use the bound producer rather than a static endpoint. The adapter regression verifies the forwarded endpoint.

vllm_omni/distributed/omni_connectors/connectors/nixl_connector.py
yuanwu2017: Fixed in 07fa639d5. Native agent construction now explicitly selects STRICT synchronization. The resolver uses the enum associated with the installed config API, which preserves compatibility with the NIXL 1.3 module split observed in the real installation. Unit coverage asserts STRICT mode, and the native two-process UCX smoke passed with this configuration.

vllm_omni/distributed/omni_connectors/connectors/nixl_connector.py
yuanwu2017: Fixed in 07fa639d5. Error handling now inspects every submitted handle and transfers the complete ownership bundle to deferred cleanup whenever any handle is nonterminal, including setup failures after earlier submissions and ERR/PROC sibling combinations. Focused lifecycle regressions verify buffers, registrations, descriptor lists, handles, and remote-agent state remain owned until all native work reaches a terminal state.

vllm_omni/distributed/omni_connectors/connectors/nixl_connector.py
yuanwu2017: Fixed in 07fa639d5. `_query_metadata_at` now performs one bounded metadata exchange and returns `None` when metadata is not ready; retry scheduling remains in the outer receive loop. The fairness regression places a missing key before a ready key and verifies the ready request is not blocked by the handshake timeout.

vllm_omni/distributed/omni_connectors/connectors/nixl_connector.py
yuanwu2017: Fixed in 07fa639d5. Supported `msgspec.Struct` payloads, including `OmniPayloadStruct`, are normalized before tensor discovery and skeleton extraction, so tensor leaves use descriptor-group transfers and preserve dtype, shape, value, and destination device. Both unit coverage and the real two-process GPU0-to-GPU1 UCX test exercise an actual `OmniPayloadStruct` plus mixed CPU/GPU leaves.

vllm_omni/engine/stage_init_utils.py
yuanwu2017: Fixed in 07fa639d5. `from_stage`/`to_stage` injection is now gated to transfer-engine connectors, preserving the established `SharedMemoryConnector` spec contract. The previously failing adapter-flow parameterizations now pass as part of the focused 235-test run.

yuanwu2017:
Addressed the latest review in `07fa639d5` and replied on each inline thread.

Validation on the updated head:

- Affected unit/transport suites: **235 passed, 3 subtests passed**.
- Native smoke: **1 passed** using independent producer and consumer processes with the real NIXL 1.3 UCX backend.
- Transfer path: CUDA GPU 0 to GPU 1 via consumer-initiated NIXL `READ`.
- Payload: an actual `OmniPayloadStruct` with GPU and CPU tensor leaves; values, dtype, shape, and destination device were verified after transfer.
- Control/lifetime: exercised ZMQ `GET_META` and `XFER_DONE`; producer pending payloads and registrations both reached zero after completion.
- Routing regressions cover TP=1/TP>1 rank-aware endpoints and concurrent request-scoped producer replicas.
- Lifecycle regressions cover timeout and non-timeout/error cases, retaining the full ownership bundle whenever any submitted native handle remains nonterminal.
- The three previously failing adapter-flow tests now pass.

Native command:

```bash
UCX_RCACHE_MAX_UNRELEASED=1024 pytest -o addopts="" -q -s tests/distributed/omni_connectors/test_nixl_connector_native.py
```

Scope note: this validates same-host cross-GPU UCX. **Inter-node UCX and Intel XPU were not validated**, and the PR description now says so explicitly.

yuanwu2017:
> I re-reviewed the current head (`dd63e8d7`). The earlier feedback fixes are mostly present, but I still think this needs changes before merge:
> 
> 1. **[P0] Stage client construction currently fails universally.** [`stage_engine_core_client.py:156`](https://github.com/vllm-project/vllm-omni/blob/dd63e8d78f32b300d18b26a6346244632f339164/vllm_omni/engine/stage_engine_core_client.py#L156) calls `_build_payload_sender_info()` before `super().__init__()`, while that helper reads `self.vllm_config` and vLLM's `MPClient.__init__` is what assigns it. This raises `AttributeError` for every `StageEngineCoreClient`/`DPLBStageEngineCoreClient`, even without NIXL configured. Please use the constructor argument directly or move this after `super()`, and add a constructor regression test.
> 2. **[P1] Producer endpoints need to remain request-scoped.** [`register_chunk_recv`](https://github.com/vllm-project/vllm-omni/blob/dd63e8d78f32b300d18b26a6346244632f339164/vllm_omni/worker/omni_connector_model_runner_mixin.py#L1094-L1100) receives endpoint data selected from `get_bound_client(request_id)`, but `update_sender_info()` overwrites one connector-wide host/port. With concurrent requests from different producer replicas, registering B redirects A's later metadata query to B. Please retain the endpoint per request and pass it to `get()` as partial source metadata.
> 3. **[P1] The async-chunk adapter drops that endpoint entirely.** [`_poll_single_request`](https://github.com/vllm-project/vllm-omni/blob/dd63e8d78f32b300d18b26a6346244632f339164/vllm_omni/distributed/omni_connectors/transfer_adapter/chunk_transfer_adapter.py#L444-L450) invokes `get()` without `entry.request.payload_sender_info`. With the documented `host: auto` receiver config, this can attempt `tcp://auto:<port>`; it also cannot route multi-replica requests. Please pass `{source_host, source_port}` from the request.
> 4. **[P1] The NIXL agent is used concurrently without enabling NIXL synchronization.** The [`nixl_agent_config`](https://github.com/vllm-project/vllm-omni/blob/dd63e8d78f32b300d18b26a6346244632f339164/vllm_omni/distributed/omni_connectors/connectors/nixl_connector.py#L103-L108) leaves `enable_listen_thread=False` and `sync_mode=None`; NIXL 1.3 resolves that to `NIXL_THREAD_SYNC_NONE`. The connector nevertheless calls the same native agent from put/get, lease, handshake, and transfer-reaper threads. The Python state lock does not cover those native calls. Please configure STRICT/RW synchronization or serialize all agent operations. See NIXL's [configuration default](https://github.com/ai-dynamo/nixl/blob/5949ccf9f55ad60f6d7f79c28200bde152cbb203/src/api/python/_api.py#L149-L216) and [`SYNC_NONE` implementation](https://github.com/ai-dynamo/nixl/blob/5949ccf9f55ad60f6d7f79c28200bde152cbb203/src/core/sync.h#L25-L53).
> 5. **[P1] Non-timeout failures can still release resources owned by active DMA.** All descriptor-group transfers are submitted before the waits, but [`get()` only defers resources for `TimeoutError`](https://github.com/vllm-project/vllm-omni/blob/dd63e8d78f32b300d18b26a6346244632f339164/vllm_omni/distributed/omni_connectors/connectors/nixl_connector.py#L354-L372). If one handle reports `ERR` while a sibling remains `PROC`, or later group setup fails after an earlier submission, `finally` deregisters destination buffers/dlists while DMA may remain active. Please defer owned resources whenever any submitted handle is nonterminal.
> 6. **[P1] Metadata lookup blocks the shared receive loop.** [`_query_metadata_at`](https://github.com/vllm-project/vllm-omni/blob/dd63e8d78f32b300d18b26a6346244632f339164/vllm_omni/distributed/omni_connectors/connectors/nixl_connector.py#L542-L563) retries for 60 seconds by default, while both consumers use one background thread to iterate pending requests and describe each poll as non-blocking. One late/abandoned key can therefore starve every ready request behind it. A single short query should return `None` and let the outer receive loop retry fairly.
> 7. **[P1] Normal `OmniPayloadStruct` chunks bypass tensor extraction.** [`_contains_tensor`](https://github.com/vllm-project/vllm-omni/blob/dd63e8d78f32b300d18b26a6346244632f339164/vllm_omni/distributed/omni_connectors/connectors/nixl_connector.py#L813-L821) only traverses dict/list/tuple, but the async adapter sends a `msgspec.Struct`. These payloads take the OBJECT path: tensor fields are copied through CPU serialization, transferred as one DRAM blob, and reconstructed as CPU tensors in a dict. Please normalize supported structs to mappings before scanning so tensor leaves use the accelerator-native path and preserve device semantics.
> 8. **[P2] Existing connector-spec tests are now red.** The unconditional [`from_stage`/`to_stage` injection](https://github.com/vllm-project/vllm-omni/blob/dd63e8d78f32b300d18b26a6346244632f339164/vllm_omni/engine/stage_init_utils.py#L1783-L1786) changes the `SharedMemoryConnector` config contract. The two parameterizations of `test_outgoing_stage_uses_sender_connector_spec` and `test_full_payload_consumer_uses_receiver_connector_spec` fail. If only transfer-engine connectors need these fields, please gate the injection; otherwise update the contract/tests here.
> 
> Validation on this head:
> 
> * Focused NIXL/coordinator/worker tests: **109 passed**.
> * Existing targeted adapter-flow tests: **3 failed** as described above.
> * `git diff --check`: passed.
> * A real NIXL transfer and representative E2E remain untested. Given the native resource-lifetime paths, I recommend at least an intra-node real-NIXL smoke test before approval.
> 
> Scope assumption: I treated ordinary payload, async-chunk, and multi-replica paths as supported because this PR explicitly changes/tests them. If the intended initial scope is only the RFC's two-stage direct-metadata consumer, the generic paths should be guarded or clearly documented as unsupported; findings 1, 4, and 5 still apply either way.

 All eight findings were addressed in https://github.com/vllm-project/vllm-omni/commit/07fa639d5d9af6384ea1b9bff5ca6ec2f9304d6e

yuanwu2017:
@hsliuustc0106 @david6666666 @xuechendi I have resolved the comments. Please help to review.

yuanwu2017:
## H3 NIXL E2E retest after main integration (2026-09-09)

The successful test used the tree now published as **0a54596420d90ed1fa15946b769734485c2e71b9** in #6264, stacked on #6093 head **7da3f90a2efc14350af1641a6e1c2b4fdba79146**. The run used `8702c8009` plus the five-file configuration fix subsequently committed as `0a5459642`; the committed diff was compared byte-for-byte with the tested patch. This is integration E2E evidence for the connector **together with the downstream stage-payload feature**, not a claim that standalone #6093 provides the H3 payload integration.

### Results

- Six same-host A100 80GB GPUs; text encoder **TP2**, diffusion **TP4 / Ulysses1 / VAE patch parallel4**.
- Directly loaded the repository's `vllm_omni/deploy/minimax_h3_disaggregated.yaml`, unchanged; only Stage 1 parallel settings were overridden through `--stage-overrides`. This does **not** validate the YAML's unmodified TP1/Ulysses4 defaults.
- Original orange-tabby request: 8 steps, seed42, 1344x768, 107 frames, 24 FPS.
- **HTTP 200**, `video/mp4`, **5,108,715 bytes**, **78.350852 seconds** request time.
- H.264 video + AAC audio; encoded duration **4.458333 seconds**. Full `ffmpeg -xerror` decode and preview extraction passed.
- One matching NIXL put/get, **358775 bytes**, key `video_sync-8be48ef374af9f47-8fca2926_0_0`; producer `Worker_TP0`, consumer `DiffusionWorker_TP0`.
- Video SHA256: `68b69e1e7d8de6bc3e6c1c7c143f39ed8cddf16f3fc53d32227b7767bcacb34c`.
- Expanded regressions on the fixed tree: **1278 passed, 6 skipped, 3 subtests passed**. Ruff lint/format and `git diff --check` passed.
- Separate native NIXL smoke on pre-config-fix `8702c8009`: **1 passed** in 31.23s. The subsequent five-file fix did not change the native connector implementation.
- All six test GPUs were released after the run.

An earlier attempt failed during text-encoder loading because an external managed model was automatically restarted onto GPU0. The successful retry above ran after that service was stopped. That failed attempt is not counted as a successful E2E.

### Update / author self-review

#6264's nine feature commits were restacked onto the current #6093 head, retaining the upstream H3 conditioning contract and native paged-KV request plumbing. The final configuration fix preserves stage payload keys through the centralized diffusion projection, serialization, and runtime mapping, with topology defaults and explicit deploy/CLI precedence (including empty overrides). It also registers/documents the two existing NIXL environment variables and updates their inventory tests. These five-file fixes are in #6264; this comment does not claim they were separately added to #6093. The native paged-KV and full-payload protocols remain distinct.

### Reproduction commands

Run from a checkout of **#6264 at `0a5459642`**, not from standalone #6093. Runtime used: local image `vllm-omni-pr6093:integration-20260908` (vLLM 0.28.0 / Python 3.12, with native NIXL/UCX installed). This image tag is a local test artifact, **not a published pullable image**; use an equivalently provisioned runtime if reproducing elsewhere. Set `HF_CACHE` to the Hugging Face cache containing the pinned H3 snapshot. Physical GPUs `0,1,3,4,5,6` must be free; they map to container logical GPUs `0..5`.

```bash
export HF_CACHE=/path/to/huggingface/cache
export IMAGE=vllm-omni-pr6093:integration-20260908
mkdir -p h3-e2e-output

docker run -d --name h3-nixl-e2e \
  --gpus '"device=0,1,3,4,5,6"' --network host --ipc host \
  -v "$PWD:/workspace:ro" -v "$PWD/h3-e2e-output:/artifacts" \
  -v "$HF_CACHE:/root/.cache/huggingface:ro" -w /workspace \
  -e PYTHONPATH=/workspace -e PYTHONDONTWRITEBYTECODE=1 \
  -e CUDA_VISIBLE_DEVICES=0,1,2,3,4,5 \
  -e HF_HOME=/root/.cache/huggingface -e VLLM_LOGGING_LEVEL=DEBUG \
  -e UCX_RCACHE_MAX_UNRELEASED=1024 \
  "$IMAGE" python -m vllm_omni.entrypoints.cli.main serve \
  /root/.cache/huggingface/hub/models--MiniMaxAI--MiniMax-H3/snapshots/42ed227ee7df40d41602854ae760620d6eb651fe/FL2VA \
  --omni --host 127.0.0.1 --port 8010 --disable-log-stats \
  --deploy-config vllm_omni/deploy/minimax_h3_disaggregated.yaml \
  --stage-overrides '{"1":{"tensor_parallel_size":4,"ulysses_degree":1}}'

trap 'docker stop --time 30 h3-nixl-e2e' EXIT
curl --noproxy '*' --fail --silent --show-error \
  --retry 180 --retry-connrefused --retry-delay 10 \
  --retry-max-time 1800 --max-time 5 http://127.0.0.1:8010/health

curl --noproxy '*' --fail --silent --show-error --max-time 1800 \
  -D h3-e2e-output/orange-cat.headers -o h3-e2e-output/orange-cat.mp4 \
  -w 'http_code=%{http_code}\ncontent_type=%{content_type}\nsize_download=%{size_download}\ntime_total=%{time_total}\n' \
  http://127.0.0.1:8010/v1/videos/sync \
  -F 'prompt=A clearly visible orange tabby cat walking through bright green grass in warm daylight, detailed fur, natural motion, vibrant colors, wide cinematic shot, with matching outdoor ambient sound.' \
  -F 'width=1344' -F 'height=768' -F 'num_frames=107' -F 'fps=24' \
  -F 'num_inference_steps=8' -F 'seed=42' \
  -F 'extra_params={"task":"t2va","duration":4.0,"aspect_ratio":"16:9","flow_shift":12.0,"audio_flow_shift":3.0}'

docker exec h3-nixl-e2e ffprobe -v error \
  -show_entries format=duration,size \
  -show_entries stream=codec_type,codec_name,width,height,nb_frames,r_frame_rate \
  -of json /artifacts/orange-cat.mp4
docker exec h3-nixl-e2e ffmpeg -v error -xerror \
  -i /artifacts/orange-cat.mp4 -f null -
docker exec h3-nixl-e2e ffmpeg -v error -ss 2 \
  -i /artifacts/orange-cat.mp4 -frames:v 1 /artifacts/preview.png
docker logs h3-nixl-e2e > h3-e2e-output/server.log 2>&1
grep -E 'NixlConnector (put|get).*video_sync' h3-e2e-output/server.log
sha256sum h3-e2e-output/orange-cat.mp4
docker stop --time 30 h3-nixl-e2e
trap - EXIT
```

The test host additionally passed its HTTP(S) proxy through to the container; configure your own proxy if required. No custom deployment YAML is needed.

Separate native smoke command (run when its two GPUs are free):

```bash
docker run --rm --gpus '"device=1,3"' --network host --ipc host \
  -v "$PWD:/workspace:ro" -w /workspace \
  -e PYTHONPATH=/workspace -e PYTHONDONTWRITEBYTECODE=1 \
  -e UCX_RCACHE_MAX_UNRELEASED=1024 \
  "$IMAGE" python -m pytest -vv -s -rA -o addopts='' -p no:cacheprovider \
  tests/distributed/omni_connectors/test_nixl_connector_native.py
```

**Scope:** real same-host CUDA/NIXL transfer with heterogeneous producer/consumer TP. Inter-node transfer, Intel XPU, and real Mooncake RDMA E2E were **not** validated. Mooncake/paged-KV coexistence is covered by regression tests, not by this H3 media run. H3 exercises an AR producer and diffusion consumer; the separate diffusion-producer/inline-drop branch is unit-tested, not exercised by this two-stage E2E.

### Expanded regression command

Run separately in the same runtime, without GPU exposure. Redirecting the CI fixture helper's generated directory to `/tmp` is necessary with the read-only source mount; otherwise fixture generation fails for environmental reasons.

```bash
docker run --rm \
  -v "$PWD:/workspace:ro" -w /workspace \
  -e PYTHONPATH=/workspace -e PYTHONDONTWRITEBYTECODE=1 \
  -e CUDA_VISIBLE_DEVICES= \
  "$IMAGE" python -c '
import sys
from pathlib import Path
import tests.helpers.stage_config as stage_config
import pytest
stage_config._CI_GENERATED_DIR = Path("/tmp/generated-ci")
raise SystemExit(pytest.main(sys.argv[1:]))
' -q -o addopts='' -p no:cacheprovider \
  tests/distributed/omni_connectors \
  --ignore=tests/distributed/omni_connectors/test_nixl_connector_native.py \
  --ignore=tests/distributed/omni_connectors/test_mooncake_transfer_engine_rdma.py \
  --ignore=tests/distributed/omni_connectors/test_bagel_mooncake_connector.py \
  --ignore=tests/distributed/omni_connectors/test_bagel_shared_memory_connector.py \
  --ignore=tests/distributed/omni_connectors/test_mooncake_tcp_completion.py \
  tests/engine/test_orchestrator_kv_sender_info.py \
  tests/worker/test_omni_connector_mixin.py \
  tests/diffusion/test_diffusion_stage_payload.py \
  tests/diffusion/test_inline_stage_diffusion_client.py \
  tests/diffusion/test_stage_payload_kv_coexistence.py \
  tests/diffusion/test_diffusion_request.py \
  tests/diffusion/test_diffusion_config_fields.py \
  tests/diffusion/diffusion_kv \
  tests/model_executor/stage_input_processors/test_minimax_h3.py \
  tests/config
```

Observed result on the fixed tree: `1278 passed, 6 skipped, 2 warnings, 3 subtests passed in 95.55s`. Apply your proxy settings to the container where model-config downloads are required. The explicitly excluded native/hardware suites are not included in this passing count.

natureofnature:
Review of `7da3f90a2efc14350af1641a6e1c2b4fdba79146`: I found three correctness/lifecycle issues and one configuration regression that should be addressed before merging.

### P1: Intermediate stages cannot publish metadata for their outgoing payloads

For a chain such as Thinker -> Talker -> Code2Wav, `get_stage_connector_spec()` selects the intermediate stage's incoming-edge configuration with `role="receiver"`. The resolver removes its local `zmq_port`, and [`_init_handshake()`](https://github.com/vllm-project/vllm-omni/blob/7da3f90a2efc14350af1641a6e1c2b4fdba79146/vllm_omni/distributed/omni_connectors/connectors/nixl_connector.py#L195) does not start a metadata listener. However, the same connector also handles outgoing `put()` calls, whose returned transfer metadata is not forwarded by the adapter/mixin send path. The next stage therefore has no way to retrieve those descriptors.

Please preserve both incoming and outgoing endpoint responsibilities for intermediate stages and add a three-stage regression test. Simply changing the role to sender would not preserve the incoming-edge contract.

### P1: Async-chunk prewarming bypasses producer endpoint propagation

The new `payload_sender_info` assignment is wired into [ordinary forwarding](https://github.com/vllm-project/vllm-omni/blob/7da3f90a2efc14350af1641a6e1c2b4fdba79146/vllm_omni/engine/orchestrator.py#L2835), but [`_prewarm_async_chunk_stages()`](https://github.com/vllm-project/vllm-omni/blob/7da3f90a2efc14350af1641a6e1c2b4fdba79146/vllm_omni/engine/orchestrator.py#L2967) constructs requests separately and leaves that field unset. The adapter/mixin receive path consequently falls back to its configured endpoint. When producer and consumer replica IDs differ, the receiver-derived port can point to the wrong producer replica, leaving the request waiting for data.

Please attach the bound upstream producer's endpoint during prewarming and test the actual submitted request with unequal producer/consumer replica IDs. Testing an adapter request with manually populated sender info does not cover this entry point.

### P1: Producer lease expiry can release memory during an active remote READ

`_handle_handshake_message()` returns descriptors without pinning or renewing the source allocation. [`_cleanup_expired_pending()`](https://github.com/vllm-project/vllm-omni/blob/7da3f90a2efc14350af1641a6e1c2b4fdba79146/vllm_omni/distributed/omni_connectors/connectors/nixl_connector.py#L687) then removes and deregisters it at the original deadline, regardless of whether the consumer has completed its READ. A consumer obtaining metadata shortly before expiry can therefore still be reading when reclamation occurs.

The receiver-side deferred-transfer handling protects destination resources, but not this producer allocation. Increasing the TTL reduces the probability without establishing memory-lifetime safety. Please distinguish unclaimed payload expiry from active-transfer ownership and cover expiry while a consumer transfer remains `PROC`.

### P2: Explicit `sender_zmq_port` overrides are discarded

[`resolve_connector_spec()`](https://github.com/vllm-project/vllm-omni/blob/7da3f90a2efc14350af1641a6e1c2b4fdba79146/vllm_omni/distributed/omni_connectors/utils/initialization.py#L109) unconditionally assigns `sender_zmq_port`. For example, a receiver specifying `sender_zmq_port=55000` but no `zmq_port` resolves to `50051` for producer stage 0/rank 0/replica 0. The previous generic initialization used `setdefault`, preserving this override. The shared resolver also affects existing transfer-engine connectors.

Please preserve explicit sender endpoints and derive defaults only for absent fields.

**Verification scope:** I inspected the current call paths and executed four dependency-stubbed control-flow probes using the unchanged method bodies. They confirmed the missing intermediate listener, missing prewarm sender info, source deregistration after metadata delivery without a completion ACK, and overwritten explicit port. I did not run native NIXL/GPU E2E tests or reproduce an actual RDMA corruption event; the probes establish the control-flow gaps, not hardware-level outcomes.

yuanwu2017:
Addressed the new review in 4146ce786a1170f30f19c64b60cf3d8c5833c371, then integrated main aff7d64948f6c0e81e9b3272234623044e215967 in **882026b9c25cc73dcdd25fbb3c1eb121b5450b79**. Both commits are signed off and pushed. Main's connector-mixin facade/split is preserved; the NIXL-specific changes are carried in the split runtime and payload-transport modules.

@natureofnature — the four findings are addressed:

1. **Intermediate stages:** retain the incoming sender endpoint plus a separate outgoing bind configuration/listener. Stage-client advertisement uses that outgoing endpoint (including default ports). Added three-stage configuration/listener regression for ordinary and async settings, plus worker routing tests; this is not a full three-stage model E2E claim.
2. **Async prewarming:** attach the request-bound upstream producer's endpoint to the actual submitted prewarm request. Regression tests use unequal producer/consumer replica IDs and concurrent bindings.
3. **Source ownership:** metadata acquisition atomically installs a generation-scoped claim before exposing descriptors, including the direct-metadata path. Only unclaimed payloads expire; claimed allocations survive expiry, cleanup, and deferred close until matching completion. Duplicate/stale ACKs cannot release another claim. Deterministic fake-agent tests hold `PROC` and cover terminal/error/timeout/unknown states, deferred cleanup and late ACKs. This deliberately favors safety over reclamation: an abandoned/no-ACK claim can retain memory indefinitely; TTL is not native READ cancellation. Current-generation direct metadata still requires the control-plane claim.
4. **Explicit endpoint overrides:** preserve `sender_zmq_port`; compute defaults only when absent, with NIXL and Mooncake regression coverage.

Also adopted @hsliuustc0106's `compute_connector_zmq_port` naming, and retained request-scoped endpoint propagation through both ordinary and async-chunk consumers. Zero-byte tensor leaves keep their shape/dtype semantics without invalid zero-byte DMA registrations.

### Validation on the final integrated tree

- Matching runtime: Python 3.12.3, **vLLM 0.29.0 / Torch 2.13.0+cu130 / NIXL 1.3.2**, derived from `vllm/vllm-openai:v0.29.0` (local test image `vllm-omni-pr6093:v029-merged-v029-20260910T083620Z-attempt3`, SHA256 `2d581fd28985715b4c7f4b53940fbe28a9adcc2f107a3be786256f81da973c50`). Local image is not a publicly pullable artifact.
- Expanded connector + worker-mixin + orchestrator sender-info + config selection: **1046 passed, 2 failed, 4 skipped, 3 subtests passed**. Both failures are the frontend-log-stats config test's `True`/`False` cases on `UnspecifiedOmniPlatform` in a CPU-only container. The same failures reproduce on pinned main in the same image/isolation (**611 passed, 2 failed, 1 skipped** for main's config directory). No new failures within this selection; not an all-suite pass.
- Real two-process/two-GPU native structured transfers: **2 passed**, covering both GET_META and direct metadata, mixed CPU/CUDA leaves, a zero-sized CUDA tensor, source retention across forced expiry/cleanup, and final zero pending registrations. Both runs recorded native states `['PROC', 'DONE']`. The test records rather than requires `PROC`, so this is not a guaranteed hardware timeout/cancellation test; deterministic lifecycle coverage is in the unit tests.
- Ruff lint/format and diff checks passed. Source was mounted read-only; all owned test containers removed. No model E2E or cross-host/RDMA-fabric validation is claimed here. The new #6264 H3 E2E remains pending.

Reproduction selections (inside the matching test environment, from repository root):

```bash
export HTTP_PROXY=http://proxy-ir.intel.com:912 HTTPS_PROXY=http://proxy-ir.intel.com:912
export http_proxy="$HTTP_PROXY" https_proxy="$HTTPS_PROXY"
export NO_PROXY=localhost,127.0.0.1,::1 no_proxy=localhost,127.0.0.1,::1
python3 -m pytest -o addopts='' -p no:cacheprovider -o faulthandler_timeout=120 \
  tests/distributed/omni_connectors \
  --ignore=tests/distributed/omni_connectors/test_nixl_connector_native.py \
  --ignore=tests/distributed/omni_connectors/test_mooncake_transfer_engine_rdma.py \
  --ignore=tests/distributed/omni_connectors/test_bagel_mooncake_connector.py \
  --ignore=tests/distributed/omni_connectors/test_bagel_shared_memory_connector.py \
  --ignore=tests/distributed/omni_connectors/test_mooncake_tcp_completion.py \
  tests/engine/test_orchestrator_kv_sender_info.py \
  tests/worker/test_omni_connector_mixin.py tests/config

# Separate run with two available GPUs:
UCX_RCACHE_MAX_UNRELEASED=1024 python3 -m pytest -o addopts='' -p no:cacheprovider -vv -s -rA \
  tests/distributed/omni_connectors/test_nixl_connector_native.py
```

For the actual read-only-container runs, an external guarded launcher redirects generated CI fixtures to `/tmp/generated-ci`; it does not override platforms or mask test failures. The same upper/lower proxy variables and localhost bypass were passed explicitly to Docker.

vllm_omni/engine/stage_client.py
yuanwu2017: `get_payload_sender_info()` advertises the request-forwarding endpoint of the selected producer StageClient. The orchestrator calls it on the request-bound upstream replica and attaches the result to the next-stage request; the worker/async adapter then queries that producer rather than deriving an endpoint from the consumer replica. This is separate from `get_kv_sender_info()` and its paged-KV endpoint. In 882026b9c, intermediate stages advertise their outgoing listener, including the default port. The sender-info tests cover explicit/default ports and replica-specific advertisement.

vllm_omni/engine/orchestrator.py
yuanwu2017: This parameter is needed for the multi-replica NIXL routing supported here: producer binding is request-specific, so static connector configuration cannot represent two concurrent requests assigned to different producers. `_build_next_stage_request()` carries that selected endpoint into the serialized request without repurposing KV metadata. The updated prewarm path now supplies it too; tests inspect the actual submitted prewarm requests with unequal producer/consumer replica IDs and concurrent bindings (4146ce786, integrated in 882026b9c). It remains optional for paths that do not need endpoint overrides.

vllm_omni/distributed/omni_connectors/utils/initialization.py
yuanwu2017: This is a small shared abstraction over the existing endpoint arithmetic, not a new discovery service. `compute_connector_zmq_port()` computes purpose + replica + TP-rank + source-stage offsets; `resolve_connector_spec()` applies it to a copied spec while preserving explicit sender overrides. KV initialization, worker/adapter creation, and StageClient advertisement use the same calculation, so the advertised endpoint and actual listener cannot drift between those paths. I adopted your suggested compute_ name in 4146ce786 and updated callers/tests. Non-transfer-engine connectors keep their existing configuration behavior.

yuanwu2017: Applied as `compute_connector_zmq_port()` in 4146ce786, with callers and tests updated. Included in published head 882026b9c.

vllm_omni/engine/__init__.py
yuanwu2017: Yes for request-scoped multi-replica routing: this optional request field transports the bound producer endpoint across the engine/scheduler boundary to the receive path. A consumer replica ID is not necessarily the producer replica ID, and storing it in shared connector configuration would cross-route concurrent requests. It is intentionally separate from native paged-KV metadata and contains endpoint information, not payload tensors. Both ordinary forwarding and async prewarm populate it; worker and adapter tests cover per-request endpoint isolation. Published in 882026b9c.

vllm_omni/distributed/omni_connectors/transfer_adapter/chunk_transfer_adapter.py
yuanwu2017: Correct, this is the async_chunk path. That is why it also needs the fix: it constructs its own connector and polls each request entry independently of the ordinary model-runner path. Without shared rank/replica-aware resolution its listener can disagree with the advertised endpoint; without passing the entry-specific sender info it can query the wrong producer. The change does not turn async_chunk into full-payload transfer. Adapter tests cover endpoint forwarding/isolation, and the new orchestrator tests cover actual prewarm submission rather than only manually populated adapter requests (head 882026b9c).

hsliuustc0106:
CI failed

yuanwu2017:
CI follow-up for `e4b93ad476127809d2238287f1d9e26b9cfd6de6`: pushed test-only fixes in `cd8a7b3248a5005b069b1820d621976d4d8db83a`.

### NVIDIA failures addressed
- [Engine/Entrypoints](https://buildkite.com/organizations/vllm/pipelines/vllm-omni/builds/14988/jobs/01a08f0e-bc5a-427f-8d9c-7366c88332b8/download.txt): `test_async_prewarm_skips_outgoing_only_stage` used a pool double missing `get_bound_client`. The double now models submission-time bindings and checks sender endpoint propagation.
- [Other](https://buildkite.com/organizations/vllm/pipelines/vllm-omni/builds/14988/jobs/01a08f0e-bc5a-4bce-bbec-2a6bede4c5d1/download.txt): the three-stage endpoint test assumed the first bounded metadata query succeeded. Discovery tests now use bounded outer retries and verify producer port/generation. Exact ownership tests use synchronous control-message dispatch to avoid lost-reply claim-count ambiguity; real ZMQ discovery/completion coverage remains. Production timeout/ownership behavior is unchanged.

Local CPU-only validation with Python 3.12.3 / vLLM 0.29.0: **401 passed, 1 skipped, 3 subtests passed**; repeated endpoint cases **100 passed**; injected first-query misses **40 passed**. The original one-shot assumption fails under the same injection. Ruff check/format and diff checks passed. This is not a remote CI or native RDMA rerun.

### Remaining accelerator CI incompatibilities
- [Intel 8081](https://buildkite.com/vllm/vllm-omni-intel-ci/builds/8081) explicitly uses `vllm/vllm-openai-xpu:v0.28.0`: 148 collection errors, including 147 direct missing-API/module errors and one separately tracked SenseNova import error.
- [AMD 11683](https://buildkite.com/vllm/vllm-omni-amd-ci/builds/11683) and [NPU 6715](https://buildkite.com/vllm/vllm-omni-npu-ci/builds/6715) also report vLLM 0.28.0 against this 0.29-based Omni tree. Missing APIs include `compute_layout_strides`, `vllm.v1.kv_cache_layout`, and the new entrypoint modules. NPU also reports a MiniCPM processor `prompt=` signature mismatch.

These failures require aligning the accelerator runtime revisions and rerunning CI; this patch does not claim those platforms pass or establish a baseline exemption. The PR diff against its merged main parent does not change the Buildkite/container/dependency configuration. No compatibility shim or broader CI-image change was added to this connector PR.

H3/#6264 E2E remains paused and incomplete; no GPU/native/E2E tests were launched for this fix.

yuanwu2017:
Follow-up: the accelerator version mismatch fixes are already available on newer main. Updated this branch in signed merge 3e746e8c691cc82fc0c4898d013983f1a8098db6 with pinned main 01a2f93256975c7ff9565c5414b0fe0225bc4765, including AMD #7395, Intel/XPU #7441 and NPU #7433. Their hardware Dockerfiles and Intel pipeline match that main revision. No custom compatibility shim was needed.

Resolved the environment-inventory conflict preserving both NIXL controls, and adapted the new upstream terminal-chunk test mock to accept forwarded metadata. Both preceding CI test fixes remain intact.

Local CPU validation: 1,069 passed, 2 skipped, plus 3 passing subtests across combined connector/orchestrator and config/AMD checks. Three additional no-device/offline config failures reproduce with identical node IDs/messages on exact main in the same runtime. Focused inventory/AMD/chunk rerun: 155 passed. Ruff and diff checks passed. No GPU/native/E2E or accelerator-image builds were run locally; updated remote CI must verify the hardware results. H3 E2E remains paused.

yuanwu2017:
## NIXL native E2E validation completed; H3 integration results

**The native two-process NIXL transport E2E matrix passed all 8 cases (236.34 seconds).** These are real native transfers, not the CPU fake-agent tests. Separately, the H3 integration generated valid video/audio for all three tasks, with an outstanding model-request log-evidence check described below.

### Tested stack / attribution

- This PR's tested revision: **3e746e8c691cc82fc0c4898d013983f1a8098db6**.
- Executed in the **#6264 integration**, now published as **7302d4c73ce269749dabd63ed995cb5793bbfbac**, not as an isolated checkout of #6093.
- Exact frozen/published tree: **b99e97a3dafb0e7e9c1bd28d3a487b33bc418d24**.
- Includes main **01a2f93256975c7ff9565c5414b0fe0225bc4765**, #6939 encoder snapshot **10ab47041faa13172e40ccea055278f1cbdbdd78**, and [asukaqaq-s/vllm-omni#1](https://github.com/asukaqaq-s/vllm-omni/pull/1) review fixes **90d24eb108d889791ac935387ce1374b212db85a**. The current #6939 merge head has equivalent dependency-tree content, but its merge commit is not an ancestor of this integration.

### Native transport matrix — PASS

`tests/distributed/omni_connectors/test_nixl_connector_native.py::test_native_two_process_structured_mixed_device_transfer`

| Payload | Queried metadata | Direct metadata |
|---|---|---|
| Generic structured mixed-device payload | PASS | PASS |
| H3 T2VA payload | PASS | PASS |
| H3 FL2VA payload | PASS | PASS |
| H3 Ref2VA payload | PASS | PASS |

Native NIXL/UCX, separate producer and consumer processes, mixed CPU/CUDA structured tensors, including H3 text/visual/audio/layout fields and empty optional tensors. Logs record native transfer states, received payload values/types/shapes, source retention and final cleanup. This is **same-host CUDA** evidence, not cross-node or non-CUDA validation.

The matrix ran in validation `e2e-latest6093-20260913-0606`. CPU **547 passed** and H3 config **21 passed** in that validation. The subsequent `0625` model run explicitly reused those results only after verifying unchanged source/index/runtime; it did not rerun or double-count them.

### Real H3 application generation — media PASS, transport log acceptance incomplete

| Task | HTTP | Request time | MP4 bytes | Media validation |
|---|---:|---:|---:|---|
| T2VA, FL2VA checkpoint | 200 | 76.743 s | 5,044,422 | PASS |
| FL2VA image input, FL2VA checkpoint | 200 | 65.350 s | 4,744,712 | PASS |
| Ref2VA video + embedded AAC, Ref2VA checkpoint | 200 | 200.459 s | 6,106,601 | PASS |

All outputs: H.264 1344×768, 107 frames at 24 fps, stereo 32-kHz AAC; full ffmpeg decode and preview extraction passed. Requests used 8 steps and seed42. Six physical GPUs0,1,3,4,5,6: Stage0 TP2/max_model_len32768; Stage1 TP4/Ulysses1/VAE4. Repository H3 YAML unchanged, with CLI overrides `{"0":{"max_model_len":32768},"1":{"tensor_parallel_size":4,"ulysses_degree":1}}`.

**Important boundary:** the application runner's request-correlated NIXL log check failed for all three requests: producer `put` records exist, but corresponding consumer `get` DEBUG records were absent. Consequently the overall application E2E runner exited **1**, not 0. The missing log cause remains unresolved. Successful media output and the separate native matrix must not be presented as proof that this application-level transport check passed, or as verified absence of fallback.

Runtime: Python3.12.3 / vLLM0.29.0 / Torch2.13cu130 / NIXL1.3.2; immutable image `sha256:bce11ef0dfb05cd4c9e18695bccdd751b7c200514b14ee67fa14bf06b3f981e4`. Model snapshot `42ed227ee7df40d41602854ae760620d6eb651fe`. All owned test containers were cleaned up and six test GPUs released; unrelated GPU2 workloads were untouched.

Detailed #6264 results and output hashes: https://github.com/vllm-project/vllm-omni/pull/6264#issuecomment-5651807676 . Local native logs/XML are retained under `vllm-omni-pr6264-artifacts/e2e-latest6093-20260913-0606/`; model logs, requests and videos under `e2e-latest6093-20260913-0625/`. These local paths are not public artifact attachments.

yuanwu2017:
## Final NIXL receiving acceptance PASS — correction and follow-up

Published fixes:
- **#6264:** `a7ac1bddb8a48dd4835edf9a3671e4dba0964833`, tree `d6b868b42bd4805777a6f4cb728c55bf5524df5f`. All tracked source hashes match the frozen full-model test source.
- **#6093:** `ea34100772a6af18fe20bcd225a1cc5f594d2cf0`, generic backport including the payload-only sender-info prerequisite, without H3-specific test dependencies.

### Root cause: not merely absent DEBUG logs

The earlier 0625 run attempted to bind the receiver to the producer's occupied port. Receiver initialization failed and the application fell back to the inline prompt. Those videos were therefore not application-level NIXL receive evidence. The earlier matcher correctly failed; this supersedes the previous unresolved-log interpretation.

The fix removes shared incoming-edge `zmq_port` when the manager constructs a NIXL receiver. Request-scoped and explicit standalone sender endpoints remain supported. Sender/intermediate-stage local listener behavior is unchanged. No logger workaround or weakened acceptance matcher was used.

### Fresh integration acceptance — all PASS, overall exit 0

Completed 2026-09-13 07:36:01 UTC. **670 CPU passed (1 skipped), 21 H3 config passed, 8 native UCX passed (237.44s)**. Native tests now exercise manager-created receivers against an already-bound producer endpoint and retain exact mixed-device payload/ownership checks.

| Task | HTTP | Seconds | MP4 bytes | Put/get bytes | Media / receive match |
|---|---:|---:|---:|---:|---|
| T2VA | 200 | 76.640 | 5,073,052 | 359,024 / 359,024 | PASS / PASS |
| FL2VA image | 200 | 65.479 | 4,765,076 | 11,158,136 / 11,158,136 | PASS / PASS |
| Ref2VA video + embedded AAC | 200 | 201.440 | 6,110,354 | 64,963,792 / 64,963,792 | PASS / PASS |

Exact keys recorded at both producer and consumer:
- `video_sync-8837c02bfe1c75fb-bdf52d52_0_0`
- `video_sync-b89d4865e4016bb8-a63e69cd_0_0`
- `video_sync-a2bd7e40639d1149-b77a57f2_0_0`

All outputs passed H.264 1344×768, 107 frames/24fps, stereo 32-kHz AAC, full ffmpeg decode and preview checks. Owned containers removed, six test GPUs released, unrelated GPU2 untouched.

### Exact tested stack

Full-model tests used **#6264 integration**, based on #6093 `3e746e8c691cc82fc0c4898d013983f1a8098db6` plus the receiver fix, main `01a2f93256975c7ff9565c5414b0fe0225bc4765`, #6939 snapshot `10ab47041faa13172e40ccea055278f1cbdbdd78`, and [asukaqaq-s/vllm-omni#1](https://github.com/asukaqaq-s/vllm-omni/pull/1) fixes `90d24eb108d889791ac935387ce1374b212db85a`. It was not standalone #6093. The later generic backport commit is not an ancestor of this tested integration.

Standalone #6093 backport validation: **479 CPU passed, 1 skipped, 3 passing subtests; 2/2 generic native UCX cases passed**. These two cases are separate from the integration's eight.

Runtime: Python3.12.3/vLLM0.29.0/Torch2.13cu130/NIXL1.3.2; image `sha256:bce11ef0dfb05cd4c9e18695bccdd751b7c200514b14ee67fa14bf06b3f981e4`. Same-host physical GPUs0,1,3,4,5,6: Stage0 TP2/maxlen32768/util0.6; Stage1 TP4/Ulysses1/VAE4. Stock H3 YAML and CLI overrides `{"0":{"max_model_len":32768},"1":{"tensor_parallel_size":4,"ulysses_degree":1}}`. 8 steps/seed42; model snapshot `42ed227ee7df40d41602854ae760620d6eb651fe`, FL2VA and Ref2VA partitions.

Cross-node, non-CUDA and Turbo checkpoints remain untested; remote CI and approval are separate. #6264's inherited unsigned commit remains unresolved. Local evidence: `vllm-omni-pr6264-artifacts/nixl-receive-final-20260913-071039/` and `pr6093-review-artifacts-20260910/6093receiverfix/` (local paths, not public attachments).

Issue #5244, opened 2026-07-20
[RFC]: vllm-omni multi-stage kvcache transfer manager


asukaqaq-s:
This RFC targets disaggregated AR/Prefill and DiT/Generation deployments for multimodal generative models. It focuses on three issues: 
redundant cross-stage KV cache transfers, GPU bubbles caused by transfer waits, and the lack of unified cache management and cache-aware scheduling on the DiT side. 
The proposed design introduces a unified DiT KV Cache Manager, a cache-aware scheduler, and explicit connector/session lifecycle management. Together, these components enable missing-only transfers, asynchronous transfer-compute overlap, and page-level cache reuse. The expected result is a substantial reduction in KV transfer volume and GPU idle time, improving effective DiT utilization and overall throughput.

co-author：@Acerak01-fy @zwhzzz0821 

## **Background**

- **KV cache in image-generation models:** In MoT/AR-DiT architectures such as Hunyuan and BAGEL, denoising target tokens are typically bidirectionally visible, while context tokens such as text and reference images are visible to the target in only one direction. Under this attention mask, K/V for stable context tokens can be cached.
- **Disaggregated deployment for image-generation models:** The model can be split into a Prefill/AR instance and a DiT Generation instance. Prefill computes and interprets the context, while DiT reuses the resulting prompt KV across multiple diffusion steps.
- **vLLM-Omni:** vLLM-Omni can already transfer AR KV to DiT, but the DiT side does not yet have complete ownership of the target cache. As a result, it cannot yet support target pre-allocation, missing-only transfer, and scheduler admission in the same way as vLLM P/D disaggregation.

Therefore, DiT needs a KV cache manager that manages KV cache for both context and target tokens, with prefix caching for the stable context.

### Disaggregated Management in vLLM

For vLLM's multi-instance disaggregation design, this RFC focuses on the P2P Mooncake path used by vLLM-Omni and does not discuss Mooncake Store in detail.

1. **Transfer semantics:**

   1. The Decode/receiver scheduler first allocates destination blocks through its own `KVCacheManager`, and only then starts the transfer. The data is written directly into the paged KV cache managed by Decode, typically on the GPU, although this is configurable.

   2. In the P2P path, once a Prefill request finishes computing, its KV blocks cannot be released immediately. They must remain leased until the P→D transfer has finished reading them.

```Python
The proxy concurrently sends requests to P and D with the same transfer_id
          |                              |
          v                              v
[...]

PR #6094, opened 2026-08-12
[Diffusion] Add native KV cache initialization and Scheduler-managed block allocation

zwhzzz0821:
## Purpose

This PR adds the Scheduler-side control plane for Scheduler-managed diffusion KV cache allocation described in RFC #5244. It connects Worker KV geometry discovery, native vLLM cache configuration, Scheduler allocation, metadata delivery, and request cleanup while keeping the existing dense execution path unchanged.

### Design

The full document is available at this link https://my.feishu.cn/wiki/AUR7whldviEI5mkOFu0c7kb6nNh for reference. 

```text
Initialization

loaded Worker Attention modules
        |
        | native KVCacheSpec per rank-local layer
        v
DiffusionExecutor gathers specs and available memory
        |
        | vLLM get_kv_cache_configs()
        | vLLM generate_scheduler_kv_cache_config()
        | vLLM resolve_kv_cache_block_sizes()
        v
rank-local KVCacheConfig -------> retained by each Worker
Scheduler KVCacheConfig --------> DiffusionKVCacheManager


Request admission

DiffusionKVRequest per CFG sequence
        |
        | prefix_len / target_len / seq_len
        v
DiffusionKVCacheManager
        |
        | native KVCacheManager.allocate_slots()
        | allocate every CFG sequence or roll back all
        v
DiffusionKVMetadata
        |
        v
NewRequestData -> SchedulerOutput -> Executor RPC -> Worker

finish / cancel / error / pop / close
        |
        v
native KVCacheManager.free()
```

Physical KV geometry is derived from the attention modules loaded on each Worker rank, rather than reconstructed from the Hugging Face config in the Scheduler. An opted-in attention module returns a native `FullAttentionSpec` using its resolved TP-local KV heads, head size, runtime KV dtype, attention semantics, and backend layout capability.

`DiffusionKVCacheManager` is a thin semantic wrapper around native `KVCacheManager`. It does not own another block pool or allocation algorithm. Its diffusion-specific responsibilities are:

- group multiple CFG sequences under one public request;
- allocate all CFG sequences atomically;
- roll back partial allocation on capacity pressure or errors;
- distinguish a temporarily full pool from a request that can never fit;
- produce one `DiffusionKVMetadata` allocation snapshot; and
- release the complete request allocation on every terminal path.

Each CFG sequence reserves its full first-step `seq_len`. `prefix_len` and `target_len` describe the retained and rewritten spans within that allocation:

```text
|------------------------ seq_len ------------------------|
|--------- prefix_len ---------|------ target_len ------| suffix
       retained across steps          rewritten each step
```

The resulting `DiffusionKVMetadata` contains the public request ID, allocation generation, per-sequence lengths, and native cache-group block IDs. It is attached to the existing `NewRequestData` payload and forwarded to the Worker without interpretation by the Executor.

### Reuse of vLLM

This PR directly reuses native vLLM 0.27 components as the physical source of truth:

- `KVCacheSpec` / `FullAttentionSpec`;
- `KVCacheConfig` and native cache groups;
- `get_kv_cache_configs()` and `max_model_len=-1` auto-fit;
- `generate_scheduler_kv_cache_config()`;
- `resolve_kv_cache_block_sizes()`;
- `KVCacheManager.allocate_slots()` and `KVCacheManager.free()`;
- `KVCacheBlocks.get_block_ids()`;
- native `BlockPool` accounting and refcounts; and
- native `VllmConfig.max_in_flight_tokens`.

Omni does not introduce a parallel KV spec hierarchy, block pool, refcount system, or physical cache lifecycle.

### Scope

This PR implements:

- rank-local KV spec and available-memory discovery after model loading;
- native Worker and Scheduler `KVCacheConfig` construction;
- propagation of rank-local configs back to Workers;
- Scheduler-owned diffusion KV allocation;
- atomic CFG allocation, rollback, backpressure, and cleanup; and
- `DiffusionKVMetadata` propagation through Scheduler output and Executor RPC.

This PR does not implement:

- Worker physical `kv_caches`, `BlockTable`, or slot mapping;
- paged diffusion attention execution;
- prefix hashing, lookup, publication, or CFG prefix sharing;
- independent `DiffusionKVContext` allocation;
- connector completion/delayed-free handling; or
- AR-to-DiT page transfer.

`dense_legacy` remains the default. It skips KV spec RPCs, native Scheduler cache configuration, block allocation, and metadata delivery. Hunyuan's existing dense KV execution remains active.

## Test Plan

```bash
/tmp/vllm027-runtime/bin/python -m pytest -q --tb=short \
  tests/diffusion/diffusion_kv \
  tests/diffusion/test_diffusion_scheduler.py \
  tests/diffusion/test_diffusion_worker.py \
  tests/diffusion/test_diffusion_engine_cleanup.py \
  tests/diffusion/test_diffusion_engine_rpc_routing.py \
  tests/diffusion/test_multiproc_engine_concurrency.py \
  tests/diffusion/models/hunyuan_image3/test_image_kv_cache_manager.py \
  tests/diffusion/models/hunyuan_image3/test_hunyuan_image3_step_execution.py \
  tests/diffusion/test_diffusion_model_runner.py \
  -k 'not test_execute_model_passes_single_request_batch_to_non_admission_pipeline \
      and not test_execute_model_accepts_bare_diffusion_output_from_single_request_pipeline'

/tmp/vllm027-runtime/bin/python -m pytest -q --tb=short \
  tests/diffusion/test_diffusion_config_propagation.py \
  tests/diffusion/test_diffusion_engine.py
```

HunyuanImage3 dense E2E was run on 2 x A100 80GB with TP=2, expert parallel, online FP8/Marlin, `TORCH_SDPA`, 512x512 output, and four denoise steps.

**vLLM Version:** 0.27.0

**vLLM-Omni Commit:** based on `8b0a6779d4dd485fdcf4b6a180ef75ec7bdfb95f`

## Test Result

- Diffusion KV/Scheduler/Worker/Executor/Hunyuan suite:
  `282 passed, 2 deselected`.
- Structured config and Engine suite: `166 passed, 1 skipped`.
- Ruff check, Ruff format, Python compileall, and `git diff --check`: passed.

The tests cover native vLLM request conformance, rank-local spec/config generation, native `max_model_len=-1` auto-fit, CFG atomic allocation and rollback, capacity backpressure, metadata propagation, multi-rank failure handling, and release on completion/cancel/error/pop/reinitialize/close.

Dense HunyuanImage3 E2E results:

| Version | E2E wall | Stage generation | Avg. denoise step | Model memory |
|---|---:|---:|---:|---:|
| isolated `upstream/main` | 1942.729 ms | 1941.744 ms | 485.436 ms | 42.1744 GiB |
| PR candidate | 1922.932 ms | 1921.890 ms | 480.473 ms | 42.1744 GiB |
| final candidate smoke | 1875.027 ms | 1874.161 ms | 468.540 ms | 42.1744 GiB |

The single-request smoke does not claim a performance improvement, but it shows no dense-path latency or memory regression. The online FP8/Marlin EP environment is not bitwise deterministic; visual inspection showed consistent subject, composition, and semantics with no obvious output-quality regression.

## CC List
@hsliuustc0106 @asukaqaq-s @Bounty-hunter @Acerak01-fy


hsliuustc0106:
@zwhzzz0821 please review the codes by yourself first

zwhzzz0821:
> @zwhzzz0821 please review the codes by yourself first

I self-reviewed this code before submitting the PR. I’ll go through the PR again.

vllm_omni/diffusion/worker/diffusion_worker.py
```diff
@@ -384,6 +319,58 @@ def load_model(self, load_format: str = "default", custom_pipeline_name: str | N
...
+            requested_memory = math.ceil(total_memory * self.vllm_config.cache_config.gpu_memory_utilization)
+            return max(0, requested_memory - int(process_memory))
+
+        return [
+            int(value)
+            for value in _run_and_gather_rank_values(
+                "Diffusion KV memory discovery",
+                determine_local_memory,
+            )
+        ]
```
zwhzzz0821, line 433: Upstream determines the available GPU memory through a profile_run. Could our current approach be too simplistic?

vllm_omni/diffusion/vllm_config.py
```diff
@@ -0,0 +1,171 @@
...
+
+    mode = getattr(getattr(od_config, "diffusion_kv_mode", None), "value", None)
+    if mode == "paged_scheduler":
+        raise ValueError(
+            "paged_scheduler Diffusion KV requires max_model_len or a model config "
+            "with max_position_embeddings/max_sequence_length/seq_length"
+        )
+    # Dense execution does not consume the native KV admission bound, but the
+    # shared Worker VllmConfig still requires a positive placeholder.
+    return 1
```
zwhzzz0821, line 59: It feels like this design does not sufficiently account for other diffusion models. Models such as Wan2.2 may not fit this kind of calculation, since factors like the number of frames, resolution, and VAE downsampling rate also need to be considered. We need a more general design.

hsliuustc0106:
design doc update needed or not?

zwhzzz0821:
> design doc update needed or not?

I updated Section 4.2, the Scheduler section, in the design document.

hsliuustc0106:
> > design doc update needed or not?
> 
> I updated Section 4.2, the Scheduler section, in the design document.

I didn't find it in this PR

zwhzzz0821:
> > > design doc update needed or not?
> > 
> > 
> > I updated Section 4.2, the Scheduler section, in the design document.
> 
> I didn't find it in this PR

https://my.feishu.cn/wiki/AUR7whldviEI5mkOFu0c7kb6nNh you can check it in this link and I will add it to the pr doc

vllm_omni/diffusion/sched/base_scheduler.py
```diff
@@ -104,14 +145,42 @@ def schedule(self) -> DiffusionSchedulerOutput:
...
+            if self._diffusion_kv_manager is not None:
+                if self._diffusion_kv_manager.has_request(request_id):
+                    diffusion_kv_metadata = self._diffusion_kv_manager.get_metadata(request_id)
+                else:
+                    try:
+                        allocation = self._diffusion_kv_manager.reserve_request(
+                            request_id,
+                            state.diffusion_kv_requests,
+                        )
+                    except DiffusionKVAdmissionError as exc:
```
hsliuustc0106: This lets an unexpected native `reserve_request()` error escape `schedule()`. The async engine calls `schedule()` at `diffusion_engine.py:490` outside its exception handler, so the busy-loop thread exits and any request stream waits forever. Please convert this failure into a terminal request/engine error and wake the stream, with an async regression test.

vllm_omni/diffusion/sched/base_scheduler.py
```diff
@@ -104,14 +145,42 @@ def schedule(self) -> DiffusionSchedulerOutput:
...
             if not self._running:
                 self._running_sampling_params_key = state.sampling_params_key
             state.status = DiffusionRequestStatus.RUNNING
             self._running.append(request_id)
             if was_new_request:
-                scheduled_new_reqs.append(NewRequestData.from_state(state))
+                scheduled_new_reqs.append(
+                    NewRequestData.from_state(
+                        state,
+                        diffusion_kv_metadata=diffusion_kv_metadata,
```
hsliuustc0106, line 184: This metadata makes `has_diffusion_kv_metadata` true, so the existing DLO+AllGather batch path in `multiproc_executor.py:471` is skipped and the loop at line 541 sends one scalar request per RPC. With DLO DP, every rank then executes every request instead of selecting one request by DP rank, losing the advertised DP concurrency. Please carry per-request KV metadata through the fused wave or reject this combination explicitly.

vllm_omni/diffusion/vllm_config.py
```diff
@@ -0,0 +1,171 @@
...
+    multimodal_config: Any | None = None
+    enforce_eager: bool = False
+    disable_cascade_attn: bool = False
+    enable_return_routed_experts: bool = False
+    enable_prompt_embeds: bool = False
+    logits_processors: tuple[str, ...] = ()
+    use_mla: bool = False
+    is_moe: bool = False
+
+    def is_quantized(self) -> bool:
```
hsliuustc0106, line 85: vLLM 0.27 exposes `ModelConfig.is_quantized` as a property, but this adapter exposes a method. Any native/property consumer therefore sees the bound method as truthy even for an unquantized diffusion config. Please make this a property and update the corresponding test at `tests/diffusion/diffusion_kv/test_initialization.py:217`.

Bounty-hunter:
@vllm-omni-review-bot

vllm_omni/diffusion/diffusion_kv/manager.py
```diff
@@ -0,0 +1,166 @@
...
+                    request,
+                    num_new_tokens=request.seq_len,
+                    delay_cache_blocks=True,
+                    full_sequence_must_fit=True,
+                )
+                if blocks is None:
+                    self._rollback(allocated)
+                    allocated.clear()
+                    if pool_was_empty:
+                        raise DiffusionKVAdmissionError(
```
Bounty-hunter: Since we have check `if request.seq_len <= self.max_model_len:`  is posible that failed to alloc even block pool is empty; In other words, the KV manager allocates according to max_model_len, but it can't fit a single sequence of max_model_len tokens.

vllm_omni/diffusion/sched/base_scheduler.py
```diff
@@ -49,9 +53,20 @@ def __init__(self) -> None:
...
+    def initialize(
+        self,
+        od_config: OmniDiffusionConfig,
+        *,
+        kv_cache_config: KVCacheConfig | None = None,
+        scheduler_block_size: int | None = None,
+        hash_block_size: int | None = None,
+        kv_vllm_config: VllmConfig | None = None,
+    ) -> None:
+        if self._diffusion_kv_manager is not None:
```
Bounty-hunter: Over-defensive: `initialize()` won't be called multiple times

vllm_omni/diffusion/diffusion_engine.py
```diff
@@ -189,7 +192,27 @@ def __init__(
         self.execution_mode = self._resolve_execution_mode(od_config)
         self._init_executor(od_config)
         try:
-            self._init_scheduler(od_config, scheduler)
+            kv_control_plane = initialize_diffusion_kv_control_plane(
```
Bounty-hunter, line 204: `_dummy_run` executes after `initialize_diffusion_kv_control_plane`, so 
`determine_available_kv_memory()` does not account for runtime activation 
memory. This may cause OOM during subsequent execution. Please check

Bounty-hunter:
Run the accuracy test with enable page block manager:
```
pytest -s -v test_hunyuan_image3_pixel_accuracy.py
```

vllm_omni/diffusion/diffusion_kv/manager.py
zwhzzz0821: We need to consider that a request should only be regarded as successfully allocated after all of its CFG branches have been fully allocated. Therefore, it is still possible for the free pool to be unable to accommodate an entire request. This change is intended to guarantee atomic allocation across all CFG branches.

vllm_omni/diffusion/sched/base_scheduler.py
zwhzzz0821: Fixed. The Scheduler now converts both admission failures and unexpected native allocator exceptions into terminal errors for the current request.

vllm_omni/diffusion/sched/base_scheduler.py
zwhzzz0821: Fixed. The DLO fused wave now passes the complete list[NewRequestData]. Each DP rank receives the full envelope containing both the request and its corresponding DiffusionKVMetadata, preserving DP concurrency without separating the request from its KV metadata.

vllm_omni/diffusion/vllm_config.py
zwhzzz0821: fixed

vllm_omni/diffusion/sched/base_scheduler.py
zwhzzz0821: Removed the redundant close/reset branch during re-initialization. The current Scheduler lifecycle only initializes once, so this extra defensive logic is unnecessary.

vllm_omni/diffusion/diffusion_engine.py
zwhzzz0821: It now supports running a profile_run during initialization to obtain the actual peak GPU memory usage.

zwhzzz0821:
> Run the accuracy test with enable page block manager:
> 
> ```
> pytest -s -v test_hunyuan_image3_pixel_accuracy.py
> ```

I ran the tests on 2×A100 GPUs. Since I did not have enough GPUs available, I used FP8 inference, which differs from the baseline. Both Paged Online and Dense Online fail the SSIM and PSNR checks, but I believe this is caused by FP8, as there is no significant difference between the dense and paged results.

Also, the current code does not actually modify the data plane; the changes are limited to the control plane.

## Offline result

- mean absolute difference: 0.01529885 (threshold <= 0.03)
- p99 absolute difference: 0.1176471 (threshold <= 0.30)
- SSIM: 0.976928 (threshold >= 0.97)
- PSNR: 31.419790 dB (threshold >= 30 dB)
- Result: PASS

## Online paged result

- mean absolute difference: 0.02084441 (threshold <= 0.03): PASS
- p99 absolute difference: 0.1882353 (threshold <= 0.30): PASS
- SSIM: 0.959334 (threshold >= 0.97): FAIL
- PSNR: 27.939274 dB (threshold >= 30 dB): FAIL
- Result: pixel thresholds pass; H100/BF16 similarity thresholds do not pass

## Online dense control

- mean absolute difference vs official baseline: 0.01869277
- p99 absolute difference vs official baseline: 0.1647059
- SSIM vs official baseline: 0.963104
- PSNR vs official baseline: 28.999220 dB

The dense control also misses the four-H100/BF16 SSIM and PSNR thresholds in this two-A100 online-FP8 environment.

vllm_omni/diffusion/vllm_config.py
zwhzzz0821: I think we can update the initialization logic on a per-model basis when a specific model actually requires different behavior.

zengchuang-hw:
@Bounty-hunter plz check again

Bounty-hunter (approved):
LGTM

vllm_omni/diffusion/diffusion_kv/manager.py
```diff
@@ -0,0 +1,166 @@
...
+                blocks = self.native_manager.allocate_slots(
+                    request,
+                    num_new_tokens=request.seq_len,
+                    delay_cache_blocks=True,
+                    full_sequence_must_fit=True,
+                )
+                if blocks is None:
+                    self._rollback(allocated)
+                    allocated.clear()
+                    if pool_was_empty:
```
hsliuustc0106: A request that can never fit is only detected when the pool happens to be fully idle at admission time. Under steady load the pool may not drain for a while, and until it does the oversized request head-of-line blocks the whole waiting queue. Could this be decided upfront instead — sum(ceil(seq_len / block_size)) across the CFG sequences vs `kv_cache_config.num_blocks` — rather than relying on the empty-pool heuristic?

vllm_omni/diffusion/diffusion_engine.py
```diff
@@ -860,47 +901,86 @@ def run_startup_warmup(self) -> None:
...
+        (for example, video frame count) rather than reusing it silently.
+        """
+
+        if (
+            getattr(self.od_config, "diffusion_kv_mode", DiffusionKVCacheMode.DENSE_LEGACY)
+            is not DiffusionKVCacheMode.PAGED_SCHEDULER
+        ):
+            return None
+        request = self._make_dummy_request(
+            height=1024,
```
hsliuustc0106, line 1008: The KV memory budget is derived from a fixed 1024x1024 CFG profile, but admission is bounded by max_model_len from the HF config. If a user can request a larger resolution (or the model supports sizes beyond this recipe), activation memory exceeds the profiled budget while admission still passes, and the derived KV budget is oversubscribed. Is there validation elsewhere that ties request resolution to the profiled shape?

vllm_omni/diffusion/vllm_config.py
```diff
@@ -0,0 +1,183 @@
...
+    enable_prompt_embeds: bool = False
+    logits_processors: tuple[str, ...] = ()
+    use_mla: bool = False
+    is_moe: bool = False
+
+    @property
+    def is_quantized(self) -> bool:
+        return self.quantization is not None
+
+    def is_model_moe(self) -> bool:
```
hsliuustc0106: is_quantized and is_diffusion became properties here, but is_model_moe and is_nvfp4_quantized are still plain methods. Anything reading model_config.is_model_moe as an attribute (as upstream ModelConfig allows) gets a truthy bound method. Any reason not to make them consistent?

vllm_omni/diffusion/data.py
```diff
@@ -848,6 +848,14 @@ class OmniDiffusionConfig:
     # Maximum number of sequences to generate in a batch
     max_num_seqs: int = 1
 
+    # Native vLLM KV-cache sizing inputs used only by paged_scheduler mode.
+    # These mirror the structured stage cache/scheduler config until all
+    # diffusion runtime construction consumes VllmOmniConfig directly.
+    kv_cache_memory_bytes: int | None = None
```
hsliuustc0106, line 857: Are docs for these new paged_scheduler sizing keys (kv_cache_memory_bytes / gpu_memory_utilization / max_num_batched_tokens / max_model_len) planned for the data-plane PR, or should they land here?

vllm_omni/diffusion/diffusion_kv/initialization.py
```diff
@@ -0,0 +1,82 @@
...
+        device = current_omni_platform.get_torch_device(0)
+    vllm_config = create_diffusion_vllm_config(device, od_config)
+    worker_specs = executor.get_kv_cache_specs()
+    available_memory = executor.determine_available_kv_memory(profile_request)
+    if len(worker_specs) != od_config.num_gpus or len(available_memory) != od_config.num_gpus:
+        raise ValueError(
+            "Diffusion KV initialization rank count mismatch: "
+            f"expected={od_config.num_gpus}, specs={len(worker_specs)}, memory={len(available_memory)}"
+        )
+    worker_configs, scheduler_config = build_native_kv_cache_configs(
```
hsliuustc0106: Minor: `scheduler_config` here (and in the return tuple) holds a native `KVCacheConfig` — the scheduler-side copy from `generate_scheduler_kv_cache_config` — not a `SchedulerConfig`. Upstream names this `kv_cache_config` at the same `resolve_kv_cache_block_sizes(kv_cache_config, vllm_config)` call site; a rename like `scheduler_kv_cache_config` would avoid the double-take when reading the block-size resolution below.

vllm_omni/diffusion/worker/diffusion_worker.py
```diff
@@ -384,6 +344,89 @@ def load_model(self, load_format: str = "default", custom_pipeline_name: str | N
...
+
+            with memory_profiling(
+                self.init_snapshot,
+                weights_memory=self.model_runner.model_memory_usage,
+            ) as profile_result:
+                self.model_runner.profile_run(profile_request)
+
+            available_memory = self.requested_memory - profile_result.non_kv_cache_memory
+            free_gpu_memory = profile_result.after_profile.free_memory
+            unrequested_memory = self.init_snapshot.free_memory - self.requested_memory
```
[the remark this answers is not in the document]
zwhzzz0821: Agreed and fixed.

vllm_omni/diffusion/diffusion_kv/initialization.py
zwhzzz0821: Renamed.

vllm_omni/diffusion/data.py
zwhzzz0821: I think the newly added parameters can be documented directly under stage_config so I add to it

vllm_omni/diffusion/vllm_config.py
zwhzzz0821: agreed and fixed. This was an oversight on my part.

vllm_omni/diffusion/diffusion_kv/manager.py
zwhzzz0821: Good point. I added validation for this at the scheduling stage.

vllm_omni/diffusion/diffusion_engine.py
zwhzzz0821: During the profile_run, we now also determine values such as CFG and resolution. Requests that exceed these profiled limits are then rejected during preprocessing.

zwhzzz0821:
The CI error looks not related to this pr.

hsliuustc0106:
To give the upcoming data-plane PRs a shared picture, here is a diagram of how I understand the control plane added here — drawn from reading this PR at head `c5f6159`; corrections welcome.

PR 6094 — Scheduler-managed Diffusion KV control plane

What it captures:

- **Initialization** — workers derive rank-local `FullAttentionSpec`s from opted-in attention modules; the engine builds native worker/scheduler `KVCacheConfig`s (`get_kv_cache_configs` → `generate_scheduler_kv_cache_config` → `resolve_kv_cache_block_sizes`, with `max_model_len = -1` auto-fit) and pushes the rank-local configs back to the workers; the scheduler then wraps the native `KVCacheManager` with `enable_caching = False`.
- **Admission** — `reserve_request` validates, applies the deterministic never-fit block-sum check, and allocates every CFG sequence with `full_sequence_must_fit = True` (a `None` return is atomic and means backpressure only: the request stays at the FIFO head). Requests are also gated against the startup profile envelope (CFG count, max `seq_len`, max `target_len`), and `DiffusionKVMetadata` travels inside the `NewRequestData` envelope on both the single and DLO+AllGather paths.
- **Terminals** — finish / cancel / error / pop / close all funnel into the idempotent native `free()` (reverse order, tail blocks first).
- **Coverage** — supported today vs deferred (physical data plane, cross-attention contexts, prefix sharing, paged backend, other models), all fail-fast rather than silently dense.

SVG source is [here](https://raw.githubusercontent.com/hsliuustc0106/vllm-omni-cookbook/main/docs/visualizations/pr-6094-kv-design.svg) if anyone wants to extend it for the data-plane PRs.

hsliuustc0106:
@alex-jw-brooks @DarkLight1337 @Isotr0py @ZJY0516 PTAL

vllm_omni/diffusion/executor/abstract.py
```diff
@@ -115,6 +117,39 @@ def register_failure_callback(self, callback: Callable[[], None]) -> None:
...
+        result = self.collective_rpc(
+            "determine_available_kv_memory",
+            args=(profile_request,),
+            unique_reply_rank=0,
+            exec_all_ranks=True,
+        )
+        if not isinstance(result, list) or not all(isinstance(value, int) for value in result):
+            raise TypeError("determine_available_kv_memory must return list[int]")
+        return result
+
```
Acerak01-fy, line 144: Is there a missing `get_kv_cache_config()` method here?

vllm_omni/diffusion/worker/diffusion_model_runner.py
```diff
@@ -368,6 +376,45 @@ def get_memory_context() -> AbstractContextManager[Any]:
 
         logger.info("Model runner: Initialization complete.")
 
+    def get_kv_cache_spec(self) -> dict[str, KVCacheSpec]:
```
Acerak01-fy, line 381: After PR 6102, I think this method can be placed under the `model_runner_backend.py`file in the ` diffusion_kv` directory.

vllm_omni/diffusion/diffusion_engine.py
```diff
@@ -869,47 +942,97 @@ def run_startup_warmup(self) -> None:
...
+            guidance_scale=5.0,
+            num_image_inputs=get_dummy_run_num_image_inputs(self.od_config.model_class_name),
+        )
+        if request is None:
+            raise RuntimeError("paged_scheduler requires a runnable Diffusion KV memory profile request")
+        request = self._prepare_request_for_admission(request)
+        kv_requests = request.diffusion_kv_requests
+        if not kv_requests:
+            raise RuntimeError("paged_scheduler profile preprocessing must produce Diffusion KV requests")
+        self._diffusion_kv_profile_limits = (
```
Gaohan123, line 1019: The KV memory budget is derived from profiling a single request, but the Scheduler may admit up to `max_num_seqs` compatible requests and execute them as one step batch. For example, with `step_execution=True` and `max_num_seqs=2`, two requests can each satisfy this per-request envelope while the Worker executes twice the profiled number of CFG rows. As a result, `determine_available_kv_memory()` subtracts only the single-request activation peak and can overestimate the memory available for the physical KV cache, leading to an OOM instead of Scheduler backpressure once the paged data plane allocates those tensors.

Please either profile the maximum schedulable batch for the selected execution mode, or reject `paged_scheduler` with `max_num_seqs > 1` until aggregate profiling is supported. A regression test covering `step_execution=True, max_num_seqs=2` would also be helpful. Note that DLO+DP executes one request per GPU, so the profiling rule should be execution-mode aware rather than simply multiplying by `max_num_seqs`.

Acerak01-fy (commented):
Some suggestions from the worker's perspective

vllm_omni/diffusion/executor/abstract.py
zwhzzz0821: There is no need to construct the KV config on the worker side. It should be constructed on the engine side, which is also how upstream vLLM handles it.

vllm_omni/diffusion/worker/diffusion_model_runner.py
zwhzzz0821: You may want to check #6102 first. I just looked through that PR, and it does not add any KV-spec-related code to the backend. I think these two spec methods are thin enough that keeping them here looks fine to me.

hsliuustc0106:
fix dco and make it an agent rule locally

vllm_omni/diffusion/worker/diffusion_model_runner.py
Acerak01-fy: Firstly, since the method `get_kv_cache_spec` is an operation on the modelrunner side, and the PR 6102 retains the design of the `model_runner_backend.py`, this design aims to decouple the modelrunner from the page backend. And I have defined the same method definition in it before. If there is both the modelrunnerbackend and the page backend methods defined within the modelrunner, I think it might be a bit unreasonable. Therefore, I think it is more reasonable to migrate it over.

vllm_omni/diffusion/diffusion_engine.py
zwhzzz0821: Good catch. The batched profile_run has been added.

Gaohan123 (approved):
LGTM. Thanks

vllm_omni/diffusion/diffusion_engine.py
```diff
@@ -869,47 +943,111 @@ def run_startup_warmup(self) -> None:
...
+            max(kv_request.seq_len for kv_request in kv_requests),
+            max(kv_request.target_len for kv_request in kv_requests),
+        )
+        dlo_dp_request_mode = self.execution_mode is DiffusionExecutionMode.REQUEST_BATCH and _uses_dlo_dp_concurrency(
+            self.od_config
+        )
+        profile_batch_size = 1 if dlo_dp_request_mode else _max_num_seqs(self.od_config)
+        profile_requests: list[OmniDiffusionRequest] = []
+        for index in range(profile_batch_size):
+            profile_request = copy.copy(request)
```
RuixiangMa: Profile requests share mutable attributes with the original dummy. Safe currently, but fragile if profile_run mutates shared attributes.

PR #6102, opened 2026-08-12
[Diffusion] Add native paged KV backend for diffusion workers

Acerak01-fy:
## Purpose

This PR adds the Worker-side native paged KV data plane and a generic diffusion adapter for DiT stages. It is part of RFC #5244 and builds on #5541, #5550, and #6094.

The PR integrates the control-plane, Scheduler allocation, and request metadata contracts from those earlier changes and validates the complete lifecycle with HunyuanImage3-shaped mock data:

- The Scheduler allocates logical KV blocks and produces
  `DiffusionKVMetadata`.
- The Worker consumes its rank-local native `KVCacheConfig`, allocates the
  physical KV tensors, and installs the allocation into native vLLM
  `BlockTables`.
- Omni Attention keeps its sequence-parallel pre/post hooks. The paged path
  enters the selected Omni attention backend, where
  `FlashAttentionImpl.forward_paged()` invokes vLLM's existing native
  FlashAttention implementation for the KV write and attention operation.
  No vLLM attention kernel source is modified in this PR.
- Scheduler-owned logical allocations and Worker-owned physical rows are
  retired when requests finish. The Worker only clears its rows; it does not
  free Scheduler blocks.

The implementation also:

- registers opted-in diffusion attention layers as vLLM
  `AttentionLayerBase` adapters and exposes rank-local native KV specs;
- consumes each Worker's Engine-generated `KVCacheConfig`, allocates and
  binds native KV tensors, initializes the native attention metadata builders,
  and constructs native `BlockTables`;
- validates and installs sequence/context rows from `DiffusionKVMetadata`,
  including generation idempotency, conflict detection, atomic rollback,
  terminal cleanup, and wake-time layout refresh;
- translates diffusion positions, block tables, slot mappings, and attention
  metadata to the native paged-attention interface;
- extracts only the current K/V write span when `kv_start_pos` begins inside a
  physical block, so already-installed prefix K/V is not written again; 
- keeps `DiffusionModelRunner` and `DiffusionWorker` thin by keeping the
  physical paged-KV implementation under `diffusion_kv`.

This PR intentionally excludes cross-stage KV transport, real HunyuanImage3 image-generation E2E, and W2 prefix-cache behavior. The above content will be completed in the future.

## Architecture

The paged path separates logical allocation, rank-local physical storage, and attention execution. The Scheduler never owns GPU KV tensors, and a Worker never allocates or frees Scheduler-owned logical blocks.

```text
                                        Startup control plane

            ┌────────────────────────── DiffusionEngine / Executor ───────────────────────────┐
            │ • collect each Worker rank's native KV specs and profiled memory budget         │
            │ • use native vLLM sizing to build the Scheduler config and `KVCacheConfig[r]`   │
            │ • send each rank-local config to its Worker through collective RPC              │
            └───────────────────────┬─────────────────────────────────────┬───────────────────┘
                                    │ Scheduler config                    │ `KVCacheConfig[r]`
                                    v                                     v
            ┌───────────────────────────────────────┐     ┌────────────────────────────────────┐
            │ Scheduler                             │     │ Worker rank r                      │
            │ `DiffusionKVCacheManager`             │     │ `DiffusionWorker`                  │
            │ • owns logical block allocation       │     │   -> `DiffusionModelRunner`        │
            │ • owns allocation generations         │     │     -> `DiffusionKV...Backend`     │
            │ • produces `DiffusionKVMetadata`      │     │        • physical KV tensors       │
            │ • releases logical blocks at terminal │     │        • native `BlockTables`      │
            │   request cleanup                     │     │        • request/row registry       │
            └─────────────────────┬─────────────────┘     │        • attention-batch preparation│
                                  │                       │                                    │
                                  │ request metadata      │ Model forward:                     │
                                  └── Engine / Executor ──► `Omni Attention` layer             │
                                                          │   -> Omni SP pre/post hooks        │
                                                          │   -> Omni FlashAttention backend   │
                                                          │      `FlashAttentionImpl`          │
                                                          │      `.forward_paged()`            │
                                                          │   -> vLLM native FlashAttention    │
                                                          └────────────────────────────────────┘
```

| Component | Owns | Does not own |
| --- | --- | --- |
| Engine / Executor | Native sizing orchestration, per-rank config distribution, and Worker RPC | Per-request physical rows or Scheduler allocation state |
| Scheduler | Global logical block allocation, allocation generation, and `DiffusionKVMetadata` | GPU KV tensors, native `BlockTables`, or Worker row mappings |
| Worker / ModelRunner backend | Rank-local KV tensors, `BlockTables`, request/sequence/context row mappings, metadata validation, and batch preparation | Logical block admission or logical block release |
| Omni Attention, its FlashAttention backend, and the paged adapter | SP pre/post hooks, layout translation, K/V write-span extraction, and dispatch to the selected paged-capable Omni backend | KV capacity management or block/row lifecycle |

### Request Lifecycle

```text
Scheduler
  -> reserve logical blocks for every execution sequence
  -> attach immutable `DiffusionKVMetadata` to `NewRequestData`
  -> Engine / Executor forwards the request to every Worker rank

Worker rank r
  -> install metadata into rank-local rows and native `BlockTables`
  -> prepare positions, block tables, slot mappings, and native attention metadata
  -> activate that prepared batch for the ModelRunner forward

Omni Attention
  -> existing SP `pre_attention()`
  -> paged adapter prepares the current layer context and extracts only the
     current K/V write span
  -> selected Omni `FlashAttentionImpl.forward_paged()` writes K/V by slot
     mapping and invokes vLLM's native FlashAttention implementation
  -> existing SP `post_attention()`

Terminal cleanup
  -> Scheduler releases its logical allocation
  -> Engine / Executor asks every Worker to clear only its physical rows
```

Dense requests bypass the paged adapter and continue through the existing Omni dense backend. The common Omni attention layer and its SP communication remain the execution boundary for both modes.

### Unit Tests

Focused coverage is under:

- `tests/diffusion/diffusion_kv/`
- `tests/diffusion/attention/test_attention_sp.py`
- `tests/diffusion/attention/test_ulysses_uaa.py`
- `tests/diffusion/test_diffusion_step_pipeline.py`
- `tests/diffusion/test_diffusion_engine_cleanup.py`

It covers:

- paged-KV configuration, rank-local spec discovery/config selection, and Worker request/RPC contracts;
- native KV initialization, `BlockTables`, row identity/lifecycle,  generation handling, atomic rollback, cleanup, wake refresh, and dense-path preservation;
- adapter batch metadata, DCP lengths, block-table gathering, slot mappings, Q/K/V layout validation, native metadata forwarding, and K/V writes whose start position is not block-aligned;
- Omni pre-attention -> paged adapter -> Omni post-attention ordering;
- strict Ulysses integration and rejection of unsupported paged SP modes; and
- existing dense diffusion attention/SP and terminal-cleanup behavior.

### Local Mock Integration Test

It runs the real Scheduler, Worker, ModelRunner, diffusion KV backend, Omni Attention dispatcher, and cleanup boundaries while replacing only CUDA/native kernel execution with recording doubles.

Mock data:

| Field | Value |
| --- | --- |
| Model shape | HunyuanImage3-shaped DiT request with two CFG sequences |
| Sequence count | 2 |
| `prefix_len` | 33 and 37 |
| Current write length | 17 tokens per sequence |
| `seq_len` | 50 and 54 |
| `kv_start_pos` | 33 and 37 |
| Native block size | 16 tokens |
| Worker rows per request | 2 |
| `max_model_len` | 128 tokens |
| `max_num_seqs` | 1 |
| `max_num_batched_tokens` | 128 |
| KV dtype | `torch.bfloat16` |
| Mock attention geometry | 2 KV heads, head size 8 |
| GPU ranks | 1 |

The K/V input is a packed full-sequence tensor containing 104 tokens (50 + 54). The adapter extracts only the two 17-token write spans, and both spans start inside a physical block (offsets 1 and 5). This exercises native slot-mapping and page-table updates for non-aligned writes without writing the prefix a second time.

The harness does not load model, text-encoder, or VAE weights and does not run the complete DiT image-generation pipeline.

### Static Checks

```bash
ruff check <changed production and diffusion-KV test paths>
ruff format --check <same paths>
python -m compileall <same paths>
git diff --check
```

**vLLM Version:** `0.27.0` 

**vLLM-Omni Commit:** `1d0d502f` 

## Test Result

The focused CPU verification currently available in the local environment
completed successfully:

- `tests/diffusion/diffusion_kv/ -m "not gpu"`: `141 passed, 2 deselected`;
- SP tests: `10 passed, 7 skipped`;
- mock Scheduler -> Worker -> Omni paged backend -> cleanup flow: `PASS`;
- Ruff lint and format checks: passed;
- Python compile check and `git diff --check`: passed; and
- all staged pre-commit hooks: passed.

## CC List

@hsliuustc0106 @asukaqaq-s @Bounty-hunter @Gaohan123 @zwhzzz0821 @fhfuih


tests/diffusion/diffusion_kv/test_startup_control_plane.py
```diff
@@ -0,0 +1,754 @@
+from __future__ import annotations
+
```
Acerak01-fy: Can the code be simplified?

vllm_omni/config/omni_config.py
```diff
@@ -697,6 +702,13 @@ def __post_init__(self) -> None:
             )
 
         self.diffusion_kv_mode = parse_diffusion_kv_cache_mode(self.diffusion_kv_mode)
+        if (
+            self.diffusion_kv_mode is DiffusionKVCacheMode.PAGED_SCHEDULER
+            and self.diffusion_kv_max_rows_per_request is None
+        ):
+            raise ValueError("paged_scheduler requires diffusion_kv_max_rows_per_request to be set")
```
Acerak01-fy, line 824: What is the relationship between the verification here and the verification in config.py? Is it a duplication?

vllm_omni/diffusion/executor/abstract.py
```diff
@@ -115,6 +117,71 @@ def register_failure_callback(self, callback: Callable[[], None]) -> None:
         """
         return None
 
+    def get_kv_cache_specs(self) -> list[dict[str, KVCacheSpec]]:
```
Acerak01-fy, line 120: Why does the abstract method under Executor have a series of operations such as get_kv_cache_specs and get_kv_cache_config? I don't understand why this shouldn't be returned by the engine? Is this the same design used in vllm?

vllm_omni/diffusion/diffusion_engine.py
```diff
@@ -238,6 +240,54 @@ def _init_executor(self, od_config: OmniDiffusionConfig) -> None:
...
+            for rank, available_bytes in enumerate(available_memory)
+            if not isinstance(available_bytes, int) or available_bytes <= 0
+        ]
+        if invalid_memory_ranks:
+            raise RuntimeError(
+                "paged_scheduler requires a positive native KV memory budget on every Worker; "
+                f"invalid rank(s) {invalid_memory_ranks}"
+            )
+
+        kv_cache_configs = self.executor.get_kv_cache_configs(
```
Acerak01-fy: Why does the engine need to call the executor's method to obtain the kvcache specs? Is vllm designed in this way as well? Why do you want to design it this way?

Acerak01-fy (commented):
The first round of pre-review

vllm_omni/diffusion/diffusion_kv/model_runner_backend.py
```diff
@@ -0,0 +1,453 @@
...
+
+    def _get_max_model_len(self) -> int:
+        max_len = getattr(self.vllm_config.model_config, "max_model_len", None)
+        if type(max_len) is not int or max_len <= 0:
+            raise ValueError("paged_scheduler requires a positive model_config.max_model_len")
+        return max_len
+
+    def get_kv_cache_spec(self) -> dict[str, KVCacheSpec]:
+        """Discover native KV cache specs from the loaded forward context."""
+        from vllm.v1.worker.gpu.attn_utils import get_kv_cache_spec
```
Acerak01-fy: Can this be moved to the beginning section?

vllm_omni/diffusion/diffusion_kv/model_runner_backend.py
```diff
@@ -0,0 +1,453 @@
...
+
+        return get_kv_cache_spec(self.vllm_config)
+
+    def initialize_kv_cache(self, kv_cache_config: KVCacheConfig) -> None:
+        """Allocate and bind native attention KV tensors for this rank."""
+        if self.kv_cache_config is not None:
+            raise RuntimeError("Diffusion native KV cache is already initialized")
+        from vllm.config import set_current_vllm_config
+        from vllm.v1.worker.gpu.attn_utils import init_attn_backend, init_kv_cache
+        from vllm.v1.worker.gpu.block_table import BlockTables
```
Acerak01-fy: Isn't it already imported at the beginning?

vllm_omni/diffusion/diffusion_kv/paged_attention_adapter.py
```diff
@@ -0,0 +1,510 @@
...
+            layer_causality = {
+                not bool(getattr(self.layers[layer_name].spec, "non_causal", False))
+                for layer_name in cache_group.layer_names
+            }
+            if len(layer_causality) != 1:
+                raise ValueError(f"Paged attention cache group {group_index} mixes causal and non-causal layers")
+            causal_by_group[group_index] = layer_causality.pop()
+        return causal_by_group
+
+    def _resolve_reorder_batch_threshold(self) -> int | None:
```
Acerak01-fy, line 294: Are these methods necessary for our current situation?

vllm_omni/diffusion/diffusion_kv/paged_attention_adapter.py
```diff
@@ -0,0 +1,510 @@
...
+                    )
+
+    def prepare_batch(
+        self,
+        rows: Sequence[DiffusionPagedAttentionRow],
+    ) -> PreparedDiffusionPagedAttentionBatch:
+        rows = tuple(rows)
+        if not rows:
+            raise ValueError("Paged attention batch must contain at least one row")
+        if self._active_batch is not None:
```
Acerak01-fy, line 397: Is the "batch" here specifically supported by vLLM? Does the "diffusion" feature support this?

zengchuang-hw:
@vllm-omni-review-bot

vllm_omni/diffusion/data.py
```diff
@@ -944,6 +947,15 @@ def __post_init__(self):
...
         self.diffusion_kv_mode = parse_diffusion_kv_cache_mode(self.diffusion_kv_mode)
+        if self.diffusion_kv_max_rows_per_request is not None and (
+            type(self.diffusion_kv_max_rows_per_request) is not int or self.diffusion_kv_max_rows_per_request <= 0
+        ):
+            raise ValueError("diffusion_kv_max_rows_per_request must be a positive integer when set")
+        if (
+            self.diffusion_kv_mode is DiffusionKVCacheMode.PAGED_SCHEDULER
+            and self.diffusion_kv_max_rows_per_request is None
+        ):
+            raise ValueError("paged_scheduler requires diffusion_kv_max_rows_per_request to be set")
```
hsliuustc0106, line 1004: The paged_scheduler checks here catch config-shape issues, but nothing gates the platform: on NPU the mode passes validation, the control plane starts, and the failure only surfaces later inside the worker (native GPU attention backend selection and the Triton-based BlockTables slot mapping under `vllm.v1.worker.gpu.*`). Since the only validated lane in this series is CUDA, consider rejecting non-CUDA platforms at config time the way `mxfp4_config.py` does for its NPU-only modes — the gate can then be lifted per platform as lanes validate.

vllm_omni/diffusion/data.py
```diff
@@ -668,6 +668,9 @@ class OmniDiffusionConfig:
 
     # Local Diffusion KV ownership and cache-layout mode.
     diffusion_kv_mode: DiffusionKVCacheMode = DiffusionKVCacheMode.DENSE_LEGACY
+    # Maximum number of native BlockTable rows one public request can own
+    # (sequences plus independent contexts). The model adapter defines it.
+    diffusion_kv_max_rows_per_request: int | None = None
```
hsliuustc0106, line 719: After this PR the mode is settable end to end (control plane, worker pool, scheduler admission), but no production model consumes it yet: `DiffusionPagedAttentionAdapter.prepare_batch/activate/forward` have no callers outside tests until the W2a dispatch seam lands, so enabling `paged_scheduler` today reserves the native KV pool and still runs the dense path, silently. Two suggestions: a one-line warning at worker init while the data plane is waiting on the model-side integration, and a docs entry for the series' config surface (`diffusion_kv_mode`, `diffusion_kv_max_rows_per_request`, `kv_cache_memory_bytes`, `gpu_memory_utilization`, `max_model_len`) landing with W2a — `docs/` currently has zero mentions of any of it.

zengchuang-hw:
UT failed

vllm_omni/diffusion/data.py
Acerak01-fy: Thanks for pointing this out. This is the intended W1/W2 split: this PR provides the generic Worker data plane and paged-attention adapter, while the HunyuanImage3 model dispatch integration and real end-to-end coverage will land in W2a.

  I added a Worker startup warning when `paged_scheduler` is selected. It now makes clear that native paged KV storage is initialized, but production model attention remains on the dense path until the model-side integration lands.

  The latest main branch already documents the generic sizing fields `kv_cache_memory_bytes`, `gpu_memory_utilization`, and `max_model_len` in `docs/configuration/stage_configs.md`. W2a will document the
  diffusion-specific `diffusion_kv_mode` and `diffusion_kv_max_rows_per_request` configuration surface together with its supported model path.

vllm_omni/diffusion/data.py
Acerak01-fy: fixed.

vllm_omni/diffusion/diffusion_kv/paged_attention_adapter.py
```diff
@@ -0,0 +1,510 @@
...
+        self,
+        *,
+        layer_name: str,
+        layer: Any,
+        spec: AttentionSpec,
+        vllm_config: VllmConfig,
+        device: torch.device,
+    ) -> None:
+        with set_current_vllm_config(vllm_config):
+            attn_backend = get_attn_backend(
```
Bounty-hunter: It can select attention from `vllm_omni/diffusion/attention/backends/` or just use vllm attention backend.

Acerak01-fy: Just use vllm attention backend. Since the attention backend in Omni expects continuous dense K/V inputs without paging, and both kvcache and tensors are stored as contiguous blocks, this means that the Omni attention backend does not support block tables or slot mapping. If we want to perform paging while still using Omni's attention mechanism, we would need to gather page KV data before feeding it into the Omni attention backend, which introduces additional overhead. Therefore, I think it would be better to use VLLM's attention backend when handling paged KV.

Bounty-hunter: My concern is that many accelerate feature rely on Omni-Attention, e.g., sp.

Acerak01-fy:
> UT failed

fixed.

vllm_omni/diffusion/diffusion_kv/paged_attention_adapter.py
Acerak01-fy: This is indeed a problem. I have made the correction. This time, it follows the same path as the original `dense` and both use the `omni attention backend`. Then, under this `omni backend`, in the `flash_atten.py` file, I added a forward method that supports pages. It follows the `vllm flash_attention kernel`.

vllm_omni/diffusion/attention/backends/flash_attn.py
```diff
@@ -246,6 +247,48 @@ def _forward_varlen_dense(
         # (b s) h d -> b s h d
         return out.reshape(batch_size, q_len, *out.shape[1:])
 
+    def forward_paged(self, paged_kv_context) -> torch.Tensor:
```
Bounty-hunter, line 251: Additionally, `piecewise_attn` needs to be adapted to support paged attention. It is a wrapper around FlashAttention and is required for mixed attention masks (such as Hunyuan). I think this adaptation should be fairly straightforward: `piecewise_attn` splits the query into multiple segments (shards), with each segment corresponding to a different KV endpoint.

vllm_omni/diffusion/diffusion_kv/config.py
```diff
@@ -36,4 +36,18 @@ def parse_diffusion_kv_cache_mode(value: object) -> DiffusionKVCacheMode:
             "Diffusion KV diffusion_kv_mode 'paged_worker_local' is reserved but not implemented; "
             "use 'dense_legacy' or 'paged_scheduler'"
         )
+    if mode is DiffusionKVCacheMode.PAGED_SCHEDULER:
+        # Native BlockTables and attention execution are validated only on CUDA.
+        from vllm_omni.platforms import current_omni_platform
+
+        if not current_omni_platform.is_cuda():
```
Bounty-hunter: This should be hardware-platform agnostic, right? At the very least, NPU should support it as well.

vllm_omni/diffusion/attention/backends/flash_attn.py
Acerak01-fy: Yes,  I have completed the page adaptation for `piecewise_attn`.

vllm_omni/diffusion/diffusion_kv/config.py
Acerak01-fy: Yes, agreed. I removed the CUDA-only restriction and made the paged KV path platform-agnostic in the latest revision.
  > - BlockTables and attention metadata creation now go through platform hooks.
  > - NPU uses AscendBlockTables and vLLM-Ascend’s native metadata builder with ChunkedPrefill.
  > - Backend selection accepts the platform-native attention backend instead of requiring CUDA FlashAttention.
  > - KV cache updates respect forward_includes_kv_cache_update, so Ascend keeps its inline cache-update path while CUDA can use the separate writer.
  > - Piecewise paged attention also passes segment-specific Q/K/V and slot mappings to the native backend.

 I added tests for native backend selection and the Ascend-style inline KV-update contract. I have not run it on physical NPU hardware in my current environment, so hardware validation is still pending.

hsliuustc0106 (commented):
The remaining model/adapter boundary issues are inline. I’m treating them as W2 promotion blockers because production HunyuanImage-3 activation is explicitly deferred from this W1 PR.

vllm_omni/diffusion/diffusion_kv/paged_attention_adapter.py
```diff
@@ -0,0 +1,855 @@
...
+
+        query_flat, query_token_shape, query_has_head_dims = self._flatten_tensor(
+            query,
+            num_heads=layer.num_heads,
+            head_size=layer.head_size,
+            name="query",
+        )
+        key_flat, key_token_shape, _ = self._flatten_tensor(
+            key,
+            num_heads=layer.num_kv_heads,
```
hsliuustc0106, line 671: HunyuanImage-3’s non-AllGather path repeats K/V from the KV-head count to the Q-head count before calling Omni Attention. A 32Q/8KV layer therefore reaches this code as `[B, S, 32, D]`, but it is parsed using `layer.num_kv_heads == 8`, so `_flatten_tensor` raises before paged attention runs. Please keep K/V compressed when the paged adapter is active, or explicitly collapse validated repeats, and cover the real 32Q/8KV boundary.

vllm_omni/diffusion/diffusion_kv/paged_attention_adapter.py
```diff
@@ -0,0 +1,855 @@
...
+            if len(token_shape) == 2:
+                batch_size, tokens_per_row = token_shape
+                if batch_size != len(batch.rows) or any(row.query_len != tokens_per_row for row in batch.rows):
+                    raise ValueError(
+                        f"Paged attention {name} query-aligned layout does not match prepared rows: "
+                        f"shape={token_shape}, row_query_lens={tuple(row.query_len for row in batch.rows)}"
+                    )
+            return tensor
+
+        full_tokens = sum(row.seq_len for row in batch.rows)
```
hsliuustc0106: Heterogeneous H3 batching pads restored K/V to `max_prefix + max_current`, while these rows retain compact scheduler lengths. For prefixes 33/37 plus 17 current tokens, K/V flattens to 108 tokens but `full_tokens` is 104, so this raises. Accepting 108 alone would still slice row 0 at 33 although its current tokens start at padded offset 37, and its shifted span ends at 54 while logical `seq_len` is 50. Please compact tensors/spans before this boundary or carry an explicit padded-to-logical mapping, with unequal-prefix H3 coverage.

vllm_omni/diffusion/diffusion_kv/paged_attention_adapter.py
```diff
@@ -0,0 +1,855 @@
...
+
+    @staticmethod
+    def _flatten_tensor(
+        tensor: torch.Tensor,
+        *,
+        num_heads: int,
+        head_size: int,
+        name: str,
+    ) -> tuple[torch.Tensor, tuple[int, ...], bool]:
+        if tensor.ndim >= 2 and tensor.shape[-2:] == (num_heads, head_size):
```
hsliuustc0106: A flattened one-token tensor for a single-head layer has shape `[1, head_size]`, so this branch treats its only two dimensions as `[heads, head_size]` and returns an empty token shape; `_validate_query_layout` then rejects an otherwise valid packed input. Require at least three dimensions for the explicit-head form and add a one-token/single-head regression.

vllm_omni/diffusion/data.py
hsliuustc0106: The new NPU hooks address the original GPU-specific BlockTables/metadata assumptions, but the capability gate is still too late. `paged_scheduler` is accepted before the platform and selected Omni backend are validated; on NPU without MindIE-SD, selection falls back to `TORCH_SDPA`, which does not opt into `supports_paged_kv`, so the first rejection happens inside `Attention._forward_impl` after model and native-cache setup. Platforms without overrides also inherit the GPU BlockTables/metadata defaults. Please add an explicit platform opt-in and validate every marked layer’s resolved backend before paged-cache allocation.

vllm_omni/diffusion/diffusion_kv/paged_attention_adapter.py
Acerak01-fy: Thanks for pointing this out. The issue was caused by an adapter-side assumption that all rows shared one full [prefix | current] K/V layout. To support that layout, the adapter tried to reconstruct the current write range from padded tensors using kv_start_pos, which breaks when rows have different prefix lengths.
  
 I removed this padding/full-sequence extraction path entirely. Paged callers now provide only the current Q/K/V tokens in packed (or uniform batched) form, while the prefix remains in the native BlockTable KV cache. The adapter passes native per-row metadata such as query_start_loc, seq_lens, positions, and slot mappings to the vLLM backend.
  
   This restores vLLM’s native heterogeneous paged-batch behavior: each request can have its own prefix and sequence length without padding the K/V inputs. For piecewise attention, rows still need matching segment topology and attention mode so they can share a kernel call. The dense path is unchanged.
 
  The regression test covers rows with prefix=33, current=17 and prefix=37, current=17, using only 34 current K/V tokens and per-row seq_lens=[50, 54].

vllm_omni/diffusion/diffusion_kv/paged_attention_adapter.py
Acerak01-fy: Thanks for catching this. You are right that HunyuanImage-3 produces 32 query heads and 8 KV heads, while the existing non-AllGather dense path eagerly expands K/V to 32 heads with repeat_kv(). That conflicts with the paged adapter’s native num_kv_heads=8 contract.

 I chose to keep K/V compressed on the paged path rather than trying to collapse repeated heads inside the generic adapter. Hunyuan now only calls repeat_kv() for the existing dense path. When Scheduler-managed paged KV is active, K/V remain [B, S, 8, D]; the adapter preserves them as [T, 8, D], and the native implementation receives num_heads=32 and num_kv_heads=8, allowing vLLM FlashAttention to execute the 4:1 GQA mapping natively. The dense behavior remains unchanged.

 I also added regressions using the actual 32Q/8KV geometry: one verifies that dense still expands K/V while paged keeps them compressed, and another verifies that the paged adapter preserves [T, 32, D] query and [T, 8, D] key/value shapes.

vllm_omni/diffusion/diffusion_kv/paged_attention_adapter.py
Acerak01-fy: Thanks for catching this edge case. The root cause was that a 2-D tensor such as [1,head_size] could be mistaken for an explicit [num_heads, head_size] layout when num_heads == 1.
  
I now only recognize the explicit head-dimension layout when the tensor has at least three dimensions. Therefore, 2-D [tokens, hidden_size] inputs are handled as flattened hidden-dimension tensors, including the single-token/single-head case.
  
 I added a regression test covering this shape and verified that the output is restored to the original input layout.

vllm_omni/diffusion/data.py
hsliuustc0106: Verified the platform-agnostic rework statically — the NPU hooks line up against vllm-ascend main (AscendBlockTables constructor parity, `seq_lens_cpu`→`seq_lens_np` conversion, the identity scales the Ascend impl asserts, FLASH_ATTN→AscendAttentionBackend resolution), and moving the SP-constraint checks into `register_kv_cache_layers` so they fail at startup is a good step. The remaining gap from the follow-up above still stands though: `supports_paged_kv` is only consulted at first attention dispatch (attention/layer.py), and platforms without overrides still inherit the GPU BlockTables/metadata defaults silently — a startup validation of platform + resolved backend for every marked layer would close that. An npu-test lane run would also be the first real exercise of the Ascend path.

vllm_omni/diffusion/data.py
hsliuustc0106: Warning verified, thanks. Two notes on the current head: the dispatch seam now exists in `Attention._forward_impl`, but it stays dormant until something calls `prepare_batch`/`activate` (still no production caller — W2a), so the warning text remains accurate. On docs: `kv_cache_memory_bytes` landed in docs/configuration/stage_configs.md, but `diffusion_kv_mode` / `paged_scheduler` / `diffusion_kv_max_rows_per_request` still have no docs mention, so the mode-level entry is still open for W2a.

vllm_omni/diffusion/diffusion_kv/paged_attention_adapter.py
hsliuustc0106: Verified — the ≥3-dimension requirement for the explicit head form plus the single-token/single-head regression both check out on 580e88cf.

vllm_omni/diffusion/diffusion_kv/paged_attention_adapter.py
hsliuustc0106: Verified — the padded full-sequence extraction path is gone; Q/K/V now validate against the current-write batch and `test_omni_paged_backend_runs_unequal_hunyuan_prefixes_without_padding` covers the heterogeneous-prefix boundary.

vllm_omni/diffusion/diffusion_kv/paged_attention_adapter.py
hsliuustc0106: Verified — `repeat_kv` is now gated on `is_paged_kv_active` and the compressed 32Q/8KV case has its regression test. Keeping K/V compressed on the paged path is the right call vs collapsing repeats in the generic adapter.

vllm_omni/diffusion/diffusion_kv/model_runner_backend.py
```diff
@@ -0,0 +1,635 @@
...
+        try:
+            with set_current_vllm_config(self.vllm_config):
+                for layer_name, (layer, spec) in layers.items():
+                    adapters[layer_name] = DiffusionPagedAttentionLayerAdapter(
+                        layer_name=layer_name,
+                        layer=layer,
+                        spec=spec,
+                        vllm_config=self.vllm_config,
+                        device=self.device,
+                        ulysses_degree=self._get_paged_attention_ulysses_degree(layer),
```
hsliuustc0106, line 118: The SP constraints now fail fast here at registration, which is the right place — but the backend-capability dimension is still only checked at first attention dispatch (`supports_paged_kv` in `Attention._forward_impl`), so a paged deployment whose marked layers resolve to a non-paged Omni backend (e.g. the NPU `TORCH_SDPA` fallback) builds the model and the native cache first and only fails at the first forward. Since `layer` is in hand at this call site, could the resolved `layer.attn_backend`'s `supports_paged_kv` (and the platform having BlockTables/metadata overrides) be validated here alongside the SP checks, so every marked layer is rejected at startup instead?

hsliuustc0106 (changes requested):
Re-review complete at 580e88cf — the Omni-backend routing rework, piecewise paged plans, and platform hooks all check out, and every finding from the earlier rounds is verified fixed with regression tests. Two items before this is approve-able:

1. Validate paged capability at startup rather than at first attention dispatch (inline comment).
2. The updated GPU piecewise-vs-reference test has not been executed on this head (the current local environment is CPU-only), so the `cuda-test` lane will be its first run — please let that complete (and an `npu-test` run would be the first real exercise of the Ascend path).

hsliuustc0106 (approved):

PR #6105, opened 2026-08-12
OmniVoice: Validate instructions 

clodaghwalsh17:
## Purpose

Add instructions validation for the OmniVoice model currently missing. OmniVoice expects instructions as a comma seperated list, where the [voice instructions](https://github.com/k2-fsa/OmniVoice/blob/38e992bc60f85548faeb77e8fa70158ba71deb30/omnivoice/models/omnivoice.py#L1492 ) must fall into a prefined category.  

Created ```prompt_utils.py``` for OmniVoice. Instruction validatation functions called in the TTS adapater and updated instructions validation in ```examples/offline_inference/text_to_speech/omnivoice/end2end.py```.

## Test Plan

Created ```tests/model_executor/models/omnivoice/test_validate_instructions.py```

**vLLM Version:** 0.28.0
**vLLM-Omni Commit:** cdd4638b
## Test Result

 - Previous tests pass

```
pytest -sv tests/e2e/online_serving/test_omnivoice_expansion.py
19 passed, 16 warnings in 111.78s (0:01:51)
```
- Confirmed instruction validation when running ```examples/online_serving/text_to_speech/omnivoice/speech_client.py```.

```
python examples/online_serving/text_to_speech/omnivoice/speech_client.py --text "Hello how are you" --instructions "boy"

Log Snippet
Model: k2-fsa/OmniVoice
Text: Hello how are you
Generating audio...
Error: 400
{"error": {"message": "Unsupported instruct items found in 'boy':\n  'boy' -> 'boy' (unsupported)\nValid English items: american accent, australian accent, british accent, canadian accent, child, chinese accent, elderly, female, high pitch, indian accent, japanese accent, korean accent, low pitch, male, middle-aged, moderate pitch, portuguese accent, russian accent, teenager, very high pitch, very low pitch, whisper, young adult\nValid Chinese items: \u4e1c\u5317\u8bdd\uff0c\u4e2d\u5e74\uff0c\u4e2d\u97f3\u8c03\uff0c\u4e91\u5357\u8bdd\uff0c\u4f4e\u97f3\u8c03\uff0c\u513f\u7ae5\uff0c\u56db\u5ddd\u8bdd\uff0c\u5973\uff0c\u5b81\u590f\u8bdd\uff0c\u5c11\u5e74\uff0c\u6781\u4f4e\u97f3\u8c03\uff0c\u6781\u9ad8\u97f3\u8c03\uff0c\u6842\u6797\u8bdd\uff0c\u6cb3\u5357\u8bdd\uff0c\u6d4e\u5357\u8bdd\uff0c\u7518\u8083\u8bdd\uff0c\u7537\uff0c\u77f3\u5bb6\u5e84\u8bdd\uff0c\u8001\u5e74\uff0c\u8033\u8bed\uff0c\u8d35\u5dde\u8bdd\uff0c\u9655\u897f\u8bdd\uff0c\u9752\u5c9b\u8bdd\uff0c\u9752\u5e74\uff0c\u9ad8\u97f3\u8c03\nTip: Use only English or only Chinese instructs. English instructs should use comma + space (e.g. 'male, indian accent'),\nChinese instructs should use full-width comma (e.g. '\u7537\uff0c\u6cb3\u5357\u8bdd').", "type": "BadRequestError", "param": null, "code": 400}}
```

- New tests pass

```
pytest -sv tests/model_executor/models/omnivoice/test_validate_instructions.py
7 passed, 15 warnings in 0.03s
```


vllm_omni/model_executor/models/omnivoice/prompt_utils.py
```diff
@@ -0,0 +1,138 @@
...
+    _INSTRUCT_EN_TO_ZH,
+    _INSTRUCT_MUTUALLY_EXCLUSIVE,
+    _INSTRUCT_VALID_EN,
+    _INSTRUCT_VALID_ZH,
+    _INSTRUCT_ZH_TO_EN,
+    _ZH_RE,
+)
+
+
+def validate_instruction(instruct: Any) -> str | None:
```
alex-jw-brooks: Can you please ensure we don't use `any` types hints where it's unnecessary?

vllm_omni/model_executor/models/omnivoice/prompt_utils.py
```diff
@@ -0,0 +1,138 @@
...
+            f"Unsupported instruct items found in {instruct_str}:\n"
+            + "\n".join(lines)
+            + "\n\nValid English items: "
+            + ", ".join(sorted(_INSTRUCT_VALID_EN))
+            + "\nValid Chinese items: "
+            + "，".join(sorted(_INSTRUCT_VALID_ZH))
+            + "\n\nTip: Use only English or only Chinese instructs. "
+            "English instructs should use comma + space (e.g. "
+            "'male, indian accent'),\nChinese instructs should use full-width "
+            "comma (e.g. '男，河南话')."
```
alex-jw-brooks, line 57: I think this should actually not raise an error and should instead be a warning. An architecture may have multiple variants, which could have different ideal instruction guidance that isn't necessarily rigidly enforced, and there are also cases like simplified vs traditional chinese characters, which could cause confusing behaviors in the validation

alex-jw-brooks (commented):
Some thoughts!

vllm_omni/model_executor/models/omnivoice/prompt_utils.py
clodaghwalsh17: @alex-jw-brooks I've changed the logic so a warning is raised if invalid instructions are provided

vllm_omni/entrypoints/openai/serving_speech.py
```diff
@@ -3564,6 +3568,13 @@ async def _create_diffusion_speech(
         """Handle speech generation for pure diffusion TTS models (e.g. OmniVoice)."""
         from vllm_omni.outputs import OmniRequestOutput
 
+        try:
+            # Assume that this will follow the same pattern of adapter.validate and adapter.build
+            # used in _prepare_speech_generation once all RFC is implemented
+            self._adapter.validate(request)
```
linyueqian: [important] This call does more than validate, and it runs before this path has read the state it changes.

`OmniVoiceAdapter.validate()` ends with `return self.ctx.server._apply_uploaded_speaker(request)`, which populates `request.ref_audio` for an uploaded voice. Ten lines below, `has_inline_ref_audio = request.ref_audio is not None` now sees the value this call just wrote, so an uploaded voice is recorded as inline audio. The consequence is at the prompt-building step: `voice_name` and `voice_created_at` are only set when `voice_lower in self.uploaded_speakers and not has_inline_ref_audio`, so that metadata is now dropped for exactly the requests it exists for. `_apply_uploaded_speaker` also runs a second time immediately afterwards, so the work is duplicated as well.

Capturing `has_inline_ref_audio` before this call, or having the adapter validate without applying the speaker, would restore the previous behaviour.

Separately, this validates but never normalizes. `validate_instruction` computes a normalized form internally and does not write it back, and the prompt is still built from `request.instructions` verbatim further down. So an instruction that is accepted but non-canonical, differing in case or separator, reaches OmniVoice exactly as the client sent it. If normalization is meant to be part of the contract, it needs to be applied to the request; if it is not, the normalized form is only ever used to decide acceptance and that is worth a comment saying so.

vllm_omni/entrypoints/openai/tts_adapters/omnivoice.py
```diff
@@ -25,6 +26,9 @@ class OmniVoiceAdapter(ARTTSAdapter):
     def validate(self, request: "OpenAICreateSpeechRequest") -> str | None:
         if not request.input or not request.input.strip():
             return "Input text cannot be empty"
+
+        validate_instruction(request.instructions)
```
linyueqian: [important] The same unsupported instruction produces two different outcomes depending on which serving path the deployment uses, and the AR one is the harsher of the two.

`validate_instruction` raises `SyntaxWarning` in three places rather than returning a message. The diffusion path wraps its call in `try: ... except Warning as w: logger.warning(...)` at `serving_speech.py:3571-3576`, so there the instruction is dropped with a log line and generation continues. The AR path calls `adapter.validate(request)` unguarded at `serving_speech.py:3172` (and again at `:1474`), where the contract is an error string or `None`. A raised `SyntaxWarning` is neither, so it propagates out of validation, `adapter.build()` never runs, and the caller gets a generic speech-generation failure instead of the carefully worded message this code builds.

That message is the part worth saving. It names the offending item, suggests a correction, lists valid English and Chinese items and explains the separator rule, which is exactly what a client needs in a 400. As written, the only deployment that surfaces it is the one that throws it away into a log.

Returning the message as a string would fit the existing adapter contract and let both paths render it consistently. If raising is preferred, then a dedicated exception type caught at both call sites and mapped to a 400 would work too, but `SyntaxWarning` is a poor carrier: it is a builtin warning category, so anything that catches broadly will treat it as a warning rather than the request-rejecting error it currently is.

linyueqian (changes requested):
Reviewed at `7c899861`. The validation itself is well built: the item table, the "did you mean" suggestions and the language-consistency rules are careful, and the message it constructs is genuinely useful to a caller. The problem is the contract around it. The same unsupported instruction is a logged warning on one serving path and a hard request failure on the other, and the new call site on the diffusion path mutates the request before that path has finished reading it. Both inline.

On testing, the description lists vLLM-Omni commit `7cd826d2`, and that commit is an ancestor of this branch's merge-base `728640f4`, so it predates every commit in the PR. Whatever was run there did not exercise the code under review. Worth replacing with a commit on this branch, since the behaviour most in question here is exactly the serving-path integration.

The unit tests for `validate_instruction` itself are good and I am not asking for changes to them: they cover valid English and Chinese items, unknown items, the dialect and accent conflict, and the separator rules. What they do not cover is what a caller experiences, which is where both findings live. A test that drives `_prepare_speech_generation` with an unsupported instruction and asserts the resulting status code would have caught the inconsistency.

Validation: static reads only, no code from this head executed. I read `validate_instruction` end to end and confirmed it raises `SyntaxWarning` at three separate points, traced both serving call sites, and confirmed the ancestry of the commit named in the description with `git merge-base --is-ancestor`. I did not run the OmniVoice model or exercise either endpoint, so the behavioural claims below are read from the code rather than reproduced.

vllm_omni/entrypoints/openai/serving_speech.py
clodaghwalsh17: Removed ```self.ctx.server._apply_uploaded_speaker(request)``` from the adapter validate function as ```_create_diffusion_speech``` already performs this operation.  

Good spot on the normalisation not being enforced. Ive updated the build function in the adapter to call this to keep this scoped to the specific model. Note I call ```build``` within ```_create_diffusion_speech``` to perform this check and allow others like it be introduced in future.

Is it worth changing the OmniVoiceAdapter subclass to ```DiffusionTTSAdapter```, and establishing the pattern that the call ```self.ctx.server._apply_uploaded_speaker(request)```  is not necessary in the adapters for diffusion models since it is done elsewhere.

vllm_omni/entrypoints/openai/tts_adapters/omnivoice.py
clodaghwalsh17: My original implementation raised a 400 error if there was a validation issue. An earlier review suggested to raise a warning instead so I made that change. But I have now converted it back to raise the 400 error as it seems like the right behaviour and as you said it better matches the existing adapter contract.

vllm_omni/entrypoints/openai/serving_speech.py
```diff
@@ -417,6 +418,9 @@ def for_diffusion(
             allowed_media_domains=allowed_media_domains,
         )
         instance._tts_model_type = "omnivoice"
+        # Set adapter to OmniVoice as it is currently the only diffusion TTS model
+        # Temporary assignment until https://github.com/vllm-project/vllm-omni/issues/4327 is completed
+        instance._adapter = OmniVoiceAdapter(SpeechServingContext(server=instance, engine_client=None))
```
linyueqian: [blocking] This assignment is undone five lines later. The rebase kept main's `instance._adapter = None` at line 428 while adding this `OmniVoiceAdapter(...)` line, so the diffusion instance ends construction with no adapter.

Line 3446 then calls `self._adapter.validate(request)` unconditionally, which raises `AttributeError: 'NoneType' object has no attribute 'validate'` inside the request handler. Every `/v1/audio/speech` request to an OmniVoice diffusion instance with non-empty input fails with a generic 500, including requests that pass no `instructions` at all, so this is not limited to the new validation path.

Deleting line 428 is probably the whole fix, but the comment above it explains why main set it, so the two intents need reconciling rather than one silently winning.

tests/model_executor/models/omnivoice/test_validate_instructions.py
```diff
@@ -0,0 +1,52 @@
...
+    def test_chinese_instruction(self) -> None:
+        """Test valid chinese instruction"""
+        instructions = "男， 河南话"
+        validate_instruction(instructions)
+
+    def test_free_form_instruction(self) -> None:
+        """Test free form instruction raises warning"""
+        instructions = "happy little scottish boy"
+
+        with pytest.raises(SyntaxWarning, match="Unsupported instruct items found"):
```
linyueqian: [blocking] `validate_instruction` returns its message rather than raising: every error path in `prompt_utils.py` ends in `return warning`, and the adapter treats a non-None return as the error string. So `pytest.raises(SyntaxWarning, ...)` cannot succeed and this test fails with `DID NOT RAISE`.

The same pattern is at lines 32, 39 and 51, so four of the seven tests in this file fail as written. They should assert on the returned string, for example `assert 'Unsupported instruct items found' in validate_instruction(...)`. Worth noting these tests never ran in CI: the PR has no `ready` label, so no per-PR Buildkite lane has reported and the green rollup is only DCO, pre-commit, build and docs.

vllm_omni/model_executor/models/omnivoice/prompt_utils.py
```diff
@@ -0,0 +1,138 @@
...
+        )
+        return warning
+
+    # --- Language consistency: dialect forces Chinese, accent forces English ---
+    if has_dialect(normalised) and has_accent(normalised):
+        warning = (
+            "Cannot mix Chinese dialect and English accent in a single instruct"
+            + "Dialects are for Chinese speech, accents for English speech."
```
linyueqian: [suggestion] Missing separator between the two concatenated fragments, so the message renders as `...in a single instructDialects are for Chinese speech, accents for English speech.` A space or newline before `Dialects` fixes it. This string reaches the client as a 400 body, so it is user-visible.

linyueqian (changes requested):
The uploaded-speaker work from round one holds up. Two problems block this head, both of which come from the rebase rather than from the feature itself, and neither is visible in the checks currently shown on the PR.

Root cause of the first is that the rebase left two conflicting intents side by side in the diffusion instance constructor: this PR's adapter assignment and main's reset of the same attribute, five lines apart. The result is not a narrow edge case, because the new `validate` call is unconditional, so the diffusion OmniVoice endpoint returns 500 for every non-empty request whether or not `instructions` is set.

The second is a contract mismatch rather than a logic error. `validate_instruction` is written in the return-an-error-string style the adapter layer expects, while its tests were written against a raise-based validator. Both halves are individually reasonable, which is why this survived: only running them together shows it.

Which brings up the thing worth fixing beyond the code. This PR carries no `ready` label, so no per-PR Buildkite lane has ever run on it, and the green rollup you see is only DCO, pre-commit, build and docs. Both blocking problems above would have been caught by the general lane on the first push. Adding `ready` is worth doing before the next round so the fixes get real coverage.

The Test Plan also needs refreshing: it cites commit `728640f4`, reports five E2E cases where this head adds seven, and still describes the old warning-plus-200 behaviour rather than the 400 the validator now produces.

One question rather than a finding: `OmniVoiceAdapter.validate` no longer calls `_apply_uploaded_speaker`, which every other adapter does as its first step and which `base.py` documents as the reason `has_inline_ref_audio` is captured before `validate`. On the diffusion path that looks harmless, since line 3458 still calls it explicitly, but if an OmniVoice request can ever reach the AR path at line 3105 then uploaded-speaker resolution and its error reporting are silently dropped there. Worth confirming which paths this adapter serves.

Validation: static review of the seven-file diff at `25994e99` against merge base `44c8d1c4`, the adapter and serving call order, `validate_instruction`'s return paths, the four test expectations, and the `_apply_uploaded_speaker` contract in `base.py` and the seven sibling adapters. I did not execute the branch.

clodaghwalsh17:
@linyueqian I have resolved the issues you highlighted and updated the unit tests and the test plan in the PR description. 

On this point:
> One question rather than a finding: `OmniVoiceAdapter.validate` no longer calls `_apply_uploaded_speaker`, which every other adapter does as its first step and which `base.py` documents as the reason `has_inline_ref_audio` is captured before `validate`. On the diffusion path that looks harmless, since line 3458 still calls it explicitly, but if an OmniVoice request can ever reach the AR path at line 3105 then uploaded-speaker resolution and its error reporting are silently dropped there. Worth confirming which paths this adapter serves.
> 

Is there any case where the OmniVoice model does go through the AR path since the model pipeline explicitly states it is a single stage diffusion model? To make it clear that the `_apply_uploaded_speaker` is called later in the diffusion path should OmniVoiceAdapter inherit the DiffusionTTSAdapter class, and establish the pattern that the call `self.ctx.server._apply_uploaded_speaker(request)` is not necessary in the adapters for diffusion models since it is done elsewhere.

vllm_omni/entrypoints/openai/tts_adapters/omnivoice.py
```diff
@@ -25,7 +26,8 @@ class OmniVoiceAdapter(ARTTSAdapter):
     def validate(self, request: "OpenAICreateSpeechRequest") -> str | None:
         if not request.input or not request.input.strip():
             return "Input text cannot be empty"
-        return self.ctx.server._apply_uploaded_speaker(request)
+
+        return validate_instruction(request.instructions)
```
linyueqian, line 34: [blocking] Replacing the body of `validate` dropped the `_apply_uploaded_speaker` call, and on the AR path nothing else makes it.

`serving_speech.py` now calls `_apply_uploaded_speaker` in exactly one place, line 2081, which is inside the diffusion request path right after the new `validate` call. The AR paths call `adapter.validate(request)` at lines 1063 and 1740 and nowhere else, so for AR OmniVoice this method was the only thing resolving an uploaded speaker.

Two things regress there. An uploaded voice never gets resolved into `request.ref_audio`, so a request naming a stored speaker silently proceeds without it, and an invalid uploaded-speaker name no longer produces its validation error. `base.py` documents the coupling explicitly: `has_inline_ref_audio` is captured before `validate()` runs precisely because `_apply_uploaded_speaker` mutates `request.ref_audio` in place, so the ordering here is load-bearing rather than incidental.

Every other adapter still does this as the first step of its own `validate`: audio8_tts, cosyvoice3, fish_speech, glm_tts, higgs_audio_v2, higgs_audio_v3, moss_tts and voxcpm2. OmniVoice is now the only one that does not, which is what makes this look unintended rather than a deliberate divergence.

Keeping both, returning the speaker error first and then `validate_instruction`, restores the AR path without affecting the diffusion path, which calls `_apply_uploaded_speaker` separately anyway. I cannot read the Buildkite step logs, so I am not asserting this is the exact assertion that fails, but it is the strongest candidate for the `Simple · Engine&Entrypoints Test` red.

vllm_omni/model_executor/models/omnivoice/prompt_utils.py
```diff
@@ -0,0 +1,138 @@
...
+            "'male, indian accent'),\nChinese instructs should use full-width "
+            "comma (e.g. '男，河南话')."
+        )
+        return warning
+
+    # --- Language consistency: dialect forces Chinese, accent forces English ---
+    if has_dialect(normalised) and has_accent(normalised):
+        warning = (
+            "Cannot mix Chinese dialect and English accent in a single instruct."
+            + "Dialects are for Chinese speech, accents for English speech."
```
linyueqian: [suggestion] Still concatenated. The period added to the previous line helps, but with no space or newline before `Dialects` the message renders as `...in a single instruct.Dialects are for Chinese speech, accents for English speech.` This string reaches the client as a 400 body.

linyueqian (changes requested):
Both blocking findings from round two are fixed, and fixed the right way. The duplicate `instance._adapter = None` is gone from the diffusion constructor, so the adapter assigned two lines earlier now survives and the unconditional `validate` call at `serving_speech.py:2069` no longer dereferences `None`. The four tests that expected `pytest.raises(SyntaxWarning)` now assert on the returned string instead, matching the validator's actual return-an-error-string contract.

The general lane is still red though, and it is this PR rather than the tree. Build 14886 ran on the current head `ad2f12b4` and hard-failed `Simple · Engine&Entrypoints Test`, while general builds 14911 and 14884 from the same window on unrelated PRs had no hard failures at all, so the tree is fine and this is not the inherited signature.

The likely cause is the one thing I raised last round as a question rather than a finding, and I was too soft on it. `OmniVoiceAdapter.validate` no longer calls `_apply_uploaded_speaker`, and it turns out that method now has exactly one call site in `serving_speech.py`, inside the diffusion path. The AR path reaches it only through `adapter.validate`, so for AR OmniVoice uploaded-speaker resolution and its error reporting are simply gone. Every one of the eight sibling adapters still calls it first thing in their own `validate`, which is what makes this read as an accident of the rewrite rather than an intended change.

The remaining suggestion is cosmetic and unchanged from last round.

Validation: static review of the seven-file author diff at `ad2f12b4` against merge base `cdd4638b`, isolating it from the main merge that dominates a raw head-to-head comparison, plus the call graph for `_apply_uploaded_speaker` and `adapter.validate` across both request paths and all nine adapters, and the Buildkite step outcomes for builds 14886, 14911 and 14884. I did not execute the branch.

clodaghwalsh17:
@linyueqian the entrypoint testing is failing on this assertion, since I have set the adapter on the diffusion path https://github.com/vllm-project/vllm-omni/blob/d8d4062547899d71b9073649f404e61658136a68/tests/entrypoints/openai_api/test_serving_speech.py#L753

vllm_omni/entrypoints/openai/tts_adapters/omnivoice.py
clodaghwalsh17: @linyueqian I resolved this by calling ```_apply_uploaded_speaker``` to the validator in the adapter so it is applied regardless of AR or diffusion path. Ive set ```has_inline_ref_audio``` before calling the validator in ```_create_diffusion_speech```. I removed the duplicate ```_apply_uploaded_speaker``` a few lines down and left a comment stating that the function is called as part of the validator so the uploaded speaker is called as expected without being overwritten at any point.

vllm_omni/model_executor/models/omnivoice/prompt_utils.py
clodaghwalsh17: New line character added for better formatting

vllm_omni/entrypoints/openai/serving_speech.py
```diff
@@ -336,7 +337,9 @@ def for_diffusion(
         # Diffusion-only instances don't have a TTS stage; set None so any
         # ``_is_tts_model()`` / ``_tts_stage`` access doesn't raise AttributeError.
         instance._tts_stage = None
-        instance._adapter = None
+        # Set adapter to OmniVoice as it is currently the only diffusion TTS model
+        # Temporary assignment until https://github.com/vllm-project/vllm-omni/issues/4327 is completed
+        instance._adapter = OmniVoiceAdapter(SpeechServingContext(server=instance, engine_client=None))
```
linyueqian: [blocking] Constructing the adapter here breaks `test_diffusion_create_speech_with_unknown_voice`, and the test is not in this diff so nothing points at it.

Two assertions in `tests/entrypoints/openai_api/test_serving_speech.py` pin the old behaviour. Line 751 is `assert server._adapter is None` immediately after `for_diffusion`, which cannot hold now that the factory assigns an `OmniVoiceAdapter`. Line 756 is `assert b"Invalid voice" in response.body` for a request with `voice="test-voice"` and no `ref_audio`; that string comes from `_get_normalized_voice`, but `_apply_uploaded_speaker` now runs first inside `validate` and returns `Unknown voice 'test-voice'. OmniVoice has no built-in speakers. Upload a voice first via POST /v1/audio/voices, or use ref_audio + ref_text.`, so `_get_normalized_voice` is never reached.

The request is still a 400 either way, so this is a test-expectation problem rather than a product regression, but it is a hard failure: the test is marked `core_model` and `cpu`, so it runs in per-PR CI. It has not surfaced yet only because no Buildkite lane has run on this head.

Updating both assertions is enough: expect an `OmniVoiceAdapter` instance, and expect the OmniVoice unknown-voice text. If you would rather keep the old message on the diffusion path, moving `request.voice = self._get_normalized_voice(request.voice)` above the `validate` call restores the previous order, but then `has_inline_ref_audio` still has to be captured before `validate` for the reason `base.py` documents.

vllm_omni/model_executor/models/omnivoice/prompt_utils.py
```diff
@@ -0,0 +1,138 @@
...
+    normalised = []
+
+    for raw in raw_items(instruct_str):
+        n = raw.strip().lower()
+        if n in _INSTRUCT_ALL_VALID:
+            normalised.append(n)
+    return normalised
+
+
+def prepare_instruct(instruct_str) -> str:
```
linyueqian, line 110: [important] `prepare_instruct` skips `unify_language`, so the string you validate and the string you send to the model can be in different languages.

`validate_instruction` normalizes and then unifies at line 71 before its conflict check, so validity is decided against a single-language token list. `prepare_instruct` calls only `normalise` and joins, so `"male, 河南话"` reaches the model as `"male，河南话"` where the unified form is `"男，河南话"`. `has_dialect` already forced the Chinese branch during validation, so the two functions disagree about the canonical form of the same accepted input.

Upstream selects the instruction language and translates every mapped item before building the prompt, which is what `unify_language` exists to do; it is defined at line 117 and currently has exactly one caller. This reaches both paths now that the diffusion prompt also goes through `build`, so a valid voice-design request can be sent in a form the model was not trained to read, and the attributes are quietly ignored or misapplied rather than erroring.

Calling `unify_language(normalised)` inside `prepare_instruct` before the separator choice lines the two up. The separator logic can stay as it is, since `has_zh` on the unified list gives the same answer.

vllm_omni/entrypoints/openai/serving_speech.py
linyueqian: [suggestion] `SpeechServingContext` documents that exactly one of `engine_client` and `diffusion_engine` is set, matching the deployment's serving backend. This call sets neither: `engine_client=None` is passed explicitly and `diffusion_engine` is left at its default.

Nothing reads either field on this path, so there is no bug today. `validate` and `build` use only `ctx.server`, and `load_capabilities`, which is what would dereference `engine_client`, is never called on the diffusion instance. It is worth passing `diffusion_engine=instance.diffusion_engine` anyway so the invariant in the docstring stays true, because the next adapter to migrate is likely to trust it.

linyueqian (changes requested):
The round-three blocking finding is fixed, and fixed the right way. `OmniVoiceAdapter.validate` calls `_apply_uploaded_speaker` first, returns its error before anything else runs, and only then checks the instruction, which is exactly the shape every sibling adapter uses. The AR path gets uploaded-speaker resolution and its error reporting back. The round-three cosmetic suggestion is fixed too: the dialect and accent warning now carries a newline before `Dialects`.

Round four did more than that, though, and the new problems are all in the part that grew beyond the ask. Routing `_create_diffusion_speech` through `self._adapter` is a reasonable direction and the `build` output is equivalent to the prompt dict it replaces, but moving `_apply_uploaded_speaker` ahead of `_get_normalized_voice` changes which of the two unknown-voice errors a diffusion client sees, and an existing test pins the old one. That test is not in this diff, so nothing here flags it and no lane has run to catch it.

Worth being explicit about the CI situation, because it is why this needs a push rather than just edits. No Buildkite lane has ever reported on `97733f56`: the `ready` label was already on the branch when you pushed, and a push onto a branch that already carries `ready` fires nothing. The branch is also `CONFLICTING` against main now. Main's own general lane was red from 2026-09-08 to 2026-09-11 on an unrelated realtime event-name break and is green again as of build 15056, so a rebase onto current main is worth doing now in a way it was not last week: it resolves the conflict and gets you a real four-lane result on the fixed code in one push.

Validation: static review of the seven-file author diff at `97733f56` against merge base `cdd4638b`, the round-four delta from `ad2f12b4`, the call graph for `_apply_uploaded_speaker`, `_get_normalized_voice`, `_validate_ref_audio_format` and `build` across both request paths, the `SpeechServingContext` field uses, and the existing assertions in `test_serving_speech.py`. Fork head, so the branch was not executed. A four-model panel reviewed the same head; two found nothing, and the two findings below are theirs, confirmed against the source before being raised.

clodaghwalsh17:
@linyueqian can you reapply the ready label to trigger the tests. BTW I choose to to move the ``` _get_normalised_voice``` call up so that any errors with voice assignment can be caught early and the test can be written in a more generalised way without being tied to the omnivoice model.

tests/model_executor/models/omnivoice/test_validate_instructions.py
```diff
@@ -0,0 +1,50 @@
...
+    def test_non_existent_instruction_english(self) -> None:
+        """Test non existent instruct raises warning"""
+        instructions = "male, english accent"
+        warning = validate_instruction(instructions)
+        assert warning.startswith("Unsupported instruct items found")
+
+    def test_non_existent_instruction_chinese(self) -> None:
+        """Test non existent instruct rejected"""
+        # TODO ask a native speaker to provide and validate this case
+        return
```
lishunyang12: Replace this no-op with an unsupported Chinese instruction and assert the validation error.

clodaghwalsh17: @lishunyang12 done

clodaghwalsh17:
@linyueqian @lishunyang12 can you please approve the changes I made to address your latest comments, thank you

vllm_omni/model_executor/models/omnivoice/prompt_utils.py
```diff
@@ -0,0 +1,138 @@
...
+    normalised = normalise(instruct_str)
+    has_zh = any(any("\u4e00" <= c <= "\u9fff" for c in n) for n in normalised)
+    separator = "，" if has_zh else ", "
+    return separator.join(unify_language(normalised))
```
linyueqian, line 114: [important] The prior language-normalization finding is fixed incompletely: translation now happens after the separator has already been selected. For the accepted input `"男, american accent"`, `has_zh` selects `，`, then `unify_language` translates the items to English, producing `"male，american accent"` instead of the canonical `"male, american accent"`. This affects accepted Chinese attributes combined with an English accent. Both serving paths and the offline example use this helper, and the model prompt builders tokenize the instruction verbatim, so nothing downstream corrects the conditioning format. Assign `normalised = unify_language(normalise(instruct_str))` before computing `has_zh`, and cover both accent-forced English and dialect-forced Chinese preparation in tests.

linyueqian (changes requested):
The author pushed substantive fixes, and the earlier serving-path regressions remain resolved. One normalization fix is incomplete and needs correction before merge.

The previous test-evidence concern still applies: the description cites cdd4638b, which is an ancestor of the current merge-base and contains neither prompt_utils.py nor test_validate_instructions.py. The reported results therefore cannot establish coverage of this PR at the documented commit. Update the tested SHA and associated results. The supplied checks show passing DCO, builds, docs and pre-commit, but provide no current unit-test or E2E lane result.

vllm_omni/model_executor/models/omnivoice/prompt_utils.py
clodaghwalsh17: @linyueqian I reassigned normalised to ```normalised = unify_language(normalise(instruct_str))``` as suggested. I created a series of unit tests for the ```prepare_instruct``` function. I renamed ```tests/model_executor/models/omnivoice/test_validate_instructions.py``` to ```tests/model_executor/models/omnivoice/test_prompt_utils.py``` to hold all the prompt utils related tests. I also rebased and reran all tests against https://github.com/vllm-project/vllm-omni/commit/cb5f508212befb16b48e5760f3b7508c457d6413 and updated the PR description.

clodaghwalsh17:
CI failing for the same reasons as https://github.com/vllm-project/vllm-omni/issues/8038 and https://github.com/vllm-project/vllm-omni/issues/8036

linyueqian (approved):
Round 6 at `bbd198c9`. The last code finding is fixed: `prepare_instruct` now unifies the language before choosing the separator, so a mixed instruction such as `男, american accent` normalises to `male, american accent` and `male, 河南话` to `男，河南话`, and the new `test_prompt_utils.py` cases cover the accent-forces-English and dialect-forces-Chinese paths. The earlier fixes stand (uploaded-speaker metadata captured before adapter validation and applied once, validation failures returned as bad requests on both serving paths). Approving. One note on the description: its test evidence cites `cb5f5082`, which is the current merge-base and predates `prompt_utils.py` and its tests, so please point it at the commit you actually tested. Static read of the delta against merge-base `cb5f5082`; fork head, no PR code executed.

PR #6106, opened 2026-08-12
[Model][IndexTTS] Add continuous batching for IndexTTS 2.5

Sy0307:
## Purpose

IndexTTS-2.5's one-shot S2Mel stage runs a complete CFM trajectory before admitting more work. This adds an opt-in continuous recipe that keeps per-request Euler state and batches compatible active requests between scheduler ticks. The standard deployment YAML remains unchanged.

The continuous path uses model-owned scheduling and runner policy, an explicit zero-token request-ID channel, sparse completed-audio outputs, bounded reference-conditioning reuse, and optional fused BigVGAN activation. Private MPS is supported by the launcher. Async vocoder overlap and singleton admission delay remain experimental and disabled by default.

## Refresh and correctness fixes

- Merge current main through `087ac46`, including vLLM 0.28 configuration/serving changes, without discarding upstream zero-token encoder/KV-transfer or external-launcher DP handling.
- Adapt reference conditioning to the three-value audio resolver and use its exact returned cache key for artifact lookup.
- Fix an H200 first-request crash in the bundled-FA2 fallback environment (without `fa3-fwd`): vLLM's bundled FlashAttention varlen kernels have no Meta implementation. Keep the forward kernel opaque to `torch.compile` through a shared custom-op boundary while preserving eager dispatch and kernel numerics. Dense and packed CUDA regressions reproduce the failure before the fix and pass afterward.
- Prevent the standard non-continuous decoder from accumulating finished request IDs in a cleanup queue it never drains. The new regression fails before the guard and passes afterward.
- Fix CUDA parity test prerequisites: missing `nvcc` skips kernel-build coverage; compilation or numerical failures with a toolchain still fail.
- Refresh test marks, test-helper types, SPDX headers, formatting, runtime documentation, and the H200 recipe environment. Wire the compiled FlashAttention regression into CUDA ready CI.

## Validation

H200, physical GPU 1 in the existing container; Python 3.12.3, vLLM 0.28.0, PyTorch 2.13.0+cu130, Transformers 5.14.1, CUDA toolkit 13.2. Model weights are the cached ModelScope `IndexTeam/IndexTTS-2.5` native bundle, including its auxiliary models. Tests and ASR checks used an otherwise unused GPU separately from inference measurements. The initial validation below used bundled FA2 because the pre-existing environment lacked `fa3-fwd`. The newer origin/main comparison installs the CUDA-required `fa3-fwd==0.0.3` in an isolated overlay and verifies the actual provider. Native main runs successfully in that FA3 environment; the fallback failure must not be generalized to a fully installed main environment.

- Expanded CPU regression suite on the final cleanup fix: **1250 passed, 1 skipped** (7 non-CPU cases deselected).
- Final targeted CUDA tests: **4 passed**, covering fused activation parity/error propagation and compiled dense/packed FlashAttention. The wider FlashAttention suite passed **39 tests**, with one optional-backend skip.
- Real two-stage Chinese/English speech requests, concurrent workloads at 4 and 16 clients, native speed control, and explicit emotion vectors produced finite, non-silent 22.05 kHz mono WAVs.
- Twelve client cancellations followed by a successful recovery request passed in both the continuous path and the experimental async-vocoder + private-MPS path. In the latter, recovery completed in 0.49 s.
- Project-native `vllm bench serve --omni --backend openai-audio-speech` smoke with async vocoder + MPS: **12/12 successful**, streaming PCM, 55.05 s generated audio, reported continuity OK rate 100%. This checks the PCM endpoint; it is not a claim of incremental CFM audio streaming.
- Wheel build passed and the wheel contains the fused `.cpp` and `.cu` sources.
- CI-equivalent full pre-commit passed. Changed-file hooks passed except the existing production mypy backlog: **57 errors on this branch versus 59 in the same two files on clean main**. No ignore/allowlist expansion was used to hide those errors.

### Actual origin/main comparison, with required FA3 installed

Pinned sources: `origin/main@c704aec` versus PR `d72c936`. Main source was not patched. Both use H200 GPU 1, vLLM 0.28.0, Torch 2.13.0+cu130, Transformers 5.14.1, `fa3-fwd==0.0.3`, and DiT compilation enabled; actual dense/varlen provider is `fa3_fwd_interface`. MPS is off throughout. Each concurrency gets 64 warmup requests followed by three runs of 128 requests, with no test/ASR job overlapping measurements. Workload: the same four short texts, reference WAV, and per-text seeds used by the replay harness below.

| Configuration | C=4 mean req/s | C=16 mean req/s |
| --- | ---: | ---: |
| main native recipe, sequence/CFM cap 4 | 4.57 | 6.14 |
| main, sequence/CFM cap aligned to 32 | 4.53 | 7.06 |
| PR, cache/fused optimizations with one-shot generation, cap 32 | 6.38 | 11.17 |
| PR continuous recipe, cap 32 | 5.49 | 12.14 |

With caps matched, the complete PR continuous configuration improves over main by **21.3% at C=4 and 71.9% at C=16**. At C=16, generated-audio throughput is 22.92 versus 40.01 audio-seconds/s and mean P95 is 3.00 versus 1.60 s.

The continuous scheduler's own increment is smaller: relative to the optimized one-shot path within the PR, it changes throughput by **-14.0% at C=4 and +8.6% at C=16**. Cache/fused and other supporting changes already provide most of the gain; the entire main-to-PR gain must not be attributed to continuous scheduling. The low-concurrency tradeoff is why this remains opt-in.

The earlier +18.6% result was a PR-internal stepwise-on/off ablation in the bundled-FA2 fallback environment, **not** an origin/main comparison and not the current FA3 result. It is superseded here as the primary performance evidence.

### Quality and performance scope

The controlled HTTP replay uses four short fixed Chinese/English texts, one repeated reference recording, fixed per-text seeds, identical precision, and matched CFM/sequence caps. It is deliberately a cache-hit workload, not a general TTS or unique-reference performance claim. Cold compilation is excluded from steady measurements and can take tens of seconds for new shapes.

Whisper-large-v3 with beam size 5 was run on samples from the FA3 matrix (16 outputs per configuration/concurrency). English WER was 0 in all sampled groups. Chinese CER at C=4/C=16 was 0%/10.34% for matched-cap main, 5.17%/10.34% for optimized PR one-shot, and 6.90%/10.34% for PR continuous. Errors concentrate on homophones in one short sentence. The C=4 difference also occurs without continuous scheduling and is not yet attributable to a specific optimization. This tiny repeated-text corpus does **not** establish quality equivalence; broader diverse-reference WER/SIM/UTMOS and fixed-code/noise component comparisons remain necessary before making that claim.

A bounded four-Euler-step scheduling quantum was also tried. It improved low-concurrency scheduling overhead but reduced concurrency-16 throughput relative to one-step scheduling, so the experiment and its extra configuration field were removed. No old L20X aggregate speedup is carried forward as evidence for this head. MPS and async-vocoder exploratory timings are not presented as independent performance gains.

## Reproduction

From a Linux CUDA checkout with the repository's runtime/dev and IndexTTS optional dependencies installed:

```bash
HF_HUB_OFFLINE=1 TRANSFORMERS_OFFLINE=1 pytest -q \
  tests/core/sched tests/worker \
  tests/entrypoints/openai_api/test_tts_adapter.py \
  tests/model_executor/models/indextts2 \
  tests/config/test_omni_config.py tests/config/test_config_factory.py \
  tests/engine/test_async_omni_engine_stage_init.py \
  -m 'core_model and cpu' --run-level core_model

CUDA_VISIBLE_DEVICES=0 pytest -q \
  tests/model_executor/models/indextts2/test_fused_alias_free_cuda.py \
  tests/diffusion/attention/test_flash_attn_compile.py \
  -m 'core_model and cuda' --run-level core_model
```

Serve the native ModelScope bundle:

```bash
CUDA_VISIBLE_DEVICES=0 MODEL_VERSION=2.5 MODEL=/path/to/IndexTTS-2.5 \
  INDEXTTS_MPS=0 PORT=8092 \
  DEPLOY_CONFIG=vllm_omni/deploy/indextts2_5_continuous.yaml \
  bash examples/online_serving/text_to_speech/indextts2/run_server.sh
```

For the repository benchmark, expose a reference WAV at `REF_AUDIO_URL` and pass a local Qwen tokenizer from the native bundle:

```bash
MODEL=/path/to/IndexTTS-2.5
REF_AUDIO_URL=http://your-reference-server/reference.wav
VLLM_OMNI_BENCH_AUDIO_SAMPLE_RATE=22050 vllm bench serve --omni \
  --host 127.0.0.1 --port 8092 --model "$MODEL" \
  --tokenizer "$MODEL/qwen0.6bemo4-merge" \
  --backend openai-audio-speech --endpoint /v1/audio/speech \
  --dataset-name seed-tts-text \
  --dataset-path benchmarks/build_dataset/seed_tts_smoke --seed-tts-locale en \
  --num-prompts 12 --num-warmups 2 --max-concurrency 4 --request-rate inf \
  --seed 6106 --extra-body "{\"ref_audio\":\"$REF_AUDIO_URL\",\"extra_params\":{\"lang\":\"en\"},\"seed\":6106}" \
  --percentile-metrics e2el,audio_rtf,audio_ttfp,audio_duration \
  --save-result --result-dir results
```

<details>
<summary>Exact non-streaming WAV replay harness and controlled-run commands</summary>

The WAV harness preserves outputs for paired ASR and checks finite/non-silent audio. The repository benchmark above separately covers the streaming PCM endpoint. Save this as `replay.py`; it requires `requests`, `numpy`, and `soundfile`.

```python
import argparse, base64, concurrent.futures, io, json, pathlib, time
import numpy as np
import requests
import soundfile as sf

p = argparse.ArgumentParser()
p.add_argument("--n", type=int, default=4)
p.add_argument("--concurrency", type=int, default=1)
p.add_argument("--label", default="smoke")
p.add_argument("--port", type=int, default=18106)
p.add_argument("--model", required=True)
p.add_argument("--ref-audio", required=True)
p.add_argument("--output-dir", default="results")
a = p.parse_args()
root = pathlib.Path(a.output_dir) / a.label
root.mkdir(parents=True, exist_ok=True)
ref = pathlib.Path(a.ref_audio).read_bytes()
model = a.model
texts = [
    ("en", "Hello, this is a speech synthesis test."),
    ("zh", "你好，这是语音合成测试。"),
    ("en", "The weather is pleasant today, and we are going for a walk."),
    ("zh", "我们正在验证连续批处理的正确性和稳定性。"),
]


def run(i):
    lang, text = texts[i % len(texts)]
    t = time.perf_counter()
    r = requests.post(
        f"http://127.0.0.1:{a.port}/v1/audio/speech",
        json={
            "model": model,
            "input": text,
            "ref_audio": "data:audio/wav;base64," + base64.b64encode(ref).decode(),
            "response_format": "wav",
            "seed": 1234 + i % len(texts),
            "extra_params": {"lang": lang},
        },
        timeout=600,
    )
    elapsed = time.perf_counter() - t
    result = {
        "i": i,
        "status": r.status_code,
        "elapsed_s": elapsed,
        "lang": lang,
        "text": text,
    }
    if r.status_code == 200:
        audio, sr = sf.read(io.BytesIO(r.content))
        assert audio.size and np.isfinite(audio).all()
        assert sr == 22050
        assert np.max(np.abs(audio)) > 1e-4
        result.update(
            samples=len(audio),
            sample_rate=sr,
            audio_s=len(audio) / sr,
            rms=float(np.sqrt(np.mean(audio**2))),
        )
        (root / f"{i}.wav").write_bytes(r.content)
    else:
        result["error"] = r.text[:1500]
    print(json.dumps(result, ensure_ascii=False), flush=True)
    return result


start = time.perf_counter()
with concurrent.futures.ThreadPoolExecutor(max_workers=a.concurrency) as pool:
    rows = list(pool.map(run, range(a.n)))
wall = time.perf_counter() - start
summary = {
    "label": a.label,
    "n": a.n,
    "concurrency": a.concurrency,
    "wall_s": wall,
    "req_s": a.n / wall,
    "audio_s_s": sum(r.get("audio_s", 0) for r in rows) / wall,
    "p50_s": float(np.percentile([r["elapsed_s"] for r in rows], 50)),
    "p95_s": float(np.percentile([r["elapsed_s"] for r in rows], 95)),
    "rows": rows,
}
(root / "result.json").write_text(json.dumps(summary, ensure_ascii=False, indent=2))
print(
    json.dumps({k: v for k, v in summary.items() if k != "rows"}, ensure_ascii=False),
    flush=True,
)
assert all(r["status"] == 200 for r in rows)
```

Use the tracked `tests/assets/indextts2/ref_audio.wav` reference (SHA256 `e33e6ee0107a1dd58e1d66dd90c13df3d55a8683047cc3d7ea206dad84ed3fc8`). Run the same loop once per server configuration with a distinct `MODE` label (main-native, main-cap32, pr-one-shot, or pr-continuous):

```bash
for C in 4 16; do
  python replay.py --model "$MODEL" --ref-audio tests/assets/indextts2/ref_audio.wav \
    --port 8092 --concurrency "$C" --n 64 --label "$MODE-c$C-warmup"
  for R in 1 2 3; do
    python replay.py --model "$MODEL" --ref-audio tests/assets/indextts2/ref_audio.wav \
      --port 8092 --concurrency "$C" --n 128 --label "$MODE-c$C-r$R"
  done
done
```

For main-native, use main's shipped `indextts2_5.yaml`. For main-cap32, change only both stages' `max_num_seqs` and Stage 1's `s2mel_cfm_batch_size` to 32. For pr-one-shot, copy the PR continuous YAML, set Stage 1's `hf_overrides.stepwise_generation` to `false`, and remove its `worker_cls` override. For pr-continuous, use the PR's shipped continuous YAML. The primary table requires `fa3-fwd==0.0.3` and DiT compilation on both sources. Disable MPS with `INDEXTTS_MPS=0` on both. The tested model/core dependency versions are listed above; use an idle GPU and exclude cold/warmup runs from the summary.

</details>


hsliuustc0106:
This PR touches **vllm_omni/model_executor/, tests/model_executor/, vllm_omni/core/** (30 files). Based on CODEOWNERS coverage of the changed files, the most-related reviewers appear to be:

> @NickCao @gcanlin @linyueqian

Sy0307:
PTAK @linyueqian

tests/model_executor/models/indextts2/test_fused_alias_free_cuda.py
```diff
@@ -0,0 +1,125 @@
...
+        actual = fused(hidden)
+
+    error = (actual.float() - expected.float()).abs()
+    relative_l2 = torch.linalg.vector_norm(error) / torch.linalg.vector_norm(expected.float()).clamp_min(1e-12)
+    cosine = torch.nn.functional.cosine_similarity(
+        actual.float().flatten(),
+        expected.float().flatten(),
+        dim=0,
+    )
+    assert fused.fused_activation_loaded is True
```
linyueqian, line 120: [Minor] This test fails rather than skips when there is no CUDA toolchain available.

The guard above is `@pytest.mark.skipif(not torch.cuda.is_available(), ...)`, but the assertion requires that the JIT extension actually **built**, which additionally needs `nvcc` and a matching toolchain at run time. On my H200 run the GPU is present and `torch.cuda.is_available()` is true, but `nvcc` is not on `PATH`, so the load falls back:

```
INFO  load.py:24 Loading IndexTTS fused BigVGAN activation for sm90
WARNING activation.py:65 Official BigVGAN fused alias-free CUDA extension is unavailable; using eager activation
...
assert fused.fused_activation_loaded is True
E   assert False is True
```

GPU CI containers frequently ship the runtime without the compiler, so this will be red there for a reason unrelated to the code under test, and it will look like a kernel regression.

Two options depending on intent. If the fused path is required, assert that and let it fail, but then also gate the skip on toolchain availability so the failure is honest. If the eager fallback is a supported path, skip when `fused.fused_activation_loaded` is False and keep the numerics comparison for the case where it did load, so the test still verifies the kernel matches eager whenever it is actually built.

linyueqian (commented):
Reviewed `c90420c3`. I ran the PR's own test surface on an H200 (vLLM 0.26.0, torch 2.11+cu130): **178 passed, 1 failed**. No correctness findings in the continuous-batching path from that run.

The one failure is environmental on my box rather than a defect in the change, but it points at a real gap in the test's guard, which is my only substantive comment (inline).

pre-commit is red on formatting only: ruff-format wants the `_GENERATION_SCHEDULER` string in `vllm_omni/model_executor/models/indextts2/pipeline.py` on one line. Mechanical.

Scope note, not a blocker: this adds a vendored CUDA extension (`s2mel/modules/alias_free_cuda/`, ~440 lines of `.cu`/`.cpp`/loader) alongside the continuous-batching work. Those are two independently reviewable changes, and the kernel is the part that most needs its own build and numerics evidence.

linyueqian:
fix pre commit

examples/online_serving/text_to_speech/indextts2/run_server.sh
```diff
@@ -29,6 +32,86 @@ else
...
+        exit 2
+    fi
+    export CUDA_MPS_PIPE_DIRECTORY="$MPS_ROOT/pipe"
+    export CUDA_MPS_LOG_DIRECTORY="$MPS_ROOT/log"
+    mkdir -p -- "$CUDA_MPS_PIPE_DIRECTORY" "$CUDA_MPS_LOG_DIRECTORY"
+    MPS_DIR_CREATED=1
+    chmod 700 "$MPS_ROOT" "$CUDA_MPS_PIPE_DIRECTORY" "$CUDA_MPS_LOG_DIRECTORY"
+    trap cleanup_mps EXIT
+    trap 'exit 130' INT
+    trap 'exit 143' TERM
```
lishunyang12, line 95: Run the server in the background; forward TERM before cleaning up MPS.

PR #6110, opened 2026-08-12
[Bugfix][MUSA][Qwen Image] Avoid complex RoPE alias guards

yeahdongcn:
## Summary

Qwen Image regional compilation on MUSA can fail when TorchDynamo lowers the
complex RoPE property chain (`freqs.real.to(...)` / `freqs.imag.to(...)`) and
creates duplicate tensor alias guards.

The fix adds a small `current_omni_platform.is_musa()` branch inside
`QwenImageCrossAttention.forward`: MUSA uses the equivalent functional
`torch.real` / `torch.imag` form, while the existing `.real/.imag` expressions
remain unchanged for CUDA and other platforms.

## Validation

- Exact image:
  `registry.mthreads.com/mcconline/inference/vllm:v0.24.0-ph1-5.2.0-torch2.11.0.post1-20260813`
  (digest `sha256:71db9e3888b64def9d483780dc1975a6ada5fb46c55d4c5c8553275cc2edcc8c`).
- Runtime: `torch 2.11.0.post1+musa5.2.0`, MUSA `50200`.
- The legacy property path reproduces `Duplicate tensors found` on CPU and
  MUSA.
- The functional MUSA path passes the standalone `torch.compile` probe with
  exact value and dtype parity against the legacy implementation.
- Formatting, import-order, syntax compilation, diff, DCO, Python 3.11/3.12
  build, and pre-commit checks pass.

### DIFF
--- a/vllm_omni/diffusion/models/qwen_image/qwen_image_transformer.py
+++ b/vllm_omni/diffusion/models/qwen_image/qwen_image_transformer.py
@@ -642,10 +642,10 @@ class QwenImageCrossAttention(nn.Module):
         txt_query = self.norm_added_q(txt_query)
         txt_key = self.norm_added_k(txt_key)
 
-        img_cos = vid_freqs.real.to(img_query.dtype)
-        img_sin = vid_freqs.imag.to(img_query.dtype)
-        txt_cos = txt_freqs.real.to(txt_query.dtype)
-        txt_sin = txt_freqs.imag.to(txt_query.dtype)
+        img_cos = torch.real(vid_freqs).to(img_query.dtype)
+        img_sin = torch.imag(vid_freqs).to(img_query.dtype)
+        txt_cos = torch.real(txt_freqs).to(txt_query.dtype)
+        txt_sin = torch.imag(txt_freqs).to(txt_query.dtype)
 
         img_query = self.rope(img_query, img_cos, img_sin)
         img_key = self.rope(img_key, img_cos, img_sin)


fhfuih:
Is it only related to MUSA platform, or triggered with a specific feature toggle?

yeahdongcn:
> Is it only related to MUSA platform, or triggered with a specific feature toggle?

I only have MUSA devices to test. If you want to limit the scope of the impact, I can add an Omni platform check so that this behavior only applies to MUSA. What do you think?

yeahdongcn:
Hi @fhfuih Could you take a look again? Thanks!

fhfuih:
> I only have MUSA devices to test

Please attach a list of steps to reproduce, what to do and what/where to observe, what changes before and after the fix. So we can evaluate if this problem is common on other platforms. I did a quick search and could not find relevant info. And the same code pattern is not unique to Qwen Image series. Many other pipelines also do similar things

yeahdongcn:
> > I only have MUSA devices to test
> 
> Please attach a list of steps to reproduce, what to do and what/where to observe, what changes before and after the fix. So we can evaluate if this problem is common on other platforms. I did a quick search and could not find relevant info. And the same code pattern is not unique to Qwen Image series. Many other pipelines also do similar things

Launching `qwen-image-edit-2511` with default parameters (`--enforce-eager` will work around this on MUSA) is enough to reproduce this issue on MUSA devices. The changes are now scoped to MUSA only, so other platforms will not be affected.

fhfuih (commented):
> Launching qwen-image-edit-2511 with default parameters is enough 

In this case, I can confirm NV GPU platform doesn't trigger such a "Duplicate tensors found" warning/error.

But since your edited version is more universal, I also help tested on GPU platform (H200). The are indeed the same ✅. **So I'd suggest you don't add another `if`, but change it universally (go back to your previous version).**

```
SSIM R:1.000000 (inf) G:1.000000 (inf) B:1.000000 (inf) All:1.000000 (inf)
```

| input | before (`tensor.real.to`) | after (`torch.real(tensor).to`) |
| ----- | ----- | ----- |
| `-F "image=@./dragon.png"` <br /> `-F "prompt='this dragon is wearing sportwear. holding a basketball, and bending one leg.'"` <br /> `-F "size=1024x1024" -F "seed=42"` <br /> `-F "output_format=png"` | Image | Image |

Thanks for keeping following up!

fhfuih:
Also CC @NumberWan @spencerr221  maybe related to Qwen Image t2v and Qwen Image Layered as well

yeahdongcn:
@fhfuih Thanks for testing this on an NVIDIA GPU! I can revert to the previous implementation.

PR #6112, opened 2026-08-12
[Bugfix] Compat: accept legacy str-based Auto* register calls on transformers >= 5.3

KuaaMU:
## Purpose

`vllm serve` crashes at startup (SIGABRT) for models whose **remote** processing files
use the pre-5.3 auto-registration signature, e.g. MiniCPM-o 4.5's
`processing_minicpmo.py`:

```python
AutoImageProcessor.register("MiniCPMVImageProcessor", MiniCPMVImageProcessor)
```

transformers >= 5.3 changed `AutoProcessor` / `AutoImageProcessor` /
`AutoFeatureExtractor` / `AutoTokenizer` `register()` to take a **config class** as the
first argument and dereference `config_class.__module__` internally. A `str` key raises:

```
AttributeError: 'str' object has no attribute '__module__'
```

vLLM imports remote model modules during model-architecture inspection
(`ModelRegistry.inspect_model_cls`, which runs a subprocess `python -m vllm.model_executor.models.registry`),
so the uncaught AttributeError aborts that subprocess and the parent process dies
with SIGABRT (`corrupted size vs. prev_size` glibc abort on some builds) — before the
server ever listens. The official A3 (Ascend NPU) evaluation image ships
transformers 5.13.0, so this breaks out-of-the-box reproduction of MiniCPM-o 4.5
serving.

## Fix

Wrap the four `Auto*` `register()` methods in `vllm_omni/patch.py`:

- every call is forwarded unchanged (new-style `(config_class, cls)`, used by
  vLLM-Omni's own vendored processors — passes through untouched);
- a legacy `(str, cls)` call is only dropped when the installed transformers
  rejects it with the `__module__` AttributeError. The registration is
  redundant: `AutoProcessor.from_pretrained()` resolves the class at load time
  from the checkpoint's `preprocessor_config.json` `auto_map` regardless of the
  registry. On older transformers where str-keyed registration is still
  accepted, behavior is byte-identical to upstream.

## Test Plan

**vLLM Version:** 0.1.dev1+ge5588e49b (vllm-omni v0.25.0 image)

**vLLM-Omni Commit:** `bb0a513d` (minicpm-challenge HEAD)

- New unit tests `tests/model_executor/models/minicpmo_4_5/test_auto_register_compat.py`
  (8 cases: legacy str + new config-class signatures x 4 Auto classes): **8 passed**.
- End-to-end on the official Ascend A3 image (transformers 5.13.0): with the
  **unmodified** checkpoint file (`processing_minicpmo.py` line 408, original
  `AutoImageProcessor.register("MiniCPMVImageProcessor", ...)` call) and only the
  vllm-omni patch applied, `vllm serve` starts cleanly — all 3 stages initialize,
  `/v1/models` returns 200, no SIGABRT / AttributeError in the serve log.

## Test Result

- Unit tests: 8/8 passed (transformers 5.13.0).
- Serve startup (A3 NPU, transformers 5.13.0, original unpatched checkpoint):
  clean start, `/v1/models` 200, 3 stages initialized, zero SIGABRT /
  AttributeError in the serve log.
- Seed-TTS 16 regression (16 requests, same protocol as official harness):
  WER 0.0097 / SIM 0.8468 (reference run: 0.0097 / 0.8467 — no degradation),
  0 failed. No throughput/latency regression observed.
- Regression: new-style config-class registrations (e.g. vendored
  `minicpmo_4_5_omni_llm.py`, `gr00t` processors) pass through unchanged.

Team: **KuaaMU**

### DIFF
--- a/tests/model_executor/models/minicpmo_4_5/test_auto_register_compat.py
+++ b/tests/model_executor/models/minicpmo_4_5/test_auto_register_compat.py
@@ -1,0 +1,66 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+
+"""
+Tests for the transformers 5.x auto-registration compatibility shim in
+vllm_omni.patch.
+
+Remote model files written against the pre-5.3 signature
+``AutoImageProcessor.register(str_name, cls)`` crash on transformers >= 5.3
+with ``AttributeError: 'str' object has no attribute '__module__'``. vLLM
+imports remote model modules during model-architecture inspection (in a
+subprocess), so this crash breaks ``vllm serve`` startup. The shim in
+vllm_omni.patch forwards every call unchanged and drops a legacy str
+registration only when the installed transformers rejects it; new-style
+config-class registrations pass through untouched.
+"""
+
+import pytest
+from transformers import AutoFeatureExtractor, AutoImageProcessor, AutoProcessor, AutoTokenizer
+from transformers.image_processing_utils import BaseImageProcessor
+
+from vllm_omni import patch
+
+pytestmark = [pytest.mark.core_model, pytest.mark.cpu]
+
+
+@pytest.fixture(autouse=True)
+def _apply_compat_shim():
+    patch._patch_auto_register_compat()
+
+
+class _StubImageProcessor(BaseImageProcessor):
+    """Stand-in for a remote image processor class."""
+
+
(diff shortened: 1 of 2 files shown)


KuaaMU:
@yenuo26  @vllm-omni-review-bot

KuaaMU:
Team: **KuaaMU** — pinging @yenuo26 for review when available. Thanks!

PR #6114, opened 2026-08-12
docs: add concise API server endpoint guide

hsliuustc0106:
## Summary

- add a central API Server guide with endpoint selection, startup checks, minimal requests, related routes, and compatibility notes
- split the Serving navigation into Core APIs, Streaming and Realtime, and Standalone Experimental Servers
- add dedicated, implementation-grounded guides for turn-based realtime audio, full duplex, streaming video output, and OpenPI robot policy inference
- document the separate PersonaPlex compatibility server and JoyVL interaction server, including their routes and wire protocols
- rewrite Chat Completions as a general conversational and multimodal reference and move diffusion details into compact compatibility guidance
- update the quickstart and examples to prefer task-specific endpoints and consistent serving commands

## Audited API server coverage

- **1 unified API server** started with `vllm serve <model> --omni`
- **6 core HTTP task families:** Chat Completions, speech, audio generation, image generation, image editing, and videos
- **6 unified WebSocket routes:** `/v1/audio/speech/stream`, `/v1/video/chat/stream`, `/v1/realtime`, `/v1/realtime/video`, `/v1/realtime/robot/openpi`, and `/v1/duplex`
- **2 standalone experimental servers:** PersonaPlex and JoyVL

`/v1/realtime?duplex=1` is a duplex mode on the existing `/v1/realtime` route, not a seventh WebSocket path.

## Important protocol distinctions

- `/v1/realtime` is turn-based JSON audio streaming; full duplex requires an enabled duplex deployment and a model serving adapter.
- `/v1/realtime?duplex=1` falls back to ordinary realtime when no duplex handler exists, while `/v1/duplex` returns an unsupported error. The guide tells clients to verify the capability payload.
- `/v1/realtime/video` combines JSON controls with binary fragmented MP4 frames.
- `/v1/realtime/robot/openpi` uses binary MessagePack with NumPy extensions, not the OpenAI Realtime JSON schema.
- PersonaPlex and JoyVL run as separate processes; their routes are not installed on every unified server.

## Why

API usage was spread across endpoint pages, model examples, and experimental source trees. The old navigation mixed unrelated routes at one level, while Chat Completions foregrounded a diffusion-specific request. Users had no concise way to identify the correct endpoint or distinguish similarly named realtime protocols.

## Validation

- `mkdocs build --strict`
- pre-commit checks on all seven files in this update
- `pytest -q tests/entrypoints/openai_api/test_diffusion_request_utils.py tests/entrypoints/openai_api/test_serving_chat_sampling_params.py` — 64 passed
- focused realtime, video-output, and OpenPI tests — 55 passed, 2 skipped
- focused duplex configuration and protocol tests — 7 passed
- generated HTML exists for all five new serving pages


hsliuustc0106:
This PR touches **docs/serving/, docs/, docs/examples/** (8 files). Based on CODEOWNERS coverage of the changed files, the most-related reviewers appear to be:

> @Gaohan123 @david6666666 @ywang96

docs/serving/openpi_api.md
```diff
@@ -0,0 +1,109 @@
...
+```bash
+pip install openpi-client websockets opencv-python
+
+hf download YangshenDeng/vllm-omni-dreamzero-assets \
+  --repo-type dataset \
+  --local-dir outputs/dreamzero/assets
+
+python examples/online_serving/dreamzero/openpi_client.py \
+  --host 127.0.0.1 \
+  --port 8000 \
```
hsliuustc0106: change every port default to 8091

docs/serving/full_duplex_api.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Native Protocol
+
+`WS /v1/duplex` exposes lower-level lifecycle names including
+`session.create`, `input_audio_buffer.append`, `turn.signal`, `playback.ack`,
+and `session.close`. It returns native session, input, response, overlap, and
+error events. This protocol is experimental and may evolve with the runtime;
+applications should use the provided Realtime client where possible.
+
+See the [MiniCPM-o example](https://github.com/vllm-project/vllm-omni/tree/main/examples/online_serving/minicpmo),
```
hsliuustc0106, line 113: Minor: the MiniCPM-o README this links to still starts the server with `vllm-omni serve` (examples/online_serving/minicpmo/README.md:50), while this page and the serve-command unification in #6221 use `vllm serve`. A one-line sweep of the example READMEs linked from the new serving pages would keep the funnel consistent — fine as a follow-up too.

docs/serving/openpi_api.md
hsliuustc0106: Addressed in 159ac216 — every serving example in the PR now uses 8091 (JoyVL keeps 8092 only for its separate backend process). Verified across all five new pages; safe to resolve.

linyueqian (approved):
Reviewed the full diff at a79dcdee against merge base f41451f6. This is a solid restructuring, and the parts that are easy to get wrong in a docs change of this size are right. I verified the factual claims against the code rather than reading them for plausibility, since a route table that drifts from the server is worse than no route table at all.

Every endpoint named in the new overview resolves to a real registration: the core JSON and multipart routes, the batch routes, the voices routes, the video job lifecycle routes, the sleep and wakeup routes, and all six WebSocket routes. Every one of the thirteen serving pages referenced from `.nav.yml` exists at this head, as does the `sleep_mode.md` target of the cross-link. The duplex paragraph is accurate down to the fallback semantics: `api_server.py` dispatches to the duplex handler only when a handler is configured and the query value matches, and otherwise falls through to the ordinary realtime handler, which is precisely what the page tells clients to expect.

The quickstart rewrite is the change most likely to be pasted verbatim by a new user, so I checked it field by field against `ImageGenerationRequest`. `prompt`, `size` in `WIDTHxHEIGHT` form, and `seed` are all real top-level fields, and the validator accepts exactly the two `response_format` values used across the new pages, `b64_json` and `file`. Moving the example off `/v1/chat/completions` with a nested `extra_body` onto `/v1/images/generations` also makes the quickstart agree with the guidance the rest of this PR gives about preferring task-specific endpoints. Correcting `vllm-omni serve` to `vllm serve` in the audio generation page fixes a command that could not have worked as written.

The one inline note is a small completeness gap rather than an error, and does not need to block merge.

Validation: static review only. I confirmed route registrations, the image request schema, the duplex query-parameter handling, and the existence of every navigation and cross-link target by reading the code and tree at the reviewed head. I did not build or serve the documentation site, so rendering, anchor resolution within pages, and the tabbed `===` blocks are unverified by me.

PR #6115, opened 2026-08-12
[Docs] Classify PD disaggregation as an experimental feature

hsliuustc0106:
## Summary

- move Prefill-Decode disaggregation from **Configuration** to **Features → Runtime and Stage Execution**
- mark the Qwen3-Omni PD page as experimental in the navigation, title, and visible warning
- replace the obsolete legacy `stage_args` launch recipe with a source-grounded description of the current PD design:
  - runtime topology and request lifecycle
  - control, KV data, and downstream model paths
  - topology invariants and validation
  - current implementation and migration boundaries
- redirect the former Configuration URL to the new Features URL
- align the Configuration and Features landing pages with the new taxonomy

## Why

The page was exposed under Configuration because it lived in `docs/configuration`, whose navigation uses a wildcard. PD is a runtime and stage-execution feature rather than a general configuration concept.

Source validation also showed that the old page advertised a legacy `stage_args` YAML to `vllm serve`, while the current serve CLI accepts the deploy-based `stages:` schema and the PD deploy overlay is not yet migrated. Presenting the page as an experimental design reference is more accurate than retaining a non-runnable recipe.

## User impact

Users will find PD with other runtime features, see its experimental status before relying on it, and get an accurate view of what is implemented versus still under development. Existing links to the former page continue to work through a docs redirect.

## Validation

- `pre-commit run --files mkdocs.yml docs/.nav.yml docs/configuration/README.md docs/features/README.md docs/features/pd_disaggregation.md`
- `mkdocs build --strict`
- inspected the generated navigation to confirm PD appears once under **Runtime and Stage Execution**
- inspected the generated redirect from `configuration/pd_disaggregation/` to `features/pd_disaggregation/`
- checked the documented runtime contracts against `pd_utils.py`, `async_omni_engine.py`, `orchestrator.py`, the Qwen3 stage processor, and the temporarily disabled PD tests
- completed the InferMatrixCopilot Direct review gate without findings


docs/configuration/pd_disaggregation.md
```diff
@@ -1,4 +1,8 @@
-# Prefill-Decode (PD) Disaggregation
```
hsliuustc0106, line 1: move the md from configuration to feature design

PR #6117, opened 2026-08-12
[minicpmo] Add opt-in streaming audio input for native duplex

Fyrgo8:
Part of #5069

## Summary

- Wire `streaming_audio_chunk_size` from the MiniCPM-o config into the existing native full-duplex session path.
- Keep the released-model default at 50 frames / 1000 ms; a 25-frame probe configuration advertises and schedules 500 ms input units.
- Preserve append ordering, residual padding, per-session Stage 0 state, and model-owned `listen` handling without adding a serving path.
- Measure each audio-encoder invocation with CUDA events and expose the positive scalar latency as `response.metadata.vllm_omni.stage_metrics["0"].audio_encoder_latency_ms` without leaking the internal tensor payload.

## Live acceptance

Both probes used the native duplex Realtime endpoint, two 600 ms PCM16 appends, `commit(final=true)`, a post-tail heartbeat, and a final `/health` check.

| Configuration | Chunk period | Listen units | Stage 0 TTFT (ms) | Encoder latency (ms) | GPU0 peak | GPU1 peak |
| --- | ---: | ---: | --- | --- | ---: | ---: |
| 25-frame probe | 500 ms | 3 | 554.51, 579.63, 604.81 | 25.879, 10.154, 7.527 | 26183 MiB | 22574 MiB |
| Released default | 1000 ms | 2 | 594.68, 616.92 | 29.803, 10.048 | 26185 MiB | 22574 MiB |

- Every `response.listen` contained a positive `audio_encoder_latency_ms` under Stage 0 metrics.
- No client event contained the internal `meta.audio_encoder_latency_ms` tensor key.
- Heartbeat acknowledgement and `/health=200` passed after both probes.
- Both probes started from 26181 MiB on GPU0 and 22574 MiB on GPU1; request-window peaks were +2/+0 MiB for 500 ms and +4/+0 MiB for 1000 ms.
- Server logs contained no traceback, engine-dead, OOM, or error entries during either acceptance run.

## Regression coverage

- Full duplex handler regression: `204 passed, 16 warnings`.
- Earlier focused CUDA suite: `307 passed, 16 warnings`.
- Ruff and format checks pass for the final core-code delta.
- DCO passes. The current Read the Docs status fails before project installation: its PyTorch CPU-index bootstrap cannot resolve `flit_core` while building `typing-extensions`; no PR code or docs build command is reached.

Tests, probe scripts, machine-specific logs, and branch-sync archives are intentionally excluded from this PR.

### DIFF
--- a/vllm_omni/experimental/fullduplex/openai/serving.py
+++ b/vllm_omni/experimental/fullduplex/openai/serving.py
@@ -914,6 +914,7 @@ class OmniDuplexSessionHandler(
             session.replace_capabilities(
                 self._serving_runtime_adapter.capabilities(
                     max_sessions=self._duplex_session_config.max_sessions,
+                    runtime_config=runtime_config,
                 )
             )
             session.replace_runtime_config(runtime_config)
--- a/vllm_omni/experimental/fullduplex/openai/serving.py
+++ b/vllm_omni/experimental/fullduplex/openai/serving.py
@@ -44,6 +44,7 @@ from vllm_omni.experimental.fullduplex.openai.runtime_adapter import (
     ServingRuntimeConfigError,
     ServingRuntimeSessionState,
     load_serving_runtime_adapter,
+    runtime_capabilities,
     validate_serving_runtime_adapter,
 )
 from vllm_omni.experimental.fullduplex.openai.runtime_bridge import (
@@ -912,7 +913,8 @@ class OmniDuplexSessionHandler(
         session = self._registry.create(config=config, session_id=session_id)
         if use_native_runtime:
             session.replace_capabilities(
-                self._serving_runtime_adapter.capabilities(
+                runtime_capabilities(
+                    self._serving_runtime_adapter,
                     max_sessions=self._duplex_session_config.max_sessions,
                     runtime_config=runtime_config,
                 )
--- a/vllm_omni/model_executor/models/minicpmo_4_5/minicpmo_4_5_omni.py
+++ b/vllm_omni/model_executor/models/minicpmo_4_5/minicpmo_4_5_omni.py
@@ -26,6 +26,7 @@ import torch
 import torch.nn as nn
 import torch.nn.functional as F
 from vllm.config import VllmConfig
+from vllm.distributed.parallel_state import get_tp_group
 from vllm.logger import init_logger
 from vllm.model_executor.models.interfaces import SupportsMRoPE, SupportsMultiModal, SupportsPP
 from vllm.model_executor.models.utils import init_vllm_registered_model, maybe_prefix
@@ -641,7 +642,8 @@ class MiniCPMO45OmniForConditionalGeneration(nn.Module, SupportsMultiModal, Supp
             return None
 
         timing_rows = [pending.pop(str(req_id), None) if isinstance(pending, dict) else None for req_id in req_ids]
-        if not any(timing_rows):
+        has_timing_rows = [timings is not None for timings in timing_rows]
+        if not any(has_timing_rows):
             return None
 
         # Synchronizing the last encoder event waits only for the measured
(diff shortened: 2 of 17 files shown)


hsliuustc0106:
This PR touches **vllm_omni/experimental/, vllm_omni/, vllm_omni/engine/** (14 files). Based on CODEOWNERS coverage of the changed files, the most-related reviewers appear to be:

> @Gaohan123 @hsliuustc0106 @ywang96

Fyrgo8:
@Gaohan123 @hsliuustc0106 @ywang96 Could one of you take a look when you get a chance? Thanks!

Sy0307:
This branch is currently conflicting and 202 commits behind `main`, with no added regression test for the new frame/window configuration. Before merge, please rebase and add a production-path test covering the configured 500 ms/1000 ms boundaries, including the short final tail, then rerun the native duplex E2E.

hsliuustc0106:
Please fix pre-commit; `ruff-format` modifies `stage_engine_core_client.py` and `gpu_ar_model_runner.py`.

vllm_omni/experimental/fullduplex/openai/serving.py
```diff
@@ -914,6 +914,7 @@ async def _open_session(
             session.replace_capabilities(
                 self._serving_runtime_adapter.capabilities(
                     max_sessions=self._duplex_session_config.max_sessions,
+                    runtime_config=runtime_config,
```
hsliuustc0106, line 919: This unconditionally changes the generic adapter call contract. Adapters injected via `serving_runtime_adapter` or loaded by dotted path are only checked for callability, so an existing `capabilities(*, max_sessions)` implementation now raises an uncaught `TypeError` during `session.create`. Please preserve the old call path and expose runtime-derived capabilities through a signature-checked/versioned optional hook, with an old-signature handshake regression.

vllm_omni/model_executor/models/minicpmo_4_5/minicpmo_4_5_omni.py
```diff
@@ -625,6 +635,31 @@ def forward(
...
+        for timings in timing_rows:
+            if timings:
+                timings[-1][1].synchronize()
+
+        latency_rows: list[torch.Tensor | None] = []
+        for timings in timing_rows:
+            if not timings:
+                latency_rows.append(None)
+                continue
+            latency_ms = sum(float(start.elapsed_time(end)) for start, end in timings)
```
hsliuustc0106: This sums only the local worker’s events. With the supported TP=4 Stage 0 config, the executor exposes TP rank 0’s output, so a slower peer is invisible and `audio_encoder_latency_ms` under-reports the encoder critical path. Please MAX-reduce the summed latency across the TP group before attaching it (with every rank participating), or explicitly define and name this as a rank-0-only metric.

Fyrgo8:
@Sy0307 Addressed in the latest update: rebased onto `main`, added the 500/1000 ms boundary and short-tail regression test, and reran the native duplex E2E successfully. Please take another look.

vllm_omni/model_executor/models/minicpmo_4_5/minicpmo_4_5_omni.py
Fyrgo8: @hsliuustc0106 Thanks. The generic adapter contract issue is addressed by preserving the legacy capabilities(max_sessions) call path and adding a signature-checked, versioned optional hook with a legacy handshake regression test.\n\nFor the TP latency fix, I validated the MAX reduction with TP=2, with both ranks participating and consistent reduced latency. Would TP2 validation be sufficient here, or would you prefer a real TP4 Stage 0 run? I can add the TP4 validation if needed.

PR #6121, opened 2026-08-12
[Bugfix] Fix diffusion TTS adapter lookup in speech serving

### LINKED ISSUE #6118: [Bug] Diffusion-path TTS speech requests 500 with "'OmniOpenAIServingSpeech' object has no attribute '_adapter'"
### What happens

On current main, any speech request to a diffusion-path TTS model returns HTTP 500:

```
Speech generation failed: 'OmniOpenAIServingSpeech' object has no attribute '_adapter'
```

Reproduced with OmniVoice (`vllm-omni serve k2-fsa/OmniVoice --omni --deploy-config vllm_omni/deploy/omnivoice.yaml`, vLLM 0.26.0). The server starts cleanly and fails on the first request, every time.

### Traceback

```
File "vllm_omni/entrypoints/openai/serving_speech.py", line 3655, in _create_diffusion_speech
    speed=self._audio_encode_speed(request),
File "vllm_omni/entrypoints/openai/serving_speech.py", line 524, in _audio_encode_speed
    if self._uses_native_speed_control():
File "vllm_omni/entrypoints/openai/serving_speech.py", line 520, in _uses_native_speed_control
    ...  -> self._get_tts_adapter() -> self._adapter
```

### What's confusing about it

`self._adapter = None` at `serving_speech.py:494` is unconditional inside `__init__` (starts at 424), there's no early `return`/`raise` between them, no class-level `_adapter` default, only one construction site (`api_server.py:1125`) and no subclass — so on the face of it the attribute should always exist by the time a request lands. Someone who knows this path will spot the construction/ordering issue faster than I will; the recent commits touching this file are #4795 (forced-aligner timestamps, adds the `forced_aligner_enabled` kwarg to that constructor) and #5957 (IndexTTS 2.5), on top of the adapter framework from #4330.

Regardless of the cause, `_get_tts_adapter()` reading `self._adapter` before it is guaranteed to exist is worth hardening — a class-level `_adapter: TTSModelAdapter | None = None`, or `getattr(self, "_adapter", None)` at the read site, would turn this 500 into correct behaviour.

### Not a regression from any open PR

I hit this while re-testing #5174 and checked the control: main at `38a2500b` **without** that PR fails identically, and the PR only touches `omnivoice_generator.py` / `configs/omnivoice.py`. Also worth noting OmniVoice served fine on this same box earlier today (main around `ee118ab4`), so this landed within the last few hours.

HaningZS:
## Purpose

Fixes #6118.

### What broke

Speech requests to pure-diffusion TTS models such as OmniVoice returned HTTP 500:

```text
Speech generation failed: 'OmniOpenAIServingSpeech' object has no attribute '_adapter'
```

### Reproduction

Start an OmniVoice server:

```bash
vllm serve k2-fsa/OmniVoice --omni --port 8091 --trust-remote-code
```

Then send a text-only speech request:

```bash
python examples/online_serving/text_to_speech/omnivoice/speech_client.py \
  --api-base http://localhost:8091/v1 \
  --text "Hello, how are you?"
```

### Root cause

`OmniOpenAIServingSpeech.for_diffusion()` constructs the handler with `__new__`
and intentionally bypasses the regular initializer. The diffusion speech encoding
path recently started checking adapter-native speed control, which attempted to
resolve an AR-stage TTS adapter and read `_adapter` on this specialized instance.

### Fix

Keep pure-diffusion speech on its dedicated request path by returning no AR-stage
adapter before adapter resolution or AR engine state is accessed. Normal AR TTS
instances retain their existing adapter and native-speed behavior.

## Test Plan

- Regress diffusion speech speed handling at the adapter boundary.
- Exercise the public `create_speech()` handler path and assert HTTP 200,
  `audio/wav`, and the generated response body.
- Run the complete speech serving L1 CPU suite.
- Run the TTS adapter and model-detection L1 CPU suites.
- Run diff-scoped pre-submit checks and Ruff.

**vLLM Version:** `0.27.0`

**vLLM-Omni Commit:** `eb4b37922a7f33b20bbd6dbd1d204b5fb0b9864b`

Test environment: Python 3.12, PyTorch 2.12.1+cu126,
Transformers 5.15.0. Tests were CPU-only and did not allocate a GPU.

## Test Result

```text
tests/entrypoints/openai_api/test_serving_speech.py
214 passed, 15 warnings in 5.34s

132 passed, 15 warnings in 3.05s

Focused #6118 regression tests
2 passed, 15 warnings in 3.04s

Ruff check: passed
Ruff format --check: passed
check_test_marks.py: passed
check_tts_adapter.py: passed
check_pickle_imports.py: passed
git diff --check: passed
```

A real-weight OmniVoice GPU E2E was not run. The failure occurs before model
inference in the speech request-routing/adapter boundary, and the regression is
covered deterministically by the L1 CPU tests above.

### DIFF
--- a/tests/entrypoints/openai_api/test_serving_speech.py
+++ b/tests/entrypoints/openai_api/test_serving_speech.py
@@ -737,6 +737,19 @@ class TestTTSMethods:
         yield server
         server.shutdown()
 
+    def test_diffusion_audio_encode_speed_skips_ar_adapter(self, mocker: MockerFixture):
+        """Pure-diffusion speech keeps generic speed handling outside AR adapters."""
+        server = OmniOpenAIServingSpeech.for_diffusion(
+            diffusion_engine=mocker.MagicMock(),
+            model_name="test-model",
+        )
+        resolve_adapter = mocker.patch.object(serving_speech_module, "resolve_adapter")
+
+        speed = server._audio_encode_speed(OpenAICreateSpeechRequest(input="Hello", speed=1.25))
+
+        assert speed == 1.25
+        resolve_adapter.assert_not_called()
+
     def test_is_tts_detection_no_stage(self, speech_server):
         """Test TTS model detection when no TTS stage exists."""
         # Fixture creates server with stage_configs = [] -> _is_tts should be False
--- a/tests/entrypoints/openai_api/test_serving_speech.py
+++ b/tests/entrypoints/openai_api/test_serving_speech.py
@@ -680,7 +680,7 @@ class TestSpeechAPI:
 
     @pytest.mark.asyncio
     async def test_create_diffusion_speech_extra_params(self, mocker: MockerFixture):
-        """Test that extra_params are correctly applied to sampling_params_list in diffusion mode."""
+        """Test public diffusion speech success and extra_params propagation."""
         # Mock the engine client
         mock_engine = mocker.MagicMock()
 
@@ -704,7 +704,11 @@ class TestSpeechAPI:
 
         req = OpenAICreateSpeechRequest(input="Hello", extra_params={"new_arg": 123, "existing_arg": "new_value"})
 
-        await server._create_diffusion_speech(req)
+        response = await server.create_speech(req)
+
+        assert response.status_code == 200
+        assert response.media_type == "audio/wav"
+        assert response.body == b"dummy"
 
         # Verify generate was called
         mock_engine.generate.assert_called_once()
--- a/vllm_omni/entrypoints/openai/serving_speech.py
+++ b/vllm_omni/entrypoints/openai/serving_speech.py
@@ -502,11 +502,17 @@ class OmniOpenAIServingSpeech(OpenAIServing, AudioMixin):
     def _get_tts_adapter(self):
         """Return the per-model serving adapter for the current ``_tts_model_type``.
 
+        Pure-diffusion speech uses its dedicated request path and does not
+        resolve adapters for AR-stage TTS models.
+
         Resolved lazily (rebuilt if ``_tts_model_type`` changed since the cached
         instance was built) so callers that set ``_tts_model_type`` after
         construction still dispatch to the matching adapter. In production
         ``_tts_model_type`` is fixed at init, so the cached instance is reused.
         """
+        if self._diffusion_mode:
+            return None
+
         adapter_cls = resolve_adapter(self._tts_model_type)
         if adapter_cls is None:
             self._adapter = None


NickCao (approved):
LGTM, special casing of _diffusion_mode matches how we handle this in other parts of serving_speech.py, and good test coverage.

NickCao:
Tested `pytest -sv tests/e2e/online_serving/test_omnivoice_expansion.py tests/e2e/offline_inference/test_omnivoice_expansion.py -m "slow and L4 and tts" --run-level "full_model"`
```
FAILED tests/e2e/online_serving/test_omnivoice_expansion.py::TestOmniVoiceVoiceCloning::test_voice_clone_ref_audio_only[omni_server0] - AssertionError: Transcript doesn't match input: similarity=0.80, transcript=' It's nice today. The weather for a walk in the park.'

1 failed, 5 passed, 14 warnings in 138.42s (0:02:18)
```

Before this PR:
```
FAILED tests/e2e/online_serving/test_omnivoice_expansion.py::TestOmniVoiceTTS::test_speech_auto_voice[omni_server0] - openai.InternalServerError: Error code: 500 - {'error': {'message': "Speech generation failed: 'OmniOpenAIServingSpeech' object has no attribute '_adapter'", 'type': 'server_error', 'param': None, 'code': 500}}
FAILED tests/e2e/online_serving/test_omnivoice_expansion.py::TestOmniVoiceSeed::test_speech_auto_voice_seed_deterministic[omni_server0] - openai.InternalServerError: Error code: 500 - {'error': {'message': "Speech generation failed: 'OmniOpenAIServingSpeech' object has no attribute '_adapter'", 'type': 'server_error', 'param': None, 'code': 500}}
FAILED tests/e2e/online_serving/test_omnivoice_expansion.py::TestOmniVoiceSeed::test_speech_auto_voice_seed_non_deterministic[omni_server0] - openai.InternalServerError: Error code: 500 - {'error': {'message': "Speech generation failed: 'OmniOpenAIServingSpeech' object has no attribute '_adapter'", 'type': 'server_error', 'param': None, 'code': 500}}
FAILED tests/e2e/online_serving/test_omnivoice_expansion.py::TestOmniVoiceVoiceCloning::test_voice_clone_ref_audio_only[omni_server0] - openai.InternalServerError: Error code: 500 - {'error': {'message': "Speech generation failed: 'OmniOpenAIServingSpeech' object has no attribute '_adapter'", 'type': 'server_error', 'param': None, 'code': 500}}
FAILED tests/e2e/online_serving/test_omnivoice_expansion.py::TestOmniVoiceVoiceCloning::test_voice_clone_ref_audio_and_text[omni_server0] - openai.InternalServerError: Error code: 500 - {'error': {'message': "Speech generation failed: 'OmniOpenAIServingSpeech' object has no attribute '_adapter'", 'type': 'server_error', 'param': None, 'code': 500}}

5 failed, 1 passed, 14 warnings in 59.71s
```

NickCao:
`tests/e2e/online_serving/test_omnivoice_expansion.py::TestOmniVoiceVoiceCloning::test_voice_clone_ref_audio_only` is flaky, passes on the second and third run.

linyueqian:
Late post-merge look since I filed #6118 — the fix is right, and thanks for the fast turnaround. `for_diffusion()` building via `__new__` was the piece I couldn't see from outside; I'd verified `_adapter = None` was unconditional in `__init__` and found only one `OmniOpenAIServingSpeech(` call site, so an alternate `__new__` constructor was invisible to me.

Three things I checked before saying it's good:

- The guard reads `self._diffusion_mode`, which is a class-level attribute, so it exists on both construction paths. Worth noting because a guard on any `__init__`-only attribute would have reproduced the exact same `AttributeError` one line earlier.
- Returning `None` matches what `_get_tts_adapter()` already did when `resolve_adapter` came back empty, so callers were already `None`-safe.
- No second landmine on that path: `for_diffusion()` sets 7 attributes where `__init__` sets 20, but none of the remaining 16 (`model_name`, `supported_languages`, `precomputed_speakers`, the various tokenizers, …) are read by `_create_diffusion_speech` / `_audio_encode_speed` / `_uses_native_speed_control`. So nothing else is currently reachable.

One non-blocking thought for whenever this area is touched again: that last point holds today but isn't enforced by anything. Any new read of an `__init__`-only attribute from a diffusion-reachable path reintroduces the same failure, and it'll surface as a runtime 500 rather than at construction. The class already shows the durable pattern with `_diffusion_mode: bool = False` and `_tts_executor: ... = None` — promoting `_adapter` (and ideally the rest of that set) to class-level defaults would make the whole family impossible instead of guarding call sites one at a time.

Also a genuine improvement worth calling out: switching the test from `_create_diffusion_speech` to the public `create_speech()` is exactly why this escaped in the first place — the private-method test never traversed the code that touched `_adapter`.

PR #6122, opened 2026-08-12
[Bugfix] Respect media redirect policy for image references

### LINKED ISSUE #6036: [Bug]: image_reference.image_url rejects any redirecting URL — httpx client missing follow_redirects=True
### Your current environment

vLLM-Omni `v0.22.0` (container `vllm/vllm-omni:v0.22.0`), serving `Wan-AI/Wan2.1-VACE-14B-diffusers`. The same code is present on `main` today.

### Describe the bug

`image_reference={"image_url": "..."}` on `POST /v1/videos` fails for any URL that responds with a redirect, and reports it as a download failure.

`decode_image_url` in `vllm_omni/entrypoints/openai/video_api_utils.py` builds its client without `follow_redirects=True`:

```python
if image_url.startswith(("http://", "https://")):
    async with httpx.AsyncClient(timeout=60) as client:
        try:
            response = await client.get(image_url)
            response.raise_for_status()
        except httpx.HTTPError as exc:
            raise InvalidInputReferenceError(
                "Invalid image_reference.image_url: failed to download image."
            ) from exc
```

httpx does not follow redirects by default, and `raise_for_status()` raises on a 3xx that was not followed. So a perfectly valid image URL is rejected with:

```
Invalid image_reference.image_url: failed to download image.
```

The message points at connectivity, which sends you looking in the wrong place — the host is reachable and the image is fine. Most CDN-backed URLs redirect, including Hugging Face resolve links, `raw.githubusercontent.com` and many presigned object-store URLs, so in practice this rejects a large share of real inputs.

Verified inside the running container, using a Hugging Face image URL:

```
no follow_redirects:   status=302 bytes=1016
raise_for_status:      HTTPStatusError: Redirect response '302 Found' for url '...'
WITH follow_redirects: status=200 bytes=1049491
```

And through vLLM-Omni's own `decode_image_url`:

```
redirecting URL         FAIL  Invalid image_reference.image_url: failed to download image.
direct (resolved) URL   OK    (1024, 704)
data: URL (base64)      OK    (1024, 704)
```

So the decode path is sound; only redirect handling is missing.

### Suggested fix

```diff
-        async with httpx.AsyncClient(timeout=60) as client:
+        async with httpx.AsyncClient(timeout=60, follow_redirects=True) as client:
```

It would also help to distinguish the failure modes in the error message — a redirect that was not followed, a non-2xx response, and a genuine connection error currently collapse into one string.

If following redirects is a deliberate SSRF precaution, that is worth stating in the message and the docs, since the current text gives no hint that a redirect was the cause.

### Workarounds

- upload the image as `input_reference` instead of referencing it by URL
- resolve the redirect client-side and pass the final URL
- inline the image as `image_reference={"image_url": "data:image/png;base64,<...>"}`

- [x] I have searched for similar issues.

HaningZS:
## Purpose

Closes #6036.

### What broke

`POST /v1/videos` rejected a valid `image_reference.image_url` whenever the
server returned a redirect. The public error reported a generic download
failure even though the host was reachable.

### Reproduction

The added deterministic HTTPX transport tests exercise the same 302 response
without external network access:

```bash
pytest -q tests/entrypoints/openai_api/test_video_api_utils.py -k decode_image_url
```

With only those tests applied to upstream `5358f5db`, the result was `3 failed,
2 passed`: allowed redirects, disabled-redirect diagnostics, and HTTP status
diagnostics failed. The connection-failure and data-URL contracts passed.

### Root cause

`decode_image_url()` created `httpx.AsyncClient` without `follow_redirects`.
HTTPX therefore returned the initial 3xx response, and `raise_for_status()`
mapped it to the same generic error as genuine network failures.

### Fix

- Pass vLLM's `VLLM_MEDIA_URL_ALLOW_REDIRECTS` value to HTTPX.
- When redirects are disabled, reject the 3xx before requesting its target and
  name the controlling setting in the error.
- Distinguish final non-success HTTP statuses from network/request failures.
- Preserve the existing local data-URL path.
- Document the setting, default, and cross-host redirect security boundary.

### Security behavior

This does not bypass the existing operator control. Redirects are followed only
when `VLLM_MEDIA_URL_ALLOW_REDIRECTS` allows them; the disabled-policy test also
asserts that the redirect target is never requested. The documentation notes
that redirects can cross hosts and that vLLM-Omni does not yet fully implement
upstream vLLM's media URL allowlist protection.

## Test Plan

The change is an L1 CPU request-boundary fix. GPU/model inference would not
exercise redirect policy, so no GPU was allocated.

**vLLM Version:** 0.27.0

**vLLM-Omni Commit:** `5358f5db01e512dc20fcc29a1d122e82c8546780`

**Environment:** Python 3.12.3, PyTorch 2.12.1+cu126, HTTPX 0.28.1

## Test Result

```text
# Focused redirect, policy, error, and data-URL contracts
5 passed, 13 deselected

# Public endpoint plus all video API CPU regression files
204 passed

# Static checks
ruff check: passed
ruff format --check: 3 files already formatted
pytest marker check: passed
git diff --check: passed
```

The public endpoint regression sends a redirecting URL through
`POST /v1/videos`, asserts HTTP 200 and task completion, and verifies the engine
receives the decoded image.


HaningZS:
@vllm-omni-review-bot please review the VLLM_MEDIA_URL_ALLOW_REDIRECTS behavior, request-scoped error mapping, and regression coverage for both the helper and public /v1/videos path.

hsliuustc0106 (approved):

vllm_omni/entrypoints/openai/video_api_utils.py
```diff
@@ -172,11 +173,21 @@ async def decode_image_url(image_url: str) -> Image.Image:
         return _decode_base64_image(image_url, source="image_reference.image_url")
 
     if image_url.startswith(("http://", "https://")):
-        async with httpx.AsyncClient(timeout=60) as client:
+        allow_redirects = envs.VLLM_MEDIA_URL_ALLOW_REDIRECTS
+        async with httpx.AsyncClient(timeout=60, follow_redirects=allow_redirects) as client:
```
hsliuustc0106, line 177: The redirect policy doesn't reach the sibling fetchers on the same `/v1/videos` route: `decode_video_url` (line 252) and `decode_audio_url` (line 299) below still construct `httpx.AsyncClient(timeout=60)` without `follow_redirects`, so a redirecting `video_reference.video_url` or `audio_reference.audio_url` (api_server.py:3439) keeps failing with the same generic "failed to download" error #6036 reports. `_load_input_images` for the image-edit endpoint (api_server.py:2681) has the same construction. The fix is one line per site — any reason not to cover them here, or should the rest be tracked in a follow-up issue?

PR #6124, opened 2026-08-12
[Doc] Add recipe for Cosmos3 video generation pinned to start and end boundaries

weishao-nv:
Add an inference-only Cosmos3 recipe and client for generating a video pinned to start and end boundaries (frame or short clip), with the middle filled from an event prompt (FL2V).

## Purpose

Stock Cosmos3 V2V locks a clean prefix of latents (typically `[0, 1]`). This PR documents and ships a client path that locks a contiguous latent range at **both** ends via existing request fields (`condition_frame_indexes_vision` + a clean head/tail reference video). No vLLM-Omni source or weight changes.

Includes:
- `recipes/cosmos3/Cosmos3-FL2V.md` — setup, serve/run, how it works, troubleshooting
- `tools/cosmos3_fl2v/` — HTTP client, required SHM patch for large FL2V payloads, demo seeds/prompts
- `.gitignore` allow-list for the demo prompt JSON files

Prefer `nvidia/Cosmos3-Super` for quality; `nvidia/Cosmos3-Nano` also works on the same client (verified).

## Test Plan

**vLLM Version:** 0.26.0

**vLLM-Omni Commit:** this PR branch (`add-cosmos3-fl2v`)

- Apply `tools/cosmos3_fl2v/patch_vllm_shm.py` in the server env
- Serve `nvidia/Cosmos3-Super` (or Nano) with `vllm serve … --omni`
- Run `python fl2v_generate_vllm.py --dry-run` (client-only sanity)
- Run frame+frame demo against a warm server; confirm start/end PSNR and output MP4

## Test Result

- Dry-run: `latent_frames=48`, conditioning latents `[0, 47]`, reference frames `189`
- Nano on H100 and Super on B200/H200: generation completed; boundaries pinned
- Without the SHM patch, large FL2V payloads hang silently (as documented)


weishao-nv:
Updated this branch with `main` (which picks up the Read the Docs fix from #6129) and pushed a follow-up commit to satisfy `pre-commit`:

- Applied `ruff format` to `tools/cosmos3_fl2v/fl2v_generate_vllm.py` and `tools/cosmos3_fl2v/patch_vllm_shm.py`.
- Reworded two strings that the `typos` hook rewrote incorrectly (`PNGs` -> `ONGs` in the recipe, and `LOD` -> `LOAD` inside a demo negative prompt). The prompt now spells out "level-of-detail", which keeps the original meaning and avoids needing a `typos` allow-list.

All reported checks are green now: `pre-commit`, `build (3.11)`, `build (3.12)`, `DCO`, and `docs/readthedocs.org`. Docs preview: https://vllm--6124.org.readthedocs.build/projects/vllm-omni/en/6124/

`buildkite/vllm-omni` is a required check that still shows "Expected — waiting for status to be reported". Could a maintainer approve the pending workflow runs so it can start? Happy to run L1/L2 locally and attach the results, or do anything else needed for the `ready` label.

For context, this change is inference-only and touches no vLLM-Omni source or model weights: it is a recipe plus a client that locks a contiguous latent range at both ends using existing request fields.

tools/cosmos3_fl2v/fl2v_generate_vllm.py
```diff
@@ -0,0 +1,527 @@
...
+        "model": args.model,
+        "prompt": prompt,
+        "negative_prompt": negative,
+        "size": f"{args.width}x{args.height}",
+        "num_frames": str(args.num_frames),
+        "fps": str(args.fps),
+        "num_inference_steps": str(args.steps),
+        "guidance_scale": str(args.guidance),
+        "flow_shift": str(args.flow_shift),
+        "max_sequence_length": str(args.max_sequence_length),
```
lishunyang12: Send max_sequence_length inside extra_params; the multipart parser ignores this field.

tools/cosmos3_fl2v/patch_vllm_shm.py
```diff
@@ -0,0 +1,155 @@
...
+        return 0
+    if ORIGINAL not in text:
+        raise SystemExit(
+            f"could not find the code to patch in {target}\n"
+            "This vLLM version differs from the one this fix targets (0.26.0).\n"
+            "Check MessageQueue.dequeue(): the overflow branch must not pass the\n"
+            "caller's timeout to MessageQueue.recv()."
+        )
+    backup = target.with_name(target.name + BACKUP_SUFFIX)
+    if not backup.exists():
```
lishunyang12: Refresh stale backups before reapplying; --revert currently restores pre-upgrade vLLM code.

weishao-nv:
Thanks for the review @lishunyang12 — both were real bugs, fixed in 49fb3715.

**`max_sequence_length`**: you're right that the top-level form field was dropped,
so `--max-sequence-length` silently had no effect. Moved it into `extra_params`
as an int, matching how `examples/offline_inference/image_to_video/image_to_video.py`
passes it via `--extra-body` and how Cosmos3 reads it from the model extras.
Verified with `--dry-run --max-sequence-length 2048` that the value now reaches
the payload.

**Stale backups**: also correct, and worse than I'd realised — after a vLLM
upgrade the old backup survived, so `--revert` would overwrite the current
install with pre-upgrade code. `apply()` now always refreshes the backup, since
the target is known to be unpatched at that point and is therefore this install's
pristine copy. I reproduced the upgrade → re-apply → revert sequence in a temp
directory and confirmed `--revert` restores the current version; it now also
prints `refreshed` rather than `backed up` so the behaviour is visible.

Also merged current `main` and brought the PR up to date with the tooling that
landed since:

- Resolved a `.gitignore` conflict — kept the FL2V testdata allow-list and took
  `main`'s removal of the SoulX-Singer entry (#6362).
- Added the SPDX headers and markdownlint table style required by the hooks
  added in #6273.

`pre-commit run` passes locally on all files this PR touches.

PR #6125, opened 2026-08-12
[XPU][SDXL] Fix text encoder input device under CPU offload

Joshna-Medisetty:
## Problem
With `--enable-cpu-offload`, SDXL crashes in `encode_prompt`:

```
RuntimeError: Expected all tensors to be on the same device, but got index
is on cpu, different from other tensors on xpu:0 (wrapper_XPU__index_select)
```

Under offload the text encoder's parameters report `cpu`, so
`next(text_encoder.parameters()).device` returns `cpu` and the input ids are
placed on `cpu`. At forward time the offload hook has already moved the weights
to the accelerator, causing a device mismatch in the embedding lookup.

## Fix
Place inputs on the pipeline's compute device (`self.device`) instead of the
offloaded parameter's device — matching the pattern already used by FLUX, SD3,
and HiDream pipelines.

## Test
SDXL text-to-image runs with `--enable-cpu-offload` on XPU. Backend-agnostic
(`self.device` resolves to cuda/xpu/cpu as assigned).


fhfuih (commented):
Can confirm `self.device` is the correct way do to it (align with Qwen Image, Wan, etc.)

But I notice this pipeline's text_encoder is not moved to `self.device` (unlike most models: QWwen Image, Wan, Flux 2, ...) It is expected?

NickCao:
Added a regression test for this in https://github.com/vllm-project/vllm-omni/pull/6169

Joshna-Medisetty:
@NickCao   Thanks for adding the regression test, appreciate it.

@fhfuih That wasn't intentional. I pushed a commit that moves text_encoder and text_encoder_2 to self.device in __init__, matching SD3/Qwen Image/Flux/Wan. Now the encoder weights and inputs stay on the same device in both the offload and non-offload paths.

NickCao:
> But I notice this pipeline's text_encoder is not moved to `self.device` (unlike most models: QWwen Image, Wan, Flux 2, ...) It is expected?

This is in fact redundant (but harmless), pipelines are constructed inside `with target_device:` in `diffusers_loader.py`, so the weights are loaded on the correct devices by default.

NickCao (approved):

PR #6130, opened 2026-08-12
[Bugfix][FLUX.2-klein] Fix duplicated image position ids 

Joshna-Medisetty:
## Problem
FLUX.2-klein image editing crashes at warm-up (dummy run) for both
klein-4B and klein-9B:

    RuntimeError: Dummy run failed: The size of tensor a (2560) must match
    the size of tensor b (3584) at non-singleton dimension 1

## Root cause
During input preparation, `image_latent_ids` is already concatenated onto
`latent_ids`:

    latent_ids = torch.cat([latent_ids, image_latent_ids], dim=1)

Later, inside the denoising loop, the image-edit branch concatenated the
image ids **again**:

    latent_image_ids = torch.cat([latent_ids, image_latent_ids], dim=1)

so the image block was counted twice. The resulting position-id length
(3584) no longer matched the hidden-states sequence length (2560), which
is `cat([latents, image_latents])`.

The sibling `reference_image_latents` branch two lines above already does
the right thing (`latent_image_ids = latent_ids`) — the image-edit branch
was simply inconsistent. The duplicate concat was introduced when the
up-front pre-concat was added for inpaint support, without updating this
branch.

## Fix
Use the already-assembled `latent_ids` directly:

    -                latent_image_ids = torch.cat([latent_ids, image_latent_ids], dim=1)
    +                latent_image_ids = latent_ids

Platform-agnostic (pure shape/count bug). No other pipeline is affected:
`flux2` base does a single concat (no pre-concat), and other pipelines use
a different id layout.

## Test
Image editing runs to completion on XPU with CPU offload for both models:

- `FLUX.2-klein-4B` (256×256, 28 steps): generated in ~55s, correct edit.
- `FLUX.2-klein-9B` (512×512, 28 steps): generated in ~100s, correct edit,
  no VAE NaN.


NickCao (commented):
Reproduced this on CUDA by forcing the use of forward_native (the forward_cuda implementation calls vllm.vllm_flash_attn.layers.rotary.apply_rotary_emb, which doesn't check the shape of the tensor). The fix is correct, but `latent_image_ids = latent_ids` in the if branches can all be dropped, as another assignment already exists right before the if.

vllm_omni/diffusion/models/flux2_klein/pipeline_flux2_klein.py
```diff
@@ -1020,10 +1020,8 @@ def forward(self, req: DiffusionRequestBatch) -> DiffusionOutput:
 
             if reference_image_latents is not None:
                 latent_model_input = torch.cat([latents, reference_image_latents], dim=1)
-                latent_image_ids = latent_ids
             elif image_latents is not None:
                 latent_model_input = torch.cat([latents, image_latents], dim=1).to(self.transformer.dtype)
-                latent_image_ids = torch.cat([latent_ids, image_latent_ids], dim=1)
```
fhfuih, line 1026: Quick question (although not related to your edit but it's adjacent): any particular reason the reference image input is not moved to `self.transformer.dtype` and the other image latent input is?

Joshna-Medisetty: I think it is not intent. It is just inconsistency/oversight. But it hasn't caused any problems.

xuechendi (approved):

PR #6137, opened 2026-08-13
[Bugfix] Fix fallback loading for IndexTTS external models

### LINKED ISSUE #6123: [Bug]: IndexTTS 2.5 fails to load
IndexTTS 2.5 fails to load in current vllm-omni:

Command:  bash examples/online_serving/text_to_speech/indextts2/run_server.sh
Model: https://huggingface.co/IndexTeam/IndexTTS-2.5/tree/main

Using AttentionBackendEnum.TRITON_ATTN backend.
(StageEngineCoreProc_stage0_replica0) [indextts2_talker.py:1208] Loaded 453 weights for IndexTTS2TalkerForConditionalGeneration
(StageEngineCoreProc_stage0_replica0) ERROR [stage_engine_core_proc.py:190] StageEngineCoreProc failed to start.
(StageEngineCoreProc_stage0_replica0) ERROR [stage_engine_core_proc.py:190] Traceback (most recent call last):
(StageEngineCoreProc_stage0_replica0) ERROR [stage_engine_core_proc.py:190] File vllm-omni/vllm_omni/model_executor/models/indextts2/indextts2_talker.py, line 1271, in _warmup_external_models
(StageEngineCoreProc_stage0_replica0) ERROR [stage_engine_core_proc.py:190]     load_wav2vec2(self.model_path, device)
(StageEngineCoreProc_stage0_replica0) ERROR [stage_engine_core_proc.py:190]     ~~~~~~~~~~~~~^^^^^^^^^^^^^^^
(StageEngineCoreProc_stage0_replica0) ERROR [stage_engine_core_proc.py:190] File vllm-omni/vllm_omni/model_executor/models/indextts2/preprocess_utils.py", line 106, in load_wav2vec2
(StageEngineCoreProc_stage0_replica0) ERROR [stage_engine_core_proc.py:190]   raise FileNotFoundError(f"IndexTTS Wav2Vec2-BERT assets are missing from local bundle {model_path!r}")
(StageEngineCoreProc_stage0_replica0) ERROR [stage_engine_core_proc.py:190] FileNotFoundError: IndexTTS Wav2Vec2-BERT assets are missing from local bundle

sphinxkkkbc:
## Purpose

Fix #6123.

Allow IndexTTS external model dependencies to fall back to their Hugging Face repositories when they are not included in the local IndexTTS model snapshot.

## Root Cause

After the IndexTTS repository ID is resolved by Hugging Face, `model_path` points to a local snapshot directory. The external model loaders treated any local `model_path` as a complete offline bundle:

```python
if os.path.isdir(model_path):
    raise FileNotFoundError(...)
```

As a result, when an external dependency was absent from the IndexTTS snapshot, loading failed immediately instead of using the existing Hugging Face fallback.

This affected:

- BigVGAN
- CAMPPlus
- Wav2Vec2-BERT
- RepCodec

The RepCodec case was not exposed in #6123 but had the same root cause.

The loaders now continue to prefer bundled assets, while falling back to the corresponding Hugging Face repository when those assets are missing.

## Test Plan

```bash
pytest -v tests/model_executor/models/indextts2/test_s2mel_v2_5.py
pytest -v tests/model_executor/models/indextts2/test_enhanced_codec.py
```

**vLLM Version:**  
0.27.0

**vLLM-Omni Commit:**  
591c90dd043bf52278948706add59242ad207a17

## Test Result

```text
50 passed
```

**BEFORE SUBMITTING:** read [[CONTRIBUTING.md](https://github.com/vllm-project/vllm-omni/blob/main/CONTRIBUTING.md)](https://github.com/vllm-project/vllm-omni/blob/main/CONTRIBUTING.md) and run the [[precheck-pr skill](https://github.com/vllm-project/vllm-omni/blob/main/.claude/skills/precheck-pr/SKILL.md)](https://github.com/vllm-project/vllm-omni/blob/main/.claude/skills/precheck-pr/SKILL.md) with the code agent for a self-check against project conventions.

### DIFF
--- a/tests/model_executor/models/indextts2/test_enhanced_codec.py
+++ b/tests/model_executor/models/indextts2/test_enhanced_codec.py
@@ -82,3 +82,31 @@ def test_model_asset_resolver_supports_native_checkpoints_layout(tmp_path):
     checkpoint.touch()
 
     assert preprocess_utils.resolve_model_file(str(tmp_path), "codec.pth") == str(checkpoint)
+
+
+def test_campplus_falls_back_to_hub_for_local_model_dir(tmp_path, monkeypatch):
+    import huggingface_hub
+
+    from vllm_omni.model_executor.models.indextts2.utils.campplus import dtdnn
+
+    checkpoint = tmp_path / "downloaded-campplus.bin"
+    checkpoint.touch()
+    downloads = []
+
+    class FakeCAMPPlus(torch.nn.Module):
+        def load_state_dict(self, state, strict=True):
+            self.loaded_state = (state, strict)
+
+    monkeypatch.setattr(dtdnn, "CAMPPlus", lambda **_kwargs: FakeCAMPPlus())
+    monkeypatch.setattr(
+        huggingface_hub,
+        "hf_hub_download",
+        lambda repo_id, filename: downloads.append((repo_id, filename)) or str(checkpoint),
+    )
+    monkeypatch.setattr(preprocess_utils.torch, "load", lambda *_args, **_kwargs: {"weight": torch.ones(1)})
+    preprocess_utils._campplus_cache.clear()
+
+    model = preprocess_utils.load_campplus(str(tmp_path), torch.device("cpu"))
+
+    assert isinstance(model, FakeCAMPPlus)
+    assert downloads == [("funasr/campplus", "campplus_cn_common.bin")]
--- a/tests/model_executor/models/indextts2/test_s2mel_v2_5.py
+++ b/tests/model_executor/models/indextts2/test_s2mel_v2_5.py
@@ -838,6 +838,16 @@ def test_v25_resolved_vocoder_source_is_cached(monkeypatch):
     assert calls == [("/models/indextts25", "fake-bigvgan")]
 
 
+def test_bigvgan_source_falls_back_to_hub_repo_for_local_model_dir(tmp_path):
+    assert (
+        indextts2_s2mel_decoder._resolve_bigvgan_source(
+            str(tmp_path),
+            "nvidia/bigvgan_v2_22khz_80band_256x",
+        )
+        == "nvidia/bigvgan_v2_22khz_80band_256x"
+    )
+
+
 def test_stage1_preload_does_not_swallow_cuda_runtime_failure(monkeypatch):
     decoder = object.__new__(IndexTTS2S2MelDecoder)
     torch.nn.Module.__init__(decoder)
--- a/vllm_omni/model_executor/models/indextts2/indextts2_s2mel_decoder.py
+++ b/vllm_omni/model_executor/models/indextts2/indextts2_s2mel_decoder.py
@@ -78,8 +78,6 @@ def _resolve_bigvgan_source(model_path: str, vocoder_name: str) -> str:
     )
     if local_dir is not None:
         return local_dir
-    if os.path.isdir(model_path):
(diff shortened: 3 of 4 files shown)


linyueqian:
This is reviewed and ready from my side, and the only thing standing between it and an approval is a stuck CI build. A rebase onto current main will clear it.

The AMD lane has never reported at this head. Build 11399 was created at 18:52 and has sat at three steps finished and seventeen `waiting_for_dependencies` ever since, with `state: started` and no `finished_at`. It is not queued behind other work: AMD builds created hours later for other PRs have run to completion in under ten minutes. That build is wedged on its own dependency graph and will not recover on its own, and because it is still in a running state, cycling the `ready` label does not create a replacement, it is deduplicated against the one already in flight. A new commit is what produces a fresh build.

Everything else at this head is in good shape. The general lane and NPU are both green. The Intel lane is red, but it never executed a test body: both Intel steps hard-failed during collection with a `ModuleNotFoundError` on a test module unrelated to this PR, exiting status 2. PR #5730 hit the identical failure in build 11401 within two minutes of yours, on a different missing module and from a completely disjoint file set, which puts that break in the XPU image rather than in either branch.

So: please rebase onto current main. That gets you a fresh AMD build, and assuming it comes back consistent with the other lanes I will approve immediately. Your change is confined to IndexTTS external asset resolution and loading, plus its tests, with no device-specific or kernel code, so I do not expect ROCm to have an opinion about it, but I would rather have the lane actually report than approve around a gap.

linyueqian (approved):
The fallback behavior is consistent across BigVGAN, CAMPPlus, Wav2Vec2-BERT, and RepCodec: bundled assets still win, while a resolved local IndexTTS snapshot no longer prevents downloading a missing external dependency from its public repository. The focused tests cover both source resolution and loading, and I did not find a material regression in the diff.

Validation: static review of the four-file diff at `d9d16781` against its current merge base. No fork code was executed locally.

This head is a force-pushed rebase of the previously reviewed `591c90dd`. I verified that rather than assuming it: comparing the PR's own content at each head against its own merge base with main, the two are byte-identical, so the commit message on the squashed commit describes the original work rather than any new change. The review above stands unchanged.

CI at this head, all four lanes terminal. The general lane, build 14662, Intel build 7817, and NPU build 6415 are green. Worth noting that Intel is now green: at the previous head it failed at collection with `ModuleNotFoundError: No module named 'examples.offline_inference'`, which I attributed to the XPU image rather than this branch, and the rebase onto current main confirms that reading by clearing it without any change to the PR's own files.

AMD build 11415 hard-failed only `mi300_1: Diffusion · Model Test`, on the four SenseNova paged-decode tests, each raising `ImportError: vllm.vllm_flash_attn requires the CUDA flash attention extensions (_vllm_fa2_C or _vllm_fa3_C). On ROCm, use upstream flash_attn.`, closing `4 failed, 20 passed, 1 skipped, 2375 deselected`. That is the ROCm image lacking the CUDA flash-attention extensions, it reproduces identically on unrelated PRs, and this branch touches four files all under `vllm_omni/model_executor/models/indextts2/` and its tests, so it runs no diffusion code.

The rebase existed to replace an AMD build that wedged: build 11399 sat at three steps finished and seventeen `waiting_for_dependencies` for over five hours with no progress, while builds created later for other PRs completed in under ten minutes, and a build stuck in a running state prevents its own replacement because Buildkite deduplicates against it. Build 11415 came up running immediately after the push, which is what the rebase was for.

Issue #5279, opened 2026-07-21
[RFC]: Refactor vLLM-Omni AR & Generation Scheduler


R2-Y:
## Context

The AR and generation stage schedulers in `vllm_omni/core/sched/` extend
upstream vLLM V1 but contain unused paths, duplicated lifecycle code,
model-specific decisions, and large update methods that are difficult to review
against upstream vLLM changes.

## Goals

1. Remove confirmed dead code.
2. Remove redundant code.
3. Extract small shared abstractions.
4. Replace model-specific scheduler logic with validated capabilities or
  extension points.
5. Improve readability without changing scheduling behavior.

## Scope

This issue covers only `OmniARScheduler`, `OmniARAsyncScheduler`,
`OmniGenerationScheduler`, and their directly related shared helpers, configuration, runners, connectors, and Orchestrator paths.

### 1. Remove unused coordinator and wrapper code

- Remove the unused chunk path from `OmniSchedulingCoordinator`; production
async-chunk scheduling uses `OmniChunkTransferAdapter`.
- Remove `process_pending_full_payload_inputs_legacy()`.
- Remove the unused scheduler-local `KVCacheTransferData`, its API-doc entry,
the generation no-op stop-reason assignment, and the thin AR `_free_blocks`
override.
- Preserve the live full-payload path and resolve the
`requests_with_ready_chunks` guard before deleting that state.

### 2. Complete LLM scheduler deduplication

- Share AR/generation I/O initialization, new-request conversion, queue
restoration, and finish/abort sequencing.
- Remove generation's double `OmniNewRequestData` conversion.
- Extract narrowly scoped helpers for KV failure handling, event publication,
connector stats, scheduler stats, and output assembly.
- Keep AR KV transfer/streaming logic and generation one-shot completion
explicit.

### 3. Make inter-stage input transport generic

- Replace `_FULL_PAYLOAD_INPUT_STAGES` with a resolved stage transport
capability.
- Validate that the selected runner/platform implements the required connector
path.
- Move `code_predictor_codes`, `left_context_size`, and other model metadata  
interpretation out of scheduler internals into a typed adapter contract.

## Required sequencing

1. Add characterization coverage for current queue, output, streaming, KV, and
  fallback contracts.
2. Remove confirmed dead code.
3. Land low-risk shared initialization, conversion, and finish helpers.
4. Split `update_from_output()` into responsibility-based helpers.
5. Add the validated transport capability and migrate existing models.
[...]

PR #6149, opened 2026-08-13
[Refactor][2/N Scheduler]Replace _FULL_PAYLOAD_INPUT_STAGES with a resolved stage transport capability.

wy17003:
Replace _FULL_PAYLOAD_INPUT_STAGES with a resolved stage transport capability.

## Purpose
Finish part of 3rd step of https://github.com/vllm-project/vllm-omni/issues/5279, make inter-stage input transport generic.

- Replace _FULL_PAYLOAD_INPUT_STAGES with a resolved stage transport capability.
- Validate that the selected runner/platform implements the required connector path.

## Test Plan

**vLLM Version:**
0.27.0
**vLLM-Omni Commit:**
0.27.0

#### Qwen3-Omni (async_chunk = true)
```text
vllm serve /data/Qwen3-Omni-30B-A3B-Instruct/ --omni --host 0.0.0.0 --port 9000 --deploy-config path/vllm-omni/vllm_omni/deploy/qwen3_omni_moe.yaml
```

```text
response=$(curl -s http://localhost:9000/v1/chat/completions \
  -d '{
    "model": "/data/Qwen3-Omni-30B-A3B-Instruct/",
    "messages": [{"role": "user", "content": "Describe vLLM in brief."}],
    "modalities": ["text", "audio"]
  }')

echo "$response" | jq -r '.choices[0].message.content'
echo "$response" | jq -r '.choices[1].message.audio.data' | base64 -d > output.wav

vLLM is an open-source library designed for efficient inference and serving of large language models (LLMs). It optimizes memory usage and throughput by introducing innovative techniques such as PagedAttention, which dynamically manages attention keys and values in memory, reducing fragmentation and enabling higher batch sizes. This allows vLLM to achieve faster generation speeds and support larger models than traditional inference frameworks. It is particularly effective for serving LLMs in production environments, offering high performance and scalability.
```

#### Qwen2.5-Omni (async_chunk = false)
```text
vllm serve /data/Qwen2.5-Omni-7B/ --omni --host 0.0.0.0 --port 9000 --deploy-config path/vllm-omni/vllm_omni/deploy/qwen2_5_omni.yaml
```

```text
response=$(curl -s http://localhost:9000/v1/chat/completions \
  -d '{
    "model": "/data/Qwen2.5-Omni-7B/",
    "messages": [{"role": "user", "content": "Describe vLLM in brief."}],
    "modalities": ["text", "audio"]
  }')

echo "$response" | jq -r '.choices[0].message.audio.data' | base64 -d > output2.5.wav
```


vllm_omni/core/sched/omni_scheduler_mixin.py
```diff
@@ -73,14 +71,19 @@ class OmniSchedulerMixin:
...
-            OmniChunkTransferAdapter(self.vllm_config) if getattr(model_config, "async_chunk", False) else None
-        )
-        self.input_coordinator = (
-            OmniSchedulingCoordinator(stage_id=getattr(model_config, "stage_id", 0))
-            if uses_full_payload_input_coordinator(model_config)
-            else None
-        )
+        transport = getattr(model_config, "stage_input_transport", None)
+
+        if transport == StageInputTransport.ASYNC_CHUNK:
```
hsliuustc0106: [P1] Preserve the async adapter for producer stages. Stage 0 has no `input_sources`, so `_resolve_stage_input_transport()` assigns it `NONE`; this branch then drops its `OmniChunkTransferAdapter`. But `OmniARScheduler.update_from_output()` only calls `save_async()` when that adapter exists, and the adapter owns the connector `put()` to the next stage. In an `async_chunk` pipeline the downstream stage waits for chunks that stage 0 never sends. Please retain sender initialization based on `async_chunk` / the async producer hook, or represent send and receive capabilities separately, and add a stage-0 sender plus stage-1 receiver regression test.

vllm_omni/config/stage_config.py
```diff
@@ -855,6 +896,7 @@ def _build_engine_args(
         engine_args["model_subdir"] = ps.model_subdir
     if ps.tokenizer_subdir:
         engine_args["tokenizer_subdir"] = ps.tokenizer_subdir
+    engine_args["stage_input_transport"] = stage_input_transport
```
hsliuustc0106: [P1] Keep this topology-resolved value authoritative. `_parse_stage_deploy()` accepts `stage_input_transport` as a flat unknown key, under `engine_args`, or under `engine_extras`; then line 913 overwrites this assignment with `ds.engine_extras` after producer/topology validation has already run. A deploy can therefore disable a required wait or enable a wait mode whose producer was never validated. Please reject this field from legacy free-form overrides (matching the structured path) and cover all three YAML spellings.

hsliuustc0106 (commented):
Two blocking transport-contract issues found in the static review.

hsliuustc0106:
review the PR by yourself with explict human inline comment

hsliuustc0106:
fix conflicts

wy17003:
> fix conflicts
Thanks for the review. I’ve addressed all the comments in the latest commit.

hsliuustc0106:
Nit (post-rebase leftover): `tests/config/test_config_factory.py:2730` — the `test_ming_flash_omni_topology` docstring still cites ``_FULL_PAYLOAD_INPUT_STAGES``, which this PR deletes. Rephrase to the new gating terms (stage 0 has no `custom_process_next_stage_input_func`, stage 1 sets no `requires_full_payload_input=True`).

Both earlier P1s verified fixed at d108e2b6: the stage-0 async adapter is now keyed on `async_chunk` alone (`omni_scheduler_mixin.py:109`), and `requires_full_payload_input` is a pipeline-owned structured field that deploy `engine_extras`/CLI overrides cannot clobber (rejected with "no structured config owner", tested in `test_from_pipeline_config_rejects_full_payload_input_capability_overrides`).

hsliuustc0106 (approved):
Both earlier P1s verified fixed at this head: the stage-0 async adapter is now keyed on `async_chunk` alone, and `requires_full_payload_input` is a pipeline-owned structured field that deploy/CLI overrides cannot clobber. Per-pipeline declarations exactly cover the old allowlists (including audex variants and the nemotron token-path talker exclusion), and the legacy-gate equivalence test pins the semantics. Only remaining item is the P3 docstring nit above — non-blocking.

wy17003:
> Both earlier P1s verified fixed at this head: the stage-0 async adapter is now keyed on `async_chunk` alone, and `requires_full_payload_input` is a pipeline-owned structured field that deploy/CLI overrides cannot clobber. Per-pipeline declarations exactly cover the old allowlists (including audex variants and the nemotron token-path talker exclusion), and the legacy-gate equivalence test pins the semantics. Only remaining item is the P3 docstring nit above — non-blocking.

The NIT comments have been addressed and the CI gate checks are all passing locally. Could you please re-run the required workflows? Thanks!

wy17003:
> > Both earlier P1s verified fixed at this head: the stage-0 async adapter is now keyed on `async_chunk` alone, and `requires_full_payload_input` is a pipeline-owned structured field that deploy/CLI overrides cannot clobber. Per-pipeline declarations exactly cover the old allowlists (including audex variants and the nemotron token-path talker exclusion), and the legacy-gate equivalence test pins the semantics. Only remaining item is the P3 docstring nit above — non-blocking.
> 
> The NIT comments have been addressed and the CI gate checks are all passing locally. Could you please re-run the required workflows? Thanks!

Apologies for the oversight — I realized I missed the pre-commit ruff format and DCO checks. Both have been fixed and pushed. All pre-commit hooks now pass locally.

vllm_omni/worker/omni_connector_validation.py
```diff
@@ -0,0 +1,42 @@
+"""Validation for Omni connector support on selected LLM workers."""
```
hsliuustc0106, line 1: is this a util function? we seldom name the scripts with support suffix

vllm_omni/worker/gpu_generation_worker.py
```diff
@@ -27,6 +27,8 @@ class GPUGenerationWorker(OmniWorkerMixin, OmniGPUWorkerBase):
     ``execution_type=StageExecutionType.LLM_GENERATION``.
     """
 
+    model_runner_cls = GPUGenerationModelRunner
```
hsliuustc0106, line 30: why we need to make this equivalent changes?

wy17003: Thanks for asking. This change is required by the runner capability validation introduced in this PR.
`OmniEngineArgs.__post_init__` validates the worker selected by the current platform before the worker is instantiated. Since the connector implementation lives on the model runner, the worker needs to expose its runner through `model_runner_cls`.

vllm_omni/worker/omni_connector_validation.py
wy17003: Yes, it is a stateless validation helper, but it is specific to the worker/model-runner connector contract rather than a general utility. I think the `worker` package is still the appropriate ownership boundary because it resolves `worker_cls.model_runner_cls` and validates it against `OmniConnectorModelRunnerMixin`. I will rename the module to `omni_connector_validation.py` to make its responsibility explicit.

Issue #5811, opened 2026-08-05
[Feature]: Image (Text-to-Image / Image-to-Image) Prometheus Metrics


Bounty-hunter:
## Motivation

- vLLM-Omni already exposes 17 `vllm_omni:*` families (6 pipeline + 7 audio + 4 transfer), yet `docs/design/metrics.md` states that "diffusion / image / video families land in a follow-up".

### Already available (reuse, zero code change)

| Metric | Type | Labels | Source | Purpose |
|---|---|---|---|---|
| `vllm_omni:e2e_request_latency_s` | Histogram | `model_name` | finalize `request_succeeded()` (`entrypoints/omni_base.py:547`) | image e2e latency (arrival → completion) |
| `vllm_omni:num_requests_running` / `num_requests_waiting` | Gauge | `model_name` | orchestrator counters (`metrics/prometheus.py`) | load level |
| `vllm_omni:requests_success_total` | Counter | `model_name, finished_reason` | finalize | completion distribution (stop/length/abort) |
| `vllm_omni:prompt_tokens_total` / `generation_tokens_total` | Counter | `model_name` | stage token accounting + finalize | input/output scale |
| `vllm_omni:transfer_size_bytes` / `transfer_tx_s` / `transfer_rx_s` / `transfer_in_flight_s` | per definitions | `model_name, from_stage, from_replica, to_stage, to_replica` | `metrics/transfer.py` | cross-stage transfer, e2e gap attribution |
| upstream `vllm:*` (AR stages: TTFT/TPOT/ITL/KV cache, etc.) | —

### New — data already produced (read existing values → export)

| Metric | Type | Labels | Source | Purpose |
|---|---|---|---|---|
| `vllm_omni:stage_gen_time_s` | Histogram | `model_name, stage` | `StageRequestStats.stage_gen_time_ms` (`metrics/stats.py:37`, `engine/stage_pool.py:595`) | diffusion stage total time (submit → output, **includes in-stage queueing**), core of e2e attribution |
| `vllm_omni:request_queue_wait_s` | Histogram | `model_name` | `pipeline_timings["queue_wait_ms"]` (`orchestrator.py:681`) | orchestration-layer queueing wait, overload alerting |
[...]

PR #6150, opened 2026-08-13
text and image metric 

AbelSara:

related to #5811 
co-auth: @Bounty-hunter 

## Purpose

Add 19 image and diffusion metrics under the `vllm_omni:` Prometheus namespace. Metric emission is controlled by `--log-stats` (off by default). Optional metrics such as profiler timing, KV transfer, and peak memory are only emitted when their source data is available.

The implementation uses the existing scheduler, engine output, and orchestrator paths. It does not add a new metrics channel or extra GPU synchronization.

### Stage service metrics

- `vllm_omni:stage_gen_time_s` (`model_name`, `stage`, `stage_type`): stage submission to finished stage output.
- `vllm_omni:request_queue_wait_s` (`model_name`): request enqueue to initial stage submission. A measured zero is retained; a missing timestamp is not reported as zero.
- `vllm_omni:stage_waiting_requests` (`model_name`, `stage`): sum of the latest scheduler waiting snapshots from live replicas in the stage.
- `vllm_omni:num_inference_steps` (`model_name`): inference-step distribution for diffusion stages only.
- `vllm_omni:image_count_total` (`model_name`): cumulative number of image outputs. Non-image outputs do not contribute.
- `vllm_omni:image_pixels` (`model_name`): pixel-count distribution for image outputs.
- `vllm_omni:peak_memory_mb` (`model_name`, `stage`): latest positive peak-memory value reported by a completed stage.
- `vllm_omni:stage_in_queue_s` (`model_name`, `stage`): diffusion scheduler enqueue to initial admission as `RUNNING`.

`stage_in_queue_s` is measured directly in the diffusion scheduler. It is no longer estimated from `stage_gen_time - diffusion_engine_exec`, which included scheduler waiting in the value being subtracted.

### Failure, KV wait, and forward timing

- `vllm_omni:requests_failed_total` (`model_name`, `reason`): failed requests grouped into `client_abort`, `client_disconnect`, `stage_error`, and `unknown`.
- `vllm_omni:kv_wait_s` (`model_name`, `connector_type`): time spent waiting for the AR scheduler's KV extraction acknowledgement.
- `vllm_omni:diffusion_forward_s` (`model_name`, `stage`, `replica`): profiler-reported diffusion forward/denoise-loop time.

Failure counting is centralized in `_record_request_failure_once()`. Request state prevents exception handling, abort, and cleanup paths from counting the same request more than once.

KV wait starts when `_free_request()` moves a request into `waiting_for_transfer_free` and ends on the extraction acknowledgement. The duration and connector type reuse `OmniEngineCoreOutput.kv_transfer_params`; no new message type is introduced. Incomplete timestamps are removed on abort, error, and request cleanup.

`diffusion_forward_s` reuses the existing `DiffusionPipelineProfiler` `*.diffuse` duration and is absent when that profiler data is unavailable.

### Diffusion engine timing

The diffusion engine attaches `*_ms` values to its existing output metrics. `OrchestratorAggregator` converts and accumulates them in seconds, then `_observe_diffusion_finalize()` emits:

- `vllm_omni:diffusion_request_time_s`
- `vllm_omni:diffusion_request_time_per_step_s`
- `vllm_omni:diffusion_preprocess_s`
- `vllm_omni:diffusion_postprocess_s`
- `vllm_omni:diffusion_kv_load_s`

All five use `model_name`, `stage`, and `replica` labels.

`diffusion_request_time_s` measures from immediately before scheduler insertion until the engine receives the output. It includes scheduler waiting, execution, and asynchronous output wait, but excludes preprocessing and postprocessing. `diffusion_request_time_per_step_s` is this value divided by `num_inference_steps`; it is a service-time-per-step metric, not pure GPU denoise latency.

`diffusion_kv_load_s` is only present when the diffusion request receives upstream KV data.

### VAE, denoise, and image TTFP

- `vllm_omni:vae_decode_s` (`model_name`, `stage`, `replica`): VAE decode time extracted from profiler stage durations.
- `vllm_omni:denoise_step_latency_s` (`model_name`, `stage`, `replica`): `forward_time_s / num_inference_steps` for image diffusion outputs.
- `vllm_omni:image_ttfp_s` (`model_name`, `stage`, `replica`): image-stage submission to first non-empty image output.

`denoise_step_latency_s` is emitted only when forward timing exists, the step count is positive, and the output type is `image`. It excludes scheduler waiting, preprocessing, postprocessing, VAE decode, and KV load.

`image_ttfp_s` is stage-scoped rather than request-entry-scoped. It excludes API queueing and preceding stages such as thinker inference.

### Emission rules

Stage metrics are emitted when a stage finishes, while diffusion breakdown metrics go through `_observe_diffusion_finalize()`. Finished output messages are tracked so replaying the same result does not update counters or histograms twice.

Audio diffusion stages emit both diffusion and audio metrics. Image and video diffusion stages emit diffusion metrics, with image-only families guarded by the output type. Text-only stages continue to use the wrapped upstream `vllm:*` metrics.

## Test Plan
pytest -s -v \
  tests/diffusion/test_diffusion_engine_metrics.py \
  tests/diffusion/test_diffusion_scheduler.py \
  tests/engine/test_orchestrator.py \
  tests/entrypoints/test_omni_entrypoints.py \
  tests/metrics/test_emit_calls.py \
  tests/metrics/test_modality.py

**vLLM Version:**
0.26.0

## Test Result
| Item | Result |
| --- | --- |
| Test level | L1 CPU unit test |
| Static checks | PASS |
| Tests collected | 283 |
| Passed | 283 |
| Failed | 0 |
| Skipped | 0 |
| Overall result | **PASS** |


hsliuustc0106:
review this PR by yourself first with explicit comments and fixes

AbelSara:
@Bounty-hunter @hsliuustc0106 @Gaohan123 this pr is ready for review

Gaohan123:
@AbelSara Please resolve conflicts. Thanks

AbelSara:
> @AbelSara Please resolve conflicts. Thanks

I've resolve conflict by `main` branch

vllm_omni/diffusion/diffusion_engine.py
```diff
@@ -188,6 +189,37 @@ def _move_tensor_tree_to_cpu(value: object) -> object:
     return value
 
 
+def _sum_stage_durations_ms(output: Any, suffix: str) -> float | None:
```
Gaohan123: I suggest we move these functions into utils

vllm_omni/entrypoints/omni_base.py
```diff
@@ -39,6 +39,32 @@
 
 logger = init_logger(__name__)
 
+_FAILURE_REASONS = frozenset({"client_abort", "client_disconnect", "stage_error", "unknown"})
+
+
+def _normalize_failure_reason(reason: str | None) -> str | None:
```
Gaohan123: The same

vllm_omni/diffusion/diffusion_engine.py
```diff
@@ -456,14 +491,24 @@ async def step_streaming(self, request: OmniDiffusionRequest) -> AsyncGenerator[
...
-                        "diffusion_engine_exec_time_ms": exec_total_time * 1000,
-                        "diffusion_engine_total_time_ms": step_total_ms,
-                        "postprocess_time_ms": postprocess_time * 1000,
-                    }
-                )
+                metrics_update = {
+                    "preprocess_time_ms": preprocess_time * 1000,
+                    "diffusion_engine_exec_time_ms": exec_total_time * 1000,
+                    "postprocess_time_ms": postprocess_time * 1000,
+                    metric_defs.DIFFUSION_SCHEDULER_WAITING_KEY: self._scheduler_num_waiting_reqs,
```
Gaohan123: [P2] Refresh the diffusion waiting gauge on abort/error paths

The scheduler waiting snapshot is attached only after `postprocess_output()` succeeds. However, `postprocess_output()` raises for aborted or errored outputs, so if the last queued request terminates through either path, the zero snapshot never reaches the orchestrator. As a result, `vllm_omni:stage_waiting_requests` can remain stuck at an earlier non-zero value and mislead queue alerts or autoscaling.

Please publish the scheduler snapshot before post-processing can raise, or refresh the gauge through a terminal cleanup/status path that also runs for aborts and errors.

vllm_omni/engine/orchestrator.py
```diff
@@ -1078,6 +1111,26 @@ async def _handle_stage_error(self, stage_id: int, output: Any) -> None:
...
+        """Drop a dead replica's snapshot and refresh the stage total."""
+        self._stage_replica_waiting.pop((stage_id, replica_id), None)
+        self._set_stage_waiting_total(stage_id)
+
+    def _set_stage_waiting_total(self, stage_id: int) -> None:
+        if self._prom_metrics is None:
+            return
+        total = sum(
+            n_waiting
+            for (snapshot_stage_id, _), n_waiting in self._stage_replica_waiting.items()
```
Gaohan123, line 1162: [P2] Exclude normally unregistered replicas from the waiting total

This sums every historical entry in `_stage_replica_waiting`. The dead-replica path removes its snapshot, but the normal `UnregisterRemoteReplicaMessage` path detaches the client through `MembershipController` without clearing this map. Therefore, a replica that last reported a non-zero waiting count continues to contribute after it is no longer live, which violates the documented “sum of live replicas” semantics.

Please either filter the entries using `self.stage_pools[stage_id].live_replica_ids()`, or remove the corresponding snapshot and republish the gauge during normal unregister.

vllm_omni/engine/stage_pool.py
```diff
@@ -643,17 +642,10 @@ def build_stage_metrics(
         )
         image_pixels = self._count_image_pixels(request_outputs) if output_unit_type == "image" else 0
         num_inference_steps = coerce_positive_int_scalar(getattr(sampling_params, "num_inference_steps", None)) or 0
-        denoise_step_latency_ms = (
-            defs.compute_denoise_step_latency(stage_gen_time_ms, num_inference_steps)
-            if output_unit_type == "image"
-            else 0.0
-        )
```
ZacheryAU, line 660: The numerator of `denoise_step_latency` is narrowed in this PR, but this attribute is consumed in benchmarks. It is better to fetch `forward_time` and update the calculation logic here rather than to delete this?

vllm_omni/engine/stage_pool.py
```diff
@@ -643,17 +642,10 @@ def build_stage_metrics(
...
         has_output_timestamps = bool(output_timestamps)
         first_ts = output_timestamps[0] if has_output_timestamps else now
         serving_time_to_first_output_ms = (
-            max((non_empty_first_output_ts - request_timestamp) * 1000.0, 0.0)
-            if non_empty_first_output_ts is not None
-            else 0.0
+            max((non_empty_first_output_ts - submit_ts) * 1000.0, 0.0) if non_empty_first_output_ts is not None else 0.0
         )
```
ZacheryAU, line 670: `serving_time_to_first_output_ms` is also consumed in benchmarks. It is not recommended to modify it only because of one new metric. And this change seems to calculate `image_ttfp`, which is not included in #5811 currently. `time_to_first_*` are usually used in streaming modalities. If `image_ttfp` is still needed, it would be better to add a new attribute for `max((non_empty_first_output_ts - submit_ts) * 1000.0, 0.0)` and keep original `serving_time_to_first_output_ms` and `request_timestamp`.

ZacheryAU (commented):
The PR is related to some attributes which are also consumed in benchmarks.

docs/design/metrics.md
```diff
@@ -167,6 +180,39 @@ second.
 | `vllm:omni_requests_success_total` | Counter | `model_name`, `finished_reason` | Total requests by completion reason ({stop, length, abort, ...}); aborts cover client-disconnect / cancellation paths in addition to upstream `FinishReason.ABORT` |
 | `vllm:omni_e2e_request_latency_s` | Histogram | `model_name` | Pipeline-global end-to-end latency in seconds |
 
+### Image and diffusion service-level metrics
+
+| Metric | Type | Labels | Description |
+|--------|------|--------|-------------|
+| `vllm:omni_stage_gen_time_s` | Histogram | `model_name`, `stage`, `stage_type` | Stage submit to finished output; includes in-stage queueing |
```
ZacheryAU: Could you update `vllm:omni_` as `vllm_omni:` in this file, like current `docs/usage/metrics.md`?

Gaohan123 (approved):
LGTM. Thanks

vllm_omni/metrics/stats.py
```diff
@@ -583,6 +594,9 @@ def _as_stage_request_stats(
             if req_id in self.diffusion_metrics
             else None
         )
+        forward_time_s = (stats.diffusion_metrics or {}).get("forward_time_s")
```
hsliuustc0106, line 598: [P2] `denoise_step_latency_ms` reads 0 in default runs (pipeline profiler off)

`forward_time_ms` is only attached when `enable_diffusion_pipeline_profiler` is on (default off, `omni_config.py:722`): with it off the profiler mixin never initializes `stage_durations`, so `diffusion_engine.py:485-487` attaches nothing, this branch never fires, and the snapshot ships 0.0. The bench CLI consumes this key (`benchmarks/metrics/metrics.py:819`, `benchmarks/patch/patch.py:1109`), where base always computed it from `stage_gen_time_ms` — default bench runs now report a silently-zero column, on top of the numerator narrowing discussed in the resolved thread.

Suggestion: fall back to `stage_gen_time_ms / num_inference_steps` when `forward_time_s` is absent (profiler value wins when present), or document the bench column as requiring the profiler flag.

vllm_omni/diffusion/diffusion_engine.py
```diff
@@ -444,7 +452,14 @@ async def step_streaming(self, request: OmniDiffusionRequest) -> AsyncGenerator[
                     )
                     raise
             postprocess_start_time = time.perf_counter()
-            formatted_outputs = self.postprocess_output(request, output)
+            scheduler_metrics = diffusion_scheduler_waiting_metrics(getattr(self, "_scheduler_num_waiting_reqs", 0))
```
hsliuustc0106, line 455: [P3] Defensive `getattr` on own-initialized attributes can be direct access

`self._scheduler_num_waiting_reqs` is set in `__init__` here, `output.kv_recv_ms` is a `DiffusionOutput` field (:488), and `_clear_kv_wait_starts` does the same for `_kv_wait_start_ts` (`omni_ar_scheduler.py:162`). Direct access is shorter and catches typos instead of silently returning the default.

Issue #5450, opened 2026-07-27
[RFC]: Worker/ModelRunner correctness fixes (ModelRunner Refactor 2/N)


tzhouam:
## 1. Summary

Seven correctness items in the GPU worker runners, found by the Worker/ModelRunner audit: **six defect fixes plus one item whose original audit premise did not survive the code check** and lands as a convention-pinning test instead. Proposed as one PR (`fix/runner-correctness`), **one commit per defect (9 commits** — the robustness trio stays three independent commits), each with its own regression test.

Part of the ModelRunner refactor series (1/N: #5101; test safety net: #5333). Sequencing rationale: these must land **before** the structural dedup steps (shared `execute_model` preamble, `_dummy_run` unification), so the merged code paths don't enshrine the bugs. Behaviour changes are deliberate; everything else is gated unchanged by the #5333 safety net (the qwen3-omni text+wav hash baseline must not move — none of these paths is exercised by the greedy batch-1 baseline, and each commit states why).

Anchors: `origin/main @ 5b230541`.

## 2. Proposed Change — the seven items

| # | Site | Defect | Action |
|---|---|---|---|
| C1 | `gpu_generation_model_runner.py:167` | zero-token early return shadows the DP / KV-connector block | fix |
| C2 | `gpu_ar_model_runner.py:1942` | `clamp(max=logits_vocab)` flagged as OOB — premise revised | pin with test |
| C3 | `gpu_ar_model_runner.py:500`, `:862` | sparse-audio marker `bool(tensor)` crash + silent wrong-request `v[0]` fallback | fix |
| C4 | `gpu_ar_model_runner.py:1015-1020` | in-place mutation of engine-shared `scheduler_output` | fix |
| C5 | `gpu_generation_model_runner.py:437-446` | `sample_tokens` silently mishandles `num_reqs > 1` | fix |
| C6 | `gpu_generation_model_runner.py:565` | `_dummy_run` missing `has_preprocess` input branch (capture≠replay) | fix |
| C7 | `gpu_ar_model_runner.py:1021`, `:444`, `:1450` | robustness trio: swallowed KV-metadata exception, stale memoization, silent sampler fallthrough | fix (3 commits) |

### C1 — generation zero-token early return shadows the DP/KV-connector path
[...]

PR #6157, opened 2026-08-13
[Model] Add Audio8 TTS Preview 0.6B (DualAR, 44.1 kHz codec)

NancyFyong:
## Summary

Adds serving support for [`Audio8/Audio8-TTS-Preview-0.6b`](https://huggingface.co/Audio8/Audio8-TTS-Preview-0.6b), a 0.6B DualAR TTS model (24-layer Slow AR + 4-layer Fast AR, 10 codebooks, 44.1 kHz codec at ~21.5 fps, 11 languages, zero-shot voice cloning, Apache-2.0).

## Design

Audio8's arktts is explicitly Fish Audio S2 Pro-inspired, so this follows the in-tree `fish_speech` two-stage layout: `audio8_tts_slow_ar` (text → semantic tokens + codec codes) → `audio8_tts_codec_decoder` (codes → waveform), streaming via `async_chunk`.

Places it diverges on purpose:

| Area | Audio8 | Why not reuse fish_speech's version |
|---|---|---|
| Backbone | `attention_qkv_bias=true`, no q/k norm → `Qwen2Model` | S2 Pro uses a Qwen3-shaped backbone; reusing it silently mismatches |
| Codebook embeddings | summed, no `1/sqrt(Q+1)` rescale | rescaling audibly changes prosody |
| Sampling | top-k/top-p on raw logits, then temperature, then Gumbel-max | reverse of the usual order — `common/nucleus_ras_sampling.py` applies temperature first, so it doesn't fit |
| Codec | vendored `ArkttsCodec`, 44.1 kHz / 2048 samples/frame | different codec entirely |
| `trust_remote_code` | false | `transformers` prefers `auto_map` over registered `AutoConfig`, which would bypass the Qwen2-shaped config that `Qwen2Model` needs |

Serving uses the `tts_adapters` registry — one adapter with `stage_keys = {"audio8_tts_slow_ar"}`, zero new `self._tts_model_type == ...` branches, so the `check_tts_adapter.py` ratchet counters stay flat. Prompt building follows the `qwen3_tts` pattern: helpers in `serving_speech.py`, dispatched from the adapter.

## Files

- Model: `vllm_omni/model_executor/models/audio8_tts/` — slow AR, fast AR, codec + utils, config, sampling, prompt utils, pipeline
- Streaming: `vllm_omni/model_executor/stage_input_processors/audio8_tts.py`
- Serving: `vllm_omni/entrypoints/openai/tts_adapters/audio8_tts.py`, Audio8 helpers in `serving_speech.py`
- Registration: `models/registry.py`, `config/pipeline_registry.py`, `transformers_utils/configs/audio8_tts.py`, `deploy/audio8_tts.yaml`
- Examples: `examples/{offline_inference,online_serving}/text_to_speech/audio8_tts/` (client, server, Gradio streaming demo)
- Tests: `tests/model_executor/models/audio8_tts/`, `tests/model_executor/stage_input_processors/test_audio8_tts_async_chunk.py`, `tests/entrypoints/openai/test_audio8_tts_adapter.py`, `tests/e2e/{offline_inference,online_serving}/test_audio8_tts.py`
- Docs/bench: `docs/models/supported_models.md`, `benchmarks/tts/{model_configs.yaml,README.md}`

## Validation

Component parity against the HF reference (throwaway scripts; numbers reproducible from the notes below):

- Codec encode is bit-identical (1 element in 220 differs, TF32 only).
- Codec decode: 105.5 dB SNR with `cudnn.allow_tf32=False`, 49.7 dB with TF32 on. Traced to TF32 convolutions, not wiring.
- `build_text_only_prompt_ids` is byte-identical to the reference `ArkttsProcessor` prefix (27 tokens on a sample sentence).
- Fast AR on captured hidden states: greedy codes agree 22/30 across three frames with identical argmax; logit deltas max 0.5 / mean 0.116 vs reference std 5.675. Bf16 accumulation.
- End-to-end offline: 44.1 kHz WAV, plausible speech stats, duration within ~5% of the reference for the same text.

Tests, 1x H20 on current main:

```bash
# unit + adapter contract
pytest -q tests/model_executor/models/audio8_tts/ \
          tests/model_executor/stage_input_processors/test_audio8_tts_async_chunk.py \
          tests/entrypoints/openai/test_audio8_tts_adapter.py                     # 34 passed

# offline e2e
AUDIO8_TTS_MODEL_PATH=<local> pytest -q -s tests/e2e/offline_inference/test_audio8_tts.py \
    --run-level core_model                                                        # 2 passed (dummy weights)
AUDIO8_TTS_MODEL_PATH=<local> pytest -q -s tests/e2e/offline_inference/test_audio8_tts.py \
    --run-level advanced_model                                                    # 2 passed (real weights)

# online e2e: non-streaming WAV, streaming PCM, voice clone, 2 error paths
AUDIO8_TTS_MODEL_PATH=<local> pytest -q -s tests/e2e/online_serving/test_audio8_tts.py \
    --run-level advanced_model                                                    # 5 passed

pre-commit run --from-ref origin/main --to-ref HEAD                               # all hooks pass
```

Two harness issues surfaced while getting the e2e suite green, both fixed here:

- `core_model` runs on dummy weights. `stage_config_path_for_run_level` patches every stage to `load_format: dummy` at L1/L2, so the AR never emits EOS and always generates to `max_tokens` (1023 frames, 47.5 s). Duration bounds now only apply at `advanced_model`/`full_model`; L1/L2 keeps the structural assertions.
- The HNR speech check assumed 24 kHz. `_compute_pcm_hnr_db` always used 24 kHz, which on 44.1 kHz audio moves the autocorrelation pitch search from 80–400 Hz to 147–735 Hz and misses the fundamental — the same waveform scores -0.08 dB at 24 kHz vs 1.83 dB at 44.1 kHz, so the streaming test failed with "Audio distortion detected" on clean audio. The helper now takes the rate from `expected_sample_rate` (default unchanged); the streaming test declares it.

### Benchmark

1x H20, bf16, TP1, `max_model_len=2048`, 2 warmups, 20 prompts from `seed_tts_smoke/en` (mean audio 4.5 s), deploy defaults. The GPU was shared with an unrelated job holding ~80% SM, so these are a lower bound; a clean re-measurement on an idle GPU is still outstanding.

```bash
export VLLM_OMNI_BENCH_AUDIO_SAMPLE_RATE=44100
vllm bench serve --omni --host 127.0.0.1 --port 8092 \
    --model Audio8/Audio8-TTS-Preview-0.6b \
    --backend openai-audio-speech --endpoint /v1/audio/speech \
    --dataset-name seed-tts-text --dataset-path benchmarks/build_dataset/seed_tts_smoke \
    --seed-tts-locale en --num-prompts 20 --num-warmups 2 --max-concurrency <C> \
    --percentile-metrics ttft,e2el,audio_rtf,audio_ttfp,audio_duration,audio_underrun
```

| | HF reference (batch=1) | c=1 | c=4 | c=8 |
|---|---|---|---|---|
| RTF (mean) | 1.008 | **0.19** | 0.30 | 0.54 |
| Throughput (req/s) | 0.23 | 1.18 | 2.84 | **2.93** |
| E2E latency (mean, ms) | 4375 | 850 | 1349 | 2361 |
| First packet (mean, ms) | n/a (no streaming) | 63 | 105 | 1124 |
| Underrun p99 / continuity | n/a | 0 / 100% | 0 / 100% | 0 / 100% |

~5.3x lower per-request RTF than the reference at c=1; ~12.8x aggregate throughput at c=8. A stage-0 `max_num_seqs` sweep (4 vs 8, table in `examples/online_serving/text_to_speech/README.md`) gives +27% throughput and 4.5x lower first-packet latency at c=8 for `max_num_seqs: 8`, at the cost of a 0.22 s worst-case buffer deficit in 1 of 20 requests. The shipped default stays at 4, where every point had zero underrun.

Raw JSONs, server logs, and the HF baseline script: [`audio8-tts-benchmark-data.tar.gz`](https://github.com/NancyFyong/vllm-omni/releases/download/audio8-tts-samples/audio8-tts-benchmark-data.tar.gz) (13 KB, [readme](https://github.com/NancyFyong/vllm-omni/releases/download/audio8-tts-samples/BENCHMARK_README.md)).

### Audio samples

Same 4 texts + 1 voice clone through both stacks — same weights, same sampling (`temp=0.7 / top_k=50 / top_p=0.9`), first output each time. Any audible difference isolates the serving stack, not the checkpoint. Voice-clone reference is Audio8's own earlier output (row 01), not a third-party recording. GitHub doesn't render inline audio for API-uploaded files, so these are release assets on my fork.

| # | Input | vllm-omni | HF reference |
|---|---|---|---|
| 01 | "The weather is nice today, perfect for a walk in the park." | [wav](https://github.com/NancyFyong/vllm-omni/releases/download/audio8-tts-samples/01_english.wav) 3.16 s | [wav](https://github.com/NancyFyong/vllm-omni/releases/download/audio8-tts-samples/hf_ref_01_english.wav) 3.30 s |
| 02 | "今天天气很好，很适合去公园散步。" | [wav](https://github.com/NancyFyong/vllm-omni/releases/download/audio8-tts-samples/02_chinese.wav) 3.30 s | [wav](https://github.com/NancyFyong/vllm-omni/releases/download/audio8-tts-samples/hf_ref_02_chinese.wav) 3.20 s |
| 03 | "今日はいい天気ですね、公園を散歩するのにぴったりです。" | [wav](https://github.com/NancyFyong/vllm-omni/releases/download/audio8-tts-samples/03_japanese.wav) 4.18 s | [wav](https://github.com/NancyFyong/vllm-omni/releases/download/audio8-tts-samples/hf_ref_03_japanese.wav) 4.27 s |
| 04 | two sentences describing the model itself | [wav](https://github.com/NancyFyong/vllm-omni/releases/download/audio8-tts-samples/04_english_long.wav) 12.26 s | [wav](https://github.com/NancyFyong/vllm-omni/releases/download/audio8-tts-samples/hf_ref_04_english_long.wav) 11.33 s |
| 05 | zero-shot voice clone, reference = 01 from vllm-omni | [wav](https://github.com/NancyFyong/vllm-omni/releases/download/audio8-tts-samples/05_voice_clone.wav) 5.15 s | [wav](https://github.com/NancyFyong/vllm-omni/releases/download/audio8-tts-samples/hf_ref_05_voice_clone.wav) 5.02 s |

Speech statistics fall in the same range across the two stacks (rms 0.06–0.21, zcr 0.05–0.14). Sampling is stochastic so single-run metrics differ run to run; please judge by ear, especially zh/ja.

## Drive-by: benchmark harness assumes 24 kHz

`vllm bench serve --omni` derives `audio_duration` from streamed-PCM byte count using `DEFAULT_AUDIO_SAMPLE_RATE = 24000`, so every 44.1 kHz model (this one, Fish Speech S2 Pro, Ming-omni-tts) reports RTF 1.84x too optimistic unless `VLLM_OMNI_BENCH_AUDIO_SAMPLE_RATE` is set, and `audio_underrun` / continuity are measured against inflated chunk budgets. `benchmarks/tts/README.md` now covers it with a sanity check. The same 24 kHz assumption sat in the test-side HNR check (above), so both are fixed here — can split either off if that's cleaner.

## Known gaps

- Greedy bit-exact comparison isn't possible: the reference with `do_sample=False` emits EOS immediately and produces 0 frames, so parity is per-component.
- Reproducing reference parity needs a workaround: with recent transformers the reference's meta-device init leaves its non-persistent RoPE buffers unmaterialised, so `freqs_cis` / `fast_freqs_cis` come back as NaN. Oracle scripts must recompute them with the module's `_precompute_rope` after loading.
- Voice-clone streaming and multi-language output were exercised manually, not in CI.
- `max_num_seqs=16` measurements were discarded (an unrelated job took the GPU mid-run, visible as single-request RTF degrading 2.2x). Raising stage-1 `max_num_seqs` to remove the c=8 underrun is untested.
- CI marks are `core_model` (L1/L2); L3/L4 tiers were not run locally.

### DIFF
--- a/examples/offline_inference/text_to_speech/audio8_tts/end2end.py
+++ b/examples/offline_inference/text_to_speech/audio8_tts/end2end.py
@@ -1,0 +1,237 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM-Omni project
+
+"""Offline inference demo for Audio8 TTS Preview 0.6B via vLLM Omni.
+
+Usage:
+    # Text-only synthesis
+    python end2end.py --text "Welcome to Audio8 TTS."
+
+    # Zero-shot voice cloning (the transcript must match the reference audio)
+    python end2end.py --text "Welcome to Audio8 TTS." \
+        --ref-audio reference.wav --ref-text "The exact transcript of the reference recording."
+
+    # Streaming (per-chunk TTFA / inter-chunk timings)
+    python end2end.py --text "Welcome to Audio8 TTS." --streaming
+"""
+
+import asyncio
+import os
+import time
+
+import numpy as np
+import soundfile as sf
+import torch
+
+os.environ["VLLM_WORKER_MULTIPROC_METHOD"] = "spawn"
+
+from vllm_omni import AsyncOmni, Omni
+from vllm_omni.model_executor.models.audio8_tts.codec_utils import estimate_reference_code_frames
+from vllm_omni.model_executor.models.audio8_tts.prompt_utils import (
+    build_text_only_prompt_ids,
+    estimate_voice_clone_prompt_len,
+    normalize_text,
+)
+from vllm_omni.utils.tracking_parser import TrackingArgumentParser
+
+DEFAULT_MODEL = "Audio8/Audio8-TTS-Preview-0.6b"
+
+
+def build_prompt(
+    text: str,
+    ref_audio_path: str | None = None,
+    ref_text: str | None = None,
+    model_name: str = DEFAULT_MODEL,
+) -> dict:
+    """Build an engine prompt, using the same protocol as online serving."""
+    from transformers import AutoTokenizer
+
+    tokenizer = AutoTokenizer.from_pretrained(model_name)
+    if ref_audio_path is None and ref_text is None:
+        prompt_ids, normalized_text = build_text_only_prompt_ids(tokenizer, text)
+        return {
+            "prompt_token_ids": prompt_ids,
+            "additional_information": {"text": [normalized_text]},
+        }
+
+    if not ref_audio_path or not ref_text:
+        raise ValueError("Audio8 TTS voice cloning requires both --ref-audio and --ref-text")
+
+    normalized_text = normalize_text(text)
+    normalized_ref_text = normalize_text(ref_text, add_default_speaker=True)
+    ref_audio_wav, ref_audio_sr = sf.read(ref_audio_path, dtype="float32", always_2d=False)
+    if ref_audio_wav.ndim > 1:
+        ref_audio_wav = ref_audio_wav.mean(axis=-1)
+    ref_frames = estimate_reference_code_frames(int(ref_audio_wav.shape[0]), int(ref_audio_sr))
+    placeholder_len = estimate_voice_clone_prompt_len(tokenizer, normalized_text, normalized_ref_text, ref_frames)
+
+    # The clone prompt embeds the reference audio's own codec codes, so the
+    # model builds it in preprocess(); we only reserve the exact length here.
+    return {
+        "prompt_token_ids": [1] * placeholder_len,
+        "additional_information": {
+            "text": normalized_text,
+            "ref_text": normalized_ref_text,
+            "ref_audio_wav": torch.from_numpy(np.asarray(ref_audio_wav, dtype=np.float32)),
+            "ref_audio_sr": int(ref_audio_sr),
+            "audio8_structured_voice_clone": True,
+        },
+    }
+
+
+def extract_audio(multimodal_output: dict) -> tuple[torch.Tensor, int]:
+    """Pull the waveform and sample rate out of a multimodal output.
+
+    The output processor concatenates the per-step delta tensors under
+    ``model_outputs``; ``audio`` is the older alias. Never use ``a or b`` on
+    these values -- truthiness on a tensor raises.
+    """
+    audio = multimodal_output.get("model_outputs")
+    if audio is None:
+        audio = multimodal_output.get("audio")
+    if audio is None:
+        raise ValueError(f"No audio in multimodal_output: {sorted(multimodal_output)}")
+    if isinstance(audio, list):
+        audio = torch.cat([torch.as_tensor(chunk).reshape(-1) for chunk in audio], dim=0)
+    audio = torch.as_tensor(audio).reshape(-1)
+
+    sr_raw = multimodal_output.get("sr")
+    if isinstance(sr_raw, list):
+        sr_raw = sr_raw[-1] if sr_raw else None
+    if sr_raw is None:
+        raise ValueError("Missing sample rate in multimodal_output")
+    sample_rate = int(sr_raw.item()) if hasattr(sr_raw, "item") else int(sr_raw)
+    return audio, sample_rate
+
+
+def _save_wav(output_dir: str, request_id: str, audio: torch.Tensor, sample_rate: int) -> None:
+    out_wav = os.path.join(output_dir, f"output_{request_id}.wav")
+    sf.write(out_wav, audio.float().cpu().numpy(), samplerate=sample_rate, format="WAV")
+    duration = audio.numel() / sample_rate
+    # print(), not logging: vLLM reconfigures logging on engine start and
+    # disables loggers created before that, which would swallow these lines.
+    print(f"Request {request_id}: saved {out_wav} (sr={sample_rate}, {duration:.2f}s)")
+
+
+def _omni_kwargs(args, model_name: str) -> dict:
+    kwargs = {
+        "model": model_name,
+        "log_stats": args.log_stats,
+        "stage_init_timeout": args.stage_init_timeout,
+    }
+    if args.deploy_config:
+        kwargs["deploy_config"] = args.deploy_config
+    return kwargs
+
+
+def main(args) -> None:
+    os.makedirs(args.output_dir, exist_ok=True)
+    model_name = args.model or DEFAULT_MODEL
+    prompts = [
+        build_prompt(text, args.ref_audio, args.ref_text, model_name) for text in ([args.text] * args.num_prompts)
+    ]
+
+    omni = Omni(**_omni_kwargs(args, model_name))
+    t_start = time.perf_counter()
+    total_audio_seconds = 0.0
+    for stage_outputs in omni.generate(prompts):
+        request_output = stage_outputs
+        if request_output is None or not request_output.outputs:
+            continue
+        audio, sample_rate = extract_audio(request_output.outputs[0].multimodal_output)
+        assert audio.numel() > 0, "Audio8 TTS produced an empty waveform"
+        total_audio_seconds += audio.numel() / sample_rate
+        _save_wav(args.output_dir, request_output.request_id, audio, sample_rate)
+    elapsed = time.perf_counter() - t_start
+    rtf = elapsed / max(total_audio_seconds, 1e-6)
+    print(f"Total {elapsed * 1000:.1f} ms for {total_audio_seconds:.2f} s of audio (RTF={rtf:.3f})")
+
+
+async def main_streaming(args) -> None:
+    os.makedirs(args.output_dir, exist_ok=True)
+    model_name = args.model or DEFAULT_MODEL
+    prompt = build_prompt(args.text, args.ref_audio, args.ref_text, model_name)
+
+    omni = AsyncOmni(**_omni_kwargs(args, model_name))
+    request_id = "0"
+    chunks: list[torch.Tensor] = []
+    sample_rate = None
+    t_start = time.perf_counter()
+    t_prev = t_start
+    chunk_index = 0
+
+    async for stage_output in omni.generate(prompt, request_id=request_id):
+        multimodal_output = stage_output.outputs[0].multimodal_output
+        if not multimodal_output:
+            continue
+        audio = multimodal_output.get("model_outputs")
+        if audio is None:
+            audio = multimodal_output.get("audio")
+        if audio is not None:
+            if isinstance(audio, list):
+                chunks.extend(torch.as_tensor(chunk).reshape(-1) for chunk in audio)
+            else:
+                chunks.append(torch.as_tensor(audio).reshape(-1))
+        sr_raw = multimodal_output.get("sr")
+        if isinstance(sr_raw, list):
+            sr_raw = sr_raw[-1] if sr_raw else None
+        if sr_raw is not None:
+            sample_rate = int(sr_raw.item()) if hasattr(sr_raw, "item") else int(sr_raw)
+
+        if stage_output.finished:
+            break
+        now = time.perf_counter()
+        if chunk_index == 0:
+            print(f"chunk 0: TTFA={(now - t_start) * 1000:.1f} ms")
+        else:
+            print(f"chunk {chunk_index}: inter_chunk={(now - t_prev) * 1000:.1f} ms")
+        t_prev = now
+        chunk_index += 1
+
+    if not chunks or sample_rate is None:
+        raise RuntimeError("Audio8 TTS streaming produced no audio")
+    _save_wav(args.output_dir, request_id, torch.cat(chunks, dim=0), sample_rate)
+    print(f"Streaming done: total={(time.perf_counter() - t_start) * 1000:.1f} ms chunks={chunk_index}")
+
+
+def parse_args():
+    parser = TrackingArgumentParser(description="Audio8 TTS Preview offline inference with vLLM Omni")
+    parser.add_argument("--model", type=str, default=DEFAULT_MODEL, help="Model path or HuggingFace repo ID.")
+    parser.add_argument(
+        "--text",
+        type=str,
+        default="Welcome to Audio8 TTS, a compact multilingual text to speech model.",
+        help="Text to synthesize.",
+    )
+    parser.add_argument("--ref-audio", type=str, default=None, help="Reference audio path for voice cloning.")
+    parser.add_argument(
+        "--ref-text",
+        type=str,
+        default=None,
+        help="Transcript of --ref-audio; required for voice cloning and must match the recording.",
+    )
+    parser.add_argument("--output-dir", type=str, default="output_audio", help="Directory for output WAV files.")
+    parser.add_argument(
+        "--num-prompts",
+        type=int,
+        default=1,
+        help="Duplicate the prompt N times to smoke-test concurrent requests.",
+    )
+    parser.add_argument(
+        "--deploy-config",
+        type=str,
+        default=None,
+        help="Override the deploy config path; defaults to vllm_omni/deploy/audio8_tts.yaml via model_type.",
+    )
+    parser.add_argument("--streaming", action="store_true", help="Use AsyncOmni and report per-chunk latency.")
+    parser.add_argument("--log-stats", action="store_true", help="Enable engine stats logging.")
+    parser.add_argument("--stage-init-timeout", type=int, default=600, help="Per-stage init timeout in seconds.")
+    return parser.parse_args()
+
+
+if __name__ == "__main__":
+    parsed_args = parse_args()
+    if parsed_args.streaming:
+        asyncio.run(main_streaming(parsed_args))
+    else:
+        main(parsed_args)
--- a/tests/e2e/offline_inference/test_audio8_tts.py
+++ b/tests/e2e/offline_inference/test_audio8_tts.py
@@ -1,0 +1,112 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM-Omni project
+"""E2E offline tests for Audio8 TTS Preview (DualAR, 44.1 kHz codec).
+
+The prompt is pre-tokenized here (not built from ``mm_processor_kwargs``)
+because Audio8 TTS shares Fish Speech's protocol: text-only prompts are plain
+token ids, and voice-clone prompts are a placeholder of the exact final length
+whose real embeddings are built model-side from the encoded reference audio.
+"""
+
+import os
+
+os.environ["VLLM_WORKER_MULTIPROC_METHOD"] = "spawn"
+
+import pytest
+
+from tests.helpers.mark import hardware_test
+from tests.helpers.runtime import OmniRunner, OmniRunnerHandler
+from tests.helpers.stage_config import get_deploy_config_path
+
+MODEL = os.environ.get("AUDIO8_TTS_MODEL_PATH", "Audio8/Audio8-TTS-Preview-0.6b")
+DEPLOY_CONFIG = get_deploy_config_path("audio8_tts.yaml")
+SAMPLE_RATE = 44100
+TEXT = "The weather is nice today, perfect for a walk in the park."
+
+# Four distinct inputs for the concurrency test: distinct text is required so a
+# leaked per-request codec buffer (which would make two requests decode to the
+# same audio) actually fails the pairwise-difference assertion.
+CONCURRENT_TEXTS = [
+    "The weather is nice today, perfect for a walk in the park.",
+    "She sold seashells by the seashore all through the summer.",
+    "Our train departs at a quarter past nine tomorrow morning.",
+    "He planted rows of tomatoes and basil behind the old house.",
+]
+
+# The checkpoint registers its own `arktts` config in vllm-omni, so remote code
+# must stay disabled (transformers would otherwise prefer the checkpoint's
+# auto_map and bypass the Qwen2-shaped view Qwen2Model needs).
+_OMNI_RUNNER_PARAM = (MODEL, DEPLOY_CONFIG, {"trust_remote_code": False})
+
+pytestmark = pytest.mark.parametrize("omni_runner", [_OMNI_RUNNER_PARAM], indirect=True)
+
+
+def _text_only_prompt(text: str = TEXT) -> dict:
+    from transformers import AutoTokenizer
+
+    from vllm_omni.model_executor.models.audio8_tts.prompt_utils import build_text_only_prompt_ids
+
+    tokenizer = AutoTokenizer.from_pretrained(MODEL)
+    prompt_ids, normalized_text = build_text_only_prompt_ids(tokenizer, text)
+    return {"prompt_token_ids": prompt_ids, "additional_information": {"text": [normalized_text]}}
+
+
+# At ``core_model`` (L1/L2) the harness patches every stage to
+# ``load_format: dummy``, so the AR runs on random weights: it never emits EOS
+# and always generates to ``max_tokens`` (1024 frames ~= 47.5 s). Duration
+# bounds are therefore only meaningful once real weights load, at
+# ``advanced_model`` / ``full_model``.
+_REAL_WEIGHT_RUN_LEVELS = frozenset({"advanced_model", "full_model"})
+
+
+def _duration_bounds(run_level: str) -> dict[str, float]:
+    if run_level not in _REAL_WEIGHT_RUN_LEVELS:
+        return {}
+    return {"min_duration_s": 0.5, "max_duration_s": 30.0}
+
+
+@pytest.mark.core_model
+@pytest.mark.advanced_model
+@pytest.mark.tts
+@hardware_test(res={"cuda": "L4"}, num_cards=1)
+def test_text_to_audio_001(omni_runner: OmniRunner, run_level: str) -> None:
+    """Default deploy, single request, text-only synthesis.
+
+    Deploy Setting: audio8_tts.yaml
+    Input Modal: text
+    Output Modal: audio
+    """
+    prompt = _text_only_prompt()
+    request_config = {
+        "input": TEXT,
+        "prompt_token_ids": prompt["prompt_token_ids"],
+        "additional_information": prompt["additional_information"],
+        "response_format": "wav",
+        "expected_sample_rate": SAMPLE_RATE,
+        **_duration_bounds(run_level),
+    }
+    OmniRunnerHandler(omni_runner).send_tokenized_tts_request(request_config)
+
+
+@pytest.mark.advanced_model
+@pytest.mark.tts
+@hardware_test(res={"cuda": "L4"}, num_cards=1)
+def test_text_to_audio_concurrent_002(omni_runner: OmniRunner, run_level: str) -> None:
+    """Four concurrent requests with distinct text: stage 0 runs ``max_num_seqs: 4``.
+
+    Per-request codec state is keyed by request id. The four inputs are
+    deliberately distinct and the decoded outputs are asserted pairwise
+    different, so leaked codec state (which would make two requests decode to
+    identical audio) fails here, and only here.
+    """
+    prompts = [_text_only_prompt(text) for text in CONCURRENT_TEXTS]
+    request_config = {
+        "input": CONCURRENT_TEXTS[0],
+        "prompt_token_ids": [p["prompt_token_ids"] for p in prompts],
+        "additional_information": [p["additional_information"] for p in prompts],
+        "response_format": "wav",
+        "expected_sample_rate": SAMPLE_RATE,
+        "assert_distinct_outputs": True,
+        **_duration_bounds(run_level),
+    }
+    OmniRunnerHandler(omni_runner).send_tokenized_tts_request(request_config)
--- a/tests/e2e/online_serving/test_audio8_tts.py
+++ b/tests/e2e/online_serving/test_audio8_tts.py
@@ -1,0 +1,166 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM-Omni project
+"""E2E online tests for Audio8 TTS Preview via /v1/audio/speech.
+
+Real model inference (no mocks): text-only synthesis, streaming, and zero-shot
+voice cloning. The reference audio is inlined as a base64 data URL because CI
+hosts cannot fetch external URLs.
+"""
+
+import os
+
+os.environ["VLLM_WORKER_MULTIPROC_METHOD"] = "spawn"
+
+import pytest
+
+from tests.helpers.mark import hardware_test
+from tests.helpers.media import get_asset_path
+from tests.helpers.runtime import OmniServerParams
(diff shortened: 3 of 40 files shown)


hsliuustc0106:
This PR touches **vllm_omni/model_executor/, examples/online_serving/, tests/model_executor/, vllm_omni/entrypoints/, .buildkite/cuda/** (41 files). Based on CODEOWNERS coverage of the changed files, the most-related reviewers appear to be:

> @NickCao @gcanlin @Sy0307

NancyFyong:
> This PR appears to belong to: **docs/design/module/benchmarking.md**, **docs/design/module/entrypoints.md**, **docs/design/module/input_output_modality_contracts.md**.
> 
> Module owners: @alex-jw-brooks @Bounty-hunter @linyueqian @NickCao @Sy0307 @amy-why-3459 @Gaohan123
> 
> @NancyFyong, please review your own changes and leave a short self-review comment describing what you checked. PRs without author self-review may not be assigned a reviewer.
> 


**Verified locally on H20 against current main.** 34 CPU unit + adapter tests, offline e2e at `core_model` and `advanced_model`, online e2e 5/5 at `advanced_model`, pre-commit clean on the diff range. Privacy grep on the diff and PR body: 0 hits. 5 vllm-omni + 5 HF-reference audio samples for the same prompts are linked side-by-side in the PR body.

Two 24 kHz-hardcoded bugs in the harness are fixed in this PR — happy to split either off.

linyueqian:
resolve conflicts please

Sy0307:
We need to accelerate this PR as this model is popular among the community. Please clean up and self-review, then I will review and merge it. Also cc @hsliuustc0106

tests/helpers/runtime.py
```diff
@@ -3105,6 +3105,107 @@ def send_audio_speech_request(self, request_config: dict[str, Any]) -> OmniRespo
...
+        e2e_latency = time.perf_counter() - start_time
+
+        assert len(outputs) == len(prompts), f"Expected {len(prompts)} outputs, got {len(outputs)}"
+        expected_sr = request_config.get("expected_sample_rate")
+        min_duration = request_config.get("min_duration_s")
+        max_duration = request_config.get("max_duration_s")
+
+        result = OmniResponse()
+        for index, stage_output in enumerate(outputs):
+            request_output = stage_output.request_output
```
linyueqian, line 3159: [blocking] `OmniRequestOutput` has no `request_output` attribute, so this raises `AttributeError` on every call.

It subclasses vLLM's `RequestOutput` and exposes `outputs` directly (`vllm_omni/outputs/__init__.py:106`). The two existing call sites in this file, lines 2943 and 3083, both go straight to `stage_output.outputs[0].multimodal_output`; `.request_output` appears only here in the whole harness.

Measured at this head: `tests/e2e/offline_inference/test_audio8_tts.py` is 2 failed at `--run-level core_model` and 2 failed at `--run-level advanced_model`. Dropping this line and using `stage_output.outputs[0]` directly gives 2 passed at both levels (88s / 73s). Worth noting `test-ready.yml` only runs the online file, so this would first have gone red at merge.

vllm_omni/model_executor/models/audio8_tts/audio8_tts_slow_ar.py
```diff
@@ -0,0 +1,753 @@
...
+                self._recent_local_ids(sampling_metadata.output_token_ids, num_reqs, device),
+                temperature=temperature,
+                top_k=top_k,
+                top_p=top_p,
+                ras_temperature=self._ras_temperature,
+                ras_top_p=self._ras_top_p,
+                num_semantic_ids=self._num_semantic_ids,
+                # One shared generator per batch: per-request generators would
+                # force a Python loop over requests.
+                generator=sampling_metadata.generators.get(0),
```
linyueqian, line 363: [important] Slot 0's generator drives sampling for the entire batch. `ras_sample_semantic` passes it into `gumbel_argmax_sample` over the full `[num_reqs, vocab]` tensor (`sampling.py:103` and `:111`), so one RNG stream produces the noise for every row.

Two consequences: a per-request `seed` is honored only for whichever request happens to occupy slot 0, and the same prompt yields different audio depending on which requests it is batched with. That undercuts the `seed: 42` in `deploy/audio8_tts.yaml`. If slot 0 is ever unseeded, `get(0)` returns `None` and the whole batch silently falls back to the global RNG.

`cosyvoice3.py:698` and `glm_tts.py:1150` both index per request. If you want to keep the fast path, gating the shared generator on `num_reqs == 1` and looping otherwise would avoid the Python loop in the common single-request case.

tests/e2e/offline_inference/test_audio8_tts.py
```diff
@@ -0,0 +1,99 @@
...
+def test_text_to_audio_concurrent_002(omni_runner: OmniRunner, run_level: str) -> None:
+    """Four concurrent requests: stage 0 runs ``max_num_seqs: 4``.
+
+    Per-request codec state is keyed by request id; a shared buffer shows up
+    here as truncation or crosstalk, and only here.
+    """
+    prompt = _text_only_prompt()
+    request_config = {
+        "input": TEXT,
+        "prompt_token_ids": [prompt["prompt_token_ids"]] * 4,
```
linyueqian, line 93: [important] These are four identical prompts, so this test cannot detect the crosstalk the docstring says it catches "and only here". If request 2's codec state leaked into request 3, both would still decode to the same audio and every assertion would pass.

`send_tokenized_tts_request` also asserts each response independently and never compares them, so nothing here would notice. Distinct text per request plus an assertion that the four outputs differ would make it real, and it would cover the shared-generator issue in `audio8_tts_slow_ar.py:363` as well.

vllm_omni/model_executor/stage_input_processors/audio8_tts.py
```diff
@@ -0,0 +1,162 @@
...
+        if pending < initial_chunk_frames and not finished:
+            return None
+        new_frames = min(pending, initial_chunk_frames)
+        left_context_size = max(0, length - new_frames)
+        window = frames[:length]
+    else:
+        initial_coverage = (
+            (chunk_frames // initial_chunk_frames) * initial_chunk_frames if initial_chunk_frames > 0 else 0
+        )
+        pending = (length - initial_coverage) % chunk_frames
```
linyueqian, line 137: [suggestion] Not blocking, and inherited rather than introduced here, but flagging it since this port carries it forward.

This branch has no re-emission guard, unlike the `already_sent` check in the branch above. When `pending == 0` and the terminal call arrives carrying no valid codec frame (`make_omni_output` returns `multimodal_outputs={}` at `audio8_tts_slow_ar.py:389`, and the all-zero heuristic in `extract_last_frame` also yields `None`), `length` is unchanged, so `new_frames` falls back to `chunk_frames` and the previous window ships a second time.

Replaying with your shipped config (chunk 25 / left 25 / initial 4) over stream lengths 1..129: frame counts 49, 74, 99 and 124 emit their last 25 frames twice, about 1.16 s of duplicated audio. Nothing is lost. Running the identical harness against `fish_speech.slow_ar_to_dac_decoder_async_chunk` reproduces it at exactly the same lengths, so it is a shared DualAR chunker issue. Happy to open a separate issue for both.

linyueqian (changes requested):
Ran the e2e suite on 1x H200-class GPU at `c3e0f7a9c` (vllm 0.27.0, torch 2.13.0, transformers 5.14.1). The model side looks good: online is 5/5 green, transcripts come back at cosine 1.000 for both English and the zero-shot clone, and streaming PCM scores 2.19 dB HNR at sr=44100. The 34 CPU unit + adapter tests match your numbers, and `check_tts_adapter.py` stays at exit 0, so the adapter ratchet claim holds.

The offline suite does not run at all, though. `send_tokenized_tts_request` reads an attribute that `OmniRequestOutput` does not have, so both offline tests fail at `core_model` and at `advanced_model` before any model code executes. Removing that one hop takes them to 2 passed at both levels. That path is only exercised by `test-merge.yml`, not by `test-ready.yml`, which is likely why it survived to here.

Two other things worth a look: the slow AR samples the whole batch from one generator, and the concurrency test sends four identical prompts so it cannot observe the crosstalk it documents.

On the async_chunk chunker, I replayed it against your shipped config and it duplicates the final 25 frames when the frame count is 24 mod 25 and the last step carries no codec frame. I ran the same harness against `fish_speech` and got byte-identical results, so that is inherited, not yours. Not blocking this PR, happy to file it separately.

Also needs a rebase, it is currently CONFLICTING.

tests/helpers/runtime.py
NancyFyong: Thank you for your reply, I'll fix it quickly

NancyFyong:
Hi, @linyueqian @Sy0307 
I Rebased onto main (single commit) and fixed the three:                            
                                                                                   
 - 1: the helper reads outputs[0].multimodal_output now — the stale               
   .request_output was the AttributeError.                                         
 - 2: sampling is per-request. num_reqs==1 keeps the batched fast path, otherwise 
   it loops and each row uses its own generator. Added a CPU test that pins each   
   row to its generator.                                                           
 - 3: the concurrency test sends 4 different texts and asserts the outputs        
   actually differ.                                                                
                                                                                   
 The async_chunk re-emission (4) reproduces on fish_speech too, so it's not       
 something this PR introduces — happy to file a separate issue for both.           
                                                                                   
 Also dropped the two Audio8 buildkite blocks since main removed the other         
 per-model TTS E2E blocks; the CPU/adapter tests ride the generic Simple steps now 
 (they're core_model + cpu).                                                       
                                                                                   
 36 CPU tests pass and offline e2e at advanced_model is green (2 passed) on 1x     
 H20. Online e2e still to run.

vllm_omni/model_executor/models/audio8_tts/audio8_tts_slow_ar.py
```diff
@@ -0,0 +1,753 @@
...
+        bsz = int(input_ids.shape[0])
+        device = input_embeds.device
+
+        token_ids = input_ids.reshape(bsz).to(dtype=torch.long, device=device)
+        past_hidden = last_talker_hidden.reshape(bsz, -1).to(dtype=torch.bfloat16, device=device)
+
+        temperature = kwargs.get("temperature")
+        temperature = DEFAULT_TEMPERATURE if temperature is None else float(temperature)
+        do_sample = kwargs.get("do_sample")
+        generator = kwargs.get("generator")
```
congw729, line 627: Non-blocking question: does the current talker_mtp path guarantee per-request seed semantics? Slow AR now routes sampling through each request generator, but Fast AR still accepts a single generator here, while the runner currently calls talker_mtp(ids, emb, hid, ts) without passing request generators or seeds. This appears to make residual codebook sampling use shared or global RNG, so the full codec output for a request may depend on batch composition or order even with a fixed seed. If per-request reproducibility is not supported by the common runner yet, this may be better documented and tracked as a shared follow-up rather than blocking Audio8. A regression test comparing the same prompt and seed standalone versus in a mixed batch would clarify the intended contract.

congw729 (commented):
Non-blocking suggestion: could we move the generic test-harness changes in `tests/helpers/assertions.py` and `tests/helpers/runtime.py` into a separate PR, or at least a clearly isolated commit?

The sample-rate fix and `send_tokenized_tts_request` are useful beyond Audio8, but they also change shared test infrastructure and make this already-large model integration harder to review and revert independently. A separate PR would keep the Audio8 diff focused and allow the harness behavior to be validated against existing TTS models.

That said, I would not block this PR on the split if creating another PR would materially delay the model integration; an isolated commit with dedicated harness tests would be a reasonable compromise.

NancyFyong:
Thanks @congw729 — both good catches.

**1. Fast AR seeding — fixed.** The Slow AR was already per-request via `sampling_metadata.generators`, but you're right that the residual codebooks go through `talker_mtp`, which reads `tts_local_seed`, and serving only set that for `qwen3_tts`. So an unseeded Audio8 request left the Fast AR on the global RNG and its full codec output could depend on batch composition.

The Audio8 adapter now defaults `tts_local_seed` from the stage-0 deploy seed in `apply_sampling_overrides` (the same thing serving does for `qwen3_tts`), so the runner's existing per-row path (`_talker_mtp_forward`'s scalar fallback / `talker_mtp_accepts_per_row_generators`) takes over and a request no longer depends on its batch mates. An explicit `request.seed` still overrides it. I kept it in the adapter rather than `serving_speech.py` so it doesn't add a `_tts_model_type` branch to the ratchet, and added a CPU regression test pinning the behavior.

**2. Splitting the harness changes.** I'd like to keep them here for now — Sy0307 asked to keep this a single commit and merge quickly — but I agree they're useful beyond Audio8. Happy to pull the `assertions.py` sample-rate fix and the `send_tokenized_tts_request` helper into a dedicated follow-up PR right after this lands. One note: upstream moved `OmniRunnerHandler` into `client.py` as `OfflineOmniClient` since I first wrote this, so on the rebase the helper now lives there (the `OmniRunnerHandler` alias is unchanged, so callers don't move).

Also rebased onto current main as a single clean commit, which clears the conflict.

NancyFyong:
Hi, @linyueqian @Sy0307 @congw729, I’ve updated this PR based on your suggestions. Could you take a look when you have a moment? Thank you!

vllm_omni/entrypoints/openai/serving_speech.py
```diff
@@ -2667,6 +2686,82 @@ def _build_fish_speech_prompt(
         prompt["additional_information"] = additional_information
         return prompt
 
+    # ---- Audio8 TTS helpers ----
```
sphinxkkkbc: Could you move the model-specific helpers and `self._audio8_tts_tokenizer` into `tts_adapters/audio8_tts.py`? PR #6186 (MiniMax Music 3) already follows this pattern. Some older models have not been fully migrated yet, but following their existing implementation would keep adding model-specific logic to `serving_speech.py`.

NancyFyong: done

examples/offline_inference/text_to_speech/audio8_tts/end2end.py
```diff
@@ -0,0 +1,237 @@
...
+    model_name = args.model or DEFAULT_MODEL
+    prompts = [
+        build_prompt(text, args.ref_audio, args.ref_text, model_name) for text in ([args.text] * args.num_prompts)
+    ]
+
+    omni = Omni(**_omni_kwargs(args, model_name))
+    t_start = time.perf_counter()
+    total_audio_seconds = 0.0
+    for stage_outputs in omni.generate(prompts):
+        request_output = stage_outputs.request_output
```
linyueqian: [important] Still unresolved at this head, and re-verified against current main rather than carried over. `omni.generate(...)` yields the request output directly, so there is no `.request_output` attribute to read and this line raises `AttributeError` before any audio is written. Every peer example in `examples/offline_inference/text_to_speech/` confirms the shape: fish_speech does `request_output = stage_outputs` and then reads `request_output.outputs[0]`, and moss_tts_nano iterates `for req_output in stage_outputs` directly. `.request_output` appears nowhere else in the TTS examples tree. Dropping the attribute access, so the loop variable is used as the request output, matches the convention and fixes the crash. The same access is repeated at line 164 in the streaming path.

vllm_omni/model_executor/models/audio8_tts/audio8_tts_slow_ar.py
```diff
@@ -0,0 +1,753 @@
...
+        if generator is None and seed is not None:
+            generator = torch.Generator(device=device)
+            generator.manual_seed(int(seed))
+
+        audio_codes = self.fast_ar(
+            slow_ar_hidden=past_hidden,
+            semantic_token_id=token_ids,
+            do_sample=(temperature >= SAMPLING_EPS) if do_sample is None else bool(do_sample),
+            temperature=temperature,
+            top_k=int(kwargs.get("top_k") or DEFAULT_TOP_K),
```
linyueqian: [suggestion] Still unresolved. `kwargs.get("top_k") or DEFAULT_TOP_K` treats an explicit `top_k=0` as unset and silently substitutes the default, so a caller asking for greedy selection gets sampling instead. Worth noting your own batched path at line 351 already gets this right with `DEFAULT_TOP_K if sampling_metadata.top_k is None else ...`, so the two paths currently disagree on what `top_k=0` means. An `is None` check here would make them consistent.

linyueqian (changes requested):
Re-reviewed at 1c6ddf06. The delta since the round 1 review is a merge of current main plus a single commit carrying the model work; neither round 1 finding has been addressed, so the verdict stays request_changes.

I re-derived both findings against this head rather than assuming they still applied, since main moved a long way underneath this branch. The example crash is confirmed by the surrounding convention rather than by the earlier note alone: every other end-to-end example in the TTS examples tree treats the object yielded by `omni.generate` as the request output itself, and the attribute this example reads appears in no other example there.

Nothing in the new merge changes the model code under review, and CI is green at this head.

Validation: static review only. I did not execute the fork head, load the checkpoint, or run the example, so the crash is established from the API shape and peer usage rather than from running it. The 44.1 kHz codec path and DualAR numerics remain unverified by me.

tests/e2e/online_serving/test_audio8_tts.py
```diff
@@ -0,0 +1,166 @@
...
+"""
+
+import os
+
+os.environ["VLLM_WORKER_MULTIPROC_METHOD"] = "spawn"
+
+import pytest
+
+from tests.helpers.mark import hardware_test
+from tests.helpers.media import load_test_audio_data_url
```
linyueqian: [blocking] This import does not resolve, so the whole module errors at collection: `ImportError: cannot import name 'load_test_audio_data_url' from 'tests.helpers.media'`. That helper is defined nowhere under `tests/helpers/`; the module exports `generate_synthetic_audio`, `generate_synthetic_video`, and their private helpers, but nothing by this name. Because it fails during collection rather than at run time, none of the tests in this file have ever executed, so the e2e coverage this PR adds is currently unexercised. Either add the helper to `tests/helpers/media.py` or build the data URL from `generate_synthetic_audio`.

vllm_omni/entrypoints/openai/tts_adapters/audio8_tts.py
```diff
@@ -0,0 +1,190 @@
...
+
+if TYPE_CHECKING:
+    from vllm_omni.entrypoints.openai.protocol.audio import OpenAICreateSpeechRequest
+
+
+@register_tts_adapter
+class Audio8TTSAdapter(ARTTSAdapter):
+    """Audio8 TTS Preview 0.6B: text-only synthesis and zero-shot voice cloning."""
+
+    stage_keys = frozenset({"audio8_tts_slow_ar"})
```
linyueqian, line 37: [blocking] Claiming this stage key fails an existing completeness test: `test_tts_detection.py::test_pipeline_stage_list_is_complete` reports `stage keys claimed by adapters but absent from _PIPELINE_STAGES: ['audio8_tts_slow_ar']`. That test exists precisely to catch a new model landing without registering its stage, and its `_PIPELINE_STAGES` list at `tests/entrypoints/openai_api/test_tts_detection.py:126` is not touched by this PR. Anchoring here because this is the declaration the PR owns; the edit itself belongs in that list.

linyueqian (changes requested):
Both round 2 findings are fixed, thank you. The example now treats the object yielded by `omni.generate` as the request output, matching the convention every other end-to-end example in the TTS tree follows, and the `.request_output` access is gone from the streaming path as well. The sampling defaults now use `is None` checks, so this path and the batched path at line 351 finally agree on what `top_k=0` means.

This PR had never run per-PR CI, though. It carried no `ready` label, so the only checks against it were the GitHub ones and the Buildkite lanes had produced nothing at any point. I applied the label rather than approving on the strength of the diff, and the general lane came back red on two failures that are both this branch's own and neither of which had a chance to surface before now. They are inline.

Neither is deep. One is a missing test helper that stops the new e2e module from importing at all, and the other is a stage key registered in the adapter but not in the list a dedicated test checks for exactly this situation. What they have in common is that they are invisible until the lanes run, which is the argument for getting the label on early rather than at review time.

For completeness on the other lanes: AMD and Intel fail the same two items as the general lane, and NPU passes here, which is the reverse of the usual pattern on this repo and worth noting only so nobody reads the NPU green as the meaningful signal.

Validation: static review of the delta plus the CI logs for general build 14391. I did not run anything locally, and the e2e file still has no successful execution anywhere, so its content remains unverified beyond importing cleanly once the helper exists.

vllm_omni/entrypoints/openai/tts_adapters/audio8_tts.py
NancyFyong: done

tests/e2e/online_serving/test_audio8_tts.py
NancyFyong: done

vllm_omni/model_executor/models/audio8_tts/audio8_tts_slow_ar.py
```diff
@@ -0,0 +1,755 @@
...
+            yield "model.norm.weight", tensor
+            continue
+
+        yield name, tensor
+
+
+class Audio8TTSSlowARForConditionalGeneration(nn.Module):
+    """Stage 0: text -> semantic tokens + residual codec codes."""
+
+    prefer_model_sampler = True
```
linyueqian, line 142: [blocking] This `prefer_model_sampler = True` declaration is what turns the general and AMD lanes red at 143a17e7. `tests/worker/test_gpu_ar_model_runner.py::TestPreferModelSamplerNoneFallback::test_in_tree_declarer_inventory_is_exact` pins the exact set of in-tree declarers as a deliberate consent gate, and `audio8_tts` is not in its `expected` set, so the sweep fails with "The set of models declaring `prefer_model_sampler` changed: added: ['audio8_tts']". The general lane (build 14396) and AMD (build 11198) fail on exactly this one assertion and nothing else PR-caused; both logs confirm it verbatim. The declaration itself looks intentional, `sample()` is typed `SamplerOutput | None` and has several `return None` decline sites, so the model genuinely relies on the documented None-fallback contract (RFC #5450 C7.3). The fix is the one the assertion message prescribes: add "audio8_tts" to the `expected` set in that test, which is the conscious sign-up the gate exists to force. After that, the remaining reds are inherited: NPU fails the same four MiniCPM-o and Qwen3-TTS steps it fails on unrelated PRs, and AMD Diffusion is the known FlashAttention tolerance flake.

linyueqian (changes requested):
Both round 3 findings are fixed at 143a17e7 and stay fixed: the e2e file imports get_asset_path, which is genuinely defined in tests/helpers/media.py, and audio8_tts_slow_ar is present in _PIPELINE_STAGES, with the entrypoints CI step now green accordingly.

The re-fired lanes surface one remaining PR-caused failure, and it is a mechanical one: the new Stage 0 model opts into the model-sampler contract, and the repo guards that opt-in with an exact-inventory consent gate that this PR has not signed. Details and the one-line remedy are inline. I read the general and AMD logs directly; both fail on that single assertion, 2868 other tests in the sweep pass, and every other red lane (NPU's four steps, AMD Diffusion) reproduces on unrelated PRs and predates this branch, so nothing else here is attributable to this PR.

Requesting changes only for the inventory signature. Once that lands this is approvable on my side: the model code itself has passed three rounds, and the None-decline sites in sample() are exactly what the contract the gate protects is for.

NancyFyong:
Done in e1b97241 — added `audio8_tts` to the `expected` set in `test_in_tree_declarer_inventory_is_exact`.

To be explicit that this is a real sign-up and not just quieting the test: the Slow AR does rely on the None-fallback. `sample()` hands the step back to the default sampler whenever logprobs, penalties or bad_words are requested, and only runs the RAS path when it can own the whole step — so opting into the C7.3 contract is intentional here.

While I was in there I rebased onto current main as a single commit, so the stray merge commit is gone as well. Left the NPU and AMD-Diffusion lanes alone since they're the inherited reds you already flagged.

linyueqian (approved):
The round 4 item is resolved: audio8_tts is signed into the prefer_model_sampler inventory at test_gpu_ar_model_runner.py, the declaration at audio8_tts_slow_ar.py:142 is unchanged, and the general lane (build 14409) that failed on exactly that assertion is green at this head, as is the AMD step that failed the same way.

Full lane picture at e1b97241, so nothing here blocks on a misread: general green; AMD build 11207 fails only the Diffusion step, the known main-side FlashAttention tolerance flake that predates this branch; NPU build 6178 fails the same four MiniCPM-o and Qwen3-TTS steps it fails on unrelated PRs; Intel build 7614 died in the Docker image build fetching pytest-mock from the USTC mirror (connection refused, zero tests ran), an infra outage worth a plain retry from the Buildkite side.

Approving. Across five rounds every finding was fixed as prescribed: the streaming request_output regression, the top_k semantics, the missing test helper import, the pipeline stage key, and now the sampler consent gate. The model code has been stable since round 2; the last three rounds were all consent and wiring, and the runs behind them are green.

NancyFyong:
Thanks — fair flag. Short version: this is a new-model integration PR, not a perf PR. It ships no performance change; the deploy default is unchanged (stage-0 `max_num_seqs: 4`). The table is operator tuning guidance, and I've reworded the README so it no longer reads as a measured speedup (dropped the bold "+27% / 4.5x", added that it's a single run on a contended GPU).

On your three questions:

1. **Bottleneck.** From single-request torch-profiler traces of the two stages on 1x H20: Stage 0 (Slow + Fast AR decode loop) is host-sync / launch-overhead bound, ~50% GPU idle — dominated by per-step D2H copies (`aten::copy_` was the top CPU op across ~3.4k calls) and many tiny kernels, not GEMM/attention. Stage 1 (codec decoder) is ~88% GPU idle and only serializes under concurrency because its deploy `max_num_seqs: 1`. That split is exactly why raising **stage-0** `max_num_seqs` helps admission at client c=8 while **stage-1** stays the continuity bottleneck.

2. **Value.** Only for an operator running client concurrency above the stage-0 limit: at c=8, raising 4→8 admits all 8 requests instead of queueing half, trading a rare sub-300 ms buffer gap (continuity 95% vs 100%) for higher throughput and lower first-packet latency. It's documented as a knob, not enabled by default.

3. **A/B / ablation.** The table is the ablation — only stage-0 `max_num_seqs` varies (4 vs 8), stage 1 fixed at 1, same 20 `seed_tts_smoke/en` prompts, 2 warmups. Its weakness is exactly what you'd expect me to call out: it's a single run on a GPU shared with an unrelated job (~80% SM), no repeats, so no spread. That's why I've demoted it to directional guidance rather than a claim. If it's worth a clean re-measure on an idle GPU with repeats + SHAs, I'm happy to do that as a follow-up, but it doesn't gate this model PR since nothing here changes the defaults.

PR #6162, opened 2026-08-13
[diffusion][feature] Add offline SVDQuant W4A4 support

lishunyang12:
## Summary

Add reusable offline SVDQuant W4A4 checkpoint loading and execution for
vLLM-Omni diffusion models. MiniMax-H3 FL2VA is the first validated consumer,
and the quantization config plus linear method remain model-independent.

This PR now also covers the MiniMax-H3 encoder fallback path needed to make
`text_encoder: fp8` work on top of the existing checkpoint transformer
quantization, without overwriting the checkpoint's transformer-side `svdquant`
settings.

The loader:

- detects `quant_method: svdquant` from transformer checkpoint metadata;
- loads packed NVFP4 weights, block scales, smoothing factors, and a BF16
  low-rank correction;
- supports optional per-output scales, including fused Q/K/V projections;
- preserves vLLM Column, Row, and QKV tensor-parallel sharding;
- reuses compatible vLLM NVFP4 GEMM providers.

## Runtime contract

- Offline checkpoint only; no calibration or conversion at model load.
- BF16 inputs with NVFP4 W4A4 base GEMM and BF16 rank correction.
- SM103 is the currently validated and enabled target.
- FlashInfer, CUTLASS, and FBGEMM NVFP4 tensor layouts are accepted.
- A forced backend with a different tensor ABI fails during model loading with
  an actionable error instead of failing on the first inference request.
- Each tensor-parallel input partition must be divisible by the NVFP4 block
  size of 16.

The checkpoint schema and tensor shapes are documented in
`docs/user_guide/quantization/svdquant.md`.

## Validation

- `16 passed`: MiniMax-H3 quantization regression suite, including the new
  transformer checkpoint fallback for `text_encoder: fp8`.
- `129 passed`: MiniMax-H3 contract suite.
- `git diff --check`: clean.
- B300 e2e measurement on MiniMax-H3 FL2VA:
  - BF16 baseline: 36.84 s, 69,838 MiB peak
  - `text_encoder: fp8` on top of checkpoint transformer `svdquant`:
    28.28 s, 48,158 MiB peak
  - speedup: 1.30x
  - peak memory reduction: 31.0%

This leaves the loader-only compatibility path intact, while also recording the
current encoder-FP8 result that is usable on B300.


hsliuustc0106:
This PR touches **vllm_omni/quantization/, tests/diffusion/, benchmarks/diffusion/, examples/offline_inference/, vllm_omni/diffusion/** (12 files). Based on CODEOWNERS coverage of the changed files, the most-related reviewers appear to be:

> @Isotr0py @david6666666 @NickCao

vllm_omni/quantization/component_config.py
```diff
@@ -103,7 +104,12 @@ def get_name(self) -> str:
     def get_quant_method(self, layer: torch.nn.Module, prefix: str) -> QuantizeMethodBase | None:
         config = self.resolve(prefix)
         if config is None:
-            return None
+            # `None` in the component dict means "this component is
+            # unquantized" (per the class docstring's `"vae": None`
+            # example). vLLM's `LinearBase.__init__` treats a `None`
+            # return as "linear doesn't support this quant method" and
+            # raises; instead surface an explicit unquantized method.
+            return UnquantizedLinearMethod()
```
hsliuustc0106: This return is layer-type-blind and breaks embedding consumers of `ComponentQuantizationConfig`. `get_quant_method` on a CQC does reach non-linear layers: `voxtral_tts_audio_generation.py:916-941` passes the CQC itself into `init_vllm_registered_model`, so `language_model.model.embed_tokens` (`VocabParallelEmbedding`) and `language_model.lm_head` (`ParallelLMHead`) call this with a prefix that resolves to `None` whenever the user's component dict doesn't cover `language_model`. Upstream's `VocabParallelEmbedding.__init__` only falls back to `UnquantizedEmbeddingMethod()` on `None`; with an `UnquantizedLinearMethod` returned instead, the exact-type embedding check raises `NotImplementedError: UnquantizedLinearMethod must implement the 'embedding' method` at construction, and `ParallelLMHead` skips that check only to hit an argument-shifted `create_weights` (embedding signature vs linear signature) with a confusing failure. This also flips `tests/diffusion/quantization/test_component_routing.py::test_get_quant_method_returns_none_for_unmatched`, which still asserts `is None` and isn't updated in this PR. Smallest fix: return `UnquantizedLinearMethod()` only when `isinstance(layer, LinearBase)`, keep `None` for everything else so upstream's per-layer fallbacks stay in charge.

vllm_omni/quantization/svdquant_config.py
```diff
@@ -0,0 +1,424 @@
...
+        return (
+            f"DiffusionSVDQuantConfig(rank={self.rank}, precision={self.precision!r}, act_unsigned={self.act_unsigned})"
+        )
+
+    @classmethod
+    def get_name(cls) -> QuantizationMethods:
+        return "svdquant"
+
+    @classmethod
+    def get_supported_act_dtypes(cls) -> list[torch.dtype]:
```
hsliuustc0106, line 129: `get_supported_act_dtypes` declares fp16, but every nvfp4 path assumes BF16 activations: the fused operator raises per-forward (`svdquant_flashinfer.py:221`), the compatible path dies on `torch.mm(x_2d, layer.proj_down)` — fp16 activations against the bf16 LoRA matrices (`svdquant_flashinfer.py:261`) — and the nvfp4 Nunchaku path also locks `lora_dtype` to bf16. int4 is fine (LoRA inherits `params_dtype`). So an fp16 run of an nvfp4 checkpoint only fails at the first forward, and outside the fused path with an unrelated dtype error. Since the classmethod can't see `precision`, the simplest contract is to cast activations to bf16 at `apply()` entry for nvfp4, or reject fp16 early when `precision == "nvfp4"`.

vllm_omni/quantization/svdquant_flashinfer.py
```diff
@@ -0,0 +1,680 @@
...
+        layer.output_channel_scale = None
+
+    _nvfp4_kernel().process_weights_after_loading(layer)
+
+    layer._svdquant_use_fused_flashinfer = _can_use_fused_svdquant(layer)
+    if layer._svdquant_use_fused_flashinfer:
+        logger.info_once("SVDQuant W4A4 is using FlashInfer's native fused residual-GEMM and rank-correction kernel.")
+        channel_scale = getattr(layer, "output_channel_scale", None)
+        if channel_scale is not None:
+            epilogue_alpha = (layer.alpha.float() * channel_scale.detach().float().reshape(-1)).contiguous()
```
hsliuustc0106: Evidence gap rather than a code issue: the PR's E2E numbers were produced at 14f0f05a, one commit before head, and this commit both adds the per-output-alpha fold and swaps the H3 MLP to `SiluAndMul`. `test_native_fused_svdquant_contract` validates the folded arguments (alpha ⊙ channel_scale, 1/alpha on `proj_up`) against a fake kernel, so a semantic mismatch inside FlashInfer's vector-alpha epilogue itself would still pass, and the compat-path numerical test can't see it either. Before marking ready, could you rerun the E2E quality comparison at head, or add a real-kernel fused-vs-compat comparison for a channel-scaled (QKV-shaped) layer?

lishunyang12:
Author self-review after the H3-only rewrite:\n\n- Rebased the branch onto current `main` and collapsed it to one self-contained MiniMax-H3 SVDQuant commit.\n- Audited both the local patch and GitHub file list: no Z-Image model, example, benchmark, converter branch, CLI option, or documentation entry remains in the PR diff. Generic SVDQuant runtime files are retained only because MiniMax-H3 exercises them.\n- Ran the complete changed-file pre-commit suite: passed.\n- Reran converter/config/FlashInfer/H3-contract/encoder-routing tests on the final commit: 212 passed, 4 skipped.\n- DCO, pre-commit, and Python 3.11/3.12 wheel builds are green; ReadTheDocs is still pending at the time of this review.

vllm_omni/quantization/svdquant_flashinfer.py
hsliuustc0106: The fused path is gone at `c06c286b`, but the validation gap remains. The new numerical test substitutes `_FakeNvfp4Kernel`, and `test_minimax_h3_quantization_quality.py` still configures FP8, so the reported 50-step SVDQuant smoke only proves that the real path completes. Please provide a frozen-head BF16 vs SVDQuant run on B300 with identical inputs and seeds, reporting video LPIPS/PSNR/MAE, audio spectral cosine/RMS, latency, and peak VRAM (1 warmup + at least 3 measured runs).

docs/user_guide/quantization/svdquant.md
```diff
@@ -0,0 +1,50 @@
+# SVDQuant W4A4
```
hsliuustc0106, line 1: update the feature design doc as well

hsliuustc0106:
Reviewed `c06c286b`. I don't see a new code-level blocker: the implementation fits the current quantization architecture by adding an Omni factory override and vLLM `QuantizationConfig`/`LinearMethod`, while reusing the existing NVFP4 kernel ABI and preserving Column/Row/QKV TP loading.

I would still hold approval until the existing frozen-head BF16-vs-SVDQuant quality/performance request is addressed. Please also update both `docs/design/feature/quantization.md` and `docs/user_guide/quantization/overview.md` so the backend list, hardware matrix, model support, and component scope include SVDQuant/MiniMax-H3.

hsliuustc0106 (approved):
lgtm

Issue #6144, opened 2026-08-13
[Bug][NPU][CI Env]: Nightly / CI failed - MiniCPM-o 4.5 Accuracy Test - seed-TTS whisper ASR eval model fails to load on NPU node, all 1088 samples 'ASR / WER failed' (env issue, reproduces on retry)


congw729:
### Your current environment

NPU nightly CI, main branch, commit 5358f5db01e5 (build [vllm-omni-npu-ci #5291](https://buildkite.com/vllm/vllm-omni-npu-ci/builds/5291), scheduled nightly 2026-08-12 18:00 UTC).

### 🐛 Describe the bug

`Omni · MiniCPM-o 4.5 · Accuracy Test` fails in both the first attempt and a manual retry (2026-08-13 02:00 UTC) with the same root cause: the whisper ASR model used by the seed-TTS WER evaluation never loads on the NPU node, so every sample fails ASR and the accuracy gate trips.

```text
ERROR [seed_tts_eval.py:730]   File ".../vllm_omni/benchmarks/data_modules/seed_tts_eval.py", line 462, in _transcribe_en_f32_16k
ERROR [seed_tts_eval.py:730]     assert _en_processor is not None and _en_model is not None and _device is not None
ERROR [seed_tts_eval.py:730] AssertionError

===== Seed-TTS eval (seed-tts-eval protocol) =====
Evaluated (WER, lower is better):        0
Request failed:                          0
No PCM captured:                         0
ASR / WER failed:                        1088
==================================================

ACCURACY CHECK FAILED:
  - [Seed-TTS] seed_tts_content_evaluated is 0 (enable --seed-tts-wer-eval and check PCM capture).
```

Failed tests (identical in both attempts):

```text
FAILED tests/e2e/accuracy/minicpmo_4_5/test_minicpmo_4_5.py::test_minicpmo_4_5_daily_omni_accuracy_bench[omni_server0] - subprocess.CalledProcessError: Command '['/usr/local/python3.12.13/bin/vllm', 'bench', 'serve', ...]'
FAILED tests/e2e/accuracy/minicpmo_4_5/test_minicpmo_4_5.py::test_minicpmo_4_5_seed_tts_wer_bench[omni_server0] - assert 1 == 0
```

[...]

Issue #6143, opened 2026-08-13
[Bug][NPU]: Nightly / CI failed - tests/e2e/online_serving/test_minicpmo_4_5_duplex.py::test_duplex_single_session_response_required[three-stage-single-gpu] - TimeoutError waiting for response.created (fails on retry too)


congw729:
### Your current environment

NPU nightly CI, main branch, commit 5358f5db01e5 (build [vllm-omni-npu-ci #5291](https://buildkite.com/vllm/vllm-omni-npu-ci/builds/5291), scheduled nightly 2026-08-12 18:00 UTC).

### 🐛 Describe the bug

`Omni · MiniCPM-o 4.5 · Duplex Test` fails on the same test in both the first attempt and a manual retry (2026-08-13 02:00 UTC), so this is deterministic rather than flaky:

```text
FAILED tests/e2e/online_serving/test_minicpmo_4_5_duplex.py::test_duplex_single_session_response_required[three-stage-single-gpu] - TimeoutError: Timed out waiting for  response.created
1 failed, 3 passed, 1 deselected, 15 warnings in 818.59s (0:13:38)
```

First attempt (18:14 UTC) and retry (02:00 UTC) show the identical failure: the duplex session never receives the `response.created` event within the timeout, while the other 3 tests in the job pass.

Note the build includes commit 7cd826d2 (#6095, MiniCPM-o duplex alignment fix), so the failure persists after that change. The CUDA counterpart job fails on a different test (`test_duplex_soft_interrupt`, tracked in #5962).

### Links

- Build: https://buildkite.com/vllm/vllm-omni-npu-ci/builds/5291
- First attempt job: https://buildkite.com/vllm/vllm-omni-npu-ci/builds/5291#019ff72d-3d0b-43f1-b4bc-eff51937f725
- Retry job: https://buildkite.com/vllm/vllm-omni-npu-ci/builds/5291#019ff8d8-9930-4fb5-8d4c-1c00112a8b16

congw729:
Update from the 2026-08-14 NPU nightly report (vllm-omni commit [90a08c0edaa1](https://github.com/vllm-project/vllm-omni/commit/90a08c0edaa15eda9f0b7bc740fd226703b77470)): the same single-session duplex timeout is still reproducible in the next nightly.

- Build: https://buildkite.com/vllm/vllm-omni-npu-ci/builds/5332
[...]

Issue #6067, opened 2026-08-11
[Bug]: Qwen3-Omni seed-TTS — model speaks the task description / answers the prompt instead of reading it; ~2x worse on the vLLM 0.27 rebase


tzhouam:
## Summary

The nightly `Omni · Accuracy Test` (`test_qwen3_omni_seed_tts_wer_bench`) fails on `dev/vllm-align` (the vLLM 0.27 rebase, PR #5976) with mean WER ~0.38–0.51 against the 0.35 gate, while `main` scores 0.27–0.29.

I reproduced the gap locally as a controlled A/B and inspected every synthesized utterance. **The failures are not audio corruption — Qwen3-Omni stops behaving as a TTS engine for a subset of prompts and instead speaks the task description, or answers the reference sentence, or babbles.** Because the WER used by the gate is uncapped, a few such clips dominate the mean.

Importantly, **this failure mode already exists on `main`; the rebase roughly doubles its frequency.** So there are two problems: a latent instruction-following bug (product) and a real regression on the rebase branch (blocking PR #5976).

## Reproduction (local, controlled A/B)

Same box (2 GPUs), same dataset (`zhaochenyang20/seed-tts-eval`, en, 1088 utterances), same harness invocation, same deploy config `vllm_omni/deploy/qwen3_omni_moe.yaml`, `--seed-tts-wer-save-items` + `SEED_TTS_WER_SAVE_AUDIO_DIR` so every utterance is scored and kept:

| stack | vLLM / torch | mean WER | median | clips with WER>1.0 |
|---|---|---|---|---|
| `origin/main` | 0.26.0 / 2.11 | **0.3047** | 0.0000 | 89 |
| `dev/vllm-align` @ `25b80bb9` | 0.27.0 / 2.13 | **0.3821** | 0.0000 | 151 |

Per-utterance comparison over the same 1088 prompts:

- **80** of main's 89 broken clips are broken on dev too (a stable hard-clip set)
- **71** clips regress (fine on main → broken on dev); only **9** improve — an 8:1 asymmetry, not a symmetric numerics reshuffle
- net **+0.0774** mean WER per clip

CI shows the same direction and a slightly larger gap (main 0.2665/0.2773/0.2788/0.2932 vs dev 0.3756/0.3826/0.3916/0.5114), so this is not specific to my hardware (I ran on L20X; CI uses H100).

## What the model actually does

[...]

PR #5754, opened 2026-08-04
[Ascend][MiniCPM-o] Add competition baseline and public proxy benchmarks

gerayking:
## Summary

- add the single-card Ascend 910C MiniCPM-o 4.5 deployment baseline and competition workflow
- add reproducible correctness, performance, profiling, Daily-Omni, Seed-TTS, and Video-MME runners
- add a deterministic 400-sample public proxy benchmark with machine-readable summaries
- document measured Ascend baseline and public-data proxy results

## Public proxy validation

- 400/400 requests completed at concurrency 1
- Daily-Omni: 73/100 on the fixed public sample
- Seed-TTS English: 100/100 generated; median TTFP 1.490 s; median RTF 0.604
- Seed-TTS Chinese: 100/100 generated; median TTFP 1.563 s; median RTF 0.584
- Video-MME: 65/100 on the stratified available public subset

These values are public-data proxy evidence, not official competition scores. The official Starter Kit, evaluator subsets, score weights, and final package schema were unavailable for this validation.

## Tests

- 28 competition/profile/public-dataset tests passed
- 65 scheduler, Code2Wav, CosyVoice2, StepAudio2, and NPU worker tests passed
- 13 focused competition and public benchmark tests passed after the final benchmark tooling update
- changed-file Ruff, compileall, Bash syntax, diff whitespace, and credential scans passed

PR #6170, opened 2026-08-13
[Bugfix][MiniCPM-o] Isolate duplex handoff failures and barge-in

### LINKED ISSUE #5962: [Bug]: Nightly / CI failed - tests/e2e/online_serving/test_minicpmo_4_5_duplex_expansion.py::test_duplex_soft_interrupt[three-stage-single-gpu] - AssertionError: {
### Your current environment

<details>

```text
ci env
```

</details>

### Your code version

<details>

```text
(not found in Buildkite step log)
```
</details>
<details>

```text
615de4e041d1
```
</details>

### 🐛 Describe the bug

**Buildkite build:** https://buildkite.com/vllm/vllm-omni/builds/13351
**Buildkite step:** [:full_moon: Omni · MiniCPM-o 4.5 Duplex Test](https://buildkite.com/vllm/vllm-omni/builds/13351#019fe7b2-5616-424a-b5d3-69e671217a58)

**Failure kind:** pytest FAILED
**Test node:** `tests/e2e/online_serving/test_minicpmo_4_5_duplex_expansion.py::test_duplex_soft_interrupt[three-stage-single-gpu]`

**Log reason:**
AssertionError: {

**Analysis:**
Category: assertion failure. Expected vs actual mismatch — check test logic, edge cases, and data setup.

**Error log excerpt:**

```text
omni_server = <tests.helpers.runtime.OmniServerStageCli object at 0x78404dff0d70>, model_prefix = ''

tmp_path = PosixPath('/tmp/pytest-of-root/pytest-0/test_duplex_soft_interrupt_thr0')



    @hardware_test(res={"cuda": "H100"}, num_cards=2)

    @pytest.mark.parametrize("omni_server", SERVER_PARAMS, indirect=True)

    def test_duplex_soft_interrupt(omni_server, model_prefix: str, tmp_path: Path) -> None:

        input_wav = validated_soft_interrupt_wav()

        result = asyncio.run(

            run_soft_interrupt(

                SimpleNamespace(

                    url=realtime_url(omni_server),

                    model=omni_server.model,

                    input_wav=str(input_wav),

                    ref_audio=str(resolve_ref_audio(model_prefix)),

                    output_dir=str(tmp_path / "soft_interrupt"),

                    summary_output=None,

                    chunk_ms=200,

                    timeout_s=180.0,

                    require_audio=True,

                    no_realtime_pacing=False,

      

...(truncated for GitHub URL length; see HTML report for full log)
```

---
*Generated from a nightly HTML report. Redact secrets before submitting; complete the checkboxes on GitHub.*

### LINKED ISSUE #6099: [Bug]: MiniCPM-o 4.5 native duplex crashes orchestrator when Thinker output has no latent/hidden_states
## Environment

| Item | Value |
| --- | --- |
| Model | `openbmb/MiniCPM-o-4_5` |
| vLLM-Omni | commit `e0d5f1ee6cdf973cf7371a2a555ee35a50c6c771` |
| vLLM | `0.26.0` |
| Python | `3.12.13` |
| PyTorch / CUDA | `2.11.0+cu130` / `13.0` |
| GPU | NVIDIA L20X |
| Deploy config | `vllm_omni/deploy/minicpmo_4_5_duplex.yaml` |

## Bug

In a MiniCPM-o 4.5 native full-duplex realtime session, Stage 0 can emit a control/listen/terminal output that has neither `multimodal_output["latent"]` nor `CompletionOutput.hidden_states`. The Thinker-to-Talker adapter currently treats this no-speech handoff as a fatal error:

```python
latent = mm_output.get("latent", None)
if latent is None:
    latent = output.hidden_states if hasattr(output, "hidden_states") else None
    if latent is None:
        raise ValueError("No latent or hidden_states found in thinker output")
```

The exception escapes `llm2tts`, terminates the shared stage runtime/orchestrator, and makes the service unavailable for every active session. A no-speech decision should not kill the server.

## Reproduction

1. Start the repository's native duplex profile:

   ```bash
   vllm-omni serve openbmb/MiniCPM-o-4_5 \
     --omni \
     --deploy-config vllm_omni/deploy/minicpmo_4_5_duplex.yaml \
     --trust-remote-code \
     --host 127.0.0.1 --port 8099
   ```

2. Connect to `ws://127.0.0.1:8099/v1/realtime?duplex=1&model=openbmb%2FMiniCPM-o-4_5&minicpmo45_native_duplex=1&autostart=0`.
3. Send `session.update` with `modalities: ["audio", "text"]` and `extra_body: {"auto_response": true, "minicpmo45_native_duplex": true}`.
4. Continuously append 16 kHz PCM16 audio in 200 ms chunks. Trigger a unit where the native duplex Thinker decides to keep listening (or otherwise emits no speech-conditioning payload). It is easier to reproduce while continuing microphone input across a response/cancel/terminal boundary.
5. Observe the exception below. The API health endpoint stops responding because the orchestrator exits.

The official client can be used as the base for the reproducer:

```bash
python examples/online_serving/minicpmo/realtime_duplex_demo.py \
  --url 'ws://127.0.0.1:8099/v1/realtime?duplex=1' \
  --input-wav /path/to/16khz-mono-pcm16.wav \
  --require-audio
```

## Observed traceback

Paths and process IDs are removed; the exception itself is unchanged.

```text
Traceback (most recent call last):
  ...
  File "vllm_omni/model_executor/stage_input_processors/minicpmo_4_5_omni.py", line 754, in llm2tts
    raise ValueError("No latent or hidden_states found in thinker output")
ValueError: No latent or hidden_states found in thinker output
```

After this exception, the shared orchestrator exits rather than failing or skipping only that handoff.

## Expected behavior

- If a native duplex output has no speech-conditioning payload because there is nothing for the Talker to synthesize, skip that downstream handoff and keep the session/server alive.
- If hidden states are missing for an output that is expected to speak, return a request-scoped error with context; do not terminate the shared orchestrator.

A minimal defensive behavior in `llm2tts` is to `continue` for a confirmed no-speech native-duplex handoff before dereferencing `latent`. The existing later guard (`if is_native_duplex_handoff and not handoff_ids: continue`) already follows the same idea, but it runs only after hidden states are required.

## Impact

One normal realtime control/listen/terminal output can take down all active duplex sessions and requires a full model-service restart.

- [x] Searched open and closed issues for the exact exception and related MiniCPM-o duplex/Thinker hidden-state failures; no duplicate was found.

### LINKED ISSUE #6101: [Bug]: MiniCPM-o 4.5 native duplex cannot reliably interrupt active responses
## 🐛 Describe the bug

MiniCPM-o 4.5 native duplex does not reliably stop the active response when the user starts speaking. The server currently declares:

```python
supports_model_native_turn_policy=True
supports_barge_in=False
target_barge_in_latency_ms=None
```

With `overlap_policy=barge_in_on_speech`, the serving layer defers overlapping input instead of immediately interrupting the active generation and playback. This breaks the expected full-duplex behavior of `/v1/realtime`: the user can speak over the assistant, but the current response continues or the new input is delayed.

The issue should be fixed in the server rather than requiring every client to implement its own unreliable RMS/VAD interruption logic.

## Expected behavior

1. Server-side VAD continues observing user audio while a response is active.
2. Confirmed user speech atomically fences and aborts every running stage of the active response.
3. No stale audio/text delta is emitted after cancellation; exactly one consistent terminal event is produced.
4. A short audio pre-roll is preserved so the interrupting utterance becomes the next turn without clipping its beginning.
5. Explicit `response.cancel` and VAD-triggered interruption use the same cancellation path.
6. The capability reports `supports_barge_in=True` with a measured `target_barge_in_latency_ms` after the behavior is implemented and tested.

## How to reproduce

1. Start a MiniCPM-o 4.5 native-duplex session through `/v1/realtime`.
2. Send continuous microphone audio and ask a question that produces a spoken response.
3. While the response is generating or playing, speak a second request.
4. Observe that the first response is not reliably stopped immediately; overlapping input is deferred because barge-in is marked unsupported.

## Suggested regression coverage

- Interrupt an active native-duplex response with user speech and verify exactly one cancellation.
- Verify no late asynchronous stage output revives the cancelled response.
- Verify the interrupting utterance is retained and produces the next response.
- Verify repeated interruptions do not leak requests, block the engine, or trigger `Duplex session pending input exceeds server limit`.
- Measure latency from server VAD confirmation to cancellation/last newly emitted delta.

## Environment and relevant code

Observed in the current implementation at commit `e0d5f1ee6cdf973cf7371a2a555ee35a50c6c771`:

- `vllm_omni/model_executor/models/minicpmo_4_5/duplex/capabilities.py`
- `vllm_omni/entrypoints/openai/duplex/serving.py`

Related discussions: #5644, #5179, #3745, and #5867. Issue #5867 focuses on turn-based models; native duplex currently bypasses that VAD path, so it does not fix this MiniCPM-o 4.5 failure.

- [x] I searched existing issues and documentation.

Sy0307:
## Purpose

MiniCPM-o 4.5 native duplex needs to keep one resumable Stage 0 stream across audio units, including while the model is speaking, while isolating control-only handoffs and optional hard barge-in from ordinary serving paths.

This PR:

- preserves Stage 0 KV/runtime state across native duplex units and feeds the previous sampled terminal token into the resumed unit;
- skips only confirmed empty/control-only native handoffs when speech conditioning is absent, while retaining errors for malformed speech handoffs;
- converts stage input-processor exceptions into request-scoped errors so one bad duplex request does not terminate the shared orchestrator;
- keeps `listen_only` as the default model-decided listen/speak mode;
- supports one explicit opt-in overlap mode, `barge_in_on_speech`, exposed through `turn_detection.type=server_vad` with `interrupt_response=true`;
- rejects `server_vad + interrupt_response=false` because a non-interrupting third mode is outside this design and does not implement a complete server-VAD turn lifecycle;
- makes Realtime VAD `session.update` transactional: a candidate detector is committed only after the session/runtime update is acknowledged, and is discarded on rejection;
- routes native speech overlap through epoch-fenced cancellation, retains interrupting PCM as next-epoch pre-roll, and filters stale output.

The ordinary non-duplex serving path is not unconditionally changed. The shared orchestrator change is limited to request-scoped failure isolation.

## Merge Status and Follow-ups (2026-09-05)

This PR merged into `main` on 2026-08-28 as `009b992c5b5d36acc55759dba7ed5acfee90d324`.

- Native MiniCPM Stage 0 continuation, scoped failure isolation, transactional VAD updates, and opt-in hard barge-in are landed.
- #6716 is closed after #6821 fixed playback ACK/history ordering; keep current-main H100/A100 soft-interrupt coverage as the regression gate.
- The threshold-boundary assertion remains a small post-merge test follow-up.
- Qwen3 turn-based VAD (#7055/#5867/#6618) is a separate compatibility/endpointing track, not this PR's native barge-in contract.

## Modes

1. `listen_only` (default): MiniCPM-o decides whether to listen or speak. No Silero model is created or invoked.
2. `barge_in_on_speech` (explicit opt-in): per-session Silero VAD confirms user speech and hard-cancels the active response.

There is no supported `server_vad + interrupt_response=false` third mode. Explicit `response.cancel` is a command, not a separate overlap mode.

## Test Plan

CPU and protocol coverage used for the H200 validation before the final test consolidation:

```bash
python -m pytest -q \
  tests/e2e/features/fullduplex/test_client.py \
  tests/e2e/online_serving/test_minicpmo_realtime_duplex_drivers.py \
  tests/engine/test_orchestrator_error_handling.py \
  tests/entrypoints/openai/test_duplex_protocol.py \
  tests/entrypoints/openai_api/test_duplex_handler.py \
  tests/model_executor/stage_input_processors/test_minicpmo_4_5_omni.py
```

Historical H200 result: `356 passed, 17 warnings in 26.58s`. The final cleanup consolidates the standalone server-VAD tests into `test_duplex_handler.py`; the current-head handler suite passes all 199 tests on H200, and the real-model lifecycle, soft-interrupt, and server-VAD hard-interrupt cases pass on GPU 6. Fresh CI is [Buildkite #13996](https://buildkite.com/vllm/vllm-omni/builds/13996): Engine&Entrypoints and both ready/nightly MiniCPM duplex jobs pass; the NPU MiniCPM duplex job also passes. Ruff format/check, Python compilation, YAML parsing, and `git diff --check` pass on the final head.

### Real-model hard barge-in acceptance

The reviewer-requested server-VAD path is now an automated L4 test:

```bash
CUDA_VISIBLE_DEVICES=6 \
python -m pytest -s -v \
  tests/e2e/online_serving/test_minicpmo_4_5_duplex_expansion.py::test_duplex_server_vad_hard_interrupt \
  --run-level full_model
```

H200 result: `1 passed, 15 warnings in 422.65s`. The public-event acceptance gate observed:

- exactly one `response.done(status=cancelled, reason=turn_detected)`;
- zero old-response audio/text deltas after that terminal fence;
- one `speech_started`, one `speech_stopped`, and a committed interrupting utterance;
- one completed follow-up response with 101 audio deltas;
- the follow-up transcript addresses the interrupting utterance's `四大发明` topic;
- zero errors and no synthetic `conversation.item.truncated`;
- 2.495 ms from server `speech_started` to cancelled `response.done` (server reaction time, not total acoustic-onset latency).

The existing byte-level regression also verifies that the 200 ms confirming pre-roll is retained before the remainder of the interrupting PCM in the new epoch.

### Default-mode regression/performance guard

The default `listen_only` real-model test was rerun on the same H200/GPU/input:

```bash
CUDA_VISIBLE_DEVICES=6 \
python -m pytest -s -v \
  tests/e2e/online_serving/test_minicpmo_4_5_duplex_expansion.py::test_duplex_soft_interrupt \
  --run-level full_model
```

Result: `1 passed, 15 warnings in 285.03s`, with 3 completed responses, 0 cancellations, 0 truncations, 12 listen decisions, and the same three deterministic transcripts as the saved pre-test-change run. The first response was created at 9.012s versus 8.474s in the saved run, while the final response completed at 20.354s versus 20.585s. A single cold run includes graph-capture and scheduling noise, so these values are evidence of stable behavior rather than a throughput claim.

This is not presented as a speedup claim. The final cleanup also rejects `threshold <= 0.15`, preserving a non-overlapping onset/silence hysteresis band; this is validation-only and adds no inference work. In the overall PR, Silero inference remains behind explicit `server_vad` opt-in and is lazy-loaded, so the default model path does not execute neural VAD work.

The L4 hard-interrupt test is collected by the existing MiniCPM duplex nightly step. Its driver is also included in the ready/merge source dependencies so edits trigger the standard duplex gate.

## Dependency note

#6274 independently adapts MiniCPM-o streaming audio cache handling to the Transformers 5.x cache API and removes the Code2Wav cold Dynamo compile regression. It is already in `main`; #6170 consumes that base behavior rather than carrying another copy.


hsliuustc0106:
This PR touches **vllm_omni/experimental/, tests/entrypoints/, tests/engine/, tests/model_executor/, vllm_omni/engine/** (10 files). Based on CODEOWNERS coverage of the changed files, the most-related reviewers appear to be:

> @Gaohan123 @NickCao @yenuo26

TaffyOfficial:
﻿Thanks for working on this. After comparing this PR with the `minicpm-challenge` branch, I think there are two different interruption semantics here.

For MiniCPM-o native duplex, I personally prefer keeping interruption model-owned by default. While the assistant is speaking, incoming audio units can continue to be appended to the existing Stage 0 resumable stream, and the model can decide whether to listen or speak through its native `<|listen|>` / `<|speak|>` decisions. A listen/control-only output should simply be treated as a successful no-op handoff rather than being forwarded to the Talker.

Conceptually:

```python
if native_auto_response:
    append_audio_unit_to_stage0(audio)

    decision = model_native_turn_decision()
    if decision == LISTEN:
        # No Talker handoff.
        return
    if decision == SPEAK:
        forward_hidden_states_to_talker()
```

This preserves the model's semantic understanding of whether the user is actually interrupting, acknowledging, or speaking in the background.

A server-side hard interruption is a different policy:

```python
if overlap_policy == BARGE_IN_ON_SPEECH and vad_detects_speech(audio):
    cancel_active_response()
    advance_epoch()
    retain_interrupting_audio_as_preroll()
```

That can be useful when deterministic cancellation is required, but it also introduces server-side VAD thresholds, cancellation/history semantics, and a behavior different from MiniCPM-o's native duplex policy. I think it would be better to get community alignment on that behavior, and possibly treat it as an explicit opt-in policy or a separate feature, rather than changing the default native-duplex semantics.

The control-only handoff handling and per-request orchestrator error isolation in this PR still look useful independently of the hard-barge-in policy.

Gaohan123:
> Thanks for working on this. After comparing this PR with the `minicpm-challenge` branch, I think there are two different interruption semantics here.
> 
> For MiniCPM-o native duplex, I personally prefer keeping interruption model-owned by default. While the assistant is speaking, incoming audio units can continue to be appended to the existing Stage 0 resumable stream, and the model can decide whether to listen or speak through its native `<|listen|>` / `<|speak|>` decisions. A listen/control-only output should simply be treated as a successful no-op handoff rather than being forwarded to the Talker.
> 
> Conceptually:
> 
> ```python
> if native_auto_response:
>     append_audio_unit_to_stage0(audio)
> 
>     decision = model_native_turn_decision()
>     if decision == LISTEN:
>         # No Talker handoff.
>         return
>     if decision == SPEAK:
>         forward_hidden_states_to_talker()
> ```
> 
> This preserves the model's semantic understanding of whether the user is actually interrupting, acknowledging, or speaking in the background.
> 
> A server-side hard interruption is a different policy:
> 
> ```python
> if overlap_policy == BARGE_IN_ON_SPEECH and vad_detects_speech(audio):
>     cancel_active_response()
>     advance_epoch()
>     retain_interrupting_audio_as_preroll()
> ```
> 
> That can be useful when deterministic cancellation is required, but it also introduces server-side VAD thresholds, cancellation/history semantics, and a behavior different from MiniCPM-o's native duplex policy. I think it would be better to get community alignment on that behavior, and possibly treat it as an explicit opt-in policy or a separate feature, rather than changing the default native-duplex semantics.
> 
> The control-only handoff handling and per-request orchestrator error isolation in this PR still look useful independently of the hard-barge-in policy.

Agreed. For MiniCPM-o 4.5 like models, which are native duplex. We should just use the first mode of native duplex. The second mode with hard barge-in should be set as optional and verify on some non full duplex models like Qwen3-Omni

Sy0307:
Self-review / CI triage for d80d7c2a:

- PR-scoped checks are green: pre-commit, DCO, Python 3.11/3.12 wheels, docs, Intel, and the CUDA MiniCPM-o core duplex job.
- CUDA full-model soft-interrupt failed because the model emitted one complete long response while the nightly assertion requires two. The same test fails on main nightly #13679 and is tracked by #5962.
- NPU duplex timed out waiting for the second-turn response.created. The identical failure is present on main nightlies #5291, #5332, #5371, and #5417 and is tracked by #6143.
- NPU MiniCPM accuracy is the existing ASR-evaluator initialization failure tracked by #6144.
- The other CUDA failure is Qwen3-Omni Seed-TTS WER (0.417924 > 0.35), tracked by #6067; it does not execute this PR's MiniCPM duplex path.
- AMD failed three LTX2 diffusion tests with rocm_unquantized_gemm dispatched on CPU. The exact failures are also present on the PR base commit in main AMD build #10628; this PR does not touch diffusion code.

I found no PR-introduced CI regression to patch. The branch is clean, mergeable, and still based on the current origin/main (baba7d1e).

Build references: CUDA PR #13691 / main #13679; NPU PR #5435 / main #5417; AMD PR #10635 / main #10628.

tests/e2e/online_serving/run_minicpmo_realtime_duplex_server_vad.py
```diff
@@ -0,0 +1,318 @@
+"""Live MiniCPM-o Realtime acceptance probe for the two overlap modes."""
+
+from __future__ import annotations
+
```
yenuo26, line 4: There already seem to be many related run_xxxx scripts in this directory, and run_minicpmo_realtime_duplex_demo_pair.py appears to have no other references except for test_minicpmo_realtime_duplex_drivers.py, which is specifically for testing this script. Actually, I think having tests for test scripts is somewhat redundant. Can we delete test_minicpmo_realtime_duplex_drivers.py and run_minicpmo_realtime_duplex_demo_pair.py?

TaffyOfficial:
The missing fix is the model-owned Stage 0 resume path, not another interruption policy.
Please restore the native duplex flow used by minicpm-challenge@4105c717:
Continue appending each prepared audio unit to the same resumable Stage 0 stream, including while the model is speaking.
When processing the next unit, preserve the Stage 0 KV/runtime state and feed back the previous unit’s sampled terminator. Do not create a new independent stream, cancel the active response, or advance the epoch.
If the native decision is <|listen|>, treat it as a successful control-only result: emit response.listen, skip the Talker/llm2tts handoff, and do not require hidden_states.
If the decision is <|speak|>, forward its hidden states to the Talker normally. A later <|speak|> after listening should start another naturally completed response.
In short:
unit = input_buffer.prepare_append(audio)
stage0.resume(
    unit,
    preserve_kv_state=True,
    previous_terminator=previous_sampled_terminator,
)

decision = stage0.native_decision
if decision == LISTEN:
    emit_response_listen()
    return  # successful no-op; no Talker handoff

forward_hidden_states_to_talker()
The current PR only handles the control-only no-op and adds optional server VAD, but it does not restore the resumable Stage 0 state handoff. Please compare the v0.27 scheduler/model-runner resume behavior with the v0.25 minicpm-challenge implementation.
The acceptance test should use the same interruption WAV and require:
completed_response_count >= 2
cancelled_count == 0
truncated_count == 0
response_listen_count >= 1
Server-VAD cancellation should remain a separate opt-in path and should not be used to make this native-duplex test pass.

BruceLoveDecimal:
> The missing fix is the model-owned Stage 0 resume path, not another interruption policy. Please restore the native duplex flow used by minicpm-challenge@4105c717: Continue appending each prepared audio unit to the same resumable Stage 0 stream, including while the model is speaking. When processing the next unit, preserve the Stage 0 KV/runtime state and feed back the previous unit’s sampled terminator. Do not create a new independent stream, cancel the active response, or advance the epoch. If the native decision is <|listen|>, treat it as a successful control-only result: emit response.listen, skip the Talker/llm2tts handoff, and do not require hidden_states. If the decision is <|speak|>, forward its hidden states to the Talker normally. A later <|speak|> after listening should start another naturally completed response. In short: unit = input_buffer.prepare_append(audio) stage0.resume( unit, preserve_kv_state=True, previous_terminator=previous_sampled_terminator, )
> 
> decision = stage0.native_decision if decision == LISTEN: emit_response_listen() return # successful no-op; no Talker handoff
> 
> forward_hidden_states_to_talker() The current PR only handles the control-only no-op and adds optional server VAD, but it does not restore the resumable Stage 0 state handoff. Please compare the v0.27 scheduler/model-runner resume behavior with the v0.25 minicpm-challenge implementation. The acceptance test should use the same interruption WAV and require: completed_response_count >= 2 cancelled_count == 0 truncated_count == 0 response_listen_count >= 1 Server-VAD cancellation should remain a separate opt-in path and should not be used to make this native-duplex test pass.

Hi @TaffyOfficial  I believe the assertion failed: completed_response_count >= 2 issue has a separate root cause. It’s unrelated to handoff failure handling or barge-in; instead, it appears to be caused by a performance regression. I’ll open a draft PR with a fix.

natureofnature:
Thanks for keeping model-owned listen/speak as the default and making hard barge-in explicitly opt-in.
The reported real-model E2E validates native Stage 0 continuation, but it has `cancelled_count = 0`, so it does not exercise the new Silero server-VAD hard-interruption path. Is it better to add a real-model test that verifies exactly one cancelled terminal, no post-fence stale deltas, preservation of the interrupting utterance, and successful subsequent responses? Otherwise, splitting server VAD into a separate PR may make the scope easier to review.

Sy0307:
> Thanks for keeping model-owned listen/speak as the default and making hard barge-in explicitly opt-in. The reported real-model E2E validates native Stage 0 continuation, but it has `cancelled_count = 0`, so it does not exercise the new Silero server-VAD hard-interruption path. Is it better to add a real-model test that verifies exactly one cancelled terminal, no post-fence stale deltas, preservation of the interrupting utterance, and successful subsequent responses? Otherwise, splitting server VAD into a separate PR may make the scope easier to review.

Agree with it.

Sy0307:
@natureofnature Thanks — this was a real coverage gap. I kept the opt-in server-VAD path in this PR and added an automated real-model H200 test that now gates the exact contract you described:

- exactly one cancelled terminal (`turn_detected`);
- no target-response delta after the terminal fence;
- `speech_started -> speech_stopped -> input_audio_buffer.committed`, plus the existing byte-level pre-roll regression;
- a completed follow-up response whose transcript addresses the interrupting `四大发明` utterance.

H200 result: `1 passed` in 422.65s; 1 cancelled terminal, 0 stale deltas, 1 committed interrupt, 1 completed follow-up with 101 audio deltas, 0 errors, and 2.495ms server reaction time.

I also narrowed the standalone server-VAD probe to this one acceptance scenario, wired it into the L4 MiniCPM duplex test and CI source dependencies, and kept Silero lazy/opt-in. The comment-driven patch adds no production Python changes relative to the prior PR tip. #6274 is now merged and this branch is updated to current main.

vllm_omni/experimental/fullduplex/openai/vad.py
```diff
@@ -0,0 +1,216 @@
...
+
+        started = False
+        stopped = False
+        start_ms: int | None = None
+        end_ms: int | None = None
+        max_probability = 0.0
+        contained_speech = self._speech_active
+        min_speech_samples = int(self.config.min_speech_duration_ms * self._SAMPLE_RATE_HZ / 1000)
+        min_silence_samples = int(self.config.silence_duration_ms * self._SAMPLE_RATE_HZ / 1000)
+        negative_threshold = max(0.0, self.config.threshold - self._NEGATIVE_THRESHOLD_GAP)
```
congw729, line 146: One small edge case: when threshold <= 0.15, negative_threshold becomes 0. Since probability is clamped to a non-negative value, the silence branch can never be reached after speech starts, so speech_stopped may never be emitted. Would it make sense to either restrict the accepted threshold range or keep the negative threshold above zero?

TaffyOfficial:
I've run the tests and the issue is resolved. LGTM.

TaffyOfficial:
@vllm-omni-review-bot

TaffyOfficial:
I agree with [congw729](https://github.com/congw729) . Considering the real‑world code modifications are approximately 600 lines, the test volume appears excessive.

Sy0307:
The new server-VAD validation accepts `threshold=0.1`, but `vad.py` computes `negative_threshold=max(0.0, threshold - 0.15)`, i.e. zero. Since probabilities are clamped to `[0, 1]`, silence can never satisfy the stop condition, so `speech_stopped` never fires. I reproduced this with the streaming VAD scorer on the H200 checkout. Please reject thresholds below the hysteresis gap or guarantee a positive negative threshold.

vllm_omni/experimental/fullduplex/openai/realtime_session.py
```diff
@@ -58,6 +58,7 @@ async def receive_internal_event_text(self, websocket: WebSocket) -> str:
         while True:
             if not self._pending_outbound.empty():
                 return json.dumps(await self._pending_outbound.get())
+            await self.wait_for_realtime_turn_detection_update()
```
linyueqian, line 61: This stalls the session when the runner drops a `session.update` without answering it.

The pending flag is cleared only by an outbound `session.created`/`session.updated`/`error` reaching `encode_outbound_event`. In `session_runner.py:1057`, `if not await wait_for_native_append_tail(): continue` emits nothing, so a mid-session `session.update` carrying `turn_detection` leaves `_turn_detection_update_pending` set and this line blocks every subsequent `websocket.receive_text()`.

Reproduced with a harness driving the protocol directly: open the session, stage a second `session.update`, emit nothing, then have the client keep sending.

```
FAILED test_mid_session_update_without_ack_stalls_the_reader
E   Failed: STALLED: receive_internal_event_text never returned.
```

The reader is wrapped in `asyncio.wait_for(..., idle_timeout_s)` in `serving.py:1263`, so it is not infinite. The effect is that a client still streaming audio gets nothing processed for `idle_timeout_s` (default 300s) and the session then dies on `__timeout__`.

The trigger needs a failed or cancelled in-flight native append concurrent with the update, so it is narrow, but the same `session.update` machinery is what this PR advertises as transactional and no test covers a mid-session update. Smallest fix: resolve the staged update on that `continue` as well, either by emitting the error or by calling `reject_realtime_turn_detection_update()` before continuing.

vllm_omni/experimental/fullduplex/openai/realtime_session.py
```diff
@@ -97,11 +98,22 @@ async def receive_internal_event_text(self, websocket: WebSocket) -> str:
...
         return payloads
 
+    def _resolve_realtime_turn_detection_update(self, data: dict[str, Any]) -> None:
+        """Publish or discard a staged VAD with the Realtime session ACK."""
+        if not self._turn_detection_update_pending:
+            return
+        event_type = data.get("type")
+        if event_type in {"session.created", "session.updated"}:
+            self.commit_realtime_turn_detection_update()
+        elif event_type == "error":
```
linyueqian: Any outbound `error` rejects the staged detector, with no correlation to the `session.update` that staged it.

An error already queued from an earlier operation (a `bad_audio` from a previous `input_audio_buffer.append`, or a concurrent response-task error) lands inside the pending window and discards a perfectly valid candidate. The runner then still emits `session.updated`, so the commit is a no-op:

```
after unrelated error:      pending=False  vad_is_previous=True
after real session.updated: pending=False  vad_is_previous=True
```

The result is a silent divergence: the session config reports the new `turn_detection` and `overlap_policy=barge_in_on_speech`, while the previous detector (or `None`) keeps running. The client believes server VAD is active and it is not, so barge-in falls back to the RMS heuristic or stops working.

Correlating on the staging event's `event_id` and resolving only on the response to that transaction would close both this and the stall above.

vllm_omni/experimental/fullduplex/openai/realtime_input.py
```diff
@@ -275,16 +283,62 @@ async def _to_duplex_event(self, event: dict[str, object]) -> dict[str, object]
...
+                else self._input_looks_like_speech(event, audio=audio, fmt=fmt)
+            )
             self._input_audio_buffer_has_audio = self._input_audio_buffer_has_audio or (
                 looks_like_speech and isinstance(audio, str) and bool(audio)
             )
             self._input_audio_buffer_had_non_speech = self._input_audio_buffer_had_non_speech or (
                 not looks_like_speech and isinstance(audio, str) and bool(audio)
             )
-            if looks_like_speech:
+            if vad_result is not None and vad_result.speech_started:
```
linyueqian, line 327: When one append carries both edges, the second onset is swallowed and the client's speech state desynchronizes.

`SileroStreamingVAD.process()` can return `speech_started=True` and `speech_stopped=True` from the same call when an append spans the end of one utterance and the start of the next. Measured on a single append:

```
started=True stopped=True active_after=True start_ms=3156 end_ms=2016
```

This block emits started first, but `_emit_input_speech_started` early-returns on the `_input_speech_started` guard at line 1410 because speech was already active. Then `_emit_input_speech_stopped` runs and sets `_input_speech_started = False`. The client receives only a `speech_stopped` carrying the *first* utterance's `audio_end_ms`, never learns the second utterance began, and since the flag is now `False` while the detector is `speech_active=True`, no later append re-emits it either.

Reaching it needs an append of roughly 700ms or more (500ms `silence_duration_ms` plus 96ms `min_speech_duration_ms`), which is under the 1000ms `chunk_period_ms` this model advertises in `protocol.py:131`.

Barge-in itself stays correct, since `_overlap_decision` reads `payload["vad"]` rather than the emitted events, so this is a wire-contract defect rather than a cancellation one. Emitting stopped before started, and not clearing `_input_speech_started` when `speech_active` is still true, would fix both the ordering and the lost onset.

vllm_omni/experimental/fullduplex/openai/vad.py
```diff
@@ -0,0 +1,218 @@
...
+            raw = base64.b64decode(audio, validate=True)
+        except (binascii.Error, ValueError) as exc:
+            raise ValueError("Silero server VAD received invalid base64 audio") from exc
+        if len(raw) % 4:
+            raise ValueError("Silero server VAD received an incomplete pcm_f32le frame")
+        samples = np.frombuffer(raw, dtype="<f4").astype(np.float32, copy=False)
+        rate = int(sample_rate_hz) if isinstance(sample_rate_hz, int | float) else self._SAMPLE_RATE_HZ
+        if rate <= 0:
+            raise ValueError("Silero server VAD requires a positive sample rate")
+        if rate != self._SAMPLE_RATE_HZ and samples.size > 1:
```
linyueqian: Minor: the `samples.size > 1` guard makes a one-sample append skip resampling entirely, so it is consumed as if it were already 16 kHz.

Measured at 48 kHz: 1024 one-sample appends carry 21.3ms of audio but advance the detector clock 64.0ms, 3x too fast, which skews onset debounce, silence hysteresis and the reported timestamps.

Only a pathological client does this, and normal chunks are fine. I measured a constant, non-accumulating -0.27% across 200ms / 100ms / 20.8ms / 6.9ms chunks at 48 kHz, so there is no drift problem for real traffic. Dropping the `> 1` condition and letting `target_size` clamp to a minimum of 1 would remove the edge case.

linyueqian (commented):
Validated on H200 against head `50ded5c71` (vLLM 0.27.0), PR head in a detached worktree.

**What reproduces.** The CPU suites you list pass (264 passed across `test_duplex_handler.py`, `test_duplex_protocol.py`, `test_orchestrator_error_handling.py`, `test_minicpmo_4_5_omni.py`, `test_client.py`). Both real-model runs pass at `--run-level full_model` on one H200: `test_duplex_server_vad_hard_interrupt` in 424.38s and the default-mode `test_duplex_soft_interrupt` in 329.64s. Silero loads from the wheel with no network access, so the new L4 test does not add a `torch.hub` dependency to CI. Against the repo's own `soft_interrupt_16k.wav`, the detector produces 5 clean onset/offset pairs with sane `speech_start_ms`, the 8 kHz resample path works, `reset()` clears model state, and malformed audio raises `ValueError` on every branch I tried.

**CI reds are unrelated.** amd-ci #10880 is `CosyVoice3-TTS E2E Test` exit 124 (timeout). npu-ci #5754 has exactly one failing job, `MiniCPM-o 4.5 Accuracy Test`; both MiniCPM duplex NPU jobs pass on that same build. I could not read the NPU job log anonymously, so that one is worth a glance before merge.

**On the two open `vad.py` threads.** The latch is real in the class: `SileroVADConfig.from_turn_detection({"threshold": 0.10})` gives `negative_threshold=0.0`, and I measured 1 onset with 0 offsets, still latched after 5s of pure silence. It is not reachable through the API. `_validate_realtime_turn_detection` rejects `threshold <= 0.15` on both `turn_detection` and `audio.input.turn_detection`, and `realtime_input.py:1160` is the only construction site. Those threads look resolvable. A range clamp inside `from_turn_detection` would still be worth it, since the invariant currently lives in a validator far from the class that depends on it.

Four findings below. The first two look blocking to me; both are proved by harness rather than inspection.

Separately, the removal of `DuplexOverlapPolicy.AUTO`, the `force_listen` propagation at `serving.py:524`, and the three-valued `_vad_speech_started` logic in both new branches all check out. `AUTO` was never referenced on main, and `force_listen` stays `False` for default `listen_only` and for `/v1/duplex`.

Sy0307:
PTAK @linyueqian @Gaohan123

vllm_omni/experimental/fullduplex/openai/realtime_session.py
Sy0307: Fixed in `8638ac37`: a failed append now emits `session_update_aborted` with the originating Realtime `event_id` and explicitly rejects the staged detector before continuing. `test_realtime_vad_update_rejects_after_failed_append_without_stalling` covers this path and verifies that the reader does not stall.

vllm_omni/experimental/fullduplex/openai/realtime_session.py
Sy0307: Fixed in `8638ac37`: `encode_outbound_event` no longer resolves VAD candidates; the runner now commits or rejects them only inside the matching `session.update` transaction. The regression test injects an unrelated outbound error while the update is blocked and verifies that the candidate remains pending until the actual runtime ACK.

vllm_omni/experimental/fullduplex/openai/realtime_input.py
Sy0307: Fixed in `8638ac37`: when one append contains both edges, `speech_stopped` is emitted before `speech_started`, so the client state transitions back to active correctly. The regression test supplies both flags in one result and asserts the two wire events and their order.

vllm_omni/experimental/fullduplex/openai/vad.py
Sy0307: Fixed in `8638ac37`: resampling now carries a streaming remainder across tiny chunks, including one-sample appends, instead of treating them as 16 kHz. The regression test verifies that 1,024 one-sample appends at 48 kHz advance the 16 kHz detector by exactly 341 samples.

vllm_omni/experimental/fullduplex/openai/session_runner.py
```diff
@@ -1065,14 +1101,16 @@ async def _run() -> None:
...
                                 session,
                                 candidate_config,
                             )
                             if runtime_update_error is not None:
-                                await emit_event(runtime_update_error)
+                                await emit_update_event(runtime_update_error)
+                                reject_update()
                                 continue
                             try:
                                 candidate_runtime_config = self._runtime_config_for_session_update(
```
linyueqian, line 1144: **Blocker.** This is the only early exit in the `session.update` handler that does not resolve the staged turn-detection candidate. The five sibling branches around it all call `reject_update()`; the `except ServingRuntimeConfigError` at L1120 goes straight to `continue`, so `_pending_turn_detection_update` stays set and the reader gate in `receive_internal_event_text` (`realtime_session.py:61`) never returns.

One Realtime event reaches it. Sent before the first append, so the new `native_context_locked` guard at `serving.py:1205` does not catch it and `MiniCPMO45NativeDuplexServingAdapter.runtime_config_for_update` raises on the instructions change:

```json
{"type": "session.update", "session": {"turn_detection": {"type": "server_vad"}, "instructions": "..."}}
```

Measured at ac2fa41c on vLLM 0.27.0, same harness for all three rows, `idle_timeout_s` lowered to 8s so the wedge is observable:

| client sends | adapter raises | next client event processed |
| --- | --- | --- |
| `turn_detection` only | no | yes |
| `instructions` only | yes | yes |
| both | yes | **no** |

Only the combination wedges, and the difference between rows 2 and 3 is the unresolved candidate. In production the session then holds until `_DEFAULT_IDLE_TIMEOUT_S` (300s, `serving.py:74`) and dies on idle timeout, dropping every event the client sends in between.

Adding `reject_update()` here, and routing the error through `emit_update_event` so it carries `realtime_event_id` like its siblings, returns row 3 to `yes` with the handler suite still at 206 passed.

Neither `test_minicpmo_native_session_update_rejects_instructions_before_first_real_append` nor `..._rejects_native_duplex_mode_flip` covers this: both feed the internal `turn.signal` form directly, which bypasses `RealtimeInputTranslator`, so no candidate is ever staged.

vllm_omni/experimental/fullduplex/openai/realtime_input.py
```diff
@@ -275,15 +284,49 @@ async def _to_duplex_event(self, event: dict[str, object]) -> dict[str, object]
...
+                else self._input_looks_like_speech(event, audio=audio, fmt=fmt)
+            )
             self._input_audio_buffer_has_audio = self._input_audio_buffer_has_audio or (
                 looks_like_speech and isinstance(audio, str) and bool(audio)
             )
             self._input_audio_buffer_had_non_speech = self._input_audio_buffer_had_non_speech or (
                 not looks_like_speech and isinstance(audio, str) and bool(audio)
             )
-            if looks_like_speech:
+            if vad_result is not None and vad_result.speech_stopped:
```
linyueqian, line 318: **High, regression against 50ded5c7.** Emitting the offset before the onset is correct only when the chunk ends inside a new utterance. When a single append contains a complete utterance (both edges present, `speech_active` False afterwards), the stop runs first, hits `if not self._input_speech_started: return` at L1378 and is dropped. `_input_speech_started` then latches True, which also shifts every later stop by one utterance.

Same probe on both heads, one append of 192ms speech followed by 320ms silence:

| head | VAD verdict | events to client | `_input_speech_started` after |
| --- | --- | --- | --- |
| 50ded5c7 | `started=True stopped=True active=False start_ms=0 end_ms=192` | `speech_started`, `speech_stopped` | False |
| ac2fa41c | identical | `speech_started` only | **True** |

At the advertised `chunk_period_ms=1000`, a short barge-in ("stop", "yes") lands entirely inside one append, so this is a normal case rather than a corner one.

Neither fixed order is correct on its own. `speech_active` is what selects between them:

```python
if vad_result.speech_stopped and vad_result.speech_active:
    await emit_stop()   # offset belongs to the previous utterance
if vad_result.speech_started:
    await emit_start()
if vad_result.speech_stopped and not vad_result.speech_active:
    await emit_stop()   # utterance opened and closed inside this chunk
```

I ran that shape: both cases emit in wire order and `test_realtime_vad_edge_order_and_tiny_chunk_resampling` stays green at 206 passed. That test presets `_input_speech_started = True` and monkeypatches `speech_active=True`, so it only pins the case that already works. A second case with `speech_active=False` and a clean flag would catch this.

linyueqian (commented):
The three new commits fix everything from the last round: the append-tail path now emits `session_update_aborted` and rejects the candidate instead of a bare `continue`, the outbound-ACK resolver is replaced by explicit commit/reject at the real decision points correlated by `realtime_event_id`, and the edge reorder fixes the dropped second `speech_started`. The transaction shape is right now.

Two of those fixes are incomplete. Details inline.

- Blocker, `session_runner.py:1120`: one `continue` still leaves the staged VAD unresolved, wedging the reader for the full 300s idle window. Reachable from the public Realtime socket with a single `session.update`.
- High, `realtime_input.py:314`: the edge reorder trades the bug it fixed for the mirror-image one. Regression against 50ded5c7.

Both fixes are small and I verified each against `tests/entrypoints/openai_api/test_duplex_handler.py`.

On CI: the `buildkite/vllm-omni` #14171 red is not a test failure. `Omni · MiniCPM-o 4.5 Duplex Test` has no `started_at`, no agent and an empty log (exit -1), so it never got scheduled. The same job ran and passed on #14166 and #14160. Since that is the job covering this change, it is worth a re-run rather than reading the red as signal.

The rest of the diff looks good. `pcm_reservation.rollback()` on `CancelledError` is safe because `_rollback_reservation` no-ops on an inactive reservation, the `speech_start_ms`/`speech_end_ms` fields dropped from the wire hint have no consumers, and the threshold validation still keeps the `vad.py` latch unreachable.

vllm_omni/experimental/fullduplex/openai/session_runner.py
Sy0307: Fixed in 52a6f1a8. The runtime-config exception now uses the correlated update emitter and rejects the staged VAD candidate, while the regression test verifies the reader continues and preserves the original event_id.

vllm_omni/experimental/fullduplex/openai/realtime_input.py
Sy0307: Fixed in 52a6f1a8. Stop-before-start is now used only when speech_active remains true; a complete utterance inside one append emits start then stop and leaves the client speech state cleared, with both cases covered by regression tests.

linyueqian (approved):
Both findings from my last round are fixed at 52a6f1a8f, and I re-ran the probes that caught them:

- `session_runner.py:1121` now rejects the staged candidate and routes the error through `emit_update_event`. A `session.update` carrying both `turn_detection` and changed `instructions` no longer wedges the reader; the next client event is processed instead of the session sitting dead until the 300s idle timeout.
- `realtime_input.py:314` now picks the emission order from `speech_active`. A single append containing a complete utterance emits `speech_started` then `speech_stopped` with `_input_speech_started` left clean, and the previous-utterance case still emits the stop first.

The regression tests are the right ones: the edge-order test now covers `speech_active=False`, and `test_realtime_vad_update_rejection_does_not_stall` asserts `_pending_turn_detection_update is None` on both rejection paths. CPU suites green at 269 passed (`test_duplex_handler.py`, `test_duplex_protocol.py`, `test_orchestrator_error_handling.py`, `e2e/features/fullduplex/test_client.py`) on vLLM 0.27.0.

Confirming the admission check dropped in 655fa7eba was genuinely stale and not a regression here: main `5de6f7c37` (#6619) raised `max_sessions` from 2 to 4 in the MiniCPM deploy YAMLs, so the probe's 2 accepted sessions plus 1 overflow now all fit under the cap and no `resource_exhausted` can be produced. Retargeting that constant to 4 would preserve the admission-rejection coverage, but that is a follow-up rather than a blocker.

Approving on the code. For the record: `Omni · MiniCPM-o 4.5 Duplex Test` has not yet reported on any head carrying these fixes. It never got an agent on ac2fa41c4 (exit -1, empty log) and is still pending on 52a6f1a8f. Everything I verified was CPU-level against a fake engine, so please confirm that job is green before merging.

One non-blocking nit: 52a6f1a8f also dropped the `threshold=0.15` boundary assertion from `test_native_realtime_protocol_rejects_server_vad_without_interrupt`. The validation at `realtime_input.py:1084` is unchanged, but that bound is what keeps the `vad.py` onset latch unreachable, so it is worth keeping a case on it.

Sy0307:
CUDA Buildkite #14176 has one failure in the unrelated Qwen3 test `test_qwen3_omni_expansion.py::test_speaker_002`. One generated sample produced a Whisper/text cosine similarity of 0.787 against the 0.8 threshold; the same job passed 45 other tests, and this Function Test passed on the earlier #14171 run of this PR.

PR #6170 does not change the Qwen3 serving or assertion paths. Its generic orchestrator change is entered only when duplex input processing raises, while this Qwen3 request completed successfully and failed only the stochastic audio/ASR quality assertion. I therefore classify this as an unrelated quality-test fluctuation, not a regression from this PR.

The PR-specific CUDA coverage is green: the ready MiniCPM-o 4.5 Offline, Online, and Duplex jobs passed, and the nightly MiniCPM-o 4.5 Duplex job passed all 3 tests.

tests/e2e/online_serving/run_minicpmo_realtime_duplex_server_vad.py
Sy0307: After the cleanup, `run_minicpmo_realtime_duplex_demo_pair.py` and `test_minicpmo_realtime_duplex_drivers.py` are no longer part of this PR diff. The remaining server-VAD runner is directly exercised by the MiniCPM GPU acceptance test, so it is retained.

vllm_omni/experimental/fullduplex/openai/session_runner.py
```diff
@@ -1249,7 +1254,16 @@ async def _run() -> None:
                     buffer_overlap_audio = True
                     if self._uses_native_input_append(session):
```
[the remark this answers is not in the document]
Sy0307: Addressed on the current head: both initial session setup and `session.update` pass through `_runtime_session_candidate_update_error`, and non-native runtimes receive `server_vad_requires_native_duplex` before audio is accepted. `test_realtime_server_vad_runtime_failures_are_terminal` covers the initial generic-session path.

Sy0307:
## Merge Readiness Checkpoint (2026-08-27)

The latest reviewer round verified the two blocking transaction fixes and approved the current code. Latest visible build, pre-commit, DCO, and documentation checks are green, and the PR-specific MiniCPM CUDA/ready/nightly coverage was reported green.

GitHub currently reports `MERGEABLE` with merge state `BLOCKED`; the remaining action is the required maintainer/branch-protection merge gate. The dropped `threshold=0.15` boundary assertion is retained as a non-blocking follow-up. The unrelated stochastic Qwen3 audio-quality failure in CUDA Buildkite #14176 is documented separately and does not exercise this PR's changed serving path.

TaffyOfficial:
@vllm-omni-review-bot

Sy0307:
CUDA Buildkite [#14235](https://buildkite.com/vllm/vllm-omni/builds/14235/list) validates the PR-specific paths after the vLLM 0.28 fix. All MiniCPM jobs passed: [Ready Offline](https://buildkite.com/vllm/vllm-omni/builds/14235/list?sid=01a04580-959f-4574-aed0-e589d610be2c&tab=output), [Online](https://buildkite.com/vllm/vllm-omni/builds/14235/list?sid=01a04580-95a2-4ce4-a7d1-88847493869f&tab=output), [Duplex](https://buildkite.com/vllm/vllm-omni/builds/14235/list?sid=01a04580-95a3-420e-8ab9-77cf344c9f78&tab=output), [Perf](https://buildkite.com/vllm/vllm-omni/builds/14235/list?sid=01a04580-95a3-4cbf-8646-52313dbd5f7a&tab=output), [nightly Duplex](https://buildkite.com/vllm/vllm-omni/builds/14235/list?sid=01a04580-cca3-42aa-87f9-149926bea7f8&tab=output), [Accuracy](https://buildkite.com/vllm/vllm-omni/builds/14235/list?sid=01a04580-ccab-4b1d-8719-cb6f053a4866&tab=output), [nightly Perf](https://buildkite.com/vllm/vllm-omni/builds/14235/list?sid=01a04580-ccaf-4502-93b7-9cc15081e2dd&tab=output), and [Duplex Seed-TTS Perf](https://buildkite.com/vllm/vllm-omni/builds/14235/list?sid=01a04580-ccb0-40e7-bdc5-7836f97e5df8&tab=output). Accuracy passed all four tests: regular Seed-TTS WER 0.0266 and both duplex variants 0.0171, with zero request/ASR failures.

The only red step is the [H100 Omni Function job](https://buildkite.com/vllm/vllm-omni/builds/14235/list?sid=01a04580-cc9f-4b27-b2dc-c6f2cf6ab574&tab=output): Nemotron VoiceChat real-model duplex passed, while Qwen3 produced one stochastic audio/text similarity failure (0.605 < 0.8) and one NVFP4 engine-startup error (45 passed, 10 skipped overall). #6170 changes no Qwen3 model/test/assertion path, and the same NVFP4 startup error plus another Qwen3 similarity flake reproduces on the [main nightly #14213 H100 job](https://buildkite.com/vllm/vllm-omni/builds/14213/list?sid=01a0460d-32af-47fc-9449-cdc2a163d842&tab=output), so these failures are not attributable to this PR.

TaffyOfficial:
LGMT

PR #4765, opened 2026-06-27
[Model][TTS] Add dots.tts (rednote-hilab): continuous-AR 48kHz TTS

Moore-Z:
## Summary
Closes #4218.

Adds `dots.tts` (rednote-hilab) — continuous-AR 48 kHz TTS built on
Qwen2.5-1.5B + autoregressive flow-matching DiT head + 48 kHz AudioVAE.

Follows the VoxCPM2 pipeline pattern (vLLM-native AR backbone +
side-path acoustic/vocoder modules), adapted for:
- Qwen2.5-1.5B-Base LM instead of MiniCPM4
- Single DiT flow-matching head with 10-step Euler ODE (no FSQ / residual LM)
- 48 kHz AudioVAE with BigVGAN-style causal decoder
- CAM++ x-vector speaker encoder (zero-shot path only this PR)

### New files
`vllm_omni/model_executor/models/dots_tts/`
- `dots_tts_talker.py` — main pipeline (LLM forward → DiT N-step Euler → patch_encoder AR loopback → AudioVAE decode → eos_proj)
- `dots_tts_vocoder.py` — vendored AudioVAE
- `dots_tts_dit.py` — vendored DiT
- `dots_tts_patch_encoder.py` — vendored semantic encoder
- `dots_tts_speaker_encoder.py` — vendored CAM++ x-vector
- `dots_tts_prompt.py` — prompt template
- `pipeline.py` — single-stage AR pipeline config

Plus `transformers_utils/configs/dots_tts.py`, `deploy/dots_tts.yaml`,
`examples/offline_inference/text_to_speech/dots_tts/end2end.py`, and
3 registry entries.

### Vendor strategy
Vendored from upstream @ \`a393d2e\` (release 2026-06-12).  Single-file
per module, full public API preserved (ming_flash_omni convention).
Upstream not on PyPI — vendoring is the only path.

### Diagnostics (env-gated, 0 production overhead)
- \`DOTS_TTS_BETA_TRACE=1\` — per-patch rms/max trace
- \`DOTS_TTS_HIDDEN_DUMP=<dir>\` — per-step LLM hidden \`.pt\` dump

### Functional smoke test
Requires \`rednote-hilab/dots.tts-soar\` cached or accessible via HF.

\`\`\`bash
python examples/offline_inference/text_to_speech/dots_tts/end2end.py \\
  --text "Hello, this is a test of dots TTS running on vLLM Omni."
\`\`\`

[...]

PR #6174, opened 2026-08-13
[CI/Build] Add dots.tts weekly e2e coverage

linyueqian:
## Summary

dots.tts landed in #4765 without any GPU coverage. Nothing under `tests/e2e/` exercises it, so the DiT flow-matching sampler, the streaming AudioVAE decode and the patch-encoder AR loopback currently have no CI signal on real hardware.

This adds `tests/e2e/offline_inference/test_dots_tts_expansion.py` marked `slow` + `tts` + L4, which the existing weekly **TTS · L4** step already collects via `-m "slow and L4 and tts"`. No pipeline YAML change is needed: the weekly TTS steps are marker driven, unlike the per-model reliability steps.

## Cases

| Test | Guards |
|---|---|
| `test_dots_tts_zero_shot_001` | Synthesis produces finite, non-silent audio of a plausible duration. |
| `test_dots_tts_prefill_decode_mixed_batch_002` | `compute_logits` fills `logits[i]` positionally from `_results_queue`, so every request must push exactly one entry per step, prefills included. Guards the row pairing raised in review of #4765. |
| `test_dots_tts_deterministic_flow_matching_noise_003` | The DiT noise comes from a per-request `blake2b(seed:request_key:noise_step)` generator rather than the global CUDA RNG. Guards the seeding fix from the same review. |

Note that dots.tts cannot use a bare `{"prompt": "..."}` string like most TTS models here: prefill needs `[文本]<text>[文本对应语音]<|audio_gen_start|>` as real Qwen2 token ids, so the helper goes through `build_dots_tts_prompt`.

## Test Plan

Selector behaviour verified against the pipelines' own expressions:

```
-m "slow and L4 and tts"        ->  3 collected      (weekly TTS · L4)
-m "full_model and L4 and tts"  ->  0 collected      (nightly, intentionally out of scope)
```

`ruff check` and `ruff format` clean at the repo's pinned v0.14.10.

The three cases have **not** been executed on an L4 yet, since they need the weekly runner. Adding the `weekly-test` label to this PR will validate them on the right hardware before merge.

## Context

Related to this, #4765's CPU unit tests were also never collected: the file carried no markers, so the ready pipeline's `-m 'core_model and cpu'` deselected all five. That was fixed in `8209be4` on the same PR. This one closes the remaining GPU-side gap.

cc @yenuo26 @congw729


Sy0307 (approved):
LGTM

PR #6049, opened 2026-08-11
[Misc] Build canonical image task prompts

hsliuustc0106:
## Purpose

Make shared text-to-image and image-to-video examples own their canonical task envelopes, while keeping `vllm_omni/model_extras` responsible only for genuinely model-specific prompt construction.

- build `modalities=["image"]` and `modalities=["video"]` in the shared `x_to_y.py` runners
- make the model-extra `build_*_prompt` APIs consume those canonical envelopes
- remove duplicated generic LingBot and Cosmos prompt builders
- retain the Bagel, MammothModa2, Ming, and VACE model-specific builders
- generalize the online Python clients and server launch scripts while preserving the existing model-specific curl recipes
- keep model-specific validation and normalization in the model pipeline
- leave the public online serving handler unchanged

This establishes the agreed ownership boundary without adding model-specific Python examples or changing public server request behavior.

## Test Plan

**vLLM Version:** 0.26.0

**vLLM-Omni Commit:** 0db924bc5444bd99c4ee6fd2714e1b4e9d9518f3

- Run pre-commit on all changed files.
- Validate the online-serving shell scripts with `bash -n`.
- Run the prompt, model-extra, LingBot request-validation, and Cosmos pipeline CPU tests:
  ```bash
  PYTHONPATH=$PWD /home/hsliu/tmp/venvs/vllm-omni-pr-5851-vllm-0.26.0/bin/python -m pytest -q \
    tests/model_extras/test_model_extras.py \
    tests/examples/offline_inference/test_image_task_prompts.py \
    tests/diffusion/models/lingbot_video/test_request_utils.py \
    tests/diffusion/models/cosmos3/test_cosmos3_pipeline.py
  ```
- Smoke-test LingBot-Video-Dense-1.3B text-to-image and image-to-video generation on one NVIDIA L20X.
- Smoke-test the Qwen-Image pipeline contract with a compatible synthetic/random checkpoint on one NVIDIA L20X.
- Exercise the Bagel, MammothModa2, Ming, Cosmos3, LingBot, VACE, and unknown-pipeline prompt paths in unit tests.

## Test Result

- Quick `precheck-pr`: 0 blocking findings, 0 warnings.
- Latest CPU suite: `192 passed, 16 warnings in 3.16s`.
- Both curl recipes, `serving_video.py`, and its general endpoint tests match the PR base.
- Diff hygiene and online server-script syntax passed.
- LingBot T2I: generated a valid 320x192 RGB PNG.
[...]

PR #6175, opened 2026-08-13
[Bugfix][XPU][Tests]Scope XPU pytest to explicit paths to fix collection crash

Joshna-Medisetty:
## Problem

The XPU job runs marker-only selections with no path, e.g.
`pytest -m "core_model and xpu and B60"`. With no path, pytest collects the
**entire** `tests/` tree to evaluate the marker, importing every test module.
`tests/examples/offline_inference/test_image_task_prompts.py` (#6049) does
`from examples... import ...`, which isn't importable, so collection errors and
aborts the whole run (exit 2) — even though it's not an XPU test. CUDA/AMD/NPU
never hit this because they scope pytest to explicit paths.

This is the 2nd unrelated collection failure to break the XPU job recently
(same class as #5780), so this fixes the root cause.

## Before

​```
tests/examples/offline_inference/test_image_task_prompts.py:9: in <module>
    from examples.offline_inference.image_to_video.image_to_video import build_image_to_video_prompt
E   ModuleNotFoundError: No module named 'examples.offline_inference'
=========================== short test summary info ============================
ERROR tests/examples/offline_inference/test_image_task_prompts.py
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
========== 3 skipped, 9896 deselected, 23 warnings, 1 error in 36.53s ==========
​```

## Fix

Scope the XPU pytest runs to explicit paths, like the CUDA/AMD pipelines:

```
XPU_TEST_PATHS="tests/diffusion tests/dfx tests/e2e"
pytest -v -s $XPU_TEST_PATHS -m "core_model and xpu and B60"
pytest -v -s tests/diffusion/quantization/test_mxfp8_config.py
pytest -v -s $XPU_TEST_PATHS -m "advanced_model and xpu and B60"
pytest -v -s $XPU_TEST_PATHS -m "omni and xpu and B60"
```

Scoped collection never imports unrelated files (e.g. tests/examples), so they can't break the job. The B60 marker only exists under these three dirs, so no coverage is lost  verified on build #6989, where core_model and xpu and B60 selected 7 tests, all under tests/diffusion.

Test
XPU collection no longer errors; the same xpu and B60 tests are selected and run.


tests/examples/__init__.py
```diff
@@ -0,0 +1 @@
+"""Example code tests."""
```
xuechendi: why this is needed?

Joshna-Medisetty: Without it the XPU CI crashes during pytest collection:

```
  ModuleNotFoundError: No module named 'examples.offline_inference'
  Interrupted: 1 error during collection  (exit 2)
```

`tests/examples/offline_inference/test_image_task_prompts.py` (from #6049) does `from examples... import ...`. Since `tests/examples/` had no `__init__.py`, pytest's `prepend` mode imports it as a top-level module and never adds the repo root to `sys.path`, so `examples` isn't importable. Adding `tests/examples/__init__.py` makes it resolve as `tests.examples.offline_inference.*`, which puts the repo root on `sys.path`. `tests/` already has `__init__.py`, so this just follows the existing layout — same fix as #5780.

Only XPU hits it because it collects the whole `tests/` tree in one run.

xuechendi: Ok, do you know which line? can we follow other HW to go deeper instead of using tests/?
image

I want to align with other HW for our test plan

Joshna-Medisetty: Good idea! Changed the PR purpose and aligned CI with other HW

xuechendi (approved):

xuechendi:
image

PR #6181, opened 2026-08-14
[Bugfix] Fix LongCat TeaCache CFG negative-branch guidance kwargs

yzong-rh:
## Purpose

With TeaCache enabled and CFG (`guidance_scale > 1`), LongCat Image / Edit fails on the negative branch. https://gist.github.com/yzong-rh/b9deacd8178fc0a1295c6382c2f8b059

```bash
python repro_teacache_sp.py --model longcat --ulysses-degree 1
```

```text
TypeError: extract_longcat_context() missing 1 required positional argument: 'guidance'
```

`extract_longcat_context` required `guidance` and did not match `LongCatImageTransformer2DModel.forward`. Negative CFG kwargs omitted `guidance` while positive kwargs included it, so TeaCache's hook crashed on the negative branch.

## Test Plan

```bash
python repro_teacache_sp.py --model longcat --ulysses-degree 1
```

```
pytest tests/model_tests/diffusion/test_common_offline.py -sv -k test_pipeline_on_supported_tasks[LongcatImagePipeline
```

**vLLM Version:**
v0.27.1

**vLLM-Omni Commit:**
0.27.0rc2.dev15+gfcf108539 (git sha: fcf108539)

## Test Result

 4 passed, 17 deselected, 15 warnings, 6 subtests passed in 114.57s

No more crash. Image generates cleanly.

image

```
================================================================================
Model                With Cache      Without Cache   Speedup   
--------------------------------------------------------------------------------
longcat                   15.91s         26.66s      1.68x
================================================================================
```

cc @alex-jw-brooks @princepride 

### DIFF
--- a/vllm_omni/diffusion/cache/teacache/extractors.py
+++ b/vllm_omni/diffusion/cache/teacache/extractors.py
@@ -723,11 +723,11 @@ def extract_flux2_klein_context(
 def extract_longcat_context(
     module: nn.Module,  # LongCatImageTransformer2DModel
     hidden_states,
-    timestep,
-    guidance,
-    encoder_hidden_states,
-    txt_ids,
-    img_ids,
+    encoder_hidden_states=None,
+    timestep=None,
+    img_ids=None,
+    txt_ids=None,
+    guidance=None,
     **kwargs,
 ) -> CacheContext:
     """Extract the cache context for LongCat Image.
--- a/tests/model_tests/diffusion/diff_model_builders.py
+++ b/tests/model_tests/diffusion/diff_model_builders.py
@@ -54,14 +54,18 @@ def _shrink_flux_t5_text_encoder(config: dict) -> dict:
 
 
 def _shrink_qwen_text_encoder_config(config: dict, hidden_size: int = 64) -> dict:
-    text_config = config["text_config"]
+    """Shrink a Qwen2.5-VL text encoder (nested Qwen-Image or flat LongCat layout)."""
+    # Nested: text fields live under text_config, with top-level mirrors.
+    # Flat: text fields live at the top level.
+    text_config = config["text_config"] if "text_config" in config else config
     old_head_dim = text_config["hidden_size"] / text_config["num_attention_heads"]
     text_config["num_hidden_layers"] = 2
     text_config["intermediate_size"] = 64
     text_config["hidden_size"] = hidden_size
     text_config["num_attention_heads"] = 2
     text_config["num_key_value_heads"] = 2
-    text_config["layer_types"] = text_config["layer_types"][:2]
+    if "layer_types" in text_config:
+        text_config["layer_types"] = text_config["layer_types"][:2]
     factor = old_head_dim / (hidden_size / 2)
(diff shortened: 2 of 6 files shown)


vllm_omni/diffusion/cache/teacache/extractors.py
```diff
@@ -723,11 +723,11 @@ def postprocess(h):
...
-    timestep,
-    guidance,
-    encoder_hidden_states,
-    txt_ids,
-    img_ids,
+    encoder_hidden_states=None,
+    timestep=None,
+    img_ids=None,
+    txt_ids=None,
+    guidance=None,
```
yzong-rh, line 730: Match `LongCatImageTransformer2DModel.forward()`

https://github.com/vllm-project/vllm-omni/blob/90a08c0edaa15eda9f0b7bc740fd226703b77470/vllm_omni/diffusion/models/longcat_image/longcat_image_transformer.py#L676-L685

alex-jw-brooks (approved):
Thanks @yzong-rh - can you add a tiny model builder (see this[https://github.com/yzong-rh/vllm-omni/blob/main/tests/model_tests/diffusion/diff_model_builders.py]) for LongCat & and a test for this? I don't think this model really has tests, but would be nice to add

yzong-rh:
The ~~test~~ tiny model tests don't catch this regression because they don't set `guidance_scale` so the negative branch isn't taken. We should either allow model overrides in the testing framework or change `guidance_scale` for all diffusion models.

yzong-rh:
Self-review from my agent
```
What I checked:
- Root cause: TeaCache calls `extract_longcat_context(module, **kwargs)`; negative CFG kwargs omitted `guidance` while the extractor required it → TypeError.
- Fix: align extractor signature with `LongCatImageTransformer2DModel.forward`; pass `guidance` on both CFG branches in T2I and Edit.
- Also fixed T2I `text_processor` load to use the hub `text_processor` subfolder (edit pipeline was already correct).
- Added tiny LongCat builder + TEA_CACHE alignment; TeaCache kept in its own test group (not combined with SP).
- Gap: current model tests don't set `guidance_scale > 1`, so they won't catch this exact regression yet.

Validation:
- `repro_teacache_sp.py --model longcat --ulysses-degree 1`: no crash; TeaCache ~1.68× vs uncached.
- Tiny LongCat diffusion tests including TEA_CACHE.
```

alex-jw-brooks:
> The test model tests don't catch this regression because it doesn't set guidance_scale so the negative branch isn't taken. We should either allow model overrides in the testing framework or change guidance_scale for all diffusion models.

Thanks! And yes, I agree with this 🙂 It would be helpful to add this to the tiny model testing framework for models that support it, and we can just test multiple requests with different CFG values etc to make sure things are propogated correctly. 

It is probably better to do in separate PR though, since that bug may be a copy paste error from other models that need the same fix, and this won't be TeaCache specific once we have finished the ongoing refactor with @akshatvishu to remove the extractor patterns

alex-jw-brooks (approved):
thanks for the fix!

PR #6182, opened 2026-08-14
[Bugfix] Preserve pipeline sampling constraints

### LINKED ISSUE #6177: [Bug]: caller-supplied SamplingParams replaces the stage defaults, dropping the pipeline's sampling_constraints
### Summary

A pipeline declares `sampling_constraints` in its `PipelineConfig`, and they reach a request only through the stage defaults. Any caller-supplied `SamplingParams` replaces those defaults wholesale, so the constraints — stop conditions included — are silently dropped. Passing nothing but a seed is enough to trigger it.

### Where it happens

Constraints enter the stage defaults at `vllm_omni/config/stage_config.py:853`:

```python
sampling.update(ps.sampling_constraints)
```

`vllm_omni/entrypoints/omni_base.py:314` then picks between the defaults and the caller's params without merging:

```python
if sampling_params_list is None:
    normalized = self.default_sampling_params_list   # constraints live here
elif isinstance(sampling_params_list, Sequence) and not isinstance(sampling_params_list, (str, bytes)):
    normalized = sampling_params_list                # returned verbatim
```

(Line numbers at `ddf1a9cf`.)

### Impact

Concrete case, Gepard-1.0 (#5666). Its constraints:

```python
sampling_constraints={
    "detokenize": False,
    "stop_token_ids": [_GEPARD_STOP_TOKEN_ID],
}
```

Generation ends on a learned binary stop head whose token is that `stop_token_ids` entry. A caller passing `SamplingParams(seed=1234)` loses it, vLLM never sees a stop signal, and the request runs to the stage's `max_tokens: 1000` — 1000 frames x 1024 samples / 22050 Hz, so **~46 s of noise in place of ~2 s of speech**. Nothing raises; the output is simply wrong.

This is not Gepard-specific. **27 files** under `vllm_omni/` declare `sampling_constraints`, voxcpm2, cosyvoice3, glm_tts, fish_speech and moss_tts_nano among them. Any pipeline whose constraints carry a stop condition, a `detokenize` flag, or anything else load-bearing has the same exposure.

### Question

Is replace-rather-than-merge deliberate — caller params take over completely by design — or should `sampling_constraints` be re-applied over caller-supplied params so a pipeline's non-negotiables survive?

If the latter, the change looks small, and it belongs in its own PR rather than riding along with a model.

### Workaround

Pass no `SamplingParams` and let the stage defaults stand. #5666's example and e2e tests do exactly that, with a comment at both sites explaining why.

maithilijoshi20:
## What broke

Pipeline-level `sampling_constraints` were included only in the default stage parameters. Supplying `SamplingParams(seed=1234)` replaced those defaults, so required settings such as `stop_token_ids` and `detokenize=False` were lost.

## Reproduction

```python
omni.generate(prompt, sampling_params_list=SamplingParams(seed=1234))
```

For a pipeline with `sampling_constraints={detokenize: False, stop_token_ids: [42]}`, the request previously omitted both constraints.

## Root cause

`OmniBase.resolve_sampling_params_list` selected caller parameters wholesale and did not retain the pipeline constraints that were merged into stage defaults during configuration resolution.

## Fix

Carry each stage's resolved `sampling_constraints` alongside its runtime config and overlay only those required values onto a shallow copy of caller-provided parameters. Normal caller overrides, such as `seed` and `max_tokens`, remain intact.

## Testing

- `python -m pytest tests/entrypoints/test_omni_entrypoints.py -q` : `62 passed, 14 warnings` in an isolated WSL Python 3.12 CPU environment.
- `python -m pre_commit run --files tests/entrypoints/test_omni_entrypoints.py vllm_omni/config/stage_config.py vllm_omni/entrypoints/omni_base.py`
- `ruff check` and `ruff format --check` on changed Python files.

Fixes #6177


vllm_omni/entrypoints/omni_base.py
```diff
@@ -223,6 +224,19 @@ def __init__(
...
+        self.sampling_constraints_list = [
+            dict(constraints)
+            if isinstance(
+                constraints := getattr(
+                    getattr(stage_config, "stage_pipeline_config", stage_config), "sampling_constraints", {}
+                ),
+                Mapping,
+            )
+            else {}
+            for stage_config in self.engine.stage_configs
```
hsliuustc0106: Tracing the production path, `engine.stage_configs` elements are OmegaConf DictConfigs, not `StageConfig` dataclasses: `merge_pipeline_deploy` builds the dataclasses, but `load_stage_configs_from_model` immediately converts them via `stage.to_omegaconf()` (`vllm_omni/entrypoints/utils.py:418`), and `to_omegaconf()` exports a fixed key set plus `yaml_extras` (`vllm_omni/config/stage_config.py:1029-1046`) — the new `sampling_constraints` field (`stage_config.py:978`) is never exported, and a DictConfig has no `stage_pipeline_config` attribute. So the getattr chain here resolves to `{}` for every stage (missing DictConfig keys fall through to the getattr default), `_apply_sampling_constraints` short-circuits on empty constraints, and the #6177 repro (`sampling_params_list=SamplingParams(seed=1234)` dropping Gepard's stop token) behaves exactly as before on a real engine. The new test only passes because it assigns `sampling_constraints_list` directly, bypassing this wiring.

Smallest fix: export the key in `to_omegaconf()` next to the `yaml_extras` merge (`config_dict["sampling_constraints"] = dict(self.sampling_constraints)`) — DictConfig is a `Mapping`, so the `isinstance` check here accepts it. Extending the test to build the list from `to_omegaconf()` output would keep this wiring covered.

Two related nits: the `stage_pipeline_config` branch only matches `BaseVllmOmniStageConfig`-style objects, which don't appear in `engine.stage_configs` today, so after the export fix a single `getattr(stage_config, "sampling_constraints", {})` would cover everything; and the `len != num_stages` reset below is unreachable (`engine.num_stages = len(engine.stage_configs)`) and would silently reproduce the exact bug it guards against if it ever fired — I'd drop it or raise.

vllm_omni/entrypoints/omni_base.py
```diff
@@ -339,8 +354,26 @@ def resolve_sampling_params_list(
...
         return normalized
 
+    @staticmethod
+    def _apply_sampling_constraints(params: Any, constraints: Mapping[str, Any]) -> Any:
+        """Apply pipeline-required sampling settings without mutating caller input."""
+        if not constraints:
+            return params
+        if isinstance(params, Mapping):
+            return {**params, **constraints}
+        constrained_params = copy.copy(params)
```
hsliuustc0106: `setattr` on an already-constructed `SamplingParams` skips its `__post_init__`, so derived state diverges: `_all_stop_token_ids` is populated only at construction (vllm `sampling_params.py:500` at v0.26.0) and by `update_from_generation_config` (`.add(eos)`, `:641`), and it's what the min_tokens ban and structured-output backends consume (`vllm/v1/sample/logits_processor/builtin.py:195`) — a `stop_token_ids` constraint applied here never reaches those. Token stopping itself is fine, since the scheduler checks `stop_token_ids` directly (`vllm/v1/core/sched/utils.py:108`).

Rebuilding the params with the constraint keys applied at construction time (e.g. via `SamplingParams.from_optional` with the caller's fields plus the constraints, so `__post_init__` runs) would keep them consistent; alternatively refresh the derived set explicitly after the loop. A rebuild also avoids `copy.copy` sharing the caller's `_all_stop_token_ids` set, which the engine later mutates in place — so eos would leak into the caller's object as-is.

vllm_omni/entrypoints/omni_base.py
```diff
@@ -325,6 +339,7 @@ def resolve_sampling_params_list(
         sampling_params_list: Sequence[Any] | Any | None,
         allow_delta_coercion: bool = False,
     ) -> Sequence[Any]:
+        """Resolve request sampling parameters without dropping stage constraints."""
```
hsliuustc0106: The acceptance matrix in the #6177 thread (foma-agent's comment) asked for a typed error when the caller explicitly conflicts with a constrained field, with documented pipeline-wins as the acceptable fallback. This implements silent pipeline-wins: a caller who explicitly passes `stop_token_ids=[...]` on a Gepard stage gets their value silently replaced by the constraint. Should conflicting explicit values be rejected instead? If silent pipeline-wins is the intended contract, worth stating it in this docstring so the behavior is discoverable.

maithilijoshi20:
Hi @hsliuustc0106, thanks for the detailed review. I have addressed the feedback:

- Exported `sampling_constraints` through `StageConfig.to_omegaconf()` so runtime stage configs retain them.
- Normalized OmegaConf values to native Python containers before applying constraints.
- Rebuild `SamplingParams` rather than mutating it, so derived stop-token state remains consistent.
- Added regression coverage for constraint propagation, caller conflicts, immutability, and the OmegaConf path.

Validation completed:

- `tests/entrypoints/test_omni_entrypoints.py`: 62 passed
- `tests/config/test_config_factory.py`: 299 passed

Could you please take another look when you have a chance?

hsliuustc0106:
@alex-jw-brooks PTAL

vllm_omni/entrypoints/omni_base.py
hsliuustc0106: Verified — the `to_omegaconf()` export lands the key on the DictConfigs `engine.stage_configs` actually holds, the extraction is a single getattr with `OmegaConf.to_container`, and the test now covers the wiring through the real conversion.

vllm_omni/entrypoints/omni_base.py
hsliuustc0106: Verified — the rebuild through the constructor recomputes `_all_stop_token_ids` (the test asserts it), and since the caller's object is never copied or touched, the shared-set aliasing is gone too.

vllm_omni/entrypoints/omni_base.py
hsliuustc0106: Documented pipeline-wins works for me — that was the acceptable fallback in the #6177 thread. Thanks.

hsliuustc0106 (approved):
lgtm

PR #6183, opened 2026-08-14
[Model][Feat] LTX HQ pipelines

mglyn:
## Purpose

- Add the Res2s high-quality two-stage pipeline for LTX-2.3 and LTX-2.5.
- Expose `LTX2TwoStageHQPipeline` for T2V and first-frame I2V through offline inference and OpenAI-compatible `/v1/videos` serving.
- Reuse the existing two-stage component profiles and sidecar weights while selecting an independent HQ execution recipe.
- Extend LTX accuracy coverage and fix BWE vocoder precision.

Based on #6070 and #5500.

## What this PR adds

| Pipeline | Model | Components | Sampling recipe | Coverage |
|---|---:|---|---|---|
| `LTX2TwoStageHQPipeline` | 2.3 | 2.3 dev Transformer, x2 spatial upsampler, distilled LoRA 384 | Res2s 15 + 3; LoRA strengths `0.25` / `0.5` | T2V, I2V, offline, online, similarity |
| `LTX2TwoStageHQPipeline` | 2.5 | `transformer_full/`, x2 spatial upsampler, distilled LoRA 450 | Res2s 15 + 3; LoRA strengths `0.25` / `0.5` | T2V, I2V, offline, online |

The implementation adds the token-count-dependent sigma schedule, second-order midpoint prediction, deterministic joint audio/video noise streams, SDE injection, explicit normalized-sigma guidance, SP-safe audio padding, and I2V conditioning restoration at each Res2s substep. Version-specific recipes select the correct Transformer and sidecars without introducing another component profile or another set of weights.

## H200 full-resolution qualification

Cold single-process runs used one H200, BF16 `CUDNN_ATTN`, eager mode, seed 42, 1920x1088, 121 frames, and 24 FPS. TP, VAE tiling/slicing, CPU offload, and layerwise offload were disabled. Generation excludes initialization and MP4 encoding.

| Model | Pipeline | Schedule | Generation | Peak GPU memory reserved |
|---|---|---:|---:|---:|
| LTX-2.3 | Ordinary Full two-stage | 30 + 3 | 96.687 s | 119,672 MiB (116.87 GiB) |
| LTX-2.3 | HQ Res2s two-stage | 15 + 3 | 97.199 s | 123,316 MiB (120.43 GiB) |
| LTX-2.5 | Ordinary Full two-stage | 30 + 3 | 96.711 s | 120,680 MiB (117.85 GiB) |
| LTX-2.5 | HQ Res2s two-stage | 15 + 3 | 98.315 s | 122,280 MiB (119.41 GiB) |

For this request, HQ changes latency by `+0.53%` on 2.3 and `+1.66%` on 2.5, and peak reserved memory by `+3,644 MiB` and `+1,600 MiB`, respectively.

## Demo videos

The demos use the model-card prompt, seed 42, 1920x1088, 121 frames, and 24 FPS. Ordinary two-stage is on the left and HQ Res2s is on the right; each MP4 contains 48 kHz stereo audio.

> A cinematic shot of a red fox walking through a snowy forest at dawn, the camera tracking alongside, snow crunching underfoot.

https://github.com/user-attachments/assets/225036cf-90d5-4db1-b4cb-f9529d1e75d5

https://github.com/user-attachments/assets/04b41b2b-b72e-4abc-a089-d5c5ab32ffe9

The demos show the sampling-recipe difference; they are not pixel-parity comparisons because Res2s changes the stochastic trajectory.

### Similarity coverage and vocoder precision

The similarity suite is simplified around representative two-stage coverage:

- Remove the redundant LTX-2.0/LTX-2.3 one-stage cases; the retained Distilled and Full two-stage cases cover their Stage-1 denoising paths at larger representative shapes.
- Remove the duplicate LTX-2.5 HQ similarity case because the HQ-only Res2s behavior is already covered by LTX-2.3; LTX-2.5 keeps Distilled and Full/SFT coverage, while HQ dispatch remains covered by serving E2E.
- Use one strict threshold profile for all retained LTX similarity cases: SSIM mean/min `>= 0.99`, PSNR `>= 40 dB`, audio relative L2 `<= 0.10`, and audio cosine `>= 0.99`.

LTX-2.3+ uses a bandwidth-extension vocoder with a long convolution stack. Running that stack in BF16 accumulated enough waveform error for the LTX-2.5 Distilled T2V case to reach audio relative L2 `0.331759` and cosine `0.947790`. The BWE vocoder now runs under FP32 autocast and converts the final waveform back to the input dtype; the base vocoder is unchanged. The same case now reaches relative L2 `0.013621` and cosine `0.999908`.

<details>
<summary><strong>Recent H200 similarity results</strong></summary>

All retained cases below pass the single strict profile: SSIM mean/min `>= 0.99`, PSNR `>= 40 dB`, audio relative L2 `<= 0.10`, and audio cosine `>= 0.99`.

| Version | Pipeline / task | SSIM mean | SSIM min | PSNR | Audio relative L2 | Audio cosine | Result |
|---|---|---:|---:|---:|---:|---:|---|
| LTX-2.0 | Distilled T2V | 0.999568 | 0.999457 | 62.466 dB | 0.000000 | 1.000000 | Pass |
| LTX-2.0 | Full two-stage T2V | 0.999635 | 0.999515 | 63.068 dB | 0.000000 | 1.000000 | Pass |
| LTX-2.3 | Distilled I2V | 0.999476 | 0.999151 | 62.408 dB | 0.033802 | 0.999429 | Pass |
| LTX-2.3 | Full two-stage I2V | 0.999523 | 0.999237 | 62.599 dB | 0.045554 | 0.998973 | Pass |
| LTX-2.3 | HQ Res2s T2V | 0.999762 | 0.999702 | 64.473 dB | 0.029472 | 0.999578 | Pass |
| LTX-2.5 | Distilled two-stage T2V | 0.997656 | 0.996499 | 46.961 dB | 0.013621 | 0.999908 | Pass |
| LTX-2.5 | Distilled two-stage I2V | 0.997642 | 0.997191 | 47.914 dB | 0.026768 | 0.999643 | Pass |
| LTX-2.5 | Full one-stage T2V | 0.999910 | 0.999842 | 66.769 dB | 0.003933 | 0.999994 | Pass |
| LTX-2.5 | Full one-stage I2V | 0.999798 | 0.999749 | 64.975 dB | 0.004341 | 0.999992 | Pass |
| LTX-2.5 | Full two-stage T2V | 0.999726 | 0.999657 | 59.788 dB | 0.003854 | 0.999994 | Pass |
| LTX-2.5 | Full two-stage I2V | 0.999819 | 0.999766 | 66.230 dB | 0.007642 | 0.999972 | Pass |

</details>

### DIFF
--- a/vllm_omni/diffusion/models/ltx2/ltx2_components.py
+++ b/vllm_omni/diffusion/models/ltx2/ltx2_components.py
@@ -228,7 +228,7 @@ _COMPONENT_PROFILES: dict[tuple[str, str], LTXComponentProfile] = {
 
 def resolve_ltx_checkpoint_kind(pipeline_kind: str) -> LTXCheckpointKind | None:
     """Derive checkpoint requirements from the execution contract."""
-    if pipeline_kind in {"one_stage", "two_stage"}:
+    if pipeline_kind in {"one_stage", "two_stage", "two_stage_hq"}:
         return "regular"
     if pipeline_kind in {"distilled_one_stage", "distilled_two_stage"}:
         return "distilled"
--- a/vllm_omni/diffusion/models/ltx2/ltx2_components.py
+++ b/vllm_omni/diffusion/models/ltx2/ltx2_components.py
@@ -213,6 +213,8 @@ _COMPONENT_PROFILES: dict[tuple[str, str], LTXComponentProfile] = {
     ("two_stage", "2"): LTX2_TWO_STAGE_COMPONENT_PROFILE,
     ("two_stage", "2.3"): LTX23_TWO_STAGE_COMPONENT_PROFILE,
     ("two_stage", "2.5"): LTX25_TWO_STAGE_COMPONENT_PROFILE,
+    ("two_stage_hq", "2.3"): LTX23_TWO_STAGE_COMPONENT_PROFILE,
+    ("two_stage_hq", "2.5"): LTX25_TWO_STAGE_COMPONENT_PROFILE,
     ("distilled_one_stage", "2"): LTX2_DISTILLED_ONE_STAGE_COMPONENT_PROFILE,
     ("distilled_one_stage", "2.3"): LTX23_DISTILLED_ONE_STAGE_COMPONENT_PROFILE,
     ("distilled_one_stage", "2.5"): LTX25_DISTILLED_ONE_STAGE_COMPONENT_PROFILE,
--- a/vllm_omni/diffusion/models/ltx2/ltx2_components.py
+++ b/vllm_omni/diffusion/models/ltx2/ltx2_components.py
@@ -639,14 +639,18 @@ def initialize_pipeline_components(pipeline: Any, od_config: Any) -> None:
         upsampler_config = os.path.join(model, "latent_upsampler", "config.json")
         if os.path.isfile(upsampler_config) or not local_files_only:
             try:
-                pipeline.latent_upsampler = _load_component(
-                    LTX2LatentUpsamplerModel,
-                    model,
-                    "latent_upsampler",
-                    local_files_only=local_files_only,
-                    dtype=dtype,
-                    revision=revision,
-                )
+                # Diffusers builds the rational-resampler kernel with Long
+                # tensors; constructing it under a CUDA default-device context
+                # attempts an unsupported Long addmm before weights are loaded.
+                with torch.device("cpu"):
+                    pipeline.latent_upsampler = _load_component(
+                        LTX2LatentUpsamplerModel,
+                        model,
+                        "latent_upsampler",
+                        local_files_only=local_files_only,
+                        dtype=dtype,
+                        revision=revision,
+                    )
             except (OSError, ValueError):
                 if profile.latent_upsampler_filename is None or profile.artifact_repo_id is None:
                     raise
--- a/vllm_omni/diffusion/models/ltx2/ltx2_guidance.py
+++ b/vllm_omni/diffusion/models/ltx2/ltx2_guidance.py
@@ -386,21 +386,96 @@ class LTXGuidanceExecutor:
         )
         return velocity_from_x0(sample, guided, sigma)
 
-    def predict_parallel_denoised(
+    @staticmethod
+    def _pass_contexts(
+        prompt: Any,
+        passes: tuple[LTXDenoisePass, ...],
+    ) -> tuple[list[torch.Tensor], list[torch.Tensor]]:
+        video_contexts: list[torch.Tensor] = []
+        audio_contexts: list[torch.Tensor] = []
+        for denoise_pass in passes:
+            video_context = (
+                prompt.negative_connector_prompt_embeds
+                if denoise_pass.negative_video_context
+                else prompt.positive_connector_prompt_embeds
+            )
+            audio_context = (
+                prompt.negative_connector_audio_prompt_embeds
+                if denoise_pass.negative_audio_context
+                else prompt.positive_connector_audio_prompt_embeds
+            )
+            if video_context is None or audio_context is None:
+                raise ValueError("Negative prompt context is required when LTX CFG is enabled.")
+            video_contexts.append(video_context)
+            audio_contexts.append(audio_context)
+        return video_contexts, audio_contexts
+
+    def _execute_model_passes(
         self,
         pipeline: Any,
-        plan: LTXGuidancePlan,
-        index: int,
+        execution_plan: LTXGuidancePlan,
         timestep: torch.Tensor,
         state: LTXAVState,
         forward_ctx: LTXForwardContext,
         denoise_ctx: LTXDenoiseContext,
-        *,
-        video_sigma: torch.Tensor,
-        audio_sigma: torch.Tensor,
-    ) -> tuple[torch.Tensor, torch.Tensor]:
-        """Run guidance-parallel prediction and return guided x0 tensors."""
-        del index
+    ) -> tuple[tuple[torch.Tensor, ...], tuple[torch.Tensor, ...]]:
+        """Execute one rank's guidance passes and return raw velocity slots."""
+        prompt = forward_ctx.prompt_context
+        video_contexts, audio_contexts = self._pass_contexts(prompt, execution_plan.passes)
+        pass_count = len(execution_plan.passes)
+        video_input = _repeat_batch(state.video, pass_count).to(prompt.positive_connector_prompt_embeds.dtype)
+        audio_input = _repeat_batch(state.audio, pass_count).to(prompt.positive_connector_audio_prompt_embeds.dtype)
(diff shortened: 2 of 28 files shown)


hsliuustc0106:
resolve conflicts

vllm_omni/diffusion/models/ltx2/ltx2_components.py
```diff
@@ -213,6 +213,8 @@ class LTXComponentProfile:
     ("two_stage", "2"): LTX2_TWO_STAGE_COMPONENT_PROFILE,
     ("two_stage", "2.3"): LTX23_TWO_STAGE_COMPONENT_PROFILE,
     ("two_stage", "2.5"): LTX25_TWO_STAGE_COMPONENT_PROFILE,
+    ("two_stage_hq", "2.3"): LTX23_TWO_STAGE_COMPONENT_PROFILE,
```
hsliuustc0106, line 216: These profile entries don't complete the registration: `resolve_ltx_checkpoint_kind` (`ltx2_components.py:229`) has no `two_stage_hq` branch, so it falls through to `ValueError: Unsupported LTX pipeline kind: 'two_stage_hq'.` — and it is called unconditionally during component loading (`ltx2_components.py:689`), after the VAE/encoders load. Every real `LTX2TwoStageHQPipeline` construction crashes at init on this head.

The unit tests only exercise the recipe/profile resolvers (which is why the focused suite is green), and the accuracy guards are manual-only, but the new `two_stage_hq` online-serving smokes (`tests/e2e/online_serving/test_ltx.py`, `test_ltx25.py`) run in nightly and will crash — and the manual parity/qualification runs in the PR description can't have been produced by this exact tree.

Smallest fix: add `"two_stage_hq"` to the `"regular"` branch alongside `"two_stage"` (same regular checkpoint + sidecar artifacts). Worth also asserting `resolve_ltx_checkpoint_kind("two_stage_hq")` in `test_ltx2_hq_recipe.py`, which covers the other two resolvers but misses this one.

vllm_omni/diffusion/models/ltx2/ltx2_guidance.py
```diff
@@ -473,6 +594,96 @@ def predict_parallel_guidance(
             ),
         )
 
+    def predict_denoised(
```
hsliuustc0106, line 598: Nit, non-blocking: `predict_denoised` and `predict_parallel_denoised` duplicate the pass-context build / `_repeat_batch` / perturbation / `_build_transformer_kwargs` / transformer-call blocks of `predict_noise` and `predict_parallel_guidance` — only the final x0-vs-velocity combine differs. A shared helper that executes the plan passes and returns the raw per-pass splits would keep future attention-kwarg or context changes in one place instead of four.

vllm_omni/diffusion/models/ltx2/ltx2_components.py
mglyn: This was missed during the semantic rebase onto #6070. I’ve added two_stage_hq to the regular-checkpoint branch in resolve_ltx_checkpoint_kind() and added a regression assertion covering the HQ component/recipe resolution. HQ shares the same regular two-stage checkpoint and component profile, so "regular" is the intended result.

vllm_omni/diffusion/models/ltx2/ltx2_guidance.py
mglyn: Agreed. I’ve consolidated prompt-context selection, perturbation construction, transformer execution, and output splitting into shared helpers used by both the Euler velocity path and the Res2s denoised-x0 path, including guidance-parallel execution. The public methods now only handle the genuinely different final combination: Euler consumes guided velocity, while Res2s requires guided x0 at an explicit sigma. This removes the duplicated model-pass logic without changing either numerical path.

linyueqian (commented):
First round, static review at 1c0b7240. I read ltx2_res2s.py in full and the integration surfaces at depth: run_res2s_phase and the scheduler-stage wiring in ltx2_denoise.py, predict_denoised and the modality-guidance refactor in ltx2_guidance.py, the recipe validation and version map in ltx2_recipes.py, and the registry entries across diffusion/registry.py, model_extras/registry.py and pipeline_ltx2_two_stage.py.

No defects survived verification, and several things I went in expecting to flag turned out to be handled. The euler-path guidance refactor is operation-for-operation identical to the base (x0_from_velocity at model_sigma, combine, velocity_from_x0 at sigma, contiguous preserved), so existing LTX-2/2.3/2.5 outputs are not at regression risk from the reuse. The always-zero step index passed by the res2s predict_x0 callback is fine because predict_denoised keys everything off the explicit sigma and discards the index. video_token_count=latents.shape[1] is correct because latents are packed 3D at that call site. refine_res2s_anchor degrades to a no-op at iterations=0, so the permissive >=0 guard is coherent. The adapter slot/strength pairing validation is satisfied at every construction site including the pre-existing official recipe, and the two_stage_hq version map covers both 2.3 and 2.5 with a clean error for anything else. The unit tests pin golden values for the token-dependent schedule, the phi coefficients, the legacy SDE step, and a full executor interval, plus RNG seed and video-then-audio order contracts and the SP-padding exclusion for audio noise, which is exactly the coverage an official-parity sampler needs.

What this round cannot certify is runtime behavior: the per-PR lanes have never fired on this branch (no ready label, so only per-commit checks exist), and the accuracy similarity tier needs the GPU lanes. I have added the ready label to start them. Approval follows green lanes with any red matched against the known inherited signatures. Note the release freeze blocks feature merges through 2026-08-31 regardless, so nothing is lost by that gate.

linyueqian:
Lane results at 1c0b7240, triaged so you do not have to guess. The general CUDA lane failed exactly one test, tests/diffusion/cache/test_teacache_extractors.py TestMiniMaxH3Extractor::test_extractor_matches_native_forward, a numeric-parity check on a model this PR does not touch; it passed on other branches' builds hours earlier and 3367 other tests passed alongside it, so I read it as a tolerance flake that will re-roll. Intel is green.

AMD is the one that needs your action: all sixteen steps die at collection with "ImportError: cannot import name 'DiffusionOutput' from partially initialized module 'vllm_omni.diffusion.data' (most likely due to a circular import)". Your added imports are all LTX-package-internal, so the cycle is not something this diff introduces directly; the branch base is 66 diffusion commits behind main, and vllm_omni/diffusion/data.py has been restructured three times since (#6037, #6749, #6340), so the ROCm import path is exercising a stale tree that current main no longer has. The NPU lane also picked up a Platforms Test failure beyond its usual four inherited MiniCPM-o/Qwen3-TTS reds, consistent with the same import problem.

Updating the branch onto current main should dissolve the AMD and NPU collection failures and re-roll the flake in one go. The review itself found nothing to change (see my first-round comment), so from my side this is the only thing between here and an approval pass over green lanes. Happy to re-trigger CI once the update lands.

PR #6184, opened 2026-08-14
[NPU][MiniCPM-o 4.5] Async omni output — deferred builder + D2H snapshot

a973826287:
## Purpose

Port the GPU `OmniAsyncGPUModelRunnerOutput` deferred-builder pattern to the NPU AR model runner, completing the NPU parity for the **Async Output** item on the omni optimization roadmap (PPT Slide 11/17: "NPU runner 未接入非阻塞 payload 路径").

Currently the NPU runner builds the full `OmniModelRunnerOutput` synchronously in `sample_tokens()` before wrapping it in `AsyncGPUModelRunnerOutput` — only the sampled-token D2H is overlapped, while the expensive omni payload assembly (hidden_states D2H + multimodal D2H + connector serialization) blocks the worker main thread on the TTFT critical path.

Implementation:
- `hidden_states` + `multimodal_outputs` are D2H-snapshotted on the main thread's dedicated `npu.Stream` (async), recorded as an `npu.Event`.
- A background daemon thread waits on the event and assembles the pooler payload + `OmniModelRunnerOutput` from the already-CPU snapshots (pure CPU work), overlapping with the next scheduler step.
- Extracted `_build_omni_output_from_tensors()` to hold the shared sync/async assembly path, mirroring GPU's `_build_omni_model_runner_output_from_snapshot`.

Two NPU-specific pitfalls fixed:
1. `torch.cuda.set_device` is NOT redirected by vllm-ascend's `_torch_cuda_wrapper` (only `Event`/`Stream`/`current_stream` are), so the builder thread must call `torch.npu.set_device` explicitly.
2. The builder must not re-read `self.input_batch` / `self.query_start_loc` (mutated by the next `execute_model` step under async scheduling); the per-request scheduling metadata is snapshotted on the main thread and passed in.

## Test Plan

- Unit tests: `tests/platforms/npu/worker/test_npu_ar_model_runner.py` (3 tests, CPU-only via `object.__new__`)
- Seed-TTS benchmark (single-concurrency + multi-concurrency 4/8)
- Seed-TTS WER/ASV accuracy verification

**vLLM Version:** 0.1.dev1+ge5588e49b

**vLLM-Omni Commit:** 009b80d6 (base) / 37c831b1 (head)

## Test Result

Single-concurrency (Ascend 910C, 5-run mean, very low variance):

| Metric | Before | After | Δ |
|--------|--------|-------|---|
| TTFT | 317 ms | 266 ms | **-16%** |
| TTFP | 776 ms | 757 ms | -2.5% |
| RTF | 0.39 | 0.40 | +2.5% |
| WER | 0.0107 | 0.0107 | unchanged |
| ASV | 0.8433 | 0.8433 | unchanged |

The RTF +2.5% is a real (not noise) trade-off: the snapshot `clone()` competes for HBM bandwidth in the memory-bound Talker decode loop (arithmetic intensity ~20 FLOPs/byte << the ~100 ridge). At multi-concurrency RTF regresses further (snapshot overhead + thread contention), so the feature is intended for the single-concurrency serving path.

Enabled by default (matches GPU). Opt out via `VLLM_OMNI_NPU_ASYNC_OUTPUT=0`.

### DIFF
--- a/tests/platforms/npu/worker/test_npu_ar_model_runner.py
+++ b/tests/platforms/npu/worker/test_npu_ar_model_runner.py
@@ -1,0 +1,100 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+"""Unit tests for the NPU async omni output (OmniAsyncNPUModelRunnerOutput).
+
+Mirrors the GPU tests in tests/worker/test_gpu_ar_model_runner.py but for the
+NPU deferred-builder class. Bypass __init__ via object.__new__ so the tests
+run on CPU without an NPU device.
+"""
+
+from types import SimpleNamespace
+
+import pytest
+import torch
+
+from vllm_omni.outputs import OmniModelRunnerOutput
+from vllm_omni.platforms.npu.worker.npu_ar_model_runner import (
+    OmniAsyncNPUModelRunnerOutput,
+)
+
+pytestmark = [pytest.mark.core_model, pytest.mark.cpu]
+
+
+def _make_async_output(
+    builder=None,
+    *,
+    sampled_token_ids=(7,),
+    max_gen_len=None,
+):
+    async_output = object.__new__(OmniAsyncNPUModelRunnerOutput)
+    async_output._model_runner_output = None
+    async_output._model_runner_output_builder = builder
+    async_output._background_thread = None
+    async_output._background_exception = None
+    async_output._invalid_req_indices = []
+    async_output._has_fault = None
+
+    ids = torch.tensor([list(sampled_token_ids)], dtype=torch.long)
+    async_output.sampled_token_ids_cpu = ids
+    async_output.async_copy_ready_event = SimpleNamespace(synchronize=lambda: None)
+    async_output._sampled_token_ids = ids
+    async_output._logprobs_tensors = None
+    async_output._logprobs_tensors_cpu = None
+    async_output._routed_experts = None
+    async_output._routed_experts_cpu = None
+    async_output.vocab_size = 10
+    return async_output
+
+
+def test_omni_async_npu_model_runner_output_builds_lazily_once():
+    calls = []
+
+    def builder():
+        calls.append("build")
+        return OmniModelRunnerOutput(req_ids=["r1"], req_id_to_index={"r1": 0})
+
+    async_output = _make_async_output(builder)
+
+    output = async_output.get_output()
+
+    assert calls == ["build"]
+    assert async_output._model_runner_output_builder is None
+    assert output.req_ids == ["r1"]
+    assert output.sampled_token_ids == [[7]]
+
+
+def test_omni_async_npu_model_runner_output_reraises_background_exception():
+    async_output = object.__new__(OmniAsyncNPUModelRunnerOutput)
+    joined = []
+
+    class FakeThread:
+        def join(self):
+            joined.append("join")
+
+    async_output._background_thread = FakeThread()
+    async_output._background_exception = RuntimeError("background failed")
+    async_output._model_runner_output = None
+    async_output._model_runner_output_builder = lambda: None
+    async_output._has_fault = None
+
+    with pytest.raises(RuntimeError, match="background failed"):
+        async_output.get_output()
+
+    assert joined == ["join"]
+    assert async_output._background_thread is None
+
+
+def test_omni_async_npu_model_runner_output_builder_runs_once():
+    # The builder runs exactly once; get_output() finalizes the sampled token
+    # ids via the superclass (AsyncGPUModelRunnerOutput.get_output).
+    calls = []
+
+    def builder():
+        calls.append("build")
+        return OmniModelRunnerOutput(req_ids=["r1"], req_id_to_index={"r1": 0})
+
+    async_output = _make_async_output(builder)
+    async_output.get_output()
+
+    assert calls == ["build"]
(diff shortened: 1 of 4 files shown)


Junbo2002:
Tried to reproduce this on an Atlas A3 (910C) and ended up with an attribution
rather than a benchmark number, because **as shipped it does not start on this
machine**. Posting the whole chain, since it seems more useful than "did not
run" — and the last table is the number this PR is actually about.

Setup for everything below: `ecd9d99da` plus the same 73-line local startup
workaround in every arm (a model-inspection subprocess aborts during teardown on
these machines, unrelated to this PR), the repo's own harness (Seed-TTS zh,
`openai-chat-omni`, 32 prompts, concurrency 1), the organizer deploy config; two
arms at once, one per die, each round repeated with the dies swapped.

## 1. Stage 0 cannot allocate its KV cache

| tree | stage-0 available KV cache | starts? |
| --- | ---: | --- |
| base | 8.00–8.01 GiB (4 arms) | yes |
| this PR, `VLLM_OMNI_NPU_ASYNC_OUTPUT=1` | **1.53–1.54 GiB** | **no** |
| this PR, variable unset | **1.53–1.54 GiB** | **no** |

```
ValueError: To serve at least one request with the model's max seq len (32768),
4.5 GiB KV cache is needed, which is larger than the available KV cache memory
(1.53 GiB). ... the estimated maximum model length is 11136.
```

Those last two rows are the same state rather than a treated/control pair:
`_should_use_async_omni_output` returns False only when the variable is
explicitly `0`/`false`/`no`, so **the feature is on by default**. That is worth
saying in the description either way — a default-on performance feature is
exactly the scoping point #6616 was asked about.

## 2. The missing 6.5 GiB is not the async output path

`minicpmo_4_5_omni_llm.py`'s `__init__` also gained an **unconditional INT8
online quantization** of the Thinker's Linear layers —
`torch_npu.npu_dynamic_quant` + `maybe_trans_nz`, registered as `_int8_weight` /
`_int8_weight_scale` buffers with a patched `forward`. It sits behind no gate:
not the async-output check, and not `vllm_config.quant_config` (its own comment
says it bypasses `quant_description` validation deliberately). And it keeps the
BF16 weight intact by design, so **both copies stay resident** — which is where
the memory goes.

## 3. Attribution — the same diff with only that block removed

One complete `try`/`except`, 3,464 characters, deleted; nothing else touched:

| tree | stage-0 available KV cache | starts? |
| --- | ---: | --- |
| base | 8.01 GiB | yes |
| this PR minus the INT8 block | **8.01 GiB** | **yes** |

Byte-for-byte the control's figure, in both rounds. So the device memory is
entirely the quantization, and the async-output change costs none of it.

## 4. And now the number this PR is about

Same diff minus the INT8 block, against the untouched base:

| round | dies (ctl / PR) | ctl RTF | PR RTF | ΔRTF | ctl TTFT | PR TTFT | **ΔTTFT** |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: |
| 1 | 1 / 0 | 0.46064 | 0.45454 | −1.32% | 298.8 ms | 246.0 ms | **−52.8 ms** |
| 2 | 0 / 1 | 0.45742 | 0.48350 | +5.70% | 290.0 ms | 242.8 ms | **−47.2 ms** |

32/32 completed, 435 output tokens, 135.64 s of audio in all four arms.

**RTF says nothing here** — the sign follows the die, and two identically
configured arms drift ±4–6% between rounds on these machines. **TTFT does**:
−52.8 and −47.2 ms, same sign and nearly the same size under both die
assignments, about −17%. You report a +2.5% RTF regression; I can neither
confirm nor refute that at this resolution, but the deferred builder is worth a
consistent ~50 ms of first-token latency on this workload, and that is a
defensible headline for it.

## Suggestion

Split the INT8 quantization into its own PR. It is an unrelated idea, it is the
single thing that stops this one from starting on a 64 GiB die, and on its own
it deserves its own gate and its own accuracy numbers. What is left here is a
clean latency win that currently cannot be measured by anyone who applies the
branch as-is.

vllm_omni/platforms/npu/worker/npu_ar_model_runner.py
```diff
@@ -1056,6 +1255,226 @@ def propose_draft_token_ids(sampled_token_ids):
...
+        # background builder thread must NOT read self.input_batch /
+        # self.query_start_loc (mutated by the next execute_model step).
+        num_scheduled_tokens_np = getattr(self, "_omni_num_scheduled_tokens_np", None)
+        if num_scheduled_tokens_np is None:
+            req_ids = self.input_batch.req_ids
+            num_scheduled_tokens_np = np.array(
+                [scheduler_output.num_scheduled_tokens[rid] for rid in req_ids],
+                dtype=np.int32,
+            )
+        query_start_loc_cpu = self.query_start_loc.cpu
```
lishunyang12, line 1268: Copy scheduling metadata and pass captured request IDs into background postprocessing.

tests/platforms/npu/worker/test_npu_ar_model_runner.py
```diff
@@ -0,0 +1,100 @@
...
+run on CPU without an NPU device.
+"""
+
+from types import SimpleNamespace
+
+import pytest
+import torch
+
+from vllm_omni.outputs import OmniModelRunnerOutput
+from vllm_omni.platforms.npu.worker.npu_ar_model_runner import (
```
lishunyang12, line 16: Isolate NPU imports so these CPU tests collect without vllm_ascend.

PR #5990, opened 2026-08-10
[Kernel] Fuse Q/K RMSNorm and RoPE

bobboli:
## Summary

- add a shared Q/K RMSNorm + packed RoPE custom-op boundary
- fuse normalization and rotation into one Triton pass per Q/K tensor
- avoid materializing normalized Q/K and rotary-product intermediates
- retain an eager fallback for unsupported devices, dtypes, and geometries

The CUDA fast path is integrated into the diffusion attention path used by MiniMax H3 and currently specializes its BF16 `head_dim=128`, `rotary_dim=96` geometry. The public layer API and fallback are model-independent.

## Performance

Minimax H3 on 4x NVIDIA B300 SXM6 with TP1, Ulysses4, Ring1, dense BF16 `TRTLLM_ATTN`, 1344x768 output, 243 frames, 50 configured inference steps, and seed 0, steady denoising latency decreases from **109.687 s to 106.205 s** (**-3.18%**, **1.033x**). Each revision used one warmup followed by one measured request.

## Nsight Systems

Unfused:
image

Fused:
image

## Testing

- `pytest -c /dev/null -q tests/diffusion/layers/test_fused_qk_norm_rope.py` (`3 passed`)
- `ruff check` on changed Python files
- `ruff format --check` on changed Python files
- `git diff --check`

Related to #5700.

PR #6186, opened 2026-08-14
[Model] Add MiniMax Music 3 text-to-music

linyueqian:
## Purpose

Adds [`MiniMaxAI/MiniMax-Music3`](https://huggingface.co/MiniMaxAI/MiniMax-Music3): lyrics plus a style caption in, a 32 kHz stereo song out, served on `/v1/audio/speech`.

This is a text-to-music model on the speech endpoint, not a TTS model with a musical voice. There is no speaker, no reference audio and no temperature: what a caller controls is the lyrics, the caption, the seed and the length.

| Component | Spec |
|---|---|
| Backbone | Qwen3 decoder, 36 layers, hidden 4096, GQA 32/8 |
| Depth decoder | 4 layers; 7 residual codebooks of 1024 |
| Frame | 8 codebooks at 25 fps; `c0` in the backbone vocabulary, `c1..c7` in the depth decoder |
| Guidance | Classifier-free, scale 1.5, `c0` masked to the conditioned branch's top 50 |
| Acoustic stage | Flow-matching transformer (36 layers, dim 2048) plus a DAC-style decoder |
| Window | 200 frames per acoustic chunk on a 100-frame hop |
| Output | 32 kHz stereo |

## Design notes

**Guidance is mandatory, so every request is two engine rows.** Per-request state is therefore keyed by CFG pair rather than by request. The cond/uncond pair-locking previously owned by Audex is promoted to `models/common/cfg_pairing.py`, and its install gate is generalized from a logits-processor class-name substring match to a declared capability, so a model that owns its own sampling can use the pairing without owning a logits processor. Audex behaviour is unchanged.

**No new per-model branch in shared serving.** The model ships one adapter under `tts_adapters/`; `serving_speech.py` is net one line shorter, and the TTS adapter ratchet passes.

**Weights come from the repository's diffusers-style component folders.** `language_model/` is a plain `Qwen3ForCausalLM`, so the backbone loads through vLLM natively; the audio embedding table, depth decoder and the three acoustic components are loaded explicitly and strictly.

## Engine fixes

Two bugs surfaced here that affect any classifier-free-guidance model, not just this one:

* The CFG companion never received `upgrade_to_omni_request`, so it reached the worker with no `additional_information` and therefore no `global_request_id`, and could not be matched to its partner.
* Pair ids derived from `request.request_id` used the internal UUID assigned at admission, while the companion suffix survives only on `external_req_id`, so the two members registered as unrelated pairs and were never held for each other.

## Test plan

Verified on one H200-class card, single GPU:

| frames | audio | wall |
|---|---|---|
| 250 | 10.00 s | 5.84 s |
| 750 | 30.02 s | 20.57 s |
| 1500 | 52.38 s | 38.88 s |

Durations are exact: `max_new_tokens` counts audio frames at 25 per second.

**Correctness against the reference implementation.** Compared against a golden set generated from the upstream model, the sampled AR codes match on **all eight codebooks for the first four frames**, then diverge as floating-point differences between fused and eager kernels compound. Four consecutive exact frames, with `c0` drawn from a 16,384-way vocabulary, pins the prompt construction, the code offset, the guidance blend and mask, the seeded sampler, and the depth decoder's weight mapping. Bit-identical end-to-end reproduction is not achievable while using vLLM's fused kernels, and is not claimed.

**Determinism.** Same seed gives byte-identical audio across runs. The depth decoder pins SDPA to flash/math because cuDNN's fused attention is not bit-reproducible for grouped-query shapes on this hardware, where one bfloat16 ULP flips a sampled code.

**Unit tests:** 89 CPU tests covering prompt normalization, the CFG null-token contract, window arithmetic, the frame-buffer lifecycle, row-to-request binding and the pairing patches.

**E2E:** weekly rather than per-PR, since the checkpoint is 27 GB across five components and a request occupies two decode rows. Two success cases in `tests/e2e/online_serving/test_minimax_music3.py` and seven rejection cases in `tests/dfx/reliability/invalid_param_test/test_invalid_audio_speech.py`; all are collected by the existing marker-driven weekly sweeps, so no pipeline YAML changes.

```bash
pytest -sv tests/e2e/ -m "slow and H100 and tts" --run-level full_model
pytest -sv tests/dfx/reliability/invalid_param_test/ -m "slow and H100"
```

## Known limitations

* **Concurrency scaling is the weak point.** At 8 concurrent requests the acoustic stage needs roughly 39 s after the AR stage finishes, against 1.5 s at concurrency 1; it runs float32 and eager and now caps concurrent throughput. Improving it is follow-up work.
* `max_model_len` is 10,240, so a maximum-length prompt (5,000 tokens) and a maximum-length song (9,000 frames) cannot both fit. This is a property of the checkpoint; long captions shorten the maximum song, and the deploy config says so.
* A request ended by the engine mid-song, for instance on the context limit, loses its final window. The frame budget is reached first in practice.
* Per-frame decode cost is around 26 ms against a roofline near 8 ms and grows with song length. Stage-1 contention on a shared card is a plausible cause but was not isolated.

cc @Sy0307


Sy0307 (changes requested):
## Request Changes

Thanks for the PR. However, there are still two correctness blockers that should be addressed before this can be merged.

### 1. Batched requests receive metadata from the last request in the batch

In `vllm_omni/model_executor/models/minimax_music3/talker.py:937`, `latent` and `codes` are kept as per-request lists, while all per-request metadata is merged into a single dictionary:

```python
merged = {}
for meta in metas:
    merged.update(meta)
```

After `flatten_payload()` and `to_payload_element()`, these scalar metadata fields are no longer partitioned by request index. As a result, the last request's `audio_seed`, `num_processed_tokens`, and `last_chunk` are broadcast to every request in the batch.

I reproduced this on the PR HEAD using two requests with seeds `111/222`, positions `100/200`, and different terminal flags. Both requests received the second request's values: `222/200/true`.

This can contaminate request seeds, misplace acoustic windows, and allow one request's `last_chunk=True` to terminate another request prematurely. Please preserve metadata as batch-aligned per-request values and add a regression test with different seeds, positions, and terminal flags in the same batch.

### 2. CFG parent and companion admission is not atomic

In `vllm_omni/engine/async_omni_engine.py:1289`, the parent request is enqueued before its CFG companion is expanded and processed. If `prompt_expand_func` fails, the error is logged and the function returns. If companion input processing raises an exception, the parent has already been admitted and is not rolled back.

After 512 scheduling steps, the scheduler deregisters the pair and claims that the request will continue with unguided decoding. However, MiniMax Music 3 has no such fallback: `_resolve_pairs()` excludes incomplete pairs, and `sample()` emits only `_HOLD_TOKEN` when no complete pair is available.

This can leave an orphan request consuming scheduler and KV-cache resources without producing audio. The caller may also receive an exception while the parent request continues running in the background. Parent and companion requests should be fully constructed and validated before admission, so the entire group either succeeds or fails and is cleaned up atomically. Tests should cover prompt expansion failure, companion processing failure, and single-member preemption.

**Conclusion: Request Changes.** Both issues violate request-level correctness under batching or failure conditions, and neither is covered by the current single-request happy-path tests.

Sy0307:
Follow-up: I opened #6242 as a stacked Draft PR for the H200 acoustic-performance work. It depends on #6186 and will remain draft. After the correctness blockers here are addressed and #6186 is merged, I will rebase #6242 onto main, reduce it to the optimization-only diff, and rerun version-aligned performance and quality validation.

vllm_omni/model_executor/models/registry.py
```diff
@@ -445,6 +445,17 @@
         "qwen3_vl",
         "AuraQwen3VLForConditionalGeneration",
     ),
+    ## MiniMax-Music3 (text-to-music; AR talker -> flow-matching acoustic decoder)
```
hsliuustc0106, line 460: [P1] This PR silently removes dots.tts entirely, but is titled and described as a pure model addition

Beyond the MiniMax Music 3 addition, the diff deletes the whole `DotsTTSForConditionalGeneration` stack (~5,300 lines): the 9 model files, `transformers_utils/configs/dots_tts.py`, `deploy/dots_tts.yaml`, `recipes/rednote-hilab/dots.tts.md`, the offline example, its unit tests, and the `supported_models.md` row. The PR body does not mention any of this.

This is a breaking change for anyone serving `rednote-hilab/dots.tts-soar` (unknown-architecture error on upgrade), and it collides with open work: #6235 is adding dots.tts online serving and #6174 adds its weekly e2e coverage, both against code this PR deletes. Nothing on main deprecates the model, so there is no sanctioned removal to align with.

Smallest fix: restore all dots_tts deletions in this PR. If removal is actually intended, it needs its own PR with maintainer sign-off, a deprecation cycle or release-note entry, and coordination with #6235/#6174.

hsliuustc0106: Correction — retracting this comment, it was based on a wrong comparison base. The branch forked from main at `dbc0dd6d` (Aug 13), i.e. before dots.tts (#4765) and the fused QK-norm+RoPE kernel (#5990) landed, and was never rebased. Diffed against the actual merge base, this PR deletes nothing; merging keeps both intact. Sorry for the noise.

The actionable residue: please rebase onto current main before merge. The engine and config files here were written against the pre-#4765/#5990 tree, and the two engine bugs fixed in this PR were reproduced and patched in code paths that main has since continued to evolve (33 commits, including the `stage_args` YAML loader removal in #6200).

docs/models/supported_models.md
```diff
@@ -70,6 +70,7 @@ th {
 | `SoulXSingerPipeline` | SoulX-Singer (SVS) | `Soul-AILab/SoulX-Singer` | ✅︎ | | | | — |
 | `SoulXSingerSVCPipeline` | SoulX-Singer-SVC | `Soul-AILab/SoulX-Singer` (`model-svc.pt`) | ✅︎ | | | | — |
 | `AudioXPipeline` | AudioX | `zhangj1an/AudioX` | ✅︎ | ✅︎ | | | — |
+| `MiniMaxMusic3ForConditionalGeneration` | MiniMax Music 3 (text-to-music) | `MiniMaxAI/MiniMax-Music3` | ✅︎ | | | | — |
```
hsliuustc0106, line 81: The architecture string here (`MiniMaxMusic3ForConditionalGeneration`) matches neither registry key this PR adds (`MiniMaxMusic3TalkerForConditionalGeneration` / `MiniMaxMusic3AcousticForConditionalGeneration`). Every other row in this table names an architecture that resolves through `_OMNI_MODELS`. Worth aligning so the column stays machine-greppable — either the per-stage keys or both, since deployment actually keys off `model_type=minimax_music3`.

linyueqian:
Thanks @Sy0307, both confirmed and fixed in b0350f2.

**1. Batched metadata broadcast.** Correct, and the diagnosis was exact. `latent` and `codes` were per-request lists, but metadata was merged into one dict, and the payload is partitioned by indexing list values, so a scalar reached every request unchanged. Metadata is now a list per field aligned to the same batch rows the tensors use.

I reproduced your case as a regression test before fixing, and confirmed it fails on the old code with exactly the values you saw:

```
E       assert 222 == 111     # audio_seed
E       assert 200 == 100     # num_processed_tokens
5 failed, 3 passed
```

and passes after. `tests/model_executor/models/test_minimax_music3_batch_payload.py` covers seeds, window offsets, terminal flags, span length, and a partial batch where one row carries no span.

**2. CFG admission not atomic.** Also correct. The parent was enqueued before the companion was built, and a build failure was logged and swallowed, so the parent decoded alone. As you note this model has no unguided fallback, so it holds capacity for the scheduler hold budget and then produces nothing.

Companions are now built before the parent is admitted, in `_build_cfg_companions`, so the group is all-or-nothing and a failure reaches the caller with nothing enqueued. `_enqueue_cfg_companions` is kept for callers that admit the parent first and cannot roll it back. Tests cover prompt-expansion failure, companion-processing failure, and the empty-expansion path, each asserting nothing was enqueued.

On single-member preemption: that path is still the scheduler dropping the pair, after which `_resolve_pairs` holds the surviving row rather than raising. I have not added a test for it because I could not construct preemption deterministically at unit level; if you know a reliable way to force it I will add one.

Verified: 180 unit tests pass, including the existing Audex CFG lifecycle and serving-guard tests, since the engine change is shared. E2E on one H200-class card is unchanged, 250 frames still gives exactly 10.00 s with an identical spectral fingerprint.

Also noted #6242, thanks for stacking it.

Sy0307 (approved):
LGTM.

PR #6189, opened 2026-08-14
[Diffusion] Add LTX-2.5 Diffusion VAE decoder support

mglyn:
based on #6070

## Summary

This PR makes the canonical LTX-2.5 **Diffusion VAE decoder** the default for all four full/distilled, one/two-stage pipelines. The convolutional VAE remains available through the startup-only `ltx2_use_conv_vae: true` model extra and is still used for I2V encoding.

The implementation reuses Diffusers 0.40's upstream LTX-2.5 DiffVAE model, decoder core, NATTEN processor, and serial tiling. vLLM-Omni keeps the canonical Native-checkpoint adapter, the short-clip fix, and model-specific distributed tile execution. It reads `Lightricks/LTX-2.5/vae/ltx-2.5-video-vae-bf16.safetensors`, converts only decoder tensors in memory, and does not rely on the separately published 834 MB Diffusers decoder weights.

The Native decoder, latent upsampler, and two-stage LoRA use repository-scoped pinned revisions. Video latent normalization now follows the active decoder's own statistics. `kernels==0.14.1` is a runtime dependency; DiffVAE loads the `shi-labs/natten` Hub kernel during startup. Both decoders are untiled by default. `vae_use_tiling` enables serial tiling, while matching `--usp N` and `--vae-patch-parallel-size N` distributes DiffVAE tiles over existing diffusion worker ranks. DiffVAE distributed decode accepts tile mode only.

## Visual comparison

All comparisons below were generated end-to-end with vLLM-Omni's full `LTX2TwoStagePipeline`. Within each pair, the input image, prompt, seed, denoising result, and audio are identical; only the video decoder changes. **Left: convolutional VAE. Right: Diffusion VAE decoder.**

### Reduced structured aliasing

On sharply focused textures containing dense, partly regular spatial frequencies, Diffusion VAE reduces the structured crawling and deformation patterns produced by ConvVAE.

#### Painting

https://github.com/user-attachments/assets/271b80e2-f610-4a79-8264-8cb4ad91e152

#### Watch

https://github.com/user-attachments/assets/43b6db12-0dcb-4a18-8880-2fd91f676cf7

#### Cherry blossoms

https://github.com/user-attachments/assets/38199e8a-ea83-4a8e-aca0-bfa4c20f0dd5

### More stable fine edges

On repeated high-contrast boundaries and thin line structures, Diffusion VAE keeps edge detail more coherent across frames.

#### Screen text

https://github.com/user-attachments/assets/e783de84-ffc6-476d-a109-5e63df05eec5

#### Guilloche

https://github.com/user-attachments/assets/b88a8499-418a-4413-bf4a-54bdd9477590

#### Speaker grid

https://github.com/user-attachments/assets/703d542b-f547-4143-9751-40153319d11c

## Full decoder benchmark

Measured on 4x NVIDIA H200 with BF16 using `1x128x21x24x32` latents and complete `1x3x161x768x1024` output. The Conv/untiled rows retain the cherry-blossom latent results; tiled rows were rerun on `ce44e1246` after the Diffusers 0.40 refactor with a fixed BF16 latent of the same shape. Model loading and CPU copies are excluded; each row has one warmup and three measured runs.

| Decoder | Measured times (s) | Mean (s) | Median (s) | Speedup vs tiled PP=1 | Max-rank peak allocated memory |
| Conv VAE baseline, PP=1, untiled | 0.746 / 0.720 / 0.725 | **0.731** | 0.725 | - | 14.41 GiB |
| Diffusion VAE, PP=1, untiled | 4.960 / 4.954 / 4.957 | **4.957** | 4.957 | **1.33x** | 47.72 GiB |
| Diffusion VAE, PP=1, tiled | 6.572 / 6.565 / 6.575 | **6.571** | 6.572 | 1.00x | 20.16 GiB |
| Diffusion VAE, PP=2, tiled | 3.450 / 3.444 / 3.448 | **3.447** | 3.448 | **1.91x** | 19.10 GiB |
| Diffusion VAE, PP=4, tiled | 1.899 / 1.895 / 1.893 | **1.896** | 1.895 | **3.47x** | 19.10 GiB |

Untiled PP=1 is 24.6% faster than tiled PP=1 but uses 2.37x the peak memory. PP=4 cuts tiled DiffVAE latency by 71.2%; padded gather and rank-0 merge keep max-rank memory near 20 GiB.

## Official decoder parity

The official implementation and vLLM-Omni decoded the same canonical Native weights and normalized latent state with the same decoder RNG, BF16 dtype, and complete Full two-stage geometry (`1x128x16x34x60` to `1x3x121x1088x1920`). The reference was Lightricks `ltx-core@7954dcb0` in its default `CHUNKED_EAGER` mode; vLLM-Omni used Diffusers 0.40.

| Metric | Mean | Worst frame |
|---|---:|---:|
| SSIM | **0.999491** | **0.999009** |
| PSNR | **52.91 dB** | **47.60 dB** |

The pinned official runtime selected its supported Triton neighborhood-attention fallback, while vLLM-Omni used NATTEN; the remaining difference is therefore not expected to be bit-exact.

## End-to-end official pipeline parity

At `70a6e476e`, the pinned official pipeline and vLLM-Omni both used the canonical Native DiffVAE weights with BF16 cuDNN SDPA. The matrix covers T2V and I2V for distilled two-stage (1024x576, 25 frames, 8+3 steps), Full one-stage (512x384, 25 frames, 30 steps), and Full two-stage (512x384, 25 frames, 30+3 steps).

| Pipeline | Task | SSIM mean / min | PSNR | Audio relative L2 | Audio cosine |
|---|---|---:|---:|---:|---:|
| Distilled two-stage | T2V | 0.998733 / 0.998426 | 52.04 dB | 0.01545 | 0.999881 |
| Distilled two-stage | I2V | 0.999204 / 0.999155 | 58.08 dB | 0.00402 | 0.999993 |
| Full one-stage | T2V | 0.998593 / 0.997921 | 45.77 dB | 0.00399 | 0.999993 |
| Full one-stage | I2V | 0.998710 / 0.998676 | 54.90 dB | 0.00452 | 0.999991 |
| Full two-stage | T2V | 0.996187 / 0.994834 | 41.64 dB | 0.00423 | 0.999992 |
| Full two-stage | I2V | 0.998681 / 0.998635 | 54.28 dB | 0.00630 | 0.999981 |

All six cases pass the shared guards: SSIM mean/min >= 0.99, PSNR >= 40 dB, audio relative L2 <= 0.10, and audio cosine >= 0.99. PR #6342's FP32 BWE vocoder path also removes the previous BF16 audio drift.

## Validation

- Pre-merge affected LTX CPU suite: **276 passed**; targeted pre-commit passed.
- Post-merge conflict-targeted VAE/runtime tests: **25 passed**.
- Official end-to-end DiffVAE matrix above: **6/6 passed** on NVIDIA H200.
- Real 2-GPU NATTEN smoke test: PP=1 and PP=2 were element-for-element identical.
- Full 4-GPU benchmark: PP=1/2/4 had the same output SHA256 and were element-for-element identical.


lishunyang12:
Please rebase.

vllm_omni/diffusion/models/ltx2/ltx2_diffusion_decoder.py
```diff
@@ -0,0 +1,1175 @@
+# Copyright 2026 Lightricks and The HuggingFace Team. All rights reserved.
+#
+# Copied from Hugging Face Diffusers' LTX-2.5 diffusion decoder. Keep this
```
hsliuustc0106: [P2] No numerical comparison against the canonical reference decoder — ltx2_diffusion_decoder.py:3

The PR body validates quality visually (conv vs DiffVAE side-by-sides) and parallel determinism numerically (tiled PP=1 vs 2/4 element-exact), but nothing compares output against the upstream diffusers/Lightricks decoder this file is adapted from. The adaptation carries several deliberate deltas — device-adjusted fp32 RoPE frequencies vs the reference's float64, token-tiled SwiGLU, NATTEN via the `kernels` Hub package, and the AdaLN gates that are "folded into the following linear weights at conversion time" (LTX2VideoVaeAdaLNZero docstring, ~line 419) — each a silent-drift risk that visual inspection would not catch, and the last one leaves it unclear whether the published `diffusion_decoder` checkpoint loads as-is or was re-converted.

Since the pinned diffusers doesn't expose the component yet, could you add a fixed-latent / fixed-seed small-clip comparison against the upstream implementation at a pinned revision (or state why it can't run and what the closest runnable reference is)? Untiled-vs-tiled exactness is not expected by design, so only a reference comparison closes copy fidelity.

vllm_omni/diffusion/models/ltx2/ltx2_components.py
```diff
@@ -604,6 +627,30 @@ def initialize_pipeline_components(pipeline: Any, od_config: Any) -> None:
...
+            local_files_only=local_files_only,
+            dtype=dtype,
+            revision=revision,
+            prefetch_list=component_subfolders,
+        )
+        # The published decoder is trained and validated with NATTEN. Its
+        # portable FlexAttention fallback materializes an impractically large
+        # block mask at production video sizes, so fail early if the matching
+        # Hub kernel cannot be loaded instead of failing later during decode.
+        pipeline.diffusion_decoder.set_attn_processor(LTX2VideoVaeNeighborhoodNattenProcessor())
```
hsliuustc0106, line 666: [P3] NATTEN failure messages suggest a fallback this integration never allows — ltx2_components.py:644

Boot with `ltx2_use_diffusion_decoder: true` unconditionally installs the NATTEN processor, but both early failures in `LTX2VideoVaeNeighborhoodNattenProcessor.__init__` (ltx2_diffusion_decoder.py:251-262) tell the user to "use the default LTX2VideoVaeNeighborhoodAttnProcessor (FlexAttention) instead" — and there is no config or serve surface that can select it here; that processor is reachable only from tests. A user without `kernels`/a supported GPU gets a message whose suggested remedy is a dead end.

Smallest fix: reword the message to the real remedy (install `kernels`, use a supported GPU — flex is not viable at production grid sizes per the comment in this file), or expose a knob if flex should stay selectable.

vllm_omni/diffusion/models/ltx2/ltx2_diffusion_decoder_distributed.py
```diff
@@ -0,0 +1,233 @@
...
+                    crop_trailing_ghost=is_trailing,
+                    noise_generator=tile_generator,
+                    x_t=x_t,
+                )
+            )
+
+        return tasks, GridSpec(
+            split_dims=(2, 3, 4),
+            grid_shape=(len(plan.temporal_tiles), len(plan.height_tiles), len(plan.width_tiles)),
+            tile_spec={"plan": plan},
```
hsliuustc0106, line 354: [P3] Tile output shapes are already known here — skip the base executor's meta packing — ltx2_diffusion_decoder_distributed.py:167

`_distributed_tile_split` computes every tile's exact decoded shape via `_tiled_pixel_shape_from_features`, but `tile_spec` only carries the plan, so `execute()` takes the base `DistributedVaeExecutor` path: meta tensors plus padded `all_gather`. That's the exact overhead `LTX2VaeExecutor` (autoencoder_kl_ltx2.py:22) was added to skip for the conv VAE — it uses `tile_spec["max_tile_output_shape"]` + `tile_spec["tile_output_shapes"]` to pack without metadata and unpack by known shape.

Filling those two keys here (and constructing the executor accordingly, as the conv VAE does) should be free and trims the padded-gather memory/traffic the benchmark table attributes to the ~20 GiB plateau. Not a correctness issue — the base path is fully general.

vllm_omni/engine/async_omni_engine.py
```diff
@@ -1039,6 +1039,7 @@ def _create_default_diffusion_stage_cfg(kwargs: dict[str, Any]) -> list:
             "streaming_output": kwargs.get("diffusion_streaming_output", False),
             "enable_ar_profiler": kwargs.get("enable_ar_profiler", False),
             "extras": {
+                **(kwargs.get("extras") or {}),
                 "auxiliary_text_encoder": kwargs.get("auxiliary_text_encoder", None),
```
hsliuustc0106: [P3] User-supplied extras keys lose to the top-level kwargs' defaults — async_omni_engine.py:1043

Now that caller extras are merged in, a value set only via `extras` (e.g. `extras={"default_llama_model_id": ...}`) is silently overridden by `kwargs.get("default_llama_model_id", "meta-llama/...")` even when the top-level kwarg was never passed — the hardcoded entries always win the dict merge. Falling back through extras when the top-level kwarg is absent (`kwargs.get(k, user_extras.get(k, default))`) keeps both entry points honest.

vllm_omni/diffusion/models/ltx2/ltx2_diffusion_decoder_distributed.py
mglyn: Thanks for the suggestion. I profiled the two isolated executor communication paths on 2 GPUs using the exact packed-tile geometry from the PR benchmark (1024x768x161, PP=2, six tiles, three per rank, 810 MiB packed per rank). The generic path was 4.253 ms median versus 3.916 ms for the known-shape path, saving 0.337 ms; measured peak allocation differed by only about 2 KiB. Both paths allocate and all-gather the same padded RGB tile tensor, so the optimization only removes the 6-int shape all-reduce and tiny metadata all-gather; it does not reduce the dominant padded-gather memory or traffic. Against the 3.620 s median PP=2 full decode this is about 0.009%. Given the negligible end-to-end benefit and the added DiffVAE-to-ConvVAE-executor coupling/shape bookkeeping, I am keeping the model-agnostic executor here. A rank-0-only or streamed gather would be the meaningful follow-up for the approximately 20 GiB plateau.

vllm_omni/diffusion/models/ltx2/ltx2_components.py
mglyn: Fixed in 6f7ed69c. The production error messages no longer suggest the unavailable FlexAttention fallback. They now state the actionable requirements: allow the trusted Hub kernel download and install the project's pinned kernels dependency.

vllm_omni/engine/async_omni_engine.py
mglyn: Fixed in 6f7ed69c. Promoted extras now follow explicit non-None top-level argument > user extras > built-in default. Added regression coverage for preserving user extras when the top-level value is absent/None and for explicit top-level overrides.

vllm_omni/diffusion/models/ltx2/ltx2_diffusion_decoder.py
mglyn: Thanks for raising this. I reran the comparison with the weight source held constant.

The published `Lightricks/LTX-2.5-Diffusers/diffusion_decoder` weights are numerically different from the decoder inside the current standalone `Lightricks/LTX-2.5/vae/ltx-2.5-video-vae-bf16.safetensors`: after the required layout conversion, 405/407 tensors differ. A comparison that uses each repository published weights therefore includes checkpoint skew and cannot isolate implementation fidelity. This may be an upstream publication/version-skew issue; it is tracked in [LTX-2#290](https://github.com/Lightricks/LTX-2/issues/290) and [the model discussion](https://huggingface.co/Lightricks/LTX-2.5-Diffusers/discussions/12).

With the current standalone Native weights (SHA256 `847e14ca7f3355debca0cea4eaa24ac0fbcdf0061da054ac89ca638a869ddba3`) loaded strictly into both the official `ltx-core` decoder and this Omni implementation, using the same BF16 latent, seed, untiled decode, and combined NATTEN path, the rerun produced:

- SSIM mean/min: `0.999853` / `0.999825`
- PSNR mean: `60.40 dB`
- cosine similarity: `0.999836`
- load result: 0 missing and 0 unexpected keys

The PR body keeps the user-facing quality evidence to the visual ConvVAE-vs-DiffVAE comparisons, where the shown ConvVAE artifacts are visibly reduced. Both weight repositories are official publications, so I do not want to speculate about the intended canonical snapshot or add a temporary checkpoint adapter that may become obsolete if the published Diffusers component is updated.

mglyn: Follow-up: Lightricks has clarified that the decoder in the Native single-file checkpoint is canonical. The implementation now reads that official checkpoint and converts only the decoder tensors at startup, instead of using the separately published Diffusers decoder weights.
Using the same Native weights, input latent, RNG, BF16 dtype, and Full two-stage output geometry, we compared vLLM-Omni against the decoder path used by the official pipeline (CHUNKED_EAGER). The results were:
- SSIM: 0.999472 mean / 0.999118 worst frame
- PSNR: 52.63 dB mean / 48.69 dB worst frame

This confirms that the converted weights and decoder implementation align closely with the canonical official pipeline.

vllm_omni/diffusion/models/ltx2/ltx2_components.py
hsliuustc0106: Extending this thread: `kernels==0.14.1` is still declared only under `[project.optional-dependencies].dev`, while this line runs unconditionally for the new default LTX-2.5 path. A normal runtime install therefore still fails during startup when `kernels` is absent. Please add it to the supported runtime requirements or provide a viable fallback.

vllm_omni/diffusion/models/ltx2/ltx2_components.py
```diff
@@ -604,6 +652,27 @@ def initialize_pipeline_components(pipeline: Any, od_config: Any) -> None:
...
+        vae_patch_parallel_size = int(
+            getattr(getattr(od_config, "parallel_config", None), "vae_patch_parallel_size", 1)
+        )
+        if vae_patch_parallel_size > 1 or getattr(od_config, "vae_use_tiling", False):
+            pipeline.diffusion_decoder.enable_tiling()
+        pipeline.diffusion_decoder.set_parallel_size(
+            vae_patch_parallel_size,
+            mode=getattr(getattr(od_config, "parallel_config", None), "vae_parallel_mode", "tile"),
```
hsliuustc0106, line 674: [P2] `vae_parallel_mode` is silently ignored here. `spatial_shard_height`/`spatial_shard_width` are accepted by the shared config, but `DistributedLTX2VideoDiffusionDecoderModel.tiled_decode()` always dispatches the tile executor. Please reject unsupported modes for this decoder or implement the requested spatial-shard path.

vllm_omni/diffusion/models/ltx2/ltx2_components.py
```diff
@@ -519,18 +537,47 @@ def _load_component(
...
+    revision: str | None,
+) -> DistributedLTX2VideoDiffusionDecoderModel:
+    """Load canonical Native weights into the local Diffusers-compatible class."""
+    config = DistributedLTX2VideoDiffusionDecoderModel.load_config(
+        model,
+        subfolder=_LTX2_DIFFUSION_DECODER_SUBFOLDER,
+        local_files_only=local_files_only,
+        revision=revision,
+    )
+    checkpoint_path = resolve_ltx_artifact(
```
hsliuustc0106, line 567: [P2] Preserve offline and revision semantics for the canonical artifact. `resolve_ltx_artifact()` calls `hf_hub_download()` without `local_files_only` or `revision`; a local `LTX-2.5-Diffusers` snapshot therefore still requires network access for this file, and a pinned Diffusers config can be paired with unpinned native weights. Please thread these options through or make the artifact requirement explicit.

hsliuustc0106 (changes requested):
Please address the two inline issues below. I also extended the existing NATTEN thread with the runtime-dependency blocker.

mglyn:
Latest update at `70a6e476e`:

- Fixed LTX-2.5 video latent normalization/denormalization to use the active decoder's statistics instead of always reusing ConvVAE statistics.
- Pinned Native DiffVAE, latent-upscaler, and two-stage LoRA sidecars with repository-scoped revisions; local model roots no longer incorrectly force independently hosted sidecars into local-only resolution.
- Moved `kernels==0.14.1` into runtime dependencies, added actionable NATTEN startup errors, constrained distributed DiffVAE execution to tile mode, and constructed the latent upsampler on CPU.
- Merged #6342: the LTX BWE vocoder now runs in FP32 like the official pipeline, eliminating the previous BF16 audio drift; official similarity guards are now shared and stricter.

I reran the complete LTX-2.5 official comparison matrix on NVIDIA H200. Both sides used the canonical Native DiffVAE decoder weights; I2V uses ConvVAE only for encoding.

| Pipeline | Task | SSIM mean / min | PSNR | Audio rel. L2 | Audio cosine |
|---|---|---:|---:|---:|---:|
| Distilled two-stage | T2V | 0.998733 / 0.998426 | 52.04 dB | 0.01545 | 0.999881 |
| Distilled two-stage | I2V | 0.999204 / 0.999155 | 58.08 dB | 0.00402 | 0.999993 |
| Full one-stage | T2V | 0.998593 / 0.997921 | 45.77 dB | 0.00399 | 0.999993 |
| Full one-stage | I2V | 0.998710 / 0.998676 | 54.90 dB | 0.00452 | 0.999991 |
| Full two-stage | T2V | 0.996187 / 0.994834 | 41.64 dB | 0.00423 | 0.999992 |
| Full two-stage | I2V | 0.998681 / 0.998635 | 54.28 dB | 0.00630 | 0.999981 |

Result: **6/6 passed** the shared thresholds (SSIM mean/min >= 0.99, PSNR >= 40 dB, audio relative L2 <= 0.10, audio cosine >= 0.99). The previously failing near-silent distilled T2V audio improved from cosine 0.94248 to 0.99988 after the FP32 BWE fix.

vllm_omni/diffusion/models/ltx2/ltx2_components.py
mglyn: Fixed in `d4f1af955`. `kernels==0.14.1` has been moved from the dev extra to `requirements/common.txt`, so normal runtime installs include it. The initialization error remains actionable for unsupported GPUs or disabled remote code. I reran all six LTX-2.5 official comparison cases on the updated runtime; the Omni DiffVAE NATTEN path initialized successfully and all 6/6 passed.

vllm_omni/diffusion/models/ltx2/ltx2_components.py
mglyn: Fixed in `d4f1af955`. The LTX-2.5 distributed DiffVAE now overrides `set_parallel_size()` and raises `ValueError` for every mode other than `tile`, so `spatial_shard_height` and `spatial_shard_width` can no longer be silently accepted. Added a positive tile test and negative height/width tests.

vllm_omni/diffusion/models/ltx2/ltx2_components.py
mglyn: Fixed in `d4f1af955` and `9454a6f82`. Canonical Native artifacts are now pinned to `8a4ff96f581e72bedc1b44367581c49d544a05f1`; model and artifact revisions are repository-scoped, so a pinned Diffusers revision is not reused for `Lightricks/LTX-2.5`. Resolution first checks `<model>/<filename>`, then calls `hf_hub_download(..., revision=<artifact revision>)`. We deliberately do not infer strict offline mode merely from `model` being a local path, because the primary Diffusers model and canonical sidecars are separate repositories; actual offline behavior is preserved through Hugging Face `HF_HUB_OFFLINE`, which uses cached artifacts or fails without network. Tests cover local-file preference, local-model/missing-sidecar Hub fallback, and revision routing.

vllm_omni/diffusion/models/ltx2/ltx2_components.py
```diff
@@ -69,6 +76,18 @@
...
+_LTX2_DIFFUSION_DECODER_SUBFOLDER = "diffusion_decoder"
+
+
+def _ltx2_use_diffusion_decoder(od_config: Any, model_version: str) -> bool:
+    """Select DiffVAE by default for LTX-2.5, with an explicit ConvVAE opt-in."""
+    extras = getattr(od_config, "extras", {}) or {}
+    use_conv_vae = extras.get(_LTX2_CONV_VAE_EXTRA, False)
+    if not isinstance(use_conv_vae, bool):
+        raise TypeError(f"{_LTX2_CONV_VAE_EXTRA} must be a bool, got {type(use_conv_vae)!r}")
+    return model_version == "2.5" and not use_conv_vae
```
hsliuustc0106, line 89: [P1] Making DiffVAE the default changes the valid request boundary. The checked-in 128x128 request passes the current divisibility-only validation, but our online Full one-stage T2V and I2V smokes both returned HTTP 500 during decode: latent (T,H,W)=(4,4,4) is smaller than the NATTEN (3,7,7) kernel. Please reject unsupported DiffVAE sizes before denoising and change the shared e2e smoke to 256x256, which passed T2V and I2V across all four pipelines locally.

recipes/LTX/LTX-2.5.md
```diff
@@ -28,6 +36,41 @@ latent upsampler, and run a three-step refinement stage. Select the class with
...
+```bash
+vllm serve Lightricks/LTX-2.5-Diffusers \
+  --omni \
+  --model-class-name LTX2Pipeline \
+  --stage-overrides '{"0":{"extras":{"ltx2_use_conv_vae":true}}}'
+```
+
+Both decoders are untiled by default. Set `vae_use_tiling` for memory-saving
+serial tiling; DiffVAE tiles only above 80 frames or 768 pixels in either spatial
+dimension. For distributed decode, use matching `--usp N` and
```
hsliuustc0106: [P2] --usp is a vllm serve alias only. The offline T2V and I2V scripts documented below expose --ulysses-degree, so following this sentence for an offline run produces an unrecognized-argument error. Please distinguish serving --usp N from offline --ulysses-degree N; both pair with --vae-patch-parallel-size N.

hsliuustc0106:
Local validation at frozen head `70a6e476e609d81e88fa3624c52dd9590711cd26`:

- All GPU processes were launched through `gpu run`.
- Hardware: 1x NVIDIA L20X, 143771 MiB, driver 570.133.20.
- Runtime: Python 3.12, vLLM 0.27.0, PyTorch 2.13.0+cu130, Diffusers 0.40.0, Transformers 5.14.1, `kernels==0.14.1`.
- Weights/cache: `/data/model`; the missing canonical Native DiffVAE checkpoint (~1.47 GB) was downloaded before the matrix.
- Server configuration: default DiffVAE, layerwise offload, cuDNN DiT attention. Requests used 256x256, 9 frames, 24 fps, seed 42, and the canonical Stage-1 step counts (30 Full / 8 distilled); two-stage refinement retained its fixed 3 steps.

This complements the official-reference matrix by exercising all four public classes through the online video path, including distilled one-stage:

| Pipeline | T2V | I2V |
| --- | ---: | ---: |
| Full one-stage | PASS (~22 s) | PASS (~22 s) |
| Full two-stage | PASS (~30 s) | PASS (~30 s) |
| Distilled one-stage | PASS (~8 s) | PASS (~6 s) |
| Distilled two-stage | PASS (~10 s) | PASS (~8 s) |

All 8 requests produced playable 9-frame MP4 files with 48 kHz stereo audio. Timings are single-request smoke timings after server readiness, not canonical-resolution benchmarks.

CPU evidence on the same head:

- Full affected CPU suite: **415 passed, 1 skipped**.
- Focused recipe/pipeline rerun: **164 passed**.
- Config/entrypoint rerun: **173 passed, 1 skipped**.
- The four class -> recipe -> component-profile mappings and documented defaults match the pinned official behavior. Full two-stage deliberately routes video from Stage 2 and audio from Stage 1; distilled two-stage routes both from Stage 2.

Boundary reproduction:

- The checked-in 128x128 online request is accepted by request validation.
- Full one-stage T2V and I2V both fail after denoising with HTTP 500 because DiffVAE reaches `(T,H,W)=(4,4,4)` with kernel `(3,7,7)`.
- 256x256 passes all eight cases above. A useful regression would assert an early client error for unsupported DiffVAE geometry and use 256x256 for the shared online smoke.

vllm_omni/diffusion/models/ltx2/ltx2_components.py
mglyn: Fixed in `c9d2dce91`. When DiffVAE is active, request validation now derives the minimum latent height/width from the first decoder-stage NATTEN kernel and rejects undersized spatial latents before denoising, including both request and latent dimensions in the error. The ConvVAE opt-in keeps its existing size boundary. The shared four-pipeline online smoke now uses 256x256, and unit coverage verifies the DiffVAE rejection plus the unaffected ConvVAE path. The LTX CPU test file passes 111/111 and affected-file pre-commit passes.

recipes/LTX/LTX-2.5.md
mglyn: Fixed in `c9d2dce91`. I removed the entrypoint-specific flags from the LTX recipe. It now states that DiffVAE supports distributed VAE decode and links the shared VAE Parallelism Guide, which documents rank reuse and the supported online/offline configuration surfaces.

vllm_omni/diffusion/models/ltx2/ltx2_components.py
hsliuustc0106: Verified at 9b0b52074: `kernels==0.14.1` now ships in `requirements/common.txt` (removed from the dev extra), the NATTEN init failure raises the actionable requirements message (`ltx2_components.py:_create_ltx25_natten_processor`), and `test_ltx25_natten_failure_has_actionable_omni_remedy` pins both. Runtime-install failure path closed.

vllm_omni/engine/async_omni_engine.py
hsliuustc0106: Verified at 9b0b52074: promoted extras follow explicit top-level > caller extras > default (`async_omni_engine.py:1044-1056`), and the same file now also folds stage-0 `extras` from `--stage-overrides` into the unregistered fallback (:1205-1227) so the recipe's ConvVAE opt-in works on local checkpoints; both behaviors have regression tests in `test_async_omni_diffusion_config.py`. I traced the registered path separately: `stage_0_extras` survives `build_stage_runtime_overrides` (extras is not blacklisted) and lands in `engine_args["extras"]` via `to_omegaconf`, consumed by `OmniDiffusionConfig.from_kwargs` (stage_init_utils.py:1471). Both entry points are honest.

vllm_omni/diffusion/models/ltx2/ltx2_components.py
hsliuustc0106: Verified at 9b0b52074: the DiffVAE executor overrides `set_parallel_size()` and raises `ValueError` for any non-`tile` mode (`ltx2_diffusion_decoder_distributed.py:89-92`), with positive and negative coverage. Silent-ignore closed.

vllm_omni/diffusion/models/ltx2/ltx2_components.py
hsliuustc0106: Verified at 9b0b52074: `resolve_ltx_artifact` (`ltx2_components.py:259`) prefers a local `<model>/<filename>` sidecar, uses the model revision only when model and artifact share a repo, otherwise the independently pinned artifact revision, and wraps Hub failures in an actionable `FileNotFoundError`. The accuracy harness pins `LTX25_OFFICIAL_MODEL_REVISION = 8a4ff96...` and asserts `video_vae_path` provenance. Offline/revision semantics closed.

vllm_omni/diffusion/models/ltx2/ltx2_components.py
hsliuustc0106: Verified at 9b0b52074: when DiffVAE is active, `__call__` derives the minimum latent H/W from `config.decoder_stage_kernels[0]` and rejects undersized requests via `validate_pipeline_request(min_video_latent_spatial_size=...)` before denoising (`ltx2_runtime.py:256-272`, `ltx2_request.py:101-115`); the shared online smoke now runs 256x256 across all four pipelines, with unit coverage for both the rejection and the ConvVAE boundary. HTTP-500-at-decode closed.

recipes/LTX/LTX-2.5.md
hsliuustc0106: Verified at 9b0b52074: entrypoint-specific flags are gone from the recipe; it links the VAE Parallelism Guide (target exists), and I checked its remaining concrete claims against code — tiling engages above the diffusers-default 80 frames / 768 px, and `decode_timestep`/`decode_noise_scale` are skipped on the DiffVAE path (`ltx2_runtime.py:554-561`). Doc drift closed.

hsliuustc0106 (approved):
Re-reviewed at 9b0b52074. All open threads verified fixed with pinning tests: kernels shipped in runtime requirements, non-tile parallel modes rejected, artifact resolution revision-pinned and local-first, undersized DiffVAE requests rejected before denoising (smoke at 256x256), recipe rewritten against the shared VAE Parallelism Guide — its concrete claims cross-checked against diffusers v0.40.0 defaults (80 frames / 768 px tiling) and the runtime conditioning path. The stage-0 extras folding into the unregistered fallback is verified end-to-end for both serve and offline entry points. Non-model surface (deps, docs, tests, engine wiring) clean. LGTM.

PR #6190, opened 2026-08-14
[Refactor] Rename batch chat completion as legacy fallback

khairulkabir1661:
## Summary
- Rename `create_batch_chat_completion` → `_create_batch_chat_completion_legacy`
- Add new `create_batch_chat_completion` that delegates to the legacy method
- Prepares for an optimized path that submits directly to the engine (follow-up PRs)

## Test plan
- [x] All 9 existing tests pass unchanged - pytest tests/entrypoints/openai/test_batch_chat_completions.py -v


khairulkabir1661:
Consolidated into #6193 for easier review and to avoid sequential dependency chain.

PR #6191, opened 2026-08-14
[Perf] Add render_batch_chat_request for validate-once batch preprocessing

khairulkabir1661:
## Summary
- Add `render_batch_chat_request()` that validates all 12 shared checks once per batch
- Preprocess each conversation individually after validation
- Avoids O(N) redundant validation (model, engine, LoRA, tokenizer, template, modalities, etc.)
- 3 new unit tests covering validation path

Depends on #6190

## Test plan
- [x] 17 tests pass (9 existing + 3 new render tests) - pytest tests/entrypoints/openai/test_batch_chat_completions.py -v


khairulkabir1661:
Consolidated into #6193 for easier review and to avoid sequential dependency chain.

PR #6194, opened 2026-08-14
[Model] Add NeuTTS-Air TTS support

XCai-1204:
# Summary

Add end-to-end NeuTTS-Air support to vLLM-Omni, including offline inference, the OpenAI-compatible Speech API, asynchronous chunked streaming, request isolation, and a production-oriented single-GPU deployment configuration.

## Architecture

The integration uses a two-stage pipeline:

1. Stage 0 reuses vLLM's native Qwen2 implementation to generate NeuTTS speech-token IDs.
2. The stage input processor converts speech-token IDs into NeuCodec code IDs and prepares synchronous or streaming decoder windows.
3. Stage 1 lazily loads the external NeuCodec runtime and decodes the codes into 24 kHz waveform chunks.

The request path is:

```text
target text + reference text + reference audio
  -> NeuTTS-Air processor
  -> phonemes + reference NeuCodec codes
  -> NeuTTS prompt token IDs
  -> native vLLM Qwen2 generation
  -> speech-token filtering and code conversion
  -> NeuCodec decoding
  -> 24 kHz audio
```

Reference audio is normalized to mono 16 kHz before NeuCodec encoding. Callers may provide precomputed reference codes to avoid repeated CPU encoding.

## What Changed

- Register NeuTTS-Air model detection, model classes, and pipeline configuration.
- Add the NeuTTS prompt processor while retaining vLLM's native Qwen2 execution path.
- Add synchronous and asynchronous Stage 0 to Stage 1 transfer.
- Add stateful streaming NeuCodec decoding with lookback, lookahead, overlap-add, terminal flushing, and abort cleanup.
- Add a NeuTTS-Air Speech API adapter with request validation and sampling overrides.
- Add offline inference and online serving documentation.
- Add the `neutts-air` optional dependency extra.
- Add focused tests for discovery, prompting, transfer, decoding, serving, streaming, and cleanup.

## Interfaces

- Offline inference through `examples/offline_inference/text_to_speech/neutts_air/end2end.py`.
- OpenAI-compatible `/v1/audio/speech` requests using `input`, `ref_audio`, and `ref_text`.
- Synchronous full-waveform generation.
- Asynchronous PCM/WAV chunk streaming.
- Concurrent request isolation and cleanup after completion, abort, and request errors.

## Correctness Evidence

Five representative English cases were checked across prompt construction, Stage 0 generation, and audio decoding.

- Phoneme and prompt-token parity: 5/5 exact against the original implementation.
- Stage 0 structure: 5/5 produced valid speech-token IDs and a terminal stop token.
- Fixed-code NeuCodec parity: exact waveform equality, maximum absolute difference `0.0`.
- Streaming decoder parity: maximum absolute difference `7.276e-12` with complete terminal flushing.
- Async end-to-end run: six chunks, terminal `finished=true`, finite 24 kHz audio, and no clipping.
- Two concurrent requests completed independently.
- Aborting one request did not affect another request, and the engine remained reusable afterward.

BF16 Stage 0 output is not expected to be token-for-token identical to Hugging Face generation. The first candidate differences have small score margins and are amplified by autoregressive generation. FP32 eager execution restored same-batch reproducibility, locating this behavior in the low-precision optimized Qwen2 execution path rather than prompt construction, stage transfer, or NeuCodec decoding.

## Performance Evidence

Measurements were collected on one RTX 5090 after warmup. Core benchmarks exclude startup, reference-audio encoding, and optional watermark postprocessing.

| Metric | Original NeuTTS-Air | vLLM-Omni |
|---|---:|---:|
| Median latency | 3.370 s | 0.828 s |
| Median RTF | 0.629 | 0.152 |
| Approximate latency speedup | - | 4.07x |

Additional vLLM-Omni measurements:

- Batch 4: 3.773 requests/s and 20.845 generated audio-seconds/s.
- Streaming with precomputed reference codes: median TTFA 0.151 s, median total latency 0.780 s, nine chunks.
- Stage 0: 282 generated codes in 0.465 s, approximately 607 tokens/s.
- Stage 1: 355 codes decoded in approximately 6.7 ms.
- Default 512 MiB Stage 0 KV cache deployment: approximately 6.6 GiB peak GPU memory in the validated batch-4 run.
- CPU reference encoding for a 13.06-second clip: median 4.02 s; precomputed reference codes avoid this cost.

## Tests

```bash
python -m pytest -q   tests/config/test_neutts_air_pipeline.py   tests/entrypoints/openai_api/test_neutts_air_tts_adapter.py   tests/model_executor/models/test_neutts_air_talker.py   tests/model_executor/models/test_neutts_air_code2wav.py   tests/model_executor/stage_input_processors/test_neutts_air_stage_input_processors.py   tests/distributed/omni_connectors/test_chunk_transfer_adapter.py
```

Result: `98 passed`.

Relevant pre-commit checks also pass, including Ruff check/format, YAML validation, whitespace and line-ending checks, typos, pytest markers, TTS adapter migration, and pickle import checks. No GitHub workflow files are changed.

## Limitations

- English synthesis only.
- Voice cloning requires one reference-audio and reference-text pair.
- Exact prompt parity requires the expected eSpeak backend behavior.
- Raw reference-audio encoding currently runs on CPU and can dominate first-request latency.
- BF16 generation is structurally correct but not bitwise identical to the Hugging Face baseline.
- The core benchmarks exclude optional watermark postprocessing.
- Performance and memory figures above are from a single RTX 5090.

### DIFF
--- a/docs/user_guide/examples/online_serving/text_to_speech.md
+++ b/docs/user_guide/examples/online_serving/text_to_speech.md
@@ -22,6 +22,7 @@ For the full list of supported architectures across all modalities, see
 | higgs-audio v2 | `bosonai/higgs-audio-v2-generation-3B-base` | ✓ (`ref_audio`+`ref_text`) | ✓ (codec_streaming) | — | — |
 | GLM-TTS | `zai-org/GLM-TTS` | ✓ (`ref_audio`+`ref_text`, required) | ✓ (PCM stream) | — | ✓ |
 | OmniVoice | `k2-fsa/OmniVoice` | (offline only) | — | — | — |
+| NeuTTS-Air | `neuphonic/neutts-air` | ✓ (`ref_audio`+`ref_text`, required) | ✓ (PCM stream) | — | — |
 | Qwen3-TTS | `Qwen/Qwen3-TTS-12Hz-1.7B-{CustomVoice,VoiceDesign,Base}` | ✓ (Base) | ✓ (PCM + WebSocket) | ✓ (presets + `/v1/audio/voices` upload) | ✓ (standard + FastRTC) |
 | VoxCPM2 | `openbmb/VoxCPM2` | ✓ | ✓ (AudioWorklet via gradio) | — | ✓ |
 | Voxtral TTS | `mistralai/Voxtral-4B-TTS-2603` | ✓ (gated upstream) | ✓ | ✓ (presets) | ✓ |
@@ -303,6 +304,38 @@ The client supports `--api-base`, `--model`, `--text`, `--response-format`, `--l
 - Voice cloning and voice design require offline inference; see the [offline OmniVoice section](https://github.com/vllm-project/vllm-omni/tree/main/examples/offline_inference/text_to_speech/README.md#omnivoice).
 
 ---
+
+## NeuTTS-Air
+
+### Prerequisites
+
+```bash
+pip install -e '.[neutts-air]'
+```
+
+### Launch
+
+```bash
+vllm serve neuphonic/neutts-air --omni --host 0.0.0.0 --port 8000
+```
+
+### Voice cloning
+
+```bash
+curl http://localhost:8000/v1/audio/speech \
+  -H "Content-Type: application/json" \
+  -d '{
+    "model": "neuphonic/neutts-air",
+    "input": "Hello from NeuTTS-Air.",
+    "ref_audio": "file:///path/to/reference.wav",
+    "ref_text": "Transcript of the reference audio.",
+    "response_format": "wav",
+    "stream": true
+  }' \
+  --output neutts_air.wav
+```
+
+NeuTTS-Air currently supports English synthesis and produces 24 kHz audio. Each request requires one reference-audio and reference-text pair. Reference audio is encoded on CPU unless precomputed reference codes are supplied. The default deployment supports up to four concurrent sequences.
 
 ## Qwen3-TTS
 
--- a/vllm_omni/entrypoints/openai/tts_adapters/neutts_air.py
+++ b/vllm_omni/entrypoints/openai/tts_adapters/neutts_air.py
@@ -1,0 +1,103 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+"""NeuTTS-Air TTS serving adapter."""
+
+import copy
+from typing import TYPE_CHECKING
+
+import numpy as np
+
+from vllm_omni.entrypoints.openai.tts_adapters import register_tts_adapter
+from vllm_omni.entrypoints.openai.tts_adapters.base import (
+    ARTTSAdapter,
+    PreparedRequest,
+)
+
+if TYPE_CHECKING:
+    from vllm_omni.entrypoints.openai.protocol.audio import (
+        OpenAICreateSpeechRequest,
+    )
+
+
+NEUTTS_MIN_NEW_TOKENS = 50
+NEUTTS_DEFAULT_MAX_NEW_TOKENS = 512
+NEUTTS_SPEECH_GENERATION_END_TOKEN_ID = 151670
+
+
+@register_tts_adapter
+class NeuTTSAirAdapter(ARTTSAdapter):
+    """Translate OpenAI speech requests into the NeuTTS-Air pipeline input."""
+
+    stage_keys = frozenset({"neucodec"})
+    name = "neutts_air"
+    max_new_tokens_min = NEUTTS_MIN_NEW_TOKENS
+
+    def validate(self, request: "OpenAICreateSpeechRequest") -> str | None:
+        if not request.input or not request.input.strip():
+            return "Input text cannot be empty"
+        if request.ref_audio is None:
+            return "NeuTTS-Air requires 'ref_audio' (reference audio for voice cloning)"
+        if isinstance(request.ref_audio, list):
+            return "NeuTTS-Air requires exactly one 'ref_audio'"
+
+        fmt_err = self.ctx.server._validate_ref_audio_format(request.ref_audio)
+        if fmt_err:
+            return fmt_err
+        if not request.ref_text or not request.ref_text.strip():
+            return "NeuTTS-Air requires 'ref_text' (transcript of the reference audio)"
+        if request.max_new_tokens is not None:
+            if request.max_new_tokens < self.max_new_tokens_min:
+                return f"max_new_tokens must be at least {self.max_new_tokens_min}"
+            if request.max_new_tokens > self.max_new_tokens_max:
+                return f"max_new_tokens cannot exceed {self.max_new_tokens_max}"
+        return None
+
+    async def build(
+        self,
+        request: "OpenAICreateSpeechRequest",
+        sampling_params_list: list,
+        has_inline_ref_audio: bool,
+    ) -> PreparedRequest:
+        del sampling_params_list, has_inline_ref_audio
+        assert isinstance(request.ref_audio, str)
+        wav_list, sample_rate = await self.ctx.server._resolve_ref_audio(request.ref_audio)
+        prompt = {
+            "prompt": request.input,
+            "multi_modal_data": {
+                "audio": (np.asarray(wav_list, dtype=np.float32), sample_rate),
+            },
+            "mm_processor_kwargs": {
+                "ref_text": request.ref_text,
+            },
+        }
+        return PreparedRequest(
+            prompt=prompt,
+            tts_params={},
+            model_type="neutts_air",
+        )
+
+    def apply_sampling_overrides(
+        self,
+        sampling_params_list: list,
+        request: "OpenAICreateSpeechRequest",
+    ) -> list:
+        if not sampling_params_list:
+            return sampling_params_list
+
+        params = copy.deepcopy(sampling_params_list)
+        stage0 = params[0]
+        requested_max = request.max_new_tokens
+        if requested_max is not None:
+            stage0.max_tokens = int(requested_max)
+        elif stage0.max_tokens is None or stage0.max_tokens < NEUTTS_MIN_NEW_TOKENS:
+            stage0.max_tokens = NEUTTS_DEFAULT_MAX_NEW_TOKENS
+        stage0.min_tokens = NEUTTS_MIN_NEW_TOKENS
+        stage0.detokenize = False
+        stage0.stop_token_ids = [NEUTTS_SPEECH_GENERATION_END_TOKEN_ID]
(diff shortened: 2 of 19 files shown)


XCai-1204:
Author self-review: I reviewed all 19 changed files, including model discovery, native Qwen2 reuse, prompt construction, synchronous and asynchronous stage transfer, NeuCodec decoding, streaming state cleanup, Speech API validation, optional dependencies, documentation, and tests. I specifically checked request isolation, terminal flushing, abort cleanup, local-path leakage, and debug artifacts. The focused suite passes with 98 tests, fixed-code NeuCodec decoding is exact, and real streaming parity and concurrent-request tests pass. Known limitations such as English-only synthesis, CPU reference encoding, and BF16 non-bitwise generation are documented. @vllm-omni-review-bot please run an automated review.

vllm_omni/model_executor/models/neutts_air/neutts_air_talker.py
```diff
@@ -0,0 +1,364 @@
...
+        try:
+            from phonemizer.backend import EspeakBackend
+            from phonemizer.backend.espeak.wrapper import EspeakWrapper
+
+            if not _configure_explicit_espeak_assets(EspeakWrapper):
+                try:
+                    import espeakng_loader
+
+                    EspeakWrapper.set_library(espeakng_loader.get_library_path())
+                    EspeakWrapper.set_data_path(espeakng_loader.get_data_path())
```
Sy0307, line 112: [P1] This makes the standard `.[neutts-air]` installation fail on the first request. The PR declares `phonemizer>=3.0.0`, but the real `phonemizer` `EspeakWrapper` provides `set_library()` and does not provide `set_data_path()`; `data_path` is a read-only instance property. Once `espeakng_loader` imports successfully, this line raises `AttributeError`. The current fake-wrapper test adds this nonexistent API by hand, so it does not cover the real dependency. Please use an API or environment variable supported by phonemizer, and add a real optional-dependency smoke test.

vllm_omni/entrypoints/openai/tts_adapters/neutts_air.py
```diff
@@ -0,0 +1,103 @@
...
+        sampling_params_list: list,
+        request: "OpenAICreateSpeechRequest",
+    ) -> list:
+        if not sampling_params_list:
+            return sampling_params_list
+
+        params = copy.deepcopy(sampling_params_list)
+        stage0 = params[0]
+        requested_max = request.max_new_tokens
+        if requested_max is not None:
```
Sy0307, line 90: [P1] This sampling override is not connected to the production serving path. `serving_speech._prepare_speech_generation()` currently calls only the adapter `validate()` and `build()` methods; it never calls `apply_sampling_overrides()`, and `neutts_air` is also missing from `_SAMPLING_MAX_TOKENS_TTS_MODEL_TYPES`. Therefore an online request with `max_new_tokens=256` passes validation here, but Stage 0 still uses `max_tokens=512` from the deploy YAML. This implementation is currently dead code that is only exercised by a direct unit test. Please wire it into the orchestrator and add a test through the real `_prepare_speech_generation()` path.

vllm_omni/model_executor/models/neutts_air/neutts_air_code2wav.py
```diff
@@ -0,0 +1,398 @@
...
+        self.has_preprocess = False
+        self.has_postprocess = False
+        self.enable_update_additional_information = True
+        self.requires_raw_input_tokens = True
+
+        hf_config = vllm_config.model_config.hf_config
+        self._codec_repo = str(getattr(hf_config, "neutts_codec_repo", DEFAULT_NEUCODEC_REPO))
+        self._codec: nn.Module | None = None
+        self._output_sample_rate = NEUTTS_SAMPLE_RATE
+        self._stream_states: dict[str, _StreamAudioState] = {}
```
Sy0307, line 112: [P1] The state key and the cleanup callback use different request-ID namespaces. Stage 0 writes `external_req_id` into the payload metadata, and `_stream_metadata()` below uses it as the `_stream_states` key; however, the worker passes scheduler internal request IDs to `on_requests_finished()`. This class also does not declare `requires_request_ids=True`, so it cannot receive the internal IDs from the runner. On abort or client disconnect, `pop()` therefore misses the actual entry and the cached GPU waveform frames remain allocated. Please pass internal IDs explicitly, or maintain an internal/external ID mapping, and test abort cleanup with different internal and external IDs.

vllm_omni/model_executor/stage_input_processors/neutts_air.py
```diff
@@ -0,0 +1,253 @@
...
+    reference_codes = _extract_reference_codes(prompt_token_ids)
+    generated_ids = all_token_ids[len(prompt_token_ids) :]
+    target_codes = filter_speech_codes(generated_ids)
+    finished = _request_finished(request, is_finished)
+
+    state_by_request = getattr(transfer_manager, "code_prompt_token_ids", None)
+    if state_by_request is None:
+        state_by_request = {}
+        transfer_manager.code_prompt_token_ids = state_by_request
+    consumed_codes = state_by_request.setdefault(request_id, [])
```
Sy0307, line 154: [P1] The sender-side state created here is not cleaned up on abort or send failure. The generic adapter reaches `cleanup_sender()` only after a terminal payload is sent successfully; the scheduler abort path merely marks the request as cancelled, and a failed `connector.put()` does not clean up `code_prompt_token_ids`, the counters, or pending payloads either. If a request is cancelled after at least one chunk has been sent, this per-request code list remains in the long-lived process. Please perform cleanup for both internal and external IDs on abort, retry exhaustion, and send failure, and add regression tests for those paths.

vllm_omni/model_executor/models/neutts_air/neutts_air_code2wav.py
```diff
@@ -0,0 +1,398 @@
...
+    ) -> None:
+        del hidden_states, sampling_metadata
+        return None
+
+    def _split_request_ids(
+        self,
+        ids: torch.Tensor,
+        seq_token_counts: list[int] | None = None,
+    ) -> list[torch.Tensor]:
+        if seq_token_counts is not None and len(seq_token_counts) > 1:
```
Sy0307, line 159: [P2] `seq_token_counts` is used only for multi-request batches, so a single request with padded input is decoded in full. The GenerationModelRunner can pass padded `input_ids` together with a one-element `seq_token_counts` containing the actual length; this function then returns the entire `ids` tensor, sending padding tokens to NeuCodec. The default YAML setting `enforce_eager=true` only avoids some cases under the current deployment and cannot be the model contract. Please apply the counts whenever `seq_token_counts` is present and add a single-request padding test.

vllm_omni/entrypoints/openai/tts_adapters/neutts_air.py
```diff
@@ -0,0 +1,103 @@
...
+        if isinstance(request.ref_audio, list):
+            return "NeuTTS-Air requires exactly one 'ref_audio'"
+
+        fmt_err = self.ctx.server._validate_ref_audio_format(request.ref_audio)
+        if fmt_err:
+            return fmt_err
+        if not request.ref_text or not request.ref_text.strip():
+            return "NeuTTS-Air requires 'ref_text' (transcript of the reference audio)"
+        if request.max_new_tokens is not None:
+            if request.max_new_tokens < self.max_new_tokens_min:
```
Sy0307, line 49: [P2] This validates `max_new_tokens` only against the generic upper bound and does not account for the actual NeuTTS-Air context budget. Stage 0 has `max_model_len=2048`, while the processor concatenates reference codec tokens, reference phonemes, and target phonemes into the prompt; this adapter nevertheless allows values up to 4096. A request with long reference audio and long text will first pay the CPU codec-encoding cost and only then fail in Stage 0 because the context is too large. Please reject requests against the actual prompt budget early, or adjust the deployment limit and enforce a matching constraint.

docs/user_guide/examples/online_serving/text_to_speech.md
```diff
@@ -304,6 +305,38 @@ The client supports `--api-base`, `--model`, `--text`, `--response-format`, `--l
...
+```bash
+curl http://localhost:8000/v1/audio/speech \
+  -H "Content-Type: application/json" \
+  -d '{
+    "model": "neuphonic/neutts-air",
+    "input": "Hello from NeuTTS-Air.",
+    "ref_audio": "file:///path/to/reference.wav",
+    "ref_text": "Transcript of the reference audio.",
+    "response_format": "wav",
+    "stream": true
```
Sy0307, line 333: [P2] This curl example will not save a WAV response. In this protocol, `stream=true` defaults to SSE; raw PCM/WAV byte streaming requires `stream_format: "audio"`. With the current payload, `--output neutts_air.wav` writes SSE text events into a file named `.wav`. Please add `"stream_format": "audio"`, or remove `stream` and use a non-streaming request.

vllm_omni/model_executor/models/neutts_air/neutts_air_code2wav.py
```diff
@@ -0,0 +1,398 @@
...
+from torch import nn
+from vllm.config import VllmConfig
+from vllm.forward_context import get_forward_context, is_forward_context_available
+from vllm.logger import init_logger
+
+from vllm_omni.model_executor.models.output_templates import OmniOutput
+
+logger = init_logger(__name__)
+
+DEFAULT_NEUCODEC_REPO = "neuphonic/neucodec"
```
Sy0307, line 22: [P3] NeuTTS-Air protocol constants are duplicated across multiple modules: `DEFAULT_NEUCODEC_REPO`, the speech-token offset, the end-token ID, and the chunk size are separately defined in the talker, adapter, pipeline, stage processor, and decoder. These values are one model contract, so keeping multiple copies will eventually create silent drift. Please move them into a single constants/config module and import them everywhere else.

Sy0307 (changes requested):
Please address the individual inline findings below before resubmitting. The current review conclusion is Request Changes.

vllm_omni/model_executor/models/neutts_air/neutts_air_code2wav.py
```diff
@@ -0,0 +1,398 @@
...
+
+        stride_samples = chunk_frames * NEUTTS_HOP_LENGTH
+        if state is None:
+            state = _StreamAudioState(stride_samples=stride_samples)
+            self._stream_states[request_id] = state
+        elif state.stride_samples != stride_samples:
+            raise RuntimeError("NeuTTS-Air streaming chunk size changed within one request.")
+
+        state.frames.append(cropped)
+        mixed = _linear_overlap_add(state.frames, state.stride_samples)
```
lishunyang12, line 286: Keep only the overlap tail instead of remixing the entire audio history every chunk.

Issue #6195, opened 2026-08-14
[RFC]: Decouple DLO host-weight storage from DP request scheduling


hsliuustc0106:
## Implementation status

- **Phase A merged:** #6213 landed the loader-owned `HostWeightPlan`, fail-closed direct checkpoint mmap, and bounded no-AllGather staging. The shared-mmap optimization is intentionally scoped to TP=1; TP > 1 remains a functional ordinary TP-aware loader fallback and does not receive the shared-mmap host-memory benefit.
- **Next:** Phase B remains open: define and implement the normalized runtime mmap cache for equivalent node-local runtime layouts, including one cache entry per TP coordinate where appropriate. This remains independent of DP request scheduling and orchestration.

### Motivation.

Distributed Layerwise Offload (DLO) currently combines three different concerns:

1. **Host-weight storage**: private loader tensors, DP-sharded pinned tensors, raw checkpoint mmap pages, or a normalized runtime mmap cache.
2. **Per-layer transfer**: complete-block H2D, or shard H2D followed by AllGather.
3. **Request orchestration**: whether replicas must enter synchronized request waves.

These concerns should be separable.

The default AllGather path has a clear memory model: every rank keeps a different `1 / group_size` pinned shard and reconstructs each block collectively. Across the group, the persistent private shards total approximately one model copy.

The current no-AllGather fallback has a different problem. Since each replica must transfer a complete block without a weight collective, the standard loader materializes a complete private host model in every process. With pure DP on one node, this scales host memory as:

```text
private host memory ~= data_parallel_size * model_size
```

However, replicas with the same runtime weight layout do not need separate persistent copies of immutable weights. They can share either the original safetensors pages or a node-local runtime-format cache through the OS page cache. no-AllGather requires every rank to stage and transfer a complete block, but it does not inherently require one private full-model host copy per rank.

This distinction also clarifies ownership:

- the **model loader** should own checkpoint discovery, checkpoint-to-runtime transformation, and host-weight-plan construction;
- **DLO** should consume prepared host weights and own block staging and transfer;
- a higher-level deployment layer such as vLLM Router or Dynamo should own admission, load balancing, and request orchestration across independent diffusion engines.

This RFC is a follow-up to:

- #5396, which introduced DLO with H2D + AllGather overlap;
- #5494, which discusses the broader unified-DP architecture;
- #5911, which explores independent rank-local request dispatch; and
- #6198, which prototypes rank-local checkpoint mmap and bounded staging.

[...]

PR #6198, opened 2026-08-14
[Diffusion] Add rank-local mmap storage for DLO

hsliuustc0106:
## Summary

Refs #6195.

This PR makes node-local mmap a DLO host-weight storage option independent of the AllGather transfer protocol.

- retain compatible checkpoint tensors as immutable file-backed sources for rank-local DLO
- stream complete blocks through two bounded pinned host staging slots per worker
- preserve the existing DP-sharded AllGather path and standard-loader fallbacks
- add MiniMax-H3 TP1 mmap support with grouped-QKV layout adaptation
- document the storage/transfer/orchestration boundaries and add a DP2 lifecycle mode
- make no scheduler, executor, worker-dispatch, or request-routing changes

## Motivation

DLO currently has two transfer modes with different host-memory behavior:

| Mode | Persistent host storage | Per-block transfer |
| --- | --- | --- |
| AllGather | a private `1 / group_size` shard per rank | shard H2D + AllGather |
| no-AllGather, before this PR | a private complete model per rank | complete-block H2D |
| no-AllGather, this PR | node-shared file-backed pages + bounded private staging | complete-block H2D |

no-AllGather requires every rank to transfer a complete block, but it does not require every rank to retain a persistent anonymous copy of the same immutable checkpoint. Processes on one node can fault the same clean safetensors pages through the OS page cache.

This is deliberately a storage/lifecycle change. Cross-replica admission and load balancing remain outside DLO and can be owned by vLLM Router, Dynamo, or another deployment layer.

### Shared selection policy

The diffusion loader and DLO backend call the same `should_select_dlo_mmap()` policy. The loader skips ordinary weight materialization only when the backend will enter the mmap path.

Rank-local mmap initially requires:

- an explicitly compatible model;
- TP=1;
- HSDP disabled; and
- no online quantization.

Other no-AllGather layouts retain the regular loader.

### Exact component source mapping

The mmap loader indexes each declared component source, including safetensors index files, prefixes keys into the pipeline namespace, and applies an optional checkpoint-key remap. This avoids recursively conflating unrelated checkpoint partitions.

### Retained source and bounded staging

Each rank-local hook retains detached checkpoint views and replaces runtime parameters with offload placeholders. All streamed and resident DiT layers in one worker share two host staging slots sized to the largest block.

A prefetch waits for prior slot use, packs one block, applies any declared layout adapter, copies the complete block to the rotating device slot, and records an event protecting host-slot reuse. The safetensors mappings remain open until DLO is disabled.

CPU packing is synchronous in this first implementation. The pinned H2D transfer can overlap with current-block compute.

### MiniMax-H3

MiniMax-H3 explicitly opts into TP1 mmap loading. Its checkpoint interleaves Q/K/V rows by query group, while the fused runtime linear expects contiguous Q, K, then V rows. A per-parameter adapter applies the same reorder as the regular loader while packing one bounded block; the file-backed source remains unchanged.

Quantized MiniMax-H3 continues to use the regular loader.

## Usage

```bash
vllm serve /path/to/model --omni \
  --data-parallel-size 2 \
  --dlo-no-use-allgather
```

For an explicitly compatible TP1 model, this selects rank-local mmap automatically. Unsupported models and parallel layouts retain existing behavior.

## Scope and limitations

- Multi-rank AllGather behavior is unchanged.
- With an effective DLO group size of one, AllGather configuration now correctly uses rank-local mmap staging, closing a pre-existing loader/backend mismatch that could leave parameters on `meta`.
- no-AllGather still performs a complete-block H2D copy per rank.
- The sharing boundary is one OS node, not the cluster.
- The first scope is the streamed DiT and planned DiT submodules; ancillary encoders/VAEs and fallback paths may retain private host weights.
- TP, HSDP, and online quantization keep their normal loader contracts.
- This PR changes no request scheduling or DP orchestration code.

### CPU and static checks

```text
234 passed
ruff check: passed
ruff format --check: passed
git diff --check: passed
```

The suite covers the DLO backend, diffusion loader, MiniMax-H3 contracts and quantization behavior, diffusion-engine regressions, and entrypoint configuration.

### MiniMax-H3 DP2 no-AllGather E2E

Exact final patch, 2x NVIDIA L20X, MiniMax-H3 FL2VA:

```text
data_parallel_size=2
tensor_parallel_size=1
dlo_use_allgather=false
535/535 checkpoint tensors mapped
52 streamed blocks across 2 hook groups
2 x 1.23 GiB pinned host staging slots per worker
initial GPU model memory: 3.25 GiB per worker
frames: [107, 256, 256, 3]
audio: [1, 2, 142400]
peak GPU memory: 13,226 MiB
request wave: 41.57 s
active_children after shutdown: []
```

A page-sharing inspection of the same path found 13 transformer mappings covering approximately 63.2 GiB. During denoising, each process reported roughly 50–51 GiB `Shared_Clean`, approximately half-sized PSS, and zero `Private_Dirty` for those mappings, confirming shared clean checkpoint pages rather than per-process anonymous full-model copies.

### DIFF
--- a/vllm_omni/diffusion/model_loader/diffusers_loader.py
+++ b/vllm_omni/diffusion/model_loader/diffusers_loader.py
@@ -363,22 +363,39 @@ class DiffusersPipelineLoader:
             else:
                 model = self._init_from_load_format(load_format, target_device, custom_pipeline_name, is_hsdp=False)
 
-                # Skip load_weights only for DLO+AllGather when the model
-                # supports mmap loading and no online quantization is active.
-                # This condition MUST match the gate in
-                # DistributedLayerwiseOffloadBackend.enable() so that the
-                # loader skips ⟺ enable() uses mmap.
-                from vllm_omni.diffusion.offloader.offload_plan import supports_mmap_loading
+                # Skip load_weights when DLO can keep checkpoint tensors as
+                # mmap-backed host masters.  Storage selection is independent
+                # of whether DLO reconstructs weights with AllGather.  The
+                # no-AllGather path can use mmap only when the checkpoint is
+                # already the rank-local runtime layout (currently TP=1 and no
+                # HSDP); TP/HSDP still require the regular loader callbacks.
+                # The loader and backend share should_select_dlo_mmap() so
+                # skipping load_weights always selects mmap in enable().
+                from vllm_omni.diffusion.offloader.offload_plan import (
+                    should_select_dlo_mmap,
+                    supports_mmap_loading,
+                )
 
                 _dist_offload = getattr(self.od_config, "enable_distributed_layerwise_offload", False)
                 _use_ag = getattr(self.od_config, "dlo_use_allgather", True)
                 _has_online_quant = self._has_online_quant(model)
                 _supports_mmap = supports_mmap_loading(model)
+                _tp_size = int(getattr(self.parallel_config, "tensor_parallel_size", 1))
+                _use_hsdp = bool(getattr(self.parallel_config, "use_hsdp", False))
 
-                _skip_load = _dist_offload and _use_ag and _supports_mmap and not _has_online_quant
+                _skip_load = _dist_offload and should_select_dlo_mmap(
+                    mmap_supported=_supports_mmap,
+                    dlo_use_allgather=_use_ag,
+                    has_online_quant=_has_online_quant,
+                    tensor_parallel_size=_tp_size,
+                    use_hsdp=_use_hsdp,
+                )
 
                 if _skip_load:
-                    logger.info("DLO+AllGather active: skipping load_weights (will load via mmap in enable())")
+                    logger.info(
+                        "DLO mmap storage active (%s): skipping load_weights (will map checkpoint weights in enable())",
+                        "AllGather" if _use_ag else "rank-local",
+                    )
                 else:
                     if _dist_offload and _use_ag and _has_online_quant:
                         raise ValueError(
--- a/vllm_omni/diffusion/model_loader/diffusers_loader.py
+++ b/vllm_omni/diffusion/model_loader/diffusers_loader.py
@@ -458,11 +458,11 @@ class DiffusersPipelineLoader:
         """Whether any layer uses an online-quant method that defers weight
         materialization onto the ``meta`` device (upstream vLLM
         ``uses_meta_device=True``, e.g. online FP8)."""
-        for module in model.modules():
-            quant_method = getattr(module, "quant_method", None)
-            if getattr(quant_method, "uses_meta_device", False):
-                return True
-        return False
+        from vllm_omni.diffusion.offloader.offload_plan import (
+            has_online_quantization,
+        )
+
+        return has_online_quantization(model)
 
     def _apply_skip_softmax_calibration(self, model: nn.Module) -> None:
         from vllm_omni.diffusion.attention.backends.trtllm_calibration import (
--- a/vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
+++ b/vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
@@ -532,13 +532,10 @@ class MiniMaxH3Pipeline(
         encoder_block_attrs={"text_encoder": ("vision.blocks", "text_model.layers")},
         on_demand_component_paths=frozenset({"text_encoder", "video_vae", "audio_vae"}),
     )
-    # H3's regular loader performs checkpoint-layout conversions (grouped QKV
-    # and fused MLP). Keep it on that path until mmap can run the same loader
-    # callbacks for every affected parameter.
-    _supports_mmap_loading: ClassVar[bool] = False
-    # TODO(offload): Re-enable after the generic rank-local mmap path can run
-    # this model's grouped-QKV reorder and fused-MLP packing before TP sharding.
-    # Do not bypass the regular loader until that equivalence is tested.
+    # TP1 DLO mmap applies the grouped-QKV adapter while packing each block.
+    # The fused MLP checkpoint is already gate-then-up, matching its TP1
+    # runtime layout. TP>1 and quantized layouts stay on the regular loader.
+    _supports_mmap_loading: ClassVar[bool] = True
     _PROFILER_TARGETS: ClassVar[list[str]] = [
         "_prepare_reference_videos",
         "encode_prompt",
@@ -648,6 +645,10 @@ class MiniMaxH3Pipeline(
             od_config.quantization_config,
             "transformer",
         )
+        if transformer_quant_config is not None:
+            # Quantized parameters may have packed weights and scale tensors
+            # whose runtime layout is owned by quant-method loader callbacks.
+            self._supports_mmap_loading = False
         self.transformer = MiniMaxH3DiTModel(
             od_config,
             quant_config=transformer_quant_config,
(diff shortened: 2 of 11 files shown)


vllm_omni/diffusion/offloader/offload_plan.py
```diff
@@ -60,15 +60,46 @@ def get_offload_plan(pipeline: nn.Module) -> OffloadPlan | None:
...
+    tensor_parallel_size: int,
+    use_hsdp: bool,
+) -> bool:
+    """Whether loader and backend should enter DLO's mmap path.
+
+    AllGather preserves its existing fail-closed topology validation inside
+    the mmap loader. The extra TP/HSDP gate applies to rank-local mmap only.
+    """
+    rank_local_compatible = tensor_parallel_size == 1 and not use_hsdp
+    return mmap_supported and not has_online_quant and (dlo_use_allgather or rank_local_compatible)
```
hsliuustc0106, line 110: With `dlo_use_allgather=True` and `dp_size=1` this now selects mmap — the old gate in `enable()` required `dp_size > 1`, so that config skipped `load_weights` in the loader but never ran `_load_weights_via_mmap()` and was left on meta params. Nice that this closes that hole, but since the PR body says AllGather behavior is unchanged, maybe add a sentence there acknowledging the `dp_size=1` corner.

vllm_omni/diffusion/model_loader/diffusers_loader.py
```diff
@@ -363,22 +363,39 @@ def load_model(
...
+                # The loader and backend share should_select_dlo_mmap() so
+                # skipping load_weights always selects mmap in enable().
+                from vllm_omni.diffusion.offloader.offload_plan import (
+                    should_select_dlo_mmap,
+                    supports_mmap_loading,
+                )
 
                 _dist_offload = getattr(self.od_config, "enable_distributed_layerwise_offload", False)
                 _use_ag = getattr(self.od_config, "dlo_use_allgather", True)
                 _has_online_quant = self._has_online_quant(model)
```
hsliuustc0106, line 381: `enable()` computes `has_online_quant` by scanning `quant_method.uses_meta_device` over `pipeline.modules()`, while the loader uses `self._has_online_quant(model)`. If the two ever disagree, `load_weights` is skipped but mmap is not selected, leaving the model on meta params. Pre-existing for AllGather — since both paths now hinge on this, could the detection live in one shared helper?

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
```diff
@@ -952,27 +1200,33 @@ def _load_weights_via_mmap(self, pipeline: nn.Module, modules) -> None:
...
             # Filter out params inside submodules that will be loaded later
             # by _load_module_weights_from_mmap (non-block submodules)
             dit_param_names = set()
-            for dit_module in modules.dits:
-                for pname, _ in dit_module.named_parameters():
-                    dit_param_names.add(pname)
+            for dit_name, dit_module in zip(modules.dit_names, modules.dits):
+                for pname, _ in chain(dit_module.named_parameters(), dit_module.named_buffers()):
+                    dit_param_names.add(f"{dit_name}.{pname}")
             truly_missing = [n for n in remaining_meta if n in dit_param_names]
```
hsliuustc0106, line 1231: This only warns, so an opt-in model whose checkpoint keys do not match runtime names (no `_remap_ckpt_key` to catch it) fails later and less clearly — via the CPU-device check in `_collect_mmap_sources` under rank-local, or cryptically under AllGather. Any reason not to raise when `truly_missing` is non-empty for explicitly opted-in models?

vllm_omni/diffusion/offloader/base.py
```diff
@@ -32,6 +32,8 @@ class OffloadConfig:
     dlo_use_allgather: bool = True
     dlo_resident_layers: int = 0  # leading DiT layers kept on device
     model_path: str | None = None  # checkpoint path for mmap weight loading
+    # Appended to preserve the positional order of the pre-existing fields.
```
hsliuustc0106: why we need to add tp size here ? is seems not related

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
```diff
@@ -532,13 +532,10 @@ class MiniMaxH3Pipeline(
         encoder_block_attrs={"text_encoder": ("vision.blocks", "text_model.layers")},
```
hsliuustc0106, line 532: for other models, what do we change to adapt to this feature?

vllm_omni/diffusion/offloader/offload_plan.py
hsliuustc0106: Thanks — clarified in both the PR body and the design doc. More precisely, the corner is an effective DLO group size of one, not simply `DP=1`: with `DP=1, SP>1`, DLO can still AllGather over the SP group. Multi-rank AllGather behavior is unchanged; an effective group size of one now uses rank-local mmap staging and closes the previous loader/backend mismatch that could leave parameters on `meta`.

vllm_omni/diffusion/model_loader/diffusers_loader.py
hsliuustc0106: Done in `76e257b0`. I added shared `has_online_quantization()` policy in `offload_plan.py`; both `DiffusersPipelineLoader._has_online_quant()` and `DistributedLayerwiseOffloadBackend.enable()` now delegate to it. A regression test covers both the false and `uses_meta_device=True` cases.

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
hsliuustc0106: Agreed. In `76e257b0`, explicitly opted-in pipelines (`_supports_mmap_loading=True`) now raise a clear `RuntimeError` immediately when expected DiT tensors remain on `meta`, including the count and first missing names. The warning is retained only for legacy remap-based mmap models, and a regression test covers an incomplete explicit component source.

vllm_omni/diffusion/offloader/base.py
hsliuustc0106: The TP size is needed for host-storage selection, not for creating the DLO AllGather group. Rank-local mmap bypasses TP-aware `weight_loader` callbacks, so no-AllGather with `TP>1` must retain the standard loader. The loader reads `parallel_config` directly, while the backend receives `OffloadConfig`; carrying the configured TP size lets both call the same `should_select_dlo_mmap()` policy. I added this rationale next to the field in `1c3d5979`.

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
hsliuustc0106: I added an explicit “Adding mmap support to another model” contract to the DLO design doc in `1c3d5979`. In short: provide exact safetensors `weights_sources` prefixes; opt in with `_supports_mmap_loading=True` when names match or provide `_remap_ckpt_key` when they do not; attach `mmap_weight_transform` for TP1 layout conversions; and test mapping completeness, equivalence with the regular loader, and TP/HSDP/online-quant fallbacks. The opt-in flag alone now fails fast if any expected DiT tensor is unmatched.

vllm_omni/diffusion/offloader/base.py
hsliuustc0106: why we only care about TP?

hsliuustc0106:
close this PR since we have updated the design in #6195

Issue #5227, opened 2026-07-20
[RFC][Refactor]: Entrypoints & api_server Refactor Roadmap


herotai214:
[Updated on 27 July]
## Purpose

The main goal is readability, maintainability, and scalability of the API server layer.

Today, `vllm_omni/entrypoints/openai/api_server.py` is not only an app assembly file. It also owns OpenAI route bodies, Omni-specific route bodies, WebSocket entrypoints, request parsing helpers, media helpers, background video job handling, error conversion, app-state wiring, and model-specific endpoint logic.

That makes small endpoint changes risky. A reviewer has to inspect whether a change intended for images also affects chat, video, speech, realtime, duplex, OpenPI, `/v1/models`, or server management routes.

This refactor aims to:

- align with upstream vLLM's current router/factory style where practical;
- make `api_server.py` mostly app construction, lifespan, middleware, app-state bootstrap, and router registration;
- move endpoint behavior into endpoint-family packages;
- keep shared helpers in neutral homes only when more than one endpoint family uses them;
- reduce cross-family coupling such as images behavior living inside chat serving code;
- keep OpenAI-compatible HTTP/WebSocket surfaces under `entrypoints/openai`;
- remove temporary migration shims after endpoint packages are stable.

## Naming Convention

Use `Refactor [P<phase>.<step>]: <short title>`.

## Roadmap

Line counts are rough estimates only. They communicate review size, not a commitment.

### Phase 0: API Server Foundation

Phase 0 is intentionally split into small PRs so the first refactor does not mix tests, helper movement, and route migration.

| Refactor Step | PR # | Focus | Why |
| --- | --- | --- | --- |
| P0.1 | ✅#6202 | API server guardrails | Add focused tests for route registration, missing-handler behavior, error envelopes, and app-state availability before moving code. |
| P0.2 | ⏳#5453 | API server helper migration | Move non-route helpers and request models out of `api_server.py` so route bodies no longer export helper implementation. |
| P0.3 | TBD | API server route-owner extraction | Move decorated route bodies from `api_server.py` into endpoint-owned routers/factories. |

P0.2 may temporarily make `api_server.py` import helpers from their new owners. That is acceptable because P0.3 immediately follows and moves the route bodies beside those helpers.

### Phase 1: Endpoint Families

| Refactor Step | PR # | Focus | Notes |
| --- | --- | --- | --- |
[...]

PR #4755, opened 2026-06-26
[Metrics] Add Diffusion Metrics

vraiti:
PLEASE FILL IN THE PR DESCRIPTION HERE.

## Purpose
Add Diffusion Engine metrics:
* `vllm_omni:diffusion_exec_s`: E2E time of core diffusion steps
* `vllm_omni:diffusion_preprocess_s:` E2E time of diffusion preprocess function
* `vllm_omni:diffusion_postprocess_s:` E2E time of diffusion postprocess function

### E2E Tests
On an EC2 `g7e.2xlarge`, run 

```
vllm serve --omni --disable-log-stats Qwen/Qwen-Image-2512 --port 8000
```

Then,

```
vllm serve --omni --log-stats Qwen/Qwen-Image-2512 --port 8000
```

Run  `POST /v1/images/generation` 5 times and record E2E latency of each request.

Observe:
* `vllm_omni:diffusion_exec_time_s` correctly captures E2E latency up to HTTP serialization/server overhead
* No performance regression up to noise between `--disable-log-stats` and `--log-stats`
* Correct response to HTTP `GET /metrics` for both deployments

### Unit Tests
- test_diffusion_exec_observed — Observes 1.25s on exec histogram, asserts _sum equals 1.25.
- test_diffusion_preprocess_observed — Observes 0.03s on preprocess histogram, asserts _sum matches with pytest.approx.
- test_diffusion_postprocess_observed — Observes 0.8s on postprocess histogram, asserts _sum equals 0.8.
- test_log_stats_false_skips — Creates OmniModalityMetrics(log_stats=False), calls all three diffusion observe methods, asserts nothing is
  written to the registry.
- test_diffusion_path_full — Passes a complete diffusion_metrics dict with output_type="image", asserts all three observe methods are
  called with correct values.
- test_diffusion_metrics_none_skipped — Passes diffusion_metrics=None, asserts no diffusion observe methods are called.
- test_diffusion_zero_values_observed — Passes 0.0 for all three timing keys, asserts all three are still observed.
- test_non_audio_output_type_skips_audio_but_still_checks_diffusion — Calls finalize with output types "text", "image", "video", asserts
  no audio methods are called.

**vLLM Version:**
v0.23.0

**vLLM-Omni Commit:**
d6f1d3d61ca462ff8e89312ebf0fdb675edad104

### E2E Tests
`--log-stats` `GET /metrics`:
[metrics-snapshot.txt](https://github.com/user-attachments/files/29395010/metrics-snapshot.txt)

[...]

PR #6202, opened 2026-08-14
[Tests][Refactor] P0.1: Add API server surface guardrails

herotai214:

# [Tests][Refactor] P0.1: Add API server surface guardrails

This is `Refactor [P0.1]` for #5227, ahead of #5453 (P0.2 helper migration).
This PR only add 1 single test file WITHOUT changing any other files.

## Purpose

Land **behavior-preserving surface guards** for `vllm_omni/entrypoints/openai/api_server.py` **before** helpers/routes move.

Today `api_server.py` mixes app assembly, route bodies, helpers, and bootstrap. P0.2/P0.3 (and later family PRs) will relocate code; without a CI fence, regressions are easy to miss (dropped route, changed unavailable JSON, missing `app.state` key, broken upstream override).

This PR adds CPU-only tests that freeze the **observable** contract:

- route census (Omni router + profiler)
- worker assembly / upstream override
- unavailable / missing-handler envelopes (HTTP + WebSocket)
- shared engine-error JSON fields
- pure-diffusion vs multi-stage `app.state` key presence
- light OpenAPI path/method checks (**no** checked-in golden JSON)

**No production code moves** in this PR — tests only.

## What this unlocks

| Next | Benefit |
|---|---|
| **P0.2** (#5453) | Helper migration can prove “surface unchanged” in CI |
| **P0.3** | Route extraction keeps the same census / envelopes / assembly behavior |
| **Later family PRs** | Shared floor against cross-family route/app-state accidents |

## Out of scope

- Helper body AST identity (P0.2)
- Route-owner extraction (P0.3)
- Deep per-endpoint success paths (existing family UTs)
- GPU / live speech byte-identity checks

## Test Plan

```bash
pytest -q tests/entrypoints/openai_api/test_api_server_guards.py
```

Markers: `core_model` + `cpu` (no GPU, no weights).

## vLLM version / vLLM-Omni commit

| Item | Value |
|---|---|
| **vLLM** | `0.27.0` |
| **vLLM-Omni base** | `origin/main` @ `c98517ce` (`[Metrics] Add Diffusion Metrics (#4755)`) Aug 13, 2026 |
| **Change** | `tests/entrypoints/openai_api/test_api_server_guards.py` |

## Test result

**Result: 19 passed**

Coverage map (14 functions; WS case ×6 → 19 items):

1. Router + profiler route census  
2. OpenAPI covers HTTP census paths/methods  
3. Assembly: upstream override, Omni/profiler mount, storage, engine exception handlers  
4. Timestamp middleware: HTTP stamped, WebSocket untouched  
5–10. WS unavailable frames (speech / video-chat / realtime-video / realtime / openpi / duplex)  
11. `/health` no engine → 503 envelope  
12. `/v1/models` no handler → empty list  
13–14. Images missing engine / missing chat handler → stable 503 details  
15. Speech missing handler → 404 detail  
16–17. Engine generate/dead error JSON keeps `request_id` / stage fields  
18–19. Pure-diffusion + multi-stage `app.state` key snapshots  

---

cc @linyueqian  


herotai214:
@vllm-omni-review-bot

tests/entrypoints/openai_api/test_api_server_guards.py
```diff
@@ -0,0 +1,733 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+"""API-server surface guards for the OpenAI Omni entrypoint.
+
+Purpose
```
herotai214: test

linyueqian:
@vllm-omni-review-bot

tests/entrypoints/openai_api/test_api_server_guards.py
```diff
@@ -0,0 +1,733 @@
...
+        if isinstance(route, APIRoute):
+            for method in sorted((route.methods or set()) & _HTTP_METHODS):
+                manifest.add((method, route.path))
+        elif isinstance(route, APIWebSocketRoute):
+            manifest.add(("WEBSOCKET", route.path))
+    return manifest
+
+
+def _route_count(routes, method: str, path: str) -> int:
+    return sum(1 for item in _route_manifest(routes) if item == (method, path))
```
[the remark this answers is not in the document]
herotai214: Fixed; Now it first use a list instead of a set to check if any duplicates; Then check Omni ownership using lines like `is api_server.create_chat_completion`

tests/entrypoints/openai_api/test_api_server_guards.py
```diff
@@ -0,0 +1,763 @@
...
+    monkeypatch.setattr(api_server, "create_streaming_video_handler", lambda **_k: _marker("streaming_video"))
+    monkeypatch.setattr(api_server, "OpenAIServingRealtime", _FakeCtor)
+    monkeypatch.setattr(api_server, "OmniOpenAIServingVideo", _FakeCtor)
+    monkeypatch.setattr(api_server, "should_enable_duplex_endpoint", lambda *_a, **_k: False)
+
+    state = State()
+    await api_server.omni_init_app_state(engine, state, _minimal_args())
+
+    present = {key for key in _MULTISTAGE_APP_STATE_KEYS if hasattr(state, key)}
+    assert present == _MULTISTAGE_APP_STATE_KEYS
```
linyueqian: [important] This misses the failure mode the module docstring leads with.

`present` is built by filtering `_MULTISTAGE_APP_STATE_KEYS`, so it is a subset of that set by construction and this asserts only that every expected key exists. In this codebase "not wired" is written as an explicit `None` (`api_server.py:788`, `:801`, `:827`, `:1157`, `:1180`), and `hasattr` is true for those, so the "endpoint registered but handler never wired" case in lines 31 and 32 is exactly what slips through.

Confirmed by mutation. Both of these leave all 19 tests green:

```
state.openai_serving_video = None      -> 19 passed
state.serving_tokenization = None      -> 19 passed
```

Both would start returning 404 or a changed envelope in production.

Suggest pinning values, not just names, for the keys that must be live. Something like a `_MUST_BE_WIRED` subset asserted with `getattr(state, key) is not None`, keeping the existing presence check for the rest. The four keys that already pin a concrete value (lines 690 to 693, 762, 763) are the right idea, just applied to only four keys today.

Same applies to line 689 for the diffusion case.

tests/entrypoints/openai_api/test_api_server_guards.py
```diff
@@ -0,0 +1,763 @@
...
+    monkeypatch.setattr(api_server, "build_openai_app", fake_build_openai_app)
+    monkeypatch.setattr(api_server, "build_async_omni", fake_build_async_omni)
+    monkeypatch.setattr(api_server, "omni_init_app_state", fake_omni_init_app_state)
+    monkeypatch.setattr(
+        api_server,
+        "shutdown_unsupported_routes",
+        lambda app, restrictions: captured.setdefault("restrictions", restrictions),
+    )
+    monkeypatch.setattr(api_server, "STORAGE_MANAGER", SimpleNamespace(start=fake_storage_start))
+    monkeypatch.setattr(api_server, "_get_vllm_config", fake_get_vllm_config)
```
linyueqian: [important] The guard fails on exactly the refactor it exists to permit.

`monkeypatch.setattr` raises `AttributeError` when the target attribute is absent, so relocating any of these private helpers out of `api_server.py`, which is what P0.2 is for, turns the fence red with no observable surface change. Measured, each mutation applied alone and reverted:

```
_get_vllm_config relocated                      -> 4 failed, 15 passed
_create_engine_error_json_response relocated    -> 2 failed, 17 passed
_register_omni_exception_handlers relocated     -> 1 failed, 18 passed
```

The affected references are `_get_vllm_config` here and at lines 451, 667 and 730, `_should_enable_profiler_endpoints` at 450, `_create_engine_error_json_response` at 614, `_register_omni_exception_handlers` at 631, and `served_wrapper._inner` at 381 and 382.

The practical risk is not the red build, it is that the P0.2 author edits this file to get green, and a fence that gets edited under pressure stops being a fence. Where a public path exists, anchoring there instead would keep the guard honest across the move: exercise the error envelope through the assembled app's registered handler rather than calling `_create_engine_error_json_response` directly, and let the profiler and vllm-config fakes hang off the assembly seam rather than the private names.

If you would rather keep these as is, that is defensible, but then it is worth a comment saying the file is expected to be updated in lockstep with P0.2 so the next reader knows the red is a rename and not a regression.

tests/entrypoints/openai_api/test_api_server_guards.py
```diff
@@ -0,0 +1,763 @@
...
+    "openai_serving_video",
+    "openai_streaming_video_output",
+    "openai_serving_duplex",
+    "openai_streaming_speech",
+    "openai_streaming_video",
+    "openai_serving_realtime_robot",
+    "enable_server_load_tracking",
+    "server_load_metrics",
+}
+_MULTISTAGE_APP_STATE_KEYS = {
```
linyueqian, line 133: [suggestion] `serving_tokenization` is missing from the multi-stage set, though it is in the diffusion set at line 108.

It is assigned by multi-stage init at `api_server.py:1062` and read as the fallback base server by five multi-stage handlers: `api_server.py:1222`, `:1297`, `:1352`, `:1390`, `:1433`. Dropping it in a later refactor changes the fallback envelope and nothing here notices.

Codex also flagged `openai_serving_responses`, the completion/pooling/embedding/classification/score handlers, transcription/translation, Anthropic messages and tokens as unguarded. Those feel more like a judgement call about which keys route owners actually depend on, so I would not insist, but `serving_tokenization` has concrete readers.

tests/entrypoints/openai_api/test_api_server_guards.py
```diff
@@ -0,0 +1,763 @@
...
+    http_scope = {"type": "http", "state": {}}
+    ws_scope = {"type": "websocket"}
+
+    async def _receive():
+        return {"type": "http.disconnect"}
+
+    async def _send(_message):
+        return None
+
+    async def inner(scope, receive, send):
```
linyueqian, line 503: [suggestion] This stand-in records nothing, so the test still passes if the middleware stops calling its inner application at all. A refactor that stamps the timestamp correctly and then drops every request would stay green here.

Setting a flag inside `inner` and asserting it ran for both the HTTP and WebSocket scopes would close that off in two lines.

To be clear about what is fine here: replacing `_inner` does not weaken the timestamp assertions themselves, since the stamping happens before the inner call, and the line 475 WebSocket assertion is meaningful rather than vacuous because `ws_scope` starts with no `state` key at all.

tests/entrypoints/openai_api/test_api_server_guards.py
```diff
@@ -0,0 +1,763 @@
...
+    """Lock the Omni + profiler router census.
+
+    Fails if a route is added/removed/renamed, a method changes, or a
+    WebSocket path disappears during helper/router moves. Also fails if the
+    same public route is registered more than once, which a set manifest alone
+    would collapse.
+    """
+    _assert_unique_route_entries(api_server.router.routes)
+    _assert_unique_route_entries(api_server.profiler_router.routes)
+    assert _route_manifest(api_server.router.routes) == _EXPECTED_ROUTER_ROUTES
```
linyueqian, line 327: [question] This freezes the census on `api_server.router` specifically, rather than on the assembled application.

If P0.3 extracts routes into owner-specific routers and mounts them straight onto the app, the assembled surface can be byte-identical while this assertion fails. If instead the sub-routers are aggregated back with `router.include_router(...)`, this keeps working unchanged.

Not a defect today, and the assembled-app census at line 405 already covers the surface that actually matters to clients. Mostly I want to know which shape P0.3 is heading for, since it decides whether this line survives contact with it.

tests/entrypoints/openai_api/test_api_server_guards.py
```diff
@@ -0,0 +1,763 @@
...
+from starlette.datastructures import State
+from starlette.requests import Request
+from starlette.websockets import WebSocketDisconnect
+from vllm.v1.engine.exceptions import EngineDeadError, EngineGenerateError
+
+from vllm_omni.entrypoints.openai import api_server
+
+pytestmark = [pytest.mark.core_model, pytest.mark.cpu]
+
+_HTTP_METHODS = {"GET", "POST", "DELETE", "PUT", "PATCH"}
```
linyueqian, line 64: [suggestion] `HEAD` and `OPTIONS` are not in this set, so adding `@router.head("/health")`, or adding `HEAD` to an existing route's methods, changes the observable method surface while the manifest, uniqueness, assembly and OpenAPI assertions all stay green.

tests/entrypoints/openai_api/test_api_server_guards.py
```diff
@@ -0,0 +1,763 @@
...
+
+    exc = EngineGenerateError("boom")
+    exc.error_stage_id = 2  # type: ignore[attr-defined]
+    response = api_server._create_engine_error_json_response(req, exc)
+    payload = json.loads(response.body)
+
+    assert "error" in payload
+    assert payload["error"]["request_id"] == "req-123"
+    assert payload["error"]["error_stage_id"] == 2
+    assert isinstance(response.status_code, int)
```
linyueqian: [suggestion] `JSONResponse.status_code` is always an `int`, so this assertion cannot fail. Line 621 already carries the real check.

linyueqian (approved):
Nice piece of work. I reviewed this as a fence rather than as ordinary tests, so instead of only checking that it passes I mutation-tested it: I injected each regression the module docstring claims to catch and checked whether the guard actually fired.

Environment: head `2dfdfcd1f`, vllm 0.27.0, CPU only.

**It passes and it does not disturb anything.** 19/19 in 2.6 s. Running the exact ready command, `pytest -sv tests/entrypoints tests/engine -m 'core_model and cpu'`, gives 1923 passed with this file against 1904 without it, so exactly the 19 new tests and no interference with neighbours (+18 s wall). Markers are right, and `.buildkite/cuda/test-ready.yml:23` already covers `tests/entrypoints` with `core_model and cpu`, so no CI wiring change is needed.

**The fence fires on real regressions, 7 out of 7.** Each row is a separate injected change to `api_server.py`, suite re-run, then reverted:

| injected regression | guard fired |
|---|---|
| route path renamed (`DELETE /v1/videos/{video_id}`) | yes, 3 tests |
| route method changed (`GET /health` to POST) | yes, 3 tests |
| WebSocket unavailable frame wording changed (duplex) | yes |
| `/health` unhealthy reason string changed | yes |
| upstream `/v1/models` override dropped | yes |
| `app.state` key removed (`openai_serving_video`) | yes |
| speech 404 detail changed | yes |
| new route added without census update | yes (intended) |

No false positives on the unmutated baseline. That is a genuinely effective guard, and worth landing.

Two things I would fix, both small, and the reasons are inline. Short version:

1. The failure mode the docstring leads with, "endpoint registered but handler never wired", is the one case the state guards miss, because in this file not-wired is written as `state.X = None` and `hasattr` is still true. I confirmed it: wiring `openai_serving_video` to `None` and wiring `serving_tokenization` to `None` both leave all 19 tests green.
2. The guard currently fails on exactly the moves P0.2 exists to make. Relocating `_get_vllm_config` breaks 4 tests, `_create_engine_error_json_response` breaks 2, and `_register_omni_exception_handlers` breaks 1, all with identical behaviour, because `monkeypatch.setattr` raises when the attribute is gone. My worry is less the red build than what it teaches: the next author edits the fence to get green, and a fence that gets edited under pressure stops being a fence.

Neither is a regression against main, so I am not blocking on them. Approving so this can go into the release, and I am happy to re-review quickly if you fix them here.

For transparency on method: I also ran this past two external models. They independently found the `None`-wiring gap and the private-symbol coupling, and they corrected one of my own readings (the `ws_scope` assertion is not vacuous, and Starlette keeps state in `State._state` rather than `vars(state)`).

tests/entrypoints/openai_api/test_api_server_guards.py
herotai214: Agreed. In this file “not wired” is an explicit None, so hasattr stayed true. 

Fixed. Both snapshots now use _MUST_BE_WIRED (is not None) plus _MUST_BE_NONE for the intentional Nones: diffusion vllm_config / serving_tokenization / duplex / streaming speech+video; multi-stage duplex (this fixture disables it) and openai_serving_realtime_robot. state.openai_serving_video = None or state.serving_tokenization = None should now fail.

tests/entrypoints/openai_api/test_api_server_guards.py
herotai214: Agreed. 

Generate-error envelope now goes through app.exception_handlers[EngineGenerateError] instead of calling _create_engine_error_json_response directly. 
Assembly/init vllm-config hangs off the fake engine’s get_vllm_config, so relocating _get_vllm_config is no longer a false red. 

The two envelope tests still call _register_omni_exception_handlers; commented TODO that a red there in P0.2 is a rename — retarget the call, don’t rewrite the fence.

tests/entrypoints/openai_api/test_api_server_guards.py
herotai214: Added `serving_tokenization` to the multi-stage key set and _MUST_BE_WIRED. 

Left the Codex extras (responses / pooling / embedding / Anthropic / …) out; those don’t have the same Omni route-owner readers. Happy to add a specific one if a later family PR actually depends on it.

tests/entrypoints/openai_api/test_api_server_guards.py
herotai214: Fixed. inner now sets a called flag; we assert it ran for both the HTTP and WebSocket scopes.

tests/entrypoints/openai_api/test_api_server_guards.py
herotai214: P0.3 will extract family routers and app.include_router(...). 

The assembled-app census is the client-facing lock that must survive. This api_server.router check is current-main structure; 

Commented TODO and we’ll retarget it in P0.3.

tests/entrypoints/openai_api/test_api_server_guards.py
herotai214: Correct that an explicit @router.head / @router.options would stay green today. 

We’re not adding HEAD to the set now: Starlette already auto-adds HEAD on GET routes, so widening the set would invent ("HEAD", "/health") etc. and fail the current census with no real surface change. 

We’ll extend the method set and expected census if those decorators actually appear.

tests/entrypoints/openai_api/test_api_server_guards.py
herotai214: Removed. The >= 400 check is the real one.

herotai214:
> Nice piece of work. I reviewed this as a fence rather than as ordinary tests, so instead of only checking that it passes I mutation-tested it: I injected each regression the module docstring claims to catch and checked whether the guard actually fired.
> 
> Environment: head `2dfdfcd1f`, vllm 0.27.0, CPU only.


Thanks for your detailed review @linyueqian ; I have fixed most of the issues.
Some lines still base on current structure before migrations; 
I intentionally commented TODOs for some lines to be edited in later P0.2 and P0.3;

PR #6203, opened 2026-08-14
[Perf][DreamZero] Add DreamZero online FP8 support

ZJLi2013:
## Purpose

- `quantization: fp8` was a no-op on DreamZero: `CausalWanModel` built its linears without a
  `quant_config`, so every GEMM ran in bf16 and nothing errored.
- Wiring only — no kernel, no `forward` change. `pipeline_dreamzero.py` passes
  `od_config.quantization_config`; `causal_wan_model.py` threads `quant_config` + `prefix`
  through `CausalWanModel`, `CausalWanAttentionBlock`, `CausalWanSelfAttention`, both
  cross-attention variants, the FFN and `MLPProj`. Same pattern as
  [`wan2_2_transformer.py`](https://github.com/vllm-project/vllm-omni/blob/main/vllm_omni/diffusion/models/wan2_2/wan2_2_transformer.py#L488-L570).
- Coverage: 402 linears (i2v) / 320 (t2v). Live per block in steady state: self-attn `qkv`/`o`,
  cross-attn `q`/`o`, `ffn.0`/`ffn.2`; cross-attn `k`/`v`/`k_img`/`v_img` come from
  `crossattn_cache`.
- No `ignored_layers` needed: the embeddings, head and action encoder are plain `nn.Linear` /
  `nn.Parameter` / `Conv3dLayer`.
- FFN prefixes are `ffn.0` / `ffn.2` — `nn.Sequential`'s child names and the checkpoint keys.
- Kernel choice stays upstream's `init_fp8_linear_kernel()`: ROCm →
  `Fp8PerTensorOnlineLinearMethod` + `ROCmFP8ScaledMMLinearKernel` (`torch._scaled_mm` →
  hipBLASLt, `e4m3fn`); CUDA → CUTLASS per-token. Measured on gfx950 only.
- Enable per stage: `quantization: fp8`.
- Not addressed: `init_weights()` calls `nn.init.uniform_` on parallel linears and would fail on
  an fp8-*serialized* checkpoint. The online path never reaches it.

## Test Plan

**vLLM Version:** 0.27.1 (torch 2.11.0+gitd0c8b1f, ROCm)

**vLLM-Omni Commit:** `a9578400`, on top of `e1aa6eae`

- **Unit, L1 / CPU** — `pytest tests/diffusion/models/dreamzero/test_dreamzero_quantization.py -v`,
  6 cases, markers `core_model` / `cpu` / `diffusion`, no Buildkite change. Asserts the set of
  linears that received a `quant_config` (402 / 320) and `prefix == module path`. A missing
  `quant_config` and a misspelled prefix both fail silently, so the set is what catches them;
  reverting the wiring turns all 6 red.
- **End-to-end A/B** — 2x MI350X (gfx950), DreamZero-DROID, TP=1 + CFG parallel=2, 15 chunks /
  16 steps, step 0 dropped as warmup, two configs differing by the `quantization` line alone.
  Step cache **off**, so both arms run the same number of DiT forwards. Production shape
  reported separately with each arm's forward count.
- **Determinism** — four real DiT shapes x four probes (activation quantization alone,
  `torch._scaled_mm` alone, the two chained, bf16 `torch.mm` control), fixed input, 20 repeats,
  `torch.equal`.
- **Output** — same-frame visual comparison against the bf16 arm.

## Test Result

```
collected 6 items

test_dreamzero_quantization.py::test_every_gemm_linear_gets_the_quant_config[1] PASSED [ 16%]
test_dreamzero_quantization.py::test_every_gemm_linear_gets_the_quant_config[2] PASSED [ 33%]
test_dreamzero_quantization.py::test_prefix_matches_module_path PASSED           [ 50%]
test_dreamzero_quantization.py::test_t2v_has_no_image_projections PASSED         [ 66%]
test_dreamzero_quantization.py::test_default_is_unquantized_and_prefixes_still_set PASSED [ 83%]
test_dreamzero_quantization.py::test_plain_linear_projections_stay_full_precision PASSED [100%]

6 passed, 15 warnings in 0.41s
```

Steady-state ms/step, step cache off:

| arm | ms/step | delta |
|---|---|---|
| bf16 | 1374.9 | — |
| bf16, repeated | 1380.1 | drift 5.2 |
| **fp8** | **1129.4** | **−245.5 (−17.9%)** |

All 16 per-step differences negative (−206.5 … −271.4, sd ≈ 15).

Production shape, step cache on:

| arm | ms/step | DiT forwards |
|---|---|---|
| bf16 | 494.3 | ~190 (inferred) |
| **fp8** | **398.3** | 184 |

Raw gap −96.0. Three steps each save ~78 ms more than the rest — one DiT forward the fp8 arm
skips. The other 12 steps mean **−80.4 ms/step, fp8's own contribution** (−79.8 on a second
tree).

Memory, rank 0, production shape:

| | bf16 | fp8 |
|---|---|---|
| model loading | 42.79 GiB | 27.70 GiB |
| peak reserved | 71.37 GiB | 58.24 GiB |
| peak allocated | 67.15 GiB | 54.12 GiB |

−15.09 GiB at load = halving a ~14B-parameter DiT's weights.

Determinism: bit-identical on all four shapes x four probes over 20 repeats.
`VLLM_OMNI_DETERMINISTIC` sets `cudnn.deterministic`, a different library from the hipBLASLt
fp8 GEMM, so this was checked separately.

Output: no artifacts, no structural drift.

Headroom: the GEMMs are 1.68–1.99x bf16 (−116.7 ms/step); the activation-quantization pass
costs ~45 ms/step back, so this realizes ~45% of the naive halving. Closing it means fusing
quantization into the producer epilogue.


vllm_omni/diffusion/models/dreamzero/pipeline_dreamzero.py
```diff
@@ -450,6 +450,7 @@ def __init__(self, *, od_config: OmniDiffusionConfig, prefix: str = "") -> None:
         transformer_kwargs["action_dim"] = ah_config["action_dim"]
         transformer_kwargs["max_state_dim"] = ah_config["max_state_dim"]
         transformer_kwargs["num_frame_per_block"] = ah_config["num_frame_per_block"]
+        transformer_kwargs["quant_config"] = getattr(od_config, "quantization_config", None)
```
hsliuustc0106: Siblings (`pipeline_flux_kontext.py`, `pipeline_ernie_image.py`) use `od_config.quantization_config` directly. The `getattr(..., None)` here would silently disable quantization if the field were ever renamed — direct attribute access matches the established pattern and fails loudly.

tests/diffusion/models/dreamzero/test_dreamzero_quantization.py
```diff
@@ -0,0 +1,179 @@
...
+    plain = {path for path, mod in model.named_modules() if isinstance(mod, nn.Linear)}
+    assert {
+        "text_embedding.0",
+        "text_embedding.2",
+        "time_embedding.0",
+        "time_embedding.2",
+        "time_projection.1",
+        "head.head",
+    } <= plain
+    assert not any(isinstance(mod, _FakeLinear) for mod in model.head.modules())
```
hsliuustc0106, line 181: Leaving `head.head` (and the text/time embeddings) in bf16 looks deliberate, but the module docstring says "Every GEMM-carrying linear must receive the pipeline's `quant_config`" — these are `nn.Linear` GEMMs too. Worth rewording the invariant to "every vLLM parallel linear gets `quant_config`; small plain-`nn.Linear` projections stay bf16", and a word on why the output head is excluded (output sensitivity?) would help the next reader.

hsliuustc0106:
Reviewed the wiring and the new CPU test — both look correct, and the A/B evidence in the description is thorough (drift control, determinism, memory).

One gap: nothing has actually executed `test_dreamzero_quantization.py` yet — CI on this PR ran only build/pre-commit/DCO, and the Buildkite test runs are label-gated. Could you run the test once on your side (`pytest tests/diffusion/models/dreamzero/test_dreamzero_quantization.py`) and paste the result, so we know the 6-passed count includes it?

vllm_omni/diffusion/models/dreamzero/pipeline_dreamzero.py
ZJLi2013: fixed

tests/diffusion/models/dreamzero/test_dreamzero_quantization.py
ZJLi2013: fixed and added

hsliuustc0106:
fix precomments please

vllm_omni/diffusion/models/dreamzero/pipeline_dreamzero.py
hsliuustc0106: Verified fixed at a9578400 — direct attribute access, matches the flux_kontext/ernie_image pattern.

tests/diffusion/models/dreamzero/test_dreamzero_quantization.py
hsliuustc0106: Verified fixed at a9578400 — invariant reworded and the head-head exclusion rationale added.

hsliuustc0106 (approved):
lgtm

lishunyang12 (changes requested):
Upd doc for fp8.md and show visual comparision, metrics such as PSNR or SSIM for eval of precision

vllm_omni/diffusion/models/dreamzero/pipeline_dreamzero.py
```diff
@@ -454,6 +454,7 @@ def __init__(self, *, od_config: OmniDiffusionConfig, prefix: str = "") -> None:
         transformer_kwargs["action_dim"] = ah_config["action_dim"]
         transformer_kwargs["max_state_dim"] = ah_config["max_state_dim"]
         transformer_kwargs["num_frame_per_block"] = ah_config["num_frame_per_block"]
+        transformer_kwargs["quant_config"] = od_config.quantization_config
```
lishunyang12, line 457: Resolve the transformer component before passing its config into these unprefixed blocks.

PR #6205, opened 2026-08-14
[Feature][Wan2.2] Per-transformer quantization for Wan2.2 I2V

kTorp:
## Purpose

Wan2.2 uses two transformers; the first acts as a coarse pass while the second handles finer details. Due to this architectural split it is often useful to set different quantization levels for each of the transformers, in order to not degrade the final output quality.

Using the `--diffusion-quantization-config` flag an end-user is able to specify quantization details, such as the method, activation scheme, and components to target.

However in the current codebase Wan2.2 is only able to specify a flat quantization config

```
--diffusion-quantization-config '{"method":"fp8","activation_scheme":"dynamic"}'
```

which further only targets the first transformer. The first transformer reads it's quantization settings from the `od_config.quantization_config` while the second reads from the transformer config directly. If the user tries to specify a specific transformer they are met with this message:
```
ValueError: All linear layers should support quant method.
```

This occurs because the unresolved `ComponentQuantizationConfig` is passed into the first transformer, whose internal layer names use blocks.* prefixes and therefore cannot resolve the requested transformer component. 

This PR adds the ability to set and experiment with per transformer quantization, allowing independent selection of existing methods.

## Test Plan

**vLLM Version:0.27.0**

**vLLM-Omni Commit: 5d50bd**

Tested with B200 USP=8. Temporary logging functions were used to inspect the transformer quantization.

#### Base command
```
vllm serve Wan-AI/Wan2.2-I2V-A14B-Diffusers \
--omni \
--ulysses-degree 8 \
--diffusion-quantization-config '{<See specific examples below>}'
```

#### Default flat config
```
--diffusion-quantization-config \
'{"method":"fp8","activation_scheme":"dynamic"}'

[quant-debug] transformer quant_config=<vllm.model_executor.layers.quantization.fp8.Fp8Config object at 0x7d57486e0ad0>
[quant-debug] transformer_2 quant_config=None
```

#### Default flat config
```
--diffusion-quantization-config \
'{"method":"fp8","activation_scheme":"dynamic"}'

[quant-debug] transformer_2 quant_config=<vllm.model_executor.layers.quantization.fp8.Fp8Config object at 0x716e218cc8c0>
[quant-debug] transformer quant_config=<vllm.model_executor.layers.quantization.fp8.Fp8Config object at 0x74b848d54a70>
```

#### Target first transformer
```
 --diffusion-quantization-config '{"transformer":{"method":"fp8","activation_scheme":"dynamic"}}'

[quant-debug] transformer quant_config=<vllm.model_executor.layers.quantization.fp8.Fp8Config object at 0x7d18f9528ad0>
[quant-debug] transformer_2 quant_config=None
```

#### Target second transformer
```
--diffusion-quantization-config '{"transformer_2":{"method":"fp8","activation_scheme":"dynamic"}}'

[quant-debug] transformer quant_config=None
[quant-debug] transformer_2 quant_config=<vllm.model_executor.layers.quantization.fp8.Fp8Config object at 0x783f84290920>
```

#### Target both transformers
```
--diffusion-quantization-config  '{"transformer":{"method":"fp8","activation_scheme":"dynamic"},"transformer_2": {"method":"int8","activation_scheme":"dynamic"}}'

[quant-debug] transformer quant_config=<vllm.model_executor.layers.quantization.fp8.Fp8Config object at 0x741823bcc770>
[quant-debug] transformer_2 quant_config=<vllm_omni.quantization.int8_config.DiffusionInt8Config object at 0x741823bcc800>
```

### DIFF
--- a/vllm_omni/diffusion/models/wan2_2/pipeline_wan2_2_i2v.py
+++ b/vllm_omni/diffusion/models/wan2_2/pipeline_wan2_2_i2v.py
@@ -281,7 +281,6 @@ class Wan22I2VPipeline(
             od_config.quantization_config,
             "transformer",
         )
-        logger.warning("[quant-debug] transformer quant_config=%r", transformer_quant_config)
         self.transformer = create_transformer_from_config(
             transformer_config,
             quant_config=transformer_quant_config,
@@ -292,7 +291,6 @@ class Wan22I2VPipeline(
                 od_config.quantization_config,
                 "transformer_2",
             )
-            logger.warning("[quant-debug] transformer_2 quant_config=%r", transformer_2_quant_config)
             self.transformer_2 = create_transformer_from_config(
                 transformer_2_config,
                 quant_config=transformer_2_quant_config,
--- a/vllm_omni/diffusion/models/wan2_2/pipeline_wan2_2_i2v.py
+++ b/vllm_omni/diffusion/models/wan2_2/pipeline_wan2_2_i2v.py
@@ -53,6 +53,13 @@ logger = logging.getLogger(__name__)
 DEBUG_PERF = False
 
 
+def _resolve_component_quant_config(quant_config, component: str):
+    component_configs = getattr(quant_config, "component_configs", None)
+    if component_configs is not None:
+        return component_configs.get(component, quant_config.default_config)
+    return quant_config
+
+
 def get_wan22_i2v_post_process_func(
     od_config: OmniDiffusionConfig,
 ):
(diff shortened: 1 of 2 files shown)


hsliuustc0106:
is this online fp8?

kTorp:
> is this online fp8?

Hi, yes the examples use online FP8: the BF16/FP16 transformer weights are converted to FP8 at model load time. With this PR I wanted to update the config routing so that transformer and transformer_2 could be targeted independently for quantization, as there are some use-cases where this is valuable in order to keep an acceptable level of output quality.

vllm_omni/diffusion/models/wan2_2/pipeline_wan2_2_i2v.py
```diff
@@ -304,24 +315,23 @@ def __init__(
...
-            if isinstance(t2_quant, dict) and "quant_method" in t2_quant:
-                from vllm_omni.quantization.factory import build_quant_config
-
-                method = t2_quant["quant_method"]
-                kwargs = {k: v for k, v in t2_quant.items() if k != "quant_method"}
-                t2_quant = build_quant_config(method, **kwargs)
-            else:
-                t2_quant = None
+            transformer_2_quant_config = _resolve_component_quant_config(
+                od_config.quantization_config,
```
lishunyang12, line 329: Keep auto-detected configs local; transformer_2 must retain its own quantization metadata.

kTorp:
It seems that a solution has landed in upstream with #7210, which now enables per-expert quantization for Wan2.2 on main.

That PR adds a `component` argument to `create_transformer_from_config` and a
`resolve_wan_transformer_quant_config` helper that selects a component config by exact
expert name before the checkpoint's storage contract is applied. This covers the same
gap this PR was opened for — `transformer_2` previously never received
`od_config.quantization_config` and could only read its own `config.json`

Therefore I will close this PR.

Thanks for the review @lishunyang12 and @hsliuustc0106

PR #6030, opened 2026-08-11
[Perf][Diffusion] Add auto stacked tiling for MiniMax-H3 VAE

baonudesifeizhai:
Depends on #6014. Tested together with #5979; this PR currently includes #6014's commit and will be rebased after it lands.

## Summary

- Add `--vae-stack-tiling {auto,true,false}` (default: `false`).
- Add a capability-gated stacked-tiling hook and wire MiniMax-H3's native implementation.
- In `auto`, stack decode tiles only when every VAE-parallel rank receives at least two tiles; otherwise keep sequential decoding.

## Benchmark

MiniMax-H3 FL2VA on B300 GPUs, eager + TORCH_SDPA, 8.7 s video, 2 denoising steps, 1 warmup + 5 measured runs. Video VAE is the mean; E2E is the median. Explicit `false` and `true` modes were compared.

| GPUs | Output | Tiles/rank | Video VAE (seq → stack) | VAE speedup | E2E (seq → stack) | E2E speedup |
|---:|---:|---:|---:|---:|---:|---:|
| 1 | 384×224 | 2 | 0.526s → 0.459s | 12.6% | 1.481s → 1.401s | 5.4% |
| 1 | 672×384 | 8 | 2.333s → 1.932s | 17.2% | 4.708s → 4.269s | 9.3% |
| 1 | 1024×576 | 15 | 4.377s → 3.551s | 18.9% | 10.701s → 9.908s | 7.4% |
| 1 | 1344×768 | 28 | 8.153s → 6.552s | 19.6% | 22.335s → 20.781s | 7.0% |
| 2 | 384×224 | 1 | 0.292s → 0.291s | 0.4% | 1.287s → 1.303s | -1.2% |
| 2 | 672×384 | 4 | 1.318s → 1.090s | 17.3% | 4.296s → 4.072s | 5.2% |
| 2 | 1024×576 | 7–8 | 2.386s → 1.973s | 17.3% | 12.504s → 12.086s | 3.3% |
| 2 | 1344×768 | 14 | 4.605s → 3.676s | 20.2% | 30.254s → 29.328s | 3.1% |
| 4 | 384×224 | N/A | N/A | N/A | N/A | N/A |
| 4 | 672×384 | 2 | 0.674s → 0.584s | 13.4% | 2.898s → 2.824s | 2.6% |
| 4 | 1024×576 | 3–4 | 1.333s → 1.119s | 16.1% | 8.248s → 8.086s | 2.0% |
[...]

PR #6131, opened 2026-08-12
[Perf][MiniMax-H3] Add online FP8 path for video VAE decoder

baonudesifeizhai:
PLEASE FILL IN THE PR DESCRIPTION HERE.

## Purpose
This PR builds on #5910, which has been merged.

## Summary

- Routes the `video_vae` component through the quantization configuration introduced by #5910.
- Applies online FP8 to 144 linear layers in the video VAE decoder TransformerBlocks:
  - `attn.to_qkv`
  - `attn.to_out`
  - `ff.w1`
  - `ff.w2`
- Keeps convolutions, norms, control flow, and input/output projections at their existing precision.
- Preserves the video VAE's FP16 input/output contract.
- Supports both single-GPU and 4-GPU VAE tile parallel execution.

## Benchmark

NVIDIA B300, 672×384, 4 seconds, 10 steps, eager mode.  
DiT and text encoder use FP8 in both cases; only the video VAE configuration changes.

| GPUs | Video VAE | Decode | E2E | Peak memory |
|---:|---|---:|---:|---:|
| 1 | Baseline | 1.205s | 4.988s | 72,278 MiB |
| 1 | Online FP8 | 1.104s | 4.874s | 66,602 MiB |
| 4 | Baseline | 0.367s | 2.203s | 71,686 MiB |
| 4 | Online FP8 | 0.343s | 2.183s | 66,214 MiB |

| GPUs | Decode improvement | E2E improvement | Memory reduction |
|---:|---:|---:|---:|
| 1 | 8.4% | 2.3% | 5,676 MiB |
| 4 | 6.7% | 0.9% | 5,472 MiB |

The 4-GPU case uses USP=4 and VAE tile parallel size 4. The VAE weights remain replicated across ranks, so FP8 reduces memory on every GPU.

## Quality

| Configuration | PSNR | SSIM |
|---|---:|---:|
| 1-GPU FP8 vs baseline | 46.55 dB | 0.98535 |
| 4-GPU FP8 vs baseline | 46.54 dB | 0.98530 |

Steady-state measurements exclude the first request because online FP8 kernel initialization adds cold-start overhead.

## Validation

- Single-GPU  
 
GPU |   | VAE Decode | E2E | Peak memory
-- | -- | -- | -- | --
1 | Baseline | 1.205s | 4.988s | 72,278 MiB
1 | FP8 + compile + stacked | 0.447s | 3.444s | 66,566 MiB

 
- 121 MiniMax-H3 contract and quantization tests passed.
[...]

PR #6206, opened 2026-08-14
[Core][Diffusion] Add independent MiniMax-H3 VAE process groups

linzhenpl07:
## Purpose

Implements rollout step 4 of #5948: allow MiniMax-H3 native tile-parallel VAE execution to use a deterministic process group smaller than the DiT group.

This PR is intentionally scoped to process-group composition. It does not implement component timing, stacked tiling, or regional VAE compilation, which are covered by #6014, #6030, and #5979. It does not modify H3 model code, model weights, latent shapes, or latent contracts.

## What changes

- Limits independent-group capability to MiniMaxH3Pipeline and MiniMaxH3ModularPipeline.
- Validates the requested VAE size and tile-only mode before distributed initialization.
- Requires the VAE size to fit and evenly divide both the worker world and DiT process-group topology.
- Creates deterministic contiguous VAE subgroups in one identical global order on every rank.
- Creates a dedicated communicator only when 1 < VAE size < DiT group size; size 1 remains local and full DiT size reuses the existing group.
- Uses a scoped ContextVar override only while binding the H3 VAE, then restores the original DiT group on normal and exceptional exits.
- Destroys the VAE group with model-parallel state and clears partial local state when initialization fails.
- Documents the supported topology and rejection behavior.

## Validation

- 25 passed: focused independent VAE process-group contract tests.
- 111 passed: unchanged MiniMax-H3 contract regression suite.
- 4-rank NCCL world with two 2-rank VAE subgroups: rank mapping, subgroup all-reduce, and teardown passed on 4x RTX 5090 32 GB.
- 4-rank Gloo world with two 2-rank VAE subgroups: passed.
- H3 FL2VA VAE group size 2 versus size 4, identical seed and 50-step request: MP4 files were byte-identical; decoded video and audio errors were exactly zero; a follow-up request returned HTTP 200.
- Ruff check, Ruff format check, compileall, and git diff check passed.
- vllm_omni/diffusion/models and tests/diffusion/models have zero diff from upstream.

The subgroup comparison measured 313.412 s at VAE size 2 versus 309.799 s at size 4 for one 1344x768, 5-second, 50-step request. This PR therefore claims process-group independence and configuration flexibility, not an end-to-end speedup.

Part of #5948.


linzhenpl07:
Author self-review completed.

What I checked:

- Initialization order: H3 topology validation runs before distributed initialization or model loading; the dedicated VAE group is created only after the base model-parallel groups exist.
- Collective safety: every rank creates the same deterministic contiguous subgroup list in the same order, and only retains its own group.
- Composition: size 1 remains local, a smaller valid H3 tile group gets a dedicated communicator, full DiT size reuses the existing group, and invalid worker-world/DiT/VAE divisibility or spatial mode fails early.
- Binding correctness: the scoped DiT-group override is active only during H3 VAE configuration. The H3 adapter persists the selected group and subgroup-local rank in its native parallel state; normal, nested, and exceptional exits restore the original DiT group.
- Lifecycle: duplicate initialization is rejected, partial local state is cleaned on creation failure, and model-parallel teardown destroys the VAE group. Collective initialization failures are propagated instead of retried on potentially damaged state.
- Scope: the final diff is 7 files with zero changes under vllm_omni/diffusion/models or tests/diffusion/models. Timing, stacked tiling, and VAE compile are intentionally excluded to avoid overlap with #6014, #6030, and #5979.

Validation:

- 25 focused process-group tests passed.
- 111 unchanged MiniMax-H3 contract tests passed.
- Real 4-rank Gloo and NCCL worlds with two 2-rank subgroups passed rank mapping, all-reduce, and teardown.
- H3 group-size 2 and 4 outputs were byte-identical with zero decoded video/audio error; the postcheck request passed.
- Ruff check/format, compileall, diff check, DCO, pre-commit, and Python 3.11/3.12 wheel builds passed.

Performance is not claimed: the qualified size-2 run was about 1.2% slower than size 4 in a single 50-step sample. The benefit here is independent VAE process-group sizing and topology isolation.

linzhenpl07:
A natural follow-up after this independent VAE process-group work is to investigate VAE/DiT pipeline overlap.

The current change isolates VAE communication and sizing, but it does not yet overlap VAE decode/postprocess with subsequent DiT denoising. A possible follow-up could evaluate:

- double-buffered VAE decode outputs;
- CUDA-event based handoff and readiness tracking;
- overlapping VAE decode/postprocess with the next DiT denoise step;
- reducing process-group barriers and unnecessary synchronization;
- synchronizing only when the output is actually consumed.

This should be driven by a stage-level profile first, since the end-to-end benefit depends on how much of the request is spent in VAE versus DiT. It may be useful to keep this as a separate follow-up PR so that the process-group topology change here remains independently reviewable.

hsliuustc0106:
> Author self-review completed.
> 
> What I checked:
> 
> * Initialization order: H3 topology validation runs before distributed initialization or model loading; the dedicated VAE group is created only after the base model-parallel groups exist.
> * Collective safety: every rank creates the same deterministic contiguous subgroup list in the same order, and only retains its own group.
> * Composition: size 1 remains local, a smaller valid H3 tile group gets a dedicated communicator, full DiT size reuses the existing group, and invalid worker-world/DiT/VAE divisibility or spatial mode fails early.
> * Binding correctness: the scoped DiT-group override is active only during H3 VAE configuration. The H3 adapter persists the selected group and subgroup-local rank in its native parallel state; normal, nested, and exceptional exits restore the original DiT group.
> * Lifecycle: duplicate initialization is rejected, partial local state is cleaned on creation failure, and model-parallel teardown destroys the VAE group. Collective initialization failures are propagated instead of retried on potentially damaged state.
> * Scope: the final diff is 7 files with zero changes under vllm_omni/diffusion/models or tests/diffusion/models. Timing, stacked tiling, and VAE compile are intentionally excluded to avoid overlap with [[Observability] Add MiniMax-H3 video/audio VAE decode timings #6014](https://github.com/vllm-project/vllm-omni/pull/6014), [[Perf][Diffusion] Add auto stacked tiling for MiniMax-H3 VAE #6030](https://github.com/vllm-project/vllm-omni/pull/6030), and [[Perf] Regionally compile MiniMax H3 video VAE transformer blocks #5979](https://github.com/vllm-project/vllm-omni/pull/5979).
> 
> Validation:
> 
> * 25 focused process-group tests passed.
> * 111 unchanged MiniMax-H3 contract tests passed.
> * Real 4-rank Gloo and NCCL worlds with two 2-rank subgroups passed rank mapping, all-reduce, and teardown.
> * H3 group-size 2 and 4 outputs were byte-identical with zero decoded video/audio error; the postcheck request passed.
> * Ruff check/format, compileall, diff check, DCO, pre-commit, and Python 3.11/3.12 wheel builds passed.
> 
> Performance is not claimed: the qualified size-2 run was about 1.2% slower than size 4 in a single 50-step sample. The benefit here is independent VAE process-group sizing and topology isolation.

please review with inline comments to make sure you know what you have submitted:)

linzhenpl07:
I took another manual pass over the full diff and checked the integration against the latest `main`.

The primary four-rank configuration used during validation (`WORLD=4`, DiT=4, VAE=2) looks sound:

* All ranks create the subgroups in the same deterministic order.
* The VAE process-group binding is correctly scoped.
* The original DiT process group is restored on both normal and exceptional exits.
* Subgroup state is cleaned up during initialization failure and teardown.

However, I found one topology-validation case that should be tightened before merge. The validator currently accepts `WORLD=8 / DiT=4 / VAE=2`, and the unit test treats this configuration as supported. However, the current `init_dit_group(4)` creates only the `[0,1,2,3]` DiT group; it does not create a second DiT partition for ranks `[4,5,6,7]`. Meanwhile, the VAE initializer would still create subgroups across all eight ranks. Therefore, the accepted configuration does not match the actual runtime DiT topology.

For this PR, I think the safest fix is to:

* Require the worker `WORLD` size to equal the DiT process-group size.
* Change the `8/4/2` test into a rejection case.

Support for multiple DiT partitions should be handled separately in future work, with an explicit rank layout.

I also noticed a minor inconsistency in the CLI help text. It currently states that every H3 VAE size greater than one uses an independent process group. However, when the VAE size equals the DiT size, the implementation intentionally reuses the existing DiT process group. The help text should instead state:

> An independent process group is created only when the VAE size is valid and strictly smaller than the DiT size.

Apart from these two items, I did not find any other correctness issues in the following areas:

* Process-group creation order
* Scoped process-group override
* Cleanup on initialization failure
* Teardown cleanup
* Non-H3 fallback behavior

vllm_omni/diffusion/worker/diffusion_worker.py
```diff
@@ -264,6 +270,21 @@ def init_device(self) -> None:
         """Initialize the device and distributed environment."""
         world_size = self.od_config.num_gpus
         rank = self.rank
+        parallel_config = self.od_config.parallel_config
+        supports_independent_vae_group = supports_independent_vae_process_group(self.od_config.model_class_name)
+        if supports_independent_vae_group:
```
hsliuustc0106: This validation is gated on the model supporting independent VAE groups, so it also rejects H3 serves that never use the feature. Config allows `num_gpus > parallel_config.world_size` (only the smaller case raises, in the `OmniDiffusionConfig` post-init), and on main an H3 serve with `--num-gpus 8` over a tp2 x usp2 topology and `vae_patch_parallel_size=1` starts fine with a local VAE. Here every worker dies in `init_device` with "independent MiniMax-H3 VAE process groups currently require DiT process group size (4) to equal diffusion world_size (8)" — for a feature the user never requested. The `vae_patch_parallel_size == world_size` reuse-DiT-group mode hits the same rejection.

Gate on the request instead, e.g. `if parallel_config.vae_patch_parallel_size > 1 and supports_independent_vae_group:` — feature-off serves keep main behavior, and the doc wording ("not supported by this path") stays accurate.

vllm_omni/diffusion/distributed/vae_parallel_state.py
```diff
@@ -0,0 +1,152 @@
+# SPDX-License-Identifier: Apache-2.0
```
hsliuustc0106, line 1: @Bounty-hunter PTAL whether this abstraction is useful and valid

hsliuustc0106:
I drew a diagram of how this PR's process-group composition reads end-to-end, to confirm my understanding and to make the one open concern easier to discuss (the image is tall — open it full-size):

PR 6206 — independent MiniMax-H3 VAE process groups: rank topology, subgroup creation discipline, three operating modes, startup and binding sequence, the scoped DiT-group override, teardown, and the validation contract

Short readback of the parts that carry the correctness:

- The worker world partitions into deterministic contiguous VAE subgroups (`[0,1]` / `[2,3]` at size 2 in a 4-rank world); every rank creates every subgroup in one identical global order and retains only its own — the right `dist.new_group` discipline.
- Binding composes because `MiniMaxH3VideoVAE.set_parallel_size` captures `get_dit_group()` eagerly and stores it in the checkpoint's native parallel state. The scoped `override_dit_group` ContextVar swaps that single lookup for the duration of the call, so decode-time VAE collectives follow the subgroup while the pipeline's broadcasts and `_dit_rank_world()` keep using the real DiT group.
- Size 1 and full-DiT-size keep the existing behavior; teardown rides `destroy_model_parallel` with null-then-destroy, so it is idempotent and cannot accidentally destroy WORLD.

The remaining concern is the inline comment at `diffusion_worker.py:275`, and it is the flip side of the `WORLD == DiT` tightening from your self-review pass: the validator is gated on `supports_independent_vae_process_group(...)` — model capability — rather than on the feature being requested. So an H3 serve with `vae_patch_parallel_size=1` (or `== world_size`) over `--num-gpus 8` with a parallel world of 4, which main accepts, now fails in `init_device` with the "independent ... must equal diffusion world_size" error — for a feature the user never enabled.

Gating on the request still rejects the `8/4/2` case you found:

```python
if parallel_config.vae_patch_parallel_size > 1 and supports_independent_vae_group:
    validate_independent_vae_parallel_config(...)
```

Everything else in the composition (creation order, scoped override, failure cleanup, teardown, non-H3 fallback) reads as correct to me.

<sub>Editable HTML source of the diagram: [pr-6206-vae-groups-design.html](https://github.com/hsliuustc0106/vllm-omni-cookbook/blob/main/docs/visualizations/pr-6206-vae-groups-design.html)</sub>

linzhenpl07:
Thanks for the diagram and the readback — that made the concern easy to act on. Pushed `b03f25ef`.

## What changed

Gated the validation on `requires_independent_vae_process_group` rather than on model capability:

```python
use_independent_vae_group = requires_independent_vae_process_group(
    self.od_config.model_class_name,
    parallel_config.world_size,
    parallel_config.vae_patch_parallel_size,
)
if use_independent_vae_group:
    validate_independent_vae_parallel_config(...)
```

It already decides whether a dedicated group is created and was already computed a few lines below, so validation now runs exactly when the contract applies, and no new predicate is needed. Net `+38/-8`.

## One deviation from the suggested snippet, with the reason

I checked `vae_patch_parallel_size > 1 and supports_independent_vae_group` against the cases you listed. It does not cover the second one — at `vae == DiT size` the size is above one, so validation still runs and the DiT-equals-WORLD rule still fires:

| config (feature not requested) | this PR before | `> 1` gate | pushed fix |
|---|---|---|---|
| `--num-gpus 8` over tp2 x usp2, `vae=1`, tile | RAISES | skipped | skipped |
| **`vae == DiT size` (reuse-DiT-group)** | RAISES | **RAISES** | **skipped** |
| `vae=1` with `spatial_shard_height` | RAISES | skipped | skipped |
| `vae=2` dedicated group, `8/4/2` (must reject) | RAISES | RAISES | RAISES |

Row 3 is a third instance of the same class that the `mode != "tile"` rule produced: `spatial_shard_height` / `spatial_shard_width` with a rank-local VAE were also rejected. The same gate fixes it.

Row 4 is the `8/4/2` composition from my earlier self-review pass — it requests a strictly smaller group, so it is still validated and still rejected.

Happy to switch to the inline `> 1` form if you prefer it, but it would leave the reuse-DiT-group case failing.

## Validation

End-to-end on the real `init_device`, constructing `num_gpus=4 > parallel_config.world_size=2` with `vae_patch_parallel_size=1`:

```
before: ValueError: independent MiniMax-H3 VAE process groups currently require
                    DiT process group size (2) to equal diffusion world_size (4)
after:  passes the gate, proceeds to device setup
```

(The first attempt at this via CLI flags was not a valid repro — `--num-gpus 4` is reconciled down to the parallel world size by the stage layer, so `num_gpus == world_size` and the check never fired. Constructing `OmniDiffusionConfig` directly is what reaches the state, which matches your note that only the `num_gpus <` case raises in post-init.)

Plus: `tests/diffusion/distributed/test_vae_process_group.py` 28 passed; MiniMax-H3 contract tests 172 passed / 10 deselected; ruff check and format clean. All CPU-only.

New tests pin both no-dedicated-group modes, alongside a guard that the contract does reject those topologies once a dedicated group is actually requested.

## Still open

@Bounty-hunter's read on whether the abstraction is useful and valid — that one is independent of this fix, and I have not heard back yet. If it would help, I can write up why an independently sized VAE group is a prerequisite for the VAE-disaggregation item in the #5700 roadmap (section 3.9), rather than only a topology-isolation knob.

linzhenpl07:
On the open question about whether the abstraction in `vae_parallel_state.py` is useful and valid — I went back through the codebase and concluded the capability is real but I had given it the wrong carrier. Pushed `ff442d42`, net `+36/-3`.

## What was wrong

The module kept its own model-name allowlist:

```python
_INDEPENDENT_VAE_GROUP_MODELS = frozenset({"MiniMaxH3Pipeline", "MiniMaxH3ModularPipeline"})
```

That is a second implementation of something `vllm_omni/diffusion/io_support.py` already does, in the same shape, for `support_image_input`, `support_audio_input`, `support_audio_output`, `color_format`, and `dummy_run_num_frames`:

```python
model_cls = DiffusionModelRegistry._try_load_model_cls(model_class_name)
return bool(getattr(model_cls, "support_audio_output", False))
```

`MiniMaxH3Pipeline` already declares `_supports_mmap_loading: ClassVar[bool] = False` for the same kind of per-model decision. So the allowlist was not a new abstraction, it was a parallel one.

## What it is now

Declared where the property lives:

```python
# pipeline_minimax_h3.py, next to _supports_mmap_loading
supports_independent_vae_process_group: ClassVar[bool] = True
```

and resolved through the existing idiom:

```python
def supports_independent_vae_process_group(model_class_name: str) -> bool:
    from vllm_omni.diffusion.registry import DiffusionModelRegistry  # registry imports this module

    model_cls = DiffusionModelRegistry._try_load_model_cls(model_class_name)
    return model_cls is not None and bool(getattr(model_cls, "supports_independent_vae_process_group", False))
```

Both H3 registry names map to the same class in `registry.py`, so the one declaration covers both, and there is now a test pinning that. A model that has never heard of the feature stays off by default, and a model that wants it declares one attribute instead of being added to a list inside the distributed layer.

## Why this cannot reuse the check already in registry.py

`registry.py:398` answers the neighbouring question with the object:

```python
is_distributed_vae = hasattr(model, "vae") and isinstance(model.vae, DistributedVaeMixin)
```

That is the better check, but it is only available after construction, and process groups must exist before the model is loaded. The class is all that is in hand at `init_device` time, which is precisely what `_try_load_model_cls` gives. I put that reasoning in the docstring so the next reader does not have to rediscover why the object-based check was not reused.

One related inconsistency I did not touch here: `registry.py` treats an unsupported `vae_patch_parallel_size` as warn-and-ignore, while this path raises. Both are defensible — a silently ignored patch-parallel size costs performance, an invalid process-group topology costs a hang — but they should probably be one documented policy. That belongs with the capability-discovery step of #5948 rather than in this PR, and it is the piece the other in-flight H3 VAE PRs (#6014, #6030, #5979, #6131) would each otherwise have to invent separately.

## Validation

- `tests/diffusion/distributed/test_vae_process_group.py`: 30 passed (28 before, plus one that the declaration lives on the pipeline class and both registry names resolve to it, and one that an unregistered name does not claim the capability).
- MiniMax-H3 tests: 147 passed, 10 skipped, unchanged.
- Import-order smoke test from a fresh interpreter importing `vae_parallel_state` first, i.e. the reverse of `registry.py`'s import, to confirm the lazy import does not cycle: H3 and the modular name resolve `True`, `WanPipeline` and an unregistered name `False`, and `requires_...` still gives `True/False/False` for `4/2`, `4/1`, `4/4`.
- ruff check and format clean.

No behavior change: the same two H3 names resolve to the same answer as the allowlist did, so the gating fix in `b03f25ef` and the topology contract are untouched.

linzhenpl07:
Synced `main` in `dacdd4f4` — the branch went `CONFLICTING` right after the previous push and is `MERGEABLE` again.

One real conflict, in `pipeline_minimax_h3.py`: #6213 removed `_supports_mmap_loading` from the class, and the new declaration had landed next to it. Resolved by keeping main's removal and leaving only `supports_independent_vae_process_group`. So the "next to `_supports_mmap_loading`" in my comment above is now stale — that attribute is gone from main; the reasoning about following the same class-attribute idiom is unaffected, and `io_support.py` remains the live precedent.

`registry.py` and `diffusion_worker.py` auto-merged; I checked that the `requires_independent_vae_process_group` gate from `b03f25ef` and its `initialize_vae_parallel_group` call site both came through intact.

Re-ran on the merge commit: `tests/diffusion/distributed` 241 passed / 52 skipped, MiniMax-H3 and VAE process-group tests 194 passed / 10 skipped, ruff clean.

hsliuustc0106:
is it related to #5885

Issue #5912, opened 2026-08-08
[Feature]: ROCm parity for verified vllm omni recipes


functionstackx:
### 🚀 The feature, motivation and pitch

hi @andyluo7 @chunfangamd

AMD's blog post here says that AMD support vllm-omni project as first class day 0 https://rocm.blogs.amd.com/software-tools-optimization/vllm-omni/README.html#vllm-omni-high-performance-omni-modality-on-amd

but when looking at verified recipes at [recipes/README.md](https://github.com/vllm-project/vllm-omni/blob/44e38d7b3879107e501acfd833765ef07c60ace6/recipes/README.md?plain=1#L33), AMD only has 2.5% parity with CUDA 

what is the timelines for when verified recipe will be over 95%?

### Data from https://github.com/vllm-project/vllm-omni/blob/44e38d7b3879107e501acfd833765ef07c60ace6/recipes/README.md?plain=1#L33 as of August 7 2026 7pm PT

| Recipe | Task | Hardware | Verified NVIDIA recipe | Verified ROCm recipe |
|--------|------|----------|---------------|-------------|
| [`audiox/AudioX.md`](./vllm-omni/recipes/audiox/AudioX.md) | Offline + online unified text/video→audio diffusion | 1x L4 24GB | ✅ | ❌ |
| [`Baidu/ERNIE-Image.md`](./vllm-omni/recipes/Baidu/ERNIE-Image.md) | Text-to-image online serving (ERNIE-Image 8B) | 1x or 2x RTX 4090 24GB | ✅ | ❌ |
| [`Bagel/BAGEL-7B-MoT.md`](./vllm-omni/recipes/Bagel/BAGEL-7B-MoT.md) | Text-to-image with shared online/offline examples | 1x A100 80GB / 2x CUDA GPUs | ✅ | ❌ |
| [`Boogu/Boogu-Image.md`](./vllm-omni/recipes/Boogu/Boogu-Image.md) | Text-to-image online serving (Boogu-Image-0.1-Base) | 1x A100/H100 40GB+ | ✅ | ❌ |
| [`BosonAI/Higgs-Audio-V3-TTS.md`](./vllm-omni/recipes/BosonAI/Higgs-Audio-V3-TTS.md) | Online + offline multilingual TTS with voice cloning | 1x H100 80GB | ✅ | ❌ |
[...]

PR #5766, opened 2026-08-04
[Bugfix] Fix SenseNova-U1 head dtype config compatibility

EchoHayate:
### What broke

The SenseNova-U1 weekly LoRA test fails during model initialization on vLLM 0.26 with:

```text
AttributeError: '_DiffusionVllmModelConfig' object has no attribute 'head_dtype'
```

### Reproduction

```bash
python -m pytest -q \
  tests/e2e/offline_inference/test_sensenova_u1_expansion.py::test_sensenova_u1_lora_scale_and_deactivation
```

The focused regression can be run with:

```bash
python -m pytest -q tests/diffusion/test_diffusion_vllm_model_config.py
```

### Root cause

Diffusion workers install `_DiffusionVllmModelConfig` as `VllmConfig.model_config`. vLLM 0.26's `LogitsProcessor` now reads `model_config.head_dtype` during initialization, but the diffusion compatibility config did not expose that part of the `ModelConfig` contract.

### Fix

Expose `head_dtype` from `_DiffusionVllmModelConfig`, returning the model dtype. This matches vLLM's default behavior for generation models and preserves the existing logits precision. Add a CPU regression test that installs the diffusion config in `VllmConfig` and constructs the real vLLM `LogitsProcessor`.

Fixes #5748.

## Test Plan

**vLLM Version:** 0.26.0

**vLLM-Omni Commit:** `bc1b8bb70b37ee48b2214716455e231cb4022ded`

```bash
uvx --offline ruff check \
  vllm_omni/diffusion/worker/diffusion_worker.py \
  tests/diffusion/test_diffusion_vllm_model_config.py

uvx --offline ruff format --check \
  vllm_omni/diffusion/worker/diffusion_worker.py \
  tests/diffusion/test_diffusion_vllm_model_config.py

python -m pytest -q \
  tests/diffusion/test_diffusion_vllm_model_config.py \
  tests/diffusion/test_diffusion_worker_cuda_profiler.py
```

## Test Result

```text
5 passed, 18 warnings in 0.91s
```

The focused CPU test reproduces the exact `LogitsProcessor` initialization failure on the pre-fix code and passes with this change. The full 8B H100 LoRA generation test was not rerun locally because the SenseNova-U1 checkpoint was not cached; the existing weekly E2E remains the end-to-end validation boundary.

PR #6207, opened 2026-08-14
[ROCm][Recipe]  Add MI300X coverage for verified recipes

akshatvishu:
## Summary
part of #5912

This PR adds single GPU AMD MI300X validation to the existing Ming Omni TTS, Qwen3 TTS, SenseNova U1, Stable Audio Open and MammothModa2 recipes. It also adds a new OmniVoice recipe.

All hardware runs used one AMD MI300X and the official ROCm image built from `docker/Dockerfile.rocm`. Each recipe records the tested environment, command, output check, timing and memory use.

## Validation

- Ming Omni TTS produced a valid 44.1 kHz mono WAV.
- Qwen3 TTS completed CustomVoice, VoiceDesign, and Base requests. Both the current eager setup and the setup using `MIOPEN_FIND_MODE=FAST` completed.
- SenseNova U1 produced a valid 1536 by 2720 RGB PNG with 50 inference steps.
- Stable Audio Open produced a valid 10 second, 44.1 kHz stereo WAV.
- MammothModa2 Preview produced a valid 1024 by 1024 RGB PNG with 50 inference steps.
- OmniVoice automatic voice produced a valid 24 kHz mono WAV.

## Scope

The PR changes recipe documentation only. 

The SenseNova U1 recipe depends on [PR #5766](https://github.com/vllm-project/vllm-omni/pull/5766), which fixes the missing `head_dtype` error.

## Checks
- Pre-commit on all changed recipe files
- Hardware validation on a single AMD MI300X


hsliuustc0106 (approved):
lgtm, btw, do you prefer an indepdent recipe md for ROCm?

akshatvishu:
@hsliuustc0106  Currently, we follow the same recipe format as the vLLM recipes(<https://github.com/vllm-project/recipes/>) .So, if our longer-term goal is to make hardware-specific recipes more visible and maintain separate recipes for different hardware providers, I think now would be a good time to split ROCm out. (so , per model we will have at most like cuda, ROCm and XPU .md)

We only have around ~10-12 ROCm recipes at the moment, so the migration should still be relatively easy in-case we choose to follow this up.

hsliuustc0106:
> @hsliuustc0106 Currently, we follow the same recipe format as the vLLM recipes(https://github.com/vllm-project/recipes/) .So, if our longer-term goal is to make hardware-specific recipes more visible and maintain separate recipes for different hardware providers, I think now would be a good time to split ROCm out. (so , per model we will have at most like cuda, ROCm and XPU .md)
> 
> We only have around ~10-12 ROCm recipes at the moment, so the migration should still be relatively easy in-case we choose to follow this up.

we will discuss a rule to follow

PR #6208, opened 2026-08-15
[Perf][CI] Reuse the Whisper judge worker across a test module

ShengleiFu:
## Purpose

The Whisper judge used to validate audio outputs (`tests/helpers/media.py::convert_audio_file_to_text`) built a fresh `spawn` `ProcessPoolExecutor` and reloaded the model for every transcription, so each call paid a process spawn plus `whisper.load_model` before the ~7% of its time that is actual transcription. On a transcript-gated test that judges many samples this dominates the wall time.

This PR:
- Reuses one transcription worker and caches the loaded Whisper model inside it (keyed by model size, so an ASR escalation to `large-v3` loads that once too), instead of spawning and loading per call.
- Serializes each `submit → result → cleanup` on the parent side with a call lock. The single worker already transcribes serially (`max_workers=1`), so this costs no throughput — it just makes "one call in flight at a time" explicit so a failed call can tear the worker down without racing another caller.
- Preserves the old one-process-per-call failure isolation: any failure discards the worker — and its resident model — before re-raising, including a `BaseException` such as `KeyboardInterrupt`/`SystemExit`. A dead worker (`BrokenProcessPool`) is additionally retried once.
- Releases the worker when each server/runner fixture instance tears down and before the next one constructs — factored into one `_whisper_device_free_around()` context manager — so a Whisper model never stays resident on an accelerator while the next parametrized server initializes. A module-scoped autouse fixture is the backstop for tests that transcribe without those fixtures.
- Keeps `spawn` because CUDA is not fork-safe.

## Test Plan

- [x] `pytest tests/helpers/tests/test_media.py tests/helpers/tests/test_runtime_transcriber_release.py` → 12 passed (2x RTX PRO 6000, vllm-omni + vLLM as installed)
- [x] Key behaviors verified non-vacuous by mutation (removing each guard reddens exactly its test): executor reuse, worker-side model cache, `BrokenProcessPool` retry, discard-on-failure for both `Exception` and `BaseException`, the parent call-lock serialization, and the fixture release ordering (enter/exit and throw→finally).
- [x] `ruff check` + `ruff format` + `typos` + `check-test-marks` clean on the changed files.

## Test Result

```
tests/helpers/tests/test_media.py ..........          [10 passed]
tests/helpers/tests/test_runtime_transcriber_release.py .. [2 passed]
12 passed
```

Isolated before/after of the transcription helper (2x RTX PRO 6000, whisper-small, 3.48s clip, 10 calls; the helper selects `cuda:1` on its own, matching `test_one_word_prompt_001`'s `num_cards=2`):

| | before | after |
|---|---:|---:|
| serial wall | 163.6s | 16.0s |
| 5-way concurrent wall | 46.5s | 16.0s |
| warm per-call | 16.2s | 0.25s |

The change touches only `tests/helpers`, so end-to-end server time is unchanged; the win is the judge's per-transcription cost. Not a duplicate: no open PR touches `convert_audio_file_to_text` or the Whisper judge worker; this is the "reuse the Whisper process/model" follow-up deferred from the success-rate gate work.

AI assistance was used for this change.


ShengleiFu:
**Self-review**

What this changes: the Whisper judge (`tests/helpers/media.py`) previously spawned a fresh `ProcessPoolExecutor` and reloaded the model for every transcription. This reuses one worker and caches the model in it, and releases the worker at each server/runner fixture boundary so it never sits on an accelerator while the next parametrized server initializes.

What I checked:
- **Tests**: `pytest tests/helpers/tests/test_media.py tests/helpers/tests/test_runtime_transcriber_release.py` → 17 passed (2x RTX PRO 6000). I verified each new test is non-vacuous by mutation — removing each guard reddens exactly its test (executor reuse, worker-side model cache, `BrokenProcessPool` discard+retry, the concurrent late-failure discard race, `max_workers=1`/`spawn` pinning, and the release ordering across all three backing paths `OmniServer`/`OmniServerStageCli`/`OmniRunner`).
- **Lint**: `ruff check` + `ruff format` + `typos` + `check-test-marks` clean on the changed files.
- **Concurrency-sensitive spots I scrutinized**: the worker is shared across concurrent judge calls, so `_discard_transcriber` is identity-checked — a late `BrokenProcessPool` report cannot shut down a replacement worker another caller already installed; and only worker-process death is recovered from, while a `torch` OOM raised by the transcription propagates and keeps the worker (same as before, one process per call). `spawn` is kept because CUDA is not fork-safe.
- **Perf** (isolated helper before/after, whisper-small, 3.48s clip, 10 calls; helper selects `cuda:1` on its own, matching `test_one_word_prompt_001`'s `num_cards=2`): serial 163.6s → 16.0s; 5-way concurrent 46.5s → 16.0s; warm per-call 16.2s → 0.25s.

Scope: changes are confined to `tests/helpers`, so server-side behavior is unchanged; the parametrized-server-switch release path is covered hermetically by `test_runtime_transcriber_release.py` and exercised end-to-end by the model tests in CI. AI assistance was used.

tests/helpers/fixtures/media.py
```diff
@@ -0,0 +1,24 @@
+"""Fixtures for the reused Whisper transcription worker."""
+
+import sys
+
+import pytest
```
yenuo26, line 5: This filename feels too specific. Maybe just calling it media.py would be more appropriate.

tests/helpers/tests/test_runtime_transcriber_release.py
```diff
@@ -0,0 +1,94 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+
+"""No server or runner may initialize while a Whisper worker still holds the device."""
+
+import threading
```
yenuo26, line 11: I don't think it needs a separate test file. Maybe we can just merge it into test_media.py.

tests/helpers/media.py
```diff
@@ -670,59 +671,129 @@ def _serialize_whisper_model_download(model_size: str = "small"):
...
+    pool for every later caller, so it is replaced and the call retried once. An
+    exception raised *by* the transcription, a ``torch`` OOM included, is
+    propagated to the caller with the worker left intact -- same as the caller
+    saw when each transcription had its own process.
+    """
+    for attempt in range(2):
+        executor = _get_transcriber()
+        try:
+            return executor.submit(_whisper_transcribe_in_current_process, output_path, model_size, language).result()
+        except BrokenProcessPool:
```
yenuo26: Right now, only BrokenProcessPool will drop the worker. If the transcription task raises an error on its own — including torch OOM, as noted in the comment and tested in test_transcription_error_propagates_and_keeps_the_worker — the process stays alive.

tests/helpers/fixtures/media.py
ShengleiFu: Renamed, thanks!

tests/helpers/tests/test_runtime_transcriber_release.py
ShengleiFu: I'd lean toward keeping it separate. `test_media.py` imports only `tests.helpers.media` (deliberately light — numpy/soundfile, no torch/vLLM), so it runs standalone in well under a second. `test_runtime_transcriber_release.py` drives `iter_omni_server`/`iter_omni_runner`, so it imports `tests.helpers.runtime` (torch, vLLM, psutil). Merging would pull that heavy import into `test_media.py` at module load and lose the light, quick-to-run property. They also cover different modules (`media.py` vs the `runtime.py` fixtures). Happy to merge if you'd still prefer one file — just flagging the tradeoff.

tests/helpers/media.py
ShengleiFu: Good catch! I serialized `submit` + `result` on the parent side(there is already only one child worker), then discard the executor and re-raise on any task-level exception. `BrokenProcessPool` will retain the existing retry-once behavior. This restores the old failure-isolation semantics without introducing overlapping workers.

tests/helpers/tests/test_media.py
```diff
@@ -2,6 +2,9 @@
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
 
 import sys
+import threading
```
yenuo26: 1. Suggest removing test_late_failure_does_not_discard_the_replacement_pool. After the call lock, this interleaving can no longer reach convert_audio_file_to_text. It's testing a private _discard_transcriber identity check, not the helper's public contract.
2. Suggest removing test_concurrent_callers_serialize_onto_one_worker and test_release_waits_for_an_in_flight_transcription. pytest already runs tests serially within the same worker, and fixture teardown happens after the test returns, so it doesn't race with an in-flight judge. These tests use join(timeout=0.5) to prove liveness, adding sleep/flake overhead while guarding implementation details rather than real e2e paths.
3.test_replacement_pool_failure_also_discards can be merged into the BrokenProcessPool test (second failure → both pools shutdown, _TRANSCRIBER is None). test_worker_pool_is_a_single_spawned_processcan be merged into the reuse test.

tests/helpers/tests/test_runtime_transcriber_release.py
yenuo26: I'm fine with keeping the two files, but I really don't think test helpers need to be covered at production-code granularity. _whisper_device_free_around() is just one implementation, and iter_omni_server / iter_omni_runner are both just wrappers around it. 3 backing paths × 2 scenarios = 6 tests for the same 6-line context manager. One enter/exit test plus one throw → finally test may be sufficient; no need to run it separately for OmniServer, StageCli, and Runner.

yenuo26:
@vllm-omni-review-bot

tests/helpers/tests/test_media.py
ShengleiFu: updated, ptal, thanks!

yenuo26 (approved):
LGTM

yenuo26:
@hsliuustc0106 @Gaohan123 Could you please help to add tts-test label?

TaffyOfficial:
@vllm-omni-review-bot

ShengleiFu:
Hi @hsliuustc0106 , CI passed, could you help merge it, thanks!

hsliuustc0106:
> Hi @hsliuustc0106 , CI passed, could you help merge it, thanks!

this should help reduce CI time, @yenuo26 PTAL and compare recent time and future time

ShengleiFu:
**Post-merge CI impact (measured on Buildkite nightlies)**

Now that this has been on `main` for a few nightly cycles, here's the realized wall-time impact, since the win only shows where a single pytest process judges many clips.

The affected step is the nightly full-model `:full_moon: Omni · Function Test with H100` (`pytest tests/e2e -m "full_model and H100 and omni"`), which aggregates the most Whisper transcriptions in one process:

| nightly build | vs. merge (#13837) | step wall-clock |
|---|---|---:|
| 13832 | before | 103m01 |
| 13862 | after | 60m55 |
| 13920 | after | 57m51 |
| 13948 | after | 55m06 |
| 13987 | after | 54m47 |
| 14022 | after | 57m36 |

~103m → stable ~55m, ≈45 min/night, on an unchanged 18-file test set.

Per-test timing (from the same step's log, `_bk;t=` timestamps, before vs after) confirms the cause is the judge and not a global speedup — transcription-gated tests collapse, text-only tests are flat:

| test | before | after |
|---|---:|---:|
| `test_one_word_prompt_001[default]` (the 40-request success-rate gate) | 305.6s | 13.4s |
| `test_speaker_002[default]` | 114.1s | 16.2s |
| `test_audio_in_video_002[default]` | 108.0s | 14.5s |
| `test_text_to_audio_001[default]` (synthesis, not judged) | 278.3s | 276.9s |
| `test_text_video_to_text_001[default]` (text-only) | 9.0s | 9.3s |
| sum of timed tests | 6082s | 3364s |

The timed-test sum drop (−2718s) matches the step wall-clock drop. Only #6208 touched `tests/helpers/media.py` in that window. Per-PR `ready` steps that judge only a few clips see a smaller (~1–2 min) gain.

PR #6213, opened 2026-08-15
[Diffusion] Add loader-owned host-weight plans for DLO (TP=1)

hsliuustc0106:
## Motivation

DLO currently couples checkpoint-storage decisions to the transfer backend. In particular, the no-AllGather path falls back to ordinary loader tensors, so pure-DP workers retain private full host copies even when they could map the same checkpoint pages. Loader/backend gates can also diverge and leave a model on `meta` after `load_weights()` was skipped.

This implements the Phase A design discussed in #6195 and supersedes the prototype in #6198.

## Design

- Make `DiffusersPipelineLoader` the single owner of checkpoint semantics. It preflights topology, key mapping, coverage, shapes, dtypes, and custom-loader compatibility before deciding whether to skip ordinary weight materialization.
- Transfer one exact `HostWeightPlan` from the loader to DLO. The backend consumes that plan without rescanning checkpoint files or consulting pipeline capability flags.
- Skip only dedicated DiT component sources owned by the plan. Other sources, such as MiniMax-H3's text encoder, continue through the ordinary component loader and remain part of strict load coverage; ambiguous mixed sources fail closed.
- Keep host storage independent from transfer protocol:
  - AllGather mode copies only the rank-local persistent shard, then releases mmap handles.
  - no-AllGather mode retains file-backed checkpoint views and packs one block at a time through two bounded pinned host staging slots. DP workers on the same node therefore share checkpoint pages through the OS page cache rather than retaining private full model copies.
  - An effective DLO group size of one uses the rank-local path and performs no collective, even if `dlo_use_allgather=True`.
- Put MiniMax-H3 grouped-QKV and Cosmos3 TP=1 checkpoint semantics in loader-side direct-mmap adapters. No model-pipeline capability flag or parameter-attached transform is required.
- Treat TP=1 as the Phase A direct-mmap support boundary. TP > 1 falls back before model mutation to ordinary TP-aware loading. DLO may consume those TP-local runtime tensors, but no cross-DP shared-mmap host-memory saving is claimed.
- Fail closed to the ordinary loader for other unsupported direct layouts, including HSDP, online quantization, metadata mismatch, missing tensors, unknown custom loaders, or non-dedicated DiT sources. The existing DLO+AllGather/online-quantization rejection remains unchanged.

DP request admission, wave scheduling, and replica orchestration remain out of scope for vLLM-Omni DLO; those belong in a router such as vLLM Router or Dynamo. A normalized runtime-layout mmap cache is also deferred to a follow-up from #6195.

## Validation

- `ruff check` and `ruff format --check` on all changed Python files.
- Focused loader/offloader/MiniMax/runner suite: 97 tests passed. One unrelated order-sensitive `caplog` assertion failed only in the combined invocation and passed when run alone.
- Post-review loader/offloader suite: 67 tests passed after removing the unused Phase A layout key.
- Post-rebase compatibility suite against `main` after #5910: 272 tests passed with the same order-sensitive runner test deselected; that test passed separately. This includes source-ownership, mixed DiT/text-encoder loading, fail-closed mixed-source, and FP8 noncontiguous-stride coverage.
- MiniMax-H3 DP=2, TP=1, `dlo_use_allgather=False` two-GPU E2E on implementation commit `fa2f5b45` (carried unchanged into the current head):
  - both ranks indexed and realized 535/535 plan-owned DiT tensors;
  - the encoder-owning rank ordinary-loaded the one `text_encoder.` source outside the plan;
  - ordinary DiT weight materialization was skipped;
  - two 1.2313 GiB bounded host staging slots were allocated per worker;
  - generated video `[107, 256, 256, 3]` and audio `[1, 2, 142400]`;
  - peak device memory was 13,226 MiB;
  - clean shutdown with zero active child processes.

### Four-GPU B300 topology smoke

MiniMax-H3 FL2VA was run with the same prompt, seed, CUDNN attention backend, 256x256 output, two denoising steps, and `dlo_resident_layers=0`:

| DiT configuration | Result | Warm E2E | Peak device memory | Host PSS |
|---|---|---:|---:|---:|
| DP=4, TP=1, AllGather | Passed, 4 concurrent requests | 2.87 s / 4 requests | 13.84 GiB | 211.99 GiB |
| DP=4, TP=1, no-AllGather | Passed, 1 request | 15.02 s | 13.23 GiB | 187.77 GiB |
| DP=2, TP=2, AllGather | Passed, 2 concurrent requests | 4.16 s / 2 requests | 12.50 GiB | 211.97 GiB |
| DP=2, TP=2, no-AllGather | Passed, 1 request | 3.51 s | 11.88 GiB | 314.01 GiB |

Within each topology, AllGather and no-AllGather produced byte-identical video and audio. All four runs logged no `ERROR` or traceback, and their device allocations were released after completion.

- In the TP=1 topology, no-AllGather direct mmap reduced total PSS by 24.22 GiB (11.4%) and `Private_Dirty` from 211.33 to 125.32 GiB (40.7%) relative to AllGather.
- In the TP=2 topology, preflight selected the ordinary loader as designed. no-AllGather PSS was 314.01 GiB, about 48% above AllGather, because DP replicas retained private TP-local runtime storage.
- The TP=2 rows used DiT DP=2, TP=2 with the text encoder and VAEs at TP=1. They validate the ordinary-loader fallback, not full-pipeline TP=2 or shared checkpoint mmap.

This matrix is a functional and memory smoke test, not a production performance or output-quality benchmark.

### Host-memory comparison

On one L20X node, the pre-change ordinary-loader path at `e57574ae` was compared with direct mmap using the same MiniMax-H3 FL2VA DP=2, TP=1, no-AllGather, BF16, two-step, 256x256 workload. Ordinary-loader workers were sampled after initialization; mmap workers were sampled after one completed request so the checkpoint working set was resident. Values are whole-worker `/proc/<pid>/smaps_rollup` measurements.

| Worker | Ordinary RSS | mmap RSS | Ordinary PSS | mmap PSS | PSS reduction |
| DP worker 0 | 168.27 GiB | 132.76 GiB | 167.84 GiB | 101.43 GiB | 66.40 GiB |
| DP worker 1 | 116.19 GiB | 79.97 GiB | 115.73 GiB | 48.64 GiB | 67.09 GiB |
| **Two-worker total** | — | — | **283.56 GiB** | **150.08 GiB** | **133.48 GiB (47.1%)** |

Each mmap worker reported 62.45 GiB `Shared_Clean` but only 31.20 GiB `Pss_File`, consistent with the same resident file pages being proportionally charged across two mappings. `Private_Dirty` fell by 97.29 GiB and 97.96 GiB for workers 0 and 1. RSS counts a shared page in every mapping, so summed PSS is the physical-memory comparison.

- Documentation follow-up: repository pre-commit hooks passed for both edited Markdown files, and `mkdocs build --strict` completed successfully.

Refs #6195
Supersedes #6198

### DIFF
--- a/docs/design/feature/offloader/distributed_layerwise_offload.md
+++ b/docs/design/feature/offloader/distributed_layerwise_offload.md
@@ -24,6 +24,11 @@ the same node can share immutable checkpoint pages when direct mmap is
 selected, while the ordinary-loader fallback still keeps a private runtime
 copy per process.
 
+The Phase A shared-mmap support boundary is TP1. TP greater than one is an
+ordinary-loader compatibility path: DLO can consume the resulting TP-local
+tensors, but those configurations do not use checkpoint mmap and must not be
+used to claim shared-mmap host-memory savings.
+
 The compatibility matrix below describes the current implementation. The
 unit-level guards are covered, but not every parallelism combination has a
 full model-and-hardware end-to-end test.
@@ -146,7 +151,7 @@ This mode means:
 |---|---|---|
 | **DP** | Supported primary path. DLO shards host weights across the DP group and can run DP multi-concurrency. | Supported rank-local path. Compatible TP1 replicas can share checkpoint pages on each node; fallback runtime tensors remain private. |
 | **SP** | Supported in the implementation. With DP=1, DLO uses the SP group for host-weight sharding; SP still shards sequence/activation work. | SP remains active, but DLO keeps standard-loader rank-local weights and adds no SP weight collective. |
-| **TP > 1** | Direct checkpoint mmap falls back before mutation. The ordinary loader preserves TP-local layouts before DLO applies any DP/SP host sharding. | Direct checkpoint mmap falls back to ordinary TP-aware loading; the resulting TP-local tensors are streamed without a DLO collective. Broader model and hardware validation is still needed. |
+| **TP > 1** | Outside the Phase A shared-mmap support scope. The loader falls back before mutation, preserves TP-local layouts, and DLO may apply DP/SP host sharding to those ordinary runtime tensors. | Outside the Phase A shared-mmap support scope. The ordinary TP-aware loader produces rank-local tensors, which DLO streams without an additional weight collective; DP replicas retain private runtime storage. |
 | **HSDP** | Rejected. HSDP has already sharded parameters, so DLO AllGather would double-shard them. | Accepted by configuration. HSDP owns parameter sharding and its own gathers; DLO only stages rank-local parameters. End-to-end coverage is limited. |
 
 ### Combined dimensions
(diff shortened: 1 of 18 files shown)


vllm_omni/diffusion/model_loader/host_weight_plan.py
```diff
@@ -0,0 +1,333 @@
...
+
+        _validate_source_metadata(required, bindings)
+    except (OSError, ValueError, _PlanIncompatibleError) as exc:
+        return HostWeightPlanResult(None, str(exc))
+
+    return HostWeightPlanResult(
+        HostWeightPlan(
+            backing_kind="checkpoint_mmap",
+            bindings=bindings,
+            runtime_layout_key="checkpoint-mmap:tp=1:hsdp=0:online-quant=0",
```
hsliuustc0106: Nothing reads `runtime_layout_key` yet — only the test fixtures set it. Since the normalized runtime-layout cache is deferred to the #6195 follow-up, either log the key in the loader's plan-active message so fallbacks are debuggable today, or drop it and reintroduce it alongside its consumer.

docs/design/feature/offloader/distributed_layerwise_offload.md
```diff
@@ -61,16 +88,36 @@ Buffers:    [current slot]       [prefetch slot]       [current slot]
...
 With `--dlo-no-use-allgather`, DLO forces its internal offload shard size to
-one. The regular model loader remains responsible for preparing each rank's
-weights, including TP-local tensors or HSDP-managed parameters. DLO then
-streams those rank-local tensors block by block using H2D copies only.
+one and streams complete blocks using H2D copies only. The host backing may be
+either a loader-approved checkpoint mapping or ordinary runtime tensors.
+
+For direct mmap, each process retains immutable safetensors views and uses two
+bounded pinned host staging slots. Processes on the same node that map the same
+files share physical checkpoint pages through the OS page cache. This removes
```
hsliuustc0106, line 125: The node-local page-cache sharing benefit is asserted but not measured — the validation covers the two bounded staging slots and skipped materialization, but no host-RSS numbers. A per-worker RSS before/after on the DP=2 no-AllGather config (here or in the #6195 follow-up) would put a measurement behind the headline memory claim.

examples/offline_inference/minimax_h3/dlo_dp2_lifecycle.py
hsliuustc0106, line 1: why we need this file? it only check dp2? why not include any dp size 1/2/4/8?

PR #6214, opened 2026-08-15
[Bugfix] Restore per-stage runtime env during launch

### LINKED ISSUE #6210: [Bug]: Per-stage runtime.env is parsed but no longer applied during stage launch
### Summary

Per-stage `runtime.env` configuration is still accepted and carried into each stage's runtime config, but on current `main` it does not appear to be applied when local LLM or diffusion stages are launched.

A static call-graph audit at `1260ceaae639437372797c710b61b4584d36b9ca` found that the stage-runtime refactor in #3855 removed the two launch scopes that called `stage_runtime_setup()` and replaced them with device-only scopes. The environment helper still exists, but has no production caller.

This appears to be a regression from the per-stage environment support introduced in #3235. This report is based on static inspection; a hardware reproduction has not yet been performed.

### Evidence

The configuration path is intact:

- `StageDeployConfig.env` is defined and `runtime.env` is parsed in [`stage_config.py`](https://github.com/vllm-project/vllm-omni/blob/1260ceaae639437372797c710b61b4584d36b9ca/vllm_omni/config/stage_config.py#L291-L296) and [`_parse_stage_deploy()`](https://github.com/vllm-project/vllm-omni/blob/1260ceaae639437372797c710b61b4584d36b9ca/vllm_omni/config/stage_config.py#L497-L505).
- The values are copied into `OmniStageRuntimeConfig.env` in [`omni_config.py`](https://github.com/vllm-project/vllm-omni/blob/1260ceaae639437372797c710b61b4584d36b9ca/vllm_omni/config/omni_config.py#L1429-L1448).
- [`stage_runtime_env()`](https://github.com/vllm-project/vllm-omni/blob/1260ceaae639437372797c710b61b4584d36b9ca/vllm_omni/engine/stage_init_utils.py#L725-L778) applies the values for a scoped block and restores the parent environment afterward.

However:

- Searching current production code finds definitions of `stage_runtime_setup()` and `stage_runtime_env()`, but no call sites.
- Before #3855, both the local LLM and local diffusion launch paths in `async_omni_engine.py` entered `stage_runtime_setup()`.
- After #3855, `StageRuntime._stage_device_scope()` [only applies the stage device environment](https://github.com/vllm-project/vllm-omni/blob/1260ceaae639437372797c710b61b4584d36b9ca/vllm_omni/engine/stage_runtime.py#L326-L331). The local LLM spawn path uses the device spawn scope, and the local diffusion path uses this device-only scope; neither composes it with `runtime.env`.
- The shipped [GLM-Image NPU deployment](https://github.com/vllm-project/vllm-omni/blob/1260ceaae639437372797c710b61b4584d36b9ca/vllm_omni/deploy/glm_image.yaml#L50-L79) still relies on different per-stage `HCCL_IF_BASE_PORT`, `HCCL_HOST_SOCKET_PORT_RANGE`, and `HCCL_NPU_SOCKET_PORT_RANGE` values.

### Expected behavior

For every stage/replica launch, that stage's `runtime.env` should:

1. Override the parent environment during stage initialization and child-process creation.
2. Be isolated from other stages and replicas.
3. Restore the parent environment, including previously unset keys, after the scoped launch.
4. Compose safely with device visibility scoping and launch serialization.

This matches the helper's current contract and the behavior originally added in #3235.

### Suggested fix

Restore an environment scope around both local launch paths, using `stage_runtime_env()` or `stage_runtime_setup()` as appropriate:

- local LLM child-process spawn;
- local/in-process diffusion stage initialization.

The scope should be composed explicitly with the device scope and launch lock. Remote/headless behavior should also be defined so that users know whether `runtime.env` is applied by the controller or must be provided independently on remote workers.

### Suggested tests

- A configured key is visible in the spawned stage process.
- Different values remain isolated across stages and replicas.
- Existing parent values are restored.
- Previously unset keys are removed after the scope.
- `runtime.devices` and `runtime.env` compose correctly.
- The GLM-Image NPU HCCL port configuration has a regression test or targeted launch test.
- Logs report keys only and never environment values, since values may contain secrets.

### Related issues and PRs

- #3235 — original per-stage runtime environment support
- #3855 — stage-runtime refactor where the launch scopes disappeared
- #4021 — unified `VllmOmniConfig` / per-stage runtime configuration
- #5386 — configuration and entrypoint alignment roadmap

m0g3r:
## Purpose

Per-stage `runtime.env` remains parsed into each stage's runtime config, but local LLM and diffusion launches no longer apply it. The observable result is that launch code sees the parent value (or an unset key) instead of the configured stage value.

Fixes #6210.

## Reproduction

Run the focused regression tests:

```bash
pytest -q tests/engine/test_async_omni_engine_stage_init.py \
  -k 'initialize_local_diffusion_replica_scopes_runtime_env or initialize_local_llm_replica_scopes_runtime_env'
```

Against the parent implementation, the LLM launch hook sees the parent value and the diffusion launch hook sees an unset key. Both should see the configured stage value.

## Root cause

The stage-runtime refactor replaced the old `stage_runtime_setup()` launch scopes with device-only scopes. `stage_runtime_env()` still handled conversion, key-only logging, and restoration, but had no production caller.

## Fix

- Enter `stage_runtime_env()` inside the existing serialized local LLM launch section.
- Compose `stage_runtime_env()` outside the local diffusion device scope, so `runtime.devices` remains the effective device selection during launch.
- Add L1 CPU regression coverage for value visibility, existing-parent restoration, previously-unset cleanup, and device-scope composition.

Remote replicas are unchanged: the controller only attaches to those processes and does not apply a local environment scope. Headless worker environments remain owned by their launcher/process environment.

## Test Plan

```bash
pytest -q tests/engine/test_async_omni_engine_stage_init.py \
  -k 'initialize_local_diffusion_replica_scopes_runtime_env or initialize_local_llm_replica_scopes_runtime_env'

uvx --from pre-commit==4.0.1 pre-commit run --files \
  vllm_omni/engine/stage_runtime.py \
  tests/engine/test_async_omni_engine_stage_init.py
```

**vLLM Version:** v0.27.0 target

**vLLM-Omni Commit:** `136cb27b05385cf2ce7246e6815154ee86525d0f`

## Test Result

- Full file-scoped pre-commit suite: passed.
- Python 3.12 compile check for both changed files: passed.
- Dependency-isolated probe of the actual `StageRuntime` class: reproduced parent/unset values on the base revision and observed stage values plus restoration on this branch.
- Focused pytest was not runnable on the local macOS arm64 host because vLLM v0.27.0 has no usable wheel there; Linux CI is expected to run the committed L1 CPU regressions.

AI assistance: Codex was used for implementation, regression-test design, and the repository `precheck-pr` / `vllm-omni-test` workflows.

### DIFF
--- a/tests/engine/test_async_omni_engine_stage_init.py
+++ b/tests/engine/test_async_omni_engine_stage_init.py
@@ -161,6 +161,17 @@ def _make_diffusion_plan(
     )
 
 
+def _make_stage_runtime() -> StageRuntime:
+    return StageRuntime(
+        stage_configs=[],
+        model="dummy-model",
+        config_path="dummy-config",
+        stage_init_timeout=1,
+        diffusion_batch_size=1,
+        async_chunk=False,
+    )
+
+
 def test_stage_engine_core_client_module_reload_keeps_forward_refs_deferred():
     """Regression test for forward references in make_async_mp_client."""
     import vllm_omni.engine.stage_engine_core_client as client_mod
@@ -231,6 +242,48 @@ def test_collect_initialized_clients_for_cleanup_deduplicates_clients():
     )
 
     assert cleanup_clients == [shared, extra]
+
+
+def test_initialize_local_diffusion_replica_scopes_runtime_env(monkeypatch):
+    import vllm_omni.engine.stage_runtime as runtime_mod
+    from vllm_omni.platforms import current_omni_platform
+
+    runtime = _make_stage_runtime()
+    plan = _make_diffusion_plan(0, stage_id=0).replicas[0]
+
+    runtime_env_var = "VLLM_OMNI_TEST_STAGE_RUNTIME_ENV"
+    device_env_var = current_omni_platform.device_control_env_var
+    runtime._init_visible_devices_baseline = "0,1"
+    plan.metadata.runtime_cfg = {
+        "devices": "0",
+        "env": {runtime_env_var: "stage-value"},
+    }
+    monkeypatch.delenv(runtime_env_var, raising=False)
+    monkeypatch.setenv(device_env_var, "0,1")
+
+    captured: dict[str, str | None] = {}
+    monkeypatch.setattr(runtime_mod, "inject_kv_stage_info", lambda *_: None)
+
+    def _capture_launch_diffusion_stage_replica(**_kwargs):
+        captured["runtime_env"] = os.environ.get(runtime_env_var)
+        captured["device_env"] = os.environ.get(device_env_var)
+        raise RuntimeError("stop after capturing launch environment")
+
+    monkeypatch.setattr(
+        runtime_mod,
+        "launch_diffusion_stage_replica",
+        _capture_launch_diffusion_stage_replica,
+    )
+
+    with pytest.raises(RuntimeError, match="stop after capturing launch environment"):
+        runtime._initialize_local_diffusion_replica(plan, stage_init_timeout=1)
+
+    assert captured == {
+        "runtime_env": "stage-value",
+        "device_env": "0",
+    }
+    assert runtime_env_var not in os.environ
+    assert os.environ[device_env_var] == "0,1"
 
 
 def test_initialize_local_diffusion_replica_restores_device_visibility_after_local_init(monkeypatch):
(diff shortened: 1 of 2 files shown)


m0g3r:
Self-review: I traced both local stage-launch paths and verified that stage runtime variables are visible only during child construction, while the parent environment is restored afterward, including removal of keys that were previously unset. I also checked that diffusion launch composes the stage environment with the existing device scope and that the launch lock still covers LLM process creation. The added CPU tests cover stage visibility, restoration, unset-key cleanup, and device composition. I ran compileall, Ruff check/format, and the full file-scoped pre-commit suite. I could not run the pytest file on macOS arm64 because vLLM 0.27 has no usable wheel for this platform; that limitation is documented in the PR body.

Gaohan123 (approved):
LGTM. Thanks

PR #6217, opened 2026-08-15
[Doc] Add environment variables configuration reference

### LINKED ISSUE #6211: [Doc]: Add an environment variables configuration reference
### Documentation problem

vLLM-Omni has no centralized reference for its environment variables. Environment-driven behavior is currently scattered across stage configuration, serving, storage, streaming, diffusion backends, model implementations, benchmark scripts, inherited vLLM behavior, and platform/launcher integrations.

This makes it difficult to answer basic operational questions:

- Which variables are public and supported?
- Which entrypoints, models, stages, or processes read them?
- What are their types, defaults, and accepted values?
- Do CLI arguments, deploy YAML, request fields, or environment variables take precedence?
- Are values read at import time, server startup, worker spawn, or per request?
- What happens when a value is invalid?
- Which variables are inherited from vLLM, deprecated, transitional, or internal?

### Audit snapshot

A static audit of `main` at `1260ceaae639437372797c710b61b4584d36b9ca` found:

- 133 statically identifiable runtime-related environment names in `vllm_omni/**/*.py` (113 direct literals and 20 indirect/constructed names).
- Only 39 of those names appear anywhere under `docs/`, often incidentally in examples rather than as reference documentation.
- There is no vLLM-Omni-wide environment registry comparable to vLLM's `envs.py`.
- The existing small registries/settings cover only parts of the surface: diffusion, video streaming, and server storage settings.
- The inventory mixes roughly 50 model-specific controls, 18 benchmark variables, inherited platform/upstream variables, and internal process-coordination variables. A raw generated dump would therefore expose implementation details and imply support guarantees that do not exist.
- Environment-related code is changing actively, so a hand-maintained page without a source-of-truth or drift check is likely to become stale.

The audit also found inconsistent precedence and evaluation behavior across subsystems: some environment values are fallbacks behind explicit config, some override config, some settings are environment-only, and values are read at different lifecycle points.

### Proposed placement

Environment variables are configuration shared by the CLI, deploy YAML, Python APIs, workers, and model implementations. They should therefore live under **User Guide → Configuration**, not as a subsection of CLI Reference.

```text
User Guide
└── Configuration
    ├── Configuration overview
    └── Environment Variables

CLI Reference
└── Configuration and precedence
    └── Link to Environment Variables
```

CLI Reference should remain a top-level documentation section, as proposed in #6142, and cross-link this page where command/configuration precedence is discussed.

### Proposed page structure

1. **Scope and version**
   - What vLLM-Omni owns versus what it inherits.
   - Whether the page describes `latest` or a stable release.
2. **Precedence and evaluation time**
   - Request/input fields, CLI args, deploy YAML, environment, and defaults.
   - Import-time, startup-time, worker-spawn-time, and request-time reads.
3. **Core runtime and process launch**
4. **Serving, storage, and streaming**
5. **Diffusion, attention, cache, and performance controls**
6. **Model-specific variables**
   - Prefer a compact index linking to model pages when variables are implementation-specific.
7. **Inherited vLLM variables**
   - Link the version-matched upstream vLLM reference.
   - Document only Omni-specific exceptions or unsupported behavior locally.
8. **Platform and external-tool variables**
   - CUDA/NPU/ROCm, Hugging Face, ModelScope, distributed launchers, etc.
9. **Deprecated and transitional variables**
10. **Per-stage `runtime.env`**
    - Contract, scoping, restoration, and safe handling.
    - Do not claim working behavior until #6210 is fixed or the current behavior is otherwise verified.
11. **Troubleshooting and environment collection**

### Reference entry schema

Each public variable should document, where applicable:

| Field | Meaning |
|---|---|
| Name | Exact environment variable name |
| Owner/scope | Omni, upstream vLLM, platform, model, benchmark, or internal |
| Type and accepted values | Including boolean parsing rules |
| Default | Effective default, not only the source literal |
| Read time | Import, startup, worker spawn, or per request |
| Precedence | Relationship to CLI, YAML, request fields, and explicit config |
| Applies to | Entrypoints, models, stages, and processes |
| Invalid-value behavior | Error, warning, fallback, or ignore |
| Lifecycle | Stable, experimental, deprecated, transitional, or internal |
| Security | Whether it may contain a secret and must be redacted |

### Source-of-truth and maintenance approach

Do not generate the public page directly from every `os.environ` access. First classify each discovered name as:

- public Omni configuration;
- inherited/external;
- model-specific;
- transitional/deprecated;
- internal implementation detail.

For the public set, use a small explicit metadata registry or equivalent allowlist as the source of truth. The documentation generator and environment collector can consume that metadata. CI should detect new direct environment reads that have not been classified, while allowing internal variables to remain undocumented.

The maintenance design should also:

- link to the vLLM documentation version compatible with the current vLLM-Omni release;
- record intentional Omni precedence differences;
- redact values that may contain secrets;
- avoid promoting debug/model escape hatches into stable API accidentally;
- coordinate benchmark-variable removal with #5376;
- extend `collect_env.py` to report the safe Omni-owned subset rather than relying only on the upstream vLLM registry.

This source-of-truth work may be coordinated with the broader configuration alignment in #5386; the first documentation version does not need to wait for a full configuration refactor if it uses a reviewed allowlist.

### Acceptance criteria

- [ ] An **Environment Variables** page is added under **User Guide → Configuration**.
- [ ] CLI Reference remains top-level and links to the new page from its configuration/precedence material (#6142).
- [ ] The complete public Omni-owned set is reviewed and documented; internal names are explicitly excluded rather than silently omitted.
- [ ] Each public entry covers type, default, scope, read time, precedence, lifecycle, and invalid-value behavior where applicable.
- [ ] Inherited vLLM variables link to the matching upstream version, with Omni-specific exceptions documented locally.
- [ ] Deprecated aliases and transitional benchmark variables are identified.
- [ ] Per-stage `runtime.env` behavior is documented only after #6210 is resolved or is clearly marked as a known limitation.
- [ ] Environment collection and examples redact secrets and log keys rather than values.
- [ ] CI detects unclassified new environment reads and/or drift in generated reference content.
- [ ] `mkdocs build` succeeds and all links are valid.

### Related issues

- #6210 — per-stage `runtime.env` is parsed but appears not to be applied
- #5386 — configuration/entrypoint alignment roadmap
- #6142 — CLI reference reorganization
- #5925 — `VLLM_TORCH_PROFILER_DIR` behavior diverges on the Omni diffusion path
- #5376 — remove benchmark configuration through environment variables
- #4021 — unified configuration and per-stage runtime settings
- #2887 — device visibility environment handling

hsliuustc0106:
## Summary

- add a central Environment Variables page under User Guide → Configuration and cross-link it from the top-level CLI reference
- classify 148 effective environment names by ownership, including explicit migration dispositions for all 57 model-specific switches
- document the 22 reviewed public Omni settings with defaults, precedence, read time, invalid-value behavior, lifecycle, and per-stage `env` scoping
- distinguish the classification-only inventory from an executable `vllm.envs`-style resolver and require the `VLLM_OMNI_` prefix for future public names
- extend `collect_env.py` with the safe public Omni allowlist and keep diagnostics usable when Omni import fails
- add bidirectional CI drift checks for direct environment reads regardless of key casing and stale model/benchmark inventory rows, plus public-reference coverage, classification counts, collector redaction, and future prefix exceptions
- track family-sized model migration packages and their ownership gate in #6232

## Why

vLLM-Omni currently reads environment variables across process startup, configuration, workers, model integrations, and benchmarks. Treating every discovered name as public would accidentally turn model-specific debug and compatibility switches into supported API.

This PR establishes a reviewed public boundary, records ownership for the remaining names, and gives model owners explicit migration targets without changing existing runtime behavior.

## Upstream alignment

The page follows vLLM's placement under User Guide → Configuration, reuses the reviewed inventory in environment collection, and links inherited behavior to the aligned vLLM 0.26 reference.

This PR intentionally stops at audit and governance metadata: `ENVIRONMENT_VARIABLE_INVENTORY` classifies names but does not resolve values. Existing consumers continue to own parsing, defaults, precedence, and evaluation time. Converging the 22 public Omni settings on a typed/lazy resolver is follow-up work rather than expanding this audit across all 148 variables.

Existing public names with older prefixes are grandfathered for compatibility. A regression test requires every future public Omni-owned name to use `VLLM_OMNI_`.

## Model-specific boundary

The existing model-only environment switches are classified as migration inputs, not promoted into public API:

- Promote: 33
- Request scope: 6
- External: 0
- Internalize: 11
- Deprecate/remove: 7

The exact decisions live in `vllm_omni/config/environment_variable_inventory.py`. Implementation is tracked in #6232 as family-sized packages that require a named assignee and model-owner review.

## Validation

- `uvx pre-commit run --files ...`
- 190 focused environment/configuration tests passed
- `mkdocs build --strict` (exit 0)

Fixes #6211


collect_env.py
```diff
@@ -494,6 +494,13 @@ def is_xnnpack_available():
 
 
 def get_env_vars():
+    try:
+        from vllm_omni.config.environment_variables import ENVIRONMENT_VARIABLES
+    except ImportError:
```
hsliuustc0106: The failure modes that motivate this fallback aren't all `ImportError`s. Importing `vllm_omni.config.environment_variables` executes `vllm_omni/__init__.py`, which imports `vllm_omni.patch` — a bare `import torch` (`patch.py:7`) raises `OSError` on a broken CUDA install, and the patch logic can also raise `AssertionError` (`patch.py:73`) or `RuntimeError` (`patch.py:249`); see the note in `vllm_omni/__init__.py` that patch may throw on vLLM version mismatch. None of those are `ImportError` subclasses, so `get_env_vars()` crashes on exactly the broken installs `collect_env.py` exists to diagnose — a regression vs. main, which never imported vllm_omni here. Since the allowlist is optional, suggest `except Exception` here (the `(ImportError, NameError, AttributeError, OSError)` tuple used for torch at line 26 would still miss the patch `RuntimeError`/`AssertionError`).

tests/config/test_environment_variables.py
```diff
@@ -0,0 +1,139 @@
...
+)
+
+pytestmark = [pytest.mark.core_model, pytest.mark.cpu]
+
+_REPO_ROOT = Path(__file__).parents[2]
+_PACKAGE_ROOT = _REPO_ROOT / "vllm_omni"
+_REFERENCE_PAGE = _REPO_ROOT / "docs" / "configuration" / "environment_variables.md"
+
+
+def _literal_environment_accesses(path: Path) -> set[str]:
```
hsliuustc0106: This gate only sees `os.getenv("LITERAL")`-shaped accesses, but the package already reads ~24 sites through shapes it cannot see: module-constant indirection (`os.environ.get(AUDIO_SAMPLE_RATE_ENV)` at `vllm_omni/metrics/definitions.py:367`), wrapper helpers (`_choice_env`/`is_set` in `entrypoints/openai/video_stream_envs.py` — both public video variables are read only this way), the pydantic-settings `VLLM_OMNI_SERVER_*` family (no `os.` call at all), and `import os as _os` (`modeling_moss_tts_local_depth.py:221`). A new variable added in any of these styles lands with no CI signal — the exact silent drift this inventory exists to prevent. Could the scan additionally resolve module-level all-caps string constants passed as the first argument, and match `"X" in os.environ` membership literals? If that is more machinery than wanted, the literal-only scope should at least be stated in the policy sentence at the end of the reference page so readers don't assume full coverage.

tests/config/test_environment_variables.py
```diff
@@ -0,0 +1,296 @@
...
+    public_names = {
+        name
+        for name, item in ENVIRONMENT_VARIABLE_INVENTORY.items()
+        if item.category is EnvironmentVariableCategory.PUBLIC_OMNI
+    }
+
+    assert {name for name in public_names if not name.startswith("VLLM_OMNI_")} == legacy_public_names
+
+
+def test_statically_resolvable_environment_accesses_are_classified():
```
hsliuustc0106, line 213: This direction (code → inventory) is now solid, but the reverse has no guard: when a model is deleted or refactored away from its env switches, its `MODEL_SPECIFIC`/`BENCHMARK_TRANSITIONAL` rows linger with no CI signal, and the pinned counts in `test_inventory_matches_reviewed_snapshot_counts` actively discourage cleaning them up. Since several of those names are read through indirect sites, the robust form is a substring liveness check rather than the scanner: assert each name in those two categories still appears textually somewhere under `vllm_omni/`. That makes model removal update the inventory automatically.

docs/configuration/environment_variables.md
```diff
@@ -0,0 +1,234 @@
...
+for redaction in the reviewed inventory. Never print their values in environment
+collection, examples, bug reports, or logs.
+
+## Model-specific variables
+
+The audit found 54 variables read by a single model or pipeline family. They are
+not listed as public usage options here because doing so would turn implementation
+escape hatches into an accidental compatibility contract.
+
+Every audited model-specific name has a migration disposition in the
```
hsliuustc0106, line 184: The dispositions are unenforced promises as it stands — nothing forces a `promote` to ever execute, so the 54 can only shrink if each batch has an owner. Worth opening the per-family tracking issue(s) alongside this PR so they don't live only in this tuple: the 11 `internalize` rows need no work, the 5 `deprecate_remove` rows are trivial deletions, and the 38 promote/request-scope rows collapse into ~8 family-sized batches (DYNIN 7, INTERNVLA_A1 4, GEPARD 6, COSYVOICE3/MIMO/MINICPMO 3 each, COSMOS3 2, plus singles).

hsliuustc0106:
I think we also need to update the related skills if the users want to add a new env var, where they should follow our general rule

PR #6222, opened 2026-08-15
[Bugfix] Honor height/width overrides in LongCatImageEditPipeline

yzong-rh:
## Purpose

Fix `LongCatImageEditPipeline` so explicit `height`/`width` sampling params are honored for both conditioning preprocess and generation.

Previously, `LongCatImageEditPipeline` automatically calculated latent sizes based on the size and aspect-ratio of the conditioning image. Overrides were ignored (and tiny I2I tests asserting `512x512` would fail).

Also enable tiny-model coverage for `LongCatImageEditPipeline` (`IMAGE_TO_IMAGE`), including TeaCache / SP+CacheDiT / CFG+TP+offload groups.

## Test Plan

```
pytest tests/model_tests/diffusion/test_common_offline.py -sv -k test_pipeline_on_supported_tasks[LongCatImageEdit
```

https://gist.github.com/yzong-rh/ee60fb7a4e234adfabc14d6f9ec37167

```
python repro_longcat_image_edit_size.py 
```

**vLLM Version:**

v0.27.0

**vLLM-Omni Commit:**

8ecd1f6d5cc91aab8a475a861213720b336e2f65

## Test Result

```
4 passed, 21 deselected, 15 warnings, 6 subtests passed in 98.49s (0:01:38)
```

Before:
OG aspect ratio and size is preserved.
image

After:
Sampling Param aspect ratio is honored.
image

cc @alex-jw-brooks 

### DIFF
--- a/tests/model_tests/diffusion/diff_model_builders.py
+++ b/tests/model_tests/diffusion/diff_model_builders.py
@@ -125,6 +125,22 @@ def tiny_longcat_image_builder() -> str:
     )
 
 
+def tiny_longcat_image_edit_builder() -> str:
+    return build_tiny_from_configs(
+        "LongCatImageEditPipeline",
+        "meituan-longcat/LongCat-Image-Edit",
+        transform={
+            "text_encoder": _shrink_qwen_text_encoder_config,
+            "transformer": partial(
+                _shrink_dit_rope_config,
+                num_single_layers=2,
+                default_axes_dims_rope=[16, 56, 56],
+                joint_attention_dim=64,
+            ),
+        },
+    )
+
+
 def tiny_flux_builder() -> str:
     return build_tiny_from_configs(
         "FluxPipeline",
--- a/tests/model_tests/diffusion/model_settings.py
+++ b/tests/model_tests/diffusion/model_settings.py
@@ -64,6 +64,16 @@ DIFFUSION_TEST_SETTINGS = {
             [DiffusionAccs.CFG_PARALLEL, DiffusionAccs.TENSOR_PARALLEL, DiffusionAccs.CPU_OFFLOAD],
         ],
     ),
+    "LongCatImageEditPipeline": DiffusionModelTestOpts(
+        model="meituan-longcat/LongCat-Image-Edit",
+        builder=diff_model_builders.tiny_longcat_image_edit_builder,
+        supported_tasks=[DiffusionTasks.IMAGE_TO_IMAGE],
+        extra_test_groups=[
+            [DiffusionAccs.TEA_CACHE],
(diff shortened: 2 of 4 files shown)


yzong-rh:
Self-review from my agent

```
What I checked:
- Root cause: `get_longcat_image_edit_pre_process_func` did compute `height/width = sampling_params or calculate_dimensions(1MP, source AR)` and wrote those into `sampling_params`, but then resized the cond image with `calculated_*` and stashed those in `additional_information`. `forward()` always consumed `calculated_height/width` for `check_inputs` / `prepare_latents` / `_unpack_latents`, so explicit overrides never affected cond preprocess or generation. Tiny I2I (`512x512` input + `height=width=512`) still generated at ~1024×1024 because `calculate_dimensions` targets 1MP, and `assert img.size == (512, 512)` failed.
- Fix: after the existing override fallback, `normalize_min_aligned_size(..., vae_scale_factor * 2)` and use that `height/width` for resize (`height`, `height//2` for `prompt_image`), `sampling_params`, and generation. Drop the unused `calculated_*` stash. LongCat packs cond image latents with the same spatial size as noise latents, so cond and gen must share H/W; this is not the Qwen Edit split (`calculated_*` for cond `img_shapes`, `sampling_params` for gen).
- Scope: edit pipeline only. Default (no override) is unchanged: `calculate_dimensions` already 16-aligns, and `vae_scale_factor * 2` is 16, so `normalize_min_aligned_size` is a no-op on that path. T2I `LongCatImagePipeline` already honors `sampling_params` and is untouched.
- Tests: `tiny_longcat_image_edit_builder` mirrors T2I LongCat shrink (`_shrink_qwen_text_encoder_config` + 2-layer DiT rope). `DIFFUSION_TEST_SETTINGS` adds `IMAGE_TO_IMAGE` with the same extra groups as T2I (TeaCache / SP+CacheDiT+layerwise / CFG+TP+offload). Removing `LongCatImageEditPipeline` from `EXCLUDED_MODELS` keeps `test_alignment.py` in sync. Offline base also covers determinism + multi-output; online picks up the base case only (`online_base_only=True`).
- Gap: no isolated preprocess unit test for a non-square source plus an override (tiny I2I is square 512 in and 512 requested, which is enough to catch the 1MP override bug). TeaCache and SP remain in separate groups, so TeaCache+SP deadlock (#6180 / #6228) is still not covered here. Unaligned overrides (e.g. 500) now floor via `normalize_min_aligned_size` instead of being ignored.

Validation:
- `pytest tests/model_tests/diffusion/test_common_offline.py -sv -k test_pipeline_on_supported_tasks[LongCatImageEdit` → 4 passed, 21 deselected, 6 subtests in ~98s.
- Repro: sampling-param aspect ratio/size is honored; default source-AR 1MP path still works when height/width are unset.
```

NickCao (approved):

PR #6224, opened 2026-08-15
[diffusion][bugfix][ROCm] Keep Hunyuan Image 3.0 GroupNorm on PyTorch

### LINKED ISSUE #6223: [Bug]: AITER GroupNorm returns wrong Hunyuan Image 3.0 VAE output on ROCm
## Environment

The evidence comes from diagnosis run `20260815T142547Z`.

| Component | Version |
| --- | --- |
| GPU | Four MI300 devices, `gfx942` |
| PyTorch | `2.11.0+gitd0c8b1f` |
| PyTorch source | `d0c8b1f364ecacff4dd8bc06a645d0fb9324cd37` |
| HIP | `7.2.53211` |
| AITER | `0.1.16.post3` |
| vLLM-Omni | `3b2b63611e48b7fc9abae673570f2898ebc6db73` |

## Issue

On ROCm, vLLM-Omni replaces the PyTorch GroupNorm layers in the VAE with AITER GroupNorm layers.

The first Hunyuan Image 3.0 VAE GroupNorm gets an FP16 input and BF16 weight and bias while autocast is on. PyTorch runs GroupNorm in FP32 under autocast. AITER does not apply the PyTorch autocast rule.

Our probe stopped at the first GroupNorm call on all four workers. Each worker reported the same error:

```text
shape=(1, 128, 4, 1024, 1024)
input=torch.float16
weight=torch.bfloat16
bias=torch.bfloat16
autocast_enabled=True
expected=torch.float32
actual=torch.float16
mean_error=1.0777817964553833
max_error=6.17850923538208
```

All four workers reached the failing call through the same path:

```text
prepare_model_inputs
  -> _encode_cond_image
  -> vae_encode
  -> vae.encode
  -> encoder
  -> residual block norm1
```

The probe saved the input before it called AITER. It then called the AITER kernel, waited for the GPU work to finish, and compared the output with PyTorch. The AITER kernel returned normally during the probe. The value mismatch was the same on all four workers.

## Saved tensor proof

The saved weight and bias show how AITER gets the wrong result:

| Value | Correct BF16 value | Raw bits | Same bits read as FP16 |
| --- | ---: | ---: | ---: |
| `weight[0]` | `0.9765625` | `0x3f7a` | `1.869140625` |
| `bias[0]` | `-0.002166748046875` | `0xbb0e` | `-0.8818359375` |

The first output element had the following group values:

```text
mean=-0.17093949019908905
variance=0.7333720326423645
normalized_input=-0.140500009059906
```

Using the correct BF16 weight and bias in FP32 gives `-0.1393737942`. PyTorch returned `-0.1393791437`.

Reading the same weight and bias bits as FP16 gives `-1.1444501877`. AITER returned `-1.14453125`.

The AITER output matches the calculation that reads the BF16 parameter bits as FP16.

## Why it happens

PyTorch lists `group_norm` as an FP32 autocast operation. The rule is defined in the following files at the tested PyTorch commit:

* https://github.com/pytorch/pytorch/blob/d0c8b1f364ecacff4dd8bc06a645d0fb9324cd37/aten/src/ATen/autocast_mode.h
* https://github.com/pytorch/pytorch/blob/d0c8b1f364ecacff4dd8bc06a645d0fb9324cd37/aten/src/ATen/autocast_mode.cpp

The live probe also confirmed the rule because `torch.nn.functional.group_norm` returned FP32 for the failing call.

`aiter.ops.groupnorm.GroupNorm` calls the private AITER operator directly. AITER chooses the kernel type from the input type. With an FP16 input, it runs the FP16 kernel and reads the BF16 weight and bias pointers as FP16 pointers. AITER also creates the result from the input shape and type, so it returns FP16.

The AITER GroupNorm tests use one shared type for the input, weight, and bias. They do not test the Hunyuan mixed type call under autocast.

The AITER bug and fix are tracked here:

* https://github.com/ROCm/aiter/issues/4780
* https://github.com/ROCm/aiter/pull/4779

## Using Pytorch GroupNorm instead of  AITER GroupNorm

During diagnosis, we tested a version that used PyTorch GroupNorm for the failing Hunyuan call. Compared with the failing version, only `vllm_omni/platforms/rocm/patch/worker/patch_groupnorm.py` changed at runtime. 

The PyTorch version completed all 50 inference steps and saved the image. All four workers shut down normally and the command returned exit status zero.

The generated image passed the repository limits against `tests/assets/hunyuan/hunyuan_baseline.png`:

| Metric | Result | Required |
| --- | ---: | ---: |
| Mean absolute image error | `0.0149488030` | At most `0.03` |
| P99 absolute image error | `0.1450980306` | At most `0.3` |
| SSIM | `0.9758310318` | At least `0.97` |
| PSNR | `30.1148847899` | At least `30` |

The log from the PyTorch test did not import the AITER GroupNorm module. The failing probe imported the AITER GroupNorm module once per worker. The successful 50-step test shows that PyTorch GroupNorm avoids the Hunyuan regression, but it does not show that a future AITER fix works.

## Reproduction command

```bash
HIP_VISIBLE_DEVICES=0,1,2,3 \
VLLM_ROCM_USE_AITER=1 \
VLLM_LOGGING_LEVEL=INFO \
python examples/offline_inference/hunyuan_image3/end2end.py \
  --model tencent/HunyuanImage-3.0-Instruct \
  --modality text2img \
  --deploy-config vllm_omni/deploy/hunyuan_image3_dit.yaml \
  --prompts "A brown and white dog is running on the grass." \
  --steps 50 \
  --guidance-scale 2.5 \
  --seed 42 \
  --height 1024 \
  --width 1024 \
  --bot-task none \
  --sys-type en_unified \
  --init-timeout 900 \
  --enforce-eager
```

## Temporary vLLM-Omni fix

Keep PyTorch GroupNorm for `HunyuanImage3ForCausalMM` until vLLM-Omni uses an AITER release with the fix. Other VAE models will keep the current AITER GroupNorm path.

PR #6224 adds the model check. Its focused unit test passed locally and in the ROCm container. The full 50-step generation test has not been repeated on PR #6224.

## Limits of the diagnosis

The clean probe proves the numerical GroupNorm failure before any later worker failure. It does not prove that the GroupNorm kernel made an out of bounds memory access.

An earlier run reported GPU memory faults after the model received the wrong GroupNorm output. The GPU memory faults may come from later work, so they are not part of the root cause claim.

The later TCPStore, ProcessGroupNCCL, semaphore and worker shutdown errors happened after the probe stopped one worker. They are cleanup errors.

## Open Question

The GroupNorm patch uses `is_aiter_found_and_supported()`. The vLLM code says the helper checks whether AITER can run, but it does not check `VLLM_ROCM_USE_AITER`.

Therefore, `VLLM_ROCM_USE_AITER=0` does not disable AITER GroupNorm for other VAEs. Was the capability check chosen so other models always keep the AITER GroupNorm speedup when AITER is installed?

The temporary Hunyuan fix does not change the current selection behavior because a global change would affect every VAE model.

akshatvishu:
Fixes #6223.

### What broke

Hunyuan Image 3.0 gets wrong VAE output on ROCm when vLLM-Omni replaces its PyTorch GroupNorm layers with AITER GroupNorm layers.

The first VAE GroupNorm gets an FP16 input and BF16 weight and bias while autocast is on. PyTorch returns FP32 output for this call. AITER returns FP16 output with a mean error of `1.0778` and a maximum error of `6.1785` against PyTorch.

### How to reproduce

```bash
HIP_VISIBLE_DEVICES=0,1,2,3 \
VLLM_ROCM_USE_AITER=1 \
python examples/offline_inference/hunyuan_image3/end2end.py \
  --model tencent/HunyuanImage-3.0-Instruct \
  --modality text2img \
  --deploy-config vllm_omni/deploy/hunyuan_image3_dit.yaml \
  --prompts "A brown and white dog is running on the grass." \
  --steps 1 \
  --guidance-scale 2.5 \
  --seed 42 \
  --height 1024 \
  --width 1024 \
  --bot-task none \
  --sys-type en_unified \
  --init-timeout 900 \
  --enforce-eager
```

### Root cause

PyTorch runs GroupNorm in FP32 when autocast is on. The AITER GroupNorm module calls its custom operator directly, so it does not apply the PyTorch autocast rule.

AITER chooses the kernel type from the input type. With an FP16 input, it runs the FP16 kernel and reads the BF16 weight and bias as FP16. The saved Hunyuan tensors produced the same wrong values when we read the BF16 bits as FP16.

### Fix

Keep the original PyTorch GroupNorm layers for `HunyuanImage3ForCausalMM` until vLLM-Omni uses an AITER release with the fix.

Other VAE models keep the current AITER GroupNorm path. This PR does not change how `VLLM_ROCM_USE_AITER` controls those models.

## Test Plan

**vLLM Version:** `0.26.0+rocm723`

**vLLM-Omni Commit:** `077fc6895`

```bash
pytest -q tests/platforms/test_rocm_groupnorm_patch.py --run-level core_model

```

## Test Result

```text
Local: 1 passed

```

The full Hunyuan generation test was not repeated on this branch. Issue #6223 contains the full diagnosis and exact tensor comparison from the failing run.

### DIFF
--- a/tests/platforms/test_rocm_groupnorm_patch.py
+++ b/tests/platforms/test_rocm_groupnorm_patch.py
@@ -1,0 +1,33 @@
+# SPDX-License-Identifier: Apache-2.0
+
+import importlib.util
+import sys
+import types
+from pathlib import Path
+
+import pytest
+import torch.nn as nn
+
+pytestmark = [pytest.mark.core_model, pytest.mark.cpu]
+
+_PATCH_PATH = Path(__file__).parents[2] / "vllm_omni" / "platforms" / "rocm" / "patch" / "worker" / "patch_groupnorm.py"
+_SPEC = importlib.util.spec_from_file_location("test_patch_groupnorm", _PATCH_PATH)
+assert _SPEC is not None and _SPEC.loader is not None
+patch_groupnorm = importlib.util.module_from_spec(_SPEC)
+sys.modules[_SPEC.name] = patch_groupnorm
+_SPEC.loader.exec_module(patch_groupnorm)
+patch_groupnorm._registry_mod.initialize_model = patch_groupnorm._original_initialize_model
+
+
+def test_hunyuan_image3_keeps_pytorch_groupnorm(monkeypatch):
+    model = types.SimpleNamespace(vae=nn.Sequential(nn.GroupNorm(4, 8)))
+    monkeypatch.setattr(patch_groupnorm, "_original_initialize_model", lambda _: model)
+    monkeypatch.setattr(
+        patch_groupnorm,
+        "_replace_groupnorm_with_aiter",
+        lambda _: pytest.fail("Hunyuan Image 3.0 must keep PyTorch GroupNorm"),
+    )
+
+    od_config = types.SimpleNamespace(model_class_name="HunyuanImage3ForCausalMM")
+    assert patch_groupnorm._patched_initialize_model(od_config) is model
+    assert type(model.vae[0]) is nn.GroupNorm
--- a/vllm_omni/platforms/rocm/patch/worker/patch_groupnorm.py
+++ b/vllm_omni/platforms/rocm/patch/worker/patch_groupnorm.py
@@ -42,6 +42,13 @@ def _patched_initialize_model(od_config):
     model = _original_initialize_model(od_config)
 
     if hasattr(model, "vae"):
+        # AITER GroupNorm does not preserve the PyTorch autocast behavior required
+        # by Hunyuan Image 3.0. Keep PyTorch GroupNorm until the fix is released:
+        # https://github.com/ROCm/aiter/issues/4780
+        # https://github.com/ROCm/aiter/pull/4779
+        if od_config.model_class_name == "HunyuanImage3ForCausalMM":
+            return model
+
         try:
             from vllm._aiter_ops import is_aiter_found_and_supported
 


akshatvishu:
@vllm-omni-review-bot

Issue #6180, opened 2026-08-13
[Bug]: Diffusion Teacache + SP could lead to deadlock


yzong-rh:
### Your current environment

<details>
<summary>The output of <code>python collect_env.py</code></summary>

```text
==============================
        System Info
==============================
OS                           : Red Hat Enterprise Linux 9.6 (Plow) (x86_64)
GCC version                  : (GCC) 11.5.0 20240719 (Red Hat 11.5.0-5)
Clang version                : Could not collect
CMake version                : version 3.26.5
Libc version                 : glibc-2.34

==============================
       PyTorch Info
==============================
PyTorch version              : 2.13.0+cu130
Is debug build               : False
CUDA used to build PyTorch   : 13.0
ROCM used to build PyTorch   : N/A

==============================
      Python Environment
==============================
Python version               : 3.12.12 (main, Feb  3 2026, 22:51:04) [Clang 21.1.4 ] (64-bit runtime)
Python platform              : Linux-5.14.0-570.98.1.el9_6.x86_64-x86_64-with-glibc2.34

==============================
       CUDA / GPU Info
==============================
Is CUDA available            : True
CUDA runtime version         : 13.0.88
CUDA_MODULE_LOADING set to   : 
GPU models and configuration : 
GPU 0: NVIDIA H100 80GB HBM3
GPU 1: NVIDIA H100 80GB HBM3
GPU 2: NVIDIA H100 80GB HBM3
GPU 3: NVIDIA H100 80GB HBM3
GPU 4: NVIDIA H100 80GB HBM3
GPU 5: NVIDIA H100 80GB HBM3
GPU 6: NVIDIA H100 80GB HBM3
GPU 7: NVIDIA H100 80GB HBM3

Nvidia driver version        : 580.126.09
cuDNN version                : Could not collect
HIP runtime version          : N/A
MIOpen runtime version       : N/A
Is XNNPACK available         : False

==============================
          CPU Info
==============================
Architecture:                            x86_64
CPU op-mode(s):                          32-bit, 64-bit
Address sizes:                           52 bits physical, 57 bits virtual
Byte Order:                              Little Endian
CPU(s):                                  384
On-line CPU(s) list:                     0-383
Vendor ID:                               AuthenticAMD
Model name:                              AMD EPYC 9654 96-Core Processor
CPU family:                              25
Model:                                   17
Thread(s) per core:                      2
Core(s) per socket:                      96
Socket(s):                               2
Stepping:                                1
[...]

PR #6228, opened 2026-08-15
[Bugfix] Sync TeaCache skip decision across SP to prevent deadlock

yzong-rh:
## Purpose

Fix TeaCache + sequence parallelism deadlock https://github.com/vllm-project/vllm-omni/issues/6180. Repro in issue.

Average the L1 stats across the SP group before the threshold so every rank takes the same skip/compute path. TP is not included: the residual TeaCache reads is already replicated. CFG/DP ranks are also excluded.

## Test Plan

**vLLM Version:** 0.27.0

**vLLM-Omni Commit:** 9798bc3a8d5a400b97fba7402f42510577e7302d

Longcat repro script: https://gist.github.com/yzong-rh/b9deacd8178fc0a1295c6382c2f8b059

## Test Result

Repro no longer deadlocks.

LongCat with teacache but no SP:

teacache

```
=== Summary (measured runs) ===
Mean latency:         14.83s
Median latency:       14.84s
Stddev:               0.07s
Min latency:          14.73s
Max latency:          14.91s
```

LongCat with teacache and SP=2:

teacache_sp2

```
=== Summary (measured runs) ===
Mean latency:         10.62s
Median latency:       10.59s
Stddev:               0.06s
Min latency:          10.57s
Max latency:          10.75s
```

cc @alex-jw-brooks

### DIFF
--- a/vllm_omni/diffusion/cache/teacache/hook.py
+++ b/vllm_omni/diffusion/cache/teacache/hook.py
@@ -23,8 +23,31 @@ from vllm_omni.diffusion.cache.teacache.state import TeaCacheState
 from vllm_omni.diffusion.distributed.parallel_state import (
     get_classifier_free_guidance_rank,
     get_classifier_free_guidance_world_size,
+    get_sp_group,
+    model_parallel_is_initialized,
 )
 from vllm_omni.diffusion.hooks import HookRegistry, ModelHook, StateManager
+
+
+def _average_l1_stats_across_sp(mean_diff: torch.Tensor, mean_prev: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
+    """Average TeaCache L1 stats across sequence-parallel ranks.
+
+    Skip vs compute must be identical on every rank that participates in
+    transformer SP collectives. Sequence shards otherwise produce different
+    local means and deadlock. TP is not included: the residual TeaCache reads
+    is replicated across TP ranks. CFG/DP ranks are also excluded: they do
+    not share those collectives.
+    """
+    if not model_parallel_is_initialized():
+        return mean_diff, mean_prev
+
+    sp_group = get_sp_group()
+    if sp_group.world_size <= 1:
+        return mean_diff, mean_prev
+
+    stats = torch.stack((mean_diff.detach(), mean_prev.detach()))
+    stats = sp_group.all_reduce(stats) / sp_group.world_size
+    return stats[0], stats[1]
 
 
 class TeaCacheHook(ModelHook):
@@ -196,6 +219,7 @@ class TeaCacheHook(ModelHook):
         1. Always compute first timestep
         2. For intermediate steps:
            - Compute relative L1 distance between current and previous modulated inputs
+           - Average that distance across SP so all ranks share the skip decision
            - Apply polynomial rescaling with model-specific coefficients
            - Accumulate rescaled distances
            - Compare to threshold: below = cache, above = compute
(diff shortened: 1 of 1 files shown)


yzong-rh:
Self-review from my agent

```
What I checked:

- Root cause: TeaCache’s skip decision uses a local mean over sequence-sharded modulated_inp. Under SP, ranks can disagree on skip vs compute; the skip path skips transformer collectives that the compute path enters → deadlock (#6180).
- Fix: _average_l1_stats_across_sp all-reduces (mean_diff, mean_prev) over the SP group before the threshold / accumulator update, so every SP rank shares the same L1 → same skip/compute path and synced accumulated_rel_l1_distance.
- Scope: SP only via get_sp_group(). TP excluded (TeaCache residual / modulated input already replicated across TP). CFG/DP excluded (they don’t share those transformer SP collectives; each CFG/DP rank has its own SP subgroup).
- Placement is correct in the generic hook (_should_compute_full_transformer), so all TeaCache models get the sync, not just LongCat.
- Gap: no automated TeaCache+SP coverage. Tiny-model accs still keep TEA_CACHE and SEQUENCE_PARALLEL in separate groups, so CI wouldn’t catch this. Also, unweighted average of per-rank means is exact only when shard lengths match (usual with SP padding); enough to prevent deadlock either way.

Validation:
- Original Repro no longer deadlocks.
- LongCat TeaCache, no SP: ~14.8s mean, images OK.
- LongCat TeaCache + SP=2: no longer deadlocks; ~10.6s mean (~1.4× vs TeaCache-only).
```

Happy to add Teacache + SP tests to the testing framework on a model by model basis.

alex-jw-brooks:
Thanks @yzong-rh - to clarify, is the deadlock in the repro script expected to be deterministic? I was able to reproduce the bad behavior with Flux with the split outputs that you reported, but not the deadlock on main

yzong-rh:
Hey Alex, it's non deterministic in the sense that it requires one CFG branch to have a small relative L1 (cache hit) on local data while another gets a cache miss, but I set the seed so this reproduced consistently on H100s.

However, v0.28.0 may have changed the numerics slightly and seed 1000 no longer reproduces. I was able to get consistent deadlock with 1001 though.

```
Name: vllm
Version: 0.28.0

Name: vllm-omni
Version: 0.28.0rc2.dev23+gebb3b2057

Edit repro script:
SEED = 1001

Pass:
python repro_teacache_sp.py --model longcat  --cache tea_cache --ulysses-degree 1

Deadlocks:
python repro_teacache_sp.py --model longcat  --cache tea_cache --ulysses-degree 2
```

Deadlock Log:
```
INFO 08-28 16:24:17 [stage_diffusion_client.py:353] [StageDiffusionClient] stage-0 [rep-0] add request: 0_23f95ded-2c82-4b17-a4fe-7c4043fad8de
Processed prompts:   0%|                              | 0/1 [00:00<?, ?it/s]/home/yzong-rh/vllm-omni/.venv/lib/python3.12/site-packages/torch/distributed/c10d_logger.py:83: FutureWarning: `torch.distributed.all_gather_into_tensor` is deprecated. Please use `torch.distributed.all_gather_single` instead.
  return func(*args, **kwargs)
/home/yzong-rh/vllm-omni/.venv/lib/python3.12/site-packages/torch/distributed/c10d_logger.py:83: FutureWarning: `torch.distributed.all_gather_into_tensor` is deprecated. Please use `torch.distributed.all_gather_single` instead.
  return func(*args, **kwargs)
```

yzong-rh:
No open blocker

Gaohan123 (approved):
LGTM. Thanks

Issue #5728, opened 2026-08-03
[Bug]: CLI stage-overrides don't validate and reject invalid knobs


yzong-rh:
### Your current environment

<details>
<summary>The output of <code>python collect_env.py</code></summary>

```text
Your output of `python collect_env.py` here
```

</details>

### Your code version

<details>
<summary>The commit id or version of vllm</summary>

```text
v0.25.0
```
</details>
<details>
<summary>The commit id or version of vllm-omni</summary>

```text
7762d664d8550bd21cd71f092c56b53e679d5a74
```
</details>

### 🐛 Describe the bug

`--stage-overrides` accept many values that would be dumped into engine args and ignored.

**Pipeline-wide:** `session_mode`, `connectors`, `edges`, `stages`, `platforms`, `pipeline`, `duplex_session` (nested `duplex_session.*` have no separate CLI key).

**Per-stage:** `env`, `input_connectors`, `output_connectors`.

`default_sampling_params` is also ignored for multi-stage deployments (does not overlay the deploy field), but it is read for the single-stage diffusion builder.

https://github.com/vllm-project/vllm-omni/pull/3900 is related, addresses the `env` override specifically.

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://vllm-omni.readthedocs.io), which can answer lots of frequently asked questions.

hsliuustc0106:
@lishunyang12 PTAL

MciG-ggg:
> [@lishunyang12](https://github.com/lishunyang12) PTAL

I’d like to work on this issue. Could you assign it to me?

Gaohan123:
> > [@lishunyang12](https://github.com/lishunyang12) PTAL
> 
> I’d like to work on this issue. Could you assign it to me?

@MciG-ggg Please go ahead. Thanks

MciG-ggg:
solved at https://github.com/vllm-project/vllm-omni/pull/6230, this issue should be closed

yzong-rh:
Thanks for the work! Closing.

yzong-rh:
On second look, looks like field name validation isn't implemented per reviewer comment. This leaves some foot guns where certain fields (mentioned in the issue above) may be unexpectedly not applied in `--stage-overrides`.

Ideally, we have a blacklist warning the user when they attempt to pass on of those fields. Feel free to close as Not Planned if that is the case

PR #6200, opened 2026-08-14
[Refactor] Remove legacy stage_args YAML loader

zwhzzz0821:
## Purpose

This PR continues the configuration cleanup tracked by #5266 and RFC #4021. All in-tree `stage_args` YAML files have now been migrated to the pipeline registry plus `vllm_omni/deploy/`, so the legacy YAML loading path is no longer needed.

It removes `load_stage_configs_from_yaml()` and the fallback in `load_stage_configs_from_model()` that attempted to resolve and parse a legacy stage config. Models that are not resolved by `StageConfigFactory` now return an empty stage list, preserving the existing caller-owned default diffusion-stage fallback.

The legacy loader tests are removed and replaced with coverage asserting that unresolved models no longer fall back to YAML path resolution. The PR also updates executable examples that still passed the removed `stage_configs_path` kwarg to use `deploy_config`, and removes stale calls to the deleted `nullify_stage_engine_defaults()` helper.

To prevent the legacy interface from being silently reintroduced, runtime resolution now rejects the removed `stage_configs_path` and `stage_configs` Python arguments, while the deploy parser rejects YAML files using the removed top-level `stage_args` schema with migration guidance.

## Test Plan

**vLLM Version:** 0.25.0

**vLLM-Omni Commit:** This PR

```bash
env HF_HUB_OFFLINE=1 TRANSFORMERS_OFFLINE=1 \
  .venv-config/bin/python -m pytest -q \
  tests/entrypoints/test_utils.py \
  tests/entrypoints/test_resolve_dreamzero_config.py \
  tests/entrypoints/test_async_omni_diffusion_config.py \
  tests/config/test_config_factory.py \
  -k 'not test_dreamzero_enrich_config_preserves_explicit_model_class_name'

.venv-config/bin/python -m py_compile \
  vllm_omni/entrypoints/utils.py \
  tests/entrypoints/test_utils.py \
  examples/offline_inference/text_to_speech/qwen3_tts/word_timestamps.py \
  examples/offline_inference/nemotron_voicechat/end2end.py \
  examples/offline_inference/aura_omni/end2end.py \
  examples/offline_inference/lance/end2end.py

.venv-config/bin/python examples/offline_inference/nemotron_voicechat/end2end.py --help
.venv-config/bin/python examples/offline_inference/aura_omni/end2end.py --help
.venv-config/bin/python examples/offline_inference/lance/end2end.py --help
```

## Test Result

- Config and entrypoint tests: `355 passed, 1 deselected`.
[...]

PR #6230, opened 2026-08-16
[Bugfix] Validate --stage-overrides shape in parse_stage_overrides (drop field allowlist)

MciG-ggg:
## Purpose

`parse_stage_overrides` (the CLI parser behind `--stage-overrides`)
previously ran only `json.loads` on the input string (issue #5728).
Four malformed inputs parsed cleanly and were silently accepted
downstream:

- top-level not a dict (e.g. `[1,2,3]`)
- stage_id not an integer string (e.g. `{"abc": {...}}`)
- negative stage_id (e.g. `{"-1": {...}}`)
- value not a dict (e.g. `{"0": "oops"}`)

This change adds structural (shape) validation at parse time:

- top-level must be a dict
- keys must be non-negative integer strings
- values must be dicts

Field-name validation is **not** enforced at parse time. The downstream
resolver (`load_stage_configs_from_model`) forwards every surviving
`stage_<id>_<key>` into `cli_overrides`, where `StageConfigFactory`
consumes what it knows and ignores the rest; surface area for "unknown
field" is there, not at the parser. Type / range validation is
deliberately left to `StageDeployConfig` instantiation, and stage-id
existence is left to `load_stage_configs_from_model`.

The earlier draft also rejected unknown field names against an allowlist
(`StageDeployConfig` fields ∪ `deploy_runtime_override_keys()`). Reviewer
feedback noted that allowlist was narrower than the existing override
contract: it rejected the structured `extras` key read by the
default-diffusion fallback in `async_omni_engine.py` (regression for
recipes like LTX-2.5), and stage-scoped engine args the resolver
forwards verbatim (`kv_cache_dtype`, `seed`, `swap_space`, `block_size`,
`stage_connector_spec`, ...). Listing an allowlist here would duplicate
the engine-arg contract and miss new fields on day one. The allowlist
is removed; parser is shape-only.

On rejection, `ValueError` cites `--stage-overrides` and the bad shape
so the user sees a targeted error instead of a confusing downstream
failure.

## Test Plan

7 new tests in `tests/entrypoints/test_serve.py` (prefix
`test_parse_stage_overrides_`): happy-path JSON parse, `None` / empty /
empty-dict passthrough, non-string pass-through, invalid-JSON raise,
non-dict top-level raise, non-integer / negative stage_id raise,
non-dict value raise, and one accepting-style test for `extras` plus the
engine-arg keys the resolver forwards (`kv_cache_dtype`, `seed`,
`swap_space`, `block_size`, `stage_connector_spec`).

Run locally:

```bash
pytest tests/entrypoints/test_serve.py -k parse_stage_overrides -v
```

**vLLM Version:** not pinned in `pyproject.toml` (no explicit dep entry);
assumed latest vllm available at merge time.

**vLLM-Omni Commit:** fb44dd2545d50b9ab2572f3e897ea9df64e0c52c

## Test Result

Logic-checked offline (sandbox lacks a vLLM install): the parser
function body was exercised in isolation against 16 cases — 7
happy-path (including `extras`, engine-arg keys, and a future-unknown
field), 8 expected rejections (invalid JSON, non-dict top level,
non-integer / negative / float keys, non-dict value, integer-key
already-parsed mapping). All 16 match expected outcomes. Upstream CI
matrix expected to exercise the new tests once the head is refreshed.

Pre-PR self-check (`.claude/skills/precheck-pr`) passes on all
dimensions:

- `git diff --stat`: 2 files, +55 / −28
- code-quality sweep on diff (kwargs / broad-except / Any / clone-deepcopy /
  blocking-IO): all empty
- examples-policy: no new model-specific Python examples
- PR title / sign-off / rebase / diff scope: ✓


MciG-ggg:
Self-review:

- Pre-flight: `.claude/skills/precheck-pr` (0 blocking / 0 warnings).
  Diff-scoped code-quality sweep (kwargs / broad-except / Any / clone /
  blocking) all empty.

- Bug: `parse_stage_overrides` previously only ran `json.loads`,
  silently accepting 5 malformed input shapes (non-dict top-level,
  non-integer / negative stage_id, non-dict value, unknown field). All
  now raise `ValueError` at parse time citing `--stage-overrides`.

- Scope: structural validation only. Type / range deferred to
  `StageDeployConfig` dataclass instantiation; stage_id existence
  deferred to `load_stage_configs_from_model`. No duplicated type info.

- Tests: 8 new regression tests in `tests/entrypoints/test_serve.py`
  covering happy-path, `None` / empty / empty-dict passthrough, non-string
  pass-through, invalid-JSON raise, non-dict top-level raise, non-integer /
  negative stage_id raise, non-dict value raise, unknown-field raise.
  Verified existing valid forms (`{"0": {"devices": "0,1"}}`,
  `{"1": {"enforce_eager": true}}`) still parse — no behavior change.

- Rebased onto `upstream/main` (`baba7d1e`); resolved one trivial
  import-block conflict where upstream #6200 had already removed the
  unused `merge_configs` import.

- CI all green: DCO, pre-commit (after `ruff check --fix` + `ruff format`),
  build 3.11, build 3.12, docs.

Diff: +82 / -4 across 2 files.

MciG-ggg (commented):
Inline self-review (line-anchored):

- Allowlist construction (utils.py:544): built once at import, union of StageDeployConfig fields minus stage_id plus deploy_runtime_override_keys().
- Validation loop (utils.py:579): shape-first fail-fast, str.isdigit() covers letters/signs/floats, integer-key path exercised via direct dict pass.
- Unknown-field test (test_serve.py:196): asserts offending key surfaces in error message, verified across all 5 previously-silent input shapes.

vllm_omni/entrypoints/utils.py
```diff
@@ -535,23 +539,60 @@ def filter_stages(
         return stage_configs
 
 
+# Positive contract for stage override keys. Build once at import
+# so we don't re-frozenset on every parse call.
+_STAGE_OVERRIDE_ALLOWED_KEYS: frozenset[str] = (
```
MciG-ggg, line 544: Built once at import to avoid re-frozenset on every parse. Field set = StageDeployConfig fields minus stage_id (not an override target) UNION deploy_runtime_override_keys() (engine-level runtime options). Cross-checked against existing valid uses: {devices}, {gpu_memory_utilization}, {enforce_eager}.

vllm_omni/entrypoints/utils.py
```diff
@@ -535,23 +539,60 @@ def filter_stages(
...
+
+    if not isinstance(parsed, dict):
+        raise ValueError(
+            f"--stage-overrides must be a JSON object mapping stage_id -> overrides, "
+            f"got {type(parsed).__name__}: {parsed!r}"
+        )
+    if not parsed:
+        return None
+
+    for stage_id_str, overrides in parsed.items():
```
MciG-ggg, line 579: Order matters: shape check first (fail-fast on wrong type), then per-stage. str.isdigit() rejects letters ("abc"), signs ("-1"), and floats ("1.5"). Integer (non-string) keys bypass json.loads string normalization - exercised via parse_stage_overrides({1: {}}) in the test.

tests/entrypoints/test_serve.py
```diff
@@ -165,6 +165,43 @@ def test_parse_stage_overrides_invalid_json_raises() -> None:
...
+    for bad_key in bad_string_keys:
+        with pytest.raises(ValueError, match="non-negative integer stage ids"):
+            parse_stage_overrides("{" + bad_key + ": {}}")
+    # Integer (non-string) key: must reach the structural check, not the JSON
+    # parser, so pass an already-parsed mapping directly.
+    with pytest.raises(ValueError, match="non-negative integer stage ids"):
+        parse_stage_overrides({1: {}})
+
+
+def test_parse_stage_overrides_rejects_unknown_field() -> None:
```
MciG-ggg, line 196: Asserts bogus_field appears in the error message so the user sees the offending key, not just a generic 'unknown fields' complaint. Verified the 5 previously-silent input shapes (non-dict top, non-integer / negative stage_id, non-dict value, unknown field) each surface a targeted message.

vllm_omni/entrypoints/utils.py
```diff
@@ -538,23 +542,60 @@ def filter_stages(
...
+        if not isinstance(stage_id_str, str) or not stage_id_str.isdigit():
+            raise ValueError(
+                f"--stage-overrides keys must be non-negative integer stage ids (as strings), got {stage_id_str!r}"
+            )
+        if not isinstance(overrides, dict):
+            raise ValueError(
+                f"--stage-overrides[{stage_id_str!r}] must be an object, got {type(overrides).__name__}: {overrides!r}"
+            )
+        unknown = set(overrides) - _STAGE_OVERRIDE_ALLOWED_KEYS
```
linyueqian: [blocking] Keep supported `extras` overrides valid here. The existing default diffusion-stage path passes an `extras` mapping for stage 0, but this allowlist rejects it before the configuration can preserve it, so both the general and AMD per-PR builds fail. Include or translate `extras` according to the stage merge contract while retaining validation for genuinely unknown fields.

vllm_omni/entrypoints/utils.py
```diff
@@ -538,23 +542,60 @@ def filter_stages(
         return stage_configs
 
 
+# Positive contract for stage override keys. Build once at import
+# so we don't re-frozenset on every parse call.
+_STAGE_OVERRIDE_ALLOWED_KEYS: frozenset[str] = (
```
linyueqian: [important] Do not define valid stage overrides only from the deploy-YAML schema. This CLI previously accepted stage-scoped engine arguments, and the resolver still forwards every surviving `stage_<id>_<key>` into that stage's engine args. This allowlist now rejects supported values such as `kv_cache_dtype`, `seed`, `swap_space`, `block_size`, and `stage_connector_spec`. Include the actual engine-argument contract, plus structured keys such as `extras`, or introduce an explicit migration instead of hard-failing existing inputs.

linyueqian (changes requested):
Reviewed at `cea58a61`. The shape validation is useful, but the field allowlist is narrower than the existing override contract. It rejects the supported `extras` field before the default diffusion-stage configuration can preserve it, which fails the same existing regression in both the general and AMD per-PR builds. It also rejects stage-scoped engine arguments that the resolver and current tests support, such as `kv_cache_dtype`. I am requesting changes.

Validation: static review of the two-file patch, the full override propagation path, current tests, and the failed Buildkite traces. Claude Opus independently confirmed the `extras` blocker and identified the broader engine-argument regression; both were verified in source before inclusion. No fork code was executed locally.

vllm_omni/entrypoints/utils.py
linyueqian: [blocking] `extras` is still rejected here even though the default diffusion-stage path reads and merges it. Keep that supported input valid while rejecting genuinely unknown fields.

vllm_omni/entrypoints/utils.py
linyueqian: [important] This allowlist is still narrower than the supported stage override contract. Values such as `kv_cache_dtype` are forwarded to per-stage engine arguments and covered by existing tests, but are rejected here because they are absent from the deploy schema. Validate against the actual engine-argument contract or preserve compatible pass-through handling.

linyueqian (changes requested):
Reviewed at `97532ef4`. This is a clean merge from main with no conflict-resolution changes in the PR files. The two override regressions from the prior review remain: supported `extras` and stage-scoped engine arguments are still rejected before they reach the existing consumers. I am requesting changes.

Validation: exact-head static review, merge-tree comparison, current source and test coverage, `git diff --check`, and an independent Codex audit. No fork code was executed.

linyueqian:
Re-reviewed the two follow-up commits at 1d8e2f7f. The ASCII guard on stage ids closes a real hole: str.isdigit accepts fullwidth digits that the downstream stage_<id>_ pattern would then misroute as a global key, and the inline comment says exactly that. The warning on dropped unknown engine-arg keys gives a typo a visible trace instead of silent loss. Both come with tests. No new finding; the approval I drafted at 4411d5ad still holds on content.

Two mechanical steps before it can post. Pre-commit fails on ruff format at this head while every other hook passes, so please run ruff format on the two touched files and push. Per-PR Buildkite has not run on this head either; I will re-fire the lanes on that push and approve once all four report green.

linyueqian (approved):
Approving at acde7ea5. Both round-2 findings stay resolved, and the two follow-up commits are good changes on their own: the ASCII guard on stage ids closes a real hole, since str.isdigit accepts fullwidth digits that the downstream ASCII-only stage_<id>_ pattern would then misroute as a global key, and the warning in filter_dataclass_kwargs gives a mistyped override a visible trace instead of silent loss. Both come with tests, and the last two pushes were a ruff format pass and a clean merge of main with the patch unchanged.

Validation: exact-head static review of the full patch against the current resolver and its consumers, plus per-PR Buildkite at this head. General, Intel, and NPU are green. AMD fails only its Diffusion Model Test step, on four SenseNova paged-decode tests that raise ImportError because the ROCm image has no CUDA flash attention extension. This PR changes three entrypoints files and no diffusion code, so that lane failure is inherited from the AMD environment rather than caused here. No fork code was executed.

Issue #6227, opened 2026-08-15
[RFC] Real-Time Stateful Video Generation Serving Runtime


linzhenpl07:
## Summary

This is an umbrella/integration RFC for making **real-time stateful video generation** a first-class serving mode in vLLM-Omni.

The intent is **not** to re-spec existing components. Several important pieces already exist or are being developed independently:

- #1987: world-model support and multi-turn realtime interaction;
- #4480: generalized session memory/state management;
- #5491: typed realtime AR-Diffusion tick/session execution;
- #3632 / #5527: streaming output and structured mid-stream interaction;
- #2280: model-depth PP + Stream Batch for real-time video execution;
- #4707: replica-level DP / StagePool scaling for video generation;
- #290: chunk/step-level SLO-aware diffusion scheduling;
- #2089: VAE as a separate stage;
- #6226: session-affinity routing for replicated realtime AR-Diffusion.

What is still missing is a **single production-oriented vertical contract and rollout plan** connecting these pieces from long-lived session state to incremental decoded video delivery and multi-session/multi-replica execution.

This aligns with #4901 (vLLM-Omni Q3 Roadmap), which explicitly calls out **World Model Support and Diffusion Generation Serving**, diffusion KV-cache management, DP/EP/PP support, and refactoring the `DiffusionEngine` path including **streaming**.

cc @hsliuustc0106 @Gaohan123

---

## Motivation

Traditional diffusion video serving is request-oriented:

```text
request
  -> encode
  -> full denoising loop
  -> VAE decode
  -> return complete video
  -> request state is released
```

Realtime causal / AR video generation has a different lifecycle:

```text
create session
  -> accept prompt/action/control update
  -> generate next chunk
  -> commit persistent state
  -> decode and emit that chunk
  -> accept next update
  -> generate next chunk
  -> ...
  -> reset / close
```

For this workload, low single-request latency is necessary but not sufficient. The serving runtime also needs:

- persistent session state across requests/ticks;
- bounded rolling KV / non-KV memory;
- chunk-wise incremental media output;
- interaction at chunk boundaries;
- low-latency multi-GPU execution for short causal chunks;
- multi-session scheduling and backpressure;
- multi-replica placement with session locality;
- explicit SLO and memory-admission semantics.

Today these capabilities are being built in separate RFCs/PRs, but there is no umbrella definition of what constitutes a production-ready realtime video serving path or how the components compose.

---

### 1. Realtime session/control plane — #5491 / #5527

[...]

PR #5491, opened 2026-07-28
[Core] Add realtime AR-Diffusion tick sessions for LingBot World 2.0

Jack47:
## Purpose

Add a typed realtime AR-Diffusion tick/session path for LingBot World 2.0.

This PR builds on:

- #5271 (merged): generalized AR-Diffusion KV session capabilities.
- #5022 (open draft): LingBot World 2.0 model and causal pipeline.

The change introduces:

- a model-neutral `ARDiffusionTickRequest` contract carried through `sampling_params.extra_args["ar_diffusion_tick"]`;
- one-request/one-AR-block execution with deterministic `session_id`, `request_id`, `event_id`, and `chunk_index` semantics;
- LingBot camera/prompt conditioning updates at chunk boundaries without a private public session protocol;
- an internal session manager and tick consumer connecting lifecycle RPCs to `AsyncOmni` and `ARDiffusionEngine`;
- a standard output metadata envelope at `multimodal_output["metadata"]["ar_diffusion"]`;
- fail-closed session behavior: a failed tick releases runner state and requires explicit reset or close;
- bounded multi-session residency and post-commit paged-KV pruning;
- explicit mutable paged-KV inputs for CUDA Graph correctness;
- continuous camera motion across chunk boundaries;
- direct/paged attention alignment so the offline path remains a numerical oracle.

This is a ready-for-review PR. If #5022 lands first, the overlapping LingBot model commits can be removed through a normal rebase.

## Scope and non-goals

This PR provides the internal, experimental session/control-plane and model execution path needed for realtime world-model inference. It does **not** add a public HTTP or WebSocket API. A transport adapter can build an SGLang-style interactive API on top of the session manager without introducing a LingBot-specific protocol into the core engine. Huang Zeyu's draft [#5527](https://github.com/vllm-project/vllm-omni/pull/5527) is the companion public structured-interaction/API proposal.

Each realtime request returns exactly one latent AR block plus typed chunk metadata. Stateful streaming VAE decode and incremental pixel-frame delivery are not included; the offline path can still decode a complete latent sequence through the VAE.

The current scoped implementation does **not** implement SGLang-style blank-tail image-condition repetition for unbounded sessions. The fixed 117-pixel-frame image-condition horizon therefore allows at most **10 realtime ticks per generation epoch** (`chunk_index` 0–9); reset or create a session to start a new world. The example rejects an 11th event, and the LingBot pipeline independently fails fast before prompt encoding or model execution.

[...]

Issue #4707, opened 2026-06-25
[RFC]:  Data Parallelism for Video Generation Models in vLLM-Omni


evanchueng:
### Motivation.

# RFC: Data Parallelism for Video Generation Models in vLLM-Omni

## Abstract

This RFC proposes a comprehensive, model-agnostic Data Parallelism (DP) extension for vLLM-Omni targeting video generation models, building upon the foundational work in **PR #4460** ("Omni-Replica + vLLM-DP/EP by YAML"). PR #4460 introduces YAML-based DP/EP configuration for the AR stage, fixes critical KV-transfer port collision bugs, and enables multi-replica deployments with inner-DP support. This RFC extends that foundation to video generation models (e.g., HunyuanVideo 1.5, 8.3B), providing a generic DP system that enables linear throughput scaling across NVIDIA H100 and Ascend 910B platforms.

---

### 1.1 The Throughput Challenge

Video generation models are growing rapidly in size. Production deployments with high request concurrency require throughput scaling beyond what a single device can provide. Data Parallelism (DP) addresses this by replicating the model across multiple devices and processing independent requests in parallel.

| Model | Parameters | Architecture |
|---|---|---|
| HunyuanVideo 1.5 | 8.3B | Diffusion Transformer (DiT) |

### 1.2 Current State in vLLM-Omni

vLLM-Omni already supports multiple parallelism strategies:

| Strategy | Primary Use Case | Throughput Gain |
|---|---|---|
| Tensor Parallelism (TP) | Large model memory reduction | Low-Medium |
| Sequence Parallelism (SP) | Long sequences | High |
| Pipeline Parallelism (PP) | Multi-stage pipelines | Medium |
| **Data Parallelism (DP)** | **High request throughput** | **Linear** |

vLLM-Omni provides foundational DP infrastructure:
- `_DIT` GroupCoordinator for DiT-specific DP groups
- DP rank tracking in `GPUARWorker.init_device()`
- `data_parallel_backend` configuration supporting "mp" and "ray" backends
- **PR #4460**: YAML-based DP/EP configuration, multi-replica support, and KV-transfer fixes

### 1.3 PR #4460 Foundation

PR #4460 establishes the core infrastructure that this RFC builds upon:

| PR #4460 Feature | Description | Relevance to This RFC |
|---|---|---|
| **YAML-based DP/EP** | Support vLLM DP and EP by YAML config for AR stage | Enables declarative DP configuration |
[...]

Issue #4480, opened 2026-06-16
[RFC]: Generalized session memory manager for world models


wjsuijlenh:
# [RFC] Generalized session memory manager for world models

## Summary

Extract the custom per-session caches of three AR-DiT world models — DreamZero (`state_dreamzero.py`, in tree), Cosmos3 (per-generation `cached_kv`/`reset_cache` on the `Cosmos3VFMTransformer` instance, in tree), and Lingbot-World-Fast (`LingbotWorldFastState` / `state_lingbot_world_fast.py`, proposed in **open** PR #3701 — not yet merged) — into a single **session memory manager** that (1) backs self-attention KV with paged blocks per #2305/#4366, and (2) models a session as a set of **typed memory objects**, so non-KV world-model memory (context latents, compressed history, spatial/recurrent state, retrieval pools) plugs in without redesign.

Related: #1987 (World Model support), #2305 and #4366 (paged KV cache for AR diffusion, **landed as PR #4534** — `PagedKV` below builds *on* its engine rather than re-speccing it), PR #3701, PR #4657 (Cosmos3 Phase-0 port), #4907/#4909 (copy-on-write KV fork on the landed engine), #4901 (Q3 roadmap: "diffusion KV cache manager"), #4497 (WMMS follow-up semantics).

**Layering and dependency.** Three layers, bottom-up: (1) **#4366 / PR #4534 — the AR-Diffusion engine, MERGED to `main`** (commit `86bdcaf3`, `vllm_omni/experimental/ar_diffusion/`): the KV block allocator (`ARDiffusionKVCache` owning a vLLM `KVCacheManager` / `BlockPool` as a library, `ChunkWindowManager` windowed eviction, per-session `ARDiffusionKVState` handles), plus #4909's copy-on-write KV fork proposed at this layer; (2) **this RFC** — typed session-memory objects + a cross-session byte budget + session-level fork, with `PagedKV` wrapping the engine's per-session state; (3) downstream world-model memory *semantics* (geometric/pose-indexed retrieval, world-state checkpoint/fork) as a separate follow-up under #1987/#4497. Phase 0 is allocator-independent (dense buffers); Phase 1 (`PagedKV`) sits on the now-landed engine and is unblocked.

## Motivation

[...]

PR #6233, opened 2026-08-16
[Core] Add session-affinity routing for replicated AR-Diffusion

linzhenpl07:
## Purpose

Phase 0 of [RFC #6226](https://github.com/vllm-project/vllm-omni/issues/6226): keep every tick of a live realtime AR-Diffusion session on the replica that owns its worker-local state, so the AR stage can be replicated for Replica-DP scaling.

#5491 landed the realtime tick/session path but pinned the AR stage to exactly one replica, because session state is worker-local while `StagePool` routing is request-oriented:

```python
# vllm_omni/experimental/ar_diffusion/consumer.py (before)
if replica_count != 1:
    raise ValueError(
        "AR-Diffusion realtime sessions currently require exactly one "
        "replica for the selected diffusion stage; session-affine "
        f"routing is not implemented (got num_replicas={replica_count})."
    )
```

This PR implements that routing and lifts the restriction.

## Design

`StagePool` gains a session-scoped route alongside its existing request-scoped ones (`_request_bindings` / `_affinity`):

```text
bind_session(session_id, replica_id)
get_session_replica_id(session_id) -> replica_id | None
release_session(session_id)
session_ids_for_replica(replica_id)
orphan_replica_sessions(replica_id)
```

The first tick of a session is placed by the stage's existing load-balancing policy and binds the session; every later tick routes back to the owner. `session_id` is threaded from `AsyncOmni.generate()` through `StageSubmissionMessage` and `submit_initial()` into **both** dispatch paths — distributed `pick()` and legacy `select_replica_id()`.

Requests that pass no `session_id` route exactly as before; this is covered by a test.

Sessions are kept in their own key namespace rather than reusing the request-scoped maps: a session outlives every request bound to it, is released on reset/close instead of on request completion, and must not be able to collide with a `request_id`.

### Failure semantics: fail-closed

Session state is neither migrated nor rebuilt in this version. If the owning replica dies, the route row is dropped (nothing points at a dead replica) and the session is **tombstoned**, so later ticks raise `SessionOwnerLostError` rather than silently landing on a replica that holds none of its state — which would corrupt the rollout without any error surfacing.

`SessionOwnerLostError` subclasses `StageUnavailableError`, so the orchestrator's existing dispatch guard fails just that one request and keeps the server serving. Only an explicit reset/close clears the tombstone.

This answers open questions 2 and 5 in the RFC with the conservative option; an explicit rebind/rebuild hook can be layered on later.

### Lifecycle

Reset and close release the route through `ARDiffusionWorkerLifecycle`, **after** the worker RPC succeeds — a session whose state could not be released must stay pinned to the replica still holding it. That RPC already fans out to every live replica via `Orchestrator._handle_collective_rpc`, and the worker-side release is idempotent for unknown session ids (`self._sessions.pop(session_id, None)`), so replicated cleanup is correct without the caller tracking placement.

### Concurrency

The RFC requires that concurrent first ticks for one session cannot create two owners. In `pick()` the fast path from the session lookup to the bind contains no `await`, so it is atomic with respect to the event loop. The bounded retry loop (entered only when no replica is serviceable) is the one suspension point, so the session is re-resolved after each sleep instead of racing to the bind. Covered by a test that parks eight concurrent first ticks and then brings replicas up.

## Out of scope

Per the RFC's Phase 0 boundary: state migration between replicas, CPU/NVMe offload, byte-budget admission control, EDF/SLO scheduling, recompute-aware placement, session fork / CoW placement, and PP/Stream-Batch scheduling changes.

## Test Plan

New `tests/engine/test_stage_pool_session_affinity.py` (CPU, no GPU required) covers:

- ticks of one session stay on the owner replica, on both dispatch paths;
- new sessions are still spread by the existing load-balancing policy;
- stateless routing is unchanged, and stateless requests still reach every replica while a session is pinned;
- release on reset/close leaves no entry, is idempotent, and allows re-placement;
- lost owner fails closed, and keeps failing until the session is explicitly released;
- `invalidate_addr` / `mark_replica_unavailable` orphan only that replica's sessions;
- unrelated sessions are unaffected by a replica death;
- concurrent first ticks converge on a single owner.

`tests/diffusion/ar_diffusion/test_consumer.py` is updated: the old "rejects replicated stage" test becomes "accepts replicated stage", plus new coverage that the tick carries its `session_id` and that the route release is scoped to the consumer's diffusion stage.

`tests/engine/test_orchestrator.py` pins the plumbing in between, which nothing else exercised: the stage behavior is covered directly and the consumer is covered with fakes, but the `session_id` hop from `StageSubmissionMessage` through the orchestrator into `submit_initial` was untested. The new test runs real `StagePool`s with two replicas and asserts that a second tick of the same session lands on the owner where round-robin alone would have moved it, that a request without a `session_id` still load-balances, and that release frees the id. I verified it fails (`assert None == 0`) when the orchestrator stops forwarding `msg.session_id`.

`tests/diffusion/ar_diffusion/test_session.py` covers `route_release` itself, which also had no test: it runs after a successful reset/close, and does **not** run when the RPC fails.

## Known limitations

Stating these up front rather than leaving them to be found:

1. **A lifecycle built without `route_release` leaks routing entries.** `StagePool.release_session` is the only path that drops a binding, and it only runs through that hook. The hook is optional for backwards compatibility, so a caller that omits it leaves one entry per session behind for the process lifetime, keyed by a client-supplied session id. The only in-tree wiring (the LingBot realtime example) passes it; construction now warns when it is missing. Making it required would be the stricter fix and I am happy to switch if that is preferred.
2. **Lost-owner tombstones are unbounded.** `_lost_sessions` is cleared only by an explicit reset/close, which is what makes the fail-closed guarantee hold — a client that disappears after its owner died leaves a tombstone behind. Bounding it means choosing an expiry policy, which felt like it belonged with the residency/byte-accounting work in #4480 rather than here.
3. **Worker-side LRU eviction is invisible to the routing table.** `ARDiffusionModelRunner` can evict a session (`reason="lru_eviction"`) without the engine side releasing its binding, so the next tick routes to a replica that no longer holds the state. This is pre-existing rather than a regression — the same tick fails the chunk-contiguity check at one replica — and re-placing would not help, since no replica holds that state either. A worker→engine eviction notification would be the real fix.

## Test Result

Local, all CPU-only:

| file | result |
| --- | --- |
| `tests/engine/test_stage_pool_session_affinity.py` | 18 passed |
| `tests/diffusion/ar_diffusion/` | 148 passed |
| `tests/engine/test_orchestrator.py` + the above | 218 passed |

`ruff check` and `ruff format --check` clean on all touched files.

I do not have a GPU-capable environment for this branch, so I have not run the full Buildkite suite locally — the changed dispatch paths are exercised by the CPU tests above. Happy to iterate on anything CI surfaces.

Related: #6226 (this RFC), #6227 (umbrella realtime serving RFC), #5491 (consumer), #4707 (replica routing infrastructure), #4480 (future session residency/byte accounting).

cc @hsliuustc0106 @Gaohan123 @Jack47

### DIFF
--- a/docs/design/feature/realtime_ar_diffusion.md
+++ b/docs/design/feature/realtime_ar_diffusion.md
@@ -151,13 +151,23 @@ reports the measured 603 MiB per-session Wan VAE causal-convolution state.
 The current runner executes one request at a time:
 
 - `max_num_seqs` must be one;
-- request batching and diffusion step execution are rejected;
-- one AR block is generated per engine request; and
-- an AR-Diffusion stage must have exactly one replica.
+- request batching and diffusion step execution are rejected; and
+- one AR block is generated per engine request.
 
-The single-replica restriction is intentional. Multi-replica support requires
-session-affine routing so every tick, reset, and close reaches the worker that
-owns the session. Random or round-robin routing is not correct.
+An AR-Diffusion stage may be replicated. Each tick carries its `session_id`
+into dispatch, and `StagePool` pins a session to the replica that owns its
+state: the first tick is placed by the stage's ordinary load-balancing policy,
+every later tick is routed back to that replica, and the route is released on
+session reset or close. Reset and close themselves fan out to every live
+replica through the existing collective RPC and are idempotent for unknown
+session ids, so they reach the owner without the caller tracking placement.
+
+Placement is fail-closed. Session state is worker-local and is neither
+migrated nor rebuilt, so if the owning replica dies, later ticks for that
+session raise `SessionOwnerLostError` (a `StageUnavailableError`, so the
+orchestrator fails just that request) instead of silently landing on a replica
+that holds none of its state. The session stays in that state until it is
+explicitly reset or closed.
 
 Multiple resident sessions may share one runner. Capacity selection accounts
 for all persistent pools, and the runner performs LRU eviction through the same
@@ -186,7 +196,9 @@ This contract does not currently provide:
   frontend is tracked separately in
   [#5527](https://github.com/vllm-project/vllm-omni/pull/5527));
 - camera/action semantics shared by every world model;
-- session migration or replication across workers;
+- session-state migration between replicas, or rebuild/recovery after the
+  owning replica is lost;
+- byte-aware or HBM-aware session admission and placement;
 - retry of an ambiguous partially executed chunk;
 - cross-stage KV transfer; or
 - stateful streaming VAE decode.
@@ -204,7 +216,10 @@ CPU contract tests cover:
 - capacity one under the default fraction for shipped LingBot geometry;
 - capacity expansion under a tuned fraction;
 - DreamZero model-owned state accounting;
-- runner session reuse, reset, close, LRU eviction, and forward cleanup; and
+- runner session reuse, reset, close, LRU eviction, and forward cleanup;
+- session-to-replica affinity: stickiness across ticks on both dispatch paths,
+  release on reset/close, fail-closed routing after the owner is lost, and a
+  single owner under concurrent first ticks; and
 - LingBot model registration, imports, camera/action reduction, and one-block
   output metadata.
 
--- a/examples/offline_inference/diffusion/lingbot_world_v2_realtime.py
+++ b/examples/offline_inference/diffusion/lingbot_world_v2_realtime.py
@@ -173,7 +173,14 @@ async def run(argv: Sequence[str] | None = None) -> Path:
     )
     manager = ARDiffusionSessionManager(
         tick_consumer=consumer,
-        lifecycle=ARDiffusionWorkerLifecycle(engine, stage_ids=[0], timeout=180.0),
+        lifecycle=ARDiffusionWorkerLifecycle(
+            engine,
+            stage_ids=[0],
+            timeout=180.0,
+            # Ticks are pinned to the replica owning the session's AR state;
+            # reset/close hand the id back to the load balancer.
+            route_release=consumer.release_session_route,
+        ),
         max_pending_events=32,
         control_reducer_factory=LingBotCameraControlReducer,
     )
--- a/examples/offline_inference/diffusion/lingbot_world_v2_realtime.py
+++ b/examples/offline_inference/diffusion/lingbot_world_v2_realtime.py
@@ -1,10 +1,12 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
-"""Run LingBot-World 2.0 as an in-process realtime AR-Diffusion session.
+"""Run one or more LingBot-World 2.0 realtime AR-Diffusion sessions.
 
 This example exercises the internal session API. It is not a public HTTP or
 WebSocket protocol. Each JSONL event advances the world by one three-latent
-frame AR block and writes the latent plus its identity metadata. The current
+frame AR block and writes the latent plus its identity and replica-route
+metadata. With ``--deploy-config`` and ``--num-sessions 2`` this is also the
+hardware driver for session-affine Replica-DP validation. The current
 scoped implementation supports at most ten ticks per generation epoch because
 its image condition is bounded to 117 pixel frames; reset or create a session
 to start a new world.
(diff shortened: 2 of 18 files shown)


linzhenpl07:
Added the missing real-hardware reproduction path in c1af39f.

### 4-GPU experiment

Topology: 2 independent replicas x TP=2, with replica 0 on GPUs 0,1 and replica 1 on GPUs 2,3.

The checked-in driver now creates two sessions, submits every round concurrently, and fails if either session has no route, changes owner, shares an owner when distinct placement is required, produces non-finite latents, or leaks its route on close.

Run:

```bash
python examples/offline_inference/diffusion/lingbot_world_v2_realtime.py \
  --model robbyant/lingbot-world-v2-14b-causal-fast-diffusers \
  --deploy-config examples/offline_inference/diffusion/lingbot_world_v2_replica_dp_tp2.yaml \
  --image /absolute/path/to/input.png \
  --prompt "A vehicle travels through an outdoor scene" \
  --events examples/offline_inference/diffusion/lingbot_world_v2_realtime_events.jsonl \
  --output-dir runs/lingbot_replica_dp_tp2 \
  --session-id world \
  --num-sessions 2 \
  --require-distinct-replicas
```

Full procedure and pass criteria: examples/offline_inference/diffusion/lingbot_world_v2_replica_dp.md

The summary records per-session replica ids and chunk latency, per-round wall time, total tick wall time, aggregate chunks/s, and aggregate latent-frames/s. The last metric is generation throughput, not delivered RGB FPS or delivery RTF; streaming VAE remains a separate end-to-end validation.

Local CPU validation after this addition: 30 passed across the new example tests, StagePool session-affinity tests, and AR-Diffusion consumer tests. Ruff check/format and diff checks pass. This commit validates the recipe and routing contract on CPU; H200 measurements are still intentionally unclaimed until someone runs the command above.

linzhenpl07:
Closing this in favor of the direction agreed in #6227.

## Why

Following the discussion with @fhfuih in #6227, LingBot World should converge on the existing step-wise streaming contract (`SupportsStepExecution` / `StepRequestState` / `post_decode()` emitting `DiffusionOutput(finished=False)`) rather than growing a second serving stack around the experimental tick/session API.

Under that contract, **one interactive session is one streaming request**. `StepRequestState` is Runner-owned and stays resident in the worker across chunks, so there is no per-tick re-entry into the engine and therefore nothing to re-route. The problem this PR solves does not exist on that path. It exists only because the experimental AR API submits a new request per tick and needs a second, session-scoped affinity layer to compensate for that.

Merging it would encode the session-vs-request divergence into core routing (`stage_pool.py`, `orchestrator.py`, `messages.py`, `entrypoints/async_omni.py`) - precisely the divergence the #6227 discussion concluded should be removed - and would then need deprecating after the migration.

No measured results are lost by closing: as stated in my previous comment, H200 numbers were intentionally unclaimed.

## What I checked before closing

I wanted to know whether the fail-closed part (`SessionOwnerLostError`) survives the migration on its own merits, since "a stateful long-lived request must not be silently re-routed when its owner is gone" is not a session-specific claim.

Current main is closer to fail-closed than this PR assumed. `MembershipController.unregister_remote_replica()` calls `StagePool.invalidate_addr()` and then emits `ErrorMessage("stage replica disappeared")` for every affected request, so in-flight requests bound to a departing replica are failed rather than silently re-routed.

The residual hazard is narrower than this PR assumed, but it is real and sits on a path an interactive session uses:

```python
# StagePool.submit_update() and StagePool.submit_interaction()
replica_id = self.get_bound_replica_id(request_id)
if replica_id is None or self.clients[replica_id] is None:
    replica_id = await self._pick_or_select(request_id)
```

Both entry points re-pick a fresh replica when the binding is missing, without distinguishing "never bound" from "was bound, binding lost". `submit_interaction()` is the camera/action path a world session would use. For a stateless request, re-picking is correct recovery; for a request whose model state is worker-local, it delivers state-dependent input to a replica holding none of that state.

I could not demonstrate a routine, non-racy trigger on current main: the loud path emits an error first, and `mark_replica_unavailable()` - which drops bindings without notifying anyone - currently has no callers in `vllm_omni/`. So this is a latent hazard on a live code path, not an observed failure, and I am recording it as such rather than claiming a bug.

## What I will re-propose separately

A small, model-neutral, request-level change: `submit_update()` / `submit_interaction()` should distinguish a request that was never bound from one whose binding was lost, and fail the latter instead of re-picking. That needs no `session_id` namespace and applies to any request with worker-local state, not only AR-Diffusion.

## What carries over

- `examples/offline_inference/diffusion/lingbot_world_v2_replica_dp_tp2.yaml` (2 replicas x TP=2) and the pass criteria in the accompanying `.md` are independent of the routing implementation, and will be re-pointed at the standard streaming path.
- The driver example targets the experimental API and will be rewritten against `/v1/realtime/video`.

The underlying reasoning moves to the admission/affinity item in #6227: when model state is worker-local, routing must be sticky and a lost owner must fail closed rather than silently re-route. The migration makes the sticky half structural; the fail-closed half is the piece worth keeping, and I will raise it on its own.

PR #3279, opened 2026-04-30
[Quantization] Redo Z-Image text encoder FP8 online quantization

Isotr0py:
PLEASE FILL IN THE PR DESCRIPTION HERE ENSURING ALL CHECKLIST ITEMS (AT THE BOTTOM) HAVE BEEN CONSIDERED.

## Purpose
- Redo https://github.com/vllm-project/vllm-omni/pull/1338, which is reverted because of a broken import before.

## Test Plan
```
python examples/offline_inference/text_to_image/text_to_image.py --model /mnt/data0/LLM/Z-Image-Turbo/ --width 512 --height 512 --quantization fp8 --tensor-parallel-size 2
```

## Test Result
```
INFO 04-30 18:20:24 [diffusers_loader.py:355] Loading weights took 8.20 seconds
INFO 04-30 18:20:25 [diffusion_model_runner.py:142] Model loading took 7.7360 GiB and 11.881425 seconds
INFO 04-30 18:20:25 [diffusion_model_runner.py:147] Model runner: Model loaded successfully.
INFO 04-30 18:20:25 [diffusion_model_runner.py:142] Model loading took 7.7371 GiB and 11.893948 seconds
INFO 04-30 18:20:25 [diffusion_model_runner.py:147] Model runner: Model loaded successfully.
INFO 04-30 18:20:25 [diffusion_model_runner.py:86] Model runner: transformer compiled with torch.compile.
INFO 04-30 18:20:25 [diffusion_model_runner.py:188] Model runner: Initialization complete.
INFO 04-30 18:20:25 [diffusion_model_runner.py:86] Model runner: transformer compiled with torch.compile.
INFO 04-30 18:20:25 [diffusion_model_runner.py:188] Model runner: Initialization complete.
INFO 04-30 18:20:26 [diffusion_worker.py:199] Worker 0: Process-scoped GPU memory after model loading: 8.25 GiB.
INFO 04-30 18:20:26 [manager.py:96] Initializing DiffusionLoRAManager: device=cuda:0, dtype=torch.bfloat16, max_cached_adapters=1, static_lora_path=None
INFO 04-30 18:20:26 [diffusion_worker.py:97] Worker 0: Initialization complete.
INFO 04-30 18:20:26 [diffusion_worker.py:703] Worker 0: Scheduler loop started.
INFO 04-30 18:20:26 [diffusion_worker.py:613] Worker 0 ready to receive requests via shared memory
[...]

PR #6234, opened 2026-08-16
[Bugfix] Prevent LTX-2.5 from silently loading unindexed Diffusers shards

mglyn:
# [Bugfix] Prevent LTX-2.5 from silently loading unindexed Diffusers shards

## Summary

Fix Diffusers component loading when a subfolder contains indexed shards plus stale/unindexed `.safetensors` files. The existing path duplicated the subfolder while locating the index, so filtering failed open and every shard was loaded.

Because duplicate parameter names are accepted and later shards overwrite earlier values, initialization still succeeds with a mixed checkpoint. This causes silent output-accuracy degradation instead of a load error.

Affected baseline: vLLM `0.27.0`. The shared Omni subfolder/index regression was introduced by `0a8204cb` (#3279); LTX-2.5 began triggering it when support landed in `ce6ab0ca` (#6070).

## Reproduction

From an empty HF cache:

```bash
vllm serve Lightricks/LTX-2.5-Diffusers \
  --omni --model-class-name LTX2DistilledOneStagePipeline --enforce-eager
```

The current loader requests repository-wide `*.safetensors` (31 files, about 164 GB), reports that `transformer/...index.json` is missing, and loads all 12 Transformer shards. Diffusers resolves the same index to 4 shards. In our default four-thread loading reproduction, 4,090 parameter names were loaded twice and 2,500 final values came from unindexed shards; a sampled BF16 parameter differed in 19/32 elements.

## Fix

- Resolve the component index first from a local path or the configured HF cache/remote, preserving `revision` and offline mode.
- Treat its `weight_map` as the authoritative manifest: download/load exactly those shards and fail if any are missing.
- If an offline/legacy cache has no index but contains conflicting shard families (for example both `of-00004` and `of-00008`), fail closed instead of mixing them.
- Preserve explicit `allow_patterns_overrides` as the highest-priority file selection policy.
- Cover online, offline-cache, local-path, explicit-override, and stale-shard behavior.

Validation: `tests/diffusion/model_loader/test_diffusers_loader.py` passes (17 tests), and pre-commit passes for both changed files.


vllm_omni/diffusion/model_loader/diffusers_loader.py
```diff
@@ -164,11 +167,22 @@ def _prepare_weights(
...
+        possible_index_files = (
+            [(index_file, repo_relative_path(index_file)) for index_file in INDEX_FILES]
+            if allow_patterns_overrides is None
+            else []
+        )
         available_index_file = [
-            f for f in possible_index_files if file_exists(model_name_or_path, f, revision=revision)
+            index_file
+            for index_file, repo_path in possible_index_files
+            if file_exists(model_name_or_path, repo_path, revision=revision)
```
lishunyang12: `file_exists` returns `False` in `HF_HUB_OFFLINE` mode even when this index is present in the local Hub cache, so `index_file` stays empty and the stale-shard filtering is skipped. Please probe the cache (for example with `hf_hub_download(..., local_files_only=True)`) before falling back, otherwise offline launches can still silently mix the checkpoint this PR is meant to protect.

tests/diffusion/model_loader/test_diffusers_loader.py
```diff
@@ -91,6 +92,77 @@ def _iter_weights(_model):
...
+    folder, files, use_safetensors = loader._prepare_weights(
+        "org/model",
+        subfolder="transformer",
+        revision="revision",
+        fall_back_to_pt=True,
+        allow_patterns_overrides=None,
+    )
+
+    assert folder == str(transformer)
+    assert [str(transformer / filename) for filename in indexed_files] == files
```
lishunyang12, line 145: `glob.glob` does not guarantee directory-entry order, so this exact list comparison is flaky; the focused test failed for me with shard 2 returned before shard 1. Please compare sorted lists (or sets) so the test checks selection rather than filesystem ordering.

vllm_omni/diffusion/model_loader/diffusers_loader.py
mglyn: Fixed in dd9121a1. Index resolution is now manifest-first and probes the configured HF cache with `hf_hub_download(..., local_files_only=True)` in offline mode, while preserving `cache_dir` and `revision`. The regression test now runs in both online and offline modes.

tests/diffusion/model_loader/test_diffusers_loader.py
mglyn: Fixed in dd9121a1. Indexed shards now come directly from a sorted `weight_map` manifest rather than `glob`, so the assertion no longer depends on filesystem directory-entry order.

lishunyang12 (approved):

lishunyang12:
please rebase.

mglyn:
@lhsliuustc0106 The latest pre-commit failed after the new forbidden Hugging Face repo API check was enabled. I updated the index download to use the shared, vLLM-tagged hf_api().hf_hub_download() path in 7ccabd647 and adjusted the mocks. The CI-equivalent full pre-commit and all 19 loader tests pass locally. PTAL, thanks!

mglyn:
@hsliuustc0106 ready.

PR #6235, opened 2026-08-16
[Feat][dots.tts] Add Online Serving Support

sphinxkkkbc:
## Purpose

Support online serving for dots.tts, introduced in #4765, by wiring a dots.tts adapter into `serving_speech`.

The dots.tts entry stage uses the same `latent_generator` stage key as VoxCPM2. To avoid routing dots.tts requests to `VoxCPM2Adapter`, this change:

- identifies dots.tts by `DotsTTSForConditionalGeneration`;
- sets `detect_priority = 5` so its architecture is checked before VoxCPM2;
- leaves `stage_keys` empty so it does not claim VoxCPM2 requests by their shared stage key.

## Test Plan

Unit tests for adapter detection, validation, and serving integration:

```bash
pytest -v \
  tests/entrypoints/openai_api/test_tts_detection.py \
```

Full-model online-serving E2E tests:

```bash
pytest -v \
  tests/e2e/online_serving/test_dots_tts_expansion.py \
  --run-level full_model
```

## Test Environment

**vLLM Version:** 0.27.0  
**vLLM-Omni Commit:** 9cd7f7681ea32aa6e1ab6fd36eed9a90d41b5e90

## Test Results

- Unit tests: 352 passed
- E2E tests: 2 passed

**BEFORE SUBMITTING:** Read [[CONTRIBUTING.md](https://github.com/vllm-project/vllm-omni/blob/main/CONTRIBUTING.md)](https://github.com/vllm-project/vllm-omni/blob/main/CONTRIBUTING.md) and run the [[precheck-pr skill](https://github.com/vllm-project/vllm-omni/blob/main/.claude/skills/precheck-pr/SKILL.md)](https://github.com/vllm-project/vllm-omni/blob/main/.claude/skills/precheck-pr/SKILL.md) with the code agent for a self-check against project conventions.


sphinxkkkbc:
@linyueqian, @Moore-Z PTAL,  Thanks!

sphinxkkkbc:
@vllm-omni-review-bot

linyueqian (commented):
Reviewed the full diff at 9cd7f768 against merge base 5d09cf27. The routing design is the interesting part and it holds up: dots.tts and VoxCPM2 genuinely share the `latent_generator` stage key, and resolving the conflict by leaving `stage_keys` empty and matching on `DotsTTSForConditionalGeneration` with `detect_priority = 5` is the right shape, because it lets dots.tts claim its own architecture without ever competing for VoxCPM2's stage key. The added detection test pins both directions of that resolution, which is exactly the regression that would otherwise be easy to reintroduce.

I checked the validation guards against the request schema rather than reading them at face value, since `x_vector_only_mode is not None` would reject every request if that field defaulted to `False`. It is declared `bool | None` with `default=None`, so the guard is correct, and the same holds for `ref_audio`, `ref_text`, and `speaker_embedding`. Accepting `voice` only as `None` or `"default"` is consistent with the endpoint's own default-voice normalization and with the placeholder-only support this PR documents.

The documentation changes are accurate about what the integration does and does not do, and I appreciate that the recipe's "Known limitations" section was narrowed rather than deleted: voice cloning is still called out as unwired, and the offline-versus-online split is stated precisely.

One inline finding is worth resolving before merge, and it is about test fidelity rather than the serving code. Both new tests refer to the checkpoint by an id that no other file in the repo uses, and because the e2e file is marked `full_model` the disagreement cannot show up in the currently green checks.

Validation: static review only. I did not execute the fork head, did not launch a server, and did not run either test file, so the end-to-end claims in the recipe and README rest on your reported runs rather than my own. Adapter registration, detection priority, and the request-schema defaults were all confirmed by reading the code at the reviewed head.

tests/e2e/online_serving/test_dots_tts_expansion.py
```diff
@@ -0,0 +1,89 @@
...
+
+from tests.helpers.mark import hardware_test
+from tests.helpers.runtime import OmniServerParams
+from tests.helpers.stage_config import get_deploy_config_path
+
+pytestmark = [pytest.mark.full_model, pytest.mark.tts]
+
+os.environ["VLLM_WORKER_MULTIPROC_METHOD"] = "spawn"
+
+MODEL = "dots-studio/dots.tts-soar"
```
linyueqian, line 19: [important] This model id disagrees with every other reference to the checkpoint in the repo, including the two README tables and the recipe this same PR updates, which all say `rednote-hilab/dots.tts-soar`. The offline example defaults to `rednote-hilab/dots.tts-soar` and the talker source cites that repo for its config values, while `dots-studio` appears nowhere on main and is introduced only by this PR (here and in the `test_serving_speech.py` mock). One of the two is wrong. The unit-test mock is harmless because `AutoTokenizer.from_pretrained` is patched there, but this constant drives a real `OmniServerParams` launch, so if `rednote-hilab` is the correct id then the `full_model` lane will fail to resolve the checkpoint the first time it runs this file. Since the marker keeps it out of the currently green checks, the mismatch will surface only on the TTS hardware lane. Please align this with the documented id or, if `dots-studio` really is the canonical repo now, update the READMEs and recipe in the same change.

vllm_omni/entrypoints/openai/tts_adapters/dots_tts.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+
+    stage_keys = frozenset()
+    model_archs = frozenset({"DotsTTSForConditionalGeneration"})
+    name = "dots_tts"
+    detect_priority = 5
+
+    def __init__(self, ctx):
+        super().__init__(ctx)
+        self.tokenizer = AutoTokenizer.from_pretrained(
```
linyueqian: [suggestion] This is the only adapter in the package that builds its own tokenizer in `__init__`. The sibling `latent_generator` adapter, VoxCPM2, keeps no tokenizer and delegates prompt construction to `server._build_voxcpm2_prompt`, reusing what the server already loaded. Because every `_get_tts_adapter()` call site is on a request path and the instance is created lazily on first use, this `from_pretrained` runs inside an async handler and blocks the event loop while it resolves; it also keeps a second copy of a tokenizer the engine already holds. The blocking window is small once the HF cache is warm, which it will be if the model loaded, so this is not a correctness problem, but the repo has an explicit convention here (see the `TestTTSAsyncOffloading` coverage for event-loop-safe offloading of blocking TTS work). Building the tokenizer lazily inside `build`, or routing it through the server the way VoxCPM2 does, would match that convention.

tests/e2e/online_serving/test_dots_tts_expansion.py
sphinxkkkbc: Resolved in feabc561dbbeb232e69ada3a628afcc72410e4af by unifying the repo name to `dots-studio/dots.tts-soar`. Even if you use `rednote-hilab/dots.tts-soar`, it will still be routed to dots-studio.

vllm_omni/entrypoints/openai/tts_adapters/dots_tts.py
sphinxkkkbc: Resolved in feabc561dbbeb232e69ada3a628afcc72410e4af by lazily loading the tokenizer through a `make_async` wrapper.

For completeness, VoxCPM2 also maintains `_voxcpm2_tokenizer` in serving_speech; its first initialization currently occurs synchronously in `_voxcpm2_encode()`.

linyueqian (approved):
Approving at 743fbbd9. Both points from the first round are resolved, and this time there is CI behind the verdict: the general lane and Intel are both green at this head, and the NPU red is the environment-wide triton `constexpr_type` import break that fails main's own scheduled nightly, so it is not attributable to this branch. AMD was still running.

On the checkpoint id, you resolved the inconsistency in the direction I had offered as the alternative, making `dots-studio/dots.tts-soar` canonical rather than reverting the test. That is the better fix if that is genuinely the right repository, and you carried it through completely: the supported-models table, both example READMEs, the recipe, the offline `end2end.py` default, the talker source comments, and the config module all moved together, and `rednote-hilab/dots.tts-soar` no longer appears anywhere in the tree. The e2e file and the docs now agree, which was the actual problem.

The adapter change went further than the suggestion asked for. I only proposed deferring the tokenizer construction out of `__init__`; you also routed `_build_prompt` through `make_async` on the server's TTS executor, so neither the first `from_pretrained` nor the prompt build occupies the event loop. That matches the offloading convention the rest of the serving path follows rather than just moving the cost to the first request.

Validation: static review of the delta plus the CI results at this head. I did not execute the fork head, launch a server, or run the e2e file, so the `full_model` lane remains the first real exercise of the new test.

Issue #6231, opened 2026-08-16
[RFC]: Define DLO runtime-cache compatibility across parallelism dimensions


hsliuustc0106:
Parent RFC: #6195

Phase A implementation: #6213

## Motivation

Phase B of #6195 proposes a node-local normalized runtime mmap cache built through the ordinary model loader. Sharing is correct only when processes have equivalent final CPU runtime-weight layouts. The active compatibility scope is DP, TP, and SP: DP and most SP implementations replicate weights, while TP changes the rank-local layout.

This sub-issue should define that compatibility contract before the runtime cache is implemented. It is specifically about **host-weight layout equivalence and cache sharing**, not request scheduling or distributed inference orchestration.

## Compatibility model

A cache entry is shared by processes on the same node that resolve to the same runtime-layout identity.

| Dimension | Expected weight relationship | Proposed cache treatment |
| --- | --- | --- |
| Pure DP | Complete replicated weights | Exclude DP rank from the identity; equivalent replicas share one entry |
| SP | Usually complete replicated weights; sequence/activations are sharded | Exclude SP rank after proving the model does not mutate or shard weights; conservatively guard on SP implementation and world size |
| TP | Different rank-local shards and sometimes different transformed contents | Include TP world size and TP rank; build one entry per TP coordinate |
| CFG parallel | Usually replicated weights, but component roles may differ | Share only after component ownership and final-layout equivalence are proven |
| Multiple nodes | No common physical page cache | Use an independent cache namespace and builder election on each node |

For example, `DP=2, TP=2, SP=2` should normally produce two entries: the TP-rank-0 entry is shared by four DP/SP processes, and the TP-rank-1 entry is shared by the other four.

## Runtime-layout identity

The pre-build coordination identity should include at least:

```text
cache schema version
model identity, revision, and checkpoint fingerprint
loader/checkpoint-adapter implementation version
cached component identity and ownership boundary
runtime dtype and quantization configuration
merged adapter identity, when applicable
TP world size and TP rank
SP/CFG implementation and world-size guards
settings that change parameter names, shapes, contents, or ownership
```

The published manifest must then validate the final tensor names, shapes, dtypes, byte lengths, layout/stride policy, persistent buffers, and any alias/tied-weight semantics. A mismatch must invalidate the entry and fail closed rather than allowing processes to map incompatible bytes.

The coordination identity and the post-build manifest are separate concerns: peers need a stable key before a builder runs, while the manifest proves what was actually published.

[...]

PR #6240, opened 2026-08-16
[Diffusion] Add node-local host weight cache and direct H2D for DLO

hsliuustc0106:
## Summary

Add loader-owned host-backing selection for no-AllGather DLO, including an opt-in node-local final-layout host weight cache discussed in #6231 and an optional direct-H2D registration path. The registration contract is platform-neutral, with CUDA as the first backend:

- retain a compatible direct-checkpoint mmap plan when one already represents the required layout and the final-layout host cache is not explicitly selected;
- otherwise retain ordinary pinned loader weights unless cache use is explicitly selected, then build or join the host weight cache after the complete post-load/calibration lifecycle;
- publish final CPU DiT tensors as a content-addressed, sharded safetensors entry on node-local disk;
- let equivalent workers mmap the same immutable entry and release private allocator-backed copies;
- with the existing `pin_cpu_memory=True` policy, ask the active platform backend to register the complete final mapping and copy directly to the existing rotating device buffers;
- retain the existing bounded two-slot host-staging path when pinned memory is disabled or registration is unavailable;
- document component ownership and startup, inference, failure, and teardown coordination in the [DLO architecture](https://github.com/vllm-project/vllm-omni/blob/6bcd350301d71fb69aca4c82ec29dee75e05d3e9/docs/design/feature/offloader/distributed_layerwise_offload.md#architecture).

This PR includes the implementation formerly stacked in #6257. It extends the Phase A direct-checkpoint mmap work from #6213 without changing DLO AllGather behavior or adding an inference-time collective.

Addresses #6231. Part of #6195.

## Motivation

Phase A can share checkpoint pages only when the checkpoint layout already matches the TP1 runtime layout. TP-aware or otherwise transformed layouts fall back to the ordinary loader, so independently scheduled no-AllGather replicas keep private host copies.

The host weight cache normalizes final weights *after* ordinary loading. Matching workers can therefore share the final host layout without joining one DLO process group or coupling request schedules. The primary deployment is repeated independent TP engines on one node:

```text
engine A: TP0-A, TP1-A
engine B: TP0-B, TP1-B

cache: TP0 entry <- TP0-A, TP0-B
       TP1 entry <- TP1-A, TP1-B
```

The pageable host-cache fallback can share immutable host pages, but every streamed block must still be packed from mmap storage into staging slots before H2D. Under concurrent TP workers, that recurrent DDR work can delay ranks unevenly before ordinary TP collectives. Registering the shared final mappings removes the recurrent host copy without adding a DLO collective.

### Shared control plane

CLI and engine settings are validated into `OmniDiffusionConfig`. `DiffusersPipelineLoader` exclusively owns host-backing selection and hands the runner an optional `HostWeightPlan`; the runner passes that contract once to `DistributedLayerwiseOffloadBackend`. The backend consumes the existing topology and model `OffloadPlan`, allocates mode-specific resources, and installs per-block hooks. `DiffusionWorker` calls backend teardown before destroying the distributed environment.

Host backing (ordinary tensors, checkpoint mmap, or host-weight-cache mmap) remains independent from transfer protocol. Replica placement, independent-engine routing, and cross-node cache distribution remain outside DLO.

See the [shared architecture](https://github.com/vllm-project/vllm-omni/blob/6bcd350301d71fb69aca4c82ec29dee75e05d3e9/docs/design/feature/offloader/distributed_layerwise_offload.md#architecture) for component ownership and lifecycle boundaries.

### AllGather mode

AllGather behavior is unchanged. The loader provides ordinary tensors or a compatible checkpoint plan, and backend setup normalizes that source into persistent local shards over the existing DP group, or the existing SP group when DP is one. The hot path is local-shard H2D followed by DLO AllGather into the alternate device slot. Host-weight-cache mmap is not selected, TP is not a DLO sharding group, and participating ranks retain the lockstep block-order contract.

The [AllGather architecture diagram](https://github.com/vllm-project/vllm-omni/blob/6bcd350301d71fb69aca4c82ec29dee75e05d3e9/docs/design/figures/dlo/dlo_allgather_sharded_host.svg) contrasts process-owned pinned shards and per-TP-coordinate DP collectives with the no-AllGather shared-host layout.

See the separate [AllGather design](https://github.com/vllm-project/vllm-omni/blob/6bcd350301d71fb69aca4c82ec29dee75e05d3e9/docs/design/feature/offloader/distributed_layerwise_offload.md#allgather-design).

### No-AllGather mode

No-AllGather keeps complete rank-local block sources and creates no DLO weight group or DLO collective. Without explicit cache selection, the loader first returns a compatible checkpoint mmap plan when available and otherwise retains ordinary pinned tensors. Explicit cache use normalizes supported final tensors into the host weight cache and falls back to ordinary tensors on cache failure. Under `pin_cpu_memory=True`, DLO prefers registration and direct copies into the rotating device slots; disabling pinned memory or an unavailable registration backend retains bounded staging. Workers may schedule independently, while ordinary TP and SP model collectives remain unchanged.

The [shared-host architecture diagram](https://github.com/vllm-project/vllm-omni/blob/6bcd350301d71fb69aca4c82ec29dee75e05d3e9/docs/design/figures/dlo/dlo_no_allgather_shared_host.svg) distinguishes the one physical host payload per TP coordinate from process-local mappings, registration state, hooks, streams, and GPU slots.

See the separate [no-AllGather design](https://github.com/vllm-project/vllm-omni/blob/6bcd350301d71fb69aca4c82ec29dee75e05d3e9/docs/design/feature/offloader/distributed_layerwise_offload.md#no-allgather-design).

### TP-coordinate runtime host layout (v1)

The cache identity represents the final loader-produced host layout, not the DLO transfer topology. It combines the final tensor content digest with loader/model inputs and implementation fingerprints.

- TP world size/rank: included; each TP coordinate has a distinct entry
- DP size/rank: excluded; equivalent DP replicas may share an entry
- SP rank: excluded; equivalent SP ranks may share an entry
- SP world size/implementation: included conservatively as a compatibility guard
- PP, HSDP/DTensor, expert parallelism, and CFG parallelism: rejected in v1
- existing local model paths: expanded and resolved, so absolute, `~`, relative, and symlink spellings converge
- unsupported or process-unstable identity values: fail closed to ordinary host weights

The physical target is one cache entry per TP coordinate, shared by equivalent DP/SP consumers on a node. `RuntimeHostLayout` describes loader-produced tensor ownership and representation. `DLOTransferLayout` describes local versus collective movement and is not cached in v1; AllGather retains its existing process-owned host shards. This PR therefore implements a TP-coordinate runtime host layout, not a general parallel-layout cache.

Final names, kinds, shapes, dtypes, byte lengths, contiguous layout, and bytes remain authoritative. A loader or configuration change that changes the final bytes creates a different entry.

### Publication and failure behavior

- strict per-layout POSIX `flock`;
- stale temporary cleanup only while holding that lock;
- sharded safetensors plus JSON manifest;
- per-file SHA-256 and final mapped-tensor digest validation;
- disk-space preflight, `fsync`, and same-filesystem atomic directory rename;
- configurable writer wait and stable fallback codes.

A timeout, unsupported layout, publication error, or validation error retains the already-valid ordinary tensors. Host weight cache failure never leaves the model on `meta`.

### Host weight cache transfer

Explicit host-cache selection uses the existing pinned-memory policy. With `pin_cpu_memory=True`, the backend asks the platform registration backend to make the complete final mapping directly H2D-copyable. Setting `pin_cpu_memory=False`, or an unavailable registration backend, retains the bounded two-slot staging path. A positive `--dlo-host-weight-cache-pin-limit-gib` is only an optional safety ceiling; zero applies no additional ceiling. CUDA is the first implementation and:

1. selects the transform-complete final host weight cache, including at TP1;
2. page-aligns and coalesces source ranges by backing file;
3. when a positive ceiling is configured, verifies the complete byte count against it before any platform mutation;
4. requires CUDA read-only host-registration capability, then registers every range with the read-only flag;
5. verifies that every mapped tensor is pinned;
6. copies directly from mapped tensor views into the rotating device buffers.

Registration is all-or-nothing. Partial failure rolls back registered ranges and uses bounded staging. If rollback cannot complete, initialization aborts instead of unmapping memory CUDA may still own. Shutdown synchronizes outstanding device work, unregisters all ranges, then releases the safetensors mappings and distributed environment.

Raw checkpoint mmap is not registered because deferred checkpoint transforms, such as grouped-QKV reordering, would retain recurrent CPU work. The registered path deliberately uses final ordinary-loader tensors.

The DLO backend owns only a platform-neutral registration lifecycle. The CUDA backend queries `cudaDevAttrHostRegisterReadOnlySupported` before registering immutable mappings. Unsupported platforms or devices safely retain bounded staging; vLLM-Omni does not substitute writable or copy-on-write CUDA-specific mappings.

## Compatibility boundary

Compatible checkpoint mmap remains automatic for supported no-AllGather layouts. Final-layout host weight cache selection is explicit, while AllGather retains its existing process-owned host-shard path. Host weight cache v1 rejects quantization, non-contiguous tensors, tied/shared storage, device-only or distributed tensors, ambiguous ownership, HSDP, expert parallelism, CFG parallelism, and PP.

Host weight cache identity is defined per TP coordinate and includes conservative SP world-size and implementation guards. DP orchestration and independent-engine placement remain outside vLLM-Omni.

The cache targets steady-state node-local page sharing, not startup peak reduction: every worker still performs ordinary loading and full content validation. Sharing is node-local; v1 has no automatic eviction.

## Usage

```bash
vllm serve /path/to/model --omni \
  --tensor-parallel-size 2 \
  --dlo-no-use-allgather \
  --dlo-use-host-weight-cache \
  --dlo-host-weight-cache-pin-limit-gib 80
```

No-AllGather DLO automatically uses a compatible checkpoint mmap plan when the final-layout cache is not selected. When checkpoint mmap is unavailable, ordinary pinned loader weights remain the default. `--dlo-use-host-weight-cache` builds or joins the transform-complete final-layout cache and, under `pin_cpu_memory=True`, prefers registered direct H2D. By default, workers use `~/.cache/vllm-omni/dlo-host-weights`; programmatic configuration can override the cache root and writer lock timeout for advanced deployments. Workers that should share must see the same node-local, disk-backed directory. A positive pin limit is an optional ceiling that must cover the complete page-aligned layout; zero applies no additional ceiling. Registration remains capability-dependent and safely falls back to staging.

## Validation status

Historical hardware tables and fixed test counts are intentionally omitted because they can drift as the implementation changes. Revalidation must run against and identify the final commit, hardware topology, model and configuration, and measurement method before results are attached or used for performance claims. Track the final matrix and commit-pinned artifacts in #6231.


vllm_omni/diffusion/model_loader/diffusers_loader.py
```diff
@@ -516,7 +518,107 @@ def load_model(
...
+            "lora_backend": getattr(self.od_config, "lora_backend", None),
+        }
+        sp_guard = {
+            "sequence_parallel_size": int(getattr(parallel_config, "sequence_parallel_size", 1)),
+            "ulysses_degree": int(getattr(parallel_config, "ulysses_degree", 1)),
+            "ring_degree": int(getattr(parallel_config, "ring_degree", 1)),
+            "allgather_degree": int(getattr(parallel_config, "allgather_degree", 1)),
+            "ulysses_mode": getattr(parallel_config, "ulysses_mode", "strict"),
+        }
+        model_identity = getattr(self.od_config, "model", None)
```
hsliuustc0106, line 597: The runtime-cache identity takes the raw `model` string here, while `cache_root` is canonicalized with `expanduser().resolve()` (`runtime_weight_cache.py:628`). Two engines loading the same weights via different-but-equivalent path spellings (`~/models/h3` vs `/home/u/models/h3`, a symlink, a relative path) get different cache keys, so each publishes its own full-size entry and node-local sharing silently doesn't happen — the primary deployment this targets. Since the final content digest is already authoritative in the key, canonicalizing local paths (`os.path.realpath(os.path.expanduser(...))`) can only remove duplicate entries, never share incompatible weights.

Related identity-stability note: `_json_normalize`'s last-resort fallback is `str(value)` (`runtime_weight_cache.py:246-249`). A `loader_inputs` leaf that isn't a primitive/dataclass/pydantic/Mapping and whose repr embeds an object address would make the key process-unstable: silent perpetual miss plus a fresh entry on every engine start. `model_config: dict[str, Any]` holds primitives today — narrowing the accepted leaf types (or asserting determinism) would make that failure mode unreachable.

hsliuustc0106: Fixed in `233b5972`. Existing local model paths are now expanded and resolved before key construction, so absolute, `~`, relative, and symlink spellings converge while non-local model IDs remain unchanged. `_json_normalize` now rejects unsupported values and non-string mapping keys with the fail-closed `unstable_identity` code instead of serializing a potentially address-bearing `str(value)`. Added regression coverage for both cases.

docs/user_guide/diffusion/offloader/distributed_layerwise_offload.md
```diff
@@ -62,6 +80,10 @@ omni = Omni(
 | `--data-parallel-size N` | DP ranks and AllGather weight-sharding group | `1` |
 | `--dlo-use-allgather` | Shard host weights and reconstruct with AllGather | `true` |
 | `--dlo-no-use-allgather` | Stream complete rank-local blocks without a DLO weight collective | `false` |
+| `--dlo-enable-runtime-cache` | Normalize final ordinary-loader DiT weights into a node-local mmap cache; requires no-AllGather | `false` |
```
hsliuustc0106: do we need to add so many cli args? this will bring difficulty for users

docs/user_guide/diffusion/offloader/distributed_layerwise_offload.md
```diff
@@ -5,7 +5,11 @@ deployments. With AllGather enabled, each rank stores roughly `1 / dp_size` of
 the host weights and reconstructs each layer at runtime. Without AllGather,
```
hsliuustc0106: the user guide should be concise and not overlap with feature design doc

vllm_omni/diffusion/offloader/cuda_host_registration.py
```diff
@@ -0,0 +1,205 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+"""Bounded CUDA registration for existing read-only host mappings."""
```
hsliuustc0106: can we make it platfrom agnostic

vllm_omni/diffusion/model_loader/runtime_weight_cache.py
```diff
@@ -0,0 +1,724 @@
...
+
+from vllm_omni.diffusion.model_loader.host_weight_plan import (
+    HostWeightPlan,
+    HostWeightPlanResult,
+    TensorBinding,
+)
+
+logger = init_logger(__name__)
+
+RUNTIME_CACHE_SCHEMA_VERSION = 1
```
hsliuustc0106: why we need these env vars?

hsliuustc0106:
Re: https://github.com/vllm-project/vllm-omni/pull/6240#discussion_r3792954414 — verified at `233b5972`. `_canonicalize_existing_local_path` converges absolute, `~`, relative, and symlink spellings only when the path exists (non-existent repo IDs pass through untouched), and `_json_normalize` now fails closed with `unstable_identity` for unknown types and non-string mapping keys instead of the `str(value)` fallback — raised inside the plan builder's try, so it degrades to ordinary tensors like every other fallback code. `test_equivalent_local_model_paths_reuse_one_entry` (four spellings → one entry) and `test_unstable_loader_identity_fails_closed` cover both. Resolved.

Re-review note for `f1abb1395` (registered direct H2D): reviewed the full delta — all-or-nothing page-aligned registration with budget checked before any CUDA call, read-only flag on immutable pages, `is_pinned()` post-condition with rollback, CleanupError aborting rather than unmapping ranges CUDA may own, unregister-before-unmap on disable, and worker-shutdown ordering (offloader → kv prefetch → distributed env, tested). The no-AllGather transfer group is `None` in the modes where registration applies, so the DP-group prefetch branch can't reach registered mode without staging buffers. No findings.

vllm_omni/diffusion/model_loader/host_weight_cache.py
```diff
@@ -0,0 +1,724 @@
...
+def _resolve_pipeline_tensor(pipeline: nn.Module, runtime_name: str) -> torch.Tensor | None:
+    parent_path, _, leaf_name = runtime_name.rpartition(".")
+    parent = pipeline.get_submodule(parent_path)
+    tensor = parent._parameters.get(leaf_name)
+    if tensor is None:
+        tensor = parent._buffers.get(leaf_name)
+    return tensor
+
+
+def _collect_runtime_tensors(
```
hsliuustc0106, line 120: add docs for this function, I cannot understand

vllm_omni/diffusion/model_loader/runtime_weight_cache.py
```diff
@@ -0,0 +1,724 @@
...
+    candidate = Path(original).expanduser()
+    try:
+        if candidate.exists():
+            return str(candidate.resolve())
+    except OSError:
+        pass
+    return original
+
+
+def _json_normalize(value: Any) -> Any:
```
hsliuustc0106: can we move it to utils in this folder?

do not use Any type if possible

docs/user_guide/diffusion/offloader/distributed_layerwise_offload.md
hsliuustc0106: Condensed the user guide from 262 to 100 lines in `1a86e135`. It now focuses on mode selection, usage, relevant options, operational notes, scheduling constraints, and limitations; loader/cache mechanics and benchmark detail remain in the feature-design document.

docs/user_guide/diffusion/offloader/distributed_layerwise_offload.md
hsliuustc0106: Reduced the runtime-cache CLI additions from four controls to two in `1a86e135`. The cache directory and writer lock timeout remain advanced programmatic configuration fields. Cache enablement and the registration budget stay separate because shared mmap is useful with a zero pin budget, while direct H2D requires an explicit per-worker page-lock budget.

vllm_omni/diffusion/offloader/cuda_host_registration.py
hsliuustc0106: Added a platform-neutral `HostRegistration` lifecycle, shared errors, and dispatch factory in `1a86e135`. The DLO backend no longer imports the CUDA implementation directly. CUDA is the first backend; unsupported platforms cleanly retain bounded host staging, covered by fallback and dispatch tests.

vllm_omni/diffusion/model_loader/runtime_weight_cache.py
hsliuustc0106: These are internal cache-format and I/O defaults rather than environment variables. I added comments in `1a86e135` explaining the schema version, shared default cache root, bounded publisher wait, shard size, and hashing chunk sizes. The root and lock timeout remain configurable; format and chunk sizes stay private.

vllm_omni/diffusion/model_loader/host_weight_cache.py
hsliuustc0106: Added a detailed docstring in `1a86e135`: the collector verifies pipeline ownership, complete contiguous CPU storage, supported dtype, and absence of aliases inside or outside the DiT modules, then returns stable name ordering for hashing and publication.

vllm_omni/diffusion/model_loader/runtime_weight_cache.py
hsliuustc0106: Moved the identity/path/JSON helpers into `model_loader/runtime_cache_utils.py` in `1a86e135`. The helpers now accept `object` and return a recursive `JsonValue` type instead of using `Any`; unsupported values still fail closed through `IdentityNormalizationError`.

vllm_omni/diffusion/model_loader/host_weight_cache.py
```diff
@@ -0,0 +1,681 @@
+# SPDX-License-Identifier: Apache-2.0
```
hsliuustc0106, line 1: what's the main function of this script? can we move many of the functions into utils.py?

vllm_omni/diffusion/model_loader/runtime_weight_cache.py
```diff
@@ -0,0 +1,681 @@
...
+    if not temp_root.is_dir():
+        return
+    for path in temp_root.glob(f"{cache_key}.*"):
+        if path.is_dir():
+            shutil.rmtree(path, ignore_errors=True)
+        else:
+            path.unlink(missing_ok=True)
+
+
+def build_runtime_weight_cache_plan(
```
hsliuustc0106: is this the main function for cache runtime? if so, please explicitly add doc strings for this function

hsliuustc0106 (commented):
what if the host memory is not engouh? do we have a warning with fallback plan?

vllm_omni/diffusion/model_loader/host_weight_cache.py
hsliuustc0106: Fixed in `a1ad2408`. The module doc now identifies `build_runtime_weight_cache_plan()` as the single entry point. I moved the shared runtime tensor/error types plus nine reusable fingerprinting, hashing, locking, fsync, sharding, and stale-temp helpers into `runtime_cache_utils.py`; the entry-point module now retains tensor discovery, atomic publication, mmap validation, and orchestration.

vllm_omni/diffusion/model_loader/runtime_weight_cache.py
hsliuustc0106: Yes, this is the main runtime-cache entry point. Added an explicit docstring in `a1ad2408` covering when it runs, unsupported-layout rejection, content identity construction, existing-entry join versus locked publication, the returned mmap bindings, and fail-closed behavior that leaves ordinary-loader tensors intact.

lishunyang12:
## TP2 no-AllGather hardware validation (`d8db2f631990c5a15b7089e7aa039e368e397266`)

Tested the current head on:

- 2x NVIDIA B300 SXM6, PyTorch `2.13.0+cu130`, vLLM `0.27.0`;
- MiniMax-H3 `FL2VA`, BF16, `DP=1`, `TP=2`, `SP=1`, text-encoder TP2, VAE patch parallel 2;
- DLO enabled with `dlo_use_allgather=False`, zero resident layers, eager execution, and `CUDNN_ATTN`;
- 448x256 T2VA, 4 seconds, 2 inference steps, seed 2000.

Focused host-cache/registration/no-AllGather tests passed: **34 passed, 73 deselected**.

### Default bounded-staging path: PASS

With `dlo_host_weight_cache_pin_limit_gib=0` and an empty cache:

- direct checkpoint mmap correctly reported `unsupported_tp` and used the ordinary TP loader;
- two distinct cache entries were published, one for each TP coordinate:
  - TP rank 0: `e9e4143b4049`, 535 tensors, 33,140,803,648 tensor bytes;
  - TP rank 1: `166b40bcb550`, 535 tensors, 33,140,803,648 tensor bytes;
- both ranks remapped all 535 final DiT tensors and allocated two bounded pinned staging buffers;
- logs explicitly reported `streaming complete rank-local blocks (no DLO shard or AllGather)`; ordinary model TP collectives were still present as expected;
- generation completed with video shape `[107, 256, 448, 3]` and audio shape `[1, 2, 142400]`;
- request time was 71.3 s, reported peak GPU memory was 12.1 GiB, and shutdown completed with `active_children=[]`.

### Warm registered path: FAILS after advertised fallback

I reran against the same two entries with a 40 GiB per-worker registration budget. Both ranks correctly reused their TP-coordinate entries, but `cudaHostRegister(read-only)` returned `cudaErrorNotSupported` on this system/mapping.

The backend logged `using bounded host staging` and allocated the staging buffers, but the next ordinary CUDA operation (`torch.zeros` in MiniMax-H3 prompt distribution) failed with the same `CUDA error: operation not supported`, aborting the request. This sequence is consistent with the failed runtime API call leaving a pending CUDA error: `_error_message()` calls `cudaGetErrorString`, but the error path does not consume it with `cudaGetLastError()` before returning to PyTorch. Thus the positive-budget path does not currently satisfy its promised safe fallback when registration is unsupported.

The cache was placed on tmpfs because local disk was full and the large NFS mount was not writable, so this run validates TP2 cache identity/remapping and functional no-AllGather execution, but not disk-backed PSS sharing. The reason registration was unsupported may be specific to this mapping/environment; the failure to continue through the documented fallback is still reproducible on both ranks.

hsliuustc0106:
@lishunyang12 can you help test it again?

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
```diff
@@ -1748,6 +1993,7 @@ def disable(self) -> None:
 
         if self._using_rank_local_mmap:
             current_omni_platform.synchronize()
+        registration_released = self._release_registered_mmap()
```
hsliuustc0106, line 2018: [P1] Synchronize all DLO streams before teardown

`DiffusionWorker.shutdown()` now calls this path before `destroy_distributed_env()`, but `disable()` only calls `current_omni_platform.synchronize()` when `_using_rank_local_mmap` is true. In AllGather/DP mode that flag is false, while `prefetch_layer()` queues H2D and AllGather work on dedicated copy/communication streams. Removing hooks and destroying process groups here can race with in-flight work. Please synchronize the DLO streams (or all device work) for every mode before teardown.

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
```diff
@@ -1103,6 +1168,151 @@ def _load_weights_via_mmap(
...
+                    )
+
+                if binding.file_path not in file_cache:
+                    file_cache[binding.file_path] = safe_open(
+                        binding.file_path,
+                        framework="pt",
+                        device="cpu",
+                    )
+                tensor = file_cache[binding.file_path].get_tensor(binding.storage_key)
+                if tensor.shape != target.shape or tensor.dtype != target.dtype:
```
hsliuustc0106, line 1242: [P2] Revalidate cache bytes or fall back before remapping

The loader validates the manifest, file digest, and mapped-tensor content before returning the plan, but this later open checks only shape and dtype. A same-shape shard changed after plan creation is therefore consumed silently; I reproduced this locally by rewriting a shard to all 9s after plan creation. If mapping fails, `enable()` also raises instead of retaining the ordinary tensors that were already loaded. Carry the expected digest/manifest into the plan and revalidate here, or catch the failure and fall back before `set_tensor_storage()`.

vllm_omni/config/omni_config.py
```diff
@@ -576,6 +576,9 @@ class keeps the diffusion-only knobs from ``OmniDiffusionConfig`` without
     enable_layerwise_offload: bool = False
     enable_distributed_layerwise_offload: bool = False
     dlo_use_allgather: bool = True
+    dlo_host_weight_cache_dir: str | None = None
+    dlo_host_weight_cache_lock_timeout: float = Field(default=600.0, gt=0)
```
hsliuustc0106, line 693: [P2] Reject non-finite structured config values

The `gt/ge` constraints accept `+inf`, unlike `OmniDiffusionConfig.__post_init__`, which explicitly rejects non-finite values. I reproduced acceptance of `inf` for both fields locally. An infinite pin budget reaches `int(pin_limit_gib * 1024**3)` and raises `OverflowError`, while an infinite lock timeout defeats the bounded-wait contract. Add finite-value validation to this structured projection and regression tests.

tests/diffusion/model_loader/test_host_weight_cache.py
```diff
@@ -0,0 +1,248 @@
...
+    assert list(tmp_path.rglob("*.safetensors")) == []
+
+
+def test_concurrent_processes_publish_one_immutable_entry(tmp_path):
+    context = mp.get_context("spawn")
+    queue = context.Queue()
+    processes = [context.Process(target=_multiprocess_builder, args=(str(tmp_path), queue)) for _ in range(2)]
+    for process in processes:
+        process.start()
+    results = [queue.get(timeout=30) for _ in processes]
```
hsliuustc0106: [P2] Avoid a fixed 30-second spawn timeout

This test failed locally with `queue.Empty` under the targeted suite. A standalone probe showed both child processes eventually published/reused the entry and exited successfully, but only after about 36 seconds while importing the full vLLM package. The hard 30-second queue timeout makes the concurrency test flaky under normal startup contention. Please use a startup-aware/larger timeout or reduce the child import cost.

hsliuustc0106 (commented):
Local validation at 31b47e0 found the actionable issues in the inline comments below. I did not run a full MiniMax-H3 TP2 inference on this head.

vllm_omni/diffusion/model_loader/diffusers_loader.py
```diff
@@ -516,7 +533,105 @@ def load_model(
                 logger.info("Quantization complete, offloaded model back to CPU")
 
         self._apply_skip_softmax_calibration(model)
-        return model.eval()
+        model.eval()
+        self._maybe_build_host_weight_cache(model, load_format)
```
hsliuustc0106, line 540: [P1] Preserve an opt-out from the known slow cache path

Every no-AllGather load without a checkpoint mmap reaches this call. With the default pin budget of `0`, TP>1 users are automatically moved to pageable mmap plus recurrent CPU staging; the #6231 A/B measured 8.96/12.30 s versus 3.50/3.60 s request latency and 69.3% lower throughput, with HBM unchanged. The registered path on this head is still unmeasured. Preserve ordinary pinned weights unless cache/registration is explicitly selected, or provide final-head TP1 and dual-TP2 A/B covering startup, PSS/`Private_Dirty`/locked bytes, peak HBM, latency/throughput, and output equality.

hsliuustc0106 (commented):
Focused host-memory, HBM, and latency re-review at `be8e0b5`.

david6666666:
## Local code review and TP=2 validation (`1a9efdc1`)

### Verdict

- **PASS** for cache publish, final-layout mmap remap, warm cache reuse, bounded-staging fallback, inference, and clean teardown.
- **Direct H2D was not exercised on this host.** Both ranks returned `cudaHostRegister(read-only) failed: operation not supported`; the current code consumed the CUDA error and safely fell back to two bounded host staging buffers per rank. The server remained healthy and produced a correct response.
- I found **no additional blocking correctness issue** in the current head.

### Code review

I traced the complete config/CLI -> async engine -> stage config -> DLO loader/offloader flow and reviewed the cache identity, atomic publish, manifest/digest validation, mmap lifetime, CUDA registration, fallback, and worker shutdown paths. In particular, the current head:

- keeps the cache explicitly opt-in (including selection by a positive pin budget);
- rejects non-finite cache/pin-limit values;
- revalidates runtime content before remapping and validates all mappings before mutating the pipeline, retaining ordinary tensors on failure;
- consumes handled CUDA registration errors and aborts unsafe cleanup instead of unmapping live registrations;
- synchronizes every registration mode before teardown and disables the backend before distributed shutdown.

This also appears to address the earlier review findings around implicit enablement, digest revalidation, non-finite config, sticky CUDA errors, and mmap-only teardown synchronization.

Non-blocking operational observations from this run:

- TP=2 created two coordinate-specific entries, each containing 535 tensors / 33,140,803,648 tensor bytes in seven safetensors files (62 GiB total on disk). There is no automatic eviction, as documented.
- On hardware where read-only host registration is unsupported, opting into the cache is materially slower for this tiny two-step request because the fallback copies through bounded staging. The default path is unaffected because the feature is opt-in.

### Environment

- 2x NVIDIA B300 SXM6 AC selected from an 8-GPU host; driver `610.43.02`
- Python from `/home/zjy/code/david/.venv`
- PyTorch `2.13.0+cu130`, vLLM `0.27.0`
- PR worktree injected with `PYTHONPATH`; HEAD `1a9efdc1d28022926971a5b9c11b065a266366b1`
- Model: local MiniMax-H3 `FL2VA`; TP=2, USP=1, ring=1, DLO enabled, no AllGather, zero resident layers

### Serve and request commands

Feature run (a positive pin limit opts into the host cache and requests the direct-H2D registration path):

```bash
source /home/zjy/code/david/.venv/bin/activate
export PYTHONPATH=/home/zjy/code/david/worktree/vllm-omni-pr6240-review
export CUDA_VISIBLE_DEVICES=0,1
export VLLM_OMNI_VIDEO_SYNC_TIMEOUT=14400
export VLLM_OMNI_SERVER_STORAGE__PATH=/home/zjy/code/david/tmp/pr6240-validation-1a9efdc/server-storage
export FLASHINFER_DISABLE_VERSION_CHECK=1

vllm serve /home/zjy/code/david/tmp/h3-prverify/model-h3 \
  --omni --host 127.0.0.1 --port 8624 --trust-remote-code \
  --task-type fl2va --num-gpus 2 --tensor-parallel-size 2 \
  --usp 1 --ring 1 --text-encoder-tp-size 2 \
  --vae-patch-parallel-size 2 --vae-parallel-mode tile --vae-use-tiling \
  --enable-distributed-layerwise-offload --dlo-no-use-allgather \
  --dlo-resident-layers 0 --enforce-eager \
  --diffusion-attention-backend CUDNN_ATTN \
  --dlo-host-weight-cache-pin-limit-gib 40 \
  --stage-init-timeout 1800 --init-timeout 1800
```

Baseline used the same serve command with `--dlo-host-weight-cache-pin-limit-gib 40` omitted. I also restarted once with `--dlo-use-host-weight-cache` (and no pin budget) to verify reuse of both published entries.

Request used for both measured runs:

```bash
curl -sS -D headers.txt -o output.mp4 --max-time 14400 \
  -w '\nhttp_code=%{http_code} time_total=%{time_total}\n' \
  -X POST http://127.0.0.1:8624/v1/videos/sync \
  -F 'prompt=At night, three cats march into a bedroom playing tiny brass instruments, then abruptly file out, with synchronized room ambience.' \
  -F 'width=448' -F 'height=256' -F 'aspect_ratio=16:9' \
  -F 'fps=24' -F 'num_inference_steps=2' -F 'flow_shift=12' \
  -F 'seed=2000' -F 'quality=lossless' \
  -F 'extra_params={"task":"t2va","duration":4.0,"audio_flow_shift":3.0}'
```

### Local test results

Targeted tests:

```bash
pytest -q \
  tests/diffusion/model_loader/test_diffusers_loader.py \
  tests/diffusion/model_loader/test_host_weight_cache.py \
  tests/diffusion/offloader/test_cuda_host_registration.py \
  tests/diffusion/offloader/test_distributed_layerwise_backend.py \
  tests/diffusion/test_worker_wrapper_base.py \
  tests/entrypoints/test_async_omni_diffusion_config.py
```

Result: **290 passed, 1 skipped** in 24.10 s. `git diff --check` also passed, and the PR's current remote CI is green.

| Result | Baseline (ordinary host tensors) | Host cache + requested registration |
|---|---:|---:|
| HTTP | 200 | 200 |
| Server inference header | 13.791 s | 54.696 s |
| curl total | 13.794398 s | 54.700501 s |
| Peak-memory header | 12,100 MB | 12,100 MB |
| MP4 size | 107,336 B | 107,336 B |
| MP4 SHA-256 | `b2517a8ac4634cf0013ff6b2030dde380d8b3ac420d034aa5c24e77ba6a142b6` | same |

The fallback run was 3.97x slower for this two-step request (74.8% lower throughput). Startup times are not compared because the runs had different filesystem page-cache state.

Warm restart evidence: both TP entries were logged as reused (`5abb4d53a9e5...` and `46760d9b26c0...`), 535 tensors per rank were remapped, and all mmap handles were released during graceful shutdown. No serving process, listening port, or GPU allocation remained afterward.

### Precision comparison

- The two encoded MP4 files are byte-for-byte identical (`cmp` exit 0 and identical SHA-256 above).
- Decoded RGB video SHA-256: `4ae511263195fb98112d4faf2c93594b0daef68d7f4fdce199a5e047b58ce812` for both.
- Decoded float32 audio SHA-256: `e92c9eafd2a720197599cb36cbf7cea88c2337dc8c18e26e954ab939c8aa9544` for both.
- FFmpeg video comparison: PSNR Y/U/V/average = `inf`; SSIM All = `1.000000`.
- Output probe: H.264 448x256 at 24 fps, 107 frames / 4.458333 s; AAC 32 kHz stereo / 4.45 s.

So the cache/fallback path is precision-identical to the baseline for this deterministic request. A host that supports the read-only CUDA registration mode is still needed to validate the actual direct-H2D path rather than its safe fallback.

vllm_omni/diffusion/offloader/cuda_host_registration.py
```diff
@@ -0,0 +1,245 @@
...
+        except Exception as exc:
+            raise HostRegistrationError(f"cannot access the CUDA runtime: {exc}") from exc
+
+        registered: list[_AddressRange] = []
+        try:
+            for region in regions:
+                error = runtime.cudaHostRegister(
+                    region.start,
+                    region.size,
+                    _CUDA_HOST_REGISTER_READ_ONLY,
```
david6666666, line 214: **Compatibility issue reproduced on B300: please add a direct-H2D-compatible fallback.**

This unconditional cudaHostRegisterReadOnly makes the direct path unavailable when cudaDevAttrHostRegisterReadOnlySupported is 0. On our NVIDIA B300 / driver 610.43.02 / CUDA 13.4 host, the device attributes were HostRegisterSupported=1, PageableMemoryAccessUsesHostPageTables=0, and HostRegisterReadOnlySupported=0. Both TP ranks therefore returned cudaErrorNotSupported (801) here and fell back to bounded staging.

A 4 KiB local reproducer showed:
- read-only mmap + cudaHostRegisterReadOnly -> cudaErrorNotSupported (801)
- the same read-only mmap + cudaHostRegisterDefault -> cudaErrorInvalidValue (1)
- writable-private/COW mmap + cudaHostRegisterDefault -> success; unregister -> success

Please add capability-aware registration rather than treating this hardware as staging-only: query cudaDevAttrHostRegisterReadOnlySupported before selecting the cache mapping/flag and, when it is false, use a writable-private/COW mapping with cudaHostRegisterDefault (or an equivalent compatible mapping). Please preserve the current error-consumption/rollback guarantees and add a test for this capability branch.

Without this, a positive pin budget cannot exercise the direct-H2D path on B300 and pays the staging cost (54.70 s versus 13.79 s for the same two-step request in my run), even though ordinary host registration is supported.

hsliuustc0106: Addressed the capability gap in `390b61d7`. The CUDA backend now queries `cudaDevAttrHostRegisterReadOnlySupported` before attempting registration; devices such as this B300 return a clear unsupported result and retain the existing bounded-staging fallback without issuing the failing CUDA call. I did not add a writable-private/COW mapping: that changes the immutable host-layout contract and may create per-worker private pages during registration, defeating the shared-cache goal. A generally safe way to register immutable file-backed mappings belongs in the CUDA/PyTorch platform layer rather than a vLLM-Omni-specific mapping workaround. Added tests for capability absence and query-error consumption, and documented that a positive budget requests—but cannot guarantee—direct H2D.

vllm_omni/diffusion/model_loader/host_weight_plan.py
```diff
@@ -40,6 +40,9 @@ class HostWeightPlan:
     backing_kind: str
     bindings: dict[str, TensorBinding]
     planned_source_prefixes: frozenset[str] = frozenset()
+    runtime_layout_key: str | None = None
+    post_load_complete: bool = False
```
yuanheng-zhao: The scenarios of `checkpoint_mmap`+`post_load_complete True` and `host_weight_cache` + `post_load_complete False` seem never come across.

The importing of `post_load_complete` seems to make current branches too complicated. We might want to clean up a bit, maybe to find some way eliminating `post_load_complete` and organizing the these structs and attrs

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
```diff
@@ -1625,6 +1868,17 @@ def enable(self, pipeline: nn.Module) -> None:
...
+            if (
+                self.enabled
+                and self._using_rank_local_mmap
+                and self.host_weight_plan.backing_kind == "host_weight_cache"
+            ):
+                self._using_registered_mmap = self._try_register_host_weight_cache_mmap()
+                self._host_weight_cache_mapped_sources.clear()
+                assert self._resident_layer_group is not None
+                self._resident_layer_group.registered_mmap = self._using_registered_mmap
+                if self._using_registered_mmap:
```
yuanheng-zhao: This part seems to be nearly duplicately implemented with the following branch `if self._using_rank_local_mmap and self.host_weight_plan.backing_kind == "host_weight_cache":`? There're several exit paths and we might prevent this redundancy

hsliuustc0106:
## Current-head registered-mmap validation

Validated commit `6bcd350301d71fb69aca4c82ec29dee75e05d3e9` against a matched private-pinned control on four NVIDIA L20X GPUs.

| Parallelism | Pairs | Private-pinned mean | Registered-mmap mean | Paired delta | 95% paired t interval |
| TP1 | 10 | 3.6855 s | 3.7063 s | +20.8 ms (+0.56%) | [-68.5, +110.2] ms |
| TP2 | 10 | 3.1354 s | 3.1631 s | +27.6 ms (+0.88%) | [-107.3, +162.5] ms |

Both intervals cross zero. The supported conclusion is **no measurable steady-state regression in this smoke workload**, not a tight equivalence claim: with ten pairs, the upper interval bounds are approximately +3.0% for TP1 and +5.2% for TP2.

### Method and correctness

- MiniMax-H3 FL2VA, CUDNN attention, eager mode, 256×256, two diffusion steps, four-second output, `dlo_resident_layers=0`.
- Two warmups per engine, followed by ten paired waves per TP degree.
- Device assignments were swapped halfway through to control for GPU placement.
- All selected GPUs and more than 99.88% of sampled resident pages were on NUMA node 0.
- Monitoring took one initialized-state snapshot and remained quiescent during warmup and measurement.
- All 20 paired outputs had byte-identical audio/video hashes, shapes, dtypes, and seeds.
- Every engine and monitor returned 0; no timed run timed out.

### TP2 transfer diagnosis

The registered path preserves H2D payload but issues 480 additional tensor-level H2D copies per rank after the dominant TP broadcast. Device H2D time increases by approximately 32 ms per rank, while profiled compute is unchanged. This explains the small positive point estimate and identifies transfer coalescing/packed cache layout as a possible follow-up; the paired steady-state run does not detect an end-to-end TP2 regression.

The profile also confirms that this is not a new DLO collective. The additional TP wait is ordinary rank skew around the existing TP broadcast.

### Memory

| Metric | Private pinned | Registered mmap | Change |
| --- | ---: | ---: | ---: |
| TP2 process-tree PSS | 181.02 GiB | 140.08 GiB | -40.94 GiB (-22.6%) |
| TP2 Private_Dirty | 180.34 GiB | 77.67 GiB | -102.67 GiB (-56.9%) |
| Per-request peak HBM | 11,884 MiB | 11,884 MiB | no change |

### Separate startup limitation

Cache-hit startup remains expensive. Ordinary loading, post-load work, and validation took 263.57 s, while registering 61.73 GiB itself took 4.45 s. This is a separate cache-hit loading/validation issue and is not part of the timed inference comparison.

Some private-control shutdowns logged the existing orchestrator cleanup timeout after all measured waves and summaries completed. These warnings did not alter return codes or timed results.

hsliuustc0106:
## Architectural review: make `HostWeightStore` the reusable boundary

I reviewed the frozen head `58eebee14b5dac0206a93f765fe065a1382ed287`. CI is green, `git diff --check` passes, and the focused cache/loader/registration tests passed locally (51 tests plus 12 backend cache/registration tests). I also checked the existing review threads and hardware reports; I do not have a new direct-H2D correctness finding beyond those already discussed.

My assumption is that the host cache added here is intended to become the common node-local backing for follow-up DLO work, especially #6374 and #6399. If the intent is instead a deliberately private, no-AllGather-only implementation, we should state that and avoid treating it as the shared cache foundation. Given the active follow-ups, I recommend the shared-foundation interpretation.

### The current lookup contract is too late for a reusable store

The main issue is in `build_host_weight_cache_plan()`:

1. the ordinary loader has already materialized and finalized the complete DiT;
2. `_collect_runtime_tensors()` walks those live tensors;
3. `runtime_content_digest(records)` hashes all final bytes;
4. that digest is placed in `layout_identity` and therefore in the cache key;
5. even a cache hit is validated against `expected_records`, including another mapped-content digest.

So a worker cannot know which entry to open until it has already paid for the complete ordinary load and full content hash. This is correct for the narrow steady-state page-sharing goal documented by this PR, but it is not a reusable warm-start cache contract. The current L20X report makes the consequence visible: cache-hit ordinary loading/post-load/validation took 263.57 s, while registration itself took 4.45 s.

It also differs from the contract already written in #6231: the **pre-build coordination identity** and **post-build manifest** are separate concerns. Peers need a stable key before the producer runs; the manifest proves the bytes that were actually published. In the current code those concerns are collapsed by putting `runtime_content_sha256` into the lookup key.

This matters immediately for #6399: its value comes from looking up a canonical FP8 artifact *before* BF16 materialization and online quantization. It cannot reuse the current API without giving up its warm-start benefit, so it is likely to grow a second implementation of locking, atomic publication, corruption recovery, manifests, mmap ownership, and fallback behavior.

### Proposed ownership boundary

I suggest extracting a small, representation-neutral `HostWeightStore` contract:

- `lookup(identity) -> HostWeightLease | None`: resolve a pre-build identity, validate the manifest/artifact, and return immutable mapped tensors without requiring a live model;
- `get_or_build(identity, producer) -> HostWeightLease`: elect one builder, publish atomically, let peers join, and preserve stable fallback/error codes;
- `HostWeightLease`: own mmap/file-handle lifetime plus tensor bindings, metadata, file/content digests, and artifact identity;
- `invalidate()/inspect()`: bounded corruption recovery and observable hit/miss/build/fallback state;
- no eviction or cross-node distribution in v1.

The producer should own how runtime bytes are created:

- #6240: ordinary-loader final BF16/FP16 TP-coordinate layout;
- #6399: canonical FP8 qweights/scales produced with bounded scratch;
- future formats: another producer/schema namespace, without another cache runtime.

The transport/offloader should continue to own everything device- and execution-specific:

- CUDA/platform host registration;
- pinned staging slots;
- streams, events, direct H2D;
- chunking and AllGather;
- device buffers and teardown ordering.

In other words, the store returns a lease over immutable host backing; the transport decides how to move that backing. The store should not import DLO hooks or CUDA registration, and the transport should not parse manifests or take publication locks.

### Relationship to the follow-up PRs

- **#6240 should provide the store foundation.** Its existing identity normalization, locking, sharded safetensors publication, validation, fallback codes, and mmap work are good raw material. The no-AllGather feature then becomes one producer plus one consumer of the store.
- **#6399 should depend on the store API.** It needs a different producer and artifact schema, but should not own a second node-local cache runtime.
- **#6374 should depend only on the lease/plan interface, not the cache implementation.** It is a transport enhancement. It should be able to consume ordinary tensors, checkpoint mmap, or a store lease and then construct its chunk manifest/local shards. Registration and chunked H2D/AllGather stay on that side of the boundary.

### Requested change before merge

At minimum, I think this PR should separate:

1. a pre-build key derived from model/revision/checkpoint fingerprint, loader/producer version, component ownership, dtype/quantization/layout policy, and TP/SP guards;
2. a published manifest containing the authoritative final tensor metadata and content digest;
3. a lease that keeps validated mappings alive independently of the DLO backend;
4. the current ordinary-loader path as a producer invoked only on a miss.

Please add a test where a warm hit does **not** call the producer or materialize/hash the ordinary runtime tensors, plus the existing concurrent-one-builder, corruption, and lease-teardown coverage.

If introducing that boundary makes this PR too large, the cleaner sequence is to split out the store foundation first, land #6240's no-AllGather producer/consumer on it, and keep direct registration as an orthogonal transport change. I would avoid merging the current one-shot builder as the shared abstraction and then asking #6399 to work around it; that would lock in two cache runtimes just as the DLO follow-up work is converging.

PR #6241, opened 2026-08-16
[Perf] Enable async schedule for MOSS-TTS

Sy0307:
## Purpose

Improve `OpenMOSS-Team/MOSS-TTS-Local-Transformer-v1.5` serving throughput while keeping the implementation confined to MOSS-TTS Local model code, and keep request-level sampling overrides in the MOSS adapter by reusing the existing generic helper.

This PR:

- Keeps Stage-0 code rows fixed-shape, avoiding CUDA boolean-index output-shape synchronization, and opts this model into the existing asynchronous Omni output snapshot path.
- Reuses the shared `apply_max_new_tokens` helper from the MOSS adapter so an explicit speech request `max_new_tokens` is honored without adding a branch to the central serving module.
- Preserves prefill behavior when the scheduler presents a one-token prompt chunk.
- Precomputes and reuses the Local Depth RoPE values for positions `0..11`.
- Runs nucleus filtering, softmax, and multinomial over the retained `top_k` candidates, then maps the sample back to the 1024-token vocabulary.

The RoPE cache and compact sampler are direct implementation paths, not process-level feature switches. Compact top-k preserves the intended categorical distribution when the top-k boundary has no ties, but it is not seed- or bit-equivalent to sampling a width-1024 tensor because multinomial maps the same RNG state differently at width 25 and finite-precision boundary ties can select a different retained set.

### Environment and provenance

- GPU: 1x NVIDIA H200, Stage 0 and Stage 1 colocated
- Driver: 580.173.02
- CUDA: 13.0
- PyTorch: 2.13.0+cu130
- vLLM: 0.27.0
- vLLM-Omni package version: 0.26.0
- Base commit: `baba7d1e2ef9d326486493146b43157f7cd4f8d0`
- Measured optimization commit: `1797c7290c5797eebd033b13947a54dc875eec4e`
- Current PR commit: `226762e2fc62478d25796c8258e8d335d95b973c`
- Current PR tree: `97b335fe856bf7fe8bfc75db55dd9f92b5026ef0`
- Model revision: `be7766a6735b98bd793f7c79fb720b4d0f5d13b8`
- Offline loading: `HF_HUB_OFFLINE=1`, `TRANSFORMERS_OFFLINE=1`

The measured service enabled compact top-k, used the default-on RoPE cache, and sent `--output-len 512`. The current head makes those optimized paths direct and replaces the model-local frame counter with the MOSS adapter's use of the generic `max_new_tokens` helper requested in review. The saved end-to-end table below is retained as optimization evidence rather than relabeled as an exact-current-tree rerun.

The vLLM-Omni 0.26 / vLLM 0.27 compatibility warning was present on both A/B services.

### Throughput command

The 64-row Seed-TTS EN subset was repeated eight times. A separate 512-request c64 pass first warmed reference codes. The measured pass used 2048 requests at c64 with streaming 48 kHz stereo output.

```bash
HF_HUB_OFFLINE=1 \
TRANSFORMERS_OFFLINE=1 \
SEED_TTS_EVAL_DEVICE=cpu \
VLLM_OMNI_BENCH_AUDIO_SAMPLE_RATE=48000 \
VLLM_OMNI_BENCH_AUDIO_CHANNELS=2 \
python benchmarks/tts/bench_tts.py \
  --host 127.0.0.1 \
  --port 8123 \
  --model OpenMOSS-Team/MOSS-TTS-Local-Transformer-v1.5 \
  --task voice_clone \
  --locale en \
  --dataset-path /root/vllm-omni-workspace/profile-data/seed64x8 \
  --num-prompts 2048 \
  --concurrency 64 \
  --output-len 512 \
  --output-dir /tmp/moss-tts-c64
```

### Focused validation on the current tree

```text
MOSS model tests: 6 passed, 14 warnings in 0.09s
MOSS adapter tests: 2 passed, 23 deselected, 15 warnings in 1.14s
ruff check: passed
ruff format --check: passed
git diff --check: passed
custom pre-commit checks: passed
```

The focused tests cover fixed-shape output and mixed-batch stop routing, one-token prefill behavior, RoPE numerical parity and storage reuse, and compact top-k candidate constraints.

### Paired quality gate

A full Seed-TTS EN paired gate evaluated 1088 utterances with the original and compact samplers.

| Metric | Original sampler | Compact sampler |
|---|---:|---:|
| WER | 3.8522% | 3.3838% |
| WavLM SIM | 0.867135 | 0.867457 |
| UTMOS | 3.360601 | 3.359941 |
| Mean audio duration | 4.3746 s | 4.3283 s |
| Requests evaluated | 1088/1088 | 1088/1088 |
| PCM / ASR / evaluator failures | 0 | 0 |

Paired bootstrap 95% confidence intervals for WER, SIM, UTMOS, duration, RMS, peak amplitude, and clipping all include zero. Neither side produced duplicate PCM groups, streaming discontinuities, or frame-cap-length output.

### End-to-end c64 A/B

| **H200 Configuration** | **Concurrency** | **Baseline req/s** | **Current req/s** | **req/s Improvement** | **Baseline audio-s/s** | **Current audio-s/s** | **Audio Improvement** | **RTF Baseline → Current** | **WER Baseline → Current** |
|---|---:|---:|---:|---:|---:|---:|---:|---|---|
| 1 GPU | 1 | 1.860 | 2.119 | **+13.9%** | 7.860 | 9.171 | **+16.7%** | 0.130 → 0.112 (**-13.8%**) | 3.14% → 3.20% (**+0.06 pp**) |
| 1 GPU | 8 | 8.430 | 11.075 | **+31.4%** | 36.400 | 47.688 | **+31.0%** | 0.220 → 0.169 (**-23.2%**) | 3.14% → 2.67% (**-0.47 pp**) |
| 1 GPU | 16 | 12.370 | 16.968 | **+37.2%** | 52.840 | 72.152 | **+36.5%** | 0.300 → 0.224 (**-25.3%**) | 3.06% → 3.19% (**+0.13 pp**) |
| 1 GPU | 32 | 16.410 | 25.374 | **+54.6%** | 70.220 | 109.342 | **+55.7%** | 0.460 → 0.294 (**-36.1%**) | 4.25% → 3.99% (**-0.26 pp**) |
| 1 GPU | 64 | 21.880 | 34.239 | **+56.5%** | 93.020 | 149.684 | **+60.9%** | 0.680 → 0.418 (**-38.5%**) | 4.28% → 3.05% (**-1.23 pp**) |

### Isolated compact-sampler profile

With the other MOSS runtime fixes held constant, compact sampling improved request throughput by 6.95% and audio throughput by 6.06%:

- Stage-0 kernel time: 5.440 s -> 4.575 s (-15.9%)
- Local sort kernels: 242.0 ms -> 103.5 ms
- Softmax kernels: 137.1 ms -> 42.3 ms
- Multinomial/scan kernels: 116.9 ms -> 57.5 ms
- Steady allocated GPU memory: 105971 MiB -> 106135 MiB (+164 MiB, +0.15%)

## Scope decisions

The following experiments remain excluded:

- Reference singleflight and persistence: no steady-state benefit and broader cache-identity risk.
- Local Depth incremental KV cache: slower at the production shape.
- Stage-1 asynchronous D2H and batched waveform: broader shared-path changes with lower measured throughput.
- Dynamic Local Depth compilation: normalized audio-throughput gain below 5% without the same paired quality closure.
- Stage-1 arbitrary-mask attention rewrites: FlashAttention rejects the current mask, while chronological repacking was slower.

### DIFF
--- a/vllm_omni/model_executor/models/moss_tts/modeling_moss_tts_talker.py
+++ b/vllm_omni/model_executor/models/moss_tts/modeling_moss_tts_talker.py
@@ -6,7 +6,8 @@
 from __future__ import annotations
 
 import copy
-from collections.abc import Iterable, Sequence
+import os
+from collections.abc import Iterable, Mapping, Sequence
 from typing import Any
 
 import torch
@@ -1268,6 +1269,8 @@ class MossTTSLocalTalkerForGeneration(nn.Module):
     has_preprocess: bool = True
     has_postprocess: bool = True
 
+    _DEFAULT_MAX_AUDIO_FRAMES = 512
+
     def __init__(self, *, vllm_config: VllmConfig, prefix: str = "") -> None:
         super().__init__()
         self.vllm_config = vllm_config
@@ -1283,6 +1286,25 @@ class MossTTSLocalTalkerForGeneration(nn.Module):
 
         self.audio_assistant_slot_token_id: int = int(self.config.audio_assistant_slot_token_id)
         self.im_end_token_id: int = int(self.config.im_end_token_id)
+
+        max_audio_frames_raw = os.getenv(
+            "MOSS_TTS_LOCAL_MAX_AUDIO_FRAMES",
+            str(self._DEFAULT_MAX_AUDIO_FRAMES),
+        ).strip()
+        try:
+            self.max_audio_frames = int(max_audio_frames_raw)
+        except ValueError as exc:
+            raise ValueError(
+                f"MOSS_TTS_LOCAL_MAX_AUDIO_FRAMES must be a non-negative integer, got {max_audio_frames_raw!r}"
+            ) from exc
+        if self.max_audio_frames < 0:
+            raise ValueError(
+                "MOSS_TTS_LOCAL_MAX_AUDIO_FRAMES must be non-negative; use 0 to disable the server-side safety cap"
+            )
+        logger.info(
+            "MOSS-TTS Local server-side audio-frame safety cap: %s",
+            self.max_audio_frames or "disabled",
+        )
 
         # Qwen3 backbone. get_text_config() is sufficient for vLLM to size
         # KV cache / heads correctly, as long as AutoConfig resolves to the
@@ -1323,7 +1345,11 @@ class MossTTSLocalTalkerForGeneration(nn.Module):
         self.mtp_hidden_size = hidden_size
         self.talker_mtp_graph_safe = True
         self.talker_mtp_output_key = ("audio_codes", "current")
-        self.use_async_omni_output = False
+        # ``make_omni_output`` keeps code rows fixed-shape and performs all
+        # state updates eagerly, so the runner can safely pack/snapshot them
+        # before CUDA-graph buffers are reused.  This moves the pinned D2H and
+        # output materialization off the Stage-0 execution thread.
+        self.use_async_omni_output = True
         self.eager_omni_postprocess_before_async_output = True
         self.omni_pooler_payload_include_hidden = False
         self.postprocess_uses_multimodal_outputs = False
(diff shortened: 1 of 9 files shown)


Sy0307:
PTAK @gcanlin

vllm_omni/model_executor/models/moss_tts/modeling_moss_tts_talker.py
```diff
@@ -1284,6 +1287,25 @@ def __init__(self, *, vllm_config: VllmConfig, prefix: str = "") -> None:
         self.audio_assistant_slot_token_id: int = int(self.config.audio_assistant_slot_token_id)
         self.im_end_token_id: int = int(self.config.im_end_token_id)
 
+        max_audio_frames_raw = os.getenv(
+            "MOSS_TTS_LOCAL_MAX_AUDIO_FRAMES",
```
gcanlin: We don't need this env var. Please cherry-pick this https://github.com/gcanlin/vllm-omni/commit/e2524db7121b4ea821bb9ab053fb7d78c9613dbe commit to support max_tokens.

gcanlin (commented):
And other env var `MOSS_TTS_LOCAL_DEPTH_ROPE_CACHE` and `MOSS_TTS_LOCAL_DEPTH_COMPACT_TOPK` can be removed as well.

gcanlin:
## Async schedule

Before:

image


After:

image

## Remove boolean-index

`codes[should_continue]` is waiting `should_continue`, which is computed by GPU. So it's leading to the sync waiting in cpu and then blocks CPU so that async schedule can't continue.

Before:

image


image

After:

image

gcanlin (approved):
LGTM. I think the core benefit is from async schedule. Removing boolean-index is releasing the limit for async schedule.

PR #6242, opened 2026-08-16
[Model] Optimize MiniMax Music 3 acoustic inference

Sy0307:
## Dependency and draft status

> [!IMPORTANT]
> This is a stacked **Draft PR** based on #6186 by @linyueqian. It must not be merged before #6186 is fixed and merged.

The branch currently contains the two parent commits from #6186, so the GitHub diff against `main` temporarily includes the full MiniMax Music 3 integration. For this draft, the optimization-only commit is `0a04148b`.

Before marking this PR ready, I will:

1. rebase it onto `main` after #6186 lands, including the correctness fixes requested in its review;
2. confirm that the resulting diff contains only the performance changes and tests;
3. rerun the end-to-end throughput and quality suite in a version-aligned environment.

## Summary

This draft improves MiniMax Music 3 acoustic-stage throughput without reducing the 30-step solver count:

- batches independent acoustic windows across requests, including multi-window spans processed in window rounds while preserving per-request overlap state;
- runs the DiT velocity network under FP16 autocast on Hopper while keeping conditioning, initial noise, and Euler accumulation in FP32;
- forces the Hopper FP16 path through PyTorch SDPA because the available FA3 backend does not support FP16;
- optionally keeps the 148 DiT Linear layers in FP16 while leaving norms and solver state in FP32;
- compiles the shared DiT blocks with a real solver warmup, caches the invariant zero tensor, and reuses the batched CFG solver buffer;
- enables the cuDNN heuristic vocoder path without shape-specific `cudnn.benchmark` startup spikes;
- increases stage admission and Stage-0 CUDA Graph capture sizes for the validated H200 deployment;
- clones explicit solver noise at the ownership boundary so warmup, Euler updates, and prompt pinning cannot mutate caller-owned noise.

## Benchmark

Hardware and workload:

- 1x NVIDIA H200 141 GB, both stages colocated on one GPU;
- 250 AR frames per request, producing 9.996 seconds of 32 kHz stereo WAV;
- identical prompt, seed, and frame budget across runs;
- one warmup request before the measured concurrency sweep.

Software used for the current draft measurements:

- PyTorch `2.11.0+cu130`;
- vLLM `0.27.0`;
- FlashInfer `0.6.16.post3`;
- the checkout reports vLLM-Omni `0.26.0`, so a version-aligned rerun is still required before this leaves draft.

| Concurrency | Original #6186 | This draft | This draft throughput |
|---:|---:|---:|---:|
| 1 | 5.22 s | 3.73 s | 0.268 req/s |
| 8 | 21.79 s | 9.86 s | 0.811 req/s |
| 16 | — | 16.73 s | 0.956 req/s |
| 32 | — | 32.96 s | 0.971 req/s |

At concurrency 8, aggregate throughput improves from `0.367` to `0.811 req/s` (`2.21x`). At concurrency 32, the single H200 produces about `9.70` seconds of audio per wall-clock second.

Component measurements:

- B8, 30-step DiT: `7.816 s -> 3.571 s` with FP16 autocast (`2.19x`);
- FP16-resident Linear weights: an additional `3.1%` reduction in B8 DiT time;
- B8 vocoder: `0.575 s -> 0.272 s` with cuDNN heuristic selection;
- Stage-1 model-loading memory: `9.36 GiB -> 5.44 GiB` with FP16 Linear weights.

## Quality and correctness

This PR does **not** claim byte-identical output to the full-FP32 path.

- Four end-to-end FP32/FP16 seed pairs all produced the same full WAV length.
- The worst measured waveform SNR was `25.61 dB`; this can be audible and must be treated as an explicit quality/performance tradeoff.
- FP16 Linear residency alone measured `54.3 dB` SNR relative to the FP32-weight autocast path.
- The solver now clones explicit noise before in-place Euler and prompt-pinning updates; a regression test verifies that caller-provided noise is unchanged.

The batching correctness issues reported on #6186 must be fixed there before these throughput numbers can be treated as evidence for mixed-seed or staggered-terminal workloads. The current throughput workload uses the same seed and length, which masks that parent-PR bug.

## Tests

```text
69 passed, 14 warnings in 0.27s
```

Covered files:

```bash
python -m pytest -q \
  tests/model_executor/models/test_minimax_music3_prompt.py \
  tests/model_executor/models/test_minimax_music3_frame_state.py \
  tests/model_executor/models/test_minimax_music3_row_binding.py
```

`git diff --check` and Python bytecode compilation also pass.

## Remaining draft work

- [ ] Wait for #6186 to address its request-level correctness blockers and merge.
- [ ] Rebase onto the merged parent and remove the stacked parent commits from this PR.
- [ ] Repeat every concurrency level for multiple measured rounds and report ranges.
- [ ] Rerun performance, peak VRAM, and four-seed audio comparison with aligned vLLM/vLLM-Omni versions.
- [ ] Decide whether `dit_fp16_linear_weights` should remain enabled by default or be an opt-in quality/performance knob.

### DIFF
--- a/vllm_omni/engine/async_omni_engine.py
+++ b/vllm_omni/engine/async_omni_engine.py
@@ -775,6 +775,14 @@ class AsyncOmniEngine:
                 params=companion_params,
                 supported_tasks=self.supported_tasks,
             )
+            # Same restore the parent request gets: the upstream input
+            # processor drops omni-only prompt fields, so without this the
+            # companion reaches the worker with no additional_information at
+            # all. That is where ``global_request_id`` lives, and it is what
+            # was just injected above, so skipping it silently undoes the
+            # injection: the model sees the companion row with no id and
+            # cannot match it to its conditioned partner.
+            request = upgrade_to_omni_request(request, companion_prompt)
             request.external_req_id = cid
             # Companions are stage-0-final for ordinary downstream payloads,
             # but diffusion still needs their CFG KV caches.
--- a/vllm_omni/model_executor/models/minimax_music3/dit.py
+++ b/vllm_omni/model_executor/models/minimax_music3/dit.py
@@ -1,0 +1,484 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+"""Condition encoder and flow-matching transformer for MiniMax Music 3.
+
+Stage 1 turns the AR stage's conditioning frames into a DAC latent in two
+steps. The condition encoder collapses the eight 4096-wide layer slices of a
+frame into one 2048-wide vector and stretches the result from the AR frame
+grid onto the vocoder's latent grid. The transformer is then solved as a
+flow-matching velocity field: a fixed-step Euler integrator walks Gaussian
+noise to the latent the vocoder expects, under classifier-free guidance.
+
+Module names mirror the checkpoint's ``condition_encoder/`` and
+``transformer/`` component folders key for key, with one deliberate
+exception: the per-block ``to_q``/``to_k``/``to_v`` projections are fused into
+a single ``to_qkv`` so each block runs one GEMM instead of three.
+:func:`remap_transformer_state` performs that fusion, and is the only place
+the checkpoint's layout is reinterpreted.
+
+Attention runs through the project's own diffusion attention layer, so this
+model picks up the same backend selection, sequence-parallel plumbing and
+KV-quant policy as every other transformer in the tree. That layer resolves
+its backend from the diffusion config, which an ``LLM_GENERATION`` stage such
+as this one does not set; it tolerates that and falls through to the platform
+default. If it cannot be built at all the block runs plain SDPA instead, a
+choice made once when the blocks are built and never inside ``forward``.
+"""
+
+from __future__ import annotations
+
+import math
+
+import torch
+from torch import Tensor, nn
+from torch.nn import functional as F
+from vllm.logger import init_logger
+
+from .constants import (
+    AR_HIDDEN_SIZE,
+    BACKBONE_HIDDEN_SIZE,
+    CONDITION_INPUT_HOP,
+    CONDITION_INPUT_SAMPLE_RATE,
+    CONDITION_OUTPUT_HOP,
+    CONDITION_OUTPUT_SAMPLE_RATE,
+    DEFAULT_DIT_CFG_SCALE,
+    DEFAULT_DIT_STEPS,
+    DIT_LATENT_CHANNELS,
+    NUM_CODEBOOKS,
+)
+
+logger = init_logger(__name__)
+
+__all__ = [
+    "MiniMaxMusic3ConditionEncoder",
+    "MiniMaxMusic3FlowMatchingDiT",
+    "MiniMaxMusic3Transformer1DModel",
+    "remap_transformer_state",
+]
+
+# Fixed by transformer/config.json. Declared rather than read from the file so
+# the skeleton can be built before any checkpoint is on disk, which is what
+# vLLM's memory profiler needs at startup.
+_DIM = 2_048
+_NUM_LAYERS = 36
+_HEAD_DIM = 64
+_FF_INNER_DIM = 8_192
+_FOURIER_DIM = 256
+_ROTARY_DIM = 32
+_ROTARY_BASE = 10_000.0
+_CONDITION_DIM = 2_048
+# The solver feeds the transformer the current latent, a zero block of the
+# same width, and the condition: 128 + 128 + 2048 = 2304.
+_TRANSFORMER_IN_DIM = DIT_LATENT_CHANNELS * 2 + _CONDITION_DIM
+
+# The prompt interpolation floor from the reference sampler. Sigma never
+# reaches exactly zero, so the noise term never fully vanishes.
+_SIGMA_FLOOR = 1e-6
+
+
+class FourierFeatures(nn.Module):
+    """Random-Fourier timestep features: ``[B, 1] -> [B, out_features]``."""
+
+    def __init__(self, in_features: int, out_features: int) -> None:
+        super().__init__()
+        self.weight = nn.Parameter(torch.empty(out_features // 2, in_features))
+
+    def forward(self, value: Tensor) -> Tensor:
+        f = 2.0 * math.pi * value @ self.weight.T
+        return torch.cat((f.cos(), f.sin()), dim=-1)
+
+
+class TimestepEmbedding(nn.Module):
+    """The checkpoint's ``time_embed``: Linear, SiLU, Linear."""
+
+    def __init__(self, in_dim: int, dim: int) -> None:
+        super().__init__()
+        self.linear_1 = nn.Linear(in_dim, dim)
+        self.act = nn.SiLU()
+        self.linear_2 = nn.Linear(dim, dim)
+
+    def forward(self, x: Tensor) -> Tensor:
+        return self.linear_2(self.act(self.linear_1(x)))
+
+
+def _rotate_half(x: Tensor) -> Tensor:
+    x1, x2 = x.chunk(2, dim=-1)
+    return torch.cat((-x2, x1), dim=-1)
+
+
+def _apply_rope(x: Tensor, rope_cos: Tensor, rope_sin: Tensor) -> Tensor:
+    """Rotate the leading ``rope_cos.shape[-1]`` channels of ``[B, S, H, D]``.
+
+    Only the first 32 of each head's 64 channels are rotated; the rest pass
+    through untouched, which is what this checkpoint was trained with. The
+    tables are cached at the exact sequence length, so there is nothing to
+    slice off the front.
+    """
+    rot_dim = rope_cos.shape[-1]
+    rotated = x[..., :rot_dim]
+    rotated = rotated * rope_cos + _rotate_half(rotated) * rope_sin
+    return torch.cat((rotated, x[..., rot_dim:]), dim=-1)
+
+
+# Set to the failure reason the first time the native attention layer cannot be
+# built, so the 36 blocks agree on one path and the reason is logged once.
+_NATIVE_ATTENTION_UNAVAILABLE: str | None = None
+
+
+def _build_native_attention(*, num_heads: int, head_size: int, softmax_scale: float, prefix: str) -> nn.Module | None:
+    """Build the project's diffusion attention layer, or ``None`` to use SDPA.
+
+    Anything that goes wrong here is a legitimate reason to run plain SDPA,
+    and is reported once rather than once per block.
+    """
+    global _NATIVE_ATTENTION_UNAVAILABLE
+    if _NATIVE_ATTENTION_UNAVAILABLE is not None:
+        return None
+    try:
+        from vllm_omni.diffusion.attention.layer import Attention as DiffusionAttention
+
+        return DiffusionAttention(
+            num_heads=num_heads,
+            head_size=head_size,
+            causal=False,
+            softmax_scale=softmax_scale,
+            prefix=prefix,
+            role="self",
+            role_category="self",
+        )
+    except Exception as exc:  # noqa: BLE001 - any failure means: use SDPA
+        _NATIVE_ATTENTION_UNAVAILABLE = repr(exc)
+        logger.warning(
+            "MiniMax Music 3 DiT could not build the diffusion attention layer "
+            "(%s); running torch SDPA instead. The stage decodes in float32, "
+            "for which the diffusion layer dispatches to SDPA anyway, so this "
+            "changes performance bookkeeping and not the audio.",
+            exc,
+        )
+        return None
+
+
+class Attention(nn.Module):
+    """Bidirectional self-attention with partial rotary position embeddings.
+
+    Q, K and V are kept in ``[B, S, H, D]``, which is the layout the diffusion
+    attention backends consume and return.
+    """
+
+    def __init__(self, dim: int, *, head_dim: int, prefix: str = "") -> None:
+        super().__init__()
+        self.num_heads = dim // head_dim
+        self.head_dim = head_dim
+        self.softmax_scale = float(head_dim**-0.5)
+        # Fused from the checkpoint's separate to_q/to_k/to_v; see
+        # remap_transformer_state for the concatenation order.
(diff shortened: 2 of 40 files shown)


Sy0307:
PTAK @linyueqian after #6186 merged.

vllm_omni/model_executor/models/minimax_music3/talker.py
```diff
@@ -0,0 +1,986 @@
...
+            return OmniOutput(text_hidden_states=model_outputs, multimodal_outputs=None)
+        payload: dict[str, Any] = {"latent": latents}
+        if any(isinstance(entry, torch.Tensor) and entry.numel() for entry in codes):
+            payload["codes"] = {"audio": codes}
+        # The span's start frame and terminal flag ride along with it: without
+        # them the acoustic stage cannot tell which window it is looking at,
+        # and therefore cannot crop it correctly.
+        merged: dict[str, Any] = {}
+        for meta in metas:
+            merged.update(meta)
```
linyueqian, line 939: [Critical] This re-introduces a bug that #6186 already fixed, and the test that guarded it is deleted in this PR.

The previous form was `{key: [meta.get(key) for meta in metas]}`. That was deliberate: `to_payload_element` in `vllm_omni/utils/mm_outputs.py` partitions the batch payload per request by **indexing list values**, and hands **scalars** to every request unchanged. Collapsing the per-row dicts with `merged.update(meta)` means every row now receives the last row's `audio_seed`, `num_processed_tokens` and `last_chunk`.

Concrete failure in a batch of N: seeds cross between songs, acoustic windows are placed at the wrong frame offsets, and one request's terminal flag ends another request's stream early. Only the final request in the batch decodes correctly, and this is the multi-request batching path the PR is optimizing.

This PR also removes the comment that documented the requirement:

> Every field is a list with one entry per batch row, in the same order as `latent`. The payload is partitioned per request by indexing list values, so a scalar here would be broadcast to the whole batch and every request would decode with the last request's seed, window offset and terminal flag.

and deletes `tests/model_executor/models/test_minimax_music3_batch_payload.py`, whose docstring describes this exact failure in the past tense because it was hit once already during #6186 and locked with `test_seeds_are_not_crossed`, `test_window_offsets_stay_with_their_request` and `test_terminal_flag_does_not_leak_across_requests`. `test_minimax_music3_prompt.py` covers prompt assembly, not payload partitioning, so nothing replaces that coverage.

Fix: restore the per-row list construction and keep the regression test.

vllm_omni/engine/async_omni_engine.py
```diff
@@ -775,6 +775,14 @@ def _enqueue_cfg_companions(
...
                 supported_tasks=self.supported_tasks,
             )
+            # Same restore the parent request gets: the upstream input
+            # processor drops omni-only prompt fields, so without this the
+            # companion reaches the worker with no additional_information at
+            # all. That is where ``global_request_id`` lives, and it is what
+            # was just injected above, so skipping it silently undoes the
+            # injection: the model sees the companion row with no id and
+            # cannot match it to its conditioned partner.
+            request = upgrade_to_omni_request(request, companion_prompt)
```
linyueqian, line 785: [Major] Anchoring here because the relevant reorder is not visible in this PR's diff, and that is worth flagging on its own: #6242 restores main's original ordering, so a diff against main shows nothing. The regression only appears against this PR's actual base, #6186.

`add_request` now calls `self.request_queue.sync_q.put(msg)` for the parent **before** `_enqueue_cfg_companions`, and this method wraps `prompt_expand_func` in `try/except Exception` and returns. #6186 deliberately built companions first via `_build_cfg_companions` so the group was all-or-nothing: expansion failure raised, and nothing was admitted.

With the new ordering, when expansion raises the parent is already queued and its companions never arrive. Guidance is mandatory for this checkpoint (`talker.py` `_resolve_pairs`, and `stage_input_processors/minimax_music3.py` `expand_cfg_prompts` both state it), so the pair cannot complete: the row waits out the pairing hold, is released unguided, emits hold tokens until it exhausts its token budget, then returns empty audio. The client sees no error at all, where before the exception reached it immediately and nothing had been admitted.

Nothing reaps it. `cfg_companion_tracker.py` carries an explicit TODO noting there is no liveness owner, and the deferral path in `orchestrator.py` only engages when `has_companions(req_id)` is true, which is false here because the companion was never registered.

The four deleted tests in `tests/engine/test_async_omni_engine_input.py` were the only coverage of this invariant: `test_cfg_companion_build_failure_admits_nothing`, `test_cfg_companion_processing_failure_admits_nothing`, `test_cfg_companion_build_returns_messages_without_enqueueing`, `test_cfg_expansion_returning_nothing_is_not_an_error`. The surviving `test_cfg_companion_suppresses_payload_but_forces_kv_transfer` asserts KV metadata only.

Fix: restore the build-before-enqueue ordering so the group is atomic, or if the reordering is needed for another reason, abort the already-admitted parent when expansion or companion input processing fails.

vllm_omni/model_executor/models/minimax_music3/dit.py
```diff
@@ -0,0 +1,559 @@
...
+        cached = self._rope_cache.get(key)
+        if cached is None:
+            t = torch.arange(seq_len, device=self.inv_freq.device, dtype=torch.float32)
+            freqs = torch.einsum("i,j->ij", t, self.inv_freq)
+            freqs = torch.cat((freqs, freqs), dim=-1).to(dtype=dtype, device=device).unsqueeze(1)
+            cached = (freqs.cos(), freqs.sin())
+            self._rope_cache[key] = cached
+        return cached
+
+    def _latent_zeros(self, x: Tensor) -> Tensor:
```
linyueqian, line 275: [Minor] This cache has no eviction and its key space is not bounded.

`chunk_windows` truncates the final window of a song, so its AR frame count varies per request, and `aligned_mel_length()` maps that to a distinct `T_mel`. The key is `(shape, dtype, device)`, so a long-running server accumulates one permanent zeros tensor per distinct (batch size x final-window length) pair rather than the handful you would get from fixed shapes.

Each entry is only a couple of MB, so this is slow growth rather than a blowup, and there is no aliasing problem — the tensor is only ever read by `torch.cat` and never mutated. But since it is only ever read by `torch.cat`, the allocation it avoids is cheap relative to the DiT forward. Simplest fix is to drop the cache, or keep only the most recent key.

linyueqian (commented):
Reviewed `91df0968`. The DiT batching work here is good, and I checked the batched CFG row layout carefully (`cond_cfg[0::2]`, `cfg_x[0::2]/[1::2]`, `t.repeat_interleave(2)`) — the guidance math is correct and I have no objection to it.

Two blockers though, and both revert fixes that are already on this PR's base (#6186), along with the tests that locked them. Since this is stacked on #6186, the reviewable delta is `+484/-362` over that base rather than the `+6983` the diff shows.

Heads-up on rebasing: I have just pushed #6186 (now `f430bee3a`), which merges current main and fixes a dangling `dots_tts` config import that made `import vllm_omni` fail on the old base. Rebasing onto the new #6186 head will save you from that.

One scope note: the change in `async_omni_engine.py` is engine-level admission semantics with no relation to acoustic optimization. Even once corrected, it would be easier to review as its own PR.

linyueqian:
Correction to my review of `91df0968`. Two of my three findings were wrong and I want to retract them plainly before anything else.

## Retracted: [Critical] on `talker.py:939` and [Major] on `async_omni_engine.py:785`

Neither is a change you made. This branch was cut from `bb62a3faf` (2026-08-13), and the fixes I said you reverted landed in `b0350f2f1` (2026-08-16), three days later. `merged.update(meta)` is my own pre-fix code, and `git diff b0350f2f1^ bb62a3faf` shows `async_omni_engine.py` byte-identical between your base and my pre-fix tree, with `talker.py` differing only by a ruff line-wrap. You never touched either line, and you are the person who reported the metadata bug to me on #6186 in the first place.

My own comment carried the tell and I read it the wrong way round: I wrote that the reorder "is not visible in this PR's diff". It is not visible because it is not there. Sorry for the noise.

The rebase is clean, for what it is worth. `git cherry-pick bb62a3faf..91df09684` onto current main applies with no conflicts (the first ruff commit goes empty). The reviewable delta afterwards is +459/-54 across five files: `acoustic.py`, `dit.py`, both deploy YAMLs, and `test_minimax_music3_prompt.py`. The CONFLICTING flag is branch divergence, not content.

## The batched decode path is correct

I built a differential harness with a row-wise stub velocity network and stub vocoder, so batching cannot change a single float and any difference is a logic bug rather than floating-point noise. Nine scenarios, all bit-identical to the serial path: equal-length co-scheduling, ragged spans, different-length songs, and staggered joins at +1 and +2 engine steps. The per-row seed derivation and the window bookkeeping in `_decode_spans_batched` both hold up. Happy to hand you that harness as a test if it is useful.

## Measurements

On an H200 with real weights, per 200-frame window, batch 1, 30 Euler steps. One methodology note for anyone reproducing this: the fp32 baseline needs `_configure_backends()` called first. Without TF32 the baseline is about 4x too slow and fp16 appears to be an 8.7x win.

| variant | ms | vs main | waveform SNR vs fp32 |
|---|---|---|---|
| DiT fp32 (main) | 1181 | 1.00x | reference |
| DiT `force_torch_sdpa` at fp32 | 1191 | 0.99x | bit-identical |
| DiT fp16 autocast | 716 | 1.65x | 34.6 dB |
| DiT fp16 + fp16 Linear | 596 | 1.98x | 34.8 dB |
| DiT fp16 + fp16 Linear + `compile_dit` | 445 | 2.65x | |
| vocoder cuDNN on | 35 (from 67) | 1.91x | 66.0 dB |

Whole acoustic window: 1248 ms on main to 480 ms here, 2.6x. Your numbers hold up. I measured 34.6 dB against your stated 35.3 dB, and `compile_dit` costs 5.6 s at startup with no recompile stall at batch 1 or 4, full or truncated window, so the eager-free claim is correct.

Worth noting that `force_torch_sdpa()` is a no-op at fp32: the output is bit-identical because vLLM's diffusion attention layer already overrides FLASH_ATTN to SDPA for float32. It only takes effect under the fp16 autocast, which is exactly where you apply it.

## [Major] The batching buys about 10%, for 220 lines that nothing executes

Per-request speedup at stage-1 batch 2/4/8 is 1.08x / 1.11x / 1.11x at fp32, and 1.08x / 1.09x / 1.10x at fp16. The DiT is already compute-saturated at batch 1, so there is little headroom to recover. The flags alone give the 2.6x above in roughly 40 lines.

The coverage gap matters more than the ratio. The weekly e2e sends one request at a time, so `len(batched_entries) > 1` is never true in CI, and the new unit tests cover `_cast_linear_weights` and solver noise ownership rather than the batched path. `_decode_spans_batched` also carries a second implementation of the window-index validation with a different formula from `_decode_span`, so the two can drift apart silently.

Would you consider landing the flags first and taking the batching as its own PR with a test? I verified it works today, but nothing in the repo will notice if that changes.

## [Major] `_latent_zeros`, upgrading my earlier [Minor]

I said "a couple of MB, slow growth". That understated it, so here are numbers.

The key is `(shape, dtype, device)` and the shape now includes the batch dimension, so the reachable key space is (stage-1 batch) x (distinct `T_mel`). `chunk_windows` truncates the final window of a song, so `T_mel` is effectively user-controlled through song length. The ceiling is 8.97 GiB at stage-1 `max_num_seqs: 16` and 34.8 GiB at 32, against a stage budget of `gpu_memory_utilization: 0.2`.

What it buys, measured: 2.7 to 5.0 us per `cat` call, which is 0.45 ms per 10-second song. `torch.zeros_like` comes from the caching allocator and the value is copied by `torch.cat` immediately afterwards. I would delete the cache.

## [Medium] The comment on `compile_dit` in `minimax_music3.yaml` names another serving stack

Convention here is to cite the model or the vendor instead.

## [Medium] The shipped defaults now change every user's audio

`vocoder_cudnn`, `dit_autocast_dtype: float16` and `dit_fp16_linear_weights` are all on in `minimax_music3.yaml`, so the default deployment moves off the validated float32 path at 34.6 dB SNR. 66 dB for the vocoder is benign, but the DiT figure is low enough to want a perceptual check before it becomes the default. Would you consider shipping the flags off and documenting an H200 profile alongside them?

## [Low] Two smaller things

The four new connector flags are documented nowhere outside the two YAMLs. And `minimax_music3_2gpu.yaml` flips stage-0 `enforce_eager: false` and doubles `max_num_seqs` to 64, which is a reasonable change but is not exercised by anything in this PR. Did you run that profile?

Two things I could not check: the full serve e2e, because the GPUs I have are held by other tenants and the shipped YAML wants 0.8 of a card, and anything perceptual. Every quality statement above is a numerical comparison, not a listening test.

vllm_omni/model_executor/models/minimax_music3/talker.py
linyueqian: **Retracted.** This is not your change. `merged.update(meta)` is my own pre-fix code: your base `bb62a3faf` is dated 2026-08-13 and my fix `b0350f2f1` landed 2026-08-16, so you branched three days before the fix and the guarding test existed. You never touched this line, and you are the person who reported the bug to me on #6186. Sorry for the misattribution. Rebasing onto current main resolves it with no conflict. Details in https://github.com/vllm-project/vllm-omni/pull/6242#issuecomment-5335199866

vllm_omni/engine/async_omni_engine.py
linyueqian: **Retracted.** Same cause as the `talker.py` thread: `git diff b0350f2f1^ bb62a3faf` shows this file byte-identical between your base and my pre-fix tree, so the ordering here is mine, not a change you made. I even wrote in this comment that the reorder "is not visible in this PR's diff"; it is not visible because it is not there, and I read that the wrong way round. The scope note about engine-level admission belonging in its own PR no longer applies either, since this PR does not touch the engine.

vllm_omni/model_executor/models/minimax_music3/dit.py
linyueqian: Upgrading this one to **[Major]**, since I measured it and my "couple of MB" estimate was too generous.

The key now includes the batch dimension, so the reachable key space is (stage-1 batch) x (distinct `T_mel`), and `T_mel` is effectively user-controlled through song length. Ceiling is **8.97 GiB at stage-1 `max_num_seqs: 16` and 34.8 GiB at 32**, against a stage budget of `gpu_memory_utilization: 0.2`.

Measured benefit on an H200: 2.7 to 5.0 us per `cat` call, i.e. **0.45 ms per 10-second song**. I would delete the cache.

PR #6255, opened 2026-08-17
[Bugfix] Make the async-output wait bound configurable and default it higher

ivanusto:
## Purpose

`_ASYNC_OUTPUT_TIMEOUT` is hardcoded at 30 s and used at two sites in `DiffusionEngine`. It bounds the wait for one step's background D2H/SHM copy — a copy that finishes in milliseconds, but that is queued behind the GPU work for that step, so the wall-clock wait tracks step time rather than copy time.

On a single GPU, large shapes legitimately run tens of seconds per step, which trips a 30 s bound and aborts the request even though the denoise completed. #5793 reported a 49-step render reaching 49/49 after 31 minutes and being discarded anyway; #5821 reported the same on 44–48 s/step. There is currently no way to raise the bound short of patching the installed package.

A tight bound also buys very little: a genuinely hung engine is surfaced elsewhere — the worker monitor and `check_health()` catch worker death, `shutdown()` settles pending futures, and the pump's dequeue loop breaks on `_is_failed`. What this wait catches, in practice, is a healthy render that is merely slow.

This PR:

- reads `VLLM_OMNI_ASYNC_OUTPUT_TIMEOUT`, defaulting to **600 s** to match `_DLO_DP_WAVE_TIMEOUT_S` in the same subsystem (which uses the same `float(os.environ.get(...))` shape);
- names the environment variable in the timeout log line, so an operator who hits the bound learns that the knob exists without reading the source;
- resolves it per call rather than at import, so the value is never frozen in a module constant, a malformed value cannot break module import, and the knob is testable with `monkeypatch` rather than a module reload — an earlier reload-based draft of this test measurably polluted `tests/diffusion/test_diffusion_engine.py` when the two ran in the same session;
- parses defensively: because the resolution now happens on the request path, a non-numeric or non-positive value is ignored with a `warning_once` and the default applies, rather than raising `ValueError` mid-generation;
- updates `docs/design/feature/async_diffusion_output.md`, which documented the old 30.0s value.

Requested in #5793 (suggested fix 2) and #5821 (suggested fix 3). Note this is only the trigger, not the crash: with #5983 merged the pump now survives the cancellation this timeout causes. #6253 covers the remaining pump-robustness gaps. The three are independent.

## Test Plan

New `tests/diffusion/test_async_output_timeout.py` (CPU-only, no GPU): default value, integer and float overrides, that the value is re-read per call rather than frozen at import, and that malformed values (non-numeric, empty, zero, negative) fall back to the default.

**vLLM Version:** 0.26.1rc1.dev608+g99a10304d

**vLLM-Omni Commit:** baba7d1e2ef9d326486493146b43157f7cd4f8d0

## Test Result

```
tests/diffusion/test_async_output_timeout.py    11 passed
tests/diffusion/test_diffusion_engine.py         \
tests/diffusion/test_result_pump.py               \  133 passed, 1 failed
tests/diffusion/test_diffusion_engine_cleanup.py  /
```

The single failure is `test_move_tensor_tree_moves_nested_cuda_tensors_to_cpu`, which needs a CUDA device (`RuntimeError: Found no NVIDIA driver on your system`) and fails identically on unmodified `main` in the same container.

`ruff check` and `ruff format --check` pass on both changed Python files.


ivanusto:
Self-review.

**What I checked**

- Both call sites of the old constant are converted (`step_streaming` and the sync path in `add_req_and_wait_for_response`); `grep _ASYNC_OUTPUT_TIMEOUT` leaves only the new `_ENV`/`_DEFAULT` names and the helper.
- Confirmed the bound is not load-bearing for failure detection before raising the default, which is the part worth challenging. Worker death is caught by the worker monitor and `check_health()`; `shutdown()` settles pending futures with an exception; the pump's dequeue loop breaks on `_is_failed`. So a longer bound delays no real failure signal — it only stops aborting slow renders.
- Checked the default against the neighbouring knob rather than picking a round number: `_DLO_DP_WAVE_TIMEOUT_S` in `multiproc_executor.py` is also 600 s from an env var, in the same subsystem.
- Grepped for other references to the old 30 s value and found the design doc, which is updated in this PR.
- `ruff check` / `ruff format --check`, plus the neighbouring diffusion suites for regressions.

**Review notes addressed in the latest commit**

- Defensive parsing. The point about `float()` raising on the request path is a good one and it is worse than it looks: resolving per call turned what would have been a startup crash into a mid-generation failure. A non-numeric or non-positive value now falls back to the default with a `warning_once`, covered by parametrized tests.
- Log wording: "raise VLLM_OMNI_ASYNC_OUTPUT_TIMEOUT" did read as raising an exception, and `%.0f` contradicted the float-accepting knob. Now "set ... to a larger value", with `%.1f`.
- The test docstring claiming operators could widen the bound without a restart was simply wrong — editing a running process's environment does not generally work that way. Reworded to what the test actually demonstrates: the value is not frozen in a module constant.

**Where I would welcome a second opinion**

- Whether 600 s is the right default, or whether the default should stay at 30 s with only the knob added. I argued for raising it because the reported cases (#5793, #5821) are ordinary single-GPU workloads rather than exotic ones, so the current default is wrong for them by default — but this is a behaviour change and it is a maintainer's call.
- The shape of the helper. The nearby precedent is a module-level `float(os.environ.get(...))`; I used a function because it keeps the parse off the import path and testable without a module reload. An earlier reload-based version of these tests measurably broke `test_make_engine_runs_startup_warmup` when both files ran in the same session, which is what pushed me here.

**Not covered**

No GPU test on this branch. The behaviour was originally observed on a single GB10 running MiniMax-H3 FL2VA at 44–48 s/step (#5821), where the 30 s bound aborted a completed denoise; the change here is confined to how that bound is resolved.

hsliuustc0106:
@SamitHuang PTAL

hsliuustc0106:
One reference to the old constant survived the update: `recipes/MiniMaxAI/MiniMax-H3-4090.md:114` still explains the dropped four-GPU Ref2VA repeats with "the engine's hardcoded 30 s async output wait (`_ASYNC_OUTPUT_TIMEOUT` in `diffusion_engine.py`)". After this PR that name no longer exists — a one-line note that the wait is now `VLLM_OMNI_ASYNC_OUTPUT_TIMEOUT` (default 600 s) would keep the recipe greppable for anyone hitting the same behavior.

hsliuustc0106:
On the two questions in the self-review:

- Raising the default to 600 s: agree. The bound only aborts renders whose denoise already completed, and the failures it might be assumed to guard are surfaced earlier by the worker monitor (`_start_worker_monitor`, `multiproc_executor.py:368`) and `check_health()` (`multiproc_executor.py:973`), so a longer bound delays no real failure signal. Anchoring the default to `_DLO_DP_WAVE_TIMEOUT_S` (same subsystem, same 600 s) is the right call.
- Helper shape: the per-call function is right here. The nearest precedent resolves `float()` at import, which turns a typo in the env var into a startup crash — acceptable there, but this knob resolves on the request path and needs the per-call read plus the defensive fallback. It's only the second consumer of the pattern in the tree, so I wouldn't generalize it into a shared helper yet.

One question on the test report: the body shows "133 passed, 1 failed" across the four neighboring suites without naming the failure. Which test failed, and does it also fail on the base commit? Given the self-review's note that an earlier draft of this file polluted `test_diffusion_engine.py` in-session, it's worth ruling out residual state leakage from the current version.

hsliuustc0106:
any updates?

ivanusto:
Sorry for the delay, and thanks for the careful review.

Pushed 62fface, which fixes the stale reference you found. `recipes/MiniMaxAI/MiniMax-H3-4090.md` was the only remaining mention of `_ASYNC_OUTPUT_TIMEOUT` in the tree (`grep -rn '_ASYNC_OUTPUT_TIMEOUT' --include='*.md'` now returns nothing but the new env var). I kept the historical fact rather than rewriting it — the wait really was a hardcoded 30 s when those numbers were taken — and named the replacement so the recipe stays greppable:

> That wait was a hardcoded 30 s when these numbers were taken; it is now `VLLM_OMNI_ASYNC_OUTPUT_TIMEOUT` (default 600 s), so a rerun on current main should not lose these repeats to the bound.

On your two points, both of which I agree with and neither of which needs a change now:

- **600 s default.** Your reasoning is the argument I wanted and could not make as precisely: the bound only aborts renders whose denoise already completed, and genuine failures surface earlier through `_start_worker_monitor` and `check_health()`, so a longer bound delays no real signal. Anchoring to `_DLO_DP_WAVE_TIMEOUT_S` keeps the two waits in the same subsystem consistent.
- **Per-call helper.** Agreed, and it is what makes the test possible at all — a module-level constant resolved at import cannot be monkeypatched without a reload.

The two review-bot comments from 08-17 were addressed in the commits before the merge from main, so they may read as outstanding above:

- The timeout log now reads `"Timed out after %.1fs waiting for async output; set %s to a larger value to allow slower steps."` — "set ... to a larger value" rather than "raise", and `%.1f` rather than `%.0f`, so a fractional value prints as given.
- The test docstring no longer claims an operator can widen the bound on a running server. It now states the actual property: the value is resolved per call rather than captured in a module constant, so it never goes stale relative to `os.environ` and no module reload is needed.

CI was green on the previous head; happy to rebase or squash if you would prefer this as a single commit.

ivanusto:
> One question on the test report: the body shows "133 passed, 1 failed" across the four neighboring suites without naming the failure. Which test failed, and does it also fail on the base commit?

Answer, with a fresh run rather than the older report.

**The failure is `tests/diffusion/test_diffusion_engine.py::test_move_tensor_tree_moves_nested_cuda_tensors_to_cpu`**, and it is a missing-GPU failure, not a regression. It is decorated `@hardware_test(res={"cuda": "L4"}, num_cards=1)` and its first statement is `torch.arange(8, dtype=torch.float32, device="cuda")` (`test_diffusion_engine.py:707`), so in a CPU-only container it raises:

```
E   RuntimeError: Found no NVIDIA driver on your system. Please check that you have an
    NVIDIA GPU and installed a driver from http://www.nvidia.com/Download/index.aspx
/usr/local/lib/python3.12/dist-packages/torch/cuda/__init__.py:529: RuntimeError
```

**It fails identically on the base commit.** Same container, same command, `-p no:randomly`, four suites:

| commit | result |
| --- | --- |
| `62fface` (this branch) | `1 failed, 125 passed, 17 warnings in 6.51s` |
| `8dc3d504` (merge-base with `main`) | `1 failed, 125 passed, 17 warnings in 6.57s` |

Same single failure on both sides, so the delta from this PR is zero. (The counts differ from the 133 in the description because that report was taken at `baba7d1e`, before the 08-19 merge from `main`; the tree has moved since. The failing test is the same one.)

**On the state-leakage concern** — that was the right thing to check, and it is clean:

- `test_async_output_timeout.py` and `test_diffusion_engine.py` in one session: `TestRequestBatchCapability::test_make_engine_runs_startup_warmup` **PASSED** (`1 failed, 44 passed` — the failure is again the CUDA test above). That is the exact test the earlier `importlib.reload`-based draft used to break, which is why the current version resolves the value in a function and tests it with `monkeypatch` instead.
- Per-file, in isolation, on this branch: `test_async_output_timeout.py` 11 passed · `test_diffusion_engine.py` 1 failed 33 passed · `test_result_pump.py` 21 passed · `test_diffusion_engine_cleanup.py` 13 passed · `test_multiproc_engine_concurrency.py` 58 passed. The four-suite numbers are exactly the sum of the isolated ones, so collection order changes nothing.

Two small things while you are here:

- **#6253** (the pump-robustness half of this pair) has no `buildkite/vllm-omni` check at all, because it never got the `ready` label that this PR has. DCO, both builds, pre-commit and readthedocs are green on it and `mergeable` is `true`, so it is only waiting on the label plus a reviewer. Could you add `ready`? It is 61 commits behind `main` now — happy to merge `main` into it first if you would rather see a fresh run.
- The offer from my last comment still stands: say the word and I will squash this branch into a single commit, and/or bring it up to date with `main`.

hsliuustc0106 (approved):
lgtm

PR #6261, opened 2026-08-17
[Examples] Harden the shared video-audio runner

hsliuustc0106:
## Summary

- define a canonical image/audio-to-video request envelope in the shared `x_to_video_audio.py` runner
- load complete audio inputs by design and represent them as `(waveform, sample_rate)`
- keep official-fixture prompt cleanup in the example while moving DreamID-specific request adaptation, input sample rate, and sampling extras into `model_extras`
- replace model-name output branches and hard-coded frame layouts/rates with model-provided metadata, declared tensor ranges, and the shared video encoder
- add model-neutral CPU tests and document the request/output contract

No example scripts or model folders are deleted.

## Why

The shared runner encoded DreamID behavior directly: it always selected a fixed 1–3 second conditioning window and selected output tensor conversion, FPS, and sample-rate defaults by model type. That made new video+audio models require another branch and prevented callers from conditioning on the complete audio input.

This change makes the example own the generic request/output contract while keeping model-specific compatibility at the model boundary.

## User impact

Existing CLI flags remain available. Input audio is consumed in full by design; DreamID users should pre-trim long recordings when they need a bounded conditioning window because its conditioning sequence, memory use, and denoising time grow with input duration. Pipelines must report output FPS and audio sample rate, or users can provide explicit `--fps` and `--audio-sample-rate` overrides.

## Validation

- `66 passed`: shared runner, model extras, and video API utility tests
- includes real MP4 encode/decode coverage for a declared `[-1, 1]` channel-first output
- all pre-commit hooks passed
- `git diff --check` passed

Full-model GPU inference was not run.


linyueqian (commented):
Reviewed the refactor plus the DreamID and MagiHuman paths. The direction is right: the runner owning a canonical envelope while `model_extras` owns model translation is a real improvement over two hard-coded branches, and it removes the frame-layout assumptions that would have needed a third branch for the next model. I reproduced your `65 passed` exactly.

I went in expecting to find output corruption and did not. Recording that first since it is the largest part of the change: deleting the explicit `(C, F, H, W)` transpose and the `[-1, 1]` rescale looked unsafe, because `_coerce_video_to_uint8_frames` ends in `np.clip(frame, 0.0, 1.0) * 255`. It is safe. `_normalize_single_video_array` (`video_api_utils.py:376-391`) already transposes channel-first layouts and rescales `[-1, 1]` before frames are split. I ran a random `[-1, 1]` `(C, F, H, W)` array through the old and new paths and the uint8 output is byte-identical. Delegating to the shared encoder is correct here.

Also checked and clean: DreamID reports `fps: 24` and `audio_sample_rate: 16000` (`pipeline_dreamid_omni.py:52-57`), so the new hard failure on missing metadata does not regress the default invocation; dropping the `--model-type` default is safe because `download_dreamid_omni.py:84` writes `_class_name` and `resolve_model_class_name` picks it up; the MagiHuman `--extra-body` path still reaches `pipeline_magi_human.py:2191` through `apply_declared_extra_args`; and unwrapping `(waveform, sample_rate)` back to an array matches what `torch.from_numpy` at `pipeline_dreamid_omni.py:236` expects.

Two things I would want addressed before this leaves draft, both inline. The `--model-class-name` override introduced by this PR cannot take effect, and removing the audio crop is a larger behavioural change than the description suggests: the upstream DreamID-Omni engine crops to a fixed one-second window, so cropping is deliberate in the reference implementation rather than an accident of this runner.

One note on the tests, also inline. They pin the envelope well, but the encode path is mocked out, so none of them would have caught the corruption I went looking for. That is worth closing given full-model inference has not been run.

examples/offline_inference/x_to_video_audio/x_to_video_audio.py
```diff
@@ -220,64 +297,82 @@ def main() -> None:
...
+    if args.model_class_name is not None:
+        omni_kwargs["model_class_name"] = args.model_class_name
     if args.quantization is not None:
         omni_kwargs["quantization"] = args.quantization
     omni = Omni(**omni_kwargs)
-    start = time.perf_counter()
-    outputs = omni.generate(prompt, sampling_params)
-    elapsed = time.perf_counter() - start
+    try:
+        model_class_name = get_model_class_name(omni) or args.model_class_name
```
linyueqian: This makes `--model-class-name` inert, so the override added by this PR never reaches adapter selection.

`get_model_class_name(omni)` reads `engine.od_config` first, but the client-side engine has no such attribute, so it always falls through to `get_diffusion_od_config()`. That resolves the class from the model files only:

```python
model_class_name = resolve_model_class_name(self.model)   # async_omni_engine.py:282
```

`resolve_model_class_name` returns `_class_name` from `model_index.json` (`data.py:559-561`), and `download_dreamid_omni.py:84` writes that file, so resolution succeeds for any normal checkpoint and the `or` never fires. Meanwhile line 311 still forwards the override into `omni_kwargs`, so the worker can load the user's class while this runner selects adapters for the file-derived one.

That divergence is the only case where the flag is meaningful: passing a value equal to the resolved one is a no-op, and passing a different one silently splits the two. Concretely, overriding a checkpoint to `DreamIDOmniPipeline` skips `build_dreamid_x_to_video_audio_prompt`, so audio stays as `(waveform, sample_rate)` tuples and `torch.from_numpy` raises at `pipeline_dreamid_omni.py:236`.

```suggestion
        model_class_name = args.model_class_name or get_model_class_name(omni)
```

The `resolved or local` idiom is copied from `text_to_video.py:517`, but this is the first place it is paired with a user-facing override flag, which is what inverts the intended precedence.

examples/offline_inference/x_to_video_audio/x_to_video_audio.py
```diff
@@ -111,20 +160,85 @@ def parse_args() -> argparse.Namespace:
...
+            images.append(img.convert("RGB"))
 
-    for path in audio_paths:
-        audio_array, sr = load_audio(path, sr=16000)
-        audio_array = audio_array[int(sr * 1) : int(sr * 3)]
-        audio.append(audio_array)
-    return image, audio
+    for path in audio_paths or []:
+        waveform, sample_rate = load_audio(path, sr=audio_sample_rate)
+        audios.append((waveform, sample_rate))
```
linyueqian, line 202: Loading the full file is a bigger change than "stop silently cropping", and I do not think the description's framing holds.

The upstream DreamID-Omni engine crops before encoding:

```python
audio_array, sr = librosa.load(path, sr=16000)
audio_array = audio_array[int(sr * 1): int(sr * 2)]
```

That is a fixed one-second conditioning window. This runner used two seconds, so it already diverged from upstream, but in both cases the bound is deliberate rather than an accident.

Nothing crashes without it: `ref_audio_lengths` is derived per input and `audio_noise_len = self.audio_latent_length + ref_audio_length` (`pipeline_dreamid_omni.py:551`) scales dynamically, with full attention on the audio branch. The cost is that the denoising sequence now grows without bound with input duration. A 30-second file is roughly 15x the previous conditioning prefix, which changes peak memory and step time, and conditions generation on different content than either this runner or upstream produced before.

Could this keep a bounded default and expose the window explicitly, say `--audio-window-seconds` defaulting to the current 1-3 s, so full-file conditioning is opt-in? That keeps the "do not silently modify user input" goal without making unbounded attention the default. If full-file really is intended as the default, it would be worth saying so in the docs along with the memory implication, and dropping the "silently crops" framing from the PR description.

tests/examples/offline_inference/test_x_to_video_audio.py
```diff
@@ -0,0 +1,133 @@
...
+    video = object()
+    audio = object()
+    output = x_to_video_audio.XToVideoAudioOutput(video, audio, 23.976, 48000)
+    received: list[tuple[Any, ...]] = []
+
+    def fake_encode(*args: Any) -> bytes:
+        received.append(args)
+        return b"encoded"
+
+    monkeypatch.setattr(x_to_video_audio, "encode_video_bytes", fake_encode)
```
linyueqian: Mocking `encode_video_bytes` means no test executes the path this PR is actually relying on.

The runner now delegates all layout and range handling to the shared encoder, which is the load-bearing assumption of the refactor. This test asserts only that the four fields arrive in the right argument positions, and `test_output_contract_keeps_model_video_layout_opaque` passes the payload through without encoding. Its `np.zeros((3, 2, 4, 4))` fixture is also entirely within `[0, 1]`, so it would not exercise the `[-1, 1]` rescale even if it did reach the encoder.

So if the transpose or the rescale were wrong, this suite would still be green. With full-model inference not run, one test that puts a real `(C, F, H, W)` float array in `[-1, 1]` through the real `encode_video_bytes` and asserts the decoded frames are the expected shape and mid-grey where the input was `0.0` would cover the assumption cheaply. That is the check I ended up writing by hand to convince myself this was safe.

examples/offline_inference/x_to_video_audio/x_to_video_audio.py
```diff
@@ -220,64 +297,82 @@ def main() -> None:
...
+                extra_body.setdefault(key, value)
+        if declared_extra_body_params:
+            apply_declared_extra_args(sampling_params, declared_extra_body_params, extra_body)
+        elif extra_body:
+            sampling_params.extra_args.update({key: value for key, value in extra_body.items() if value is not None})
+
+        start = time.perf_counter()
+        outputs = omni.generate(prompt, sampling_params)
+        elapsed = time.perf_counter() - start
+        output = extract_x_to_video_audio_output(
```
linyueqian, line 405: Minor, non-blocking, and reviewers disagreed on whether it is worth doing.

The `[-1, 1]` rescale the encoder applies keys off the data rather than the contract:

```python
if video_array.min() < 0.0 or video_array.max() > 1.0:   # video_api_utils.py:390
```

A `[-1, 1]` clip whose samples all happen to be non-negative skips the rescale and renders about half as bright. I measured a max error of 127/255 on a synthetic case. Real generated video essentially always contains a sample below mid-grey, so the practical risk is low, which is why I am not flagging it as a defect.

The reason to mention it is that the repo already has the explicit form: `get_output_tensor_range()` returns `negative_one_to_one` for `DreamIDOmniPipeline`, and `text_to_video.py:519` uses it. Since this PR is about replacing model-name branches with model-provided metadata, consuming that declaration here would fit the goal and remove the reliance on a value heuristic.

vllm_omni/model_extras/dreamid_omni.py
```diff
@@ -0,0 +1,64 @@
...
+        r"\[AUDIO_DESCRIPTION_START].*?\[AUDIO_DESCRIPTION_END]",
+        "",
+        prompt,
+        flags=re.DOTALL,
+    ).strip()
+    prompt = re.sub(r"\[[A-Z_]+\]", "", prompt)
+    return re.sub(r"\n\s*\n", "\n", prompt).strip()
+
+
+def build_x_to_video_audio_prompt(
```
hsliuustc0106, line 13: this does not belong to this PR only, we should move it under examples and I figure out there already exists inn x_to_video_audio.py

PR #6046, opened 2026-08-11
[CI/Build] Add examples policy to PR skills

hsliuustc0106:
## Summary

- define a canonical, diff-scoped Python examples policy in the `precheck-pr` skill
- make the `review-pr` skill reuse that policy through a direct relative reference
- block newly added, copied, or renamed model-specific Python examples while grandfathering existing example debt
- route reusable model support and workflows to the appropriate non-example locations

## Impact

This is advisory skill guidance only. It does not delete existing examples, change runtime behavior, or add a hard CI gate.

## Validation

- `quick_validate.py .claude/skills/precheck-pr`
- `quick_validate.py .claude/skills/review-pr`
- verified the cross-skill policy link resolves
- `git diff --check`

Issue #4327, opened 2026-06-10
[RFC]: Unify per-model TTS dispatch in serving_speech.py via a mandatory adapter base class


linyueqian:
## Motivation.

`vllm_omni/entrypoints/openai/serving_speech.py` (2,592 lines) dispatches per-model TTS behaviour through ad-hoc branching scattered across ~10 sites, with **three** boolean flags (`_is_tts`, `_is_fish_speech`, `_is_cosyvoice3`) coexisting with a `_tts_model_type` string discriminator. Adding a model is shotgun surgery on the single file that every model PR touches — the in-repo TTS integration skill calls it *"the most common source of rebase conflicts."*

The dispatch is also split across **two serving backends**: the AR `engine_client` path (`_prepare_speech_generation`) and the pure-diffusion `_diffusion_engine` path (`for_diffusion` / `_create_diffusion_speech`, selected at deploy time in `api_server.py`), with OmniVoice's prompt-building duplicated across both.

This RFC proposes a mandatory `TTSModelAdapter` ABC plus a stage-keyed registry so each model declares its full behaviour (request normalization, validation, prompt-building, generation-param mapping, sampling overrides, output policy, and serving backend) in **one file**. `serving_speech.py` collapses to a thin generic orchestrator, and adding a model becomes *"write one adapter file and import it"* — zero edits to the dispatch core, removing the rebase-conflict hot spot.

It is the speech-side companion to #3572 (diffusion pipelines). It does **not** compete with it: for the one pure-diffusion TTS model served here (OmniVoice), the adapter *consumes* #3572's `EXTRA_BODY_PARAMS` contract rather than reinventing parameter handling. It is also complementary to — and independent of — the partial normalization in #2757.

## Proposed Change.

The full design document is inlined below. It covers the scattered-site inventory, the two-backend reality, the `TTSModelAdapter` contract (with `ARTTSAdapter` / `DiffusionTTSAdapter` specializations), the registry, the `SpeechServingContext`, a `[N/...]` strangler migration plan, and the testing/equivalence strategy. It has been cross-checked against the source at `HEAD` and adversarially reviewed.

---

# RFC: Mandatory Adapter Base Class for TTS Serving Models with Unified Per-Model Dispatch

[...]

PR #6262, opened 2026-08-17
ci: add examples/ Python path ratchet to enforce examples policy

### LINKED ISSUE #6260: [Help Wanted][Examples] Ratchet Python examples toward shared task runners
## Motivation

[#6046](https://github.com/vllm-project/vllm-omni/pull/6046) added a diff-scoped policy that blocks new model-specific Python examples in the precheck/review skills. Its own impact statement correctly notes that the policy is advisory and does not add a hard CI gate.

Since that policy merged, six net-new model-specific Python paths and one policy-relevant rename have landed under `examples/`. This is a process gap, not a criticism of those contributions: a rule that exists only in review guidance is easy to miss.

[#6076](https://github.com/vllm-project/vllm-omni/pull/6076) demonstrated the intended migration pattern for LingBot-Video: preserve model defaults and behavior behind the shared `text_to_video.py` runner, use the canonical request envelope and `model_extras` contracts, add model-neutral tests and hardware parity evidence, repoint documentation, then retire the dedicated script.

This issue turns that pattern into an enforceable, incremental cleanup plan and asks the community to help with independently claimable PRs. It complements [#6116](https://github.com/vllm-project/vllm-omni/issues/6116), which governs examples documentation and navigation. #6116 may catalog grandfathered examples, but it does not exempt new model-specific Python files.

## Target state

- One maintained Python entrypoint per user task or protocol.
- Model-specific defaults, prompt/request adaptation, and output contracts live in production model modules or `vllm_omni/model_extras/`.
- Model commands, hardware notes, and validation evidence live in task documentation or recipes.
- Existing example debt is grandfathered and migrated deliberately; there is no bulk deletion.

## Work plan

Please follow this dependency order. Comment before starting and claim one unchecked item. If a migration exposes a missing generic capability, add that capability in a small prerequisite PR instead of adding model-name branches to the shared runner.

### 1. Make the examples policy enforceable

- [ ] Add a small CI ratchet for Python paths introduced under `examples/`.

Acceptance criteria:

- Compare the PR head with its merge base and inspect added, copied, and renamed destination paths (`ACR`) under `examples/**/*.py`.
- Grandfather paths that exist when the ratchet lands. Modifications and deletions pass; the grandfathered baseline may only shrink.
- Reject every newly introduced Python path by default, including copies and renames.
- Permit a genuinely model-neutral, user-facing task/protocol runner only through an exact-path exception with a written justification and explicit maintainer review. Do not allow directory globs or name-based loopholes.
- Test additions, copies, renames, modifications, deletions, non-Python files, and approved exceptions.
- Ensure CI has enough base history and fails clearly rather than silently skipping the check.
- Point failures toward shared runners, production adapters/`model_extras`, recipes, tests, tools, apps, or benchmarks as appropriate.
- Do not migrate or delete existing examples in the ratchet PR.

### 2. Harden the existing X-to-video/audio runner

- [ ] Harden [`x_to_video_audio.py`](https://github.com/vllm-project/vllm-omni/blob/main/examples/offline_inference/x_to_video_audio/x_to_video_audio.py) without deleting any existing model example.

The current runner silently resamples audio to 16 kHz and keeps only seconds 1–3, branches request construction on DreamID versus Magi, and assumes model-specific frame layouts, ranges, FPS, and audio sample rates during export. No CPU test directly exercises the shared runner.

Acceptance criteria:

- Define an importable canonical request builder: `prompt`, optional `negative_prompt`, `multi_modal_data` containing only supplied `image`/`audio`, and optional `additional_information`. Model knobs use explicit config/extra channels, not model-name conditionals.
- Remove implicit audio cropping and resampling. Any trim or resample must be explicit, documented, and tested.
- Resolve one canonical media representation. Preferred: decoded RGB images and `(waveform, sample_rate)` audio with cardinality preserved. An acceptable alternative is unmodified paths with all decoding owned by production model processors. Do not silently mix representations by model.
- Normalize `OmniRequestOutput` by documented structure and metadata rather than model name. Cover PIL frame sequences, THWC `uint8`, THWC float `[0, 1]`, and CTHW float `[-1, 1]`; reject ambiguous shapes/ranges.
- Read generated audio, FPS, and sample rate from `multimodal_output`, with explicit CLI fallbacks. Muxing conditioning audio must be opt-in, not inferred from a model name.
- Extract pure request, sampling, frame-normalization, and mux-metadata helpers; keep orchestration thin and close `Omni` in `finally`.
- Add model-neutral CPU tests for text/image/audio combinations, repeated media, untrimmed audio, output variants, optional audio, metadata precedence, invalid output, and one fake-`Omni` CLI path.
- Keep the documented DreamID and Magi commands working except for the intentional removal of hidden audio cropping. Retain existing model E2E coverage.
- Do not migrate Wan or LongCat in this PR.

### 3. Migrate one model per PR

- [ ] Migrate Wan from [`speech_to_video.py`](https://github.com/vllm-project/vllm-omni/blob/main/examples/offline_inference/speech_to_video/speech_to_video.py) to the hardened shared runner.
- [ ] Migrate LongCat-Video-Avatar from [`longcat_video/end2end.py`](https://github.com/vllm-project/vllm-omni/blob/main/examples/offline_inference/longcat_video/end2end.py) in a separate PR.

For each model migration:

- Preserve supported modes, defaults, and user-visible behavior through presets, production adapters, and explicit extras. Do not grow a model-dispatch ladder in the shared runner.
- Add CPU request/output contract tests and retain the relevant model E2E tests.
- Record a reproducible hardware smoke/parity command with model, commit, seed, sampling settings, media metadata, and result.
- Repoint README, recipe, generated-doc, benchmark, and test references before removing the dedicated entrypoint.
- Keep the PR limited to one model and include a DCO `Signed-off-by` trailer.

Wan should go first because its request is already close to the canonical `prompt` / `negative_prompt` / `multi_modal_data` envelope. LongCat is a later, maintainer-assisted migration: its multi-speaker JSON cases, conditioning-audio merge, and mux behavior must become production-owned or genuinely reusable capabilities, not copied into the generic runner.

## TTS is a separate follow-up project

The offline `text_to_speech` hub currently has 17 model-specific folders. Do not bundle those migrations into this tracker. Open a separate design issue for a shared `text_to_speech.py` contract first, land the model-neutral runner and CPU contract tests without deleting model scripts, then migrate exactly one model per PR.

## Coordination

- Comment with the checklist item you want to claim and wait for confirmation so work is not duplicated.
- Start a migration only after its shared-runner prerequisite has merged.
- Link this issue from each PR and report any generic-contract gap before implementing a model-specific workaround.
- Small reviewable PRs are preferred; unrelated cleanup belongs elsewhere.

MSaiRam10:
Closes #6260 (item 1)

## Summary

Adds `tools/pre_commit/check_examples_policy.py` - a CI ratchet that blocks new model-specific Python files added under `examples/` per the governance policy introduced in #6046.

## What it does

- Diffs PR head vs merge base using `--diff-filter=ACR` to catch added, copied, and renamed destination paths under `examples/**/*.py`
- Grandfathers all existing paths in `tools/pre_commit/examples_policy_baseline.txt`
- Approved model-neutral task/protocol runners listed in `APPROVED_EXCEPTIONS` with written justifications
- Baseline may only shrink as models are migrated to shared runners
- Points failures toward shared runners, production adapters, model_extras, recipes, and docs

## Precedent

Follows the budget-file pattern from `tools/pre_commit/check_tts_adapter.py` (RFC #4327/#4855), extended with diff-vs-merge-base semantics for ACR paths as noted by the maintainer.

## Tests

14 tests covering: happy path (no new paths), grandfathered paths, approved exceptions, new model-specific paths blocked, modifications pass, deletions pass, non-Python files ignored, multiple blocked paths reported together.

## What this does NOT do

Does not migrate or delete any existing examples. That is items 2-4 in #6260.


tools/pre_commit/check_examples_policy.py
```diff
@@ -0,0 +1,139 @@
...
+        print("check_examples_policy: git diff failed", file=sys.stderr)
+        sys.exit(1)
+    return [
+        line.strip()
+        for line in result.stdout.splitlines()
+        if line.strip().startswith("examples/") and line.strip().endswith(".py")
+    ]
+
+
+def main(argv: list[str] | None = None) -> int:
```
hsliuustc0106, line 103: **[P1]** As delivered this ratchet never runs: there is no hook entry for it in `.pre-commit-config.yaml` (the precedent `check_tts_adapter.py` is wired there at line 66), and no workflow invokes it — at the merge base the only workflows are `build_wheel.yml` and `pre-commit.yml`, neither of which calls into `tools/pre_commit/`. Merging as-is enforces nothing, which defeats the PR's stated purpose (#6260 item 1).

Smallest fix: add a local hook entry (`pass_filenames: false`, `files: ^(examples/.*\.py|tools/pre_commit/check_examples_policy\.py|tools/pre_commit/examples_policy_baseline\.txt)$`), plus the checkout fix noted on `_get_merge_base` so the wired hook can actually pass in CI.

tools/pre_commit/check_examples_policy.py
```diff
@@ -0,0 +1,139 @@
...
+def _load_baseline() -> set[str]:
+    if not BASELINE_FILE.is_file():
+        return set()
+    lines = BASELINE_FILE.read_text().splitlines()
+    return {line.strip() for line in lines if line.strip() and not line.startswith("#")}
+
+
+def _get_merge_base(base_ref: str) -> str:
+    result = subprocess.run(
+        ["git", "merge-base", "HEAD", base_ref],
```
hsliuustc0106, line 74: **[P1]** `.github/workflows/pre-commit.yml` checks out with `actions/checkout@v6` at the default `fetch-depth: 1`, so no `origin/main` ref exists in that checkout — `git merge-base HEAD origin/main` exits nonzero and the script exits 1. Once this hook is wired into `.pre-commit-config.yaml`, every PR will fail the pre-commit lane, not just ones violating the policy.

The cited precedent avoids this by design: `check_tts_adapter.py` is AST-only over the working tree and needs no git state. Smallest fix: `fetch-depth: 0` on that checkout step (or a `git fetch origin main` step before pre-commit), plus one CI run of this PR to prove the wired hook passes.

tools/pre_commit/examples_policy_baseline.txt
```diff
@@ -0,0 +1,134 @@
+# Grandfathered examples/ Python paths as of the ratchet introduction.
+# These paths existed before the policy was enforced.
+# This list may only SHRINK as paths are migrated or removed.
+# Do not add new paths here - use APPROVED_EXCEPTIONS in check_examples_policy.py instead.
+
+examples / offline_inference / audex / audio_qa.py
```
hsliuustc0106, line 6: **[P1]** Every entry is written with spaces around the slashes (`examples / offline_inference / ...`), so `_load_baseline()` returns strings that `git diff --name-only` output can never match: `path in baseline` is always False and grandfathering is entirely dead — every A/C/R path under `examples/` gets blocked unless it happens to be in `APPROVED_EXCEPTIONS` (which is written with plain slashes, unlike this file).

This also means `test_baseline_only_python_paths` (asserts `path.startswith("examples/")`) fails against the shipped file, so the 14 tests in the PR description cannot all have passed against this state.

Fix: regenerate as `git ls-tree -r HEAD --name-only examples/ | grep '\.py$'` — I diffed that against this file ignoring the spaces, and the path set is correct except line 22 (`image_to_image/image_to_image.py`), which does not exist in the tree and would silently grandfather a future file at that path; drop it while regenerating.

tools/pre_commit/tests/test_check_examples_policy.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def test_no_new_paths_passes():
+    # No new paths added - should pass
+    with patch("check_examples_policy._get_added_paths", return_value=[]):
+        result = main(["--base-ref", "origin/main"])
+    assert result == 0
+
+
+def test_grandfathered_path_passes():
```
hsliuustc0106, line 51: **[P2]** Two gaps. First, nothing runs these tests: no workflow invokes `pytest tools/pre_commit/tests/`, so they are local-only — which is exactly how the malformed baseline slipped through. Wire them into an existing lane.

Second, coverage: this test takes its fixture from the same baseline it is verifying, so it is self-consistent with the spacing bug — derive fixtures from `git ls-tree` instead, and add a "baseline ⊇ tree" assertion so formatting regressions and coverage gaps fail loudly. Also uncovered: `_get_added_paths` itself (note `--diff-filter=C` never emits copies without `--find-copies`, so the `C` is currently a no-op) and the merge-base failure path — one tmp-repo test driving real `git` would cover both.

PR #6263, opened 2026-08-17
[Draft][HunyuanImage3] Fuse GroupNorm+SiLU and adaptive GroupNorm,

Bounty-hunter:

## Test Plan

**vLLM Version:**

**vLLM-Omni Commit:**

## Test Result

### DIFF
--- a/tests/model_executor/models/common/test_fused_adaptive_group_norm.py
+++ b/tests/model_executor/models/common/test_fused_adaptive_group_norm.py
@@ -1,0 +1,257 @@
+"""Unit tests for fused_adaptive_group_norm operator.
+
+Tests numeric correctness against PyTorch native implementation:
+    F.group_norm(x, num_groups, weight, bias, eps) * (1 + scale) + shift
+"""
+
+import pytest
+import torch
+import torch.nn.functional as F
+
+from vllm_omni.model_executor.models.common.ops import fused_adaptive_group_norm
+
+# Skip tests if CUDA not available
+pytestmark = pytest.mark.skipif(
+    not torch.cuda.is_available(),
+    reason="CUDA required for Triton kernels"
+)
+
+
+def _reference(x, weight, bias, scale, shift, num_groups, eps):
+    """AdaGN reference, computed in fp32.
+
+    As with fused_group_norm_silu, a same-dtype reference is unusable: the
+    GroupNorm affine can cancel to near zero, and rounding its O(1) terms to
+    bf16 wipes out the remaining significant digits.
+    """
+    B, C = x.shape[:2]
+    normed = F.group_norm(
+        x.float(), num_groups, weight.float(), bias.float(), eps
+    )
+    scale_b = scale.float().view(B, C, *([1] * (x.ndim - 2)))
+    shift_b = shift.float().view(B, C, *([1] * (x.ndim - 2)))
+    return normed * (1 + scale_b) + shift_b
+
+
+def _tolerance(dtype):
+    if dtype == torch.float32:
+        return 1e-5, 1e-5
+    if dtype == torch.float16:
+        return 2e-3, 2e-3
+    return 1e-2, 1e-2  # bfloat16
+
+
+@pytest.mark.parametrize("batch_size", [1, 2])
+@pytest.mark.parametrize("channels", [32, 64, 128])
+@pytest.mark.parametrize("spatial_size", [(16, 16), (32, 32)])
+@pytest.mark.parametrize("num_groups", [8, 32])
+@pytest.mark.parametrize("dtype", [torch.float32, torch.float16, torch.bfloat16])
+def test_fused_adaptive_group_norm_correctness(
+    batch_size, channels, spatial_size, num_groups, dtype
+):
+    """Test numeric correctness against PyTorch native ops."""
+    if channels % num_groups != 0:
+        pytest.skip(f"channels {channels} not divisible by num_groups {num_groups}")
+
+    H, W = spatial_size
+    eps = 1e-6
+    torch.manual_seed(42)
+    kw = dict(device="cuda", dtype=dtype)
+
+    x = torch.randn(batch_size, channels, H, W, **kw)
+    weight = torch.randn(channels, **kw)
+    bias = torch.randn(channels, **kw)
+    scale = torch.randn(batch_size, channels, **kw)
+    shift = torch.randn(batch_size, channels, **kw)
+
+    ref_out = _reference(x, weight, bias, scale, shift, num_groups, eps).to(dtype)
+    fused_out = fused_adaptive_group_norm(
+        x, weight, bias, scale, shift, num_groups, eps
+    )
+
+    rtol, atol = _tolerance(dtype)
+    torch.testing.assert_close(
+        fused_out, ref_out,
+        rtol=rtol, atol=atol,
+        msg=f"Mismatch for batch={batch_size}, C={channels}, "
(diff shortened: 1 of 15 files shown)


tests/model_executor/models/common/test_fused_adaptive_group_norm.py
```diff
@@ -0,0 +1,257 @@
+"""Unit tests for fused_adaptive_group_norm operator.
```
hsliuustc0106, line 1: open the design issue first

vllm_omni/diffusion/models/hunyuan_image3/autoencoder_blocks.py
```diff
@@ -0,0 +1,43 @@
+# SPDX-License-Identifier: Apache-2.0
```
hsliuustc0106, line 1: why not directly import from nvidia?

vllm_omni/diffusion/models/hunyuan_image3/transformer_blocks.py
```diff
@@ -0,0 +1,149 @@
+# SPDX-License-Identifier: Apache-2.0
```
hsliuustc0106, line 1: why not direct import from nvidia?

vllm_omni/diffusion/models/hunyuan_image3/autoencoder_blocks.py
```diff
@@ -0,0 +1,43 @@
...
+
+class ResnetBlock(nn.Module):
+    def __init__(self, in_channels: int, out_channels: int):
+        super().__init__()
+        self.in_channels = in_channels
+        out_channels = in_channels if out_channels is None else out_channels
+        self.out_channels = out_channels
+
+        self.norm1 = nn.GroupNorm(num_groups=32, num_channels=in_channels, eps=1e-6, affine=True)
+        self.conv1 = nn.Conv3d(in_channels, out_channels, kernel_size=3, stride=1, padding=1)
```
hsliuustc0106, line 22: On main, `ResnetBlock` in `autoencoder.py` builds its convs from the chunked `Conv3d` (the 2 GiB temporal-chunking subclass) on every platform. This default block uses plain `nn.Conv3d`, so NPU/XPU/CPU lose the memory bounding that the CUDA variant keeps (`nvidia/autoencoder_blocks.py:26`). That is a silent capacity/numerics change for non-CUDA VAE decode at large resolutions. If dropping the chunking off-CUDA is intentional, worth stating why with NPU memory numbers; otherwise the default block should use the same chunked `Conv3d`.

tests/model_executor/models/common/test_fused_group_norm_silu.py
```diff
@@ -0,0 +1,295 @@
...
+"""
+
+import pytest
+import torch
+import torch.nn.functional as F
+
+from vllm_omni.model_executor.models.common.ops import fused_group_norm_silu
+
+# Skip tests if CUDA not available
+pytestmark = pytest.mark.skipif(
```
hsliuustc0106, line 14: Both new test files are currently deselected in every CI lane: `tests/model_executor` runs under `-m 'core_model and cpu'` (`.buildkite/cuda/test-ready.yml:13`) and `-m 'core_model and cuda'` (`:87`), and neither file carries the `core_model` marker that every sibling under `tests/model_executor/models/` uses. Adding `pytestmark = [pytest.mark.core_model, pytest.mark.cuda]` (keeping the skipif) gets them picked up by the cuda lane.

vllm_omni/diffusion/models/hunyuan_image3/nvidia/transformer_blocks.py
```diff
@@ -0,0 +1,178 @@
...
+
+        if self.out_channels == self.in_channels:
+            self.skip_connection = nn.Identity()
+        elif use_conv:
+            self.skip_connection = conv_nd(dims, self.in_channels, self.out_channels, 3, padding=1, **factory_kwargs)  # noqa: N802
+        else:
+            self.skip_connection = conv_nd(dims, self.in_channels, self.out_channels, 1, **factory_kwargs)  # noqa: N802
+
+    def forward(self, x, emb) -> torch.Tensor:
+        with cudnn_settings(benchmark = True, deterministic = True):
```
hsliuustc0106, line 142: Is this per-forward `cudnn_settings(benchmark=True, deterministic=True)` wrapper load-bearing? The Triton kernels don't consult cuDNN flags, so it only affects the convs — and it overrides whatever the engine/user set globally, forces `deterministic=True` (a conv-perf cost), and `benchmark=True` adds per-shape autotune warmup with varying video shapes. If it matters for the measured speedups, setting the flags once at init and saying why would be cleaner; otherwise I'd drop it. Same wrapper in `nvidia/autoencoder_blocks.py:81`.

vllm_omni/diffusion/models/hunyuan_image3/nvidia/autoencoder_blocks.py
```diff
@@ -0,0 +1,98 @@
...
+                        value=0,
+                    )
+                    if i > 0:
+                        padded_chunk[:, :, : self.padding[0]] = chunks[i - 1][:, :, -self.padding[0] :]
+                    if i < len(chunks) - 1:
+                        padded_chunk[:, :, -self.padding[0] :] = chunks[i + 1][:, :, : self.padding[0]]
+                else:
+                    padded_chunk = chunks[i]
+                padded_chunks.append(padded_chunk)
+            padding_bak = self.padding
```
hsliuustc0106, line 55: `self.padding` stays zeroed if any chunk's conv raises (e.g., a retryable OOM mid-decode) — the restore on line 60 never runs, and every later forward of this loaded block silently convolves without temporal padding. A try/finally around the loop (or `F.conv3d` with explicit padding instead of mutating the module) closes it.

vllm_omni/diffusion/models/hunyuan_image3/nvidia/transformer_blocks.py
```diff
@@ -0,0 +1,178 @@
...
+def zero_module(module):
+    """
+    Zero out the parameters of a module and return it.
+    """
+    for p in module.parameters():
+        p.detach().zero_()
+    return module
+
+
+class ResBlock(nn.Module):
```
hsliuustc0106, line 73: `_DefaultResBlock` is imported on line 25 but never used — the whole `__init__` is duplicated below, while the module docstring says "Only forward is overridden -- __init__ is inherited". `class ResBlock(_DefaultResBlock)` overriding only `forward` would delete ~70 lines plus the duplicated `conv_nd`/`normalization`/`linear`/`zero_module` helpers, and remove the risk that a future edit to the default `__init__` desyncs CUDA checkpoint keys from the default block's.

vllm_omni/model_executor/models/common/ops/fused_group_norm_silu.py
```diff
@@ -0,0 +1,207 @@
+"""Fused GroupNorm + SiLU operator.
+
+This operator fuses GroupNorm followed by SiLU activation into a single kernel,
+reducing memory traffic and kernel launch overhead. The implementation uses
+Triton for CUDA/ROCm compatibility, and falls back to native PyTorch ops when
```
hsliuustc0106, line 5: Minor: the docstring (here and in `fused_adaptive_group_norm.py`) promises CUDA/ROCm coverage, but the dispatch in `hunyuan_image3/__init__.py` selects these only when `is_cuda()` — false for the ROCm enum — so ROCm always gets the default blocks. Either add `is_rocm()` to the dispatch or reword the docstring.

PR #6272, opened 2026-08-17
[Bugfix][MiniCPM-o] Stop Talker mid-utterance truncation and load cod…

amy-why-3459:
Summary

1. MiniCPM-o 4.5 non-duplex Talker used text_len * 10 as a hard codec budget. A 15-token sentence was capped at 150 codes (~6s) and cut off mid-speech before EOS.
2. Stage-1 max_tokens is raised from 2048 to 4096 to match Talker max_position_embeddings.
3. Codec sampling (temperature, top_k, top_p, repetition_penalty, seed, min_tokens, max_tokens) now comes from deploy YAML codec_sampling_params, not model hardcoded fallbacks. Missing YAML keys fail fast.
4. This is separate from stage-1 default_sampling_params, which still only controls vLLM’s binary continue/stop row.

### DIFF
--- a/tests/config/test_omni_config.py
+++ b/tests/config/test_omni_config.py
@@ -382,6 +382,7 @@ def test_sub_config_fields_match_rfc_scopes():
         "enable_sleep_mode",
         "default_sampling_params",
         "subtalker_sampling_params",
+        "codec_sampling_params",
         "has_sampling_extra_args",
         "custom_voice_dir",
         "task_type",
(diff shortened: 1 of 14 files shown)


amy-why-3459:
@vllm-omni-review-bot

natureofnature:
@codex review

Sy0307:
On the current native-duplex H200 path, a two-turn short-input case (600 ms then 1200 ms) crashed Stage 1 with `ValueError: could not broadcast input array from shape (4111,) into shape (4096,)`. This is not covered by the non-duplex codec budget tests. Please either verify that duplex prompt growth is bounded by this change or explicitly keep it out of scope with a follow-up, because short-segment inputs can currently kill the Talker stage.

PR #6277, opened 2026-08-17
[Feature][Diffusion] Add scheduler registry and injection seam for diffusion pipelines

knlnguyen1802:
Fix [Feature][Diffusion] Add scheduler registry and injection seam for diffusion pipelines

## Purpose
Add a supported way to inject a custom diffusion scheduler class into stock pipelines.

- New `vllm_omni.diffusion.models.schedulers.registry`: `register_scheduler`, `resolve_scheduler_cls` (registry name / dotted path / class), and `build_pipeline_scheduler`. Resolution order: explicit arg > `od_config.scheduler` > pipeline `default_builder`.
- External packages can advertise classes via the empty `vllm_omni.schedulers` entry-point group in `pyproject.toml`.
- New `OmniDiffusionConfig.scheduler` / `scheduler_kwargs`, plumbed through `OmniEngineArgs` and `AsyncOmniEngine` stage kwargs. Same pattern as `custom_pipeline_args` (Python API / deploy YAML engine args; no new CLI flags).
- Mechanical swap of the five hardcoded sites (`flux`, `qwen_image`, `sd3`, `ltx2`, `wan2_2`) to `build_pipeline_scheduler`. Unconfigured path calls each pipeline's existing `default_builder` unchanged.
- `wan2_2` skips its per-request `sample_solver` / `flow_shift` rebuild (with a warning) when an injected scheduler is active, so the injected object is not replaced.

Example:

```python
engine = AsyncOmni(
    model="Qwen/Qwen-Image",
    scheduler="my_pkg.FlowMatchSDEDiscreteScheduler",  # registry name or dotted path
    scheduler_kwargs={...},  # optional, forwarded to from_pretrained
)
```
## Test Plan

**vLLM Version:**

**vLLM-Omni Commit:**

L1 (CPU) — registry / factory / contract:

```bash
pytest -s -v tests/diffusion/models/schedulers/test_registry.py -m "core_model and cpu"
```

L2 (CUDA L4, `tiny-random/Qwen-Image`) — inject SDE scheduler into stock Qwen-Image; control run without injection:

```bash
pytest -s -v tests/e2e/features/scheduler_injection/test_scheduler_injection.py \
  -m "core_model and diffusion" --run-level=core_model
```

Prerequisites: CUDA L4, `tiny-random/Qwen-Image` cached locally. L1 is already collected by the Simple Unit Test step (`tests/diffusion -m 'core_model and cpu'`).

## Test Result

- L1 `tests/diffusion/models/schedulers/test_registry.py`: added; covers register/resolve/entry points, `build_pipeline_scheduler` resolution order, kwargs merge, default-path bit-identical fallback, and `set_timesteps` `sigmas` signature (required by `retrieve_timesteps`).
- GPU e2e on L4 + `tiny-random/Qwen-Image`:
  - `test_default_scheduler_path_without_injection` passed (default construction path does not touch the injected class).
  - Injected path: dummy warmup and image generate succeed with `scheduler=` set to `FlowMatchSDEDiscreteSchedulerForTest` on stock Qwen-Image (no `custom_pipeline_args`). Marker file asserts `constructed` / `stepped` after env-gated hooks on that class.

cc: @SamitHuang @zhtmike @wtomin 

### DIFF
--- a/tests/e2e/features/helpers/custom_pipeline.py
+++ b/tests/e2e/features/helpers/custom_pipeline.py
@@ -27,6 +27,7 @@ from diffusers.utils import BaseOutput
 from diffusers.utils.torch_utils import randn_tensor
 from msgspec import field
 
+from tests.e2e.features.helpers.custom_scheduler import mark_scheduler_event
 from vllm_omni.diffusion.data import DiffusionOutput, OmniDiffusionConfig
 from vllm_omni.diffusion.distributed.utils import get_local_device
 from vllm_omni.diffusion.models.qwen_image import QwenImagePipeline
@@ -69,7 +70,17 @@ class FlowMatchSDEDiscreteSchedulerForTest(FlowMatchEulerDiscreteScheduler):
     """SDE version of the FlowMatchEulerDiscreteScheduler.
     The implementation is based on FlowGRPO paper (https://arxiv.org/abs/2505.05470)
     and diffusers v0.37 branch.
+
+    When ``VLLM_OMNI_CUSTOM_SCHEDULER_MARKER`` is set, construction and
+    ``step`` append marker lines so scheduler-injection tests can observe
+    this class from the pytest process. ``set_timesteps`` is inherited
+    unchanged so Qwen-Image's ``retrieve_timesteps`` still sees ``sigmas``.
     """
+
+    @classmethod
+    def from_pretrained(cls, *args, **kwargs):
+        mark_scheduler_event("constructed")
+        return super().from_pretrained(*args, **kwargs)
 
     def step(
         self,
@@ -122,6 +133,7 @@ class FlowMatchSDEDiscreteSchedulerForTest(FlowMatchEulerDiscreteScheduler):
             return_logprobs (`bool`, *optional*, defaults to True):
                 Whether to return log probabilities of the previous sample.
         """
+        mark_scheduler_event("stepped")
 
         if isinstance(timestep, (int, torch.IntTensor, torch.LongTensor)):
             raise ValueError(
--- a/vllm_omni/diffusion/data.py
+++ b/vllm_omni/diffusion/data.py
@@ -46,6 +46,17 @@ def normalize_omni_diffusion_kwargs(kwargs: Mapping[str, Any]) -> dict[str, Any]
     """Normalize legacy diffusion kwargs before config construction."""
     normalized = dict(kwargs)
 
+    # OmniEngineArgs / deploy YAML use diffusion_scheduler so users can
+    # distinguish it from stage scheduler_cls (request batching).
+    if normalized.get("diffusion_scheduler") is not None:
+        normalized["scheduler"] = normalized.pop("diffusion_scheduler")
+    else:
+        normalized.pop("diffusion_scheduler", None)
+    if normalized.get("diffusion_scheduler_kwargs") is not None:
+        normalized["scheduler_kwargs"] = normalized.pop("diffusion_scheduler_kwargs")
+    else:
+        normalized.pop("diffusion_scheduler_kwargs", None)
+
     # Backwards-compatibility: older callers may use a diffusion-specific
     # "static_lora_scale" kwarg. Normalize it to the canonical "lora_scale".
     if "static_lora_scale" in normalized:
@@ -758,9 +769,11 @@ class OmniDiffusionConfig:
     # Custom pipeline arguments for custom pipelines
     custom_pipeline_args: dict[str, Any] | None = None
 
-    # Scheduler injection: registry name (see vllm_omni.diffusion.models.schedulers)
-    # or dotted class path of a scheduler class to use instead of the pipeline's
-    # default scheduler. Constructed via cls.from_pretrained(model, subfolder="scheduler").
+    # Diffusion sampling-scheduler injection (not the request scheduler).
+    # Registry name (see vllm_omni.diffusion.models.schedulers) or dotted
+    # class path. Constructed via build_pipeline_scheduler / from_pretrained.
+    # Set from OmniEngineArgs.diffusion_scheduler. Pipelines that do not
+    # consume this field fail at load (ensure_scheduler_consumed).
     scheduler: str | None = None
     # Extra kwargs forwarded to the injected scheduler's from_pretrained().
     scheduler_kwargs: dict[str, Any] | None = None
--- a/vllm_omni/diffusion/data.py
+++ b/vllm_omni/diffusion/data.py
@@ -758,6 +758,13 @@ class OmniDiffusionConfig:
     # Custom pipeline arguments for custom pipelines
     custom_pipeline_args: dict[str, Any] | None = None
 
+    # Scheduler injection: registry name (see vllm_omni.diffusion.models.schedulers)
+    # or dotted class path of a scheduler class to use instead of the pipeline's
+    # default scheduler. Constructed via cls.from_pretrained(model, subfolder="scheduler").
+    scheduler: str | None = None
+    # Extra kwargs forwarded to the injected scheduler's from_pretrained().
+    scheduler_kwargs: dict[str, Any] | None = None
+
     # Diffusion model loading format
     # "default", "custom_pipeline", "dummy", "diffusers" (HF diffusers adapter)
     diffusion_load_format: str = "default"
--- a/vllm_omni/diffusion/models/flux/pipeline_flux.py
+++ b/vllm_omni/diffusion/models/flux/pipeline_flux.py
@@ -29,6 +29,7 @@ from vllm_omni.diffusion.models.dmd2 import DMD2PipelineMixin
 from vllm_omni.diffusion.models.flux import FluxTransformer2DModel
 from vllm_omni.diffusion.models.flux.flux_pipeline_mixin import FluxPipelineMixin
 from vllm_omni.diffusion.models.interface import SupportsComponentDiscovery
+from vllm_omni.diffusion.models.schedulers import build_pipeline_scheduler
 from vllm_omni.diffusion.models.t5_encoder import T5EncoderModel
 from vllm_omni.diffusion.profiler.diffusion_pipeline_profiler import DiffusionPipelineProfilerMixin
 from vllm_omni.diffusion.utils.tf_utils import get_transformer_config_kwargs
@@ -106,8 +107,11 @@ class FluxPipeline(
         # Check if model is a local path
         local_files_only = os.path.exists(model)
 
-        self.scheduler = FlowMatchEulerDiscreteScheduler.from_pretrained(
-            model, subfolder="scheduler", local_files_only=local_files_only
+        self.scheduler = build_pipeline_scheduler(
+            od_config,
+            default_builder=lambda: FlowMatchEulerDiscreteScheduler.from_pretrained(
+                model, subfolder="scheduler", local_files_only=local_files_only
+            ),
         )
         self.text_encoder = CLIPTextModel.from_pretrained(
             model, subfolder="text_encoder", local_files_only=local_files_only
--- a/vllm_omni/diffusion/models/flux/pipeline_flux.py
+++ b/vllm_omni/diffusion/models/flux/pipeline_flux.py
@@ -112,6 +112,7 @@ class FluxPipeline(
             default_builder=lambda: FlowMatchEulerDiscreteScheduler.from_pretrained(
                 model, subfolder="scheduler", local_files_only=local_files_only
             ),
+            local_files_only=local_files_only,
         )
         self.text_encoder = CLIPTextModel.from_pretrained(
             model, subfolder="text_encoder", local_files_only=local_files_only
--- a/vllm_omni/diffusion/models/ltx2/ltx2_components.py
+++ b/vllm_omni/diffusion/models/ltx2/ltx2_components.py
@@ -30,6 +30,7 @@ from vllm_omni.diffusion.distributed.autoencoders.autoencoder_kl_ltx2 import Dis
 from vllm_omni.diffusion.distributed.utils import get_local_device
 from vllm_omni.diffusion.model_loader.diffusers_loader import DiffusersPipelineLoader
 from vllm_omni.diffusion.model_loader.hub_prefetch import from_pretrained_with_prefetch, prefetch_subfolders
+from vllm_omni.diffusion.models.schedulers import build_pipeline_scheduler
 from vllm_omni.diffusion.offloader.module_collector import ModuleDiscovery
 
 if TYPE_CHECKING:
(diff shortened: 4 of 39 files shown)


vllm_omni/engine/arg_utils.py
```diff
@@ -231,6 +235,11 @@ def _add_omni_specific_args(cls, parser: argparse.ArgumentParser) -> argparse.Ar
     output_modalities: list[str] | None = None
     log_stats: bool = False
     custom_pipeline_args: dict[str, Any] | None = None
+    # Diffusion scheduler injection (forwarded to OmniDiffusionConfig). No CLI
+    # flags are added for these (same as custom_pipeline_args); set them via
+    # the Python API or a deploy YAML's engine args.
+    scheduler: str | None = None
```
hsliuustc0106: users cannot distuinghsi this works only for diffusion

vllm_omni/diffusion/data.py
```diff
@@ -758,6 +758,13 @@ class OmniDiffusionConfig:
     # Custom pipeline arguments for custom pipelines
     custom_pipeline_args: dict[str, Any] | None = None
 
+    # Scheduler injection: registry name (see vllm_omni.diffusion.models.schedulers)
+    # or dotted class path of a scheduler class to use instead of the pipeline's
+    # default scheduler. Constructed via cls.from_pretrained(model, subfolder="scheduler").
+    scheduler: str | None = None
```
hsliuustc0106, line 777: `od_config.scheduler` is accepted on every pipeline, but the seam is only wired into 5 construction sites. On the other stock pipelines that hardcode `FlowMatch*Scheduler.from_pretrained` — qwen_image_edit (pipeline_qwen_image_edit.py:262), qwen_image_layered (:236), flux_kontext (:112), flux2 (:384), plus glm_image, hidream_image, longcat_*, hunyuan_video_1_5*, ernie_image, boogu_image, krea2, lingbot_video — the option is silently ignored and the stock scheduler runs. For the verl-omni use case that is a silent loss of the injected sampler (log-prob collection) rather than a loud failure. An accepted config field should be applied, explicitly rejected, or visibly warned about. Either extend the mechanical swap to the remaining sites (the qwen_image_edit/layered siblings at minimum), or warn/fail at pipeline init when `od_config.scheduler` is set but unconsumed.

vllm_omni/diffusion/models/ltx2/ltx2_components.py
```diff
@@ -670,11 +671,14 @@ def initialize_pipeline_components(pipeline: Any, od_config: Any) -> None:
     quant_config = getattr(od_config, "quantization_config", None)
     pipeline.transformer = create_transformer_from_config(transformer_config, quant_config=quant_config)
     _place_aux_components(pipeline)
-    pipeline.scheduler = FlowMatchEulerDiscreteScheduler.from_pretrained(
-        model,
-        subfolder="scheduler",
-        local_files_only=local_files_only,
-        revision=revision,
+    pipeline.scheduler = build_pipeline_scheduler(
```
hsliuustc0106, line 674: Two post-swap overwrites silently discard an injected scheduler:

- The `ltx2_5_full` profile sets `scheduler_use_dynamic_shifting=True` (line 127), so the injected scheduler constructed here is immediately replaced by the stock `FlowMatchEulerDiscreteScheduler.from_config(...)` at line 683 — only its `.config` survives.
- `WanT2VDMD2Pipeline.__init__` (pipeline_wan2_2.py:1074) calls `__init_dmd2__()`, which unconditionally overwrites `self.scheduler` with a `DMD2EulerScheduler` (dmd2/mixin.py:31).

The wan2_2 forward path already handles this pattern correctly: it warns and leaves the injected scheduler untouched. Please apply the same guard to both overwrites (skip + warn when `od_config.scheduler` is set), or document these pipeline variants as unsupported.

vllm_omni/diffusion/models/schedulers/registry.py
```diff
@@ -0,0 +1,125 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+
+"""Scheduler registry and construction seam for diffusion pipelines.
```
hsliuustc0106, line 4: This ships a new user-facing option (`scheduler` / `scheduler_kwargs` on the Python API and deploy YAML) plus an entry-point group, but no `docs/` coverage — the stated precedent `custom_pipeline_args` has docs/features/custom_pipeline.md. Could you add a short docs page covering the option, the entry-point group, the injected-class contract, and which pipelines/variants actually honor the seam?

vllm_omni/diffusion/models/schedulers/registry.py
```diff
@@ -0,0 +1,125 @@
...
+    cls = resolve_scheduler_cls(scheduler_cls) or resolve_scheduler_cls(getattr(od_config, "scheduler", None))
+    if cls is None:
+        assert default_builder is not None
+        return default_builder()
+    kwargs = dict(getattr(od_config, "scheduler_kwargs", None) or {})
+    kwargs.update(scheduler_kwargs or {})
+    return cls.from_pretrained(
+        od_config.model,
+        subfolder="scheduler",
+        local_files_only=getattr(od_config, "local_files_only", False),
```
hsliuustc0106: `OmniDiffusionConfig` has no `local_files_only` attribute, so this always evaluates to `False`, while the stock sites being replaced pass `os.path.exists(model)` (and ltx2 additionally passes `revision`). No live-path divergence today (stock passes True only for local-dir paths, where `from_pretrained` reads the directory regardless), but the docstring's construction contract (`local_files_only=od_config.local_files_only`) does not match reality, and a custom class keying off `local_files_only` would behave differently from the stock path. Either thread the pipeline-resolved `local_files_only`/`revision` through the factory, or document that injected loaders always see `local_files_only=False` and no revision.

tests/e2e/features/helpers/custom_pipeline.py
zhtmike, line 1: should we still use custom pipeline for rl? or normal inference pipline with scheduer injection?

if rl is serverd as first-class citzen, I suggest to depreciate custom pipeline, use normal pipeline directly.

vllm_omni/diffusion/models/flux/pipeline_flux.py
```diff
@@ -106,8 +107,11 @@ def __init__(
         # Check if model is a local path
         local_files_only = os.path.exists(model)
 
-        self.scheduler = FlowMatchEulerDiscreteScheduler.from_pretrained(
-            model, subfolder="scheduler", local_files_only=local_files_only
+        self.scheduler = build_pipeline_scheduler(
+            od_config,
+            default_builder=lambda: FlowMatchEulerDiscreteScheduler.from_pretrained(
+                model, subfolder="scheduler", local_files_only=local_files_only
+            ),
```
zhtmike, line 114: nice idea, I think other components should also be properly hooked. But it is beyond this PR's scope

tests/e2e/features/helpers/custom_pipeline.py
knlnguyen1802: For current scope of this PR yes, only removed when all component ready

hsliuustc0106:
submit an issue to the weekly meeting for this PR please

vllm_omni/diffusion/models/schedulers/registry.py
```diff
@@ -0,0 +1,125 @@
...
+    """Single construction seam replacing hardcoded scheduler ``from_pretrained`` sites.
+
+    Resolution order: explicit ``scheduler_cls`` arg > ``od_config.scheduler``
+    > pipeline default. When no scheduler is configured, ``default_builder``
+    is invoked and its return value is passed through untouched, preserving
+    the pipeline's existing construction bit-for-bit.
+    """
+    cls = resolve_scheduler_cls(scheduler_cls) or resolve_scheduler_cls(getattr(od_config, "scheduler", None))
+    if cls is None:
+        assert default_builder is not None
```
SamitHuang: build_pipeline_scheduler passes local_files_only=getattr(od_config, 'local_files_only', False) to cls.from_pretrained(). OmniDiffusionConfig does not possess a local_files_only attribute (so it always evaluates to False), whereas stock pipelines compute local_files_only = os.path.exists(model) and pass revision parameters. When an injected scheduler is configured, local offline directory loads or revisions will ignore local file and revision settings.

vllm_omni/diffusion/models/ltx2/ltx2_components.py
```diff
@@ -670,11 +671,14 @@ def initialize_pipeline_components(pipeline: Any, od_config: Any) -> None:
...
+        od_config,
+        default_builder=lambda: FlowMatchEulerDiscreteScheduler.from_pretrained(
+            model,
+            subfolder="scheduler",
+            local_files_only=local_files_only,
+            revision=revision,
+        ),
     )
     if profile.scheduler_use_dynamic_shifting:
```
SamitHuang, line 685: When profile.scheduler_use_dynamic_shifting is True in LTx2, pipeline.scheduler is unconditionally replaced by FlowMatchEulerDiscreteScheduler.from_config(...) at line 684, discarding the injected scheduler object. Similarly, DMD2PipelineMixin.__init_dmd2__() unconditionally overwrites self.scheduler with DMD2EulerScheduler(...). These post-construction overwrites should check whether an injected scheduler is active before replacing it.

vllm_omni/diffusion/models/wan2_2/pipeline_wan2_2.py
```diff
@@ -689,9 +692,20 @@ def forward(self, req: DiffusionRequestBatch) -> list[DiffusionOutput]:
         sample_solver = resolve_wan_sample_solver(first_request, default=self._sample_solver)
         flow_shift = resolve_wan_flow_shift(first_request, self.od_config)
         if sample_solver != self._sample_solver or abs(flow_shift - self._flow_shift) > 1e-6:
-            self.scheduler = build_wan_scheduler(sample_solver, flow_shift)
-            self._sample_solver = sample_solver
-            self._flow_shift = flow_shift
+            if self.od_config.scheduler is not None:
```
SamitHuang: While pipeline_wan2_2.py now guards against replacing an injected scheduler on per-request solver/shift updates, its sibling pipeline_wan2_2_i2v.py (lines 295 and 561) still uses build_wan_scheduler directly and lacks this guard in forward(). Wan2.2 Image-to-Video runs will still overwrite self.scheduler when per-request solver or flow shift parameters change.

vllm_omni/diffusion/models/schedulers/registry.py
```diff
@@ -0,0 +1,125 @@
...
+
+def build_pipeline_scheduler(
+    od_config: "OmniDiffusionConfig",
+    scheduler_cls: str | type | None = None,
+    scheduler_kwargs: dict[str, Any] | None = None,
+    default_builder: Callable[[], Any] | None = None,
+):
+    """Single construction seam replacing hardcoded scheduler ``from_pretrained`` sites.
+
+    Resolution order: explicit ``scheduler_cls`` arg > ``od_config.scheduler``
```
SamitHuang, line 149: Using assert default_builder is not None as a control flow check in production code will fail when Python runs with optimization flags (-O), which strip assert statements. If stripped, return default_builder() will raise a TypeError when no scheduler is configured rather than a descriptive error. Replace this assertion with an explicit if default_builder is None: check and raise a ValueError.

SamitHuang (commented):
This PR introduces `vllm_omni.diffusion.models.schedulers.registry` and `build_pipeline_scheduler` to allow injecting custom diffusion scheduler classes via `OmniDiffusionConfig.scheduler` and `scheduler_kwargs`.

#### Key Findings & Action Items
1. **`local_files_only` & `revision` Handling in `build_pipeline_scheduler`**: `build_pipeline_scheduler` looks up `getattr(od_config, "local_files_only", False)`. Since `OmniDiffusionConfig` lacks a `local_files_only` field, `cls.from_pretrained` always receives `local_files_only=False` and drops any pipeline-specific `revision` argument.
2. **Injected Scheduler Overwrites in LTx2 and DMD2**: LTx2 dynamic shifting (`ltx2_components.py:683`) and `DMD2PipelineMixin.__init_dmd2__()` unconditionally overwrite `self.scheduler` with stock scheduler instances, discarding injected schedulers.
3. **Missing Guard in Wan2.2 Image-to-Video (`pipeline_wan2_2_i2v.py`)**: `pipeline_wan2_2.py` adds `build_pipeline_scheduler` and a `forward()` guard, but `pipeline_wan2_2_i2v.py` was omitted and will still overwrite `self.scheduler` on per-request parameter changes.
4. **`assert` Statement in Production Code**: `build_pipeline_scheduler` uses `assert default_builder is not None`, which gets stripped when Python is executed with optimization flags (`-O`).

knlnguyen1802:
Resolve conflict

knlnguyen1802:
@SamitHuang PTAL when you have time

vllm_omni/diffusion/data.py
```diff
@@ -56,6 +56,17 @@ def normalize_omni_diffusion_kwargs(raw_kwargs: Mapping[str, Any]) -> dict[str,
     elif not isinstance(dtype, str):
         raise TypeError(f"Provided dtype must be a string or torch.dtype, got {type(dtype).__name__}")
 
+    # OmniEngineArgs / deploy YAML use diffusion_scheduler so users can
+    # distinguish it from stage scheduler_cls (request batching).
+    if normalized.get("diffusion_scheduler") is not None:
```
lishunyang12: Use config_kwargs throughout this block; normalized is undefined.

vllm_omni/diffusion/models/schedulers/registry.py
```diff
@@ -0,0 +1,189 @@
...
+
+def is_injected_scheduler(od_config: "OmniDiffusionConfig") -> bool:
+    """True when ``od_config.scheduler`` selects a class other than the pipeline default."""
+    return getattr(od_config, "scheduler", None) is not None
+
+
+def mark_scheduler_consumed(od_config: "OmniDiffusionConfig") -> None:
+    """Record that this config's ``scheduler`` field was handled by a construction site."""
+    if is_injected_scheduler(od_config):
+        _consumed_scheduler_configs.add(od_config)
```
lishunyang12: Use WeakValueDictionary keyed by id; OmniDiffusionConfig is unhashable.

vllm_omni/diffusion/data.py
knlnguyen1802: Fixed in c1cb4f99: replaced `normalized` with `config_kwargs` throughout the diffusion_scheduler normalization block.

vllm_omni/diffusion/models/schedulers/registry.py
knlnguyen1802: Fixed in c1cb4f99: switched consumed-config tracking from `WeakSet` to `WeakValueDictionary[int, Any]` keyed by `id(od_config)` since `OmniDiffusionConfig` is unhashable.

PR #6278, opened 2026-08-17
[Core] Add AllGather-KV backend capability

0z5a:
## Purpose

AllGather-KV compatibility is an attention-backend contract. The current
validation checks `get_name() == "TRTLLM_ATTN"`, which couples the attention
layer to one implementation name and cannot represent renamed or third-party
backends with the same limitation.

This change:

- adds a `supports_allgather_kv` capability to diffusion attention backends;
- marks `TRTLLM_ATTN` as incompatible with AllGather-KV sequence parallelism;
- validates the selected backend by capability while preserving the compatible
  default for existing duck-typed custom backends.

No existing open PR was found for this backend-capability change. AI assistance
was used to prepare and pre-check the implementation.

## Test Plan

**vLLM Version:** N/A; this changes only the vLLM-Omni diffusion attention
backend contract.

**vLLM-Omni Implementation Commit:** `8409be6a38c5bbbe9827bbef8e0f0710e56c6477`

- Run the targeted backend capability and attention configuration tests.
- Run the neighboring sequence-parallel attention tests.
- Compare the existing two-rank AllGather-KV path against its single-rank
  baseline on two A100 GPUs.
- Run all pre-commit hooks on the final diff against `origin/main`.

## Test Result

- `pytest -q tests/diffusion/attention/test_attention_config.py tests/diffusion/attention/test_trtllm_attn.py::test_backend_does_not_support_allgather_kv`
  - 55 passed
- `pytest -q tests/diffusion/attention/test_attention_sp.py -k "not test_sequence_parallel" tests/diffusion/attention/test_trtllm_attn.py`
  - 36 passed, 10 skipped, 4 deselected
- Two NVIDIA A100-PCIE-40GB GPUs, CUDA 13.0, PyTorch 2.11.0, NCCL 2.28.9:
  - ran the existing `allgather-kv` sequence-parallel case directly with
    `allgather_degree=2` and `TORCH_SDPA`;
  - single-rank and two-rank outputs matched exactly: `max_abs_diff=0` and
    `max_relative_diff=0`;
  - NCCL reported direct CUDA P2P channels: `isAllDirectP2p=1` and
    `P2P/CUMEM/read`.
- `uvx pre-commit run --from-ref origin/main --to-ref HEAD`
  - all applicable hooks passed.

The parametrized pytest node is decorated for four GPUs because it shares a
test function with the 2x2 Ulysses/Ring case. The two-GPU AllGather-KV parameter
was invoked directly with the unchanged test body and oracle.


0z5a:
Self-review complete. I verified that the capability check runs only when AllGather-KV is active, that the default remains compatible for existing and duck-typed backends, and that TRTLLM alone opts out through the new capability. The tests cover the default, inactive, renamed-incompatible, and legacy-backend cases; the documented two-GPU parity result and current CI are green. I found no additional issue in the final diff.

hsliuustc0106 (commented):
LGTM — zero findings. Behavior-preserving swap from the `TRTLLM_ATTN` name check to a capability attribute: the layer.py site is the only AllGather-KV backend-compat gate, duck-typed custom backends keep the compatible default, and the new CPU-lane tests cover the full dispatch matrix including a renamed backend.

Dong1017:
**General review: zero findings.** Behavior-preserving swap from the `TRTLLM_ATTN`
name check to a capability attribute; the removed TODO is resolved.

Verified against head `3f0812c5` (base `01a2f932`):

- The gate at `vllm_omni/diffusion/attention/layer.py:167-176` stays double-gated
  (`allgather_degree > 1` and `not skip_sequence_parallel`), and `skip_sequence_parallel`
  layers also route through `NoParallelAttention` at runtime — the capability is only
  consulted when AllGather-KV is actually active.
- `getattr(cls, "supports_allgather_kv", True)` keeps duck-typed backends on the
  compatible default, and the rejection set is unchanged (only `TrtllmAttentionBackend`
  declares `False`), so behavior is equivalent to the old name check.
- The check runs after the paged-KV fallback reassignment (`layer.py:154-155`),
  so it validates the finally selected backend. No other name-based AllGather-KV
  gates remain in the repo.
- No docs change needed: `docs/user_guide/diffusion/attention_backends/trtllm.md`
  already describes the user-visible restriction accurately.

This implements a minimal instance of RFC #7226 (its motivation cites exactly this
TODO, and this matches its migration step "replace backend-name checks with resolved
capability checks"); worth cross-referencing in the PR body. When #7226's path-specific
capability resolution lands, this class-level flag and its `True` default should be
folded in there.

CI green (build 3.11/3.12, pre-commit, DCO).

PR #6281, opened 2026-08-17
[Perf] Fuse MiniMax H3 modulation with FP32 accumulation

gcanlin:
## Summary

This PR fuses the MiniMax H3 modulation operations into dedicated Triton kernels using FP32 accumulation. It removes redundant gather, elementwise, RMSNorm, and intermediate BF16 operations from each transformer block.

## Performance

Benchmarked on MiniMax H3 Ref2VA with USP=4, eager execution, a 5-second output, and 49 denoising iterations. Both configurations were warmed up before profiling and end-to-end measurement.

### End-to-End Performance

| Metric | Baseline | FP32 Fused Modulation | Improvement |
| --- | ---: | ---: | ---: |
| Engine E2E latency | 136.961 s | 134.707 s | 2.253 s / 1.65% |
| Diffusion engine execution | 134.695 s | 132.199 s | 2.496 s / 1.85% |
| Reported step latency | 2.739 s | 2.694 s | 45.1 ms / 1.65% |

### Single DiT Forward Profile

| Metric | Baseline | FP32 Fused Modulation | Improvement |
| --- | ---: | ---: | ---: |
| GPU wall time | 2,575.64 ms | 2,534.49 ms | 41.16 ms / 1.60% |
| CUDA kernel count | 1,969 | 1,234 | 735 / 37.33% |
| Total CUDA kernel duration | 2,571.83 ms | 2,531.96 ms | 39.87 ms / 1.55% |

The fusion removes 294 gather kernels, 490 BF16 elementwise kernels, 98 in-place add kernels, and 98 standalone RMSNorm kernels per DiT forward, replacing them with 147 fused modulation kernels.

## Accuracy

The implementation uses FP32 accumulation and casts the final results back to the destination tensor dtype. It is therefore not bit-exact with the baseline because intermediate BF16 rounding is removed.

| Metric | Result |
| --- | ---: |
| PSNR | 39.55 dB |
| SSIM | 0.9750 |

Before:

https://github.com/user-attachments/assets/414c48de-5db8-4757-acf6-baaa75a75246

After:

https://github.com/user-attachments/assets/e927dc58-b6e2-4d82-93e8-8fb964c70fe6

### DIFF
--- a/vllm_omni/diffusion/attention/ops/minimax_h3_modulation.py
+++ b/vllm_omni/diffusion/attention/ops/minimax_h3_modulation.py
@@ -155,6 +155,9 @@ def indexed_scale_shift_(
     indices: torch.Tensor,
 ) -> torch.Tensor:
     """Apply indexed scale/shift in-place to a disposable contiguous input."""
+    if x.is_cpu:
+        x.copy_((x * (1.0 + scale.index_select(0, indices)) + shift.index_select(0, indices)).to(x.dtype))
+        return x
     rows, hidden_size = x.shape
     if rows == 0:
         return x
@@ -182,6 +185,8 @@ def indexed_gate(
     indices: torch.Tensor,
 ) -> torch.Tensor:
     """Return ``x + gate[indices] * other`` without indexed temporaries."""
+    if x.is_cpu:
+        return (x + gate.index_select(0, indices) * other).to(x.dtype)
     output = torch.empty_like(x)
     rows, hidden_size = x.shape
     if rows == 0:
@@ -213,6 +218,13 @@ def rms_norm_indexed_scale_shift(
     eps: float,
 ) -> torch.Tensor:
     """Fuse H3 RMSNorm with its indexed AdaLN affine transform."""
+    if x.is_cpu:
+        input_dtype = x.dtype
+        normalized = x.float()
+        variance = normalized.pow(2).mean(-1, keepdim=True)
+        normalized = normalized * torch.rsqrt(variance + eps)
+        normalized = (weight.float() * normalized).to(input_dtype)
+        return (normalized * (1.0 + scale.index_select(0, indices)) + shift.index_select(0, indices)).to(input_dtype)
     output = torch.empty_like(x)
     rows, hidden_size = x.shape
     if rows:
@@ -246,6 +258,17 @@ def indexed_gate_rms_norm_scale_shift(
     eps: float,
 ) -> tuple[torch.Tensor, torch.Tensor]:
     """Fuse gated residual, the following RMSNorm, and AdaLN affine."""
+    if residual.is_cpu:
+        input_dtype = residual.dtype
+        residual_out = (residual + gate.index_select(0, indices) * branch).to(input_dtype)
+        normalized = residual_out.float()
+        variance = normalized.pow(2).mean(-1, keepdim=True)
+        normalized = normalized * torch.rsqrt(variance + eps)
+        normalized = (weight.float() * normalized).to(input_dtype)
+        modulated_out = (normalized * (1.0 + scale.index_select(0, indices)) + shift.index_select(0, indices)).to(
+            input_dtype
+        )
+        return residual_out, modulated_out
     residual_out = torch.empty_like(residual)
     modulated_out = torch.empty_like(residual)
     rows, hidden_size = residual.shape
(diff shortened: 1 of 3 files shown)


gcanlin:
cc @hsliuustc0106 @david6666666 @lishunyang12

Sy0307 (approved):
LGTM

vllm_omni/diffusion/attention/ops/minimax_h3_modulation.py
```diff
@@ -0,0 +1,305 @@
...
+
+
+def indexed_scale_shift_(
+    x: torch.Tensor,
+    shift: torch.Tensor,
+    scale: torch.Tensor,
+    indices: torch.Tensor,
+) -> torch.Tensor:
+    """Apply indexed scale/shift in-place to a disposable contiguous input."""
+    if x.is_cpu:
```
hsliuustc0106, line 158: All four wrappers gate the Triton path on `x.is_cpu` only, so NPU (and XPU/MUSA) tensors fall through to a raw Triton launch. MiniMax H3 is served on Ascend — this same transformer file opts NPU into `ascend_laser_attention` and the packed `npu_attn_varlen` path, and these blocks currently run `norm1`/`norm2` through `RMSNorm.forward_npu` → `torch_npu.npu_rms_norm`. After this change an NPU forward reaches `_rms_norm_indexed_scale_shift_kernel` with torch_npu tensors on the first block and fails at the launch. The npu-ci lane doesn't run H3, so CI can't catch this.

The in-repo precedent is `_fused_cuda_supported` (`vllm_omni/diffusion/layers/fused_qk_norm_rope.py:133`): fuse only when `HAS_TRITON and current_platform.is_cuda() and x.is_cuda` (+ dtype check), otherwise run the eager reference. Since the eager bodies already exist here, I'd extend the gate the same way; on the non-CUDA path it's also worth routing the norm through the `RMSNorm` module so NPU keeps `npu_rms_norm` instead of the bf16 eager fallback.

vllm_omni/diffusion/attention/ops/minimax_h3_modulation.py
```diff
@@ -0,0 +1,305 @@
...
+        gate.stride(0),
+        other.stride(0),
+        indices.stride(0),
+        block_n=triton.next_power_of_2(hidden_size),
+        num_warps=8,
+    )
+    return output
+
+
+def rms_norm_indexed_scale_shift(
```
hsliuustc0106, line 212: No test covers the new module. A CPU test in the style of `tests/diffusion/models/minimax_h3/test_minimax_h3_rmsnorm.py` would pin the fallback math almost for free. Worth stating in the test (or the module docstring) that the CPU fallback intentionally keeps the old two-step numerics — norm rounded to input dtype, then affine — while the Triton kernel single-rounds in FP32, so CPU asserts validate the fallback and the PR-level PSNR is the GPU-parity evidence.

vllm_omni/diffusion/attention/ops/minimax_h3_modulation.py
```diff
@@ -0,0 +1,305 @@
...
+
+    x = tl.load(x_ptr + row * stride_x_row + columns, mask=mask, other=0.0).to(tl.float32)
+    weight = tl.load(weight_ptr + columns, mask=mask, other=0.0).to(tl.float32)
+    variance = tl.sum(x * x, axis=0) / hidden_size
+    normalized = x * tl.rsqrt(variance + eps) * weight
+
+    shift = tl.load(shift_ptr + index * stride_shift_row + columns, mask=mask, other=0.0).to(tl.float32)
+    scale = tl.load(scale_ptr + index * stride_scale_row + columns, mask=mask, other=0.0).to(tl.float32)
+    tl.store(
+        output_ptr + row * hidden_size + columns,
```
hsliuustc0106, line 101: Minor: the rms-norm kernels write outputs via `row * hidden_size` and all loads/stores assume unit column stride, so the wrappers only accept row-major contiguous inputs. Today's callers pass fresh contiguous tensors, but a strided input would silently read/write the wrong elements. An `x.is_contiguous()` assert (or a stated requirement in the module docstring) would pin the assumption.

hsliuustc0106:
amd-ci is red on head (build 10670) while the CUDA/Intel/NPU lanes pass; the logs are behind auth on my side. The AMD lane runs the CPU-marked diffusion tests on a ROCm host, and the CPU fallbacks here should cover those, so I can't correlate the failure from the diff — could you check whether it's related or infra?

Issue #6271, opened 2026-08-17
[RFC]: Serve OpenVLA-7B as an autoregressive robot policy


MrlixiangWE:
### Motivation.

Every robot policy vLLM-Omni hosts today produces actions by denoising: DreamZero,
InternVLA-A1, GR00T-N1.7, Cosmos3, and now π0 (#4222). OpenVLA does not. It has no action
head at all — it reserves the last 256 ids of the Llama-2 vocabulary as bins and emits a
7-DoF action as seven ordinary generated tokens, so it runs as a plain AR stage.

#1987 already anticipates this. Its text says that "VLA models (Pi0, OpenVLA) do NOT require
the state or engine changes — they work with existing infrastructure plus **action I/O
support**", and OpenVLA appears in its robotics scenario table. This RFC proposes to build
that action I/O support, with OpenVLA as the first model through it — 80.6k downloads,
second only to `lerobot/smolvla_base` among open VLAs on the Hub.

**The model itself needs no new code here, and I am not claiming otherwise.** Upstream vLLM
v0.27.0 — the version vLLM-Omni pins — already ships a complete, registered
`OpenVLAForActionPrediction` (`vllm/model_executor/models/openvla.py`, plus a config shim
and a timm-free image processor), and `OmniModelRegistry` merges `_VLLM_MODELS` wholesale.
A stage declaring `model_arch="OpenVLAForActionPrediction"` resolves to that class with
**zero** new `_OMNI_MODELS` entries. This proposal is only the robot-facing half.

That half is missing in both repositories:

| gap | evidence |
|---|---|
| No action de-tokenisation: `bin = unpadded_vocab_size - token_id - 1` (31999, from the *unpadded* 32000, not the config's padded 32064), clip into `bin_centers`, then un-normalise with the checkpoint's per-embodiment `q01/q99` — except where `mask` is `False`, which in every one of openvla-7b's 25 embodiments is the gripper dimension. | vLLM reads `n_action_bins` at `openvla.py:402` and never uses it again, and has no `unnorm_key` anywhere. vLLM-Omni *does* have per-embodiment de-normalisation — DreamZero's `_denormalize_action`, and π0's `norm_stats` handling from #4222 — but both map a continuous action vector out of a diffusion head. Nothing anywhere maps discrete token ids onto bins, which is the step OpenVLA needs. |
[...]

PR #6286, opened 2026-08-17
[Model] Serve OpenVLA-7B as an autoregressive robot policy

MrlixiangWE:
## Purpose

Serve `openvla/openvla-7b` as a robot policy over `/v1/realtime/robot/openpi`. Follows RFC #6271.

## Propose

No model code is added — vLLM 0.27.0 already ships a registered `OpenVLAForActionPrediction` and
`OmniModelRegistry` merges `_VLLM_MODELS` wholesale, so this adds **zero** `_OMNI_MODELS` entries.
It is the robot-facing half that was missing:

(1) **Action de-tokenisation.** `bin = unpadded_vocab_size - token_id - 1` (31999, from the *unpadded*
32000, not the config's padded 32064), clip into `bin_centers`, un-normalise with the checkpoint's
per-embodiment `q01/q99` except where `mask` is False.

(2) **The trailing empty token.** Training wrapped the turn as `"In: {msg}\nOut: "`; the prompt builder
right-strips it, so `predict_action` re-inserts token id `29871`.

(3) **An AR path through the OpenPI endpoint,** which was diffusion-only.

(4) **A `robot_adapter` dotted path on `StagePipelineConfig`,** like the existing
`custom_process_input_func` / `scheduler_cls`. The generic serving layer holds no model names.

Each of those fails silently — a wrong vocab size shifts every bin by 64, a missing `29871` or an
ignored `mask` still returns a well-formed 7-vector.

## Test Plan

```bash
pytest tests/model_executor/models/test_openvla_action_decode.py \
       tests/model_executor/models/test_openvla_registration.py \
       tests/entrypoints/openai_api/test_openpi_ar_action_policy.py \
       tests/entrypoints/openai_api/test_openpi_serving.py

pytest tests/config tests/engine/test_stage_engine_args.py   # on main and on this branch

pytest -s -v tests/e2e/online_serving/test_openvla_openpi.py --run-level full_model
```

**vLLM Version:** 0.27.0

**vLLM-Omni Commit:** 170d4987

## Test Result

1x A800 80GB, bf16, TP=1, greedy.

| suite | result |
| --- | --- |
| unit tests (31 new + 14 existing) | 45 passed |
| `tests/e2e/online_serving/test_openvla_openpi.py` (new) | 2 passed in 51.29s |
| `pre-commit` | clean (`actionlint` skipped) |

Shared `stage_config.py` change, same machine both sides:

| config suites | result |
| --- | --- |
| unmodified main (`170d4987`) | 10 failed, **553 passed**, 2 skipped |
| this branch | 10 failed, **557 passed**, 2 skipped |

Identical failure sets, both non-empty, pre-existing and network-bound. The +4 are this PR's registry
parametrisations.

### agreement with the reference implementation

| check | result |
| --- | --- |
| over the real WebSocket, `bridge_orig` episodes | **17 of 18** bit-identical to `predict_action` |
| offline, 10 observations | 8/10 identical |
| offline, teacher-forced next-token argmax | 69/70 positions identical |
| the five real photographs | 5/5 identical |

Every disagreement is a one-bin flip where the reference's own top-1/top-2 logprob gap is 0.0625–0.125,
i.e. bf16 rounding resolution.

### latency, distinct image per request

vLLM caches by image hash, so reusing one frame skips the vision encoder and reports 89 ms against a
true 145 ms.

| per action | reference | offline engine | over this endpoint |
| --- | --- | --- | --- |
| one robot | 206.6 ms (4.84 Hz) | 145.4 ms (6.88 Hz) | 183.4 ms (5.45 Hz) |

### concurrency

The reference raises `ValueError("Generation with batch size > 1 is not currently supported!")`, so N
robots there need N model copies.

| N | 1 | 2 | 4 | 8 | 16 |
| --- | --- | --- | --- | --- | --- |
| aggregate | 6.9 actions/s | 11.5 | 17.8 | 26.0 | 32.9 |
| per robot | 6.88 Hz | 5.77 Hz | 4.44 Hz | 3.25 Hz | 2.06 Hz |

cc @tzhouam @gcanlin

### DIFF
--- a/docs/models/supported_models.md
+++ b/docs/models/supported_models.md
@@ -25,6 +25,7 @@ th {
 | `BagelForConditionalGeneration` | BAGEL (DiT-only) | `ByteDance-Seed/BAGEL-7B-MoT` | ✅︎ | ✅︎ | | ✅︎ | — |
 | `InternVLAA1Pipeline` | InternVLA-A1 | `InternRobotics/InternVLA-A1-3B` | ✅︎ | ✅︎ | | | — |
 | `Gr00tN1d7Pipeline` | GR00T N1.7 | `nvidia/GR00T-N1.7-3B` | ✅︎ | | | | — |
+| `OpenVLAForActionPrediction` | OpenVLA | `openvla/openvla-7b` | ✅︎ | | | | [Repository](https://github.com/vllm-project/vllm-omni/blob/main/recipes/OpenVLA/OpenVLA-7B.md) |
 | `HunyuanImage3ForCausalMM` | HunyuanImage3.0 (DiT-only) | `tencent/HunyuanImage-3.0`, `tencent/HunyuanImage-3.0-Instruct` | ✅︎ | ✅︎ | ✅︎ | ✅︎ | — |
 | `QwenImagePipeline` | Qwen-Image | `Qwen/Qwen-Image` | ✅︎ | | | | [Published](https://recipes.vllm.ai/Qwen/Qwen-Image) |
 | `QwenImagePipeline` | Qwen-Image-2512 | `Qwen/Qwen-Image-2512` | ✅︎ | | | | [Repository](https://github.com/vllm-project/vllm-omni/blob/main/recipes/Qwen/Qwen-Image-2512.md) |
--- a/recipes/OpenVLA/OpenVLA-7B.md
+++ b/recipes/OpenVLA/OpenVLA-7B.md
@@ -1,0 +1,153 @@
+# OpenVLA-7B
+
+> OpenVLA-7B served as an autoregressive robot policy over the OpenPI WebSocket protocol
+
+## Summary
+
+- Vendor: OpenVLA (Stanford / Google DeepMind / Toyota Research Institute)
+- Model: `openvla/openvla-7b`
+- Task: Vision-Language-Action (VLA) inference for robot manipulation
+- Mode: Online serving via the OpenPI WebSocket endpoint
+- Hardware used for the numbers below: 1x NVIDIA A800 80GB
+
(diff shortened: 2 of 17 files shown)


MrlixiangWE:
Self-review.

Four existing files are touched. Every new branch is gated on `robot_adapter` / `ar_adapter` being
non-None, which is None for every existing deployment; the 13 existing OpenPI tests pass unmodified.

The four silent-failure modes each have a test: unpadded vs padded vocab size, the trailing `29871`,
the per-dimension `mask`, and `q01/q99` from the wrong embodiment.

The e2e case pins no golden action vector on purpose — the model's own top-1/top-2 margin drops to
0.125 nat on the synthetic frames, inside bf16 rounding, so it asserts the action grid instead.

@vllm-omni-review-bot

MrlixiangWE:
All GitHub checks are green here — DCO, pre-commit, both wheel builds, Read the Docs.
`buildkite/vllm-omni` is the only one outstanding, and it only runs behind the `ready` label.

Test evidence is in the body: 45 unit tests pass, the registry-wide config suites go
553 -> 557 passed with the same 10 pre-existing failures on both sides, actions are
bit-identical to the reference implementation on 17 of 18 real WebSocket frames, and the
new in-tree e2e case boots a real server over the msgpack socket (2 passed in 51.29s).
Happy to run anything else that would help.

@hsliuustc0106 PTAL when you have a chance.

MrlixiangWE:
All checks are green; `buildkite/vllm-omni` is the only one left and it only runs behind the `ready`
label. @Gaohan123 @hsliuustc0106 PTAL when you have a chance.

PR #6287, opened 2026-08-17
[ROCm] [CI] Fix circular import for rocm due to patch

tjtanaa:

## Purpose

**Fix ROCm circular import during vLLM-Omni initialization**

This change fixes a circular import that occurs during `vllm_omni` initialization on ROCm.

The issue is triggered by ROCm-specific patches that eagerly import the diffusion registry while the current platform is still being initialized. The diffusion LoRA implementation then imports `LoRAConfig` through the `vllm_omni.config` package, whose package initializer eagerly loads the pipeline registry. The pipeline registry imports the PI0 diffusion pipeline, which attempts to import symbols from the still-partially-initialized `vllm_omni.diffusion.data` module.

The fix makes the internal diffusion LoRA modules import `LoRAConfig` directly from its canonical upstream location, `vllm.config.lora`, instead of going through the `vllm_omni.config.lora` compatibility re-export.

## Problem

Running the following tests on ROCm failed during pytest startup and before test collection completed:

```bash
pytest -sv tests/entrypoints tests/engine -m 'core_model and cpu'
```

The relevant error was:

```text
ImportError: cannot import name 'DiffusionOutput' from partially initialized
module 'vllm_omni.diffusion.data' (most likely due to a circular import)
```

## Root cause

The circular dependency is caused by the following import sequence:

```text
vllm_omni import
  -> vllm_omni.patch
  -> model_executor.layers.rotary_embedding.mrope
  -> imports current_omni_platform
  -> constructs RocmOmniPlatform
  -> RocmOmniPlatform.__init__
  -> apply_patches
  -> ROCm worker patches
  -> patch_groupnorm
  -> imports vllm_omni.diffusion.registry
  -> imports vllm_omni.diffusion.data
  -> imports vllm_omni.diffusion.lora.manager
  -> imports vllm_omni.config.lora
  -> executes vllm_omni.config.__init__
  -> imports config_factory
  -> imports pipeline_registry
  -> imports the PI0 diffusion pipeline
  -> imports vllm_omni.diffusion.data again
```

At the final step, `vllm_omni.diffusion.data` is already present in `sys.modules`, but its initial execution has not completed. In particular, `DiffusionOutput` has not yet been defined. The PI0 pipeline therefore cannot import it from the partially initialized module.

### Why importing a submodule executes `config/__init__.py`

The diffusion LoRA code imported:

```python
from vllm_omni.config.lora import LoRAConfig
```

Even though `vllm_omni.config.lora` is a small module, Python must initialize its parent package before loading the submodule. This means Python first executes `vllm_omni/config/__init__.py`.

That package initializer exposes the broader public configuration API and imports modules including `config_factory` and `pipeline_registry`. Consequently, importing a single compatibility re-export unexpectedly loads all registered pipelines and re-enters the diffusion package.

### ROCm

`RocmOmniPlatform` applies ROCm patches from its constructor:

```python
def __init__(self):
    super().__init__()
    apply_patches()
```

The ROCm worker patch package imports `patch_groupnorm`, which immediately imports the complete diffusion registry so that it can wrap `initialize_model`:

```python
import vllm_omni.diffusion.registry as _registry_mod
```

As a result, resolving `current_omni_platform` on ROCm enters the diffusion import graph while the top-level `vllm_omni` package is still initializing. This early entry exposes the circular dependency.

### CUDA

`CudaOmniPlatform` does not have a constructor that eagerly applies an equivalent diffusion registry patch. Constructing the CUDA platform therefore does not import `vllm_omni.diffusion.registry` through this path.

The CUDA platform module imports the diffusion attention-backend registry, but that is a smaller and separate module. It does not enter the `diffusion.data -> diffusion.lora -> vllm_omni.config` chain involved in this failure.

### Intel XPU

The XPU platform performs some platform setup, such as configuring the sampler behavior and memory information function, but it also does not import the complete diffusion registry while the platform object is being constructed.

It therefore does not enter the problematic dependency chain during initial platform resolution.

## Solution

The internal diffusion LoRA modules now import `LoRAConfig` directly from vLLM:

```python
from vllm.config.lora import LoRAConfig
```

instead of importing it through the vLLM-Omni compatibility module:

```python
from vllm_omni.config.lora import LoRAConfig
```

This change was made in:

- `vllm_omni/diffusion/lora/manager.py`
- `vllm_omni/diffusion/lora/utils.py`

The resulting ROCm initialization path is reduced to:

```text
diffusion.data
  -> diffusion.lora.manager
  -> vllm.config.lora
```

It no longer executes `vllm_omni.config.__init__`, so it does not load `pipeline_registry` and does not re-enter the partially initialized `diffusion.data` module.

## Why this fix is safe

`vllm_omni.config.lora.LoRAConfig` is only a compatibility re-export of the upstream class:

```python
from vllm.config.lora import LoRAConfig
```

Therefore, both import paths resolve to the same `LoRAConfig` class. This change does not alter configuration semantics, object types, or runtime behavior. It only prevents internal code from initializing the much larger `vllm_omni.config` package unnecessarily.

The public compatibility API remains unchanged, so external users can continue to use:

```python
from vllm_omni.config.lora import LoRAConfig
```

## Test Plan

**vLLM Version:**

**vLLM-Omni Commit:**

## Test Result

Pass CI: https://buildkite.com/vllm/vllm-omni-amd-ci/builds/10662/list


yenuo26:
@vllm-omni-review-bot

yenuo26 (approved):
LGTM

Gaohan123 (approved):
LGTM. Thanks

PR #6288, opened 2026-08-17
[Model][Frontend] MiniMax-H3: Add opt-in planar video response encoding

MosCloud:
## Purpose

Reduce CPU latency and peak host memory when producing non-streaming MP4 video responses. Compatible frame layouts are converted directly into PyAV planar `gbrp` frames, avoiding a full interleaved RGB video buffer before H.264 encoding.

Encoder selection is internal and capability-based. There is no new CLI option, model allowlist, or request parameter.

## Design

The common dispatcher normalizes and validates the video payload once. It selects direct planar encoding only when all of these runtime conditions hold:

- every frame has the same HWC shape, with positive height and width and three or four channels;
- the common dtype is `uint8`, boolean, or floating point; 
- each frame's R, G, and B channel view is C-contiguous.

Other layouts use the existing encoder before a planar PyAV container is opened. The fallback reuses the prepared frames, so capability detection does not normalize or copy the video twice. Invalid inputs retain their validation errors, and errors after direct encoding begins are propagated rather than retried.

The direct path quantizes one channel at a time into PyAV's G/B/R plane order, handles padded plane strides, and reuses one channel-sized scratch array. H.264/yuv420p and AAC settings, timestamps, codec options, and response formats remain unchanged. Raw MP4 and base64 responses use the same dispatcher; fragmented streaming output is unchanged.

Every encoding call logs the selected route and effective media parameters. Fallback logs include a stable reason: `unsupported_shape`, `unsupported_dtype`, or `non_contiguous_rgb_planes`. When audio is present without a sample rate, a shared one-time message records the 24 kHz default, while each route log records the effective sample rate.

## Scope

- Encoding and muxing: `vllm_omni/entrypoints/openai/video_api_utils.py` and `vllm_omni/diffusion/utils/media_utils.py`.
- Serving integration: `vllm_omni/entrypoints/openai/serving_video.py`.
- Tests: `tests/entrypoints/openai_api/test_video_api_utils.py` and `tests/entrypoints/openai_api/test_video_server.py`.
- Reviewer benchmark: `tests/benchmarks/benchmark_video_response_encoding.py`.
- Measured deployment notes: `recipes/MiniMaxAI/MiniMax-H3.md` and `recipes/MiniMaxAI/MiniMax-H3-NPU.md`.

## Test Plan

Run the two CPU suites that cover the encoder and serving integration:

```bash
export ASCEND_RT_VISIBLE_DEVICES=
export VLLM_TARGET_DEVICE=cpu
python -m pytest -q tests/entrypoints/openai_api/test_video_api_utils.py
python -m pytest -q tests/entrypoints/openai_api/test_video_server.py
```

Run the deterministic legacy-versus-automatic benchmark:

```bash
ASCEND_RT_VISIBLE_DEVICES= VLLM_TARGET_DEVICE=cpu \
python tests/benchmarks/benchmark_video_response_encoding.py \
  --frames 124 --height 768 --width 1344 \
  --fps 24 --audio-sample-rate 32000 \
  --warmup 1 --rounds 3 --seed 20260817 \
  --output-json ${ARTIFACT_DIR:-/tmp}/pr1-video-encoding.json
```

Run static checks on the PR range:

```bash
PR_MERGE_BASE=199ef7779c1b7650a0e9df0951e0a2b9da4e0190
PR_HEAD=c947547e7ac3ccdcff8be2554c40e206bca9edfa
CHANGED_PY="$(git diff --name-only "${PR_MERGE_BASE}..${PR_HEAD}" -- '*.py')"
ruff check ${CHANGED_PY}
ruff format --check ${CHANGED_PY}
git diff --check "${PR_MERGE_BASE}..${PR_HEAD}"
```

For the production-route check, run one MiniMax-H3 `FL2VA`/`t2va` request on the Atlas A2 eight-NPU setup documented in the recipe: five seconds, 1344x768, 24 FPS, seed 1101, and 50 requested steps. Verify `selected_path=direct_planar`, HTTP 200, the expected H.264/AAC stream properties, and a full FFmpeg decode with `-xerror`. This is a correctness check, not a performance measurement.

**vLLM Version:** `0.26.0+empty`

**vLLM-Omni revision:** merge base `199ef7779c1b7650a0e9df0951e0a2b9da4e0190`; PR head `c947547e7ac3ccdcff8be2554c40e206bca9edfa`.

## Test Result

Validation on the current PR head:

- `test_video_api_utils.py`: 48 passed;
- `test_video_server.py`: 86 passed;
- Ruff check, Ruff format check, and `git diff --check`: passed;
- all legacy and automatic benchmark outputs were byte-identical.

Reproducible benchmark results on a 192-CPU Kunpeng-920 host at PR head `c947547e`:

| Metric | Legacy | Automatic planar |
| --- | ---: | ---: |
| Wall time, mean +/- sample stddev | 11.291 +/- 1.124 s | 7.232 +/- 0.122 s |
| Median wall time | 11.617 s | 7.216 s |
| Process CPU time, mean +/- sample stddev | 15.498 +/- 1.242 s | 11.592 +/- 0.263 s |
| Median process CPU time | 15.737 s | 11.599 s |

The paired wall-time saving was 4.060 +/- 1.053 s, or 35.56% +/- 5.99 percentage points. The median paired reduction was 36.63%. Means and sample standard deviations were calculated from the raw per-round records; the checked-in benchmark summary reports median, minimum, and maximum. All six measured outputs were byte-identical. The run used Python 3.12.13, PyTorch 2.10.0+cpu, NumPy 1.26.4, PyAV 18.0.0, and FFmpeg 4.4.2.

[Independent validation](https://github.com/vllm-project/vllm-omni/pull/6288#discussion_r3828656200) of the same planar conversion core on a Xeon 8480C at PR head `333ef4b0` measured response-encoding wall time decreasing from 1.805 s to 1.394 s (22.8%), process CPU time decreasing by 11.2%, and peak RSS decreasing from 3182 MiB to 2794 MiB. Outputs were byte-identical and passed media validation. This community measurement predates the automatic-dispatch update and is reported as cross-architecture evidence for the conversion itself.

An Atlas A2 eight-NPU MiniMax-H3 `FL2VA`/`t2va` correctness run used runtime commit `c7c3dfb0` (the current head adds documentation only), 50 requested steps (49 DiT forwards), and a five-second, 1344x768 request. The runtime selected `direct_planar` for 124 float32 frames with stereo 32-kHz audio. The response was HTTP 200, and the 7,027,664-byte output passed full decode with H.264/yuv420p video at 24 FPS and AAC audio. This cold request validates production routing and output integrity; the performance measurements above isolate CPU response encoding.


MosCloud:
Self-review completed. I checked:

- `legacy` remains the default, while `auto` is limited to a single diffusion stage whose metadata opts in; unknown, multi-stage, and incompatible-layout inputs retain the legacy path.
- The planar path preserves RGB-to-GBR plane mapping, padding, H.264/AAC options, timestamps, and cleanup/error behavior.
- Raw MP4 and base64 response paths propagate the resolved policy consistently.
- Added tests cover policy resolution, CLI/config propagation, planar conversion, layout fallback, audio/no-audio equivalence, and container cleanup.
- The focused CPU suites, Ruff checks, reviewer benchmark, byte-identity checks, `ffprobe`, and full FFmpeg decode all passed. The documented performance claim is limited to CPU response encoding and does not attribute any gain to DiT or stage 0.

I did not find any additional issues in the submitted diff.

vllm_omni/entrypoints/openai/serving_video.py
```diff
@@ -85,10 +108,36 @@ def __init__(
...
+            video_response_encoding_mode,
+            model_prefers_optimized=model_metadata.prefers_optimized_video_response_encoding,
+        )
+        logger.info(
+            "Video response encoding mode resolved: requested=%s optimized=%s",
+            video_response_encoding_mode,
+            self._video_response_encoding_config.optimized,
+        )
+
+    def _get_diffusion_model_metadata(self) -> DiffusionModelMetadata:
```
hsliuustc0106: This is the third copy of the od-config resolution block in this class (`preserves_reference_image_size` / `supports_mixed_reference_inputs` below), and `entrypoints/openai/utils.py` already exposes `resolve_diffusion_od_config` / `is_single_stage_diffusion` for the same lookup. Could this consolidate on those helpers (or a small private accessor), here or in a follow-up? One behavioral difference to preserve if consolidating: this path checks the constructor-injected `stage_configs` first, while `is_single_stage_diffusion` only looks at the engine.

MosCloud: Thanks, good catch. I consolidated the three `od_config` lookups behind a private `_resolve_diffusion_od_config()` accessor. It preserves the existing `get_diffusion_od_config()`-first behavior with the `od_config` attribute fallback, and it leaves constructor-injected `stage_configs` ahead of the engine client configs rather than switching to `is_single_stage_diffusion`. I added regression coverage for both stage-config precedence directions, getter priority, and attribute fallback. Focused server tests, the video API utility/CLI tests, Ruff, formatting, and `git diff --check` all pass.

hsliuustc0106:
does it only work for kunpeng CPU?

MosCloud:
> does it only work for kunpeng CPU?

I think the answer is No. However I just test it on the Atlas A2 910B with Kunpeng-920.
The implementation is CPU-architecture agnostic: it uses NumPy and PyAV/FFmpeg's planar gbrp frames, with no Kunpeng/AArch64-specific checks or instructions. The auto policy is gated by model metadata and diffusion-stage topology, not CPU type. Inputs with an incompatible layout fall back to the legacy encoder. Kunpeng-920 is only the CPU used for the reported performance and Atlas A2 E2E validation, so the PR does not claim the same speedup on other CPU platforms. Other platforms supported by PyAV/libx264 should be compatible, but their performance still needs independent validation.

hsliuustc0106:
Independent CPU A/B validation on PR #6288 (exact base 2fdbf223... vs current head 333ef4b09e026f39c8954c0283f3e7c3965461b7). The latest head commit only consolidates serving config resolution; the encoder and benchmark path are unchanged.

Using the PR's 124 x 768 x 1344, 24 FPS, audio-enabled input with one warmup and seven measured rounds on 2x Xeon 8480C (PyAV 18.0.0, NumPy 2.3.5, vLLM 0.26.0):

- Legacy: 1.805 s median wall time / 3.613 s process CPU time
- Planar: 1.394 s / 3.207 s
- Measured reduction: 22.8% wall time and 11.2% process CPU time
- Peak RSS: 3,182 MiB -> 2,794 MiB (-387 MiB, 12.2%)

A 1-8.7 second video sweep stayed around 21.8-22.6% median wall-time reduction. The PR's 37.78% paired wall-time claim was not reproduced on this host/runtime, so it appears environment-specific rather than portable.

Correctness checks passed: legacy and planar outputs were byte-identical, both passed ffprobe and ffmpeg -xerror decoding, and the focused suites pass 154 tests on the current head. This validates the encoding path only; we could not run an end-to-end MiniMax-H3 Atlas/NPU test or reproduce the captured H3 fixture.

Please consider reporting the 37.8% result as host-specific and adding mean +/- standard deviation plus a reproducible H3 fixture (and an end-to-end number if the broader serving speedup is intended). I did not find a correctness issue in the tested CPU path.

MosCloud:
Thanks for taking the time to run this on the Xeon 8480C and for sharing the detailed results.

The optimization is not limited to Kunpeng. Your validation is helpful because it confirms the output correctness and memory reduction on a different CPU architecture, while also showing that the exact speedup depends on the host and runtime. The 22.8% wall-time reduction, 11.2% process CPU reduction, and 387 MiB peak-RSS reduction are meaningful independent results.

I have updated the PR description accordingly:

- the 37.78% result is identified specifically as the measurement from the Kunpeng-920 environment;
- your Xeon result is included as independent cross-architecture validation;
- the Kunpeng results now include arithmetic mean and sample standard deviation alongside the existing median and range;
- the checked-in benchmark is described as a deterministic MiniMax-H3-compatible fixture with the same video shape, planar-backed layout, dtype, frame rate, and audio path, so it can be reproduced directly from the PR checkout.

The performance claim remains limited to CPU MP4 response encoding. It does not claim a DiT, stage-0, or general end-to-end serving speedup.

If an end-to-end measurement would be useful for the review, I can also provide a separate Atlas A2 MiniMax-H3 result with stage 0, MP4 response encoding, and total request latency reported independently.

hsliuustc0106:
@alex-jw-brooks @NickCao PTAL of the entrypoint changes

vllm_omni/entrypoints/openai/serving_video.py
hsliuustc0106: Verified — `_resolve_diffusion_od_config()` preserves the getter-first precedence and the injected-`stage_configs`-ahead-of-engine ordering, and the three new tests pin exactly those semantics (the getter-vs-attribute case would fail if precedence flipped). Thanks!

hsliuustc0106 (approved):
lgtm

yiminghub2024:
image
lgtm @hsliuustc0106 @MosCloud

vllm_omni/entrypoints/openai/video_api_utils.py
```diff
@@ -541,6 +610,59 @@ def _encode_video_bytes(
...
+        )
+
+    audio_np = _coerce_audio_to_numpy(audio) if audio is not None else None
+    return mux_av_video_audio_bytes(
+        _iter_planar_video_frames(frames, common_dtype),
+        width=frame_shape[1],
+        height=frame_shape[0],
+        audio_waveform=audio_np,
+        fps=float(fps),
+        audio_sample_rate=audio_sample_rate or 24000,
```
alex-jw-brooks: Can you set `DEFAULT_AUDIO_SAMPLE_RATE=24000` as a const and use it both here and in the legacy `_coerce_video_to_uint8_frames` call, and also add a log if it's `None` to both paths so that it's clear the default sampling rate is being used?

vllm_omni/entrypoints/openai/video_api_utils.py
```diff
@@ -520,18 +536,71 @@ def _coerce_video_to_uint8_frames(video: Any) -> np.ndarray:
...
+    frame_shape = frames[0].shape
+    if any(frame.shape != frame_shape for frame in frames[1:]):
+        raise ValueError("All video frames must have the same shape.")
+
+    common_dtype = np.result_type(*(frame.dtype for frame in frames))
+    return frames, frame_shape, common_dtype
+
+
+def _supports_direct_planar_mux(frames: list[np.ndarray], frame_shape: tuple[int, ...]) -> bool:
+    if len(frame_shape) != 3 or frame_shape[-1] not in (3, 4):
```
alex-jw-brooks: Please add a log with the `frame_shape` here so that it's clear why direct planar mux is unsupported, in case it's unexpected

vllm_omni/entrypoints/openai/video_api_utils.py
```diff
@@ -541,6 +610,59 @@ def _encode_video_bytes(
     )
 
 
+def _encode_video_bytes_optimized(
+    video: Any,
+    fps: int,
+    audio: Any | None = None,
```
alex-jw-brooks, line 693: Please add the expected types here instead of `Any`, and also remove the unused default values since the only caller always passes all of the keyword args here

vllm_omni/entrypoints/openai/video_api_utils.py
```diff
@@ -520,18 +536,71 @@ def _coerce_video_to_uint8_frames(video: Any) -> np.ndarray:
...
+
+    frame_shape = frames[0].shape
+    if any(frame.shape != frame_shape for frame in frames[1:]):
+        raise ValueError("All video frames must have the same shape.")
+
+    common_dtype = np.result_type(*(frame.dtype for frame in frames))
+    return frames, frame_shape, common_dtype
+
+
+def _supports_direct_planar_mux(frames: list[np.ndarray], frame_shape: tuple[int, ...]) -> bool:
```
alex-jw-brooks: Since we can check this directly, why not just always try the optimized path and fallback to legacy format if it's unsupported?

Since we can already fall back to the legacy path if it is not supported, it would be best to not make this user facing IMO. It would make potential standardization of the format and eventual removal of the legacy path easier since we wouldn't need to deprecate the API, as long as the outputs are expected to be bit identical

vllm_omni/entrypoints/openai/serving_video.py
```diff
@@ -85,10 +108,41 @@ def __init__(
...
+            video_response_encoding_mode,
+            model_prefers_optimized=model_metadata.prefers_optimized_video_response_encoding,
+        )
+        logger.info(
+            "Video response encoding mode resolved: requested=%s optimized=%s",
+            video_response_encoding_mode,
+            self._video_response_encoding_config.optimized,
+        )
+
+    def _resolve_diffusion_od_config(self) -> Any:
```
alex-jw-brooks: Thanks for making this common! Please update the return type to be an optional omni diffusion config

vllm_omni/entrypoints/openai/serving_video.py
```diff
@@ -40,6 +46,23 @@
...
+    mode: VideoResponseEncodingMode,
+    *,
+    model_prefers_optimized: bool,
+) -> VideoResponseEncodingConfig:
+    """Resolve a server mode without coupling the encoder to model names."""
+    if mode == "legacy":
+        return LEGACY_VIDEO_RESPONSE_ENCODING
+    if mode == "optimized":
+        return OPTIMIZED_VIDEO_RESPONSE_ENCODING
+    if mode == "auto":
```
alex-jw-brooks: Similar to above, I think it would be best to not make this a choice for users, and just infer whether or not the optimal path is supported, since then we would not need this

vllm_omni/entrypoints/openai/video_api_utils.py
MosCloud: Thanks, I agree that this should not remain a user-facing choice. I propose selecting the path internally through two gates: a model capability in DiffusionModelMetadata, defaulting to False, and the existing runtime frame-layout check. MiniMax-H3 would declare the capability because its output has been validated; compatible payloads would use the planar path, while unsupported layouts would still fall back to the legacy path before encoding starts. Other models would remain on the legacy path until their outputs have equivalent correctness and byte-identity coverage. This keeps the behavior automatic and avoids a future public-API deprecation. Does this direction work for you? I will wait for your confirmation before changing the policy in this PR.

vllm_omni/entrypoints/openai/serving_video.py
MosCloud: Agreed. I described the proposed automatic, model-capability-plus-layout-gate policy here: https://github.com/vllm-project/vllm-omni/pull/6288#discussion_r3826755906. I will wait for your feedback there before changing the user-facing mode and its plumbing.

vllm_omni/entrypoints/openai/video_api_utils.py
MosCloud: Updated in 188578d0. I added `DEFAULT_AUDIO_SAMPLE_RATE = 24000` and a shared resolver used by both the legacy and optimized encoders. When audio is present and the sample rate is `None`, it logs once that the default is being used; video-only responses do not emit a misleading audio-rate log. The tests cover both encoder paths, the default value, and preservation of an explicit sample rate.

vllm_omni/entrypoints/openai/video_api_utils.py
MosCloud: Updated in 188578d0. The optimized encoder now logs one fallback message containing `frame_shape` and distinguishes an unsupported shape from a non-contiguous channel layout. The fallback still occurs before opening the PyAV container, and the interleaved-layout test now verifies the diagnostic and legacy path.

vllm_omni/entrypoints/openai/video_api_utils.py
MosCloud: Updated in 188578d0. I added concrete video and audio input types covering Tensor/NumPy/PIL frame inputs and mono or stereo numeric audio lists. I also removed the unused defaults from the private optimized helper; its only caller already passes every argument explicitly.

vllm_omni/entrypoints/openai/serving_video.py
MosCloud: Updated in 188578d0. The accessor now has an optional concrete return type: `OmniDiffusionConfig | SimpleNamespace | None`. The `SimpleNamespace` member reflects the lightweight config view returned by the out-of-process engine, while the full in-process path returns `OmniDiffusionConfig`. The full config import is guarded by `TYPE_CHECKING` to avoid adding a runtime dependency.

vllm_omni/entrypoints/openai/video_api_utils.py
MosCloud: Updated in `c7c3dfb0`. The non-streaming dispatcher now selects the path entirely from the prepared payload. It uses direct planar encoding only for equal positive HWC RGB/RGBA shapes with a supported common dtype and C-contiguous RGB channel planes. Other layouts log a stable fallback reason and reuse the already prepared frames in the legacy encoder before a planar container is opened. Invalid input still raises, and failures after direct encoding starts propagate rather than retrying. Fixed-input tests assert complete MP4 byte equality with and without audio.

vllm_omni/entrypoints/openai/serving_video.py
MosCloud: Updated in `c7c3dfb0`. I removed `--video-response-encoding-mode`, its configuration plumbing, and the model metadata preference. Raw MP4 and base64 responses now use the same internal runtime capability decision, while fragmented streaming output is unchanged. Route selection is logged as `selected_path=direct_planar` or `selected_path=legacy_fallback`, together with the fallback reason and key media parameters.

MosCloud:
Implemented the automatic selection follow-up in `c7c3dfb0`.

- Removed the public mode flag, configuration plumbing, and model metadata opt-in.
- Added strict runtime shape, dtype, and channel-contiguity gating with logged fallback reasons.
- Reused prepared frames for fallback to avoid a second normalization and copy.
- Kept direct-path failures non-retriable and preserved byte-identical output coverage.

Validation:

- 48 video utility tests passed; 24 focused serving and CLI tests passed; the final no-retry regression passed after its mock target was corrected.
- Ruff check, Ruff format check, and `git diff --check` passed.
- Legacy and automatic benchmark outputs were byte-identical.
- An Atlas A2 eight-NPU MiniMax-H3 five-second run completed 49/49 DiT forwards, selected `direct_planar`, returned HTTP 200, and produced a valid H.264/AAC MP4 that passed full FFmpeg decode.

The E2E run is routing and correctness evidence; the performance claim remains isolated to CPU response encoding.

hsliuustc0106:
> Implemented the automatic selection follow-up in `c7c3dfb0`.
> 
> * Removed the public mode flag, configuration plumbing, and model metadata opt-in.
> * Added strict runtime shape, dtype, and channel-contiguity gating with logged fallback reasons.
> * Reused prepared frames for fallback to avoid a second normalization and copy.
> * Kept direct-path failures non-retriable and preserved byte-identical output coverage.
> 
> Validation:
> 
> * 48 video utility tests passed; 24 focused serving and CLI tests passed; the final no-retry regression passed after its mock target was corrected.
> * Ruff check, Ruff format check, and `git diff --check` passed.
> * Legacy and automatic benchmark outputs were byte-identical.
> * An Atlas A2 eight-NPU MiniMax-H3 five-second run completed 49/49 DiT forwards, selected `direct_planar`, returned HTTP 200, and produced a valid H.264/AAC MP4 that passed full FFmpeg decode.
> 
> The E2E run is routing and correctness evidence; the performance claim remains isolated to CPU response encoding.

Now, do users need to add cli flags or it's been made as default?

recipes/MiniMaxAI/MiniMax-H3-NPU.md
```diff
@@ -118,6 +118,21 @@ On Atlas 800I A3 (64 GB HBM per device) the combined service does not fit at
 above) or HSDP — see
 [§ Memory and attention optimizations](#memory-and-attention-optimizations-a3).
 
+### CPU MP4 response encoding (Atlas A2)
```
hsliuustc0106, line 121: can you also update in the GPU based on the experiment result from the community?

MosCloud:
No CLI flag is required. The automatic dispatcher is now the default for non-streaming MP4 responses. It selects the direct planar path only when the runtime frame shape, common dtype, and RGB channel-plane contiguity satisfy the required conditions; otherwise it falls back to the existing legacy encoder before opening the PyAV container. Streaming fMP4 remains unchanged.

recipes/MiniMaxAI/MiniMax-H3-NPU.md
MosCloud: Updated in c947547e. I added a CPU MP4 response-encoding section to the general MiniMax-H3 recipe used by CUDA and ROCm deployments. It explains that non-streaming MP4 encoding runs on the host CPU, documents the automatic direct-planar/fallback behavior, and includes the community Xeon 8480C results: 22.8% lower median wall time, 11.2% lower process CPU time, and 387 MiB lower peak RSS. The section labels these as host-CPU encoding results rather than GPU, DiT, or stage-0 gains. I also credited you @hsliuustc0106 as a co-author for contributing the independent benchmark data.

alex-jw-brooks (approved):
LGTM. Thanks :)

PR #6289, opened 2026-08-17
[XPU] Fix Dockerfile.xpu build

Joshna-Medisetty:
## Fix XPU Dockerfile build failure
The XPU image was failing to build. During the vLLM build, `import torch` crashed with:
`libsycl.so.9: undefined symbol: urDeviceWaitExp, version LIBUR_LOADER_0.12`

Two small changes fix it:
1. **Update driver versions to match vLLM v0.27.0.** The Intel UMD/driver `.deb`s were older than what v0.27.0 (which this Dockerfile builds) expects, so I bumped IGC, compute-runtime, and level-zero to the same versions upstream vLLM uses.
2. **Add `/opt/venv/lib` to `LD_LIBRARY_PATH`.** This is the real fix. torch's own `libsycl`/`libur_loader` live in `/opt/venv/lib`, but that path wasn't on `LD_LIBRARY_PATH`, so torch was picking up an older `libur_loader` from the base image's oneAPI libs and failing. Putting `/opt/venv/lib` first makes torch use its matching one.


hsliuustc0106:
@xuechendi PTAL

hsliuustc0106 (approved):
lgtm

xuechendi:
@hsliuustc0106 , I am asking @Joshna-Medisetty  doing more broader model test locally before merging it. Will merge once we validated

xuechendi (approved):

xuechendi:
@hsliuustc0106 @Gaohan123 , may you help to merge this PR, I tried get ready on/off multi times, but can't get CI triggered

hsliuustc0106:
> @hsliuustc0106 @Gaohan123 , may you help to merge this PR, I tried get ready on/off multi times, but can't get CI triggered

intel-ci triggered

xuechendi:
hello, @Joshna-Medisetty , starting from 0.28, vllm-xpu has its own docker image, I added a new PR here - https://github.com/vllm-project/vllm-omni/pull/6727

PR #5527, opened 2026-07-29
[Frontend][Core] Structural Data Interaction (e.g. camera) for Streaming Diffusion Generation

fhfuih:
## Purpose

> This PR is going to be repurposed as the phase 2 of #5120.
>
> We can land #6294 first (which is a subset of this PR)

This feature extends the previous "prompt update" #4652.
- It depends on the landing of at least one model with camera interactivity (e.g., LingBog World 2).
- It generalize the existing prompt update to an `InteractionCoordinator`, which will routes interactions in other future modalities (**note: this part is moved to #6294**)
- It registers camera interaction as the second available modality to `InteractionCoordinator`

### Architecture overview

```mermaid
flowchart TB
  subgraph api ["API / serving"]
    WS["serving_video_output_stream"]
  end

  subgraph runner ["DiffusionModelRunner"]
    SI["submit_interaction(request_id, interaction)"]
  end

  subgraph pipeline ["Pipeline + InteractionMixin"]
    APPLY["apply_interaction_at_chunk_boundary(state, ctx)"]
    STEP["step / denoise loop"]
    ENC["encode_prompt(...)"]
  end

  subgraph handlers ["Coordinator & Handler (co-owned by runner and pipeline)"]
    COORD["InteractionCoordinator\n(registers concrete modality handlers, route modality payload/action to concrete handlers)"]
    CH["CameraModalityHandler subclass\n(modality=camera)"]
    PH["PromptInteractionHandler\n(modality=prompt)"]
  end

  subgraph state ["StepRequestState.extra (per request)"]
    PS["prompt_session\npending, lerp, version, lock"]
    CS["camera_session\ntimeline, velocity, pending"]
    COND["conditioning\ncamera tensor, etc."]
    META["chunk_metadata\nstarted/active/completed ids"]
  end

  WS -->|session.interaction| SI
  SI -->|enqueue| COORD
  COORD -->|route enqueue & apply| CH
  COORD -->|route enqueue & apply| PH
  ENC -.->|injected at build| PH


  PH -->|read/write| PS
  CH -->|read/write| CS

  STEP --> APPLY
  APPLY -->|apply_interaction_at_chunk_boundary| COORD
  PH -->|mutates prompt_embeds + PS| state
  CH -->|writes COND + CS| state
  STEP -->|reads prompt_embeds, conditioning| state
```

## Test Plan

TBA once un-drafted

**vLLM Version:**

**vLLM-Omni Commit:**

### pytest

TBA once un-drafted

### Local test

[...]

Issue #5120, opened 2026-07-15
[RFC]: Generic Structural Data Updates for Streaming Diffusion (e.g., camera and robot actions)


fhfuih:
# Motivation

LingBot World 2.0 and ABot World show that streaming video generation is moving beyond text-only prompt updates. The support realtime video generation with camera control, where user interaction changes the generated world over time.

Similar capabilities are emerging across the ecosystem:

- SGLang has documented LingBot World 2.0 support with realtime camera control: https://docs.sglang.io/cookbook/diffusion/LingBot-World/LingBot-World-2.0
- FlashDreams (NVIDIA's world model inference framework) supports interactive camera.
- Reactor exposes LingBot World 2.0 as a realtime world-building model: https://www.reactor.inc/lingbot-world-2

**Generalizability**: This pattern will likely generalize beyond camera control. Future streaming generation models, such as robot-control or world-model systems like a possible Cosmos 3-style model, may need continuous structural updates: robot joint states, navigation goals, object transforms, scene constraints, user controls, or planner outputs.

## Our current status

We currently have prompt-based interaction and streaming video generation for Helios model. (RFC #3632, PR #4652). The next step is to generalize the mechanism from text prompt updates to structured data updates.

Meanwhile, we have merged LingBot 2.0 support in #5491 . **But its current interaction implementation is an experimental and less elegant hack. And it uses a completely different call path than the Helios ones.**:

- It wraps `AsyncOmni.generate` with external "tick" (aka chunk) manager.
- `AsyncOmni.generate` only generates one tick (aka chunk) and output latents. The external manager calls it repeatedly.
- **Problems**
  - `AsyncOmni.generate` should be **the** entrypoint, not to be further wrapped
  - When `AsyncOmni` is wrapped, it cannot be plugged in to the `/realtime/video` or any other API endpoint to really make it "realtime and streaming" for end users.
  - Semantically, `AsyncOmni.generate` produces one video (given initial prompt). Interactions and steering are after all optional.
- **Potential Blocks**
  - The Lingbot world 2 model itself needs to reuse the current Helios call path (or the interaction module that this RFC is proposing). **Including to implement step-wise execution**.
  - **Before** the model itself is plugged in to the current call path, **maybe need to investigate whether Lingbot's KV and stuff works well on the current path**
- **Roadmap (how to push forward and verify this RFC)**
  - Currently, can refactor Helio's prompt-update implementation to adapt this framework #6294
[...]

PR #4652, opened 2026-06-23
[Frontend] Add midway prompt update for streaming video generation

fhfuih:
## Purpose

Adds midway prompt update ("injection"/"steering") for chunked streaming video generation (phase 2 of #3632).

- After `video.start`, clients can send `session.prompt_update` to change the active prompt during generation. The server acknowledges with `session.prompt_update.accepted`.
- The video generation pipeline (Helios) applies updates at chunk boundaries with a configurable `transition_duration_chunks` (default 3 = Helios model's official configuration).
- Wiring spans the WebSocket handler, `AsyncOmni.add_prompt_update_async`, orchestrator/stage clients, diffusion runner, `PromptUpdateMixin`, and `InputBatch` cache invalidation via `prompt_update_version`.

https://github.com/user-attachments/assets/f66e91dc-cbe0-480e-a0e9-03ca6ea1899b

## Test Plan

- `tests/diffusion/test_prompt_update.py`
  - `TestPromptUpdateExecution`: the interoperation of pipeline inference, prompt update transition/interpolation/queuing, and input batch/scheduler
  - `TestPromptUpdateIntegration`: integration test from entrypoint to the pipeline
- `tests/entrypoints/test_async_omni_prompt_update.py`: focus on the AsyncOmni contract
- `tests/entrypoints/openai_api/test_serving_video_output_stream.py`: focus on the websocket API contract

**vLLM Version:** 0.23.0

**vLLM-Omni Commit:**

```
commit 465f1464d25cb5e20d61d7bd18d8a8f0c5d82c9a (upstream/main, origin/main, origin/HEAD, main)
Author: Alex Brooks <albrooks@redhat.com>
Date:   Tue Jun 23 21:02:29 2026 -0600

    [Feat] Support Pipeline Resolution Based on HF Config & Make Pipeline Registry Eager (#3760)

    Signed-off-by: Alex Brooks <albrooks@redhat.com>
```

### Pytest

All passed

### Performance

**Environment**: single H200, the example input, measuring client side latency

**The first chunk is excluded when computing the accumulated FPS**, as Time-To-First-Chunk is particularly slower than subsequent Chunks for Helios model (discussed in #3737).

Overall, the generation remains stable with the prompt update feature

image

**Raw log**

```
[chunk 001] bytes=1170480 frames=32 elapsed=3.00s total_bytes=1170480 total_frames=32 total_elapsed=3.00s
[...]

PR #6294, opened 2026-08-18
[Core][Diffusion] Refactor prompt updates into a modality-ready interaction framework

fhfuih:
Replace PromptUpdateMixin with a modality-handler architecture (InteractionCoordinator / InteractionMixin / PromptInteractionHandler) while keeping mid-stream interactions prompt-only.

- This is a newly extracted *phase 1* of RFC #5120 (that RFC extends the previous "prompt update" #4652)
- Phase 2 (the complete picture) is at #5527 (which is the old PR that lands camera / `multi_modal_data`). Will open a new one after this is merged

## Purpose

Builds on prompt-update (#4652) and #5120: replace `PromptUpdateMixin` / `prompt_update.py` with a pluggable interaction stack while **keeping runtime behavior prompt-only**.

### What changes and doesn't change in this PR

- Add `vllm_omni.diffusion.interaction` (`InteractionCoordinator`, `InteractionMixin`, handler ABC, prompt handler, shared types/registry).
  - Currently there is only one modality (the prompt) registered to the coordinator.
- Rename protocol from `SupportsPromptUpdate` to `SupportsInteractionApply`; runner applies interactions at chunk boundaries via Helios `InteractionMixin` and ACK event ids via stepwise metadata.
- Move related tests under `tests/diffusion/streaming_interactive_generation/`.
- User interface is not changed. Only the internal handlers change to get ready for future generalization

### Out of scope (phase 2 — #5527)

Camera handlers, `event.multi_modal_data` admission in serving/runner, and camera unit tests. Until then, non-prompt / MM payloads stay rejected as today.

Architecture for the full multi-modality design (incl. camera) is documented in #5527; this PR only wires the prompt track through that coordinator.

### Design Diagram (subset of #5527's diagram)

```mermaid
flowchart TB
  subgraph api ["API / serving"]
    WS["serving_video_output_stream"]
  end

  subgraph runner ["DiffusionModelRunner"]
    SI["submit_interaction(request_id, interaction)"]
  end

  subgraph pipeline ["Pipeline + InteractionMixin"]
    APPLY["apply_interaction_at_chunk_boundary(state)"]
    PREP["prepare_next_chunk(state)"]
    STEP["step / denoise loop"]
    ENC["encode_prompt(...)"]
  end

  subgraph handlers ["Coordinator & Handler (co-owned by runner and pipeline)"]
    COORD["InteractionCoordinator\n(registers concrete modality handlers, route modality payload/action to concrete handlers)"]
    OH["OtherFutureHandler\n(modality=other)"]
    AltPH["FutureAltPromptHandler\n(modality=prompt)"]
    PH["PromptInteractionHandler\n(modality=prompt)"]
    style OH stroke-dasharray: 5 5,opacity:0.35
    style AltPH stroke-dasharray: 5 5,opacity:0.35
  end

  subgraph state ["StepRequestState (per request)"]
    PS["interaction_sessions['prompt']\nPromptSession: pending/active event, version, lock"]
    META["interaction_chunk_metadata\nstarted/active/completed event ids"]
    PE["prompt_embeds"]
  end

  WS -->|session.interaction| SI
  SI -->|upon input, call enqueue| COORD
  COORD -->|route enqueue & apply| OH
  COORD -->|route enqueue & apply| AltPH
  COORD -->|route enqueue & apply| PH
  ENC -.->|injected at build| AltPH
  ENC -.->|injected at build| PH

  OH -.-o state
  AltPH -.-o state
  linkStyle 2,3,5,7,8 stroke-dasharray: 5 5,opacity:0.35
  linkStyle 6 stroke-dasharray: 5 5
  PH -->|read/write| PS
  PH -->|writes blended embeds| PE
  APPLY -->|writes ACK ids| META

  STEP --> APPLY
  APPLY --> PREP
  APPLY -->|apply_at_chunk_boundary| COORD
  STEP -->|reads prompt_embeds| PE
```

```mermaid
sequenceDiagram
  autonumber
  participant Serving as serving_video_output_stream
  participant Runner as DiffusionModelRunner
  participant Coord as InteractionCoordinator
  participant Handler as PromptInteractionHandler
  participant Pipeline
  participant State as StepRequestState

  Serving->>Runner: submit_interaction(request_id, interaction)
  Runner->>Coord: enqueue(modality="prompt")
  Coord->>Handler: enqueue(...)
  Handler->>Pipeline: encode_prompt(new prompt)
  Handler->>State: pending_event = QueuedPromptEvent (LWW)
  Note over Serving: reply session.interaction.queued

  rect rgb(235, 245, 255)
    Note over Runner,State: denoise step, not yet chunk boundary
    Runner->>Pipeline: step_scheduler(state)
    Pipeline->>State: read prompt_embeds
    Note over Runner: chunk_denoise_completed = false, skip post_decode
  end

  rect
    Note over Runner,Serving: denoise step at chunk boundary
    Runner->>Pipeline: step_scheduler(state)
    Runner->>Pipeline: post_decode(state)
    Runner->>State: attach started/active/completed ids to DiffusionOutput

    alt mid-chunk (request_denoise_completed = false)
      Runner->>Pipeline: apply_interaction_at_chunk_boundary(state)
      Pipeline->>Coord: apply_at_chunk_boundary(...)
      Coord->>Handler: apply_at_chunk_boundary(...)
      Handler->>State: start/advance lerp, bump version
      Pipeline->>State: set interaction_chunk_metadata
      Runner->>Pipeline: prepare_next_chunk(state) (a hook for pipeline---default to no-op)
      Runner-->>Serving: video chunk + event ACK ids
    else last chunk (request_denoise_completed = true)
      Note over Runner,Pipeline: skip apply/prepare, generation finished
      Runner-->>Serving: final video chunk + event ACK ids
    end
  end
```

### pytest UT

```bash
pytest \
  tests/diffusion/streaming_interactive_generation/test_prompt_interaction.py \
  tests/diffusion/streaming_interactive_generation/test_diffusion_streaming_output.py \
  tests/entrypoints/openai_api/test_serving_video_output_stream.py \
  -m 'cpu'
```

### pytest e2e

```
pytest -sv tests/e2e/accuracy/test_video_streaming_output_similarity.py
```

### pytest UT

42 passed

### pytest e2e

- passed
- also locally ran the same test on main, and compared the streaming-mode output between HEAD and main: `SSIM Y:1.000000 (inf) U:1.000000 (inf) V:1.000000 (inf) All:1.000000 (inf)`


fhfuih:
@vllm-omni-review-bot

linzhenpl07:
The prompt-only scoping and the SSIM 1.000000 comparison against main make this easy to read as a pure refactor — thanks for splitting it out. Two observations on the capability-declaration side, both about phase 2 rather than about anything broken here.

Disclosure of bias: I had almost exactly this question raised on my own PR (#6206), where a reviewer asked whether the abstraction I had written was valid. My answer there was that it was not, so I may be pattern-matching. Take these accordingly.

## 1. `build()` resolves two modalities two different ways

`coordinator.py` decides the prompt handler by asking the pipeline, and structured handlers by looking the model up in a central table, five lines apart:

```python
if supports_interaction_apply(pipeline):
    handlers["prompt"] = PromptInteractionHandler.from_pipeline(pipeline)

for modality, handler_cls in STRUCTURED_HANDLER_REGISTRY.get(model_class_name or "", {}).items():
    handlers[modality] = handler_cls()
```

`registry.py`'s own docstring says "A pipeline only declares the modalities that it supports", which is the first mechanism, not the second — under `STRUCTURED_HANDLER_REGISTRY` the declaration lives in the interaction layer keyed by `od_config.model_class_name`, and a model landing camera support has to also land an edit here.

The reason I bring it up rather than shrugging: on #6206 I had written the same shape (`_INDEPENDENT_VAE_GROUP_MODELS = frozenset({...})`) and it turned out the codebase already had one idiom for this — `io_support.py` resolves `support_image_input`, `support_audio_output`, `color_format` and friends as class attributes via `DiffusionModelRegistry._try_load_model_cls(...)` + `getattr`. I had a genuine reason not to use the object-based check there (process groups are created before the model is loaded, so only the class exists at that point). `build()` takes the pipeline instance, so that constraint does not apply here — the second mechanism could be the same as the first.

If the concern is layering — not wanting `models/` to import from `interaction/` — a class attribute naming the modality strings, with handler resolution staying in the interaction layer, would keep the dependency pointing the same way it does now.

Concretely, this is what makes #5527 need an edit in `interaction/registry.py` at the moment a camera model lands, rather than the model declaring it.

## 2. The explicit opt-in on the protocol is gone

Before:

```python
class SupportsPromptUpdate(Protocol):
    supports_prompt_update: ClassVar[bool] = True
```

After, `SupportsInteractionApply` has no flag, so `supports_interaction_apply` is a purely structural check for three method names. `InteractionMixin` supplies all three:

- `peek_chunk_media` — raises `NotImplementedError`, but the attribute exists
- `apply_interaction_at_chunk_boundary` — concrete
- `prepare_next_chunk` — documented no-op default

So any pipeline that inherits `InteractionMixin` passes `isinstance`, including one that never overrides `peek_chunk_media`. The failure mode is that the interaction is accepted at admission and then fails at the chunk boundary with `NotImplementedError`, rather than being rejected up front.

This is not a live bug — Helios is the only pipeline mixing it in, and it overrides what it needs. It is only that the guardrail that used to make the opt-in explicit is no longer there, right before phase 2 widens who participates.

There is a precedent for the ambiguity being uncomfortable: `supports_step_execution` is already structural-only (`isinstance(pipeline, SupportsStepExecution)`), which means `LingBotVideoPipeline`'s explicit `supports_step_execution: ClassVar[bool] = False` reads authoritative but is not actually consulted. It happens to agree with the structural answer today, so nothing is wrong — but a reader cannot tell that from the declaration.

Keeping an explicit `ClassVar` opt-in on `SupportsInteractionApply`, or having `peek_chunk_media` not be pre-supplied by the mixin, would restore the property that a pipeline has to say yes.

Neither of these blocks the prompt-only behavior this PR ships.

fhfuih:
@linzhenpl07 I can see your review from email but not on the website. Strange...

Anyway, your points do make sense, and the latest commits tries to resolve them:

1. Move prompt interaction to the registry. Now every modality reads from it
2. Removed peek_chunk_media from the mixin, so that now if a model doesn't

fhfuih:
@linzhenpl07 I can see your comments from email but not on the website. Your points do make sense and my latest commits try to resolve them

1. Move prompt update to the registry so that now every modality's interaction handler should come from the registry
2. Remove stub implementation of peek_chunk_media from the mixin, so that now a pipeline has to implement it itself to pass SupportInteractionApply check.

And note on why a registry exist: because there can be more than one handler for one modality, and a pipeline can choose the handler(s) that work best with it
- For prompt, although now we only have an interpolation-based transition, there are other potential alternatives like https://github.com/GordonChen19/Prompt-Relay that uses "concatenated prompt embedding + attention mask"
- For camera (as you can see in #5527), we are already aware of two types of interaction handlers

fhfuih:
Discussion with @hsliuustc0106 during the community meeting raise a question of whether the entire interaction module can be abstracted to upper layers: become diffusion/AR-agnostic, and unify with recent duplex call path for understanding models.

After thorough investigation, my conclusion is to **not** merge the two execution paths in this PR or at the current stage. A path-agnostic interaction *runtime* is not realistic yet. What is only shareable later is a thin envelope (event id, ack, opaque payload, orchestrator dispatch) for the control plane. The apply/clock/session semantics are different enough that a shared `Coordinator`/`Mixin` would be mostly adapters hiding two backends.

### Full-Duplex today is experimental, coupled with AR clock and ownership

Duplex today is another control plane (`open` / `append` / `signal` / `close`, fences, leases, epoch/turn). Apply is scheduler resume + KV continuity + model policy (`plan_append`, listen/speak, barge-in), not denoise chunk boundaries. JoyVL is even further away: a sidecar in front of plain `vllm serve`, not the Omni diffusion worker.

The entire duplex engine is still experimental today (`experimental/fullduplex`). The risk is high to plug in the entire diffusion-style control to it today. The current "interaction" system only extends today's stable diffusion call-path.

To elaborate, they have the following **different clock and ownership**:

| | This PR | Duplex |
|---|---|---|
| Lifetime | request (generation already running) | session (outlives one generate) |
| Mid-stream control is... | patch conditions at chunk boundary | append / turn / barge-in |
| Effect | condition (prompt embeds; later camera) | tokens / KV / speak-listen |
| Input and start | initial prompt kicks off rollout, all following input (interactions) are optional | must continue to repeatedly append input over time |
| Output pattern | continuous generation from initial state | incremental response to new input/state |
| state owner | StepRequestState on one request | session + KV/resumable scheduler state |
| Stability | aiming at stable diffusion path | experimental |

`apply_at_chunk_boundary(..., num_frames, fps)` is a diffusion timeline. Duplex `on_input` / `plan_append` is another. Moving the module up a directory does not create one path.

### What *could* be unified (may block this PR and to explore alternative development pathway)

In the experimental AR-Diffusion engine (`experimental/ar_diffusion`), its own session/tick contract actually support camera interaction from complex offline launching interface (manually launching `AsyncOmni` and use other experimental tooling on the user side). The call path can be unified with the current prompt update component.

I will continue to explore the engine-level opportunities to integrate these two; and that will land in another PR. This PR will be still diffusion-scoped, and it may stay here and get merged, or be supersedes once a more unified design is proposed

vllm_omni/diffusion/interaction/mixin.py
```diff
@@ -0,0 +1,68 @@
...
+            if chunk_index is None:
+                chunk_index = state.chunk_index
+            if num_frames is None:
+                num_frames = media.num_frames
+            if fps is None:
+                fps = media.fps
+
+        if self._interaction_coordinator is None:
+            raise RuntimeError(
+                "interaction coordinator is not initialized; "
```
SamitHuang, line 33: The runner calls this with no `chunk_index`/`num_frames`/`fps`, so this branch always runs `peek_chunk_media()` after every non-final streamed chunk. `ChunkMediaSpec` then rejects `fps <= 0`, while `PromptInteractionHandler.apply_at_chunk_boundary` immediately `del`s those args. This couples a prompt-only apply path to a camera-timeline contract that is not required yet and is not exercised by current tests (they pass `fps=0` explicitly and skip peek). Either pass a spec that cannot fail for prompt-only, or do not peek until a handler actually needs media time.

vllm_omni/diffusion/models/helios/pipeline_helios.py
```diff
@@ -533,10 +534,17 @@ def prepare_encode(
                 "zero_steps": int(extra.get("zero_steps", 1)),
             }
```
SamitHuang, line 536: `OmniDiffusionSamplingParams.fps` is `int | None = None`, and Helios extra_body does not inject a default fps. `float(state.sampling.fps or 0.0)` becomes `0.0`, then `ChunkMediaSpec` raises `fps must be > 0`. Because the runner always peeks here, streaming Helios can fail at the first chunk boundary with no prompt update submitted. Use a real Helios default fps, or stop constructing `ChunkMediaSpec` until a modality needs the timeline.

vllm_omni/diffusion/interaction/registry.py
```diff
@@ -0,0 +1,21 @@
...
+
+from __future__ import annotations
+
+from vllm_omni.diffusion.interaction.modality_handlers.base import InteractionHandler
+from vllm_omni.diffusion.interaction.modality_handlers.prompt import PromptInteractionHandler
+
+STRUCTURED_HANDLER_REGISTRY: dict[str, dict[str, type[InteractionHandler]]] = {
+    # Pipeline class name -> modality -> handler class.
+    "HeliosPipeline": {
+        "prompt": PromptInteractionHandler,
```
SamitHuang, line 19: `InteractionCoordinator.build` looks up `od_config.model_class_name` exactly. `HeliosPyramidPipeline` maps to the same `HeliosPipeline` class in `vllm_omni/diffusion/registry.py` (distilled/mid), but it is missing here, so `has_modality("prompt")` is false and `submit_interaction` raises. The old mixin applied on the instance and did not depend on this alias. Register the pyramid name (or key by pipeline type / `type(pipeline).__name__`) and add a UT with `model_class_name="HeliosPyramidPipeline"`.

vllm_omni/diffusion/worker/diffusion_model_runner.py
```diff
@@ -1094,10 +1107,15 @@ def _execute_stepwise_core(
                                 clear_pipeline_stage_durations(self.pipeline)
                                 result = self.pipeline.post_decode(req)
                                 if result is not None:
-                                    self._attach_stepwise_metrics(
+                                    self._attach_stepwise_metadata(
                                         req,
                                         result,
                                     )
+                                    # After consuming this chunk's interaction metadata, apply pending interactions and
```
SamitHuang, line 1296: Apply and `prepare_next_chunk` moved here from `HeliosPipeline.post_decode`, and ACK ids now land via `_attach_stepwise_metadata` on the *previous* boundary's `interaction_chunk_metadata`. There is no test that drives this runner loop: existing UTs call `PromptInteractionHandler` / `apply_interaction_at_chunk_boundary(..., fps=0)` and never check `DiffusionOutput.started_event_ids` after `post_decode`. Add a stepwise runner test that submits an interaction, completes a chunk through this path with `sampling.fps is None`, and asserts embeds + ACK ids. Without that, the peek/`ChunkMediaSpec` crash and ACK wiring are untested.

SamitHuang (changes requested):
Phase-1 coordinator split is scoped, but it is not behavior-preserving on the production Helios streaming path.

1. `DiffusionModelRunner` now calls `apply_interaction_at_chunk_boundary(req)` with no media args after every non-final chunk. The mixin then requires `peek_chunk_media()` to return a `ChunkMediaSpec` with `fps > 0`. Helios builds that spec as `float(state.sampling.fps or 0.0)`, and `OmniDiffusionSamplingParams.fps` defaults to `None`. Prompt lerp does not use fps, so this can raise on the first chunk boundary even when no interaction was submitted.
2. Handler registration is keyed by exact `od_config.model_class_name`. Only `HeliosPipeline` is registered; `HeliosPyramidPipeline` is the same class in `vllm_omni/diffusion/registry.py` (distilled/mid). Previously `PromptUpdateMixin` applied to the instance regardless of the alias, so pyramid serving would now hit `prompt_update is not supported`.
3. Apply + event ACK moved out of `HeliosPipeline.post_decode` into the runner, but UTs still call the handler/mixin with explicit `fps=0` and never assert `DiffusionOutput` ACK ids through `_execute_stepwise_core`. That is why (1) is invisible in the reported 42 passing tests.

Please fix the peek/fps contract, register (or type-key) both Helios names, and add a runner-path regression that calls apply the same way production does.

vllm_omni/diffusion/worker/diffusion_model_runner.py
```diff
@@ -1173,42 +1191,49 @@ def submit_interaction(
...
 
-    def _submit_prompt_update_interaction(
-        self,
-        request_id: str,
-        interaction: OmniInteractionPrompt,
-    ) -> None:
-        """Queue a prompt-update interaction for an active stepwise request."""
-        if not supports_prompt_update(self.pipeline):
+        event = interaction.get("event")
+        multi_modal_data = event.get("multi_modal_data") if isinstance(event, dict) else None
```
wtomin: This should reject structural payloads by presence of the `multi_modal_data` key, not only by a non-empty dict value. With `{"event": {"prompt": "updated", "multi_modal_data": {}}}`, `has_mm` is false and the request is accepted as prompt-only, whereas the old path rejected any interaction carrying `multi_modal_data` and this PR says MM payloads remain out of scope. Reject the key here (or strip to a prompt-only event in serving) and add empty/non-dict `multi_modal_data` cases to the runner test.

wtomin:
@fhfuih  Do you have an API design document? And if new video-streaming model other than helios needs to support this framework, how to adapt it to new models/pipelines?

vllm_omni/diffusion/interaction/modality_handlers/prompt.py
```diff
@@ -0,0 +1,271 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+"""Prompt-track interaction handler (midway prompt updates)."""
+
+from __future__ import annotations
+
+import logging
+from collections.abc import Callable, Sequence
+from dataclasses import dataclass
+from typing import Any, ClassVar, override
```
david6666666: typing.override is only available in Python 3.12, but this project still supports 3.10/3.11. Importing this module fails on those versions before Helios can load. Please use typing_extensions.override here and in pipeline_helios.py.

fhfuih:
> @fhfuih Do you have an API design document? 

API design is in the RFC: #5120

> And if new video-streaming model other than helios needs to support this framework, how to adapt it to new models/pipelines?

1. implement stepwise execution for streaming generation (this allows RPC messages of interaction inputs to be digested at step/chunk boundary)
2. Mix `InteractionMixin`
3. Implement `peek_chunk_media(StepRequestState)` so the framework know the number of frames in the yet-to-generate chunk at runtime
4. If needed to prepare per-chunk generation routine internally, implement a `prepare_next_chunk(StepRequestState)` hook, which will be invoked at chunk boundary
5. Register supported modalities of interactions, and that modality's preferred interaction handlers

vllm_omni/diffusion/interaction/modality_handlers/prompt.py
fhfuih: Oh but our doc (user guide) says we only supports 3.12 🤔 After your hint, I do see that existing `@overrides` are all imported from typing_extensions. But I think that user guide doc can be separate issue

vllm_omni/diffusion/interaction/mixin.py
fhfuih: Added fps check in peek_media, and only call peek media if at least one of the interaction handlers needs it.

In the future, camera interaction will need it, causing future models that supports camera interaction to peek.

Prompt update itself doesn't need it. So models like Helios never triggers ChunkMediaSpec peeking.

Consequently, make `apply_at_chunk_boundary`'s params optional. chunk_index and num_frames and stuff may or may not be passed to concrete handlers (depending on whether the coordinator decides to peek). And a handler shall validate their presence if the handler needs this info

vllm_omni/diffusion/models/helios/pipeline_helios.py
fhfuih: Fixed and explained in the mixin's comment

vllm_omni/diffusion/worker/diffusion_model_runner.py
fhfuih: fixed

vllm_omni/diffusion/interaction/registry.py
fhfuih: fixed

vllm_omni/diffusion/worker/diffusion_model_runner.py
fhfuih: added a UT

vllm_omni/diffusion/interaction/modality_handlers/prompt.py
```diff
@@ -0,0 +1,273 @@
...
+        fps: float | None = None,
+    ) -> InteractionChunkMetadata | None:
+        """Advance or start prompt interpolation before the next chunk."""
+        # Prompt lerp is chunk-LWW; media timeline and caller chunk_index unused.
+        del chunk_index, num_frames, fps
+        session = _get_prompt_session(state)
+        # Prompt ignores frame scheduling but still records the boundary clock.
+        session.last_boundary_at = boundary_at
+
+        with session.lock:
```
RuixiangMa: pending_event is correctly protected by the lock. But active_event is both read and written without the lock. While apply_at_chunk_boundary is intended to be single-threaded, this is fragile — any future concurrent access would be a race. The old code had the same pattern, but the new framework's broader scope makes it more likely to be hit.

vllm_omni/inputs/data.py
```diff
@@ -43,11 +43,13 @@ class OmniInteractionEvent(OmniTextPrompt, total=False):
     pass
 
 
-class OmniInteractionPrompt(TypedDict, total=False):
-    """Mid-stream interaction payload."""
+class OmniInteractionPrompt(TypedDict):
```
RuixiangMa, line 49: lient code omitting event_id or event will fail type checking. The comment says "API layer always guarantees its presence," but this is an undocumented breaking contract change.

vllm_omni/diffusion/interaction/modality_handlers/prompt.py
```diff
@@ -0,0 +1,273 @@
...
+def prompt_update_versions(states: Sequence[StepRequestState]) -> tuple[int, ...]:
+    """Return per-request prompt-update versions for batch cache comparison."""
+    versions: list[int] = []
+    for state in states:
+        session = state.interaction_sessions.get("prompt")
+        versions.append(session.version if isinstance(session, PromptSession) else 0)
+    return tuple(versions)
+
+
+def _get_prompt_session(state: StepRequestState) -> PromptSession:
```
RuixiangMa: The old code used setdefault (GIL-atomic). Two threads (enqueue from the interaction thread + apply_at_chunk_boundary from the generation thread) could both create separate PromptSession instances with separate Lock objects, and one overwrites the other's pending_event.

vllm_omni/inputs/data.py
fhfuih: Now documented

vllm_omni/diffusion/interaction/modality_handlers/prompt.py
fhfuih: Resolve this comment by moving `active_event` and all other stuff related to session into `with session.lock`

vllm_omni/diffusion/interaction/modality_handlers/prompt.py
fhfuih: I can verify that there are two places to try-get-or-set state.interaction_sessions. Now consolidate the call path and remove defensive code:

- This function is removed and inlined
- In `enqueue` (when the first interaction input enters), get-or-set corresponding modality's interaction_session
- In `apply` (always called in streaming output mode's chunk boundary, without knowing if there are queued interaction inputs or not), only try to get. Since the session data is lazy-initialized, the coordinator will skip those not-yet-created sessions (not present == this modality's enqueue is not called yet). Then when coordinator actually fans out to a concrete interaction handler's `apply` function, `handler.apply` will assert that the session data is valid

fhfuih:
Commits about resolved RuixiangMa' comment. Also, the conversation/clarification with wtomin is added to Adding a Diffusion Model doc

> > And if new video-streaming model other than helios needs to support this framework, how to adapt it to new models/pipelines?
> 
> 1. implement stepwise execution for streaming generation (this allows RPC messages of interaction inputs to be digested at step/chunk boundary)
> 2. Mix `InteractionMixin`
> 3. Implement `peek_chunk_media(StepRequestState)` so the framework know the number of frames in the yet-to-generate chunk at runtime
> 4. If needed to prepare per-chunk generation routine internally, implement a `prepare_next_chunk(StepRequestState)` hook, which will be invoked at chunk boundary
> 5. Register supported modalities of interactions, and that modality's preferred interaction handlers

fhfuih:
Commits above moves around the doc page for video streaming generation API to adapt to the new docs hierarchy. Due to the recent doc refactors, this API doc page is not eventually picked up on the website.

With the commits above, now the API is at `Serving -> OpenAI-Compatible API -> Video Streaming Generation API` (next to other API pages). Also, the recently-added `Features -> Runtime and Stage Execution -> Execution Modes and Streaming -> # Send Requests` is able to link to this page.

SamitHuang:
Pls resolve the conflicts

fhfuih:
> Pls resolve the conflicts

Done. Please take a look. Thanks

SamitHuang (approved):
LGTM

Issue #6305, opened 2026-08-18
[RFC]: Diffusion Operator Boundary


Bounty-hunter:
### Motivation.

Upstream vLLM has settled, in [#42770](https://github.com/vllm-project/vllm/issues/42770), on a new direction: **stop evolving `CustomOp` / vLLM IR (keep them for legacy models), and move to "hand-written fused ops + per-hardware code separation"** for new models. The diffusion side has **no unified operator boundary**, this RFC proposes **per-hardware code separation** and three-level operator boundary that gives each case its own home: 
(1) **base ops (legacy CustomOp/IR)**
(2) **cross-model fused ops (`common/ops/`, pure functions)**
(3) **a model's own large fusions (`<model>/<vendor>/ops`)**

### Proposed Change.

---

## 1. Background: the upstream direction has changed

vLLM RFC #42770 (WoosukKwon, 2026-05) announced three changes to model development. The two that matter here:

1. **Remove the full-graph `torch.compile` requirement** — fusion moves from "done by the compiler" to "written by hand in model code."
2. **Separate model code by hardware vendor** — different hardware favors different fusion boundaries, requiring hardware-specific model definitions.
> Each model definition can then evolve independently for its target hardware without risk of breaking the others. Just as importantly, each vendor can design fusions around the kernels available to them, rather than conforming to a shared abstraction that fits no one perfectly.

Upstream evidence (`vllm/models/deepseek_v4/`): `nvidia/model.py` (1548 lines) and `amd/model.py` (1018 lines) are **two complete model definitions with different fusion boundaries**:

| Path | Fused unit |
| --- | --- |
| NVIDIA / tilelang | `post + pre + norm` in one kernel |
| AMD / tilelang | `post + pre` in one kernel |
| AMD / AITER | `pre` and `post` as two kernels |

```python
# amd/model.py:407 — give up the post+pre fusion when AITER's smaller-boundary
# kernels are available and the shape constraint holds
self.use_fused_mhc = HAS_TILELANG_MHC and not (
    HAS_AITER_MHC and self.hidden_size % 256 == 0
)
```
A shape predicate (`hidden_size % 256 == 0`) choosing between fusion boundaries is not something a shared abstraction can hide.

We take the *conclusion* from this — fusion boundaries are hardware-specific and must be expressible in code — without taking the *cost*: we are not proposing per-vendor copies of whole models. See §4.4.

---

### Case 1: fused ops have no home — hand-written twice

[...]

PR #6306, opened 2026-08-18
[Perf][Hunyuan] Optimize for vae and patch_embed

Bounty-hunter:

## Purpose
Optimize for `vae` and `patch_embed`

## Propose
(1) Add Fused GroupNorm+SiLU / AdaGN+SiLU Triton kernels 
(2) follow https://github.com/vllm-project/vllm-omni/issues/6305 to manager fused kernel and model isolation by hardware for high performance(enable cudnn.benchmark for conv3d)

### accuracy
```bash
# fused kernel unit tests (1 GPU)
pytest -q tests/model_executor/models/common/test_fused_group_norm_silu.py \
          tests/model_executor/models/common/test_fused_adaptive_group_norm_silu.py

# end-to-end pixel accuracy + it2i (4 GPUs)
pytest -v -s tests/e2e/accuracy/test_hunyuan_image3_pixel_accuracy.py
pytest -v -s tests/e2e/accuracy/test_hunyuan_image3.py
```

### performance benchmark
```bash
pytest -s -v tests/dfx/perf/scripts/run_diffusion_benchmark.py \
  --test-config-file tests/dfx/perf/tests/test_hunyuan_image_tp4.json

pytest -s -v tests/dfx/perf/scripts/run_diffusion_benchmark.py \
  --test-config-file tests/dfx/perf/tests/test_hunyuan_image_tp2sp2.json

pytest -s -v tests/dfx/perf/scripts/run_diffusion_benchmark.py \
  --test-config-file tests/dfx/perf/tests/test_hunyuan_image_tp2cfgp2.json

pytest -s -v tests/dfx/perf/scripts/run_diffusion_benchmark.py \
  --test-config-file tests/dfx/perf/tests/test_hunyuan_image_it2i.json
```

**vLLM Version:**

**vLLM-Omni Commit:**

### accuracy
| suite | result |
| --- | --- |
| test_fused_group_norm_silu.py + test_fused_adaptive_group_norm_silu.py | 365 passed |
| test_hunyuan_image3_pixel_accuracy.py | 2 passed |
| test_hunyuan_image3.py (it2i) | 1 passed, 3 skipped |

### performance
| bench | before qps | after qps | qps delta |
| --- | --- | --- | --- |
| dit tp4 | 0.4205 | 0.4398 | **+4.60%** |
| dit tp2+sp2 | 0.3982 | 0.4155 | **+4.32%** |
| dit tp2+cfgp2 | 0.4483 | 0.4596 | **+2.52%** |
| it2i (AR TP2 + DiT TP2) | 0.3983 | 0.4140 | **+3.95%** |

per-stage 耗时（ms，before -> after）：

| bench | patch_embed | vae.decode |
| --- | --- | --- |
| dit tp4 | 140.2 -> 83.4 (−40.5%) | 305.1 -> 246.8 (−19.1%) |
| dit tp2+sp2 | 146.9 -> 82.5 (−43.9%) | 305.9 -> 245.6 (−19.7%) |
| dit tp2+cfgp2 | 34.4 -> 31.3 (−9.0%) | 305.9 -> 245.7 (−19.7%) |


vllm_omni/diffusion/models/hunyuan_image3/nvidia/transformer_blocks.py
```diff
@@ -0,0 +1,161 @@
...
+    """
+    Make a standard normalization layer.
+    :param channels: number of input channels.
+    :return: a nn.Module for normalization.
+    """
+    return nn.GroupNorm(32, channels, **kwargs)
+
+
+def zero_module(module):
+    """
```
ZJY0516: These should lie in common folders

vllm_omni/model_executor/models/common/ops/fused_adaptive_group_norm_silu.py
```diff
@@ -0,0 +1,193 @@
...
+"""
+
+import torch
+import torch.nn.functional as F
+from vllm.triton_utils import HAS_TRITON, tl, triton
+
+
+if HAS_TRITON:
+    @triton.jit
+    def _adaptive_group_norm_silu_kernel(
```
Isotr0py:
```suggestion
@triton.jit
def _adaptive_group_norm_silu_kernel(
```
I remember that `triton_utils` has guarded triton import, which means that we don't need the `if` statement here.

vllm_omni/diffusion/models/hunyuan_image3/hunyuan_image3_transformer.py
```diff
@@ -79,6 +79,11 @@
 from vllm_omni.diffusion.layers.fused_moe import FusedMoE
 from vllm_omni.diffusion.layers.norm import RMSNorm
 from vllm_omni.diffusion.layers.rope import RotaryEmbedding
+
+# ResBlock is platform-dispatched in the package __init__: CUDA gets fused
+# GroupNorm+SiLU and AdaGN kernels, every other backend gets the plain PyTorch
+# block. UNetDown and UNetUp below instantiate whichever one this resolves to.
+from vllm_omni.diffusion.models.hunyuan_image3 import ResBlock
```
Isotr0py: Hmmm, I feel this can easily case circular if import from `vllm_omni.diffusion.models.hunyuan_image3` directly. 

I don't think we would introduce hardware ops for each components, should we add an extra folder like `layers`  or something else there?
```python3
from vllm_omni.diffusion.models.hunyuan_image3.layers import ResBlock
```
Then we can place triton/native layer into `hunyuan_image3/layers/nvidia/autoencoder_blocks.py` and `hunyuan_image3/layers/native/autoencoder_blocks.py` etc.

Bounty-hunter: Agreed — The platform-specific blocks under a `huanyuan_iamge3/layers/` with the dispatch done in `layers/__init__.py` is the cleaner layout.

Isotr0py (approved):

vllm_omni/model_executor/models/common/ops/fused_group_norm_silu.py
```diff
@@ -0,0 +1,187 @@
...
+
+            x_val = tl.load(x_ptr + base + offsets, mask=mask, other=0.0)
+            x_val = x_val.to(tl.float32)
+
+            mean_acc += tl.sum(x_val, axis=0)
+            var_acc += tl.sum(x_val * x_val, axis=0)
+
+    group_total = group_size * spatial_size
+    mean = mean_acc / group_total
+    var = var_acc / group_total - mean * mean
```
Gaohan123: [P2] Use a numerically stable variance calculation

Computing the variance as `E[x^2] - E[x]^2` is subject to catastrophic cancellation even with fp32 accumulation. For example, for a valid input around `10000 ± 0.1`, this expression can round the variance to zero (or even a negative value depending on the reduction order), while the actual variance is about `0.01`. This can produce incorrectly normalized output or NaNs in `sqrt(var + eps)`, whereas PyTorch GroupNorm remains stable.

Please use Welford accumulation or another stable centered/two-pass variance algorithm. The same calculation is also present in `fused_adaptive_group_norm_silu.py`, so both kernels should be fixed. It would also be useful to add tests with a large constant offset plus small variance for fp32, fp16, and bf16 inputs.

vllm_omni/diffusion/models/hunyuan_image3/autoencoder.py
```diff
@@ -70,46 +69,6 @@ def custom_forward(*inputs):
...
-    """
-    Perform Conv3d on patches with numerical differences from nn.Conv3d within 1e-5.
-    Only symmetric padding is supported.
-    """
-
-    def forward(self, input):
-        B, C, T, H, W = input.shape
-        memory_count = (C * T * H * W) * 2 / 1024**3
-        if memory_count > 2:
-            n_split = math.ceil(memory_count / 2)
```
Gaohan123, line 83: [P2] Preserve the memory fallback when VAE tiling is disabled

This PR replaces the memory-aware `Conv3d` wrapper with plain `nn.Conv3d` throughout the VAE. The removed wrapper split large activations once its estimate exceeded approximately 2 GiB, but VAE tiling is still disabled by default (`use_spatial_tiling = False` and `use_temporal_tiling = False` around lines 373-374, and `vae_use_tiling` also defaults to false at the API/config level).

Consequently, high-resolution requests that previously relied on the Conv3d chunking path can now OOM when tiling is disabled or when the VAE is used directly. Please retain a memory-aware Conv3d fallback, or ensure tiling is enabled for every relevant entry point before removing it. A regression test covering a large input with `vae_use_tiling=False` would help protect this behavior. This concern applies to the other `nn.Conv3d` replacements in this file and in the new native/NVIDIA ResnetBlock implementations as well.

vllm_omni/diffusion/models/hunyuan_image3/layers/nvidia/_cudnn.py
```diff
@@ -0,0 +1,27 @@
...
+
+    Used to scope ``torch.backends.cudnn.benchmark``/``deterministic`` to the
+    few blocks whose convolutions actually benefit from autotune, instead of
+    setting them globally at import time (which would silently change every
+    other model's conv behaviour in the same process).
+    """
+    old_benchmark = torch.backends.cudnn.benchmark
+    old_deterministic = torch.backends.cudnn.deterministic
+
+    torch.backends.cudnn.benchmark = benchmark
```
Gaohan123, line 21: [P2] Avoid changing process-global cuDNN flags inside forward paths

`torch.backends.cudnn.benchmark` and `torch.backends.cudnn.deterministic` are process-global settings. Saving and restoring them in a context manager does not make the operation thread-safe. If two forwards overlap, one context may restore the flags while the other is still executing, and the final restore may leave the process in a state different from the original one.

This also weakens the model-isolation goal of the hardware-specific split. Please configure these settings once at an appropriate worker/process boundary, or use an algorithm-selection mechanism that does not mutate process-global state from each block's `forward`.

vllm_omni/diffusion/models/hunyuan_image3/layers/__init__.py
```diff
@@ -0,0 +1,22 @@
+# SPDX-License-Identifier: Apache-2.0
```
hsliuustc0106, line 1: try to move this to model_executor/models please and check whether we can reuse diffusion engine for stage 1

vllm_omni/model_executor/models/common/ops/fused_group_norm_silu.py
Bounty-hunter: fixed, change to welford parallel.

vllm_omni/diffusion/models/hunyuan_image3/autoencoder.py
Bounty-hunter: We removed it deliberately for two reasons: (1) the chunking `Conv3d` was a hack — a hardcoded ~2 GiB threshold that ignores the actual available memory and pays chunking overhead even when it fits; (2) we want high-res VAE memory handled by the unified framework-side VAE tiling, which enables tiling parallelism and related features the per-conv split can't.

vllm_omni/diffusion/models/hunyuan_image3/layers/nvidia/_cudnn.py
Bounty-hunter: The concern is valid for concurrent forwards. In this pipeline the diffusion stage runs sequentially (single stream, one forward at a time per worker), so the save/restore never overlap and the flags stay consistent.

vllm_omni/diffusion/models/hunyuan_image3/layers/__init__.py
Bounty-hunter: AR and DiT share the same structure, so some accelerations can be reused across both. The some submodule, however, is still implemented differently in places to fit the diffusion engine, so keeping `layers/` under the DiT model for now is the pragmatic call — it also gives us a natural home for future work like attention-module optimizations. 
In the future, key models are move into `model_executor/models`; consolidation will unify this into a single `layers/` directory covering both the shared submodule and the AR/DiT-specific accelerate submodule.

PR #5531, opened 2026-07-29
[REFACT]Refactor diffusion parallel state

bjf-frz:
PLEASE FILL IN THE PR DESCRIPTION HERE.

## Purpose

The diffusion parallel-state implementation had accumulated several parallel-group abstractions that no longer matched the actual execution model:

- `_DIT` duplicated the global diffusion rank topology without representing an independent parallel dimension.
- `_FS` was initialized as a communication group but was not used by the FSDP2/HSDP execution path.
- HSDP group ownership was split between custom process groups and FSDP2, making the source of truth unclear.
- Parallel initialization could leave partially initialized process groups after a failure.
- `data_parallel_size`, `WORLD_SIZE`, and intermediate layout sizes had overlapping and ambiguous responsibilities.
- Several unsupported HSDP combinations were accepted until they failed later during model loading or execution.

This PR refactors the diffusion parallel-state implementation so that each abstraction corresponds to an actual execution or communication responsibility.

The refactor contains six main changes:

1. **Remove the redundant `_DIT` group**
 - Diffusion workers already execute within `WORLD`.
 - VAE patch parallelism now selects ranks directly from `WORLD`.
 - No separate DiT communication group or DiT-specific global size is maintained.

2. **Remove the unused `_FS` group**
 - HSDP/FSDP2 process groups are now owned exclusively by the FSDP2 `DeviceMesh`.
 - The custom parallel state no longer creates an unused fully-sharded communication group.

3. **Make parallel initialization atomic**
 - Parallel groups are either initialized completely or cleaned up on failure.
 - Partially initialized diffusion and vLLM process-group state is no longer retained after an exception.

4. **Simplify WORLD and DP resolution**
 - `WORLD_SIZE` is treated as the actual runtime process count.
 - When HSDP is disabled, an omitted DP size is inferred as:
   `DP = WORLD_SIZE / (TP × SP × PP × CFG)`.
 - An explicitly configured DP size must match the inferred value.

5. **Separate HSDP from the regular orthogonal rank layout**
 - HSDP is no longer represented as an axis inside `RankGenerator`.
 - `RankGenerator` only handles the real orthogonal diffusion dimensions: TP, SP, PP, CFG, and DP.
 - HSDP overlays the same WORLD ranks used by supported SP and CFG layouts.

6. **Add explicit compatibility validation**
 - HSDP is supported with SP, CFG, and VAE Patch Parallel.
 - HSDP combinations with TP, ordinary DP, PP, and EP are rejected early.
[...]

PR #6307, opened 2026-08-18
[Core] Make diffusion worker titles topology-aware

### LINKED ISSUE #6297: [Feature]: Make diffusion worker process titles and log prefixes topology-aware

Diffusion worker process titles currently expose only the global worker rank:

```text
vLLM-Omni::DiffusionWorker-0
vLLM-Omni::DiffusionWorker-1
vLLM-Omni::DiffusionWorker-2
vLLM-Omni::DiffusionWorker-3
```

This makes workers visible in `ps`, `htop`, and tools such as `nvidia-smi`, but it does not identify which parallel groups each worker belongs to. The same four process names could represent DP=4, TP=4, SP=4, TP=2×CFG=2, or another topology.

The current title is assigned before the distributed groups are initialized:

https://github.com/vllm-project/vllm-omni/blob/main/vllm_omni/diffusion/worker/diffusion_worker.py

Diffusion model-parallel groups are later initialized with rank order `tp-sp-pp-cfg-dp`:

https://github.com/vllm-project/vllm-omni/blob/main/vllm_omni/diffusion/distributed/parallel_state.py

#### Upstream precedent

After model-parallel initialization, upstream vLLM reads the initialized process groups and updates both the process title and log prefix. For example:

```text
VLLM::Worker_DP0_PP0_TP1_EP1
```

Implementation:

https://github.com/vllm-project/vllm/blob/main/vllm/v1/executor/multiproc_executor.py

Original upstream change:

https://github.com/vllm-project/vllm/pull/22205

vLLM-Omni PR #3602 introduced the current diffusion process title. Its follow-up discussion already noted that multi-DP diffusion deployments should adopt the upstream naming approach:

https://github.com/vllm-project/vllm-omni/pull/3602#issuecomment-4466904703

#### Proposed change

Refactor diffusion process-title construction into a shared helper that runs after `initialize_model_parallel()` and before model loading.

The helper should preserve the recognizable prefix and append each active parallel rank:

```text
vLLM-Omni::DiffusionWorker_TP1
vLLM-Omni::DiffusionWorker_DP0_CFG1_TP1
vLLM-Omni::DiffusionWorker_DP1_PP0_SP1_CFG0_TP0
```

As in upstream vLLM, dimensions whose world size is 1 should be omitted.

At minimum, include:

- data parallelism (`DP`)
- pipeline parallelism (`PP`)
- sequence parallelism (`SP`)
- classifier-free-guidance parallelism (`CFG`)
- tensor parallelism (`TP`)

Expert-parallel and HSDP/fully-sharded configurations should also remain uniquely identifiable when enabled.

Ranks should be read from the initialized process groups rather than recomputed from the global rank. The same topology identifier should be used for the worker log prefix where practical.

Use the upstream `vllm.utils.system_utils.set_process_title` helper with the `vLLM-Omni` prefix instead of importing `setproctitle` directly.

#### Cross-platform behavior

Process-title construction should be accelerator-neutral. CUDA, ROCm, Intel XPU, MUSA, and Ascend NPU diffusion workers share the common worker lifecycle, so the implementation must not depend on a vendor-specific API.

The title should be observable through `ps`/`htop` on supported Linux deployments. Accelerator-specific tools such as `nvidia-smi` may also display it where their process views expose OS titles.

#### Acceptance criteria

- Titles are updated after distributed/model-parallel initialization and before model loading.
- Every active DP/PP/SP/CFG/TP dimension appears with its local group rank.
- Dimensions with world size 1 are omitted.
- Supported EP/HSDP layouts still produce unambiguous worker identifiers.
- The `vLLM-Omni::DiffusionWorker` prefix remains stable for existing reliability and diagnostic tooling.
- Worker log prefixes use the same topology identifier.
- Process-title handling is accelerator-neutral.
- Missing process-title support remains a non-fatal no-op.
- Titles remain reasonably readable in standard process-monitoring output.
- There is no scheduling, communication, output, or performance behavior change.

#### Testing

Add CPU-friendly unit coverage for the title-building helper with mocked group sizes and ranks:

- all parallel sizes equal to 1
- TP-only
- DP-only
- TP × CFG
- DP × SP × TP
- applicable EP/HSDP configurations
- unavailable process-title support

Add a Linux subprocess test that checks the resulting title through `ps`. Optional platform smoke tests may verify the corresponding accelerator monitoring tool when it exposes process names.

### Alternatives

1. Keep the global-rank-only title and manually derive topology from the deploy configuration. This is cumbersome during live debugging and incident response.
2. Log the topology once during startup. This helps postmortem analysis but does not identify live processes in `ps`, `htop`, or accelerator monitoring tools.
3. Recompute all dimensions from the global rank. This duplicates topology logic and may drift from the actual initialized process groups.

### Additional context

The standard diffusion topology is:

```text
world_size = DP × CFG × SP × PP × TP
rank order = TP-SP-PP-CFG-DP
```

The existing global-rank suffix identifies the worker but not its role within that topology.

A duplicate search found no issue specifically tracking topology-aware diffusion worker names. Open PR #5441 refactors stage process/client structure but does not implement this process-title change.

- [ ] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://vllm-omni.readthedocs.io), which can answer lots of frequently asked questions.

prettygirlisnotme:
## Summary

- Replace global-rank-only diffusion worker titles with topology-aware names derived from initialized parallel groups.
- Include active DP, PP, SP, CFG, TP, FS (HSDP), and EP ranks while omitting singleton dimensions.
- Reuse the same topology identifier for the OS process title and log prefix through the upstream vLLM helpers.
- Keep the early generic title, then refresh it after model-parallel initialization and before model loading.

Closes #6297

## User impact

Diffusion workers can be identified unambiguously in ps, htop, nvidia-smi, and logs across mixed parallel topologies. Missing optional process-title support remains non-fatal.

## Validation

- CPU unit and Linux ps integration tests: 12 passed, 15 warnings
- Ruff check: passed
- Ruff format check: passed
- Python compilation check: passed
- Git diff check: passed

Test environment: Python 3.12, vLLM 0.27.0+cpu, PyTorch 2.13.0+cpu, pytest 9.1.1, setproctitle 1.3.7. The local host exposes an NVIDIA GPU while using a CPU-only vLLM wheel, so Omni GPU auto-detection was disabled only in the pytest bootstrap.

### DIFF
--- a/tests/diffusion/test_diffusion_worker_process_title.py
+++ b/tests/diffusion/test_diffusion_worker_process_title.py
@@ -1,0 +1,218 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+
+import importlib.util
+import shutil
+import subprocess
+import sys
+import textwrap
+import time
+from dataclasses import dataclass
+from unittest.mock import Mock
+
+import pytest
+from pytest_mock import MockerFixture
+
+import vllm_omni.diffusion.worker.diffusion_worker as worker_module
+
+pytestmark = [pytest.mark.core_model, pytest.mark.diffusion, pytest.mark.cpu]
+
+
+@dataclass(frozen=True)
+class _FakeGroup:
+    world_size: int = 1
+    rank_in_group: int = 0
+
+
+_GROUP_GETTERS = (
+    "get_dp_group",
+    "get_pp_group",
+    "get_sp_group",
+    "get_cfg_group",
+    "get_tp_group",
+    "get_fs_group",
+    "get_ep_group",
+)
+
+
+def _patch_groups(
+    mocker: MockerFixture,
+    groups: dict[str, _FakeGroup],
+) -> dict[str, Mock]:
+    patched_getters = {}
+    for getter_name in _GROUP_GETTERS:
+        patched_getters[getter_name] = mocker.patch.object(
+            worker_module,
+            getter_name,
+            return_value=groups.get(getter_name, _FakeGroup()),
+        )
+    return patched_getters
+
+
+def test_setup_uses_default_name_before_model_parallel_init(
+    mocker: MockerFixture,
+) -> None:
+    mocker.patch.object(
+        worker_module,
+        "model_parallel_is_initialized",
+        return_value=False,
+    )
+    group_getters = _patch_groups(mocker, {})
+    set_process_title = mocker.patch.object(worker_module, "set_process_title")
+    decorate_logs = mocker.patch.object(worker_module, "decorate_logs")
+
+    worker_module._setup_diffusion_worker_proc_title_and_log_prefix(enable_ep=True)
+
+    set_process_title.assert_called_once_with(
+        name="DiffusionWorker",
+        prefix="vLLM-Omni",
+    )
+    decorate_logs.assert_called_once_with("DiffusionWorker")
+    for group_getter in group_getters.values():
+        group_getter.assert_not_called()
+
+
+@pytest.mark.parametrize(
+    ("groups", "enable_ep", "expected_name"),
+    [
+        ({}, False, "DiffusionWorker"),
+        ({"get_tp_group": _FakeGroup(2, 1)}, False, "DiffusionWorker_TP1"),
+        ({"get_dp_group": _FakeGroup(4, 2)}, False, "DiffusionWorker_DP2"),
+        ({"get_pp_group": _FakeGroup(2, 1)}, False, "DiffusionWorker_PP1"),
+        (
+            {
+                "get_cfg_group": _FakeGroup(2, 1),
+                "get_tp_group": _FakeGroup(2, 1),
+            },
+            False,
+            "DiffusionWorker_CFG1_TP1",
+        ),
+        (
+            {
+                "get_dp_group": _FakeGroup(2, 1),
+                "get_sp_group": _FakeGroup(4, 2),
+                "get_tp_group": _FakeGroup(2, 1),
+            },
+            False,
+            "DiffusionWorker_DP1_SP2_TP1",
+        ),
+        (
+            {
+                "get_dp_group": _FakeGroup(4, 3),
+                "get_fs_group": _FakeGroup(2, 1),
+            },
+            False,
+            "DiffusionWorker_DP3_FS1",
+        ),
+        (
+            {"get_ep_group": _FakeGroup(4, 2)},
+            True,
+            "DiffusionWorker_EP2",
+        ),
+        ({}, True, "DiffusionWorker"),
+    ],
+    ids=[
+        "all-singleton",
+        "tp-only",
+        "dp-only",
+        "pp-only",
+        "tp-cfg",
+        "dp-sp-tp",
+        "standalone-hsdp",
+        "expert-parallel",
+        "singleton-ep",
+    ],
+)
+def test_setup_uses_initialized_parallel_groups(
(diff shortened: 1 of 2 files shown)


prettygirlisnotme:
Self-review:

- Checked the lifecycle placement: the early worker setup keeps a generic title, then the title/log prefix is refreshed immediately after `initialize_model_parallel()` and before model loading.
- Checked that active ranks come from the initialized DP/PP/SP/CFG/TP groups, with singleton dimensions omitted; FS/HSDP and EP are included when applicable, and EP is only queried when enabled.
- Checked that the process title and log prefix use the same topology identifier via the upstream vLLM helpers. Missing `setproctitle` remains non-fatal.
- Reviewed the diff for behavior changes outside observability; there are no scheduling, communication, model-output, or performance-path changes.
- Focused coverage includes singleton, TP/DP/PP, TP×CFG, DP×SP×TP, HSDP/FS, EP, missing process-title support, and a Linux `ps` subprocess check.

vllm_omni/diffusion/worker/diffusion_worker.py
```diff
@@ -283,6 +323,7 @@ def init_device(self) -> None:
                 hsdp_replicate_size=parallel_config.hsdp_replicate_size if parallel_config.use_hsdp else 1,
                 enable_expert_parallel=parallel_config.enable_expert_parallel,
             )
+            _setup_diffusion_worker_proc_title_and_log_prefix(enable_ep=parallel_config.enable_expert_parallel)
```
fhfuih, line 326: Any reason now the title is set twice? (Another call is below at the original location)

prettygirlisnotme: Yes, this is intentional. The first call keeps the generic `DiffusionWorker` name before model-parallel init. After `initialize_model_parallel()`, the second call updates it with the real group ranks. This follows the current vLLM `WorkerProc` flow, so the worker is still identifiable during early startup.

tests/diffusion/test_diffusion_worker_process_title.py
```diff
@@ -0,0 +1,218 @@
...
+    )
+    process = subprocess.Popen(
+        [sys.executable, "-c", child_script],
+        stderr=subprocess.PIPE,
+        text=True,
+    )
+
+    observed_title = ""
+    try:
+        for _ in range(100):
```
hsliuustc0106, line 196: [P3] ps-title poll budget may be too short on cold CI runners

The loop gives the child ~11.5s (100 iterations x 0.1s sleep plus a ps spawn each) to start Python, import vllm, and set the title. A fresh CI container's first `import vllm` pulls in torch and can exceed that on a cold FS cache, flaking this otherwise-Linux-gated test with `expected 'vLLM-Omni::DiffusionWorker_TP1' in process title, got ''`. Consider raising the iteration count, or polling against a wall-clock deadline (e.g. 60s) instead.

hsliuustc0106 (approved):

hsliuustc0106:
reverted due to CI https://buildkite.com/vllm/vllm-omni/builds/14277/canvas

prettygirlisnotme:
Thanks for the follow-up and revert. I checked the failure and understand the issue now — #5531 changed the parallel-state layout while #6307 was open, which left the "get_fs_group" dependency stale at merge time.

Thanks for catching this.

hsliuustc0106:
I handled it by myself with ACK to your contribution due to the time limitation for new stable release. thanks anyway

PR #6308, opened 2026-08-18
[Core] Add UniProc diffusion executor for single-GPU

rahul-steiger-nv:
A 1-GPU diffusion worker has nothing to distribute but still paid for a subprocess, MessageQueues, and `/dev/shm` output segments. Select `"uni"` when the backend is omitted and `num_gpus == 1`, matching vLLM's `world_size == 1` default. Explicit `"mp"` keeps process isolation.

## Purpose

Add `UniProcDiffusionExecutor`: run the diffusion worker in-process on a single GPU so the engine does not spawn a subprocess, allocate MessageQueues, or create `/dev/shm` output segments.

- Omitted `distributed_executor_backend` + `num_gpus == 1` → `"uni"`.
- Explicit `"mp"` is unchanged (process isolation / RPC timeouts).
- `num_gpus > 1` still defaults to `"mp"`.
- Deploy YAMLs that already pin `"mp"` are left alone.

This is the diffusion counterpart of vLLM's `UniProcExecutor`. Fatal CUDA faults latch `is_failed` so the engine goes dead, matching multiproc process-death. Recoverable errors (OOM, bad request) stay per-request.

## Test Plan

```bash
pytest tests/diffusion/test_uniproc_executor.py -m 'core_model and cpu' -q
pytest tests/entrypoints/test_async_omni_diffusion_config.py -q
```

**vLLM Version:** latest compatible with current `main`

**vLLM-Omni Commit:** `ee67bb3be4`

## Test Result

- `tests/diffusion/test_uniproc_executor.py`: 15 passed (selection, explicit `mp`, multi-GPU default, RPC / CUDA-fault latch).
- `test_async_omni_diffusion_config.py`: omitted backend stays `None` through the serve path (no coerce to `"mp"`).

## Measured A/B (uni vs pinned mp), 1x GB300, Cosmos3-Nano

Frozen pair: same commit, same process, only `distributed_executor_backend`
differs (`"mp"` pinned reproduces the pre-PR behavior; omitted takes the new
`"uni"` default). bf16, guardrails off, seed 42 both arms, 3 interleaved
rounds per resolution (1 warmup + 2 measured, alternating uni/mp to cancel
run-to-run drift).

Reproduction (recipe's official 720p T2V params; the offline example script
doesn't expose the backend flag, so pin it directly through the `Omni` API):

```python
from vllm_omni.inputs.data import OmniDiffusionSamplingParams
import torch, time

omni = Omni(
    model="nvidia/Cosmos3-Nano",
    model_config={"guardrails": False},
    # distributed_executor_backend="mp",   # present: mp arm; absent: uni arm
)
gen = torch.Generator(device="cuda").manual_seed(42)
sp = OmniDiffusionSamplingParams(
    height=720, width=1280, num_frames=189, num_inference_steps=35,
    guidance_scale=6.0, generator=gen,
    extra_args={"flow_shift": 10.0, "max_sequence_length": 4096,
                "use_resolution_template": False, "use_duration_template": False},
)
t0 = time.perf_counter()
out = omni.generate({"prompt": "A robot arm is cleaning a plate in the kitchen.",
                      "negative_prompt": "blurry, distorted, low quality, jittery, deformed",
                      "modalities": ["video"]}, sp)
print(time.perf_counter() - t0)
```

| Metric | uni (this PR) | mp (pinned, pre-PR default) | Delta |
|---|---|---|---|
| Latency, 1280x720 / 189f / 35 steps | 82.00 s | 83.08 s | -1.3%, ranges non-overlapping across rounds |
| Latency, 480x832 / 189f | 24.64 s | 24.37 s | no measurable difference |
| Peak VRAM (reserved) | 47.9 GB | 50.8 GB | -2.9 GB |
| Engine init (first ready) | 16.6 s | 50.9 s | ~3x |
| Process-start-to-ready | 44 s | 78 s | import time is noisy (16-33s), not attributable to the executor |
| Processes | 2 | 4 | - |
| `/dev/shm` segments | 1 | 12 | - |
| Output correctness | md5-identical to mp arm, same `np.ndarray float32` type | - | bit-identical at fixed seed |


vllm_omni/diffusion/executor/uniproc_executor.py
```diff
@@ -0,0 +1,240 @@
...
+        """Whether the CUDA context survived the failure we just caught.
+
+        Running the worker inline means a worker fault no longer kills a
+        process, so nothing else notices it. Recoverable faults (OOM, an
+        invalid request) leave the context intact and must not take the
+        deployment down. A sticky fault (illegal memory access, ECC error)
+        poisons the context: every later launch fails, so the engine needs
+        to be restarted. ``synchronize()`` re-raises a pending sticky error.
+        """
+        if not torch.cuda.is_available() or not torch.cuda.is_initialized():
```
hsliuustc0106: _device_is_usable_ only probes CUDA, so on NPU the \`torch.cuda.is_available()\` guard returns False, the probe short-circuits to True, and a sticky device fault never latches \`_is_failed\` — \`check_health\` keeps reporting healthy while every request fails per-request, losing the process-death → \`EngineDeadError\` parity this latch exists to preserve. Probing through \`torch.accelerator\` (\`torch.accelerator.synchronize()\`, as in \`worker/base.py\`) dispatches to the registered accelerator on NPU/ROCm and also resolves the TID251 \`torch.cuda.synchronize\` failure currently failing pre-commit.

vllm_omni/diffusion/executor/uniproc_executor.py
```diff
@@ -0,0 +1,240 @@
...
+                f"UniProcDiffusionExecutor supports a single GPU only, got num_gpus={num_gpus}. "
+                'Use distributed_executor_backend="mp" for multi-GPU deployments.'
+            )
+
+        # The worker still initializes a one-rank distributed environment so
+        # that group lookups (`get_dp_group()` and friends) resolve. At world
+        # size 1, vLLM's GroupCoordinator skips the device communicator and
+        # the shared-memory MessageQueue, and every collective short-circuits.
+        worker_cls_path = current_omni_platform.get_diffusion_worker_cls()
+        self.driver_worker = WorkerWrapperBase(
```
hsliuustc0106, line 67: Every single-GPU diffusion deployment now boots through this in-process path by default (most `vllm_omni/deploy/*.yaml` don't pin a backend), but the new tests mock `WorkerWrapperBase` end-to-end and no GPU lane has run the PR. The invariants only exercised live are the in-process `init_distributed_environment` (resolved `master_port`, one-rank group init), real `DiffusionOutput` flow through the inline/stage clients, and worker shutdown. Could you add one real single-GPU diffusion serve smoke (CI lane or documented local run) before merge?

vllm_omni/diffusion/executor/uniproc_executor.py
rahul-steiger-nv: updated

vllm_omni/diffusion/executor/uniproc_executor.py
rahul-steiger-nv: added

vllm_omni/diffusion/executor/uniproc_executor.py
```diff
@@ -0,0 +1,246 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+
+"""In-process diffusion executor for single-GPU deployments.
+
+Counterpart to :class:`MultiprocDiffusionExecutor` for ``num_gpus == 1``.
+The worker is constructed and driven in the engine process, so none of the
+multiproc machinery is created:
+
+* no spawned worker subprocess (and no second model load / CUDA context)
```
hsliuustc0106, line 10: The motivation is a resource/latency claim — subprocess, MessageQueues, and /dev/shm segments are pure overhead at world size 1 — but it has no measured evidence behind it. Could you add an A/B on one GPU, same commit with `distributed_executor_backend: mp` pinned vs omitted: startup time to first ready, peak memory (weights + context + IPC buffers, e.g. `max_memory_reserved` plus process RSS), and per-request latency over a fixed prompt set with warmup and repetitions? Exact commands and observed variance per the perf A/B contract, and a fixed-seed output comparison would double as the correctness guard.

rahul-steiger-nv: I added the comparison to the PR description

hsliuustc0106 (approved):
lgtm

hsliuustc0106:
fix precommits please

rahul-steiger-nv:
> fix precommits please

Fixed, sorry for the delay! Thanks for your help with this PR!

rahul-steiger-nv:
@hsliuustc0106 Fixed the memory leak. Let me know what you think!

hsliuustc0106:
> @hsliuustc0106 Fixed the memory leak. Let me know what you think!

can you try to check the mp leak problem? #6413

rahul-steiger-nv:
> can you try to check the mp leak problem? #6413

Sure, I will look into it today.

hsliuustc0106:
> > can you try to check the mp leak problem? #6413
> 
> Sure, I will look into it today.

I will merge this PR first, please open a seperate PR for MP mode. thanks for you help:)

PR #6310, opened 2026-08-18
[Diffusion] Implement paged AR↔DiT KV connector ("v1", reusing vllm's native mooncake)

fhfuih:
> [!NOTE]
> this PR was previously described as a stub component and only a portion of a new AR-DiT KV Cache connector implementation (a page-supported connector that reuses vLLM's native Mooncake connector instead of going through in-house `OmniKVTransferManager`).
>
> The scope is now expanded to actually implement this new connector, including the native request/metadata path and kicking off Mooncake load on the DiT side.
>
> Still, final page-tensor registration against `DiffusionKVModelRunnerBackend.kv_caches_by_layer` is stubbed until #6102 lands. This PR does **not** plug the new connector to the DiT side, nor prove DiT's internal KV cache consumption is end-to-end runnable yet.

## Purpose

This PR partially fulfills #5244 & implements a new page-supported KV cache connector for AR-to-DiT:

- Diffusion config and Scheduler admission already on this branch remain the entry point
- Worker connector init uses the real rank-local `KVCacheConfig` when paged mode is configured; page registration stays stubbed for #6102
  - If this PR lands first, future PRs may need to pay attention to the following stub points: https://github.com/vllm-project/vllm-omni/pull/6310#pullrequestreview-4980160702
- Router mints one shared `transfer_id` and passes source/target params
- DiT side receives `kv_connector_metadata` and calls `start_load_kv`
- AR source lifecycle is adjusted so Mooncake can actually send the final pages

### 1. Diffusion native connector assembly

- Keep `OmniDiffusionConfig.kv_transfer_config` as the entry point for native vLLM `MooncakeConnector`
- Create Scheduler-role and Worker-role connector instances for the Diffusion stage
- Validate `engine_id` eagerly so Mooncake handshake identity is not silently auto-generated

### 2. Worker connector init (registration deferred to #6102)

- Native Mooncake is gated by `diffusion_kv_mode=paged_scheduler` **and** `kv_transfer_config`; `dense_legacy` keeps OmniKVTransferManager only and rejects `kv_transfer_config`
- Initialize the Worker connector in `set_kv_cache_configs` with the real rank-local `KVCacheConfig` (not during `init_device` with a placeholder)
- Keep `maybe_register_vllm_kv_caches(...)` as a helper for layer-name tensors
- Leave `DiffusionWorker.set_kv_cache_configs()` call sites for `initialize_kv_cache` / `maybe_register_vllm_kv_caches` as #6102 stubs

### 3. Scheduler admission to native pages

- Reuse Scheduler-side external lookup / reserve / `update_state_after_alloc()` flow already present on this branch
- Continue building one opaque `kv_connector_metadata` payload per scheduling step after all `update_state_after_alloc()` calls
- Keep requests in `RUNNING`; this PR still ignores `load_async=True`

### 4. Router handshake and diffusion stage ingress

- Before submitting AR and DiT work, the Orchestrator mints one shared `transfer_id`
- Build matching source/target `kv_transfer_params` for Mooncake producer/consumer roles
- Pass target params through diffusion stage ingress so `OmniDiffusionRequest.kv_transfer_params` is populated and `_bind_vllm_kv_request_fields()` can copy the public bag onto native per-sequence requests
- Preserve the existing `kv_sender_info` / packed-tensor path for stage topologies that do not use native `kv_transfer_config`

### 5. DiT-side start-load entry

- When `DiffusionSchedulerOutput.kv_connector_metadata` is present, the DiT side binds that metadata and calls `start_load_kv`
- This is intentionally limited to **starting** the remote load on the consumer side
- This PR does not add blocking wait / completion aggregation before model forward

### 6. AR source lifecycle needed for handoff

- For the native-KV path, AR successful EOS is shimmed to `FINISHED_LENGTH_CAPPED` so Mooncake actually sends and delay-frees source pages
- The AR scheduler uses only the first return from `_connector_finished()` / `request_finished()` (`delay_free_blocks`)
- This PR does not use `request_finished_all_groups()`'s second return value as a handshake channel

## What This PR Does Not Do

- Does not allocate or register final DiT page tensors itself (`DiffusionKVModelRunnerBackend` belongs to `#6102`)
- Does not implement DiT-side blocking wait before forward
- Does not aggregate `finished_recving` / `finished_sending` across ranks
- Does not attach `KVConnectorOutput` onto diffusion runner outputs
- Does not bypass legacy OmniKV receive/prefetch paths as part of a full native e2e switch
- Does not implement a new DiT scheduler state dedicated for KV waiting
- Does not implement `ActiveKVConnector.no_forward()`

### CPU unit tests in repo

- [x] `tests/diffusion/diffusion_kv/test_v1_connector.py`
  - config roundtrip
  - scheduler create/shutdown
  - registration helper skip / call
  - transfer-id helper
- [x] `tests/diffusion/diffusion_kv/test_worker_kv_connector.py`
  - metadata -> `start_load_kv`
- [x] `tests/diffusion/diffusion_kv/test_worker_contract.py`
  - request RPC metadata routing stays intact
- [x] `tests/diffusion/diffusion_kv/test_scheduler_connector.py`
  - scheduler-side public/native handshake plumbing
- [x] `tests/diffusion/models/hunyuan_image3/test_diffusion_kv_request.py`
  - Hunyuan preprocess keeps `kv_transfer_params`
- [x] `tests/engine/test_orchestrator_native_kv_handshake.py`
  - Router mints the same `transfer_id` for AR and DiT
- [x] `tests/core/sched/test_omni_ar_scheduler_native_kv.py`
  - EOS shim to `FINISHED_LENGTH_CAPPED`
  - native path uses connector delay-free semantics

Details: https://github.com/vllm-project/vllm-omni/pull/6310#pullrequestreview-4980120988

### Command run

```text
pytest \
  tests/diffusion/diffusion_kv/test_v1_connector.py \
  tests/diffusion/diffusion_kv/test_scheduler_connector.py \
  tests/diffusion/diffusion_kv/test_worker_kv_connector.py \
  tests/engine/test_orchestrator_native_kv_handshake.py \
  tests/core/sched/test_omni_ar_scheduler_native_kv.py \
  tests/diffusion/models/hunyuan_image3/test_diffusion_kv_request.py \
  tests/diffusion/diffusion_kv/test_worker_contract.py \
  tests/engine/test_orchestrator_kv_sender_info.py \
  tests/diffusion/diffusion_kv/test_initialization.py \
  tests/diffusion/diffusion_kv/test_config.py \
  -q --tb=short
```

Result:

- `72 passed`

### GPU landing test

```text
pytest \
  tests/diffusion/diffusion_kv/test_native_kv_mooncake_landing.py \
  -q --tb=short
```

Result:

- `1 passed`

## Notes / Dependencies

- Final page ownership and Worker-side `register_kv_caches` wiring still depend on `#6102`
- CPU tests in this PR validate the handoff/control path using fakes and mocks; they do not replace live Mooncake GPU verification
- Unconfigured stages (`kv_transfer_config` omitted) keep previous behavior

**vLLM Version:** 0.27.0


fhfuih:
CC @asukaqaq-s @Bounty-hunter @Acerak01-fy @zwhzzz0821

fhfuih (commented):
> @fhfuih, please review your own changes and leave a short self-review comment describing what you checked.

This PR involves file conflict with just-merged 6094. Also contains some getattr/hasattr calls. Fixing them...

vllm_omni/diffusion/diffusion_kv/v1/connector.py
```diff
@@ -0,0 +1,163 @@
+# SPDX-License-Identifier: Apache-2.0
```
hsliuustc0106, line 1: why we call it native? is this a base connector?

fhfuih: Native as in vLLM's native KV connector, routing away from our existing `OmniKVTransferManager` implementation. What do you think of this word choice? What about "v1/connector.py", "upstream_connector", "vllm_connector"? 🤔

fhfuih: Decide to rename based on these rules:

- In general, this new connector system is "v1". Classes and functions related to the creation, operation, cleanup of this system is named ".._v1"
- The file `native_connector.py` is renamed to `connector.py`
- Some functions and variables related to creating and configuration of underlying vLLM's native (Mooncake) KV transfer system is not renamed to ".._v1". Instead, "vllm_.."

vllm_omni/diffusion/worker/diffusion_worker.py
```diff
@@ -362,9 +361,6 @@ def load_model(self, load_format: str = "default", custom_pipeline_name: str | N
         if load_format != "dummy":
             assert self.model_runner.pipeline is not None
 
-        if self.model_runner is not None:
-            maybe_register_native_kv_caches(self.model_runner.get_native_kv_caches_dict())
-
```
fhfuih: NOTE: The latest commit removes this part (apart from renaming)

- `get_native_kv_caches_dict` is completely removed. To align with #6102, I expect the same thing to reside at its `DiffusionKVModelRunnerBackend.kv_caches_by_layer`
- Since that part in #6102 doesn't have a guard or stub for now, if I call `maybe_register_native_kv_caches`, it will enter the call path of the new paged KV connector, which is out of the scope of this PR

fhfuih (commented):
Some items being discovered and refined during self-review

- A previous version is introducing call path for `dense_legacy + kv_transfer_config` (specifying old KV connector, but providing new connector's config). This is removed and guarded
- Some not-so-elegant `hasattr`
- Missing an integration test to create KV from AR side and land it onto DiT side.

fhfuih (commented):
I'll stage the current state of this PR as "complete", self-reviewed and ready for review

Most lines of code changes are in the test folder.

```
❯ git diff HEAD~8..HEAD --shortstat -- tests
 16 files changed, 1258 insertions(+), 40 deletions(-)
```

To ease the code review pressure, below is a table to describe each test case's purpose (at the stage of writing this review)

#### `tests/diffusion/diffusion_kv/test_v1_connector.py`

| Case | Purpose |
|------|---------|
| `test_kv_transfer_config_roundtrip_to_worker_vllm_config` | YAML/`OmniDiffusionConfig` KV transfer settings are preserved into worker `VllmConfig` |
| `test_unconfigured_does_not_create_scheduler_connector` | Without `kv_transfer_config`, no scheduler connector is created |
| `test_scheduler_creates_kv_connector_v1_when_configured` | Scheduler init creates a SCHEDULER-role connector with the real `KVCacheConfig` |
| `test_set_kv_cache_configs_inits_worker_connector` | Worker connector is initialized in `set_kv_cache_configs` with the rank-local config |
| `test_maybe_register_vllm_kv_caches` | Registration helper skips `None`/`{}` and registers named layer tensors when present |
| `test_shutdown_is_idempotent` | Shutdown is safe with or without an active transfer group |
| `test_parse_kv_transfer_config` | Empty mapping → `None`; blank/missing `engine_id` raises |
| `test_kv_transfer_rejects_dense_legacy` | Native transfer requires `diffusion_kv_mode=paged_scheduler` |
| `test_scheduler_close_shuts_down_kv_connector_v1` | `scheduler.close()` shuts down and clears the connector |

#### `tests/diffusion/diffusion_kv/test_scheduler_connector.py`

| Case | Purpose |
|------|---------|
| `test_no_params_skips_connector_metadata` | Connector present but no handshake params → no metadata; request still runs |
| `test_cfg_sequences_each_call_connector_and_build_meta_once` | CFG (2 seqs): per-seq connector calls, one `build_connector_meta`, enters RUNNING |
| `test_public_params_are_copied_per_sequence` | Public `kv_transfer_params` are deep-copied onto each sequence |
| `test_missing_prompt_token_ids_are_filled_to_seq_len` | Missing prompt tokens are zero-filled to `seq_len` for admission |
| `test_ext_tokens_none_frees_and_stays_waiting` | Connector not-ready (`ext_tokens=None`) → stay WAITING and free reserved pages |

#### `tests/diffusion/diffusion_kv/test_worker_kv_connector.py`

| Case | Purpose |
|------|---------|
| `test_start_remote_kv_load_binds_metadata` | Model runner binds scheduler connector metadata and calls `start_load_kv` |

#### `tests/engine/test_orchestrator_native_kv_handshake.py`

| Case | Purpose |
|------|---------|
| `test_orchestrator_mints_matching_transfer_id_on_ar_and_dit` | Orchestrator stamps the same `transfer_id` and crossed engine roles on AR/DiT |

#### `tests/core/sched/test_omni_ar_scheduler_native_kv.py`

| Case | Purpose |
|------|---------|
| `test_native_eos_shims_to_length_capped_and_uses_delay_free_only` | Native EOS → `FINISHED_LENGTH_CAPPED` and delay-free (blocks not freed yet) |
| `test_native_error_does_not_shim_and_may_free` | ERROR path does not shim and may free blocks immediately |

vllm_omni/diffusion/worker/diffusion_worker.py
```diff
@@ -440,7 +449,16 @@ def set_kv_cache_configs(self, kv_cache_configs: list[KVCacheConfig]) -> None:
                 "Diffusion KVCacheConfig rank count mismatch: "
                 f"expected={self.od_config.num_gpus}, got={len(kv_cache_configs)}"
             )
-        self.model_runner.set_kv_cache_config(kv_cache_configs[self.rank])
+        rank_config = kv_cache_configs[self.rank]
+        self.model_runner.set_kv_cache_config(rank_config)
+        # self.model_runner.diffusion_kv_backend.initialize_kv_cache(rank_config)  # PR 6102
```
fhfuih: 🦄 stub here

vllm_omni/diffusion/worker/diffusion_worker.py
```diff
@@ -440,7 +449,16 @@ def set_kv_cache_configs(self, kv_cache_configs: list[KVCacheConfig]) -> None:
...
+        if (
+            self.vllm_config is not None
+            and self.vllm_config.kv_transfer_config is not None
+            and not has_kv_transfer_group()
+        ):
+            init_worker_kv_connector_v1(self.vllm_config, kv_cache_config=rank_config)
+        # maybe_register_vllm_kv_caches(self.model_runner.diffusion_kv_backend.kv_caches_by_layer)  # PR 6102
```
fhfuih: 🦄 stub here

vllm_omni/diffusion/worker/diffusion_model_runner.py
```diff
@@ -148,6 +149,8 @@ def __init__(
         # and BlockTable binding remain the responsibility of the paged
         # data-plane initialization path.
         self.kv_cache_config: KVCacheConfig | None = None
+        # Populated by DiffusionKVModelRunnerBackend (#6102). Empty until then.
+        # self.diffusion_kv_backend: Any | None = None
```
fhfuih: 🦄 stub here

vllm_omni/diffusion/sched/base_scheduler.py
```diff
@@ -203,6 +243,23 @@ def schedule(self) -> DiffusionSchedulerOutput:
...
+            # other consumer connectors we assemble today) ignore it and only
+            # drain internal recv queues, so a dummy object is enough to
+            # snapshot metadata. Passing this stub does not populate any
+            # SchedulerOutput fields and must not be treated as a real
+            # schedule payload. Mooncake currently drains internal recv
+            # queues and ignores scheduled-request fields, so the dummy
+            # remains valid. Call this once after every update_state_after_alloc.
+            kv_connector_metadata = self._kv_connector_v1.build_connector_meta(
+                object()  # type: ignore[arg-type]
+            )
```
fhfuih, line 256: 🦄 stub here

fhfuih (commented):
~~Below are the stubs that future PR like #6102 needs to pay attention to and resolve when finally plugging in the new connector.~~

CC @Acerak01-fy

Gaohan123:
@fhfuih Please resolve conflicts

asukaqaq-s:
> @fhfuih Please resolve conflicts

doing

vllm_omni/diffusion/worker/diffusion_worker.py
```diff
@@ -440,7 +446,29 @@ def set_kv_cache_configs(self, kv_cache_configs: list[KVCacheConfig]) -> None:
...
             )
-        self.model_runner.set_kv_cache_config(kv_cache_configs[self.rank])
+        rank_config = kv_cache_configs[self.rank]
+        self.model_runner.set_kv_cache_config(rank_config)
+        if (
+            self.vllm_config is not None
+            and self.vllm_config.kv_transfer_config is not None
+            and not has_kv_transfer_group()
+        ):
+            init_worker_kv_connector_v1(self.vllm_config, kv_cache_config=rank_config)
```
Gaohan123: [P1] Register the physical DiT KV pages before enabling remote loads

The Worker initializes the Mooncake connector here, but the corresponding `register_kv_caches()` call is still commented out. As a result, the production Worker never provides Mooncake with the actual DiT page tensors or their addresses, so a remote load cannot reliably land in the Scheduler-managed page pool.

The GPU landing test does not cover this production path because it manually allocates and registers the pages through `_stub_6102_allocate_and_register_pages()`.

Now that #6102 has been merged, could we initialize the physical `DiffusionKVModelRunnerBackend` pages first, build the layer-name-to-tensor mapping, and call `register_kv_caches()` before any connector metadata can reach the Worker? If registration remains intentionally out of scope, the new configuration path should fail explicitly instead of starting a transfer with no registered destination pages.

vllm_omni/diffusion/worker/diffusion_model_runner.py
```diff
@@ -186,6 +184,15 @@ def _validate_diffusion_kv_metadata(
...
             )
 
+    def _maybe_start_remote_kv_load(self, scheduler_output: DiffusionSchedulerOutput | None) -> None:
+        """Bind metadata and kick off native remote load on the DiT worker side."""
+        if scheduler_output is None:
+            return
+        metadata = scheduler_output.kv_connector_metadata
+        if metadata is None or not scheduler_output.scheduled_new_reqs:
+            return
+        start_load_kv(metadata)
```
Gaohan123: [P1] Wait for Mooncake receive completion before the first model forward

`start_load_kv()` only submits asynchronous receive work to Mooncake; it does not guarantee that the destination pages have been populated when this method returns. Both request mode and step mode proceed directly into model execution after this call, so the pipeline can read stale or uninitialized KV pages.

The GPU landing test already demonstrates the required synchronization by polling `get_finished()` until `finished_recving` contains the request before checking the destination tensors.

Could we keep the request in a KV-waiting state, or add a rank-aggregated completion barrier, before its first forward? This also needs to aggregate `finished_recving` across all participating ranks rather than allowing one rank to proceed independently.

vllm_omni/engine/orchestrator.py
```diff
@@ -2381,6 +2387,84 @@ async def _prewarm_async_chunk_stages(
...
+        if req_state.final_stage_id < 1:
+            return
+        dit_cfg = None
+        ar_cfg = self._stage_kv_transfer_config(0)
+        for stage_id in range(1, req_state.final_stage_id + 1):
+            if self.stage_pools[stage_id].stage_type == "diffusion":
+                dit_cfg = self._stage_kv_transfer_config(stage_id)
+                if dit_cfg is not None:
+                    break
+        if dit_cfg is None:
```
Gaohan123: [P1] Reject incomplete or one-sided native KV handshakes

This path only requires a configured DiT endpoint. It still creates and attaches source/target parameters when `ar_cfg` is `None`, and it does not require the AR bootstrap address to be present.

Mooncake's consumer-side `update_state_after_alloc()` requires `transfer_id`, `remote_engine_id`, and `remote_bootstrap_addr`. When one is missing, Mooncake logs a warning and skips the receive. However, `get_num_new_matched_tokens()` can already have reported the remote tokens as available, so the request may continue into the DiT forward without any KV having been transferred.

Please validate both AR and DiT endpoints, their roles, engine IDs, and the required bootstrap address before mutating either request. An incomplete configuration should fail fast. It would also be useful to add tests for a DiT-only configuration and for a missing AR bootstrap address.

asukaqaq-s:
We now support m:n AR-to-DiT transfer through the Mooncake Connector. This means AR and DiT can use different numbers of instances. However:

1. Performance has not been further optimized yet and will be improved in follow-up PRs. Since these optimizations have not been integrated, transferring large KV caches currently introduces more bubbles than the KVTransfer-based approach.
2. Since this PR is quite large, please let us know if it should be split. Currently, it is an atomic feature that integrates the Mooncake Connector and completes Sender Push transfers.

PTAL @fhfuih @hsliuustc0106 @Gaohan123 @Bounty-hunter @Acerak01-fy @zwhzzz0821 



# PR-1 HunyuanImage3 MoE KV Transfer Benchmark

Date: 2026-08-23  
Commit: `512851ba`

## 1. Deployment

Test configuration:

- 8 × NVIDIA RTX PRO 6000 Blackwell GPUs.
- HunyuanImage-3.0 MoE (`pipeline: hunyuan_image_3_moe`).
- All Dataset C requests are 512 × 512 with two denoising steps.
- Two AR replicas and two DiT replicas.
- Maximum concurrency: 4.
- Dataset C: 50 T2I and 50 I2I records, each replayed twice per backend.

Prepare the environment:

```bash
cd /root/omni/vllm-omni
source .venv/bin/activate

export HUNYUAN_MODEL=/root/.cache/modelscope/models/Tencent-Hunyuan--HunyuanImage-3.0/snapshots/master
export LD_LIBRARY_PATH=/root/omni/vllm-omni/.venv/lib/python3.12/site-packages/nvidia/cu13/lib:${LD_LIBRARY_PATH:-}
export VLLM_USE_FLASHINFER_SAMPLER=0
```

### Mooncake KVConnector

```bash
export MC_CUSTOM_TOPO_JSON=/root/omni/mooncake-tcp-topology.json

vllm serve "$HUNYUAN_MODEL" \
  --omni \
  --deploy-config /root/omni/vllm-omni/vllm_omni/deploy/hunyuan-image-3-mooncakeconnector.yaml \
  --host 127.0.0.1 \
  --port 8091 \
  --trust-remote-code
```

### OmniKVTransfer

```bash
vllm serve "$HUNYUAN_MODEL" \
  --omni \
  --deploy-config /root/omni/vllm-omni/vllm_omni/deploy/hunyuan_image_3_moe.yaml \
  --host 127.0.0.1 \
  --port 8091 \
  --trust-remote-code
```

## 2. Benchmark client

Run the following commands against each backend service.

### T2I

```bash
python benchmarks/diffusion/diffusion_benchmark_serving.py \
  --base-url http://127.0.0.1:8091 \
  --endpoint /v1/images/generations \
  --model "$HUNYUAN_MODEL" \
  --dataset custom \
  --dataset-path /root/omni/hunyuan_dataset_c_50.jsonl \
  --task t2i \
  --num-prompts 100 \
  --request-rate inf \
  --max-concurrency 4 \
  --num-inference-steps 2
```

### I2I

```bash
python benchmarks/diffusion/diffusion_benchmark_serving.py \
  --base-url http://127.0.0.1:8091 \
  --endpoint /v1/images/edits \
  --model "$HUNYUAN_MODEL" \
  --dataset custom \
  --dataset-path /root/omni/hunyuan_dataset_c_i2i_50.jsonl \
  --task i2i \
  --bot-task think_recaption \
  --num-prompts 100 \
  --request-rate inf \
  --max-concurrency 4 \
  --num-inference-steps 2
```

## 3. Results

### Mooncake KVConnector

| Workload | Success | Duration | Throughput | Mean | P95 | P99 |
| T2I | 100/100 | 28.693 s | 3.485 req/s | 1.132 s | 1.421 s | 1.809 s |
| I2I | 100/100 | 139.043 s | 0.719 req/s | 5.487 s | 7.150 s | 8.963 s |

| Transfer data | T2I | I2I |
| --- | ---: | ---: |
| HTTP 200 | 100 | 100 |
| Receive starts across two SP ranks | 200 | 200 |
| Receive completions across two SP ranks | 200 | 200 |
| Errors | 0 | 0 |

### OmniKVTransfer

| Workload | Success | Duration | Throughput | Mean | P95 | P99 |
| T2I | 100/100 | 31.295 s | 3.195 req/s | 1.238 s | 1.593 s | 1.802 s |
| I2I | 100/100 | 132.033 s | 0.757 req/s | 5.160 s | 6.414 s | 7.155 s |

| Transfer data, T2I + I2I | Count |
| --- | ---: |
| HTTP 200 | 200 |
| KV serialization events across two TP ranks | 400 |
| Receiver waits across two TP ranks | 400 |
| Successful receives across two TP ranks | 400 |
| Errors | 0 |

asukaqaq-s:
After excluding tests/ and the three renamed files:

38 existing files were modified
3 new files were added
0 files were renamed

The diff contains approximately 2,000 lines added and 819 lines modified.

Some of these changes are needed because `OmniKVTransfer` cannot be completely removed yet. I introduced a unified `TransferBackend` entry point so that the orchestration/control plane and `MooncakeConnector` can share the same code path.

`OmniKVTransfer` will be completely removed in a follow-up PR.

vllm_omni/diffusion/data.py
```diff
@@ -716,9 +716,7 @@ class OmniDiffusionConfig:
 
     # Local Diffusion KV ownership and cache-layout mode.
     diffusion_kv_mode: DiffusionKVCacheMode = DiffusionKVCacheMode.DENSE_LEGACY
-    # Maximum number of native BlockTable rows one public request can own
-    # (sequences plus independent contexts). The model adapter defines it.
-    diffusion_kv_max_rows_per_request: int | None = None
+    diffusion_kv_max_sequences_per_request: int | None = None
```
RuixiangMa: diffusion_kv_max_rows_per_request → diffusion_kv_max_sequences_per_request is a hard rename. Existing YAML deployments using the old name will silently lose the config or error out. Consider adding a deprecation alias or startup warning.

vllm_omni/diffusion/executor/abstract.py
```diff
@@ -72,8 +77,79 @@ def get_class(od_config: OmniDiffusionConfig) -> type[DiffusionExecutor]:
...
+                raise RuntimeError(
+                    "KV Connector all-rank completion is incomplete; missing sequence IDs: "
+                    f"{sorted(expected_ids - finished)}"
+                )
+            return kv_output
+        except Exception as exc:
+            if connector_metadata is None:
+                raise
+            # Quiesce every Worker before target pages can be recycled.
+            try:
```
RuixiangMa: The fail-closed rationale is sound (prevent DMA to recycled pages), but self.shutdown() is irreversible — the stage cannot recover from a transient network hiccup. Please document that this is intentionally non-recoverable, or consider a less destructive recovery path if feasible

vllm_omni/diffusion/diffusion_kv/kv_connector.py
```diff
@@ -0,0 +1,241 @@
...
+                    expected_finished_count=last_output.expected_finished_count,
+                )
+            elapsed = time.monotonic() - start
+            if elapsed >= timeout_s:
+                missing = sorted(expected_ids - finished_recving)
+                raise KVConnectorFatalError(
+                    f"KV Connector receive timed out after {elapsed:.3f}s on rank {rank}; "
+                    f"missing sequence IDs: {missing}"
+                )
+            time.sleep(min(0.01, max(0.0, timeout_s - elapsed)))
```
RuixiangMa: For large transfers that take seconds, this wastes CPU. The max(0.0, ...) is also redundant. Consider exponential backoff or event-based completion notification.

vllm_omni/diffusion/diffusion_kv/kv_connector.py
```diff
@@ -0,0 +1,241 @@
...
+    )
+    return connector
+
+
+def init_worker_kv_connector_v1(
+    vllm_config: VllmConfig,
+    kv_cache_config: KVCacheConfig,
+    kv_caches_by_layer: dict[str, Any],
+) -> ActiveKVConnector | None:
+    if vllm_config.kv_transfer_config is None:
```
RuixiangMa, line 141: NIT:Inconsistent with the declared return type and codebase convention. Minor, but linters may flag it.

vllm_omni/diffusion/diffusion_kv/kv_connector.py
```diff
@@ -0,0 +1,241 @@
...
+        raise TypeError(
+            f"Diffusion stage kv_transfer_config must be KVTransferConfig, got {type(kv_transfer_config)!r}"
+        )
+
+    if vllm_config is None or kv_cache_config is None:
+        raise ValueError("Native Diffusion Mooncake requires the sized VllmConfig and Scheduler KVCacheConfig")
+    if kv_cache_config.num_blocks <= 0 or not kv_cache_config.kv_cache_groups:
+        raise ValueError("Native Diffusion Mooncake requires a non-empty Scheduler KV cache plan")
+    connector_vllm_config = vllm_config
+    if connector_vllm_config.kv_transfer_config is None:
```
RuixiangMa: If the same VllmConfig is reused across stages or replicas, this mutation persists. Please document that this is intentional and safe, or create a copy.

fhfuih:
Int the latest commits, I cherrypicked the code changes by @asukaqaq-s and remove those that (1) plugs in the new KV connector to the call path, and (2) involves Hunyuan's adaptation (e.g., attention)

Gaohan123:
@asukaqaq-s @fhfuih Please resolve conflicts

asukaqaq-s:
I’ve cleaned up the code. This PR mainly removes redundant code and introduces the metadata and abstractions required by the Mooncake Connector. The end-to-end flow is not connected yet and will be completed in the next PR.

PTAL.


@Gaohan123 @hsliuustc0106 @RuixiangMa

Bounty-hunter (approved):
LGTM

RuixiangMa:
LGTM

vllm_omni/engine/orchestrator.py
```diff
@@ -2998,6 +3000,48 @@ async def _prewarm_async_chunk_stages(
...
         return True
 
+    def _maybe_attach_native_kv_transfer_params(self, req_state: OrchestratorRequestState) -> None:
+        """Attach one opaque ticket when a native Diffusion target is configured."""
+
+        if req_state.final_stage_id < 1 or not req_state.sampling_params_list:
+            return
+        has_native_target = any(
+            self.stage_pools[stage_id].stage_type == "diffusion"
+            and getattr(getattr(self.stage_pools[stage_id], "stage_vllm_config", None), "kv_transfer_config", None)
```
hsliuustc0106: Production diffusion pools never populate `stage_vllm_config`: `StageRuntime._assemble_stage_pools()` assigns it only when `stage_type != "diffusion"`. This condition is therefore always false, so no native transfer ID reaches either the AR params or DiT submission; the `SimpleNamespace` test models a state production cannot create. Please detect the retained diffusion config (or carry an explicit capability on the pool) and cover the real StageRuntime → Orchestrator path.

vllm_omni/diffusion/vllm_config.py
```diff
@@ -227,6 +227,8 @@ def configure_diffusion_vllm_config(vllm_config: VllmConfig, od_config: OmniDiff
         max_num_batched_tokens = getattr(od_config, "max_num_batched_tokens", None)
         if max_num_batched_tokens is not None:
             vllm_config.scheduler_config.max_num_batched_tokens = int(max_num_batched_tokens)
+    if od_config.kv_transfer_config is not None:
```
hsliuustc0106: This direct access breaks five existing CPU tests with `AttributeError` because their supported partial config objects do not define the new optional field. The focused affected suite currently reports 53 passed and 5 failed. Please keep this optional via `getattr(od_config, "kv_transfer_config", None)` as the surrounding optional fields do, or update the full config contract and fixtures.

vllm_omni/config/omni_config.py
```diff
@@ -740,6 +743,7 @@ class keeps the diffusion-only knobs from ``OmniDiffusionConfig`` without
     worker_extension_cls: str | None = None
     custom_pipeline_args: dict[str, Any] | None = None
     additional_config: dict[str, Any] = field(default_factory=dict)
+    kv_transfer_config: KVTransferConfig | None = None
```
hsliuustc0106, line 755: Pydantic materializes the raw mapping into `KVTransferConfig` here before `parse_kv_transfer_config()` sees it, and upstream `__post_init__` auto-generates `engine_id`. I reproduced the direct raw path raising as intended while the structured path accepts the same missing ID; the connector and diffusion projections also generate different UUIDs. Validate the raw mapping before constructing either projection and add a resolve-to-materialize regression test.

hsliuustc0106:
The PR description and test evidence need updating for the current head: they still describe the removed end-to-end start-load, AR EOS, and GPU-landing paths and report `72 passed`, while those paths/tests were removed and the focused CPU suite currently has five failures.

vllm_omni/engine/orchestrator.py
fhfuih: I suggest to not add this part, as this PR only introduces a new KV connector but not to plug it in to the production call path

vllm_omni/config/omni_config.py
fhfuih: Fixed with a pydantic field hook.

vllm_omni/diffusion/vllm_config.py
fhfuih: ~~I suggest to not do `getattr` in production code (for linting and mypy checking/CI comprehensiveness), but to complete the test mocks.~~ This field is triggering error to too many places. So for this case adding `getattr` is a better choice.

Relevant test files (and confirm they all pass with the latest commits):

```
tests/config/test_omni_config.py
tests/diffusion/test_diffusion_engine.py
tests/diffusion/test_diffusion_engine_cleanup.py
tests/diffusion/test_diffusion_engine_rpc_routing.py
tests/diffusion/test_diffusion_scheduler.py
tests/diffusion/test_diffusion_step_pipeline.py
tests/e2e/features/fullduplex/openai/test_runtime_adapter_boundary.py
tests/engine/test_duplex_import_boundary.py
tests/diffusion/test_diffusion_scheduler.py
tests/diffusion/diffusion_kv/test_initialization.py

423 passed, 1 skipped, 4 deselected
```

hsliuustc0106:
@Acerak01-fy is this related to your previous work? are they aligned in architecture design?

vllm_omni/diffusion/data.py
```diff
@@ -1089,6 +1092,17 @@ def __post_init__(self):
                 "paged_scheduler Diffusion KV does not support imported AR KV; "
                 "disable need_recv_cache until connector-aware import is implemented"
             )
+
+        if self.kv_transfer_config is not None:
```
SamitHuang, line 1102: This only rejects `need_send_cache` / `need_recv_cache`; default `diffusion_kv_mode` is still `dense_legacy`. `create_scheduler_kv_connector` and `init_worker_kv_connector` then assemble Mooncake whenever `kv_transfer_config` is set. Fail closed unless mode is `paged_scheduler`, and add a regression test for `dense_legacy + kv_transfer_config` — the suite that used to cover this is no longer in the PR.

vllm_omni/diffusion/diffusion_kv/kv_connector.py
```diff
@@ -0,0 +1,180 @@
...
+    )
+    return connector
+
+
+def init_worker_kv_connector(vllm_config: VllmConfig) -> None:
+    """Initialize the Worker-role connector after distributed setup."""
+
+    if vllm_config.kv_transfer_config is None:
+        return
+    ensure_kv_transfer_initialized(vllm_config, _empty_kv_cache_config())
```
SamitHuang: `ensure_kv_transfer_initialized` is one-shot and this passes `KVCacheConfig(num_blocks=0, kv_cache_tensors=[], kv_cache_groups=[])`. Enabling `kv_transfer_config` will start Mooncake with no pages, and #6102 cannot later register real rank-local tensors on the same transfer group. Defer Worker init until real `KVCacheConfig` exists, or document that this call must be removed before registration.

tests/engine/test_orchestrator_native_kv_metadata.py
```diff
@@ -0,0 +1,60 @@
...
+pytestmark = [pytest.mark.core_model, pytest.mark.cpu]
+
+
+def _orchestrator_with_native_diffusion_stage() -> Orchestrator:
+    orchestrator = object.__new__(Orchestrator)
+    orchestrator.stage_pools = [
+        SimpleNamespace(stage_type="llm", stage_vllm_config=SimpleNamespace(kv_transfer_config=None)),
+        SimpleNamespace(
+            stage_type="diffusion",
+            stage_vllm_config=SimpleNamespace(kv_transfer_config=SimpleNamespace(kv_connector="MooncakeConnector")),
```
SamitHuang: `StageRuntime._assemble_stage_pools()` leaves `stage_vllm_config=None` for every diffusion pool. This fixture plants a Mooncake `kv_transfer_config` on that field, so the test is green for a state production cannot create and `_maybe_attach_native_kv_transfer_params` never mints a ticket. Detect the retained diffusion config (or an explicit pool capability) and cover the real StageRuntime → Orchestrator path, or drop this orchestrator wiring until that exists.

vllm_omni/engine/orchestrator.py
```diff
@@ -2998,6 +3000,48 @@ async def _prewarm_async_chunk_stages(
...
+        req_state: OrchestratorRequestState,
+    ) -> dict[str, Any]:
+        submit_kwargs: dict[str, Any] = {
+            "kv_sender_info": self._build_kv_sender_info(
+                list(getattr(next_client, "engine_input_source", None) or [src_stage_id]),
+                request_id=request_id,
+            )
+        }
+        if req_state.native_kv_transfer_id is not None:
+            submit_kwargs["kv_transfer_params"] = {"transfer_id": req_state.native_kv_transfer_id}
```
SamitHuang: Even if detection is fixed, the AR extra_args and DiT submit bag are only `{"transfer_id": ...}`. `build_source_kv_transfer_params` / `build_target_kv_transfer_params` (roles, `remote_engine_id`, bootstrap) are unused, so Mooncake never gets a producer/consumer handshake. Either call those helpers here or delete them until the landing PR; a transfer_id-only ticket is not a native KV handoff.

vllm_omni/config/omni_config.py
```diff
@@ -231,6 +232,7 @@ class _ParallelEngineOverrides(_ParallelConfigEngineOverrides, total=False):
 
 class _ConnectorEngineOverrides(TypedDict, total=False):
     omni_kv_config: dict[str, Any]
+    kv_transfer_config: KVTransferConfig
```
SamitHuang: `kv_transfer_config` is now owned by both `_ConnectorEngineOverrides` / `OmniStageConnectorConfig` and `_DiffusionConfigProjection`. Structured YAML therefore stores it on `connector_config`, while the Scheduler/Worker read `OmniDiffusionConfig.kv_transfer_config` and the Orchestrator looks at `stage_vllm_config`. Pick one owner, flatten it into `build_diffusion_config`, and add a YAML → `OmniDiffusionConfig` roundtrip test so the field is not silently dropped.

SamitHuang (changes requested):
The current head is a connector-assembly stub, not the AR↔DiT handoff described in the PR body. The listed GPU landing, EOS shim, `start_load_kv`, and `test_v1_connector.py` paths are gone; remaining CPU tests mock objects that production never constructs.

Blocking gaps: native `kv_transfer_config` is not gated to `paged_scheduler`; Worker/Mooncake is initialized once with an empty `KVCacheConfig`; Orchestrator handshake detection and payload do not match StageRuntime or the unused source/target helpers. Update the PR description and add a StageRuntime/YAML roundtrip test before this lands.

vllm_omni/config/omni_config.py
asukaqaq-s: Fixed. A mode="before" validator now validates the raw mapping before Pydantic constructs KVTransferConfig, so a missing engine_id cannot be silently auto-generated.

vllm_omni/config/omni_config.py
asukaqaq-s: Fixed. _DiffusionConfigProjection is now the single structured-config owner of kv_transfer_config; the duplicate fields were removed from  _ConnectorEngineOverrides and OmniStageConnectorConfig.

vllm_omni/engine/orchestrator.py
asukaqaq-s: Removed the incomplete Orchestrator wiring from this PR.

vllm_omni/engine/orchestrator.py
asukaqaq-s: Removed the incomplete Orchestrator wiring from this PR.PR0 no longer generates or forwards transfer-ID-onlymetadata;

vllm_omni/diffusion/diffusion_kv/request.py
```diff
@@ -108,6 +111,11 @@ def __init__(
         # vLLM 0.26+ uses this optional boundary when publishing cached blocks;
         # zero means that no sparse-retention boundary is active.
         self.shared_prefix_boundary = 0
+        # Opaque request-scoped parameters consumed by an upstream KVConnector.
+        self.kv_transfer_params = kv_transfer_params
+        # Mooncake derives the remote-load length from this list. PR0 carries
+        # the field without creating a placeholder or starting a load.
+        self.prompt_token_ids = prompt_token_ids
```
wtomin, line 118: `DiffusionKVRequest` now accepts `prompt_token_ids`, with the comment in `kv_connector`/`request.py` saying Mooncake derives the remote-load length from this list. However, `BaseScheduler._make_request_state` only copies `kv_transfer_params` onto sequence requests; it does not propagate or populate `prompt_token_ids`.

Where is `prompt_token_ids` actually set in the real request path? If it remains `None`, follow-up Mooncake loading cannot derive the remote-load length it is supposed to use.

vllm_omni/diffusion/diffusion_kv/kv_connector.py
```diff
@@ -0,0 +1,180 @@
...
+    params["do_remote_prefill"] = True
+    params["do_remote_decode"] = False
+    if remote_engine_id:
+        params["remote_engine_id"] = remote_engine_id
+    if remote_bootstrap_addr:
+        params["remote_bootstrap_addr"] = remote_bootstrap_addr
+    return params
+
+
+def bootstrap_addr_from_kv_transfer_config(kv_transfer_config: KVTransferConfig | None) -> str | None:
```
wtomin, line 77: `mint_transfer_id`, `build_source_kv_transfer_params`, `build_target_kv_transfer_params`, and `bootstrap_addr_from_kv_transfer_config` are added but not referenced anywhere in this diff except tests indirectly.

vllm_omni/diffusion/diffusion_kv/kv_connector.py
```diff
@@ -0,0 +1,180 @@
...
+    from vllm_omni.diffusion.vllm_config import create_diffusion_vllm_config
+
+    kv_transfer_config = getattr(od_config, "kv_transfer_config", None)
+    assert isinstance(kv_transfer_config, KVTransferConfig)
+    vllm_config = create_diffusion_vllm_config(torch.device("cpu"), od_config)
+    vllm_config.kv_transfer_config = kv_transfer_config
+    return vllm_config
+
+
+def _validate_kv_transfer_config(config: KVTransferConfig) -> None:
```
wtomin, line 175: `_validate_kv_transfer_config` permits `kv_role` to be `None`. But `create_scheduler_kv_connector` explicitly passes `KVConnectorRole.SCHEDULER`, while `init_worker_kv_connector` delegates to upstream `ensure_kv_transfer_initialized` without explicitly choosing the worker role.

If `kv_role` is `None`, how does `ensure_kv_transfer_initialized` know whether the worker side should act as consumer/producer/both? Is `None` actually valid for a diffusion-worker config, or should the diffusion projection require a concrete role?

vllm_omni/diffusion/diffusion_kv/kv_connector.py
```diff
@@ -0,0 +1,180 @@
...
+    )
+
+
+def shutdown_kv_connector(*, scheduler_connector: KVConnectorBase_V1 | None = None) -> None:
+    """Shutdown Worker and Scheduler connector objects idempotently."""
+
+    ensure_kv_transfer_shutdown()
+    if scheduler_connector is not None:
+        scheduler_connector.shutdown()
+
```
wtomin:
```suggestion
    if scheduler_connector is not None:
        scheduler_connector.shutdown()
    ensure_kv_transfer_shutdown()

```

Reversing the cleanup order is more robust because the scheduler-owned connector should release its resources before the process-wide KV transfer runtime is torn down.

asukaqaq-s: done, thanks

vllm_omni/diffusion/diffusion_kv/kv_connector.py
asukaqaq-s: prompt_token_ids is intentionally not populated in PR0. This PR only assembles the native connector and does not start remote KV loading. The follow-up e2e PR1 populates it in prepare_kv_connector_request(), using num_transfer_tokens/prefix_len before starting the load. We keep this field as the request contract for PR1.

These helpers are intentionally unused in PR0 because the AR-to-DiT handoff is out of scope here. The follow-up PR1 Orchestrator uses them to build the producer/consumer Mooncake metadata. Removing them from the shared base would require reintroducing the same API in PR1.

asukaqaq-s: Upstream already rejects configurations where `kv_connector` is set but `kv_role` is `None`. We now mirror this validation in `_validate_kv_transfer_config()` for pre- materialized config objects, so diffusion worker configurations must provide an explicit role. No additional change to the Worker initialization path is needed.

asukaqaq-s:
@Gaohan123 @hsliuustc0106  PTAL，

hsliuustc0106:
fix ci failure:


```
=================================================================== short test summary info ====================================================================
--
  | FAILED tests/diffusion/diffusion_kv/test_kv_connector.py::test_config_roundtrip_to_kv_vllm_config - ValueError: paged_scheduler Diffusion KV requires max_model_len or a model config with max_position_embeddings/max_sequence_length/seq_length
  | FAILED tests/diffusion/diffusion_kv/test_kv_connector.py::test_scheduler_assembles_kv_stub_and_shuts_it_down - ValueError: paged_scheduler Diffusion KV requires a native Scheduler KVCacheConfig
  | ===================================== 2 failed, 4470 passed, 31 skipped, 385 deselected, 31 warnings in 656.01s (0:10:56) ======================================
  | /usr/lib/python3.12/multiprocessing/resource_tracker.py:254: UserWarning: resource_tracker: There appear to be 1 leaked shared_memory objects to clean up at shutdown
  | warnings.warn('resource_tracker: There appear to be %d '
  | 🚨 Error: The command exited with status 1
  | user command error: exit status 1
```

Issue #6237, opened 2026-08-16
[RFC]: DTPS (DiT-load-aware Type-Priority Scheduling) for AR+DiT Unified Multimodal Comprehension/Generation Serving


yibeicanghai:
### Motivation.
---
AR+DiT unified pipelines such as HunyuanImage-3.0 serve image understanding (`i2t`) and image generation (`t2i`/`it2i`) from a single AR stage, where comprehension tasks finish at AR while generation tasks proceed to the DiT (diffusion) stage. The two families **contend for the same AR scheduler**, and the default FCFS queue is oblivious to this asymmetry:

- When **comprehension tasks arrive first**, they occupy the AR stage and block generation tasks from reaching DiT, leaving the DiT stage **idle** and dropping end-to-end throughput.
- When **generation tasks arrive first**, they pile up in the DiT stage and **delay comprehension-task scheduling**, since i2t has no downstream stage to absorb but must wait behind a generation-heavy queue.

The root cause is that the AR scheduler has **no visibility into DiT-stage load** -- AR and DiT run as separate subprocesses, so `schedule()` cannot infer how loaded DiT is and cannot decide whether admitting the next request would feed DiT or just queue behind it. There is also no mechanism to cap admission when DiT is saturated, so a burst of generation requests floods DiT before the AR stage can react.

This RFC proposes **DTPS (DiT-load-aware Type-Priority Scheduling)**, which reorders the AR waiting queue by **task family + a cross-process DiT load signal** (pushed via the existing UTILITY ZMQ channel). It defines a load metric **T** -- generation tasks already scheduled by AR but not yet executing diffusion -- against a threshold `dit_load_threshold`: when DiT is busy (`T >= threshold`) it prioritizes comprehension; when DiT is idle (`T < threshold`) it prioritizes generation **with a per-step admission cap** so DiT is fed exactly to the threshold without piling up, plus a hardcoded `i2t_aging_s` starvation guard (100s). The change is **opt-in, AR-stage-only, and resource-neutral** -- it reorders admission order only and never alters model outputs.

To evaluate it, the RFC also adds a **mixed-task benchmark** (`benchmarks/omni/`) since the existing `benchmarks/diffusion/` covers only pure generation and cannot mix `i2t` with `t2i`/`it2i` or report per-task stats -- a gap that today makes this contention invisible and unmeasurable.

[...]

PR #6315, opened 2026-08-18
[Core] DTPS DiT-load-aware Type-Priority Scheduling

yibeicanghai:
Related to #6237

Implements the DTPS RFC (#6237): a dynamic, load-aware AR waiting-queue reordering strategy for mixed AR+DiT deployments (e.g. HunyuanImage-3.0).

## Goal

Prevent DiT starvation, avoid DiT queue overload, and guarantee no task starves — while staying fully model-agnostic and opt-in (absent the config, behavior is identical to `main`).

## Dynamic 3-tier priority (per schedule cycle)

- **L0 (highest — starving):** **any** request (ar_only or ar_downstream) waiting longer than `aging_s` (hardcoded 100s). Anti-starvation overrides all load-awareness.
- **L1 (budgeted DiT):** when DiT is hungry (`min_waiting < threshold`), a dynamic admission budget is granted; `ar_downstream` within budget is prioritized to feed the downstream, sorted by `num_prompt_tokens` ASC (shorter AR dwell → DiT starts sooner).
- **L2 (standard AR):** remaining `ar_only` requests (FCFS). When DiT is busy or the budget is exhausted, these tasks are prioritized.

Over-budget `ar_downstream` tasks are demoted below L2 (also sorted by `num_prompt_tokens` ASC) to prevent downstream pile-ups.

## Architecture

- `vllm_omni/core/sched/dtps_scheduler.py` — `DTPSScheduler`, one instance per AR stage replica, invoked once per `schedule()` via `maybe_reorder_waiting`. `OmniARScheduler` constructs it only when `dit_load_aware: true` is set on the AR stage; otherwise pure FCFS.
- `vllm_omni/core/sched/dit_load_state.py` — thread-safe in-process aggregator (per-replica unions of waiting/running ids). `update()` returns a boolean indicating whether the data changed, so the Orchestrator can skip redundant ZMQ pushes when the load snapshot is unchanged.
- The Orchestrator polls each DiT replica's load (throttled to 0.1 s/replica), aggregates per-replica `(waiting, running, ids)` into a `DitLoadSnapshot`, and pushes the snapshot to all AR stage replicas via the existing **UTILITY ZMQ channel** (`call_utility_async`). The AR subprocess receives it through `StageEngineCoreProc.omni_update_dit_load`, which vLLM's `getattr(self, method_name)` discovers via MRO — zero vLLM patch, zero new socket. Colocated deployments use `ipc://` transport; distributed deployments use `tcp://` transparently. All orchestrator DTPS code paths (state init, polling, pushing, replica eviction) are gated by `dit_load_aware` — zero overhead when disabled.

### Task classification — model-agnostic

Requests are classified as `ar_only` (terminates at this AR stage) or `ar_downstream` (proceeds to DiT) by reading the topology-resolved `omni_final_stage_id` field from `additional_information`. This field is stamped by the engine layer based on the request's output modality and the pipeline's `final_output_type` declarations — no model-specific task-type strings (e.g. `i2t`/`t2i`/`it2i`) are used.

### Blind in-flight tracking

When a downstream request finishes AR but hasn't yet surfaced in a DiT poll, its id is tracked in a blind-spot set so the phase decision isn't fooled by poll-lagged DiT-load reports. Blind ids are evicted by:
- **De-dup:** the id appears in DiT's reported waiting or running ids (it has reached DiT).
- **Age timeout:** the id exceeds `_DIT_INFLIGHT_MAX_AGE_S` (guards a dead DiT).

## Configuration (per-stage, opt-in)

```yaml
stages:
  - stage_id: 0          # AR stage
    dit_load_aware: false
    dit_load_threshold: 2
```

Two per-stage knobs on the AR stage's deploy config:
- `dit_load_aware` — opt-in switch (default `false`/`None`)
- `dit_load_threshold` — DiT busy/idle threshold + admission budget cap (default `2` when `None`)

The starvation-guard timeout (`aging_s`) is hardcoded at 100s in the scheduler — no user configuration needed. Fields flow through the existing `StageDeployConfig` → `engine_args` → `OmniEngineArgs` → `OmniModelConfig` plumbing — no dict wrapping, no `from_config` parsing, same pattern as `step_execution`.

### Gating

DTPS is enabled only when **both** conditions are met:
1. **User opt-in:** `dit_load_aware: true` on the AR stage in the deploy YAML
2. **Topology check:** the AR stage has a downstream DIFFUSION stage that directly consumes its output (checked via `input_sources` in `merge_pipeline_deploy`)

Any AR+DiT two-stage pipeline that has the AR stage feeding into a DiT stage can enable DTPS — no model-specific declaration needed.

> **Validation scope:** Although the design allows any AR+DiT two-stage pipeline to enable DTPS, the feature has only been validated on **HunyuanImage-3.0-Instruct**. All performance data below is based on this model. Other AR+DiT models should be validated before enabling in production.

## Performance

vllm-omni v0.23.0, Ascend A3, AR+DiT unified deployment, HunyuanImage-3.0-Instruct. Mixed task: **20 concurrent**, AR `max_num_seqs=3`, **i2t : t2i : it2i = 14 : 3 : 3**; `it2i` sub-mix `recaption : think : think_recaption = 1 : 1 : 1`; generation resolutions `512² : 1024² : 1280×720 = 1 : 1 : 1`; randomized request order *and* randomized inputs, **10 rounds (seeds 1–10)**, 100 ms send interval for reproducibility.

**End-to-end time per round (seconds) — FCFS baseline vs DTPS**

| Round | FCFS baseline | DTPS | Δ vs baseline |
|---:|---:|---:|---:|
| R1  | 204.31 | 183.48 | +10.20 % |
| R2  | 256.00 | 210.79 | +17.66 % |
| R3  | 163.89 | 155.16 | +5.33 %  |
| R4  | 234.17 | 177.79 | **+24.08 %** |
| R5  | 193.78 | 186.82 | +3.59 %  |
| R6  | 181.77 | 158.56 | +12.77 % |
| R7  | 221.86 | 180.22 | +18.77 % |
| R8  | 204.11 | 184.59 | +9.56 %  |
| R9  | 178.41 | 175.23 | +1.78 %  |
| R10 | 200.86 | 179.32 | +10.72 % |
| **Mean** | **203.92** | **179.20** | **+12.12 %** |

## Fail-safe design

DTPS is a throughput optimization and degrades gracefully at every boundary: UTILITY push failures, scheduler construction failure, and DiT-load poll failures all log at `debug`/`warning` and fall back to pure FCFS — the AR scheduler is never blocked.

## Lint / DCO

- All pre-commit hooks pass (ruff-check, ruff-format, typos, check-pickle-imports, signoff).
- DCO `Signed-off-by` present.

## Follow-ups (not in this PR)

- Regression tests: `DitLoadState` update/snapshot round-trip + `maybe_reorder_waiting` 3-tier layering — to be added in a follow-up.
- Benchmark numbers (RTF/latency at 2+ concurrency, hardware, software versions) — to be attached separately.


yibeicanghai:
Self-review for #6315:

**What this PR does:** Adds DTPS (DiT-load-aware Type-Priority Scheduling) for AR+DiT deployments — opt-in via the `dtps:` deploy block; behavior is identical to `main` when absent/disabled. 4-tier priority (starving AR > budgeted DiT > standard AR > excess DiT); cross-process DiT-load snapshot over a seqlock shared-memory segment; AR subprocess reattaches by name.

**What I checked:**
- Lint/DCO: all pre-commit hooks green (ruff-check, ruff-format, typos, check-pickle-imports, signoff).
- Self code-review of the diff: every `except Exception:` is logged (`debug`/`exception`) and confined to best-effort degrade boundaries — no silent swallow on the request/classification path; tightened type hints (`Request`, `list[Request] | None`, `DTPSScheduler | None`); removed a dead return (`_ar_proxy_parts`→`_ar_proxy`) and a stale docstring ref; lifted a function-body import to top-level.
- Fail-safety: DTPS construction / SHM attach-snapshot-write / DiT-load poll failures all fall back to pure FCFS — the AR scheduler is never blocked; SHM reads are seqlock + field-validated.
- Config: `omni_dtps_config`/`dtps` validated in `_build_dtps_config_block` (malformed entries warn + default, never break deploy); `hunyuan_image_3_moe.yaml` is the reference example with DTPS disabled by default (flip `dtps.enabled` to enable).

**Not in this PR (follow-ups):** regression tests (`DitLoadSharedState` seqlock round-trip + `maybe_reorder_waiting` 4-tier layering) and benchmark numbers (RTF/latency at 2+ concurrency, hardware, software versions).

cc @tzhouam @yinpeiqi @fake0fan @Sy0307 @Gaohan123

---

@vllm-omni-review-bot please review.

yinpeiqi:
Thanks for the detailed write-up and for iterating on this! The mixed AR+DiT starvation problem is real, and the RFC/PR makes the intent clear.

However, the change is much more invasive than a scheduler policy. DTPS is not confined to `OmniARScheduler`: it adds a new control-plane loop through the Orchestrator, StagePool, diffusion subprocess IPC, UTILITY RPC into `StageEngineCoreProc`, plus chat-serving task tagging. That is a lot of architecture complexity for one scheduling heuristic. It's too heavy for the orchestrator hot path.

I also doubt that the target scenario is a strong, fairly research-specific assumption: mixed `i2t` / `t2i` / `it2i` on a shared AR+DiT pipeline. Most pipelines we actually ship are not this mixed-task shape. Task buckets are hardcoded (`i2t`/`t2t` vs `t2i`/`it2i`), classification is stamped only on the chat path. That is not a general omni scheduling primitive.

A few concrete issues on top of that:
1. **Not actually opt-in on the orch path.** `DitLoadState` is always constructed, and `_orchestration_loop` polls every diffusion replica on a 0.1s throttle, then `_push_dit_load_to_ar` fans out to all `stage_type == "llm"` pools. Disabling `dtps.enabled` still pays poll + UTILITY traffic. If this stays, the entire orch/client path has to be gated the same way as scheduler construction.
2. **New cross-stage coupling without a general API.** If we want DiT-load / backpressure to drive AR admission, it should be a reusable stage-load or backpressure mechanism, not a DTPS-specific poll/push wired through Orchestrator. Otherwise the next model-specific policy will fork the same layers again.
3. **Eval / test gap.** No unit tests (called out as follow-up). Numbers are one model, one mix (14:3:3), one platform, wall-clock round time with large run-to-run spread. That is not enough to take a core scheduling + orch change.

I'd rather keep this out of tree / as an experiment until there is a much thinner, general hook. Happy to discuss what that hook could look like, but I don't think this should land in core in its current form.

yibeicanghai:
> Thanks for the detailed write-up and for iterating on this! The mixed AR+DiT starvation problem is real, and the RFC/PR makes the intent clear.
> 
> However, the change is much more invasive than a scheduler policy. DTPS is not confined to `OmniARScheduler`: it adds a new control-plane loop through the Orchestrator, StagePool, diffusion subprocess IPC, UTILITY RPC into `StageEngineCoreProc`, plus chat-serving task tagging. That is a lot of architecture complexity for one scheduling heuristic. It's too heavy for the orchestrator hot path.
> 
> I also doubt that the target scenario is a strong, fairly research-specific assumption: mixed `i2t` / `t2i` / `it2i` on a shared AR+DiT pipeline. Most pipelines we actually ship are not this mixed-task shape. Task buckets are hardcoded (`i2t`/`t2t` vs `t2i`/`it2i`), classification is stamped only on the chat path. That is not a general omni scheduling primitive.
> 
> A few concrete issues on top of that:
> 
> 1. **Not actually opt-in on the orch path.** `DitLoadState` is always constructed, and `_orchestration_loop` polls every diffusion replica on a 0.1s throttle, then `_push_dit_load_to_ar` fans out to all `stage_type == "llm"` pools. Disabling `dtps.enabled` still pays poll + UTILITY traffic. If this stays, the entire orch/client path has to be gated the same way as scheduler construction.
> 2. **New cross-stage coupling without a general API.** If we want DiT-load / backpressure to drive AR admission, it should be a reusable stage-load or backpressure mechanism, not a DTPS-specific poll/push wired through Orchestrator. Otherwise the next model-specific policy will fork the same layers again.
> 3. **Eval / test gap.** No unit tests (called out as follow-up). Numbers are one model, one mix (14:3:3), one platform, wall-clock round time with large run-to-run spread. That is not enough to take a core scheduling + orch change.
> 
> I'd rather keep this out of tree / as an experiment until there is a much thinner, general hook. Happy to discuss what that hook could look like, but I don't think this should land in core in its current form.

Thanks for the review! Your feedback is highly valuable. I had similar thoughts on some of these points and have already made adjustments. Here are my responses to your concerns:

1. Regarding the task classification at the API server layer:
I agree that the previous implementation wasn't generic and was too tightly coupled to i2t/t2i/it2i. In the latest implementation, I've abstracted the task classification to be more generic. Now, if a request terminates after the AR stage, it is classified into an "AR-only" queue; if the AR stage is immediately followed by a DiT stage, it is identified as an "AR+DiT" task. It is no longer bound to specific task types. Please refer to the latest code for details.

2. Regarding the gating / opt-in path:
The code now validates that the DTPS feature can only be enabled under an AR+DiT topology. I've also added strict toggle isolation at the orchestrator's DiT polling entry point and other related paths. This ensures absolutely no impact (no polling overhead) on the existing logic when the feature is disabled. You can check the latest commits for this implementation.

3. Regarding the run-to-run spread in the test data:
I realize my PR description wasn't clear enough on this point. The high variance isn't from running the same dataset 10 times. Instead, the 10 rounds were run using incremental random seeds (from 1 to 10) to ensure test diversity. (If we ran 10 times with the same seed, the variance would be very small).
Because the seeds are different, the sequence of requests sent to the server varies in each round, even though the overall ratio of understanding vs. generation tasks remains constant.

When the random sampling happens to evenly interleave understanding and generation tasks, the baseline naturally behaves similarly to the DTPS scheduling policy, so the observed improvement is relatively small.

However, when the sampled queue clusters understanding tasks first and generation tasks later continuously, DTPS demonstrates its maximum value, resulting in a massive performance lift.

4. Regarding the assumption of the mixed-task scenario:
This is actually based on a real-world deployment scenario. The 7:3 ratio (understanding vs. generation tasks) I constructed is directly derived from our actual production user traffic distribution.

5. Regarding cross-stage coupling without a general API:
Currently, the communication between the Orchestrator and the AR/DiT processes has been refactored to fully reuse the existing ZMQ channels. If we want a more generic abstraction here, it seems we would need to strictly define what future information might need to be passed between the DiT and AR stages to design a truly generic channel. For the current scope, is reusing the underlying ZMQ channel sufficient? I'd love to hear if you have any better architectural suggestions for this part.

Thanks again for your time and feedback. The latest code has been pushed—please take a look. If anything above is still unclear, I'm more than happy to discuss it further! @yinpeiqi

yibeicanghai:
@hsliuustc0106 Could you please take a look at this feature PR when you get a moment?

As a quick summary: 
1. The IPC channels for DiT load‑aware information passing have been fully reused from the existing ZMQ channels between the vllm‑omni orchestrator and the AR/DiT stages. 
2. The deployment‑facing enable switches have been streamlined – we now expose only dit_load_aware (to turn load‑aware scheduling on/off for the AR stage) and dit_load_threshold (to adjust admission control for DiT tasks). 
3. The benchmark part is being adapted to the vllm bench --omni framework and will come in a later PR.

We already covered the overall design in the previous community sync, and all the outstanding issues from that discussion have been addressed. Please let me know if you have any further questions or feedback – happy to address them promptly.

vllm_omni/core/sched/dtps_scheduler.py
```diff
@@ -0,0 +1,260 @@
...
+        """
+        rid = req.request_id
+        cached = self._ar_only_cache.get(rid)
+        if cached is not None:
+            return cached
+        result = self._compute_ar_only(req)
+        self._ar_only_cache[rid] = result
+        return result
+
+    def evict_request(self, request_id: str) -> None:
```
RuixiangMa, line 77: Finished requests' blind in-flight entries persist up to 1s until _dit_phase timeout evicts them. This inflates inflight_blind count, making DiT appear busier than it is.

RuixiangMa, line 77: Finished requests' blind in-flight entries persist up to 1s until _dit_phase timeout evicts them. This inflates inflight_blind count, making DiT appear busier than it is.

vllm_omni/core/sched/dtps_scheduler.py
```diff
@@ -0,0 +1,260 @@
...
+        if info is None:
+            return False
+        if isinstance(info, dict):
+            pass
+        else:
+            try:
+                info = deserialize_additional_information(info)
+            except Exception:
+                logger.debug("[OmniDTPS] deserialize additional_information failed", exc_info=True)
+                return False
```
RuixiangMa, line 92: Corrupt requests are classified as ar_downstream, getting DiT-priority treatment. Safer fallback is True (stays at AR stage).

vllm_omni/core/sched/dtps_scheduler.py
```diff
@@ -0,0 +1,260 @@
...
+
+from vllm.logger import init_logger
+from vllm.v1.request import Request
+
+from vllm_omni.core.sched.dit_load_state import DitLoadSnapshot
+from vllm_omni.engine.serialization import deserialize_additional_information
+
+logger = init_logger(__name__)
+
+_DEFAULT_AGING_S = 100.0
```
RuixiangMa, line 32: may be too long for latency-sensitive deployments

RuixiangMa (commented):
@yibeicanghai Thanks for your contribution, the design is mostly sound. left a few comments:

- Orchestrator SPOF: crash → AR loses DiT visibility → falls back to "always idle" → DiT overload. No graceful degradation.

- L1 sorts by num_prompt_tokens ASC — long-prompt downstream tasks could starve within L1 before aging to L0. Consider secondary aging.

yibeicanghai:
> @yibeicanghai Thanks for your contribution, the design is mostly sound. left a few comments:
> 
> * Orchestrator SPOF: crash → AR loses DiT visibility → falls back to "always idle" → DiT overload. No graceful degradation.
> * L1 sorts by num_prompt_tokens ASC — long-prompt downstream tasks could starve within L1 before aging to L0. Consider secondary aging.

Thanks for the review! @RuixiangMa 

1. Orchestrator failure: I'll check if we can fall back to the default behavior upon failure—keeping it consistent with how the system behaves when this scheduling feature is disabled.
2. L1 starvation: L1 is sorted by num_prompt_tokens ASC, but to prevent long-prompt requests from starving in L1, we use the same aging mechanism as AR-only tasks. Requests that time out get bumped to L0 priority. You can check the code to confirm this logic—long-prompt starvation shouldn't be an issue here.
3. Remaining feedback: I'm currently working through the rest of your comments and will push updates soon.

vllm_omni/config/stage_config.py
```diff
@@ -409,6 +409,10 @@ class StageDeployConfig:
     async_scheduling: bool | None = None
     disable_hybrid_kv_cache_manager: bool | None = None
     mm_processor_cache_gb: float | None = None
+    # AR-stage DiT-load-aware scheduling (DTPS): when True, the AR scheduler
+    # reorders its waiting queue by task type + downstream DiT load.
+    dit_load_aware: bool | None = None
```
lishunyang12, line 414: Wire both fields through the structured config; even the default Hunyuan deployment fails.

vllm_omni/engine/orchestrator.py
```diff
@@ -1183,6 +1209,77 @@ def _absorb_diffusion_metrics(self, stage_id: int, replica_id: int, diffusion_ou
...
+            last = self._dit_load_last_poll.get(key, 0.0)
+            if (now - last) < self._DIT_LOAD_POLL_INTERVAL_S:
+                return
+        self._dit_load_last_poll[key] = now
+
+        # Exit-driven eviction: a dead replica's last waiting=0 would falsely
+        # signal idle, so drop it before polling.
+        if not pool.check_dit_health(replica_id):
+            try:
+                state.remove(stage_id, replica_id)
```
lishunyang12, line 1237: Remove dead replicas from DitLoadState in _handle_dead_replica; this check runs after output polling.

vllm_omni/engine/stage_pool.py
```diff
@@ -1197,6 +1199,67 @@ def poll_diffusion_output(self, replica_id: int) -> Any | None:
...
+        payload = dict(snapshot)
+        for replica_id in self.live_replica_ids():
+            client = self.clients[replica_id]
+            if client is None:
+                continue
+            call_utility = getattr(client, "call_utility_async", None)
+            if call_utility is None:
+                continue
+            try:
+                await call_utility("omni_update_dit_load", payload)
```
lishunyang12, line 1254: Bound this utility call with a timeout so stalled AR replicas can't freeze output routing.

vllm_omni/core/sched/dtps_scheduler.py
```diff
@@ -0,0 +1,260 @@
...
+        downstream_reqs.sort(key=lambda r: getattr(r, "num_prompt_tokens", 0) or 0)
+
+        if budget_raw is None:
+            downstream_head = downstream_reqs
+            downstream_tail: list = []
+        else:
+            downstream_head = downstream_reqs[:budget_raw]
+            downstream_tail = downstream_reqs[budget_raw:]
+
+        ordered = starving_reqs + downstream_head + ar_only_reqs + downstream_tail
```
lishunyang12, line 225: Sort starving_reqs by arrival_time; previous DTPS ordering isn't FCFS.

vllm_omni/core/sched/dtps_scheduler.py
```diff
@@ -0,0 +1,260 @@
...
+                npt = getattr(r, "num_prompt_tokens", 0) or 0
+                tag = "AR+DiT" if not self._request_is_ar_only(r) else "AR"
+                parts.append(f"{r.request_id}({tag},{w:.1f}s,{npt}t)")
+            return "[" + ",".join(parts) + "]"
+
+        logger.debug(
+            "[OmniDTPS] reorder: before=%s | L0_starving=%s L1_budget=%s "
+            "L2_ar_only=%s L3_overflow=%s | budget=%s eff_min=%s | "
+            "after=%s",
+            _fmt(before_list),
```
lishunyang12, line 244: Guard these queue dumps with logger.isEnabledFor(DEBUG) before formatting every request.

Issue #4340, opened 2026-06-11
[RFC]: [Cosmos3 optimization] Feature Support, Quantilization, Hardware Recipes, and Scenario Optimization


bjf-frz:
### Motivation.

  Cosmos3 covers multiple inference scenarios, including T2I, T2V, I2V, video generation with sound, and action-related generation. These scenarios have different latency, memory, and hardware requirements, especially between Cosmos3-Nano and Cosmos3-Super.

  The goal of this analysis is to provide an evidence-based feature matrix for Cosmos3 across cache acceleration, CPU offload, parallelism, LoRA, quantization, device support, and hardware recipes. It also identifies which features are already supported, which are explicitly unsupported, and which still require validation.

  This document follows the same direction as the continuous diffusion acceleration tracking RFC: maintain a Model x Feature support matrix and clarify Feature x Feature compatibility gaps before recommending production configurations.

#### Features Supported Table
The following table tracks Cosmos3 feature support across cache acceleration, offload, parallelism, quantization, LoRA, and execution features. It follows the same status convention as the continuous diffusion acceleration tracking RFC.

 | Model | TeaCache | Cache-DiT | UND K/V cache | SP (Ulysses & Ring) | CFG Parallel | Tensor Parallel | Pipeline Parallel | HSDP | CPU Offload (Layerwise) | VAE Patch Parallel | Quantization | LoRA Inference | Step Execution |
  |---|---|---|---|---|---|---|---|---|---|---|---|---|---|
  | Cosmos3-Nano | ❌ | ✅ | ✅ | ✅ | ✅ | ✅ | ❌ | ✅ | ✅ | ✅ | ✅ | ❌ | ❌ |
  | Cosmos3-Super | ❌ | ✅ | ✅ | ✅ | ✅ | ✅ | ❌ | ✅ | ✅ | ✅ | ✅ | ❌ | ❌ |

#### Quantization
Cosmos3 is marked as quantization-supported at the framework level, and its transformer passes the diffusion `quantization_config` into quantizable linear layers. The current Cosmos3 recipes, however, do not include method-specific validation or recommended quantized checkpoints. Therefore, this section tracks both the framework-level quantization support and the validation status of selected quantization methods.
  | Model | Quantization Framework | FP8 | NVFP4 |
  |---|---|---|---|
  | Cosmos3-Nano | ✅ | ❓ | ❓ |
  | Cosmos3-Super | ✅ | ❓ | ❓ |

[...]

PR #6316, opened 2026-08-18
[Model][Cosmos3] Implement SupportsStepExecution step-wise pipeline

86MaxCao:
## Summary

Implements the step-wise execution protocol (RFC #4340) for `Cosmos3OmniDiffusersPipeline`, splitting the monolithic `forward()` into `prepare_encode` / `denoise_step` / `step_scheduler` / `post_decode` so Cosmos3 can be scheduled at step granularity and participate in streaming output / continuous batching of diffusion requests.

- `Cosmos3OmniDiffusersPipeline` now inherits `SupportsStepExecution` and declares `supports_step_execution = True`.
- `prepare_encode` reuses the existing request-parsing / tokenization / latent-preparation helpers, covering t2i / t2v / i2v / v2v / action / sound modes, and populates the runner-owned `StepRequestState` (`state.extra`). The pipeline keeps no per-request caches, so cancelled or failed requests cannot leak GPU tensors.
- `denoise_step` reproduces the CFG semantics of `diffuse()`: CFG-parallel branch-per-rank, guidance-interval-aware sequential CFG with transformer K/V cache save/restore, and cache-dit paired-CFG compatibility.
- Each request snapshots its own deepcopy'd `FlowUniPC` scheduler and timesteps (same pattern as qwen_image / hunyuan_image3), so interleaved step requests cannot cross-contaminate multistep scheduler state.
- Multimodal (video/action/sound) noise predictions are flattened into a single batch-dim packed tensor returned by `denoise_step`, matching the runner's per-request batch-dim slicing; `step_scheduler` unpacks per-modality predictions from recorded shapes and performs the joint packed scheduler step, matching `diffuse()` numerics.
- `post_decode` decodes the final latents, assembles mode-specific outputs (image / video / audio / actions), drops the UND K/V session, and clears `state.extra`.
- `prepare_encode` reuses `forward()`'s request validation (distilled/Edge limits, action/sound combo checks, `action_gen`/`sound_gen` module checks, image+video exclusivity, shared modalities normalization) so step mode cannot silently bypass mode restrictions.
- Pipeline init rejects `step_execution=True` with `max_num_seqs > 1` upfront (`requires max_num_seqs=1`) instead of failing mid-denoise.

## Limitations

- **Single active request**: multi-request batching in step mode is not implemented yet; init requires `max_num_seqs=1`, and `denoise_step` raises `NotImplementedError` if more than one request is active (fail loudly rather than returning wrong predictions).
- **`num_outputs_per_prompt == 1`**: `forward()` realizes T2I `num_outputs_per_prompt > 1` by looping the whole diffusion; the step path requires `num_outputs_per_prompt == 1` and raises a clear error otherwise.
- **Transfer inference** is explicitly rejected in step mode (separate code path).
- **RoboLab/OpenPI observation requests** (`extra_args["robot_obs"]` / `["observation"]`) are explicitly rejected; their action-only envelope routing is not wired into step mode yet.
- Cancelled requests that never reach `post_decode` rely on the `SessionStateManager`'s count-based eviction (max_sessions) to reclaim their UND K/V session.

## Test Plan

- [x] 19 new CPU-only tests in `tests/diffusion/models/cosmos3/test_cosmos3_pipeline.py`, including a runner-level round-trip using the real `StepRequestState` → `InputBatch.make_batch` → `denoise_step` → per-request batch-dim slicing → `step_scheduler` → `post_decode` data path.
- [x] Full file passes: 96 passed (vllm 0.26.0 / torch 2.11.0); `ruff check` / `ruff format` / `git diff --check` clean.
- [x] GPU end-to-end: real `Cosmos3-Nano` checkpoint (35GB, local path), H20 GPU, t2v request with the same parameters as the merged Cosmos3 PR's curl example (1280x720, 193 frames, 35 steps, guidance 4.0, seed 42). `step_execution=True` output is **bit-exact identical** to the `forward()` baseline (`torch.equal=True`, max abs diff 0.0).

Usage: `step_execution` is an `Omni` engine arg (defaults to `False`); the request API is unchanged — only the engine construction differs:

```python

omni = Omni(model="nvidia/Cosmos3-Nano", step_execution=True, model_config={"guardrails": False})
out = omni.generate(prompt, sampling_params)  # same call as the default forward() path
```

- [ ] GPU E2E for i2v / v2v / sound / action modes (follow-up).

### DIFF
--- a/vllm_omni/diffusion/models/cosmos3/pipeline_cosmos3.py
+++ b/vllm_omni/diffusion/models/cosmos3/pipeline_cosmos3.py
@@ -29,13 +29,15 @@ is produced from sound latents rather than from ``multi_modal_data["audio"]``.
 
 from __future__ import annotations
 
+import copy
 import json
 import math
 import os
 import time
-from collections.abc import Iterable, Mapping
+from collections.abc import Iterable, Mapping, Sequence
 from dataclasses import fields
-from typing import Any, ClassVar
+from types import SimpleNamespace
+from typing import TYPE_CHECKING, Any, ClassVar
 
 import numpy as np
 import PIL.Image
@@ -60,6 +62,7 @@ from vllm_omni.diffusion.models.interface import (
     ReferenceVideoDecodeSpec,
     SupportImageInput,
     SupportsComponentDiscovery,
+    SupportsStepExecution,
 )
 from vllm_omni.diffusion.models.progress_bar import ProgressBarMixin, _is_rank_zero
 from vllm_omni.diffusion.models.schedulers.scheduling_flow_match_euler_discrete import (
@@ -80,6 +83,10 @@ from vllm_omni.experimental.world_models.session_state import (
     resolve_session_state_config,
 )
 from vllm_omni.inputs.data import OmniDiffusionSamplingParams
+
+if TYPE_CHECKING:
+    from vllm_omni.diffusion.worker.input_batch import InputBatch
+    from vllm_omni.diffusion.worker.utils import StepRequestState
 
 from .action import (
     ACTION_MODE_FORWARD_DYNAMICS,
@@ -776,6 +783,7 @@ class Cosmos3OmniDiffusersPipeline(
     CFGParallelMixin,
     SupportImageInput,
     SupportsComponentDiscovery,
+    SupportsStepExecution,
     ProgressBarMixin,
     DiffusionPipelineProfilerMixin,
 ):
@@ -820,6 +828,7 @@ class Cosmos3OmniDiffusersPipeline(
     """
 
     support_image_input: ClassVar[bool] = True
+    supports_step_execution: ClassVar[bool] = True
     color_format: ClassVar[str] = "RGB"
     _dit_modules: ClassVar[list[str]] = ["transformer.language_model", "transformer"]
     _encoder_modules: ClassVar[list[str]] = []
@@ -1011,6 +1020,11 @@ class Cosmos3OmniDiffusersPipeline(
         # stays in sync.
         self._cache_dit_requires_paired_cfg = False
 
+        # Step-wise execution keeps per-request state on the runner-owned
+        # StepRequestState (state.extra); the pipeline holds no per-request
+        # caches so cancelled / failed requests cannot leak GPU tensors here.
+        self._validate_step_execution_config(od_config)
+
         self.setup_diffusion_pipeline_profiler(
             enable_diffusion_pipeline_profiler=self.od_config.enable_diffusion_pipeline_profiler
         )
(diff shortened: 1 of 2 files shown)


86MaxCao:
Self-review completed:

- Reviewed the full diff against `main`.
- Verified the `SupportsStepExecution` lifecycle: `prepare_encode`, `denoise_step`, `step_scheduler`, and `post_decode`.
- Checked request-local scheduler/state isolation, CFG paths, multimodal prediction packing, and cleanup behavior.
- Confirmed unsupported combinations fail explicitly.
- Ran the Cosmos3 unit tests and formatting/static checks.
- Verified that the Cosmos3-Nano T2V step path is bit-exact with the existing `forward()` path in GPU testing.

I did not find any remaining blocking issues. The PR is ready for review.

@vllm-omni-review-bot, please review this PR.

hsliuustc0106:
@vklimkov-nvidia @ayushag-nv PTAL

vllm_omni/diffusion/models/cosmos3/pipeline_cosmos3.py
```diff
@@ -1148,6 +1162,751 @@ def _remap_ckpt_key(key: str) -> str | None:
...
+        elif p.is_t2i:
+            result = {"image": video}
+        else:
+            result = {"video": video}
+
+        # Cleanup UND K/V session and release per-request tensors held in
+        # state.extra so the runner can drop the state without leaks. (If a
+        # request is cancelled before post_decode, the SessionStateManager's
+        # count-based eviction reclaims the K/V session eventually.)
+        kv_state = extra.get("kv_state")
```
hsliuustc0106, line 2052: [P2] Release session K/V on failure and cancellation

With `enable_session_state_manager=True`, an exception in video/audio decoding skips this cleanup, and a request aborted during denoising never reaches `post_decode()` at all. The runner only removes its `StepRequestState`; the session manager still owns the request's GPU K/V until count-based eviction (64 sessions by default). This allows failed/aborted requests to accumulate retained GPU memory, whereas `diffuse()` drops the session in a `finally` block. Please release the session through a terminal cleanup path covering success, failure, and abort, and add regression coverage with the session manager enabled.

vllm_omni/diffusion/models/cosmos3/pipeline_cosmos3.py
```diff
@@ -1148,6 +1162,751 @@ def _remap_ckpt_key(key: str) -> str | None:
...
+                    "Cosmos3 action video tensor must have shape [1, 3, T, H, W] "
+                    f"or [3, T, H, W], got {tuple(p.video_tensor.shape)}."
+                )
+            if p.video_tensor is not None and p.video_tensor.shape[2] < p.num_frames:
+                pad = p.video_tensor[:, :, -1:].repeat(1, 1, p.num_frames - p.video_tensor.shape[2], 1, 1)
+                p.video_tensor = torch.cat([p.video_tensor, pad], dim=2)
+            elif p.video_tensor is not None and p.video_tensor.shape[2] > p.num_frames:
+                p.video_tensor = p.video_tensor[:, :, : p.num_frames]
+
+            if p.action_mode == ACTION_MODE_INVERSE_DYNAMICS and p.video_tensor is None:
```
hsliuustc0106, line 1537: [P2] Preserve the existing action-video input fallback

An inverse-dynamics request supplying its conditioning tensor through `sampling_params.extra_args["action_video"]` works in `forward()`, which resolves that fallback before validating the input. The step path only reads `additional_information["preprocessed_video"]`, so the same request now raises here despite providing a valid conditioning video. Please share action-input resolution with `forward()`, including the missing height/width defaults, and add a parity test for this input form.

hsliuustc0106 (commented):
Two issues in the step-execution path. Reviewed statically at 0eb512eb690b4bc24aba26a6b59acc189559e96c; GPU parity was not independently verified.

vllm_omni/diffusion/models/cosmos3/pipeline_cosmos3.py
86MaxCao: Addressed in https://github.com/86MaxCao/vllm-omni/commit/21600c11be1157230c594cce8c7ad36f87078b02.

Added an optional `SupportsStepRequestCleanup` protocol (`release_step_state` hook, same pattern as `SupportsChunkedVAEDecode`). The runner now routes every terminal path — success, per-request failure, interrupt, and scheduler-side abort — through a single `_retire_step_state()` helper that pops the `StepRequestState` and invokes the hook when implemented.

Cosmos3 implements the hook via an idempotent `_release_step_resources()` shared with `post_decode()`'s success path: it drops the session from the manager **and** clears the transformer's `cached_kv` references (which alias the session tensors), so the K/V is actually freed rather than waiting for eviction.

Tests with a real `SessionStateManager` enabled: `test_release_step_state_drops_session_on_abort` (mid-denoising abort, no post_decode), `test_release_step_state_drops_session_on_step_failure`, `test_post_decode_success_releases_session_with_manager_enabled`.

Since this touches the shared runner, happy to split the protocol/runner change into a separate PR if you prefer.

vllm_omni/diffusion/models/cosmos3/pipeline_cosmos3.py
86MaxCao: Addressed in https://github.com/86MaxCao/vllm-omni/commit/21600c11be1157230c594cce8c7ad36f87078b02.

Extracted `forward()`'s inline `extra_args["action_video"]` resolution into a shared `_resolve_action_video_input()` helper (fallback → `[3,T,H,W]`→`[1,3,T,H,W]` normalization → height/width defaults derived from the tensor). The step path now calls the same helper in `_parse_request_from_state()`, and its height/width defaults prefer the conditioning tensor's dimensions (`sp.height or action_video_height or <default>`). The two error messages in `prepare_encode()` now also list `extra_args['action_video']` as a valid input form.

Parity test: `test_step_execution_accepts_extra_args_action_video_parity_with_forward` runs the same inverse-dynamics request through both `forward()` and step execution and asserts `_prepare_latents_action_video()` receives identical `(video, mode, height, width, num_frames)`.

vllm_omni/diffusion/models/cosmos3/pipeline_cosmos3.py
```diff
@@ -1226,6 +1240,826 @@ def _remap_ckpt_key(key: str) -> str | None:
...
+            if guidance_interval is None:
+                return True
+            t_scalar = float(t_val.item()) if torch.is_tensor(t_val) else float(t_val)
+            lo, hi = guidance_interval
+            return lo <= t_scalar <= hi
+
+        if cfg_parallel:
+            # Each CFG-parallel rank runs exactly one branch.
+            cfg_rank_is_negative = get_classifier_free_guidance_rank() != 0
+            step_scale = guidance_scale if _cfg_active_at(t) else 1.0
```
lishunyang12: Extract shared CFG helpers for diffuse() and _step_denoise_single().

86MaxCao: Addressed in https://github.com/vllm-project/vllm-omni/pull/6316/commits/e7a9e6dc526aa2d15a66a8bb5ce85f464c093715.

Extracted `_cfg_active_at()` and `_predict_noise_one_step()` shared by both `diffuse()` and `_step_denoise_single()`, collapsing diffuse()'s three denoising loops into a single loop. Added regression tests covering the step-path behavior (cache-context routing, per-request scheduler metadata, and terminal-state cleanup).

PR #6318, opened 2026-08-18
[Bugfix] Reject instructions changes on locked native duplex sessions

### LINKED ISSUE #6132: [Bug]: MiniCPM-o 4.5 realtime duplex ignores instruction updates after session start
﻿## Summary

In a MiniCPM-o 4.5 native Realtime duplex session, a second `session.update`
can change `session.instructions` and the server returns `session.updated`
containing the new value. However, once Stage 0 has created its per-session
model state, the new instruction is not inserted into the existing system
context/KV state. Later PCM/video appends therefore continue to follow the
original instruction.

This is specifically about the Realtime API field:

```json
{
  "type": "session.update",
  "session": {
    "instructions": "..."
  }
}
```

For MiniCPM-o native duplex, this is a **session-level system instruction**.
It is not a normal user message, a `conversation.item.create` prompt, a
response-scoped `response.create.instructions` override, or a UI-only label.
The MiniCPM adapter renders it under `<|im_start|>system` when preparing the
Stage 0 context:
[`policy.py`](https://github.com/vllm-project/vllm-omni/blob/877beeaa4957e8caab0daad31930cb274cb601cb/vllm_omni/experimental/fullduplex/minicpmo45/policy.py#L46-L57).

## Environment

```text
GPU: 4 x NVIDIA L20X
NVIDIA driver: 570.133.20
PyTorch: 2.11.0+cu130
Model: openbmb/MiniCPM-o-4_5
Endpoint: /v1/realtime?duplex=1
Native mode: minicpmo45_native_duplex=1
```

Observed on:

```text
vLLM-Omni: e0d5f1ee6cdf973cf7371a2a555ee35a50c6c771
tag description: v0.26.0-82-ge0d5f1ee
```

The same relevant control/data-plane behavior is present on:

```text
877beeaa4957e8caab0daad31930cb274cb601cb
```

## Standalone A/B reproducer

The script runs two cases with the same reference voice, book image, and PCM
stream:

1. **same_session**: initialize Stage 0 with an instruction that requires
   `OLD_CONTEXT`, stream one model unit to create the per-session state, update
   the open session to require `UPDATE_APPLIED`, wait for the matching
   `session.updated`, then send the book image.
2. **fresh_session**: open a new session with `UPDATE_APPLIED` as the initial
   instruction, then send the same book image.

This control run distinguishes an ignored mid-session update from a model/image
capability problem.

### Requirements

```bash
pip install websockets
```

Use the checkpoint reference voice at
`<MiniCPM-o-4_5>/assets/HT_ref_audio.wav` and a JPEG/PNG that clearly contains
a book.

### Run

```bash
python repro_minicpmo45_session_update.py \
  --url ws://127.0.0.1:8000/v1/realtime \
  --ref-audio /path/to/MiniCPM-o-4_5/assets/HT_ref_audio.wav \
  --book-image /path/to/book.jpg
```

The script writes `summary.json`, both event streams, and both output WAV files
under `session_update_repro_output/`.

### Failure signature

When the bug reproduces:

```json
{
  "verdict": "REPRODUCED",
  "same_session": {
    "transcript": "OLD_CONTEXT"
  },
  "fresh_session": {
    "transcript": "UPDATE_APPLIED"
  },
  "same_session_update_ack_instructions": "When you see a book, immediately say exactly \"UPDATE_APPLIED\" and nothing else."
}
```

The important comparison is that the same-session arm acknowledges
`UPDATE_APPLIED` but behaves according to `OLD_CONTEXT`, while the fresh-session
control follows `UPDATE_APPLIED`.

<details>
<summary>repro_minicpmo45_session_update.py</summary>

```python
"""Reproduce MiniCPM-o 4.5 native-duplex session.update instruction loss.

The script runs two cases with the same book image:

1. same_session: establish Stage 0 with OLD_CONTEXT, then update the open
   session to UPDATE_APPLIED before sending the image.
2. fresh_session: open a new session with UPDATE_APPLIED as its initial
   instruction, then send the same image.

A typical reproduction says OLD_CONTEXT in case 1 and UPDATE_APPLIED in case 2,
even though case 1 received session.updated containing the new instruction.
"""

import argparse
import asyncio
import base64
import json
import re
import time
import wave
from pathlib import Path
from urllib.parse import parse_qsl, urlencode, urlsplit, urlunsplit

import websockets
from websockets.exceptions import ConnectionClosed


SAMPLE_RATE = 16_000
BYTES_PER_SAMPLE = 2
OLD_MARKER = "OLD_CONTEXT"
NEW_MARKER = "UPDATE_APPLIED"
OLD_INSTRUCTION = (
    f'When you see a book, immediately say exactly "{OLD_MARKER}" and nothing else.'
)
NEW_INSTRUCTION = (
    f'When you see a book, immediately say exactly "{NEW_MARKER}" and nothing else.'
)


def realtime_url(url: str, model: str) -> str:
    parts = urlsplit(url)
    query = dict(parse_qsl(parts.query, keep_blank_values=True))
    query.update(
        {
            "duplex": "1",
            "model": model,
            "minicpmo45_native_duplex": "1",
            "autostart": "0",
        }
    )
    return urlunsplit((parts.scheme, parts.netloc, parts.path, urlencode(query), parts.fragment))


def data_url(path: Path, mime: str) -> str:
    return f"data:{mime};base64," + base64.b64encode(path.read_bytes()).decode("ascii")


def normalized(text: str) -> str:
    return re.sub(r"[^A-Z0-9]+", "", text.upper())


class EventLog:
    def __init__(self, label: str) -> None:
        self.label = label
        self.events: list[dict[str, object]] = []

    def add(self, event: dict[str, object]) -> None:
        event = dict(event)
        event["_client_received_at"] = time.time()
        self.events.append(event)
        event_type = event.get("type")
        if event_type in {
            "session.created",
            "session.updated",
            "response.created",
            "response.listen",
            "response.done",
            "input_audio_buffer.committed",
            "error",
        }:
            print(f"[{self.label}] {event_type}: {json.dumps(event, ensure_ascii=False)}")
        elif event_type in {"response.audio_transcript.delta", "response.output_text.delta"}:
            delta = event.get("delta", "")
            if delta:
                print(f"[{self.label}] transcript: {delta}", flush=True)

    async def wait_for(self, predicate, *, after: int, timeout: float, label: str) -> dict[str, object]:
        deadline = time.monotonic() + timeout
        while time.monotonic() < deadline:
            for event in self.events[after:]:
                if predicate(event):
                    return event
                if event.get("type") == "error":
                    raise RuntimeError(f"server error while waiting for {label}: {event}")
            await asyncio.sleep(0.02)
        raise TimeoutError(f"timed out waiting for {label}")

    def transcript_after(self, cursor: int) -> str:
        return "".join(
            str(event.get("delta", ""))
            for event in self.events[cursor:]
            if event.get("type")
            in {"response.audio_transcript.delta", "response.output_text.delta"}
        )

    def audio_after(self, cursor: int) -> bytes:
        chunks: list[bytes] = []
        for event in self.events[cursor:]:
            if event.get("type") != "response.audio.delta":
                continue
            encoded = event.get("delta") or event.get("audio")
            if isinstance(encoded, str):
                chunks.append(base64.b64decode(encoded))
        return b"".join(chunks)


async def reader(ws, events: EventLog) -> None:
    try:
        async for raw in ws:
            if not isinstance(raw, str):
                continue
            event = json.loads(raw)
            if isinstance(event, dict):
                events.add(event)
    except ConnectionClosed:
        pass


async def send_session_update(
    ws,
    events: EventLog,
    *,
    instruction: str,
    model: str | None = None,
    ref_audio: str | None = None,
    timeout: float,
) -> dict[str, object]:
    session: dict[str, object] = {"instructions": instruction}
    if model is not None:
        session.update(
            {
                "model": model,
                "modalities": ["audio", "text"],
                "input_audio_format": "pcm16",
                "output_audio_format": "pcm16",
                "turn_detection": None,
                "overlap_policy": "listen_only",
                "playback_commit_policy": "ack_only",
                "extra_body": {
                    "auto_response": True,
                    "minicpmo45_native_duplex": True,
                    "force_listen_count": 0,
                },
            }
        )
    if ref_audio is not None:
        session["ref_audio"] = ref_audio

    cursor = len(events.events)
    await ws.send(json.dumps({"type": "session.update", "session": session}))
    return await events.wait_for(
        lambda event: event.get("type") == "session.updated"
        and isinstance(event.get("session"), dict)
        and event["session"].get("instructions") == instruction,
        after=cursor,
        timeout=timeout,
        label=f"session.updated for {instruction!r}",
    )


async def send_silence(
    ws,
    *,
    duration_ms: int,
    chunk_ms: int,
    start_audio_ms: int,
    frame_b64: str | None,
    realtime: bool,
) -> int:
    audio_end_ms = start_audio_ms
    sent_ms = 0
    next_frame_ms = 0
    while sent_ms < duration_ms:
        this_ms = min(chunk_ms, duration_ms - sent_ms)
        samples = SAMPLE_RATE * this_ms // 1000
        chunk = b"\x00\x00" * samples
        audio_end_ms += this_ms
        sent_ms += this_ms
        event: dict[str, object] = {
            "type": "input_audio_buffer.append",
            "audio": base64.b64encode(chunk).decode("ascii"),
            "input_audio_format": "pcm16",
            "sample_rate_hz": SAMPLE_RATE,
            "duration_ms": this_ms,
            "audio_end_ms": audio_end_ms,
        }
        if frame_b64 is not None and sent_ms > next_frame_ms:
            event["video_frames"] = [frame_b64]
            next_frame_ms += 1000
        await ws.send(json.dumps(event))
        if realtime:
            await asyncio.sleep(this_ms / 1000)
    return audio_end_ms


async def run_case(
    *,
    label: str,
    url: str,
    model: str,
    ref_audio: str,
    book_frame: str,
    update_existing_session: bool,
    warmup_ms: int,
    trigger_ms: int,
    chunk_ms: int,
    settle_seconds: float,
    timeout: float,
    realtime: bool,
) -> tuple[EventLog, int, dict[str, object]]:
    events = EventLog(label)
    async with websockets.connect(url, max_size=64 * 1024 * 1024) as ws:
        reader_task = asyncio.create_task(reader(ws, events))
        try:
            initial_instruction = OLD_INSTRUCTION if update_existing_session else NEW_INSTRUCTION
            await send_session_update(
                ws,
                events,
                instruction=initial_instruction,
                model=model,
                ref_audio=ref_audio,
                timeout=timeout,
            )

            audio_end_ms = 0
            if update_existing_session:
                # One full model unit is enough to create the per-session Stage 0 state.
                audio_end_ms = await send_silence(
                    ws,
                    duration_ms=warmup_ms,
                    chunk_ms=chunk_ms,
                    start_audio_ms=audio_end_ms,
                    frame_b64=None,
                    realtime=realtime,
                )
                await asyncio.sleep(settle_seconds)
                update_ack = await send_session_update(
                    ws,
                    events,
                    instruction=NEW_INSTRUCTION,
                    timeout=timeout,
                )
            else:
                update_ack = next(
                    event
                    for event in reversed(events.events)
                    if event.get("type") == "session.updated"
                )

            result_cursor = len(events.events)
            await send_silence(
                ws,
                duration_ms=trigger_ms,
                chunk_ms=chunk_ms,
                start_audio_ms=audio_end_ms,
                frame_b64=book_frame,
                realtime=realtime,
            )
            await ws.send(json.dumps({"type": "input_audio_buffer.commit", "final": True}))
            await asyncio.sleep(settle_seconds)
            await ws.send(json.dumps({"type": "session.close"}))
            try:
                await events.wait_for(
                    lambda event: event.get("type") == "session.closed",
                    after=result_cursor,
                    timeout=timeout,
                    label="session.closed",
                )
            except TimeoutError:
                pass
        finally:
            reader_task.cancel()
            await asyncio.gather(reader_task, return_exceptions=True)
    return events, result_cursor, update_ack


def write_artifacts(output_dir: Path, label: str, events: EventLog, cursor: int) -> dict[str, object]:
    output_dir.mkdir(parents=True, exist_ok=True)
    (output_dir / f"{label}.events.jsonl").write_text(
        "".join(json.dumps(event, ensure_ascii=False) + "\n" for event in events.events),
        encoding="utf-8",
    )
    transcript = events.transcript_after(cursor)
    audio = events.audio_after(cursor)
    if audio:
        with wave.open(str(output_dir / f"{label}.wav"), "wb") as wav_file:
            wav_file.setnchannels(1)
            wav_file.setsampwidth(BYTES_PER_SAMPLE)
            wav_file.setframerate(24_000)
            wav_file.writeframes(audio)
    return {
        "transcript": transcript,
        "audio_bytes": len(audio),
        "event_count": len(events.events),
    }


async def main_async(args: argparse.Namespace) -> int:
    url = realtime_url(args.url, args.model)
    ref_audio = data_url(Path(args.ref_audio), "audio/wav")
    book_frame = base64.b64encode(Path(args.book_image).read_bytes()).decode("ascii")

    same_events, same_cursor, same_ack = await run_case(
        label="same_session",
        url=url,
        model=args.model,
        ref_audio=ref_audio,
        book_frame=book_frame,
        update_existing_session=True,
        warmup_ms=args.warmup_ms,
        trigger_ms=args.trigger_ms,
        chunk_ms=args.chunk_ms,
        settle_seconds=args.settle_seconds,
        timeout=args.timeout,
        realtime=not args.no_realtime,
    )
    fresh_events, fresh_cursor, fresh_ack = await run_case(
        label="fresh_session",
        url=url,
        model=args.model,
        ref_audio=ref_audio,
        book_frame=book_frame,
        update_existing_session=False,
        warmup_ms=args.warmup_ms,
        trigger_ms=args.trigger_ms,
        chunk_ms=args.chunk_ms,
        settle_seconds=args.settle_seconds,
        timeout=args.timeout,
        realtime=not args.no_realtime,
    )

    output_dir = Path(args.output_dir)
    same = write_artifacts(output_dir, "same_session", same_events, same_cursor)
    fresh = write_artifacts(output_dir, "fresh_session", fresh_events, fresh_cursor)
    same_text = normalized(str(same["transcript"]))
    fresh_text = normalized(str(fresh["transcript"]))
    if OLD_MARKER in same_text and NEW_MARKER in fresh_text and NEW_MARKER not in same_text:
        verdict = "REPRODUCED"
        exit_code = 0
    elif NEW_MARKER in same_text and NEW_MARKER in fresh_text:
        verdict = "NOT_REPRODUCED"
        exit_code = 1
    else:
        verdict = "INCONCLUSIVE_CHECK_WAV_AND_EVENTS"
        exit_code = 2

    summary = {
        "verdict": verdict,
        "url": url,
        "same_session": same,
        "fresh_session": fresh,
        "same_session_update_ack_instructions": same_ack.get("session", {}).get("instructions"),
        "fresh_session_update_ack_instructions": fresh_ack.get("session", {}).get("instructions"),
    }
    output_dir.mkdir(parents=True, exist_ok=True)
    (output_dir / "summary.json").write_text(
        json.dumps(summary, ensure_ascii=False, indent=2), encoding="utf-8"
    )
    print(json.dumps(summary, ensure_ascii=False, indent=2))
    return exit_code


def parse_args() -> argparse.Namespace:
    parser = argparse.ArgumentParser(description=__doc__)
    parser.add_argument("--url", default="ws://127.0.0.1:8000/v1/realtime")
    parser.add_argument("--model", default="openbmb/MiniCPM-o-4_5")
    parser.add_argument("--ref-audio", required=True, help="MiniCPM-o reference voice WAV")
    parser.add_argument("--book-image", required=True, help="JPEG/PNG visibly containing a book")
    parser.add_argument("--output-dir", default="session_update_repro_output")
    parser.add_argument("--warmup-ms", type=int, default=1200)
    parser.add_argument("--trigger-ms", type=int, default=6000)
    parser.add_argument("--chunk-ms", type=int, default=200)
    parser.add_argument("--settle-seconds", type=float, default=6.0)
    parser.add_argument("--timeout", type=float, default=60.0)
    parser.add_argument("--no-realtime", action="store_true")
    return parser.parse_args()


def main() -> None:
    raise SystemExit(asyncio.run(main_async(parse_args())))


if __name__ == "__main__":
    main()
```

</details>

## Expected behavior

After the second `session.update` is acknowledged, the updated session-level
system instruction should affect later native appends in that same Realtime
session.

If changing the system instruction after Stage 0 initialization is
intentionally unsupported, the update should fail with an explicit error. It
should not return a successful `session.updated` containing a value that the
model will not use.

## Actual behavior

- The second update returns `session.updated`.
- The public session object contains the new `instructions`.
- Later PCM/video appends continue normally.
- Model behavior still follows the original system instruction.
- Reopening the session with the new instruction as the initial value makes the
  model follow it.

### 1. Serving accepts and stores the new value

[`serving.py:1328-1331`](https://github.com/vllm-project/vllm-omni/blob/877beeaa4957e8caab0daad31930cb274cb601cb/vllm_omni/experimental/fullduplex/openai/serving.py#L1328-L1331):

```python
if isinstance(payload.get("instructions"), str):
    session.config.instructions = str(payload["instructions"])
elif "instructions" in payload and payload.get("instructions") is None:
    session.config.instructions = None
```

### 2. The runner signals the runtime and emits a successful acknowledgement

[`session_runner.py:1040-1094`](https://github.com/vllm-project/vllm-omni/blob/877beeaa4957e8caab0daad31930cb274cb601cb/vllm_omni/experimental/fullduplex/openai/session_runner.py#L1040-L1094)
builds the candidate configuration, calls
`_signal_runtime_session(..., session_config=candidate_config.as_dict(), ...)`,
replaces the public configuration, and emits:

```python
{
    "type": "session.updated",
    "session": session.as_public_dict(),
}
```

### 3. Later appends carry the updated configuration

[`runtime.py:138-158`](https://github.com/vllm-project/vllm-omni/blob/877beeaa4957e8caab0daad31930cb274cb601cb/vllm_omni/experimental/fullduplex/minicpmo45/runtime.py#L138-L158)
includes both the unchanged session incarnation and the latest configuration in
the data-plane prompt:

```python
"incarnation": fence.incarnation,
...
"session_config": dict(session_config),
```

Therefore, the updated value is not lost at the WebSocket/UI boundary.

### 4. Stage 0 consumes `session_config` only while creating a new state

[`minicpmo_4_5_omni.py:338-354`](https://github.com/vllm-project/vllm-omni/blob/877beeaa4957e8caab0daad31930cb274cb601cb/vllm_omni/model_executor/models/minicpmo_4_5/minicpmo_4_5_omni.py#L338-L354):

```python
session_key = (session_id, incarnation)
state = helper.sessions.get(session_key)
if state is None:
    state = _MiniCPMO45Stage0SessionState(session_id=session_id)
    helper.sessions[session_key] = state
    session_config = duplex.get("session_config")
    ...
    helper._prepare_session_context(
        state,
        session_config,
        runtime_config=runtime_config,
    )
```

In this file, the only reads of `session_config` for this path are inside
`if state is None`. A normal `session.update` keeps the same `session_id` and
`incarnation`, so later appends find the existing state and skip
`_prepare_session_context`.

### 5. `_prepare_session_context` is where the system instruction enters the model context

[`stage0.py:112-134`](https://github.com/vllm-project/vllm-omni/blob/877beeaa4957e8caab0daad31930cb274cb601cb/vllm_omni/experimental/fullduplex/minicpmo45/stage0.py#L112-L134):

```python
prefix, suffix = MiniCPMO45DuplexPolicy.session_context_texts(
    session_config.get("instructions"),
    ref_audio is not None,
    (runtime_config or {}).get("initial_user_text"),
)
...
state.context_embeds.append(...)
state.context_token_ids.append(...)
```

The context is prepared only for the initial state. No update path rewrites the
existing `context_embeds`, `context_token_ids`, or already established KV state.

## Root cause

The control plane and public session configuration support updating
`session.instructions`, and the updated configuration reaches later data-plane
appends. The MiniCPM-o Stage 0 model path, however, only applies that field
during first-state initialization. The acknowledgement therefore promises a
configuration change that is not reflected in the established model context.

## Required contract

Either of the following would be consistent:

1. safely apply the updated session-level system instruction to the existing
   MiniCPM-o native-duplex model state, with defined context/KV semantics; or
2. reject `session.instructions` changes after Stage 0 initialization with an
   explicit unsupported-update error.

The current behavior—successful `session.updated` plus continued use of the old
system context—is misleading.

anurag12-webster:
## What broke

MiniCPM-o native duplex sessions accept a `session.update` changing
`instructions` mid-session and always return a successful
`session.updated`, but Stage0 only builds the instructions into the
model's context on the session's first audio append. Later updates are
silently dropped - the client is told it worked, the model never sees it.

## Root cause

`_prepare_session_context` (stage0.py) and the Stage0 session-state
creation in `minicpmo_4_5_omni.py` only consume `session_config` when the
per-session state doesn't exist yet. A normal `session.update` doesn't
touch `session_id`/`incarnation`, so later appends reuse the existing
state and never re-read the new instructions.

## Fix

`session.update` now rejects a changed `instructions` value with an
explicit `instructions_update_unsupported` error once the session has
locked its native context (first audio append), instead of returning a
misleading success. Same fix applied to PersonaPlex, which shares the
identical code path and would otherwise crash on the new field.

## Testing

- tests/entrypoints/openai_api/test_duplex_handler.py - 200 passed
  (2 new tests added)
- test_runtime_adapter_boundary.py + PersonaPlex adapter/session tests
  - 24 passed

Actually applying the update instead of rejecting it needs GPU-validated
changes to Stage0's context building - planned as a follow-up.

Fixes #6132


anurag12-webster:
@vllm-omni-review-bot

anurag12-webster:
Self-review:

- Confirmed the bug: `session.update` with changed `instructions` returned
  `session.updated` but the model kept the old instructions, since Stage0
  only reads `instructions` when the session state is first created.
- Fix: reject the update explicitly (`instructions_update_unsupported`)
  once the session's native context is locked, instead of a fake success.
  Also fixed PersonaPlex, which shares the same code path.
- Tested: 2 new tests + full existing suite (200 + 24 tests) pass. Verified
  on GPU.
- Not in scope: actually applying the update live is a separate follow-up.

Gaohan123:
@TaffyOfficial PTAL

vllm_omni/experimental/fullduplex/openai/session_runner.py
```diff
@@ -1381,6 +1381,7 @@ async def _run() -> None:
                             continue
                         session.mark_user_input_activity()
                         native.input_since_commit = True
+                        native.native_context_locked = True
```
Gaohan123: The context is marked as locked before an append has actually reached and initialized Stage0. `prepare_append()` can return `None` when less than one model chunk has been buffered; Realtime turn mode can also defer all appends until commit, and backpressure or validation may reject the append before it reaches the runtime. In all of these cases, a subsequent instructions update is rejected even though no native context exists yet.

Could we set this flag only after `_append_runtime_input()` succeeds (for example, in the `if append_ok:` branch)? Since `session.update` already waits for `native_append_tail`, that would preserve ordering while making the flag accurately represent Stage0 initialization. Please also add a test covering: append less than one chunk, update instructions, then commit.

vllm_omni/experimental/fullduplex/openai/serving.py
```diff
@@ -1156,6 +1156,16 @@ def _runtime_session_candidate_update_error(
     ) -> dict[str, object] | None:
         if not self._uses_native_input_append(session):
             return None
+        if (
+            candidate_config.instructions != session.config.instructions
+            and self._runtime_session_state(session).native_context_locked
```
Gaohan123, line 1161: This condition still allows PersonaPlex instructions to change before the first append, but that update is not safe with the current runtime adapter. `PersonaPlexServingRuntimeAdapter.runtime_config_for_update()` updates `personaplex_persona` without recomputing `personaplex_prefill_slots`. The first append therefore reserves scheduler slots using the old persona length while Stage0 builds embeddings from the new persona; if the new persona is longer, Stage0 raises `PersonaPlex scheduler prompt reservation mismatch`.

Until the prefill slot count can be recomputed transactionally during `session.update`, could we reject PersonaPlex instructions changes unconditionally, or otherwise rebuild `personaplex_prefill_slots` before acknowledging the update? Please add a PersonaPlex test that changes instructions before the first append and uses a different token length.

TaffyOfficial:
@vllm-omni-review-bot

TaffyOfficial:
The direction looks good, but two issues remain:
1. In vllm_omni/experimental/fullduplex/openai/session_runner.py:1384, move native.native_context_locked = True into start_native_append()’s if append_ok: branch around line 359. session.update already waits for native_append_tail, so no PENDING state is needed.
2. In vllm_omni/experimental/fullduplex/personaplex/serving_adapter.py:188-197, runtime_config_for_update() changes personaplex_persona without recomputing personaplex_prefill_slots. This can trigger the reservation mismatch in personaplex/stage0.py:275-279. Please recompute the slots transactionally or reject PersonaPlex instruction updates.
Also, tests/entrypoints/openai_api/test_duplex_handler.py:1030 should explicitly send the unchanged instructions value instead of only updating temperature.

anurag12-webster:
Fixed both:

1. Lock flag now only sets after a real append succeeds (`if append_ok:`), not on receipt. Added a test for append < 1 chunk → update → commit.
2. PersonaPlex now just rejects any persona change outright, since recomputing the prefill slots mid-session isn't safe. Added a test with a much longer persona.

Also fixed the test nit about resending the same instructions explicitly.

vllm_omni/experimental/fullduplex/openai/session_runner.py
anurag12-webster: Fixed

vllm_omni/experimental/fullduplex/openai/serving.py
anurag12-webster: Fixed

anurag12-webster:
@Gaohan123 @TaffyOfficial PTAL

tzhouam:
@vllm-omni-review-bot

anurag12-webster:
Fixed the flag bypass — folded `minicpmo45_native_duplex` into PersonaPlex's
existing `_PRIVATE_RUNTIME_CONFIG_KEYS` instead of a separate check, so it
uses the same protocol-validated mechanism already in place rather than a
new one-off.

Also found and fixed the same bug on `voice` while in there — it has the
identical prefill-slot reservation risk as persona, just wasn't flagged.
Rejected the same way.

Both covered by new tests, verified RED (fails without the fix) then GREEN
on GPU.

anurag12-webster:
@Gaohan123 @TaffyOfficial @tzhouam ptal thanks!

TaffyOfficial:
1. vllm_omni/experimental/fullduplex/minicpmo45/adapter.py:63-85
Pre‑lock instruction updates still keep the stale duplex_first_append_context_tokens computed at session creation. Stage0 builds the first context from the updated instructions, while the scheduler reserves slots using the old token count. Please recompute this value transactionally, or reject MiniCPM instruction changes until that is supported. The current FakeEngine test does not cover this mismatch.


2. vllm_omni/experimental/fullduplex/openai/serving.py:1181‑1185, 1387‑1388
After updating instructions, duplex_first_append_context_tokens is not recalculated. Stage0 consumes the new instructions, yet the scheduler still reserves capacity based on the old length, which may trigger padding or truncation.
minicpmo45_native_duplex remains mutable through session.update. Changing it from true to false disables _uses_native_input_append(), so subsequent instruction updates bypass both the context lock and runtime_config_for_update(). This results in spurious session.updated. Please make this mode immutable after session creation and add a true‑>false‑>instructions change regression test.

TaffyOfficial:
@vllm-omni-review-bot

TaffyOfficial:
@vllm-omni-review-bot

anurag12-webster:
Fixed both:

1. MiniCPM-o instructions are now rejected unconditionally, same as
   PersonaPlex, since the first-append token count is never recomputed.
2. minicpmo45_native_duplex is now immutable after session creation.


@TaffyOfficial @Gaohan123 i have added the necessary fixes PTAl

TaffyOfficial:
> Fixed both:
> 
> 1. MiniCPM-o instructions are now rejected unconditionally, same as
>    PersonaPlex, since the first-append token count is never recomputed.
> 2. minicpmo45_native_duplex is now immutable after session creation.
> 
> @TaffyOfficial @Gaohan123 i have added the necessary fixes PTAl

I ran the full E2E on GPU1 with the real PersonaPlex model. Concurrent sessions, capacity limits, session reuse, and audio output all worked correctly.

One issue remains: changing the voice after session creation should report that PersonaPlex does not support changing voices mid-session. Instead, it currently reports that reference audio is missing because the request incorrectly goes through a MiniCPM-specific check.

Please limit the ref_audio_required check to MiniCPM and add a PersonaPlex realtime audio-session test for changing the voice after session creation.


10:17

anurag12-webster:
Fixed, the ref_audio check now only applies to MiniCPM sessions.
PersonaPlex voice changes correctly hit its own voice_update_unsupported
error now. Added a test for it.

anurag12-webster:
@TaffyOfficial take a look now,i have now added a check along with a test to the Personpalex for changing the voice after session creation.

TaffyOfficial:
> @TaffyOfficial现在可以看到，我已经为 Personpalex 添加了一个检查以及一个测试，用于在会话创建后更改语音。

LGTM

TaffyOfficial:
@vllm-omni-review-bot

TaffyOfficial:
@Gaohan123

vllm_omni/experimental/fullduplex/minicpmo45/adapter.py
```diff
@@ -66,7 +66,11 @@ def runtime_config_for_update(
         current: object,
     ) -> dict[str, object]:
         runtime_config = deepcopy(dict(current)) if isinstance(current, dict) else {}
-        runtime_config["instructions"] = config.instructions
+        if config.instructions != runtime_config.get("instructions"):
```
Gaohan123: It seems there is repetitive logic in both minicpm and personalplex. Please merge them into general logic

anurag12-webster: making it into a shared helper

anurag12-webster:
Merged the duplicate logic into one helper.

anurag12-webster:
@Gaohan123 ptal

TaffyOfficial:
@hsliuustc0106

Gaohan123 (approved):
LGTM. Thanks

anurag12-webster:
Thanks @TaffyOfficial @Gaohan123

PR #6325, opened 2026-08-18
Cosmos3 logging and prompt improvements

MaciejBalaNV:
## Purpose
This PR improves consistency of resolution and aspect ratio logging and prompt formatting for the Cosmos3 model. Before this change, there were several issues:

1. `-F "size=HxW"` would be ignored for Cosmos3 Transfer modality.
2. `aspect_ratio` part of the JSON prompt would be incorrectly filled out in case the input video for transfer modality would have a different aspect ratio than the one hardcoded in the initial prompt.
3. I2V and V2V modality would also keep an incorrect `aspect_ratio` in the JSON prompt.

All of these issues are fixed with this PR.

### Unit tests

Added 3 new unit tests that pass.

### Online serving

Start the server normally, e.g. `vllm serve nvidia/Cosmos3-Nano --omni`

After that, let's use [this prompt](https://github.com/NVIDIA/cosmos/blob/main/cookbooks/cosmos3/generator/transfer/assets/blur/prompt.json) and any input video that has a different aspect ratio than the one hardcoded in the prompt, e.g. 16:9 instead of 4:3.

Then use this request:

`curl  -sS -X POST http://localhost:8000/v1/videos/sync     -H "Accept: video/mp4"     -F "model=nvidia/Cosmos3-Nano"     -F "prompt=</mbala/vllm-omni/prompt.json"     -F "size=832x480"     -F "num_frames=121"     -F "fps=30"     -F "num_inference_steps=3"     -F "seed=125"     -F 'extra_params={"blur":{"control_path":"/mbala/vllm-omni/cosmos3_t2v.mp4"},"max_frames":121,"resolution":"720","num_video_frames_per_chunk":121}'     -o cosmos3_transfer_depth.mp4`

Before this change, the prompt would still have `"aspect_ratio": "4,3"` in it. After this change it's correctly set to `"aspect_ratio": "16,9"` with a warning emitted:

`WARNING 08-18 06:51:32 [pipeline_cosmos3.py:2999] Cosmos3 transfer JSON prompt aspect_ratio='4,3' conflicts with the control-selected 16,9 bucket, 1280x720 (WxH); using 16,9 for prompt metadata.`

Additionally, if the `-F "size=832x480"` argument (which isn't really needed for transfer modality, since any size will still be derived from the video input) doesn't match the video input, a warning will be emitted:

`WARNING 08-18 06:51:32 [pipeline_cosmos3.py:2981] Cosmos3 transfer ignores requested size=832x480 (WxH); control input 1280x720 (WxH) selected the 16,9 resolution=720 bucket, 1280x720 (WxH). Transfer supports canonical Cosmos buckets, not arbitrary output sizes.`

**vLLM Version:**
0.26
**vLLM-Omni Commit:**
ToT
## Test Result

Unit tests pass. Online serving described above.


MaciejBalaNV:
@david6666666 @hsliuustc0106 
PTAL

MaciejBalaNV (commented):
Verified the changes with steps described in `Online serving` part of the PR description. Overall every test passes, warnings are emitted when necessary and prompts are correctly formatted.

MaciejBalaNV:
@vllm-omni-review-bot Please review as well

vllm_omni/diffusion/models/cosmos3/pipeline_cosmos3.py
```diff
@@ -2086,13 +2137,39 @@ def _format_and_tokenize_prompts(
...
+        requested_aspect_ratio = self._get_sp_param(sp, "aspect_ratio", None)
+        prompt_aspect_ratio = _json_object_aspect_ratio(prompt)
+        metadata_aspect_ratio = requested_aspect_ratio if requested_aspect_ratio is not None else prompt_aspect_ratio
+        canvas_aspect_ratio = aspect_ratio_override or _aspect_ratio_for_dimensions(width, height)
+        normalized_metadata_ratio = _normalize_aspect_ratio(metadata_aspect_ratio)
+        if (
+            aspect_ratio_override is None
+            and metadata_aspect_ratio is not None
+            and normalized_metadata_ratio != canvas_aspect_ratio
+            and _is_rank_zero()
```
hsliuustc0106: **P1 — the conflict rewrite is rank-gated, not just the warning.** `_is_rank_zero()` is a conjunct of the `if`, so on multi-rank Cosmos3 (TP/SP) rank 0 rewrites `metadata_aspect_ratio` to the canvas ratio while ranks ≥ 1 keep the user/prompt value. That yields rank-divergent JSON prompts and token ids — there is no broadcast of `cond_ids` downstream; each rank tokenizes and embeds its own copy. Token counts stay identical, so sharded compute silently mixes different text conditioning instead of failing. Reachable via exactly the I2V/V2V case this PR targets: a JSON prompt with a stale `aspect_ratio` on TP>1 serving (the transfer path is safe — `aspect_ratio_override` applies on all ranks unguarded).

Fix: apply the rewrite unconditionally and guard only the `logger.warning(...)` with `_is_rank_zero()` — `_warn_transfer_bucket_conflicts` (the early-return at :2956) already does this correctly.

vllm_omni/diffusion/models/cosmos3/pipeline_cosmos3.py
```diff
@@ -235,6 +237,43 @@ def _format_json_object_prompt(
...
+    return f"{width},{height}"
+
+
+def _aspect_ratio_for_dimensions(width: int, height: int) -> str:
+    """Return Cosmos's canonical label for a known bucket, else the exact ratio."""
+    for sizes in VIDEO_RES_SIZE_INFO.values():
+        for aspect_ratio, size in sizes.items():
+            if size == (width, height):
+                return aspect_ratio
+    divisor = math.gcd(width, height)
```
hsliuustc0106: **P2 — non-canonical canvases get gcd labels that can never match the request ratio, so correct metadata is warned and rewritten on every request.** I2V sizing uses the max-area/mod-16 formula (:555), so a 4:3 input image produces a 1104x816 canvas; `_aspect_ratio_for_dimensions` then returns `"23,17"`, which string-compares unequal to a perfectly correct `"4,3"` prompt label → warning + rewrite to the out-of-distribution `"23,17"` on every such request (single-rank too). Near-miss explicit sizes behave the same (e.g. 832x468 → `"208,117"` vs `"16,9"`).

Suggestion: when the canvas is not a canonical bucket, snap the metadata label to the nearest `VIDEO_RES_SIZE_INFO` ratio (same idea as `find_closest_target_size`) instead of the exact gcd ratio, or compare with a small tolerance so mod-16 rounding drift does not fire.

vllm_omni/diffusion/models/cosmos3/pipeline_cosmos3.py
MaciejBalaNV: Fixed, now all ranks overwrite the aspect ratio metadata. Verified with a manual test with online serving on 4 GPUs, as well as a modified unit test.

vllm_omni/diffusion/models/cosmos3/pipeline_cosmos3.py
MaciejBalaNV: Fixed, now all aspect ratios fall into one of the "canonical", accepted aspect ratios.

MaciejBalaNV: Refactored the solution slightly not to duplicate the aspect ratio data.

ywang96 (approved):

PR #6329, opened 2026-08-18
[BugFix][TTS] Move user input from INFO to DEBUG in TTS/audio log lines

dougbtv:
## Summary

- TTS, audio generation, and diffusion chat serving paths log verbatim user input at INFO level, which can expose sensitive information (names, phone numbers, financial details) in centralized log aggregators
- The standard `/v1/chat/completions` and `/v1/completions` endpoints already handle this correctly — `RequestLogger` gates prompt content behind `--enable-log-requests` at DEBUG level and respects `--max-log-len` truncation
- The TTS/audio/diffusion paths bypass `RequestLogger` entirely and unconditionally emit user text at INFO — this PR aligns them with the established pattern

## Changes

For each of the 4 affected log sites across 3 files:

- **INFO** now logs request metadata only (request_id, model type, voice_clone, ref_images, params)
- **User content** moves to `logger.debug`, gated on `self.request_logger` (set when `--enable-log-requests` is passed) and truncated by `--max-log-len`

**Files:**
- `serving_speech.py` — TTS speech request + Diffusion TTS speech request
- `serving_audio_generate.py` — Audio generation request
- `serving_chat.py` — Diffusion chat request
- `test_serving_speech.py` — Updated log format assertion to match new metadata-only INFO line

## Test plan

- [ ] Verify INFO logs no longer contain `text=` or `prompt=` fields
- [ ] Verify DEBUG logs emit user content when `--enable-log-requests` is set
- [ ] Verify `--max-log-len` truncation is respected in DEBUG output
- [ ] Existing test suite passes (`test_serving_speech.py` assertion updated)

### DIFF
--- a/vllm_omni/entrypoints/openai/serving_speech.py
+++ b/vllm_omni/entrypoints/openai/serving_speech.py
@@ -3080,12 +3080,14 @@ class OmniOpenAIServingSpeech(OpenAIServing, AudioMixin):
                 model_type = tts_params.get("task_type", ["unknown"])[0]
             else:
                 model_type = "generic"
-        logger.info(
-            "TTS speech request %s: text=%r, model=%s",
-            request_id,
-            request.input[:50] + "..." if len(request.input) > 50 else request.input,
-            model_type,
-        )
+        logger.info("TTS speech request %s: model=%s", request_id, model_type)
+        _rl = getattr(self, "request_logger", None)
+        if _rl:
+            base_len = len(f"TTS speech request {request_id}: text=")
+            raw_max = getattr(_rl, "max_log_len", None)
+            cap = raw_max if isinstance(raw_max, int) else 200
+            text = request.input[: max(cap - base_len, 0)]
+            logger.debug("TTS speech request %s: text=%r", request_id, text)
 
         # Apply model-specific extra parameters
         if request.extra_params is not None and sampling_params_list:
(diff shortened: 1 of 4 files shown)


vllm_omni/entrypoints/openai/serving_speech.py
```diff
@@ -3080,12 +3080,11 @@ async def _prepare_speech_generation(
...
                 model_type = "generic"
-        logger.info(
-            "TTS speech request %s: text=%r, model=%s",
-            request_id,
-            request.input[:50] + "..." if len(request.input) > 50 else request.input,
-            model_type,
-        )
+        logger.info("TTS speech request %s: model=%s", request_id, model_type)
+        if self.request_logger:
+            max_len = self.request_logger.max_log_len
```
alex-jw-brooks: I think we also need to consider the actual log line, or else this would truncate because it's max len + len("TTS speech request....") right? 

IMO we should also probably have a small upper bound here in case it's the max_log_len is unset, otherwise this could be a really big log if the input is huge, even though it is at debug level

dougbtv:
@alex-jw-brooks Good call — updated in 47c95f73. Now uses `max_log_len or 200` so there's always a cap on the debug text, even when `--max-log-len` isn't set. Simplified the truncation to just `text[:max_len]` since it always applies now.

alex-jw-brooks (approved):

alex-jw-brooks:
@dougbtv can you please fix the attr error and DCO checks?

Issue #6246, opened 2026-08-17
[RFC][Diffusion] Build on #4238: Fast-Ulysses-style SymmMem Transport for Low-Latency Long Video


baonudesifeizhai:
## Motivation

PR #4238 introduces a fused, permute-free Ulysses all-to-all using NCCL symmetric memory and LSA kernels. It removes explicit relayout kernels, but still uses SM resources, requires regular splits, and executes Q/K/V exchanges separately.

Fast-Ulysses demonstrates another path: pitched Copy Engine transfers directly into peer symmetric-memory layouts, with zero SM usage, uneven sequence/head splits, and 1.66–2.17× collective-level speedups over the PyTorch baseline on 8-GPU NVLink systems.

We propose building on #4238 with a native, pluggable Ulysses transport targeting low-latency long-video diffusion.

References:

- #4238
- https://github.com/triple-mu/fast-ulysses

## Proposal

Introduce a common `UlyssesTransport` with:

- `torch_dist`: existing portable fallback
- `symm_lsa`: the fused NCCL SymmMem kernel from #4238
- `symm_ce`: Fast-Ulysses-style Copy Engine transport for large NVLink payloads
- Persistent symmetric-memory workspaces and fixed peer addresses
- Direct producer-to-consumer-final-layout transfers
- Batched Q/K/V exchange with one synchronization round
- Device-side epoch, double buffering, and microbatch slots
- Warmed-shape CUDA Graph capture and `torch.compile` custom-op support
- Deterministic capability checks and fallback

No allocation, rendezvous, JIT compilation, or backend selection should occur in the hot path.

## UAA and Attention Layouts

The transport planner should support:

- Uneven sequence splits
- Uneven head splits without mandatory padding
- MHA: `Hq == Hkv`
- GQA: `1 < Hkv < Hq`
- MQA: `Hkv == 1`
- KV replication when the Ulysses degree exceeds the number of KV heads

The first optimized benchmark target should be MHA long-video models such as Wan2.2, while the interface and correctness path must not assume `Hq == Hkv`.

## Expected Benefit

The current path performs three forward exchanges for Q/K/V and one reverse exchange for O, with additional layout copies.

The proposed path reduces this to:

1. One fused Q/K/V transport
2. One output transport

The communication volume is unchanged. Benefits come from fewer launches and barriers, removal of intermediate relayout copies, persistent buffers, and reduced SM contention.

## Scope and Fallback

[...]

PR #6340, opened 2026-08-18
[Diffusion] Add native SymmMem Fast Ulysses transport

baonudesifeizhai:

## Summary
https://github.com/vllm-project/vllm-omni/issues/6246
This PR adds an opt-in native SymmMem-based Fast Ulysses transport for diffusion sequence parallelism.

- Adds Fast Ulysses transport for Ulysses All-to-All.
- Uses PyTorch SymmetricMemory + P2P copy path to reduce NCCL/SM communication overhead.
- Enabled only when `fast_ulysses=True` and `ulysses_degree > 1`.
- Keeps the existing regular Ulysses path unchanged when Fast Ulysses is disabled.
- Waiting for ZJY's PR to land before final integration / review alignment.

## Benchmark

Cosmos3-Nano, 381 frames, 1280x720, 35 denoise steps, 4 GPUs:

| Config | Diffuse | Pipeline | E2E | Peak Memory |
|---|---:|---:|---:|---:|
| BF16 + Ulysses | 81.18s | 88.44s | 99.11s | 46.78 GiB |
| BF16 + Fast Ulysses | 76.57s | 84.08s | 92.12s | 46.61 GiB |
| FP8 + Ulysses | 76.91s | 84.43s | 92.68s | 34.16 GiB |
| FP8 + Fast Ulysses | 73.12s | 80.64s | 90.49s | 33.99 GiB |

Compared with BF16 + regular Ulysses, FP8 + Fast Ulysses improves:

- Diffuse latency: -9.9%
- Pipeline latency: -8.8%
- E2E latency: -8.7%
- Peak memory: -27.3%

## Notes

Fast Ulysses optimizes the transport layer of Ulysses attention All-to-All. It does not change the attention math or non-communication parts such as MLP, projection, VAE decode, or output transfer.

## Test Plan

for cosmos3 nano 

A BF16 + bf16Ulysses
diffuse  81.18s
pipeline 88.44s
E2E      99.11s
memory   46.78 GiB

B BF16 + Fast Ulysses
diffuse  76.57s
pipeline 84.08s
E2E      92.12s
memory   46.61 GiB

C FP8 + bf16 Ulysses
diffuse  76.91s
pipeline 84.43s
E2E      92.68s
memory   34.16 GiB

D FP8 + Fast Ulysses
diffuse  73.12s
pipeline 80.64s
E2E      90.49s
memory   33.99 GiB

for cosmos3 nano  

**vLLM Version:**

**vLLM-Omni Commit:**

## Test Result


hsliuustc0106:
This PR touches **vllm_omni/diffusion/, vllm_omni/config/, vllm_omni/engine/, examples/offline_inference/, pyproject.toml** (14 files). Based on CODEOWNERS coverage of the changed files, the most-related reviewers appear to be:

> @princepride @xuechendi @yuanheng-zhao

baonudesifeizhai:
### MiniMax-H3 E2E Benchmark

**Test conditions**

- Model: original MiniMax-H3 (non-distilled)
- Hardware: 4 GPUs
- Parallelism: Ulysses sequence parallelism across 4 GPUs
- Workload: batch size 1, 281-frame long-video generation, width 1344
- The prompt, seed, sampling parameters, and request payload were identical
- Baseline: standard Ulysses all-to-all
- SymmMem: fused permute-free Ulysses all-to-all enabled with `--ulysses-a2a-permute`
- One warmup request followed by three measured requests
- Warmup latency was excluded from the steady-state average
- All measured outputs had the same size: 16,122,793 bytes

| Configuration | Warmup | Run 1 | Run 2 | Run 3 | Steady-state average |
| Baseline | 184.512 s | 152.828 s | 152.922 s | 153.041 s | **152.930 s** |
| SymmMem | 203.250 s | 150.689 s | 151.044 s | 150.813 s | **150.849 s** |

**Result**

SymmMem reduced the average steady-state E2E latency from **152.930 s to 150.849 s**, saving **2.082 s per request**. This corresponds to a **1.36% latency reduction**, or a **1.0138× speedup**.

The improvement was consistent across all three measured runs. SymmMem had an additional **18.738 s warmup overhead**, likely caused by initialization/JIT setup, so the optimization is primarily beneficial for long-running services where this one-time cost can be amortized.

baonudesifeizhai:
We evaluated the reviewer-suggested safety changes for symmetric-memory buffer reuse, including stream/event lifetime management, synchronization semantics, and stronger cache identity.

The general event-based implementation regressed steady-state E2E latency from **150.849 s to 151.210 s** (**+0.24%**). A simplified single-stream guard reduced the regression to **151.059 s** (**+0.14%**), with the last two runs averaging **150.802 s**.

Given the small original gain (~1.36%), the additional complexity and potential overhead were not worthwhile, so these changes were reverted.

vllm_omni/diffusion/distributed/a2a_permute.py
```diff
@@ -0,0 +1,148 @@
...
+
+        here = os.path.dirname(__file__)
+        src = os.path.join(here, "csrc", "a2a_permute.cu")
+        sp = sysconfig.get_paths()["purelib"]
+        nccl_dir = os.path.join(sp, "nvidia", "nccl")
+        nccl_inc = os.path.join(nccl_dir, "include")
+        nccl_libs = glob.glob(os.path.join(nccl_dir, "lib", "libnccl.so*"))
+        if not nccl_libs:
+            raise RuntimeError("a2a_permute: could not locate nvidia-nccl libnccl.so")
+        load(
```
hsliuustc0106, line 76: The enabled path fails to JIT-build in a PyTorch 2.11.0+cu130 wheel environment: `CUDAContextLight.h` includes `cusparse.h`, but this call only adds the NCCL include directory, so nvcc exits with `fatal error: cusparse.h: No such file or directory` even though the header is under `site-packages/nvidia/{cu13,cusparse}/include`. Please resolve the complete CUDA include roots (or ship a prebuilt extension) and capability-check or fall back before the first request.

vllm_omni/diffusion/distributed/a2a_permute.py
```diff
@@ -0,0 +1,148 @@
...
+    key = (symm_shape, dtype, group_name)
+    buf = _SYMM_CACHE.get(key)
+    if buf is None:
+        pg = _resolve_process_group(group_name)
+        warm = torch.ones(1, device=device)
+        dist.all_reduce(warm, group=pg)
+        torch.accelerator.synchronize(device)
+        buf = symm_mem.empty(*symm_shape, dtype=dtype, device=device)
+        symm_mem.rendezvous(buf, group_name)
+        _SYMM_CACHE[key] = buf
```
hsliuustc0106: This retains one symmetric-memory buffer for every `(shape, dtype, group)` for the process lifetime. Request resolution or frame-count changes alter the attention shape, so a long-running online server can accumulate large GPU buffers with no bound or teardown. Please use bounded workspaces or eviction and release them during shutdown.

vllm_omni/diffusion/distributed/a2a_permute.py
```diff
@@ -0,0 +1,148 @@
...
+def ulysses_qkv_fwd(x: torch.Tensor, group_name: str, world_size: int) -> torch.Tensor:
+    _ensure_built()
+    p = world_size
+    B, s_local, H, D = x.shape
+    Hl = H // p
+    lc = Hl * D
+    rows = B * s_local
+    symm_in = _get_symm_buffer((rows, p, lc), x.dtype, x.device, group_name)
+    # (B, S_local, H, D) -> (rows, p, lc); H is row-major so column block r = heads [r*Hl:(r+1)*Hl]
+    symm_in.copy_(x.reshape(rows, p, lc))
```
hsliuustc0106, line 175: All calls sharing this key overwrite the same staging buffer, while both `copy_` and the kernel launch are asynchronous. Overlapping streams or replays can overwrite it before the previous exchange finishes and silently corrupt outputs. Please use per-slot buffers or events, or enforce and test a single-stream serialization invariant; the 0.14–0.24% E2E cost reported in the discussion is not a correctness tradeoff.

vllm_omni/config/omni_config.py
```diff
@@ -573,6 +574,7 @@ class OmniStageDiffusionParallelConfig(OmniStageParallelConfig):
     ring_degree: int = Field(default=1, ge=1)
     allgather_degree: int = Field(default=1, ge=1)
     ulysses_mode: str = "strict"
+    ulysses_a2a_permute: bool = False
```
hsliuustc0106, line 577: This changes the exact field set, but `tests/config/test_omni_config.py::test_sub_config_fields_match_structured_scopes` still expects the old set, so that test will fail. Please update the assertion and add flag-propagation coverage for the serve CLI, default-stage, and deploy-config paths.

lishunyang12 (approved):

vllm_omni/diffusion/distributed/a2a_permute.py
baonudesifeizhai: added stream_id = torch.cuda.current_stream(device).cuda_stream temporary  to gate single stream 

will remove later

hsliuustc0106:
## MiniMax-H3 performance A/B

I ran the PR merge tree (`a7dd354bdcd0de6d7088767efe7ccc8af14d42c2`) against its base (`9a88d2d21d4b3a97f7e7a5d2d751328ffda08379`) on 4× NVIDIA L20X (GPU IDs 1–4).

Configuration: native MiniMax-H3 FL2VA, 1344×768, 124 frames, 24 FPS, `num_inference_steps=4`, no LoRA, Ulysses4, text-encoder TP4, VAE tile parallelism 4, CUDNN attention, concurrency 1. The PR run used `--ulysses-a2a-permute`.

| Metric | Base | PR | Delta |
|---|---:|---:|---:|
| Client latency | 18.201 ± 1.461 s | 13.620 ± 0.547 s | **−25.2% / 1.34×** |
| Reported inference time | 18.187 ± 1.460 s | 13.605 ± 0.546 s | **−25.2%** |
| Stage-0 generation | 16.734 ± 1.442 s | 12.066 ± 0.543 s | **−27.9%** |
| Reported diffusion stage | 5.467 ± 0.170 s | 4.989 ± 0.126 s | **−8.8%** |
| Peak memory/GPU | 91,111 MB | 90,852 MB | **−0.28%** |

Values are the mean ± sample standard deviation of 3 measured requests per variant; the exact-shape warmup was excluded. The PR warmup was 56.38 s versus 43.95 s for base because the fused extension is JIT-compiled on first use; this one-time cost is not included above.

All 6 measured requests returned valid 124-frame H.264 video with 32 kHz stereo AAC. Same-seed runs were deterministic within each variant. Cross-variant decoded-video PSNR was 35.3–39.4 dB (SSIM 0.977–0.988), so there is small numerical drift but no output-format regression. This was a performance run with an output sanity check, not a formal perceptual-quality benchmark.

Software: PyTorch 2.11.0+cu130, CUDA 13.0, NCCL 2.28.9, vLLM 0.28.0, diffusers 0.38.0, transformers 5.14.1.

PR #6342, opened 2026-08-19
[Model][Bugfix] Improve LTX audio parity and similarity guards

mglyn:
## Purpose

- Improve LTX-2.3/LTX-2.5 audio parity by running the bandwidth-extension vocoder in FP32.
- Consolidate LTX-2.0/LTX-2.3 similarity coverage around representative two-stage cases.
- Apply one strict similarity profile to all retained LTX-2.0, LTX-2.3, and LTX-2.5 cases.
- Construct the LTX latent upsampler on CPU before loading weights, avoiding unsupported CUDA Long `addmm` during initialization.

## Similarity coverage and vocoder precision

The similarity suite is simplified around representative two-stage coverage:

- Remove the redundant LTX-2.0/LTX-2.3 one-stage cases; the retained Distilled and Full two-stage cases cover their Stage-1 denoising paths at larger representative shapes.
- Retain LTX-2.5 Distilled two-stage and Full/SFT one-stage and two-stage coverage.
- Use one strict threshold profile for all retained LTX similarity cases: SSIM mean/min `>= 0.99`, PSNR `>= 40 dB`, audio relative L2 `<= 0.10`, and audio cosine `>= 0.99`.

LTX-2.3+ uses a bandwidth-extension vocoder with a long convolution stack. Running that stack in BF16 accumulated enough waveform error for the LTX-2.5 Distilled T2V case to reach audio relative L2 `0.331759` and cosine `0.947790`. Following the [official BWE implementation](https://github.com/Lightricks/LTX-2/blob/9377758131b1ffde4b7f766804590a6617bf2ab9/packages/ltx-core/src/ltx_core/model/audio_vae/vocoder.py#L572-L629), the BWE vocoder now runs under FP32 autocast and converts the final waveform back to the input dtype; the base vocoder is unchanged. The same case now reaches relative L2 `0.013621` and cosine `0.999908`.

The latent upsampler is constructed inside an explicit CPU device context before its weights are loaded. This avoids creating its rational-resampler Long tensors on CUDA while preserving the loaded dtype and runtime placement.

## Validation

- 14 non-GPU unit and helper tests passed.
- 14 LTX similarity cases and helper tests collected successfully.
- Targeted pre-commit checks passed.

<details>
<summary><strong>Recent H200 similarity results</strong></summary>

All retained cases below pass the single strict profile: SSIM mean/min `>= 0.99`, PSNR `>= 40 dB`, audio relative L2 `<= 0.10`, and audio cosine `>= 0.99`.

| Version | Pipeline / task | SSIM mean | SSIM min | PSNR | Audio relative L2 | Audio cosine | Result |
|---|---|---:|---:|---:|---:|---:|---|
| LTX-2.0 | Distilled T2V | 0.999568 | 0.999457 | 62.466 dB | 0.000000 | 1.000000 | Pass |
| LTX-2.0 | Full two-stage T2V | 0.999635 | 0.999515 | 63.068 dB | 0.000000 | 1.000000 | Pass |
| LTX-2.3 | Distilled I2V | 0.999476 | 0.999151 | 62.408 dB | 0.033802 | 0.999429 | Pass |
| LTX-2.3 | Full two-stage I2V | 0.999523 | 0.999237 | 62.599 dB | 0.045554 | 0.998973 | Pass |
| LTX-2.5 | Distilled two-stage T2V | 0.997656 | 0.996499 | 46.961 dB | 0.013621 | 0.999908 | Pass |
| LTX-2.5 | Distilled two-stage I2V | 0.997642 | 0.997191 | 47.914 dB | 0.026768 | 0.999643 | Pass |
| LTX-2.5 | Full one-stage T2V | 0.999910 | 0.999842 | 66.769 dB | 0.003933 | 0.999994 | Pass |
| LTX-2.5 | Full one-stage I2V | 0.999798 | 0.999749 | 64.975 dB | 0.004341 | 0.999992 | Pass |
| LTX-2.5 | Full two-stage T2V | 0.999726 | 0.999657 | 59.788 dB | 0.003854 | 0.999994 | Pass |
| LTX-2.5 | Full two-stage I2V | 0.999819 | 0.999766 | 66.230 dB | 0.007642 | 0.999972 | Pass |

</details>

### DIFF
--- a/tests/diffusion/models/ltx2/test_ltx2_vae.py
+++ b/tests/diffusion/models/ltx2/test_ltx2_vae.py
@@ -3,12 +3,69 @@
 
 """Unit tests for LTX-2.3 VAE tiling and distributed decode behavior."""
 
+from contextlib import contextmanager
 from types import SimpleNamespace
 
 import pytest
 import torch
 
 pytestmark = [pytest.mark.core_model, pytest.mark.cpu]
+
+
+def test_ltx_bwe_vocoder_uses_fp32_autocast(monkeypatch):
+    import vllm_omni.diffusion.models.ltx2.ltx2_runtime as ltx_runtime
+
+    class FakeBWEVocoder(torch.nn.Module):
+        def __init__(self):
+            super().__init__()
+            self.weight = torch.nn.Parameter(torch.ones((), dtype=torch.bfloat16))
+            self.bwe_generator = torch.nn.Identity()
+            self.input_dtype = None
+
+        def forward(self, value):
+            self.input_dtype = value.dtype
+            return value.float()
+
+    autocast_calls = []
+
+    @contextmanager
+    def fake_autocast(*, device_type, dtype):
+        autocast_calls.append((device_type, dtype))
+        yield
+
+    monkeypatch.setattr(ltx_runtime.torch, "autocast", fake_autocast)
+    vocoder = FakeBWEVocoder()
+
+    output = ltx_runtime._run_ltx_vocoder(vocoder, torch.ones(1, dtype=torch.bfloat16))
+
+    assert vocoder.input_dtype == torch.float32
+    assert output.dtype == torch.bfloat16
+    assert autocast_calls == [("cpu", torch.float32)]
+
+
+def test_ltx_base_vocoder_keeps_native_dtype(monkeypatch):
+    import vllm_omni.diffusion.models.ltx2.ltx2_runtime as ltx_runtime
+
+    class FakeBaseVocoder(torch.nn.Module):
+        def __init__(self):
+            super().__init__()
+            self.input_dtype = None
+
+        def forward(self, value):
+            self.input_dtype = value.dtype
+            return value
+
+    monkeypatch.setattr(
+        ltx_runtime.torch,
+        "autocast",
+        lambda **_kwargs: (_ for _ in ()).throw(AssertionError("base vocoder must not use autocast")),
+    )
+    vocoder = FakeBaseVocoder()
+
+    output = ltx_runtime._run_ltx_vocoder(vocoder, torch.ones(1, dtype=torch.bfloat16))
+
+    assert vocoder.input_dtype == torch.bfloat16
+    assert output.dtype == torch.bfloat16
 
 
 class TestLTXOutputRank:
--- a/tests/diffusion/models/ltx2/test_ltx2_vae.py
+++ b/tests/diffusion/models/ltx2/test_ltx2_vae.py
@@ -3,44 +3,12 @@
 
 """Unit tests for LTX-2.3 VAE tiling and distributed decode behavior."""
 
-from contextlib import contextmanager
 from types import SimpleNamespace
 
 import pytest
 import torch
 
 pytestmark = [pytest.mark.core_model, pytest.mark.cpu]
-
-
-def test_ltx_bwe_vocoder_uses_fp32_autocast(monkeypatch):
-    import vllm_omni.diffusion.models.ltx2.ltx2_runtime as ltx_runtime
-
-    class FakeBWEVocoder(torch.nn.Module):
-        def __init__(self):
-            super().__init__()
-            self.weight = torch.nn.Parameter(torch.ones((), dtype=torch.bfloat16))
-            self.bwe_generator = torch.nn.Identity()
-            self.input_dtype = None
-
-        def forward(self, value):
-            self.input_dtype = value.dtype
-            return value.float()
-
-    autocast_calls = []
-
-    @contextmanager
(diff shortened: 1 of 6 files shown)


vllm_omni/diffusion/models/ltx2/ltx2_runtime.py
```diff
@@ -60,6 +60,29 @@
...
+            vocoder.float()
+        try:
+            return vocoder(generated_mel.float()).to(input_dtype)
+        finally:
+            if module_dtype != torch.float32:
+                vocoder.to(module_dtype)
+
+    # BF16 errors compound through the BWE model's long convolution stack.
+    # FP32 autocast upcasts each op without materializing a second FP32 model.
+    with torch.autocast(device_type=device_type, dtype=torch.float32):
```
david6666666, line 82: **[P1] FP32 autocast is disabled instead of upcasting the vocoder**

PyTorch's autocast target dtype is limited to FP16/BF16 on CUDA and CPU; passing `torch.float32` warns and disables autocast. The next line therefore feeds an FP32 mel tensor into BF16 vocoder parameters, and `Conv1d` raises a dtype-mismatch error instead of executing in FP32. Please convert the vocoder parameters/buffers to FP32 for this call (then restore them safely), or use another explicit FP32 execution mechanism.

tests/diffusion/models/ltx2/test_ltx2_vae.py
```diff
@@ -11,6 +12,62 @@
...
+            return value.float()
+
+    autocast_calls = []
+
+    @contextmanager
+    def fake_autocast(*, device_type, dtype):
+        autocast_calls.append((device_type, dtype))
+        yield
+
+    monkeypatch.setattr(ltx_runtime.torch, "autocast", fake_autocast)
```
david6666666: **[P1] Exercise the real autocast path in this regression test**

Mocking `torch.autocast` removes the exact runtime behavior this test needs to validate: real PyTorch disables `dtype=torch.float32`. The fake vocoder also never uses its BF16 parameter, so the resulting input/weight dtype mismatch cannot surface. Please keep the real autocast and run a parameterized op such as a BF16 `Conv1d`, asserting that the BWE forward succeeds and actually computes in FP32.

david6666666 (commented):
Found two issues in the new FP32 vocoder path and its unit coverage.

vllm_omni/diffusion/models/ltx2/ltx2_runtime.py
mglyn: This does not seem to be an issue on CUDA. I verified that CUDA autocast accepts FP32 and runs the BF16 vocoder operations in FP32 on H200.

tests/diffusion/models/ltx2/test_ltx2_vae.py
mglyn: Fixed in 331c750f5. The test now uses real CUDA autocast with a BF16 Conv1d and verifies that the convolution executes in FP32 before the output is restored to BF16.

tests/e2e/accuracy/ltx/test_ltx_official_similarity.py
```diff
@@ -299,12 +238,11 @@ class LTXAccuracyCase:
         seed=10,
         stg_block=28,
         enable_layerwise_offload=True,
-        thresholds=STRICT_THRESHOLDS,
         image=I2V_IMAGE,
     ),
 )
 
-CASES = (*LEGACY_CASES, *TWO_STAGE_CASES)
+WEEKLY_CASES = TWO_STAGE_CASES
```
hsliuustc0106, line 245: [P2] Keep at least one `pipeline_kind="one_stage"` case here. `WEEKLY_CASES = TWO_STAGE_CASES` removes all LTX-2/LTX-2.3 `LTX2Pipeline` parity coverage, while those supported entry points use a different recipe from the retained two-stage cases. A one-stage-only regression would now pass this suite.

mglyn: `LTX2Pipeline` is still covered directly by the LTX-2.5 Full one-stage T2V/I2V cases. In addition, the retained LTX-2/LTX-2.3 two-stage recipes are constructed from their corresponding one-stage recipes via `_official_two_stage_recipe`, so Stage 1 exercises the same version-specific guidance, schedule, and denoising path. The retained matrix therefore covers both the one-stage entry path and the LTX-2/LTX-2.3-specific execution without another expensive duplicate E2E case. If the concern is specifically the `("one_stage", version)` recipe/profile dispatch, would a focused unit test be sufficient?

hsliuustc0106 (approved):
lgtm

PR #6173, opened 2026-08-13
[Model][Performance] Optimize MiniMax-H3 strict Ulysses boundaries

mo-ke-ke:
## Purpose

MiniMax-H3's strict Ulysses path previously built the complete packed
`[S, 5376]` embedding and RoPE tensors on every rank, then immediately selected
that rank's `S / world_size` rows. After the transformer blocks it also gathered
the full BF16 hidden state before applying the final 128-channel projection.

This PR keeps those two model boundaries rank-local when, and only when, the
registered `sp_input---local_sp_prepare` hook proves the strict-Ulysses layout:

- select rank-owned image/audio latent rows before their patch projections, run
  the token refiner on its required full text context before selecting rank-owned
  text rows, construct only the local embedding rows, and slice matching RoPE
  rows;
- keep transformer output rows local through the final projection, then gather
  `[S, 128]` FP32 logits instead of `[S, 5376]` BF16 hidden states;
- fall back to the existing full-sequence path for SP=1, Ring SP, AllGather-KV,
  hybrid/advanced layouts, missing hooks, non-forward contexts, or shapes that
  cannot be divided evenly.

For MiniMax-H3, the final all-gather payload is reduced from 10,752 bytes to
512 bytes per packed row (21x smaller). This is a boundary-payload reduction,
not a 21x end-to-end performance claim.

The implementation deliberately does not change Ulysses Q/K/V all-to-all,
the attention backend, packed-prefix handling, scheduler semantics, or model
weights. Focused tests exercise the optimized contract and every fallback above.

### 8x B300 performance and quality

Hardware and software:

- 8x NVIDIA B300 SXM6, 275,040 MiB per GPU, compute capability 10.3
- driver 580.126.09, CUDA 13.2, PyTorch 2.13.0+cu132
- vLLM 0.27.0, vLLM-Omni current main `dbc0dd6d`
- official MiniMax-H3 FL2VA checkpoint, BF16, default `TRTLLM_ATTN`

Fixed workload:

- T2VA, 1344x768, 107 frames, 24 fps, 50 denoising steps, seed 42
- Ulysses 8, Ring 1, AllGather-KV 1, text-encoder TP 8, VAE patch parallel 8
- regional `torch.compile`; one excluded warmup and five measured generations
  for each engine; B1 (main) -> candidate -> B2 (main) run order
- same prompt, checkpoint, environment, and generation parameters for every run

| Metric | Current main (10 runs, B1+B2) | This PR (5 runs) | Change |
[...]

PR #6344, opened 2026-08-19
[ci][bugfix] Restore MiniMax-H3 H100 performance baselines

david6666666:
### What broke

The MiniMax-H3 performance dashboard reports a stable 11.08%-11.11% peak-memory regression for all four 4xH100 cases. The values themselves are stable: each configured H100 baseline is exactly 90% of the original measured result.

| Case | Old baseline (MB) | Original measurement / new baseline (MB) | Reported latest (MB) |
|---|---:|---:|---:|
| T2V | 54,219.6 | 60,244.0 | 60,226.048 |
| TI2V | 54,622.8 | 60,692.0 | 60,691.968 |
| V2V | 55,180.8 | 61,312.0 | 61,312.0 |
| T2V DLO | 36,218.4 | 40,242.6667 | 40,233.984 |

The same 90% transform was also applied to the throughput and latency reference values.

### Reproduction

This command shows that the peak-memory baselines introduced by the original perf-config commit are 0.9 times the recorded 4xH100 measurements:

```bash
git show a874b8e0:tests/dfx/perf/tests/test_minimax_h3_vllm_omni.json   | jq '[.[].benchmark_params[0].baseline.H100.peak_memory_mb_mean] as $b
        | [60244, 60692, 61312, 40242.6667] as $m
        | [range(0; 4) | ($b[.] / $m[.])]'
```

Output:

```text
[0.9, 0.9, 0.9, 0.9]
```

### Root cause

PR #5836 treated the reference baseline as a failure threshold and reduced every measured value by 10%. The dashboard already applies its own metric-direction-aware tolerance. For lower-is-better metrics such as latency and peak memory, a stable run against a 90% reference necessarily appears about 11.11% worse.

### Fix

Restore the original measured 4xH100 throughput, latency, and peak-memory values for T2V, TI2V, V2V, and T2V DLO. This only changes dashboard reference data; no MiniMax-H3 runtime path or generated output changes.

## Test Plan

- Run the complete T2V perf case with 1 warmup and 3 measured requests at concurrency 1 on 4x NVIDIA B300:
  ```bash
  CUDA_VISIBLE_DEVICES=0,1,2,3   VLLM_TEST_MINIMAX_H3_MODEL=/path/to/materialized/MiniMax-H3   pytest -s -v tests/dfx/perf/scripts/run_diffusion_benchmark.py     --test-config-file tests/dfx/perf/tests/test_minimax_h3_vllm_omni.json     -k 'test_minimax_h3_t2v_4gpu and not dlo'
  ```
- Run metadata unit tests.
- Run pre-commit on the changed JSON.

**vLLM Version:** 0.27.0

**vLLM-Omni Commit:** 01893b8b

## Test Result

- 4x NVIDIA B300 (274 GB each), T2V: 3/3 requests completed; peak memory 57,102 MB; mean latency 20.4931 s; throughput 0.0488 req/s.
- `tests/dfx/perf/tests/test_runner_metadata.py`: 21 passed.
- `pre-commit run --files tests/dfx/perf/tests/test_minimax_h3_vllm_omni.json`: passed.
- `jq empty tests/dfx/perf/tests/test_minimax_h3_vllm_omni.json`: passed.

### DIFF
--- a/tests/dfx/perf/tests/test_minimax_h3_vllm_omni.json
+++ b/tests/dfx/perf/tests/test_minimax_h3_vllm_omni.json
@@ -63,9 +63,9 @@
         },
         "baseline": {
           "H100": {
-            "throughput_qps": 0.0235,
-            "latency_mean": 42.5836,
-            "peak_memory_mb_mean": 60244.0
+            "throughput_qps": 0.0211,
+            "latency_mean": 38.3252,
+            "peak_memory_mb_mean": 54219.6
           }
         }
       }
@@ -136,9 +136,9 @@
         "num-input-images": 1,
         "baseline": {
           "H100": {
-            "throughput_qps": 0.0236,
-            "latency_mean": 42.3761,
-            "peak_memory_mb_mean": 60692.0
+            "throughput_qps": 0.0212,
+            "latency_mean": 38.1385,
+            "peak_memory_mb_mean": 54622.8
           }
         }
       }
@@ -208,9 +208,9 @@
         },
         "baseline": {
           "H100": {
-            "throughput_qps": 0.0085,
-            "latency_mean": 117.5009,
-            "peak_memory_mb_mean": 61312.0
+            "throughput_qps": 0.0077,
+            "latency_mean": 105.7508,
+            "peak_memory_mb_mean": 55180.8
           }
         }
       }
@@ -284,9 +284,9 @@
         },
         "baseline": {
           "H100": {
-            "throughput_qps": 0.0236,
-            "latency_mean": 42.4543,
-            "peak_memory_mb_mean": 40242.6667
+            "throughput_qps": 0.0212,
+            "latency_mean": 38.2089,
+            "peak_memory_mb_mean": 36218.4
           }
         }
       }
--- a/tests/dfx/perf/tests/test_minimax_h3_vllm_omni.json
+++ b/tests/dfx/perf/tests/test_minimax_h3_vllm_omni.json
@@ -63,9 +63,9 @@
         },
         "baseline": {
           "H100": {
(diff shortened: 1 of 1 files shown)


hsliuustc0106:
填错baseline了？

david6666666:
still check

david6666666:
Closing after correcting the diagnosis. The baseline-only change has been fully reverted, so this PR now has no net diff.

A reverse git bisect over 2026-08-13 through current main, using the same MiniMax-H3 T2V configuration (1344x768, 209 frames, 8 steps, USP4 + HSDP4), found no memory-regressing commit in that range. The range starts in the high-memory state and the first memory-fixing commit is 7b76b644 / #6173.

Local 4xB300 measurements:
- pre-fix side (bc73546c and b95c7374): 60,176 MB
- 7b76b644: 56,974 MB
- later midpoint ce6ab0ca: 56,974 MB

The public H100 history independently shows 60,242 MB on build #13582 (90a08c0e, Aug 13) and 57,024 MB on build #13732 (d7216ef5, Aug 18). Therefore the pasted report is displaying the older failing measurement; the code-side memory reduction is already merged in #6173. No additional vLLM-Omni patch is justified here.

david6666666:
Follow-up nuance from checking all four raw history series: #6173 lowered the non-DLO cases into +5.17% (T2V), +6.74% (TI2V), and +5.81% (V2V) relative to their stored baselines. DLO is still +10.93%, but it has stayed at roughly 39.2-39.3 GiB since its very first Aug 6 record; there is no DLO regression transition to attribute to a commit. The original four +11.11% values arise because #5836 explicitly stored each baseline as 90% of the measured value. Per the corrected task, no baseline values were changed.

PR #6345, opened 2026-08-19
[BugFix][Diffusion] Fix MiniMax-H3 VAE hang when decoder tiles are fewer than ranks

### LINKED ISSUE #6394: [Bug]: MiniMax-H3 explicit height/width bypasses the output canvas policy (no lower or upper bound)
## Summary

MiniMax-H3 enforces its output canvas policy — short edge fixed at 768, area capped at `768*1344` — inside `_resolve_output_canvas`. That function is only called when the request omits `height`/`width`. Supplying them explicitly skips the whole policy, leaving only three checks: floor to a multiple of 32, `min(height, width) > 0`, and aspect ratio within `[1:4, 4:1]`.

So the same knob has two policies depending on how it is spelled, and the explicit path has neither a lower nor an upper bound.

## Where

`vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py`, in `_resolve_shape`:

```python
if height is None or width is None:
    height, width = _resolve_output_canvas(aspect_ratio, short_edge)   # policy lives here
height = int(height) // 32 * 32
width = int(width) // 32 * 32
if min(height, width) <= 0:
    raise OmniClientError(f"invalid MiniMax H3 canvas {width}x{height}")
if width > 4 * height or height > 4 * width:
    raise OmniClientError("MiniMax H3 canvas aspect ratio must be in [1:4, 4:1]")
```

Both `MINIMAX_H3_OUTPUT_SHORT_EDGE` (checked at line 486) and `MINIMAX_H3_OUTPUT_MAX_PIXELS` (applied at line 495) are inside `_resolve_output_canvas`.

Nothing upstream bounds them either: `VideoRequest.resolve_video_params` (`protocol/videos.py:272`) passes `width`/`height`/`size` through unvalidated, and `serving_video_output_stream.py:526` assigns them straight onto the sampling params.

## Consequence 1 — no lower bound

`256x256` is accepted. It is a legal multiple of 32 with a 1:1 aspect ratio, and it is 4x below the trained short edge.

Until #6345 this hung the engine outright: the decoder grid for a 256x256 canvas is a single tile, and a VAE tile group with more ranks than tiles leaves the tile-less ranks raising out of the gather while the rest block in it forever. #6345 replaces the hang with a rank-local fallback, so this is no longer a hang — but the request still runs far outside the trained regime, and silently drops tile parallelism.

Decoder grid against the checkpoint's own `split_tiles` (`tile_size=256`, `tile_overlap_min=64`):

| requested canvas | decoder tiles |
| --- | ---: |
| 256x256 | 1 |
| 384x384 | 4 |
| 512x512 | 9 |
| 768x768 | 16 |

## Consequence 2 — no upper bound

The area cap is skipped too, so a request can ask for far more than the policy allows. `1536x2688` passes every remaining check (aspect 1.75, both multiples of 32) at **4x** `MINIMAX_H3_OUTPUT_MAX_PIXELS`.

Measured on 4x RTX 5090, VAE decode alone for that canvas at 111 frames peaks at **25.15 GiB per rank**, and the peak does not fall as the tile group grows — every rank materialises the full tile set in `_all_gather_tiled_results` (measurement: https://github.com/vllm-project/vllm-omni/issues/5948#issuecomment-5336436264). Nothing stops a request going further; the ceiling is whatever the card happens to survive.

## Why the two paths disagree

The policy path treats short edge as a constant and rejects anything else:

```python
if short_edge != MINIMAX_H3_OUTPUT_SHORT_EDGE:
    raise OmniClientError(f"MiniMax H3 target.short_edge must be {MINIMAX_H3_OUTPUT_SHORT_EDGE}, got {short_edge}")
```

A caller who sets `target.short_edge = 512` is rejected. A caller who sets `height=512, width=512` is accepted. Same intent, opposite outcome.

## Suggested fix

Apply the same policy on both paths — validate `height`/`width` against the short-edge and area rules instead of only when they are absent. Concretely:

- reject a canvas whose short edge is not `MINIMAX_H3_OUTPUT_SHORT_EDGE`, or accept a documented range and say what it is;
- apply `MINIMAX_H3_OUTPUT_MAX_PIXELS` to explicit canvases as well;
- keep the existing multiple-of-32 and aspect-ratio checks.

If explicit off-policy canvases are deliberately supported for experimentation, then the fix is the opposite one: relax the `target.short_edge` check so both spellings behave the same, and document that quality outside 768 is not supported.

Either resolution is fine; the current state is that the two spellings disagree, and one of them has no bounds at all.

## Notes

- Found while validating #6345. That PR fixes the hang, not this — a small canvas no longer deadlocks, but it is still accepted.
- The audio VAE is unaffected: `MiniMaxH3AudioVAE` does not inherit `DistributedVaeMixin`, is never given a parallel size, and its `decode_latent` calls `remote.decode` directly with no tiling or gather.
- Reference-video inputs are not affected either — `_resolve_reference_canvas` pins the short edge to 768 unconditionally, so a reference grid is always at least 16 tiles.

linzhenpl07:
## Purpose

Fixes a hang in MiniMax-H3 video VAE decode when a request produces fewer decoder tiles than the tile group has ranks.

## Root cause

The checkpoint splits tiles by rank and rejects an empty share inside the gather (`FL2VA/video_vae/klvae.py`):

```python
def _local_tile_indices(self, num_tiles, sp_rank, sp_size):
    return list(range(sp_rank, num_tiles, sp_size))

def _all_gather_tiled_results(self, tasks, num_tiles):
    if not tasks:
        raise ValueError(f"Found empty tasks on sp rank {sp_rank}")
    gathered = all_gather_var_shape(stacked, group=group)
```

Ranks with no tiles raise and leave the collective; ranks with tiles block in `all_gather_var_shape` forever. The stage hangs instead of failing.

Reproduced on 4x RTX 5090, `vae_patch_parallel_size=4`, latent `(1, 24, 9, 24, 16)`:

```
CASE latent 24x16 (px 384x256)  num_tiles=2x1=2  G=4
  rank2: RAISE ValueError: Found empty tasks on sp rank 2
  rank3: RAISE ValueError: Found empty tasks on sp rank 3
  -> 180 s timeout, ranks 0/1 still at 100% GPU
```

## When it triggers

Tile count is computed on pixel dimensions with `vae_tile_size=256` and `overlap=64`, so it is small only for small outputs. Verified against the checkpoint's own `split_tiles`:

| output | tile grid | hangs a tile group of |
| --- | --- | --- |
| 256x256 | 1x1 = 1 | 2 or more |
| 384x256 | 2x1 = 2 | 3 or more |
| 384x384 | 2x2 = 4 | 5 or more |
| 768x1344 (shipped) | 4x7 = 28 | 29 or more |
| 1536x2688 | 8x14 = 112 | 113 or more |

So low-resolution previews and smoke tests wedge a multi-GPU diffusion stage, while the shipped H3 resolution is unaffected.

## Fix

`decode_latent` compares the tile count against `parallel_size` and, when tiles are short, runs that one call with tiling kept on each rank:

```python
num_tiles = self._decoder_tile_count(latent)
if self.parallel_size > 1 and num_tiles < self.parallel_size:
    logger.warning_once(...)
    tiling_context = self._rank_local_tiling()
else:
    tiling_context = nullcontext()

with tiling_context:
    decoded = self.model.decode_base(latent * std + mean)
```

**Why the branch cannot diverge:** the tile grid is a pure function of the latent shape, and every rank decodes the same latent, so every rank computes the same `num_tiles` and takes the same branch. There is no path where only some ranks enter the collective.

Falling back means each rank decodes every tile, which is slower than sharing the work, but it is correct and involves no collective. `_rank_local_tiling` restores the checkpoint's native parallel state on both normal and exceptional exit.

`_native_parallel_state()` is factored out of the existing `set_parallel_size` so both places reach that state the same way.

## Test Plan

```bash
pytest tests/diffusion/models/minimax_h3/test_minimax_h3_vae_tiling.py -q
pytest tests/diffusion/models/minimax_h3 tests/diffusion/distributed -q
```

Plus a real 4-rank run of the repro above, with two regression controls that must keep the parallel path.

## Test Result

New CPU tests: 12 passed. They pin the tile-count arithmetic at four resolutions, both restore paths of the context manager, and six cases around the boundary (2 tiles / 4 ranks fires, 4 tiles / 4 ranks does not, 2 tiles / 2 ranks does not, `parallel_size=1` never fires).

The tile counts asserted in the test were checked against the loaded checkpoint:

```
latent 16x16  (px 256x256)   -> 1x1  = 1
latent 24x16  (px 384x256)   -> 2x1  = 2
latent 24x24  (px 384x384)   -> 2x2  = 4
latent 48x84  (px 768x1344)  -> 4x7  = 28   <- the shipped request
latent 96x168 (px 1536x2688) -> 8x14 = 112
```

Existing suites: 387 passed, 62 skipped. `ruff check` and `ruff format --check` clean.

Real hardware, 4x RTX 5090, `vae_patch_parallel_size=4`:

| latent | tiles | before | after |
| --- | ---: | --- | --- |
| 24x16 | 2 | hangs, 180 s timeout | `exit=0`, guard fires, all 4 ranks return |
| 16x16 | 1 | hangs | `exit=0`, guard fires |
| 24x24 | 4 | ok | `exit=0`, guard does not fire |
| 96x168 | 112 | ok | `exit=0`, guard does not fire |

The last two are the regression controls. Note that `96x168` is a 1536x2688 canvas, not the shipped 768x1344 request; @david6666666 covered the shipped 48x84 / 28-tile shape separately on 4x B300 and it stayed on the parallel path.

## Related

Found while measuring step 4 of #5948; the measurements are in https://github.com/vllm-project/vllm-omni/issues/5948#issuecomment-5336436264.

### DIFF
--- a/tests/diffusion/models/minimax_h3/test_minimax_h3_vae_tiling.py
+++ b/tests/diffusion/models/minimax_h3/test_minimax_h3_vae_tiling.py
@@ -1,0 +1,112 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+
+"""CPU tests for the decoder tile-shortage guard in the MiniMax-H3 video VAE."""
+
+import sys
+import types
+
+import pytest
+import torch
+
+from vllm_omni.diffusion.models.minimax_h3.vae import MiniMaxH3VideoVAE
+
+pytestmark = [pytest.mark.core_model, pytest.mark.cpu, pytest.mark.diffusion]
+
+TILE_SIZE, OVERLAP, RATIO = 256, 64, 16
+
+
+class _FakeCheckpointModel:
+    """The parts of the checkpoint's AutoencoderKL the guard touches."""
+
+    def __init__(self):
+        self.vae_ratio = RATIO
+        self.parallel_tiling = True
+        self.decoded_with = []
+
+    def split_tiles(self, input_len, is_decoder=False):
+        """Same grid arithmetic as the checkpoint, on pixel dimensions."""
+        if TILE_SIZE >= input_len:
+            return [0], [input_len], []
+        n = -(-input_len // TILE_SIZE)
+        while TILE_SIZE * n - OVERLAP * (n - 1) - input_len < 0:
+            n += 1
+        return list(range(n)), [TILE_SIZE] * n, [OVERLAP] * (n - 1)
+
+
+def _vae(parallel_size, state):
+    """A stand-in instance: the guard only reads .model/.remote/.parallel_size."""
+    module = types.ModuleType("fake_ckpt.parallel")
+    module.get_parallel_state = lambda: state
+    sys.modules.setdefault("fake_ckpt", types.ModuleType("fake_ckpt"))
+    sys.modules["fake_ckpt.parallel"] = module
+
+    remote = type("Remote", (), {"__module__": "fake_ckpt.klvae"})()
+
+    vae = object.__new__(MiniMaxH3VideoVAE)
+    vae.model = _FakeCheckpointModel()
+    vae.remote = remote
+    vae.parallel_size = parallel_size
+    return vae
+
+
+def _latent(h, w):
+    return torch.zeros(1, 24, 9, h, w)
+
+
+@pytest.mark.parametrize(
+    ("h", "w", "expected"),
+    [
+        (16, 16, 1),  # 256x256 px -> single tile
+        (24, 16, 2),  # 384x256 px
+        (24, 24, 4),  # 384x384 px
+        (96, 168, 112),  # 1536x2688 px, the shipped H3 resolution
+    ],
+)
+def test_tile_count_matches_the_checkpoint_grid(h, w, expected):
+    assert MiniMaxH3VideoVAE._decoder_tile_count(_vae(1, {}), _latent(h, w)) == expected
+
+
+def test_rank_local_tiling_restores_the_group_state():
+    state = {"sp_size": 4, "sp_rank": 2, "sp_enabled": True, "sp_process_group": "pg"}
+    vae = _vae(4, state)
+
+    with vae._rank_local_tiling():
+        assert state["sp_size"] == 1
+        assert state["sp_rank"] == 0
+        assert state["sp_enabled"] is False
+        assert state["sp_process_group"] is None
+        assert vae.model.parallel_tiling is False
+
+    assert state == {"sp_size": 4, "sp_rank": 2, "sp_enabled": True, "sp_process_group": "pg"}
+    assert vae.model.parallel_tiling is True
+
+
+def test_rank_local_tiling_restores_after_an_exception():
+    state = {"sp_size": 4, "sp_rank": 1, "sp_enabled": True, "sp_process_group": "pg"}
+    vae = _vae(4, state)
+
+    with pytest.raises(RuntimeError, match="decode failed"):
+        with vae._rank_local_tiling():
+            raise RuntimeError("decode failed")
+
+    assert state["sp_size"] == 4
+    assert vae.model.parallel_tiling is True
+
+
+@pytest.mark.parametrize(
+    ("parallel_size", "h", "w", "falls_back"),
+    [
+        (4, 24, 16, True),  # 2 tiles, ranks 2 and 3 would get none
+        (4, 16, 16, True),  # 1 tile
+        (4, 24, 24, False),  # 4 tiles, exactly enough
+        (4, 96, 168, False),  # 112 tiles
+        (2, 24, 16, False),  # 2 tiles across 2 ranks
+        (1, 16, 16, False),  # never parallel, nothing to guard
+    ],
+)
+def test_guard_fires_exactly_when_tiles_are_short(parallel_size, h, w, falls_back):
+    """Fewer tiles than ranks is the hang condition; equal or more is fine."""
+    vae = _vae(parallel_size, {"sp_size": parallel_size, "sp_rank": 0})
+    num_tiles = MiniMaxH3VideoVAE._decoder_tile_count(vae, _latent(h, w))
+    assert (vae.parallel_size > 1 and num_tiles < vae.parallel_size) is falls_back
--- a/tests/diffusion/models/minimax_h3/test_minimax_h3_vae_tiling.py
+++ b/tests/diffusion/models/minimax_h3/test_minimax_h3_vae_tiling.py
@@ -13,7 +13,15 @@ from vllm_omni.diffusion.models.minimax_h3.vae import MiniMaxH3VideoVAE
 
 pytestmark = [pytest.mark.core_model, pytest.mark.cpu, pytest.mark.diffusion]
 
-TILE_SIZE, OVERLAP, RATIO = 256, 64, 16
+RATIO = 16
+# The checkpoint keeps two independent grids -- ``split_tiles`` selects
+# ``decoder_tile_size``/``decoder_tile_overlap_min`` when ``is_decoder`` is set
+# and ``tile_size``/``tile_overlap_min`` otherwise. The shipped H3 config leaves
+# both at 256/64, so the two happen to coincide there, but that is a config
+# coincidence. The double deliberately gives them different values so that
+# passing the wrong flag changes the tile count and fails a test.
+TILE_SIZE, OVERLAP = 256, 64  # decoder, matching the shipped config
+ENCODER_TILE_SIZE, ENCODER_OVERLAP = 128, 32
 
 
 class _FakeCheckpointModel:
@@ -26,12 +34,14 @@ class _FakeCheckpointModel:
 
     def split_tiles(self, input_len, is_decoder=False):
         """Same grid arithmetic as the checkpoint, on pixel dimensions."""
-        if TILE_SIZE >= input_len:
+        tile = TILE_SIZE if is_decoder else ENCODER_TILE_SIZE
+        overlap = OVERLAP if is_decoder else ENCODER_OVERLAP
+        if tile >= input_len:
             return [0], [input_len], []
-        n = -(-input_len // TILE_SIZE)
-        while TILE_SIZE * n - OVERLAP * (n - 1) - input_len < 0:
+        n = -(-input_len // tile)
+        while tile * n - overlap * (n - 1) - input_len < 0:
             n += 1
-        return list(range(n)), [TILE_SIZE] * n, [OVERLAP] * (n - 1)
+        return list(range(n)), [tile] * n, [overlap] * (n - 1)
 
 
 def _vae(parallel_size, state):
@@ -67,6 +77,17 @@ def test_tile_count_matches_the_checkpoint_grid(h, w, expected):
     assert MiniMaxH3VideoVAE._decoder_tile_count(_vae(1, {}), _latent(h, w)) == expected
 
 
+@pytest.mark.parametrize(("h", "w", "decoder_tiles", "encoder_tiles"), [(16, 16, 1, 9), (24, 24, 4, 16)])
+def test_tile_count_uses_the_decoder_grid_not_the_encoder_one(h, w, decoder_tiles, encoder_tiles):
+    """Passing ``is_decoder=False`` would silently change the count, so pin it."""
+    vae = _vae(1, {})
+    assert MiniMaxH3VideoVAE._decoder_tile_count(vae, _latent(h, w)) == decoder_tiles
+
+    px_h, px_w = h * RATIO, w * RATIO
+    as_encoder = len(vae.model.split_tiles(px_h, False)[0]) * len(vae.model.split_tiles(px_w, False)[0])
+    assert as_encoder == encoder_tiles != decoder_tiles
+
+
 def test_rank_local_tiling_restores_the_group_state():
     state = {"sp_size": 4, "sp_rank": 2, "sp_enabled": True, "sp_process_group": "pg"}
     vae = _vae(4, state)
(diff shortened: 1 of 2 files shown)


tests/diffusion/models/minimax_h3/test_minimax_h3_vae_tiling.py
```diff
@@ -0,0 +1,112 @@
...
+    [
+        (4, 24, 16, True),  # 2 tiles, ranks 2 and 3 would get none
+        (4, 16, 16, True),  # 1 tile
+        (4, 24, 24, False),  # 4 tiles, exactly enough
+        (4, 96, 168, False),  # 112 tiles
+        (2, 24, 16, False),  # 2 tiles across 2 ranks
+        (1, 16, 16, False),  # never parallel, nothing to guard
+    ],
+)
+def test_guard_fires_exactly_when_tiles_are_short(parallel_size, h, w, falls_back):
```
hsliuustc0106: [P2] None of these tests execute `decode_latent`, so deleting the guard at `vae.py:356-366` leaves the whole suite green — the tile-count and restore tests keep passing because they exercise the helpers in isolation, and this predicate test recomputes the same condition rather than calling the dispatch. Could you add one CPU test that calls `decode_latent` on the fake (`parallel_size=4`, 1-tile latent, stubbed `decode_base` + `processor.revert_tensor`, plus a small `config_dict`) and asserts the stub observed the rank-local state (`sp_size == 1` / `parallel_tiling is False`)? That fails on base and pins the actual fix.

vllm_omni/diffusion/models/minimax_h3/vae.py
```diff
@@ -293,7 +347,27 @@ def decode_latent(self, latent: torch.Tensor) -> torch.Tensor:
...
         ).view(1, channels, 1, 1, 1)
-        decoded = self.model.decode_base(latent * std + mean)
+        # The checkpoint hands rank r the tiles ``range(r, num_tiles, sp_size)``
+        # and then rejects an empty share inside the gather. A rank with no
+        # tiles raises and leaves the collective while the others block in it
+        # forever, so too few tiles hangs the whole stage rather than failing
+        # it. Tile count depends only on the latent shape, so every rank takes
+        # this branch together.
+        num_tiles = self._decoder_tile_count(latent)
+        if self.parallel_size > 1 and num_tiles < self.parallel_size:
```
hsliuustc0106, line 357: [P3] The same condition is still live on the encode side: `encode_video` (`vae.py:237`) enters the checkpoint's `tiled_encode`, which gathers through the identical empty-share raise, and every rank enters it when tile parallelism is on (`pipeline_minimax_h3.py:1307`). Reference videos are normalized to a ≥768 short edge (`reference_video.py:21`), so the encoder grid is ≥16 tiles and nothing can fire for tile groups ≤16 — but `set_parallel_size` requires the full DiT group, so any world_size ≥ 17 hangs a reference-video request exactly the same way. Worth applying the same guard to `encode_video` (or recording the ≥17-rank bound) while this is fresh.

hsliuustc0106 (commented):
Verified the fix against the checkpoint's own `split_tiles`/`tiled_decode` and the shipped VAE config: the tile-count mirror is exact (it calls the checkpoint's own method, so tile sizes can't drift), the branch is a pure function of the latent shape so every rank takes it together, and the fallback path is collective-free. No blockers from me — the two comments above are one ask for a regression test that actually exercises `decode_latent`, and one note about the encode-side twin. Happy to approve once the test lands.

hsliuustc0106:
@lishunyang12 @david6666666 PTAL

linzhenpl07:
Thanks — the first one was a real hole in what I submitted. Pushed `780dc3cf`, tests only.

## [P2] Tests now go through `decode_latent`

You were right that the suite was green without the fix, and that the parametrized test recomputed the predicate instead of calling the dispatch. Rather than bolt a new test alongside it, I routed the existing trigger matrix through `decode_latent` itself and added two assertions on the state `decode_base` actually observed:

- the six-row matrix now calls `decode_latent` and asserts `parallel_tiling is False` exactly when the grid is short — `parallel_tiling` is the one flag only the guard clears, so it is the signal that it fired;
- `test_decode_latent_isolates_and_restores_the_whole_group_state` — 1 tile, `parallel_size=4`: `sp_size == 1`, `sp_enabled is False`, `sp_process_group is None` inside `decode_base`, and the group state restored on return;
- `test_decode_latent_leaves_the_group_state_alone_when_the_grid_is_large_enough` — 112 tiles, `parallel_size=4`: the guard does not fire and does not touch the state.

Verified against base by disabling only the predicate:

```
guard present   : 14 passed
guard disabled  : 3 failed, 11 passed
guard restored  : 14 passed
```

`vae.py` is byte-identical to the commit the 4-rank hardware validation in the description was run against, so that evidence still stands as written.

## [P3] Encode side — taking the "record the bound" option for now

Your analysis checks out: `MINIMAX_H3_BASE_SHORT_EDGE = 768` (`reference_video.py:21`) with `tile=256 / overlap=64` gives at least `ceil((768-64)/192) = 4` tiles per side, so the encoder grid is >= 16, and since `set_parallel_size` requires the full DiT group the first hanging configuration is `world_size >= 17`. `encode_image` is not exposed — it already forces `parallel_tiling = False`.

I had a guard for `encode_video` written, then backed it out of this PR for two reasons:

1. **I cannot currently verify the tile count on that path.** The decoder calls `split_tiles(len, True)`; the encoder would need the other flag, and I have no way to confirm the checkpoint's signature right now — a CPU test would be validating my own stub against my own call.
2. **A false positive there is not free.** `decode` is deterministic, so a spurious fallback only costs speed — I measured byte-identical output across group sizes. `encode_video` samples under `fork_rng` with a fixed seed, so a spurious fallback changes the tile split and therefore the reference-conditioning latents, and the surrounding code already warns that path is numerically delicate.

Given the configuration is unreachable below 17 ranks, shipping an unverifiable change that can silently alter conditioning seemed worse than recording the bound. Happy to open it as a follow-up once I can validate the encoder grid against the real `split_tiles` and show the guard does not fire on a normal reference video.

tests/diffusion/models/minimax_h3/test_minimax_h3_vae_tiling.py
```diff
@@ -0,0 +1,191 @@
...
+    [
+        (4, 24, 16, True),  # 2 tiles, ranks 2 and 3 would get none
+        (4, 16, 16, True),  # 1 tile
+        (4, 24, 24, False),  # 4 tiles, exactly enough
+        (4, 96, 168, False),  # 112 tiles
+        (2, 24, 16, False),  # 2 tiles across 2 ranks
+        (1, 16, 16, False),  # never parallel, nothing to guard
+    ],
+)
+def test_decode_latent_falls_back_exactly_when_tiles_are_short(parallel_size, h, w, falls_back):
```
david6666666, line 178: This closes the earlier coverage gap: the exact boundary now exercises decode_latent instead of duplicating its predicate.

linzhenpl07:
@david6666666 @hsliuustc0106 — status check, plus the `>=17-rank bound` @hsliuustc0106 asked me to record, now measured rather than estimated.

## Where this stands

- [P2] the trigger matrix now runs through `decode_latent` instead of duplicating its predicate; 3 of the 14 tests fail if the guard is removed.
- [P3] answered below, as the "record the bound" option rather than a code change.
- `vae.py` is byte-identical to the commit the 4-rank hardware validation was run against, so that evidence stands as written.
- All checks green, `MERGEABLE`.

## The bound, from the checkpoint's own code

I pulled `FL2VA/video_vae/klvae.py` and replayed `split_tiles` and `_resolve_reference_canvas` exactly.

Confirmed `split_tiles(self, input_len, is_decoder=False)` selects `decoder_tile_size` / `decoder_tile_overlap_min` for the decoder and `tile_size` / `tile_overlap_min` for the encoder. They are separate knobs; this checkpoint leaves both at 256/64, so the two grids coincide here — but that is a config coincidence, not a guarantee.

Sweeping the whole legal reference aspect range `[0.4, 2.5]` through the real canvas resolver:

```
61 distinct canvases; smallest grid = 16 tiles @ 768x832 (ratio 0.906)
-> encoder shortage needs a VAE group >= 17
-> set_parallel_size requires the full DiT group, so world_size >= 17
```

So your bound is exact. Reference videos are pinned to a 768 short edge, so a request cannot shrink the encoder grid.

The decoder is different, which is why it is worth fixing now: output height/width come straight from the request (`serving_video_output_stream.py:526`, floored to a multiple of 32, with no lower bound beyond `> 0`):

| requested canvas | tiles | hangs at VAE group |
| --- | ---: | ---: |
| 256x256 | 1 | **>= 2** |
| 384x384 | 4 | >= 5 |
| 512x512 | 9 | >= 10 |
| 768x768 | 16 | >= 17 |

A small-canvas request hangs a 2-GPU deployment today.

Two side checks while I had the real implementation: the test double matches the real tile count for every length in 32..6000, and all four asserted decoder counts (1 / 2 / 4 / 112) reproduce exactly.

## Question

Is there anything else you'd like changed here, or is it ready to go in? Happy to open the encode-side guard as a follow-up once this lands — I have the verification above, so the two things that made me hold it back (unconfirmed `split_tiles` flag, and the risk of a false positive silently changing conditioning latents) are both resolved.

linzhenpl07:
Small follow-up push, tests only — `vae.py` is still byte-identical to the commit the 4-rank validation ran against.

While reading `klvae.py` for the bound above I noticed the test double was ignoring the `is_decoder` flag it declared, so a wrong flag in production would not have failed anything. The double now gives the two grids different values (decoder 256/64 as shipped, encoder 128/32), and a new case pins that `_decoder_tile_count` uses the decoder grid:

| latent | decoder tiles | if the encoder grid were used |
| --- | ---: | ---: |
| 16x16 | 1 | 9 |
| 24x24 | 4 | 16 |

Flipping `split_tiles(..., True)` to `False` in `vae.py` now fails 9 of the 16 tests; before this it failed none. The four documented decoder counts (1 / 2 / 4 / 112) are unchanged, since the decoder side of the double still matches the shipped config.

The merge question in the previous comment still stands.

tests/diffusion/models/minimax_h3/test_minimax_h3_vae_tiling.py
```diff
@@ -0,0 +1,212 @@
...
+    return torch.zeros(1, 24, 9, h, w)
+
+
+@pytest.mark.parametrize(
+    ("h", "w", "expected"),
+    [
+        (16, 16, 1),  # 256x256 px -> single tile
+        (24, 16, 2),  # 384x256 px
+        (24, 24, 4),  # 384x384 px
+        (96, 168, 112),  # 1536x2688 px, the shipped H3 resolution
```
david6666666: The shipped 768×1344 request reaches `decode_latent` as a 48×84 latent (`height / 16`, `width / 16`), and the real checkpoint produces a 4×7 = 28-tile grid. This 96×168 case is actually 1536×2688, so could we use `(48, 84, 28)` here to make the production-shape control accurate?

david6666666 (commented):
Validated `6b5ae871` locally on 4×B300 with the real cached MiniMax-H3 VAE:

- 2 tiles / 4 ranks returned successfully; `decode_base` saw rank-local state (`sp_size=1`, SP disabled, process group cleared), and the 4-rank state was restored afterward.
- 4 tiles / 4 ranks and the shipped 768×1344 shape (28 tiles) stayed on the parallel path. Output summaries matched across all ranks exactly.
- 16 focused tests and 29 related VAE tests passed; Ruff and diff checks were clean.

The local env warns about vLLM 0.27.0 vs vLLM-Omni 0.26.1.dev. I also stopped an exploratory wider sweep after 161 passed / 3 skipped when an unrelated test ran long, so I am not counting that as a full-suite pass. The runtime fix looks sound; one test-control mismatch is noted inline.

linzhenpl07:
@david6666666 thanks for running it on real hardware — and you are right, that control was mislabelled. Fixed in `8c51aec3`.

`96x168` is a 1536x2688 canvas, not the shipped request. The shipped 768x1344 reaches `decode_latent` as `48x84` and the checkpoint's `split_tiles` gives `4x7 = 28`, exactly as you said. I replayed both against the real implementation:

```
latent 48x84  (px 768x1344)  -> 4x7  = 28   <- the shipped request
latent 96x168 (px 1536x2688) -> 8x14 = 112
```

`48x84 / 28` is now the production-shape control in both the tile-count table and the dispatch matrix, and the "guard must not fire" test runs on it. `96x168` stays as a larger canvas the request path also accepts, relabelled rather than dropped — it is still a useful upper control, and #6394 is about the fact that the request path accepts it at all.

The same mistake was in the PR description's trigger matrix, which listed `1344x768` against the 112-tile grid. Corrected there too, and the regression-control note now says explicitly that your 4x B300 run is what covers the shipped shape rather than implying mine did.

18 tests pass; ruff clean.

david6666666 (approved):
Re-reviewed the latest delta. The shipped-shape control now uses the correct 48×84 latent / 28-tile grid, and all 18 focused tests pass locally. The current AMD failures are in unrelated offloader, engine, and TTS jobs. No blockers from me.

PR #6346, opened 2026-08-19
[Bugfix][MiniCPM-o] Stop Talker mid-utterance truncation and load cod…

### LINKED ISSUE #5259: [Bug]: MiniCPM-o 4.5: chat completion text includes final sentence but returned audio truncates before speaking it
## Summary

When serving **MiniCPM-o 4.5** with vLLM-Omni via OpenAI-compatible `/v1/chat/completions` (`modalities: text+audio`), the **text response is complete**, but the **returned WAV audio stops before the last sentence**. The final closing sentence appears only in text, not in speech.

```
Chat completion output from text: 当然可以。从前有一个小男孩叫小明，他住在一个宁静的村庄里。每天早上，他都会去村边的小溪边钓鱼。有一天，他在水面上发现了一枚闪闪发光的石头。

小明捡起那块石头，惊奇地发现它竟然能说话！“谢谢你救了我！”石头说，“我是来自遥远星球的魔法宝石。”

小明惊讶极了：“你真的会说话？”他小心翼翼地问。

是啊，”宝石回答道，“但只有善良的人才能听懂我的声音。”宝石告诉小明，它可以实现三个愿望，但要小心使用它们。

小明想了想，第一个愿望是让全村人都变得快乐无比。“好主意！”村民们听了后都笑开了花儿。

接着，小明用第二个愿望帮助生病的老奶奶恢复健康。“太棒了！”老奶奶感激地说。

最后一个小愿望有些难决定。小明最终选择许愿让村子周围长满美丽的花朵和树木。“这真是个美好的世界！”大家都赞叹不已。

从那天起，小明成了村里最受欢迎的孩子之一。每当有人遇到困难时，他们总会想起那个神奇的宝石，并向小明寻求建议或帮助。

多年过去了，虽然那颗宝石已经不见了踪影，但它带来的善意与爱却永远留在了每个人心中。而小明也一直记得那个改变他一生的日子——因为有了那颗神奇的宝石，他的生活变得更加精彩纷呈！

这就是关于一颗神奇宝石和一位勇敢男孩的故事啦！希望你喜欢哦~
Audio saved to audio_chatcmpl-86f448a05b0c504e_0.wav
```

[audio_chatcmpl-86f448a05b0c504e_0.wav](https://github.com/user-attachments/files/30216559/audio_chatcmpl-86f448a05b0c504e_0.wav)

Missing spoken sentence (present in text, absent in audio):

> 这就是关于一颗神奇宝石和一位勇敢男孩的故事啦！希望你喜欢哦~

## Environment

| Item | Value |
|------|--------|
| Model | `openbmb/MiniCPM-o-4_5` (local: `/data/why/MiniCPM-o-4_5`) |
| vLLM-Omni | `0.25.0rc2.dev47+g10d3a2591` (git `81416c590`) |
| vLLM server log | `vLLM server version 0.25.0` |
| Endpoint | `POST /v1/chat/completions` |
| Deploy config | `vllm_omni/deploy/minicpmo_4_5.yaml` (`async_chunk: false`, stage0+stage1 co-located) |
| Hardware | CUDA (H20-class, single GPU layout: devices `"0"` → physical GPU 2) |
| Client | `examples/online_serving/minicpmo/openai_chat_completion_client_for_multimodal_generation.py` |

### 1. Serve

```bash
vllm serve /path/to/MiniCPM-o-4_5 --omni --port 28889 \
  --trust-remote-code \
  --interleave-mm-strings
```

### 2. Request (long Chinese story)

```bash
python examples/online_serving/minicpmo/openai_chat_completion_client_for_multimodal_generation.py \
  --model /path/to/MiniCPM-o-4_5 \
  --query-type text \
  --port 28889 \
  --prompt "你好，可以给我讲一个500字的故事吗？"
```

(Default client modalities are text+audio / TTS template enabled.)

### 3. Observe

- Print `choice.message.content`: full story **including** the closing sentence above.
- Decode `choice.message.audio.data` → WAV.
- Listen / ASR the WAV: speech ends on the previous paragraph; the closing sentence is **not** spoken.

## Actual vs Expected

| | Text (`message.content`) | Audio (`message.audio`) |
|--|--------------------------|-------------------------|
| **Actual** | Complete, ends with “这就是关于一颗神奇宝石和一位勇敢男孩的故事啦！希望你喜欢哦~” | Truncates before that sentence |
| **Expected** | Same | Spoken content should cover the full text response (or document intentional truncation) |

## Evidence from this run

**Request id:** `chatcmpl-86f448a05b0c504e`  
**Saved audio:** `audio_chatcmpl-86f448a05b0c504e_0.wav`

Audio file properties:

- 24 kHz mono PCM
- `2,298,240` samples → **~95.76 s**

Server logs (Talker / stage 1):

```text
4.5 Talker: generating speech for 310 tokens
generate_speech: inputs_embeds shape=[1, 312, 768]
generate_speech: generated 2540 audio tokens (cap=3100, text_tokens=310)
generate_speech: streaming vocoder, 2540 tokens -> 51 chunks (chunk=50)
generate_speech: waveform 2298240 samples, sr=24000
```

Stage stats (same request):

| Stage | `num_tokens_out` / notes |
|-------|--------------------------|
| 0 (Thinker) | `312` text units out |
| 1 (Talker) | `2,298,240` audio frames; `audio_generated_frames=2298240` |

Notable:

1. **Not hitting `max_new_token`:** `cap=3100`, generated `2540` → Talker stopped via **early EOS** (or equivalent), not the hard length cap in `minicpmo_4_5_omni_tts.py` (`max(2048, min(16384, num_text * 10))`).
2. **Streaming vocoder path engaged:** `2540 > STREAM_THRESHOLD` (default `2500`), so Token2Wav used chunked `stream()` (`CHUNK_SIZE=50`, 51 chunks) instead of one-shot `Token2wav.__call__`.
3. Text path and speech path diverge: Thinker finishes the full story text; speech ends early relative to that text.

## Suspected areas (for investigation)

Not claiming root cause yet; likely candidates in the MiniCPM-o 4.5 Omni TTS path:

1. **Talker AR early EOS** in `MiniCPMTTS.generate()` / `minicpmo_4_5_omni_tts.py` `generate_speech()` — audio token stream ends before covering the full `tts_bos..tts_eos` text region (310 text tokens → 2540 audio tokens ≈ 8.2×, below the 10× safety budget but still incomplete speech).
2. **Streaming Token2Wav tail drop** when `num_tokens > MINICPMO45_TTS_STREAM_THRESHOLD` — chunked `stream(..., last_chunk=...)` may clip the final spoken content even if discrete tokens were produced (worth A/B: force one-shot path via raising threshold / shorter prompt so `num_tokens <= 2500`).
3. **Thinker→Talker condition construction** — mismatch between tokens passed to Talker (`310`) vs Thinker `num_tokens_out` (`312`), or incomplete `tts_bos..tts_eos` / hidden-state merge for the trailing sentence.
4. **HF reference parity** — confirm whether the same prompt on HuggingFace `MiniCPMO.generate` / official duplex path speaks the closing sentence; if HF is complete, this is a vLLM-Omni serving bug.

## Suggested checks / workarounds to try

- Re-run with a shorter prompt (stay under streaming threshold) and check whether tail truncation still happens.
- Set `MINICPMO45_TTS_STREAM_THRESHOLD` high enough to force non-streaming vocoder and compare.
- Dump Talker discrete token count vs expected speech duration for the full Chinese text; compare against HF reference audio length.
- Add an assertion / metric: ASR(audio) should cover the final N characters of `message.content` for long story prompts.

## Impact

- **Severity:** High for voice chat / storytelling UX — users see a complete reply but hear a truncated one.
- **Scope:** Observed on MiniCPM-o 4.5 online multimodal generation (`/v1/chat/completions` with audio). Unknown whether Qwen3-Omni / other TTS pipelines share a similar “text complete, audio short” failure mode.

## Additional context

- Deploy profile uses non-streaming omni path (`async_chunk: false`).
- Stage 1 `max_num_seqs: 1` (correctness-first Talker).
- Issue appears on a **single** long-form Chinese generation request (no concurrency).

amy-why-3459:
## Purpose

Fix: https://github.com/vllm-project/vllm-omni/issues/5259

MiniCPM-o Talker was cutting speech mid-utterance for two independent reasons, and codec sampling could not be tuned from deploy YAML.

1. **Codec budget ignored the Talker prompt.** Non-duplex prefill used `text_len * 10` capped at 2048, counted raw `token_ids` rather than the padded prompt (`full_embeds` / `_omni_prompt_len`), and did not reserve those prompt positions inside the 4096-token Talker window. Decode then hit `max_model_len` and stopped on `reached_limit` instead of codec EOS.

2. **`min_tokens` could exceed the remaining budget.** EOS is masked while `step < min_tokens`. After reserving prompt positions, a leftover window smaller than the YAML floor (50) made EOS illegal for the entire generation, so the utterance was always hard-cut.

3. **Codec knobs were hardcoded** (`T=0.8`, `top_k=25`, `top_p=0.85`, `rep=1.05`, ceiling 2048) and silently ignored stage-1 YAML.

This PR:

- Loads required `codec_sampling_params` from deploy YAML (`seed`, `temperature`, `top_k`, `top_p`, `repetition_penalty`, `min_tokens`, `max_tokens`). Missing keys fail at Talker init; there is no tts_config fallback.
- Sets the codec budget to `min(yaml_max_tokens, context_len - prompt_len)` and clamps `min_tokens` to that budget so EOS stays reachable.
- Warns when the leftover window is below 64 codec steps (not recoverable without a longer `max_model_len` or a shorter Thinker text).
- Deep-merges `codec_sampling_params` so duplex / replica overlays inherit the base YAML.
- Raises stage-1 `default_sampling_params.max_tokens` from 2048 to 4096 to match Talker context.

Also in this commit, because they are required for the same MiniCPM-o path:

- **NPU duplex sampling hook.** GPU and NPU AR runners are siblings; NPU `_sample` had copied the GPU path without `prepare_duplex_sampling`, so `_minicpmo45_duplex_row_sessions` stayed empty on Ascend. The hook lives in `DuplexSamplingRunnerMixin` and is applied before `model_sample(...)`.
- **NPU nightly MiniCPM-o perf.** A3 / A2 Seed-TTS perf jobs and A3 duplex Seed-TTS perf, plus `disable_shuffle: true`.
- **Long-form e2e.** `test_text_to_audio_long_form_001` asks for a 300-word story and asserts ≥ 200 words (the regression that exposed the mid-utterance cut).

**Breaking change for out-of-tree MiniCPM-o deploy YAMLs:** stage 1 must set a complete `codec_sampling_params` block. In-tree overlays (`minicpmo_4_5_duplex.yaml` and the replica YAMLs) inherit it from `minicpmo_4_5.yaml`.

## Test Plan

**vLLM Version:** 0.26.0

**vLLM-Omni Commit:** `8cbc247010c0b55246cad010fbe2788bd4a1dae3`

Unit / config (CPU):

```bash
pytest -q tests/model_executor/models/minicpmo_4_5/test_talker_batching.py
pytest -q tests/worker/test_native_duplex_hooks.py
pytest -q tests/config/test_config_factory.py tests/config/test_omni_config.py
pytest -q tests/worker/

../.venv/lib/python3.12/site-packages/torch/jit/_script.py:365: 14 warnings
  /data/why/.venv/lib/python3.12/site-packages/torch/jit/_script.py:365: DeprecationWarning: `torch.jit.script_method` is deprecated. Please switch to `torch.compile` or `torch.export`.
    warnings.warn(

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
--- Running Summary
=============================== 2 passed, 1 deselected, 16 warnings in 790.86s (0:13:10) ================================
```

### DIFF
--- a/tests/e2e/online_serving/test_minicpmo_4_5_expansion.py
+++ b/tests/e2e/online_serving/test_minicpmo_4_5_expansion.py
@@ -1,7 +1,7 @@
 """E2E online expansion tests for MiniCPM-o 4.5.
 
 These cover modality combinations, separate Code2Wav behavior, sequential
-request isolation, longer audio output, and Chinese speech.
+request isolation, longer audio output, long-form generation, and Chinese speech.
 
 Like Qwen3-Omni expansion, ``default`` exercises ``--no-async-chunk`` and
 ``async_chunk`` exercises ``--async-chunk``.
@@ -75,6 +75,15 @@ def get_max_batch_size(size_type="few"):
     return batch_sizes.get(size_type, 5)
 
 
+# Close <think> and emit <|tts_bos|> so Talker speaks the answer, not reasoning.
+_TTS_EXTRA_BODY = {
+    "chat_template_kwargs": {
+        "use_tts_template": True,
+        "enable_thinking": False,
+    }
+}
+
+
 @hardware_test(res={"cuda": "H100", "npu": "A3"}, num_cards=1)
 @pytest.mark.parametrize("omni_server", test_params, indirect=True)
 def test_text_video_to_text_001(omni_server, openai_client) -> None:
@@ -127,6 +136,7 @@ def test_sequential_requests_independent(omni_server, openai_client) -> None:
             "model": omni_server.model,
             "messages": messages_1,
             "stream": True,
+            "extra_body": _TTS_EXTRA_BODY,
         },
         request_num=1,
     )
@@ -138,6 +148,7 @@ def test_sequential_requests_independent(omni_server, openai_client) -> None:
             "messages": messages_2,
             "stream": True,
             "key_words": {"text": ["Beijing"]},
+            "extra_body": _TTS_EXTRA_BODY,
         },
         request_num=1,
     )
@@ -164,18 +175,37 @@ def test_text_to_audio_long_output_001(omni_server, openai_client) -> None:
         "messages": messages,
         "modalities": ["text", "audio"],
         "stream": True,
-        # Delimit the assistant answer with the TTS template so the Talker
-        # does not spend its codec-token budget speaking hidden reasoning.
-        "extra_body": {
-            "chat_template_kwargs": {
-                "use_tts_template": True,
-                "enable_thinking": False,
-            }
-        },
+        "extra_body": _TTS_EXTRA_BODY,
         "key_words": {"audio": ["Beijing"]},
     }
 
     openai_client.send_omni_request(request_config, request_num=get_max_batch_size())
+
+
+@hardware_test(res={"cuda": "H100", "npu": "A3"}, num_cards=1)
+@pytest.mark.parametrize("omni_server", test_params, indirect=True)
+def test_text_to_audio_long_form_001(omni_server, openai_client) -> None:
+    """
+    Input Modal: text only (long-form generation prompt).
+    Output Modal: text, audio (default ``modalities``);
+    Input Setting: stream=True
+    Datasets: single request
+    """
+    messages = dummy_messages_from_mix_data(
+        system_prompt=get_system_prompt(),
+        content_text="帮我讲一个300字的故事.",
+    )
+
+    request_config = {
+        "model": omni_server.model,
+        "messages": messages,
+        "stream": True,
+        "extra_body": _TTS_EXTRA_BODY,
+    }
+    responses = openai_client.send_omni_request(request_config, request_num=get_max_batch_size())
+    text = responses[0].text_content if responses else ""
+    word_count = len(text.split())
+    assert word_count >= 200, f"Expected at least 200 words in long output, got {word_count}"
 
 
 @hardware_test(res={"cuda": "H100", "npu": "A3"}, num_cards=1)
@@ -193,5 +223,6 @@ def test_chinese_text_to_audio(omni_server, openai_client) -> None:
         "messages": messages,
         "stream": True,
         "key_words": {"text": ["北京"]},
+        "extra_body": _TTS_EXTRA_BODY,
     }
     openai_client.send_omni_request(request_config)
--- a/vllm_omni/deploy/minicpmo_4_5.yaml
+++ b/vllm_omni/deploy/minicpmo_4_5.yaml
@@ -64,15 +64,15 @@ stages:
     devices: "0"
     output_connectors:
       to_stage_2: connector_of_shared_memory
-    # Codec sampling comes from the checkpoint's tts_config. These parameters
-    # govern only vLLM's binary continue/stop token.
+    # Talker is a single-vocab codec LM. These knobs go to vLLM's Sampler
+    # (head_code logits). EOS is tts_config.eos_token_id (num_audio_tokens-1).
     default_sampling_params:
-      temperature: 0.0
-      top_p: 1.0
-      top_k: -1
+      temperature: 0.8
+      top_p: 0.8
+      top_k: 100
+      repetition_penalty: 1.02
       min_tokens: 50
-      max_tokens: 2048
-      stop_token_ids: [1]
+      max_tokens: 4096
       seed: 42
       detokenize: false
 
--- a/vllm_omni/model_executor/models/minicpmo_4_5/batched_token2wav.py
+++ b/vllm_omni/model_executor/models/minicpmo_4_5/batched_token2wav.py
@@ -19,6 +19,65 @@ from .cuda_graph_wrapper import CFMGraphWrapper, HiFTGraphWrapper
 logger = init_logger(__name__)
 
 _SILENCE_TOKEN = 4218
+# CosyVoice2 RelPos PE is built for max_len=5000 and `forward_chunk` never
+# calls ``extend_pe``. After the 2x upsample a single Code2Wav prefill longer
+# than this many codec tokens overflows the PE slice (matrix_ac vs matrix_bd).
+_DEFAULT_RELPOS_MAX_POS = 5000
+_MAX_ENCODE_TOKEN_CAP = 1024
+
+
+def relpos_encode_token_budget(
+    *,
+    max_pos: int,
+    stride: int,
+    cache_offset: int,
+    lookahead: int,
+    cap: int = _MAX_ENCODE_TOKEN_CAP,
+) -> int:
+    """How many codec tokens ``forward_chunk`` can take before RelPos PE wraps.
+
+    ``position_encoding(size)`` needs ``size <= max_pos``. After upsample,
+    ``size = cache_offset * stride + ~stride * token_frames`` (plus last-chunk
+    lookahead pad).
+    """
+    stride = max(1, int(stride))
+    lookahead = max(0, int(lookahead))
+    room = int(max_pos) // stride - max(0, int(cache_offset)) - lookahead - 1
+    return max(lookahead + 1, min(int(cap), room))
+
+
+def plan_token2wav_encode_slices(
+    num_frames: int,
+    *,
+    max_frames: int,
+    min_nonfinal: int,
+    last_chunk: bool,
+) -> list[tuple[int, int]]:
+    """Split a Code2Wav token span so every non-final piece has a lookahead window."""
+    if num_frames <= 0:
+        return []
+    min_nonfinal = max(1, int(min_nonfinal))
+    max_frames = max(int(max_frames), min_nonfinal)
+    last_min = 1 if last_chunk else min_nonfinal
+    if num_frames <= max_frames:
+        return [(0, num_frames)]
+
+    slices: list[tuple[int, int]] = []
+    start = 0
+    while start < num_frames:
+        remaining = num_frames - start
+        if remaining <= max_frames:
+            slices.append((start, num_frames))
+            break
+        take = max_frames
+        tail = remaining - take
+        if tail <= max_frames and tail < last_min:
+            take = remaining - last_min
+        if take < min_nonfinal:
+            take = remaining
+        slices.append((start, start + take))
+        start += take
+    return slices
 
 
 def _autocast_disabled(device: torch.device):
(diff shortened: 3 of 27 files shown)


amy-why-3459:
@natureofnature @R2-Y @y-null PTAL

tests/e2e/online_serving/test_minicpmo_4_5_expansion.py
```diff
@@ -178,6 +178,39 @@ def test_text_to_audio_long_output_001(omni_server, openai_client) -> None:
...
+            "chat_template_kwargs": {
+                "use_tts_template": True,
+                "enable_thinking": False,
+            }
+        },
+    }
+    responses = openai_client.send_omni_request(request_config, request_num=get_max_batch_size())
+    text = responses[0].text_content if responses else ""
+    word_count = len(text.split())
+    assert word_count >= 200, f"Expected at least 200 words in long output, got {word_count}"
```
R2-Y, line 208: Maybe add assertions for non-empty audio and minimum duration or ASR output. Text length alone does not catch Talker or Code2Wav truncation.

vllm_omni/model_executor/models/minicpmo_4_5/batched_token2wav.py
```diff
@@ -508,14 +603,68 @@ def decode_batch(
...
             if num_frames <= lookahead:
                 raise RuntimeError(
                     "MiniCPMO45Code2WavBatchError "
                     f'{{"reason":"chunk_below_lookahead_window","frames":{num_frames},'
                     f'"minimum":{lookahead + 1}}}'
                 )
+        # A non-async Talker dump can land thousands of codec tokens in one
+        # Code2Wav prefill. CosyVoice RelPos PE (max_len=5000) plus 2x upsample
+        # cannot score that in one ``forward_chunk`` (6968 vs 985 on NPU).
+        max_frames = self._max_encode_token_frames(states)
```
R2-Y, line 617: Each decode grows the cache, but max_frames is calculated only once. Recompute the safe frame count after every slice and test progressively shrinking cache room.

tests/e2e/online_serving/test_minicpmo_4_5_expansion.py
amy-why-3459: > Maybe add assertions for non-empty audio and minimum duration or ASR output. Text length alone does not catch Talker or Code2Wav truncation.

The extra word_count >= 200 check is text-only, agreed — it would not catch a Talker/Code2Wav cut by itself.

send_omni_request already runs assert_omni_response. Default modalities are ["text", "audio"], so that helper already requires non-empty audio and a Whisper transcript whose cosine similarity must match the generated text. A mid-utterance codec cut produces a short transcript and fails that check. Same pattern as test_text_to_audio_long_output_001 (key_words.audio).

I’d rather not add a second ASR keyword list on a 300-word story (flaky) or stack another Whisper pass. Leaving the word-count assert as a Thinker-length smoke check on top of the existing audio/ASR gate.

vllm_omni/model_executor/models/minicpmo_4_5/batched_token2wav.py
amy-why-3459: > Each decode grows the cache, but max_frames is calculated only once. Recompute the safe frame count after every slice and test progressively shrinking cache room.

relpos_encode_token_budget does shrink with cache_offset, but decode_batch does not keep the full growing cache. After each slice we trim conformer_att back to prompt_len + 100, so the offset is effectively bounded, not monotonic in total tokens. Combined with _MAX_ENCODE_TOKEN_CAP = 1024 and _ensure_relpos_pe before every forward_chunk, the one-shot slice plan stays inside RelPos for MiniCPM prompt sizes. Dropping room below the 1024 cap would need a ~2800-frame prompt.

test_relpos_encode_token_budget_leaves_room_for_upsample_and_cache already covers the offset term; a “progressively shrinking cache” loop would not match the trimmed cache we actually feed the next slice. If we ever remove that trim, we should switch to a greedy recompute. I don’t think we should do that in this PR.

vllm_omni/deploy/minicpmo_4_5.yaml
```diff
@@ -64,15 +64,15 @@ stages:
...
-      top_p: 1.0
-      top_k: -1
+      temperature: 0.8
+      top_p: 0.8
+      top_k: 100
+      repetition_penalty: 1.02
       min_tokens: 50
-      max_tokens: 2048
-      stop_token_ids: [1]
+      max_tokens: 4096
```
natureofnature, line 75: [P1] `_max_audio_tokens()` currently has no production caller. Because this YAML explicitly sets `max_tokens=4096`, `build_engine_core_request_from_tokens()` does not derive the remaining context budget (it only does so when `max_tokens is None`). For a prompt of length P, the scheduler can therefore still attempt 4096 output tokens in a 4096-token context, and `min_tokens=50` can keep EOS masked when fewer than 50 positions remain. Please apply the prompt-aware budget to the stage `SamplingParams` before request construction and clamp `min_tokens` to the effective budget.

vllm_omni/deploy/minicpmo_4_5.yaml
```diff
@@ -64,15 +64,15 @@ stages:
...
+    # Talker is a single-vocab codec LM. These knobs go to vLLM's Sampler
+    # (head_code logits). EOS is tts_config.eos_token_id (num_audio_tokens-1).
     default_sampling_params:
-      temperature: 0.0
-      top_p: 1.0
-      top_k: -1
+      temperature: 0.8
+      top_p: 0.8
+      top_k: 100
+      repetition_penalty: 1.02
```
natureofnature, line 73: [P2] The Talker scheduler prompt is `[0] * condition_length`, while vLLM's repetition penalty treats tokens appearing in either the prompt or output as already seen. This therefore penalizes valid codec id 0 from the first sampling step, unlike the previous codec-history-only behavior. Please exclude scheduler placeholder IDs from the penalty state or preserve an output-code-history-only repetition penalty.

natureofnature (changes requested):
Requesting changes for two issues in the Talker sampling migration: the prompt-aware output budget is currently dead code, leaving the context-truncation path unfixed, and vLLM's repetition penalty is seeded by the synthetic prompt token 0. Details inline.

vllm_omni/deploy/minicpmo_4_5.yaml
amy-why-3459: > [P1] `_max_audio_tokens()` currently has no production caller. Because this YAML explicitly sets `max_tokens=4096`, `build_engine_core_request_from_tokens()` does not derive the remaining context budget (it only does so when `max_tokens is None`). For a prompt of length P, the scheduler can therefore still attempt 4096 output tokens in a 4096-token context, and `min_tokens=50` can keep EOS masked when fewer than 50 positions remain. Please apply the prompt-aware budget to the stage `SamplingParams` before request construction and clamp `min_tokens` to the effective budget.

Agreed — `_max_audio_tokens()` never reached production after the Sampler migration, and YAML `max_tokens=4096` skipped the remaining-context derivation.

The stall path is the one that matters. `check_stop` returns early while `min_tokens` is unmet, so it never reaches the length cap; once the leftover window is smaller than YAML `min_tokens=50`, EOS stays masked and the scheduler can grant no further tokens.

`build_engine_core_request_from_tokens` now clamps `min_tokens` to `max(0, max_model_len - prompt_len)` for every stage. We did not also clamp `max_tokens`: vLLM already stops on `num_tokens >= max_model_len`, so rewriting the YAML ceiling would only change the finish reason. `_max_audio_tokens` itself is gone.

Covered in `tests/engine/test_stage_request_token_budget.py`.

vllm_omni/deploy/minicpmo_4_5.yaml
amy-why-3459: > [P2] The Talker scheduler prompt is `[0] * condition_length`, while vLLM's repetition penalty treats tokens appearing in either the prompt or output as already seen. This therefore penalizes valid codec id 0 from the first sampling step, unlike the previous codec-history-only behavior. Please exclude scheduler placeholder IDs from the penalty state or preserve an output-code-history-only repetition penalty.

Agreed. The scheduler prompt is not codec history — `llm2tts` fills it with `0`s on the handoff path, or with thinker token ids on the non-handoff path — but vLLM's repetition penalty scores both prompt and output.

`Talker.sample()` now replaces the entire prompt row with the penalty pad (`vocab_size`) before calling the Sampler, via `dataclasses.replace` so the runner's shared `SamplingMetadata` is not mutated. That restores the old codec-history-only penalty and also covers the non-handoff ids.

Covered in `test_talker_batching.py`.

natureofnature (approved):
LGTM

R2-Y:
LGTM. Thanks

Gaohan123 (approved):
LGTM. Thanks

PR #6279, opened 2026-08-17
[Core] Support online FP8 with DLO AllGather

hsliuustc0106:
## Purpose

Allow per-tensor online FP8 linear layers to run with distributed layerwise
offload's AllGather path.

The previous compatibility gate rejected every online quantization method.
However, the ordinary loader already finalizes `Fp8PerTensorOnlineLinearMethod`
weights and scales before DLO initialization, and the DLO backend records and
reconstructs each runtime tensor's dtype, shape, and stride. This makes the
per-tensor FP8 layout compatible with the existing sharding path, including the
transposed Cutlass weight layout.

This PR narrows the gate to allow that validated method while keeping all other
online quantization methods fail-closed. It also documents the startup-memory
tradeoff: direct checkpoint mmap remains unavailable, so each rank temporarily
materializes the complete FP8 model in host memory before retaining its DLO
shard.

## Test Plan

- Run the diffusion loader unit tests.
- Run the distributed layerwise offload backend unit tests.
- Run all pre-commit hooks on the changed files.
- Exercise the real DLO hook with two local NCCL ranks using a transposed FP8
  weight and FP32 scale.

**vLLM Version:** 0.27.0

**vLLM-Omni Commit:** 265d80b4dd87f86f7cf5cbde07d248855383ce90

## Test Result

- `tests/diffusion/model_loader/test_diffusers_loader.py`: 15 passed
- `tests/diffusion/offloader/test_distributed_layerwise_backend.py`: 61 passed
- Pre-commit: passed
- Two-rank NCCL FP8 AllGather/DLO hook smoke: passed on 2 NVIDIA L20X GPUs

A full MiniMax-H3 end-to-end run was not performed because only one GPU was
available during validation.

PR #6348, opened 2026-08-19
[Diffusion] Add normalized online FP8 cache for DLO AllGather

hsliuustc0106:
## Summary

Add the Phase-I normalized online-FP8 runtime cache for DLO AllGather.

On a cache miss, the existing ordinary online-FP8 loader finalizes the transformer weights and scales, then publishes an atomic serialized safetensors cache. On a valid cache hit, the loader selects that serialized source before DLO planning, allowing the transformer to use the direct checkpoint-mmap path without re-materializing the BF16 source.

The cache is intentionally scoped to:

- per-tensor online FP8;
- one top-level `transformer` component;
- TP=1;
- DLO AllGather;
- no HSDP and no resident-layer/no-AllGather path.

Invalid, stale, incomplete, or incompatible artifacts fail closed as cache misses. Cache publication failures retain the ordinary-loader path.

## Motivation

PR #6279 enables online FP8 with DLO AllGather, but the ordinary-loader fallback still materializes the complete runtime model before DLO retains each rank's shard. This follow-up removes that repeated warm-start materialization for the normalized transformer representation while preserving the existing DLO layout contract.

## Implementation

- Add `--dlo-fp8-cache-dir` through CLI, stage, engine, and diffusion configuration.
- Add source-fingerprint and manifest validation.
- Serialize finalized FP8 weights, generated scales, and persistent transformer buffers atomically.
- Reuse the existing loader-owned direct-mmap plan on cache hits.
- Finalize serialized FP8 parameters after mmap replacement without CUDA re-quantization.
- Preserve MiniMax-H3's post-transform QKV storage contract.
- Remove the temporary memory-profiler instrumentation from the production DLO backend.
- Document Phase-I scope and the broader compatibility follow-up in RFC #6231.

### Warm normalized-cache smoke

The normalized cache was built from the MiniMax-H3 Ref2VA transformer and contained **795 tensors** backed by **13 source safetensors files**. Subsequent launches logged a cache hit on every DLO rank and served two-step requests successfully.

Workload: vLLM 0.27.0, CUDA 12.9, four NVIDIA H100 GPUs, MiniMax-H3 Ref2VA, CUDNN attention, 256×256, two denoising steps, zero resident layers. Throughput/device is completed requests divided by wave wall time and four GPUs. Host values are peak process-tree RSS/PSS.

| Warm cache layout | Requests | Wave latency | Throughput/device | Host RSS peak | Host PSS peak | Peak GPU/device |
| DP2/SP2 | 2 | 650.88 s | 0.000768 req/s | 236.36 GiB | 231.44 GiB | 22.16 GiB |
| DP4/SP1 | 4 | 1198.89 s | 0.000834 req/s | 213.00 GiB | 187.48 GiB | 21.84 GiB |

All six requests returned HTTP 200. The server logs explicitly reported `DLO normalized FP8 cache hit` and `keeping model initialization on CPU`. These are warm-cache systems smoke results, not a cold-start speed comparison or a production-quality benchmark.

### Short memory-only comparison

For the same four-worker setup, the aggregate request-phase PSS was:

| Path | DP2/SP2 | DP4/SP1 |
|---|---:|---:|
| Ordinary online FP8 | 188.71 GiB | 145.07 GiB |
| Normalized FP8 cache hit | 213.33 GiB | 171.85 GiB |
| Native BF16 | 318.48 GiB | 222.93 GiB |

The cache-hit rows are higher than ordinary FP8 in this prototype because Phase I normalizes the transformer only; the text encoder and other components remain BF16, and file-backed cache pages are visible in RSS/PSS. This confirms the cache removes the transformer's repeated BF16 materialization but does not yet guarantee lower whole-process PSS.

## Validation

- Focused loader/DLO/cache/config suite: **115 passed, 17 warnings**.
- Non-network FP8 configuration suite: **23 passed, 4 deselected**.
- Pre-commit hooks: passed (ruff, formatting, typo checks, test markers).
- `git diff --check`: passed.

The four deselected integration cases require the uncached Hugging Face model ID `test` while this environment is offline; they are unrelated to the cache changes.

## Follow-ups

This PR does not generalize cache sharing to other components, TP coordinates, or broader DP/SP compatibility. Those contracts remain tracked in [RFC #6231](https://github.com/vllm-project/vllm-omni/issues/6231).

### DIFF
--- a/vllm_omni/diffusion/model_loader/diffusers_loader.py
+++ b/vllm_omni/diffusion/model_loader/diffusers_loader.py
@@ -8,7 +8,7 @@ import re
 import time
 from collections.abc import Generator, Iterable, Sequence
 from pathlib import Path
-from typing import cast
+from typing import Any, cast
 
 import torch
 from torch import nn
@@ -33,6 +33,12 @@ from vllm_omni.diffusion.data import OmniDiffusionConfig
 from vllm_omni.diffusion.distributed.hsdp import HSDPInferenceConfig, apply_hsdp_to_model
 from vllm_omni.diffusion.model_loader.checkpoint_adapters import (
     get_checkpoint_adapter,
+)
+from vllm_omni.diffusion.model_loader.fp8_cache import (
+    _transformer_quant_config,
+    build_normalized_fp8_cache,
+    is_online_fp8_config,
+    load_fp8_cache_manifest,
 )
 from vllm_omni.diffusion.model_loader.host_weight_plan import (
     HostWeightPlan,
@@ -141,6 +147,10 @@ class DiffusersPipelineLoader:
         self.quant_config = od_config.quantization_config
         self.parallel_config = od_config.parallel_config
         self.host_weight_plan: HostWeightPlan | None = None
+        self._fp8_cache_dir: Path | None = None
+        self._fp8_cache_requested = False
+        self._fp8_cache_hit = False
+        self._fp8_cache_manifest: dict[str, Any] | None = None
 
     def take_host_weight_plan(self) -> HostWeightPlan | None:
         """Transfer the loader-produced plan to the offload backend."""
@@ -367,11 +377,60 @@ class DiffusersPipelineLoader:
         )
 
     def _get_weight_sources(self, model: nn.Module) -> tuple["ComponentSource", ...]:
-        return tuple(
+        sources = tuple(
             cast(
                 Iterable[DiffusersPipelineLoader.ComponentSource],
                 getattr(model, "weights_sources", ()),
             )
+        )
+        if not self._fp8_cache_hit or self._fp8_cache_dir is None or self._fp8_cache_manifest is None:
+            return sources
+
+        cache_prefix = str(self._fp8_cache_manifest["source_prefix"])
+        cache_subfolder = str(self._fp8_cache_manifest.get("component_subfolder", "transformer"))
+        return tuple(
+            dataclasses.replace(
+                source,
+                model_or_path=str(self._fp8_cache_dir),
+                subfolder=cache_subfolder,
+                revision=None,
+            )
+            if source.prefix == cache_prefix
+            else source
+            for source in sources
+        )
+
+    def _prepare_fp8_cache(self) -> None:
+        """Select the Phase-I cache path before constructing the pipeline."""
+        cache_dir = getattr(self.od_config, "dlo_fp8_cache_dir", None)
+        if not cache_dir or not is_online_fp8_config(self.quant_config):
+            return
+        if not getattr(self.od_config, "enable_distributed_layerwise_offload", False):
+            logger.warning("Ignoring --dlo-fp8-cache-dir because distributed layerwise offload is disabled")
+            return
+        if not getattr(self.od_config, "dlo_use_allgather", True):
+            logger.warning("Ignoring --dlo-fp8-cache-dir because DLO AllGather is disabled")
+            return
+        if int(getattr(self.parallel_config, "tensor_parallel_size", 1)) != 1:
+            logger.warning("Ignoring --dlo-fp8-cache-dir because Phase-I cache requires TP=1")
+            return
+        if bool(getattr(self.parallel_config, "use_hsdp", False)):
+            logger.warning("Ignoring --dlo-fp8-cache-dir because Phase-I cache does not support HSDP")
+            return
+
+        self._fp8_cache_dir = Path(cache_dir)
+        self._fp8_cache_requested = True
+        self._fp8_cache_manifest = load_fp8_cache_manifest(self._fp8_cache_dir)
+        if self._fp8_cache_manifest is None:
+            logger.info("DLO normalized FP8 cache miss: %s", self._fp8_cache_dir)
+            return
+
+        self._fp8_cache_hit = True
+        self.quant_config = _transformer_quant_config()
+        self.od_config.quantization_config = self.quant_config
+        logger.info(
+            "DLO normalized FP8 cache hit: %s; selecting serialized FP8 checkpoint path",
+            self._fp8_cache_dir,
         )
 
     def _get_expected_parameter_names(self, model: nn.Module) -> set[str]:
@@ -403,6 +462,7 @@ class DiffusersPipelineLoader:
         self.host_weight_plan = None
         if load_format is None:
             load_format = "default"
+        self._prepare_fp8_cache()
         # CPU offload + quantization: for offline-quantized models (e.g., AutoRound MXFP8),
         # weights are already quantized in the checkpoint — load directly on CPU.
         # For online quantization, load on device so quantization can run on accelerator,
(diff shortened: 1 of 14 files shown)


vllm_omni/engine/async_omni_engine.py
```diff
@@ -1046,6 +1046,7 @@ def _create_default_diffusion_stage_cfg(kwargs: dict[str, Any]) -> list:
             "enable_distributed_layerwise_offload": kwargs.get("enable_distributed_layerwise_offload", False),
             "dlo_use_allgather": kwargs.get("dlo_use_allgather", True),
             "dlo_resident_layers": kwargs.get("dlo_resident_layers", 0),
+            "dlo_fp8_cache_dir": kwargs.get("dlo_fp8_cache_dir", None),
```
hsliuustc0106: why we need this

vllm_omni/diffusion/model_loader/diffusers_loader.py
```diff
@@ -367,12 +377,61 @@ def offload_completed() -> None:
...
+
+        self._fp8_cache_dir = Path(cache_dir)
+        self._fp8_cache_requested = True
+        self._fp8_cache_manifest = load_fp8_cache_manifest(self._fp8_cache_dir)
+        if self._fp8_cache_manifest is None:
+            logger.info("DLO normalized FP8 cache miss: %s", self._fp8_cache_dir)
+            return
+
+        self._fp8_cache_hit = True
+        self.quant_config = _transformer_quant_config()
```
hsliuustc0106: **P1: cache-hit config swap silently drops sibling/global component quantization.** `_transformer_quant_config()` returns a config with only a `transformer` entry (no default, no siblings) and replaces the whole `od_config.quantization_config`. With the documented plain `--quantization fp8`, a plain `Fp8Config` resolves for *every* component on the miss path (`component_config.py:42-44`), but after the swap `resolve("text_encoder")` returns `None` — e.g. MiniMax-H3/Flux2 text encoders are FP8 on the cold launch and BF16 on warm launches. An explicit per-component dict (`{"transformer": fp8, "vae": ...}`) loses its non-transformer entries the same way. This contradicts the user-guide claim that other components keep their ordinary quantization, and `_get_source_quant_config` (`:296-300`) changes adapter selection for non-transformer sources too. Note `vllm_config.quant_config` (`diffusion/vllm_config.py:156`) still holds the old object, so the two views diverge. Smallest fix: build the replacement from the existing component mapping, overriding only the `transformer` entry (preserving siblings and `default_config`) so both views stay consistent.

vllm_omni/diffusion/model_loader/diffusers_loader.py
```diff
@@ -367,12 +377,61 @@ def offload_completed() -> None:
...
         )
+        if not self._fp8_cache_hit or self._fp8_cache_dir is None or self._fp8_cache_manifest is None:
+            return sources
+
+        cache_prefix = str(self._fp8_cache_manifest["source_prefix"])
+        cache_subfolder = str(self._fp8_cache_manifest.get("component_subfolder", "transformer"))
+        return tuple(
+            dataclasses.replace(
+                source,
+                model_or_path=str(self._fp8_cache_dir),
```
hsliuustc0106: **P2: a manifest-valid hit whose mmap plan falls back loads the cache as an ordinary source instead of failing closed.** On a hit, `_get_weight_sources` rewrites the transformer source to the cache dir before plan construction, so if `build_checkpoint_mmap_plan` returns a fallback reason (e.g. a remapping pipeline where published buffer/scale keys do not round-trip through `_remap_ckpt_key`), the ordinary `load_weights` path consumes the cache artifact rather than the original model — and `_process_weights_after_loading` then runs the serialized requant path (CUDA-only `scaled_fp8_quant`) on CPU tensors. Manifest validation does not prove plan compatibility; a hit whose plan is incompatible should degrade to the original source (true cache miss), not to an ordinary load of the cache.

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
```diff
@@ -1103,6 +1106,52 @@ def _load_weights_via_mmap(
                 f"(first 5: {remaining_meta[:5]})."
             )
 
+    @staticmethod
+    def _process_normalized_fp8_cache(pipeline: nn.Module) -> None:
```
hsliuustc0106: **P2: finalizer can run CUDA-only marlin repack on CPU mmap tensors.** On sm80/86-class GPUs where cutlass FP8 scaled-mm is unavailable and a marlin kernel is selected, `quant_method.fp8_linear.process_weights_after_loading` performs `pack_fp8_to_int32` + `gptq_marlin_repack` against the CPU mmap views (cache hits keep init on CPU), so the warm boot crashes. Consider gating the cache hit away from marlin kernels or documenting the constraint. Two related notes: (1) the docstring claim "The ordinary loader normally runs this method before DLO owns the plan" is inverted for plan paths — with a host-weight plan the loader skips `_process_weights_after_loading` entirely (`diffusers_loader.py:526-549` vs `:584`), which is why this finalizer is needed at all; (2) pre-existing gap next to this code: gating on `_dlo_fp8_cache_normalized` leaves an ordinary serialized-FP8 checkpoint that passes plan preflight with no finalizer anywhere — keying off `isinstance(quant_method, Fp8LinearMethod)` over mmap-loaded modules would close both.

vllm_omni/diffusion/model_loader/diffusers_loader.py
```diff
@@ -367,12 +377,61 @@ def offload_completed() -> None:
...
+                subfolder=cache_subfolder,
+                revision=None,
+            )
+            if source.prefix == cache_prefix
+            else source
+            for source in sources
+        )
+
+    def _prepare_fp8_cache(self) -> None:
```
hsliuustc0106: **P2: the behavior-changing paths beyond manifest validation have no test coverage.** `test_fp8_cache.py` covers manifest validation and publication, but nothing exercises `_prepare_fp8_cache` gating (DLO off / AllGather off / TP>1 / HSDP each ignored with a warning), the `_get_weight_sources` rewrite on hit, or `_process_normalized_fp8_cache` (transpose + scale collapse + non-uniform-scale rejection — all CPU-testable with a stubbed kernel/quant_method). These are the paths where a regression silently changes numerics rather than failing loudly.

vllm_omni/diffusion/model_loader/fp8_cache.py
```diff
@@ -0,0 +1,349 @@
...
+            encoding="utf-8",
+        )
+        (tmp_root / CACHE_COMPLETE_NAME).write_text("phase1\n", encoding="utf-8")
+        for handle in file_cache.values():
+            close = getattr(handle, "__exit__", None)
+            if callable(close):
+                close(None, None, None)
+        try:
+            os.replace(tmp_root, cache_root)
+        except FileExistsError:
```
hsliuustc0106: **P3: the concurrent-publisher race catches the wrong errno.** On Linux, `rename(2)` of a directory onto an existing non-empty directory returns `ENOTEMPTY`, which maps to plain `OSError` — not `FileExistsError` — so the `except FileExistsError` branch at the `os.replace` is effectively dead and the losing worker logs the "cache build failed" warning with a traceback instead of the intended "published concurrently" info. Broaden the except to `OSError` with `errno in (EEXIST, ENOTEMPTY)` (or re-check the published manifest first).

PR #6354, opened 2026-08-19
[Perf][NemotronVoiceChat] Fast defaults for all three deploy profiles: 2.8x offline, realtime full-duplex audio

yuekaizhang:
> **Rebased on `main`** (post-#6089). The final #6089 revision independently
> converged on part of our earlier work (per-wake drain, duplex segment-stop
> semantics, per-request codec caches); this PR keeps upstream's versions of
> those and adds what is new.

## What

Performance work on the NemotronVoiceChat-11B pipeline (thinker → talker →
code2wav) with **zero new deploy yamls**: the three existing profiles now
default to the fast execution, and every fast setting carries a `PARITY:`
comment whose edits restore the eager fp32 NeMo-bit-exact reference
(verified: the md5-anchored acceptance WAV reproduces after applying them).

- **`nemotron_labs_voicechat.yaml` (offline sync).** bf16 + CUDA-graph
  thinker; the talker's Gemma3 backbone runs as a native vLLM model
  (`hf_overrides.use_native_talker`): PagedAttention KV (per-step attention
  cost scales with the actual session length), vLLM's own decode CUDA
  graphs, MoG sampling captured into one graph, multi-session batching.
  Fused input embeddings are verified against the reference forward to
  ~4e-6; the reference's `inference_force_speech_silence_on_eos` code
  feedback is applied.
- **`nemotron_labs_voicechat_streaming.yaml` (offline streaming).** bf16 +
  CUDA-graph thinker, the vendored per-frame EAR-TTS step captured into CUDA
  graphs (`use_talker_cuda_graphs`), 5-frame codec chunks decoded
  incrementally with a per-request causal-conv cache.
- **`nemotron_labs_voicechat_duplex.yaml` (realtime duplex).** Native talker
  + bf16 thinker: unbounded session length, one 80 ms packet per frame.
- **Optional classifier-free guidance on the native talker**
  (`hf_overrides.native_talker_guidance`, off by default): the conditional
  stream stays on vLLM paged KV, the unconditional stream mirrors it on the
  reference backbone (captured StaticCache step), blended with the reference
  `generate_step` math. Paired 4-fixture ASR matches the guided reference;
  +~8.5 ms/frame, duplex stays realtime.
- **No environment variables.** The talker prompt geometry comes from config:
  `hf_overrides.talker_init_len` (sync input processor via a new optional
  `next_stage_hf_config` kwarg), `hf_overrides.async_chunk_prewarm_prompt_len`
  (orchestrator async-chunk prewarm), and the connector-extra `native_talker`
  flag (thinker→talker chunk producer).

## Results (1x H100, 196-frame fixture / full turn_taking.wav)

| profile | default (this PR) | PARITY edits (reference) |
|---|---|---|
| sync | **3.7 s, RTF 0.24** | 10.3 s, RTF 0.66 (md5-anchored, reproduced) |
| streaming | **5.2 s total, TTFP 1.47 s** | TTFP 2.6 s |
| duplex | **TTFP 408 ms, one 80 ms packet/frame, median gap 81 ms / p95 109 ms (RTF ~1.01)** | TTFP 712 ms, delivery lags the frame clock |

Multi-session (sync, native talker): 4 concurrent sessions, aggregate RTF
0.12 (~1.9x throughput). bf16 also cuts the thinker weights ~43 GB → ~21 GB.

- **Sub-200ms duplex first-packet latency** (second commit): first input
  frame → first audio event drops 632 ms → ~120 ms on a warm session
  (NVIDIA NIM measures ~316 ms on the same caliber). Matching NeMo's
  realtime pipeline, the talker no longer sees the system prompt (it
  conditions the thinker only), the speaker prefill runs as one chunked
  prefill (`max_num_batched_tokens` 1 → 64, with the thinker→talker
  producer dripping at most one new timeline token per chunk to keep the
  per-frame code feedback strictly sequential), and the MoG sampling
  graph is captured at load time. Transcripts verified identical
  turn-for-turn; steady-state pacing unchanged.
- **Startup warmup for the realtime endpoint** (third commit): a new
  `duplex_session.warmup_frames` knob (default 2 in the duplex yaml, 0 =
  disabled) makes the server open a websocket to its own `/v1/realtime`
  at boot, run a short silent session, and explicitly close it before
  admitting real clients (`/v1/realtime` waits on the warmup event, with
  a 120 s escape hatch). This moves the remaining one-time costs
  (inductor/codec JIT, first code2wav dispatch) out of the first user
  session: first-session first-frame → first-audio drops 2.3 s → ~105 ms,
  matching every later session. Warmup failure is non-fatal — the server
  logs it and serves normally with cold-start costs.
- **Long-session hardening** (fifth commit): live sessions past the
  ~16 s regression fixtures exposed two duplex bugs. (a) The thinker's
  cache-aware perception re-ran the mel featurizer over the entire
  session audio every 80 ms frame (O(n²)); it now keeps a bounded rolling
  window (pre-encode cache + chunk + an 8-column margin that keeps
  reflect-pad/preemphasis boundary effects away from consumed columns),
  unit-tested bit-identical to full-history slicing. (b) A request abort
  racing an already-queued step (or a stage-request re-open) hit a fresh
  model session mid-stream and the resulting ValueError killed the whole
  engine core; perception now restarts from the current frame with a
  warning, and the fused-prompt prefill branch is gated on engine
  position instead of the duplex sequence. Validated with a 4-minute
  3000-frame realtime session (30 coherent responses, zero engine
  errors); the 196-frame probe transcript is unchanged.

## Validation

- PARITY edits applied to the new default sync yaml reproduce the historical
  md5-anchored NeMo-bit-exact WAV exactly.
- Every default profile re-verified end to end: sync 3.72 s / streaming
  5.18 s / duplex probe green (8/8 turns complete, coherent ASR-clean
  transcript, realtime pacing) — all without any environment variables.
- Unit suites green; probe gotchas documented in the recipe (`--max-frames`
  truncation; the done-after-commit gate is timing-sensitive for realtime
  pipelines — `--allow-incomplete-response` re-added for latency runs).

## Notes

- Known limits are documented in the deploy yamls and the recipe (CFG's
  unconditional stream is bucketed at 4096 positions; duplex allows one
  session by default).


Sy0307:
RFC #6494 status note: #6089, the parent native Nemotron serving PR, merged into `main` on 2026-08-26. This PR is now non-draft, but the latest head has a failing `pre-commit` check. Please fix the formatting failure, rebase/remove any duplicated parent changes, and rerun the performance and quality validation before review.

yuekaizhang:
@linyueqian  @Sy0307  Could you help review the PR? Thanks.

Sy0307:
Updated #6494 tracking: this remains non-draft Nemotron performance work with visible checks green but no formal approval. #6816 is closed after #6831 improved timeout diagnostics, but #6831 explicitly did not fix the observed Stage 0 → Stage 1 stall. Please keep that reliability evidence separate from the pinned H100 speed/quality claims when requesting review.

vllm_omni/engine/stage_engine_core_client.py
```diff
@@ -395,17 +396,25 @@ def process_engine_inputs(
...
         if self.custom_process_input_func is not None:
             signature = inspect.signature(self.custom_process_input_func)
-            if len(signature.parameters) >= 4:
+            extra_kwargs: dict[str, Any] = {}
+            if "next_stage_hf_config" in signature.parameters:
+                # Let a processor size the next stage's prompt from that
+                # stage's model config (e.g. a talker whose engine positions
+                # must cover a speaker-prompt prefill).
+                extra_kwargs["next_stage_hf_config"] = self._stage_hf_config
+            if "streaming_context" in signature.parameters:
```
linyueqian: This drops `streaming_context` for the registered MiniCPM-o processor because its fourth parameter is named `_streaming_context`. Without that context, `llm2tts` loses its bridge state and resends cumulative output to TTS on each handoff. Keep positional-arity compatibility or recognize `_streaming_context` while adding `next_stage_hf_config`.

vllm_omni/model_executor/models/nemotron_voicechat/nemotron_voicechat_thinker.py
```diff
@@ -300,12 +331,14 @@ def __init__(self, *, vllm_config: VllmConfig, prefix: str = "") -> None:
...
+        # request id; the runner drives preprocess per request and hands
+        # postprocess each request's own hidden-state slice, and the NemotronH
+        # backbone batches natively on paged KV. Validated at max_num_seqs<=4.
         max_num_seqs = int(getattr(vllm_config.scheduler_config, "max_num_seqs", 1))
-        if max_num_seqs != 1:
-            raise NotImplementedError(
-                f"NemotronVoiceChat thinker supports max_num_seqs=1 only (got {max_num_seqs}); "
-                "multi-request batching of the per-request fusion state is not implemented."
-            )
+        if max_num_seqs > 1:
```
linyueqian, line 340: Removing this guard exposes a shared function token: `make_omni_output` runs `function_head` on `model_outputs[-1:]`, and postprocess feeds that token to every request. Concurrent sessions can receive the last row's function decision and feed it into the next step. Return request-aligned tokens or keep `max_num_seqs=1`.

vllm_omni/model_executor/models/nemotron_voicechat/nemotron_voicechat_talker.py
```diff
@@ -324,14 +415,337 @@ def _init_session(self, request_id: str, info: dict[str, Any], device: torch.dev
...
+                flag[start:end, 0] = 1.0
+            cumulative = torch.cat(session["codes_rows"], dim=0)
+            payload_meta: dict[str, Any] = {
+                # 1-D tensor on purpose: the wire-payload builder indexes
+                # element.shape[0] (see the eager make_omni_output).
+                "nvc_logical_prompt_len": torch.tensor([session["prompt_len"]], dtype=torch.long)
+            }
+            if session["codec_streaming"]:
+                payload_meta["codec_streaming"] = torch.tensor([True], dtype=torch.bool)
+            outputs = {"codes": {"audio": cumulative}, "meta": payload_meta}
```
linyueqian: This loop overwrites `outputs` for each request and returns only the last request's code history. The generic splitter cannot recover the discarded tensors, so async-chunk and per-step transport can send earlier sessions data from the final request. Return a batch-aligned payload, including placeholders for requests with no new codes.

vllm_omni/model_executor/models/nemotron_voicechat/nemotron_voicechat_talker.py
```diff
@@ -324,14 +415,337 @@ def _init_session(self, request_id: str, info: dict[str, Any], device: torch.dev
...
+            raise RuntimeError(
+                f"NemotronVoiceChat native talker was scheduled {span} timeline positions in one "
+                "step; the per-frame code feedback is strictly sequential. The duplex producer "
+                "drips one new timeline token per chunk — more than one schedulable position "
+                "means chunks outpaced engine steps (raise the frame budget or investigate the "
+                "engine stall)."
+            )
+        t = offset - init_len + 1
+        total = int(session["timeline"].numel())
+        if t != int(session["step"]):
```
linyueqian: KV-cache recomputation rewinds engine positions, but this model-side session keeps its advanced step and code state. A full replay can cross the speaker-prefill boundary, while a partial rewind reaches this desynchronization check. Either path can kill the engine under cache pressure. Rebuild the session during recomputation or prevent and document preemption for this path.

vllm_omni/deploy/nemotron_labs_voicechat.yaml
```diff
@@ -49,12 +70,22 @@ stages:
...
+      # this stage (the vendored backbone copy stays loaded). Off by default:
+      # measured ASR-equivalent at this checkpoint's scale.
+      # native_talker_guidance: true
     output_connectors:
       to_stage_2: connector_of_shared_memory
     default_sampling_params:
       # Placeholder-token loop; the stage emits stop token 1 after the last
-      # timeline position. 16383 (not 16384): the request also carries one
-      # placeholder prompt token, and prompt + max_tokens must fit
+      # timeline position. 16383 (not 16384): the request also carries the
```
linyueqian: The native talker prompt is now 37 tokens, but this profile still sets `max_tokens: 16383` with `max_model_len: 16384`. The combined length is 16420, so the shipped default cannot honor its advertised generation budget. Reduce `max_tokens` to 16347 or derive it from the prompt length.

recipes/NVIDIA/NemotronLabs-VoiceChat.md
```diff
@@ -31,7 +31,7 @@ NeMo modules (`nemo_vendored/`), so no `nemo_toolkit` install is needed.
   [`examples/offline_inference/nemotron_voicechat/end2end.py`](../../examples/offline_inference/nemotron_voicechat/end2end.py)
 - Model modules (thinker / talker / code2wav / vendored NeMo):
   [`vllm_omni/model_executor/models/nemotron_voicechat/`](../../vllm_omni/model_executor/models/nemotron_voicechat/)
-- Staged pipeline config:
+- Staged pipeline config (NeMo bit-parity path):
```
linyueqian: This calls the YAML the bit-parity path, but the file now defaults to the fast, non-bit-exact setup with bfloat16 and the native talker. The table still says fp32, and the prompt-length note still says one placeholder instead of `talker_init_len: 37`. Call this the fast default, list the parity edits explicitly, and update the stale details.

linyueqian (changes requested):
Request changes. The new multi-session path shares function output and drops earlier requests' audio codes, so concurrent sessions can contaminate each other. The dispatch change also breaks MiniCPM-o streaming state, native talker sessions can fail after KV-cache recomputation, and the shipped talker budget no longer fits its larger prompt. The recipe's parity instructions are stale.

Validation: static review of the full diff at `ebc5126e`, relevant call sites, output splitting, and current CI. I did not run code from the untrusted fork.

vllm_omni/engine/stage_engine_core_client.py
yuekaizhang: Fixed in 5f0f99ec: the dispatch now also recognizes the underscore-prefixed `_streaming_context` parameter name, so MiniCPM-o's `llm2tts` keeps receiving its bridge state. I kept name-based matching rather than restoring positional arity: Nemotron's `thinker2talker_token_only` also has four parameters with `next_stage_hf_config` in the fourth slot, so an arity-4 rule would pass the streaming context into that argument.

vllm_omni/model_executor/models/nemotron_voicechat/nemotron_voicechat_thinker.py
yuekaizhang: Fixed in 5f0f99ec: `make_omni_output` now emits the shared `nvc_function_token` only when the batch carries exactly one request (the only case where the batch's last hidden row is attributable). With several concurrent requests the key is omitted and `postprocess` computes each request's token from its own hidden-state slice; the forced-token override moved into `postprocess` (which has the request id) so both paths honor it.

vllm_omni/model_executor/models/nemotron_voicechat/nemotron_voicechat_talker.py
yuekaizhang: Fixed in 5f0f99ec: `_native_make_omni_output` now returns batch-aligned per-request lists for `codes.audio` and both meta keys (the generic splitter hands list element idx to request idx, while a shared tensor gets cloned to every request). Requests with nothing new this step keep placeholder entries (or their own cumulative history on prefill wakes) so no slot ever falls back to another request's tensor.

vllm_omni/model_executor/models/nemotron_voicechat/nemotron_voicechat_talker.py
yuekaizhang: Fixed in 5f0f99ec: recomputation is now replayed instead of killed. Rewound positions rebuild their decode embeds deterministically from the stored per-step code history (`codes_rows` plus the retained initial speaker-prompt code) — no resampling (`pending_step` is only set for a live step, so `make_omni_output` emits nothing for replayed rows) and the manual unconditional CFG stream is not advanced (it was never rewound). Bulk prefill replays, spans crossing the speaker-prompt boundary, and mixed replay+live chunks are all handled; the desync error now only fires when history is genuinely missing. Verified the live path is unchanged: duplex probe transcript and audio_bytes are identical to the previous head.

vllm_omni/deploy/nemotron_labs_voicechat.yaml
yuekaizhang: Fixed in 5f0f99ec: `max_tokens` is now 16347 (= 16384 max_model_len − the 37-token speaker-prompt prefill), with the comment updated to show the derivation.

recipes/NVIDIA/NemotronLabs-VoiceChat.md
yuekaizhang: Fixed in 5f0f99ec: the recipe now calls the yaml the fast default (with "PARITY:" comments restoring the NeMo-bit-exact reference), the pipeline table lists the thinker as bf16 (PARITY: fp32), and the talker note reflects the 37-token `talker_init_len` prompt and the 16347 budget.

yuekaizhang:
Thanks @linyueqian for the thorough review — all six findings were real and are addressed in 5f0f99ec:

- **MiniCPM-o streaming context**: dispatch also matches the `_streaming_context` spelling (name-based matching kept, since an arity-4 rule would collide with processors whose fourth parameter is `next_stage_hf_config`).
- **Shared function token**: emitted only for single-request batches; multi-request batches compute per request in `postprocess` from each request's own hidden slice, with the forced-token override applied there.
- **Talker output overwrite**: `_native_make_omni_output` returns batch-aligned per-request lists with placeholders, matching the generic splitter's list convention.
- **KV recomputation**: replayed deterministically from stored code history (bulk prefill spans, boundary-crossing spans, mixed replay+live chunks) instead of killing the engine; the CFG unconditional stream is not advanced during replay.
- **Talker budget**: `max_tokens` 16383 → 16347.
- **Recipe**: fast-default wording, bf16 (PARITY: fp32) dtype, `talker_init_len: 37` prompt note.

Validation on 1x H100 at 5f0f99ec: duplex probe transcript and audio_bytes byte-identical to the previous head; a synthetic 5-turn realtime session answers 5/5 correctly with delta pacing median 80.8 ms / p95 82.5 ms and zero gaps >160 ms; nemotron unit suites green (57 passed); pre-commit and builds green.

vllm_omni/model_executor/models/nemotron_voicechat/nemotron_voicechat_talker.py
```diff
@@ -324,14 +415,434 @@ def _init_session(self, request_id: str, info: dict[str, Any], device: torch.dev
...
+                session["uncond_hidden"] = session["uncond_stream"].step(uncond_embed)
+        session["pending_step"] = t
+        embeds = embeds.to(self._dtype)
+        if replay_rows:
+            embeds = torch.cat([*replay_rows, embeds], dim=0)
+        if prefill_head is not None:
+            embeds = torch.cat([prefill_head.to(embeds.dtype), embeds], dim=0)
+        return input_ids, embeds, {}
+
+    def _native_replay_embeds(self, session: dict[str, Any], t: int) -> torch.Tensor:
```
linyueqian, line 701: [suggestion] This replay path is the one part of the round-1 response that changes a loud failure into a silent one, and it is the only part with no test.

Before, any KV recompute on the native path raised immediately and the request died with a clear message. Now the session is reconstructed from stored history, so an off-by-one in the index arithmetic would not raise, it would emit subtly wrong audio. I checked that arithmetic and it is right: `codes_rows[i]` holds the codes sampled at step `i + 1`, so replaying step `t` correctly reads `codes_rows[t - 2]`, `initial_code` covers `t == 1`, and the guard rejects a replay whose stored rows are short. The substitution also matches the live path exactly, same `timeline[t:t+1]` EOS compare, same `silence_codes` select, same `cas_table[t:t+1]` slice, with the unconditional stream deliberately not advanced.

So I am not asking for a change to the logic. I am asking for one unit test that pins it, because the validation you ran, the byte-identical duplex probe and the five-turn session, exercises the live path and never preempts, so nothing in the suite would notice if these indices drifted later. A session fixture with a few stored code rows, replaying a span that crosses the prefill boundary and one that does not, would cover the three cases the code distinguishes: pure replay, replay plus a live step, and the short-history guard.

linyueqian (approved):
Re-reviewed the delta at 5f0f99ec. All six findings from round one are genuinely addressed, and in each case I checked the mechanism rather than the summary.

The dispatch fix matches the underscore spelling alongside the plain one, and the comment records why name matching was kept instead of switching to arity, which would have collided with processors whose fourth parameter is the next stage's config. The shared function token is now emitted only for a single-request batch, and the multi-request path computes each request's token in postprocess from that request's own hidden slice, with the forced-token override applied there too; moving the session write after the override also means the recorded token is the one actually used. The talker output is now batch-aligned per-request lists with placeholders, which is what the generic splitter needs, since it hands element idx to request idx but clones a shared tensor to everyone. The talker budget is 16347, exactly the 16384 context minus the 37-token speaker prefill, and the recipe now describes the fast default with the parity path called out rather than the reverse.

The recompute fix is the substantial one and it holds up. Replay reconstructs each already-generated step from stored codes, and the reconstruction mirrors the live path line for line: the same EOS compare against the timeline slice, the same silence substitution, the same subword slice into the embed builder, with the unconditional stream left alone because it was never rewound. The index arithmetic is correct, and every structural failure it cannot recover from still raises with a specific message.

One suggestion inline, about test coverage rather than logic. Nothing else to raise.

Validation: static review of the full delta against the round-one findings, both talker paths compared step by step, the code-history indexing checked against where rows are appended, and the batch-alignment claim checked against the splitter's contract. GitHub checks pass. Per-PR Buildkite has never run on this branch, so I have added the ready label and the approval waits on those lanes. No fork code was executed.

vllm_omni/model_executor/models/nemotron_voicechat/nemotron_voicechat_talker.py
yuekaizhang: Added in 96fec87c: `tests/model_executor/models/test_nemotron_voicechat_talker_replay.py` pins the replay arithmetic with a faked `build_decode_embeds` that encodes (timeline step, prev-code sum, uncond flag) into each row, so the composition is asserted directly. Six cases: pure replay (`initial_code` for t=1, `codes_rows[t-2]` after), replay + live final step (`pending_step` set, latest code fed last), a span crossing the speaker-prompt boundary (prefill embeds prepended), the live-first-step boundary rejection, the short-history guard, and the outpaced-position guard.

On the Buildkite lanes: `vllm-omni-amd-ci` #11359 fails on this head, but the same lane also fails on unrelated concurrent PRs (#11358, #11356, #11355), so it looks lane-wide rather than PR-induced; the main, Intel, and NPU lanes are green.

linyueqian (approved):
Re-approving at 96fec87c. The only change since my approval is the new replay test file, and it does more than the suggestion asked for.

It pins the arithmetic I had verified only by reading. Pure replay reads step t minus one's codes, and the code values are distinct enough that a shifted index would fail rather than pass by luck. The replay plus live step case asserts the live row lands last and the session is left pending resampling, while pure replay leaves nothing pending. Both guards are covered, the short code history and the position past the session step, and so is the span that crosses the speaker-prompt boundary, including that a live first step crossing it is still rejected.

The assertion I did not ask for and am glad to see is that the unconditional stream stays untouched during replay. That was the subtlest property of the fix, since advancing it would corrupt the CFG stream for every later step in a way no output-shape check would catch, and it is now locked.

Validation: static review of the delta, which is test-only with no production change, against the replay implementation and the three cases the suggestion named. Per-PR Buildkite at this exact head reports general, Intel, and NPU green. AMD fails only its Diffusion Test step, on the two cross-attention tolerance cases where a max difference of 0.015625 exceeds a 0.01 threshold, which is a known main-side flake in diffusion attention; this PR touches the nemotron voicechat model and the engine client and no diffusion code. No fork code was executed.

PR #5102, opened 2026-07-14
[Core][Benchmark]Omniinteract for Minicpm-o4.5

R2-Y:
PLEASE FILL IN THE PR DESCRIPTION HERE ENSURING ALL CHECKLIST ITEMS (AT THE BOTTOM) HAVE BEEN CONSIDERED.

## Purpose

Add a production-ready OmniInteract workflow for MiniCPM-o 4.5 native full-duplex video/audio serving, exposed through `vllm bench serve --omni`.

This PR:

- supports the official OmniInteract layouts and the `1q1a`, `1q1a_math`, and `1qna` subsets;
- preserves real-time media pacing and exact Session completion semantics;
- writes scoreable artifacts (`.done`, `output.wav`, `wav_transcript.json`, and `official_eval_manifest.jsonl`), while failed samples remain explicitly unscoreable;
- hardens input watermarking, final-input continuation, response completion, chunk metadata replacement, and Session cleanup/reuse paths found during real H800 E2E testing;
- adds an opt-in, single-GPU Nightly E2E using the real OmniInteract dataset.

### Runtime correctness changes worth calling out

- **Code2Wav runtime snapshots:** Talker-to-Code2Wav generation chunks are complete per-step snapshots, not incremental dictionaries. The generic runner previously deep-merged a marker-only chunk into the previous terminal chunk, retaining stale `codes.audio`, `cache_epoch`, `chunk_seq`, and `last_chunk`. That could replay terminal audio or report a stale/reordered chunk. `MiniCPMO45Code2Wav` now explicitly opts into replacement through `replace_runtime_additional_information = True`; the runner replaces only that request's producer payload while retaining runner-owned `num_processed_tokens` and `resumable` state. Models without the opt-in keep the existing merge behavior.
- **Connector snapshot boundary:** generation-mode connector payloads also replace the previous chunk before reaching the runner. Diffusion and other incremental payloads retain their prior merge semantics.
- **Exact input identity through Code2Wav:** `duplex_epoch`, `duplex_turn_id`, and `duplex_input_seq` are propagated through Stage 0, Talker, Code2Wav, and the data plane. This lets Serving emit `input.processed` for the exact accepted input instead of merely observing that some output arrived.
- **Per-chunk metadata accumulation:** `duplex_input_seq` is treated as a replace-and-drain chunk snapshot. Without this rule, repeated scalar tensors accumulated into a list, were not coercible to an input sequence on non-final deltas, and could leave the final Session waiting indefinitely.
[...]

PR #4349, opened 2026-06-11
[Bugfix] Reclaim unconsumed SHM segments and sender state on request abort

gagandhakrey:
## Purpose

Fixes a `/dev/shm` and host-memory leak on request abort in the SHM chunk-transfer path.

### What breaks

`SharedMemoryConnector.cleanup()` has **no callers** on current main. SHM segments are only unlinked inside `shm_read_bytes()` during successful consumption, so a chunk that is written but never read — client disconnect, downstream early exit, consumer failure — stays allocated until process exit, along with its `/dev/shm/shm_{key}_lockfile.lock`.

`/dev/shm` is physical RAM. Once exhausted, every subsequent `put()` fails and chunk transfer breaks pipeline-wide, not per-request.

Aborted requests additionally retain sender-side state (`put_req_chunk`, `request_payload`, `code_prompt_token_ids`, `requests_num_chunks_sent`, `ramp_chunk_count`, `_pending_streaming_prefills`), which can pin CPU tensors held by TTS input processors. `cleanup_sender()` is reachable only from the natural-finish path today.

This is issue #6352, measured at ~45 KB/abort and 258 → 60 MiB/h under sustained load.

### Repro

```bash
# Any 2-stage chunk-transfer pipeline (e.g. TTS talker -> vocoder).
ls /dev/shm | wc -l          # baseline

# Issue N streaming requests, disconnect each before the final chunk.

ls /dev/shm | wc -l          # grows by ~1 segment + 1 lock file per aborted
                             # request, never returns to baseline until exit
```

### Fix

**1. Abort-time connector sweep.**
`finish_requests()` enqueues a `connector_cleanup` task onto the save queue; `_send_single_request()` dispatches it to `_run_connector_cleanup()`, which calls `connector.cleanup(external_req_id)` and then `cleanup_sender()`.

Enqueued rather than called inline: both `put()` and the sweep run on the save-loop thread, so FIFO ordering guarantees no chunk is written after its own sweep.

**2. Gated on `finished_status`.**
Only `FINISHED_ABORTED` / `FINISHED_ERROR` sweep. Any other terminal status may still have an unconsumed terminal chunk downstream, and unlinking it would leave the next stage waiting on a finish marker that no longer exists.

The `is_finished()` guard alone is no longer sufficient — #6360's resumable-`FINISHED_STOPPED` carve-out lets a finished request fall through it.

**3. Structural cleanup-key matching.**
[...]

PR #6360, opened 2026-08-19
[Bugfix] Reclaim resumable async-chunk requests on finish

natureofnature:
## Purpose

This PR is the first focused bug-fix split from #5102. It contains only
resumable async-chunk scheduler/connector cleanup; the OmniInteract benchmark
and CI workflow, final-input lifecycle, and model-pipeline changes remain in
separate follow-up work.

Fix a scheduler/connector cleanup leak for resumable async-chunk requests.

### What broke

A model-native duplex request may finish one streaming segment with
`resumable=True` and `FINISHED_STOPPED` while it is still owned by a scheduler
or chunk-transfer queue. On session close or cancellation, upstream vLLM skips
requests that already have a finished status. The request could therefore keep
its KV/model-runner slot and active-stream capacity, eventually leaving later
sessions stuck before their first chunk.

Stale `requests_origin_status` state could also restore an off-queue terminal
request, while a defensive global running-queue purge could remove an unrelated
paused resumable request.

### Fix

- Realign a targeted resumable segment stop only when a live scheduler queue
  still owns it (`running`, `waiting`, or `skipped_waiting`).
- Preserve the `WAITING_FOR_STREAMING_REQ` accounting for skipped paused
  sessions so upstream cleanup balances its counter.
- Keep unrelated resumable segments in the running queue while purging only
  targeted, non-resumable-finished, or untracked entries.
- Restore connector origin status only when connector-owned parked state proves
  ownership, then reclaim receiver state through the existing idempotent cleanup
  path.
- Materialize single-pass request-id iterators before multiple cleanup layers
  consume them.

This split intentionally contains only scheduler/connector cleanup and its
tests. It does not change serving APIs, final-input/watermark handling, model
code, benchmark code, CI, or deployment configuration.

## Test Plan

```bash
python3 -m pytest -q \
  tests/core/sched/test_omni_generation_scheduler_update_session.py \
  tests/core/sched/test_omni_scheduler_finish_requests_purge.py \
  tests/distributed/omni_connectors/test_chunk_transfer_adapter.py

python3 -m pytest -q tests/core/sched -k \
  "not test_log_failed_chunk_sends_reports_a_live_request and \
   not test_log_failed_chunk_sends_still_reports_a_departed_request and \
   not test_zero_warns_and_names_both_transports"

CHANGED_PY="$(git diff --name-only HEAD^ HEAD -- '*.py')"
ruff check ${CHANGED_PY}
ruff format --check ${CHANGED_PY}
python3 -m compileall -q ${CHANGED_PY}
git diff --check HEAD^ HEAD
```

Coverage includes both AR and generation schedulers; hidden connector,
`running`, `waiting`, and `skipped_waiting` ownership; exactly-once free;
active-stream capacity reuse; unrelated resumable preservation; off-queue
terminal preservation; and generator request-id input.

**vLLM Version:** `0.26.0`

**vLLM-Omni Commit:** `9a9b37fed1e6b9e261e3c34fa9ddde19b700c853`

## Test Result

Tested in the existing H800 development container with Python 3.12 and vLLM
0.26.0. These are CPU scheduler/connector contract tests and do not require GPU
execution.

```text
Focused cleanup matrix: 107 passed, 17 warnings in 0.36s
Broader scheduler suite: 125 passed, 3 deselected, 17 warnings in 2.54s
ruff check:              passed
ruff format --check:     passed
compileall:              passed
git diff --check:        passed
```

The three deselected logging-capture tests also fail unchanged on the clean
`main` baseline in this container; they are not touched by this PR.


natureofnature:
@R2-Y PTAL

natureofnature:
@codex review

natureofnature:
@amy-why-3459   PTAL

R2-Y (approved):

amy-why-3459:
Thanks — this is the right first split from #5102. Queue-owned
resumable FINISHED_STOPPED was being skipped by upstream
finish_requests, so KV / runner slots / active-stream capacity
leaked across session close. Realigning only when a live queue
still owns the request, restoring WAITING_FOR_STREAMING_REQ for
skipped_waiting, scoping the running purge to this finish set,
and restoring connector origin status only on connector-owned
rows is the correct cut. Approve with comments.

Should-fix (non-blocking):

1. skipped_waiting + connector-owned can miss the counter fix.
   Adapter finish_requests restores origin status first, which
   clears FINISHED_STOPPED. The mixin then only maps
   skipped_waiting → WAITING_FOR_STREAMING_REQ when
   resumable_segment_stop is still true. The new test places
   "skipped" without also parking the request in a chunk deque,
   so it passes. A same-stage request in skipped_waiting and
   waiting_for_chunk_* / _held_non_active (close right at a
   segment boundary) would keep WAITING/RUNNING, stay in
   skipped_waiting, and leak num_waiting_for_streaming_input.
   Key that branch off skipped_waiting membership, not the
   pre-restore status, and add a skipped+hidden case.

2. Document that _purge_finished_from_running() with no target
   now keeps every resumable FINISHED_STOPPED in running. Fine
   today because only mixin passes the finish set.

3. A MiniCPM-o session-close / admission-reuse E2E would make
   the high-priority report harder to regress; fine as follow-up.

Receiver cleanup_receiver, off-queue deferred-free, generator
materialization, and the exactly-once free tests all look solid.
Sender SHM reclaim (#4349) is a different layer and correctly
out of this PR. Happy to merge from the duplex-core side.

natureofnature:
Addressed the skipped_waiting counter concern in 90a0c40e.

Changes:
- Snapshot counter-bearing skipped_waiting rows before the chunk adapter can restore an older status, then give that snapshot priority when realigning to WAITING_FOR_STREAMING_REQ.
- Restore requests_origin_status only while the request is actually owned by a connector hidden queue, so stale origin data cannot overwrite a visible scheduler status.
- Document the no-target behavior of _purge_finished_from_running.

Regression coverage now includes:
- the production WFSR + skipped_waiting + stale RUNNING origin path;
- the defensive skipped_waiting + connector-hidden segment-boundary path;
- a non-streaming skipped row proving the streaming counter is not decremented accidentally;
- direct adapter coverage proving stale origin is not restored without connector ownership;
- both AR and generation schedulers, exact-once free, second-finish idempotency, and active-stream capacity reuse.

Validation on vLLM 0.27:
- 114 passed, 14 warnings
- Ruff check passed
- Ruff format check passed

natureofnature:
@codex review

amy-why-3459 (approved):

PR #6364, opened 2026-08-19
[MUSA] Fix SwiGLU to use fused op

yeahdongcn:

## Purpose
Allow SwiGLU on MUSA to use the fused op.

## Test Plan

**vLLM Version:**

**vLLM-Omni Commit:**

### Fair A/B Comparison: Baseline → Final

Version | TP8 Median | Throughput Improvement
-- | -- | --
Baseline | 105.351 s | —
#6281 only | 101.492 s | +3.802%
#6283 + #6364 | 104.801 s | +0.525%
Final combined | 101.811 s | +3.477%

### DIFF
--- a/vllm_omni/diffusion/layers/activation.py
+++ b/vllm_omni/diffusion/layers/activation.py
@@ -22,7 +22,7 @@ class SiluAndMul(CustomOp):
         return F.silu(x[..., :d]) * x[..., d:]
 
     def forward_cuda(self, x: torch.Tensor) -> torch.Tensor:
-        if x.device.type != "cuda":
+        if x.device.type == "cpu":
             return self.forward_native(x)
         d = x.shape[-1] // 2
         output_shape = x.shape[:-1] + (d,)


yeahdongcn:
This is a follow-up to https://github.com/vllm-project/vllm-omni/pull/6283
Please @gcanlin take a look. Thanks!

vllm_omni/diffusion/layers/activation.py
```diff
@@ -22,7 +22,7 @@ def forward_native(x: torch.Tensor) -> torch.Tensor:
         return F.silu(x[..., :d]) * x[..., d:]
 
     def forward_cuda(self, x: torch.Tensor) -> torch.Tensor:
-        if x.device.type != "cuda":
+        if x.device.type not in ("cuda", "musa"):
```
Bounty-hunter: just remove `if x.device.type not in ("cuda", "musa")` ?  The default behavior mean `HIP` and `MUSA` are compatible with CUDA.
```
def forward_hip(self, *args, **kwargs):
        # By default, we assume that HIP ops are compatible with CUDA ops.
        return self.forward_cuda(*args, **kwargs)

def forward_musa(self, *args, **kwargs):
        # By default, we assume that MUSA ops are compatible with CUDA ops.
        return self.forward_cuda(*args, **kwargs)
```

if  `HIP` and `MUSA` really not compatible with CUDA, we should implement `forward_hip` or `forward_musa`.

yeahdongcn: Done.

Bounty-hunter:
need to add  `forward_hip`,  It seem that self.op only init when `current_omni_platform.is_cuda() or current_omni_platform.is_rocm() or current_omni_platform.is_musa()` in `__init__`

yeahdongcn:
> need to add `forward_hip`, It seem that self.op only init when `current_omni_platform.is_cuda() or current_omni_platform.is_rocm() or current_omni_platform.is_musa()` in `__init__`

`rocm` uses `forward_hip` (which falls back to `forward_cuda` by default), so leaving it as is should work.

Bounty-hunter (commented):
LGTM

vllm_omni/diffusion/layers/activation.py
```diff
@@ -22,8 +22,6 @@ def forward_native(x: torch.Tensor) -> torch.Tensor:
         return F.silu(x[..., :d]) * x[..., d:]
 
     def forward_cuda(self, x: torch.Tensor) -> torch.Tensor:
-        if x.device.type != "cuda":
```
gcanlin, line 25: For some potential CPU UT. We don't have cpu platform so need to add this guard.

```Python
if x.device.type == "cpu":
```

yeahdongcn: Updated. Thanks!

gcanlin (approved):

PR #6367, opened 2026-08-19
[Bugfix] Fix colocate-async sleep admission race and deliver AR abort tokens

knlnguyen1802:
Follow-up to [#6084](https://github.com/vllm-project/vllm-omni/pull/6084)

## Purpose
Colocated async RL (pause → abort → sleep → train → wake → resume) still fails after EngineCore sleep/wake routing lands:

1. Pipelined `generate()` can enter EngineCore while sleep is draining / offloading.
2. AR abort does not return the tokens already generated, so the trainer restarts from the original prompt instead of resuming the prefix.

Diffusion abort is unchanged (whole-sample retry). `wake_up()` still does **not** resume admission; callers must `resume_generation()`.

Typical trainer / serving order:

```text
pause → abort → sleep(level=1) → train/idle → wake → resume_generation → generate
```

### What broke

**Bug 1 — sleep admission race** 

Qwen3-Omni MMK12 training in `omni_colocate_async` fails at the first `on_sample_end()` sleep/wake boundary. Same model and dataset work with synchronous training. Observed errors:

```text
ValueError: b'\x00\x00' is not a valid EngineCoreRequestType
AssertionError: Expected a cached item for mm_hash='...'
RuntimeError: Generation rejected: Engine is partially or fully asleep.
Currently sleeping tags: ['weights', 'kv_cache'].
```

`EngineCoreRequestType.ADD` is the single byte `b'\x00'`. The `b'\x00\x00'` frame is consistent with a request being admitted while EngineCore is already sleeping / tearing down.

**Bug 2 — AR abort drops partial tokens**

An interrupted AR rollout is returned as an empty aborted response (`finish_reason="abort"`, `token_ids=[]`). `FullyAsyncLLMServerClient` therefore retries from the original prompt instead of `original_prompt + partial_tokens`. That breaks the verl colocated-async contract: abort is expected control flow, not a restart.

### Reproduce

In-tree GPU regression (fails on `main` / #6084-only; passes on this branch). Requires 2x H100 or MI325 and HF cache for `Qwen/Qwen3-Omni-30B-A3B-Instruct`:

```bash
pytest -sv tests/e2e/offline_inference/test_qwen3_omni.py::test_colocate_async_abort_tokens_and_sleep_admission \
  -m 'advanced_model and omni' --run-level advanced_model
```

On `main` the test fails with:

- abort: `generate()` never yields `finish_reason="abort"` with the tokens produced so far (empty prefix, or the stream hangs after frontend state is dropped)
- sleep: `generate()` is admitted into EngineCore while weights/KV are offloaded, so the task errors or finishes instead of waiting for `resume_generation()`

### Fix

**Sleep admission**

- `generate()` increments `_admitting` under `_pause_cond` **before** `add_request` and releases it only after submit (or on failure).
- `sleep()` sets `_paused = True`, then `wait_for(_admitting == 0)` before any EngineCore sleep RPC.
- Best-effort P0 `reset_mm_cache()` before EngineCore offload so hash-only follow-ups cannot outlive the P1 receiver cache.
- `wake_up()` still leaves `_paused`; callers must `resume_generation()`.

**AR abort tokens**

- `OutputProcessor.abort_requests_collecting_outputs()` builds a cumulative terminal output with `finish_reason=ABORT` and the detokenizer prefix **before** EngineCore abort (mirrors vLLM).
- `StagePool.abort_requests()` returns those outputs for AR stages; diffusion returns none.
- Orchestrator attaches final-stage AR outputs to `AbortResultMessage.abort_outputs`.
- `AsyncOmni._abort` enqueues them on the per-request queue, then pops frontend state so `generate()` observes the prefix.

## Test Plan

**vLLM Version:** 0.26.0

**vLLM-Omni Commit:** *(fill after push)*

- [x] CPU: sleep waits for in-flight `generate()` admission before EngineCore RPC
- [x] CPU: pause / sleep / wake routing + abort-ack unit tests
- [x] GPU 2x H100 or MI325: Qwen3-Omni abort delivers prefix, then resume `generate`
- [x] GPU 2x H100 or MI325: Qwen3-Omni sleep blocks `generate()` until `wake_up` + `resume_generation`
- [ ] Existing sleep-mode tests still pass with explicit resume after wake

CPU:

```bash
pytest -sv tests/entrypoints/test_async_omni_pause_sleep_routing.py \
  tests/engine/test_async_omni_engine_abort_ack.py \
  tests/engine/test_orchestrator.py -k abort \
  tests/engine/test_rpc_result_router.py -k abort \
  tests/entrypoints/test_async_omni.py -k abort \
  -m 'core_model and cpu'
```

GPU (Qwen/Qwen3-Omni-30B-A3B-Instruct, `ci/qwen3_omni_moe.yaml`, 2x H100 or MI325):

```bash
pytest -sv tests/e2e/offline_inference/test_qwen3_omni.py::test_colocate_async_abort_tokens_and_sleep_admission \
  -m 'advanced_model and omni' --run-level advanced_model
```

### CPU

```text
(paste pytest -q summary here)

# expected: all selected tests pass
```

### GPU (2x H100 / MI325, Qwen3-Omni-30B-A3B-Instruct)

| Check | `main` / #6084-only | this PR |
| --- | --- | --- |
| Abort yields `finish_reason="abort"` with non-empty prefix | fail | pass |
| Resume `generate` from `prompt_token_ids + abort_prefix` | fail (empty / restart) | pass |
| `generate()` during `sleep(level=1)` waits (does not enter EngineCore) | fail (asleep / ADD race) | pass |
| `wake_up()` does not resume admission | n/a | pass (`resume_generation` required) |

```text
(paste GPU pytest -sv summary here after the run)
```

cc: @SamitHuang  


knlnguyen1802:
cc: @zhtmike

tests/engine/test_orchestrator.py
```diff
@@ -805,10 +806,74 @@ async def test_run_abort(orchestrator_factory) -> None:
...
+
+        orchestrator_fixture.request_sync_q.put_nowait(AbortRequestMessage(request_ids=["req-ack"], rpc_id="abort-1"))
+        result = await _get_rpc_message(orchestrator_fixture)
+        assert isinstance(result, AbortResultMessage)
+        assert result.rpc_id == "abort-1"
+        assert result.success is True
+        assert result.error is None
+        assert result.rpc_correlation_key == ("abort", "abort-1")
+        assert stages[0].abort_calls == [["req-ack"]]
+        assert "req-ack" not in orchestrator_fixture.orchestrator.request_states
```
SamitHuang, line 872: This test never asserts `result.abort_outputs`. `FakeOutputProcessor` has no `abort_requests_collecting_outputs`, so StagePool falls back to `abort_requests()` and always returns `[]`. The ack/rpc_id path can go green while the trainer contract (terminal `finish_reason=abort` plus prefix tokens) is completely untested on CPU. Implement collecting on the fake OP (or use the real MultimodalOutputProcessor) and assert a finished abort `OutputMessage` with the seeded token ids.

vllm_omni/engine/stage_pool.py
```diff
@@ -1192,18 +1207,32 @@ async def abort_requests(self, request_ids: list[str]) -> None:
...
+            # different internal request_id assigned by InputProcessor.
+            engine_abort_ids = list(replica_request_ids)
+            if not is_diffusion and self._output_processor is not None:
+                collect = getattr(self._output_processor, "abort_requests_collecting_outputs", None)
+                if collect is not None:
+                    collected_ids, stage_outputs = collect(replica_request_ids, internal=False)
+                    for req_out in stage_outputs:
+                        req_id = getattr(req_out, "request_id", None)
+                        if req_id:
+                            abort_outputs.append((req_id, req_out))
```
SamitHuang: Abort outputs are keyed by `req_out.request_id`. Orchestrator then does `request_states.get(orch_req_id)` and silently `continue` on a miss. Omni `make_request_output` usually writes `external_req_id`, but EngineCore/internal ids still show up on some stages. Pair each collected output with the orchestrator id from `replica_request_ids` (the id you passed as `internal=False`) instead of trusting the RequestOutput id, or abort prefixes are dropped and the frontend never unblocks.

vllm_omni/entrypoints/async_omni.py
```diff
@@ -1079,8 +1122,23 @@ async def _abort_internal_requests(self, request_id: str | Iterable[str]):
...
+            req_id = getattr(output_msg, "request_id", None)
+            if req_id is None:
+                continue
+            state = self.request_states.get(req_id)
+            if state is None:
+                logger.debug("[AsyncOmni] Dropping abort output for unknown req %s", req_id)
+                continue
+            await state.queue.put(output_msg)
         for rid in request_ids:
             state = self.request_states.pop(rid, None)
```
SamitHuang, line 1173: `generate()` only leaves `_process_orchestrator_results` when it receives a `finished` OutputMessage. After this pop, there is no other producer. If `abort_async` returns `[]` (OP not registered yet, StagePool skipped an unbound replica, or Orchestrator dropped the id), `queue.get()` waits forever — the same hang described in the PR. When `abort_outputs` is empty, enqueue a synthetic finished abort OutputMessage for each `request_ids` entry that still has frontend state.

vllm_omni/entrypoints/async_omni.py
```diff
@@ -613,6 +621,8 @@ async def generate(
             submit_ts = time.time()
             req_state.metrics.stage_first_ts[0] = submit_ts
             req_start_ts[request_id] = submit_ts
+            await self._release_generate_admission()
```
SamitHuang, line 629: `_add_streaming_input_request` returns as soon as it `create_task`s the input pump; the first `add_request_async` runs later. Admission is released here, so `sleep()` can pass `_admitting == 0` and start EngineCore offload while that pump still submits ADD. Hold the slot until the first chunk is actually submitted (or fail), not until the helper returns.

tests/e2e/offline_inference/test_qwen3_omni.py
```diff
@@ -77,3 +82,129 @@ def test_thinker_only_model_request(omni_runner, omni_runner_handler) -> None:
...
+    On ``origin/main`` this test fails:
+    - abort: ``generate()`` never yields ``finish_reason="abort"`` with the
+      tokens produced so far, so resume has an empty prefix (or the stream
+      hangs after frontend state is dropped).
+    - sleep: ``generate()`` is admitted into EngineCore while weights/KV are
+      offloaded, so the task errors (asleep / corrupted ADD frame) instead of
+      waiting for ``resume_generation()``.
+
+    On this PR both legs pass: abort returns a resumeable prefix, and sleep
+    holds ``generate()`` until wake + resume.
```
SamitHuang, line 136: The Test Result section still says “paste GPU pytest -sv summary here”, so the claim that this test passes on 2x H100/MI325 is unverified in the PR. This is also `@pytest.mark.advanced_model` / 30B-only, so it will not catch the abort-prefix or sleep-admission races in L1/L2. Keep this e2e, but add a CPU test that fails on `main` for empty abort tokens and for `generate()` entering EngineCore during sleep.

SamitHuang (changes requested):
The sleep admission gate for non-streaming `generate()` is the right shape (`_paused` then wait `_admitting == 0` before EngineCore offload). The AR abort-token path is not actually proven on CPU, and two delivery holes can recreate the hang this PR is trying to close.

`AbortResultMessage` / pause-sleep routing unit tests do not assert a non-empty prefix. The GPU Qwen3-Omni test is the only coverage, and the PR still has placeholder pytest output in Test Result. Please add a CPU test that drives `abort_requests_collecting_outputs` → StagePool → Orchestrator → `AsyncOmni._abort` and paste GPU results before merge.

zhtmike:
One contract question: `StagePool.collective_rpc` still returns engine failures as values (`{"supported": False}`) — including for the control methods this PR routes (`sleep`/`wake_up`/`pause_scheduler`/`resume_scheduler`). We hit this in colocated RL: a worker died mid-`wake_up` (native `RuntimeError`), the dict came back as a normal result, verl-omni/upstream verl never inspect it, and the trainer hung on a dead engine. vllm's own `collective_rpc` propagates exceptions, and this PR's machinery (`_engine_core_rpc`, `_paused`/`_admitting`) already assumes failures are visible.

Suggestion: make the control methods re-raise (keep the dict for capability probes):

```python
except Exception as exc:
    if method in _ENGINE_CORE_CONTROL_ASYNC_METHODS:
        raise
```

Also worth flipping the new `assert result["supported"] is False` — as written it test-blesses the swallow. Happy to contribute the hunk + tests.

@knlnguyen1802 , AI assistant review, related to https://github.com/verl-project/verl-omni/issues/433

tests/engine/test_orchestrator.py
knlnguyen1802: Fix in 88e713273f998ab74428da620715cda7735b9171

vllm_omni/entrypoints/async_omni.py
knlnguyen1802: Fix in 88e713273f998ab74428da620715cda7735b9171

vllm_omni/entrypoints/async_omni.py
knlnguyen1802: Fix in 88e713273f998ab74428da620715cda7735b9171

tests/e2e/offline_inference/test_qwen3_omni.py
knlnguyen1802: Fix in 88e713273f998ab74428da620715cda7735b9171

SamitHuang:
Pls resolve the conflicts

tests/e2e/offline_inference/test_qwen3_omni.py
```diff
@@ -77,3 +82,132 @@ def test_thinker_only_model_request(omni_runner, omni_runner_handler) -> None:
...
+        },
+    )
+
+
+@pytest.mark.advanced_model
+@pytest.mark.omni
+@hardware_test(res={"cuda": "H100", "rocm": "MI325"}, num_cards=2)
+@pytest.mark.asyncio
+async def test_colocate_async_abort_tokens_and_sleep_admission() -> None:
+    """Prove the two remaining zhtmike/vllm-omni#1 bugs are fixed.
```
zhtmike: drop the comment ~ thx

vllm_omni/outputs/output_processor.py
```diff
@@ -424,7 +445,60 @@ def abort_requests(self, request_ids, internal: bool) -> list[str]:
...
+                            self.external_req_ids.pop(external_req_id, None)
+            elif internal_ids := self.external_req_ids.pop(request_id, []):
+                internal_req_ids.extend(internal_ids)
+
+        request_ids_to_abort: list[str] = []
+        abort_outputs: list[RequestOutput | PoolingRequestOutput] = []
+        for request_id in internal_req_ids:
+            req_state = self.request_states.pop(request_id, None)
+            if req_state is not None:
+                self.lora_states.request_finished(request_id, req_state.lora_name)
```
SamitHuang: When aborting individual requests that have a `parent_req` (from parallel sampling), `req_state` is popped and added to `request_ids_to_abort`, but `self.parent_requests` retains the parent entry unless the parent itself was passed as the abort ID. This leaves stale `ParentRequest` objects in memory. Consider cleaning up `self.parent_requests.pop(req_state.parent_req.request_id, None)` when the parent's children are finished or aborted, matching the behavior in `remove_request()`.

tests/entrypoints/test_omni_sleep_mode.py
```diff
@@ -374,6 +374,7 @@ class TP=2 BagelPipeline engine rather than a second default-config load).
             # AR stages keep the frontend admission gate held after wake.
             await diffusion_engine.resume_generation(stage_ids=[0])
             await asyncio.sleep(2.0)
+            await diffusion_engine.resume_generation(stage_ids=[0])
```
SamitHuang: `await diffusion_engine.resume_generation(stage_ids=[0])` is called twice back-to-back across lines 375 and 377. Line 375 already performs the resume after `wake_up()`, so line 377 is a redundant duplicate likely left over from merge resolution. Please remove the duplicate invocation.

vllm_omni/engine/stage_pool.py
```diff
@@ -1254,9 +1278,13 @@ async def collective_rpc(
                 replica_id,
                 method,
             )
+            if isinstance(exc, TimeoutError):
+                error = f"{type(exc).__name__}: {method} timed out after {timeout}s"
+            else:
+                error = str(exc) or repr(exc)
             return {
```
SamitHuang, line 1304: `StagePool.collective_rpc` converts all exceptions into a return dictionary `{"supported": False, "error": error}`. For EngineCore control methods (`pause_scheduler`, `resume_scheduler`, `sleep`, `wake_up`), callers and external frameworks (such as verl-omni) expect worker exceptions to propagate directly rather than being silently absorbed into return values. Consider re-raising exceptions when `method in self._ENGINE_CORE_CONTROL_ASYNC_METHODS`.

vllm_omni/entrypoints/async_omni.py
```diff
@@ -1291,10 +1393,15 @@ async def sleep(
...
-        # pause the AR scheduler itself; avoid a second pause_scheduler here.
+        # _pause_cond during the drain/offload window. Wait until generate()
+        # coroutines that already passed the pause check have submitted (or
+        # failed) so EngineCore does not see ADD frames while sleeping.
         async with self._pause_cond:
             self._paused = True
+            await self._pause_cond.wait_for(lambda: getattr(self, "_admitting", 0) == 0)
+
+        # P0 sender cache must drop hashes before EngineCore.sleep clears P1.
+        await self.reset_mm_cache()
```
SamitHuang, line 1488: `_sleeping_tags` is tracked as a single global set on `AsyncOmni`, but `sleep()` and `wake_up()` support stage-specific operations via `stage_ids`. If `sleep(stage_ids=[0])` runs followed by `wake_up(stage_ids=[0])`, all tags in `_sleeping_tags` are discarded, causing a subsequent `wake_up(stage_ids=[1])` to return early without waking stage 1. Consider tracking sleeping tags per stage or only clearing a tag when all stages holding it are awake.

SamitHuang (commented):
This update addresses the core issues identified in previous reviews, including acknowledged abort routing, holding streaming admission until the initial chunk submit, synthesizing terminal abort outputs when the engine returns none, and adding CPU regression tests for abort prefixes and sleep admission.

A few remaining items require attention:
1. **Parent Request Cleanup in OutputProcessor**: Aborting child requests does not clean up `parent_requests` mappings.
2. **Exception Propagation in Control Methods**: `StagePool.collective_rpc` converts errors to dictionaries rather than raising for critical control paths.
3. **Global Sleep Tags vs Stage-Specific Operations**: `_sleeping_tags` is tracked globally, which causes partial `wake_up` calls to prematurely clear sleep state.
4. **Duplicate Resume Call**: `tests/entrypoints/test_omni_sleep_mode.py` contains a redundant duplicate `resume_generation()` call.

PTAL @knlnguyen1802

vllm_omni/outputs/output_processor.py
knlnguyen1802: Done in 30b5e3b1. Aborting a parallel-sampling child now discards it from `parent.child_requests` and pops `parent_requests` once no children remain. Checked with `python3 -m pytest --noconftest -o addopts= tests/engine/test_output_processor.py::test_abort_last_parallel_child_drops_parent_request`.

tests/entrypoints/test_omni_sleep_mode.py
knlnguyen1802: Done in 30b5e3b1. Removed the duplicate `resume_generation(stage_ids=[0])` after wake; the call on the previous line is kept.

vllm_omni/engine/stage_pool.py
knlnguyen1802: Done in 30b5e3b1. `StagePool.collective_rpc` now re-raises for EngineCore control methods (`pause_scheduler`, `resume_scheduler`, `sleep`, `wake_up`); capability probes still return `{supported: False, error: ...}`. Checked with `python3 -m pytest --noconftest -o addopts= tests/engine/test_stage_pool_collective_rpc.py`.

vllm_omni/entrypoints/async_omni.py
knlnguyen1802: Done in 30b5e3b1. Sleeping tags are tracked per stage; `wake_up(stage_ids=[0])` only clears stage 0, so a later `wake_up(stage_ids=[1])` still runs. Checked with `python3 -m pytest --noconftest -o addopts= tests/entrypoints/test_async_omni_pause_sleep_routing.py`.

vllm_omni/outputs/output_processor.py
knlnguyen1802: resolved by 30b5e3b1c1ed65b5ea7eaff1a24aaceae42a199b

tests/entrypoints/test_omni_sleep_mode.py
knlnguyen1802: resolved by 30b5e3b1c1ed65b5ea7eaff1a24aaceae42a199b

vllm_omni/engine/stage_pool.py
knlnguyen1802: resolved by 30b5e3b1c1ed65b5ea7eaff1a24aaceae42a199b

vllm_omni/entrypoints/async_omni.py
knlnguyen1802: resolved by 30b5e3b1c1ed65b5ea7eaff1a24aaceae42a199b

knlnguyen1802:
@SamitHuang Resolve all comment PTAL

vllm_omni/entrypoints/async_omni.py
```diff
@@ -731,37 +762,43 @@ async def handle_inputs() -> None:
...
                 )
             finally:
-                if not cancelled:
-                    # Send empty final request to indicate that inputs have
-                    # finished. Don't send if canceled (session was aborted).
-                    final_sampling_params_list = list(sampling_params_list)
-                    final_sampling_params_list[0] = stage0_params
-                    final_prompt = TokensPrompt(prompt_token_ids=[0])
-
-                    if has_submitted_first_chunk:
```
Gaohan123, line 748: [P1] Gate every streaming update against sleep admission

After the first streaming chunk is submitted, `generate()` releases `_admitting`, but subsequent chunks here—and the final marker at line 774—call `add_streaming_update_async()` without consulting `_pause_cond`.

If `sleep()` observes `_admitting == 0` between chunks, its sleep RPC can be processed before a later streaming update. That update is then submitted to an already sleeping EngineCore, reproducing the asleep/corrupted-ADD race this PR is intended to fix.

Please apply the same pause-aware admission protocol around every streaming ADD/update/final marker, or pause and drain/cancel the input pump before starting sleep.

vllm_omni/entrypoints/async_omni.py
```diff
@@ -603,7 +615,9 @@ async def generate(
                     final_output_stage_ids=final_output_stage_ids,
                     arrival_time=wall_start_ts,
                     lora_request=lora_request,
+                    first_chunk_submitted=first_chunk_submitted,
                 )
```
Gaohan123, line 631: [P1] Do not hold the global admission slot while waiting for an unbounded input stream

`_admitting` is incremented before the streaming input generator produces its first chunk, while this future completes only after the first chunk is submitted or the generator exits.

A realtime connection that is open but temporarily waiting for client input can therefore keep `_admitting` nonzero indefinitely, causing `sleep()` to block forever at line 1443.

Please acquire the admission slot inside the input pump immediately before the actual first ADD, rather than while waiting for user input, or explicitly coordinate/cancel pending pre-submit streams during sleep.

vllm_omni/entrypoints/async_omni.py
```diff
@@ -1109,11 +1146,34 @@ async def _abort_internal_requests(self, request_id: str | Iterable[str]):
...
+            await state.queue.put(output_msg)
+            delivered.add(req_id)
+        for rid in request_ids:
+            state = self.request_states.get(rid)
+            if state is not None and rid not in delivered:
+                queue = getattr(state, "queue", None)
+                if queue is not None:
+                    await state.queue.put(self._synthetic_abort_output_message(rid))
+                    delivered.add(rid)
         for rid in request_ids:
```
Gaohan123, line 1218: [P1] Keep the request state until the terminal abort output is consumed

This removes the request from `request_states` immediately after queuing its abort output. If an external abort runs after submission but before `generate()` enters `_process_orchestrator_results()`, the consumer reaches line 853, sees no request state, and returns without consuming the queued terminal message.

The caller then receives neither `finish_reason="abort"` nor the generated prefix.

Please retain the `ClientRequestState`—or a tombstone containing its queue—until the generator consumes the terminal output and performs normal cleanup, or pass the already-created state directly into the result-processing loop.

vllm_omni/engine/orchestrator.py
```diff
@@ -886,13 +895,48 @@ async def _handle_interaction(self, msg: InteractionMessage) -> None:
...
+                        final_output_type=final_output_type,
+                    )
+                    abort_outputs.append(
+                        OutputMessage(
+                            request_id=orch_req_id,
+                            stage_id=pool.stage_id,
+                            replica_id=pool.get_bound_replica_id(orch_req_id),
+                            engine_outputs=engine_output,
+                            metrics=None,
+                            finished=True,
```
Gaohan123: [P2] Do not mark every final-stage abort output as request-terminal

A request may have multiple `final_output_stage_ids`, such as separate text and audio stages. Every abort output is marked `finished=True` here, while `AsyncOmni._process_orchestrator_results()` stops reading after the first finished message at line 904.

Because pools are processed in stage order, a text abort can terminate the stream before a later audio-stage abort output is delivered.

Please apply the same all-final-stage convergence used by `_route_output()` and mark only the final or aggregate abort message as request-terminal.

vllm_omni/outputs/output_processor.py
```diff
@@ -424,7 +445,65 @@ def abort_requests(self, request_ids, internal: bool) -> list[str]:
...
+                            pass
+                        if not internal_ids:
+                            self.external_req_ids.pop(external_req_id, None)
+            elif internal_ids := self.external_req_ids.pop(request_id, []):
+                internal_req_ids.extend(internal_ids)
+
+        request_ids_to_abort: list[str] = []
+        abort_outputs: list[RequestOutput | PoolingRequestOutput] = []
+        for request_id in internal_req_ids:
+            req_state = self.request_states.pop(request_id, None)
```
Gaohan123: [P2] Do not destroy OutputProcessor state before the physical abort succeeds

This removes the request state, external-ID mapping, detokenizer state, and LoRA accounting before `StagePool.abort_requests()` awaits the physical `client.abort_requests_async()` call at `stage_pool.py:1238`.

If that RPC fails, the EngineCore request may still be active, but a retry can no longer recover the partial prefix or restore consistent processor state.

Please snapshot the terminal output first and commit state removal only after the physical abort succeeds, or restore the removed state and mappings when the RPC fails.

vllm_omni/engine/async_omni_engine.py
```diff
@@ -1752,12 +1752,16 @@ async def abort_async(
         self,
         request_ids: list[str],
         timeout: float | None = None,
-    ) -> None:
+    ) -> list[Any]:
```
Gaohan123: Avoid use Any

vllm_omni/entrypoints/async_omni.py
knlnguyen1802: Done in a7b6812d. Each streaming ADD/update/final marker now takes a pause-aware admission slot via `_submit_with_admission()`. Checked with `python3 -m pytest --noconftest -o addopts= tests/entrypoints/test_async_omni_pause_sleep_routing.py::test_streaming_in_flight_add_blocks_sleep`.

vllm_omni/entrypoints/async_omni.py
knlnguyen1802: Done in a7b6812d. Streaming generate no longer holds `_admitting` while waiting for the first client chunk; the pump acquires the slot immediately before each ADD. Checked with `python3 -m pytest --noconftest -o addopts= tests/entrypoints/test_async_omni_pause_sleep_routing.py::test_streaming_wait_for_first_chunk_does_not_block_sleep`.

vllm_omni/entrypoints/async_omni.py
knlnguyen1802: Done in a7b6812d. `_abort` no longer pops `request_states`; `generate()` consumes the terminal abort and cleans up in `finally`. Checked with `python3 -m pytest --noconftest -o addopts= tests/entrypoints/test_async_omni.py::test_abort_keeps_request_states_until_generate_cleanup`.

vllm_omni/engine/orchestrator.py
knlnguyen1802: Done in a7b6812d. Abort OutputMessages are built unfinished, then only the last message per request is marked `finished=True`. Checked with `python3 -m pytest --noconftest -o addopts= tests/engine/test_orchestrator.py::test_abort_marks_only_last_final_stage_output_finished`.

vllm_omni/outputs/output_processor.py
knlnguyen1802: Done in a7b6812d. Abort outputs are snapshotted with `commit_state=False`; OP mappings are dropped only after EngineCore abort succeeds. Checked with `python3 -m pytest --noconftest -o addopts= tests/engine/test_output_processor.py::test_abort_snapshot_leaves_state_until_commit tests/engine/test_stage_pool_collective_rpc.py::test_abort_requests_does_not_commit_op_state_when_engine_abort_fails`.

vllm_omni/engine/async_omni_engine.py
knlnguyen1802: Done in a7b6812d. `abort_async` now returns `list[OutputMessage]` instead of `list[Any]`. Checked with `python3 -m pytest --noconftest -o addopts= tests/engine/test_async_omni_engine_abort_ack.py`.

SamitHuang (approved):
LGTM

SamitHuang:
@knlnguyen1802 Pls fix the CI

knlnguyen1802:
@SamitHuang Ready again

knlnguyen1802:
@SamitHuang Separate CI, please help to check

Issue #5867, opened 2026-08-06
[RFC]: Add Server-Side VAD to Realtime Sessions for Turn-Based Omni Models


LHXuuu:
## 1. Motivation

vLLM-Omni already has an experimental Realtime serving stack under `experimental/fullduplex/openai`, including WebSocket I/O, OpenAI Realtime event translation, session lifecycle, conversation history, output fencing, native-duplex runtime integration, and a chat fallback path.

However, turn-based speech models such as Qwen3-Omni cannot determine user turn boundaries by themselves. Clients must explicitly commit buffered audio before inference can start, which makes it difficult to build an OpenAI-style Realtime voice experience.

This RFC proposes extending the existing Realtime stack with server-side VAD. It does not introduce a second WebSocket, session, protocol, or runtime-adapter stack.

For turn-based models, VAD converts continuous audio input into committed turns and reuses the existing `ChatFallbackProjectorMixin` path backed by `OmniOpenAIServingChat.create_chat_completion`, rather than calling `AsyncOmni.generate` directly. Model-native duplex models continue to use the existing `NativeRuntimeBridgeMixin` path and bypass server VAD.

This provides a **serving-managed Realtime conversation** for turn-based models. It does not make Qwen3-Omni natively capable of listening and speaking simultaneously. Phase 1 supports one active response and does not provide reliable barge-in.

### Phase 1 goals

- Support OpenAI-style `/v1/realtime` over WebSocket.
- Support continuous audio append for Qwen3-Omni.
- Add `server_vad` turn detection, using Silero VAD initially.
- Emit the corresponding Realtime speech, commit, and response events.
- Reuse the existing full-duplex session and chat serving infrastructure.
- Preserve the existing native-duplex path.
- Add bounded session-owned input buffering for server VAD, per-session VAD state cleanup, and minimal VAD metrics.

### Non-goals

- WebRTC or SIP.
- Video input or output.
- Realtime transcription sessions.
- MCP or function tools.
- Semantic VAD.
- Reliable `interrupt_response` or barge-in for turn-based models.
- Simultaneous model listening and generation.
- Starting inference before VAD endpoint detection.

### 3.1 Architecture

The existing Realtime stack supports both model-native duplex sessions and serving-managed turn-based sessions.

This RFC extends the turn-based path with a `ServerVADPipeline`. Continuous audio is processed by server VAD, and the detected speech endpoint is converted into a turn commit that reuses the existing chat fallback path.

[...]

PR #6372, opened 2026-08-19
[Model][Full Duplex] Add Qwen3-Omni turn-based duplex serving

JoseCarlosGarcia95:
## Summary

- Add Qwen3-Omni turn-based duplex serving through `/v1/duplex`.
- Reuse the existing chat fallback for Qwen3 instead of enabling the MiniCPM/PersonaPlex engine control plane.
- Add Qwen3 duplex policy/session adapter, deploy profile, audio normalization, and integration tests.

## Design

Qwen3-Omni does not expose MiniCPM-style native listen/speak tokens. The implementation therefore uses a serving-session adapter and the existing `ChatFallbackProjectorMixin`:

- User text/audio is committed into the existing `DuplexSession` history.
- PCM float input is normalized into a valid mono PCM16 WAV before becoming an `audio_url` chat content part.
- Qwen3 receives text/audio output modalities through the existing chat service.
- Barge-in marks the interrupted turn and cancels the active fallback response.
- Non-native Qwen3 sessions explicitly opt out of engine runtime control and engine-side append calls.

## Scope

Included:

- `qwen3omni` serving adapter, session state, policy and data-plane contract.
- Qwen3 duplex pipeline wiring and `qwen3_omni_duplex.yaml`.
- `input_audio_format` compatibility for duplex append events.
- vLLM 0.27 `ErrorResponse.error` handling.
- Audio duration/playback metadata for fallback audio deltas.
- CPU and GPU-host integration tests.

Intentionally not included:

- Generic non-native duplex session admission/backpressure redesign.
- A new generic core runtime path inside `OmniDuplexSessionHandler`.
- Broader overlap policy changes for all duplex adapters.

## Validation

- 239 relevant tests passed on an H100 host.
- Real Qwen3-Omni PCM16 audio input was understood and transcribed.
- Text turn, audio turn, audio response, barge-in and post-interruption turn were exercised on the real model.
- `ruff check` passed.
- `ruff format --check` passed.
- Existing duplex handler regression tests passed.


JoseCarlosGarcia95:
Key points checked:

- Qwen3-Omni uses the existing `ChatFallbackProjectorMixin` path, not the MiniCPM/PersonaPlex native engine control plane.
- The static duplex system prompt enforces short spoken turns, natural speech, no markdown, and stopping when interrupted.
- After a barge-in, a one-shot interruption note instructs the model to discard the interrupted reply and answer only the latest user input.
- Session-level instructions remain configurable through the duplex API.
- PCM16 and `input_audio_format` handling were verified, including normalization to a valid WAV container before creating the chat `audio_url` content.
- Audio output metadata, duration tracking, playback state, barge-in cancellation, and post-interruption turns were reviewed.
- The Qwen3 serving adapter and deploy pipeline wiring were checked.
- The real model was tested on an H100 with text input, audio input, audio output, barge-in, and a subsequent response.
- The relevant GPU test suite passed with `239 tests passed`.
- `ruff check` and `ruff format --check` passed.
- Existing duplex handler regression tests passed.

Manual browser testing used an external local HTML harness. That frontend is intentionally not included in this PR.

This PR is limited to Qwen3-Omni's turn-based chat-fallback path. Generic admission/backpressure and cross-adapter overlap-policy changes are left for separate work.

@amy-why-3459 @psv666, please review the Qwen3-Omni integration and the chat-fallback audio path.

amy-why-3459:
Thanks — chat-fallback for Qwen3-Omni (no native listen/speak tokens, no
engine control plane) is the right cut. Request changes on the live data
path and the shared-file branches.

1. mark_audio_sent() takes a cumulative timestamp and does max().
   _audio_duration_ms() returns the duration of the current payload.
   Streaming multi-chunk audio will under-count playback
   (max(100,100,100)=100). Sum, or accumulate in the fallback, and add a
   test that two 500ms chunks yield generated_ms == 1000. Don't hardcode
   24 kHz if the model can emit 16 kHz PCM.

2. DuplexSession.history is what _build_chat_request sends. The Qwen3
   session PCM buffer, state.history, USER_AUDIO_MARKER, and
   Qwen3OmniDataPlaneSession.project are not on that path (project is
   only used from the native runtime_bridge stream). Session unit tests
   therefore don't cover serving. Either wire them or delete/label them
   as handler-protocol padding. test_project_fails_closed_* only checks
   events == [] — that's drop, not fail-closed.

3. Don't special-case adapter_id == "qwen3omni" in chat_fallback.py and
   session_runner.py. Add adapter hooks (turn_policy_messages is already
   there; add on_barge_in / auto_respond_on_commit). The next non-native
   model should not copy another id check.

Should-fix:
- validate_client_extra_body is only invoked when
  _uses_native_input_append; Qwen3 never hits it.
- serving_session_adapter always starts a response on commit, even when
  auto_response is unset — document or honor response.create.
- Clear last_turn_interrupted only after the chat request is issued, not
  inside turn_policy_messages / _build_chat_request.
- Add a short recipe for --deploy-config qwen3_omni_duplex.yaml + /v1/duplex.
- is_enabled() is unconditionally True; say this adapter must not be
  hung on another pipeline.

Skipping the MiniCPM control plane, the pcm_f32le→WAV wrap, and the
barge-in-only-if-active-response tests all look right. Happy to
re-review once playback and the live history path are honest.

JoseCarlosGarcia95:
> Thanks — chat-fallback for Qwen3-Omni (no native listen/speak tokens, no engine control plane) is the right cut. Request changes on the live data path and the shared-file branches.
> 
> 1. mark_audio_sent() takes a cumulative timestamp and does max().
>    _audio_duration_ms() returns the duration of the current payload.
>    Streaming multi-chunk audio will under-count playback
>    (max(100,100,100)=100). Sum, or accumulate in the fallback, and add a
>    test that two 500ms chunks yield generated_ms == 1000. Don't hardcode
>    24 kHz if the model can emit 16 kHz PCM.
> 2. DuplexSession.history is what _build_chat_request sends. The Qwen3
>    session PCM buffer, state.history, USER_AUDIO_MARKER, and
>    Qwen3OmniDataPlaneSession.project are not on that path (project is
>    only used from the native runtime_bridge stream). Session unit tests
>    therefore don't cover serving. Either wire them or delete/label them
>    as handler-protocol padding. test_project_fails_closed_* only checks
>    events == [] — that's drop, not fail-closed.
> 3. Don't special-case adapter_id == "qwen3omni" in chat_fallback.py and
>    session_runner.py. Add adapter hooks (turn_policy_messages is already
>    there; add on_barge_in / auto_respond_on_commit). The next non-native
>    model should not copy another id check.
> 
> Should-fix:
> 
> * validate_client_extra_body is only invoked when
>   _uses_native_input_append; Qwen3 never hits it.
> * serving_session_adapter always starts a response on commit, even when
>   auto_response is unset — document or honor response.create.
> * Clear last_turn_interrupted only after the chat request is issued, not
>   inside turn_policy_messages / _build_chat_request.
> * Add a short recipe for --deploy-config qwen3_omni_duplex.yaml + /v1/duplex.
> * is_enabled() is unconditionally True; say this adapter must not be
>   hung on another pipeline.
> 
> Skipping the MiniCPM control plane, the pcm_f32le→WAV wrap, and the barge-in-only-if-active-response tests all look right. Happy to re-review once playback and the live history path are honest.

Thanks for the detailed review. I reviewed the requested changes and agree with the architectural cut: Qwen3-Omni should use the chat-fallback path because it has no native listen/speak tokens and does not require the engine control plane.

I also think this approach is interesting beyond Qwen3. Keeping `DuplexSession.history` as the canonical live history and exposing model-specific hooks for turn policy, commit behavior, barge-in, and runtime-control opt-out could provide a reusable path for making other audio-capable vLLM-Omni models duplex, without implementing a separate native runtime for each model.

This PR remains intentionally scoped to Qwen3-Omni. Generalizing and validating the hooks for additional models can be handled as follow-up work.

I reviewed the live audio path, PCM/WAV normalization, multi-chunk playback accounting, interruption state, and compatibility with the shared duplex handler. The requested changes are covered by tests and have been validated on an H100.

@amy-why-3459

Sy0307:
Tracked in #6494 as the Qwen turn-based `/v1/duplex` chat-fallback path, separate from GA `/v1/realtime` routing (#6592/#6659) and native MiniCPM serving. Current gates are the DCO failure plus the review-bot implementation/test mismatch and shared commit-policy scope finding; resolve those before using this PR as the base for #5867/#6618 VAD behavior.

JoseCarlosGarcia95:
Thanks for the triage and the detailed review. Everything is addressed in the updated head (**42398b10**), rebased onto **upstream/main (5d7fdf93)** as a single DCO-signed commit.

The duplex engine and serving modules have moved out of experimental/fullduplex/, so all Qwen3 imports now use the stable vllm_omni.entrypoints.duplex.* and vllm_omni.engine.duplex.* paths.

**A1 (tests vs implementation):** tests/e2e/features/fullduplex/test_qwen3omni_session.py has been removed. The remaining Qwen3 tests no longer use append_pcm, drain_pcm, build_audio_content_part, record_*_turn, or state.history.

Qwen3OmniServingSessionState is now explicitly documented as handler-protocol padding. The actual conversation history lives in **DuplexSession.history**. test_project_fails_closed_* now asserts RuntimeError("chat fallback") instead of relying on an empty event list.

**A2 (shared-path scope):** auto_respond_on_commit is opt-in per adapter and defaults to False. MiniCPM-o 4.5, PersonaPlex, and Nemotron VoiceChat don't define it, so their commit path is unchanged and _session_auto_responds(session) continues to control native full-duplex behavior.

**Qwen3-Omni is the only adapter returning True**, keeping this change scoped to the turn-based chat fallback tracked in #6494.

**Review item 1 — playback duration:** mark_audio_sent now receives cumulative generated_ms from the fallback projector, with _audio_duration_ms accumulated across streaming chunks. A regression test verifies that two 500 ms PCM16 chunks at 16 kHz produce **generated_ms == 1000**. The sample rate comes from model output rather than being hardcoded to 24 kHz.

**Review item 2 — session state vs serving path:** covered by A1. test_qwen3_chat_fallback_reads_duplex_session_history verifies that _build_chat_request reads from **DuplexSession.history**.

**Review item 3 — no adapter_id special-casing:** all adapter_id == "qwen3omni" checks have been removed from chat_fallback.py and session_runner.py. Qwen3-specific behavior now goes through the generic adapter hooks turn_policy_messages, on_barge_in, on_turn_request_issued, and auto_respond_on_commit.

**Other fixes:**

- validate_client_extra_body is now called through the serving adapter path regardless of _uses_native_input_append.
- auto_respond_on_commit is a callable rather than a boolean. Qwen3 returns True, so input_audio_buffer.commit creates the response automatically; adapters without the hook keep the existing response.create behavior.
- last_turn_interrupted is cleared in on_turn_request_issued, after the chat request has been issued, rather than from turn_policy_messages / _build_chat_request.
- Added a --deploy-config qwen3_omni_duplex.yaml + /v1/duplex example to docs/serving/realtime_duplex_api.md.
- is_enabled() now always returns True, with a docstring noting that this adapter shouldn't be attached to another pipeline's PipelineConfig.

**DCO, Python 3.11/3.12 builds, and pre-commit all pass.** I also verified the tests on an RTX PRO 6000 with vLLM 0.28.0: 30 Qwen3-specific tests, 224 handler regression tests, and 15 sample-rate propagation tests. Ruff and compileall are clean.

@Sy0307 — the DCO, test/implementation mismatch, and shared commit-policy concerns should all be resolved now. This should be ready to use as the base for the #5867/#6618 VAD work. Could you re-trigger the review-bot triage on the updated head?

NolenLiang:
I reproduced the input-format problem addressed by this PR on main `d774033`, while bringing up the separate cancellation work in #7278.

Environment: Qwen3-Omni-30B-A3B-Instruct revision `26291f7`, vLLM 0.28.0, Python 3.12.3, two GB200 GPUs. The standard three-stage Qwen deployment had `duplex_session` enabled and remained turn-based. All stages loaded; `/health` returned 200.

Reproduction on `/v1/realtime?duplex=1&autostart=0`:

```text
session.update:
  session.input_audio_format = "pcm16"
  session.sample_rate_hz = 16000
  session.audio.input.sample_rate_hz = 16000
  session.turn_detection = null
  session.extra_body.auto_response = false
input_audio_buffer.append:
  audio = base64(raw mono PCM16), format = "pcm16", sample_rate_hz = 16000
  [100 ms chunks from tests/assets/minicpmo_4_5/response_required_16k.wav]
input_audio_buffer.commit: final = true, response_create = true
```

Expected: committed speech starts model inference. Observed: `input_audio_buffer.committed`, then `response.created`, then:

```text
BadRequestError: Invalid or unsupported audio file.
[Errno 1094995529] Invalid data found when processing input: '<none>'
```

The same failure occurred on #7278's `a916cbc8`. Both runs received zero audio bytes and failed before sending `response.cancel`.

The source trace matches this PR's normalization rationale: [manual append](https://github.com/vllm-project/vllm-omni/blob/d7740336ff94d23276d690e3cd74a1ddb25d23cb/vllm_omni/entrypoints/duplex/realtime_input.py#L370-L383) converts PCM16 to raw float32; [the converter](https://github.com/vllm-project/vllm-omni/blob/d7740336ff94d23276d690e3cd74a1ddb25d23cb/vllm_omni/entrypoints/duplex/audio.py#L154-L189) returns `pcm_f32le`; [commit_user_input](https://github.com/vllm-project/vllm-omni/blob/d7740336ff94d23276d690e3cd74a1ddb25d23cb/vllm_omni/entrypoints/duplex/protocol.py#L831-L857) embeds those headerless bytes directly in an `audio_url`. The server traceback reaches vLLM's file decoder and fails at `av.open`.

Could the PCM-to-WAV normalization already proposed here be split into a small main-based change for the shared manual-commit path? #5626 mentioned the same underlying error but addressed error reporting. This evidence concerns manual PCM input; it does not establish a server-VAD or native-duplex failure, and these particular runs provide no cancellation measurements.

JoseCarlosGarcia95:
Thanks for reproducing this independently and for the detailed trace.

Yes, splitting the PCM-to-WAV normalization into a small main-based fix for the shared manual-commit path makes sense to me. It is independent enough from the Qwen3-Omni adapter work in #6372, and having it land separately would also simplify the remaining diff there.

I'm happy to extract that part into a focused PR and rebase #6372 on top of it. @NolenLiang

vllm_omni/entrypoints/duplex/audio.py
```diff
@@ -189,6 +189,33 @@ def convert_input_audio_with_rate(
...
+        raw = base64.b64decode(audio.strip(), validate=False)
+    except (binascii.Error, ValueError) as exc:
+        raise ValueError("audio payload is not valid base64") from exc
+    if not raw or len(raw) % 4:
+        raise ValueError("pcm_f32le audio payload must contain complete samples")
+
+    samples = np.frombuffer(raw, dtype="<f4")
+    if not np.isfinite(samples).all():
+        raise ValueError("pcm_f32le audio payload contains non-finite samples")
+    pcm16 = np.clip(samples, -1.0, 1.0)
```
lishunyang12: Reuse encode_float32_mono_wav_base64 here to keep PCM encoding consistent.

tests/entrypoints/openai_api/test_duplex_input_audio.py
```diff
@@ -0,0 +1,45 @@
...
+import io
+import wave
+
+import numpy as np
+
+from vllm_omni.entrypoints.duplex.audio import pcm_f32le_payload_to_wav
+from vllm_omni.entrypoints.duplex.chat_fallback import _audio_metadata
+
+
+def test_pcm_f32le_payload_is_wrapped_as_wav():
```
lishunyang12, line 21: Add core_model and cpu markers so CI actually runs these audio tests.

PR #6374, opened 2026-08-19
[Diffusion] Chunked DLO weight transport with H2D/AllGather overlap (stage 1)

leonail1:
# Chunked DLO Weight Transport (PR #6374): Scope, Problem, Changes, Roadmap

@asukaqaq-s 

## 1. Where It Applies

**Models.** vLLM-Omni diffusion models requiring distributed layerwise offload (DLO), validated on Cosmos3-Super and MiniMax-H3.

**Hardware.**

- Consumer-grade GPUs with limited VRAM.
- Validated on Ascend 910B1 (CANN 9.0.0, HCCL).

**Scenario.** DLO lets diffusion models whose weights exceed device memory stream DiT blocks from host memory.

## 2. Why This PR: the Problem

The current DLO implementation has several issues:

1. On systems with inter-GPU interconnects supported by NCCL/HCCL, the next layer’s H2D transfer and AllGather block each other, preventing effective concurrent utilization of both PCIe and interconnect bandwidth.
2. Contention between the current layer’s computation/communication and the next layer’s data transfer.

To address these issues, we developed a chunk-granularity data transfer pipeline:

1. `chunk` divides each block into fixed-size chunks, overlaps H2D transfers with AllGather at the chunk level, and further hides data transfer overhead behind computation.
2. Future fine-grained scheduling can control when each chunk is issued on the communication stream to reduce contention with computation and SP communication.

## 3. What This PR Changes (by Layer)

**Offload backend** (`distributed_layerwise_backend.py`). Per-chunk pipeline: H2D and AllGather overlap for each chunk, while two device slots alternate between transfer and compute.

**Transport backend** (`weight_transport_backend.py`). Implements the chunked H2D + AllGather transport contract used by the two alternating transfer/compute slots.

## 4. What Is Next

- **Fine-grained scheduling**: Controls which chunk runs at each moment to prevent contention between data transfer and computation.
- **Fine-grained data transfer**: An Attention/MoE slot pipeline fetches the next slot while the current slot is computing, reducing exposed transfer overhead and device-buffer usage. A persistent CUDA Graph path for NCCL remains future work.

Attention/MoE slot pipeline

## 5. Experimental Results

Environment: Ascend 910B1, CANN 9.0.0/HCCL, BF16, one request at a time, and 64 MiB chunks. Each arm includes 2 warmup requests and 5 measured requests. Values are mean HTTP end-to-end latency ± sample standard deviation.

| Model | Configuration | Baseline (s) | Chunk (s) | Latency reduction | Output check |
|---|---|---:|---:|---:|---|
| Qwen-Image | 512×512, SP2, 4 steps | 15.867 ± 0.079 | 10.665 ± 0.188 | 32.79% | Matching SHA256 |
| FLUX.2-dev | 512×512, SP2, 4 steps | 15.358 ± 0.126 | 8.186 ± 0.192 | 46.70% | Matching SHA256 |
| MiniMax-H3 Ref2VA | 1344×768, SP4, 8 steps | 94.059 ± 0.258 | 92.881 ± 3.742 | 1.25% | Matching SHA256 for MP4 and decoded audio/video |
| Wan2.2-I2V-A14B | 832×480, SP4, 4 steps | 7.737 ± 0.171 | 7.178 ± 0.080 | 7.23% | Matching SHA256 |

Additional profiling attribution: Baseline → Chunk. Exposed transfer and Compute are rank 0 means across accepted profiles (absolute latency perturbation <5%). End-to-end reduction is the difference between the means of 4 adjacent unprofiled requests per arm. Exposed transfer is the duration of the H2D ∪ weight AllGather intervals that do not overlap computation.

| Model | Exposed transfer (s) | Compute (s) | End-to-end reduction (s) |
|---|---:|---:|---:|
| Qwen-Image | 15.047 → 8.847 | 1.103 → 1.270 | 6.081 |
| FLUX.2-dev | 11.014 → 6.133 | 1.518 → 1.975 | 4.345 |
| Wan2.2-I2V-A14B | 1.152 → 1.167 | 4.682 → 4.048 | 0.768 |
| MiniMax-H3 | 9.171 → 4.579 | 75.228 → 75.935 | 2.959 |


hsliuustc0106:
please submit an issue to the weekly meeting

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
```diff
@@ -1103,57 +1341,599 @@ def _load_weights_via_mmap(
...
                 "torch.distributed is not initialized. "
-                "Distributed layerwise offload with dp_size > 1 requires "
+                "Distributed layerwise offload with weight_shard_size > 1 requires "
                 "an initialized process group."
             )
 
-        from vllm_omni.diffusion.distributed.parallel_state import (
-            get_data_parallel_world_size,
-            get_dp_group,
+        if not self.config.use_hsdp:
```
hsliuustc0106, line 1657: This drops the existing DP/SP AllGather resolver. For non-HSDP DLO, `OffloadConfig` still derives `weight_shard_size` from DP (or SP when DP=1), so every established configuration with a shard size greater than one now reaches this branch and fails during `enable()`. Please keep the new FS-group path for HSDP, but retain the previous DP/SP group selection for non-HSDP, with a test that reaches `backend.enable()`.

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
```diff
@@ -1551,46 +2278,65 @@ def enable(self, pipeline: nn.Module) -> None:
...
+
+        # 1. Ownership + stable block ids (small consecutive ints so the trace
+        #    analyzer sees contiguous block ranges).
+        ownership = self._resolve_chunk_ownership(pipeline)
+
+        # 2. Plan every final layout and enforce the pin cap before allocation.
+        self._plan_chunk_manifests(ownership)
+
+        # 3. Pack Host shards; this replaces the original parameter storage.
+        self._prepare_host_storage(ownership)
```
hsliuustc0106: This also runs for `weight_shard_size == 1`, including no-AllGather and effective single-rank mmap. It copies every mapped block into private per-process host shards and then closes the mappings at lines 2383–2388, removing the shared-page plus bounded-staging path required by #6195; the two former mmap-retention tests are now skipped. Host memory therefore scales with replica count and can OOM. Keep rank-local mmap on the existing path and scope chunk packing to the sharded AllGather path.

hsliuustc0106:
The reported 1.9× E2E table is not a base-vs-head comparison: its baseline forces this PR into the pageable whole-block fallback with a tiny pin budget, while current main's AllGather path uses pinned host shards. Please provide a same-hardware base/head A/B with identical inputs, warmup, measured repetitions/variability, memory metrics, and an output-quality guard.

Please also update the DLO design and user docs before re-review. They currently say HSDP + DLO AllGather is rejected and no-AllGather retains shared mmap pages, both of which this PR changes. Given the architectural scope, the weekly design issue/RFC should define these contracts first.

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
```diff
@@ -1103,57 +1341,599 @@ def _load_weights_via_mmap(
...
+        ):
+            local = tensor.to_local() if hasattr(tensor, "to_local") else tensor
+            transform = self._mmap_transforms_by_tensor_id.get(id(tensor))
+            if transform is None:
+                # Legacy attribute path kept for tests that install the
+                # transform directly on the tensor.
+                transform = getattr(tensor, "mmap_weight_transform", None)
+                if not getattr(tensor, "mmap_weight_transform_pending", False):
+                    transform = None
+            if callable(transform):
```
RuixiangMa, line 1782: he new _block_tensor_specs applies transforms without any validation， If a transform accidentally changes dtype or shape, the manifest will contain incorrect metadata, causing silent data corruption during transport

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
```diff
@@ -1746,8 +2431,18 @@ def disable(self) -> None:
         if not self.enabled and not hasattr(self, "_mmap_file_cache"):
             return
 
-        if self._using_rank_local_mmap:
-            current_omni_platform.synchronize()
+        # Join outstanding submits and retire every bootstrap/tail ticket
+        # before closing the transport state.
+        for group in self._all_hook_groups:
```
RuixiangMa, line 2904: If any hook's drain_request() raises (e.g., from future.result()), the remaining hooks won't be drained or closed, leaking pinned memory and threads. Consider wrapping each hook's drain/close in try/except, or using a single loop that drains and closes each hook before moving to the next.

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
```diff
@@ -1103,57 +1341,599 @@ def _load_weights_via_mmap(
...
+            manifest = self._planned_manifests[id(module)]
+
+            fallback_reason = self._forced_fallback_reason
+
+            failed = 0
+            detail = fallback_reason
+            cpu_shards: dict[torch.dtype, torch.Tensor] | None = None
+            if fallback_reason is None:
+                try:
+                    cpu_shards = pack_local_shard(chunk_specs, manifest)
```
RuixiangMa, line 1895: pack_local_shard → _default_pinned_allocator → torch.empty(..., pin_memory=True) can raise RuntimeError (pinned memory exhausted, CUDA driver unavailable) or OSError. These would propagate uncaught, leaving the backend in a partially initialized state where some blocks are packed and others aren't.

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
leonail1: Fixed in 206bb6fb: `_init_weight_shard_group` keeps the FS-group path for HSDP and restores the DP/SP resolver for non-HSDP (DP > 1 → DP group; DP = 1 with SP > 1 → SP group, with a size-consistency check). Three resolver regression tests added; suite 90 passed.

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
leonail1: Fixed in 45822b24. Chunked packing is now scoped to `weight_shard_size > 1`. For `weight_shard_size == 1` with a `checkpoint_mmap` plan, the backend retains the mmap views (handles released at `disable()`, not at `enable()`) and stages one block at a time through two bounded pinned staging slots — the shared-page + bounded-staging path from #6195 is restored; no private per-block shards are allocated, and resident layers retain their mmap sources. The two skipped tests are un-skipped and passing; suite: 93 passed.

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
leonail1: Fixed in b983216f: transforms are validated at application — numel changes and dtype-family changes raise (view-only/stride transforms stay legal); regression tests added.

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
leonail1: Fixed in b983216f: `_drain_transport` and `disable()` now drain/close every hook even when one fails, then re-raise the first error after cleanup; a test proves later hooks still drain.

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
leonail1: Fixed in b983216f: all pinned-allocator failures (not only MemoryError) go through the FS vote, and an agreed failure drops the partially packed host storage and closes mmap handles before raising.

leonail1:
A/B test completed. Configuration: DP=4, 4×910B1, 4 identical concurrent requests, 40 steps, 1 warmup round + 2 measured rounds. Mean round wall across the two measured rounds: 314.6 s → 244.6 s (-22.2%, 1.29×). Outputs confirmed identical. Peak device memory 7586–7720 → 7416 MB.

leonail1:
Docs updated in 4232f6ec: the design doc now covers the HSDP support contract (single-ownership split, FS-axis-only weight collectives), adds a chunked weight transport section, and scopes chunk packing to the sharded AllGather path; the user guide documents the three new flags and the HSDP change.

hsliuustc0106:
DCO is currently `action_required` on head `4232f6ec`. Please sign the offending commit(s) with `git commit -s` and update the branch; I’ll continue the architecture review after the gate is green.

leonail1:
DCO fixed: 45822b24 was missing the trailer; the branch has been re-signed (all commits now carry Signed-off-by) and force-pushed as 8002b542. Suite still green (96 passed).

hsliuustc0106:
check #6414

hsliuustc0106:
## Independent local MiniMax-H3 validation

Tested PR head 9867b97a9bd76c3556ea6f8d871d8220e69959f0 against the local base checkout 72ee535fcde303a4802b9b2e9663fad23d6f7cdb on the same host. The important result is that the full end-to-end H3 A/B did not reach readiness on either revision, so there is no valid H3 request-level speedup or output-quality result to report. I did obtain a transport-only mechanism result below.

### Environment and configuration

- 8 x NVIDIA L20X host; the H3 attempts used four otherwise-free GPUs.
- Python 3.12.13, vLLM 0.27.0, PyTorch 2.13.0+cu129, transformers 5.14.1, local HF cache under /data/models.
- Local MiniMaxAI/MiniMax-H3 Ref2VA snapshot; BF16; 1344x768; 8 denoising steps.
- DLO configuration: DP=2, USP=2, sharded AllGather path.
- The API runner was configured for one warmup wave and three measured waves, with the same inputs for base and head.

### End-to-end failure (base and head)

Repeated API attempts on both revisions exited while waiting for server readiness, before the first request was submitted. The relevant logs are consistent:

1. Diffusion workers report: Worker N: Shutdown signal received, starting cleanup.
2. The API process reports: Rank 0 scheduler is dead, followed by Exit code: 130.
3. Stage initialization reports that it is shutting down 0 initialized clients.
4. The orchestrator then fails with EOFError while receiving from the multiprocessing connection.

The same failure was reproduced on the head with the direct MiniMax-H3 lifecycle example in two controls: a 2-GPU DLO run (2 steps) and a 1-GPU rank-local no-AllGather run (1 step). Thus the observed failure is before request execution and is not evidence of a head-only transport regression; the base revision fails in the same worker-startup phase, and the rank-local control does not exercise the new AllGather transport. The underlying worker-startup cause is still unresolved.

No client_results.json, latency/throughput samples, peak-memory comparison, generated media, or output hashes were produced. Consequently, the PR's request-level H3 performance and output-equivalence claims remain unverified locally.

### Transport-only A/B

I also ran a pure PyTorch/NCCL microbenchmark that mirrors the whole-block serial schedule and the PR's 64 MiB chunked schedule. It uses the measured MiniMax-H3 block footprint (1,291,143,680 BF16 bytes), two NCCL ranks on two L20X GPUs, five warmups, 20 measured transfers, and validates the gathered payload on every run.

| schedule | run 1 group max | run 2 group max | two-run mean |
| --- | ---: | ---: | ---: |
| whole-block serial | 14.146 ms | 14.235 ms | 14.190 ms |
| 64 MiB chunked | 12.621 ms | 11.913 ms | 12.267 ms |

- Transport-only speedup: 1.1568x, or 13.55% lower transfer time.
- Payload validation passed on all four result files.
- Explicit staging bytes in the harness: 3,873,431,040 -> 2,649,396,224 bytes (31.6% lower).

This is mechanism-level evidence only. The microbenchmark has no DiT compute, scheduler behavior, request batching, end-to-end memory pressure, or media-quality comparison, so the 1.16x number must not be interpreted as MiniMax-H3 request-level speedup.

### Additional local validation

- ruff check, ruff format --check, compileall, and git diff --check passed on the head worktree.
- Targeted offloader tests passed: 61 on the base checkout and 70 on the head; the MiniMax-H3 CPU contract/parallel tests passed 29 on the head.
- A CLI parser smoke test rejects the newly documented --dlo-chunk-size-mb, --dlo-pin-budget-gb, and --dlo-pin-failure-policy options as unrecognized, even though the corresponding fields exist in arg_utils.py.
- The HSDP+DLO+AllGather path currently imports get_fs_group from parallel_state.py, but that symbol is not exported there; a direct import reproduces ImportError before backend enable().

I added inline comments for the HSDP import, stepwise DLO lifecycle, and missing CLI registration. Given the startup blocker and the absence of a full H3 A/B, I recommend holding approval until the worker initialization is made reproducible and the full experiment is rerun with at least one warmup and three measured rounds, reporting mean/stddev, per-step or request latency, throughput, peak VRAM, and output hashes/quality checks.

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
```diff
@@ -1103,57 +1628,745 @@ def _load_weights_via_mmap(
...
-                "Distributed layerwise offload: DP=1, using SP group (world_size=%d) for weight sharding",
-                coord.world_size,
+                "Distributed layerwise offload (non-HSDP): weight_shard_size=%d, rank_in_group=%d, group_ranks=%s",
+                self.weight_shard_size,
+                self.weight_shard_rank,
+                coord.ranks,
+            )
+            return
+
+        from vllm_omni.diffusion.distributed.parallel_state import get_fs_group
```
hsliuustc0106: [P1] The newly documented HSDP + DLO + AllGather path cannot initialize here: importing get_fs_group from vllm_omni.diffusion.distributed.parallel_state raises ImportError because that module exports no such symbol. _init_weight_shard_group reaches this import when use_hsdp is true, so the failure occurs during backend enable() before any transport work. Please wire this to the actual FS-group accessor (or add/export the accessor) and add a runtime enable() test for HSDP.

vllm_omni/diffusion/worker/diffusion_model_runner.py
```diff
@@ -582,7 +633,9 @@ def _execute_request_list(
         use_hsdp = od_config.parallel_config.use_hsdp
         use_distributed_offload = getattr(self.od_config, "enable_distributed_layerwise_offload", False)
         grad_context = torch.no_grad() if (use_hsdp or use_distributed_offload) else torch.inference_mode()
-        with grad_context:
+        request_context = getattr(self.offload_backend, "request_context", None)
```
hsliuustc0106, line 665: [P1] This request_context only surrounds the request-list path. execute_stepwise() calls _execute_stepwise() separately, and that path has no request_context, so stepwise DLO requests bypass the begin/end generation lifecycle that owns the ticket and future state. A multi-step request can therefore leave transport state live across scheduler steps or request boundaries. Please wrap the complete stepwise request lifecycle, or reject DLO + stepwise mode until it is supported and tested.

vllm_omni/engine/arg_utils.py
```diff
@@ -548,6 +548,9 @@ class OrchestratorArgs:
     enable_distributed_layerwise_offload: bool = False
     dlo_use_allgather: bool = True
     dlo_resident_layers: int = 0
+    dlo_chunk_size_mb: int = 64
```
hsliuustc0106, line 551: [P2] These new configuration fields are not exposed by the public serve parser. A local OmniServeCommand smoke test rejects --dlo-chunk-size-mb, --dlo-pin-budget-gb, and --dlo-pin-failure-policy as unrecognized; serve.py registers options only through --dlo-resident-layers. This makes the documented chunk/budget/failure-policy controls inaccessible from the CLI and prevents reproducing the intended benchmark configuration. Please register all three options and add parser coverage.

hsliuustc0106 (commented):
The local MiniMax-H3 experiment and the pre-request failures are documented in the top-level comment. These inline issues also need to be addressed before the new paths can be considered ready.

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
leonail1: Fixed in f9034eca: parallel_state's FS group was removed by #5531; `_init_weight_shard_group` now builds the contiguous replica-major FS rows itself with `new_group` (device + gloo), with divisibility validation. FS-row builder tests added.

vllm_omni/diffusion/worker/diffusion_model_runner.py
leonail1: Fixed in f9034eca: `_execute_stepwise` now enters the offload backend's `request_context`, same generation lifecycle as the request-list path; a test asserts the context is entered.

vllm_omni/engine/arg_utils.py
leonail1: Fixed in f9034eca: serve now registers `--dlo-chunk-size-mb` / `--dlo-pin-budget-gb` / `--dlo-pin-failure-policy`, with a passthrough test asserting they reach `engine_args`.

hsliuustc0106:
## Authoritative MiniMax-H3 Ref2VA A/B on gpu run

This supersedes my earlier pre-readiness report. After launching through the GPU reservation system, I completed the same full request benchmark on both revisions:

- Base: bd4f9acfd30456cb8fa98af53d32f7adc34e03a0
- Head: 6e226a6ab99c2e9ca3937dd1e96e74c57078b599
- Hardware: 4 x NVIDIA L20X, allocated as GPUs 0/2/3/4
- Python 3.12.13, vLLM 0.27.0, PyTorch 2.13.0+cu129, BF16
- Local MiniMaxAI/MiniMax-H3 Ref2VA checkpoint, 1344x768, 8 steps
- DP=2, USP=2, DLO AllGather, two concurrent requests per wave
- Same inputs, one warmup wave excluded from the three measured waves

All 16 requests (2 branches x 4 waves x 2 requests) returned HTTP 200 and produced valid MP4 outputs.

| metric | base | head | head vs base |
| --- | ---: | ---: | ---: |
| measured wave wall mean +/- sample std | 136.861 +/- 9.257 s | 138.052 +/- 1.689 s | +0.87% slower |
| DLO stage-generation mean +/- sample std | 134.188 +/- 9.142 s | 135.447 +/- 1.575 s | +0.94% slower |
| per-request wall mean +/- sample std | 136.167 +/- 8.265 s | 137.407 +/- 1.645 s | +0.91% slower |
| peak device memory, max across 4 GPUs | 26,243 MiB | 27,028 MiB | +2.99% |
| peak process-tree PSS | 361.9 GiB | 373.5 GiB | +3.21% |

Measured wave wall times were base [145.858, 137.362, 127.363] s and head [139.888, 137.702, 136.565] s. Corresponding base/head warmup and measured outputs had identical SHA256 hashes (the two deterministic measured output hashes were b7632bbc... and 1e67a96f...), identical byte sizes, and matching video/audio shapes. This is a bit-identical output guard, not just a visual comparison.

Interpretation: this H3 run does not reproduce a positive end-to-end speedup; the head is effectively flat while the base drops across successive waves. That trend indicates one warmup wave is insufficient to remove page-cache/compile effects, so the mean comparison should be treated as provisional rather than as a precise regression measurement. A follow-up with at least three warmup waves and randomized/interleaved branch order is needed for a stable estimate; my queued longer pass could not acquire four GPUs without displacing other reserved jobs.

The transport-only microbenchmark remains positive at 1.1568x / 13.55% lower transfer time, but that mechanism result does not translate into a measured H3 request-level gain on this hardware/configuration. The PR's published 1.29x result is on a different 4x910B1/Cosmos3 workload and is not independently confirmed by this H3 A/B.

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
```diff
@@ -598,59 +951,164 @@ def get_weights(self, slot: int) -> dict[torch.dtype, torch.Tensor] | None:
...
+            self._last_use_events[slot] = evt
+        evt.record(compute_stream)
+        self._output_slot_events[slot] = evt
+
+        prev = self._prev_hook
+        if prev is not None:
+            ticket = prev.ready_tickets[slot]
+            if ticket is not None and prev.transport_state.is_current(ticket):
+                prev.transport.record_last_use(ticket, evt)
+                prev.ready_tickets[slot] = None
```
hsliuustc0106, line 975: [P1] Clear prev.ready_events[slot] together with prev.ready_tickets[slot]. get_weights() falls back to the previous hook's ready event when its own event is absent, then validates that same owner's ticket. After this path retires the ticket, a later three-block ring reuse can observe the stale event with a missing ticket and fail instead of treating the slot as reusable. The overwrite path above already clears both fields; this retirement path should do the same.

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
```diff
@@ -1747,8 +2887,36 @@ def disable(self) -> None:
...
+        # the slot state machine is locked against further use.
+        for group in self._all_hook_groups:
+            for hook in group:
+                try:
+                    if hook.transport is not None:
+                        hook.transport.close()
+                except BaseException as exc:  # noqa: BLE001
+                    if first_error is None:
+                        first_error = exc
+                hook.cpu_shards = {}  # drop pinned-memory references
```
hsliuustc0106, line 2922: [P1] disable() drops hook.cpu_shards, but the backend's _prepared_host_parts still owns the prepared part dictionaries and their cpu_shards references. Those entries are populated by _prepare_host_storage and are only cleared by _abort_prepared_host_storage, so a normal disable retains pinned host tensors/manifests (the H3 run reports 30.780 GiB prepared) and stale transport metrics across the backend lifetime. Clear the backend-owned prepared storage/pin budget here after draining, or make ownership single-source.

hsliuustc0106 (commented):
The refreshed 4-GPU H3 A/B now completes on both revisions, but these two lifecycle issues remain independent of the benchmark result.

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
leonail1: Fixed in 4f7decb7: the retirement path in `offload_layer()` now clears `prev.ready_events[slot]` together with `prev.ready_tickets[slot]`, keeping the pair-retirement invariant used by the other two paths. A regression test asserts both fields are cleared.

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
leonail1: Fixed in 4f7decb7: `disable()` now clears `_prepared_host_parts` (which aliases the same pinned tensors as `hook.cpu_shards`), resets `_planned_manifests`, and releases the pin budget, so enable/disable cycles no longer retain pinned host memory or stale transport metrics. Covered by a new test.

hsliuustc0106:
## Independent MiniMax-H3 profiling follow-up — requesting discussion and verification

I ran an independent local check of the DLO transport claim and would like to discuss the workload regime where the proposed overlap is expected to win.

### Full Ref2VA A/B (same host, 4 x L20X)

The earlier full request A/B used one warmup plus three measured waves, DP=2, USP=2, BF16, 1344x768, 8 steps, and two concurrent Ref2VA requests. All 16 requests completed with HTTP 200 and corresponding outputs were byte-identical by SHA256.

| metric | base | head | delta |
| --- | ---: | ---: | ---: |
| measured wave wall mean +/- sample std | 136.861 +/- 9.257 s | 138.052 +/- 1.689 s | head +0.87% slower |
| peak device memory | 26,243 MiB | 27,028 MiB | +2.99% |

The base timing decreased across its three measured waves (145.858, 137.362, 127.363 s), while the head was flatter (139.888, 137.702, 136.565 s). Therefore I treat this as evidence of no net gain, but not as a perfectly stabilized estimate; more warmups or randomized branch order would be useful.

### Nsight Systems, current base/head

I captured the current PR base 17285c2f and head db7e0fdb with Nsight Systems 2026.1.3, CUDA/NVTX/OSRT/NCCL tracing, two L20X GPUs, DP=2, BF16, 512x288, and two actual denoise transitions. The model was FL2VA from the mandated /data/model cache. Temporary NVTX ranges mark denoise_step[0] and denoise_step[1].

`--steps=3` is intentional here: MiniMax-H3 executes len(sigmas)-1 transitions, so `--steps=2` produces only one actual transition.

| transition | base wall | head wall | base compute | head compute | base H2D | head H2D | base AllGather | head AllGather |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| 0 | 2,845 ms | 3,940 ms | 334 ms | 349 ms | 691 ms | 684 ms | 1,765 ms | 2,980 ms |
| 1 | 849 ms | 816 ms | 334 ms | 340 ms | 683 ms | 689 ms | 159 ms | 282 ms |

Base has effectively zero H2D/AllGather overlap. Head shows 144–260 ms overlap in transition 0 and 142–222 ms in transition 1, so the intended mechanism is visible. However, the head also launches roughly 19x as many AllGather kernels per rank (about 1,044 vs 54) and 16x as many H2D copies (about 1,054 vs 64). Across both ranks, kernel-launch API time rises from 668.8 to 986.2 ms in transition 0, and H2D API time rises from 3.4 to 134.6 ms.

### Current interpretation

The profiled bottleneck is not DiT compute. Transition 0 is dominated by chunked transport submission and AllGather overhead; transition 1 is primarily H2D-lane limited. The isolated transport microbenchmark is positive (1.1568x, 13.55% lower transfer time), but that benefit is not currently translating into a request-level H3 speedup.

### Requested discussion / verification

Could the authors and maintainers help verify the following?

1. What model size, GPU topology, chunk size, and denoise-step regime is expected to amortize the additional per-chunk launch/collective overhead?
2. Please provide a current-head Nsight trace on the intended 4x910B1/Cosmos3 workload, including per-step host launch, H2D, NCCL AllGather, compute-kernel, and overlap breakdown.
3. Please rerun the full MiniMax-H3 Ref2VA A/B on the current head after the lifecycle fixes, with at least three warmup waves and three measured waves (or randomized base/head order), reporting mean +/- stddev, peak VRAM, and output hashes/quality checks.
4. Please explain whether the expected gain is only at larger blocks/batches or whether the current H3 configuration should also improve. If the latter, the launch-count and step-0 AllGather overhead need further investigation.

Detailed local Nsight artifacts and the per-transition report are available in the reviewer environment; the results above are the reproducible summary.

leonail1:
@hsliuustc0106

Thanks for the questions. We reran the experiments on the current code and collected the following results.

## Q1: When is the chunk launch overhead amortized?

Using the same H3 model and inputs as in the comment, but on 2x910B1/HCCL, `chunk` has 3.88 ms/block of host-launch time. Even without persistent, it is hidden from the critical path.

Cosmos3 shows the same behavior: `chunk` has 99.76% H2D/compute overlap and only 0.002 ms/transition of exposed transport. `chunk+persistent` reduces host launch from 7.150 to 2.451 ms/transition, and aggregate launch time from 27.90 to 16.05 s; wall time is essentially unchanged (28.962 to 29.274 ms/transition).

Using the same CANN launch-API accounting, after skipping the first 128 transitions:

| Arm | Launch calls (per transition) | Host launch (ms/transition) |
|---|---:|---:|
| chunk | 165 | 7.150 |
| chunk+persistent | 150 | 2.451 |

`chunk+persistent` is an opt-in NPU feature: HCCL AllGather is captured once for stable buffers/shapes and replayed with NPUGraph. It errors directly when NPUGraph is unsupported.

Conclusion: on 910B/HCCL, `chunk` already hides launch; `chunk+persistent` further reduces launch and is useful on platforms where launch is expensive.

## Q2: 4x910B1/Cosmos3 trace breakdown

Configuration: 4x910B1, CANN 9.0.0, Cosmos3-Super, DP=1/SP=4, 832x480x29 frames, 4 steps, seed 42, four concurrent requests, 64 MiB chunks.

The following is the rank-0 steady-state median after skipping the first 128 transitions:

| Metric (ms/transition) | base | chunk | chunk+persistent |
|---|---:|---:|---:|
| wall | 28.526 | 28.962 | 29.274 |
| compute | 22.926 | 28.942 | 29.238 |
| H2D | 10.743 | 13.098 | 6.906 |
| HCCL AllGather | 17.088 | 28.054 | 28.516 |
| host launch | 9.937 | 7.150 | 2.451 |
| exposed transport | 4.824 | 0.002 | 0.016 |

H2D/compute overlap is 35.56% for `base`, 99.76% for `chunk`, and 99.45% for `chunk+persistent`. 910B uses HCCL; HCCL AllGather is the counterpart of the NCCL AllGather item requested in the comment.

We also reprofiled current head (`16460474`; SP=4 path equivalent to the profiled commit). We aggregate from `pipeline_cosmos3.py:2612 diffuse` to the end of consecutive `_step` calls; values below are mean +/- standard deviation across four concurrent requests. Step 0 has 256 block transitions; later steps have 128 each.

The 256 transitions in step 0 come from two cond/uncond forwards: each CFG branch has 64 UND + 64 GEN blocks, so `2 x (64+64)`; later steps reuse UND K/V and execute only `2 x 64 = 128`.

Current-head per-denoise-step aggregation (ms):

| Step | Arm | Wall | Compute | H2D | HCCL AllGather | H2D/compute overlap | AllGather/compute overlap | Host launch (calls; time) | Exposed transport |
|---:|---|---:|---:|---:|---:|---:|---:|---|---:|
| 0 | chunk | 10607.5 +/- 136.1 | 10558.8 +/- 142.4 | 2823.8 +/- 43.2 | 10352.3 +/- 139.4 | 2814.2 +/- 45.0 | 10352.3 +/- 139.4 | 39838; 2457.2 +/- 44.1 | 9.59 +/- 12.29 |
| 0 | chunk+persistent | 10835.2 +/- 718.3 | 10746.5 +/- 700.5 | 3872.0 +/- 243.0 | 10470.5 +/- 694.4 | 3840.7 +/- 238.2 | 10470.5 +/- 694.4 | 35938; 1248.2 +/- 66.7 | 31.28 +/- 5.82 |
| 1 | chunk | 5295.6 +/- 94.2 | 5290.2 +/- 95.6 | 1429.6 +/- 7.9 | 5184.3 +/- 95.4 | 1428.6 +/- 8.2 | 5184.3 +/- 95.4 | 21418; 970.0 +/- 18.3 | 0.96 +/- 0.50 |
| 1 | chunk+persistent | 5380.4 +/- 250.1 | 5358.7 +/- 249.4 | 1781.0 +/- 129.3 | 5132.3 +/- 264.2 | 1776.5 +/- 129.6 | 5132.3 +/- 264.2 | 19498; 364.7 +/- 21.5 | 4.57 +/- 1.11 |
| 2 | chunk | 5270.4 +/- 26.4 | 5264.6 +/- 27.8 | 1432.5 +/- 20.4 | 5157.0 +/- 26.8 | 1431.6 +/- 20.8 | 5157.0 +/- 26.8 | 21460; 969.9 +/- 12.0 | 0.88 +/- 0.56 |
| 2 | chunk+persistent | 5343.6 +/- 243.1 | 5315.1 +/- 246.9 | 1575.4 +/- 176.5 | 5124.1 +/- 240.1 | 1569.2 +/- 175.9 | 5124.1 +/- 240.1 | 19540; 352.5 +/- 10.0 | 6.15 +/- 2.56 |
| 3 | chunk | 5225.1 +/- 50.0 | 5213.3 +/- 53.7 | 1427.1 +/- 14.3 | 5107.0 +/- 54.2 | 1425.3 +/- 14.6 | 5107.0 +/- 54.2 | 21428.5; 1017.0 +/- 50.0 | 1.85 +/- 0.98 |
| 3 | chunk+persistent | 5249.1 +/- 139.8 | 5224.2 +/- 136.7 | 1593.0 +/- 136.4 | 5079.2 +/- 115.4 | 1586.7 +/- 135.7 | 5079.2 +/- 115.4 | 19509; 342.5 +/- 16.9 | 6.30 +/- 0.90 |

AllGather fully overlaps compute in this trace; the table also reports H2D/compute overlap and exposed transport. Denoise-step windows use profiler Python API boundaries, preserving actual overlap rather than multiplying transition counts. Current-head profiled wall: `chunk` 122.37 s, `chunk+persistent` 124.71 s; outputs are byte-identical. HSDP-specific changes are on `vllm-dlo/hsdp-dlo`; current head retains the generic SP=4 collective isolation.

## Q3: Full H3 A/B after lifecycle fixes

Configuration: 4x910B1, MiniMax-H3 Ref2VA, DP=1/SP=4, 1344x768, 8 steps, duration 5 s, seed 3100. Three warmup and three profiled requests.

Outputs are byte-identical.

| Metric | base | chunk |
|---|---:|---:|
| Profiled wall (s, mean +/- stddev) | 107.96 +/- 1.82 | 105.43 +/- 0.70 |
| Peak HBM (GiB/card) | 28.6 | 28.5 |
| Peak process-tree PSS (GiB) | 147.0 | 148.5 |

Additional H3 fine-grained trace (mechanism evidence, not a replacement for the current-head E2E): 2x910B1, CANN 9, 512x288, 3 steps, DP=2. `chunk`/`chunk+persistent` use DLO-marker aggregation; `base` uses the corresponding DiT compute window. Units: ms.

| Arm | Wall/block | Compute/block | H2D/block | HCCL AllGather/block | Launch API/block | AG/compute overlap | H2D/compute overlap | Exposed transport/block |
|---|---:|---:|---:|---:|---:|---:|---:|---:|
| base | 168.1 | 157.6 | 28.1 | 32.7 (1 kernel) | 140.0 | 100% | 86% | 4.0 |
| chunk | 38.98 | 38.92 | 7.04 | 11.08 (20.64 kernels) | 3.88 | 100% | 100% | 0.0 |
| chunk+persistent | 86.81* | 86.75* | 14.18 | 19.73 (20.64 kernels) | **0.42** | 100% | approx. 100% | 0.005 |

`chunk` has no visible transport bubble. `chunk+persistent` keeps the same AllGather kernel count but reduces host launch API from 3.88 to 0.42 ms/block (about 9.2x). `*` NPUGraph capture inflates `CAPTURE_WAIT`, so wall is not directly comparable; profiled measured wall is `chunk` 38.9 s and `chunk+persistent` 37.5 s, with identical outputs.

This supplemental trace uses a different configuration from formal Q3; it supports the H3 mechanism only. The Q3 main table remains the current-head, three-warmup/three-profile E2E result.

## Q4: Is a larger block or batch required for a gain?

No. The key factor is whether transport is exposed on the critical path: on 910B/HCCL, `chunk` hides transport and `chunk+persistent` further reduces launch; persistent benefit on L20X/NCCL has not been measured.

hsliuustc0106:
## Follow-up concern: workload and platform dependence

Our concern is about the scope and reproducibility of the performance claim, rather than advocating one particular implementation.

Our independent L20X/NCCL observations show that the chunked path does create H2D/AllGather overlap, but the additional transport submissions dominate the first denoise transition and the full Ref2VA request A/B shows no net end-to-end gain. In the two-transition Nsight trace, the head launched roughly 19x more AllGather kernels and 16x more H2D copies per rank; step 0 was 3.940 s on the head versus 2.845 s on base.

The latest author results show a different outcome on 4x910B1/CANN/HCCL: the current H3 A/B reports 107.96 +/- 1.82 s for base versus 105.43 +/- 0.70 s for chunk, with byte-identical outputs. That is useful evidence, but it is a different accelerator, interconnect, collective implementation, topology, and parallel configuration from the local L20X/NCCL result.

Taken together, the evidence suggests that the benefit is strongly workload/platform dependent. We would like to discuss how broadly the current performance claim should be interpreted and verify the discrepancy between the HCCL/910B result and the NCCL/L20X result. We are intentionally leaving the implementation choice open here; the immediate concern is that the same chunked transport is positive in one environment but not in the other.

vllm_omni/diffusion/offloader/base.py
```diff
@@ -148,9 +160,10 @@ def from_od_config(cls, od_config: OmniDiffusionConfig) -> "OffloadConfig":
                 "the backend will select mmap or standard-loader host storage"
             )
 
-        # HSDP already shards parameters into DTensors.  Running distributed
-        # layerwise offload on top would shard each to_local() again, producing
-        # incorrect reconstruction after AllGather.  Reject this combination.
+        # HSDP already shards parameters into DTensors. Running distributed
```
david6666666, line 160: Looks like the HSDP fix was lost in a later merge. This guard still rejects HSDP + AllGather, and _init_weight_shard_group() now resolves only DP/SP, while the design doc says this combination uses the FS axis. Please restore the FS-group path with an enable() regression test, or keep it documented as unsupported.

vllm_omni/diffusion/offloader/weight_transport_backend.py
```diff
@@ -0,0 +1,423 @@
...
+                torch.distributed.barrier(group=group)
+                torch.distributed.all_gather_into_tensor(full_output, local_input, group=group)
+                torch.npu.synchronize()
+                torch.distributed.barrier(group=group)
+                torch.npu.synchronize()
+                graph = torch.npu.NPUGraph()
+                graph_kwargs = {"stream": streams.communication}
+                if self._graph_pool is not None:
+                    graph_kwargs["pool"] = self._graph_pool
+                with torch.npu.graph(graph, **graph_kwargs):
```
david6666666: Only the AllGather is captured here; _submit_fs_chunk() still runs the H2D copy before this callback. So group_persistent does not replay the advertised H2D + HCCL schedule. Could we either move the copy/event dependency into the graph, or narrow the PR and docs to AllGather-only persistence?

vllm_omni/diffusion/offloader/base.py
leonail1: Thanks for catching this. Fixed in [`88f16cb5`](https://github.com/vllm-project/vllm-omni/commit/88f16cb506494895287709e952d39905c215048d): the design and user docs now explicitly reject HSDP + DLO AllGather and state that this capability is not introduced by this PR. The runtime rejection remains; we did not restore the FS-group path.

vllm_omni/diffusion/offloader/weight_transport_backend.py
leonail1: Thanks. We will treat `group_persistent` as optional follow-up scope and decide whether to include it in this PR after the current experiments. The core scope remains the non-persistent chunked backend. Until that decision, we will not claim that persistent replay captures the full H2D + HCCL schedule; if the backend is retained, its implementation and documentation will be aligned with the measured behavior.

leonail1:
@hsliuustc0106 This PR now shows a measured performance improvement on Cosmos3-Super. Please see the updated PR description for the experimental setup, end-to-end results, transition breakdown, and peak-memory data.

vllm_omni/diffusion/model_loader/diffusers_loader.py
```diff
@@ -1011,10 +1011,54 @@ def _load_model_with_hsdp(
...
+        # wrapping; a parameter with two owners would end up sharded twice.
+        chunk_ownership = None
+        _dist_offload = getattr(self.od_config, "enable_distributed_layerwise_offload", False)
+        _use_ag = getattr(self.od_config, "dlo_use_allgather", True)
+        if _dist_offload and _use_ag:
+            from vllm_omni.diffusion.offloader.block_discovery import (
+                discover_chunk_owned_blocks,
+            )
+
+            resolved = discover_chunk_owned_blocks(model)
```
hsliuustc0106, line 1026: [P3] This discovery only runs under `parallel_config.use_hsdp` + DLO + AllGather — exactly the combination `OffloadConfig.from_od_config` rejects when `enable_offload_backend` runs after `load_model` (diffusion_model_runner.py:307→341) — and the `_dlo_chunk_ownership`/`_dlo_block_ids` attributes set here have no consumer (the backend builds ownership from `_pending_block_groups`). So this branch and `apply_hsdp_to_model(chunk_owned_blocks=...)` can only execute right before a guaranteed startup abort, contradicting the docs' scope note that this PR does not introduce HSDP + AllGather. Suggest dropping both until a PR actually introduces that capability.

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
```diff
@@ -685,59 +1052,169 @@ def get_weights(self, slot: int) -> dict[torch.dtype, torch.Tensor] | None:
...
                 slot_contaminated = self._shared_slot_group[self.current_slot] != self._group_id
+            ticket = self._prev_hook.ready_tickets[self.current_slot]
+            has_current_ticket = ticket is not None and self._prev_hook.transport_state.is_current(ticket)
             if slot_contaminated:
+                if has_current_ticket:
+                    # Another group overwrote the buffer while our ticket was
+                    # still live.  Retire it so the re-prefetch really submits
+                    # instead of hitting the idempotent guard.
+                    self._prev_hook.transport.record_last_use(ticket, self._output_slot_events[self.current_slot])
+                    self._prev_hook.ready_tickets[self.current_slot] = None
```
hsliuustc0106, line 1171: [P3] This contamination repair retires `ready_tickets[slot]` without clearing `ready_events[slot]`, unlike the other two retirement sites (`_submit_prefetch`, `offload_layer`), which keep the ticket/event pair invariant. Today the following sync `prefetch_layer` overwrites the event, or the `get_weights` fail-fast trips, so the impact is only a confusing "ticket is missing" error on an already-failing path — but clear the pair here too.

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
```diff
@@ -462,26 +602,60 @@ def _shard_and_pin(
...
+    # ------------------------------------------------------------------ #
+    #  Tracing                                                            #
+    # ------------------------------------------------------------------ #
+
+    def _trace_range(self, kind: str, chunk_id: int | None = None, block_id: int | None = None) -> Any:
+        """Return a profiler range whose name the DLO timeline analyzer parses.
+
+        No-op unless ``VLLM_OMNI_DLO_TRACE`` is set.  The marker text must stay
+        in the ``dlo.<kind>.block_<id>[.chunk_<id>]`` shape consumed by
+        ``benchmarks/diffusion/extract_dlo_timeline.py``.
```
hsliuustc0106, line 652: [P3] This docstring points at `benchmarks/diffusion/extract_dlo_timeline.py`, which does not exist at this head (nor on main). Either land the analyzer or drop the file reference and keep only the marker-shape contract.

hsliuustc0106:
Deep-dive re-review at `50fff39f5`: all seven previously agreed fixes are verified in place (FS-row resolver + tests, stepwise `request_context`, CLI flag exposure, paired slot retirement, prepared-host release, transform validation, drain/vote cleanup), and the two open threads from @david6666666 are resolved — HSDP + DLO AllGather is now rejected and documented as out of scope, and the `group_persistent` backend was removed outright, so nothing claims AllGather-only replay.

One evidence-binding item (P3): the body's A/B tables cite base `030fccb2` / chunk `7ab426ec`, but `7ab426ec` is not on the current branch (pre-rebase head), and the delta to `50fff39f5` includes `5c5e9076d` (isolated DLO weight collectives — a new communicator) plus the late-August main merges. Could you re-run the headline Cosmos3/Qwen A/B on the current head, or confirm the transport path is unchanged since `7ab426ec` so the numbers still bind? Three P3 code nits are posted inline.

hsliuustc0106:
After five rounds and the new conflicts against the offloader wave on main (#7209 / #5929 / #7005), I'd like to request slimming this PR down before the rebase. The current shape is 20 files / +3,605 −582, with `distributed_layerwise_backend.py` alone at ~1,980 changed lines (~55% of the diff). Concretely:

1. **Drop the HSDP chunk-ownership handoff** (the `chunk_owned_blocks=` path in `apply_hsdp_to_model` plus the loader-side discovery in `_load_model_with_hsdp`, thread [3926973803](https://github.com/vllm-project/vllm-omni/pull/6374#discussion_r1926973803)). It can only execute immediately before the `base.py` guard aborts startup, so it is dead until the HSDP capability lands — deleting it removes most of the `hsdp.py` and `diffusers_loader.py` hunks and takes two files out of the rebase surface entirely.

2. **Land the foundation as its own PR**: `chunked_transport.py` (675 lines; manifest/pin-budget layer, imports only stdlib + torch) + `weight_transport_backend.py` (319) + the `block_discovery.py` additions + the CPU-side manifest tests moved out of `test_distributed_layerwise_backend.py` into e.g. `test_chunked_transport.py`. All-new files conflict with nothing on main, are fully CPU-testable without GPUs, and would merge quickly.

3. **The transport-backend PR** — `distributed_layerwise_backend.py` + `base.py` + the CLI/config wiring + e2e tests + docs — then carries the semantic rebase alone. All eight current conflicts live there, plus the #7209 plan-resolver re-verification and the #7005 LoRA interplay, so it benefits most from a minimal surface. It also lets the perf A/B bind directly to that PR's head, which would close the still-open evidence-binding thread in the same step.

Happy to fast-track the foundation PR once it's up.

leonail1:
In response to [hsliu’s comment](https://github.com/vllm-project/vllm-omni/pull/6374#issuecomment-5656716972):

> I'd like to request slimming this PR down before the rebase.

Thanks for the review. The PR has been reduced to five files on top of the latest main, retaining only chunked transport and the necessary configuration. Extensions such as the HSDP handoff have been removed. Please see the code changes and PR description for details.

leonail1:
@hsliuustc0106 Thanks for the review. The PR has been rebased onto the latest main and trimmed down to five files, retaining only chunked transport and the necessary configuration. Extensions such as the HSDP handoff have been removed. Please see the code changes and PR description for details.

vllm_omni/diffusion/offloader/chunked_transport.py
```diff
@@ -0,0 +1,287 @@
...
+        raise ValueError(f"divisor must be positive, got {divisor}")
+    return (value + divisor - 1) // divisor
+
+
+def round_up(value: int, alignment: int) -> int:
+    return ceil_div(value, alignment) * alignment
+
+
+def is_chunk_transport_supported(tensor: torch.Tensor) -> bool:
+    return tensor.ndim > 0
```
lishunyang12: Include scalar tensors; FP8 weight_scale is otherwise cleared and never restored.

vllm_omni/diffusion/offloader/base.py
```diff
@@ -202,13 +203,19 @@ def from_od_config(cls, od_config: OmniDiffusionConfig) -> "OffloadConfig":
                 "component in diffusion_offload_config, or disable HSDP."
             )
 
+        chunk_size_mb = int(getattr(od_config, "dlo_chunk_size_mb", 64))
```
lishunyang12: Add dlo_chunk_size_mb to the config schemas; this fallback ignores the requested size.

vllm_omni/diffusion/offloader/chunked_transport.py
```diff
@@ -0,0 +1,287 @@
...
+    return full_chunk_numel, alignment_numel
+
+
+def _storage_numel(tensor: torch.Tensor) -> int:
+    if tensor.numel() == 0:
+        return 0
+    return 1 + sum((size - 1) * st for size, st in zip(tensor.shape, tensor.stride()))
+
+
+def _flat_physical(source: torch.Tensor, storage_numel: int) -> torch.Tensor:
```
lishunyang12: Reuse physical_storage_numel and flatten_physical_storage from tensor_utils instead of duplicating them.

vllm_omni/diffusion/offloader/chunked_transport.py
leonail1: Fixed in 3c01fd46.

`build_part_manifest` now includes 0-d tensors such as FP8 `weight_scale`. They are packed, AllGathered, and restored with `shape=()` / `stride=()`. Tests cover a scalar parameter plus buffer through two offload/prefetch cycles.

vllm_omni/diffusion/offloader/base.py
leonail1: Fixed in 3c01fd46.

`dlo_chunk_size_mb` is now declared on `OrchestratorArgs`, `StageDeployConfig`, `_DiffusionConfigProjection`, and `OmniDiffusionConfig`. `OffloadConfig` reads the field directly, so a requested 32/128 MiB is no longer replaced by the 64 MiB fallback.

vllm_omni/diffusion/offloader/chunked_transport.py
leonail1: Fixed in 3c01fd46.

`chunked_transport.py` now uses `physical_storage_numel` and `flatten_physical_storage`, so strided tensors keep deterministic zero padding in storage holes.

PR #6375, opened 2026-08-19
[Bugfix][Diffusion] Aggregate async RPC completion across worker ranks

tangbinh:
## Summary

Fix a race in request-mode multiprocess diffusion RPCs where rank 0 can publish `COMPUTE_DONE` before another worker publishes an `RPC_RESULT` error. The first result pump previously popped the shared future, so a rank-zero success could silently hide the worker failure.

- successful non-output ranks now acknowledge async RPC completion;
- the executor waits for all executing ranks before resolving success;
- any rank error resolves the caller with an exception immediately;
- timeout and shutdown paths clear the new reply bookkeeping.

## Test Plan

- `pytest -q tests/diffusion/test_result_pump.py tests/diffusion/test_async_output_worker.py`
- 40 passed
- Added both success-first and error-first two-rank regression coverage.

Signed-off-by: Binh Tang <binht@netflix.com>

### DIFF
--- a/tests/diffusion/test_async_output_worker.py
+++ b/tests/diffusion/test_async_output_worker.py
@@ -164,6 +164,27 @@ class TestReturnResultAsyncPath:
         proc._return_result(DiffusionOutput())
         # No crash, no enqueue attempted
 
+    def test_non_output_rank_acknowledges_successful_async_rpc(self):
+        proc = _make_worker_proc(step_execution=False)
+        proc.recv_message = MagicMock(
+            side_effect=[
+                {
+                    "type": "rpc",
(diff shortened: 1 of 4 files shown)


tangbinh:
Self-review: I traced both result-arrival orders and checked that success waits for one reply from every rank, while any rank error still resolves immediately. Timeout and shutdown paths clear the per-request bookkeeping. Focused async RPC tests cover success-first/error-later, error-first, and worker acknowledgements; 40 focused tests passed locally. Pre-commit, wheel builds for Python 3.11/3.12, and DCO are also green.

Issue #6365, opened 2026-08-19
[Performance][VoxCPM2]: TTFA starts queueing at 2 RPS on H100


lolyhop:
### Proposal to improve performance

Under steady open-loop load, VoxCPM2 starts reporting scheduler queue time at
2 RPS. One-second Prometheus samples at this level showed 1-2 running requests
and approximately one waiting request on average. Mean
`request_queue_time_seconds` was 391 ms, while audio TTFA p95 increased from
194 ms at 1 RPS to 842 ms.

Throughput stops scaling at about 5.4 requests/s. From 6 to 10 offered RPS,
mean server inference time remains close to 1.37 s, while mean queue time grows
from 1.85 s to 12.0 s.

We would like to understand what this queue time represents for VoxCPM2. In
particular, is it expected from decode-first scheduling, or is another
admission or model-specific constraint delaying new requests?

This may be related to #3874 / #3882 and #4176. The symptom is also similar to
#5365, but this test uses vLLM-Omni 0.26.0 and fixed-rate open-loop arrivals
rather than a single concurrent burst.

### Report of performance regression

This is not a version-to-version regression report. The measurements below are
a vLLM-Omni 0.26.0 baseline.

#### Setup

- one NVIDIA H100 PCIe 80 GB
- vLLM 0.26.0
- vLLM-Omni 0.26.0
- PyTorch 2.11.0+cu129
- FlashAttention 3
- 24 vCPU cgroup quota
- VoxCPM2-compatible fine-tuned checkpoint
- direct `/v1/audio/speech` request to the vLLM-Omni server; no reverse proxy

Server command (checkpoint path omitted):

```bash
vllm-omni serve <checkpoint> --omni \
  --host 0.0.0.0 \
  --port 8010 \
  --served-model-name voxcpm2 \
  --max-num-seqs 8 \
  --max-num-batched-tokens 4096 \
  --gpu-memory-utilization 0.9 \
  --max-model-len 4096 \
  --hf-overrides '{"voxcpm2_runtime_config":{"vae_decode_every":3,"audio_emit_every":1}}'
```

The remaining VoxCPM2 runtime options use the 0.26.0 dataclass defaults. This
includes batched CFM, batched VAE decode, coalesced audio D2H, fused LocDiT
QKV/MLP and the unified decode graph.

Request body:

```json
{
  "input": "<Russian text, 140-151 characters>",
  "response_format": "pcm",
  "stream": true,
  "stream_format": "audio"
}
```

[...]

PR #6377, opened 2026-08-19
[Core] Add opt-in periodic prefill admission for VoxCPM2

lolyhop:
## Purpose

VoxCPM2's unified decode graph currently defers waiting requests whenever a
decode-ready request is running. This preserves pure decode batches, but an
open-loop TTS workload can build substantial queue time while
`max_num_seqs` still has free slots. This is the behavior reported in #6365.

This PR adds an opt-in latency policy for that case. When
`unified_decode_graph_prefill_interval` is positive, the VoxCPM2 scheduler
pauses decode scheduling for one iteration after the configured number of
decode-only iterations and uses that iteration to admit waiting prefills. The
paused requests keep their request and KV-cache state and resume on the next
iteration.

The default is `0`, so existing deployments keep the current decode-first
behavior. A value of `6` was evaluated as a latency-oriented setting in the workload
below. This is intentionally a configuration knob rather than a new default:
periodic prefill admission reduces first-packet latency below saturation, but it also reduces
peak throughput.

## Test Plan

Unit tests were run against current `main` at
`c3f80502e00b564d7b74d43a5e21fbc6b9da2aed`:

```text
21 passed: focused VoxCPM2 scheduler/config tests
40 passed: tests/core/sched/test_omni_ar_scheduler*.py
```

The GPU A/B used a source-equivalent backport of this scheduler change because
the available host has a CUDA 12.9 PyTorch runtime, while current `main` targets
vLLM 0.27.0/CUDA 13.

- one NVIDIA H100 PCIe 80 GB; 24 vCPU; 36 GiB host RAM
- vLLM 0.26.0 and vLLM-Omni 0.26.0; BF16; FlashAttention 3
- `max_num_seqs=8`, `max_num_batched_tokens=4096`, `max_model_len=4096`
- `vae_decode_every=3`, `audio_emit_every=1`
- five Russian texts, 140–151 characters
- open-loop arrivals, 30 seconds per RPS level, `burstiness=100`
- eight warmup requests before every measured level
- raw PCM16 mono at 48 kHz (`stream=true`, `stream_format=audio`)

The benchmark is the repository's `openai-audio-speech` backend. It forces raw
PCM streaming and records `AUDIO_TTFP` on the first non-empty response-body
chunk, after HTTP headers. `AUDIO_RTF` is request wall time divided by generated
audio duration. Audio throughput is generated audio seconds divided by the
complete benchmark window.

The five-row dataset was stored as `$DATASET_ROOT/en/meta.lst`:

```text
ru001|||Здравствуйте! Ваш заказ уже собран и скоро будет передан курьеру. Пожалуйста, проверьте адрес доставки и оставайтесь на связи в течение дня.
ru002|||Добрый день! Напоминаем, что сегодня заканчивается срок хранения вашей посылки. Заберите её до закрытия пункта или продлите хранение в приложении.
ru003|||Уважаемый продавец, новый заказ ожидает подтверждения. Проверьте наличие товара, подготовьте упаковку и передайте отправление в ближайший пункт приёма.
ru004|||Ваша встреча назначена на завтра, на половину одиннадцатого утра. Если планы изменились, пожалуйста, выберите другое удобное время в личном кабинете.
ru005|||Платёж успешно принят, электронный чек уже доступен в приложении. Сохраните его до получения заказа и обратитесь в поддержку, если заметите ошибку.
```

The server used the same command for both runs; only the final interval changed
between `0` (stock) and `6`:

```bash
vllm-omni serve "$MODEL" --omni \
  --host :: --port 8000 --served-model-name voxcpm2 \
  --max-num-seqs 8 --max-num-batched-tokens 4096 \
  --gpu-memory-utilization 0.9 --max-model-len 4096 \
  --hf-overrides '{"voxcpm2_runtime_config":{"vae_decode_every":3,"audio_emit_every":1,"unified_decode_graph_prefill_interval":6}}'
```

The sweep used the checked-in Omni serving benchmark. `num_prompts = RPS * 30`
keeps the offered-load window at 30 seconds:

```bash
export VLLM_OMNI_BENCH_AUDIO_SAMPLE_RATE=48000

for rps in 1 2 3 4 5 6; do
  n=$((rps * 30))
  vllm-omni bench serve --omni \
    --backend openai-audio-speech \
    --base-url http://[::1]:8000 --endpoint /v1/audio/speech \
    --model "$MODEL" --served-model-name voxcpm2 --tokenizer "$MODEL" \
    --trust-remote-code \
    --dataset-name seed-tts-text --dataset-path "$DATASET_ROOT" \
    --seed-tts-locale en --disable-shuffle \
    --num-prompts "$n" --num-warmups 8 \
    --extra-body '{"voice":"default"}' \
    --request-rate "$rps" --burstiness 100 --max-concurrency 100 \
    --metric-percentiles 50,95 \
    --percentile-metrics audio_ttfp,audio_rtf,audio_duration,audio_underrun \
    --temperature 0 --seed 0 --disable-tqdm --save-result --save-detailed
done
```

## Test Result

The full sweep used `unified_decode_graph_prefill_interval=6` for the patched
run. Values are p50/p95; all requests completed without errors.

| RPS | stock TTFP, ms | interval 6 TTFP, ms | stock RTF | interval 6 RTF | stock audio throughput, s/s | interval 6 audio throughput, s/s |
|---:|---:|---:|---:|---:|---:|---:|
| 1 | 215 / 612 | 156 / 191 | 0.12 / 0.16 | 0.11 / 0.12 | 9.34 | 9.53 |
| 2 | 636 / 1142 | 199 / 209 | 0.17 / 0.22 | 0.13 / 0.14 | 18.38 | 18.77 |
| 3 | 687 / 1195 | 214 / 232 | 0.18 / 0.24 | 0.15 / 0.16 | 28.03 | 28.24 |
| 4 | 779 / 1451 | 237 / 249 | 0.20 / 0.27 | 0.18 / 0.19 | 37.42 | 37.43 |
| 5 | 876 / 1611 | 1656 / 2975 | 0.22 / 0.30 | 0.34 / 0.47 | 45.89 | 43.02 |
| 6 | 2658 / 4553 | 4727 / 8899 | 0.40 / 0.61 | 0.64 / 1.07 | 49.34 | 43.17 |

A separate PCM sanity check looked for the first 20 ms window with RMS above
100. At 4 RPS its median first-body-byte time was 257–258 ms and median audible
time was 261 ms, so the first packet is audio rather than an HTTP/SSE artifact.

The result is a latency/throughput trade-off, not a universal improvement. At
4 RPS the policy cuts TTFP p50 by 70% and p95 by 83% with equivalent audio
throughput. Above saturation, stock decode-first scheduling reaches about 49
audio-seconds/s, while interval 6 reaches about 43. That is why this PR leaves
the policy disabled by default.

Related: #6365

### DIFF
--- a/docs/user_guide/examples/online_serving/text_to_speech.md
+++ b/docs/user_guide/examples/online_serving/text_to_speech.md
@@ -472,6 +472,17 @@ vllm serve openbmb/VoxCPM2 --omni --host 0.0.0.0 --port 8000
 ```
 Deploy config auto-loads from `vllm_omni/deploy/voxcpm2.yaml`. Pass `--deploy-config <path>` to override or `--stage-N-<field> <value>` for per-stage runtime tweaks.
 
+For latency-oriented serving, `unified_decode_graph_prefill_interval=N` admits
+waiting prefills for one scheduler iteration after every `N` decode-only
+iterations. The default `0` keeps decode-first admission. Positive values can
+reduce TTFP under concurrent load at the cost of peak throughput, so tune the
+interval for the target workload:
+
+```bash
+vllm serve openbmb/VoxCPM2 --omni \
+    --hf-overrides '{"voxcpm2_runtime_config":{"unified_decode_graph_prefill_interval":6}}'
+```
+
 ### Sending requests
 ```bash
 # Zero-shot synthesis
--- a/tests/core/sched/test_omni_ar_scheduler_unified_decode_deferral.py
+++ b/tests/core/sched/test_omni_ar_scheduler_unified_decode_deferral.py
@@ -7,6 +7,7 @@ from vllm.v1.request import RequestStatus
 
 import vllm_omni.core.sched.omni_ar_scheduler as scheduler_mod
 import vllm_omni.model_executor.models.voxcpm2.scheduler as voxcpm2_scheduler_mod
+from vllm_omni.model_executor.models.voxcpm2.runtime_config import _VoxCPM2RuntimeConfig
 from vllm_omni.model_executor.models.voxcpm2.scheduler import VoxCPM2OmniARAsyncScheduler
 
 pytestmark = [pytest.mark.core_model, pytest.mark.cpu]
@@ -28,6 +29,12 @@ class _MockQueue:
     def add_request(self, request) -> None:
         self._items.append(request)
 
+    def pop_request(self):
+        return self._items.pop(0)
+
+    def prepend_request(self, request) -> None:
+        self._items.insert(0, request)
+
     def prepend_requests(self, requests) -> None:
         self._items = list(requests) + self._items
 
@@ -47,6 +54,7 @@ class _MockRequest:
         self.num_prompt_tokens = num_prompt_tokens
         self.num_computed_tokens = num_computed_tokens
         self.num_output_placeholders = num_output_placeholders
+        self._all_token_ids = list(range(num_computed_tokens))
 
     def is_finished(self) -> bool:
         return RequestStatus.is_finished(self.status)
@@ -62,6 +70,7 @@ def _make_scheduler(
     enable_unified_decode_graph: bool | None = True,
     deterministic_cfm_noise: bool = False,
     runtime_config_on_model_config: bool = False,
+    prefill_interval: int = 6,
 ) -> VoxCPM2OmniARAsyncScheduler:
     sched = VoxCPM2OmniARAsyncScheduler.__new__(VoxCPM2OmniARAsyncScheduler)
     hf_config = SimpleNamespace()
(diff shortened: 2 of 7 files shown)


lolyhop:
Self-review:

- Verified that the default `unified_decode_graph_prefill_interval=0` preserves the existing decode-first behavior.
- Checked that a positive interval pauses decode for exactly one scheduler iteration, admits waiting prefills, and resumes the existing requests without dropping request or KV-cache state.
- Reviewed counter reset and boundary cases, including iterations without waiting prefills and transitions between prefill and decode.
- Added focused tests for runtime configuration, scheduler cadence, and request-state preservation.
- Ran the focused VoxCPM2 tests (`21 passed`) and the AR scheduler suite (`40 passed`). The build, pre-commit, and DCO checks are also green.
- Repeated a 60-second stock-vs-patched benchmark on one H100 at 4 RPS: interval 6 reduced audio TTFP p95 from 1.50 s to 260 ms at equivalent throughput (~38 SPS), with all 240 requests completing successfully.

@tzhouam @yinpeiqi @fake0fan @Sy0307 @Gaohan123, please take a look when you have a chance 🙏

lolyhop:
@linyueqian, take a look, please

PR #6378, opened 2026-08-19
[Bugfix] Build CUDA images on vLLM 0.27

tangbinh:
## Summary

Update the default CUDA image base from vLLM 0.26 to vLLM 0.27.

vLLM-Omni main now uses vLLM 0.27 APIs, including the multimodal hasher helper. Release and nightly jobs build and publish docker/Dockerfile.cuda, so retaining the 0.26 base produces an image that installs current vLLM-Omni over an incompatible runtime.

## Validation

- git diff --check: passed

## Self-review

I confirmed the release pipeline builds docker/Dockerfile.cuda and that no remaining Docker, Buildkite, requirements, or project pins reference vLLM 0.26.

### DIFF
--- a/docker/Dockerfile.cuda
+++ b/docker/Dockerfile.cuda
@@ -1,4 +1,4 @@
-ARG BASE_IMAGE=vllm/vllm-openai:v0.26.0
+ARG BASE_IMAGE=vllm/vllm-openai:v0.27.0
 FROM ${BASE_IMAGE}
 
 ARG COMMON_WORKDIR=/app


tangbinh:
Self-review: I traced the release pipeline to docker/Dockerfile.cuda, confirmed current main uses vLLM 0.27-only APIs, and searched Docker, Buildkite, requirements, and project files for remaining vLLM 0.26 pins. The changed-file hooks and diff check pass.

PR #6380, opened 2026-08-19
[Bugfix] Complete deferred engine and request cleanup

tangbinh:
## Summary

- retry DiffusionEngine.close after inline client tasks join, and keep shutdown retryable while engine teardown remains deferred
- release a preselected stage-0 replica binding when CFG companion construction fails before admission
- add regressions for deferred worker-thread teardown and multi-replica multimodal companion failures

These are both pre-admission/shutdown ownership leaks: the current inline client marks itself complete even when DiffusionEngine deliberately defers resource teardown, while a failed companion build leaves the parent request bound to a stage-0 replica even though nothing is enqueued.

## Validation

- inline client test file: 13 passed
- focused stage-binding regression: passed
- git diff --check: passed

## Self-review

I verified that successful/ordinary shutdown remains idempotent, deferred shutdown retries after the client executor joins and remains callable if still incomplete, and the binding release occurs only on companion construction failure before the parent is admitted.

### DIFF
--- a/tests/diffusion/test_inline_stage_diffusion_client.py
+++ b/tests/diffusion/test_inline_stage_diffusion_client.py
@@ -177,6 +177,37 @@ def test_inline_shutdown_runs_after_orchestrator_premark(client, mock_engine):
     mock_engine.close.assert_called_once()
 
 
+def test_inline_shutdown_retries_deferred_engine_cleanup(client, mock_engine):
+    mock_engine._shutdown_complete = False
+
+    def close_engine():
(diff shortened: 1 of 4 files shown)


tangbinh:
Self-review: I verified ordinary inline shutdown remains idempotent, deferred engine teardown is retried after client tasks join and remains retryable if still incomplete, and CFG failure releases only the pre-admission stage-0 binding before re-raising. The inline client file passes 13/13 and the focused multi-replica binding regression passes.

PR #6387, opened 2026-08-19
[Kernel] Add shared gated residual op for video DiTs

Sunbeam23333:
## Purpose

Video diffusion transformer blocks repeatedly execute the same gated residual pattern:

```python
residual + branch * gate
```

This PR introduces a shared `gated_residual(residual, branch, gate)` operator so model implementations can use one acceleration contract instead of carrying model-specific pointwise sequences.

The operator:

- uses a Triton CUDA fast path for fp16/bf16 activations;
- supports global, per-batch, and per-token gates, including strided modulation views produced by `chunk`/`unbind`;
- preserves eager's intermediate fp16/bf16 multiplication rounding before the residual add;
- registers a `torch.library` custom op with a fake implementation for `torch.compile`;
- falls back to the exact eager expression for unsupported devices, dtypes, layouts, hidden sizes, and general broadcast patterns.

The common interface is integrated at 14 hot-path call sites across:

- Wan2.2;
- HunyuanVideo 1.5;
- LTX-2 video/audio transformer blocks.

A standalone CUDA microbenchmark is included under `benchmarks/kernels/`.

This is intentionally a **Draft** because the implementation environment is CPU-only. CUDA exactness, compile coverage, and the reproducible benchmark are checked in, but NVIDIA GPU latency/VRAM results still need to be collected before marking ready.

## Test Plan

CPU contract and fallback coverage:

```bash
pytest -q tests/diffusion/layers/test_gated_residual.py -m "core_model and cpu"
```

CUDA numerical, strided-gate, and compile coverage:

```bash
pytest -q tests/diffusion/layers/test_gated_residual.py -m "core_model and cuda"
```

Reproducible kernel A/B (example Wan-shaped workload):

```bash
python benchmarks/kernels/gated_residual_benchmarks.py \
  --batch-size 1 \
  --tokens 32760 \
  --hidden-size 5120 \
  --gate-layout batch \
  --dtype bfloat16 \
  --warmup 25 \
  --iterations 100 \
  --repeats 20
```

Static and repository checks:

```bash
pre-commit run --files \
  benchmarks/kernels/gated_residual_benchmarks.py \
  tests/diffusion/layers/test_gated_residual.py \
  vllm_omni/diffusion/layers/gated_residual.py \
  vllm_omni/diffusion/models/wan2_2/wan2_2_transformer.py \
  vllm_omni/diffusion/models/hunyuan_video/hunyuan_video_15_transformer.py \
  vllm_omni/diffusion/models/ltx2/ltx2_transformer.py
```

**vLLM Version:** GPU runtime validation pending

**vLLM-Omni Commit:** `c3f80502e00b564d7b74d43a5e21fbc6b9da2aed`

## Test Result

- Focused CPU contract tests: **9 passed**.
- Full staged pre-commit suite: **passed** (Ruff check/format, typos, test marks, file hygiene).
- `git diff --check`: **passed**.
- CUDA numerical and `torch.compile` tests: **not run locally; NVIDIA GPU required**.
- Kernel latency, end-to-end video latency, and peak VRAM A/B: **pending before ready-for-review**.
- Branch freshness at push: live upstream `main=c3f80502`, **0 behind / 1 ahead**.

## AI assistance disclosure

OpenAI Codex assisted with implementation, testing, and review. The commit includes a `Co-authored-by` trailer.

### DIFF
--- a/vllm_omni/diffusion/layers/gated_residual.py
+++ b/vllm_omni/diffusion/layers/gated_residual.py
@@ -1,0 +1,224 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+
+"""Shared gated-residual operator for diffusion transformer blocks."""
+
+from __future__ import annotations
+
+import math
+
+import torch
+from torch.library import Library
+from vllm.platforms import current_platform
+from vllm.triton_utils import HAS_TRITON, tl, triton
+from vllm.utils.torch_utils import direct_register_custom_op
+
+_GLOBAL_GATE = 0
+_BATCH_GATE = 1
+_ROW_GATE = 2
+_EAGER_GATE = -1
+_MAX_FUSED_HIDDEN_SIZE = 16384
+
+
+if HAS_TRITON:
+
+    @triton.jit
+    def _gated_residual_kernel(
+        residual_ptr,
+        branch_ptr,
+        gate_ptr,
+        output_ptr,
+        hidden_size: tl.constexpr,
+        rows_per_batch: tl.constexpr,
+        gate_row_stride: tl.constexpr,
+        gate_mode: tl.constexpr,
+        product_dtype: tl.constexpr,
+        block_size: tl.constexpr,
+    ):
+        row = tl.program_id(0)
+        columns = tl.arange(0, block_size)
+        offsets = row * hidden_size + columns
+        mask = columns < hidden_size
+
+        if gate_mode == 0:
+            gate_row = 0
+        elif gate_mode == 1:
+            gate_row = row // rows_per_batch
+        else:
+            gate_row = row
+
+        gate_offsets = gate_row * gate_row_stride + columns
+        residual = tl.load(residual_ptr + offsets, mask=mask).to(tl.float32)
+        branch = tl.load(branch_ptr + offsets, mask=mask).to(tl.float32)
+        gate = tl.load(gate_ptr + gate_offsets, mask=mask).to(tl.float32)
+
+        # Match ``branch * gate`` followed by the residual add. The explicit
+        # cast preserves the intermediate rounding of eager fp16/bf16 math.
+        gated_branch = (branch * gate).to(product_dtype).to(tl.float32)
+        tl.store(output_ptr + offsets, residual + gated_branch, mask=mask)
+
+
+def _eager_gated_residual(
+    residual: torch.Tensor,
+    branch: torch.Tensor,
+    gate: torch.Tensor,
+    gate_mode: int,
+    rows_per_batch: int,
+    gate_row_stride: int,
+) -> torch.Tensor:
+    del gate_mode, rows_per_batch, gate_row_stride
+    return residual + branch * gate
+
+
+def _fused_cuda_supported(
+    residual: torch.Tensor,
+    branch: torch.Tensor,
+    gate: torch.Tensor,
+    gate_mode: int,
+) -> bool:
+    return (
+        HAS_TRITON
+        and current_platform.is_cuda()
+        and residual.is_cuda
+        and branch.is_cuda
+        and gate.is_cuda
+        and residual.dtype in (torch.float16, torch.bfloat16)
+        and branch.dtype == residual.dtype
+        and gate.dtype == residual.dtype
+        and residual.device == branch.device == gate.device
+        and residual.is_contiguous()
+        and branch.is_contiguous()
+        and gate.ndim > 0
+        and gate.stride(-1) == 1
+        and gate_mode != _EAGER_GATE
+        and gate.shape[-1] == residual.shape[-1]
+        and 0 < residual.shape[-1] <= _MAX_FUSED_HIDDEN_SIZE
+    )
+
+
+def _gated_residual_impl(
+    residual: torch.Tensor,
+    branch: torch.Tensor,
+    gate: torch.Tensor,
+    gate_mode: int,
+    rows_per_batch: int,
+    gate_row_stride: int,
+) -> torch.Tensor:
+    if not _fused_cuda_supported(residual, branch, gate, gate_mode):
+        return _eager_gated_residual(residual, branch, gate, gate_mode, rows_per_batch, gate_row_stride)
+
+    output = torch.empty_like(residual)
+    rows = residual.numel() // residual.shape[-1]
+    if rows == 0:
+        return output
+
+    product_dtype = tl.float16 if residual.dtype == torch.float16 else tl.bfloat16
+    block_size = triton.next_power_of_2(residual.shape[-1])
+    _gated_residual_kernel[(rows,)](
+        residual,
+        branch,
+        gate,
+        output,
+        hidden_size=residual.shape[-1],
+        rows_per_batch=rows_per_batch,
+        gate_row_stride=gate_row_stride,
+        gate_mode=gate_mode,
+        product_dtype=product_dtype,
+        block_size=block_size,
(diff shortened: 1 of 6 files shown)


vllm_omni/diffusion/layers/gated_residual.py
```diff
@@ -0,0 +1,224 @@
+# SPDX-License-Identifier: Apache-2.0
```
hsliuustc0106, line 1: @Bounty-hunter PTAL

Dong1017 (commented):
I picked up the GPU checks listed as pending in the PR at `7da8d1447ac0`. On L20X (vLLM 0.28, PyTorch 2.13/CUDA 13, Triton 3.7.1), the unchanged tests gave 9 CPU passes and 8 CUDA failures. The documented BF16 benchmark also stopped at its correctness check: 22.9% of outputs differed, with max absolute error 0.0625, so there are no valid latency/VRAM results from that run. The rounding and backward issues are described inline.

I loaded the unchanged operator directly because this older branch's package startup fails against vLLM 0.28. These are real CUDA/custom-op tests, but not a reproduction of the full older CI environment or model runs.

This is the existing base for the Wan/Hunyuan R1 work in [#7382](https://github.com/vllm-project/vllm-omni/issues/7382#issuecomment-5629791477). Could we also coordinate the intended shared `ops/modulation/gated_residual.py` entry with that work? The model qualification and separate LTX-2/R2 follow-up are still needed.

vllm_omni/diffusion/layers/gated_residual.py
```diff
@@ -0,0 +1,224 @@
...
+
+        gate_offsets = gate_row * gate_row_stride + columns
+        residual = tl.load(residual_ptr + offsets, mask=mask).to(tl.float32)
+        branch = tl.load(branch_ptr + offsets, mask=mask).to(tl.float32)
+        gate = tl.load(gate_ptr + gate_offsets, mask=mask).to(tl.float32)
+
+        # Match ``branch * gate`` followed by the residual add. The explicit
+        # cast preserves the intermediate rounding of eager fp16/bf16 math.
+        gated_branch = (branch * gate).to(product_dtype).to(tl.float32)
+        tl.store(output_ptr + offsets, residual + gated_branch, mask=mask)
```
Dong1017, line 58: On L20X with Triton 3.7.1, this cast still compiles to FP16/BF16 FMA, which skips the intermediate product rounding of `residual + branch * gate`. All eight CUDA tests fail their zero-tolerance checks, including strided gates and compile. In a small FP16/BF16 control, launching the same kernel with only `enable_fp_fusion=False` removed all mismatches. Could we disable that contraction, or otherwise enforce the rounding boundary, and rerun the existing tests and benchmark correctness check?

vllm_omni/diffusion/layers/gated_residual.py
```diff
@@ -0,0 +1,224 @@
...
+        broadcast_shape = torch.broadcast_shapes(tuple(residual.shape), tuple(gate.shape))
+    except RuntimeError as error:
+        raise ValueError(f"gate shape {gate.shape} is not broadcastable to residual shape {residual.shape}") from error
+    if broadcast_shape != tuple(residual.shape):
+        raise ValueError(f"gate shape {gate.shape} broadcasts beyond residual shape {residual.shape}")
+
+    gate_mode, rows_per_batch, gate_row_stride = _gate_layout(residual, gate)
+    if not _fused_cuda_supported(residual, branch, gate, gate_mode):
+        return _gated_residual_impl(residual, branch, gate, gate_mode, rows_per_batch, gate_row_stride)
+    return torch.ops.vllm_omni.gated_residual(
```
Dong1017, line 214: Could we keep the eager path when grad is enabled and any input requires gradients? Supported CUDA FP16/BF16 inputs currently reach the custom op without an autograd registration. In a direct helper test, `autograd.grad(output.float().sum(), (residual, branch, gate), allow_unused=True)` returned `(None, None, None)` with a warning, while the original expression produced all three gradients. This affects gradient-enabled helper calls rather than the normal inference path; a fallback check and a small backward regression test would preserve that behavior.

Dong1017:
@Sunbeam23333 @ztang2370 — thanks for preparing the Wan model-side regression coverage in #7491. To unblock the P2/R1 work in #7382, I suggest the following split:

- **Sunbeam / #6387:** address the existing [intermediate-rounding finding](https://github.com/vllm-project/vllm-omni/pull/6387#discussion_r3994828504) and [grad-enabled fallback finding](https://github.com/vllm-project/vllm-omni/pull/6387#discussion_r3994828507), with the corresponding regression checks and the benchmark correctness check.
- **ztang / follow-up:** build on the corrected operator for the canonical `diffusion/layers/ops/modulation/gated_residual.py` module and public exports, then Wan2.2 adoption/qualification followed by HunyuanVideo 1.5. Keep #7491 focused on its current model-side tests.

@Sunbeam23333, can you take the two operator fixes? If you prefer to include the canonical migration in #6387 itself, please say so before ztang starts that part. @ztang2370, would the separate migration follow-up fit your R1 work?

Please confirm the split and your next update point; this is a proposed coordination boundary, not an assumption that either of you has already accepted the additional migration step.

ztang2370:
Thanks @Dong1017 , the separate migration follow-up fits my R1 scope. I’ll keep #7491 focused on model-side tests, then build on the corrected #6387 operator for the canonical module/public exports, followed by Wan2.2 and HunyuanVideo 1.5 adoption/qualification.

@Sunbeam23333, please confirm whether you prefer to include the migration in #6387. Once we agree on the split, I’ll post a concrete migration plan as my next update.

ztang2370:
@Dong1017 @Sunbeam23333 — to move P2/R1 forward, I’d like to take on the operator fixes alongside the migration and consumer work we discussed.
I propose opening a new PR carrying forward the implementation from #6387, preserving Sunbeam’s authorship and credit. The initial PR would cover:
- The intermediate-rounding and grad-enabled fallback fixes, with regression tests, CUDA/compile validation, and the benchmark correctness check.
- The canonical `diffusion/layers/ops/modulation/gated_residual.py` module and public export.
- Documentation of supported inputs, rounding, and fallback behavior, including the differences from the SANA and MiniMax-H3 VAE variants.
I’ll keep #7491 focused on its existing model-side tests, then follow with Wan2.2 and HunyuanVideo 1.5 adoption and qualification, covering numerical correctness, actual fused/fallback execution, model outputs, and performance. This would put the full P2 scope with me, delivered through focused PRs.
Would this updated split work?

Dong1017:
@hsliuustc0106 @david6666666 — status on #6387: the two review findings ([intermediate rounding](https://github.com/vllm-project/vllm-omni/pull/6387#discussion_r3994828504), [grad-enabled fallback](https://github.com/vllm-project/vllm-omni/pull/6387#discussion_r3994828507)) remain unaddressed here, with no author activity since 08-20 ([stale notice](https://github.com/vllm-project/vllm-omni/pull/6387#issuecomment-5585488647)).

@ztang2370 has opened #7857 carrying this implementation forward with @Sunbeam23333's authorship preserved (first commit matches this PR's head exactly) and addresses both findings — regression tests, the canonical `diffusion/layers/ops/modulation/gated_residual.py` module, and `ops/README.md` docs included. Model call-site wiring moves to follow-up adoption PRs; #7491 keeps its model-test scope.

Could you confirm #7857 as the review target for the operator fixes + canonical migration? We'll review it there and verify the two fixes. And how would you like #6387 handled — close as superseded, or keep open for reference?

@Sunbeam23333 — flagging in case you'd prefer to handle the fixes here; otherwise we'll proceed on #7857.

PR #6388, opened 2026-08-19
[Perf][MiniCPM-o] Drop the full-vocab bincount from the Talker penalty

0xtoward:
## Purpose

Swap `torch.bincount` for `scatter_add_` in `_apply_repetition_penalty`. One-line change, frequency vector is elementwise identical.

`torch.bincount` output length is data-dependent — dispatches to AI_CPU on Ascend, forces a device sync on CUDA even with `minlength`.

vLLM upstream already avoids it in three places for the same reason:

1. `vllm/v1/attention/backends/utils.py` — the v1 metadata builder:
   > Avoid `torch.bincount` here - on CUDA it forces a sync to determine the output size (even with `minlength`, the kernel must confirm no value exceeds the bound). `scatter_add_` into a preallocated buffer is equivalent and stays async.

2. `vllm/model_executor/layers/utils.py` — `get_token_bin_counts_and_mask` uses `scatter_add_` for token histograms.

3. `vllm/v1/worker/gpu/sample/penalties.py` — v1 GPU sampler ships a hand-written Triton `_bincount_kernel` to avoid ATen bincount entirely.

## Test Results

910B, vocab 6562, 16-token window, median of 400 calls:

| config | before | after |
|---|---|---|
| f32, 1 row | 334–366 µs | 167–168 µs |
| f32, 4 rows | 344–348 µs | 171–177 µs |
| bf16, 1 row | 324–327 µs | 170–172 µs |
| bf16, 4 rows | 335–345 µs | 177 µs |

### DIFF
--- a/vllm_omni/model_executor/models/minicpmo_4_5/minicpmo_4_5_omni_tts.py
+++ b/vllm_omni/model_executor/models/minicpmo_4_5/minicpmo_4_5_omni_tts.py
@@ -76,7 +76,8 @@ def _apply_repetition_penalty(
     if penalty == 1.0 or history.numel() == 0:
         return logits
     recent = history.reshape(-1)[-window_size:].to(device=logits.device, dtype=torch.long)
-    frequencies = torch.bincount(recent, minlength=logits.shape[-1]).to(dtype=logits.dtype)
+    frequencies = torch.zeros(logits.shape[-1], device=logits.device, dtype=logits.dtype)
+    frequencies.scatter_add_(0, recent, torch.ones_like(recent, dtype=logits.dtype))
     alpha = torch.pow(torch.as_tensor(penalty, device=logits.device, dtype=logits.dtype), frequencies)
     return torch.where(logits < 0, logits * alpha, logits / alpha)
 


Junbo2002:
Reproduced on an Atlas A3 (910C) with stock vllm-ascend. Both arms are
`ecd9d99da` differing only by this PR; the repo's own harness (Seed-TTS zh,
`openai-chat-omni`, 32 prompts, concurrency 1); both arms run at once, one per
die, each round repeated with the dies swapped.

**End-to-end:** −1.58% and −1.77% RTF. 32/32 completed, 435 output tokens,
135.64 s of audio in every arm.

**In place** — the same `perf_counter_ns` wrapper in *both* arms, so it
cancels; two die-swapped rounds:

| method | calls / 32 req | round 1 | round 2 |
| --- | ---: | --- | --- |
| `_sample_audio_code` | 3557 | 1115.87 → **860.92** µs | 1203.23 → **857.39** µs |
| `sample` (untouched — control) | 3591 | 232.78 → 233.42 µs | 238.16 → 235.27 µs |

**Isolated, no vLLM** (real shapes: logits `[1, 6562]` f32, window 16, 200
alternating trials): `torch.bincount` 527.51 µs → `zeros + scatter_add_`
308.88 µs, **−218.6 µs (−41.4%)**, outputs bit-identical. The
device-inclusive medians drop by the same amount, so what goes away is time
the call spends blocking the host.

The three measurements agree: 33.6 ms per request in place ≈ 1.7% of a
~1940 ms request, against −1.58% / −1.77% end-to-end.

natureofnature:
Thanks for this optimization. Since `minicpm-challenge` is no longer being used, would you be willing to update this PR to target `main`?

We have adapted your fixed-size `scatter_add_` histogram approach to the current batched Talker repetition-penalty path in draft #7968. We'd be happy to contribute the relevant adaptation and tests to this PR, so your change can land through the original contribution. We would then remove the overlapping changes from #7968 and keep only our additional optimizations.

The current main implementation uses the batched helper, so this would require adapting the change rather than only changing the base branch. Would you be open to coordinating this?

0xtoward:
Thanks @natureofnature! Updated for main with your adaptation and co-author credit. CPU equivalence checks pass.

natureofnature (approved):
LGTM for the focused histogram change at `b3681f7a`.

The fixed-size `int64` counts preserve request-local row offsets, the existing 16-row workspace bound, and windowed frequency-penalty semantics. History handling, scalar checks, sampling, and EOS behavior remain unchanged.

Validated the actual PR export in a remote container with vLLM 0.29.0 / PyTorch 2.13.0+cu130: the complete `test_talker_batching.py` module passed (**43 tests**). Additional strict comparisons against the base helper passed **144 CPU and 432 CUDA cases** on H800, covering FP32/BF16, chunk boundaries, empty/repeated histories, per-row penalties, non-contiguous logits, and CPU/CUDA/mixed-device histories.

This validates numerical equivalence, not new E2E performance or NPU results. Thanks for keeping this port scoped; it cleanly separates the shared histogram optimization from the remaining changes in #7968.

PR #6390, opened 2026-08-19
[Diffusion] Make Flux2Pipeline text_encoder_out_layers configurable

khairulkabir1661:
## Summary

- Read `text_encoder_out_layers` from the text encoder config in `Flux2Pipeline.__init__` (same pattern as `Flux2KleinPipeline`), falling back to the original `(10, 20, 30)` when the config attribute is absent
- Update `encode_prompt` and `__call__` to use the instance attribute instead of hardcoded defaults
- Shrink the tiny model builder from 31 text encoder layers to 3 (`hidden_size=32`, `text_encoder_out_layers=[0,1,2]`, `joint_attention_dim=96`)

## Test plan

- [x] `pytest tests/model_tests/diffusion/test_common_offline.py -k "Flux2Pipeline" -v` — all 3 subtests pass (TEXT_TO_IMAGE, determinism, multi_output)
- [ ] Full-size model behavior unchanged (no `text_encoder_out_layers` in config → falls back to `(10, 20, 30)`)

### DIFF
--- a/vllm_omni/diffusion/models/flux2/pipeline_flux2.py
+++ b/vllm_omni/diffusion/models/flux2/pipeline_flux2.py
@@ -683,12 +683,8 @@ class Flux2Pipeline(
         num_images_per_prompt: int = 1,
         prompt_embeds: torch.Tensor | None = None,
         max_sequence_length: int = 512,
-        text_encoder_out_layers: tuple[int, ...] | None = None,
     ):
         device = device or self._execution_device
-
-        if text_encoder_out_layers is None:
-            text_encoder_out_layers = self.text_encoder_out_layers
 
         if prompt is None:
             prompt = ""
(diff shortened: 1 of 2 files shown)


vllm_omni/diffusion/models/flux2/pipeline_flux2.py
```diff
@@ -886,7 +895,7 @@ def forward(
         callback_on_step_end: Callable[[int, int, dict], None] | None = None,
         callback_on_step_end_tensor_inputs: list[str] = ["latents"],
         max_sequence_length: int = 512,
-        text_encoder_out_layers: tuple[int, ...] = (10, 20, 30),
+        text_encoder_out_layers: tuple[int, ...] | None = None,
```
alex-jw-brooks, line 898: Can you set this to `text_encoder_out_layers=self.text_encoder_out_layers` and not pass it as a param through the request body, similar to flux2klein ([from this PR](https://github.com/vllm-project/vllm-omni/pull/5589))?

Diffusers does allow passing it at inference time, but this should be a model level property since the model would be trained against specific extracted text encoder layers

alex-jw-brooks (changes requested):

vllm_omni/diffusion/models/flux2/pipeline_flux2.py
khairulkabir1661: Done

alex-jw-brooks (approved):
LGTM, thanks :)

PR #6392, opened 2026-08-20
[Model] Add ABot-World realtime AR-Diffusion support

tsinghua-code:
## Purpose

Add experimental offline and in-process realtime support for `acvlab/ABot-World-0-5B-LF`.

This is an additional world-model integration on the model-neutral realtime infrastructure from #5491, which builds on the existing AR-Diffusion engine and paged-KV session support. ABot-World reuses the typed tick/control/chunk protocol, session manager, tick consumer, worker lifecycle, request orchestration, model runner, and session-owned paged-KV state.

ABot does **not** reuse LingBot weights or LingBot-specific model/conditioning logic. This PR adds the ABot-specific transformer and pipeline, native checkpoint loader, UMT5/Wan2.2 VAE conversion, camera-action reducer, and tensor/KV geometry.

The official checkpoint uses a native single-directory layout without a Diffusers `model_index.json`. Config discovery now recognizes its `model_type="ti2v"` config when `ABotWorldCausalPipeline` is selected, so no synthetic index is required.

Changed areas:

- `vllm_omni/diffusion/models/abot_world/`: model, pipeline, loading, and actions;
- `vllm_omni/diffusion/{data.py,registry.py}`: native discovery and registration;
- `tests/diffusion/models/abot_world/`: CPU regression tests;
- recipe and supported-model documentation.

Current scope: internal Python realtime API, latent chunk output, one AR replica with `max_num_seqs=1`, and no Cache-DiT/additional diffusion parallel modes.

## Test Plan

```bash
pytest -q tests/diffusion/models/abot_world
```

Manual checks use the official checkpoint without `model_index.json`, then run three realtime ticks with TP=1 and TP=2. Every output must be finite with shape `[1, 48, 3, 32, 52]`.

**vLLM Version:** 0.27.1 (the validation environment warns that it is not version-aligned with this development checkout)

**vLLM-Omni Commit:** 0477b7cc

Environment: PyTorch 2.13.0+cu126, CUDA runtime 12.6, driver 570.133.20, NVIDIA A100 PCIe 40GB.

## Test Result

- CPU suite: **22 passed, 15 warnings in 3.16s**.
- Native config discovery without `model_index.json`: passed.
- TP=1, one A100 40GB: 3/3 finite chunks; 1.99s / 0.69s / 0.71s; ~28.34 GiB.
- TP=2, two A100 40GB GPUs: 3/3 finite chunks; 5.94s / 0.82s / 0.84s; ~21.89 GiB per worker.

GPU figures are smoke-test observations, not performance claims.


david6666666:
you can record a video to show this model realtime case

tsinghua-code:
Thanks for the suggestion! Here is a realtime ABot-World demo running with the implementation in this PR. The first tick took 2.20s, and subsequent ticks took approximately 0.72–0.84s on a single A100 40GB.

https://github.com/user-attachments/assets/772e6724-816a-4a90-9c12-89aa4a4314dd

tsinghua-code:
Self-review completed:

- Checked ABot-World model registration and native checkpoint loading.
- Reviewed the realtime session lifecycle and per-session paged-KV state.
- Verified prompt updates and camera-action conditioning.
- Confirmed that each realtime tick returns a finite `[1, 48, 3, 32, 52]` latent chunk with matching metadata.
- Ran the CPU regression suite: 22 passed.
- Completed TP=1 and TP=2 realtime smoke tests with the official checkpoint.
- Reviewed the documented limitations and example commands.

wtz2333:
Found a correctness bug while benchmarking the realtime path on this branch (head `1ce74ed3`):

  **Realtime AR-Diffusion sessions produce degenerate (noise) latents for every chunk after chunk 0.**

  The latents are finite with normal-looking statistics (mean ≈ 0, std ≈ 0.85), but after VAE decode only chunk 0 yields coherent content — chunks 1+ decode to gray/banded noise. Chunk 0 itself is correct: 9 coherent frames, frame 0 exactly reconstructs the condition image.

  **Setup:** `acvlab/ABot-World-0-5B-LF`, 1xH200 140GB, TP=1, realtime session via `ARDiffusionSessionManager` + `ARDiffusionOmniTickConsumer` + `ABotCameraControlReducer` at 512x832, `num_frames=9`, 4 DMD steps, `flow_shift=5.0`, `output_type="latent"`, seed 42.

  **Isolation:**

  | Experiment | Result |
  |---|---|
  | Chunk 0 decoded alone (`latents * std + mean` → Wan2.2 VAE decode, same convention as the offline path) | Coherent 9-frame video |
  | Chunk 1 decoded alone | Still noise → not a chunk-concatenation/decode artifact |
  | Realtime with **no camera actions** at all | Chunk 1+ still degenerate → not action-conditioning related |
  | Offline single request, `num_frames=21` / `117` (local cache, no paged KV), same weights/image/prompt/seed | Fully coherent, frame 0 matches the input image |

  So weights, DiT, VAE, and the decode convention are all fine — the regression is specific to the **paged-KV path across chunk boundaries**, most likely the KV commit/retrieval after chunk 0, or cross-attention text-cache reuse between chunks.

  The current GPU validation only asserts finite chunks ("3/3 finite chunks"), which this passes — finite latents with normal statistics can still decode to noise. Suggest adding a regression check that decodes chunk 1+ and asserts content continuity (e.g. PSNR/brightness vs chunk 0, or parity
  with the offline first two blocks).
abot_noact_check

vllm_omni/diffusion/models/abot_world/pipeline.py
```diff
@@ -0,0 +1,1278 @@
...
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Any, ClassVar, cast
+
+import numpy as np
+import PIL.Image
+import PIL.ImageOps
+import torch
+import torch.nn.functional as F
+from diffusers.utils.torch_utils import randn_tensor
+from huggingface_hub import snapshot_download
```
fhfuih, line 20: Drop direct `snapshot_download` — `vllm_omni/diffusion/models/abot_world/pipeline.py:20`

`from huggingface_hub import snapshot_download` fails newly added check `tools/pre_commit/check_forbidden_imports.py`, and this path is not on the allowlist. `_resolve_local_model_path` (`:161`) only needs an already-local tree; LingBot resolves via local path / shared hub helpers instead. Prefer a local-path requirement or `vllm.transformers_utils.repo_utils` / existing diffusion hub helpers

recipes/ACVLab/ABot-World.md
fhfuih, line 1: Index the recipe — `recipes/README.md`

`recipes/ACVLab/ABot-World.md` is added but never listed in Available Recipes. Model-addition policy expects both the recipe file and the README row (same pattern as `Robbyant/LingBot-World-2.0.md`).

vllm_omni/diffusion/models/abot_world/actions.py
```diff
@@ -0,0 +1,201 @@
...
+)
+from vllm_omni.experimental.ar_diffusion.tick_protocol import (
+    ARDiffusionControlInput,
+)
+
+ABOT_CAMERA_ACTION_SCHEMA = "abot.camera_actions.v1"
+_ACTION_ORDER = ("w", "a", "s", "d", "i", "j", "k", "l")
+_VALID_ACTIONS = frozenset(_ACTION_ORDER)
+_REFERENCE_HEIGHT = 480
+_REFERENCE_WIDTH = 832
```
fhfuih, line 23: These are unused?

fhfuih: And they conflict with the default resolution 512x832 in pipeline.py

recipes/ACVLab/ABot-World.md
```diff
@@ -0,0 +1,80 @@
+# ABot-World
+
+> Experimental: offline and realtime interactive world generation
+
+## Summary
+
+- Vendor: ACVLab (Amap CV Lab)
+- Model: `acvlab/ABot-World-0-5B-LF`
+- Task: image-conditioned interactive world generation
+- Modes: offline trajectory replay and in-process realtime AR-Diffusion ticks
```
fhfuih, line 10: Offline trajectory is actually not possible at the time I review this PR. The camera input is handled in the preprocess function and `if`-skipped unless using the experimental sesssion/tick tooling (hence never a trajectory list of actions submitted via `Omni.generate()` at generation start; has to be submitted one by one using experimental `ARDiffusionControlInput` and stuff)

fhfuih (commented):
Left some comments for your consideration

wtz2333:
> Found a correctness bug while benchmarking the realtime path on this branch (head `1ce74ed3`):
> 
> **Realtime AR-Diffusion sessions produce degenerate (noise) latents for every chunk after chunk 0.**
> 
> The latents are finite with normal-looking statistics (mean ≈ 0, std ≈ 0.85), but after VAE decode only chunk 0 yields coherent content — chunks 1+ decode to gray/banded noise. Chunk 0 itself is correct: 9 coherent frames, frame 0 exactly reconstructs the condition image.
> 
> **Setup:** `acvlab/ABot-World-0-5B-LF`, 1xH200 140GB, TP=1, realtime session via `ARDiffusionSessionManager` + `ARDiffusionOmniTickConsumer` + `ABotCameraControlReducer` at 512x832, `num_frames=9`, 4 DMD steps, `flow_shift=5.0`, `output_type="latent"`, seed 42.
> 
> **Isolation:**
> 
> Experiment	Result
> Chunk 0 decoded alone (`latents * std + mean` → Wan2.2 VAE decode, same convention as the offline path)	Coherent 9-frame video
> Chunk 1 decoded alone	Still noise → not a chunk-concatenation/decode artifact
> Realtime with **no camera actions** at all	Chunk 1+ still degenerate → not action-conditioning related
> Offline single request, `num_frames=21` / `117` (local cache, no paged KV), same weights/image/prompt/seed	Fully coherent, frame 0 matches the input image
> So weights, DiT, VAE, and the decode convention are all fine — the regression is specific to the **paged-KV path across chunk boundaries**, most likely the KV commit/retrieval after chunk 0, or cross-attention text-cache reuse between chunks.
> 
> The current GPU validation only asserts finite chunks ("3/3 finite chunks"), which this passes — finite latents with normal statistics can still decode to noise. Suggest adding a regression check that decodes chunk 1+ and asserts content continuity (e.g. PSNR/brightness vs chunk 0, or parity with the offline first two blocks). abot_noact_check

This bug is caused by enabling CUDA Graphs.

vllm_omni/diffusion/models/abot_world/pipeline.py
tsinghua-code: Fixed in `b428ad2`. I removed the direct `huggingface_hub.snapshot_download` import. The ABot pipeline now requires an already-downloaded local checkpoint directory and raises a clear `FileNotFoundError` otherwise.

recipes/ACVLab/ABot-World.md
tsinghua-code: Fixed in `b428ad2`. I added `ACVLab/ABot-World.md` to the Available Recipes table in `recipes/README.md`.

vllm_omni/diffusion/models/abot_world/actions.py
tsinghua-code: Fixed in `b428ad2`. These constants were unused and conflicted with the pipeline's validated `512x832` default, so I removed them. Resolution remains owned by the pipeline rather than the camera-action reducer.

recipes/ACVLab/ABot-World.md
tsinghua-code: Fixed in `b428ad2`. I corrected the recipe to state that offline `Omni.generate()` currently supports image-conditioned generation but not camera-action trajectories. Camera actions are currently supported only through the experimental in-process Session/Tick API via incremental `ARDiffusionControlInput`; offline trajectory support is deferred to a follow-up.

tsinghua-code:
### Update: ABot-specific stateful streaming VAE decode

Pushed in `f8f0c8f2`.

The realtime path previously returned latent blocks only. Decoding each block independently
resets the Wan2.2 causal-convolution feature cache and treats every chunk as the start of a
new video, producing visible diagonal grid/checkerboard artifacts at chunk boundaries.

This update:

- adds session-owned `vae_decode_cache` to `_ABotARSessionState`;
- adds an ABot-scoped `_decode_realtime_chunk()` using the existing vLLM Wan2.2 VAE
  components;
- preserves causal VAE feature state across contiguous realtime ticks and releases it when
  the session is reset or closed;
- returns decoded `frames` alongside `latents`;
- adds regression coverage for 9 opening frames, 12 continued frames, and stateless session
  restart.

Both comparisons below decode the same saved realtime chunk-1 latent with the same vLLM
Wan2.2 VAE. The only difference is whether chunk 1 inherits chunk 0's causal VAE feature
cache.

**Local decoded frame 2 — left: no memory; right: chunk 0 cache → chunk 1**

chunk1_compare_frame_02

**Local decoded frame 4 — left: no memory; right: chunk 0 cache → chunk 1**

chunk1_compare_frame_04

The stateless path shows strong diagonal grid artifacts across the grass, road, and
mountains. These artifacts disappear when the causal VAE state is preserved across chunks.

Validation:

- CPU regression tests: `5 passed`
- no-memory chunk 1 output: `[1, 3, 9, 512, 832]`
- stateful chunk 1 output: `[1, 3, 12, 512, 832]`
- retained session VAE cache: approximately 2.80 GiB
- isolated comparison peak allocated GPU memory: approximately 12.1 GiB

This decoder-boundary issue is separate from the CUDA Graph-related latent-generation
correctness issue identified in the latest discussion. Stateful VAE decoding fixes cross-
chunk decoding continuity; it does not claim to fix degenerate latents produced upstream.

Remaining work includes accounting for the retained VAE cache in runtime memory budgeting
and validating the complete engine-to-RGB delivery path.

vllm_omni/diffusion/data.py
```diff
@@ -1439,6 +1416,11 @@ def enrich_config(self) -> None:
                         self.model_class_name = "Pi0Pipeline"
                     self.set_tf_model_config(TransformerConfig())
                     self.update_multimodal_support()
+                elif model_type == "ti2v" and self.model_class_name == "ABotWorldCausalPipeline":
+                    # The official ABot checkpoint uses its native single-directory
+                    # layout and intentionally has no Diffusers pipeline index.
+                    self.set_tf_model_config(TransformerConfig())
+                    self.update_multimodal_support()
```
fhfuih, line 1423: I was about to say that adding model-specific thing here is weird. But since there are already many other model specific operations above these lines (e.g., Pi0, DreamZero, WanS2V), now I think it is OK to stuff it here, and we have another refactor PR.

After all this is just conflict between transformer- and diffusers-style metadata formats

I am still posting a comment here in case other people encounter it and forget to check out the unchanged code above this hunk

docs/models/supported_models.md
fhfuih, line 1: You can also add this model in `docs/user_guide/diffusion_features.md`

recipes/ACVLab/ABot-World.md
fhfuih, line 1: According to current experiment results, you can highlight that `enforce_eager` is required or otherwise the accuracy/quality drops, and the root cause is yet to be discovered

recipes/ACVLab/ABot-World.md
```diff
@@ -0,0 +1,86 @@
...
+- Vendor: ACVLab (Amap CV Lab)
+- Model: `acvlab/ABot-World-0-5B-LF`
+- Task: image-conditioned interactive world generation
+- Modes: offline image-conditioned generation and experimental in-process realtime AR-Diffusion ticks
+- Hardware validated: NVIDIA A100 40GB
+- Maintainer: Community
+
+The checkpoint is Apache-2.0 licensed. The vLLM-Omni integration code is also Apache-2.0.
+
+## Realtime architecture
```
fhfuih: Recipe is only about recommended usage and setup configurations. So you can remove the architecture introduction. Keep the recipe focused on usage

fhfuih (commented):
Thanks for the update. Added some comments on the doc & chores.


Briefly checked the transformer hot path: can confirm that there is no `.item()` / `.cpu()` / `.tolist()` / dict `.items()` to introduce unnecessary overhead

recipes/README.md
```diff
@@ -39,6 +39,8 @@ recipes/
 
 | Recipe | Task | Hardware |
 |--------|------|----------|
+| [`ACVLab/ABot-World.md`](./ACVLab/ABot-World.md) | Offline image-conditioned generation and experimental realtime interactive generation | 1x A100 40GB |
+| [`audiox/AudioX.md`](./audiox/AudioX.md) | Offline + online unified text/video→audio diffusion | 1x L4 24GB |
```
fhfuih, line 43: Why is this line in the diff?

recipes/ACVLab/ABot-World.md
```diff
@@ -0,0 +1,76 @@
...
+
+Raw frame counts must be `9 + 12k` (9, 21, 33, ..., 117), up to 117 frames.
+
+Camera-action trajectories are not currently supported through
+`Omni.generate()`. Camera actions are supported only by the experimental
+in-process realtime Session/Tick API, where controls are submitted
+incrementally through `ARDiffusionControlInput`. Offline trajectory input
+is planned as a follow-up.
+
+## Realtime in-process generation
```
hsliuustc0106, line 44: [P3] Evidence ask for the realtime contract (not a perf claim): the smoke run shows tick 1 at 1.99s vs 0.69s steady — the warmup/capture stall — but only 3 ticks. Could you run one realtime session for a realistic length (>=30 consecutive ticks, same `max_num_seqs=1` setup) and report per-tick times, flagging any tick that triggers a capture or bucket-growth event and its stall? Three ticks validate the first steps; a cadence that collapses mid-session is a functional failure of the realtime contract, and this is the one number the recipe's realtime section currently can't promise.

hsliuustc0106: One quality signal worth adding alongside the cadence run, and it's cheap because both paths already exist in this PR: **realtime-vs-offline parity** — submit the same control sequence through `Omni.generate()` (offline) and the Session/Tick API, then SSIM/PSNR per chunk using the in-tree ffmpeg helpers (`tests/e2e/accuracy/helpers.py`, `run_ffmpeg_similarity`). Paged-KV sessions and chunked streaming change the numerical path, so this is the one number that validates the serving integration's quality, and it doubles as the first point of a **drift-over-horizon** curve (parity vs tick index) — the metric that tells users how long a session stays usable. Action-compliance (commanded camera trajectory vs rendered motion) would be a nice follow-up. FVD/VBench sweeps belong on the model card, not here.

hsliuustc0106: Caveat to my own suggestion: the recipe notes camera-action trajectories are currently tick-API-only, so the parity run should either use the control-free horizon (offline `Omni.generate()` at `num_frames=117` vs 13 ticks from the same image condition) or land together with the planned offline-trajectory follow-up.

vllm_omni/diffusion/models/abot_world/pipeline_abot_world.py
```diff
@@ -0,0 +1,1318 @@
...
+        cache: ABotTransformerCache | None,
+        ar_cross_attention: list[ABotAttentionCache] | None,
+        progress_bar: TqdmProgressBar[Any],
+    ) -> torch.Tensor:
+        out_channels = int(self.transformer.config.out_channels)
+        current_latents = noise_latent
+        replace_first = start_frame == 0 and first_frame_latent is not None
+
+        for step_idx, (ts_val, sigma) in enumerate(schedule):
+            if not self.od_config.enforce_eager:
```
RuixiangMa, line 1065: if not self.od_config.enforce_eager: torch.compiler.cudagraph_mark_step_begin() — recipe says enforce_eager=True is required, but code doesn't enforce it. Consider adding a warning in _validate_parallel_config.

RuixiangMa (commented):
uses just pipeline.py/ transformer.py ， missing the model name prefix.

tsinghua-code:
@RuixiangMa Thanks for pointing this out. I renamed the modules to `pipeline_abot_world.py` and `abot_world_transformer.py`, and updated all ABot-World imports accordingly in 5a978b04.

vllm_omni/diffusion/models/abot_world/pipeline_abot_world.py
tsinghua-code: Thanks. Addressed in 9bc6c202 and retained in the current head: `_validate_parallel_config()` now emits a `UserWarning` when `enforce_eager=False`, before the `parallel_config is None` early return. The warning explains the `cudagraph_mark_step_begin()` path, possible NaNs or black frames, and recommends `--enforce-eager`. Tests cover both eager modes and preserve the existing quantization error behavior.

recipes/README.md
fhfuih: (I mean the audiox line)

vllm_omni/diffusion/models/abot_world/pipeline_abot_world.py
```diff
@@ -0,0 +1,1327 @@
...
+            num_kv_heads=num_local_heads,
+            head_size=int(cfg.attention_head_dim),
+            tokens_per_frame=tokens_per_frame,
+            frames_per_block=self._num_frame_per_block,
+            window_frames=_WINDOW_FRAMES,
+            sink_frames=0,
+            kv_branches=(ARDiffusionKVBranchSpec(self._AR_BRANCH, 0),),
+            session_capacity=2,
+            cross_attention=(ARDiffusionCrossAttentionKVSpec(self._AR_TEXT_CACHE, _MAX_SEQUENCE_LENGTH),),
+            model_owned_state_bytes_per_session=condition_bytes,
```
david6666666, line 724: This reservation only covers a latent-sized condition, but each realtime session also retains the decoder's convolution feature maps in `vae_decode_cache`. Those tensors survive across ticks and aren't part of the runner-owned KV pool, so admission can overestimate how many decoded sessions fit. Please include the persistent VAE cache in the per-session budget and verify residency after decoding with two live sessions.

vllm_omni/diffusion/registry.py
```diff
@@ -146,6 +146,11 @@
         "pipeline",
         "LingBotWorldCausalDMDPipeline",
     ),
+    "ABotWorldCausalPipeline": (
+        "abot_world",
+        "pipeline",
```
lishunyang12: Change the module name to pipeline_abot_world so registry loading survives the rename.

hsliuustc0106 (commented):
The 9bc6c202/5a978b04 fixes (enforce_eager UserWarning in `_validate_parallel_config`, module renames to `pipeline_abot_world.py`/`abot_world_transformer.py`) look right. Still open from my 08-30 review: the realtime cadence run (>=30 consecutive ticks with per-tick timings) and the realtime-vs-offline parity signal — neither appears in the recipe at the current head. Also pre-commit is failing on this head and the branch conflicts with main, so both a lint fix and a rebase are needed.

vllm_omni/diffusion/models/abot_world/pipeline_abot_world.py
tsinghua-code: Fixed in 1eb56697e87b985014c605b0e9fbad91f20eb496. Admission now includes persistent decoder convolution state separately from runner-owned paged KV, as well as retained image/prompt/action conditions. The full-Wan estimate accounts for **FP32 feature caches even with BF16 model weights**.

On one L20X 140GB at 512x832, actual full-Wan decoder state was **3,006,062,592 bytes/session**, within the **3,040,460,800-byte/session** decoder reservation; total model-owned reservation including conditions is **3,131,236,352 bytes/session**, in addition to KV. The optional tiny decoder measured/reserved **17,891,328 bytes/session** for its decoder state.

The GPU check uses real RGB decoding (`output_type="pt"`), not latent-only output: A0/B0/A1/B1/A2/B2 with different camera controls, compared against separate solo controls. All six chunks were bit-exact for each backend. Both caches remained bounded; closing A preserved B and closing both left no decoder state. These are **two resident/interleaved sessions**, not concurrent execution (`max_num_seqs=1`). Full-Wan peak PyTorch allocated memory was **46.078 GiB** with the configured fixed KV pool; this is not a minimum-card-memory claim. Reset/error cleanup is also covered by CPU regression tests.

Details, reproduction commands, and data: [validation recipe](https://github.com/tsinghua-code/vllm-omni/blob/1eb56697e87b985014c605b0e9fbad91f20eb496/recipes/ACVLab/ABot-World.md#measured-validation-2026-09-19).

vllm_omni/diffusion/registry.py
tsinghua-code: Fixed in 1eb56697e87b985014c605b0e9fbad91f20eb496: the ABot registry entry now points to `pipeline_abot_world`. Added a regression check exercising registry model-class loading and pre/post-processing resolution so the module rename is covered, rather than only importing the class directly. The focused ABot/shared-decoder/runner suite passed 79 tests; the submission-only AR/ABot CPU suite passed 197 tests with 8 skipped.

recipes/ACVLab/ABot-World.md
tsinghua-code: Added the long-session and quality evidence in 1eb56697e87b985014c605b0e9fbad91f20eb496. The [recipe and reproduction commands](https://github.com/tsinghua-code/vllm-omni/blob/1eb56697e87b985014c605b0e9fbad91f20eb496/recipes/ACVLab/ABot-World.md#measured-validation-2026-09-19) link the per-tick CSV, per-chunk quality CSV, summary, and plot.

On one scheduler-reserved **L20X 140GB** (not A100/H200), BF16 eager, 512x832, seed 42, and `max_num_seqs=1`, each backend/API case completed one warmup plus three measured 35-tick/417-frame rollouts. Full-Wan typed Session/Tick results: median-across-runs steady median **590.3 ms/tick**, steady p95 **605.5 ms**, and **20.712 s** summed tick intervals. These are CPU-ready RGB client wall times, excluding validation IO, MP4 encoding, and network transport; no profiler was enabled.

CUDA Graph capture is disabled (`enforce_eager=True`, `warmup_cudagraph=False`); actual KV pool storage identities/sizes remained fixed at snapshots. We do not label ordinary clean-KV forwards as captures or fixed-pool page allocation as bucket growth. The CSV retains all intervals, including an **855.6 ms** full-Wan stepwise interval at zero-based tick **22** in one measured run; its cause is not attributed to capture/growth without evidence.

For the control-free parity check, 117 frames correspond to **10 chunks here (9 + 9 x 12)**. All 40 same-backend comparisons (two backends; typed vs offline and typed vs stepwise; ten chunks each) were bit-exact, with ffmpeg SSIM=1 and PSNR=infinity. The offline reference uses the same per-frame decoder to isolate chunking/paged-KV differences. This is a single-scene 117-frame consistency check, **not** 35-tick offline quality equivalence, original-upstream whole-clip-kernel parity, or action-compliance validation. Tiny-vs-full quality is reported separately (mean per-chunk float PSNR 30.13 dB; RGB8 SSIM 0.7988), so full Wan remains the default.

tsinghua-code:
Addressed the open cadence/parity evidence from this review in commit 1eb56697e87b985014c605b0e9fbad91f20eb496.

The recipe now includes a reproducible one-warmup plus three-measured-run protocol for both typed Session/Tick and one-generate stepwise paths, each at 35 consecutive ticks / 417 decoded frames with per-tick CSV output. It also includes the control-free 117-frame realtime-vs-offline parity check (10 chunks under the 9+12k contract): 40 same-backend comparisons were exact, SSIM=1 and PSNR=infinity. The chart and raw artifacts are here: https://github.com/tsinghua-code/vllm-omni/tree/1eb56697e87b985014c605b0e9fbad91f20eb496/recipes/ACVLab/ABot-World-evidence

Measured on one scheduler-reserved L20X 140GB, BF16 eager, TP=1, `max_num_seqs=1`: full-Wan typed steady median **590.3 ms/tick**, p95 **605.5 ms/tick**, 35-tick interval sum **20.712 s**. CUDA Graph capture was disabled by `enforce_eager=True`; fixed KV-pool identities/sizes stayed unchanged. One stepwise run retained an **855.6 ms** interval at tick 22 without attributing it to capture or bucket growth absent event evidence.

Pre-commit/diff checks passed for the submitted 23-file change set, and the focused CPU suite passed 79 tests; the submission-only AR/ABot CPU suite passed 197 tests with 8 skipped. The PR still reports a merge conflict with current `main`; that is a separate integration step and is not being misreported as resolved by the evidence commit.

vllm_omni/diffusion/models/abot_world/pipeline_abot_world.py
tsinghua-code: Adding the requested chart/evidence pointer for the VAE-cache review. The submitted validation plot is here: https://github.com/tsinghua-code/vllm-omni/blob/1eb56697e87b985014c605b0e9fbad91f20eb496/recipes/ACVLab/ABot-World-evidence/abot-review-validation.png

The chart contains four panels: (1) representative per-tick wall time across the 35-tick run, (2) the complete serial rollout timeline, (3) actual persistent decoder-state versus reservation by backend/session, and (4) realtime/offline SSIM by chunk. The supporting CSV/JSON are beside the PNG in the recipe.

Key measured values: full Wan decoder cache **3,006,062,592 B/session** versus **3,040,460,800 B/session** reserved; tiny decoder **17,891,328 B/session**. The dual-session test used real RGB decoding and interleaved A0/B0/A1/B1/A2/B2 chunks, then compared each session with an isolated solo run: all six chunks were exact, cache residency stayed bounded, closing A preserved B, and closing both released decoder state. Full-Wan peak allocated memory was **46.078 GiB** with the configured fixed KV pool.

Latency evidence in the same chart/data package is one warmup plus three measured 35-tick rollouts on one L20X 140GB, BF16 eager, TP=1, `max_num_seqs=1`; full-Wan typed steady median **590.3 ms/tick**, p95 **605.5 ms/tick**, summed intervals **20.712 s**. This is CPU-ready RGB wall time, not profiler time.

tsinghua-code: Adding the requested chart/evidence pointer for the VAE-cache review. The submitted validation plot is here: https://github.com/tsinghua-code/vllm-omni/blob/1eb56697e87b985014c605b0e9fbad91f20eb496/recipes/ACVLab/ABot-World-evidence/abot-review-validation.png

The chart contains four panels: (1) representative per-tick wall time across the 35-tick run, (2) the complete serial rollout timeline, (3) actual persistent decoder-state versus reservation by backend/session, and (4) realtime/offline SSIM by chunk. The supporting CSV/JSON are beside the PNG in the recipe.

Key measured values: full Wan decoder cache **3,006,062,592 B/session** versus **3,040,460,800 B/session** reserved; tiny decoder **17,891,328 B/session**. The dual-session test used real RGB decoding and interleaved A0/B0/A1/B1/A2/B2 chunks, then compared each session with an isolated solo run: all six chunks were exact, cache residency stayed bounded, closing A preserved B, and closing both released decoder state. Full-Wan peak allocated memory was **46.078 GiB** with the configured fixed KV pool.

Latency evidence in the same chart/data package is one warmup plus three measured 35-tick rollouts on one L20X 140GB, BF16 eager, TP=1, `max_num_seqs=1`; full-Wan typed steady median **590.3 ms/tick**, p95 **605.5 ms/tick**, summed intervals **20.712 s**. This is CPU-ready RGB wall time, not profiler time.

tsinghua-code:
Addressed the open cadence/parity evidence from this review in commit 1eb56697e87b985014c605b0e9fbad91f20eb496.

The recipe now includes a reproducible one-warmup plus three-measured-run protocol for both typed Session/Tick and one-generate stepwise paths, each at 35 consecutive ticks / 417 decoded frames with per-tick CSV output. It also includes the control-free 117-frame realtime-vs-offline parity check (10 chunks under the 9+12k contract): 40 same-backend comparisons were exact, SSIM=1 and PSNR=infinity. The chart and raw artifacts are here: https://github.com/tsinghua-code/vllm-omni/tree/1eb56697e87b985014c605b0e9fbad91f20eb496/recipes/ACVLab/ABot-World-evidence

Measured on one scheduler-reserved L20X 140GB, BF16 eager, TP=1, `max_num_seqs=1`: full-Wan typed steady median **590.3 ms/tick**, p95 **605.5 ms/tick**, 35-tick interval sum **20.712 s**. CUDA Graph capture was disabled by `enforce_eager=True`; fixed KV-pool identities/sizes stayed unchanged. One stepwise run retained an **855.6 ms** interval at tick 22 without attributing it to capture or bucket growth absent event evidence.

Pre-commit/diff checks passed for the submitted 23-file change set, and the focused CPU suite passed 79 tests; the submission-only AR/ABot CPU suite passed 197 tests with 8 skipped. The PR still reports a merge conflict with current `main`; that is a separate integration step and is not being misreported as resolved by the evidence commit.

recipes/ACVLab/ABot-World.md
tsinghua-code: Addressed in `ffee46ebf35928b35a4a5dd8379ac012f6a8268d` (the runtime/evidence implementation is in `1eb56697e87b985014c605b0e9fbad91f20eb496`; `ffee46eb` corrects the evidence hardware metadata to **1 × NVIDIA H200**).

Test configuration: **1 × NVIDIA H200**, `max_num_seqs=1`, TP=1, BF16 eager, 512×832, seed 42, four DMD denoising steps plus one clean-context KV forward per tick. Each case used one warmup followed by three measured **35-tick / 417-frame** sessions. Timings are CPU-ready RGB client wall time and exclude validation I/O, MP4 encoding, and network transport; no profiler was enabled.

| Decoder / API | First chunk (ms) | Steady median (ms/tick) | Steady p95 (ms/tick) | Sum of 35 tick intervals (s) |
| Full Wan / typed Session-Tick | 578.6 | 590.3 | 605.5 | 20.712 |
| Full Wan / one-generate stepwise | 595.9 | 587.8 | 611.3 | 20.765 |
| TAEW2.2 / typed Session-Tick | 1407.2 | 304.7 | 330.4 | 11.618 |
| TAEW2.2 / one-generate stepwise | 1001.8 | 299.5 | 323.1 | 11.212 |

Full-Wan typed per-tick intervals are below. Each run is one uninterrupted 35-tick session; the last column is the per-tick median across the three measured sessions.

| Tick | Frames | Run 1 (ms) | Run 2 (ms) | Run 3 (ms) | Median (ms) |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 0 | 9 | 579.4 | 567.0 | 578.6 | 578.6 |
| 1 | 12 | 591.0 | 582.2 | 579.5 | 582.2 |
| 2 | 12 | 605.9 | 583.0 | 593.6 | 593.6 |
| 3 | 12 | 603.3 | 587.7 | 583.6 | 587.7 |
| 4 | 12 | 603.1 | 583.4 | 591.8 | 591.8 |
| 5 | 12 | 597.3 | 597.3 | 586.3 | 597.3 |
| 6 | 12 | 612.9 | 587.8 | 593.4 | 593.4 |
| 7 | 12 | 603.0 | 582.9 | 597.9 | 597.9 |
| 8 | 12 | 596.8 | 584.8 | 603.4 | 596.8 |
| 9 | 12 | 591.4 | 584.9 | 603.1 | 591.4 |
| 10 | 12 | 591.0 | 584.5 | 585.5 | 585.5 |
| 11 | 12 | 591.6 | 586.0 | 590.2 | 590.2 |
| 12 | 12 | 588.5 | 592.6 | 590.0 | 590.0 |
| 13 | 12 | 597.3 | 592.0 | 586.7 | 592.0 |
| 14 | 12 | 591.2 | 590.4 | 597.5 | 591.2 |
| 15 | 12 | 593.1 | 592.7 | 592.8 | 592.8 |
| 16 | 12 | 604.7 | 586.4 | 597.9 | 597.9 |
| 17 | 12 | 591.9 | 583.2 | 590.4 | 590.4 |
| 18 | 12 | 603.3 | 587.3 | 603.6 | 603.3 |
| 19 | 12 | 614.8 | 589.1 | 585.8 | 589.1 |
| 20 | 12 | 607.1 | 594.8 | 584.7 | 594.8 |
| 21 | 12 | 613.2 | 587.5 | 583.3 | 587.5 |
| 22 | 12 | 594.8 | 588.2 | 583.0 | 588.2 |
| 23 | 12 | 600.8 | 587.7 | 605.5 | 600.8 |
| 24 | 12 | 615.5 | 598.0 | 585.6 | 598.0 |
| 25 | 12 | 608.4 | 594.1 | 608.0 | 608.0 |
| 26 | 12 | 606.0 | 583.6 | 590.0 | 590.0 |
| 27 | 12 | 596.8 | 602.2 | 586.4 | 596.8 |
| 28 | 12 | 602.2 | 602.1 | 585.8 | 602.1 |
| 29 | 12 | 601.8 | 587.8 | 601.8 | 601.8 |
| 30 | 12 | 591.4 | 588.5 | 592.8 | 591.4 |
| 31 | 12 | 607.3 | 590.6 | 590.1 | 590.6 |
| 32 | 12 | 595.4 | 606.5 | 586.3 | 595.4 |
| 33 | 12 | 588.0 | 589.8 | 591.0 | 589.8 |
| 34 | 12 | 596.2 | 598.3 | 605.5 | 598.3 |

Event attribution:

| Event | Observation | Tick / added stall |
| --- | --- | --- |
| CUDA Graph capture | Excluded by configuration: `enforce_eager=True`, `warmup_cudagraph=False` | None / N/A |
| KV-pool growth | The paged-KV tensors are allocated once; recorded storage identities and byte sizes remained unchanged across engine-ready and run-boundary snapshots. Fixed-pool page reuse is not growth. | None observed / N/A |
| Unattributed latency outlier | One Full-Wan stepwise measured session contained an 855.6 ms interval at zero-based tick 22. There was no corresponding capture or pool-growth evidence, so it is not labeled as either event. | Tick 22 / +267.8 ms versus the 587.8 ms stepwise steady median |

ABot-World long-session and parity validation

Raw data: [`per-tick.csv`](https://github.com/tsinghua-code/vllm-omni/blob/ffee46ebf35928b35a4a5dd8379ac012f6a8268d/recipes/ACVLab/ABot-World-evidence/per-tick.csv), [`summary.json`](https://github.com/tsinghua-code/vllm-omni/blob/ffee46ebf35928b35a4a5dd8379ac012f6a8268d/recipes/ACVLab/ABot-World-evidence/summary.json), and the [reproduction recipe](https://github.com/tsinghua-code/vllm-omni/blob/ffee46ebf35928b35a4a5dd8379ac012f6a8268d/recipes/ACVLab/ABot-World.md#measured-validation-2026-09-19).

tsinghua-code: Addressing the realtime-vs-offline parity request in [this follow-up](https://github.com/vllm-project/vllm-omni/pull/6392#discussion_r3889679297): the evidence in `ffee46ebf35928b35a4a5dd8379ac012f6a8268d` uses the same image, prompt, seed, precision, resolution, and decoder backend for each paired comparison. It compares the first 117 frames chunk by chunk using exact tensor comparison plus the in-tree ffmpeg PSNR/SSIM helpers.

| Backend | Comparison | Chunks | Exact chunks | RGB8 SSIM | RGB8 PSNR |
| --- | --- | ---: | ---: | ---: | ---: |
| Full Wan2.2 | typed Session-Tick vs offline `Omni.generate()` | 10 | 10/10 | 1.000000 | infinity |
| Full Wan2.2 | typed Session-Tick vs one-generate stepwise | 10 | 10/10 | 1.000000 | infinity |
| TAEW2.2 | typed Session-Tick vs offline `Omni.generate()` | 10 | 10/10 | 1.000000 | infinity |
| TAEW2.2 | typed Session-Tick vs one-generate stepwise | 10 | 10/10 | 1.000000 | infinity |

Thus all **40 same-backend chunk comparisons were bit-exact** (`max_abs_error=0`), and the per-tick parity curve remains flat at SSIM 1.0 throughout the tested horizon. The offline reference deliberately uses the same per-frame decoder, so this isolates serving integration differences from decoder substitution.

For completeness, tiny-vs-full-Wan is reported separately because it changes the decoder: mean float PSNR is **30.13 dB** and mean RGB8 SSIM is **0.7988** over the same ten chunks. Full Wan therefore remains the default.

ABot-World long-session and parity validation

Raw per-chunk metrics: [`quality.csv`](https://github.com/tsinghua-code/vllm-omni/blob/ffee46ebf35928b35a4a5dd8379ac012f6a8268d/recipes/ACVLab/ABot-World-evidence/quality.csv).

tsinghua-code: Addressing the scope correction in [this follow-up](https://github.com/vllm-project/vllm-omni/pull/6392#discussion_r3889680390): I used the suggested **control-free** parity path because offline camera-action trajectories are not supported yet.

One arithmetic correction is needed for ABot-World's actual chunk contract: 117 decoded frames are **10 ticks**, not 13. Tick 0 emits 9 frames and ticks 1–9 emit 12 frames each:

| Tick range | Frames per tick | Cumulative frames |
| --- | ---: | ---: |
| 0 | 9 | 9 |
| 1–9 | 12 | 117 (`9 + 9 × 12`) |

The parity run therefore compares offline `Omni.generate(num_frames=117)` against ten typed realtime chunks and ten one-generate stepwise chunks for each decoder backend. As reported above, all 40 same-backend chunk comparisons are exact, with SSIM 1 and PSNR infinity.

This is intentionally scoped as a single-scene, control-free 117-frame serving-consistency check. It does **not** claim 35-tick offline equivalence, action compliance, or parity with a different upstream whole-clip decoding kernel; those remain separate follow-ups.

tsinghua-code:
Follow-up on review `pullrequestreview-5238992478`:

I rebased this PR onto current `main` (`154a1f25`) and resolved the integration conflicts. In particular, the AR-diffusion runner now uses the single stepwise lifecycle already present on current `main`; the duplicate rebased path was removed.

I also rewrote the affected commits for DCO compliance. All eight PR commits now use `tsinghua-code <zr-liu25@mails.tsinghua.edu.cn>` as the author and contain the matching `Signed-off-by` trailer. The new PR head is `926579a572375c012642ffcb7e327d65d62e0f6d`, and the DCO check is passing.

Because the runtime integration changed during the rebase, I reran the review evidence on one scheduler-reserved NVIDIA H200 (TP=1, BF16 eager, 512x832, seed 42, 4 DMD steps + 1 clean-KV forward per tick, vLLM 0.29.0). Each value is the median of three measured 35-tick runs after a separate warmup:

| Decoder / API | First chunk (ms) | Steady median (ms/tick) | Steady p95 (ms/tick) | Sum of 35 tick intervals (s) |
| Full Wan / typed | 570.1 | 614.6 | 661.3 | 21.617 |
| Full Wan / stepwise | 606.3 | 611.9 | 628.1 | 21.460 |
| TAEW2.2 / typed | 858.3 | 295.7 | 316.0 | 11.073 |
| TAEW2.2 / stepwise | 870.7 | 295.3 | 323.3 | 11.161 |

The refreshed validation also confirmed:

- both backends completed 35 ticks / 417 frames with finite RGB output;
- the two-session interleaving check was bit-exact against the corresponding solo sessions for all six checked chunks;
- closing one session preserved the other, and closing both released all VAE decode state;
- full-Wan persistent decode state was 3,006,062,592 bytes/session within its 3,040,460,800-byte reservation; the largest observed full-Wan PyTorch peak was 46.078 GiB;
- the first 117 control-free frames were bit-exact across typed, stepwise, and offline paths for both backends;
- the tiny-vs-full-Wan single-scene comparison remained 30.13 dB mean float PSNR and 0.7988 mean RGB8 SSIM;
- CUDA Graph capture was disabled, and every KV pool storage identity and size remained fixed. The largest non-warmup interval was 802.1 ms at typed tick 33, so it is retained as an unattributed tail-latency sample rather than mislabeled as capture or KV growth.

Validation completed:

- focused CPU tests: `79 passed`;
- broader diffusion CPU sweep: `224 passed, 14 skipped, 1 deselected`;
- pre-commit on every PR-changed path: passed;
- `git diff --check`: passed;
- generated `ABot-World-evidence/` artifacts remain excluded from the PR.

Current hosted checks: DCO and Python 3.12 build pass; Python 3.11, pre-commit, and Read the Docs are still pending at the time of this comment. Please take another look once the remaining hosted checks complete.

hsliuustc0106:
Thanks for the rebase and the recipe rework — the 35-tick offline-equivalence and streaming-vs-offline consistency sections cover my evidence asks. One blocker left: pre-commit is failing on the current head (926579a5); once it's green I'll flip this to approve.

Issue #6321, opened 2026-08-18
[RFC] Deployment-selectable component stages for diffusion pipelines


hsliuustc0106:
## Proposed architecture

```mermaid
flowchart LR
    subgraph Components["Single model implementation"]
        E["Encoder"]
        D["DiT / AR-DiT"]
        V["VAE decoder"]
        E --> D --> V
    end

    subgraph Mono["Monolithic deployment"]
        M["Encoder + DiT + VAE<br/>one process / stage"]
    end

    subgraph Split["Disaggregated deployment"]
        ES["Encoder pool<br/>encoder cache"]
        DS["DiT pool<br/>step scheduler + KV manager"]
        VS["VAE pool<br/>decode scheduler"]
        ES -->|"ConditioningArtifact"| DS
        DS -->|"LatentArtifact"| VS
    end
```

The model components should have one implementation. Whether they run together or as separate stages should be a deployment choice.

## Motivation

Diffusion pipelines are commonly structured as:

```text
encoder → DiT / AR-DiT → VAE decoder
```

These components increasingly have different requirements:

- Parallelism, hardware, memory, and compute characteristics.
- Scheduling and batching policies.
- Cache lifetimes, including reference encoding, multi-turn editing, conditioning KV, and AR-DiT self-KV.
- Scaling ratios in production deployments.

Monolithic deployment remains preferable for many small models and individual users. Large models and MaaS deployments may benefit from independently scaling and placing each component. Therefore, disaggregation should be supported, but not required.

### 1. Separate model composition from deployment topology

Each model declares its logical components and supported boundaries. A deployment configuration decides whether adjacent components are colocated or disaggregated.

```text
ModelPipelineSpec:
  encoder → denoiser → decoder

DeploymentPlan A:
  [encoder + denoiser + decoder]

DeploymentPlan B:
  [encoder] → [denoiser] → [decoder]
```

Model authors should not implement separate monolithic and disaggregated versions. The same component modules should be invoked either directly in-process or through stage adapters.

A model should expose a split only when benchmarks demonstrate a meaningful placement, scaling, memory, or throughput benefit.

### 2. Introduce typed component boundaries

Stage communication should use typed artifacts instead of model-specific dictionaries:

- `ConditioningArtifact`: text, reference-image, audio, or multimodal conditioning.
- `LatentArtifact`: latent tensor plus layout and decoding metadata.
- `MediaArtifact`: decoded image, video, or audio.
[...]

PR #6395, opened 2026-08-20
[Model][Cosmos3] Register co-located pipeline and add disaggregated two-tower topology

### LINKED ISSUE #6874: [Bug]: Cosmos3 is absent from OMNI_PIPELINES, so its deploy config is silently discarded (no way to disable guardrails)
### Summary

`Cosmos3OmniDiffusersPipeline` is fully implemented and wired into all four class-name dispatch tables in `vllm_omni/diffusion/registry.py`, but it ships without a `PipelineConfig`: `cosmos3_omni` is not a key in `OMNI_PIPELINES`. A deploy YAML is only read *through* that registry, so `--deploy-config` has no effect for Cosmos3. Resolution returns `None`, the engine falls back to a single stage built from CLI kwargs, and the run completes normally — nothing signals that the config was dropped.

Generation itself works today because the pipeline class is reached by `model_index.json:_class_name` dispatch, which is a separate path from deploy-config resolution. Only the config path is dead.

### Where resolution falls through

`StageConfigFactory.get_pipeline_config` (`vllm_omni/config/config_factory.py:264`) has three routes. Cosmos3 misses all three:

1.  `_get_deploy_override_pipe_config` — fires only on an explicit `pipeline:` key in the YAML. This is the one route that raises (`KeyError`) on an unknown key.
2.  Direct lookup `model_type in OMNI_PIPELINES` — `cosmos3_omni` is not a key.
3.  The `hf_architectures` fallback — no registered `PipelineConfig` claims `Cosmos3ForConditionalGeneration`.

So it returns `None`. `create_legacy_stage_configs_from_model` (`config_factory.py:388`) returns `(None, None)`, and `load_and_resolve_stage_configs` (`vllm_omni/entrypoints/utils.py:595`) substitutes `default_stage_cfg_factory()` — the single-stage CLI-defaults fallback — without distinguishing "no registry entry" from "registry entry produced no stages".

Verified on `main` @ `b81aeb7b` with vLLM 0.28.0:

```python
from vllm_omni.config.pipeline_registry import OMNI_PIPELINES
"cosmos3_omni" in OMNI_PIPELINES   # False
```

```console
$ git grep -l Cosmos3ForConditionalGeneration -- '*.py'
$   # no hits anywhere: nothing declares it in hf_architectures
```

The only runtime trace is one line among engine startup logs (`config_factory.py:291`), at WARNING, followed by a successful generation:

```text
WARNING ... Inferred model type cosmos3_omni is not registered to an Omni pipeline
```

### Reproduction

No GPU and no weights needed for the resolution half:

```python
from vllm_omni.config.config_factory import StageConfigFactory

# None, for any deploy YAML that does not carry an explicit `pipeline:` key
StageConfigFactory.get_pipeline_config(
    model="nvidia/Cosmos3-Super-Text2Image",
    trust_remote_code=True,
    deploy_config_path="/tmp/cosmos3.yaml",
)
```

End to end, with `/tmp/cosmos3.yaml` holding `stages: [{stage_id: 0, model_config: {guardrails: false}}]` — the YAML is never applied:

```bash
  --model nvidia/Cosmos3-Super-Text2Image \
  --deploy-config /tmp/cosmos3.yaml \
  --prompt "A small warehouse robot moves a blue box across a clean floor." \
  --height 1024 --width 1024 --num-inference-steps 20 --seed 42
```

### Why it matters

Cosmos3 has one setting that exists **only** in a deploy YAML: `model_config.guardrails`. It has to be `false` at pipeline build time when the gated guardrail checkpoints are unavailable, and there is no CLI flag for it. With no registry entry there is no way to set it, i.e. no way to serve Cosmos3 at all without those checkpoints.

Everything else a deploy YAML carries is also unreachable: Cosmos3 has no shipped default YAML, so no per-stage `devices`, parallelism or sampling defaults, and it cannot participate in a multi-stage topology.

The silence is the worse half. A dropped `--deploy-config` on a registered model raises (route 1); on an unregistered one it is a WARNING that a passing run buries.

### Proposed fix

1.  Add a `COSMOS3_PIPELINE` `PipelineConfig` keyed on `cosmos3_omni`, plus the co-located `deploy/cosmos3_super_t2i.yaml` it names as its `default_deploy_config_name`. Safe to declare `hf_architectures=("Cosmos3ForConditionalGeneration",)` here since the entry is keyed on the checkpoint's real HF `model_type`, so route 3 only ever reaches it for a checkpoint that genuinely is Cosmos3.
2.  Separately, consider making the fallback in `load_and_resolve_stage_configs` distinguish the two `None` cases, so that an explicitly-passed `--deploy-config` that resolves to no pipeline raises rather than being replaced by CLI defaults. Out of scope for the fix above, but it is what would have caught this at the first run.

PR #6395 implements (1), and — on top of that registration — adds a second, opt-in `cosmos3_omni_disagg` topology that runs one stage per Mixture-of-Transformers tower.

I'd like to take this one.

tthakkal:
Fixes #6874

Registers Cosmos3's already-implemented co-located pipeline in `OMNI_PIPELINES`, and adds an opt-in disaggregated topology that runs one stage per Mixture-of-Transformers tower.

## Purpose

Cosmos3 (`nvidia/Cosmos3-Super-Text2Image`) is a Mixture-of-Transformers model with two near-symmetric ~31.2 B towers: an autoregressive UND tower and a diffusion GEN tower.

**1. Registers the co-located pipeline (bug fix, #6874).** `Cosmos3OmniDiffusersPipeline` is fully implemented and wired into all four class-name dispatch tables, but shipped without a `PipelineConfig`, so no Cosmos3 T2I entry existed in `OMNI_PIPELINES` at all. Every route in `StageConfigFactory.get_pipeline_config` then misses, the single-stage CLI-defaults fallback is substituted, and the deploy YAML is **dropped with nothing but one `WARNING ... is not registered to an Omni pipeline`** — the run still produces an image, so nothing signals the config was ignored. Adds `COSMOS3_PIPELINE` plus an opt-in co-located YAML, since Cosmos3 had no deploy-YAML path at all.

It is keyed **`cosmos3_omni_colocated`**, not `cosmos3_omni`, and declares `hf_architectures=()` and no `diffusers_class_name`. Every Cosmos3 checkpoint — T2I, T2V/I2V/V2V and policy — reports the same `model_type=cosmos3_omni` and the same `model_index.json` `_class_name`, so an auto-detectable entry under the bare key would capture the video and policy checkpoints too, and this stage pins `final_output_type="image"` where the fallback correctly resolves `"video"` for them. `main`'s `cosmos3_policy` entry uses exactly this convention, and `tests/config/test_config_factory.py` asserts `"cosmos3_omni" not in OMNI_PIPELINES` for that reason. `deploy/cosmos3_super_t2i.yaml` therefore carries `pipeline: cosmos3_omni_colocated`, which reaches the entry because `_get_deploy_override_pipe_config` is consulted *before* any `model_type` lookup. Omitting `--deploy-config` keeps today's path exactly as it is.

Also in the guardrail path: `--no-guardrails` was negative-only (`store_true`), so a YAML `guardrails: false` could not be overridden back on from the CLI. It becomes one `BooleanOptionalAction` defaulting to `None`; the old spelling is unchanged.

**2. Adds a disaggregated topology, one stage per tower.** The towers are near-symmetric in size but wildly asymmetric in duty cycle — UND runs once per request, GEN once per denoising step (50 by default) — so co-located the idle UND tower's 58.1 GiB stays resident for the whole denoise loop. Split, each tower gets its own GPUs and parallelism. Only the UND tower's per-layer text K/V crosses the boundary, at **~256 KiB per token per CFG branch**, trimmed to the real prompt length; the generator builds a replay stub instead of a UND tower, so no UND weights load there.

Notes for reviewers:

- **Each stage builds only the tower it owns.** `cosmos3_owned_towers` flows into `Cosmos3VFMTransformer(owned_towers=...)`, so the unowned tower is never constructed rather than allocated and pruned; `_require_unowned_absent` asserts that at init. A generator-only stage builds `Cosmos3LanguageModel` `rope_only`, since GEN still needs the rotary path.
- **The handoff is TP-independent.** UND K/V is born TP-sharded, but the reasoner all-gathers the KV-head dimension before the payload leaves the tower and each generator rank re-slices the head range its own cross-attention owns, from its own TP rank. The two stages may run different `tensor_parallel_size`.
- **What crosses the edge is a typed contract:** `Cosmos3TextConditioning`, a frozen dataclass tagged with the code-owned `cosmos3.text_conditioning/v1`, validating branch identity, layout, dtype, batch size and per-branch UND token count. Layout is re-checked against the generator's cross-attention at install; replay entries are keyed by a sha256 fingerprint of the tokenized prompt ids, so a stage that resolved conditioning differently raises instead of producing a wrong image.
- **The seam** is `self.language_model(text_ids, freqs_und)` in `transformer_cosmos3.py`. Calling it directly bypasses the FSDP2 root pre-forward hook, so the reasoner wraps it in `unshard()` / `try-finally reshard()`.
- **Both new registry entries are opt-in**, declaring no `hf_architectures` and no `diffusers_class_name`, so each is reachable only via an explicit `pipeline:` key and neither can hijack the other's checkpoint by existing in the registry. Same convention as `cosmos3_policy` and `hunyuan_image3_ar` / `hunyuan_image3_dit`.
- **Mixed precision is now tower-aware.** `Cosmos3MixedPrecisionRuntime.install` scanned `transformer.gen_layers` unconditionally and raised when it wrapped nothing; a reasoner-only stage has no `gen_layers`, so a quantized checkpoint with `reasoner="native"` died at load with *"found no compatible FP8 or NVFP4 ModelOpt linears"*. It now filters by `owns_generator` / `owns_reasoner` and returns early when the stage has no schedulable tower. The unwrapped-linear guard still fires for a genuinely unquantized checkpoint, and the `getattr` defaults are `True` so a transformer without those attributes keeps the both-towers path.
- **The one-line SPDX header diffs** in 7 files whose content also changes are `check-spdx-header` rewriting `vLLM project` to `vLLM-Omni project` on contact. CI `SKIP`s that hook.

26 files: 11 new (`cosmos3_pipeline_config.py`, `pipeline_cosmos3_disagg.py`, `stage_input_processors/cosmos3.py`, both deploy YAMLs, `recipes/cosmos3/Cosmos3-TowerDisagg.md`, 5 test files) and 15 modified — additive registry hooks in `config/pipeline_registry.py`, `diffusion/registry.py`, `model_extras/registry.py` and `cache/cachedit/model_specific.py` + its test (the generator owns `gen_layers` and the denoise loop, so the existing Cache-DiT enabler applies verbatim; the reasoner has nothing to cache), plus the two transformers, `pipeline_cosmos3.py`, `mixed_precision/runtime.py` + its test, `entrypoints/cli/serve.py`, `cosmos3/guardrails.py` and 2 recipe docs.

## Test Plan

**(A)** 2x H200 141 GiB with the real checkpoint, for anything needing weights or a GPU. **(B)** A CPU host, for pre-commit and the CPU-only suite.

1. **Pipeline resolution** via `StageConfigFactory.get_pipeline_config`: `--deploy-config` omitted, co-located YAML, disagg YAML.
2. **Co-located e2e** on 1x H200, shipped YAML, at 1024x1024 / 1536x1536 / 2048x2048.
3. **Aggregated vs disaggregated** on the same prompt, seed and geometry at TP=1 and TP=2, comparing PNGs by sha256; plus one asymmetric run (reasoner TP=2, generator TP=1).
4. **A negative run** with `default_sampling_params: {max_sequence_length: 256}` on stage 1 only, to try to provoke a fingerprint miss.
5. **Gates and unit tests** on (B): `pre-commit run --files ...`, `pytest tests/config tests/model_executor/stage_input_processors tests/diffusion/cache tests/diffusion/quantization/test_cosmos3_mixed_precision.py`, `pytest tests/diffusion/models/cosmos3/`.

Both YAMLs set `guardrails: false` on every stage, so no per-request override is needed to run this:

```bash
# disaggregated, 2 GPUs; for the aggregated half of the comparison use
# CUDA_VISIBLE_DEVICES=0 and cosmos3_super_t2i.yaml, everything else unchanged
CUDA_VISIBLE_DEVICES=0,1 python examples/offline_inference/text_to_image/text_to_image.py \
  --model nvidia/Cosmos3-Super-Text2Image \
  --deploy-config vllm_omni/deploy/cosmos3_super_t2i_disagg.yaml \
  --prompt "A robot arm cleaning a plate in a bright kitchen" \
  --height 1024 --width 1024 --num-inference-steps 50 --seed 142 \
  --output cosmos3_disagg.png
```

**vLLM Version:** 0.29.0 (torch 2.13.0+cu130)

**vLLM-Omni Commit:** merged with `main` @ `627630d7`

## Test Result

**1. Resolution.** With `--deploy-config` omitted, resolution reaches no registered pipeline and produces the same single-stage topology such a run already got — `try_infer_model_type` and `get_pipeline_config` both return `None`, so `final_output_type` still resolves to `"video"` for video checkpoints. `--deploy-config cosmos3_super_t2i.yaml` gives `cosmos3_omni_colocated`, 1 stage, every degree 1, HSDP off; the disagg YAML gives `cosmos3_omni_disagg`, 2 stages. Neither is reachable by either auto-detection path. Asserted in `tests/config/test_cosmos3_topologies.py`, which also asserts `"cosmos3_omni" not in OMNI_PIPELINES` so this file carries `main`'s invariant too.

**2. Co-located on one card.** Both towers are 120.91 GiB of bf16 weights and fit on one H200, so the shipped default needs no collectives. Correct images at all three sizes, 20 steps:

| | 1024x1024 | 1536x1536 | 2048x2048 |
|---|---|---|---|
| peak | 129.302 GB | 136.798 GB | 140.308 GB |
| per step | 253.8 ms | 558.2 ms | 1085.1 ms |
| total | 5.08 s | 11.17 s | 21.71 s |

**3. The tower split is numerically transparent, and halves per-stage memory.** Same prompt, seed and geometry, PNGs compared by sha256:

| | aggregated | disaggregated | identical |
|---|---|---|---|
| TP=1 | 121.56 GiB | 63.43 / 62.04 GiB per stage | yes  |
| TP=2 | 63.78 GiB per rank | 34.52 / 33.13 GiB per rank | yes  |

The TP=1 row was re-run end to end on vLLM 0.29.0 after the merge with `main`, through both shipped deploy YAMLs, and both layouts produced the same 1024x1024 PNG bit for bit. The TP=2 row was measured on 0.28.0; the absolute hash moves with the vLLM version, the aggregated-vs-disaggregated equality does not. The re-run also reproduced the per-stage figures exactly — `kept 720/1425` and `kept 718/1425` tensors, 63.43 GiB and 62.04 GiB resident, a 7.5 MiB payload logged as `heads [0, 8) of 8` — at 246.6 ms/step co-located and 264.9 ms/step on the generator stage (12.3 s vs 14.2 s end to end, the reasoner stage adding 972 ms).

Each stage keeps only its own tower (`kept 720/1425` and `kept 718/1425` tensors), so the split fits two 80 GB cards; the ~2 GiB above 58.1 GiB of raw weights is the VAE and sound tokenizer each stage builds. Changing TP shifts pixels slightly but by the same amount in both layouts, so that delta is reduction order in the sharded matmuls, not the split. At TP=2 the generator ranks log `heads [0, 4) of 8` and `heads [4, 8) of 8` with the payload the same size as at TP=1, and an asymmetric run (reasoner TP=2, generator TP=1) also renders correctly.

**4. The negative run did not miss.** `max_sequence_length: 256` on stage 1 alone gave no fingerprint miss and a byte-identical image, both stages still logging 4096: every current request path replaces a diffusion stage's YAML defaults wholesale with one request-level `OmniDiffusionSamplingParams` cloned to all stages, so a value pinned on one tower is silently dropped. The YAML comment and the recipe now say so and point at per-stage `sampling_constraints` as the route that does reach the check, which is unchanged and still exercised by `test_a_stage_that_resolves_conditioning_differently_misses`.

**5. Tests.** 211 new tests across 5 new files (`core_model` + `cpu`, plus `diffusion`; no GPU or checkpoint needed): topology registration and both YAMLs (33), the reasoner -> generator bridge (26), both tower pipelines with the typed contract, fingerprinting and the replay stub (133), the K/V payload through `OmniSerde` bit-exactly (8), and the whole chain from real reasoner `forward` to real generator install and replay (11) — plus new ownership and TP cases in the existing transformer tests, and 5 mixed-precision cases covering every tower x precision combination. `TestMultiRankShardIdentity` patches `_tp_rank` and gives every (layer, head) distinguishable contents, so it fails outright under a rank-0-only handoff; forcing `kv_head_offset = 0` fails 10 tests. Reverting the mixed-precision fix alone fails `test_reasoner_only_stage_installs[native]` with the original `ValueError`, so that test is not vacuous.

Sweep of `tests/diffusion/models/cosmos3` is **326 passed / 3 skipped**; the other touched trees (`tests/config`, `tests/model_executor/stage_input_processors`, `tests/diffusion/cache`, `tests/diffusion/quantization/test_cosmos3_mixed_precision.py`) are **1363 passed / 1 skipped**.

`pre-commit run --all-files` with the CI `SKIP` set exits 0. Run over the diff without the `SKIP`s, every hook passes except two pre-existing mypy errors in `test_cosmos3_mixed_precision.py` that reproduce identically on `main`.

**6. Docs.** `recipes/cosmos3/Cosmos3-TowerDisagg.md` documents the split as experimental, is listed in `recipes/README.md`, cross-linked from the Cosmos3-Super recipe, records the measurements above in a **What has been measured** section, and passes `markdownlint-cli2`.


tthakkal (commented):
**Self-review**

Two changes: `cosmos3_omni` was implemented but never registered in `OMNI_PIPELINES`, so `resolve_pipeline_config` returned `None` and any `--deploy-config` was **silently dropped** (bug fix); plus an opt-in `cosmos3_omni_disagg` topology with one stage per MoT tower. Registry edits are additive — no other model's resolution changes.

Checked on H200 141 GiB against the real checkpoint, not mocks:

- **Resolution.** Omitted `--deploy-config` and the co-located YAML both give `cosmos3_omni` / 1 stage; the disagg YAML gives 2 stages and is unreachable by auto-detection (no `hf_architectures`, no `diffusers_class_name`), so it can't hijack a co-located deploy by merely being registered.
- **Co-located default fits one card.** I shipped this as a 2-GPU default first, then measured instead of trusting it: 120.91 GiB of weights fit on one card, correct images at 1024²/1536²/2048² (129.3 / 136.8 / 140.3 GB peak), and a second card is *slower* at 1024² — 439.4 vs 253.8 ms/step, because HSDP all-gathers each layer from the peer every step. Default is now one card, no collectives; scale-out is documented as a headroom knob (2048² leaves 3.4 GiB).
- **Disagg on 2 cards.** Correct image, byte-identical PNG across two runs, 8.2 MiB K/V at 0.72 ms across the boundary, no fingerprint miss. Tower split verified independently: safetensors headers + `_remap_ckpt_key` predict 720 / 718 of 1425 tensors, exactly what each stage logs.

Specifically went looking for: the FSDP2 seam (`self.language_model(...)` bypasses the root pre-forward hook, so the reasoner brackets it in `unshard()` / `try-finally reshard()`); CFG cross-wiring (replay keyed on a sha256 of the tokenized prompt ids, a miss raises rather than producing a subtly wrong image); and guardrails, which default on and eager-load *gated* models at build time, so both YAMLs disable them in the default path.

Deliberate trade-off: the reasoner builds a VAE it never uses (~1.3 GiB) rather than forking `Cosmos3OmniDiffusersPipeline.__init__`. Deploy YAMLs are tested for the invariants HSDP enforces at startup, not literal degrees, so rescaling a layout needs no test edit.

1098 passed / 1 skipped across `tests/config`, `tests/cosmos3`, `tests/diffusion/models/cosmos3`, `tests/model_executor/stage_input_processors`; `ruff check` and `ruff format --check` clean.

hsliuustc0106:
sorry for missing such an important PR, we recently have an early trial in minimax-h3, I will take a deep review on this one as well to align the arch design and user interface

vllm_omni/diffusion/models/cosmos3/pipeline_cosmos3_disagg.py
```diff
@@ -0,0 +1,719 @@
...
+                with transformer._offload_context("reasoner"):
+                    cached_kv_full = transformer.language_model(text_ids, freqs_und)
+
+                # Trim padding exactly as the co-located forward does. Shipping
+                # already-trimmed K/V makes the generator's own trim a no-op and
+                # keeps the payload proportional to the real prompt length.
+                # ``.cpu()`` also materializes any DTensor-backed result into a
+                # plain tensor, which is what has to cross the stage boundary.
+                table[fingerprint_text_ids(text_ids)] = [
+                    (k[:, :max_real_len].contiguous().cpu(), v[:, :max_real_len].contiguous().cpu())
```
hsliuustc0106: [P1] Preserve TP rank identity in the K/V handoff

With both stages at TP=2, each reasoner rank produces different KV heads (`Cosmos3CausalAttention` uses `ColumnParallelLinear(..., gather_output=False)`), but `MultiprocExecutor.execute_request` requests output only from rank 0. The generator then receives that same payload on every rank. Matching TP sizes make the shape checks pass, so generator rank 1 silently conditions on rank 0's KV heads instead of its own. Gather and redistribute the matching shards, or reject TP>1 until rank-aware handoff is supported. A multi-rank test with distinguishable shard contents would catch this; the current single-process replay tests cannot.

vllm_omni/diffusion/models/cosmos3/pipeline_cosmos3_disagg.py
```diff
@@ -0,0 +1,719 @@
...
+    has been read, so this saves device memory, not startup I/O; see the module
+    docstring.
+    """
+
+    def __init__(self, *, od_config: Any, prefix: str = "") -> None:
+        super().__init__(od_config=od_config, prefix=prefix)
+        # Must happen here, not in load_weights -- see the module docstring.
+        # ``weights_sources`` is assigned by the base __init__, so the loader
+        # still finds the checkpoint after this returns.
+        self._drop_unused_tower()
```
hsliuustc0106: [P1] Avoid allocating both towers before dropping one

With HSDP and CPU offload disabled, `DiffusionModelRunner.load_model` selects the CUDA device and `DiffusersPipelineLoader._init_from_load_format` constructs this pipeline under that device context. `super().__init__()` therefore allocates the complete UND and GEN towers before `_drop_unused_tower()` runs. This prevents startup on an 80 GB card even though the retained tower fits, defeating the documented two-smaller-card use case. Construct only the required tower, or defer device allocation until after pruning. The H200 run has enough memory for both towers, and the unit fixtures bypass this constructor with `object.__new__`, so neither exercises this failure.

hsliuustc0106 (commented):
Two correctness issues found by tracing the initialization and multi-rank handoff paths at 4615878b4b0a. Static review; GPU inference was not run.

hsliuustc0106:
Following up on the alignment with MiniMax-H3: this is directionally consistent with #5885, but only partially follows the newer conditioning-boundary work in #6720 and the direction proposed in #6321. This comparison is against Cosmos3 head `4615878b4b0a`.

The opt-in topology and reuse of the existing Cosmos3 transformer/denoising implementation are aligned. The main gaps are:

- **Component ownership:** H3 skips constructing its local text encoder when it consumes external conditioning. Cosmos3 constructs both towers and then deletes one. Constructing only the owned components would also address the startup-memory issue in the inline review.
- **Conditioning contract:** H3 now has a typed `MiniMaxH3TextConditioning` contract with the code-owned `minimax_h3.text_conditioning/v1` schema identifier and validation of shape, dtype, layout, token counts, and token tags. Cosmos3 currently carries a dictionary through `trajectory`, with prompt fingerprints and partial layout validation. A model-specific typed K/V contract should make branch identity, tensor layout, sharding, ownership, and lifetime explicit.
- **Independent parallelism:** H3 transfers full hidden-state conditioning, while Cosmos3 transfers TP-local K/V. The handoff therefore needs to preserve shard identity, or reject unsupported TP configurations, as described in the other inline comment.

Using a diffusion worker for Cosmos3's UND tower is reasonable: this is a prompt-encoding pass inside the existing MoT implementation. Alignment does not require copying H3's `LLM_AR` stage choice.

My recommendation is to construct only owned components, introduce a typed Cosmos3 conditioning contract, and expose conditioning production/consumption as a shared component boundary used by both layouts. The replay stub could remain as an interim adapter. Keep this model-specific for now: #6720 explicitly deferred generic stage/artifact abstractions, and #6321 remains a proposal.

The two inline findings are correctness blockers. The broader boundary cleanup above is an architectural recommendation, not a request to implement the entire RFC in this PR.

docs/features/cosmos3_tower_disaggregation.md
```diff
@@ -0,0 +1,170 @@
+# Cosmos3 Tower Disaggregation (experimental)
```
hsliuustc0106: I think this is not a new feature

vllm_omni/diffusion/models/cosmos3/pipeline_cosmos3_disagg.py
tthakkal: Confirmed — fixed by gathering rather than by rejecting TP>1.

The reasoner trims each branch to its real prompt length and all-gathers the KV-head dimension before the payload leaves the tower, so the wire always carries the full 8-head set (9.2 MiB at both TP=1 and TP=2). Each generator rank then slices `[_tp_rank() * num_kv_heads_local, +num_kv_heads_local)` at bind time, from its own rank and never from the payload; a TP that would leave a rank zero KV heads raises there, naming both TP sizes.

`TestMultiRankShardIdentity` is the multi-rank test you asked for: it patches `_tp_rank` and gives every (layer, head) distinguishable contents, so `test_the_two_ranks_do_not_receive_the_same_heads` fails outright under the old rank-0-only handoff. Forcing `kv_head_offset = 0` fails 10 tests in all.

On 2×H200 the generator ranks log `heads [0, 4) of 8` and `heads [4, 8) of 8`, and aggregated vs disaggregated images are byte-identical at TP=1 and at TP=2. Asymmetric TP (reasoner 2 / generator 1) also renders correctly.

vllm_omni/diffusion/models/cosmos3/pipeline_cosmos3_disagg.py
tthakkal: Confirmed — each stage now constructs only the tower it owns instead of pruning after the fact.

The tower pipelines declare `cosmos3_owned_towers` and `__init__` forwards it to `Cosmos3VFMTransformer(owned_towers=...)`, so the unowned tower's blocks are never built; on a generator-only stage `Cosmos3LanguageModel` is built `rope_only`, since GEN still needs the rotary path. `_drop_unused_tower` is gone, and `_require_unowned_absent` asserts the unowned `ModuleList` is empty right after construction, so the stage fails loudly at init if the flag ever stops arriving.

Startup I/O is unchanged — both stages still stream the whole checkpoint and filter the other tower's tensors, which the recipe states — but per-stage peak on H200 drops from 121.56 GiB to 63.43 / 62.04 GiB, so the 80 GB card case now starts.

tthakkal:
> Following up on the alignment with MiniMax-H3: this is directionally consistent with #5885, but only partially follows the newer conditioning-boundary work in #6720 and the direction proposed in #6321. This comparison is against Cosmos3 head `4615878b4b0a`.
> 
> The opt-in topology and reuse of the existing Cosmos3 transformer/denoising implementation are aligned. The main gaps are:
> 
>     * **Component ownership:** H3 skips constructing its local text encoder when it consumes external conditioning. Cosmos3 constructs both towers and then deletes one. Constructing only the owned components would also address the startup-memory issue in the inline review.
> 
>     * **Conditioning contract:** H3 now has a typed `MiniMaxH3TextConditioning` contract with the code-owned `minimax_h3.text_conditioning/v1` schema identifier and validation of shape, dtype, layout, token counts, and token tags. Cosmos3 currently carries a dictionary through `trajectory`, with prompt fingerprints and partial layout validation. A model-specific typed K/V contract should make branch identity, tensor layout, sharding, ownership, and lifetime explicit.
> 
>     * **Independent parallelism:** H3 transfers full hidden-state conditioning, while Cosmos3 transfers TP-local K/V. The handoff therefore needs to preserve shard identity, or reject unsupported TP configurations, as described in the other inline comment.
> 
> 
> Using a diffusion worker for Cosmos3's UND tower is reasonable: this is a prompt-encoding pass inside the existing MoT implementation. Alignment does not require copying H3's `LLM_AR` stage choice.
> 
> My recommendation is to construct only owned components, introduce a typed Cosmos3 conditioning contract, and expose conditioning production/consumption as a shared component boundary used by both layouts. The replay stub could remain as an interim adapter. Keep this model-specific for now: #6720 explicitly deferred generic stage/artifact abstractions, and #6321 remains a proposal.
> 
> The two inline findings are correctness blockers. The broader boundary cleanup above is an architectural recommendation, not a request to implement the entire RFC in this PR.

@hsliuustc0106 
Thanks — this shaped the revision. All three gaps are addressed; the RFC-scale piece deliberately is not.

**Component ownership.** `cosmos3_owned_towers` → `Cosmos3VFMTransformer(owned_towers=...)`: the unowned tower is never constructed, `_require_unowned_absent` asserts it, `_drop_unused_tower` is gone. Per-stage peak drops from 121.56 GiB to ~63 GiB on H200.

**Conditioning contract.** The dict through `trajectory` is now a frozen `Cosmos3TextConditioning` tagged `cosmos3.text_conditioning/v1`, validated on construction: branch keys (each a prompt fingerprint, empty set rejected), layout (`num_layers` / `num_kv_heads` / `head_dim` positive and matching every tensor, re-checked against the generator's cross-attention at install), dtype and batch size payload-wide, and UND token count per branch — per branch because each is trimmed to its own real prompt. Schema mismatch is reported by name before any field check. `reasoner_tp_size` is declared but constrains nothing, since the payload is gathered and therefore unsharded; it is carried to be named in layout errors. Cosmos3 has no analogue of H3's token tags — the payload is per-layer K/V for one whole formatted prompt, and branch identity lives in the keys and fingerprints.

**Independent parallelism.** Shard identity is preserved by gathering on the reasoner and re-slicing per generator rank, so the handoff is TP-independent rather than TP-restricted; verified on hardware with reasoner TP=2 / generator TP=1.

**Boundary and lifetime.** Kept model-specific, with the replay stub as the interim adapter you suggested. Toward the shared boundary this PR only lifts the two conditioning resolvers out of the co-located `forward` so all three pipelines share one implementation. Exposing conditioning production/consumption as a full shared component boundary is out of scope here — with #6720 deferring generic stage/artifact abstractions and #6321 still a proposal, I would rather follow than lead. Happy to take it as a follow-up once one lands.

docs/features/cosmos3_tower_disaggregation.md
tthakkal: Agreed — it is a deployment recipe, not a feature. Moved to `recipes/cosmos3/Cosmos3-TowerDisagg.md`, with the `docs/.nav.yml` and `docs/features/README.md` entries removed, a row added to `recipes/README.md`, and a cross-link from `Cosmos3-Super.md`. Nothing is left under `docs/features/`. It also now carries a **What has been measured** section with the per-stage memory and bit-exactness results.

vllm_omni/deploy/cosmos3_super_t2i.yaml
```diff
@@ -0,0 +1,112 @@
...
+pipeline: cosmos3_omni_colocated
+async_chunk: false
+
+stages:
+  # `devices` are LOGICAL indexes into the process's CUDA_VISIBLE_DEVICES, not
+  # physical ids: _map_device_list treats each entry as a 0-based index into the
+  # visible list and silently drops any entry >= len(visible).
+  - stage_id: 0
+    devices: "0"
+    # One request in flight. Not a model limitation -- it is what makes the
```
lishunyang12: Keep max_num_seqs at one; adding GPUs does not enable Cosmos3 request batching.

tthakkal: Good catch — fixed, thanks. 

There's no batching to enable at all: no Cosmos3 pipeline sets supports_request_batch, so DiffusionEngine raises at init for max_num_seqs > 1 on any number of GPUs. Dropped max_num_seqs from the scale-out list and reused the capability-gate wording already in cosmos3_super_t2i_disagg.yaml.

Issue #6330, opened 2026-08-18
[RFC] Quantization-aware checkpoint mmap for DLO AllGather online FP8


hsliuustc0106:
## Problem

PR [#6279](https://github.com/vllm-project/vllm-omni/pull/6279) enables per-tensor online FP8 with DLO AllGather, but the direct checkpoint-mmap path remains disabled for online quantization.

Today, the loader must use the ordinary loader to:

1. load BF16 checkpoint weights;
2. finalize online FP8 weights and scales;
3. move the quantized runtime tensors back to host memory; and
4. let DLO shard those finalized tensors.

This makes online FP8 compatible with DLO AllGather, but it still requires substantial startup materialization before DLO owns the final shard. The existing `stream_online_quant_weights_to_cpu` path bounds accelerator residency; it does not provide a quantization-aware, direct mmap path.

[...]

PR #6399, opened 2026-08-20
[RFC #6330] Add node-local runtime FP8 cache for DLO AllGather

baonudesifeizhai:

## Purpose

## Summary

Adds a node-local, canonical runtime FP8 cache for DLO AllGather online FP8.

- Streams DiT BF16/FP16 checkpoint tensors into runtime-ready FP8 qweights and scales with bounded GPU scratch space.
- Warm starts mmap the cached runtime layout and reuse the existing online FP8 linear/kernel path.
- Uses atomic publish and per-key builder locking so concurrent workers share one cold build.
- Keeps the cache scoped to CUDA, DiT-only, `Fp8PerTensorOnlineLinearMethod`, TP=1, non-HSDP, and DLO AllGather.
- Dynamic LoRA remains a runtime overlay; static LoRA bypasses the cache.

## Test Plan

- vLLM Version: `0.27.0`
- vLLM-Omni Commit: `203ba8ce`
- GPUs: 4 GPUs, DLO AllGather
- Output validation: video-frame and audio SHA256 equality

## FL2VA E2E

Online FP8 + DLO AllGather, 3 denoising steps. For every topology, baseline, cold-cache, and warm-cache produce byte-identical video frames and audio.

| Topology | Baseline init | Cold-cache init | Warm-cache init | Warm-cache improvement |
|---|---:|---:|---:|---:|
| DP2 / SP1 | 165.42s | — | 83.25s | 49.7% / 1.99x |
| DP2 / SP2 | 172.77s | 149.37s | 82.49s | 52.3% / 2.09x |
| DP4 / SP1 | 122.37s | 114.31s | 72.10s | 41.1% / 1.70x |

The DP2/SP2 cold-cache run elects one builder; the other three workers reuse the published artifact. Warm-cache workers load the runtime mmap layout directly.

## Ref2VA DP2/SP2 E2E

50 denoising steps with the official Ref2VA input video and audio. Baseline, cold-cache, and warm-cache video/audio outputs are byte-identical.

| Run | Init | Valid generation wave | Init improvement |
|---|---:|---:|---:|
| Baseline | 153.34s | 324.28s | — |
| Cold cache | 116.20s | 323.92s | 24.2% / 1.32x |
| Warm cache | 89.16s | 324.63s | 41.9% / 1.72x |

Cold start publishes one cache entry and three peer workers reuse it. Warm start hits `online_fp8_runtime_mmap` on all four workers and skips ordinary BF16 materialization.

The Ref2VA runtime cache artifact is 31 GiB on disk.

## Ref2VA DP2/SP2 Memory

Peak values aggregated across the four DLO workers.

| Phase | Baseline PSS / RSS / HBM | Cold cache PSS / RSS / HBM | Warm cache PSS / RSS / HBM |
|---|---:|---:|---:|
| Initialization | 212.7 / 246.7 / 91.6 GiB | 237.5 / 263.0 / 52.1 GiB | 235.8 / 260.9 / 57.5 GiB |
| Generation | 212.0 / 214.4 / 78.5 GiB | 213.9 / 216.2 / 78.5 GiB | 213.4 / 215.7 / 78.5 GiB |

The cache substantially reduces initialization HBM. Generation HBM and worker PSS/RSS remain effectively unchanged. Initialization PSS is higher for cache runs, so this PR does not claim OS page-cache early reclamation.

## Nsight Systems

DP2/SP1 initialization trace, baseline vs. warm cache:

| Metric | Baseline | Warm cache |
|---|---:|---:|
| Engine initialization | 144.92s | 81.15s |
| `cudaMemcpyAsync` API time | 59.31s | 8.23s |
| GPU memcpy volume | 424.5 GiB | 239.2 GiB |
| GPU memcpy operations | 10,997 | 8,025 |
| Online FP8 quantization kernels | 520 | 0 |

The startup gain comes from avoiding repeated BF16 checkpoint transport, materialization, and online FP8 quantization. Runtime GEMM execution is unchanged, as expected.

## Follow-ups

- Validate Ref2VA against the official `r2va.mp4` SSIM/PSNR accuracy threshold. Current Ref2VA coverage validates baseline/cold/warm cache parity.
- Add Ref2VA DP4/SP1 coverage.
- Add corrupt/stale/interrupted cache recovery coverage.
- Add explicit page-cache reclamation (`madvise` / `posix_fadvise`) and validate it with system page-cache metrics.
- Collect PSS/RSS/HBM metrics for FL2VA/T2VA as well.
phase2  will  support valid generation wave


baonudesifeizhai:
automated review @vllm-omni-review-bot

hsliuustc0106:
## Architecture/correctness: make FP8 a producer of the shared host-weight store

I reviewed frozen head `e8cd6570df566a9e2975a039c9b814c7013c0053`. DCO, pre-commit, and the Python builds pass; `git diff --check`, import/compile checks, and 24 focused existing mmap/online-quant/offloader tests pass locally. However, no test under `tests/` references `RuntimeFp8Cache`, the new cache key, the FP8 cache plan, or the new CLI setting.

My assumption is that this is intended to be the canonical FP8 producer for the same node-local host-weight storage layer being discussed in #6240, rather than a second cache runtime maintained independently. Under that interpretation, the streaming quantization work here is valuable, but the storage lifecycle needs to be shared.

### Cache corruption cannot be deferred

`RuntimeFp8Cache._lookup()` currently accepts an entry when:

- `COMPLETE` exists;
- `manifest.json` parses;
- `manifest["identity"] == key.identity`.

The manifest contains only `identity` and component paths. It does not record or validate shard existence, file size/digest, tensor names, shapes, or dtypes. `build_online_fp8_runtime_mmap_plan()` also bypasses `_validate_source_metadata()`, and `_load_weights_via_mmap()` installs the mapped tensor without checking it against the runtime target.

That creates two concrete failure modes:

1. a missing/truncated shard is still reported as a cache hit, then warm startup fails while building or realizing the plan instead of invalidating/rebuilding/falling back;
2. raw tensor bytes can change without changing the manifest, and the corrupted FP8 weight is consumed silently.

The publication path closes files and renames the directory, but does not `fsync` the files, manifest, temporary directory, or parent directory before/after publication. Therefore `COMPLETE` is not a durable proof after a host/power failure.

Please do not leave corrupt/stale/interrupted recovery as a follow-up. It is part of the minimum correctness contract for a persistent cache that skips ordinary BF16 loading and online quantization. Before merge, the published manifest should carry authoritative tensor metadata plus per-file/content digests; lookup should validate the complete artifact and, under the key lock, invalidate and rebuild or take an explicit safe fallback. Publication should use the same durable temp → fsync → atomic rename protocol.

### Reuse one `HostWeightStore`

This PR independently implements:

- source/producer identity;
- per-key `flock`;
- builder election;
- temporary publication and atomic rename;
- manifest lookup and invalidation;
- cache-hit source rewriting.

Those are store responsibilities, not FP8 responsibilities, and they overlap the host cache lifecycle in #6240. Maintaining both will make recovery, identity, observability, leases, and future eviction diverge—the missing integrity checks here are already an example.

I recommend this boundary:

- **`HostWeightStore`** owns pre-build identity, locking, atomic/durable publication, manifest/integrity validation, hit/miss/build/fallback reporting, and a `HostWeightLease` that keeps mapped backing alive.
- **This PR's FP8 producer** owns `_component_specs`, the two-pass per-tensor scale calculation, bounded GPU quantization, stream transforms, and streamed safetensors output.
- **The loader** asks `store.get_or_build(identity, fp8_producer)` before ordinary materialization and converts the returned lease into a `HostWeightPlan`.
- **The DLO transport** owns sharding, pinned buffers, H2D, AllGather/chunking, streams/events, and teardown. It consumes the lease/plan but never parses cache manifests or takes publication locks.
- Dynamic LoRA remains a runtime overlay, as this PR already intends.

The latest commit removed the LRU lease implementation along with eviction. That is fine for v1, but a lease is still useful as ownership of open mappings/file handles and provides the correct future seam; it does not require implementing eviction now.

This makes the dependency direction clear:

- #6240 should establish the representation-neutral store contract;
- #6399 should add the FP8 producer/schema and remain responsible for proving FP8 equivalence;
- #6374 should consume a host-backing lease/plan as a transport input, without depending on either cache implementation.

If #6240's current one-shot API is not suitable, factor the small store foundation into a separate PR and rebase both producers on it rather than merging two cache runtimes.

### Required tests

Please add focused tests proving:

- a warm hit never invokes the FP8 producer or ordinary BF16 materialization;
- concurrent workers elect exactly one producer;
- missing, truncated, byte-corrupted, wrong-shape, and wrong-dtype artifacts are rejected and rebuilt/fallback safely;
- an interrupted writer leaves no publishable entry and the next writer recovers;
- lease/mmap handles are released after DLO has copied its persistent AllGather shards;
- the FP8 producer and the #6240 final-layout producer both satisfy the same store contract.

One smaller cleanup: `--dlo-fp8-cache-max-bytes` remains in `serve.py`, but the final commit removed every config field and implementation consumer for it. It is currently a public no-op; remove it until eviction is implemented with a defined lease policy.

The reported warm-start and output-parity results are useful evidence for the producer. The store correctness and shared ownership boundary above should be resolved before this becomes another independent runtime.

baonudesifeizhai:
will start a new one base on https://github.com/vllm-project/vllm-omni/pull/6419

PR #6402, opened 2026-08-20
[Bugfix][MiniCPM-o] Align active-response final append budget

natureofnature:
## Summary

- Snapshot whether a native-duplex response is already active when an audio append is accepted.
- Avoid the 12-slot final reserve only for successful exact-unit appends accepted during an active response.
- Preserve idle/new-response behavior and add cross-layer regressions for active, precreated, post-`response.done`, overlapping, and client-spoofed states.

## Root cause

During an active response, an exact-chunk final append reserved 25 scheduler positions while MiniCPM Stage 0 materialized only 13 embeddings. Preprocessing filled the 12-position surplus with `</unit>` embeddings, which changed the KV context and delayed the model-owned semantic stop.

The serving layer now records the active-response state at input acceptance time, before a new response may be precreated. The MiniCPM runtime uses that server-owned marker to omit the unmatched reserve only for the affected path.

## Scope

This aligns the default model-owned `listen_only` path after output has begun. It does not add hard server-side barge-in, atomic cross-stage cancellation/fencing, stale audio suppression, or pre-first-output interruption.

Related to #6101.

## Validation

- Commit hooks: passed
- Focused regression: 8 passed
- Related duplex runtime/handler files: 231 passed
- Full duplex engine suite: 85 passed
- Ruff check/format and `git diff --check`: passed
- Real-model GPU matrix on aligned inputs:
  - candidate: 8/12 semantic stops
  - MiniCPM challenge reference: 6/12
  - prior main: 2/12
  - no-interrupt controls: 4/4
  - client/server errors: 0

The GPU matrix used the same model snapshot, prompts, audio inputs, sampling parameters, request order, and observation window for all compared arms.

### DIFF
--- a/tests/e2e/features/fullduplex/engine/test_duplex_runtime.py
+++ b/tests/e2e/features/fullduplex/engine/test_duplex_runtime.py
@@ -1,3 +1,4 @@
+from base64 import b64encode
 from dataclasses import FrozenInstanceError
 from types import SimpleNamespace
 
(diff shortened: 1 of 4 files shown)


natureofnature:
@codex review

PR #6404, opened 2026-08-20
[Bugfix][MiniCPM-o] Bind omni-duplex video frames to the unit they close

amy-why-3459:
## Purpose

Fix MiniCPM-o 4.5 native-duplex **video/audio pairing** on `/v1/realtime?duplex=1`, and add a Realtime video-input demo on the same session.

Official `streaming_prefill(audio_waveform=<1 s>, frame_list=[frame])` puts a second of audio and the picture captured during it into the **same** Stage0 unit. Two holes broke that pairing.

### Frame cadence

Stage0 closes unit 0 at 1030 ms of appended audio (`first_chunk_ms=1035` aligned down to hop 160), then every later unit 1000 ms after that. Frames are **not** carried across appends.

The client used to put frame 0 on the first 200 ms append. That append cannot close a unit, so Stage0 dropped the frame under a generic "audio not enough" result and every later frame landed one unit ahead of its audio.

`RealtimeDuplexClient.stream_pcm16` (and the demo helpers) now send frame `k` on the append that closes unit `k` (`duplex_unit_boundary_ms`). A clip shorter than the audio holds its last frame; a still image is a one-element list and repeats.

### Stacked frames in one unit

The wire allows at most two JPEGs per append (`stack_frames=2` also sends the previous frame so motion is visible). Stage0 used to emit **one** frame per unit and leave the extra one unmatched, fail the prefill, and still advance `audio_chunk_idx`.

The whole append's `frame_list` now goes into the unit that append closes: `<unit>`, then each frame as `<image>` + 64 resampler embeddings + `</image>`, then the audio. A frame on an append that closes no unit is rejected with an explicit reason instead of being dropped.

Vision hidden states are also cast to the token-embedding dtype before they enter the unit, matching official `get_vllm_embedding`.

### Video input demo

`examples/online_serving/minicpmo/realtime_duplex_demo.py` accepts `--input-video`. PyAV demuxes JPEG frames at `--video-fps` (default 1.0); vLLM `load_audio` extracts a 16 kHz mono WAV unless `--input-wav` overrides the soundtrack (frame count then follows the WAV duration, so a silent clip does not fail `load_audio`). `--stack-frames 2` sends the previous+current pair. Frames keep capture size (`--frame-max-side 0`); the server `process_image` normalizes at 448.

## Test Plan

**vLLM Version:** (fill)

**vLLM-Omni Commit:** (this branch)

- [x] Unit: frame `k` rides the append that closes unit `k`; frame 0 is never sent before 1030 ms
- [x] Unit: last frame is held when audio outlives the clip; still image repeats
- [x] Unit: `stack_frames=2` sends `[f0]` then `[f0, f1]` then `[f1, f2]`
- [x] Unit: `--input-wav` + `--input-video` uses WAV duration and skips video soundtrack
- [x] Unit: Stage0 puts every frame of an append into one unit (`<unit> <image>…</image> <image>…</image> <|audio|>`)
- [x] Unit: a frame on an append that closes no unit fails with "closes no model unit" and does not advance `audio_chunk_idx`
- [ ] E2E: `test_duplex_single_session_video_input` (1 fps camera track, `stack_frames=2`, 2 turns stay intact)
- [ ] Manual video demo:

```bash
python examples/online_serving/minicpmo/realtime_duplex_demo.py \
    --url ws://localhost:8099/v1/realtime?duplex=1 \
    --input-video /path/to/clip.mp4 \
    --stack-frames 2 \
    --ref-audio /path/to/MiniCPM-o-Demo/assets/ref_audio/ref_minicpm_signature.wav \
    --output-dir /tmp/minicpmo_realtime_duplex_video_demo

```


vllm_omni/experimental/fullduplex/minicpmo45/stage0.py
```diff
@@ -176,14 +188,30 @@ def _stage_prefill_embeddings_only(
...
+        # frames evenly across the units this loop will build.
         frame_blocks: list[Any] = []
         if video_frames:
             self._require_vision_token_ids()
-            frame_blocks = self._stage_vision_embeddings(video_frames)
-            if frame_blocks is None or len(frame_blocks) != len(video_frames):
-                return self._stage_prefill_result(False, start_time, "streaming vision embedding failed")
+            try:
+                frame_blocks = self._stage_vision_embeddings(video_frames)
+            except _VisionEncodeError as exc:
```
Gaohan123, line 219: [P1] Vision failure still advances the reserved span

Returning a failed prefill here does not cancel the scheduler span. MiniCPMO45OmniForConditionalGeneration.preprocess() falls back to the original input_ids/input_embeds, so malformed frames or a vision-tower failure can still feed placeholder embeddings into KV while losing the associated audio. Abort/rollback the span or fall back to audio-only, and add a test proving num_computed_tokens/KV do not advance.

vllm_omni/experimental/fullduplex/minicpmo45/input.py
```diff
@@ -248,7 +251,9 @@ def prepare_append(
         )
         frames_in = payload.get("video_frames")
         if isinstance(frames_in, list):
-            self._frame_queue.extend(frame for frame in frames_in if isinstance(frame, str) and frame)
+            group = [frame for frame in frames_in if isinstance(frame, str) and frame][:2]
```
Gaohan123, line 269: [P1] Bound the pending camera-frame queue

A frame group is appended for every audio append but only one group is consumed per approximately one-second model unit. Since the wire accepts up to two roughly 4 MB frames on every append, a fast or malformed client can grow this queue without bound. Cap pending groups/bytes or replace the current unit's pending frame instead of accumulating all groups.

tests/config/test_config_factory.py
```diff
@@ -1132,6 +1132,12 @@ def test_load_minicpmo_duplex_deploy_config(self):
         assert stages[1].yaml_extras["default_sampling_params"]["temperature"] == 0.8
         assert stages[1].yaml_extras["default_sampling_params"]["stop_token_ids"] == [6561]
         assert "codec_sampling_params" not in stages[1].yaml_engine_args
+        npu_stage2 = next(
```
Gaohan123, line 1135: [P2] Duplex graph setting contradicts the PR contract

The PR description says Code2Wav NPUGraph must be disabled because replay has no cancel hook, but this test locks the duplex deployment to graph-enabled. Either disable it as described or update the design and add an NPU mid-replay interruption-latency test proving graph replay does not defeat soft interrupt.

examples/online_serving/minicpmo/realtime_duplex_demo.py
```diff
@@ -176,13 +299,29 @@ def _ref_audio_data_url(path: str | None) -> str | None:
...
-    input_pcm16 = read_pcm16_wav(Path(args.input_wav))
+    output_dir = Path(args.output_dir)
+    try:
+        stream_writer = _StreamingOutputWriter(output_dir)
+    except PermissionError as exc:
+        raise SystemExit(f"cannot write --output-dir {output_dir}: {exc}") from exc
+
+    video_frames: list[str] = []
+    input_wav = args.input_wav
+    if args.input_video:
```
Gaohan123: [P2] External WAV does not actually override video audio

Even when input_wav is supplied, this branch unconditionally calls _extract_video_input(), which invokes load_audio on the video and derives frame count from that soundtrack. Silent video plus an external WAV therefore fails, and differing durations misalign frames. Split frame extraction from soundtrack extraction and use the external WAV duration when it is provided.

amy-why-3459:
@natureofnature @R2-Y PTAL

amy-why-3459:
@Gaohan123 Thanks for the review — this PR is now scoped only to the omni-duplex frame/audio pairing bug (title change on Aug 26). Status on the four comments:

External WAV override: fixed. --input-wav + --input-video extracts frames against the WAV duration and never calls load_audio on the video, so a silent clip no longer fails.
NPUGraph / soft-interrupt contract: dropped from this PR with the retitle. No longer in scope here.
Unbounded frame queue: the intended client now attaches frame k only on the append that closes unit k (duplex_unit_boundary_ms), so a well-behaved stream no longer accumulates groups. _frame_queue is still unbounded against a fast/malformed client — I’ll cap pending frames (replace the current unit’s pending group instead of extending) if you want that in this PR.
Vision failure vs reserved span: agreed this is still a hole. Stage0 now fails closed (no audio_chunk_idx advance, explicit reason) and serving reports buffering, but preprocess() still falls back to the original input_ids/input_embeds and we do not yet abort/rollback the reserved scheduler span or prove num_computed_tokens/KV stay put. I’d rather take span abort + that test as a follow-up than expand this pairing fix — let me know if you want it in this PR.
Thread 1 — Vision failure still advances the reserved span

Agreed, this path is still unsafe. A failed vision encode now returns success=False without advancing audio_chunk_idx, and the serving bridge treats prefill_success=False as buffering, but preprocess() still falls back to the original input_ids/input_embeds, so the reserved span can still enter KV without the matching audio. I have not added a num_computed_tokens/KV test. This PR is now only the frame-to-unit pairing fix; I’d like to do span abort/rollback + that test in a follow-up unless you want it here.

Thread 2 — Bound the pending camera-frame queue

The intended client no longer enqueues a group on every 200 ms append: RealtimeDuplexClient.stream_pcm16 sends frame k only on the append that closes unit k. That removes the accumulation for a correct client. _frame_queue is still uncapped, so a fast or malformed client can still grow it. I’ll replace the current unit’s pending frames instead of extending, and cap the queue, if you want that in this PR.

Thread 3 — Duplex graph setting contradicts the PR contract

This no longer applies. Soft-interrupt / Code2Wav NPUGraph was dropped when the PR was retitled to the frame-binding fix. That deploy assertion is not part of this change.

Thread 4 — External WAV does not actually override video audio

Fixed. _resolve_duplex_av_inputs now takes --input-wav as the audio source, derives frame count from that WAV duration, and skips _extract_video_soundtrack / load_audio on the video. Silent clip + external WAV no longer fails, and frames stay aligned to the WAV.

yenuo26 (approved):
LGTM

PR #1851, opened 2026-03-12
[Bugfix] Remove duplicated config keyword max batch size

tzhouam:
## Purpose
This PR migrates stage concurrency configuration from `runtime.max_batch_size` to `engine_args.max_num_seqs` across code, docs, examples, and test configs, as described in #695 .

The change standardizes concurrency control under `engine_args`, aligns stage config semantics with vLLM scheduler behavior, and removes legacy runtime-to-engine arg mapping.

- Replaced `runtime.max_batch_size` with `engine_args.max_num_seqs` in:
  - stage config docs and quickstart references
  - model/executor stage YAMLs
  - platform-specific stage configs (NPU/ROCm/XPU)
  - E2E/perf test stage configs
  - offline inference example READMEs
- Updated runtime code paths to read concurrency from `engine_args.max_num_seqs`:
  - default stage construction in `omni.py` and `async_omni.py`
  - stage worker batching path in `omni_stage.py`
- Removed legacy conversion logic from `load_stage_configs_from_yaml` that copied
  `runtime.max_batch_size` into `engine_args.max_num_seqs`.
- Renamed helper parameter in dynamic chunk-size computation from
  `max_batch_size` to `max_num_seqs` for consistency.

## Test Plan
Tested on qwen 3 omni online.
## Test Result
```bash
/app/rebase/.venv/lib/python3.12/site-packages/transformers/models/auto/image_processing_auto.py:647: FutureWarning: The image_processor_class argument is deprecated and will be removed in v4.42. Please use `slow_image_processor_class`, or `fast_image_processor_class` instead
  warnings.warn(
INFO 03-12 13:19:00 [logo.py:45]        █     █     █▄   ▄█       ▄▀▀▀▀▄ █▄   ▄█ █▄    █ ▀█▀ 
INFO 03-12 13:19:00 [logo.py:45]  ▄▄ ▄█ █     █     █ ▀▄▀ █  ▄▄▄  █    █ █ ▀▄▀ █ █ ▀▄  █  █  
INFO 03-12 13:19:00 [logo.py:45]   █▄█▀ █     █     █     █       █    █ █     █ █   ▀▄█  █  
INFO 03-12 13:19:00 [logo.py:45]    ▀▀  ▀▀▀▀▀ ▀▀▀▀▀ ▀     ▀        ▀▀▀▀  ▀     ▀ ▀     ▀ ▀▀▀ 
INFO 03-12 13:19:00 [logo.py:45] 
[...]

PR #6405, opened 2026-08-20
[bugfix][diffusion] Enable request-level batching via online entrypoint for diffusion models

wtomin:
## Summary

This PR makes `vllm serve --omni --max-num-seqs N` correctly control the default diffusion request batch size when `--diffusion-batch-size` is not explicitly provided (BTW, `--diffusion-batch-size` is not a valid CLI argument for `vllm serve`).

Previously, even if users started omni serving with:

```bash
vllm serve ... --omni --max-num-seqs N
```

the diffusion model could still run with `diffusion_batch_size=1`, so the documented request batching knob did not actually take effect for diffusion workloads.

While using `AsyncOmni(diffusion_batch_size=N)` is working correctly, this PR wants to enable diffusion batching currently for online entrypoint.

## Motivation

During diffusion stage initialization, `od_config.max_num_seqs` is overwritten by the resolved diffusion batch size:

```python
od_config.max_num_seqs = batch_size
```

Because of this, the effective diffusion batching capacity is determined by `diffusion_batch_size`.

Before this change, when `diffusion_batch_size` was not explicitly provided, it defaulted to `1`. This caused the following behavior:

- `--max-num-seqs N` was accepted by the server configuration.
- Diffusion stage startup later overwrote `od_config.max_num_seqs` with `diffusion_batch_size`.
- Since `diffusion_batch_size` was still `1`, diffusion request batching remained effectively limited to batch size `1`.
- Users had to pass an extra diffusion-specific batch size option even though `--max-num-seqs` is the documented request batching knob. 

## What Changed

This PR changes omni engine argument handling so that `diffusion_batch_size` is resolved from `max_num_seqs` when it is not explicitly set:

```python
explicit_batch = kwargs.pop("diffusion_batch_size", None)
max_num_seqs = kwargs.get("max_num_seqs") or 1
diffusion_batch_size = int(
    explicit_batch if explicit_batch is not None else max_num_seqs
)
```

The new behavior is:

- If `diffusion_batch_size` is explicitly provided, it still takes precedence.
- If `diffusion_batch_size` is not provided, it defaults to `max_num_seqs`.
- If neither value is provided, the fallback remains `1`.

## Effect After This Change

After this PR, users can configure diffusion request batching with the standard serving option:

```bash
vllm serve ... --omni --max-num-seqs 4
```

In this case, the diffusion model will use an effective batch size of `4`, unless `--diffusion-batch-size` is explicitly set to another value. An example output is:
```bash
INFO 08-20 08:58:55 [diffusion_pipeline_profiler.py:32] [DiffusionPipelineProfiler] Wan22Pipeline.text_encoder.forward took 0.087888s
INFO 08-20 08:58:55 [diffusion_pipeline_profiler.py:32] [DiffusionPipelineProfiler] Wan22Pipeline.text_encoder.forward took 0.087681s
INFO 08-20 08:58:56 [diffusion_pipeline_profiler.py:32] [DiffusionPipelineProfiler] Wan22Pipeline.text_encoder.forward took 0.087087s
INFO 08-20 08:58:56 [diffusion_pipeline_profiler.py:32] [DiffusionPipelineProfiler] Wan22Pipeline.text_encoder.forward took 0.087135s
100%|███████████████████████████████████████████████████████████████████████████████████████████| 50/50 [02:33<00:00,  3.07s/it]
INFO 08-20 09:01:29 [diffusion_pipeline_profiler.py:32] [DiffusionPipelineProfiler] Wan22Pipeline.diffuse took 153.415677s
INFO 08-20 09:01:29 [diffusion_pipeline_profiler.py:32] [DiffusionPipelineProfiler] Wan22Pipeline.diffuse took 153.415595s
INFO 08-20 09:01:45 [diffusion_pipeline_profiler.py:32] [DiffusionPipelineProfiler] Wan22Pipeline.vae.decode took 16.236813s
INFO 08-20 09:01:45 [diffusion_pipeline_profiler.py:32] [DiffusionPipelineProfiler] Wan22Pipeline.forward took 169.843718s
INFO 08-20 09:01:45 [diffusion_pipeline_profiler.py:32] [DiffusionPipelineProfiler] Wan22Pipeline.vae.decode took 16.306176s
INFO 08-20 09:01:45 [diffusion_pipeline_profiler.py:32] [DiffusionPipelineProfiler] Wan22Pipeline.forward took 169.917127s
(APIServer pid=543700) INFO 08-20 09:01:45 [diffusion_engine.py:578] [RequestBatch] admission wait done waiting=4 max_batch=4 waited_ms=0.0
```

`[RequestBatch] admission wait done waiting=4 max_batch=4 waited_ms=0.0` shows the effective batch size. Before this change, when the effective batch size is 1, the `Wan22Pipeline.diffuse` took around 37s.

This makes omni serving behavior match user expectations and avoids a silent mismatch where `--max-num-seqs` appears configured but diffusion execution still runs with batch size `1`.

## Backward Compatibility

This change preserves existing behavior for users who explicitly set `diffusion_batch_size`.

The only behavior change is for the implicit/default case: diffusion serving now follows the documented `--max-num-seqs` batching setting instead of silently falling back to `1`.

### DIFF
--- a/vllm_omni/entrypoints/omni_base.py
+++ b/vllm_omni/entrypoints/omni_base.py
@@ -177,7 +177,12 @@ class OmniBase(PDDisaggregationMixin):
         # override the deploy YAML's ``async_chunk: true`` default.
         async_chunk = kwargs.get("async_chunk")
         output_modalities = kwargs.pop("output_modalities", None)
-        diffusion_batch_size: int = kwargs.pop("diffusion_batch_size", 1)
+        # Stage init overwrites ``od_config.max_num_seqs`` with this value.
+        # ``vllm serve --omni --max-num-seqs N`` is the documented request-batch
+        # knob; use it when ``diffusion_batch_size`` is not passed explicitly.
+        explicit_batch = kwargs.pop("diffusion_batch_size", None)
+        max_num_seqs = kwargs.get("max_num_seqs") or 1
+        diffusion_batch_size = int(explicit_batch if explicit_batch is not None else max_num_seqs)
 
         if "log_requests" in kwargs:
             raise TypeError("`log_requests` has been removed in Omni/AsyncOmni. Use `log_stats`.")


wtomin:
I simply use `max_num_seqs` to overwrite `diffusion_batch_size`, but if `--diffusion-batch-size` is an acceptable CLI argument to `vllm serve`, I am fine with the following solution:
- 1. add `--diffusion-batch-size` to `vllm_omni/entrypoint/cli/serve.py`;
- 2. Enable request-level  batching by `vllm serve --omni --max-num-seqs 4 --diffusion-batch-size 4`

cc @knlnguyen1802

yJader:
I hit the same issue when testing Wan 2.2. It seems #5676 allows the legacy `diffusion_batch_size=1` default to overwrite `max_num_seqs`.

Since #1851 made `max_num_seqs` the scheduler concurrency setting, should we remove the internal diffusion_batch_size / batch_size handling and just use `max_num_seqs` everywhere? The legacy Python argument could remain as a deprecated alias.

cc  @LyxWxj

LyxWxj:
I investigated the usages of `max_num_seqs` and `diffusion_batch_size`.

I recommend removing `diffusion_batch_size` from the public/runtime API. It does not represent an independent execution concept. In the current implementation, it is primarily propagated through the Stage runtime and eventually assigned to `od_config.max_num_seqs`:

```python
od_config.max_num_seqs = batch_size
````

As a result, `max_num_seqs` is the parameter actually consumed by the diffusion scheduler, execution-mode selection, KV-cache profiling, NPU capacity configuration, transfer adapters, and model-specific input processors.

## Current usage of `diffusion_batch_size`

`diffusion_batch_size` is propagated through:

- [`omni_base.py`](https://github.com/vllm-project/vllm-omni/tree/main/vllm_omni/entrypoints/omni_base.py): lines 180, 205 — extracts from kwargs (default 1), passes to `AsyncOmniEngine`
- [`async_omni_engine.py`](https://github.com/vllm-project/vllm-omni/tree/main/vllm_omni/engine/async_omni_engine.py): lines 113, 123, 300 — constructor argument, saved as member, passed to `create_stage_runtime`
- [`stage_runtime.py`](https://github.com/vllm-project/vllm-omni/tree/main/vllm_omni/engine/stage_runtime.py): lines 90, 127, 136, 662, 673, 757, 774, 903, 966, 1034, 1098, 1120, 1137 — remote stage context field, constructor parameter, member variable, passed as `batch_size` when launching local Diffusion Stage, prints batch size, `DistStageRuntime` parameter, passed to parent class, builds and saves remote Replica context, passes to remote `StageDiffusionClient`, Runtime Factory parameter, passes to `DistStageRuntime`, passes to normal `StageRuntime`
- [`stage_init_utils.py`](https://github.com/vllm-project/vllm-omni/tree/main/vllm_omni/engine/stage_init_utils.py): line 1506 — sets `od_config.max_num_seqs = batch_size`
- [`stage_engine_startup.py`](https://github.com/vllm-project/vllm-omni/tree/main/vllm_omni/engine/stage_engine_startup.py): line 1561 — sets `od_config.max_num_seqs = batch_size`

It is therefore an alias for the Stage-level `max_num_seqs`, rather than an independent configuration.

## Current usage of `max_num_seqs`

`max_num_seqs` is the canonical Stage-level capacity used by:

### Configuration definition and merging

- [`stage_config.py`](https://github.com/vllm-project/vllm-omni/tree/main/vllm_omni/config/stage_config.py): lines 310, 1011, 1024, 1035 — per-stage `StageDeployConfig.max_num_seqs`; CLI/runtime override into engine args; compatibility with old `max_batch_size`; final engine args
- [`omni_config.py`](https://github.com/vllm-project/vllm-omni/tree/main/vllm_omni/config/omni_config.py): lines 187, 445, 460 — scheduler override type; `OmniStageSchedulerConfig`; validates `max_num_batched_tokens >= max_num_seqs`
- [`diffusion/data.py`](https://github.com/vllm-project/vllm-omni/tree/main/vllm_omni/diffusion/data.py): line 894 — `OmniDiffusionConfig.max_num_seqs`
- [`config_factory.py`](https://github.com/vllm-project/vllm-omni/tree/main/vllm_omni/config/config_factory.py): line 623 — default value 1

### Diffusion Engine and Scheduler

- [`diffusion_engine.py`](https://github.com/vllm-project/vllm-omni/tree/main/vllm_omni/diffusion/diffusion_engine.py): lines 118, 262, 344, 1027 — reads max sequences; checks whether model supports request-level batching; prints final value; KV profile batch size
- [`base_scheduler.py`](https://github.com/vllm-project/vllm-omni/tree/main/vllm_omni/diffusion/sched/base_scheduler.py): line 74 — sets `max_num_running_reqs`
- [`platform.py`](https://github.com/vllm-project/vllm-omni/tree/main/vllm_omni/platforms/npu/platform.py): line 100 — sets NPU MC2 token capacity

### Transfer and input processing

- [`chunk_transfer_adapter.py`](https://github.com/vllm-project/vllm-omni/tree/main/vllm_omni/distributed/omni_connectors/transfer_adapter/chunk_transfer_adapter.py): lines 45, 567 — records `scheduler_max_num_seqs`; limits running queue
- [`qwen3_tts.py`](https://github.com/vllm-project/vllm-omni/tree/main/vllm_omni/model_executor/stage_input_processors/qwen3_tts.py): line 136 — computes chunk based on scheduling capacity
- [`fish_speech.py`](https://github.com/vllm-project/vllm-omni/tree/main/vllm_omni/model_executor/stage_input_processors/fish_speech.py): line 68 — computes load based on scheduling capacity
- [`chunk_size_utils.py`](https://github.com/vllm-project/vllm-omni/tree/main/vllm_omni/model_executor/stage_input_processors/chunk_size_utils.py): line 18 — dynamic chunk computation

The following files also use the semantics of `max_num_seqs`:

- [`gpu_model_runner.py`](https://github.com/vllm-project/vllm-omni/tree/main/vllm_omni/worker/gpu_model_runner.py)
- [`gpu_generation_model_runner.py`](https://github.com/vllm-project/vllm-omni/tree/main/vllm_omni/worker/gpu_generation_model_runner.py)
- [`gpu_ar_model_runner.py`](https://github.com/vllm-project/vllm-omni/tree/main/vllm_omni/worker/gpu_ar_model_runner.py)
- [`npu_model_runner.py`](https://github.com/vllm-project/vllm-omni/tree/main/vllm_omni/platforms/npu/worker/npu_model_runner.py)
- [`npu_generation_model_runner.py`](https://github.com/vllm-project/vllm-omni/tree/main/vllm_omni/platforms/npu/worker/npu_generation_model_runner.py)
- [`omni_generation_scheduler.py`](https://github.com/vllm-project/vllm-omni/tree/main/vllm_omni/core/sched/omni_generation_scheduler.py)
- [`omni_scheduler_mixin.py`](https://github.com/vllm-project/vllm-omni/tree/main/vllm_omni/core/sched/omni_scheduler_mixin.py)
- [`glm_tts_dit_wrapper.py`](https://github.com/vllm-project/vllm-omni/tree/main/vllm_omni/model_executor/models/glm_tts/glm_tts_dit_wrapper.py)
- [`voxcpm2_talker.py`](https://github.com/vllm-project/vllm-omni/tree/main/vllm_omni/model_executor/models/voxcpm2/voxcpm2_talker.py)
- [`modeling_moss_tts_codec.py`](https://github.com/vllm-project/vllm-omni/tree/main/vllm_omni/model_executor/models/moss_tts/modeling_moss_tts_codec.py)
- [`nemotron_voicechat`](https://github.com/vllm-project/vllm-omni/tree/main/vllm_omni/model_executor/models/nemotron_voicechat)
- [`minimax_music3`](https://github.com/vllm-project/vllm-omni/tree/main/vllm_omni/model_executor/models/minimax_music3)
- [`runner.py`](https://github.com/vllm-project/vllm-omni/tree/main/vllm_omni/experimental/ar_diffusion/runner.py)

## Proposed change

I suggest using `max_num_seqs` as the single source of truth throughout the full path:

```
CLI / deploy config
    -> per-stage engine args
    -> OmniDiffusionConfig.max_num_seqs
    -> scheduler and model runner
```

Specifically:

1. Remove the redundant `diffusion_batch_size` argument from `OmniBase`, `AsyncOmniEngine`, and `StageRuntime`.
2. Remove the two assignments that overwrite `od_config.max_num_seqs` from a separate `batch_size` argument.
3. Let each Stage use its resolved `stage_cfg.engine_args.max_num_seqs`.
4. Update the related tests and benchmark scripts.
5. If backward compatibility is required, keep `diffusion_batch_size` temporarily as a deprecated alias for `max_num_seqs`, emit a warning, and remove it in a later release.

This would avoid the current inconsistency where `--max-num-seqs 4` is accepted by the CLI but the Stage runtime still defaults to `diffusion_batch_size=1`, causing the final diffusion engine configuration to become `max_num_seqs=1`.

If maintainers agree with this direction, I can prepare a separate follow-up PR for the cleanup. I do not think it should block this PR.

**Comment by wtomin** [comment]
> I hit the same issue when testing Wan 2.2. It seems #5676 allows the legacy `diffusion_batch_size=1` default to overwrite `max_num_seqs`.

@yJader  Thanks for your comments. I believe #5676 worked fine with `diffusion_batch_size>1`, because in its test script `run_async_validation.py`, it passed `diffusion_batch_size` key value via:
```python
    omni = AsyncOmni(
        model=args.model,
        diffusion_batch_size=configured_batch_size, # configured_batch_size is defined via args.batch_size
        request_batch_max_wait_ms=250.0 if concurrency > 1 else 0.0,
        enforce_eager=True,
        dtype="bfloat16",
        boundary_ratio=0.875,
        flow_shift=5.0,
        log_stats=True,
    )
```

The major problem this PR tries to solve is when enabling request-level batching in online serving mode, `diffusion_batch_size` is always its default value (`1`), therefore,  request-level batching cannot be truly enabled.

As for whether we should remove `diffusion_batch_size`, I think we should ask @SamitHuang @knlnguyen1802 @zhtmike from the verl-omni side . I am wondering when request-level batching is enabled, is it possible that `diffusion_batch_size` and `max_num_seqs` are not equal?  For example, `max_num_seqs=8`, but `diffusion_batch_size` maybe smaller than 8 in order to get better performance?

SamitHuang:
I also noticed this bug and set 

```python
diffusion_batch_size = int(
    explicit_batch if explicit_batch is not None else max_num_seqs
)
```

in verl-omni in https://github.com/verl-project/verl-omni/pull/408. But this PR is still worth merging for root cause fixing

vllm_omni/entrypoints/omni_base.py
```diff
@@ -177,7 +177,12 @@ def __init__(
         # override the deploy YAML's ``async_chunk: true`` default.
         async_chunk = kwargs.get("async_chunk")
         output_modalities = kwargs.pop("output_modalities", None)
-        diffusion_batch_size: int = kwargs.pop("diffusion_batch_size", 1)
+        # Stage init overwrites ``od_config.max_num_seqs`` with this value.
```
linyueqian, line 180: [blocking] This comment is false on current `main`. #6525 removed the `batch_size -> od_config.max_num_seqs` overwrite and made `--max-num-seqs` flow through stage runtime overrides into the diffusion scheduler directly. The legacy value changed here is only logged by the subprocess and inline clients, so this fallback cannot increase request concurrency. The intended fix has already shipped; please close this PR as obsolete rather than rebasing an ineffective assignment.

linyueqian (changes requested):
Reviewed at `4d003faa`. Current `main` already fixes the original problem through the scheduler path added in #6525: `--max-num-seqs` reaches per-stage runtime overrides and `od_config.max_num_seqs`, which is the value the diffusion scheduler uses for request concurrency.

This PR only changes the legacy `diffusion_batch_size` value passed through `AsyncOmniEngine` and `StageRuntime`. The subprocess client logs it, and the inline client stores it only for logging. It never changes scheduler capacity. The new comment therefore contradicts the current implementation and its tests.

Please close this PR as obsolete rather than rebasing it. If there is a separate intended use for the legacy client batch value, that should be demonstrated through a current consumer and a focused test.

Validation: static review of the changed file plus the current merged tree and diffusion client/runtime call path. No PR code was executed locally.

PR #6406, opened 2026-08-20
[Bugfix][MiniCPM-o] Fix async-chunk snapshot replacement and prompt cleanup

natureofnature:
## Purpose

This is the second correctness-only split from #5102, following the resumable cleanup fix in #6360. It fixes MiniCPM-o 4.5 async-chunk snapshot replacement and explicit prompt-replacement cleanup without including the OmniInteract benchmark client, Nightly workflow, production YAML changes, final-input lifecycle work, force-listen policy, or automatic context-capacity rollover.

### What broke

1. Code2Wav runtime metadata was merged across chunks. A previous terminal audio payload could remain in the model buffer and be processed again by a later control-only boundary.
2. When a producer explicitly replaced a streaming prompt at a new assistant turn, the scheduler did not release the old KV/encoder state or clear the connector watermark. The replacement could therefore inherit cache ownership and chunk progress from the previous prompt.

### Reproduction

The targeted tests exercise both failures directly:

- send a terminal Code2Wav snapshot followed by an empty/control-only replacement and verify that the old audio is not replayed;
- explicitly replace a running or computed Talker prompt and verify re-admission, one-time KV/encoder release, stale-output fencing, and connector-watermark reset.

### Root cause

The Code2Wav path had only incremental merge semantics for runtime additional information. Separately, an explicit async prompt replacement updated the request payload without first going through the scheduler's cache-release and connector-reset path.

### Fix

- Add an explicit producer marker and a MiniCPM Code2Wav model opt-in for snapshot replacement. Unmarked generation/diffusion payloads retain their existing merge behavior.
- Emit a scheduler-ready empty replacement snapshot for duplicate terminal boundaries so old terminal audio cannot be replayed, while retaining sibling fields such as `codes.ref` when 1-D audio is represented by prompt placeholders.
- Requeue an explicitly replaced running prompt through normal scheduler admission, release its old KV/encoder state exactly once, mark in-flight output stale once, and reset the connector watermark.
- Pin the NPU runner's inherited snapshot-replacement contract with an Ascend test; `OmniNPUModelRunner` already derives from `OmniGPUModelRunner` and explicitly delegates Omni state updates to it.

### Scope decision

This revision deliberately removes the automatic Talker context-capacity rollover that was previously in this PR. We do not yet have a targeted long single-Talker-turn E2E showing that automatic replacement is required, and replacing an accumulated prompt is a model-semantics decision rather than a safe transport-only fix. If the long-video E2E demonstrates a real context-limit failure, that policy can return in a separate PR with model-quality and continuity evidence. Explicit turn-boundary replacement and its cache/watermark cleanup remain in this PR.

## Test Plan

**vLLM Version:** target pin `0.27.0`; compatibility unit-test container reports vLLM-Omni `0.25.0`

**vLLM-Omni Commit:** `d5e206cd6b4cf5744fff19ef6a5a77d07f1d3d9e`

```bash
python3 -m pytest -q \
  tests/core/sched/test_omni_ar_scheduler_streaming.py::test_explicit_streaming_payload_replaces_placeholder_prompt \
  tests/core/sched/test_omni_ar_scheduler_streaming.py::test_explicit_model_intermediate_prompt_replacement_releases_cache_and_watermark \
  tests/core/sched/test_omni_ar_scheduler_streaming.py::test_ready_async_chunk_prompt_replacement_releases_stale_kv_once \
  tests/distributed/omni_connectors/test_chunk_transfer_adapter.py::test_load_poll_ar_requeues_explicitly_replaced_running_prompt \
  tests/distributed/omni_connectors/test_chunk_transfer_adapter.py::test_load_poll_generation_segment_marker_replaces_previous_chunk \
  tests/distributed/omni_connectors/test_chunk_transfer_adapter.py::test_load_poll_generation_empty_replacement_snapshot_is_ready \
  tests/distributed/omni_connectors/test_chunk_transfer_adapter.py::test_load_poll_generation_without_snapshot_marker_keeps_incremental_state \
  tests/model_executor/models/minicpmo_4_5/test_code2wav_batching.py::test_empty_final_ignores_generation_scheduler_placeholder_token \
  tests/model_executor/stage_input_processors/test_minicpmo_4_5_async_chunk.py::test_duplex_turn_end_closes_epoch_and_next_turn_restarts_sequence \
  tests/model_executor/stage_input_processors/test_minicpmo_4_5_omni.py::test_native_duplex_continuation_appends_only_new_talker_condition \
  tests/worker/test_omni_gpu_model_runner.py::test_update_additional_information_deserializes_new_request_payload \
  tests/worker/test_omni_gpu_model_runner.py::test_streaming_new_request_marker_replaces_terminal_chunk_snapshot \
  tests/worker/test_omni_gpu_model_runner.py::test_cached_empty_marker_replaces_terminal_chunk_snapshot

python3 -m pytest -q \
  tests/core/sched/test_omni_ar_scheduler_streaming.py \
  tests/model_executor/stage_input_processors/test_minicpmo_4_5_omni.py \
  tests/worker/test_omni_gpu_model_runner.py

python3 -m pytest -q tests/model_executor/models/minicpmo_4_5/test_talker_batching.py

git diff --name-only upstream/main...HEAD -- '*.py' | xargs python3 -m ruff check
git diff --name-only upstream/main...HEAD -- '*.py' | xargs python3 -m ruff format --check
git diff --check upstream/main...HEAD
```

## Test Result

- Targeted regression: `13 passed, 17 warnings in 1.79s`
- Reviewer stale-seed regression: `2 passed, 17 warnings in 0.11s`
- Affected related suite: `181 passed, 17 warnings in 2.92s`
- Talker chunk-budget regression: `24 passed, 17 warnings in 0.33s`
- NPU snapshot inheritance contract: added under `tests/platforms/npu/` for Ascend CI
- Ruff: `All checks passed`; all changed Python files are formatted
- `compileall` and `git diff --check`: passed

The snapshot replacement path requires both an explicit payload marker and the MiniCPM Code2Wav model opt-in. Explicit prompt replacement is producer-marked; there is no automatic context-limit policy in this PR.


natureofnature:
@R2-Y PTAL

amy-why-3459:
Thanks — this is the right second split from #5102. Talker capacity
rollover (min of model/TTS len, generation reserve 26, requeue through
admission, one-shot KV/encoder free + watermark reset) looks correct
and the AR/adapter tests pin it. Snapshot replace as an explicit
producer marker plus the Code2Wav opt-in is the right default so
other generation/diffusion paths keep merging.

Request changes on the Code2Wav replay path.

The duplicate turn_end payload is only

    OmniPayloadStruct(
        meta=_MiniCPMO45MetaStruct(replace_runtime_additional_information=True),
        request_id=request_id,
    )

`finished` / `is_segment_finished` are unset. Generation poll then
increments `get_req_chunk` and hits

    if not has_new_ids and not payload_finished:
        return False

so the empty snapshot is written onto `request.additional_information`
but never marked ready. The runner buffer is only replaced when that
request is actually scheduled into `_update_additional_information`.
If Code2Wav is still running and this step is scheduled without a
ready update, `forward()` still sees the previous terminal
`codes.audio` — the replay this PR claims to fix.

The producer test only checks the marker exists; the runner test
injects a `scheduled_new_reqs` snapshot. Please mark the duplicate
boundary ready (`is_segment_finished=True`, or treat the replace
flag as ready) and add a generation-poll test that an empty
duplicate snapshot becomes ready and leaves no leftover audio in
the runner buffer. Keeping `codes.ref` under the 1-D placeholder
is fine.

Should-fix:
- NPU generation runner does not inherit
  `_replace_intermediate_buffer`; snapshot replace is CUDA-only
  unless you lift it to a shared mixin.
- Confirm `_free_request_blocks` exists on the vLLM pin (the PR
  says 0.25.0; recent schedulers have it, older ones only have
  `_free_blocks`).
- Prefix-cache / async-scheduling remain follow-up, as noted.

Happy to re-review once the empty snapshot is actually consumed.

natureofnature:
Self-review and reviewer update for `c04fc9fc`:

- Fixed the blocking replay path: a duplicate native-duplex `turn_end` now emits an empty replacement snapshot with `is_segment_finished=True`, so generation poll marks it ready and the runner consumes it instead of retaining the previous terminal `codes.audio`.
- Added regressions for the producer marker, generation-poll readiness, and both scheduled-new and scheduled-cached runner replacement. The cached-request test specifically verifies that the old terminal audio is removed.
- Confirmed the NPU path shares this implementation: `OmniNPUModelRunner` derives from `OmniGPUModelRunner`, and its `_update_states()` explicitly delegates to `OmniGPUModelRunner._update_states()`. Added an Ascend contract test that exercises the inherited snapshot replacement.
- Corrected the documented target vLLM version to `0.27.0` and verified that its scheduler provides `_free_request_blocks()` with deferred in-flight block release. The `0.25.0` value referred only to the compatibility unit-test container.

Validation after the fix:

- targeted regression: `13 passed, 16 warnings in 1.92s`
- affected related suite: `149 passed, 21 deselected, 16 warnings in 2.65s`



I also rechecked the scope: snapshot replacement remains gated by the explicit producer marker plus the MiniCPM Code2Wav model opt-in; unmarked generation/diffusion models retain merge semantics. @amy-why-3459 PTAL when convenient.

natureofnature:
@codex review

natureofnature:
@codex review

natureofnature:
@codex review

natureofnature:
@tzhouam PTAL

natureofnature:
@Gaohan123 PTAL

amy-why-3459:
Thanks for the follow-up. Rechecked d5e206cd.

The previous blocking replay path is fixed: a duplicate native-duplex
turn_end now emits an empty replacement with is_segment_finished=True,
generation poll marks it ready, and both scheduled-new and
scheduled-cached runner updates drop the previous terminal codes.audio.
The generation scheduler already forwards request.additional_information
onto cached reqs, so a still-running Code2Wav step can consume the
snapshot. NPU was my miss — OmniNPUModelRunner inherits
OmniGPUModelRunner and NPUGenerationModelRunner goes through
_preprocess() → _update_additional_information. The 0.27.0
_free_request_blocks pin is also fine.

Approve with comments.

Should-fix / follow-up:
- _clear_chunk_ready still drops replaced_streaming_prompt_ids only on
  scheduled_new_reqs. Mixin now consumes the marker, so this is leftover
  backup; please also clear it on cached admission or document that mixin
  is the only consumer.
- _update_request_as_session still does num_stale += in_flight and then
  _replace_streaming_session assigns over it. Harmless after d5e206cd,
  but the += is dead on the replace path and will double-count again if
  someone reverts the assign.
- Automatic Talker capacity rollover is correctly out of this PR. Please
  keep that as a separate #5102 follow-up if the long-turn E2E actually
  hits TTS max_position_embeddings; this split should not be read as
  having fixed that overflow.

Snapshot replace remaining opt-in (producer marker + Code2Wav flag) is
the right default. Happy to approve.

amy-why-3459 (approved):

Issue #6333, opened 2026-08-18
[Bug]: OmniVoice silently drops words on long text — no long-form chunking (upstream has audio_chunk_duration)


lazariv:
### Summary

When serving `k2-fsa/OmniVoice` with `--omni`, long inputs **silently lose words**: contiguous spans disappear from the middle of the generated speech while the response is a normal `200 OK` with full-length audio. There is no error and no warning.

Root cause is that vLLM-Omni's OmniVoice pipeline has **no long-form chunking**, while upstream `k2-fsa/OmniVoice`'s own `generate()` splits long text automatically (`audio_chunk_duration`, default 15 s). OmniVoice is non-autoregressive (no EOS, no learned duration model), so the total output length is committed *before* decoding; when the frame budget is tight the model compresses and drops content instead of extending.

### Environment

- Image `vllm/vllm-omni:v0.26.0`, single GPU (NVIDIA A40-12Q), Kubernetes
- Serve: `vllm serve k2-fsa/OmniVoice --host 0.0.0.0 --port 8000 --trust-remote-code --omni`
- Client: `POST /v1/audio/speech`, `response_format: wav`
- Code references below verified at tag **v0.26.0**

### Reproduction / measurements

Objective round-trip: synthesize with OmniVoice → transcribe with Whisper large-v3-turbo → align word sequences. ASR used **overlapping** 24 s/4 s windows so window boundaries cannot cause false gaps.

| input | coverage | observation |
|---|---|---|
| ≤ 93 words (EN) | **100 %** | clean |
| 118 words (EN) | **89.8 – 96.6 %** (seeds 42/7/123/none) | contiguous **2–10 word gaps mid-text**, position varies per seed |
| 67 words (DE) | 100 % | clean |

Example gap (seed 42), input `... Longer passages reveal how generation time scales with input length. Ultimately the best model depends on the specific product requirements. ...` → the span `time scales with input length ultimately the best` is absent from the audio.

**This is not truncation and not a length cap:**

| sentences | words | audio | expected | ratio |
|---|---|---|---|---|
| 12 | 118 | 48.3 s | 49.2 s | 0.98 |
| 16 | 150 | 60.8 s | 62.5 s | 0.97 |
| 20 | 191 | 79.6 s | 79.6 s | 1.00 |
| 24 | 236 | 96.6 s | 98.3 s | 0.98 |
| 32 | 309 | 128.0 s | 128.8 s | 0.99 |

[...]

Issue #4418, opened 2026-06-14
[Performance][Discussion]: FA2 varlen vs PyTorch SDPA varlen: performance tradeoff and numerical drift


asukaqaq-s:
Related to #4024 and #4041.
I ran a small operator-level benchmark to compare FA2 varlen and PyTorch SDPA varlen for packed varlen attention. The goal is to check whether SDPA varlen is a good backend candidate for dynamic batching.

### Setup
- A100 80GB, BF16
- heads=24, head_dim=128
- FA2: flash_attn_varlen_func
- SDPA: torch.nn.attention.varlen.varlen_attn
- real cu_seqlens, no padding
- metric: attention output max_abs, not E2E image diff

### main results 

<html>
<body>
<html><head>
            
case | lens | FA2 batch ms | SDPA batch ms | faster
-- | -- | -- | -- | --
edge_256 | 256 | 0.137 | 0.070 | SDPA 1.97x
qwen_256_single_320 | 320 | 0.127 | 0.068 | SDPA 1.88x
qwen_256_batch_320x2 | 320,320 | 0.119 | 0.067 | SDPA 1.78x
qwen_mixed_txt_256+512 | 299,1115 | 0.179 | 0.180 | tie
qwen_512+768 | 1088,2368 | 0.539 | 0.571 | FA2 1.06x
qwen_512+1024 | 1088,4160 | 1.310 | 1.441 | FA2 1.10x
bsz8_1024 | 4160 x8 | 9.040 | 10.015 | FA2 1.11x

</html>

Batch runtime comparison shows SDPA is faster for small-token cases, while FA2 becomes faster for larger and 1024-resolution cases.

<html>
<body>
<html><head>
            
case | lens | SDPA batch vs SDPA serial | FA2 batch vs FA2 serial
-- | -- | -- | --
edge_256 | 256 | 0.003906 | 0.000000
edge_319 | 319 | 0.003906 | 0.000000
qwen_256_single_320 | 320 | 0.003906 | 0.000000
edge_321 | 321 | 0.003906 | 0.000000
edge_384 | 384 | 0.001953 | 0.000000
qwen_256_batch_320x2 | 320,320 | 0.001953 | 0.000000
qwen_mixed_txt_256+512 | 299,1115 | 0.003906 | 0.000000
qwen_512+768 | 1088,2368 | 0.000000 | 0.000000
qwen_512+1024 | 1088,4160 | 0.000000 | 0.000000
[...]

PR #6408, opened 2026-08-20
[Feat][OmniVoice]Support Varlen Attn,  Request-Batch and Step-Execution

sphinxkkkbc:
***
## Purpose

Support request batching and step execution for OmniVoice by implementing model-specific helpers following the existing `execute_stepwise()` lifecycle.

Bit-exact parity with outputs generated before this PR is not guaranteed because the attention backend and packed execution layout have changed. WER results are listed below. The OmniVoice decoder is unchanged from `main`, and each request is decoded separately using exact-length codec tokens.

An earlier version of this PR implemented padded Decoder batching, but it was later reverted. With the packed varlen Generator, generated codec tokens are now kept at exact per-request lengths, so batching Decoder execution would require padding them again to the longest request. The current implementation therefore decodes completed requests individually using exact-length codec tokens. I have not benchmarked padded Decoder batching against per-request exact-length decoding, so it is unclear which approach is faster.

### 1. Add packed variable-length attention to the OmniVoice generator

The previous Generator layout stored conditional and unconditional inputs as separate rows. Since their sequence lengths differ, unconditional inputs had to be padded to the conditional sequence length. In a request batch, shorter
requests were also padded to the longest request, introducing substantial attention and Transformer overhead.

This PR changes Generator inputs to a packed, token-major 2D layout:

```text
[cond0, uncond0, cond1, uncond1, ...]
```

with tensor shapes:

```text
input_ids:     [total_seq_len, num_codebooks]
audio_mask:    [total_seq_len]
hidden_states: [total_seq_len, hidden_size]
logits:        [num_codebooks, total_seq_len, audio_vocab_size]
```

For `B` requests, CFG produces `2B` independent sequences. Their cumulative offsets are represented by `cu_seqs`, allowing PyTorch varlen attention to keep requests and cond/uncond branches mutually invisible while retaining full bidirectional attention within each sequence.

Position IDs are reset independently for every packed sequence:

```text
[0..cond0_len-1, 0..uncond0_len-1, 0..cond1_len-1, 0..uncond1_len-1, ...]
```

`_position_ids_from_cu_seqs()` derives these sequence-local positions directly from `cu_seqs`.

Per-request random generators are retained so one request does not consume or change another request's Gumbel sampling stream.

### 2. Make packed varlen attention compatible with CUDA Graphs

OmniVoice already supported CUDA Graphs, but the previous graph inputs and keys were based on padded `[batch, codebook, sequence]` tensors.

CUDA Graph keys are now: `(request_batch_size, token_bucket)`

Captured static inputs use:

```text
input_ids:  [token_bucket, num_codebooks]
audio_mask: [token_bucket]
cu_seqs:    [2 * request_batch_size + 2]
```

The `2B+2` `cu_seqs` layout represents:

- `2B` real cond/uncond sequences;
- the initial zero offset;
- one additional tail sequence used for Graph bucket padding.

In eager mode, the tail has zero length: `[..., real_total, real_total]`

In Graph mode, the last offset is changed to the selected bucket: `[..., real_total, token_bucket]`

This keeps the metadata shape static while isolating padded tail tokens from all real requests.

`max_q` and `max_k` use the fixed packed token bucket during graph capture, so the real maximum sequence length does not need to be part of the graph key.

Common request batch sizes derived from default config are pre-captured. Other batch sizes and oversized token shapes are captured lazily.

### 3. Support request-batch and step-execution lifecycles

Request-batch mode packs the `input_ids` and `audio_mask` values produced by `_prepare_request_input()` along the token dimension. `cond_lens` and `target_lens` are used to construct `cu_seqs` once before Transformer execution and to locate each request's target logits and generated codec tokens.

Step-execution stores each request's persistent latents as: `[request_seq_len, num_codebooks]`

`InputBatch.make_batch()` therefore concatenates active requests directly along dimension 0 without model-specific repadding. `denoise_step()` constructs the packed metadata, runs one varlen Generator step, and returns the same 2D layout so the Runner can scatter rows back to individual request states.

The standard lifecycle methods are used:

- `prepare_encode()` initializes request-local state and sampling metadata;
- `denoise_step()` executes one packed Generator step;
- `step_scheduler()` persists the request slice and advances its step;
- `post_decode()` decodes the completed request.

Generated codec tokens are maintained at exact per-request lengths. 

### Unit Tests

```bash
pytest -v \
  tests/model_executor/models/omnivoice/test_cuda_graph_generator.py \
  tests/model_executor/models/omnivoice/test_pipeline_batching.py \
  tests/model_executor/models/omnivoice/test_mask_dtype.py \
  tests/model_executor/models/omnivoice/test_fused_projection_load.py
```

The CUDA Graph tests continue the validation coverage from the earlier implementation for packed Generator. Additional tests were added for the varlen layout to verify dynamic `cu_seqs` updates across replays of the same graph key, and eager/Graph parity.

### E2E Tests

```bash
pytest -v \
  tests/e2e/online_serving/test_omnivoice_expansion.py \
  tests/e2e/online_serving/test_omnivoice_parity.py
```

The E2E tests verify:

- request-batch execution correctness test;
- step execution generates valid audio without an error response, batch step-execution correctness test;
- B=1 step execution produces the same seeded WAV bytes as B=1 request-mode execution.

### Perf and Acc
- **request-rate=inf**: batch vs per-req
- **request-rate=1**: batch vs step-execution

**Serve Command:**
```
vllm serve k2-fsa/OmniVoice \
  --omni \
  --port 8091 \
  --trust-remote-code \
  --max-num-seqs 8 \
  --request-batch-max-wait-ms 50 \
  --step-execution
```
Remove `--step-execution` for benchmarking request-batch.

**Benchmark Command:**

```bash
vllm bench serve --omni \
  --host 127.0.0.1 \
  --port 8091 \
  --model k2-fsa/OmniVoice \
  --backend openai-audio-speech \
  --endpoint /v1/audio/speech \
  --dataset-name seed-tts-text \
  --dataset-path /path/to/seedtts_testset \
  --seed-tts-locale en \
  --num-prompts 32 \
  --num-warmups 2 \
  --extra-body '{
    "language": "English",
    "seed": 42,
    "extra_params": {
      "num_inference_steps": 32,
      "guidance_scale": 2.0
    }
  }' \
  --max-concurrency 4 \
  --seed-tts-wer-eval \
  --request-rate inf \
  --percentile-metrics e2el,audio_rtf,audio_ttfp,audio_duration \
  --save-result \
  --result-dir ./omnivoice_results
```

**Environment:**

- **vLLM Version:** 0.28.0
- **vLLM-Omni Commit:** `c4197ffe163c23017f179bf2f766c67766aca249`

---

## Test Results

Unit Test: `1 failed, 302 passed`. The single failure is unrelated to this PR and is not addressed here.

```
FAILED tests/entrypoints/openai_api/test_serving_speech.py::TestTTSAsyncOffloading::test_prepare_speech_generation_awaits_voxtral_async - ValueError: Invalid speaker 'test'. Supported: bar
```

**E2E : `14 passed`**

#### 1. Batch vs Per-Request

**C=4**

| Metrics | C=4 Batch | C=4 per-req | ∆ |
|---|---:|---:|:---:|
| Benchmark duration (s) | 15.96 | 24.04 | -33.61% |
| Request throughput (req/s) | 2.00 | 1.33 | +50.38% |
| Mean E2EL (ms) | 1993.46 | 2852.22 | -30.11% |
| Median E2EL (ms) | 2257.81 | 2920.93 | -22.70% |
| P99 E2EL (ms) | 2290.35 | 3400.14 | -32.64% |
| Audio throughput (audio duration/s) | 7.35 | 4.88 | +50.61% |
| Total Token throughput (tok/s) | 26.12 | 17.35 | +50.55% |
| Mean AUDIO_RTF | 0.58 | 0.84 | -30.95% |
| Median AUDIO_RTF | 0.56 | 0.78 | -28.21% |
| P99 AUDIO_RTF | 0.95 | 1.57 | -39.49% |
| Mean AUDIO_TTFP (ms) | 1991.95 | 2851.77 | -30.15% |
| Median AUDIO_TTFP (ms) | 2256.20 | 2920.65 | -22.75% |
| P99 AUDIO_TTFP (ms) | 2288.25 | 3399.69 | -32.69% |

**C=8**

| Metrics | C=8 Batch | C=8 per-req | ∆ |
|---|---:|---:|:---:|
| Benchmark duration (s) | 14.75 | 24.44 | -39.65% |
| Request throughput (req/s) | 2.17 | 1.31 | +65.65% |
| Mean E2EL (ms) | 3579.96 | 5416.61 | -33.91% |
| Median E2EL (ms) | 3575.87 | 6025.53 | -40.65% |
| P99 E2EL (ms) | 4092.70 | 6495.25 | -36.99% |
| Audio throughput (audio duration/s) | 7.96 | 4.80 | +65.83% |
| Total Token throughput (tok/s) | 28.28 | 17.06 | +65.77% |
| Mean AUDIO_RTF | 1.08 | 1.60 | -32.50% |
| Median AUDIO_RTF | 1.02 | 1.54 | -33.77% |
| P99 AUDIO_RTF | 1.98 | 3.42 | -42.11% |
| Mean AUDIO_TTFP (ms) | 3577.48 | 5416.21 | -33.95% |
| Median AUDIO_TTFP (ms) | 3573.22 | 6025.31 | -40.70% |
| P99 AUDIO_TTFP (ms) | 4091.90 | 6494.96 | -37.00% |

#### 2. Batch vs Step-Execution (C=4, request-rate=1)

| Metrics | Step-Execution | Request-batch | ∆ |
|---|---:|---:|:---:|
| Benchmark duration (s) | 32.73 | 32.70 | -0.09% |
| Request throughput (req/s) | 0.98 | 0.98 | 0.00% |
| Mean E2EL (ms) | 981.70 | 943.41 | -3.90% |
| Median E2EL (ms) | 934.53 | 785.74 | -15.92% |
| P99 E2EL (ms) | 1764.82 | 1974.45 | +11.88% |
| Audio throughput (audio duration/s) | 3.59 | 3.59 | 0.00% |
| Total Token throughput (tok/s) | 12.74 | 12.75 | +0.08% |
| Mean AUDIO_RTF | 0.28 | 0.27 | -3.57% |
| Median AUDIO_RTF | 0.24 | 0.21 | -12.50% |
| P99 AUDIO_RTF | 0.84 | 0.98 | +16.67% |
| Mean AUDIO_TTFP (ms) | 981.21 | 942.79 | -3.92% |
| Median AUDIO_TTFP (ms) | 934.14 | 785.35 | -15.93% |
| P99 AUDIO_TTFP (ms) | 1764.36 | 1974.01 | +11.88% |

### Accuracy

All request passed for each benchmark with 32 requests. All benchmark result with `Median WER=0.0000`

| Mode | Mean WER |
|---|---:|
| Baseline (Original) | 0.0074 |
| Request-batch + varlen (C=4, inf) | 0.0074 |
| Request-batch + varlen (C=8, inf) | 0.0074 |
| Step-execution (C=4, rate=1) | 0.0048 |
| Request-batch (C=4, rate=1) | 0.0096 |

---

*(anything written below this line will be removed by GitHub Actions)*


sphinxkkkbc:
@linyueqian could you help review this PR? Thanks!

vllm_omni/entrypoints/omni_base.py
```diff
@@ -177,7 +177,10 @@ def __init__(
         # override the deploy YAML's ``async_chunk: true`` default.
         async_chunk = kwargs.get("async_chunk")
         output_modalities = kwargs.pop("output_modalities", None)
-        diffusion_batch_size: int = kwargs.pop("diffusion_batch_size", 1)
+        diffusion_batch_size = kwargs.pop(
+            "diffusion_batch_size",
+            kwargs.get("max_num_seqs", 1),
```
linyueqian: **[High]** This changes the diffusion batch default for every pipeline, not just OmniVoice, and the PR body does not mention it.

`diffusion_batch_size` is force-assigned onto `od_config.max_num_seqs` at `stage_init_utils.py:1444` and `stage_engine_startup.py:1561`, overwriting whatever the stage config resolved. With the old default of `1`, `--max-num-seqs N` was silently ignored by diffusion stages. Now it lands, which is what makes the `--max-num-seqs 8` e2e test in this PR work, and it does match `docs/user_guide/diffusion/execution_modes.md:232`.

The side effect: only 8 of 51 diffusion pipelines set `supports_request_batch = True`, and `diffusion_engine.py:229` raises at startup when `max_num_seqs > 1` without it. Serve commands that run today (serially) will fail to start after this. It also still clobbers a per-stage `engine_args.max_num_seqs` with the global CLI value.

Please split this into its own PR with a unit test in `tests/engine/test_async_omni_engine_stage_init.py`, or at minimum call it out in the body and in the execution-modes doc.

vllm_omni/diffusion/models/omnivoice/pipeline_omnivoice.py
```diff
@@ -274,105 +281,335 @@ def forward(self, req: DiffusionRequestBatch) -> DiffusionOutput:
...
+          {"text": "...", "ref_audio": (samples, sr), "ref_text": "...",
+           "lang": "...", "instruct": "..."}
+        """
+        prepared_requests: list[_PreparedOmniVoiceRequest] = []
+        for request in req.requests:
+            prompt = request.prompt if request.prompt else ""
+            extra = request.sampling_params.extra_args or {}
+            prepared = self._prepare_request_input(prompt, extra)
+            if isinstance(prepared, DiffusionOutput):
+                return prepared
```
linyueqian: **[High]** Returning a bare `DiffusionOutput` here now raises `RuntimeError` instead of surfacing the user error.

Declaring `supports_request_batch = True` routes OmniVoice through `execute_model_batch`, which calls `_normalize_pipeline_outputs(..., allow_single_output=False)` (`diffusion_model_runner.py:650`). That helper raises

```
RuntimeError: OmniVoicePipeline.forward returned a single DiffusionOutput;
request-batch forward must return list[DiffusionOutput].
```

at `diffusion_model_runner.py:84`, for every batch size including B=1. On `main` OmniVoice went through `execute_model` with `allow_single_output=True`, so the same prompt produced a clean per-request error.

Trigger: offline `Omni.generate` with two clips in `multi_modal_data["audio"]` (line 258 above), or a dict prompt with no text key (line 276).

Fix direction: allocate `outputs = [None] * len(req.requests)` up front, place the error `DiffusionOutput` in that request's slot, and keep the remaining requests running.

vllm_omni/diffusion/models/omnivoice/pipeline_omnivoice.py
```diff
@@ -274,105 +281,335 @@ def forward(self, req: DiffusionRequestBatch) -> DiffusionOutput:
...
+        audio_masks = torch.cat(audio_masks, dim=0)
+
+        return input_ids, audio_masks, cond_lens
+
+    def prepare_encode(self, state: StepRequestState) -> DiffusionRequestBatch:
+        prompt = state.prompt if state.prompt else ""
+        extra = state.sampling.extra_args or {}
+        prepared = self._prepare_request_input(prompt, extra)
+        if isinstance(prepared, DiffusionOutput):
+            return prepared
```
linyueqian: **[High]** In step mode this error return is discarded, and the malformed request takes down the whole step batch.

`diffusion_model_runner.py:714` calls `prepare_encode(state)` for its side effects only and drops the return value. When `_prepare_request_input` returns an error, `state.latents` is never assigned, so `InputBatch.make_batch` hits `_prepare_latents` and raises `ValueError("All requests must have 'latents' initialized.")` (`input_batch.py:347`). That happens at `execute_stepwise` line 776, outside the per-request `try` at line 810, so one bad request fails every concurrent request in the batch.

The annotation on line 371 also matches neither the interface (`interface.py:60` declares `-> StepRequestState`) nor the success path, which returns `None`. Suggest recording the error on the state and letting the runner emit it per request, then fixing the annotation.

vllm_omni/model_executor/models/omnivoice/omnivoice_generator.py
```diff
@@ -473,18 +492,16 @@ class _OmniVoiceCUDAGraphForward:
     # Default bucket count is 10; 16 gives modest headroom for edge cases
     # (seq_len > max bucket or non-CFG batch) without unbounded GPU growth.
     _MAX_LAZY_GRAPHS: int = 16
+    _DEFAULT_CAPTURE_BATCH_SIZES: tuple[int, ...] = (1, 2, 3, 4)
```
linyueqian: **[Medium]** The pre-warm set does not cover the concurrency this PR targets, so the hot path falls into exact-size lazy capture.

The graph key is now `(request_batch_size, packed_token_total)`, but this tuple is hard-coded to `(1, 2, 3, 4)` and the largest bucket is 1024 (`configs/omnivoice.py:88`).

Worked example with the parity-test prompt: `RuleDurationEstimator` gives `target_len` about 80, so the packed total per request is about 200 (`cond_len` about 120 plus `target_len` 80).

- B=4 gives about 800, bucket 1024, fine.
- B=8 gives about 1600, past every bucket. `_find_bucket` returns `None` and line 629 sets `bucket = seq_len` exactly.

So every distinct combination of request lengths becomes its own lazy capture of a 28-layer graph on the hot path, thrashing the 16-entry cache. B in {5..8} is never pre-warmed at all, and `tests/e2e/online_serving/test_omnivoice_expansion.py:34` runs this model with `--max-num-seqs 8`.

Fix direction: derive the capture batch sizes from `max_num_seqs`, and round oversized totals up to a capped ladder instead of capturing at the exact packed length.

vllm_omni/model_executor/models/omnivoice/omnivoice_generator.py
```diff
@@ -543,32 +540,25 @@ def _capture_for_key(
...
-        # memory from other vllm modules (unlike get_global_graph_pool which
-        # shares a single pool across all captured graphs and can cause memory
-        # aliasing when two graphs replay concurrently).
-        if self._pool_handle is None:
-            self._pool_handle = torch.cuda.graph_pool_handle()
-
         graph = torch.cuda.CUDAGraph()
         with torch.no_grad():
-            with torch.cuda.graph(graph, pool=self._pool_handle):
+            with torch.cuda.graph(graph, pool=current_platform.get_global_graph_pool()):
```
linyueqian: **[Medium]** This reverses a deliberate decision on `main`, and the class docstring above still describes the old behavior.

`main` used a per-instance `torch.cuda.graph_pool_handle()` with an explicit comment:

> Lazy-init per-instance pool handle: isolates OmniVoice CUDA Graph memory from other vllm modules (unlike get_global_graph_pool which shares a single pool across all captured graphs and can cause memory aliasing when two graphs replay concurrently).

That comment is gone, but the class docstring at lines 486 to 490 still says graphs "share a single per-instance pool handle, which isolates OmniVoice CUDA memory from other vllm modules". Please either restore the per-instance pool, or state why the aliasing concern no longer applies and update the docstring to match.

vllm_omni/model_executor/models/omnivoice/omnivoice_generator.py
```diff
@@ -281,57 +282,55 @@ def __init__(self, config: OmniVoiceConfig):
...
-        k = k.permute(0, 2, 1, 3)
-        v = v.permute(0, 2, 1, 3)
+            k = k.repeat_interleave(repeat_factor, dim=1)
+            v = v.repeat_interleave(repeat_factor, dim=1)
 
-        # Caller passes a float mask; materialize float form if a bool slips through.
-        sdpa_mask = attention_mask
-        if sdpa_mask is not None and sdpa_mask.dtype == torch.bool:
-            sdpa_mask = torch.zeros_like(attention_mask, dtype=q.dtype).masked_fill_(~attention_mask, float("-inf"))
+        q = q.to(torch.bfloat16)
```
linyueqian: **[Medium]** Attention silently drops to bf16 under a float32 deploy config.

`vllm_omni/deploy/omnivoice.yaml` pins `dtype: "float32"`. `_varlen_attn` needs fp16 or bf16, so the cast is necessary, but the literal `torch.bfloat16` here and the literal `torch.float32` on line 331 hard-code a precision change that the config says is not happening. Suggest `out.to(hidden_states.dtype)` on line 331 so a bf16-loaded model does not feed fp32 into `o_proj`, and a note in the model doc that attention runs in bf16 regardless of `dtype`.

Related question on the accuracy table: mean WER is 0.0074 at C=4 but 0.0178 at C=8. If attention is per-sequence isolated by `cu_seqs` and each request has its own generator, outputs should not depend on batch composition. Do you know whether that delta is GEMM non-determinism over the larger packed `M`? If so, `test_speech_auto_voice_batch`'s byte-exact assertion is fragile at higher concurrency and should be relaxed to an audio-similarity or WER check.

vllm_omni/model_executor/models/omnivoice/omnivoice_generator.py
```diff
@@ -281,57 +282,55 @@ def __init__(self, config: OmniVoiceConfig):
...
         self,
         hidden_states: torch.Tensor,
-        attention_mask: torch.Tensor | None = None,
+        cu_seqs: torch.Tensor,
         cos: torch.Tensor | None = None,
         sin: torch.Tensor | None = None,
     ) -> torch.Tensor:
-        batch_size, seq_len, _ = hidden_states.shape
+        seq_len, _ = hidden_states.shape
+        max_q = seq_len
```
linyueqian: **[Medium]** `max_q` is the packed total rather than the longest sequence, and position ids are recomputed per layer.

`max_q = seq_len` is the full packed length and is passed for both q and k on lines 327 and 328. For B=8 with sequences of about 120 tokens and a packed total of about 1600, that oversizes the kernel grid in the M dimension by roughly 13x, across 28 layers and 32 steps. `cond_lens` and `target_lens` are already available at the call site, so `max(max(cond_lens), max(target_lens))` costs nothing and can be threaded through alongside `cu_seqs`.

Separately, `_position_ids_from_cu_seqs` on line 292 runs inside all 28 attention layers on every step. Hoisting it into `_step_forward` / `_transformer_forward` and passing the result down is a free win and stays CUDA-graph safe (same shapes, same device ops).

vllm_omni/diffusion/models/omnivoice/pipeline_omnivoice.py
```diff
@@ -274,105 +281,335 @@ def forward(self, req: DiffusionRequestBatch) -> DiffusionOutput:
...
+            c_len = cond_lens[i]
+            t_len = target_lens[i]
+
+            # Extract logits for target region; upcast only the slices we actually consume.
+            request_start = sequence_offsets[i]
+            cond_end = request_start + c_len
+            uncond_start = cond_end
+
+            # Extract logits for target region; upcast only the slices we actually consume.
+            c_logits = batch_logits[:, cond_end - t_len : cond_end, :].unsqueeze(0).to(torch.float32)
```
linyueqian, line 510: **[Medium]** Lines 488 to 551 duplicate `omnivoice_generator.py:956-1016` nearly verbatim.

The CFG fuse, the `log_probs[..., mask_id] = -inf` mask, the layer penalty, the Gumbel position noise, the top-k select and the cond/uncond mirror write are copied line for line, comments included. The only differences are `guidance_scales[i]` vs `guidance_scale`, `generators[i]` vs `request_generator`, and where `sample_tokens` comes from. Any future correctness fix has to land in both places.

Extracting one `_unmask_one_request(...)` helper owned by the generator would also remove `denoise_step`'s reach into `generator._prepare_embeddings`, `_transformer_forward`, `_get_logits` and `_cuda_graph_fwd`, which are all private.

vllm_omni/diffusion/models/omnivoice/pipeline_omnivoice.py
```diff
@@ -274,105 +281,335 @@ def forward(self, req: DiffusionRequestBatch) -> DiffusionOutput:
...
+        state.extra["layer_ids"] = layer_ids
+        state.extra["generator"] = generator
+        state.extra["t_shift"] = t_shift
+        state.extra["cond_len"] = cond_len
+        state.extra["target_len"] = target_len
+        state.extra["audio_mask"] = audio_mask
+        state.extra["tokens"] = tokens
+
+    def denoise_step(self, input_batch: InputBatch, *, states: Sequence[StepRequestState] | None = None, **kwargs: Any):
+        use_cuda_graph = self.generator._cuda_graph_fwd is not None
```
linyueqian: **[Minor]** Two small things in this method.

1. This line has no `input_ids.is_cuda` guard, unlike the request-mode path at `omnivoice_generator.py:929`. Worth keeping the two consistent.
2. Line 450 reads `state.extra.get("guidance", self.guidance_scale)`, but `prepare_encode` sets `state.guidance` (line 420), not `state.extra["guidance"]`. The `.get` always falls through, so the per-request guidance plumbing is dead. Either write `state.extra["guidance"]` in `prepare_encode` or read `state.guidance` here.

vllm_omni/model_executor/models/omnivoice/omnivoice_generator.py
```diff
@@ -577,91 +567,92 @@ def _capture_for_key(
...
         entry["static_input_ids"].copy_(ids_in)
         entry["static_audio_mask"].copy_(mask_in)
-        if attn_in is not None and entry["static_attn_mask"] is not None:
-            entry["static_attn_mask"].copy_(attn_in)
+        entry["static_cu_seqs"].copy_(runtime_cu_seqs)
 
         entry["graph"].replay()
 
         output = entry["static_output"]
         if bucket is not None and bucket != seq_len:
```
linyueqian: **[Minor]** `bucket is not None` cannot be false: line 629 assigns `bucket = seq_len` when `_find_bucket` returns `None`, so the check is dead and only `bucket != seq_len` matters.

Also, `_lazy_graphs` is documented as LRU but behaves as FIFO: the hit path on line 639 never calls `move_to_end`, so a frequently used key is still evicted by `popitem(last=False)`.

vllm_omni/entrypoints/openai/serving_speech.py
```diff
@@ -3387,6 +3387,15 @@ async def _create_diffusion_speech(
                 if sampling_params_list[0].extra_args is None:
                     sampling_params_list[0].extra_args = {}
                 sampling_params_list[0].extra_args.update(extra)
+
+                sampling = sampling_params_list[0]
+
+                if "num_inference_steps" in extra:
+                    sampling.num_inference_steps = int(extra["num_inference_steps"])
+
+                if "guidance_scale" in extra:
```
linyueqian, line 3520: **[Minor]** This block is out of scope for OmniVoice and unvalidated.

OmniVoice ignores both fields: `prepare_encode` reads `self.num_step` and `self.guidance_scale` from the model config, never `state.sampling`. So this changes behavior only for other diffusion TTS pipelines, with no test in this PR.

Two details if it stays:

- `guidance_scale_provided = True` is redundant. `OmniDiffusionRequest.__post_init__` already sets it whenever `guidance_scale is not None` (`vllm_omni/diffusion/request.py:66`).
- `int(...)` and `float(...)` run on unvalidated client input. `extra_params: {"num_inference_steps": "abc"}` raises `ValueError` and surfaces as a 500 rather than a 400.

vllm_omni/model_executor/models/omnivoice/omnivoice_generator.py
```diff
@@ -21,7 +20,9 @@
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
+from torch.nn.attention.varlen import _varlen_attn
```
linyueqian: **[Minor]** `_varlen_attn` is a private torch API imported at module scope, so it becomes a hard load-time requirement for OmniVoice on every platform.

What is the minimum torch version, and is `torch.nn.attention.varlen` present on the npu, rocm and xpu images? A short note in the module docstring plus the minimum version in the model doc would help, since there is no other use of this API in the repo.

linyueqian (changes requested):
Thanks for the writeup, the packed varlen layout and the `2B+2` `cu_seqs` trick with the padded tail sequence are a clean way to keep the graph metadata static. Numbers on the request-batch path look convincing.

Three blocking items and a few smaller ones, all inline.

**Blocking**

1. `pipeline_omnivoice.py:584`: declaring `supports_request_batch = True` routes through `execute_model_batch`, where `allow_single_output=False`. Returning a bare `DiffusionOutput` on a per-request validation error now raises `RuntimeError` for every batch size including B=1, where `main` returned a clean user error.
2. `pipeline_omnivoice.py:376`: in step mode the same error return is discarded by the runner, `state.latents` stays `None`, and `_prepare_latents` raises outside the per-request `try`, failing every concurrent request in the batch.
3. `omni_base.py:182`: defaulting `diffusion_batch_size` to `max_num_seqs` changes startup behavior for all 51 diffusion pipelines. 43 of them lack `supports_request_batch` and will now fail to start under `--max-num-seqs > 1` where they previously ran serially. This is unrelated to the PR title and has no test.

**Test coverage**

`test_cuda_graph_generator.py` is the only `core_model` test, and its `_SyntheticGenerator` never calls `_varlen_attn`, so it covers the padding, slicing and replay bookkeeping but not the attention change itself. The e2e tests are `slow` plus `tts` plus L4, so they only run in the weekly sweep, which does pick up the new `test_omnivoice_parity.py` automatically through the marker-driven collection in `.buildkite/cuda/test-weekly.yml:93`. `test_omnivoice_parity.py` also sets `OMNIVOICE_CUDA_GRAPH=0` on both sides, so the CUDA-graph plus varlen combination, which is the riskiest part of this change, has no automated coverage at all. Could you add a `core_model` test that drives a small real `OmniVoiceGenerator` through `_varlen_attn` in both eager and graph mode, including a replay with a nonzero padded tail?

Reviewed at `7a096c39668eb3295c016f792a16d4b3ea21f577`. Static review only, no GPU run on my side.

vllm_omni/entrypoints/omni_base.py
sphinxkkkbc: This change was removed in d5f0bfb6ae982bd21f73d3e4756aad458202f637. #6405 is already opened, so we can support that PR.

Also, #6525 resolved the issue where `max_num_seqs` was being overridden to 1. Now we can configure this in deploy.yaml. I've set the default `max_num_seqs` to 8 in 3354dbf5f7922b58799744a85f4e45dacd2eba82.

vllm_omni/diffusion/models/omnivoice/pipeline_omnivoice.py
sphinxkkkbc: Resolved in d5f0bfb6ae982bd21f73d3e4756aad458202f637, added 
```
outputs = [None] * len(req.requests)
prepared_indices: list[int] = []
```
to handle this and preserve the original request indices to prevent slot mismatch

Also added regression test in `tests/model_executor/models/omnivoice/test_pipeline_batching.py` for changes in `pipeline_omnivoice.py`

vllm_omni/diffusion/models/omnivoice/pipeline_omnivoice.py
sphinxkkkbc: This was initially resolved in  d5f0bfb6ae982bd21f73d3e4756aad458202f637, but after merging with the latest main, I found that #5810 already fixed it a few days ago. I've reverted it in f78c2904ed97455c0f63aa47f5277153d0dee2f6 and now directly using the handling from main.

Annotation fixed.

vllm_omni/model_executor/models/omnivoice/omnivoice_generator.py
sphinxkkkbc: Resolved in `d5f0bfb6ae982bd21f73d3e4756aad458202f637`, now `_OmniVoiceCUDAGraphForward` will read `od_config` and derive buckets and batches to be captured. I configured the buckets in a ladder pattern (shown below), which properly activates the LRU cache. With `MAX_LAZY_GRAPH = 16`, lazy-captured buckets are rounded up to multiples of 128. In benchmarking (using the command described), most scenarios do not trigger lazy capture, except for three specific (batch, seq_len) combinations: (7, 1574), (8, 1511), and (7, 1511). I believe the current capture strategy is reasonable. 

This PR measures ~7 GiB peak memory after pre-capture, versus ~5 GiB on the main branch.

Long-text generation issues persist as noted in #6333. #6409 proposes a chunk-based solution with a per-request limit of ~15 seconds for ~375 tokens in a single forward pass, so capturing extremely long sequences is unnecessary.

```
B=1: [128, 192, 256, 320, 384, 448, 512]
B=2: [256, 320, 384, 448, 512, 640, 768]
B=3: [384, 448, 512, 640, 768, 896, 1024]
B=4: [512, 640, 768, 896, 1024]
B=5: [640, 768, 896, 1024]
B=6: [768, 896, 1024]
B=7: [896, 1024]
B=8: [1024]
```

Regression test added in `tests/model_executor/models/omnivoice/test_cuda_graph_generator.py` for LRU.

vllm_omni/model_executor/models/omnivoice/omnivoice_generator.py
sphinxkkkbc: Reverted in d5f0bfb6ae. This change was introduced while debugging an issue during testing, but there were two potential variables involved, and this change turned out to be one of them. After double-checking, I found no reason to keep it. I'll comment below with more details.

vllm_omni/model_executor/models/omnivoice/omnivoice_generator.py
sphinxkkkbc: The per-layer `_position_ids_from_cu_seqs` computation was removed in commit d5f0bfb6ae982bd21f73d3e4756aad458202f637. Position IDs are now computed once before entering the transformer blocks and shared by all layers.

Regarding `max_q = seq_len`: it was initially kept this way to avoid introducing additional runtime metadata into the CUDA Graph path. Preliminary correctness testing in commit 7a096c39668eb3295c016f792a16d4b3ea21f577 found no output difference.

Following your comment, I ran an additional controlled benchmark where the only changed variable was max_q: the exact longest sequence length versus the total packed sequence length. The outputs were identical (max_abs_diff = 0), peak memory usage was unchanged, and the measured latency difference stayed within approximately ±3%, which appears to be normal run-to-run variance. Based on these results, I kept `max_q = seq_len` to avoid adding extra metadata plumbing. I’m happy to revisit this if there is another case or backend where the tighter bound is expected to make a measurable difference.

<details>
  <summary>script</summary>

```python
"""Measure how the varlen-attention max sequence-length bound affects latency.

The Q/K/V tensors and cumulative sequence lengths stay identical for every
measurement. ``max_q`` and ``max_k`` always receive the same value; only that
shared maximum-sequence-length bound changes between paired measurements.
"""

import argparse
import math
import statistics
from collections.abc import Callable

import torch


_DTYPES = {
    "float16": torch.float16,
    "bfloat16": torch.bfloat16,
}

_DEFAULT_CAPTURE_BUCKETS = [128, 192, 256, 320, 384, 448, 512, 640, 768, 1024]


def _parse_int_list(value: str) -> list[int]:
    try:
        values = [int(item.strip()) for item in value.split(",") if item.strip()]
    except ValueError as exc:
        raise argparse.ArgumentTypeError(f"Expected comma-separated integers, got {value!r}") from exc
    if not values:
        raise argparse.ArgumentTypeError("Expected at least one integer")
    return values


def _load_varlen_attn() -> tuple[Callable, str]:
    try:
        from torch.nn.attention.varlen import varlen_attn

        return varlen_attn, "torch.nn.attention.varlen.varlen_attn"
    except ImportError as exc:
        raise SystemExit("This benchmark requires public torch.nn.attention.varlen.varlen_attn.") from exc


def _unwrap_output(result):
    return result[0] if isinstance(result, tuple) else result


def _time_cuda(call: Callable[[], torch.Tensor], *, warmup: int, iters: int, repeats: int) -> list[float]:
    for _ in range(warmup):
        call()
    torch.cuda.synchronize()

    samples_ms: list[float] = []
    for _ in range(repeats):
        start = torch.cuda.Event(enable_timing=True)
        end = torch.cuda.Event(enable_timing=True)
        start.record()
        for _ in range(iters):
            call()
        end.record()
        end.synchronize()
        samples_ms.append(start.elapsed_time(end) / iters)
    return samples_ms


def _peak_allocated_bytes(call: Callable[[], torch.Tensor]) -> int:
    torch.cuda.synchronize()
    before = torch.cuda.memory_allocated()
    torch.cuda.reset_peak_memory_stats()
    output = call()
    torch.cuda.synchronize()
    peak_delta = max(0, torch.cuda.max_memory_allocated() - before)
    del output
    return peak_delta


def _format_bytes(value: int) -> str:
    if value < 1024:
        return f"{value} B"
    if value < 1024**2:
        return f"{value / 1024:.1f} KiB"
    return f"{value / 1024**2:.1f} MiB"


def _default_max_seqlens(actual_max: int, total_tokens: int) -> list[int]:
    values = [actual_max]
    candidate = 1 << max(0, math.ceil(math.log2(max(1, actual_max))))
    while candidate <= total_tokens:
        values.append(candidate)
        candidate *= 2
    return sorted(set(values))


def _make_call(
    varlen_attn: Callable,
    q: torch.Tensor,
    k: torch.Tensor,
    v: torch.Tensor,
    cu_seqlens: torch.Tensor,
    max_q: int,
    max_k: int,
    *,
    causal: bool,
) -> Callable[[], torch.Tensor]:
    def call() -> torch.Tensor:
        return _unwrap_output(
            varlen_attn(
                q,
                k,
                v,
                cu_seq_q=cu_seqlens,
                cu_seq_k=cu_seqlens,
                max_q=max_q,
                max_k=max_k,
                window_size=(-1, 0) if causal else (-1, -1),
            )
        )

    return call


def _run_matrix(args, varlen_attn: Callable, *, device: torch.device, dtype: torch.dtype) -> None:
    batches = args.batches
    buckets = args.total_buckets
    if any(batch <= 0 for batch in batches):
        raise SystemExit(f"Every matrix batch size must be positive, got {batches}")
    if any(bucket <= 0 for bucket in buckets):
        raise SystemExit(f"Every total bucket must be positive, got {buckets}")

    print("Matrix semantics: total bucket tokens are distributed as evenly as possible across B logical sequences.")
    print("exact max_q=ceil(total_bucket/B); conservative max_q=total_bucket.")
    print("max_q and max_k always use the same value.")
    print()
    print(
        f"{'batch':>5}  {'bucket':>7}  {'exact q':>7}  {'exact ms':>10}  {'bucket ms':>10}  "
        f"{'slowdown':>8}  {'exact peak':>12}  {'bucket peak':>12}  {'max |diff|':>12}  {'match':>6}"
    )

    for batch in batches:
        for bucket in buckets:
            quotient, remainder = divmod(bucket, batch)
            lengths = [quotient + (index < remainder) for index in range(batch)]
            actual_max = max(lengths)

            q = torch.randn(bucket, args.heads, args.head_dim, device=device, dtype=dtype)
            k = torch.randn_like(q)
            v = torch.randn_like(q)
            cu_seqlens = torch.zeros(batch + 1, dtype=torch.int32, device=device)
            cu_seqlens[1:] = torch.tensor(lengths, dtype=torch.int32, device=device).cumsum(dim=0)

            exact_call = _make_call(
                varlen_attn,
                q,
                k,
                v,
                cu_seqlens,
                actual_max,
                actual_max,
                causal=args.causal,
            )
            bucket_call = _make_call(
                varlen_attn,
                q,
                k,
                v,
                cu_seqlens,
                bucket,
                bucket,
                causal=args.causal,
            )

            try:
                baseline_output = exact_call().detach().clone()
                exact_samples = _time_cuda(
                    exact_call,
                    warmup=args.warmup,
                    iters=args.iters,
                    repeats=args.repeats,
                )
                bucket_samples = _time_cuda(
                    bucket_call,
                    warmup=args.warmup,
                    iters=args.iters,
                    repeats=args.repeats,
                )
                bucket_output = bucket_call()
                torch.cuda.synchronize()
                exact_peak = _peak_allocated_bytes(exact_call)
                bucket_peak = _peak_allocated_bytes(bucket_call)
                max_abs_diff = float((bucket_output - baseline_output).abs().max().item())
                matches = torch.allclose(bucket_output, baseline_output, rtol=1e-3, atol=1e-3)
            except Exception as exc:  # noqa: BLE001 - report the failing matrix cell and continue
                print(
                    f"{batch:>5}  {bucket:>7}  {actual_max:>7}  ERROR: "
                    f"{type(exc).__name__}: {str(exc).splitlines()[0]}"
                )
                continue

            exact_ms = statistics.median(exact_samples)
            bucket_ms = statistics.median(bucket_samples)
            print(
                f"{batch:>5}  {bucket:>7}  {actual_max:>7}  {exact_ms:>10.4f}  {bucket_ms:>10.4f}  "
                f"{bucket_ms / exact_ms:>8.3f}  {_format_bytes(exact_peak):>12}  "
                f"{_format_bytes(bucket_peak):>12}  {max_abs_diff:>12.4g}  {str(matches):>6}"
            )


def main() -> None:
    parser = argparse.ArgumentParser(description=__doc__, formatter_class=argparse.RawDescriptionHelpFormatter)
    parser.add_argument(
        "--lengths",
        type=_parse_int_list,
        default=None,
        help="Logical sequence lengths used to construct cu_seqlens (default: 64,20,44).",
    )
    parser.add_argument(
        "--batch",
        type=int,
        default=None,
        help="Physical padded batch size. Use together with --bucket-size.",
    )
    parser.add_argument(
        "--bucket-size",
        type=int,
        default=None,
        help="Per-row padded sequence bucket. Generates Q/K/V with batch*bucket tokens.",
    )
    parser.add_argument(
        "--valid-lengths",
        type=_parse_int_list,
        default=None,
        help=(
            "Valid prefix length for every physical row in bucket mode. Each row becomes two logical documents: "
            "[valid_length, bucket_size-valid_length]. Defaults to bucket_size/2 for every row."
        ),
    )
    parser.add_argument(
        "--matrix",
        action="store_true",
        help="Sweep logical batch sizes and total-token capture buckets.",
    )
    parser.add_argument(
        "--batches",
        type=_parse_int_list,
        default=list(range(1, 9)),
        help="Logical batch sizes for --matrix (default: 1,2,3,4,5,6,7,8).",
    )
    parser.add_argument(
        "--total-buckets",
        type=_parse_int_list,
        default=_DEFAULT_CAPTURE_BUCKETS,
        help="Total-token buckets for --matrix (defaults to OmniVoice's current capture sizes).",
    )
    parser.add_argument(
        "--max-seqlens",
        type=_parse_int_list,
        default=None,
        help="max_q/max_k values to sweep. Defaults to powers of two from the real max through total tokens.",
    )
    parser.add_argument("--heads", type=int, default=16)
    parser.add_argument("--head-dim", type=int, default=128)
    parser.add_argument("--dtype", choices=sorted(_DTYPES), default="bfloat16")
    parser.add_argument("--warmup", type=int, default=50)
    parser.add_argument("--iters", type=int, default=500)
    parser.add_argument("--repeats", type=int, default=5)
    parser.add_argument("--causal", action="store_true")
    parser.add_argument("--seed", type=int, default=0)
    args = parser.parse_args()

    if not torch.cuda.is_available():
        raise SystemExit("CUDA is required for this benchmark.")
    bucket_mode = args.batch is not None or args.bucket_size is not None or args.valid_lengths is not None
    if bucket_mode:
        if args.lengths is not None:
            raise SystemExit("Use either --lengths or --batch/--bucket-size, not both")
        if args.batch is None or args.bucket_size is None:
            raise SystemExit("Bucket mode requires both --batch and --bucket-size")
        if args.batch <= 0 or args.bucket_size <= 0:
            raise SystemExit("--batch and --bucket-size must be positive")
        valid_lengths = args.valid_lengths or [args.bucket_size // 2] * args.batch
        if len(valid_lengths) != args.batch:
            raise SystemExit(
                f"--valid-lengths needs one entry per physical row: batch={args.batch}, got {valid_lengths}"
            )
        if any(length < 0 or length > args.bucket_size for length in valid_lengths):
            raise SystemExit(
                f"Every valid length must be in [0, {args.bucket_size}], got {valid_lengths}"
            )
        lengths = [segment for valid in valid_lengths for segment in (valid, args.bucket_size - valid)]
    else:
        valid_lengths = None
        lengths = args.lengths or _parse_int_list("64,20,44")

    if any(length < 0 for length in lengths):
        raise SystemExit(f"Sequence lengths must be non-negative, got {lengths}")
    if sum(lengths) <= 0:
        raise SystemExit("At least one sequence must be non-empty")
    if min(args.heads, args.head_dim, args.warmup, args.iters, args.repeats) <= 0:
        raise SystemExit("heads, head-dim, warmup, iters, and repeats must all be positive")

    varlen_attn, implementation = _load_varlen_attn()
    device = torch.device("cuda")
    dtype = _DTYPES[args.dtype]
    torch.manual_seed(args.seed)

    print(f"GPU             : {torch.cuda.get_device_name()}")
    print(f"torch           : {torch.__version__}")
    print(f"implementation  : {implementation}")
    print(f"dtype           : {dtype}")

    if args.matrix:
        if args.lengths is not None or bucket_mode:
            raise SystemExit("--matrix cannot be combined with --lengths, --batch, --bucket-size, or --valid-lengths")
        _run_matrix(args, varlen_attn, device=device, dtype=dtype)
        return

    actual_max = max(lengths)
    total_tokens = sum(lengths)
    requested_max_seqlens = args.max_seqlens or _default_max_seqlens(actual_max, total_tokens)
    max_seqlens = sorted({actual_max, *requested_max_seqlens})
    invalid = [value for value in max_seqlens if value < actual_max]
    if invalid:
        raise SystemExit(
            f"Every max-seqlen must be >= the real maximum {actual_max}; invalid values: {invalid}. "
            "A smaller value is not a valid performance comparison."
        )

    q = torch.randn(total_tokens, args.heads, args.head_dim, device=device, dtype=dtype)
    k = torch.randn_like(q)
    v = torch.randn_like(q)
    cu_seqlens = torch.zeros(len(lengths) + 1, dtype=torch.int32, device=device)
    cu_seqlens[1:] = torch.tensor(lengths, dtype=torch.int32, device=device).cumsum(dim=0)

    print(f"Q/K/V shape     : {(total_tokens, args.heads, args.head_dim)}")
    if bucket_mode:
        print(f"capture layout  : batch={args.batch}, bucket_size={args.bucket_size}")
        print(f"valid lengths   : {valid_lengths}")
    print(f"sequence lengths: {lengths}")
    print(f"cu_seqlens      : {cu_seqlens.cpu().tolist()}")
    print(f"real max / total: {actual_max} / {total_tokens}")
    print(f"sweep           : {max_seqlens}")
    print()

    baseline_output = _make_call(
        varlen_attn,
        q,
        k,
        v,
        cu_seqlens,
        actual_max,
        actual_max,
        causal=args.causal,
    )().detach().clone()
    torch.cuda.synchronize()
    rows: list[tuple[int, float, float, float, int, float, bool]] = []
    baseline_ms: float | None = None

    for max_seqlen in max_seqlens:
        call = _make_call(
            varlen_attn,
            q,
            k,
            v,
            cu_seqlens,
            max_seqlen,
            max_seqlen,
            causal=args.causal,
        )
        try:
            samples = _time_cuda(call, warmup=args.warmup, iters=args.iters, repeats=args.repeats)
            output = call()
            torch.cuda.synchronize()
            peak_bytes = _peak_allocated_bytes(call)
            max_abs_diff = float((output - baseline_output).abs().max().item())
            matches = torch.allclose(output, baseline_output, rtol=1e-3, atol=1e-3)
        except Exception as exc:  # noqa: BLE001 - keep the sweep useful when one bound is unsupported
            print(f"max={max_seqlen}: {type(exc).__name__}: {str(exc).splitlines()[0]}")
            continue

        median_ms = statistics.median(samples)
        if baseline_ms is None:
            baseline_ms = median_ms
        rows.append(
            (
                max_seqlen,
                median_ms,
                min(samples),
                max(samples),
                peak_bytes,
                max_abs_diff,
                matches,
            )
        )

    if not rows or baseline_ms is None:
        raise SystemExit("No benchmark case completed successfully")

    print(
        f"{'max_q':>8}  {'median ms':>10}  {'min ms':>10}  {'max ms':>10}  "
        f"{'vs base':>8}  {'peak alloc':>12}  {'max |diff|':>12}  {'match':>6}"
    )
    for max_seqlen, median_ms, min_ms, max_ms, peak_bytes, max_abs_diff, matches in rows:
        print(
            f"{max_seqlen:>8}  {median_ms:>10.4f}  {min_ms:>10.4f}  {max_ms:>10.4f}  "
            f"{median_ms / baseline_ms:>8.3f}  {_format_bytes(peak_bytes):>12}  "
            f"{max_abs_diff:>12.8g}  {str(matches):>6}"
        )


if __name__ == "__main__":
    main()
```

</details>

<details>
  <summary>command</summary>

```bash
python bench_varlen_max_q.py \
  --matrix \
  --heads 16 \
  --head-dim 128 \
  --warmup 100 \
  --iters 1000 \
  --repeats 7
```

</details>

vllm_omni/diffusion/models/omnivoice/pipeline_omnivoice.py
sphinxkkkbc: Resolved in d5f0bfb6ae982bd21f73d3e4756aad458202f637

vllm_omni/diffusion/models/omnivoice/pipeline_omnivoice.py
sphinxkkkbc: Both resolved in d5f0bfb6ae982bd21f73d3e4756aad458202f637, regression test for Q2 added in `tests/model_executor/models/omnivoice/test_pipeline_batching.py`

vllm_omni/model_executor/models/omnivoice/omnivoice_generator.py
sphinxkkkbc: Both issues have been fixed.

Detailed comments on the `_lazy_graphs` FIFO issue have been inlined in the bucket issue. This was an existing unresolved issue on main.

vllm_omni/entrypoints/openai/serving_speech.py
sphinxkkkbc: These two changes are primarily required for scheduler correctness:
1. `StepScheduler._get_total_steps()` reads `sampling_params.num_inference_steps`. If the value remains only in extra_args and is not promoted to the standard sampling field, the scheduler receives None and fails when evaluating int(None).

2. Request-batch compatibility is determined from the standard sampling fields, including `guidance_scale`. If the request value remains only in `extra_args`, requests with different guidance scales may receive the same compatibility key and be incorrectly admitted to the same request batch.

Regression coverage was added in `tests/entrypoints/openai_api/test_serving_speech.py`

vllm_omni/model_executor/models/omnivoice/omnivoice_generator.py
sphinxkkkbc: Resolved in 9432df3c3cbce15ade345fb6d4b66aa30f0754e7. Now using the in-tree class `diffusion.attention.layer.Attention` with an explicit `attn_mask` fallback when `varlen_attn` is not supported. 

As for the version question, the torch version I used is the one installed with vLLM(2.13), and the minimum version that supports `varlen_attn` is torch 2.10.

vllm_omni/model_executor/models/omnivoice/omnivoice_generator.py
sphinxkkkbc: The dtype issue was addressed in commit `d5f0bfb6ae982bd21f73d3e4756aad458202f637`: the attention output is converted back to `hidden_states.dtype` before `o_proj`.

You are also right that the original byte-exact batch determinism tests were fragile, and the initially reported passing results were misleading.

When `test_speech_auto_voice_batch` first appeared to hang after another test, I initially suspected two CUDA Graph-related issues:

1. The implementation was still using the per-instance `_pool_handle`, so a graph or its static buffers might have survived the preceding request and interfered with a later replay.
2. The pre-capture strategy covered only B=1, so a real B=2 execution required lazy capture on the request path. I suspected that the lazy-capture lifecycle or its interaction with the graph pool might be responsible.

I changed and tested both variables during the investigation. However, neither change actually resolved the underlying issue. The output difference could be reproduced with both the per-instance pool and the global pool, and it was still present when CUDA Graph was disabled. This ruled out both graph-pool reuse and lazy capture as the root cause.

I then found that the apparent hang was not a server or CUDA Graph deadlock. Pytest was spending a long time constructing and rendering an assertion diff for two unequal WAV byte strings of approximately 163 KB.

After identifying that issue, I revisited the earlier passing results and found another problem: concurrent submission did not reliably guarantee actual batched model execution. Depending on request-arrival timing, the requests could still be processed serially. This affected both request-batch and step-execution tests. In step execution, for example, the observed active batch sizes could be:

```
1, 2, 2, ..., 2, 1
```

This explained why some earlier runs appeared deterministic: they were using batch mode at the server level, but the model requests were effectively processed as B=1.

I then explicitly verified the model batch size and compared a true B=2 execution with an equivalent B=1 per-request execution. Requests within the same B=2 batch remained identical to each other when using the same prompt and seed. However, the B=2 output was not identical to the B=1 output, and a significant portion of the final codec tokens differed.

A step-by-step trace showed that the B=1 and B=2 token states remained identical through step 20. The first token divergence appeared at step 21, after which the iterative discrete unmasking process amplified the numerical difference over the remaining steps. This is consistent with batch-shape-dependent numerical drift in the packed transformer/attention computation. A related numerical-invariance issue is also discussed in #4418.

Based on these findings, tests comparing batched and per-request execution were changed from byte-exact determinism checks to batch correctness checks. They now verify that every concurrent request completes successfully and produces valid audio with the expected WAV format, sample rate, shape, and duration. Semantic correctness remains covered by the existing ASR/WER evaluation. B=1 request-mode versus B=1 step-execution parity remains a separate test because it does not compare different batch shapes.

I also reran the C=8 request-batch benchmark several times. The previously reported WER of `0.0178` was not reproduced; the latest observed range was 0.0074. This WER decrease is one thing that I haven't fully investigated yet, one posible direction is that `varlen_attn` been changed with existing Attention impl.

sphinxkkkbc:
Inline comments are replied. Perf number still hold.

Test coverage updates:

1. `test_omnivoice_parity.py` now verifies both graph and eager modes.
2. `test_cuda_graph_generator.py` now uses the small real `OmniVoiceGenerator` in all cases rather than the `_SyntheticGenerator`. The results and commands have also been updated with descriptions.

sphinxkkkbc:
@linyueqian  PTAL, thanks!

vllm_omni/model_executor/models/omnivoice/omnivoice_generator.py
```diff
@@ -409,11 +419,68 @@ def _precompute_rope(
...
+    cu_seqs: torch.Tensor,
+    seq_len: int,
+    *,
+    needs_sdpa_mask: bool,
+) -> AttentionMetadata:
+    """Build packed-varlen metadata and, when needed, an SDPA block mask."""
+    extra = {
+        "cu_seqlens_q": cu_seqs,
+        "cu_seqlens_k": cu_seqs,
+        "max_seqlen_q": seq_len,
```
linyueqian: [question] This is the one round-1 item that stands unchanged: max_seqlen_q/k is still the packed total rather than the longest sequence, so the kernel grid stays oversized in M for multi-request packs. I suspect that is now deliberate: the CUDA graphs are keyed on (batch, bucket) and a captured launch bakes max_seqlen in, so any per-batch max would invalidate graph reuse within a bucket. If that is the reasoning, a one-line comment here saying so would close the loop; if the eager path (use_packed_varlen without an active capture) could pass the true max from cond_lens while the captured path keeps the bucket total, that would recover the grid efficiency where it is free. Either answer is fine for approval.

linyueqian (commented):
Re-review of the delta since 7a096c39. Every item from my first round is resolved, most of them more thoroughly than asked.

The three blocking items: the global diffusion_batch_size default change is gone entirely, main's #6525 removed the max_num_seqs mapping and the OmniVoice default now lives in its own deploy yaml; the batch forward returns list[DiffusionOutput] with per-request errors slotted in place so one bad request no longer poisons the batch; and prepare_encode now raises OmniClientError into the runner's per-request try/except, which pops the state and converts it to a per-request failure with the DiT lockstep all-reduce intact.

The mediums: the hard-coded prewarm tuple became a config-driven triangular batch-aware capture plan with the plan logged at warmup; the per-instance graph pool handle is restored with its original isolation comment; attention output is cast back to hidden_states.dtype; position ids are computed once per forward from cu_seqs and passed down; and the duplicated sampling block is gone, the pipeline now calls the generator's shared _unmask_one_request and forward helpers. The private _varlen_attn import concern got the best possible answer: an in-tree attention op with a supports_multi_doc_packed_varlen capability probe and an SDPA fallback, which also settles the non-CUDA platform question from the earlier commit message. The serving_speech extra_args block now validates both fields with typed errors, and OmniVoice consumes num_inference_steps and guidance_scale from sampling, with the new regression tests covering the scheduler and extra_args passthrough. The guidance key mismatch and the dead bucket check plus FIFO-as-LRU are all fixed, move_to_end is on the hit path.

One question inline about the surviving max_seqlen choice, which I suspect is a deliberate consequence of graph capture; either a confirming comment or an eager-path refinement closes it. Per-PR CI has never run on this branch, so I have added the ready label; approval follows green lanes.

vllm_omni/model_executor/models/omnivoice/omnivoice_generator.py
sphinxkkkbc: Added a clarifying comment and exact longest-segment max_seqlen support for the eager path in a502950bc153e2f7867c3e3db40782ae1e556e54.

Local benchmarks showed no measurable performance difference between using the exact longest segment and the packed total; the observed variation was within normal run-to-run noise. The CUDA Graph path intentionally keeps the fixed token-bucket maximum. This avoids requiring the exact runtime sequence composition during pre-capture and prevents the graph cache from growing with request-specific shapes.

The latest commit's CI is all green except for one CosyVoice3 Test in amd-ci; that failure is unrelated to this PR.

linyueqian (approved):
The max_seqlen question is answered exactly the way I hoped: the eager path now passes the exact bound while the CUDA graph path keeps the fixed bucket maximum with a comment explaining why, plus a unit test for the new parameter. I verified the bound is right at the packing level, cond_ids concatenates text, optional reference audio and the target region, so cond_len always dominates the uncond segment and max(cond_lens) is the true longest packed sequence.

Lanes at this head: general, Intel and NPU all green, twice in a row across e2285aa8 and a502950b. AMD fails only the CosyVoice3-TTS E2E step, both runs, with exit 124, a step timeout during a slow compiled server bring-up with no assertion anywhere; the only surface this PR shares with that pipeline is the serving_speech extra_args plumbing, and the CosyVoice3 e2e sends none of those fields, so the path is provably inert there. Worth an infra look at that step's time budget, separately from this PR.

Approving. Every round-1 item resolved, the delta re-reviews found nothing, and the in-tree attention op with the capability probe turned the riskiest part of the original design into the cleanest.

linyueqian:
Heads up: this went conflicting against `main` a few minutes ago, so it needs a rebase or a merge from main to land.

Nothing else is outstanding on the review side. This is approved at `a502950b`, the general, Intel, and NPU lanes are green there, and the one AMD red is the `mi300_1: CosyVoice3-TTS E2E Test` step timeout (exit 124) that has now been attributed as not-this-PR at three consecutive heads: the 20 minute step budget is consumed by roughly six and a half minutes of compiled bring-up plus a single inference request that carries its own 600 second timeout, so the wall clock fires mid-request rather than any assertion failing.

Auto-merge is already armed with squash, so once the conflict is resolved and the lanes come back this will merge on its own without needing anything further from me.

vllm_omni/model_executor/models/omnivoice/omnivoice_generator.py
```diff
@@ -432,6 +440,63 @@ def _precompute_rope_table(
...
+        "cu_seqlens_k": cu_seqs,
+        "max_seqlen_q": kernel_max_seqlen,
+        "max_seqlen_k": kernel_max_seqlen,
+    }
+    if not needs_sdpa_mask:
+        return AttentionMetadata(extra=extra)
+
+    token_ids = torch.arange(seq_len, device=cu_seqs.device, dtype=cu_seqs.dtype)
+    sequence_ids = torch.searchsorted(cu_seqs[1:], token_ids, right=True)
+    attn_mask = sequence_ids.view(1, 1, seq_len, 1) == sequence_ids.view(1, 1, 1, seq_len)
```
linyueqian, line 474: [blocking] The branch does not import after the rebase. General build 14663 hard-failed `Simple · Model Executor Test` and `Engine&Model Executor Test`, both exiting status 2 at collection with `ImportError: cannot import name '_additive_float_mask' from vllm_omni.model_executor.models.omnivoice.omnivoice_generator`, raised while importing `tests/model_executor/models/omnivoice/test_mask_dtype.py`. That is a collection error, so it takes down every test in the directory, not just the one module. The cause is a semantic conflict git could not flag: this PR removes `_additive_float_mask`, and main ships `test_mask_dtype.py` which imports it, so the two only collide once the rebase puts them in the same tree.

The removal itself looks right. The new metadata builder produces a boolean mask on this line, and a bool mask carries no dtype coupling at all, which is a cleaner answer than main's additive-float form rather than a regression of it.

What needs care is the coverage, not the function. `test_mask_dtype.py` exists to pin a specific production bug: SDPA rejects an additive mask whose dtype differs from the query, so materializing it as float32 unconditionally made OmniVoice servable only in float32 and killed every half-precision deployment on the first attention call. If the bool mask makes the additive contract obsolete, please replace that test with one pinning the new contract, that the SDPA fallback receives a mask requiring no dtype coercion, rather than deleting it. Deleting it would leave the branch importable and the half-precision guarantee silently uncovered, which is the outcome the test was written to prevent.

linyueqian (commented):
Withdrawing my approval at this head. The rebase that cleared the conflict introduced a build break, and it is the kind that only appears once main and the branch are in the same tree.

To be explicit about my own error, since I re-armed auto-merge on this an hour ago: I checked that the rebase preserved the approval by diffing all eleven of this PR's files between the previously approved `a502950b` and `ab97b1d7` and confirming they were byte-identical. That check was correct as far as it went, and it was the wrong check. It compares the branch against itself and cannot see a file main added in the meantime that depends on something the branch removes. Auto-merge stays armed but cannot fire while the required general lane is red; the inline finding is what needs resolving first.

The rest of the review is unchanged from round three. The production change and the answered `max_seqlen` question still read correctly, and the eager-path exact-shape test added since then is a reasonable addition rather than something needing another pass.

Two lane notes for when this is re-run. The AMD red on the previous head was `mi300_1: Diffusion · Model Test` with the four SenseNova paged-decode `ImportError: vllm.vllm_flash_attn requires the CUDA flash attention extensions` failures, the inherited ROCm image break, and this branch touches no diffusion code. The CosyVoice3-TTS E2E step timeout attributed at the two earlier heads is also unrelated to this change.

Validation: static review of the delta at `ab97b1d7` against its merge base, a file-level comparison against the previously approved head, the general lane job logs for build 14663, and the definition and call sites of `_additive_float_mask` on `origin/main` against their absence on this head. No fork code was executed locally.

vllm_omni/model_executor/models/omnivoice/omnivoice_generator.py
sphinxkkkbc: Resolved in c4197ffe163c23017f179bf2f766c67766aca249, also restored test in `test_serving_speech.py` 

Updated test command in description, re-tested locally, all passed.

sphinxkkkbc:
Found another rebase regression in `test_fused_projection_load.py`, which was also exposed by the previous CI run. Fixed in `70a2b193d07c6dcf33c43f2b6991d04411c12210`.

linyueqian:
Lanes re-fired at `70a2b193`. A suggestion that should save you a few round trips, since you have now hit this twice.

Both breaks were the same shape: main added or kept a test that calls an API this branch changes, so there is no textual conflict for git to report and the rebase produces a tree that does not import. First it was `test_mask_dtype.py` importing `_additive_float_mask`, which you removed, and now `test_fused_projection_load.py` calling `OmniVoiceGenerator(_config())` with the old single-argument signature. Neither needs a GPU or a full test run to find, because both fail at collection.

`pytest --collect-only -q tests/model_executor/models/omnivoice tests/entrypoints tests/diffusion` finds every remaining one of these in a single local pass, in seconds, with no model weights and no accelerator. Anything that imports cleanly there will at least reach its assertions in CI. Given each Buildkite round trip is half an hour or so, that is worth running once before the next push rather than discovering them one lane at a time.

Nothing else is outstanding from my side. The `test_mask_dtype.py` rewrite was the right call, and pinning the bool mask through an actual half-precision `scaled_dot_product_attention` call keeps the guarantee the original test existed for rather than dropping it with the function. Restoring the guidance-scale batching coverage in `test_serving_speech.py` was more than I asked for and closes the gap the earlier scheduler-test removal would have left.

Once a full lane set comes back clean at a single head I will re-approve there. I am deliberately not carrying the older approval forward, because the whole lesson of build 14663 is that a rebase invalidates the CI evidence an approval rested on, even when the branch's own files are unchanged.

linyueqian:
Do not chase the general lane red on build 14666. It is not yours, and I would rather say so before you spend another round trip on it.

The failure is `tests/distributed/omni_connectors/test_mooncake_transfer_engine_rdma.py::TestStressCorrectness::test_concurrent_put_get_threaded_both_sides`, with `AssertionError: errors: ['bi6: MD5 mismatch']`, one failure against 23 passed in that step. It is already tracked as #6977, and I have added today's evidence there. Your branch cannot reach it: that test imports only hashlib, os, subprocess, threading, time, pytest, torch, `tests.helpers.mark`, `tests.helpers.runtime`, and the Mooncake connector, and the single file you touch that is anywhere near shared code, `vllm_omni/diffusion/worker/utils.py`, changed by exactly one line, the SPDX header.

The rest of the lane set at `70a2b193` is the good news. Intel 7821 and NPU 6419 are green, so both of the rebase breaks you fixed are confirmed gone: no more `_additive_float_mask` import error and no more `OmniVoiceGenerator` signature mismatch. AMD 11419 is red only on the four SenseNova paged-decode `ImportError: vllm.vllm_flash_attn requires the CUDA flash attention extensions` failures, which is the ROCm image lacking those extensions and unrelated to this branch.

So the branch itself now looks clean, and what stands between it and an approval is one required lane that failed on a known connector flake. The `--collect-only` sweep from my last comment is still worth running once before your next push, in case a third signature mismatch is hiding behind the two you have already fixed, but I am not expecting one. Once a lane set comes back without the connector flake I will re-approve at that head.

sphinxkkkbc:
Thanks for the genuinely helpful suggestions. I verified locally with `--collect-only` and found no further issues.

This rebase ended up taking two extra CI rounds to fix small integration bugs that really should have been caught locally right after the rebase, so that was on me.

The AMD CI also hit the `od_config` constructor issue in the previous round. The latest AMD run has not fully finished yet, but that issue should be fixed by the latest change.

linyueqian (approved):
Approving at `70a2b193`. Both rebase-surfaced breaks are fixed and, more to the point, verified by execution rather than by the diff reading correctly.

Build 14666's `Simple · Model Executor Test` closed `1809 passed, 1 skipped, 608 deselected` with no failures and, crucially, no collection errors. That step ran the two files that were broken: every test in `test_fused_projection_load.py`, including all five `test_a_missing_shard_raises_instead_of_loading_partially` parametrizations that exercise the constructor you re-signatured, and `test_mask_dtype.py::test_sdpa_fallback_mask_is_dtype_independent`. `test_pipeline_batching.py::test_eager_attention_metadata_honors_exact_max_seqlen` ran too. Getting past collection at all is the signal that matters here, since both earlier failures were import-time.

The `test_mask_dtype.py` rewrite was the right resolution rather than the expedient one. Deleting the file would have made the branch importable and quietly dropped the half-precision guarantee it existed to protect; instead it now pins the new contract, and the CUDA case drives a real `scaled_dot_product_attention` call in float16 and bfloat16 with the production mask and asserts the output dtype and finiteness. That is the call that used to fail with "invalid dtype for bias". Restoring the guidance-scale coverage in `test_serving_speech.py` was beyond what I asked for and closes the gap the earlier scheduler-test removal left.

Both remaining reds are attributed and neither belongs to this branch. Intel 7821 and NPU 6419 are green. The general lane fails one step, `Distributed Test`, on one test, `test_mooncake_transfer_engine_rdma.py::TestStressCorrectness::test_concurrent_put_get_threaded_both_sides`, with `AssertionError: errors: ['bi6: MD5 mismatch']` against 23 passed. That is tracked as #6977, where I have added today's evidence, and it is unreachable from here: the test imports only stdlib, torch, two `tests.helpers` modules and the Mooncake connector, while your one file anywhere near shared code, `vllm_omni/diffusion/worker/utils.py`, changed by a single SPDX header line. AMD 11419 is red only on the four SenseNova paged-decode `ImportError: vllm.vllm_flash_attn requires the CUDA flash attention extensions` failures, the ROCm image break that reproduces on unrelated PRs, and this branch runs no diffusion code.

I am not holding the approval for a fresh general run. The one failing test is a known infrastructure flake with a tracking issue, and waiting on it would block a verified change on someone else's problem.

linyueqian:
One practical follow-up to the approval, because it is approved but still cannot merge.

Auto-merge is armed, so this will land by itself once the required checks are satisfied. The blocker is that `buildkite/vllm-omni` is a required check and it is still showing build 14666, which is red on the mooncake `test_concurrent_put_get_threaded_both_sides` MD5 mismatch. That failure is not yours, as covered in the review, but GitHub does not care whose it is.

I just tried clearing it without troubling you: cycling the `ready` label re-fired the NPU lane, but the general lane stayed pinned to the old failed build. The general pipeline triggers on push, so a label cycle cannot replace a stale general result. Only a new commit will.

So when you have a moment, merging current main into the branch will give you a fresh general run, and the flake is intermittent so it should come back clean. That is also worth doing on its own terms, since main has moved since 70a2b193 and you have already been bitten twice by tests on main that call APIs this branch changes. I would rather you spend the push on a real rebase than on an empty commit.

Nothing else is outstanding from my side.

sphinxkkkbc:
Rebased cleanly with no conflicts, ran the unit tests locally with no new errors raised. The ready label needs to be cycled.

PR #6409, opened 2026-08-20
[BugFix][TTS] Support long-form OmniVoice generation

### LINKED ISSUE #6333: [Bug]: OmniVoice silently drops words on long text — no long-form chunking (upstream has audio_chunk_duration)

akshatvishu:
## Purpose

Fixes #6333.

OmniVoice can drop or replace words when it generates long text in one pass. This change follows the official OmniVoice long form behavior. It splits requests estimated above 30 seconds into about 15 second text chunks, keeps the voice reference across chunks, decodes each chunk, and joins the audio with faded boundaries.

Users can override `audio_chunk_duration` and `audio_chunk_threshold` through `extra_params`.

Chunking must use one random generator for the whole request and advance it across chunks. The diffusion worker creates that generator from the canonical `sampling_params.seed`, while diffusion speech previously stored the seed in `extra_args`. This change maps speech seeds to the canonical field before dispatch and copies a legacy `extra_args["seed"]` when needed. It preserves the model owned value in `extra_args` because models such as Ming ImageGen and Cosmos RoboLab still read it directly.

The benchmark code is on the separate [`benchmark/6333-omnivoice-longform`](https://github.com/akshatvishu/vllm-omni/tree/benchmark/6333-omnivoice-longform/benchmarks/tts/omnivoice_longform) branch so this PR stays focused on the serving fix. I can open a separate benchmark PR after this fix merges if maintainers want it.

The diffusion worker already records peak reserved GPU memory for each response, but the speech endpoint returned only the audio bytes and dropped the output metadata. This change returns the existing value in `X-Peak-Memory-MB` when available. The benchmark can therefore record per-request worker memory from outside the server without a platform-specific GPU monitor. The endpoint does not add another memory measurement.

## Test plan

Core commit: `6af2f1c0d6632cd0bba1c4039944b0f87c119374`

```bash
.venv/bin/pytest -q tests/diffusion/test_diffusion_request.py tests/diffusion/models/omnivoice/test_chunking.py --run-level core_model
.venv/bin/pytest -q tests/entrypoints/openai_api/test_serving_speech.py::TestSpeechAPI::test_create_diffusion_speech_extra_params tests/entrypoints/openai_api/test_serving_speech.py::TestSpeechAPI::test_create_diffusion_speech_rejects_invalid_extra_seed --run-level core_model
```

The selected CPU test suites passed 57 tests. Pre-commit passed on all 13 changed files.

## Benchmark result

Benchmark branch revision: [`c6c095de8437f3a6e066686cfa88d3edf34e8f27`](https://github.com/akshatvishu/vllm-omni/commit/c6c095de8437f3a6e066686cfa88d3edf34e8f27)

The results below were collected from [`bba11e8b92659107e17dfb5d9e0b0a0218d691fa`](https://github.com/akshatvishu/vllm-omni/commit/bba11e8b92659107e17dfb5d9e0b0a0218d691fa). Later benchmark branch commits contain review fixes and documentation changes.

Use a clean vLLM-Omni checkout inside a ROCm container with vLLM-Omni and the OmniVoice runtime dependencies installed. Checking out the exact revision is safer than cherry-picking the benchmark because the result depends on the server and benchmark code from the same commit.

```bash
git fetch https://github.com/akshatvishu/vllm-omni.git benchmark/6333-omnivoice-longform
git switch --detach bba11e8b92659107e17dfb5d9e0b0a0218d691fa
bash benchmarks/tts/omnivoice_longform/run_benchmark.sh --small
```

The command uses GPU 0, model seed 42, concurrency 1, `k2-fsa/OmniVoice` at revision `c5fdb5ccb189668d56333f77ba2629f4cd7535f4`, and `openai/whisper-large-v3` at revision `06f233fe06e710322aca913c1bc4249a0d71fce1`. The script prints the result directory and writes the report to `evaluation/summary.md` under that directory.

The benchmark used the real OmniVoice and Whisper weights with 40 prompts from 100 to 507 words. All 160 measured generations succeeded.

GPU: 1x MI300x (vLLM-Omni official Docker ROCm Image)

| Prompt group | vLLM one shot | vLLM chunked | Official chunked |
| --- | ---: | ---: | ---: |
| About 120 words | 89.60% | 96.33% | 95.23% |
| About 200 words | 72.28% | 95.48% | 94.87% |
| About 300 words | 12.31% | 95.02% | 94.34% |
| About 500 words | 0.96% | 92.86% | 91.92% |

The table reports speech coverage.

Chunking reduced mean peak reserved GPU memory from 11.84 GiB to 9.45 GiB and reduced the maximum from 15.69 GiB to 11.44 GiB. Mean real time factor increased from 0.053 to 0.072.


akshatvishu:
### Self-Review:
Tested locally on both short (<30s, one-shot) and long (>30s, chunked) prompts. Confirmed speaker consistency across chunks for unconditioned runs, bounded GPU memory allocation via chunk-wise D2H offloading and deterministic seed progression. Unit tests passing.

clodaghwalsh17:
@akshatvishu I think it would be helpful it you could add a note to the docs on what is the recommended upper and lower bounds for the 2 new extra args (audio_chunk_duration and audio_chunk_threshold) so that a user can safely adjust these parameters without accidentally triggering this bug

tests/diffusion/models/omnivoice/test_chunking.py
```diff
@@ -0,0 +1,395 @@
...
+    ],
+)
+def test_chunking_seconds_rejects_invalid_values(value, allow_zero):
+    with pytest.raises(ValueError):
+        _parse_chunking_seconds("audio_chunk_duration", value, allow_zero=allow_zero)
+
+
+def test_pipeline_marks_invalid_chunking_values_as_client_errors():
+    pipeline = SimpleNamespace(
+        config=SimpleNamespace(
```
alex-jw-brooks: Can you refactor this to use the proper types where possible, e.g., the actual config type? using too many simple namespaces gets brittle if the objects evolve over time

tests/diffusion/models/omnivoice/test_chunking.py
```diff
@@ -0,0 +1,395 @@
...
+            audio_chunk_duration=15.0,
+            audio_chunk_threshold=30.0,
+        )
+    )
+    request = SimpleNamespace(
+        prompts=["Hello"],
+        sampling_params=SimpleNamespace(extra_args={}, generator=None),
+    )
+
+    with pytest.raises(RuntimeError, match="diffusion worker"):
```
alex-jw-brooks: Maybe this should be somewhere else? I.e., it's not very specific to chunking or omni voice, and is more of a test against the worker, right?

akshatvishu:
@vllm-omni-review-bot  Pr is ready to go ; I will fix the conflicts !

PR #6410, opened 2026-08-20
Perf/minimax h3 dit swiglu rope cache npu

wjialish:

## Purpose

Optimize MiniMax-H3 DiT inference on Ascend NPU.

This PR contains two complementary optimizations:

1. **Fused SwiGLU on NPU**

   - Reuse the shared diffusion `SiluAndMul(CustomOp)` introduced on main.
   - Dispatch its NPU path to `torch_npu.npu_swiglu(x, dim=-1)`.
   - Do not introduce a MiniMax-H3-specific duplicate SwiGLU layer.

2. **Request-local DiT RoPE table cache**

   - Materialize the rank-local BF16 RoPE table once before the MiniMax-H3 denoise loop.
   - Reuse the table across all denoise steps in the same request branch.
   - Cache the final packed table `[cos(theta), sin(theta)]`, rather than only caching RoPE frequencies.
   - Preserve the existing RoPE numerical path and BF16 output layout; this PR does not replace the rotary kernel.

The RoPE cache is request-local and sequence-parallel-rank-local. It is stored only in `MiniMaxH3DenoiseBranch.static_kwargs`; it is not retained on the model and is released when the request branch completes.

### TeaCache compatibility

TeaCache bypasses `MiniMaxH3DiTModel.forward()`, so its MiniMax-H3 extractor now also:

- consumes `kwargs["rope_table"]` when provided;
- validates the local table's shape, device, and BF16 dtype;
- falls back to `prepare_rope_table()` only when no cached table is supplied;
- mirrors the native forward sequence-parallel path, including `local_sp_prepare` and local-output gathering.

### 拉起服务

`#!/usr/bin/env bash
export PORT=8098
export MODEL="/mnt/weight/minimax-h3/FL2VA"
export VLLM_OMNI_VIDEO_SYNC_TIMEOUT=1800
export PYTHONDONTWRITEBYTECODE=1
export HF_HUB_OFFLINE=1
export TRANSFORMERS_OFFLINE=1
export PYTORCH_NPU_ALLOC_CONF=expandable_segments:True
export MINDIE_SD_FA_TYPE="ascend_laser_attention"
export HCCL_NPU_SOCKET_PORT_RANGE="auto"

vllm serve "${MODEL}" \
  --omni \
  --host 0.0.0.0 \
  --port "${PORT}" \
  --trust-remote-code \
  --num-gpus 8 \
  --usp 8 \
  --ring 1 \
  --text-encoder-tp-size 8 \
  --vae-parallel-mode tile \
  --vae-use-tiling \
  --vae-patch-parallel-size 8 \
  --enable-diffusion-pipeline-profiler \
  --diffusion-attention-config '{"default": {"backend": "RAINFUSION_ATTN",
      "block_sparse": {"sparsity": 0.8, "start_step": 12}}}' \
  --cache-backend cache_dit \
  --enable-cache-dit-summary \
  --cache-config '{"Fn_compute_blocks":2,"Bn_compute_blocks":1,"max_warmup_steps":4,"residual_diff_threshold":0.4,"max_continuous_cached_steps":4,"enable_taylorseer":true,"taylorseer_order":2}'`

### 推理
`
export PORT=8098
export API_URL="http://127.0.0.1:${PORT}/v1/videos/sync"
export PROMPT="For the target video, at 0.00 seconds into the target video, <Picture 1> (from [Shot 1]) is fully referenced.\n\nintegrated_multimodal_description: [Shot 1] This is a live-action, cinematic shot with a shallow depth of field. The camera holds a perfectly static shot throughout the entire eight-second duration, capturing a cozy family gathering in a traditional Japanese dining room. The scene opens with a large, intricately patterned blue and white ceramic bowl of ramen in the immediate foreground, rendered in crisp, sharp focus. The bowl sits on a smooth, polished long wooden table. Inside the bowl, a rich, oily golden-brown broth surrounds yellow wavy noodles, topped with two thick, round slices of chashu pork featuring visible fat marbling and a distinct spiral meat pattern. A generous mound of freshly chopped, bright green scallions rests in the center, and a crisp, dark green rectangular sheet of nori seaweed is tucked into the right edge. To the left of the bowl, a pair of light brown wooden chopsticks rests horizontally on a small, dark rectangular chopstick rest, near a small cylindrical ceramic teacup with blue painted patterns. On the right side of the table, a spherical paper lantern with a ribbed bamboo frame sits on a black wooden base. In the background, a large family of seven is gathered around the table, initially appearing as a soft, blurred presence. Behind them, traditional Japanese sliding shoji screens with wooden lattice frames are open, revealing a bright outdoor scene with lush green trees. Early in the clip, the thick, white steam rising from the hot ramen broth immediately intensifies, billowing upwards in thick, swirling clouds that dance continuously above the bowl. As the clip progresses into the middle seconds, the camera maintains its static position while the focus begins a deliberate, smooth shift deeper into the room. The foreground ramen bowl, its vibrant ingredients, and the rising steam gradually soften into a hazy, out-of-focus blur. Simultaneously, the family members in the background come into sharp, detailed clarity. The heavy steam continues to rise from the foreground, creating a dynamic, translucent veil between the camera and the family. With the focus now firmly locked on the background, the vibrant family dinner comes alive. The man in the dark navy blue long-sleeved shirt on the left leans forward, his mouth moving animatedly in a silent exchange. The young girl in the crisp white short-sleeved t-shirt beside him smiles brightly, looking toward the center of the table. The woman on the far left, wearing a soft light blue long-sleeved blouse, turns her head slightly, smiling gently. Across the table, the woman in the light grey button-down shirt smiles broadly, her eyes crinkling, as she rests her hands near her plate. The woman in the dark grey top further back uses her wooden chopsticks to pick up a small piece of food from a central ceramic dish filled with bright red pickled vegetables. The woman in the center back in the light grey sweater smiles gently, her hands clasped softly in front of her, observing the interaction. Throughout the remainder of the clip, the family continues their lively physical interaction, their mouths moving in continuous, silent cadences of conversation, while the thick, white steam from the blurred ramen bowl in the foreground never stops rising, adding a comforting atmosphere to the warm gathering.\n\noverall_soundscape: The soundscape begins with a quiet room tone mixed with the faint, airy rustle of the thick steam billowing from the hot ramen bowl in the foreground, accompanied by the subtle, continuous hissing and bubbling of the rich broth. As the visual focus shifts deeper into the room, the physical sounds of the bustling family dinner become dominant in the foreground. The clear, sharp clinking of ceramic bowls and wooden chopsticks touching plates is clearly heard as the family members reach for food. This is followed by the faint, muffled thud of a cup being set down on the smooth wooden table, and the subtle, rhythmic rustle of cotton and wool clothing as the family members lean forward and gesture, perfectly capturing the lively, physical atmosphere of the shared meal.\n\nnon_diegetic_music: A gentle, heartwarming acoustic guitar melody plays softly in the background, accompanied by the subtle, resonant notes of a traditional Japanese koto. The music maintains a slow, comforting tempo that enhances the cozy, nostalgic, and joyful atmosphere of the family gathering."
export FIRST_FRAME="${SCRIPT_DIR}/../sample_data/fl2va_first_frame.png"
export SEED=1101
export FLOW_SHIFT=12.0
export AUDIO_FLOW_SHIFT=3.0

## 5s
HDR_FILE=$(mktemp)
curl -sS -D "$HDR_FILE" -X POST "${API_URL}" \
  -F "prompt=${PROMPT}" \
  -F 'short_edge=768' \
  -F 'aspect_ratio=16:9' \
  -F 'fps=24' \
  -F 'num_inference_steps=50' \
  -F "flow_shift=${FLOW_SHIFT}" \
  -F "audio_flow_shift=${AUDIO_FLOW_SHIFT}" \
  -F "seed=${SEED}" \
  -F "input_references=@${FIRST_FRAME};type=image/png" \
  -F 'extra_params={"task":"fl2va","duration":5.0}' \
  -o "${OUTPUT_DIR}/fl2va_5s.mp4"

INF_TIME=$(hdr_val X-Inference-Time-S)
PEAK_MB=$(hdr_val X-Peak-Memory-MB)
STAGES=$(hdr_val X-Stage-Durations)
PEAK_GB=$(awk "BEGIN{printf \"%.2f\", (${PEAK_MB:-0})/1024}" 2>/dev/null)

echo "---------------- step1 result ----------------"
echo "output            : ${OUTPUT_DIR}/fl2va_5s.mp4"
echo "推理耗时          : ${INF_TIME} s   (X-Inference-Time-S)"
echo "峰值显存 (rank0)  : ${PEAK_MB} MB  (${PEAK_GB} GB reserved)   (X-Peak-Memory-MB)"
echo "各阶段耗时 (JSON) : (X-Stage-Durations)"

if [ -n "$STAGES" ]; then
    printf '%s\n' "$STAGES" | python3 -m json.tool 2>/dev/null || printf '%s\n' "$STAGES"
fi
rm -f "$HDR_FILE"

## 15s
HDR_FILE=$(mktemp)
curl -sS -D "$HDR_FILE" -X POST "${API_URL}" \
  -F "prompt=${PROMPT}" \
  -F 'short_edge=768' \
  -F 'aspect_ratio=16:9' \
  -F 'fps=24' \
  -F 'num_inference_steps=50' \
  -F "flow_shift=${FLOW_SHIFT}" \
  -F "audio_flow_shift=${AUDIO_FLOW_SHIFT}" \
  -F "seed=${SEED}" \
  -F "input_references=@${FIRST_FRAME};type=image/png" \
  -F 'extra_params={"task":"fl2va","duration":15.0}' \
  -o "${OUTPUT_DIR}/fl2va_15s.mp4"

INF_TIME=$(hdr_val X-Inference-Time-S)
PEAK_MB=$(hdr_val X-Peak-Memory-MB)
STAGES=$(hdr_val X-Stage-Durations)
PEAK_GB=$(awk "BEGIN{printf \"%.2f\", (${PEAK_MB:-0})/1024}" 2>/dev/null)

echo "---------------- step3 result ----------------"
echo "output            : ${OUTPUT_DIR}/fl2va_15s.mp4"
echo "推理耗时          : ${INF_TIME} s   (X-Inference-Time-S)"
echo "峰值显存 (rank0)  : ${PEAK_MB} MB  (${PEAK_GB} GB reserved)   (X-Peak-Memory-MB)"
echo "各阶段耗时 (JSON) : (X-Stage-Durations)"

if [ -n "$STAGES" ]; then
    printf '%s\n' "$STAGES" | python3 -m json.tool 2>/dev/null || printf '%s\n' "$STAGES"
fi
rm -f "$HDR_FILE"`

## image
image

**vLLM Version:**
0.26.0

### Performance

E2E Latency 5s : 65147ms -> 61691ms.

ori_5s:
main_5s

swiglu_rope_cache_5s：
swiglu_rope_cache_5s

E2E Latency 10s : 136399ms  -> 129754ms.

ori_10s:
main_10s

swiglu_rope_cache_10s：
swiglu_rope_cache_10s

E2E Latency 15s :232002ms  -> 222223ms.
ori_15s:
main_15s

swiglu_rope_cache_15s：
swiglu_rope_cache_15s

### Precision
ori_5s:

https://github.com/user-attachments/assets/9c565426-9726-4ab9-8fe1-ea745d5dbbcd

swiglu_rope_cache_5s：

https://github.com/user-attachments/assets/22794e71-7c71-4760-9c7c-2c22dc796c38

ori_10s:

https://github.com/user-attachments/assets/813bb437-32eb-4f4c-880c-1570f2ec6a05

swiglu_rope_cache_10s：

https://github.com/user-attachments/assets/9cc9b7a7-9fe0-44fc-8cca-5a7b675fe890


hsliuustc0106:
perf A/B tests

david6666666:
perf A/B tests and accuracy test，included video output

vllm_omni/diffusion/models/minimax_h3/denoise_loop.py
```diff
@@ -164,6 +180,7 @@ def minimax_h3_denoise_loop(
         raise ValueError("video/audio sigma schedules must have equal length")
     if len(sigmas_video) < 2:
         raise ValueError("sigma schedules need at least 2 entries")
+    positive.prepare_rope_table(model)
```
david6666666, line 184: TeaCache bypasses `forward()` and its MiniMax-H3 extractor still rebuilds RoPE from `img_position_ids`, so this cache is unused there. Could the extractor consume `kwargs["rope_table"]` too?

vllm_omni/diffusion/layers/swiglu.py
```diff
@@ -0,0 +1,36 @@
...
+
+"""SwiGLU layer with platform-specific fused implementations."""
+
+import torch
+import torch.nn.functional as F
+
+from vllm_omni.diffusion.layers.custom_op import CustomOp
+
+
+class SwiGLU(CustomOp):
```
david6666666: Current main already uses the shared `SiluAndMul` here. After rebasing, could we add the NPU dispatch there instead of introducing a second SwiGLU layer?

wjialish:
> perf A/B tests

Added A/B tests details.

wjialish:
> perf A/B tests and accuracy test，included video output

Done.

vllm_omni/diffusion/layers/swiglu.py
wjialish: Addressed in 76179eed.

Rebased onto current main and dropped the duplicate `SwiGLU` layer.
The Ascend implementation now reuses the shared `SiluAndMul` in
`vllm_omni/diffusion/layers/activation.py`, with
`forward_npu()` dispatched to `torch_npu.npu_swiglu`.

Added coverage for the shared activation path as well.

vllm_omni/diffusion/models/minimax_h3/denoise_loop.py
wjialish: Addressed in 76179eed.

`extract_minimax_h3_context()` now consumes `kwargs["rope_table"]`
when it is provided by the denoise loop. It validates the rank-local
table and mirrors the native `forward()` sequence-parallel path
(`local_sp_prepare` and local-output gathering).

It falls back to `prepare_rope_table()` only when no cached table is
provided. Added a regression test that fails if the TeaCache extractor
rebuilds RoPE after receiving a prepared table.

vllm_omni/diffusion/cache/teacache/extractors.py
```diff
@@ -1332,7 +1344,7 @@ def extract_minimax_h3_context(
         refiner_max_seqlen=refiner_max,
         seq_len=seq_len,
         device=device,
-        local_span=(0, seq_len),
+        local_span=local_span,
```
david6666666, line 1351: With strict SP, `decoder_input` is rank-local here, but `combined_indices` is still full-length; `run_transformer_blocks()` also gathers before TeaCache subtracts the rank-local input. The extractor therefore mixes local and gathered layouts and fails on the first full-compute step. Please keep the TeaCache state rank-local end to end, gather only for final postprocessing, and add a TeaCache + SP test.

tests/diffusion/layers/test_activation.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def test_silu_and_mul_native_matches_packed_reference() -> None:
+    packed = torch.randn(257, 256, dtype=torch.float32)
+    gate, up = packed.chunk(2, dim=-1)
+
+    torch.testing.assert_close(SiluAndMul().forward_native(packed), F.silu(gate) * up)
+
+
+@pytest.mark.skipif(not current_omni_platform.is_npu(), reason="requires Ascend NPU")
```
david6666666: This test inherits the module-level `cpu` mark and has no `npu` mark. CPU CI skips it, while the NPU jobs select `-m npu`, so the fused path never runs in CI. Please give it the NPU hardware marks and add it to an NPU job.

david6666666 (commented):
Two issues in the changed paths:

tests/diffusion/layers/test_fused_qk_norm_rope_npu.py
```diff
@@ -0,0 +1,158 @@
...
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+
+import sys
+import types
+
+import pytest
+import torch
+import torch.nn.functional as F
+
+pytestmark = [pytest.mark.core_model, pytest.mark.cpu, pytest.mark.diffusion]
```
david6666666: These are CPU mocks, so the public NPU dispatch and real torch_npu/MindIE signatures are still not exercised; the new NPU job only runs test_activation.py. Could we add a small A3 parity test through fused_qk_norm_rope() and include this file in that job?

vllm_omni/diffusion/cache/teacache/extractors.py
david6666666: The strict-SP layouts now match, but the cache decision is also rank-local: _should_compute_full_transformer() has no SP reduction, so different shards can cross the threshold differently and some ranks enter the block collectives while others skip to sp_gather(). Please synchronize the L1 stats or decision across the SP group. Also, the local_len == seq_len fallback still gathers twice (in run_transformer_blocks() and postprocess()), which expands an already-global tensor; covering both cases would close this thread.

david6666666:
LGTM now, please pass ci

david6666666 (approved):

david6666666:
please pass ci, and paste your result

tests/diffusion/cache/test_teacache_extractors.py
```diff
@@ -705,6 +729,215 @@ def counted_forward(*args, **kwargs):
...
+            def __init__(self):
+                super().__init__()
+                self.inputs: list[torch.Tensor] = []
+
+            def forward(self, tensor):
+                self.inputs.append(tensor.clone())
+                if tensor.shape[-1] == hidden_size:
+                    raise AssertionError("TeaCache must not gather hidden states in strict SP")
+                assert tensor.shape == (local_len, video_width + audio_width)
+                return tensor
```
congw729: The [`Simple · Diffusion Test`](https://buildkite.com/vllm/vllm-omni/builds/14150/list?sid=01a03d67-b7cf-4962-9ae2-e124c312d12a) currently fails in `test_teacache_strict_sp_keeps_cache_state_rank_local` with `IndexError: index out of range in self`.

It looks like this mock returns the rank-local compact logits unchanged, while `postprocess()` subsequently applies the global `infer_out_pos` / `audio_pos` indices. In this test, `local_len == 3` but `audio_pos` contains index `3`, so the failure is deterministic. Could we make this mock simulate the full SP gather so the test exercises the intended post-gather layout?

wjialish: Fixed in f182afb7. The strict-SP gather mock now simulates the all-gather of rank-local compact logits into the global [seq_len, video_width + audio_width] layout before postprocess() applies global output indices. It still rejects hidden-state gathers, so the test continues to verify that TeaCache state remains rank-local end to end.

tests/diffusion/cache/test_teacache_extractors.py
```diff
@@ -705,6 +729,222 @@ def counted_forward(*args, **kwargs):
...
+            return original_indexed_scale_shift(hidden, shift, scale, indices)
+
+        monkeypatch.setattr(minimax_h3_modulation, "indexed_scale_shift_", capture_indexed_scale_shift)
+
+        local_img_pos = torch.tensor([0, 1])
+        local_audio_pos = torch.tensor([2, 3])
+        token_tags = torch.full((seq_len,), -1, dtype=torch.long)
+        token_tags[local_img_pos] = 0
+        token_tags[local_audio_pos] = 2
+        inverse_indices = torch.arange(seq_len, dtype=torch.long) % 2
```
david6666666: This still fails the new strict-SP test. sample_inputs has one unique timestep, but % 2 lets combined_indices reach 5 while the mocked AdaLN table above has only 3 rows. Could we keep inverse_indices at zero here, or mock two timesteps and six AdaLN rows?

wjialish: Fixed in 49109643. The fixture represents a single unique timestep, so inverse_indices now stays zero instead of alternating with % 2. This keeps combined_indices within the mocked three-row AdaLN table while preserving the strict-SP layout and post-gather assertions covered by the test.

tests/diffusion/layers/test_activation.py
wjialish: Addressed in 626b7c70 and the follow-up updates.
Added an A3 hardware parity test that calls the public fused_qk_norm_rope() entry point with real NPU tensors, verifies the NPU dispatch is selected, and compares Q/K outputs against the eager RMSNorm + RoPE reference.  
The NPU ready pipeline now runs both test_activation.py and test_fused_qk_norm_rope_npu.py with the core_model and npu and A3 selection, so the real torch_npu / MindIE path is exercised in CI.

tests/diffusion/layers/test_fused_qk_norm_rope_npu.py
wjialish: Addressed in ed5890b3. Added an A3 hardware parity test through the public fused_qk_norm_rope() entry point with real NPU tensors, verifies that the NPU dispatch is selected, and compares against the eager RMSNorm + RoPE reference. The NPU ready job now includes test_fused_qk_norm_rope_npu.py, so the real torch_npu / MindIE path is exercised in CI.

vllm_omni/diffusion/cache/teacache/extractors.py
wjialish: Addressed across 626b7c70 and ed5890b3. In strict SP, TeaCache now keeps the modulation input, hidden states, and cached residuals rank-local. state_combined_indices is sliced for the rank-local cache state, while local_sp_prepare remains responsible for block-input sharding. Hidden states are gathered only in the non-strict/global fallback; strict SP gathers compact logits once during postprocessing.  
Cache decisions are now synchronized with an SP MAX all-reduce, so if any rank requires a full transformer execution, all ranks take the collective block path. Ranks forced from cache to full compute reset their accumulated L1 state. Added strict-SP coverage and a non-strict fallback test that asserts hidden states are gathered exactly once.

tests/diffusion/layers/test_activation.py
wjialish: The NPU SwiGLU test now uses @hardware_test(res={"npu": "A3"}, num_cards=1) in 626b7c70 instead of inheriting the module-level CPU mark / runtime skip. The NPU ready pipeline now includes test_activation.py and selects core_model and npu and A3, so the real Ascend torch_npu.npu_swiglu path is exercised in CI.

vllm_omni/diffusion/layers/fused_qk_norm_rope.py
```diff
@@ -11,18 +11,23 @@
   ``[cos(theta), sin(theta)]`` with ``theta`` of width ``rotary_dim // 2``.
```
hsliuustc0106, line 11: can you open a npu directory under layers? cc @Bounty-hunter

hsliuustc0106:
@Bounty-hunter please provide some suggestions to this PR in terms of how we can add the npu specific optimization

vllm_omni/diffusion/layers/fused_qk_norm_rope.py
hsliuustc0106: Refining this: `fused_qk_norm_rope` currently has only the MiniMax-H3 production caller, so please keep the hardware implementations model-local, e.g. `diffusion/models/minimax_h3/layers/{native,nvidia,npu}`, with `layers/__init__.py` as the single platform-selection point (the existing HunyuanImage3 pattern). The model should import one `fused_qk_norm_rope` symbol; each selected backend retains its package/dtype/shape/layout checks and falls back to native. A global `diffusion/layers/npu` would make sense only once a second model family shares this contract. This follows #6305 and #6691.

vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
```diff
@@ -1362,18 +1412,26 @@ def forward(self, **kwargs: Any) -> tuple[torch.Tensor, torch.Tensor]:
...
-            seq_len,
-            hooks_applied=hooks_applied,
-        )
+        local_span = self._rope_local_span(seq_len)
         local_start, local_len = local_span
-        rope_position_ids = img_position_ids.narrow(1, local_start, local_len)
-        rope_table = _build_rope_table(self.rope(rope_position_ids).to(device))
+        rope_table = kwargs.get("rope_table")
+        if rope_table is None:
+            if current_omni_platform.is_npu():
```
hsliuustc0106, line 1419: Can we keep this forward path platform-neutral? The presence of `rope_table` already selects the cached path. When it is absent, always run the exact pre-PR reference builder (`narrow` → `self.rope` → `_build_rope_table`); when present, validate and reuse it. That removes the `current_omni_platform` import from the model while preserving the CUDA/CPU reference behavior.

vllm_omni/diffusion/models/minimax_h3/denoise_loop.py
```diff
@@ -241,6 +259,11 @@ def minimax_h3_denoise_loop(
         raise ValueError("video/audio sigma schedules must have equal length")
     if len(sigmas_video) < 2:
         raise ValueError("sigma schedules need at least 2 entries")
+    # Keep CUDA/CPU on the reference path: its accuracy CI compares the
+    # generated video against the official artifact. The table reuse is an
+    # Ascend-specific performance optimization and must not alter that path.
+    if current_omni_platform.is_npu():
```
hsliuustc0106, line 277: Please select the request-cache policy once instead of branching on platform here. A consumer-owned MiniMax-H3 policy can return `{}` by default and `{"rope_table": model.prepare_rope_table(...)}` for NPU; this loop then unconditionally updates `static_kwargs`. Keep selection in the model leaf adapter/registry, and invoke the same preparation seam for request, step, and batched construction paths so the tensor remains request- and rank-local.

vllm_omni/diffusion/layers/fused_qk_norm_rope.py
wjialish: Thanks for the clarification. We agree that the implementation should eventually be moved to a MiniMax-H3 model-local, platform-dispatched layers package.  
As this optimization is needed for the upcoming release, we would like to merge the current PR first. We will follow up with a separate refactoring PR to move the native, NVIDIA, and NPU implementations into diffusion/models/minimax_h3/layers/, while preserving the existing dispatch, validation, fallback behavior, and test coverage.

vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
wjialish: Thanks for the suggestion. The NPU-only guard was added intentionally after the initial implementation because our local hardware validation was limited to Ascend NPU. Keeping the cache preparation restricted to NPU ensures that CUDA/CPU continue to follow the exact pre-PR reference path until they are covered by equivalent validation.

vllm_omni/diffusion/models/minimax_h3/denoise_loop.py
wjialish: Thanks for the suggestion. The NPU-only guard was added intentionally after the initial implementation because our local hardware validation was limited to Ascend NPU. Keeping the request-cache preparation restricted to NPU ensures that CUDA/CPU continue to follow the exact pre-PR reference path until they are covered by equivalent validation.
Since this PR is scoped to the Ascend optimization and is needed for the upcoming release, we would prefer to keep the current guard for now. We can follow up with a separate refactoring PR to move the platform-specific request-cache policy into a MiniMax-H3 model-local adapter and unify the preparation seam across request construction paths.

david6666666:
## GPU accuracy validation (B300)

I ran the requested accuracy file on the PR base and head:

- Base: `13173d6bbbd0f0ddd4f6caa2debabcf2375d911d`
- Head: `49109643242cd7946256a3977129b8e14bdbbb15`
- Test: `tests/e2e/accuracy/minimax_h3/test_minimax_h3_i2va_ref2va_similarity.py`
- Result: **2 passed** on both base and head

| Case | Base vs official (SSIM / PSNR) | Head vs official (SSIM / PSNR) | Base vs head output |
|---|---:|---:|---|
| I2VA | `0.972637` / `35.074233 dB` | `0.972637` / `35.074233 dB` | byte-identical, SHA256 `25a5dc30309d25053ac861ac75734c5190b56a24d871f4da6f3e91df07886035` |
| Ref2VA | `0.980578` / `42.650344 dB` | `0.980578` / `42.650344 dB` | byte-identical, SHA256 `72211f15c0af87a41bec4b4a904c3019f05b2c9088e0febc28349ff9ee9cda0a` |

Thresholds are SSIM `>= 0.97` and PSNR `>= 34 dB`. Base completed in `1083.29s`; head completed in `1061.94s` (timings included model startup/compile and are not intended as a performance comparison).

Environment/config:

- 4x NVIDIA B300 SXM6 AC (GPU 4-7), compute capability 10.3, driver `610.43.02`
- Python `3.12.3`, PyTorch `2.13.0+cu130` (CUDA runtime `13.0`), vLLM `0.28.0`
- MiniMax-H3 snapshot `42ed227ee7df40d41602854ae760620d6eb651fe`, seed `0`, 50 inference steps
- `--num-gpus 4 --usp 4 --ring 1 --use-hsdp --hsdp-shard-size 4 --text-encoder-tp-size 4 --vae-patch-parallel-size 4 --diffusion-attention-backend FLASH_ATTN`

Compatibility note: both tested SHAs contain #6307, which fails before model loading in this environment with `ImportError: cannot import name 'get_fs_group'`. I applied the same worker-title-only revert from #6717 to both worktrees; the resulting worker file was identical on both sides and this workaround does not touch the MiniMax-H3 numerical path.

**Conclusion:** no GPU accuracy impact was observed for the requested I2VA and Ref2VA cases; both generated MP4 files are byte-identical between base and PR head.

PR #6415, opened 2026-08-20
[NPU] Support running diffusion stages without vllm-ascend

blian6:
## Summary

Split the NPU platform into a shared **interface** plus two **backend implementations**:

- **`NPUOmniPlatform`** — shared NPU interface, holding only the methods both backends use identically and vllm-ascend does not define.
- **`DiTNPUOmniPlatform`** — torch_npu-native backend for **pure diffusion stages** (e.g. Qwen-Image): runs with the torch_npu backend + mindiesd attention, **zero vllm-ascend imports**.
- **`ARNPUOmniPlatform`** — vllm-ascend-enhanced backend for **AR/generation stages** (Qwen3-Omni, TTS, etc.): keeps the original diamond-inheritance pattern with vllm-ascend's `NPUPlatform` and all Ascend enhancements.

The backend is selected **per stage** via the `VLLM_OMNI_DISABLE_VLLM_ASCEND` env var in the stage's `runtime.env` (typically in the deploy YAML's `platforms.npu.stages` section). It **defaults to `false`** — the vllm-ascend backend, matching the pre-split behavior — so existing deployments need **zero config change**, and no new shared config surface is added (`stage_config.py` is untouched).

## Motivation

Previously `vllm_omni/platforms/npu/platform.py` imported and inherited from `vllm_ascend.platform.NPUPlatform` at module scope, making the entire NPU path require vllm-ascend — even for diffusion models that never use it. This PR makes vllm-ascend an **explicit, per-stage, opt-out dependency**: AR/generation stages keep it (required, fail-fast with install guidance if missing), diffusion stages can disable it (`VLLM_OMNI_DISABLE_VLLM_ASCEND=true`) and run standalone.

## Changes

| File | Change |
|---|---|
| **`vllm_omni/platforms/npu/platform.py`** | `NPUOmniPlatform(OmniPlatform)` — shared NPU interface, decoupled from `vllm_ascend.platform.NPUPlatform`. Placement rule: only methods used identically by both backends and **not** defined by vllm-ascend's `NPUPlatform` live here (the complement of the vllm-ascend surface), so the AR platform's diamond MRO is conflict-free. Keeps `adopt_as_vllm_platform()`; adds the shared `_vllm_ascend_available()` helper |
| **`vllm_omni/platforms/npu/dit_platform.py`** *(new)* | `DiTNPUOmniPlatform(NPUOmniPlatform)` — torch_npu implementations of the vLLM `Platform` entries a diffusion stage needs (mindiesd attention, hccl distributed); zero vllm-ascend imports |
| **`vllm_omni/platforms/npu/ar_platform.py`** *(new)* | `ARNPUOmniPlatform(NPUOmniPlatform, vllm_ascend.platform.NPUPlatform)` — original diamond-inheritance pattern: vLLM `Platform` entries from vllm-ascend, class body keeps the Ascend enhancements (custom ops, ACL graph wrapper, ascend forward context, ascend config/logging, model patches, torch_npu total-memory override); module-level `vllm_ascend` import isolated to this file |
| **`vllm_omni/platforms/__init__.py`** | NPU plugin selects the backend class per stage from `VLLM_OMNI_DISABLE_VLLM_ASCEND` (true → `DiTNPUOmniPlatform`; unset/false → `ARNPUOmniPlatform`, the default); optional per-platform `adopt_as_vllm_platform` hook on lazy resolution |
| **`vllm_omni/platforms/npu/__init__.py`** | Lazily exports `NPUOmniPlatform`, `DiTNPUOmniPlatform`, `ARNPUOmniPlatform` |
| **`vllm_omni/engine/arg_utils.py`** | Adopt the Omni platform as vLLM's `current_platform` from the vllm general plugin before worker subprocesses spawn |
| **`vllm_omni/model_executor/layers/rotary_embedding/mrope.py`** | Fall back to upstream `MRotaryEmbedding` on standalone NPU |
| **`vllm_omni/patch.py`** | Lazily import `OmniMRotaryEmbedding` to keep it out of the unconditional import graph |
| **`vllm_omni/engine/stage_init_utils.py`** | Fail early with a clear message when an AR/generation stage runs on NPU without vllm-ascend |

### Per-stage backend scope

Deploy YAML declares the NPU backend per stage via the stage `runtime.env`
(no new shared config field):

```yaml
platforms:
  npu:
    stages:
      - stage_id: 1            # diffusion: no vllm-ascend needed
        runtime:
          env:
            VLLM_OMNI_DISABLE_VLLM_ASCEND: "true"
      - stage_id: 0            # AR: requires vllm-ascend (default)
        # runtime.env left unset -> VLLM_OMNI_DISABLE_VLLM_ASCEND=false
```

- `true` → disable vllm-ascend, force the standalone torch_npu backend (`DiTNPUOmniPlatform`)
- unset or `false` (**default**) → vllm-ascend backend (`ARNPUOmniPlatform`; a missing vllm-ascend raises from the AR class constructor with install guidance)

### Platform structure

`NPUOmniPlatform(OmniPlatform)` is the shared interface; `DiTNPUOmniPlatform(NPUOmniPlatform)` implements the torch_npu `Platform` entries; `ARNPUOmniPlatform(NPUOmniPlatform, vllm_ascend.platform.NPUPlatform)` keeps the original diamond-inheritance pattern.

The diamond MRO — `ARNPUOmniPlatform → NPUOmniPlatform → OmniPlatform → NPUPlatform → Platform` — resolves every entry unambiguously: omni/device helpers from the interface, vLLM `Platform` entries from vllm-ascend, Ascend enhancements from the class body.

### vLLM current_platform adoption

Without vllm-ascend, vLLM 0.26 resolves `current_platform` to `UnspecifiedPlatform`, which breaks `vllm.utils.mem_utils.DeviceMemoryProfiler` in worker subprocesses. The Omni platform adopts itself as vLLM's `current_platform` (via the general plugin and the lazy-resolution hook) and re-binds already-imported vllm modules.

### Cleanups

- `ARNPUOmniPlatform.check_and_update_config` calls `super()` first, preserving vllm-ascend's own config checks (parallel-config validation, incompatible-config fixes).
- `find_spec("vllm_ascend")` unified into one `_vllm_ascend_available()` helper (AR constructor guard, `adopt_as_vllm_platform` hook).

## Backward Compatibility

| Scenario | Before | After |
|---|---|---|
| **With vllm-ascend** (AR/generation, e.g. Qwen3-Omni, TTS) | `NPUOmniPlatform` inherits `NPUPlatform`; all vllm-ascend enhancements active | **Identical behavior.** `ARNPUOmniPlatform` inherits the interface plus vllm-ascend's `NPUPlatform`; the same vllm-ascend code paths run (`__init__` patches, `set_device` custom ops, `init_diffusion_*`, ACL graph wrapper, ascend forward context, `import_kernels`, static graph), and vllm-ascend's `check_and_update_config` validation is **preserved** via `super()`. vLLM's `current_platform` remains vllm-ascend's `NPUPlatform`. |
| **Without vllm-ascend, unset (default `false`)** | `import vllm_omni.platforms` → `ModuleNotFoundError` (hard failure) | `ARNPUOmniPlatform` raises with install guidance: install vllm-ascend, or set `VLLM_OMNI_DISABLE_VLLM_ASCEND=true` to run the standalone torch_npu backend. |
| **Without vllm-ascend, `VLLM_OMNI_DISABLE_VLLM_ASCEND=true`** (pure diffusion, e.g. Qwen-Image) | (n/a) | Runs on `DiTNPUOmniPlatform` (torch_npu backend): mindiesd attention, hccl distributed, no vllm-ascend package needed. |
| **Mixed pipeline** (AR + diffusion stages) | Requires vllm-ascend for the whole deployment | AR stages still require vllm-ascend (fail early with a clear message if missing); diffusion stages work with or without it. |

**Key guarantee**: existing vllm-ascend deployments see **zero behavior change** — `VLLM_OMNI_DISABLE_VLLM_ASCEND` **defaults to `false`**, so vllm-ascend is only skipped when a stage explicitly opts in (`true`).

### Scripts (reproducible)

- **`tools/start_serve_no_ascend.sh`** — launch a Qwen-Image serve on the standalone backend (sets `VLLM_OMNI_DISABLE_VLLM_ASCEND=true`):
  ```bash
  bash tools/start_serve_no_ascend.sh /data/models/Qwen-Image-2512 8091 2,3
  ```
- **`tools/verify_npu_standalone.py`** — assert platform resolution + diffusion chain load without vllm-ascend:
  ```bash
  VLLM_OMNI_DISABLE_VLLM_ASCEND=true python tools/verify_npu_standalone.py   # (no vllm-ascend env)
  ```
  Output (Ascend 950PR, no vllm-ascend): `platform: DiTNPUOmniPlatform`, `num_compute_units: 28`, `vllm current_platform: DiTNPUOmniPlatform`, `diffusion chain OK; ALL_OK`
- **`tools/generate_image_no_ascend.py`** — POST `/v1/images/generations` and decode the PNG:
  ```bash
  python tools/generate_image_no_ascend.py --host 127.0.0.1 --port 8091
  ```

### Test image

`tools/qwen_noascend.png` — generated by the standalone path (2×NPU, TP=2, 20 steps, seed 42, prompt "A beautiful landscape painting of mountains at sunset, vibrant colors"):

qwen_noascend

Serve stats for this generation: `denoise_step_latency_ms 249.97`, `diffusion_engine_exec_time_s 4.969`, `POST /v1/images/generations 200 OK` — output byte-identical to the vllm-ascend baseline (3,147,861 bytes).

### End-to-end on Ascend 950PR × 8

- **DiT standalone** (`VLLM_OMNI_DISABLE_VLLM_ASCEND=true`, TP2 on NPU 0,1): Qwen-Image-2512, `/v1/images/generations` **200 in ~11.8s**, 1024×1024 PNG, content-valid.
- **DiT default mode** (vllm-ascend drives diffusion, env unset): same model, `/v1/images/generations` **200 in ~11.9s**, 1024×1024 PNG, content-valid.
- **AR**: Qwen3-Omni-30B-A3B-Instruct full 3-stage pipeline (thinker/talker/code2wav): `/health` 200; text→text; audio-in→text; text→text+audio (14.7s / 24 kHz WAV). `tools/verify_ar_platform.py` → `AR_PLATFORM_OK`.

### AR platform layer

`tools/verify_ar_platform.py` on the ascend env: default mode resolves `ARNPUOmniPlatform`; vllm-ascend entries inherited via MRO (`get_device_name: Ascend950PR_9579`, `support_static_graph_mode: True`, `num_compute_units: 28`); class-body Ascend enhancements callable (ACLGraph wrapper, `set_forward_context`, `check_and_update_config`); forced `VLLM_OMNI_DISABLE_VLLM_ASCEND: true/false` selection correct; `AR_PLATFORM_OK`.

## Test Plan

- [x] Lint (ruff) + syntax (py_compile) on all changed files
- [x] Import-chain regression with vllm-ascend present (ascend path behavior unchanged: ACLGraphWrapper, AscendAttentionBackend, NPUCommunicator, static-graph True)
- [x] Import-chain regression without vllm-ascend (standalone path with `VLLM_OMNI_DISABLE_VLLM_ASCEND=true`: torch_npu interfaces, `DeviceMemoryProfiler` works)
- [x] End-to-end `vllm serve` + `/v1/images/generations` on NPU without vllm-ascend (output identical to baseline)
- [x] `tools/verify_npu_standalone.py` passes on the no-vllm-ascend environment
- [x] AR platform layer on the ascend env (`verify_ar_platform.py`: `ARNPUOmniPlatform` resolution, MRO, `true/false` backend selection, `AR_PLATFORM_OK`)
- [x] Qwen3-Omni AR end-to-end + Qwen-Image-2512 DiT (standalone & default modes) on Ascend 950PR
- [ ] CI (runs on PR)


hsliuustc0106:
@blian6 please submit an issue to the weekly meeting. In addition, please also consider about the NPU tests and docker image~ cc @FayeSpica @fan2956

vllm_omni/config/stage_config.py
```diff
@@ -942,6 +942,12 @@ def merge_pipeline_deploy(
             runtime["num_replicas"] = ds.num_replicas
             if ds.env is not None:
                 runtime["env"] = ds.env
+            if ds.enable_standalone is not None:
+                # Propagate the per-stage NPU backend scope to stage
+                # subprocesses (read by NPUOmniPlatform at lazy init).
+                stage_env = dict(runtime.get("env") or {})
+                stage_env["VLLM_OMNI_ENABLE_STANDALONE"] = str(ds.enable_standalone).lower()
+                runtime["env"] = stage_env
```
gcanlin: I think it's not necessary to add this env and field `enable_standalone`. Just fallback to dit platform when vllm-ascend is not available. Users don't need to be aware.

gcanlin (commented):
Thanks! Overall the idea is fine. Just need to check whether the performance happen regression when vllm-ascend is not available.

vllm_omni/config/stage_config.py
blian6: Adding a manually controlled parameter makes it easier to distinguish between environmental anomalies and active human control.

blian6:
> This PR appears to belong to: **docs/design/module/execution_platforms.md**.
> 
> Module owners: @gcanlin @tjtanaa @xuechendi
> 
> @blian6, please review your own changes and leave a short self-review comment describing what you checked. PRs without author self-review may not be assigned a reviewer.
> 

I have personally reviewed this PR. In line with feedback from the regular meeting, I have avoided modifying public parameters; instead, I added an NPU-specific parameter to control the scenario where vLLM Ascend is disabled. Additionally, for multi-backend platforms, I separated the platforms for AR and DiT to prevent function shadowing caused by diamond dependencies.
@vllm-omni-review-bot

blian6:
> @blian6 please submit an issue to the weekly meeting. In addition, please also consider about the NPU tests and docker image~ cc @FayeSpica @fan2956

I have already covered both AR and DiT scenarios using the qwen3-omni and qwen-image models. At the same time, I understand that, to ensure generality, there is no need to provide multiple separate Docker images.

FayeSpica:
I think we should add a CI-only image without `vllm-ascend` to cover the standalone DiT path. I’ll add this coverage later as part of the NPU CI image work tracked in #6806.

vllm_omni/engine/stage_init_utils.py
```diff
@@ -598,6 +598,20 @@ def extract_legacy_stage_metadata(stage_config: Any) -> StageMetadata:
 
     _apply_rocm_attention_backend(engine_args, stage_type)
 
+    # NPU backend scope pre-check: AR/generation stages require vllm-ascend,
```
hsliuustc0106: move this out as an indepdent fucntion

vllm_omni/platforms/npu/ar_platform.py
```diff
@@ -0,0 +1,141 @@
...
+from vllm.forward_context import BatchDescriptor
+from vllm.logger import init_logger
+from vllm_ascend.platform import NPUPlatform
+
+from vllm_omni.platforms.npu.platform import NPUOmniPlatform, _vllm_ascend_available
+
+logger = init_logger(__name__)
+
+
+class ARNPUOmniPlatform(NPUOmniPlatform, NPUPlatform):
```
hsliuustc0106, line 49: [P1] `get_diffusion_kv_block_tables_cls` / `build_diffusion_kv_attn_metadata` were deleted from the shared interface but not restored on this class, so the NPU paged-KV path (`diffusion/diffusion_kv/model_runner_backend.py:267`, `paged_attention_adapter.py:371`) now resolves to the `OmniPlatform` upstream defaults (`vllm.v1.worker.gpu.block_table.BlockTables`, `vllm.v1.worker.gpu.attn_utils.build_attn_metadata`). The deleted Ascend builders owned the `seq_lens_np` + `attn_state=ChunkedPrefill` contract this path was built on (`flash_attn.py` sets `supports_paged_kv=True`, and the `callable(...)` guard at `model_runner_backend.py:82` passes for the default, so the regression is silent). Please restore both Ascend implementations here on `ARNPUOmniPlatform` — by this PR's own placement rule they belong on the concrete subclass, since vllm-ascend does not define them; for `DiTNPUOmniPlatform`, either implement torch_npu-native paged KV or override to raise so the guard trips loudly.

vllm_omni/platforms/npu/platform.py
```diff
@@ -215,6 +163,24 @@ def get_diffusion_attn_backend_cls(
...
+        )
+
+        # Both patches import the MiniMax encoder package, whose __init__ loads
+        # pipeline_minimax_h3 → diffusion.data. Doing that during platform
+        # construction races vllm_omni/__init__.py (patch before config) and
+        # closes a cycle through pipeline_registry → PI0_PIPELINE →
+        # DiffusionOutput. Apply them only after the platform exists, before
+        # the diffusion pipeline is loaded.
+        apply_minimax_h3_qwen3vl_patch()
+        apply_minimax_h3_qwen3vl_swiglu_patch()
```
hsliuustc0106, line 195: [P3] This hunk drops the only call site of `apply_minimax_h3_qwen3vl_sdpa_patch` while keeping the swiglu/main patches, so the removal looks accidental: on NPU, MiniMax-H3's Qwen3-VL text encoder falls back to the reference `repeat_interleave`-expanded GQA in `encoder.py:_scaled_dot_product_attention` instead of the native `enable_gqa` path — same math, but extra K/V materialization (memory + latency) on both backends. Please keep the patch call, or call out the intentional removal in the PR description.

vllm_omni/model_executor/layers/rotary_embedding/mrope.py
```diff
@@ -20,7 +20,16 @@
 from vllm_omni.platforms import current_omni_platform
 
 if current_omni_platform.is_npu():
-    from vllm_ascend.ops.rotary_embedding import AscendMRotaryEmbedding as _BaseMRotaryEmbedding
+    from importlib.util import find_spec
+
+    if find_spec("vllm_ascend") is None:
```
hsliuustc0106: [P3] Gating on `find_spec("vllm_ascend")` disagrees with the platform selection, which honors `VLLM_OMNI_DISABLE_VLLM_ASCEND`: with vllm-ascend installed but the DiT backend selected, importing this module still pulls in `vllm_ascend.ops`, violating the zero-vllm-ascend-import property this PR adds for diffusion stages (and an installed-but-broken vllm-ascend would crash the import). Suggest keying off the selected backend (the same env check the plugin uses) instead of package presence.

vllm_omni/engine/stage_init_utils.py
blian6: Done — extracted into _check_npu_ar_stage_requires_ascend(stage_id, stage_type), called by extract_legacy_stage_metadata. (commit d38910aa)

vllm_omni/platforms/npu/ar_platform.py
blian6: Fixed — ARNPUOmniPlatform restores get_diffusion_kv_block_tables_cls (AscendBlockTables) and build_diffusion_kv_attn_metadata (seq_lens_np + ChunkedPrefill contract), so the paged-KV path uses the Ascend builders again; DiTNPUOmniPlatform overrides both to raise NotImplementedError. (commit d38910aa)

vllm_omni/platforms/npu/platform.py
blian6: Fixed — the removal was accidental; init_diffusion_model_runner_runtime now applies main + sdpa + swiglu, keeping MiniMax-H3's Qwen3-VL text encoder on native enable_gqa. (commit d38910aa)

vllm_omni/model_executor/layers/rotary_embedding/mrope.py
blian6: Fixed — mrope.py now keys off the same VLLM_OMNI_DISABLE_VLLM_ASCEND selection the plugin uses, so standalone DiT never imports vllm_ascend.ops even if installed; find_spec remains only as a guard for the misconfigured AR-without-vllm-ascend case. (commit d38910aa)

vllm_omni/engine/stage_init_utils.py
hsliuustc0106: Verified at 20ec458c — thanks.

vllm_omni/platforms/npu/ar_platform.py
hsliuustc0106: Verified at 20ec458c: ARNPUOmniPlatform restores both Ascend builders and DiTNPUOmniPlatform now raises with guidance, so the silent upstream fallback is gone. Thanks.

vllm_omni/platforms/npu/platform.py
hsliuustc0106: Verified at 20ec458c — sdpa patch applied again alongside the main and swiglu patches. Thanks.

vllm_omni/model_executor/layers/rotary_embedding/mrope.py
hsliuustc0106: Verified at 20ec458c — env-keyed selection matches the platform gate, and the find_spec path now only guards the misconfigured AR case. Thanks.

vllm_omni/platforms/__init__.py
```diff
@@ -75,7 +93,13 @@ def npu_omni_platform_plugin() -> str | None:
     except Exception as e:
         logger.debug("NPU OmniPlatform is not available because: %s", str(e))
 
-    return "vllm_omni.platforms.npu.platform.NPUOmniPlatform" if is_npu else None
+    if not is_npu:
+        return None
+
+    flag = os.environ.get("VLLM_OMNI_DISABLE_VLLM_ASCEND", "false").strip().lower()
```
hsliuustc0106, line 99: [P1] Resolve the platform selector after applying stage runtime.env. current_omni_platform is a process-global singleton, but stage_runtime.py imports it before stage_runtime_env(...) is entered for the stage. In the default inline or single-stage path, runtime.env: VLLM_OMNI_DISABLE_VLLM_ASCEND=true is therefore ignored and the cached AR class remains in use. That breaks the documented per-stage selection and can keep vllm-ascend in a supposedly standalone diffusion stage or fail before DiT can be selected. Please resolve or pass the platform per stage after applying the runtime environment, or make this explicitly process-global.

vllm_omni/platforms/npu/ar_platform.py
```diff
@@ -0,0 +1,231 @@
...
+"""
+
+from functools import cache
+from typing import Any
+
+import torch
+from vllm.config import CUDAGraphMode, VllmConfig
+from vllm.forward_context import BatchDescriptor
+from vllm.logger import init_logger
+from vllm_ascend.platform import NPUPlatform
```
hsliuustc0106, line 20: [P2] Keep the optional vllm_ascend import behind the dependency guard. This module-level import runs before ARNPUOmniPlatform.__init__ checks _vllm_ascend_available(). On NPU with the default selector and no vllm-ascend installed, current_omni_platform resolution raises a raw ModuleNotFoundError, so the promised install or standalone-backend guidance is unreachable. Move the import and class binding behind guarded resolution, or select DiT automatically when the dependency is unavailable.

vllm_omni/platforms/npu/dit_platform.py
```diff
@@ -0,0 +1,108 @@
...
+    Implements the vLLM ``Platform`` entries a diffusion stage needs
+    (mindiesd attention, hccl distributed) with torch_npu-native calls. The
+    torch_npu method bodies mirror the corresponding implementations in
+    vllm-ascend's ``NPUPlatform`` (vllm_ascend/platform.py); keep them in
+    sync when either side changes.
+    """
+
+    def __init__(self) -> None:
+        # 310P worker patches are torch_npu-level and apply on both backends.
+        from vllm_omni.platforms.npu._310p import apply_patches as apply_310p_patches
```
hsliuustc0106, line 35: [P2] Avoid importing vllm_ascend from the standalone backend. The DiT constructor calls the 310P helper, whose hardware probe imports vllm_ascend.utils and can then load vllm_ascend 310P worker code. Thus VLLM_OMNI_DISABLE_VLLM_ASCEND=true still couples the standalone backend to the optional AR package whenever it is installed, which conflicts with the no-vllm-ascend isolation contract. Use a torch_npu-only capability probe or keep this patch selection in the AR path.

hsliuustc0106:
Model coverage is not sufficient to establish the stated diffusion scope. Qwen-Image is the only model-specific no-vllm-ascend result described here; the changed tree contains no standalone validation scripts and no DiTNPUOmniPlatform model tests. Existing NPU Wan coverage is only I2V-A14B and does not set the standalone flag; T2V/TI2V and HunyuanVideo-1.5 coverage is CUDA-only. Please add no-ascend smoke coverage for Qwen-Image, Wan2.2 T2V/I2V/TI2V, and HunyuanVideo-1.5 T2V/I2V, with fixed inputs and startup/first-request evidence. If Hunyuan-Video means the original model rather than HunyuanVideo-1.5, that model is not registered by this repository.

vllm_omni/patch.py
```diff
@@ -305,6 +304,11 @@ def _patched_glm_image_text_config_init(self, *args, **kwargs):
     if hasattr(module, "TokensPrompt") and module.TokensPrompt == _OriginalTokensPrompt:
         module.TokensPrompt = OmniTokensPrompt
     if hasattr(module, "MRotaryEmbedding") and module.MRotaryEmbedding == _OriginalMRotaryEmbedding:
+        # Lazy import: OmniMRotaryEmbedding is NPU/AR-path specific; keep it
```
FayeSpica, line 307: I ran into an issue while testing MiniMax-H3, which requires Triton kernels on NPU and therefore depends on `triton-ascend`.

I tested this PR without `vllm-ascend`, using `quay.io/ascend/cann:9.1.0-a3-ubuntu22.04-py3.12` and vLLM v0.28.0 installed with `VLLM_TARGET_DEVICE=empty`. I installed `triton-ascend==3.2.2` alongside `triton==3.5.0`, as required by upstream vLLM. The vLLM-Omni server failed to start with:

```text
ImportError: cannot import name 'constexpr_type' from 'triton.language.core'
```

`vllm-ascend` avoids this failure by applying [the following compatibility patch](https://github.com/vllm-project/vllm-ascend/blob/d7697b7476c598ac4cdc612da6a219cb46cccb7e/vllm_ascend/__init__.py#L22-L46):

```python
_triton_available = importlib.util.find_spec("triton") is not None

if "triton.experimental" not in sys.modules:
    _experimental = ModuleType("triton.experimental")
    _experimental.__path__ = []
    sys.modules["triton.experimental"] = _experimental
for _gluon_stub in (
    "triton.experimental.gluon",
    "triton.experimental.gluon.language",
):
    if _gluon_stub not in sys.modules:
        sys.modules[_gluon_stub] = ModuleType(_gluon_stub)

# main2main compat: `_aggregate` was added to triton.language.core in
# vllm main post-0.26.0. Stub it here so vllm.triton_utils can import it
# without breaking on triton-ascend 3.2.1. Skip if triton is not
# installed at all (e.g. 310P or CPU-UT environments).
if _triton_available:
    try:
        import triton.language.core as _tl_core  # type: ignore[import-untyped]
    except Exception:
        pass
    else:
        if not hasattr(_tl_core, "_aggregate"):
            _tl_core._aggregate = lambda *a, **kw: None
```

If we want to support the standalone NPU DiT path without `vllm-ascend`, vLLM-Omni will need to maintain a `triton-ascend` compatibility patch.
@hsliuustc0106, does it make sense to maintain this patch in vLLM-Omni?

FayeSpica (commented):
**Blocked by a `triton-ascend` compatibility issue**

hsliuustc0106:
@blian6 gentle ping — this high-priority PR has seen no updates for 14 days and review feedback is waiting. Could you share a status or ETA when you get a chance? Thanks!

hsliuustc0106 (approved):
Everything from the 09-01 review is now resolved at this head: the P1 (restored Ascend builders on ARNPUOmniPlatform, explicit guidance-raising on the DiT platform instead of silent upstream fallback) and the helper-function nit were verified at 20ec458c, and both P3s are fixed here — `apply_minimax_h3_qwen3vl_sdpa_patch()` is called again, and the mrope gating keys off `VLLM_OMNI_DISABLE_VLLM_ASCEND` with the `find_spec` check demoted to the misconfiguration fallback, matching platform selection. Mergeable and checks green. Good to merge.

Issue #6414, opened 2026-08-20
[RFC] Introduce a Host Weight Runtime for reusable runtime-ready artifacts


hsliuustc0106:
## Summary

Introduce a loader-adjacent, transport-independent **Host Weight Runtime** for
immutable, runtime-ready host-weight artifacts.

The runtime provides three foundational contracts:

- `HostWeightStore`: exact lookup, coordinated production/import, validation,
  atomic publication, and lifecycle management;
- `HostWeightLease`: stable CPU tensor views and process-local mmap lifetime
  ownership; and
- `WeightProducer`: deterministic creation of final-layout BF16, FP8, or future
  representations.

A versioned `WeightRestorer` contract completes the model-loading boundary. It
reconstructs a valid model from a lease before the ordinary shared
post-load/finalization lifecycle runs.

This should be the common foundation for the final-layout BF16 cache in #6240,
the runtime FP8 cache in #6399, and future remote artifact materialization. It
must not make DLO, CUDA transport, or a particular quantization implementation
the owner of storage semantics.

## Related work and proposed sequencing

- #6195 defines the broader separation between DLO host storage and DP request
  scheduling.
- #6231 defines the supported parallel-layout and deployment-topology boundary.
- #6330 motivates a normalized online-FP8 runtime artifact.
- #6213 is merged and provides the Phase A direct-checkpoint mmap path. That
  path remains separate because checkpoint layout and deferred transforms are
  not necessarily final runtime layout.
- #6240 currently implements a DLO-specific final-layout BF16 cache and
  registered direct H2D.
- #6399 currently implements a separate runtime-FP8 cache for DLO AllGather.
- #6374 changes DLO transport scheduling and should remain orthogonal to host
  artifact storage.

Both #6240 and #6399 demonstrate useful consumers, but each currently owns
some combination of identity, locking, manifest, publication, validation, and
recovery. Landing those storage implementations independently would create two
versioned model-reconstruction paths and make later unification a migration.

Proposed order:

1. Agree on this RFC.
2. Land a small foundation PR containing the contracts and local filesystem
   store.
3. Rebase or split #6240 so it supplies the final-layout BF16 producer/restorer
   and keeps CUDA registration/H2D in DLO transport.
4. Rebase or split #6399 so it supplies the FP8 producer/restorer instead of a
   second cache implementation.
5. Add remote providers, cohort production, NUMA enforcement, and eviction as
   independent follow-ups.

[...]

PR #6419, opened 2026-08-20
[Core] Add Host Weight Runtime foundation

hsliuustc0106:
## Purpose

Introduce the top-level, neutral Host Weight Runtime foundation agreed in #6414. This PR creates the contracts and CPU local-filesystem implementation needed before DLO, runtime FP8, or another loader can consume reusable runtime-ready host weights.

The PR adds:

- exact canonical identity, manifest, publication, failure, outcome, producer, restorer, store, lease, and inspection contracts;
- a local filesystem store with stable build/artifact locks, one-builder coordination, invisible temporary construction, full publication-time hashing, fsync, and atomic publication;
- descriptor-backed safetensors mmap leases that retain a shared artifact lock and expose registration-ready mapped ranges without owning GPU transfer;
- deny/quarantine/rebuild, crash recovery, nonblocking cleanup, disk/tmpfs classification, capacity preflight, and typed read-only inspection;
- preferred/required local resolution policy, including typed initialization failure handling and structured terminal reports; and
- module and feature design documents separating semantic identity, storage, restoration, consumer integration, and GPU transport ownership.

The implementation is deliberately neutral to diffusion and DLO. It adds no CLI/config wiring, model-specific BF16/FP8 producer or restorer, CUDA registration, H2D path, remote provider, cohort producer, eviction, or change to DLO AllGather/no-AllGather behavior. Unsupported V1 capabilities fail explicitly instead of silently changing semantics.

Suggested review order:

1. `vllm_omni/host_weight_runtime/{identity,manifest,protocols,lease}.py` — semantic contracts.
2. `vllm_omni/host_weight_runtime/filesystem/` — lifecycle and durability.
3. `runtime.py`, `inspection.py`, `tests/host_weight_runtime/`, and the module/feature design documents — policy and evidence.

Part of #6414.

## Test Plan

- Canonical identity and exact-schema tests.
- Streaming BF16 and available FP8 safetensors round trips.
- Multiprocess one-builder/joined-lease coordination using `spawn`.
- Atomic invisibility during construction, bounded lock timeout, and crashed-builder recovery.
- Full-checksum and malformed-safetensors corruption detection, read-only inspection, deny, quarantine, and rebuild.
- Fork-safe lease locks, overlapping validation/deny coordination, and exact producer/restorer schema validation.
- Conflicting non-cooperative publication without silent replacement.
- Symlink rejection, descriptor cleanup, active-lease cleanup exclusion, and shared-backing fingerprints.
- Artifact/store/free-space limits, injected `ENOSPC`, domain-policy mismatch, disk/tmpfs policy, and explicit unsupported results.
- Disabled/preferred/required runtime policy, retryable versus nonretryable failures, explicit unsupported post-load publication, and observer isolation.
- Full changed-file pre-commit suite, including Python 3.10 mypy, SPDX, forbidden imports, test marks, and markdownlint.

**vLLM Version:** 0.27.0

**PR Head:** `12cbcd42c776d67aff010f4179a338b915271938`

**Base Commit:** `5730af0a95299387c233e12dfbdbe728aace0e58`

## Test Result

```text
72 passed, 14 warnings in 31.91s
```

The three spawn-based concurrency tests also passed in three additional consecutive stress runs. The review-fix regressions cover forked lock ownership, overlapping denial, failure policy, filesystem locality, restoration boundaries, malformed safetensors, exact manifest schemas, and unsupported post-load publication. All changed-file pre-commit hooks passed. The warnings are pre-existing PyTorch `torch.jit.script_method` deprecation warnings. This foundation PR makes no production performance or model-quality claim; those validations belong to each producer/restorer consumer PR.


hsliuustc0106:
Self-review completed at `418f927aff3a6ff7ee6675652b4eca319a27b98f`.

- Checked the semantic contracts and exact identity boundary: requested representations cannot be silently substituted; TP-owned bytes are explicit, replicated DP identity is excluded, and unsupported cohort/remote/NUMA/fs-verity/eviction behavior fails with typed outcomes.
- Checked filesystem lifecycle and concurrency: stable locks live outside replaceable artifact directories; mutators use build-lock → artifact-lock ordering; construction stays invisible until fsync + atomic publication; leases retain the shared artifact lock and exact descriptor-backed mappings; corruption, deny/quarantine/rebuild, capacity, `ENOSPC`, and crashed-builder paths are covered.
- Checked failure policy: preferred mode falls back only for recoverable store failures, while invalid or unsupported configuration remains visible; required mode fails without an exact lease.
- Checked package ownership and subtraction: the package has no diffusion, DLO, CUDA, distributed-state, engine, or model imports; internal writer/lock helpers are not exported, and consumer-specific producer/restorer/transport integrations remain out of this PR.
- Rebased onto current `main`; 50 focused CPU tests pass, all changed-file pre-commit hooks pass, and DCO, wheel builds, and Read the Docs are green. Platform Buildkite jobs are currently running.

I found no remaining blocking issue in the frozen patch. The intentional V1 limits are documented and return explicit unsupported results rather than silently changing behavior.

hsliuustc0106:
Follow-up self-review at `93542f53c3efadd3dea85970618c660e56fc0a9e`:

CUDA build #13883 ran 1,619 selected model-executor tests; 1,617 passed and the only failures were the two new spawn-based Host Weight Runtime tests. The worker took about 12 seconds to import vLLM on the loaded CI host, exceeding a 10-second startup bound. A second test hit a finish-at-timeout race where cleanup observed a worker as alive immediately before it exited successfully.

The fix changes only the test harness:

- process startup and whole-group completion now have explicit 60-second bounds;
- joining uses one group deadline instead of multiplying a timeout per process;
- a genuine timeout still terminates/kills workers and fails with their names/PIDs; and
- the short store lock, producer, and resolution deadlines remain unchanged.

Validation after the fix: the full package is `50 passed`; the three spawn tests passed in three additional consecutive stress runs; mypy and all changed-file pre-commit hooks pass. No product/store behavior changed.

vllm_omni/model_executor/host_weight_runtime/filesystem/store.py
```diff
@@ -0,0 +1,1085 @@
...
+                    StoreStatus.FAILED,
+                    failure=_failure(
+                        ResolutionStage.LIFECYCLE,
+                        FailureCode.QUARANTINE_FAILED,
+                        f"failed to prepare artifact rebuild: {exc}",
+                    ),
+                )
+
+            staging = self.tmp_dir / f"{identity.key}.{os.getpid()}.{uuid.uuid4().hex}"
+            staging.mkdir(mode=0o700)
```
hsliuustc0106: [P2] `staging.mkdir()` escapes the typed-failure contract — an ENOSPC/EDQUOT/EACCES here (tmpfs-full at exactly this call is the case capacity preflight tries to guard) raises a raw `OSError` out of `get_or_build` instead of a typed `StoreResult`; every other `OSError` in the method is typed. `HostWeightRuntime.resolve()` doesn't catch it either, so a preferred-mode caller would crash instead of taking canonical fallback. Smallest fix: move the `mkdir`/writer setup inside the try (typed ENOSPC/DOMAIN failure + `_discard_temp`).

vllm_omni/host_weight_runtime/lease.py
```diff
@@ -0,0 +1,143 @@
...
+
+    @property
+    def closed(self) -> bool:
+        with self._state_lock:
+            return self._closed
+
+    def close(self) -> None:
+        with self._state_lock:
+            if self._closed:
+                return
```
hsliuustc0106, line 110: [P3] Concurrent `close()` returns before teardown finishes — `_closed` is set under `_state_lock`, but the resource closers and artifact-lock release run after the lock is dropped, so a second concurrent `close()` observes `_closed` and returns while unmapping is still in flight; "close() returned ⇒ mappings released and shared lock dropped" doesn't hold for concurrent callers. Single-owner usage (the documented transport teardown) is unaffected — either hold the state lock across teardown or document single-owner close.

vllm_omni/host_weight_runtime/filesystem/store.py
```diff
@@ -0,0 +1,1085 @@
...
+                    retryable=True,
+                ),
+            )
+        except OSError as exc:
+            return StoreResult(
+                StoreStatus.FAILED,
+                failure=_failure(ResolutionStage.DOMAIN, FailureCode.DOMAIN_UNAVAILABLE, str(exc), retryable=True),
+            )
+
+        if self._deny_path(identity.key).exists():
```
hsliuustc0106, line 538: [P3] Orphaned deny marker reports INVALID forever — deny is checked before entry existence, and `cleanup()` unlinks the deny only after removing the entry (store.py:998-1005), so a crash in that window leaves deny-without-entry and subsequent lookups return `ARTIFACT_DENIED` for an artifact that no longer exists until an explicit `cleanup()`/rebuild clears it. Treating a missing entry as MISS (a deny is only meaningful while something exists to deny) would self-heal lookup-only consumers.

vllm_omni/host_weight_runtime/lease.py
hsliuustc0106: Fixed in `ae190763`. `HostWeightLease` now holds a re-entrant state lock through tensor clearing, resource teardown, and artifact-lock release, so another concurrent `close()` cannot return until teardown completes. Added a blocking-closer concurrency regression.

vllm_omni/model_executor/host_weight_runtime/filesystem/store.py
hsliuustc0106: Fixed in `ae190763`. Staging-directory creation and writer setup now execute inside the typed production `try`: `ENOSPC`/`EDQUOT` return `CAPACITY` + `ENOSPC`, failures before writer creation such as `EACCES` return `DOMAIN` + `DOMAIN_UNAVAILABLE`, and temporary cleanup is attempted on every failure. Added a parametrized regression for all three errno cases.

vllm_omni/host_weight_runtime/filesystem/store.py
hsliuustc0106: Fixed in `ae190763`. Lookup now checks artifact existence before consulting the deny marker while retaining the shared artifact lock, so deny-without-entry is a `MISS`; a subsequently coordinated build removes the stale marker as before. Added an orphan-marker regression.

vllm_omni/host_weight_runtime/filesystem/locks.py
```diff
@@ -0,0 +1,108 @@
...
+                if remaining <= 0:
+                    raise FileLockTimeoutError(f"timed out waiting for {self.path}") from exc
+                time.sleep(min(0.05, remaining))
+
+    def close(self) -> None:
+        if self._closed:
+            return
+        self._closed = True
+        try:
+            fcntl.flock(self._fd, fcntl.LOCK_UN)
```
hsliuustc0106, line 85: [P1] A forked child can release the parent’s active lease lock here. `flock` is associated with the inherited open-file description, so child `lease.close()`/`__del__` issues `LOCK_UN` for the parent too; I reproduced an exclusive acquisition while the parent lease object remained open. Track the creator PID and, in an inherited child, close the FD without `LOCK_UN`; please add a `fork` regression (the current process tests use `spawn`).

vllm_omni/host_weight_runtime/filesystem/store.py
```diff
@@ -0,0 +1,1096 @@
...
+            return StoreResult(
+                StoreStatus.INVALID,
+                failure=_failure(
+                    ResolutionStage.LOOKUP,
+                    FailureCode.ARTIFACT_DENIED,
+                    f"artifact {identity.key} is denied pending quarantine",
+                ),
+            )
+        try:
+            lease = self._open_lease(identity, entry_path, artifact_lock, validation)
```
hsliuustc0106, line 549: [P1] This can return a lease after the artifact has been denied by an overlapping full-checksum lookup. A metadata lookup can pass the initial deny check, pause here, then resume after another reader writes the deny marker; reproduction ended with `STRONG=invalid DENY=True OVERLAP_WEAK=hit`. Recheck the deny marker after `_open_lease()` while retaining the shared lock, and close the lease and return `INVALID` if denial won.

vllm_omni/model_executor/host_weight_runtime/runtime.py
```diff
@@ -0,0 +1,315 @@
...
+                ResolutionAttempt(
+                    stage=ResolutionStage.PRODUCTION,
+                    result=AttemptResult.SKIPPED,
+                    action=self._fallback_action(),
+                    elapsed_seconds=0.0,
+                    failure=unavailable,
+                )
+            )
+
+        outcome = (
```
hsliuustc0106: [P1] Preferred mode unconditionally converts every post-lookup/production failure into canonical fallback here. Nonretryable `PRODUCER_UNSUPPORTED`, `IDENTITY_COLLISION`, publication failures, etc. are therefore masked despite the policy saying only recoverable failures may fall back. Centralize stage/code/retryability-to-action policy and add a post-initialization nonretryable test.

vllm_omni/host_weight_runtime/filesystem/store.py
```diff
@@ -0,0 +1,1096 @@
...
+                    ),
+                )
+
+            staging = self.tmp_dir / f"{identity.key}.{os.getpid()}.{uuid.uuid4().hex}"
+            writer: FilesystemArtifactWriter | None = None
+            manifest: WeightManifest | None = None
+            try:
+                staging.mkdir(mode=0o700)
+                writer = FilesystemArtifactWriter(staging, capacity_check=self._check_capacity)
+                metadata = producer.produce(writer)
```
hsliuustc0106, line 862: [P1] The configured total resolution deadline cannot bound this synchronous call: the producer receives no deadline, timeout is checked only after it returns and finalizes, and publication later also uses `deadline=None`. A 50 ms deadline with a 200 ms producer returned after about 209 ms. Either execute production in enforceably bounded isolation or narrow the advertised deadline contract; add slow/hung-producer coverage.

vllm_omni/model_executor/host_weight_runtime/filesystem/store.py
```diff
@@ -0,0 +1,1096 @@
...
+            continue
+        mount_point = Path(_mount_unescape(left_fields[4]))
+        try:
+            resolved.relative_to(mount_point)
+        except ValueError:
+            continue
+        if len(str(mount_point)) > best_length:
+            best_length = len(str(mount_point))
+            best_type = right_fields[0]
+    return StorageClass.TMPFS if best_type in {"tmpfs", "ramfs"} else StorageClass.DISK
```
hsliuustc0106: [P1] This classifies NFS, CIFS, Lustre, Ceph, and every unknown non-tmpfs mount as `DISK`, so cross-node storage silently satisfies the node-local contract and changes `flock`/page-cache assumptions. Preserve filesystem type and reject known remote mounts (or require an explicit non-local backend/override) rather than defaulting every non-tmpfs mount to local disk.

vllm_omni/model_executor/host_weight_runtime/protocols.py
```diff
@@ -0,0 +1,214 @@
...
+    def produce(self, writer: ArtifactWriter) -> ProductionMetadata: ...
+
+
+@dataclass(frozen=True)
+class BuildRequest:
+    identity: WeightArtifactIdentity
+
+
+class WeightRestorePlan(Protocol):
+    def commit(self) -> object: ...
```
hsliuustc0106: [P1] The restoration contract does not define mutation or rollback semantics: `plan_restore` may mutate the model, `commit()` has an unspecified return type, and a failure could let preferred fallback reuse a partially hydrated model. Make planning validation-only, make commit the sole one-shot mutating phase, specify its return type, and require a fresh model for canonical fallback after commit failure. GPU registration and transport should remain outside this contract.

hsliuustc0106 (commented):
Please address these lifecycle and contract issues.

hsliuustc0106:
@vllm-omni-review-bot PTAL

vllm_omni/host_weight_runtime/filesystem/locks.py
hsliuustc0106: Fixed in `12cbcd42`. `FileLock` now records its creator PID; an inherited child closes its descriptor without issuing `LOCK_UN`, so it cannot release the parent’s open-file-description lock. Added a fork regression proving cleanup still reports `ACTIVE_LEASE` until the parent closes its lease.

vllm_omni/host_weight_runtime/filesystem/store.py
hsliuustc0106: Fixed in `12cbcd42`. Lookup now rechecks the deny marker after `_open_lease()` while retaining the shared artifact lock. If overlapping validation denied the artifact, the new lease is closed and lookup returns typed `INVALID`/`ARTIFACT_DENIED`. Added a deterministic overlapping weak/strong validation regression.

vllm_omni/model_executor/host_weight_runtime/runtime.py
hsliuustc0106: Fixed in `12cbcd42`. Failure-to-action selection is centralized: preferred mode falls back only for retryable failures, while nonretryable producer, identity-collision, and publication failures fail startup. Added post-initialization tests for the nonretryable cases plus a retryable-capacity control.

vllm_omni/model_executor/host_weight_runtime/filesystem/store.py
hsliuustc0106: Fixed in `12cbcd42`. The backend now preserves the kernel filesystem type, accepts only an explicit local-filesystem allowlist plus tmpfs/ramfs, rejects known remote and unknown types, and records `filesystem_type` in domain metadata and inspection. Added regressions for NFS, CIFS, Lustre, Ceph, and an unknown filesystem.

vllm_omni/host_weight_runtime/filesystem/store.py
hsliuustc0106: Fixed in `12cbcd42` by narrowing the contract instead of claiming cancellation that V1 cannot enforce. `resolution_timeout_seconds` is now `coordination_timeout_seconds` and bounds filesystem lock acquisition only; synchronous validation, production, and atomic publication run to completion. The stale post-production timeout result was removed, and a slow-producer regression plus both design docs capture the boundary.

vllm_omni/model_executor/host_weight_runtime/protocols.py
hsliuustc0106: Fixed in `12cbcd42`. `plan_restore()` is now explicitly validation-only; the returned plan has a one-shot `commit() -> None`, which is the sole model-mutating phase. The contract and both design docs require discarding a partially hydrated model and constructing a fresh model after commit failure. GPU registration and transport remain outside restoration.

vllm_omni/host_weight_runtime/filesystem/locks.py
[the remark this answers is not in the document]
hsliuustc0106: Fixed in `12cbcd42`. `FileLock` records its creator PID and inherited children close the FD without `LOCK_UN`; the fork regression verifies the parent lease continues to block cleanup.

vllm_omni/host_weight_runtime/filesystem/store.py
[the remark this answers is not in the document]
hsliuustc0106: Fixed in `12cbcd42`. Lookup rechecks denial after opening the lease, closes the lease if denial won, and returns typed `INVALID`/`ARTIFACT_DENIED`. A deterministic overlapping-validation regression covers the race.

vllm_omni/model_executor/host_weight_runtime/runtime.py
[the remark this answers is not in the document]
hsliuustc0106: Fixed in `12cbcd42`. Failure policy is centralized and preferred mode now canonical-falls back only for retryable failures. Nonretryable producer, identity-collision, and publication failures fail startup, with focused regressions.

vllm_omni/host_weight_runtime/filesystem/store.py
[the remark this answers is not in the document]
hsliuustc0106: Fixed in `12cbcd42` by making the supported V1 contract precise. The option is now `coordination_timeout_seconds` and bounds lock acquisition only; it does not claim to cancel an in-process producer or atomic publication. Removed the misleading post-production timeout and added slow-producer coverage and documentation.

vllm_omni/model_executor/host_weight_runtime/filesystem/store.py
```diff
@@ -0,0 +1,1096 @@
...
+            return StoreResult(StoreStatus.HIT, lease=lease)
+        except HostWeightError as exc:
+            try:
+                if record_invalid:
+                    with contextlib.suppress(OSError):
+                        self._write_deny(identity.key, exc.failure)
+            finally:
+                artifact_lock.close()
+            return StoreResult(StoreStatus.INVALID, failure=exc.failure)
+        except (OSError, ValueError, ManifestValidationError) as exc:
```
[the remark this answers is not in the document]
hsliuustc0106: Fixed in `12cbcd42`. Lookup now catches the specific `safetensors.SafetensorError`, maps it to typed `INVALID`/`MANIFEST_CORRUPT`, and records the deny marker under normal lookup. Added a malformed-header regression.

vllm_omni/model_executor/host_weight_runtime/protocols.py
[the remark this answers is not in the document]
hsliuustc0106: Fixed in `12cbcd42`. Restore planning is validation-only, `commit() -> None` is one-shot and is the sole mutating phase, and fallback after commit failure requires a fresh model. Transport and GPU registration remain separate.

vllm_omni/model_executor/host_weight_runtime/filesystem/store.py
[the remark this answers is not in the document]
hsliuustc0106: Fixed in `12cbcd42`. Filesystem type is preserved and checked against an explicit local allowlist; known remote and unknown filesystems are rejected, and the verified type is persisted in domain metadata/inspection. Added remote/unknown filesystem regressions.

vllm_omni/host_weight_runtime/runtime.py
```diff
@@ -0,0 +1,315 @@
...
+    ) -> ResolutionAction:
+        if self.config.remote.on_local_miss is not RemoteOnMiss.DISABLED:
+            return ResolutionAction.TRY_REMOTE
+        if self._producer_allowed(producer):
+            if result.status is StoreStatus.INVALID:
+                return ResolutionAction.QUARANTINE_THEN_PRODUCE
+            return ResolutionAction.TRY_PRODUCER
+        return self._fallback_action()
+
+    def _producer_allowed(self, producer: WeightProducer | None) -> bool:
```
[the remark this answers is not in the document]
hsliuustc0106: Fixed in `12cbcd42` by rejecting the unsupported setting rather than adding a partial post-load path. `ProductionPolicy(allow_post_load_publish=True)` now fails configuration validation explicitly; V1 remains limited to `PRE_LOAD_SAFE` producers, with contract and documentation coverage.

vllm_omni/host_weight_runtime/filesystem/store.py
```diff
@@ -0,0 +1,1096 @@
...
+            manifest_digest = hashlib.sha256(manifest.canonical_bytes).hexdigest()
+            if marker.manifest_sha256 != manifest_digest:
+                raise HostWeightError(
+                    _failure(ResolutionStage.VALIDATION, FailureCode.CHECKSUM_MISMATCH, "manifest digest mismatch")
+                )
+            if marker.artifact_content_sha256 != manifest.artifact_content_sha256:
+                raise HostWeightError(
+                    _failure(ResolutionStage.VALIDATION, FailureCode.CHECKSUM_MISMATCH, "content digest mismatch")
+                )
+            if manifest.artifact_key != identity.key or manifest.identity.canonical_bytes != identity.canonical_bytes:
```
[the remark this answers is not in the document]
hsliuustc0106: Fixed in `12cbcd42`. Lease acquisition now compares both manifest schemas with the exact producer/restorer schemas embedded in the requested identity and returns `MANIFEST_INCOMPATIBLE` on mismatch. Added parametrized producer- and restorer-schema regressions.

PR #6422, opened 2026-08-20
[BugFix][Higgs-Audio-V3] Fix transcript and concurrent sampling tests

### LINKED ISSUE #6418: [Bug]: higgs-audio v3 /v1/audio/speech Whisper similarity fails (max_new_tokens truncation, concurrent extra speech, inline control tags)
### Your current environment

<details>

```text
CI / L4 online serving job (not a local collect_env.py dump).

Model: bosonai/higgs-audio-v3-tts-4b
Deploy: vllm_omni/deploy/higgs_multimodal_qwen3.yaml
Test: tests/e2e/online_serving/test_higgs_audio_v3.py
Gate: assert_audio_speech_response Whisper transcript vs request input, threshold=0.9
  (tests/helpers/assertions.py::_assert_transcript_matches)
```

</details>

### Your code version

<details>

```text
CI image used by the TTS online-serving job.
```

</details>
<details>

```text
tests/e2e/online_serving/test_higgs_audio_v3.py against bosonai/higgs-audio-v3-tts-4b
```

</details>

### 🐛 Describe the bug

Several `/v1/audio/speech` cases in `test_higgs_audio_v3.py` fail the **full_model / advanced_model** Whisper similarity gate (`threshold=0.9`). The WAV/PCM request itself succeeds; the spoken content does not match `request_config["input"]`.

Not a hang or EngineDeadError. Neighboring cases in the same file (`test_plain_text_wav`, streaming, voice-clone) were not in this failure set.

#### Failing cases

| Test | Input (abbrev.) | Whisper transcript | similarity |
| --- | --- | --- | --- |
| `TestHiggsAudioV3OnlineHappyPath::test_plain_text_with_max_new_tokens` | `Innovation distinguishes between a leader and a follower.` + `max_new_tokens=32` | `innovation.` | **0.11** |
| `TestHiggsAudioV3OnlineHappyPath::test_concurrent_plain_text` | `it was the night before my birthday` (`request_num=3`) | `It was the night before my birthday. In this pal, how this acti...` | **0.44** |
| `test_inline_emotion_and_expressive` | `<\|emotion:amusement\|><\|prosody:expressive_high\|>Wait, that was actually hilarious.` | `Wait, that was actually hilarious.` | **0.42** |
| `test_inline_style_whispering` | `<\|style:whispering\|>It is just between you and me, alright?` | `It is just between you and me, alright?` | **0.70** |
| `test_inline_prosody_speed_and_pitch` | slow+low-pitch radar sentence | `The radar shows a storm approaching from the east.` | **0.56** |
| `test_inline_pause_mid_text` | `Hold on a moment <\|prosody:pause\|> let me think about it.` | `Hold on a moment. Let me think about it.` | **0.71** |
| `test_inline_sfx_with_onomatopoeia` | emotion + `<\|sfx:laughter\|>Hehe, ...` | `I cannot believe that just happened. I am still recovering from...` | **0.64** |

#### Two distinct problems (same gate)

1. **Truncation / extra speech (model or serving)**
   - `max_new_tokens=32` on a long sentence yields only `"innovation."` — the talker is likely hitting the token cap before the sentence finishes.
   - Concurrent `request_num=3` prepends the right sentence then appends unrelated speech (`In this pal, how this acti...`), so per-slot audio state / batched decode may be leaking or continuing past EOS.

2. **Inline control tokens vs Whisper (test contract)**
   - The class docstring says these cases only need a non-trivial WAV (`min_audio_bytes`); they do **not** claim audio-quality / transcript fidelity.
   - `assert_audio_speech_response` still compares Whisper against the **raw** `input`, including `<|emotion:...|>` / `<|style:...|>` / `<|sfx:...|>` tags. Whisper transcribes the spoken words only, so cosine vs tagged text is systematically below 0.9 even when the spoken sentence is correct (see whispering 0.70 and pause 0.71).

Until (1) is fixed and/or the speech assertion strips control tokens / uses spoken-only expected text, these seven cases should stay skipped so L4 TTS does not red on a 0.9 transcript gate.

#### Repro

```bash
cd tests
pytest -s -v e2e/online_serving/test_higgs_audio_v3.py \
  -m "full_model and tts" --run-level=full_model
```

#### Expected

- `max_new_tokens` still produces a complete-enough utterance, or the test uses a budget that matches sentence length.
- Concurrent requests do not append foreign speech.
- Inline-control tests either skip the Whisper-vs-raw-input gate or compare against the spoken text with tags stripped.

akshatvishu:
## Purpose

Closes #6418.

Seven Higgs Audio V3 speech tests failed the Whisper similarity gate for three separate reasons:

1. Inline control-token requests were compared with the raw request syntax instead of the spoken text.
2. The sentence test limited generation to 32 tokens and produced only `innovation.`.
3. The audio-code sampler ignored vLLM's request generators, so seeded output depended on batch composition. The decode delay state was also stored only by temporary batch row.

This change:

- adds `transcript_expected_text` for tests whose request syntax differs from their spoken reference;
- raises the sentence test budget to 1024 tokens;
- samples each request's eight audio codebooks with `SamplingMetadata.generators`;
- keeps the GPU decode delay state associated with request IDs when vLLM condenses the batch;
- enables the Higgs V3 online tests on the ROCm `MI325` pytest resource and selects `TRITON_ATTN` for ROCm.

The transcript field defaults to `input`, so existing speech tests keep their current behavior.

## Reproduction

```bash
HIP_VISIBLE_DEVICES=0 pytest -s -vv \
  tests/e2e/online_serving/test_higgs_audio_v3.py::TestHiggsAudioV3OnlineHappyPath::test_concurrent_plain_text \
  --run-level advanced_model
```

Before the seeded sampler fix, one of three identical concurrent requests appended unrelated speech and scored 0.247 against the expected transcript.

## Root cause

vLLM creates one `torch.Generator` for every explicitly seeded request and moves that generator when its batch row changes. Higgs flattened each request into eight codebook rows, then called `torch.multinomial` without those generators. Identical seeded requests therefore consumed a shared random stream according to the current batch shape.

The shared transcript assertion also treated model control tokens as spoken text, and the 32-token test budget was too small for its sentence.

## Test Plan

**vLLM Version:** 0.27.0 in the MI300X test image

**vLLM-Omni Commit:** `68a3e2f0fbd73cbb8d2fda1533bed2f6ce6ccdec`

Local CPU regression checks:

```bash
.venv/bin/pytest -o addopts='' -q \
  tests/helpers/tests/test_assertions.py \
  tests/config/test_config_factory.py::TestPlatformOverrides::test_higgs_audio_v3_rocm_uses_triton_attention \
  tests/worker/test_omni_gpu_model_runner.py::test_model_forward_passes_request_ids_to_decode_metadata \
  tests/model_executor/models/higgs_audio_v3/test_higgs_audio_v3.py::TestSamplerMethods \
  --run-level core_model
```

Result: `30 passed, 14 warnings in 2.54s`.

The seeded sampler regression fails against the preceding branch commit and passes with this change. It checks identical requests in one batch and generator continuity after row condensation.

## Test Result

Test environment: one AMD MI300X with 192 GB VRAM, ROCm 7.2.3, amdgpu 6.19.14, Python 3.12.13, pytest 9.1.1, and `TRITON_ATTN`.

The targeted concurrent test passed in 101.10 seconds:

```text
audio content is: It was the night before my birthday.
Cosine similarity: 1.000
audio content is: It was the night before my birthday.
Cosine similarity: 1.000
audio content is: It was the night before my birthday.
Cosine similarity: 1.000

1 passed, 14 warnings in 101.10s
```

This WAV test ran Whisper for all three responses. The warnings were non-fatal import, deprecation, and ROCm environment warnings.


akshatvishu:
@vllm-omni-review-bot

vllm_omni/model_executor/models/higgs_audio_v3/higgs_audio_v3_talker.py
```diff
@@ -355,6 +356,91 @@ def update_decode_step_metadata(
...
+        self._state_pool_delay_count = resize(self._state_pool_delay_count, 0)
+        self._state_pool_eoc_countdown = resize(self._state_pool_eoc_countdown, -1)
+        self._state_pool_generation_done = resize(self._state_pool_generation_done, False)
+
+    def _sync_decode_state_with_batch(self, req_ids: list[str]) -> None:
+        device = self._decode_has_codes.device
+        self._ensure_state_pool_capacity(len(req_ids), device)
+        previous_req_ids = getattr(self, "_last_batch_req_ids", [])
+
+        if previous_req_ids != req_ids:
```
linyueqian: Force synchronization when a current request ID no longer has a pool entry. The runner explicitly supports a finished request and a newly scheduled request sharing the same ID in one scheduler step. In that case `on_requests_finished` removes the old pool entry, but the old and new `req_ids` lists can both be `["X"]`, so this condition is false and the new request retains the finished request's active decode tensors. The pool entry is also never recreated while that batch remains unchanged. Invalidating `_last_batch_req_ids` during cleanup, or including missing current IDs in this condition, would reset the new request correctly. Please add a regression where the finished and replacement requests use the same ID.

linyueqian (changes requested):
Reviewed at `68a3e2f0` (whole footprint). The transcript expectations, larger generation budget, ROCm configuration, and request-local sampling generators match the stated root causes. However, the request-keyed decode state still mishandles a lifecycle case that the runner explicitly supports, so this needs another revision.

Validation: static reads only, no code from this head executed. I verified the finding's two premises independently: the runner's own comment in `_update_states` documents the aborted-and-resubmitted-same-ID case as two distinct requests, and the talker's `_sync_decode_state_with_batch` gates on list equality, which is exactly the case that makes both lists identical.

vllm_omni/model_executor/models/higgs_audio_v3/higgs_audio_v3_talker.py
akshatvishu: Thanks, fixed. The talker now re-syncs when a current request has no pool entry, even if the ID list is unchanged. Added a same ID regression and ran the targeted tests, all passed.

linyueqian (approved):
The round 1 finding is fixed and the fix now has runtime evidence behind it. _sync_decode_state_with_batch re-syncs when a current request id lacks a pool entry, so a finished-then-reused same-id request gets fresh decode state even though previous_req_ids equals req_ids, and the regression test mirrors that exact scenario.

E2E on an H200 under vLLM 0.28.0: the unit tier passed 118 tests (higgs talker, gpu runner, helpers, including the same-id regression), and the online serving tier passed 11 of 12, covering concurrent plain text, concurrent PCM streaming, inline control tokens with transcript assertions, and the references-alias voice clone, which is exactly the concurrent-sampling surface this PR touches.

The one serving failure is not this PR. test_voice_clone_ref_audio_ref_text fails with "Token id -100 is out of vocabulary" (VLLMValidationError raised at add_request), and it fails identically at the merge base c50b7de1 on the same machine, same venv, same error string, deterministic across two runs at the head. The ref_audio+ref_text path is broken on main independently of this change; the references-alias clone path passes on the head. Worth its own issue, and happy to file it with the repro.

Approving. The transcript and decode-state fixes do what they claim, the regression coverage is genuine, and every red signal in the run traces to main.

Issue #5631, opened 2026-07-31
[RFC]: Production-grade NVIDIA Blackwell support in vLLM-Omni


hsliuustc0106:
## Motivation.

vLLM-Omni already has substantial NVIDIA Blackwell enablement, but it does not yet have a precise, continuously validated support contract.

Today:

- diffusion attention recognizes `sm_100`, `sm_103`, `sm_120`, and `sm_121`;
- datacenter Blackwell can use TRTLLM diffusion attention, while cuDNN, FlashInfer, or SDPA provide broader fallback paths;
- upstream vLLM supplies the main autoregressive CUDA foundation; and
- selected FP8, INT8, ModelOpt NVFP4, SageAttention, and Quack paths exist.

The remaining gaps are correctness and qualification:

- importability does not prove that an optional extension contains a kernel for the active GPU. This caused the B200 ring-attention failure in #5611; #5617 adds an architecture-aware SDPA fallback;
- current CUDA CI resources cover L4 and H100, not Blackwell. #5543 starts adding B200 selection, but there is no `sm_120` lane or architecture-specific test contract;
- the Qwen3-Omni NVFP4 test explicitly does not exercise the native Blackwell path; and
- datacenter `sm_10x` and workstation/consumer `sm_12x` intentionally support different optimized providers, so testing B200 alone cannot justify a blanket “Blackwell supported” statement.

This RFC defines support as a testable production baseline rather than full optimized-kernel parity.

### Scope

Initial architecture scope:

| Family | Representative target | Validation intent |
| --- | --- | --- |
| `sm_100` | B200/GB200 | Continuous |
| `sm_103` | B300/GB300 | Periodic or release qualification |
| `sm_120` | RTX PRO 6000 or equivalent | Continuous |
| `sm_121` | Available runner | Qualification when hardware is available |

This does not automatically extend support to unlisted or future Blackwell variants.

### Support levels

- **Baseline supported:** compatible installation; safe kernel selection; explicit unsupported backends fail early; documented fallbacks are correct; representative eager, compiled, multi-GPU, and quantization paths pass recurring real-hardware CI.
- **Optimized:** an intentional accelerated provider is selected, with architecture-local accuracy and performance evidence.
- **Experimental:** opt-in and non-release-blocking, with documented limitations.

[...]

PR #6431, opened 2026-08-21
[Bugfix][Attention] Gate SAGE_ATTN_3 on the architectures it builds for

linzhenpl07:
## Purpose

The `SAGE_ATTN_3` selection guard admits any GPU with `compute_capability.major >= 10`, which is broader than the set of architectures the kernel is compiled for.

`sageattention3_blackwell/setup.py` matches `(major, minor)` exactly and raises `"Unsupported GPU"` otherwise, emitting one gencode per supported capability:

```text
(10, 0) -> -gencode arch=compute_100a,code=sm_100a     B200 / GB200
(12, 0) -> -gencode arch=compute_120a,code=sm_120a     RTX PRO 6000, RTX 50-series
(12, 1) -> -gencode arch=compute_121a,code=sm_121a     consumer Blackwell refresh
otherwise -> "Unsupported GPU"
```

`sm_103` (B300 / GB300) is not among them. Under the current guard it passes platform routing, the `sageattn3` import can succeed, and the failure only surfaces inside the kernel: a native `sm_103a` build reaches CUTLASS invalid-control-path guards because SM103 requires the tcgen05 implementation rather than the SM120 warp-level `mma.sync` path. `major >= 10` also admits hypothetical 11.x parts the kernel has never been built for.

This was reported by @mo-ke-ke while qualifying packed `SAGE_ATTN_3` on an 8x B300 node in #5700.

This PR gates on the exact `{(10,0), (12,0), (12,1)}` set instead, and names both the detected architecture and the supported list in the fallback warning, so an unsupported GPU degrades to `TORCH_SDPA` at selection time with an actionable message rather than failing later inside the kernel.

`TRTLLM_ATTN` immediately below already gates precisely on SM100/SM103 and raises a specific error, so this removes an inconsistency in the same function rather than introducing a new policy. It also moves one step toward the single internal CUDA architecture classification proposed in #5631, which notes that datacenter `sm_10x` and workstation/consumer `sm_12x` intentionally support different optimized providers.

Docs gain the per-capability support table, since "a Blackwell-class GPU" is not a sufficient description of the requirement.

## Test Plan

```bash
pytest tests/diffusion/attention/ -q
```

`tests/diffusion/attention/test_sage_attn3.py` is extended with two parametrized tests:

- supported architectures route to the backend: `(10, 0)` (pre-existing case), `(12, 0)`, `(12, 1)`;
- unsupported architectures fall back to `TORCH_SDPA`: `(9, 0)`, `(8, 9)`, `(10, 3)`, `(11, 0)`.

In the negative cases `sageattn3` is monkeypatched to be importable, so the assertion isolates the architecture gate from package availability. The pre-existing negative test used `(9, 0)` with the package absent, which conflated the two conditions and could not have caught this.

**vLLM Version:** 0.26.0

**vLLM-Omni Commit:** 73b623f2f7db092053c1c86fe796bed89eb3dc71

## Test Result

Target file:

```text
tests/diffusion/attention/test_sage_attn3.py
9 passed in 0.12s
```

Regression check — reverting only the gate to `major >= 10` and keeping the new tests:

```text
.......FF                                                        [100%]
FAILED test_cuda_platform_falls_back_when_sage_attn3_gpu_is_unsupported[sm_103-b300-not-a-build-target]
FAILED test_cuda_platform_falls_back_when_sage_attn3_gpu_is_unsupported[sm_110-unknown-future-variant]
2 failed, 7 passed
```

Exactly the two cases this change targets fail under the old guard; the other seven still pass. The assertion message is that `SageAttention3Backend` was returned where `SDPABackend` was expected.

Full attention directory:

```text
tests/diffusion/attention/
6 failed, 188 passed, 16 skipped in 33.95s
```

The 6 failures are pre-existing on this machine and unrelated to this change. On `73b623f2` without this patch, the same two files produce the same 6 failures:

```text
tests/diffusion/attention/test_kernels_hub.py tests/diffusion/attention/test_trtllm_attn.py
6 failed, 30 passed, 4 skipped in 1.86s
```

They are environment limitations: `test_kernels_hub_execution` needs the HuggingFace `kernels` package, and the five `test_trtllm_attn.py` cases need a datacenter Blackwell GPU with flashinfer trtllm-gen. The test host is `sm_120`.

No GPU-specific behavior is exercised by the new tests — `get_device_capability` is monkeypatched throughout, so they run anywhere the CUDA platform is selected.

### DIFF
--- a/docs/user_guide/diffusion/attention_backends/sage.md
+++ b/docs/user_guide/diffusion/attention_backends/sage.md
@@ -54,8 +54,22 @@ python -c "import sageattn3; print(sageattn3.__file__)"
 vllm-omni serve <model> --diffusion-attention-backend SAGE_ATTN_3
 ```
 
-`SAGE_ATTN_3` requires CUDA, an importable `sageattn3`, and a Blackwell-class
-GPU. Its kernel assumes the query-head count equals the key/value-head count.
+`SAGE_ATTN_3` requires CUDA, an importable `sageattn3`, and one of the
+architectures the kernel is built for. `sageattention3_blackwell/setup.py`
+emits one gencode per supported compute capability and rejects anything else:
+
+| Compute capability | Gencode | Example GPUs |
+| --- | --- | --- |
+| 10.0 | `sm_100a` | B200, GB200 |
+| 12.0 | `sm_120a` | RTX PRO 6000, RTX 50-series |
+| 12.1 | `sm_121a` | Consumer Blackwell refresh |
+
+Not every Blackwell GPU qualifies. `sm_103` (B300 / GB300) is absent by
+design: the kernel takes the SM120 warp-level `mma.sync` path, while SM103
+requires the tcgen05 implementation. Selecting `SAGE_ATTN_3` on an
+unsupported GPU logs a warning and falls back to `TORCH_SDPA`.
+
+Its kernel also assumes the query-head count equals the key/value-head count.
 GQA and MQA diffusion calls therefore fall back to PyTorch SDPA for
 correctness.
 
--- a/tests/diffusion/attention/test_sage_attn3.py
+++ b/tests/diffusion/attention/test_sage_attn3.py
@@ -115,19 +115,80 @@ def test_cuda_platform_selects_sage_attn3_alias(monkeypatch: pytest.MonkeyPatch)
 
 
 @pytest.mark.skipif(not current_platform.is_cuda(), reason="sage_attn3 tests require CUDA platform")
-def test_cuda_platform_falls_back_when_sage_attn3_gpu_is_unsupported(monkeypatch: pytest.MonkeyPatch):
+@pytest.mark.parametrize(
+    "capability",
+    [
+        pytest.param((12, 0), id="sm_120-rtx-pro-6000-and-rtx-50-series"),
+        pytest.param((12, 1), id="sm_121-consumer-blackwell-refresh"),
+    ],
+)
+def test_cuda_platform_selects_sage_attn3_on_supported_blackwell(
+    monkeypatch: pytest.MonkeyPatch, capability: tuple[int, int]
+):
+    """Every architecture sageattention3_blackwell builds for must route to the backend."""
     from vllm.platforms.interface import DeviceCapability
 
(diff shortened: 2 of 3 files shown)


linzhenpl07:
Self-review.

**Scope.** 3 files, +104/-7. One behavioral change: the `SAGE_ATTN_3` architecture guard in `get_diffusion_attn_backend_cls`. No other backend's selection path is touched, and the default-selection path below is unchanged — `SAGE_ATTN_3` is only reachable through explicit `--diffusion-attention-backend`, so nothing auto-routes differently.

**What I verified.**

- The supported set comes from upstream `sageattention3_blackwell/setup.py`, not from inference. I quoted the gencode mapping in the code comment so a future reader can re-check it against the source rather than trusting the constant.
- The new tests fail on the old guard and pass on the new one, so they actually cover the defect. Reverting only the gate produces exactly 2 failures — `sm_103` and `sm_110` — with the other 7 cases green.
- The 6 failures in the wider attention directory reproduce identically on `73b623f2` without this patch. I ran that comparison specifically to avoid attributing environment gaps to this change.
- I deliberately made `sageattn3` importable in the negative tests. The pre-existing negative case used `(9, 0)` with the package missing, so a passing result did not prove the architecture gate was what rejected it.

**Behavior change worth flagging for review.** A user on `sm_103` who previously got `SAGE_ATTN_3` selected — and then a kernel-level failure — now gets `TORCH_SDPA` plus a warning. That is the intent, but it is a silent fallback rather than a hard error. I chose fallback over raising to match the existing `SAGE_ATTN_3` branch, which already falls back when the package is missing. `TRTLLM_ATTN` raises instead. If maintainers prefer consistency with `TRTLLM_ATTN` here, raising is a small change and I am happy to make it.

**Not covered.** I have no `sm_100`/`sm_103` hardware, so the runtime behavior on datacenter Blackwell is not verified end to end — only the selection logic, which is monkeypatched. The `sm_103` claim rests on @mo-ke-ke's 8x B300 report in #5700 and on `setup.py` itself.

**Conventions checked.** PR title prefix, SPDX headers, no new `torch.cuda` call sites, no forbidden imports, no new broad `except`, no `Any` hints, no new model-specific example, test file carries CI level and hardware marks, markdown tables surrounded by blank lines, all lines within the 120-char ruff limit.

PR #6434, opened 2026-08-21
[CI/Build] Add omni serving benchmark for unified understanding+generation models

### LINKED ISSUE #6237: [RFC]: DTPS (DiT-load-aware Type-Priority Scheduling) for AR+DiT Unified Multimodal Comprehension/Generation Serving

yibeicanghai:
## Summary

Adds a serving benchmark tool (`benchmarks/omni/`) that evaluates mixed i2t (understanding) / t2i (generation) / it2i (editing) concurrent workloads on unified AR+DiT models (e.g. HunyuanImage-3.0). This benchmark fills the gap left by `benchmarks/diffusion/` (which only covers pure generation tasks) by testing the mixed understanding+generation concurrency scenario that the DTPS scheduler is designed to optimize.

## Motivation

Currently there is no benchmark for unified understanding+generation models under mixed task concurrency. This tool is designed to measure DTPS scheduling strategy effectiveness and evaluate scheduling optimizations under realistic mixed i2t/t2i/it2i workloads.

Closes #6237
Related feature PR: #6315 (DTPS DiT-load-aware Type-Priority Scheduling)

## Key Features

- **Configurable task mix ratios** (`--num-i2t`/`--num-t2i`/`--num-it2i`) — e.g. 14:2:4 for 7:1:2 ratio
- **it2i bot_task weighted sampling** (think/recaption/think_recaption) with proportional or random allocation
- **True random send order** with reproducible seed — interleaves task types instead of sending in blocks
- **Per-task / per-bot_task / per-stage (AR/DiT) statistics** — overall + per-task + it2i-by-bot_task three-layer bucketing
- **Streaming text-output TTFT/TPOT/ITL** (vLLM-style) — i2t AR answer text and it2i AR thinking text latency
- **Input randomization** (`--randomize-input`) — diverse prompt lengths, input image resolutions, and content per request
- **Generation resolution weighted allocation** — mix 512x512 / 1024x1024 / 1280x720 outputs
- **Client dispatch order audit trail** — distinguishes client send order from server-side DTPS re-dispatch

## Files

| File | Description |
|------|-------------|
| `omni_benchmark_serving.py` | Main entry point — CLI parsing, request dispatch, metrics aggregation, JSON output |
| `omni_backends.py` | Async HTTP senders (chat stream/non-stream, images-edits stream/non-stream) + output parsing |
| `omni_dataset.py` | Request list construction — task mix, bot_task sampling, resolution allocation, input randomization |
| `_test_randomize_cli.py` | CLI-level dry-run test for --randomize-input |
| `_test_randomize_input.py` | Unit tests for dataset construction logic |
| `run_benchmark_loop.sh` | Multi-round benchmark loop helper |
| `README.md` | Full documentation with usage examples |
| `__init__.py` | Package init |

## Quick Smoke

```bash
# Dry-run: print the request plan without sending
python benchmarks/omni/omni_benchmark_serving.py \
    --host 127.0.0.1 --port 8099 \
    --num-i2t 14 --num-t2i 2 --num-it2i 4 \
    --it2i-bot-task-weights "recaption=2,think=1,think_recaption=1" \
    --shuffle --seed 3 --dry-run
```

## Naming Rationale

The directory is named `omni` (not `mixed`) to align with the codebase's existing terminology (`vllm_omni`, `--omni` serving flag, `OmniARScheduler`) and to follow the capability-based naming convention used by sibling benchmark directories (`diffusion/`, `tts/`). The name reflects that this benchmark applies to any unified understanding+generation model, not just HunyuanImage-3.0.

## Pre-check

- [x] PR title follows `[CI/Build]` convention
- [x] All pre-commit hooks pass (ruff check, ruff format, typos, DCO)
- [x] No model-specific Python examples under `examples/`
- [x] No dead code or unused imports
- [x] No event-loop blocking or hot-path copies

### DIFF
--- a/benchmarks/omni/omni_backends.py
+++ b/benchmarks/omni/omni_backends.py
@@ -1,0 +1,803 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM-Omni project
+"""Request I/O types and async request functions for the omni serving benchmark.
+
+Reuses ``diffusion/backends.py`` for the base ``RequestFuncInput`` /
+``RequestFuncOutput`` shapes and small image-encoding helpers, but keeps all
+omni-specific state and senders here so the diffusion benchmark stays
+unmodified:
+
+* ``OmniRequest`` — a ``RequestFuncInput`` subclass carrying the client-side
+  task label (``i2t`` / ``t2i`` / ``it2i``), the sampled ``bot_task``, and the
+  send-time image caches (``image_data_urls`` / ``image_bytes``). None of these
+  are sent to the server; the server resolves the task type itself from
+  ``modalities`` plus the presence of a reference image (see
+  ``serving_chat._resolve_omni_task_type``).
+* ``OmniRequestFuncOutput`` — a ``RequestFuncOutput`` subclass adding the
+  text-streaming metrics (``ttft`` / ``itl`` / ``output_tokens`` /
+  ``generated_text``) captured by the streaming senders; non-stream senders
+  leave them at their defaults.
+* ``async_request_omni_chat`` — a thin wrapper over
+  ``async_request_chat_completions`` that lifts ``modalities`` (and a few
+  generation knobs) from ``extra_body`` into the JSON payload so the same
+  ``/v1/chat/completions`` endpoint can carry i2t (``modalities=["text"]``) and
+  t2i/it2i (``modalities=["image"]``) traffic, and that parses stage metrics
+  from both the image-output (``content`` is a list) and text-output
+  (``content`` is a string) response shapes.
+* ``async_request_image_edits`` — an omni /v1/images/edits (non-stream) sender
+  (the diffusion one is not imported) that reuses the shared form builder.
+"""
+
+from __future__ import annotations
+
+import base64
+import json
+import os
+import sys
+import time
+from collections.abc import AsyncGenerator
+from dataclasses import dataclass, field
+from typing import Any
+
+import aiohttp
+from tqdm import tqdm
+
+# ``backends`` lives in the sibling ``diffusion`` benchmark directory. Make that
+# directory importable regardless of the caller's CWD.
+_DIFFUSION_DIR = os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__))), "diffusion")
+if _DIFFUSION_DIR not in sys.path:
+    sys.path.insert(0, _DIFFUSION_DIR)
+
+from backends import (  # noqa: E402  (import after sys.path tweak)
+    DEFAULT_EDITS_BOT_TASK,
+    RequestFuncInput,
+    RequestFuncOutput,
+    _encode_image_as_data_url,
+    _guess_mime_type,
+)
+
+# Top-level request fields the OpenAI chat endpoint understands and that the
+# omni benchmark drives through ``extra_body``.
+_CHAT_TOPLEVEL_FROM_EXTRA = ("modalities", "max_tokens", "temperature", "top_p", "seed")
+
+# Tasks the it2i bot_task can take on. Keep in sync with
+# hunyuan_image3/prompt_utils._BOT_TASK_PRESETS (minus ``vanilla``/``None`` which
+# are plain-mode and not part of the "thinking intensity" sweep).
+IT2I_BOT_TASKS = ("recaption", "think", "think_recaption")
+
+
+@dataclass
+class OmniRequest(RequestFuncInput):
+    """A chat/edits request tagged with its client-side task type.
+
+    ``task_type`` and ``bot_task`` are benchmark-side bookkeeping only — they
+    drive per-task / per-bot_task statistics and never enter the HTTP payload.
+    The server classifies the request from ``modalities`` (+ reference image).
+
+    ``image_data_urls`` / ``image_bytes`` are send-time caches populated by
+    ``prepare_request_images`` (pre-read + pre-encode the input images once) so
+    the senders skip file I/O and base64 under concurrency. They live here on
+    the omni subclass — the diffusion ``RequestFuncInput`` is left untouched.
+    """
+
+    task_type: str = ""
+    bot_task: str | None = None
+    # Bookkeeping only (never sent). Under --randomize-input this records the
+    # generated input image's "WxH" so the tally / JSON can show the realized
+    # input-resolution distribution without re-reading the image file.
+    input_image_size: str | None = None
+    # Send-time caches (never sent). Populated by prepare_request_images.
+    image_data_urls: list[str] | None = None
+    image_bytes: list[bytes] | None = None
+
+
+@dataclass
+class OmniRequestFuncOutput(RequestFuncOutput):
+    """Omni-benchmark output — adds text-streaming metrics on top of the
+    diffusion ``RequestFuncOutput``.
+
+    ``ttft`` / ``itl`` / ``output_tokens`` / ``generated_text`` are only set by
+    the streaming senders (AR-text TTFT/ITL); non-stream senders leave the
+    defaults so the aggregate stats can read them uniformly (``o.ttft <= 0.0``
+    marks a request that produced no AR text). Kept on this subclass so the
+    diffusion ``RequestFuncOutput`` is left untouched.
+    """
+
+    ttft: float = 0.0
+    itl: list[float] = field(default_factory=list)
+    output_tokens: int = 0
+    generated_text: str = ""
+
+
+def prepare_request_images(req: OmniRequest) -> None:
+    """Pre-read + pre-encode ``req``'s input images into on-request caches.
+
+    Run once in a prep pass (outside the timed send) so the senders skip file
+    I/O and base64 at send time. This keeps the event loop from blocking under
+    concurrency and — together with ordered dispatch in the benchmark driver —
+    makes the actual send order match the dry-run order exactly, instead of
+    being raced by per-task prep cost: t2i has no image, it2i builds a multipart
+    form, i2t base64-encodes a data URL, so without pre-encoding the
+    fastest-prep group reaches the server first regardless of list order.
+    Idempotent; a no-op for requests without input images.
+    """
+    if not req.image_paths:
+        return
+    if req.image_data_urls is not None and req.image_bytes is not None:
+        return
+    urls: list[str] = []
+    raws: list[bytes] = []
+    for p in req.image_paths:
+        if not os.path.exists(p):
+            raise FileNotFoundError(p)
+        with open(p, "rb") as f:
+            raw = f.read()
+        raws.append(raw)
+        mime = _guess_mime_type(p)
+        urls.append(f"data:{mime};base64,{base64.b64encode(raw).decode('utf-8')}")
+    if req.image_data_urls is None:
+        req.image_data_urls = urls
+    if req.image_bytes is None:
+        req.image_bytes = raws
+
+
+def _parse_chat_stage_metrics(resp_json: dict[str, Any]) -> tuple[dict[str, float], float]:
+    """Extract (stage_durations, peak_memory_mb) from a chat-completions body.
+
+    Image-output responses put them on ``choices[0].message.content[0]``; the
+    text-output path (i2t) puts them on the root-level ``metrics`` object.
+    """
+    stage_durations: dict[str, float] = {}
+    peak_memory_mb = 0.0
+
+    choices = resp_json.get("choices", [])
+    if isinstance(choices, list) and choices:
+        content = (choices[0] or {}).get("message", {}).get("content")
+        if isinstance(content, list) and content:
+            first = content[0]
+            if isinstance(first, dict):
+                sd = first.get("stage_durations")
+                if isinstance(sd, dict):
+                    stage_durations = {str(k): float(v) for k, v in sd.items()}
+                pm = first.get("peak_memory_mb")
+                if isinstance(pm, (int, float)):
+                    peak_memory_mb = float(pm)
+
+    if not stage_durations or peak_memory_mb == 0.0:
+        metrics = resp_json.get("metrics")
+        if isinstance(metrics, dict):
+            if not stage_durations:
+                sd = metrics.get("stage_durations")
+                if isinstance(sd, dict):
+                    stage_durations = {str(k): float(v) for k, v in sd.items()}
+            if peak_memory_mb == 0.0:
+                pm = metrics.get("peak_memory_mb")
+                if isinstance(pm, (int, float)):
+                    peak_memory_mb = float(pm)
+
+    return stage_durations, peak_memory_mb
+
+
+def _first_message_content(resp_json: dict[str, Any]) -> Any:
+    """Return ``choices[0].message.content`` (str for i2t, list for t2i/it2i) or None."""
+    choices = resp_json.get("choices", [])
+    if isinstance(choices, list) and choices:
+        return (choices[0] or {}).get("message", {}).get("content")
+    return None
+
+
+def extract_chat_outputs(resp_json: dict[str, Any]) -> tuple[str | None, str | None, int]:
+    """Extract (returned_text, cot_output, num_images) from a chat-completions body.
+
+    * i2t (text output): ``content`` is a string -> the model's answer.
+    * t2i / it2i (image output): ``content`` is a list of ``image_url`` items;
+      returned_text is None and num_images counts the image items.
+
+    ``cot_output`` (the AR-stage "thinking" text) is NOT exposed by the current
+    chat image response, so it is returned as None on this path. The lookup is
+    kept tolerant: if a future server change surfaces it on a content item it
+    will be picked up without a benchmark change. Use the /v1/images/edits
+    endpoint (see :func:`extract_edits_outputs`) to obtain the AR thinking text
+    today — its response carries ``cot_output`` at the root.
+    """
+    content = _first_message_content(resp_json)
+    if isinstance(content, str):
+        return content or None, None, 0
+    if isinstance(content, list):
+        cot_output: str | None = None
+        num_images = 0
+        for item in content:
+            if not isinstance(item, dict):
+                continue
+            if item.get("type") == "image_url":
+                num_images += 1
+            if cot_output is None:
+                co = item.get("cot_output")
+                if isinstance(co, str) and co.strip():
+                    cot_output = co
+        return None, cot_output, num_images
+    return None, None, 0
+
+
+def extract_edits_outputs(resp_json: dict[str, Any]) -> tuple[str | None, int]:
+    """Extract (cot_output, num_images) from a /v1/images/edits body.
+
+    The edits response is an ``ImageGenerationResponse``: ``data`` is a list of
+    ``{b64_json}`` image items and ``cot_output`` sits at the root — this is the
+    only endpoint that currently surfaces the AR-stage "thinking" text.
+    """
+    cot_output = resp_json.get("cot_output") if isinstance(resp_json, dict) else None
+    if not (isinstance(cot_output, str) and cot_output.strip()):
+        cot_output = None
+    data = resp_json.get("data", []) if isinstance(resp_json, dict) else []
+    num_images = sum(1 for d in data if isinstance(d, dict) and d.get("b64_json"))
+    return cot_output, num_images
+
+
+async def async_request_omni_chat(
+    input: OmniRequest,
+    session: aiohttp.ClientSession,
+    pbar: tqdm | None = None,
+) -> OmniRequestFuncOutput:
+    """POST /v1/chat/completions for any of i2t / t2i / it2i.
+
+    Mirrors ``async_request_chat_completions`` but hoists ``modalities`` (and a
+    few generation knobs) from ``extra_body`` to the payload root, and parses
+    stage metrics for both image- and text-output response shapes.
+    """
+    output = OmniRequestFuncOutput()
+    output.start_time = time.perf_counter()
+
+    extra_body = dict(input.extra_body)
+    modalities = extra_body.pop("modalities", None)
+
+    # Build the message content. Image-conditioned tasks (i2t, it2i) attach the
+    # input image; t2i is plain text.
+    content: list[dict[str, Any]] = []
+
+    if input.image_paths:
+        cached_urls = input.image_data_urls
+        for idx, img_path in enumerate(input.image_paths):
+            if cached_urls is not None and idx < len(cached_urls):
+                url = cached_urls[idx]
+            else:
+                if not os.path.exists(img_path):
+                    output.error = f"Image file not found: {img_path}"
+                    output.success = False
+                    if pbar:
+                        pbar.update(1)
+                    return output
+                url = _encode_image_as_data_url(img_path)
+            content.append(
+                {
+                    "type": "image_url",
+                    "image_url": {"url": url},
+                }
+            )
+
+    if input.prompt:
+        content.append({"type": "text", "text": input.prompt})
+
+    messages = [{"role": "user", "content": content}]
+
+    payload: dict[str, Any] = {
+        "model": input.model,
+        "messages": messages,
+        "max_tokens": 512,
+    }
+    if modalities is not None:
+        payload["modalities"] = modalities
+    if input.width and input.height:
+        payload["height"] = input.height
+        payload["width"] = input.width
+    if input.num_inference_steps is not None:
+        payload["num_inference_steps"] = input.num_inference_steps
+    if input.seed is not None:
+        payload["seed"] = input.seed
+
+    # Generation knobs the caller may have parked in extra_body.
+    for key in _CHAT_TOPLEVEL_FROM_EXTRA:
+        if key in extra_body and key not in payload:
+            payload[key] = extra_body[key]
+    # Anything else (e.g. negative_prompt, guidance_scale, bot_task for it2i)
+    # goes through extra_body as the diffusion path expects.
+    rest = {k: v for k, v in extra_body.items() if k not in _CHAT_TOPLEVEL_FROM_EXTRA}
+    if rest:
+        payload["extra_body"] = rest
+
+    try:
+        async with session.post(input.api_url, json=payload) as response:
+            if response.status == 200:
+                resp_json = await response.json()
+                output.response_body = resp_json
+                output.success = True
+                output.stage_durations, output.peak_memory_mb = _parse_chat_stage_metrics(resp_json)
+            else:
+                output.error = f"HTTP {response.status}: {await response.text()}"
+                output.success = False
+    except Exception as e:
+        output.error = str(e)
+        output.success = False
+
+    output.latency = time.perf_counter() - output.start_time
+    if output.success and input.slo_ms is not None:
+        output.slo_achieved = (output.latency * 1000.0) <= float(input.slo_ms)
+
+    if pbar:
+        pbar.update(1)
+    return output
+
+
+def _build_chat_payload(input: OmniRequest) -> dict[str, Any]:
+    """Build the /v1/chat/completions JSON body shared by the stream / non-stream senders."""
+    extra_body = dict(input.extra_body)
+    modalities = extra_body.pop("modalities", None)
+
+    content: list[dict[str, Any]] = []
+    if input.prompt:
+        content.append({"type": "text", "text": input.prompt})
+    if input.image_paths:
+        cached_urls = input.image_data_urls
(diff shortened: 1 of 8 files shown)


hsliuustc0106:
we prefer users to use `vllm bench serve --omni` and we will remove benchmarks/diffusion/ folder later after `vllm bench serve --omni` can support diffusion models as well, cc @NickCao

yibeicanghai:
> we prefer users to use `vllm bench serve --omni` and we will remove benchmarks/diffusion/ folder later after `vllm bench serve --omni` can support diffusion models as well, cc @NickCao

Got it. I'll extend and add new features under the vllm bench serve --omni framework. I used the benchmarks/diffusion/ folder as a reference when implementing, but I didn't know it was going to be removed. @hsliuustc0106

NilouBao:
My solution is to create a task scheduling class named _MultiModalTaskDataset_ under **/vllm_omni/benchmarks/data_modules**.

It schedules OmniRandomMultiModalDataset instances to obtain the capability of generating random underlying multimodal data. The MultiModalTaskDataset class receives parameters to control the request generation behavior in detail: 
1. _**--num-i2t/--num-t2i/--num-ti2i**_: explicitly specifies the number of requests of each type. 
2. _**--it2i-bot-task-weights**_: allocates the bot_task of ti2i by weight. 
3. _**--shuffle/--no-shuffle**_: controls whether to shuffle the request list. 
4. _**--gen-resolution-weights**_: controls the image resolution allocation. 
5. _**--randomize-input**_: randomizes the prompt length, input image resolution, and image content based on the seed.

The overall workflow is as follows: In the sample() method, a placeholder list of various tasks is generated based on the ratio. Then, the bot_task is allocated to the ti2i task and the resolution is allocated to the t2i/ti2i task in polling or random mode. Finally, the shuffle flag determines whether to shuffle the entire list or concatenate the tasks in group order. 

I plan to use this new class to assist in generating hybrid task requests. Please take a look and see if this idea is appropriate and feasible @NickCao

benchmarks/omni/omni_benchmark_serving.py
```diff
@@ -0,0 +1,1136 @@
...
+        idx = 0
+        async for req in iter_requests(requests_list, args.request_rate):
+            tasks.append(asyncio.create_task(limited(req, idx, pbar)))
+            idx += 1
+            # Yield to the loop so the just-scheduled task acquires the
+            # semaphore and reaches session.post (the request is pre-encoded,
+            # so there is no prep delay) before the next task is created. With
+            # the FIFO semaphore this makes the server arrival order match the
+            # list / dry-run order even under --request-rate inf + concurrency.
+            await asyncio.sleep(0.1)
```
lishunyang12, line 827: Change this to asyncio.sleep(0); the fixed delay caps dispatch at ten requests per second.

benchmarks/omni/omni_benchmark_serving.py
```diff
@@ -0,0 +1,1136 @@
...
+
+    bot_task_weights = parse_bot_task_weights(args.it2i_bot_task_weights)
+    group_order = _parse_group_order(args.no_shuffle_order)
+    gen_resolution_weights = parse_gen_resolution_weights(args.gen_resolution_weights)
+
+    config = OmniConfig(
+        width=args.width,
+        height=args.height,
+        num_inference_steps=args.num_inference_steps,
+        seed=args.seed,
```
lishunyang12, line 675: Pass args.seed_gen into OmniConfig so the generation seed actually reaches requests.

benchmarks/omni/omni_backends.py
```diff
@@ -0,0 +1,803 @@
...
+                            image_content.extend(content)
+                        if choice.get("finish_reason"):
+                            finish_reason = choice.get("finish_reason")
+                    usage = data.get("usage")
+                    if isinstance(usage, dict) and usage.get("completion_tokens"):
+                        output_tokens = int(usage["completion_tokens"])
+                    m = data.get("metrics")
+                    if isinstance(m, dict):
+                        metrics = m
+                output.success = True
```
lishunyang12, line 456: Check SSE error events before marking the request successful.

benchmarks/omni/omni_benchmark_serving.py
```diff
@@ -0,0 +1,1136 @@
...
+    print(
+        f"{indent}text-output (n={n}, out_tokens mean={out_tok['mean']:.1f} "
+        f"median={out_tok['median']:.1f} max={out_tok['max']}):"
+    )
+    for key, label in (("ttft", "TTFT"), ("tpot", "TPOT"), ("itl", "ITL")):
+        s = ts[key]
+        print(
+            f"{indent}  {label:<5} n={s['count']:<5} "
+            f"mean={s['mean_ms']:.2f}ms median={s['median_ms']:.2f}ms "
+            f"p50={s['p50_ms']:.2f}ms p95={s['p95_ms']:.2f}ms p99={s['p99_ms']:.2f}ms"
```
lishunyang12, line 333: Print the requested percentiles instead of hardcoding p50, p95, and p99.

PR #221, opened 2025-12-05
[DNM][CI] Test H100 on CI

ywang96:
PLEASE FILL IN THE PR DESCRIPTION HERE ENSURING ALL CHECKLIST ITEMS (AT THE BOTTOM) HAVE BEEN CONSIDERED.

## Purpose

Example buildkite setup to use H100 on CI

## Test Result

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft.
</details>

**BEFORE SUBMITTING, PLEASE READ <https://github.com/vllm-project/vllm-omni/blob/main/CONTRIBUTING.md>** (anything written below this line will be removed by GitHub Actions)

PR #6439, opened 2026-08-21
[Bugfix][Diffusion] Drop aborted request's orphaned async output (fix #6413)

### LINKED ISSUE #6413: [Bug] Aborted diffusion requests orphan async outputs in MultiprocDiffusionExecutor._completed_outputs (engine-process memory leak)
### Summary

With the multiproc executor in request mode (`step_execution=False`, async output enabled by default), every aborted diffusion request permanently retains one fully-unpacked output in the engine process. `_completed_outputs` in `MultiprocDiffusionExecutor` has no eviction path for outputs whose consumer never calls `wait_output_ready`.

Verified on `main` @ 273ffce13.

### Mechanism

1. Worker finishes the forward and replies COMPUTE_DONE with a fresh `async_output_id`; the background D2H/SHM thread will send OUTPUT_READY later (`diffusion_worker.py`, `_return_result`).
2. Request is aborted mid-generation (client cancel / API abort — the common case for 10-80s generations): `_abort_requests` marks scheduler state `FINISHED_ABORTED` (`diffusion_engine.py:1238`).
3. `_finalize_finished_request` returns a synthetic `DiffusionOutput(aborted=True)` from the ABORTED branch (`diffusion_engine.py:1253`) **before** the `async_output_id` branch (`diffusion_engine.py:1265`), so the pending `async_output_id` is dropped.
4. The consumer only waits when `output.async_output_id` is set (`diffusion_engine.py:387-388`), so `wait_output_ready` is never called for that id.
5. The worker still sends OUTPUT_READY. The result pump finds no pending future, unpacks the tensors anyway, and caches them: `multiproc_executor.py:913` (single) and `:938` (batch split).
6. Nothing evicts the entry: only `wait_output_ready` (`multiproc_executor.py:960`) and the early-batch pop (`:647`) remove from `_completed_outputs`, and `shutdown()` clears `_rpc_futures`/`_output_futures`/`_batch_split_map` but **not** `_completed_outputs` (`multiproc_executor.py:1004-1006`).

### Impact

Unbounded engine-process growth proportional to abort traffic — tens of MB per aborted image request, potentially hundreds of MB per aborted video request (unpacked frame tensors). Long-lived serving deployments with client cancels/timeouts accumulate these for the process lifetime. `describe_pending_state()` shows it as a monotonically growing `cached=` count. Batch splits leak one entry per aborted member.

### Reproduction sketch

Serve any request-mode diffusion model with `distributed_executor_backend: mp`, start generations, abort them mid-flight (client disconnect or API abort) in a loop, and watch engine RSS grow while `describe_pending_state().cached` climbs.

### Fix directions

* Bound + evict `_completed_outputs` in the result pump (drop oldest past a small bound, debug-log), and clear it in `shutdown()`.
* Or plumb aborts to an explicit executor `drop_output(async_output_id)` (the scheduler knows the ids).
* Or have `_finalize_finished_request` still return the `async_output_id`-bearing output for aborted requests so the existing wait path drains it (needs care around the aborted-output error flow).

### Relation to #6308

Found while auditing executors around #6308 (the UniProc executor needed explicit `gc.collect()`/`empty_cache()` teardown because in-process workers lose process-exit cleanup). The uniproc executor is immune to this leak by construction — no async envelope, no futures caches, outputs returned synchronously. Step mode (`step_execution=True`) bypasses the pump and is also unaffected.

@rahul-steiger-nv — since you have been deep in the executor/async-output machinery for #6308, could you double-check this chain (especially steps 3-5) and whether a bound/evict or a `drop_output` hook is the preferred fix here?

zhang-keliang:
## Summary

Fixes #6413.

Request-mode diffusion with the multiproc executor uses an **async output path**: the worker finishes the forward, replies `COMPUTE_DONE` with a fresh `async_output_id`, and a background D2H/SHM thread later sends `OUTPUT_READY` carrying the tensors. The consumer is expected to call `wait_output_ready(async_output_id)` to receive them.

When a request is **aborted** mid-generation (client cancel / API abort — the common case for 10–80 s generations), that contract breaks and the output leaks:

1. `_abort_requests` sets the scheduler status to `FINISHED_ABORTED`.
2. On the next step, `_finalize_finished_request` hits the `FINISHED_ABORTED` branch and returns a synthetic aborted `DiffusionOutput` **before** the `async_output_id` branch, so the pending id is silently dropped.
3. Because the returned aborted output has no id, the consumer **never calls `wait_output_ready`** — no waiter is ever registered (and in the common abort case the consumer coroutine is already gone).
4. The worker's background thread still sends `OUTPUT_READY`. In the result pump, with no pending future, the tensors are unpacked and a completed `Future` is cached in `self._completed_outputs[async_output_id]`.
5. Nothing ever evicts it: `_completed_outputs` is only popped by `wait_output_ready` (never called for this id) and the early-batch pop. `shutdown()` clears `_rpc_futures` / `_output_futures` / `_batch_split_map` but **not** `_completed_outputs`.

Result: unbounded engine-process growth proportional to abort traffic (unpacked image/video frame tensors), for the process lifetime; the cache also survives shutdown.

## Fix

Host-side bookkeeping only, minimal and scoped to avoid the contested `OUTPUT_READY` caching block that other in-flight pump PRs touch:

- **`MultiprocDiffusionExecutor.drop_output(async_output_id)`** (new): under `_futures_lock`, drains an async output that will never be waited on, handling both arrival orderings —
  - result already arrived → pop and drop the cached future from `_completed_outputs`;
  - result not yet arrived → register a placeholder `Future` in `_output_futures` so the pump resolves *into it* (and lets it be freed) instead of caching;
  - a genuine already-registered waiter is left untouched (drains normally).
- **`shutdown()`** also `self._completed_outputs.clear()` inside the `_futures_lock` block, so cached tensors do not survive shutdown.
- **`_finalize_finished_request()`**: in the `FINISHED_ABORTED` branch, if `runner_output.async_output_id` is set, call `self.executor.drop_output(...)` (guarded via `getattr`, since non-multiproc executors like UniProc do not have it) *before* returning the aborted output. This is the root-cause fix: the engine stops silently dropping the id and instead tells the executor to reclaim it.

## Test

CPU-only unit tests on the exact host-side bookkeeping where the leak lives (uses the existing `object.__new__` executor scaffold — no process spawn, no GPU):

- `tests/diffusion/test_result_pump.py`: `TestDropOutput` (already-cached drop; not-yet-arrived placeholder prevents caching; genuine waiter untouched) and `TestShutdownClearsCompletedOutputs`.
- `tests/diffusion/test_diffusion_engine_cleanup.py`: aborted-request finalize drops the pending async output (asserts `drop_output` called with the id) + negative guard when there is no async output.

Results:
- **RED** (source reverted, tests only): `5 failed, 1 passed` — the 5 new positive assertions fail (`drop_output` missing / not called / cache not cleared).
- **GREEN** (with fix): `test_result_pump.py` + `test_diffusion_engine_cleanup.py` → **40 passed**.
- Surrounding regression: `test_diffusion_engine.py` + `test_async_output_worker.py` → **49 passed, 1 failed**; the single failure is `test_move_tensor_tree_moves_nested_cuda_tensors_to_cpu` (`@hardware_test(res={"cuda":"L4"})`, "Torch not compiled with CUDA enabled") — pre-existing GPU-only, unrelated to this change.
- `ruff check` on all changed files → clean.

The runtime RSS-growth reproduction (serve a request-mode diffusion model with `distributed_executor_backend: mp`, abort in a loop, watch RSS) needs GPU + weights and was **not** run here, so no measured MB figure is claimed; the fix is validated by the unit tests on the host-side bookkeeping that actually leaks.

## Abort contract (added per review)

After a request is aborted — either its `wait_output_ready()` waiter was cancelled, or the abort path called `drop_output(async_output_id)` — that `async_output_id` is terminal: its late `OUTPUT_READY` is unpacked (so SHM segments are unlinked) and then discarded, never cached. A later `wait_output_ready()` on the same id does not hang and does not return the old tensors; it returns an already-failed Future (`RuntimeError: ... was dropped ...`). Clients that abort must therefore treat the abort as a whole-sample retry with a new request (e.g. verl-omni fully-async / `partial_rollout` cannot merge pixels from the aborted generation). Dropped ids are remembered in a bounded LRU (`_DROPPED_OUTPUT_IDS_MAX`) so this fail-fast behavior does not itself grow unbounded under abort traffic.

### DIFF
--- a/tests/diffusion/test_multiproc_shutdown_race.py
+++ b/tests/diffusion/test_multiproc_shutdown_race.py
@@ -18,6 +18,7 @@ after shutdown still has to be unpacked (just never cached).
 import queue
 import threading
 import time
+from collections import OrderedDict
 from unittest.mock import MagicMock
 
 import pytest
@@ -47,6 +48,7 @@ def _make_executor():
     executor._rpc_futures = {}
     executor._output_futures = {}
     executor._completed_outputs = {}
+    executor._dropped_output_ids = OrderedDict()
     executor._batch_split_map = {}
     executor._futures_lock = threading.RLock()
     executor._pump_running = False
--- a/tests/diffusion/test_multiproc_shutdown_race.py
+++ b/tests/diffusion/test_multiproc_shutdown_race.py
@@ -1,0 +1,232 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+"""Regression tests for MultiprocDiffusionExecutor shutdown / delayed-pump race.
+
+Reproduces hsliuustc0106's #6439 review finding: `shutdown()` joins each result
+pump thread with a 2s timeout and only warns if it survives. A pump blocked
+inside `unpack_diffusion_output_shm` can resume AFTER `_completed_outputs` was
+cleared and repopulate the dictionary, reintroducing the leak this PR was meant
+to close.
+
+Both single-output and batch-split code paths must discard deliveries once
+`_closed` is set; `drop_output()` must be a no-op after shutdown.
+"""
+
+import queue
+import threading
+import time
+from unittest.mock import MagicMock
+
+import pytest
+
+from vllm_omni.diffusion.data import (
+    AsyncDiffusionOutput,
+    AsyncOutputKind,
+    DiffusionOutput,
+)
+
+pytestmark = [pytest.mark.core_model, pytest.mark.cpu]
+
+
+def _make_executor():
+    """Minimal MultiprocDiffusionExecutor-like object; mirrors test_result_pump helper."""
+    from vllm_omni.diffusion.executor.multiproc_executor import (
+        MultiprocDiffusionExecutor,
+    )
+
+    od_config = MagicMock()
+    od_config.step_execution = False
+
+    executor = object.__new__(MultiprocDiffusionExecutor)
+    executor.od_config = od_config
+    executor._rpc_id_counter = 0
+    executor._rpc_id_lock = threading.Lock()
+    executor._rpc_futures = {}
+    executor._output_futures = {}
+    executor._completed_outputs = {}
+    executor._batch_split_map = {}
+    executor._futures_lock = threading.RLock()
+    executor._pump_running = False
+    executor._pump_stop = threading.Event()
+    executor._sync_result_buffer = queue.Queue()
+    executor._result_mq = MagicMock()
+    executor._result_mqs = []
+    executor._broadcast_mq = MagicMock()
+    executor._closed = False
+    executor._is_failed = False
+    executor._finalizer = MagicMock()
+    executor._shutdown_cleaner = None
+    executor._processes = []
+    return executor
+
+
+class _FakeBatchOutput:
+    """Batch-level output exposing per-request results (mirrors test_result_pump helper)."""
+
+    def __init__(self, results):
+        self._results = results
+
+    def get_request_output(self, req_id):
+        result = self._results.get(req_id)
+        if result is None:
+            return None
+        return MagicMock(result=result)
+
+
+def _run_pump_blocked_until_shutdown(executor, msg, unpack_gate):
+    """Drive the pump thread with one *msg*.
+
+    The pump enters `unpack_diffusion_output_shm` and blocks on *unpack_gate*.
+    Caller is expected to close the executor, then release the gate. Returns
+    the pump thread so the test can join it.
+    """
+    call_count = [0]
+
+    def mock_dequeue(timeout=None):
+        call_count[0] += 1
+        if call_count[0] == 1:
+            return msg
+        # After the single msg, block until pump_stop is set so the pump
+        # loops out cleanly (mirrors _feed_one_msg_to_pump behaviour).
+        while not executor._pump_stop.is_set():
+            time.sleep(0.02)
+        raise TimeoutError
+
+    executor._result_mq.dequeue = mock_dequeue
+    t = threading.Thread(target=executor._result_pump, daemon=True)
+    t.start()
+    # Give the pump time to dequeue and reach unpack_gate.wait().
+    time.sleep(0.1)
+    return t
+
+
+class TestResultPumpDelayedAfterShutdown:
+    """Reproduce hsliuustc0106's finding: a pump blocked inside SHM unpack can
+    resume after shutdown clears `_completed_outputs` and repopulate it.
+
+    Without gating, the assertions here fail (leak reintroduced through the
+    late-write path). With gating on `_closed`, deliveries after shutdown are
+    discarded.
+    """
+
+    def test_result_pump_delayed_after_shutdown_single(self, mocker):
+        """Single-output pump path: late unpack must not repopulate the dict."""
+        executor = _make_executor()
+
+        unpack_gate = threading.Event()
+
+        def _blocked_unpack(*_args, **_kwargs):
+            unpack_gate.wait(timeout=10.0)
+
+        mocker.patch(
+            "vllm_omni.diffusion.executor.multiproc_executor.unpack_diffusion_output_shm",
+            side_effect=_blocked_unpack,
+        )
+
+        async_output_id = "abc-single"
+        msg = AsyncDiffusionOutput(
+            kind=AsyncOutputKind.OUTPUT_READY,
+            async_output_id=async_output_id,
+            output=DiffusionOutput(output="late"),
+        )
+
+        pump_thread = _run_pump_blocked_until_shutdown(executor, msg, unpack_gate)
+
+        # Simulate shutdown() clearing state while the pump is still stuck
+        # inside unpack. `_closed = True` mirrors shutdown()'s first line.
+        executor._closed = True
+        with executor._futures_lock:
+            executor._completed_outputs.clear()
+
+        # Release the pump: it resumes past unpack and tries to write into
+        # _completed_outputs. Without the gate, this reintroduces the leak.
+        unpack_gate.set()
+
+        # Give the pump time to finish delivering and hit the gate.
+        time.sleep(0.2)
+        executor._pump_stop.set()
+        pump_thread.join(timeout=2.0)
+        assert not pump_thread.is_alive(), "pump thread did not exit"
+
+        # The core assertion: the delayed pump must not have repopulated
+        # `_completed_outputs` after shutdown cleared it.
+        assert executor._completed_outputs == {}, (
+            f"delayed pump repopulated dict after shutdown: "
+            f"{list(executor._completed_outputs)}"
+        )
+
+    def test_result_pump_delayed_after_shutdown_batch_split(self, mocker):
+        """Batch-split pump path: late unpack must not repopulate the dict."""
+        executor = _make_executor()
+
+        unpack_gate = threading.Event()
+
+        def _blocked_unpack(*_args, **_kwargs):
+            unpack_gate.wait(timeout=10.0)
+
+        mocker.patch(
+            "vllm_omni.diffusion.executor.multiproc_executor.unpack_diffusion_output_shm",
+            side_effect=_blocked_unpack,
+        )
+
+        batch_id = "batch-delayed"
+        # Pre-populate batch split map so pump goes down the batch path,
+        # WITHOUT registering per-request waiters (they are the ones aborted).
+        with executor._futures_lock:
+            executor._batch_split_map[batch_id] = {
+                f"{batch_id}/r-a": "r-a",
+                f"{batch_id}/r-b": "r-b",
+            }
+
+        outputs = {
+            "r-a": DiffusionOutput(output="late-a"),
+            "r-b": DiffusionOutput(output="late-b"),
+        }
+        msg = AsyncDiffusionOutput(
+            kind=AsyncOutputKind.OUTPUT_READY,
+            async_output_id=batch_id,
+            output=_FakeBatchOutput(outputs),
+        )
+
+        pump_thread = _run_pump_blocked_until_shutdown(executor, msg, unpack_gate)
+
+        # Simulate shutdown during unpack.
+        executor._closed = True
+        with executor._futures_lock:
+            executor._completed_outputs.clear()
+
+        # Release the pump so it enters _deliver_batch_split with no waiters
+        # registered; without gating, both per-request results land in
+        # _completed_outputs.
+        unpack_gate.set()
+
+        time.sleep(0.2)
+        executor._pump_stop.set()
+        pump_thread.join(timeout=2.0)
+        assert not pump_thread.is_alive(), "pump thread did not exit"
+
+        assert executor._completed_outputs == {}, (
+            f"delayed batch-split pump repopulated dict after shutdown: "
+            f"{list(executor._completed_outputs)}"
+        )
+
+
+class TestDropOutputAfterShutdown:
+    """`drop_output()` runs from the abort path and must be a no-op once the
+    executor has entered shutdown, so it does not race the pump into an
+    already-torn-down `_output_futures` / `_completed_outputs`.
+    """
+
+    def test_drop_output_after_shutdown_is_noop(self):
+        executor = _make_executor()
+        executor._closed = True
+
+        # Before the gate, `drop_output` would register a placeholder waiter
+        # in `_output_futures`. After shutdown, that would leak past teardown.
+        executor.drop_output("abc-abort")
+
+        assert executor._output_futures == {}, (
+            f"drop_output registered waiter after shutdown: "
+            f"{list(executor._output_futures)}"
+        )
+        assert executor._completed_outputs == {}
--- a/tests/diffusion/test_multiproc_shutdown_race.py
+++ b/tests/diffusion/test_multiproc_shutdown_race.py
@@ -151,8 +151,7 @@ class TestResultPumpDelayedAfterShutdown:
         # The core assertion: the delayed pump must not have repopulated
         # `_completed_outputs` after shutdown cleared it.
         assert executor._completed_outputs == {}, (
-            f"delayed pump repopulated dict after shutdown: "
-            f"{list(executor._completed_outputs)}"
+            f"delayed pump repopulated dict after shutdown: {list(executor._completed_outputs)}"
         )
 
     def test_result_pump_delayed_after_shutdown_batch_split(self, mocker):
@@ -206,8 +205,7 @@ class TestResultPumpDelayedAfterShutdown:
         assert not pump_thread.is_alive(), "pump thread did not exit"
 
         assert executor._completed_outputs == {}, (
-            f"delayed batch-split pump repopulated dict after shutdown: "
-            f"{list(executor._completed_outputs)}"
+            f"delayed batch-split pump repopulated dict after shutdown: {list(executor._completed_outputs)}"
         )
 
 
@@ -226,7 +224,6 @@ class TestDropOutputAfterShutdown:
         executor.drop_output("abc-abort")
 
         assert executor._output_futures == {}, (
-            f"drop_output registered waiter after shutdown: "
-            f"{list(executor._output_futures)}"
+            f"drop_output registered waiter after shutdown: {list(executor._output_futures)}"
         )
         assert executor._completed_outputs == {}
--- a/tests/diffusion/test_multiproc_shutdown_race.py
+++ b/tests/diffusion/test_multiproc_shutdown_race.py
@@ -9,7 +9,10 @@ cleared and repopulate the dictionary, reintroducing the leak this PR was meant
 to close.
 
 Both single-output and batch-split code paths must discard deliveries once
-`_closed` is set; `drop_output()` must be a no-op after shutdown.
+`_closed` is set; `drop_output()` must be a no-op after shutdown. Discarding
+must NOT skip SHM unpack: `unpack_diffusion_output_shm()` is the only
+receive-side path that unlinks named SHM segments, so an OUTPUT_READY dequeued
+after shutdown still has to be unpacked (just never cached).
 """
 
 import queue
@@ -73,12 +76,14 @@ class _FakeBatchOutput:
         return MagicMock(result=result)
 
 
-def _run_pump_blocked_until_shutdown(executor, msg, unpack_gate):
+def _run_pump_with_one_msg(executor, msg):
     """Drive the pump thread with one *msg*.
 
-    The pump enters `unpack_diffusion_output_shm` and blocks on *unpack_gate*.
-    Caller is expected to close the executor, then release the gate. Returns
-    the pump thread so the test can join it.
+    After the single msg, dequeue blocks until ``_pump_stop`` is set so the
+    pump loops out cleanly. Returns the pump thread. The caller finishes the
+    pump with ``_finish_pump``; the message's delivery completes before the
+    second dequeue call, so joining the thread is a deterministic barrier for
+    "delivery handled".
     """
     call_count = [0]
 
@@ -86,8 +91,6 @@ def _run_pump_blocked_until_shutdown(executor, msg, unpack_gate):
         call_count[0] += 1
         if call_count[0] == 1:
             return msg
-        # After the single msg, block until pump_stop is set so the pump
-        # loops out cleanly (mirrors _feed_one_msg_to_pump behaviour).
         while not executor._pump_stop.is_set():
             time.sleep(0.02)
         raise TimeoutError
@@ -95,9 +98,13 @@ def _run_pump_blocked_until_shutdown(executor, msg, unpack_gate):
     executor._result_mq.dequeue = mock_dequeue
     t = threading.Thread(target=executor._result_pump, daemon=True)
     t.start()
-    # Give the pump time to dequeue and reach unpack_gate.wait().
-    time.sleep(0.1)
     return t
+
+
+def _finish_pump(executor, pump_thread):
+    executor._pump_stop.set()
+    pump_thread.join(timeout=5.0)
+    assert not pump_thread.is_alive(), "pump thread did not exit"
 
 
 class TestResultPumpDelayedAfterShutdown:
@@ -106,16 +113,20 @@ class TestResultPumpDelayedAfterShutdown:
 
     Without gating, the assertions here fail (leak reintroduced through the
     late-write path). With gating on `_closed`, deliveries after shutdown are
-    discarded.
+    discarded. `entered_unpack` proves the pump is inside the mocked unpack
+    BEFORE `_closed` is set, so the post-unpack race path is really exercised
+    (a fixed sleep could let the pump take the early closed gate instead).
     """
 
     def test_result_pump_delayed_after_shutdown_single(self, mocker):
         """Single-output pump path: late unpack must not repopulate the dict."""
         executor = _make_executor()
 
+        entered_unpack = threading.Event()
         unpack_gate = threading.Event()
 
         def _blocked_unpack(*_args, **_kwargs):
+            entered_unpack.set()
             unpack_gate.wait(timeout=10.0)
 
         mocker.patch(
@@ -130,7 +141,11 @@ class TestResultPumpDelayedAfterShutdown:
             output=DiffusionOutput(output="late"),
         )
 
-        pump_thread = _run_pump_blocked_until_shutdown(executor, msg, unpack_gate)
+        pump_thread = _run_pump_with_one_msg(executor, msg)
+
+        # The pump must be INSIDE unpack (past the early closed gate) before
+        # shutdown starts, otherwise the post-unpack race is not exercised.
+        assert entered_unpack.wait(timeout=5.0), "pump never reached SHM unpack"
 
         # Simulate shutdown() clearing state while the pump is still stuck
         # inside unpack. `_closed = True` mirrors shutdown()'s first line.
@@ -142,11 +157,9 @@ class TestResultPumpDelayedAfterShutdown:
         # _completed_outputs. Without the gate, this reintroduces the leak.
         unpack_gate.set()
 
-        # Give the pump time to finish delivering and hit the gate.
-        time.sleep(0.2)
-        executor._pump_stop.set()
-        pump_thread.join(timeout=2.0)
-        assert not pump_thread.is_alive(), "pump thread did not exit"
+        # Joining the pump is a deterministic barrier: the delivery is fully
+        # handled before the pump blocks on the second dequeue and exits.
+        _finish_pump(executor, pump_thread)
 
         # The core assertion: the delayed pump must not have repopulated
         # `_completed_outputs` after shutdown cleared it.
@@ -158,9 +171,11 @@ class TestResultPumpDelayedAfterShutdown:
         """Batch-split pump path: late unpack must not repopulate the dict."""
         executor = _make_executor()
 
+        entered_unpack = threading.Event()
         unpack_gate = threading.Event()
 
         def _blocked_unpack(*_args, **_kwargs):
+            entered_unpack.set()
             unpack_gate.wait(timeout=10.0)
 
         mocker.patch(
@@ -187,7 +202,9 @@ class TestResultPumpDelayedAfterShutdown:
             output=_FakeBatchOutput(outputs),
         )
 
-        pump_thread = _run_pump_blocked_until_shutdown(executor, msg, unpack_gate)
+        pump_thread = _run_pump_with_one_msg(executor, msg)
+
+        assert entered_unpack.wait(timeout=5.0), "pump never reached SHM unpack"
 
         # Simulate shutdown during unpack.
         executor._closed = True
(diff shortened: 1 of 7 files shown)


rahul-steiger-nv:
Verified against #6413: the abort-path and shutdown fixes work, and the targeted
tests pass (`40 passed`).

One additional case should be covered: if an output future is cancelled, the
late result is re-cached because `pending.done()` is true.

Desired behavior when `OUTPUT_READY` arrives:

- no registered waiter → cache for a later `wait_output_ready()`;
- registered pending future → resolve it;
- registered cancelled/done future → discard the result.

This applies to both single-output and batch-split delivery. The existing
cancelled-future tests can assert that `_completed_outputs` remains empty.

zhang-keliang:
Thanks for the careful review, @rahul-steiger-nv — you're right, that was a real gap.

Fixed in `8b68728`. The pump previously used `if pending is not None and not pending.done(): resolve() else: cache()`, so a registered-but-cancelled/done waiter fell into the caching branch and re-cached the late `OUTPUT_READY` into `_completed_outputs` — reintroducing the exact leak this PR set out to fix. I've made the three cases explicit on **both** delivery paths (single-output in `_result_pump` and per-member in `_deliver_batch_split`):

- no waiter registered → cache for a later `wait_output_ready()`;
- waiter pending → resolve it;
- waiter cancelled/done → discard the result, do **not** re-cache.

Tests (CPU, RED→GREEN):
- New `TestResultPumpDiscardsDoneFuture` with a genuinely cancelled `Future` (real `done()==True`) for single-output and batch-split, asserting `_completed_outputs == {}`. These failed before the fix (the result was re-cached) and pass after.
- Extended the existing `TestResultPumpCancelledFutureRace` cases (single, output-ready, batch-split) with `assert _completed_outputs == {}` as you suggested.
- Targeted suites now `42 passed` (was 40); surrounding `test_diffusion_engine.py` + `test_async_output_worker.py` `49 passed` (the one failing case is the pre-existing `@hardware_test(cuda:L4)` `test_move_tensor_tree_...`, GPU-only, unrelated). `ruff` clean.

Note the existing racy tests use a `done()`-always-false stub, so they exercised the *pending* branch and never hit this cache path — that's why the gap slipped through; the new tests use a real cancelled future to cover it.

vllm_omni/diffusion/diffusion_engine.py
fhfuih, line 1: No need to use `getattr` if it is only for test mocks and the production code is already typed. You can add a stub or abstract definition of `drop_output` on base DiffusionExecutor as well (so it is well typed if the end user extends their own executor), remove the getattr in production code, and adjust only the mock if needed

fhfuih (commented):
The PR overall looks good to me. The abort drain, cancelled-future discard, and `shutdown()` clear of `_completed_outputs` match the leak chain.

Small coding style consideration:

zhang-keliang:
Thanks @fhfuih. Done in a91c94e: added a default no-op `drop_output` on the base `DiffusionExecutor` (mirroring `register_failure_callback`, so it's well typed for user-defined executors too) and dropped the `getattr` in the abort branch for a direct `self.executor.drop_output(...)`. `MultiprocDiffusionExecutor` keeps its real override; the abort/no-async tests still pass.

vllm_omni/diffusion/executor/multiproc_executor.py
```diff
@@ -1004,5 +1034,8 @@ def shutdown(self) -> None:
                 self._rpc_futures.clear()
                 self._output_futures.clear()
                 self._batch_split_map.clear()
+                # Cached async outputs hold unpacked tensors; drop them so they
+                # do not survive shutdown (issue #6413).
+                self._completed_outputs.clear()
```
hsliuustc0106, line 1181: `shutdown()` can return while a result-pump thread is still alive: each join is capped at 2 seconds and a surviving pump is only warned about. That pump can then repopulate the dictionary this line just cleared. I reproduced this by blocking SHM unpack: the cache was empty when shutdown returned, then contained the late ID after the pump resumed. Please discard deliveries once `_closed` is set, make `drop_output()` a no-op after shutdown, and add delayed-pump regressions for the single and batch paths.

zhang-keliang: Thanks for the detailed reproduction, addressed in `bef011d`.

Added `self._closed` gates guarded by `_futures_lock` at every state-write site so a delayed pump / drop can no longer repopulate cleared state:

- `_result_pump`: early-continue after dequeue when `_closed` is set — a delayed message never reaches either write path.
- Single-output write site (line 910): re-check `_closed` under the lock after unpack, since `shutdown()` can slip in while the lock is briefly released.
- `_deliver_batch_split`: entry check + per-iteration re-check under the lock (the per-request loop drops and re-acquires the lock, so a mid-loop shutdown must not repopulate the cleared dict).
- `drop_output`: entry check under `_futures_lock` — abort path is a no-op after shutdown.

Regression added at `tests/diffusion/test_multiproc_shutdown_race.py` (3 pytests, CPU, no GPU deps): drives the pump into `unpack_diffusion_output_shm` via a `threading.Event` gate, sets `_closed` and clears `_completed_outputs` mid-flight, then releases the gate and asserts nothing repopulates — for both single and batch-split paths — plus a third case for `drop_output()` post-shutdown.

Local RED@parent → 0/3 (all three leak paths trigger exactly as you described); GREEN@bef011d → 3/3.

hsliuustc0106:
fix precommits please

zhang-keliang:
Done in 6bf67b6 — ruff-format on the new test file. pre-commit, build (3.11/3.12) and DCO are all green now.

vllm_omni/diffusion/executor/multiproc_executor.py
```diff
@@ -863,6 +863,12 @@ def _result_pump(self, result_mq: MessageQueue | None = None) -> None:
                 self._sync_result_buffer.put(msg)
                 continue
 
+            # If shutdown started while we were dequeuing / unpacking, drop
+            # this delivery so it cannot repopulate _completed_outputs after
+            # shutdown() cleared it (issue #6413 / #6439 review).
+            if self._closed:
```
hsliuustc0106: [P1] Preserve SHM cleanup before this closed-time discard

`OUTPUT_READY` may already carry named POSIX SHM handles. If shutdown sets `_closed` after dequeue, this `continue` skips `unpack_diffusion_output_shm()`, which is the only receive-side path that calls `shm.unlink()`. Large image/video segments then survive in `/dev/shm` for the lifetime of the parent process. Please still unpack/unlink `OUTPUT_READY` before discarding it (the later `_closed` checks already prevent cache writes), or add a cleanup-only handle walker, plus an above-threshold SHM regression.

tests/diffusion/test_multiproc_shutdown_race.py
```diff
@@ -0,0 +1,229 @@
...
+        # loops out cleanly (mirrors _feed_one_msg_to_pump behaviour).
+        while not executor._pump_stop.is_set():
+            time.sleep(0.02)
+        raise TimeoutError
+
+    executor._result_mq.dequeue = mock_dequeue
+    t = threading.Thread(target=executor._result_pump, daemon=True)
+    t.start()
+    # Give the pump time to dequeue and reach unpack_gate.wait().
+    time.sleep(0.1)
```
hsliuustc0106: [P2] Synchronize this interleaving instead of sleeping

This never proves the pump reached the mocked unpack call. On a delayed thread, the test can set `_closed` first and pass through the early closed gate without exercising the post-unpack race. Have the mock set an `entered_unpack` event, require it before closing, then release it with a second event; that can replace both fixed sleeps.

vllm_omni/diffusion/executor/multiproc_executor.py
zhang-keliang: Good catch — fixed in 469ab59. The early gate now only drops non-OUTPUT_READY messages; OUTPUT_READY still flows through the dispatch so `unpack_diffusion_output_shm()` unlinks the segments, and the closed-time re-checks under `_futures_lock` keep any cache write out. Added a regression (`test_output_ready_after_shutdown_still_unpacks_shm`): message dequeued after `_closed` is set → unpack called exactly once, `_completed_outputs`/`_output_futures` stay empty. It fails on the previous gate (unpack skipped) and passes now.

tests/diffusion/test_multiproc_shutdown_race.py
zhang-keliang: Done in 469ab59 — the mocked unpack now sets an `entered_unpack` event which the test requires (assert, 5s timeout) before flipping `_closed`, and a second event releases it; both fixed sleeps are gone. The post-release barrier is deterministic too: the pump fully handles the delivery before blocking on the second dequeue, so `_pump_stop` + join replaces the trailing sleep.

vllm_omni/diffusion/executor/multiproc_executor.py
```diff
@@ -967,6 +996,32 @@ def wait_output_ready(self, async_output_id: str) -> concurrent.futures.Future[D
...
+        with self._futures_lock:
+            if self._closed:
+                # Executor is torn down; abort path must not repopulate the
+                # cleared state (issue #6413 / #6439 review).
+                return
+            if self._completed_outputs.pop(async_output_id, None) is not None:
+                return
+            if async_output_id in self._output_futures:
+                return
+            self._output_futures[async_output_id] = concurrent.futures.Future()
```
SamitHuang: `drop_output` parks a placeholder in `_output_futures`, but `wait_output_ready` still always does `_output_futures[id] = new Future()` when the id is not in `_completed_outputs`. If the pump resolves the placeholder first (abort finalized, `OUTPUT_READY` arrives, then a consumer waits), the new waiter never completes.

verl-omni fully async abort (`abort_replicas` / HTTP cancel overlapping `step_streaming`) can hit abort-finalize → `drop_output` → late wait on the same `async_output_id`. Reuse an existing `_output_futures` entry (and return a completed placeholder if the pump already resolved it) instead of overwriting.

vllm_omni/diffusion/executor/multiproc_executor.py
```diff
@@ -967,6 +996,32 @@ def wait_output_ready(self, async_output_id: str) -> concurrent.futures.Future[D
             self._output_futures[async_output_id] = fut
```
SamitHuang, line 1106: This assignment is the other half of the hang: it clobbers the `drop_output` placeholder without checking `if async_output_id in self._output_futures`. After reuse, `wait_output_ready` must return that future under `_futures_lock` so drop-then-wait and wait-then-drop agree.

Without that, fully async `FullyAsyncDiffusionLLMServerClient` cannot safely wait after an abort that already drained the id, even if the client still holds `async_output_id` from `COMPUTE_DONE`.

vllm_omni/diffusion/executor/multiproc_executor.py
```diff
@@ -898,19 +908,27 @@ def _result_pump(self, result_mq: MessageQueue | None = None) -> None:
...
                                     fut.set_exception(exc)
                                 else:
                                     fut.set_result(output_result)
                                 self._completed_outputs[batch_id] = fut
+                            elif not pending.done():
+                                if exc is not None:
+                                    try_set_exception(pending, exc)
+                                else:
+                                    try_set_result(pending, output_result)
+                            # else: waiter registered but already cancelled/done
```
SamitHuang: Discarding late `OUTPUT_READY` when the waiter is already cancelled/done is required to close #6413, and it is the correct behavior for verl-omni abort-all during D2H (the generate task is cancelled; retry is a new request).

It is a breaking change for any client that cancelled `wait_output_ready` and expected to claim the same `async_output_id` later from `_completed_outputs`. Document that contract: after abort/cancel, the old id is gone; `partial_rollout` resume for diffusion cannot merge pixels from the aborted generation.

tests/diffusion/test_result_pump.py
```diff
@@ -492,3 +493,129 @@ def test_batch_split_racing_cancel_does_not_crash_pump(self):
...
+        fut.set_result(DiffusionOutput(output="leaked"))
+        with executor._futures_lock:
+            executor._completed_outputs["aid-1"] = fut
+
+        executor.drop_output("aid-1")
+
+        with executor._futures_lock:
+            assert "aid-1" not in executor._completed_outputs
+
+    def test_prevents_caching_when_output_not_yet_arrived(self):
```
SamitHuang, line 581: This covers drop-then-pump with no later wait. Add the ordering that hangs today: `drop_output(id)` → feed `OUTPUT_READY` → `wait_output_ready(id)` must not return a future that never completes (timeout/assert `done()`).

Also cover `drop_output` then `wait_output_ready` before the pump, asserting both calls share one future. Those two cases are the overlap fully async abort vs `step_streaming` wait will exercise.

SamitHuang (changes requested):
Host-side abort drain for orphaned `OUTPUT_READY` is the right fix for #6413, and the cancelled-waiter / shutdown `_closed` follow-ups match the leak chain. I did not approve.

**verl-omni fully async ([RFC #221](https://github.com/verl-project/verl-omni/issues/221)):** this does **not** block the streaming rollouter/trainer skeleton (`staleness_threshold=0`). Colocated/NCCL weight sync does `abort_replicas()` then `sleep(level=1)` / wake while the multiproc executor stays alive — without `drop_output`, abort traffic would keep unpacked image/video tensors in `_completed_outputs` for the replica lifetime, which is exactly the leak fully async will hit. `sleep()` does not set `_closed`, so the new shutdown gates do not apply there; abort-path drain is the mechanism that matters.

Phase 3 `partial_rollout=True` / `FullyAsyncDiffusionLLMServerClient` must treat diffusion abort as **whole-sample retry**, not reclaim of a late `OUTPUT_READY` on the old `async_output_id`. After this PR, a cancelled waiter or `drop_output`+pump discards those tensors; a later `wait_output_ready` on the same id can hang (see inline). That is compatible with HTTP abort → new generate(), and incompatible with merging pixels from the aborted request.

Please fix `wait_output_ready` so it reuses a `drop_output` placeholder instead of overwriting it, and add that race to `TestDropOutput`.

vllm_omni/diffusion/executor/multiproc_executor.py
zhang-keliang: Fixed in 1336ba1. `wait_output_ready` now reuses an existing `_output_futures` entry (the `drop_output` placeholder or any already-resolved future) instead of clobbering it with a fresh one. Also added the two tests you specified: `drop → wait` shares one future, `drop → pump → wait` returns done.

vllm_omni/diffusion/executor/multiproc_executor.py
zhang-keliang: Fixed in the same commit: when the pump resolves a pending future, it now also caches it in `_completed_outputs` (guarded by `pending.done() and not pending.cancelled()` to avoid caching a racing cancel) so the `drop → pump → wait` ordering finds the result via the `cached = _completed_outputs.pop(...)` path in `wait_output_ready`.

vllm_omni/diffusion/executor/multiproc_executor.py
zhang-keliang: Agreed — this is the intended contract: after abort/cancel, the old id is gone and cannot be reclaimed from `_completed_outputs`. I'll add a note in the PR body to document this explicitly ("abort consumed; `partial_rollout` resume must use a new request").

tests/diffusion/test_result_pump.py
zhang-keliang: Both orderings added in 1336ba1:
- `test_drop_then_pump_then_wait_reuses_placeholder`: `drop → pump → wait` must return a `done()` future (was a hang before the fix)
- `test_drop_then_wait_share_same_future`: `drop → wait` (before pump) shares one future object

vllm_omni/diffusion/executor/multiproc_executor.py
hsliuustc0106: Verified at 1336ba1 — the closed gates cover the pump and both write paths; thanks.

vllm_omni/diffusion/executor/multiproc_executor.py
hsliuustc0106: Verified at 1336ba1 — the gate now exempts OUTPUT_READY so unpack_diffusion_output_shm() still unlinks the segments, with the regression test in place. Thanks.

tests/diffusion/test_multiproc_shutdown_race.py
hsliuustc0106: Verified at 1336ba1 — entered_unpack/unpack_gate events make the interleaving deterministic and the fixed sleeps are gone. Thanks.

vllm_omni/diffusion/executor/multiproc_executor.py
```diff
@@ -898,19 +908,33 @@ def _result_pump(self, result_mq: MessageQueue | None = None) -> None:
...
+                                if exc is not None:
+                                    try_set_exception(pending, exc)
+                                else:
+                                    try_set_result(pending, output_result)
+                                # Cache if the resolve succeeded so a late
+                                # wait_output_ready (drop → pump → wait) can
+                                # still retrieve the result. Skip if a racing
+                                # cancel won the set (pending.cancelled()).
+                                if pending.done() and not pending.cancelled():
+                                    self._completed_outputs[batch_id] = pending
```
hsliuustc0106: [P1] Do not re-cache an output that already has a waiter

In the normal waiter-before-delivery path, step_streaming() has already called wait_output_ready() and holds pending. After this branch resolves that Future, this assignment stores the same Future—and its DiffusionOutput tensors—in _completed_outputs. There is no second wait_output_ready() call to pop it, so each single-request OUTPUT_READY delivered after waiter registration remains retained until shutdown. The drop placeholder follows the same branch: drop_output(id) → OUTPUT_READY with no late wait now re-caches the aborted output, recreating #6413. The updated tests at test_result_pump.py:270 and :582 currently assert this retention; the latter is still named test_prevents_caching_when_output_not_yet_arrived.

Please keep a genuine registered waiter on the direct resolve/no-cache path and restore assertions that wait → pump → consume and drop → pump with no later wait leave _completed_outputs empty. If drop → pump → late wait must remain supported, distinguish a discard placeholder from a claimed waiter and define bounded terminal cleanup; caching every resolved pending Future cannot support that without retaining tensors. Please apply the same contract to _deliver_batch_split, which currently does not re-cache resolved pending Futures.

zhang-keliang: You're right, and thanks for spelling out both consequences — the re-cache in 1336ba1 retained every normally-waited output until shutdown and re-cached aborted ones, undoing #6413. Reworked in fd0de83 along the lines you suggested:

- `drop_output()` now registers a distinct `_DropPlaceholder` (a `Future` subclass), so the pump can tell an abort's placeholder from a claimed waiter.
- New `_finish_output()` is the single delivery contract for both the single-output path and `_deliver_batch_split`: no waiter → cache; genuine pending waiter → resolve directly, never cache; `_DropPlaceholder` → discard the tensors, terminate the placeholder with a "dropped" `RuntimeError` (so a caller that already reused it via `wait_output_ready()` wakes up), and remember the id; cancelled/done waiter → discard.
- Bounded terminal cleanup: dropped ids go into an `OrderedDict` capped at `_DROPPED_OUTPUT_IDS_MAX=4096` (LRU eviction). A late `wait_output_ready()` on a dropped id consumes the record and returns an already-failed Future instead of a fresh one — so the drop → pump → late-wait ordering can't hang, and nothing holds tensors.
- Restored the assertions that wait → pump → consume and drop → pump (no late wait) leave `_completed_outputs` empty (`test_result_pump.py:270`, and `test_prevents_caching_when_output_not_yet_arrived` again means what its name says). Added drop → pump → late wait, drop → wait → pump, cached → drop → wait, bounded-memory, and a batch-split drop-member case. 51 passed locally across the three diffusion executor test files.

vllm_omni/diffusion/executor/multiproc_executor.py
zhang-keliang: Documented in the PR body under "Abort contract" — after abort/cancel the id is terminal, a late wait fails fast (no hang, no old tensors), and diffusion abort has to be treated as whole-sample retry. Note the drop → late-wait behavior changed slightly since your review: instead of retaining the resolved placeholder (which re-introduced the #6413 retention, see hsliuustc0106's comment above), a late `wait_output_ready()` now gets an already-failed Future via a bounded dropped-id record (fd0de83).

lishunyang12:
Hello, any progress?

zhang-keliang:
Hi @lishunyang12 — all review feedback has been addressed:

- **SamitHuang's Sep 1 comments** (wait/drop ordering hang, re-caching leak): fixed in `fd0de83` with `_DropPlaceholder` and a single `_finish_output()` delivery contract. Replied inline.
- **hsliuustc0106's Sep 3 P1** (re-cache after waiter resolution): also fixed in `fd0de83`; hsliuustc0106 verified the earlier fixes at `1336ba1`.
- **fhfuih's getattr suggestion**: fixed in `a91c94e` with an abstract base `drop_output`.

CI is fully green (build 3.11 ✅, build 3.12 ✅, pre-commit ✅, DCO ✅, docs ✅). Ready for re-review whenever convenient. Thanks!

vllm_omni/diffusion/executor/multiproc_executor.py
```diff
@@ -957,19 +989,51 @@ def _result_pump(self, result_mq: MessageQueue | None = None) -> None:
...
+                fut.set_result(result)
+            self._completed_outputs[async_output_id] = fut
+        elif isinstance(pending, _DropPlaceholder):
+            try_set_exception(pending, _dropped_output_error(async_output_id))
+            self._remember_dropped(async_output_id)
+        elif not pending.done():
+            if exc is not None:
+                try_set_exception(pending, exc)
+            else:
+                try_set_result(pending, result)
```
SamitHuang, line 1044: `_finish_output` discards a cancelled/done waiter without `_remember_dropped`. After the pump pops that future, a later `wait_output_ready` on the same id allocates a fresh Future and waits forever.

That is the fully-async abort overlap: `step_streaming` has already called `wait_output_ready` (so `drop_output` leaves the waiter alone), `asyncio.wrap_future` cancel marks it done, then `OUTPUT_READY` is discarded here. The PR body says a cancelled waiter makes the id terminal and a late wait fails fast; this branch does not. Call `_remember_dropped` on the cancelled/done path (single and batch both go through here) and assert `wait_output_ready` after cancel→pump returns a completed failed Future with `_completed_outputs` empty.

vllm_omni/diffusion/executor/multiproc_executor.py
```diff
@@ -1017,15 +1082,59 @@ def _next_rpc_id(self) -> str:
...
+        with self._futures_lock:
+            if self._closed:
+                # Executor is torn down; abort path must not repopulate the
+                # cleared state (issue #6413 / #6439 review).
+                return
+            if self._completed_outputs.pop(async_output_id, None) is not None:
+                self._remember_dropped(async_output_id)
+                return
+            if async_output_id in self._output_futures:
+                return
```
SamitHuang, line 1169: This early return treats any registered future as a live waiter, including one that is already cancelled/done. Abort-after-cancel then never installs a `_DropPlaceholder` and never records the id, so the pump takes the discard-without-remember branch above.

If the existing future is already done/cancelled, pop it and `_remember_dropped` (or replace it with `_DropPlaceholder`) instead of returning. Leave a still-pending genuine waiter untouched so an in-flight `wait_for` can still complete.

vllm_omni/diffusion/executor/multiproc_executor.py
```diff
@@ -1017,15 +1082,59 @@ def _next_rpc_id(self) -> str:
...
             cached = self._completed_outputs.pop(async_output_id, None)
             if cached is not None:
                 return cached
+            # Share an already-registered Future (a genuine waiter or a
+            # drop_output placeholder) rather than clobbering it.
+            existing = self._output_futures.get(async_output_id)
+            if existing is not None:
+                return existing
+            if async_output_id in self._dropped_output_ids:
+                del self._dropped_output_ids[async_output_id]
```
SamitHuang: Deleting the LRU entry makes fail-fast one-shot: a second `wait_output_ready` on the same dropped id misses `_dropped_output_ids` and hangs on a new Future. Keep the id in the bounded OrderedDict (the 4096 cap already bounds abort traffic) so every late wait on a dropped id fails the same way.

SamitHuang (dismissed):
Re-review of cc005a5. The Sep 1 hang is closed: `wait_output_ready` reuses an existing future, `_DropPlaceholder` keeps abort distinct from a claimed waiter, and `drop → pump → late wait` fails fast without re-caching tensors. `_finish_output` is shared by single-output and batch-split, and the `_closed` / SHM-unpack gates from earlier review still hold.

The abort contract in the PR body is still not true for the cancelled-waiter path that fully-async abort actually hits (`step_streaming` `wrap_future` cancel, then a late wait on the same `async_output_id`). That ordering can still hang. Please mark cancelled/done waiters as dropped the same way `_DropPlaceholder` is, and keep the dropped-id record so fail-fast is stable.

zhang-keliang:
Addressed the three P1s in `4acf8d93`:

1. `_finish_output` (L1029) — cancelled/done waiters call `_remember_dropped` before being discarded, so late waits fail-fast instead of resolving.
2. `wait_output_ready` (L1135) — an existing done/cancelled future is now popped + remembered rather than returned as a live waiter.
3. `_dropped_output_ids` — kept as LRU with no delete; entries survive across arbitrary late waits on the same id.

Follow-on I hit while writing the test (`b52d5aec`): `_finish_output` could still cache tensors into `_completed_outputs` for an id already in the dropped set, so a later wait would see a stale success. Added a guard + a second regression test for that timeline.

pytest for the async-abort path: 4/4 GREEN locally (CPU). RED before either fix, GREEN after.

Ready for another look when you have time.

vllm_omni/diffusion/executor/multiproc_executor.py
```diff
@@ -957,19 +989,74 @@ def _result_pump(self, result_mq: MessageQueue | None = None) -> None:
...
+            fut: concurrent.futures.Future = concurrent.futures.Future()
+            if exc is not None:
+                fut.set_exception(exc)
+            else:
+                fut.set_result(result)
+            self._completed_outputs[async_output_id] = fut
+        elif isinstance(pending, _DropPlaceholder):
+            try_set_exception(pending, _dropped_output_error(async_output_id))
+            self._remember_dropped(async_output_id)
+        elif not pending.done():
```
Gaohan123, line 1040: [P2] Remember cancellation that races with result delivery

If the consumer cancels its Future after `pending.done()` returns false but before `try_set_result()` or `try_set_exception()` resolves it, the helper swallows `InvalidStateError` and this branch exits without recording the ID as dropped. The waiter has already been removed from `_output_futures`, so a subsequent `wait_output_ready()` creates a fresh Future that cannot complete, violating the documented fail-fast contract. Both single-output and batch-split delivery use this path.

I reproduced this with a real Future by pausing delivery after the `done()` check, cancelling the Future from another thread, and then resuming delivery. The subsequent waiter remained pending and the ID was absent from `_dropped_output_ids`.

Please record the dropped ID when cancellation prevents delivery—for example, have the resolution helpers report whether they succeeded—and add a regression covering this interleaving. This extends the cancellation case in https://github.com/vllm-project/vllm-omni/pull/6439#discussion_r4002843498; the already-cancelled branch is fixed, but cancellation between the check and resolution still loses the terminal state.

zhang-keliang:
Addressed the P2 in `63ff36d3`. `try_set_result` / `try_set_exception` now return whether they actually delivered; `_finish_output` calls `_remember_dropped` when a cancel races between the `.done()` check and set. Other four callers deliberately discard the bool — they're best-effort signaling contexts (rpc/shutdown/batch-split-via-_finish_output) with no live waiter contract to preserve.

Regression: `test_cancel_between_done_check_and_set_result_records_dropped` uses a `Future` subclass that cancels itself in `set_result` — deterministically forces the race window without sleeps. RED without the guard, GREEN with. Full pytest for the diffusion result-pump path: 36/36.

hsliuustc0106 (approved):
Confirming at the current head: the three shutdown findings were verified at 1336ba1 earlier, and the follow-up re-cache P1 from 09-03 is resolved by the _DropPlaceholder mechanism in fd0de83 — drop_output registers a distinct placeholder instead of re-caching, aborted outputs are discarded without undoing the #6413 fix, and the bounded remembered-ids set keeps memory in check. Mergeable and green. Good to merge.

Bounty-hunter (commented):
LGTM — the abort-drain, cancelled-waiter, shutdown, and bounded-ID state transitions are coherent, and the targeted tests cover the important orderings.

Gaohan123:
Please resolve CI failure

vllm_omni/diffusion/executor/multiproc_executor.py
zhang-keliang: Good catch — fixed in e2a7cb4d. try_set_result/try_set_exception now return a bool so _finish_output's genuine-pending-waiter branch can detect the mid-check cancellation and call _remember_dropped. Test: test_cancel_between_done_check_and_set_result_records_dropped uses a _RaceCancellingFuture subclass that cancels inside set_result.

Also rebased onto latest main (939e6ba56) to clear the flaky CI — the buildkite failures were present on main itself, not from this PR.

vllm_omni/diffusion/executor/multiproc_executor.py
zhang-keliang: Fixed — the cancelled/done branch now calls _remember_dropped(async_output_id) after discarding the waiter. Also added try_set_result/try_set_exception returning bool to detect mid-check cancellation (Gaohan123's P2). Test: test_cancelled_waiter_then_pump_then_late_wait_fails_fast_twice.

vllm_omni/diffusion/executor/multiproc_executor.py
zhang-keliang: Fixed — wait_output_ready now checks: if existing is _DropPlaceholder or done/cancelled, evict it and record as dropped before installing a fresh future. See line ~1134. Test: test_wait_output_ready_replaces_stale_cancelled_existing_future.

vllm_omni/diffusion/executor/multiproc_executor.py
zhang-keliang: Fixed — _dropped_output_ids entry is now kept (move_to_end) instead of deleted, so repeated late waits on the same dropped id fail fast every time. The 4096 cap already bounds memory. Test: test_drop_then_pump_then_late_wait_fails_fast_without_caching now asserts the LRU entry persists and a second wait also fails.

Gaohan123:
Please fix precommit

Gaohan123 (approved):
LGTM. Thanks

PR #6440, opened 2026-08-21
[skip ci][Doc] Module design doc for diffusion runtime

fhfuih:
## Purpose

Document diffusion runtime (from engine to model runner) in the new design module doc. 

This PR does not change the draft status of this doc, so it can still receive other's completion before published.

## Test Plan

NA. Doc only.


fhfuih:
CC @hsliuustc0106

docs/design/module/diffusion/diffusion_runtime.md
```diff
@@ -8,29 +8,335 @@ owners:
...
+error outputs. A dead worker group is different: the executor marks itself
+failed, health checks raise `EngineDeadError`, and the owning stage client
+handles the stage-level failure.
+
+## Execution modes
+
+The engine resolves one mode at startup and binds the matching scheduler and
+executor call.
+
+| Mode | Scheduler | Executor call | Runner path |
```
hsliuustc0106, line 193: do we have the streaming mode?

docs/design/module/diffusion/diffusion_runtime.md
```diff
@@ -8,29 +8,335 @@ owners:
...
+
+Only compatible requests enter the same running wave. The scheduler compares a
+mode-specific sampling-parameter key and stops at the first incompatible
+waiting request. This is intentionally conservative and can cause
+head-of-line blocking.
+
+## Executor and IPC
+
+`DiffusionExecutor` is the backend interface. The current built-in backend is
+`MultiprocDiffusionExecutor`, selected by `distributed_executor_backend="mp"`
```
hsliuustc0106, line 249: there is a new unitexecutor PR, please check. In addition, there is a PR working on memory leak #6413

docs/design/module/diffusion/diffusion_runtime.md
fhfuih: Added streaming mode description after this table. This table is for batching mode (request and step, only these two). Streaming depents on step mode. So it is elaborated below

docs/design/module/diffusion/diffusion_runtime.md
fhfuih: Updated the doc for uniproc executor. Also amended the doc to align with the better-established cleanup procedure as in #6413 ( PR #6439 & #6580 )

docs/design/module/diffusion/diffusion_runtime.md
fhfuih: diff in https://github.com/vllm-project/vllm-omni/pull/6440/commits/958f39c3b99e937b7d62dd17b47772b5dfd317d6

docs/design/module/diffusion/diffusion_runtime.md
fhfuih: diff in https://github.com/vllm-project/vllm-omni/pull/6440/commits/958f39c3b99e937b7d62dd17b47772b5dfd317d6

docs/design/module/diffusion/diffusion_runtime.md
```diff
@@ -8,29 +8,387 @@ owners:
...
+  - tests/diffusion/test_async_output_worker.py
+  - tests/diffusion/test_diffusion_ipc.py
+  - tests/diffusion/test_diffusion_streaming_output.py
+  - tests/diffusion/test_stage_diffusion_proc.py
+  - tests/diffusion/test_inline_stage_diffusion_client.py
 upstream_refs:
   - diffusers.DiffusionPipeline
-last_reviewed: 2026-07-16
+last_reviewed: 2026-08-25
+last_verified_commit: f7e96541dd11c21dce50fdcd9fc5b1269353ee2a
```
SamitHuang: `last_verified_commit` is `f7e96541`, a documentation commit on this PR (and not even HEAD after `958f39c`). Sibling module docs such as `engine_orchestration.md` point at the code tree that was checked. Point this at the diffusion runtime SHA you actually verified (`UniProcDiffusionExecutor`, dummy-run skip, #6413 cleanup), then update it whenever those sections change.

docs/design/module/diffusion/diffusion_runtime.md
```diff
@@ -8,29 +8,387 @@ owners:
...
+| --- | --- | --- | --- |
+| Request batch | `RequestScheduler` | `execute_batch()` | `execute_model()` for one request, or `execute_model_batch()` for a fused batch |
+| Step batch | `StepScheduler` | `execute_step()` | `execute_stepwise()` |
+
+Request mode runs a complete pipeline forward for each scheduled wave. A
+single-request wave is the conservative path. A multi-request wave usually
+uses fused `execute_model_batch()` and requires the pipeline to declare
+request-batch support. One exception is distributed layerwise offload with
+AllGather: the engine may schedule one request per DP replica and keep the
+per-request `execute_model()` path instead of a fused batch.
```
SamitHuang: This reads as scheduler/engine policy ("the engine may schedule one request per DP replica"). In `MultiprocDiffusionExecutor.execute_batch()`, the scheduler still emits a multi-request `DiffusionSchedulerOutput`; the executor then routes that wave to `execute_request()` / `execute_model` with one envelope per DP rank instead of fused `execute_model_batch()`. Describe it as an executor dispatch rule so it does not contradict the policy-vs-execution split above.

docs/design/module/diffusion/diffusion_runtime.md
```diff
@@ -8,29 +8,387 @@ owners:
...
+## Request state and identity
+
+The scheduler is the source of truth for an admitted request. Its normal state
+flow is:
+
+```mermaid
+stateDiagram-v2
+    [*] --> WAITING: add_request
+    WAITING --> RUNNING: schedule
+    RUNNING --> PREEMPTED: preempt
```
SamitHuang: `RUNNING --> PREEMPTED` is not on the engine control loop. `BaseScheduler.preempt_request()` exists, but the only in-tree caller is `tests/diffusion/test_diffusion_scheduler.py`. Drop this from the "normal state flow" diagram, or label it as an unused scheduler API, so readers do not assume the runtime preempts in-flight denoising work.

SamitHuang (changes requested):
Checked this draft against current `vllm_omni/diffusion` on main (`UniProcDiffusionExecutor`, scheduler, multiproc DLO dispatch, abort/cleanup). Three accuracy issues should be fixed before this is treated as the runtime source of truth: `last_verified_commit` points at a doc SHA from this PR, the DLO+AllGather exception is described as scheduler policy, and `PREEMPTED` is drawn as a normal engine path though only tests call `preempt_request()`.

fhfuih:
> Checked this draft against current `vllm_omni/diffusion` on main (`UniProcDiffusionExecutor`, scheduler, multiproc DLO dispatch, abort/cleanup). Three accuracy issues should be fixed before this is treated as the runtime source of truth: `last_verified_commit` points at a doc SHA from this PR, the DLO+AllGather exception is described as scheduler policy, and `PREEMPTED` is drawn as a normal engine path though only tests call `preempt_request()`.

Thanks for the careful review. Resolved all of them in the latest commit. Also rebased onto latest main

docs/design/module/diffusion/diffusion_runtime.md
```diff
@@ -8,29 +8,391 @@ owners:
...
+`MultiprocDiffusionExecutor`:
+
+1. creates a broadcast queue shared by all workers;
+2. spawns one worker process per configured device;
+3. waits until every worker reports ready;
+4. sends generation and control calls as RPC messages;
+5. applies rank-aware reply rules so only expected ranks respond; and
+6. monitors worker process sentinels and fails the executor if a worker dies.
+
+In request mode on this backend, workers always return a lightweight
```
hsliuustc0106: [P2] Do not guarantee that `COMPUTE_DONE` wins this race

`WorkerProc._return_result()` queues work to the async-output thread before it enqueues `COMPUTE_DONE`, so that thread can publish `OUTPUT_READY` first; `execute_batch()` explicitly handles this early-output ordering. Document the messages as independently produced and correlated by `async_output_id`, with either arrival order valid.

docs/design/module/diffusion/diffusion_runtime.md
hsliuustc0106: [P2] `19da238` still does not contain #6439 or #6580, while lines 351-357 describe both cleanup behaviors as current. Please either merge/rebase this document after those dependencies and set `last_verified_commit` to a tree that actually contains them, or mark those paragraphs as pending.

docs/design/module/diffusion/diffusion_runtime.md
fhfuih: The latest one commit removes this guarantee.

docs/design/module/diffusion/diffusion_runtime.md
fhfuih: > [P2] `19da238` still does not contain #6439 or #6580, while lines 351-357 describe both cleanup behaviors as current. Please either merge/rebase this document after those dependencies and set `last_verified_commit` to a tree that actually contains them, or mark those paragraphs as pending.

I rephrased relevant paragraphs in the current doc and remove clearly affirmative tone that we *does* it. If this is to be merged before those two PRs, I think the current tone is more suitable (and you can open a new comment if you disagree, lest I miss the thread here)

If you plan to merge it after those two PRs, signal me to modify these paragraphs later if you feel needed

docs/design/module/diffusion/diffusion_runtime.md
```diff
@@ -8,29 +8,394 @@ owners:
...
+  - tests/diffusion/test_async_output_worker.py
+  - tests/diffusion/test_diffusion_ipc.py
+  - tests/diffusion/test_diffusion_streaming_output.py
+  - tests/diffusion/test_stage_diffusion_proc.py
+  - tests/diffusion/test_inline_stage_diffusion_client.py
 upstream_refs:
   - diffusers.DiffusionPipeline
-last_reviewed: 2026-07-16
+last_reviewed: 2026-08-31
+last_verified_commit: 15c6ab8fa7cd12c55aa12ce9c2aedca25c2c016d
```
SamitHuang: `last_verified_commit` is still `15c6ab8` (2026-08-28). `main` is 69 commits ahead and already changed runtime files this page lists, including `BaseScheduler.initialize()` reading `max_num_seqs` (#6484), plugin load inside `StageDiffusionProc` (#6750), and runner paged-KV (#6563). Point this SHA at the diffusion tree you actually re-read (`get_class` uni/mp default, dummy-run skip, DLO dispatch), and name `max_num_seqs` as the capacity knob if that initialize() path is in the verified tree.

docs/design/module/diffusion/diffusion_runtime.md
```diff
@@ -8,29 +8,394 @@ owners:
...
+
+Request mode runs a complete pipeline forward for each scheduled wave. A
+single-request wave is the conservative path. A multi-request wave usually
+uses fused `execute_model_batch()` and requires the pipeline to declare
+request-batch support. Under distributed layerwise offload with AllGather,
+the engine raises running capacity to `dp_size` so the scheduler can admit one
+request per DP replica in a wave. The multiproc executor still receives that
+multi-request `DiffusionSchedulerOutput`, but routes it through
+`execute_request()` / per-rank `execute_model` envelopes instead of fused
+`execute_model_batch()`.
```
SamitHuang: This still under-specifies DLO+AllGather concurrency. `_init_runtime_state()` also sets `dp_concurrent` and, when `request_batch_max_wait_ms == 0`, writes `500.0` so the scheduler can fill a `dp_size` wave. `MultiprocDiffusionExecutor.execute_request()` then rejects mixed `num_inference_steps` / `extra_args` (AllGather deadlock), and `DiffusionWorker.execute_model()` selects `req[dp_rank % len(req)]`. State the wait window, the homogeneity check, and the rank pick here, not only the `execute_model_batch()` bypass.

docs/design/module/diffusion/diffusion_runtime.md
```diff
@@ -8,29 +8,394 @@ owners:
...
+
+## Lifecycle and cleanup
+
+### Startup
+
+`DiffusionEngine.make_engine()` resolves the engine class, constructs the
+executor and scheduler, then normally runs a small dummy request to warm up the
+model. Some distributed offload layouts skip this request because every rank
+must enter the same collective. If initialization or warmup fails, the runtime
+closes scheduler state and worker resources before returning the error.
```
SamitHuang, line 374: `run_startup_warmup()` does not skip for generic distributed offload. It returns early only when `enable_distributed_layerwise_offload` and `dlo_use_allgather` are set and `max(data_parallel_size, sequence_parallel_size) > 1` (SP>1 is included because OffloadConfig can derive shard size from SP). Write that predicate; otherwise readers will skip dummy-run on every multi-rank offload layout.

docs/design/module/diffusion/diffusion_runtime.md
```diff
@@ -46,14 +411,42 @@ admitting, reordering, or forwarding requests independently.
 ### DIFF-RUNTIME-INV-003: Terminal cleanup is complete
 
 **Rule:** Every terminal path MUST release request state, temporary tensors,
-hooks, and runtime-owned resources.
+hooks, and runtime-owned resources, including async-output bookkeeping when a
+request ends before its output is consumed.
```
SamitHuang: INV-002 (unchanged text just above) is stronger than the DLO path. Scheduler output is still the admission decision, but workers do not execute that list as a fused batch: each rank independently takes `req[dp_rank % len(req)]` so every rank enters the same AllGather. Carve that mapping out of INV-002, or the invariant will be read as forbidding the dispatch this page documents in Execution modes.

SamitHuang (changes requested):
Re-checked the Aug 31 draft against current `vllm_omni/diffusion` on main (`DiffusionExecutor.get_class`, `UniProcDiffusionExecutor`, scheduler, DLO dispatch, abort/cleanup). PREEMPTED, COMPUTE_DONE ordering, and the executor-side DLO routing note from the last pass are fixed. Remaining accuracy gaps: `last_verified_commit` is 69 commits behind a tree that already changed runtime files, DLO concurrency is still under-specified, dummy-run skip is vaguer than the predicate in code, and INV-002 does not admit the per-rank request pick.

docs/design/module/diffusion/diffusion_runtime.md
```diff
@@ -8,29 +8,394 @@ owners:
...
+| `UniProcDiffusionExecutor` | Single in-process worker for `num_gpus == 1`; no IPC or async output pump | Multi-GPU execution |
+| `MultiprocDiffusionExecutor` | Worker processes, shared-memory message queues, result dispatch, worker monitoring | In-process single-GPU path |
+| `WorkerProc` | One `mp` worker process's IPC loop and reply rules | Scheduling policy; unused by `uni` |
+| `DiffusionWorker` | Device/distributed setup, LoRA activation, profiling, sleep/wake, runner delegation | Request admission |
+| `DiffusionModelRunner` | Pipeline loading, request-local model state, cache/compile setup, request or step execution | Queueing and cross-stage routing |
+
+The boundary between policy and execution is
+`DiffusionSchedulerOutput`. It contains newly admitted request payloads,
+IDs for requests whose runner state is already cached, finished IDs, and
+optional KV-prefetch work. Executors and workers consume this output; they do
```
hsliuustc0106: This inventory predates paged-KV: `NewRequestData` now carries `diffusion_kv_metadata` (`vllm_omni/diffusion/sched/interface.py:177`), which both executors forward (`multiproc_executor.py:548`, `uniproc_executor.py:127`). Worth a clause here so the boundary-object list matches the #6563 tree.

docs/design/module/diffusion/diffusion_runtime.md
```diff
@@ -8,29 +8,394 @@ owners:
...
+cleanup.
+
+### Shutdown and worker failure
+
+`DiffusionEngine.close()` is idempotent. It stops the busy loop, wakes pending
+streams with an error, closes scheduler state, and shuts down the executor.
+
+- **`mp`**: the executor asks workers to stop, waits for them, then terminates
+  workers that miss the grace period. Shutdown also clears pending and cached
+  async-output state.
```
hsliuustc0106: At current main, `shutdown()` clears the pending maps (`_rpc_futures` / `_output_futures` / `_batch_split_map`) but not the `_completed_outputs` cache — that dict only dies with the executor object (`multiproc_executor.py:979-1008`). Either drop "cached" here, or say the cache is dropped along with the executor.

docs/design/module/diffusion/diffusion_runtime.md
hsliuustc0106: Verified against current main (`a65e89cb`) to save a lookup — with one difference from the comment above:

- Gate `_uses_dlo_dp_concurrency` = `dp_size > 1 and enable_distributed_layerwise_offload and dlo_use_allgather` (`diffusion_engine.py:181-188`), which sets `scheduler.max_num_running_reqs = dp_size` and `dp_concurrent = True` (`:365-366`).
- The wave-fill wait is opt-in, not defaulted: with `request_batch_max_wait_ms <= 0.0` (the default), `get_admission_wait_decision` returns `should_wait=False` (`request_scheduler.py:69-71`). I do not find the `500.0` write on main — `diffusion_engine.py:370` only logs the configured value. Under `dp_concurrent` the stable window widens to `min(0.3s, wait/2)` (`request_scheduler.py:76`).
- Homogeneity check confirmed: mixed sampling key → `ValueError` ("DLO DP multi-concurrency requires compatible shape, CFG, denoise schedule, output count, and LoRA settings"), mixed `extra_args` → `ValueError` citing AllGather deadlock (`multiproc_executor.py:484-500`).
- Rank pick confirmed: `DiffusionWorker.execute_model` takes `req[dp_rank % len(req)]` (`diffusion_worker.py:641-643`).

docs/design/module/diffusion/diffusion_runtime.md
hsliuustc0106: Answering the open question above: the requirement tone is fine to keep, but at current main nothing releases this bookkeeping at abort/consumer-drop time — release happens when the late `OUTPUT_READY` reaches the pump, and an unconsumed result is parked in `_completed_outputs` by design (#6023). The orphan-after-abort case is #6413, and #6253 / #6439 / #6580 are all still open. Suggest keeping the sentence but tagging it "(pending #6253/#6439/#6580)" so it does not read as current behavior.

docs/design/module/diffusion/diffusion_runtime.md
hsliuustc0106: Verified at 62f3c987 — the page now documents that either message may arrive first and that both orderings must be handled. Thanks.

docs/design/module/diffusion/diffusion_runtime.md
fhfuih: > Suggest keeping the sentence but tagging it "(pending https://github.com/vllm-project/vllm-omni/pull/6253/https://github.com/vllm-project/vllm-omni/pull/6439/https://github.com/vllm-project/vllm-omni/pull/6580)" so it does not read as current behavior.

Done

fhfuih:
@SamitHuang @hsliuustc0106 Resolved your recent reviews about DLO, new KV connector's metadata, and cleanup behaviors

Seems nothing to add to `diffusion_runtime.md` from today’s GMT+8 main commits till now.

Cleanup PRs (6253 / 6439 / 6580) are still open — no change to the pending tags.

PR #6441, opened 2026-08-21
[Core] Preserve configured diffusion batch size

0z5a:
## Summary

- preserve a diffusion stage's configured `max_num_seqs` when the legacy client batch size is smaller
- use the same effective batch size for inline and multiprocess stage startup
- report the effective capacity in diffusion stage initialization logs

## Motivation

Server launches still pass the legacy `diffusion_batch_size` default of one. That value currently overwrites a larger per-stage `max_num_seqs` from the deploy YAML, so the diffusion scheduler cannot accept the configured number of concurrent requests.

## Tests

```text
pytest -q tests/engine/test_async_omni_engine_stage_init.py \
  -k "merges_configured_and_client_batch_sizes"

4 passed, 34 deselected
```

### DIFF
--- a/tests/engine/test_async_omni_engine_stage_init.py
+++ b/tests/engine/test_async_omni_engine_stage_init.py
@@ -399,11 +399,20 @@ def test_initialize_local_llm_replica_scopes_runtime_env(monkeypatch):
     assert os.environ[runtime_env_var] == "parent-value"
 
 
-def test_initialize_diffusion_stage_applies_client_batch_size_to_engine(monkeypatch):
+@pytest.mark.parametrize(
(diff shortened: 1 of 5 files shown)


wtomin:
I think this PR solves the same problem as #6405. Pls also take a look @knlnguyen1802

wtomin:
@0z5a  Do you use `AsyncOmni(...)` or the `vllm serve` entrypoint? I would like to know the behavior changes before and after this PR,

0z5a:
> @0z5a Do you use `AsyncOmni(...)` or the `vllm serve` entrypoint? I would like to know the behavior changes before and after this PR,

ok, I see.

PR #6444, opened 2026-08-21
[Benchmark] Add a multi-session realtime AR-Diffusion load harness

linzhenpl07:
## Summary

There is no way to measure realtime AR-Diffusion serving in tree today. `examples/offline_inference/diffusion/lingbot_world_v2_realtime.py` is a single-session sequential loop, and `benchmarks/` has nothing session-oriented, so none of the numbers a streaming video deployment is gated on have a baseline.

This adds `benchmarks/ar_diffusion`, the Phase 0 harness from [#6227](https://github.com/vllm-project/vllm-omni/issues/6227) (deliverables 1 and 4: *common benchmark harness* and *explicit TTFF/chunk-latency/RTF metrics*).

| Module | Needs a device? |
|---|---|
| `realtime_metrics.py` — metric definitions, pure functions | no |
| `realtime_harness.py` — load model, arrivals, pacing, recording | no |
| `run_realtime_benchmark.py` — CLI; argument-to-config path has no engine imports | no |
| `engine_binding.py` — AsyncOmni + session manager wiring | **yes** |

Only the last one needs a checkpoint. Everything the benchmark actually measures is tested on CPU.

## Design decisions worth reviewing

**Session arrival rate is the only open-loop axis.** Per-session tick rate is not one: ticks of a session are strictly ordered and state-dependent — tick `k+1` reads the state committed by tick `k` and cannot be issued before chunk `k` returns — so raising it only grows a queue. Active session count is derived state, reported as `peak_concurrent_sessions` rather than driven.

**Two load modes, both required.** `saturating` measures the ceiling and has no deadlines. `paced` measures continuity and is the only mode in which an SLO is defined. Running only the first produces a throughput number no deployment can use; running only the second hides the ceiling, so a miss cannot be attributed to capacity rather than scheduling.

**CPR is reported next to RTF, not derived from it.** RTF is a ratio of aggregates: a session can hold `RTF = 0.9` over a minute and still stutter visibly, because chunks arrive in bursts that individually miss their playout instants. CPR is macro-averaged over sessions so one lagging session is not diluted by a busier neighbour — there is a test for exactly that.

**Deadline model.** Playback starts once `--buffer-chunks` chunks are buffered; chunk `buffer_chunks - 1` anchors the grid, and every later chunk is due one release period after its predecessor. Chunks inside the prebuffer have no deadline — their cost is start latency, which TTFC already reports.

**`compare_runs()`** derives `per_tick_switching_cost_s` as the latency beyond what pure serialization predicts. With state concurrency `N` and execution concurrency 1, ticks serialize, so `aggregate FPS(N) ≈ FPS(1)` and `RTF(N) ≈ N × RTF(1)`. A measurable drop in aggregate FPS is per-tick switching cost — state bind/unbind, KV pool paging, conditioning rebuild — and that number is the baseline any cross-session batching work has to beat.

**Run conditions are mandatory.** At least one `--note` is required and the values are copied verbatim into the summary. A latency without model, checkpoint, resolution, denoising steps and hardware is not a result.

## One gap this surfaced

The harness is model-neutral: the chunk shape is read from the pipeline's declared `ARDiffusionKVCacheSpec` and no model name appears in the directory.

But **one conversion cannot come from the capability today.** `frames_per_block`, `window_frames` and `sink_frames` are all in *latent* frames, and `ARDiffusionKVCacheSpec` declares nothing that converts them to delivered frames. So no runtime component can compute the playout grid — release period, deadlines, RTF, generated FPS — from the capability alone.

**And the conversion is not a multiplication.** A causal video decoder expands a session's first latent frame to one raw frame and every later one to the full temporal factor:

| | latent frames | raw frames |
|---|---|---|
| chunk 0 | 3 | **9** |
| chunk k > 0 | 3 | 12 |
| K chunks | 3K | **12K − 3** |

The reference pipeline's own geometry confirms it — `num_latent_frames = (num_frames - 1) // factor + 1`, and its `_MAX_RAW_FRAMES = 117` horizon is exactly 10 chunks under this mapping.

I had this wrong in the first revision of this PR: the profile multiplied a constant, so every run over-counted by the difference. Fixed in `58357fc`, which also corrects the deadline model — it walked a constant release period, crediting the shorter opening chunk with a full period and giving a deeper buffer no extra slack.

The practical consequence for the spec: **the missing piece is a mapping, not a factor**, so declaring a single integer would not close it. `vae_temporal_factor` arrives as a flag (`--vae-temporal-factor`, default `1`; `--non-causal-decoder` selects the uniform mapping) and `frames_per_chunk_from_spec` returns the pair.

Worth noting how small the fix would be if the spec should carry this: `LingBotWorldCausalDMDPipeline.ar_diffusion_kv_cache_spec()` already has `self.vae_scale_factor_temporal` in scope — it uses it two lines earlier to compute `horizon_latent_frames` and `condition_channels`. Happy to add it here or separately if there is a preferred shape.

Relatedly, `resident_bytes_per_session` is reported from `model_owned_state_bytes_per_session` only; runner-owned KV bytes are not in the spec, so that field under-reports total residency.

## Test

`tests/diffusion/ar_diffusion/test_realtime_benchmark.py` — **32 tests, CPU only**, ruff check and format clean. The load model runs on a virtual clock against fake sessions, so the suite spends no wall time and needs no engine, device or checkpoint.

Coverage:

- **Causal geometry** — the opening chunk is shorter; the uniform profile stays uniform by default; the opening chunk is not credited with a full period of playback; spec derivation produces both values.
- **Metric definitions** — release period from the profile rather than a flag; nearest-rank percentiles (a reported P99 should be a latency that occurred); deadline anchoring and prebuffer exclusion; deeper buffers moving the grid; CPR counting only chunks that had a deadline; saturating mode reporting no continuity metrics; RTF; macro- vs micro-averaged CPR; worst-case latency across rather than within sessions; frames-per-GPU-second normalization.
- **Load model** — saturating issues the next tick immediately; paced holds the release period when generation is faster; paced records misses when generation is too slow; a lost session is recorded as a result and still released; peak concurrency reflects arrivals; staggered arrivals lower it; Poisson arrivals deterministic per seed.
- **Outputs** — events JSONL carries deadlines in paced mode and omits them for prebuffer chunks; the summary is JSON-serializable and carries its conditions.
- **The N=1 vs N=2 comparison** the harness exists to produce — serialized ticks keep aggregate FPS flat and double per-session RTF (asymptotically, since the first chunk meets no contention), and switching cost is the latency beyond pure queueing.

## Follow-up

This is the harness only. Producing the N=1 and N=2 numbers needs a card that holds two live sessions, which by the memory arithmetic in [#6227](https://github.com/vllm-project/vllm-omni/issues/6227#issuecomment-5364804077) is ~75 GB for the reference workload — 80 GB devices leave nothing for activations. Those numbers will be posted to the RFC once run.

### DIFF
--- a/benchmarks/ar_diffusion/engine_binding.py
+++ b/benchmarks/ar_diffusion/engine_binding.py
@@ -1,0 +1,88 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+"""Wire the benchmark harness to a real AR-Diffusion engine.
+
+This is the only module here that needs a device, a checkpoint and a CUDA
+build, and therefore the only one the CPU tests cannot cover. Everything the
+benchmark actually measures -- the load model, the deadline model and every
+metric -- lives in ``realtime_harness`` and ``realtime_metrics`` and is tested
+without it.
+
+The wiring mirrors ``examples/offline_inference/diffusion/lingbot_world_v2_realtime.py``
+but drops everything model-specific: no control reducer, no action schema, no
+fixed frame count. A pipeline needing per-tick controls should extend
+``prompt_provider``/``control_reducer_factory`` rather than change the harness.
+"""
+
+from __future__ import annotations
+
+import argparse
+from dataclasses import dataclass
+from typing import Any
+
+
+@dataclass(frozen=True)
+class RealtimeBackend:
+    """The session manager under test, plus the capability it declared."""
+
+    manager: Any
+    spec: Any
+    engine: Any
+
+
+async def build_realtime_backend(args: argparse.Namespace) -> RealtimeBackend:
+    """Build an AsyncOmni engine and a session manager from CLI arguments."""
+    from vllm_omni.entrypoints.async_omni import AsyncOmni
+    from vllm_omni.experimental.ar_diffusion.consumer import ARDiffusionOmniTickConsumer
+    from vllm_omni.experimental.ar_diffusion.session import (
+        ARDiffusionSessionManager,
+        ARDiffusionWorkerLifecycle,
+    )
+    from vllm_omni.inputs.data import OmniDiffusionSamplingParams
+
+    engine = AsyncOmni(
+        model=args.model,
+        engine_backend="vllm_omni.experimental.ar_diffusion.engine.ARDiffusionEngine",
+        enforce_eager=args.enforce_eager,
+        tensor_parallel_size=args.tensor_parallel_size,
+        max_num_seqs=1,
+        model_config={
+            "ar_diffusion_kv_config": {
+                "gpu_memory_fraction": args.gpu_memory_fraction,
+                "warmup_cudagraph": not args.enforce_eager,
+            },
+        },
+    )
+
+    sampling = OmniDiffusionSamplingParams(seed=args.seed, output_type="latent")
+
+    def prompt_provider(tick: Any) -> dict[str, Any]:
+        prompt: dict[str, Any] = {"prompt": tick.prompt or args.prompt}
+        if args.image is not None:
+            prompt["multi_modal_data"] = {"image": str(args.image)}
+        return prompt
+
+    consumer = ARDiffusionOmniTickConsumer(
+        engine,
+        prompt_provider=prompt_provider,
+        sampling_params_list=[sampling],
+        diffusion_stage_id=0,
+    )
+    manager = ARDiffusionSessionManager(
+        tick_consumer=consumer,
+        lifecycle=ARDiffusionWorkerLifecycle(engine, stage_ids=[0], timeout=180.0),
+        max_pending_events=args.max_pending_events,
+    )
+    return RealtimeBackend(manager=manager, spec=_declared_spec(engine), engine=engine)
+
+
+def _declared_spec(engine: Any) -> Any:
+    """Fetch the pipeline's declared AR-Diffusion KV spec from the engine."""
+    pipeline = getattr(getattr(engine, "engine", None), "pipeline", None)
+    spec_fn = getattr(pipeline, "ar_diffusion_kv_cache_spec", None)
+    if spec_fn is None:
+        raise RuntimeError(
+            "The selected model does not implement SupportsARDiffusionPipeline, "
+            "so its chunk shape and session capacity are not declared."
+        )
+    return spec_fn()
--- a/benchmarks/ar_diffusion/README.md
+++ b/benchmarks/ar_diffusion/README.md
@@ -1,0 +1,97 @@
+# Realtime AR-Diffusion session benchmark
+
+Multi-session load harness for realtime AR-Diffusion serving: session arrivals,
+playout deadlines, and the metrics a streaming video deployment is actually
+gated on.
+
+## Load modes
+
+| Mode | Question it answers | Deadlines |
+|---|---|---|
+| `saturating` | How fast can this configuration go at all? Aggregate generated FPS, frames per GPU-second. | none |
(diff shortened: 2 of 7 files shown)


linzhenpl07:
Self-review.

**What I verified locally**

- 32 tests pass on CPU, `ruff check` and `ruff format --check` clean. The suite drives the whole load model on an injected virtual clock against fake sessions, so it spends no wall time and needs no engine, device or checkpoint.
- Every metric has a test that would fail if the definition were wrong, not just if the code crashed. The ones I care most about: CPR macro- vs micro-averaged (micro reports ~0.95 where macro reports 0.5 for the same data), worst-case latency taken across sessions rather than within one, and saturating mode reporting `None` for continuity rather than a misleading number.
- `test_serialized_ticks_keep_aggregate_fps_flat_and_double_rtf` encodes the prediction this harness exists to check, so a wrong result on hardware will be legible rather than ambiguous.

**What I could not verify**

`engine_binding.py` — it needs a device and a checkpoint. It is deliberately the only module with those imports, and it contains no measurement logic; everything reported comes from the two tested modules. It mirrors the wiring in `examples/offline_inference/diffusion/lingbot_world_v2_realtime.py` with the model-specific parts removed. Treat it as unvalidated until Step B runs.

Also unverified: the produced numbers themselves. This PR is the harness only. Running it needs a device that holds two live sessions concurrently, which for the reference workload is ~75 GB.

**A bug I found in my own first revision**

The initial version computed delivered frames as `chunks * frames_per_chunk`. That is wrong for a causal video decoder: the first latent frame of a session expands to one raw frame, so an opening block of 3 latent frames delivers 9 raw frames and every later block delivers 12 — `12K - 3` for K chunks, not `12K`. Every run over-counted.

I caught it by checking the mapping against the reference pipeline rather than trusting my own earlier assumption, and `_MAX_RAW_FRAMES = 117` confirms it at exactly 10 chunks. Fixed in `58357fc`, along with the deadline model, which had the same cause: it walked a constant release period, so the shorter opening chunk was credited with a full period and a deeper `--buffer-chunks` granted no additional slack.

**What I would most like a reviewer to push back on**

1. The deadline model. `deadline(i) = t_ready(buffer_chunks - 1) + cumulative_frames(i) / target_fps` assumes playback begins when the buffer fills and never stalls afterwards. That is the standard model and it matches the CPR definition this follows, but a deployment that re-buffers after a miss would want different semantics.
2. `per_tick_switching_cost_s` in `compare_runs()` attributes all latency beyond `N x baseline` to switching. That is the right first-order model when ticks serialize, but it charges any other source of growth to the same bucket.
3. Whether the latent-to-delivered-frames mapping belongs in `ARDiffusionKVCacheSpec`. It is a flag here on purpose; the PR description explains why a single integer cannot express it.

benchmarks/ar_diffusion/engine_binding.py
```diff
@@ -0,0 +1,88 @@
...
+        tick_consumer=consumer,
+        lifecycle=ARDiffusionWorkerLifecycle(engine, stage_ids=[0], timeout=180.0),
+        max_pending_events=args.max_pending_events,
+    )
+    return RealtimeBackend(manager=manager, spec=_declared_spec(engine), engine=engine)
+
+
+def _declared_spec(engine: Any) -> Any:
+    """Fetch the pipeline's declared AR-Diffusion KV spec from the engine."""
+    pipeline = getattr(getattr(engine, "engine", None), "pipeline", None)
```
david6666666, line 81: `AsyncOmni.engine` is an `AsyncOmniEngine`, which has no `pipeline`; the pipeline lives in the diffusion worker. This therefore raises even for LingBot, before the benchmark can start. Please retrieve the capability through worker RPC/engine metadata and cover this binding with a real-engine smoke test.

PR #6427, opened 2026-08-21
[Core] Add explicit post-load Host Weight Runtime publication

hsliuustc0106:
## Purpose

Follow up on merged #6419 and the sequencing agreed in #6414 with the small shared-runtime contract required by post-load Host Weight Runtime producers.

A final-layout producer may depend on a canonically loaded and finalized model, so it cannot honestly run through the pre-load `resolve()` path. Calling the store directly would also bypass runtime-owned policy and observability.

This PR therefore:

- adds explicit synchronous `publish_after_load()` support for `POST_LOAD_ONLY` producers;
- keeps the completed pre-load resolution unchanged and reports publication separately through `PostLoadPublicationReport`;
- makes post-load publication warm future startups only: the runtime closes the validated store lease before returning and never restores, rebinds, or mutates the canonically loaded model serving the current startup;
- keeps pre-load `allow_local_build` and post-load `allow_post_load_publish` policy independent;
- reports disabled, policy-disabled, published, already-present, joined, and typed failure outcomes through the publication observer; and
- documents the publication, restoration, ownership, and failure boundaries.

Pre-load resolution behavior, storage semantics, model restoration, and GPU transport remain unchanged.

This PR also renames the Host Weight Runtime resolution-policy test module to a unique basename, avoiding the pytest import collision reported in #6436.

## Test Plan

**vLLM Version:** 0.27.0

**vLLM-Omni Commit:** 33976104

## Test Result

- `python -m pytest -o addopts='' -q tests/host_weight_runtime`: 81 passed, 14 warnings in 36.26s.
- `python -m pytest -o addopts='' --collect-only -q tests/e2e/features/fullduplex/test_runtime.py tests/host_weight_runtime/test_runtime_resolution.py`: 30 tests collected without an import mismatch.
- All changed-file pre-commit hooks passed, including Ruff, formatting, mypy 3.10, Markdown lint, test-mark validation, SPDX, forbidden imports, and the `torch.cuda` policy check.

Relates to #6414.

Unblocks replacing #6240's private cache contract with the shared Host Weight Runtime.

Fixes #6436.

PR #6445, opened 2026-08-21
[Diffusion] Add final-layout BF16 Host Weight Runtime artifacts

hsliuustc0106:
## Purpose

This is PR 1 of the replacement series for the private DLO host-weight cache in #6240. It builds on the neutral Host Weight Runtime foundation in #6419 and the warm-only post-load publication contract in #6427.

The PR defines representation-independent final-layout identity and restoration mechanics for diffusion models, plus the first concrete BF16-with-preserved-FP32 artifact policy and producer for MiniMax H3. It does not activate the artifact in the model loader or DLO transport.

### Shared final-layout contracts

- Adds typed `FinalLayoutRequest`, loader identity, and parallel identity contracts. DP rank, SP rank, device identity, DLO transfer mode, registration policy, and store path cannot enter through an open metadata bag.
- Splits canonical source resolution and transaction guards into a reusable source-identity module, including explicit checkpoint-adapter ABI identity.
- Requires every owned tensor to bind to exactly one canonical source. Longest-prefix matching resolves intentional broad/specific overlap, equal-specificity ties fail, and target-to-source bindings participate in source identity.
- Adds `FinalLayoutArtifactSpec` and `FinalLayoutTensorPolicy` so representation, tensor roles, runtime layout, manifest schema, and producer/restorer ABI are policy-defined.
- Uses an explicit canonical implementation ABI descriptor instead of reflective `inspect.getsource()` hashing.
- Makes tensor discovery representation-independent while retaining complete alias, storage, shape, stride, device, role, and ownership validation.
- Adds a generic exact-identity `FinalLayoutTensorRestorer` whose planning phase is validation-only and whose commit is one-shot.
- Maps source and producer contract failures into stable Host Weight Runtime failure codes.

### Concrete BF16 policy

- Adds a `POST_LOAD_ONLY`, `SINGLE_PROCESS` BF16 producer that writes finalized CPU tensors through the store-scoped artifact writer.
- Preserves BF16 parameters plus model-required FP32 parameters and persistent buffers.
- Makes MiniMax H3 declare a dtype-neutral, versioned `FinalLayoutModelContract` and revalidate its mixed-precision invariants after restore.
- Keeps BF16 concrete only in `FinalLayoutBF16Policy` and `FinalLayoutBF16Producer`. Future FP8/Cutlass layouts must provide their own scale roles, quantization metadata, physical-layout policy, and producer ABI.

## Scope boundary

This PR intentionally does not add:

- loader activation or ordinary-load skipping;
- public CLI or DLO configuration;
- disabled/preferred/required consumer precedence;
- lease handoff to DLO;
- bounded staging, CUDA registration, or H2D changes;
- checkpoint-mmap changes;
- an FP8 or quantized producer; or
- any DLO AllGather behavior change.

PR 2 will own the complete cold/warm loader transaction and bounded no-AllGather staging. PR 3 will own registered mmap/direct H2D and ordered teardown.

## Validation

- `python -m pytest -o addopts='' -q tests/host_weight_runtime tests/diffusion/model_loader/test_final_layout_host_weights.py tests/diffusion/models/minimax_h3/test_minimax_h3_parallel.py`
  - 118 passed, 14 warnings in 34.72s.
- All changed-file pre-commit hooks passed, including Ruff, formatting, mypy 3.10, Markdown lint, test-mark validation, SPDX, forbidden imports, and the `torch.cuda` policy check.

The CPU suite covers:

- CPU/meta identity equivalence and immutable source resolution;
- complete multi-component source coverage, deterministic longest-prefix resolution, and ambiguous-overlap rejection;
- typed loader, checkpoint-adapter, model, producer, and restorer ABI invalidation;
- TP/SP semantic separation and structural DP/transport exclusion;
- a synthetic second representation policy proving the shared path is not BF16-specific;
- ownership discovery against a reduced real MiniMax-H3 model;
- mixed BF16/FP32 tensor ownership and unsupported-layout rejection;
- deterministic publication across physical shard layouts;
- source-race rejection with stable failure codes;
- validation-only planning, exact-identity restoration, one-shot commit, and publication-lease cleanup.

No GPU benchmark is claimed or required because this PR does not connect a transport consumer.

Relates to #6414.

Replacement series for #6240.


vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
```diff
@@ -889,6 +890,12 @@ def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
 
 
 class MiniMaxH3DiTModel(nn.Module):
+    # Loading is tensor-complete: constructor state plus final-layout
+    # parameters and persistent buffers is sufficient to reconstruct a ready
+    # inference model. The model-specific validator below checks the preserved
+    # FP32 portion after a lease-backed restore commits.
+    host_weight_restore_contract = "vllm-omni.diffusion.final-layout-bf16-v1"
```
hsliuustc0106: why we fix it as bf16?

vllm_omni/diffusion/model_loader/host_weights/producers/final_layout_bf16.py
```diff
@@ -0,0 +1,206 @@
...
+    layout = identity.layout.metadata.to_value()
+    if not isinstance(component, dict) or not isinstance(layout, dict):
+        raise FinalLayoutContractError(
+            "identity_incompatible",
+            "final-layout BF16 identity metadata must use JSON objects",
+        )
+    return component, layout
+
+
+def validate_final_layout_bf16_identity(
```
hsliuustc0106: do we have other layouts?

vllm_omni/diffusion/model_loader/host_weights/identity_adapter.py
```diff
@@ -0,0 +1,471 @@
...
+
+
+def _require_positive_int(name: str, value: object) -> int:
+    if isinstance(value, bool) or not isinstance(value, int) or value < 1:
+        raise ValueError(f"{name} must be a positive integer")
+    return value
+
+
+@dataclass(frozen=True)
+class FinalLayoutBF16Request:
```
hsliuustc0106: why the adapter needs to name with dtype? this is quite not scalable

vllm_omni/diffusion/model_loader/host_weights/identity_adapter.py
```diff
@@ -0,0 +1,471 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM-Omni project
+"""Exact pre-load identity for final-layout BF16 diffusion artifacts."""
```
hsliuustc0106: do we only have the bf16 layout?

vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py
hsliuustc0106: Agreed. The model ABI is now dtype-neutral through `FinalLayoutModelContract` (`final-layout-tensors-v1`); BF16 remains only in the concrete policy and producer.

vllm_omni/diffusion/model_loader/host_weights/producers/final_layout_bf16.py
hsliuustc0106: Yes. `FinalLayoutArtifactSpec` and `FinalLayoutTensorPolicy` now make layout, tensor roles, schemas, and ABI policy-defined; FP8/Cutlass can supply its own scale/layout policy without duplicating identity or restoration.

vllm_omni/diffusion/model_loader/host_weights/identity_adapter.py
hsliuustc0106: Agreed. The shared API is now `FinalLayoutRequest`, `FinalLayoutIdentityContext`, and `build_final_layout_identity`; dtype appears only in the concrete representation policy/producer.

vllm_omni/diffusion/model_loader/host_weights/identity_adapter.py
hsliuustc0106: BF16 is now only the first concrete policy. The generic path is also exercised with a synthetic FP32 policy, while future quantized layouts provide their own roles, metadata, and producer ABI.

hsliuustc0106:
@vllm-omni-review-bot

vllm_omni/diffusion/model_loader/host_weights/source_identity.py
```diff
@@ -0,0 +1,307 @@
...
+
+def _blob_content_id(path: Path) -> str | None:
+    if not path.is_symlink():
+        return None
+    try:
+        target = os.readlink(path)
+    except OSError:
+        return None
+    blob_name = Path(target).name
+    if _IMMUTABLE_REVISION_RE.fullmatch(blob_name) is None:
```
hsliuustc0106, line 177: `_blob_content_id()` trusts any symlink whose target basename is 40/64 hex as immutable. That is false for arbitrary local roots: rewriting a same-sized target under the same hex-looking name produced the same identity on a fresh resolution, so a warm startup can return a stale artifact; the process-local change guard only protects the old context. Please restrict this shortcut to validated Hugging Face snapshot→blob links, or hash every other symlink target, and add a cross-startup regression test.

hsliuustc0106 (commented):
One source-identity blocker.

PR #6446, opened 2026-08-21
[Docs] Restore the Star History chart

congw729:
## Summary

- restore the live Star History chart in the README
- use a sealed GitHub token so Star History can access the repository's stargazer data under GitHub's API restrictions
- support light and dark color schemes

## Testing

- `env npm_config_cache=/private/tmp/codex-vllm-omni-npm-cache pre-commit run --files README.md`
- `git diff --check`
- verified the chart endpoint returns HTTP 200 with `image/svg+xml`


congw729:
@ywang96 Hi Roger

Issue #6448, opened 2026-08-21
[Performance]: GR00T-N1.7 never uses torch.compile (always runs eager), ~30% slower than it could be


hyunnnchoi:
### Proposal to improve performance

GR00T-N1.7 always runs eager, and there is no config that changes it.
`enforce_eager: false` is a no-op for this model.

The runner looks for `setup_compile()`, `_dit_modules`, or `.transformer` on the pipeline:

```python
# vllm_omni/diffusion/worker/diffusion_model_runner.py:315-330
if not self.od_config.enforce_eager:
    if hasattr(self.pipeline, "setup_compile"):
        self.pipeline.setup_compile()
    else:
        transformer_attrs = getattr(self.pipeline, "_dit_modules", None) or ("transformer", "transformer_2")
        for attr_name in transformer_attrs:
            self._compile_transformer(attr_name)
```

`Gr00tN1d7Pipeline` has none of the three. It holds `self.policy`, a `Gr00tPolicy` around an HF `AutoModel`, so every branch does nothing.

That matters because the model is launch-bound. nsys on one steady-state request, 1xA30, DP=1, captured after 10 warmup requests:

| section | GPU busy | kernels |
|---|---|---|
| preprocess (host) | 0 ms | 0 |
| backbone (Qwen3-VL) | 33.8 ms | 1,925 |
| denoise (4 steps) | 34.1 ms | 3,099 |
| whole request | 70.1 ms | 5,086 |

Median kernel duration is 6.0 us. The request takes 143.7 ms, so the GPU is idle for about half of it.

I added a `setup_compile()` that compiles the action head and measured it. 
30 requests, warmup 20.

| | latency | throughput |
|---|---|---|
| eager | 146.9 ms | 7.49 req/s |
| compile, `mode=default` | 102.9 ms | 10.69 req/s |
| compile, `mode=reduce-overhead` | 99.3 ms | 11.06 req/s |

Actions still match the reference values in `tests/e2e/online_serving/test_gr00t_openpi_expansion.py` at atol 1e-2 in all three cases.

Most of that is inductor fusion rather than CUDA graphs: fusion alone accounts for 44 ms of the 47.6 ms. That fits the kernel mix, since 3,446 of the 5,086 launches are ~6 us elementwise and layernorm kernels doing only 29% of the GPU work.

[...]

PR #6450, opened 2026-08-21
[Model] GR00T-N1.7: add setup_compile() for the action head

hyunnnchoi:
## Purpose

Fixes the problem reported in #6448: GR00T-N1.7 always runs eager because `Gr00tN1d7Pipeline` implements none of the hooks the runner looks for, so `enforce_eager: false` does nothing.

```python
# vllm_omni/diffusion/worker/diffusion_model_runner.py:315-330
if not self.od_config.enforce_eager:
    if hasattr(self.pipeline, "setup_compile"):
        self.pipeline.setup_compile()
    else:
        transformer_attrs = getattr(self.pipeline, "_dit_modules", None) or ("transformer", "transformer_2")
        for attr_name in transformer_attrs:
            self._compile_transformer(attr_name)
```

The pipeline holds `self.policy`, a `Gr00tPolicy` around an HF `AutoModel`, so all three branches are no-ops. `_dit_modules` cannot express it either, since `_compile_transformer` does a flat `getattr(self.pipeline, attr_name)` and the hot modules live at `policy.model.action_head.{model,action_encoder,action_decoder}`. That leaves `setup_compile()`.

Changes:

- `pipeline_gr00t.py`: add `setup_compile()`, compiling `action_head.model`, `action_head.action_encoder`, and `action_head.action_decoder` with `dynamic=False`. Each target is guarded, so a missing or failing module logs a warning and stays eager.
- `Gr00tN1d7.yaml`: comment only. `enforce_eager` stays `true`, so nothing changes for existing users.
- `recipes/NVIDIA/GR00T-N1.7.md`: an optional section with the overlay, the numbers, and the caveats.

Opting in:

```yaml
base_config: vllm_omni/deploy/Gr00tN1d7.yaml
stages:
  - stage_id: 0
    enforce_eager: false
    compile_mode: default          # or reduce-overhead
```

`compile_mode` is read from the stage rather than from `model_config`. An overlay that sets `model_config:` replaces the base block instead of merging it, since `_DEEP_MERGE_KEYS` covers `engine_extras` and `engine_args` but not `model_config`. That drops `policy_server_config`, and the server then logs `Robot OpenPI serving disabled` at INFO, still accepts the WebSocket, and the client dies on the msgpack handshake. A stage-level key rides in `engine_extras` and merges correctly. The recipe warns about this; `model_config.compile_mode` still works for direct edits of the base YAML.

## Test Plan

**vLLM Version:** 0.26.0+cu129

**vLLM-Omni Commit:** based on `0168613`

1xA30, DP=1, 30 requests per client, warmup 20. Both paths were exercised: the shipped YAML untouched, and the opt-in overlay. Actions were checked against the reference values in `tests/e2e/online_serving/test_gr00t_openpi_expansion.py` at atol 1e-2 before each benchmark, with `GR00T_NOISE_SEED=42`.

## Test Result

| | latency | throughput |
|---|---|---|
| eager (shipped default) | 146.9 ms | 7.49 req/s |
| overlay, `compile_mode: default` | 102.9 ms | 10.69 req/s |
| overlay, `compile_mode: reduce-overhead` | 99.3 ms | 11.06 req/s |

Actions match the reference values in all three.

The default path is untouched. Serving with the bundled YAML emits no `setup_compile` line and reproduces the eager baseline:

```
[default] setup_compile log lines: 0
RESULT: PASS
clients   done   thru/s     mean      p50      p95
      1     25     7.09    141.0    140.7    146.6
```

The opt-in path works through the overlay alone, with no environment variables:

```
GR00T setup_compile: compiled action head with mode=default.
=== reference check (atol=0.01, steps [0, 4, 39]) ===
  eef_9d             max|diff| = 6.147e-03  ok
  gripper_position   max|diff| = 3.906e-03  ok
  joint_position     max|diff| = 7.047e-03  ok
  RESULT: PASS
clients   done   thru/s     mean      p50      p95
      1     25     9.56    104.5    104.3    105.6
```

## Notes

`default` rather than `reduce-overhead` is the documented starting point. Fusion alone accounts for 44 ms of the 47.6 ms, CUDA graphs add the last 3.6 ms, and `default` has none of the graph static-address constraints. `reduce-overhead` stays available for anyone who measures a reason to use it.

For `reduce-overhead` the code turns off `cudagraph_trees`. With it on, every inference fails:

```
torch/_inductor/cudagraph_trees.py:2608 dealloc_current_path_weakrefs
AssertionError: len(node.tensor_weakrefs) == len(node.stack_traces)
```

Three compiled regions run back to back inside one `torch.inference_mode()` block, and the warmup teardown trips on that pattern. The flat cudagraphify path keeps no such bookkeeping. Seen on torch 2.11.0+cu129, and only applied when `reduce-overhead` is selected.

The Qwen3-VL backbone is left eager on purpose. CUDA graph capture fails on it (`compile_fx.py:1966 assert dst.data_ptr() == src.data_ptr()`) because its token count varies per observation, and fusion alone moves `eef_9d` to 1.5e-2 against the 1e-2 tolerance while being slower than compiling the action head alone (107.7 ms vs 99.3 ms).

Compilation happens on the first inference, so the first request after enabling the overlay is slow. The default path does not compile at all, so CI should be unaffected.

One incidental line: the SPDX pre-commit hook rewrote `pipeline_gr00t.py`'s copyright header from `vLLM project` to `vLLM-Omni project`. Happy to drop it if you would rather keep the diff clean.

### DIFF
--- a/vllm_omni/diffusion/models/gr00t/pipeline_gr00t.py
+++ b/vllm_omni/diffusion/models/gr00t/pipeline_gr00t.py
@@ -1,5 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
-# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM-Omni project
 
 from collections.abc import Iterable, Mapping
 from typing import Any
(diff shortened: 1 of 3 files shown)


hyunnnchoi:
Self-review.

`Gr00tN1d7Pipeline` implemented none of the hooks the runner looks for, so `enforce_eager: false` did nothing — GR00T ran eager regardless of what the config said. This adds `setup_compile()` so that knob does what it claims. That's the whole change. The default stays eager and the YAML diff is comments only, so nothing moves for existing users unless they opt in.

What I checked: with the shipped YAML, `setup_compile()` is never called and the run reproduces the eager baseline (141.0 ms / 7.09 req/s). With the overlay, actions still match the golden values in `test_gr00t_openpi_expansion.py` at atol 1e-2, on both `default` and `reduce-overhead`. Each compile target is guarded individually, so a missing or failing module logs a warning and stays eager instead of taking model load down with it. Only the action head is compiled — its shapes are fixed by the checkpoint, so `dynamic=False` holds. The backbone stays eager on purpose; reasons and numbers are in the description.

Two calls I'd rather a maintainer made:

- `reduce-overhead` requires setting `torch._inductor.config.triton.cudagraph_trees = False`, which is process-global rather than local to this pipeline. `default` carries 92% of the win without it, so I'm happy to cut `reduce-overhead` and keep this to `default` only.
- The SPDX hook rewrote this file's header to `vLLM-Omni project`. It matches ~120 other files in the repo so I left it, but I'll revert if you want the diff limited to the feature.

cc @fhfuih @Bounty-hunter @wtomin @RuixiangMa

hyunnnchoi:
Rebased onto `main` (was 30 commits behind) and force-pushed: `176dbf3` -> `5698bc2`. Clean rebase, no conflicts — nothing upstream had touched any of the three files — and the diff is unchanged from before.

cc @hsliuustc0106

vllm_omni/diffusion/models/gr00t/pipeline_gr00t.py
```diff
@@ -49,6 +49,87 @@ def __init__(self, *, od_config: OmniDiffusionConfig, prefix: str = "") -> None:
...
+        #
+        # Prefer the stage-level ``compile_mode`` (an unrecognised stage key, so it
+        # rides in ``engine_extras``, which base_config overlays deep-merge). A
+        # ``model_config.compile_mode`` is honoured too for direct edits of this
+        # model's deploy YAML, but note that an overlay setting ``model_config:``
+        # replaces the base block wholesale rather than merging it -- only the
+        # keys in _DEEP_MERGE_KEYS are merged -- which would drop
+        # ``policy_server_config`` and silently disable OpenPI serving.
+        self.compile_mode = str(
+            getattr(od_config, "compile_mode", None) or model_config.get("compile_mode") or "default"
```
lishunyang12: Add compile_mode to the structured config; the documented stage overlay currently fails validation.

PR #6452, opened 2026-08-21
docs: document OmniVoice online voice cloning

Emre-Akgul:
## Summary

- mark OmniVoice voice cloning as available in the online serving support tables
- remove stale claims that voice cloning and voice design are offline-only
- document the online `--ref-audio`/`--ref-text` client invocation and supported client flags
- align the OmniVoice recipe with the implemented online serving behavior


Emre-Akgul:
### Self-review
Verified the documentation changes against the OmniVoice online serving implementation and client arguments. I also tested an online /v1/audio/speech request with ref_audio and ref_text using vLLM/vLLM-Omni 0.26.0 on an RTX 3090; the request returned HTTP 200 and produced a valid 24 kHz WAV.

Emre-Akgul:
@hsliuustc0106 @Gaohan123 @david6666666 @ywang96 

Please take a look when you have a chance

PR #6461, opened 2026-08-21
[Diffusion] Add a Host Weight Runtime FP8 producer for DLO AllGather

baonudesifeizhai:

## Purpose
## Test Plan
 
<h2>MiniMax-H3 E2E</h2><p>Environment:</p><ul><li>vLLM 0.27.0</li><li>4x NVIDIA B300</li><li>online per-tensor FP8</li><li>DLO AllGather, DP2 / Ulysses-SP2</li><li>FL2VA T2VA, 448x256, 4 seconds, 3 denoising steps</li><li>fixed prompt and seeds</li></ul><div><div><div>
Run | Engine initialization | Valid generation wave | Init improvement
-- | -- | -- | --
Baseline, runtime disabled | 132.73s | 36.55s | —
Cold production | 122.64s | 28.49s | 7.6%
Warm hit | 97.28s | 26.70s | 26.7% / 1.36x

MiniMax-H3 FL2VA, online per-tensor FP8, DLO AllGather, DP4/SP1, GPUs 4–7:
Ref2VA functional validation
The warm server was also exercised using the Ref2VA reference video and audio inputs:
- Resolution: 448×256
- Inference steps: 3
- Requested duration: 4 seconds
- Seed: 42
- Response: HTTP 200
- Request E2E latency: 72.55 seconds
- Output: H.264 video with 32 kHz stereo AAC audio
- Encoded duration: 4.458 seconds
- FFmpeg decode: passed
- All workers exited cleanly and all four GPUs were released
 
Controlled baseline/cold comparison:
Mode | Initialization time | Improvement vs. baseline
-- | -- | --
Baseline, HWR disabled | 134.70 s | —
Cold, fresh HWR cache | 116.60 s | 18.10 s / 13.4% / 1.16x
 
 A second fresh-cache run was used to measure cold and warm with the same artifact and process-to-health timing:
Mode | Startup-to-health | Improvement vs. cold
-- | -- | --
Cold, artifact build | 103.608 s | —
Warm, artifact reuse | 77.460 s | 26.148 s / 25.2% / 1.34x
 
**vLLM Version:**

**vLLM-Omni Commit:**

## Test Result


hsliuustc0106 (commented):
Please address the storage identity, runtime-policy, and validation issues below.

vllm_omni/host_weight_runtime/filesystem/writer.py
```diff
@@ -338,51 +363,55 @@ def finalize(
...
+                        shape=spec.shape,
+                        dtype=str(spec.dtype),
+                        stride=_contiguous_stride(spec.shape),
+                        nbytes=spec.nbytes,
                     )
-                    tensor_entries.append(entry)
+                )
 
         ordered_tensor_entries = tuple(sorted(tensor_entries, key=lambda entry: entry.name))
         content_digest = hashlib.sha256()
```
hsliuustc0106, line 380: `artifact_content_sha256` now includes physical file entries, so shard names and boundaries become part of semantic content. I merged this branch with #6445 and ran the combined Host Weight Runtime consumer tests: `test_independent_publications_are_content_deterministic` fails because the same tensors published with 16-byte and 1024-byte shard limits produce different content digests. The store compares this digest when deciding whether a publication with the same identity is a collision, so two valid physical layouts can be rejected as different semantic content. Please keep the aggregate digest based on canonical tensor content and independent of file sharding; retain file hashes separately for physical integrity. The manifest-v2/writer optimization should be isolated in a foundation PR rather than changing this contract inside the FP8 consumer.

vllm_omni/diffusion/model_loader/diffusers_loader.py
```diff
@@ -445,30 +446,44 @@ def load_model(
...
-                        model,
-                        dit_modules=tuple(zip(modules.dit_names, modules.dits)),
-                        sources=weight_sources,
-                        model_path=str(getattr(self.od_config, "model", "")) or None,
-                        tensor_parallel_size=_tp_size,
-                        use_hsdp=_use_hsdp,
-                        online_quantization=_has_online_quant,
-                    )
-                    self.host_weight_plan = plan_result.plan
+                    if self.od_config.host_weight_runtime_mode != "disabled":
```
hsliuustc0106: This invokes the FP8 runtime for every non-disabled mode under DLO before checking AllGather, effective group size, TP1, HSDP/LoRA, model support, or the online quantization method. Conversely, `mode="required"` without DLO never reaches this branch, so startup silently succeeds through the canonical loader without obtaining the required lease. Unsupported online methods can also enter `resolve_runtime_fp8_plan()` instead of taking the documented preferred fallback, and preflight/`host_plan()` exceptions currently escape. Please make one explicit eligibility decision before resolution: the initial path should require the supported MiniMax-H3 per-tensor FP8 layout, TP1, no HSDP/LoRA, and multi-rank DLO AllGather. In `preferred`, unsupported/preflight failures must fall back before mutation with a typed reason; in `required`, they must fail startup. Add coverage for required-without-DLO and preferred-with-unsupported-layout.

vllm_omni/diffusion/model_loader/runtime_fp8.py
```diff
@@ -0,0 +1,523 @@
...
+        Fp8PerTensorOnlineLinearMethod,
+    )
+
+
+def _scale_name(weight_name: str) -> str:
+    parent_path, _, _ = weight_name.rpartition(".")
+    return f"{parent_path}.weight_scale"
+
+
+def _source_identity(
```
hsliuustc0106: This creates a second, incomplete source/identity contract beside #6445. It fingerprints physical path/stat information but does not bind the final tensor names/shapes/dtypes/roles, model/loader ABI, or checkpoint-adapter transform ABI. More importantly, `produce()` never verifies that these files remain unchanged after identity construction, so a checkpoint replaced during the transaction can be streamed and published under the old key. Physical paths, device IDs, and inodes also make equivalent sources in different local domains semantically different. Please rebase this producer onto #6445's representation-neutral `PreparedWeightSource`, `FinalLayoutRequest`, `FinalLayoutArtifactSpec`, tensor-policy, source-guard, and restorer contracts (after its coverage fix), and validate the source before and after production rather than maintaining a separate FP8 identity/restoration path.

vllm_omni/diffusion/model_loader/runtime_fp8.py
```diff
@@ -0,0 +1,523 @@
...
+
+    def _quantize_weight(
+        self,
+        binding: TensorBinding,
+        outputs: Mapping[str, Any],
+        runtime_name: str,
+        target: torch.Tensor,
+    ) -> None:
+        scale: torch.Tensor | None = None
+        with safe_open(binding.file_path, framework="pt", device="cpu") as handle:
```
hsliuustc0106: Closing the `safe_open` mapping after this two-pass quantization does not evict the consumed BF16 pages from the kernel page cache. Issue #6330 explicitly requires per-tensor `madvise(MADV_DONTNEED)` or `posix_fadvise(POSIX_FADV_DONTNEED)` so checkpoint residency tracks the working set rather than accumulating toward the complete checkpoint. As written, GPU/CPU scratch allocation is bounded, but the cold builder can still retain the entire BF16 checkpoint in page cache. Please add range-aware page release after the second pass and report cold/warm RSS, PSS, and HBM evidence demonstrating that the intended host-memory bound is achieved.

hsliuustc0106:
Before marking this ready, please replace the remaining PR-template placeholder and complete the validation record. The current tables appear to be single measurements and do not include exact commands, the vLLM-Omni commit, warmup policy, mean ± standard deviation over repeated runs, peak RSS/PSS/HBM, or a baseline-versus-cold-versus-warm output-quality comparison using identical inputs and seeds. HTTP 200 plus successful FFmpeg decoding proves request/container validity, but not FP8 weight/scale or model-output parity.

For reference, the PR head passed the targeted CPU suite (`88 passed`). I also tested the branch together with #6445: `99 passed, 1 failed`, where #6445's deterministic-publication test exposes the shard-dependent content digest described inline. Please include the combined consumer suite after rebasing onto the shared final-layout contracts.

baonudesifeizhai:
#6445  foundation commits：+2700 / -5   base on that

PR #5052, opened 2026-07-12
examples: replica data-parallel recipe + benchmark for video DiT (#4707)

linzhenpl07:
### Summary

A runnable **replica data-parallel** recipe + benchmark for the video DiT path, plus a small message fix. This is the concrete first slice for #4707 (request-throughput scaling for video generation), built on the existing `runtime.num_replicas` fan-out.

**New** `examples/online_serving/replica_data_parallel/`:
- `wan2_2_ti2v_dp.yaml` — Wan2.2-TI2V-5B diffusion stage config; `runtime.num_replicas` + `runtime.devices` fan out N independent engine replicas (one GPU per replica for `tensor_parallel_size=1`).
- `run_server.sh` — parametrizes the replica count (`NUM_REPLICAS` / `DEVICES`) and serves.
- `bench_replica_dp.py` — replica-agnostic load driver: throughput (videos/min) vs replica count, per-request latency, and an md5 isolation check.
- `README.md` — how to run + measured numbers.

**Fix** `vllm_omni/entrypoints/cli/serve.py`: the `--omni` prohibited-args error only mentioned `--omni-dp-size-local` for replica count, which doesn't apply to the single-runtime path. Now points at the per-stage `num_replicas` config field (single-runtime) as well.

### Measured scaling

Wan2.2-TI2V-5B (832×480, 33f, 30 steps), 4× A800-80GB NVLink, one GPU per replica:

| Replicas | videos/min | Scaling | Efficiency |
|----------|-----------|---------|------------|
| 1 | 4.71 | 1.00× | — |
| 2 | 9.20 | 1.95× | 98% |
| 4 | 18.02 | 3.83× | 96% |

Per-request latency flat (~13–14 s); seed-fixed outputs byte-identical across replica counts (replicas isolated, no cross-talk).

### Interface question (for discussion)

`num_replicas` is accepted by both the legacy `stage_args` path and the new `StageDeployConfig`, but no bundled `vllm_omni/deploy/*.yaml` exposes it and the new deploy-schema docs don't list it. This recipe therefore serves via `--stage-configs-path`. **How should replica-DP be expressed under the newer `--deploy-config` schema?** Happy to move the example over once that's settled — flagging so it aligns with the ongoing serving-config refactor (#4901 / #4843 / #5031). cc @fake0fan @hsliuustc0106

Related to #4707.

PR #6463, opened 2026-08-21
[AR-Diffusion] Allocate K and V separately so the compiled path stops cloning the pool

linzhenpl07:
## Summary

The AR-Diffusion paged-write custom op declares both `key_pool` and `value_pool` as mutated:

```python
@torch.library.custom_op(
    "vllm_omni::ar_diffusion_paged_write_attn",
    mutates_args=("key_pool", "value_pool"),
)
```

They were the two halves of one allocation:

```python
kv = torch.empty(2, num_blocks, block_size, num_kv_heads, head_dim, ...)
k_pools.append(kv[0].reshape(...))
v_pools.append(kv[1].reshape(...))
```

so they alias a single storage. Inductor's reinplace pass can then re-inplace only the first of the two mutated arguments:

```
For node auto_functionalized_v2, attempted to reinplace range(0, 2).
We were unable to reinplace [1]
```

The second hits the "mutated arg aliases a graph input" case in `_inductor/fx_passes/reinplace.py`, so `auto_functionalized_v2`'s clone of the **entire V pool** survives into the compiled graph:

```python
buf13 = empty_strided_cuda((545587200, ), (1, ), torch.bfloat16)   # 1.02 GiB
# Topologically Sorted Source Nodes: [ar_diffusion_paged_write_attn]
triton_poi_fused_1.run(arg10_1, buf13, 545587200, stream=stream0)
```

and in the FX graph:

```
%as_strided_default = aten.as_strided(%arg10_1, [545587200], [1], 0)
%clone_default      = aten.clone(%as_strided_default)
```

That is one full pool copied **per compiled region per denoising step**. At 480x768 with 40 layers and 4 DMD steps that is roughly **163 GiB of copies per committed chunk**.

## Fix

Allocate K and V as separate tensors. `kv_pools[layer][0]` and `[1]` still address K and V, so `key_cache()` / `value_cache()` and every other caller are unchanged.

## Measured

One RTX PRO 6000 Blackwell (sm_120), LingBot-World 2.0 at 480x768 bf16, single session, steady state over chunks 6-9 — the sliding window (`sink=9` + `window=9` frames, 3 frames per block) fills at chunk 6, and latency is flat within 0.4% after that:

| | chunk latency | RTF |
|---|---|---|
| eager | 4.049 s | 5.40 |
| compiled, before this change | **crashed** | — |
| compiled, after | 3.668 s | 4.89 |

**1.10x**, and the compiled path runs at all.

Before the change, the clone kernel faulted during inductor's Triton config selection:

```
File ".../torch/_inductor/runtime/triton_heuristics.py", line 1176, in autotune_to_one_config
torch.AcceleratorError: CUDA error: an illegal memory access was encountered
```

confirmed under `CUDA_LAUNCH_BLOCKING=1`, so it is the clone kernel itself rather than an async report from an earlier one. On other architectures the copy presumably completes and is simply slow, which is likely why this went unnoticed.

The plateau still occurs at chunk 6 after the change, so compilation introduces nothing that grows with session length.

### Confirmed on Hopper, where it does not crash

The paragraph above guesses that on other architectures the copy completes and
is simply slow. An Nsight Systems trace of an unrelated H200 run -- 2x H200,
TP=2, one session, saturating -- shows exactly that, and puts a number on it.

Steady state, 16 s window, 99,248 kernels per rank. The GPU is **90.3% busy**,
so nothing here is launch overhead; the cost is work the GPU actually performs:

| category | share of GPU time |
|---|---:|
| GEMM (cuBLAS `nvjet*`) | 35.7% |
| attention | 22.4% |
| **`triton_poi_fused_7` + `triton_poi_fused_copy__8`** | **19.4%** |
| NCCL all-reduce | 12.1% |
| other elementwise / layernorm | 10.3% |

The two pointwise kernels carry the clone's signature:

```
triton_poi_fused_7        gridX 1,404,000 x block 256 = 359,424,000 elements   1167 us x 1870
triton_poi_fused_copy__8  gridX   351,000 x block 128 =  44,928,000 elements    335 us x 1871
a normal per-layer pointwise                            2,995,200 elements     35-51 us
```

120x the elements of an ordinary per-layer op, ~1.44 GiB read+written per call
in bf16, **2.7 TB moved for this one kernel in 16 seconds**. Removing that work
takes busy time from 14.44 s to 11.63 s, so at an unchanged idle fraction the
step rate improves by **1.21x**.

Two caveats on that number. The attribution is by signature -- magnitude, cost
and the pointwise/copy pairing -- not from that run's inductor output, so it is
consistent with this clone rather than proof of it. And the run is TP=2 at an
unrecorded resolution, so the 1.21x is an estimate for that shape, not a general
claim. It does establish that the cost is not specific to sm_120, and that on a
card where the copy succeeds it is roughly a fifth of all GPU time.

### Ruled out along the way

Recorded because each of these looks plausible and none of them is the cause — all four act on a different layer than the one that decides the clone:

| hypothesis | result |
|---|---|
| allocator holding cached blocks from the fp32 -> bf16 load | `torch.cuda.empty_cache()` released 0.00 GiB |
| `dynamic=True` producing multiple Triton configs | `dynamic=False` still crashes |
| `torch._dynamo.mark_static_address()` on the pools | clone still generated |
| memory pressure during autotune | 12 GiB more headroom, still crashes |
| int32 byte-offset overflow past 2 GiB | pool reduced to 1.53 GiB/layer, still crashes |

The clone is a post-grad FX pass decision; none of the above can influence it. `TORCH_LOGS="+torch._inductor.fx_passes.reinplace"` names the failing argument directly and is what located it.

## Test

`tests/diffusion/ar_diffusion/` — 145 pass, ruff check and format clean.

- `test_paged_pool_layout_exposes_flat_slot_views` updated for the new layout, keeping its real invariant: the flat slot views alias their block-shaped cache, so a slot write is visible through the layout the attention kernel reads.
- `test_key_and_value_caches_do_not_share_storage` added, pinning the **mechanism** this fix introduces — K and V have distinct storages per layer, the flat views point at the right one, and a negative control that filling V leaves K untouched.
- `test_compiled_paged_write_does_not_clone_the_pool` added, pinning the **effect** that mechanism buys. Separate storage is what lets inductor re-inplace both pools, but only the generated code says whether it did, so this compiles a stand-in op with the same `mutates_args` contract over pools from the real allocator and asserts no pool-sized allocation appears. On the shared allocation it fails and names the buffer:

  ```
  AssertionError: the compiled graph allocates a pool-sized buffer, so the pool
  is being cloned rather than written in place: [16384]
  ```

  It disables inductor's caches explicitly. Compiled code survives `torch._dynamo.reset()`, so a second compile in the same process returns the first one's code; without that guard the assertion silently describes a graph other than the one under test.
- `test_real_paged_write_op_does_not_clone_the_pool_on_cuda` added, `hardware_test(res={"cuda": "L4"})`. The CPU test above compiles a stand-in, so the real signature and Triton codegen stayed unverified — and Triton is the path this regression was observed on. This compiles the real op over pools from the real allocator, with the same positive control. On sm_120:

  ```
  shared     pool-sized allocations [65536]   12.0 KB of generated code
  separate   pool-sized allocations []         4.8 KB
  ```

  65536 is the whole `(2, ...)` pool. It passes the flat views rather than the block-shaped caches, since the op indexes by slot and unflattens internally, so those are the tensors `mutates_args` actually names.

## Note

The `(2, num_blocks, ...)` shape was described as following FlashAttention's block-table layout. The kernel takes `k` and `v` as separate arguments (`flash_attn_varlen_func(q=..., k=key_cache, v=value_cache, block_table=...)`), so the combined allocation was a convenience rather than a requirement; each cache keeps its own `(num_blocks, block_size, num_kv_heads, head_dim)` shape. If there is a reason to keep one allocation that I have missed, the alternative would be to declare only one of the two as mutated and write V through a different path — but that seems worse.


linzhenpl07:
This turns out to be more than a memory optimisation: without it, NVFP4 plus
`torch.compile` does not run at all.

On an NVFP4 (ModelOpt) LingBot World v2 checkpoint, 1x RTX PRO 6000 Blackwell
(sm_120), the compiled path dies with `CUDA error: an illegal memory access
was encountered`. With `CUDA_LAUNCH_BLOCKING=1` the crash lands inside the
generated code, and the generated code is exactly what this PR describes:

```python
buf15 = empty_strided_cuda((501350400, ), (1, ), torch.bfloat16)   # ~1 GB
triton_poi_fused_4.run(arg14_1, buf15, 501350400, stream=stream0)  # clone the pool
buf16 = torch.ops.vllm_omni.ar_diffusion_paged_write_attn.default(
    ..., reinterpret_tensor(buf15, (48960, 40, 128), (5120, 128, 1), 250675200), ...)
```

501350400 is 2 x 250675200 -- the whole K+V allocation -- and the offset hands
the attention a view into the *copy*, once per compiled region per denoising
step.

Applying this branch's `allocate_kv_pool_with_views` makes it run:

| config | 768x480 steady chunk | result |
| --- | --- | --- |
| NVFP4 + compile, before | -- | CUDA illegal memory access |
| NVFP4 + compile, after | **2.845 s** | runs, 0 errors over 8 chunks |
| NVFP4 + eager (the workaround) | 3.348 s | runs |
| bf16 + compile, before | 3.668 s | runs |
| bf16 + compile, after | 3.626 s | runs |

So bf16 barely notices the clone, which is presumably why this looked like a
pure memory win. Quantised weights do notice: the fix is what makes compile
usable there at all, and compile is then worth 1.18x over eager.

Measured across resolutions after the fix (compile, gpu-memory-fraction 0.60):

| resolution | bf16 | NVFP4 | ratio |
| --- | --- | --- | --- |
| 832x480 | 4.051 s | 3.253 s | 1.245x |
| 768x480 | 3.626 s | 2.845 s | 1.275x |
| 320x256 | 0.627 s | 0.442 s | 1.418x |

I am not asking for anything beyond what the PR already proposes -- just
recording that the failure mode is a crash, not only a wasted gigabyte.

linzhenpl07:
Pushed `71c7469e`. Before asking for review I went back over the inline feedback
on my earlier PRs here and checked this one against the patterns that came up
more than once. Two of them applied and are now fixed; recording the rest so a
reviewer does not have to re-derive that they were considered.

### The test could have passed for the wrong reason

Raised twice in review on #5052 — identical hashes cannot prove replica
isolation when every request uses the same prompt and seed, and `got` being a
strict subset of `baseline` still reported PASS.

The clone test asserted `not pool_sized_allocations(generated_code)`. A detector
that matches nothing satisfies that. If inductor stops emitting
`empty_strided_*` in the form the regex expects, the assertion holds while
testing nothing, and it fails open rather than closed.

The shared layout is now compiled first as a positive control, in the same
process and torch version, and the test fails if that control does **not**
produce a pool-sized allocation:

```
AssertionError: the compiled graph allocates a pool-sized buffer, so the pool is
being cloned rather than written in place: [16384] (control saw [16384])
```

### The stand-in op could drift from the real one

Raised in review on #6345 — none of the tests executed `decode_latent`,
so deleting the guard left the suite green — and in review on #5255,
where FlashAttention on an L4 meant the Ring regression test never reached
`pytorch_attn_forward()`.

What drives the reinplace pass is the `mutates_args` declaration, and the test
models it with a stand-in because the real op needs CUDA and a flash-attention
kernel. If the real op ever declares a different set, the test would keep
passing against a contract nobody uses. It now reads the real schema:

```python
real_mutated = [arg.name for arg in real_schema.arguments if arg.alias_info and arg.alias_info.is_write]
assert real_mutated == ["key_pool", "value_pool"], (
    f"the real paged-write op now mutates {real_mutated}; this test still models two pools"
)
```

That narrows the gap rather than closing it: the stand-in still is not the real
op. A CUDA test that compiles the real path would close it properly, and I am
happy to add one if you would rather have that than the CPU test.

### Checked and did not apply

- **No CI job collects the file** (review on #5255, on
  `tests/diffusion/attention/`). `tests/diffusion/ar_diffusion/` is collected by
  `pytest -sv tests/diffusion -m 'core_model and cpu'` in `test-ready.yml`, and
  the new test is CPU-only and adds about 1 s.
- **Cold measurement** (review on #5052). The numbers in the description
  are chunks 6-9, after the sliding window fills, with the plateau argued in the
  text rather than assumed.
- **Same bug still live nearby** (review on #6345, on the encode side).
  `torch.empty(2, ...)` for a KV pool appears nowhere else in `vllm_omni/`.
- **Test shape is not the production shape** (review on #6345). This one
  asserts on graph structure, not on numerics, so the toy shape is deliberate --
  a production-sized pool would only make the test slower.

### Drive-by

The file had two `pytestmark` assignments; the second silently overwrote the
first and dropped `pytest.mark.diffusion`. Collection was unaffected because the
CPU job selects on `core_model and cpu` plus a path, but the loss was not
intentional, so they are merged into one.

tests/diffusion/ar_diffusion/test_kv_cache.py
```diff
@@ -25,12 +26,9 @@
 )
 from vllm_omni.experimental.ar_diffusion.kv_cache.paged import chunk_window_skipped_tokens
 
-pytestmark = [pytest.mark.core_model, pytest.mark.diffusion, pytest.mark.cpu]
-
-
 BLOCK = 16
 
-pytestmark = [pytest.mark.core_model, pytest.mark.cpu]
+pytestmark = [pytest.mark.core_model, pytest.mark.diffusion, pytest.mark.cpu]
```
SamitHuang, line 30: `pytestmark` applies `cpu` to every test in this file, including `test_real_paged_write_op_does_not_clone_the_pool_on_cuda`. Ready/merge CI collects this path only via `pytest tests/diffusion -m 'core_model and cpu'`; there is no job that runs `tests/diffusion/ar_diffusion` with `-m cuda`. `hardware_test(num_cards=1)` also adds no `skipif` when CUDA is missing, so a CPU-only run of this module hits `torch.device("cuda")` and fails. If you later remove `cpu` from the CUDA test, it disappears from CI entirely. Split the CUDA compile into its own file (or a job that selects `core_model and cuda`) and keep this module CPU-only, matching the file docstring.

tests/diffusion/ar_diffusion/test_kv_cache.py
```diff
@@ -155,14 +153,253 @@ def test_paged_pool_layout_exposes_flat_slot_views():
...
+            torch.empty(0, dtype=torch.long, device=device),
+            torch.arange(num_blocks, device=device, dtype=torch.int32).unsqueeze(0),
+            torch.tensor([0, num_tokens], device=device, dtype=torch.int32),
+            torch.tensor([num_tokens], device=device, dtype=torch.int32),
+        )
+        torch._dynamo.reset()
+        # See the CPU test: inductor's cache outlives torch._dynamo.reset().
+        with inductor_config.patch(force_disable_caches=True):
+            _, codes = run_and_get_code(
+                torch.compile(_Holder(key_flat, value_flat), backend="inductor", fullgraph=True), *args
```
SamitHuang: This compiles the real `ar_diffusion_paged_write_attn` with `fullgraph=True`, so inductor has to trace FlashAttention paged `flash_attn_varlen_func`, not just the `mutates_args` contract. That is a different failure mode than the clone (graph break, missing FA kernel, autotune IMA on a large control pool). The CPU stand-in already pins reinplace; if the goal is Triton codegen, compile a custom op with the real schema on CUDA without calling FA, or skip when the backend cannot lower the attention kernel.

tests/diffusion/ar_diffusion/test_kv_cache.py
```diff
@@ -155,14 +153,253 @@ def test_paged_pool_layout_exposes_flat_slot_views():
...
+        num_blocks=num_blocks,
+        block_size=BLOCK,
+        num_layers=1,
+        num_kv_heads=heads,
+        head_dim=dim,
+        dtype=torch.bfloat16,
+        device=device,
+    )
+    separate = compiled_pool_allocations(k_pools[0], v_pools[0])
+    assert not separate, (
```
SamitHuang: `numel >= pool_numel` treats any generated `empty_strided_cuda` at least as large as one toy pool (32768 elements, 64 KiB bf16) as a clone. Compiling the real attention op can emit workspace/scratch buffers in that range even after K/V are separate, which fails this assertion while the clone is gone. Tighten the detector to the control's exact size (or `2 * pool_numel`, the shared `(2, ...)` allocation you already observed as `[65536]`) and assert the separate layout does not contain that size, rather than `not any(numel >= pool_numel)`.

SamitHuang (changes requested):
The allocator change is the right fix for inductor reinplace cloning a mutated alias, and the CPU stand-in test with a shared-layout positive control actually pins that mechanism. The CUDA test does not: module-level `cpu` is the only reason CI collects it, `hardware_test` does not skip without a GPU, and the detector can fail on any FlashAttention workspace ≥ one pool. Split that test out of the CPU module (and onto a job that selects `cuda`) or drop it until it is collected on purpose.

wtz2333:
This bug causes generation failure when torch.compile is enabled. The figure below shows the resulting behavior.
Codex 图像 2026年9月1日 16_31_46

tests/diffusion/ar_diffusion/test_kv_cache.py
linzhenpl07: Split `test_real_paged_write_op_does_not_clone_the_pool_on_cuda` into its own module (`test_kv_cache_cuda.py`) without the `cpu` mark, so it's no longer swept up by `pytest tests/diffusion -m 'core_model and cpu'` (CUDA and AMD both run that job on this path). Nothing currently selects `tests/diffusion/ar_diffusion` under `cuda` -- `tests/diffusion/*.py` in test-ready.yml is non-recursive, so it wouldn't have picked this up from the old location either -- so for now it's intentionally uncollected rather than wired into a job on the side, per your second option. Verified locally: the CPU module now collects 30 (was 31) and all pass under `core_model and cpu`; the new file collects 0 under that filter and 1 under `core_model and cuda`.

tests/diffusion/ar_diffusion/test_kv_cache.py
linzhenpl07: Agreed -- the real op's own FA call is a different failure mode than the clone this test exists to catch. Registered a second stand-in in the new file that copies the real op's full argument list and `mutates_args` declaration (so reinplace sees the identical contract) but stops at the pool writes the same way `_paged_write_attn_impl` does before it calls `ar_diffusion_paged_attention`, so no FA kernel gets compiled here at all. Ran it on an actual CUDA device (L4-class): passes.

tests/diffusion/ar_diffusion/test_kv_cache.py
linzhenpl07: Tightened the detector to match the control's exact size instead of `>= pool_numel`: `clone_numel = 2 * pool_numel` (the shared `(2, ...)` allocation, same [65536] you flagged). The control must contain that exact size and the separate layout must not, rather than "no allocation at least this big". Verified on CUDA that the positive control still hits [65536.0] with the change, so the detector still distinguishes the two layouts and this isn't a check that now vacuously passes.

SamitHuang (approved):
LGTM

Issue #6453, opened 2026-08-21
[RFC]: Request-scoped incremental stage-state lifecycle


EchoHayate:
### Motivation.

Incremental stage-output processors across TTS/omni models keep mutable
per-request state while converting repeated upstream outputs into downstream
chunks. Examples include accumulated token/tensor rows, prompt conditioning,
chunk cursors, terminal flags, and codec context.

Current main already has a de facto sender-side ownership boundary:

```python
ChunkTransferAdapter.request_payload
```

and releases it after a terminal payload is sent successfully. However,
processors access it as a raw shared dictionary and independently choose:

- whether to key by scheduler/internal or connector/external request ID;
- how to namespace state;
- whether the processor or runtime removes it;
- what happens on cancellation, timeout, processor exception, or failed
  `connector.put()`;
- whether a resumable segment boundary ends the state lifecycle;
- how request-ID reuse and late queued work are handled.

This is observed duplication, not a one-model abstraction. At current main
`4f7c08c0`, at least ten processor modules directly reference
`request_payload`: Audex, CosyVoice3, GLM-TTS, Ming-TTS, MiniCPM-o 4.5,
MOSS-TTS, Nemotron VoiceChat, PersonaPlex, Qwen3-Omni, and Qwen3-TTS.

PR #5556 exposed one concrete failure mode. LLaMA-Omni2 Thinker/Talker state
was originally stored in manager-private dictionaries that canonical sender
cleanup could not see. A focused repair moved it under
`request_payload[external_request_id]`; the lifecycle selection changed from
`6 failed, 33 passed` to `39 passed`, and the complete focused suites reached
`104 passed`. That repair intentionally did not add a generic API.

#### Relationship to existing RFCs

- #4872 owns processor role/signature contracts, registry/startup validation,
  async/non-async parity, and shared processor helpers.
- #4855 owns async-chunk reliability outcomes such as send-failure
  propagation and chunk-wait deadlines.
- #5805 discusses broader full-duplex user/session state.

This RFC is a narrower lifecycle sub-contract. It defines how mutable
stage-request state is owned and released around the processor and connector
events above. It does not create another processor framework, timeout system,
or session manager.

### Proposed Change.

Define a small request-scoped, namespaced state accessor on the existing
`transfer_manager` producer boundary. Keep the initial storage backed by
`request_payload`; do not add a new global manager.

#### 1. Explicit identity

Pass both identities:

```python
@dataclass(frozen=True)
[...]

PR #4728, opened 2026-06-25
[Frontend] Add omni benchmark support for image and video endpoints

ZacheryAU:
## Purpose
As a further development of https://github.com/vllm-project/vllm-omni/pull/3628:
- implement https://github.com/vllm-project/vllm-omni/issues/4253

## Summary
This PR extends `vllm bench serve --omni` to benchmark OpenAI-compatible image and video generation endpoints more naturally.
- Adds endpoint-driven backend inference for `/v1/images/generations` and `/v1/videos`
- `/v1/images/generations`, `/v1/images/edits`, and `/v1/videos` can be benchmarked without passing endpoint-specific backend names.
- Adds benchmark request handling for `/v1/images/generations` and async `/v1/videos`, including polling video jobs until completion.
- Adds videos metrics like video duration, video frames, video throughput, video RTF, video generation latency, and also peak memory percentiles, with new metrics added in `definitions.py`.
- `--print-stage` only prints stage benchmark results when real stage data is available.
- Include a bugfix for Helios-Distilled for the experiment

## Test Plan

**vLLM Version:** 0.28.0

**vLLM-Omni Commit:** current commit against 5d20f6b900dfcf22b42068e28a3cbbdb5b4707e7
- Tests image-output model with `vllm bench serve --omni --endpoint /v1/images/generations` without `--backend`
- Tests video-output model with `vllm bench serve --omni --endpoint /v1/videos` without `--backend`
- Run `python -m pytest -m 'core_model and cpu' tests/benchmarks/ tests/metrics/ tests/entrypoints/openai_api/test_video_server.py -q`

### Qwen/Qwen-Image (`/v1/images/generations`)
<details class="admonition abstract" markdown="1">
<summary>Show more</summary>
<pre>
============ Serving Benchmark Result ============
Successful requests:                     3         
Failed requests:                         0         
Maximum request concurrency:             1         
Benchmark duration (s):                  8.42      
Request throughput (req/s):              0.36      
Peak concurrent requests:                2.00      
-------------------Peak Memory--------------------
Mean PEAK_MEMORY_MB (MB):                58832.00  
Median PEAK_MEMORY_MB (MB):              58832.00  
P99 PEAK_MEMORY_MB (MB):                 58832.00  
----------------End-to-end Latency----------------
Mean E2EL (ms):                          2805.73   
Median E2EL (ms):                        2804.90   
P99 E2EL (ms):                           2810.49   
================== Image Result ==================
Total images generated:                  3         
[...]

PR #6464, opened 2026-08-21
[Bench] Save per-request stage metrics in detailed results

EchoHayate:
## Summary

Preserve the per-request stage snapshots that the canonical benchmark client
already receives when `--save-detailed` is enabled.

Today, `--print-stage` can aggregate `MixRequestFuncOutput.stage_metrics`, but
the benchmark result builder never copies those snapshots into the detailed
JSON result. This makes aggregate stage output visible while dropping the raw
request-aligned evidence needed for attribution.

## Changes

- request `return_stage_metrics=true` for supported Omni backends during
  detailed runs;
- save one `stage_metrics` mapping per request, aligned with `input_lens`,
  `ttfts`, `itls`, `generated_texts`, and `errors`;
- preserve `{}` for requests without a snapshot instead of shifting indexes;
- keep non-detailed results unchanged;
- avoid injecting the Omni-only response field into unsupported backends.

## Scope

This only fixes canonical benchmark artifact serialization. It does not add
runtime events, change scheduler/orchestrator behavior, expose
`stage_durations`, or introduce model-specific telemetry.

## Validation

- RED: 3 focused tests failed for the missing helper, detailed request opt-in,
  and detailed-only result gate;
- GREEN: 3 focused tests passed;
- `tests/benchmarks/patch/test_patch.py`: `23 passed, 15 warnings`;
- Ruff: passed;
- `git diff --check`: passed;
- local P0-P2 review: no findings.

Related context: #6453.


EchoHayate:
Author self-review:

- Scope: this PR only preserves already-returned per-request `stage_metrics` in `--save-detailed` benchmark JSON and opts supported Omni backends into requesting them. It does not change runtime, scheduler, model adapters, or telemetry generation.
- Alignment/failure semantics: output order is preserved; a request without a snapshot is serialized as `{}` so detailed arrays keep the same index identity.
- Local verification: 3 RED tests before the implementation, then `3 passed`; full benchmark patch module `23 passed, 15 warnings`; Ruff and `git diff --check` passed. A local P0-P2 review found no issues.
- GPU artifact validation: on the fixed 2xA100 LLaMA-Omni2 c1 workload, 8/8 requests succeeded. The saved JSON contains 8 `stage_metrics` entries aligned with 8 `input_lens`, `output_lens`, `generated_texts`, and `errors`, with no empty snapshots. Six requests contain stages 0/1/2 and two contain stages 0/1; the server `OmniTiming` logs show the same 6/2 stage-pattern split, so the artifact preserves the runtime snapshots rather than inventing or dropping Stage 2.
- Current checks: DCO, pre-commit, Python 3.11/3.12 wheels, and Read the Docs are green.

Validation boundary: this fixes artifact serialization only. It does not establish a performance improvement or claim that existing stage metrics fully separate producer, connector, admission, and consumer waiting.

@alex-jw-brooks @Bounty-hunter, could you please review when convenient?

EchoHayate:
Current-head update after refreshing onto upstream `main`:

- Head: `49b70d8faf95ac1e78fa11892c45689536c0d909`.
- The refresh had one mechanical import conflict in `tests/benchmarks/patch/test_patch.py`; both upstream benchmark helpers and this PR's `benchmark_patch` import were preserved. The production scope remains the same three files / detailed-result serialization only.
- Fresh local checks on this head: Python 3.12 `py_compile`, targeted Ruff, and `git diff --check` all passed.
- GitHub checks are now all green: DCO, pre-commit, Python 3.11/3.12 wheels, and Read the Docs. The PR is currently mergeable and waiting for review.

Evidence boundary is unchanged: this preserves already-produced per-request stage snapshots in benchmark artifacts; it does not claim a runtime performance improvement or fuller stage attribution.

EchoHayate:
## Author self-review / current-main refresh

- Restacked onto `main@b1ec136b` after #4728 added image/video benchmark endpoints. The only content conflict was the stage-metrics backend set; the resolution preserves all new image backends and adds `/v1/videos` to the detailed-output opt-in matrix.
- Current-main RED: with `--save-detailed` and `backend=openai-chat-omni`, unpatched main returned `False` from `should_request_stage_metrics`, so the server was not asked for the snapshots that the detailed artifact should retain.
- Scope remains three benchmark files only. Runtime metric generation, scheduler/orchestrator behavior, and aggregate metric calculations are unchanged.
- Alignment invariant: the detailed result contains exactly one `stage_metrics` mapping per request, using `{}` when a request has no snapshot rather than shifting indexes.
- Backend coverage now explicitly tests `daily-omni`, `openai-chat-omni`, all three image endpoint/backend forms, `/v1/videos`, and an unsupported backend negative case.

Fresh current-head verification:

- focused detailed/stage-metric tests: `10 passed` (`4` original focused selections plus the expanded seven-case backend selection, with overlap)
- full `tests/benchmarks/patch/test_patch.py`: `47 passed`
- Python 3.12 byte compilation and scoped `git diff --check`: passed
- all changed-file pre-commit hooks except mypy: passed
- mypy boundary: the same `30` source + `8` test errors reproduce on a clean `main@b1ec136b`; this refresh adds no new mypy diagnostic. CI's all-files pre-commit path skips this currently non-clean hook, as documented in `.pre-commit-config.yaml`.
- independent read-only review: no remaining Critical or Important findings after adding `/v1/videos` coverage.

Evidence boundary remains benchmark artifact correctness; this PR does not claim a runtime performance improvement or complete producer/connector/admission/consumer waiting attribution.

Current head reviewed: `1ed39c8b9d935e003e5d30101b82db45790f8c64`.

@alex-jw-brooks @Bounty-hunter, this is conflict-free again and ready for review once the refreshed CI completes.

Issue #6425, opened 2026-08-21
[Bug]: Merge CI, Occasional, tests/diffusion/distributed/test_pipeline_parallel.py::test_predict_noise[pp2-seq_cfg] - DistNetworkError EADDRINUSE port 33649


yenuo26:
### Your current environment

<details>
<summary>The output of <code>python collect_env.py</code></summary>

CI

</details>

### Your code version

<details>
<summary>The commit id or version of vllm</summary>

CI

</details>
<details>
<summary>The commit id or version of vllm-omni</summary>

CI

</details>

### 🐛 Describe the bug

https://buildkite.com/vllm/vllm-omni/builds/13886

repo-existing issue

torch.distributed.DistNetworkError: The server socket has failed to listen on any local network address. port: 33649, useIpv6: false, code: -98, name: EADDRINUSE, message: address already in use
[W820 18:57:52.292435752 socket.cpp:764] [c10d] The client socket has failed to connect to [localhost]:33649 (errno: 99 - Cannot assign requested address).
File "/workdir/vllm_omni/diffusion/distributed/parallel_state.py", line 395, in init_distributed_environment
    torch.distributed.init_process_group(
tests/diffusion/distributed/test_pipeline_parallel.py::test_predict_noise[pp2-seq_cfg]

Gaohan123:
solved by PR #6468

PR #6468, opened 2026-08-21
[BugFix] Eliminate distributed test port race by switching to file:// rendezvous

NickCao:
## Purpose

torch.distributed process groups in these tests picked a free TCP port, released it, then handed the bare number to init_process_group (often after a torch.multiprocessing.spawn subprocess-startup delay). Any other process on the CI host could grab that port in the gap, causing DistNetworkError EADDRINUSE (#6425). Rendezvous via a unique file:// path instead removes the port-picking step entirely, closing the race at its root rather than narrowing the window.

Ports picked for non-torch-distributed services (mooncake_master, RDMA transfer engine, OmniDiffusionConfig's own port-stepping test) still use tests.helpers.runtime.get_open_port, since file:// doesn't apply there.

## Test Plan

**vLLM Version:** 0.27.0

**vLLM-Omni Commit:** 2a62098d47e0e2cd2fed1d7d587973621b6dea4f

TBD

## Test Result

TBD

### DIFF
--- a/tests/diffusion/attention/test_ulysses_uaa.py
+++ b/tests/diffusion/attention/test_ulysses_uaa.py
@@ -1,10 +1,9 @@
 # SPDX-License-Identifier: Apache-2.0
-# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM-Omni project
 
 from __future__ import annotations
 
 import os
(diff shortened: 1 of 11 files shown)


NickCao:
@vllm-omni-review-bot

lishunyang12:
@yenuo26

lishunyang12 (approved):

PR #6473, opened 2026-08-22
[Diffusion] Add model-declared LoRA runtime with MiniMax-H3 Turbo

mglyn:
## Motivation and scope

Diffusion models publish LoRA adapters with model-specific checkpoint layouts, packed modules, and application semantics. The existing diffusion LoRA paths do not provide a clean model-owned contract for these differences, and combining their migration with MiniMax-H3 Turbo made the change difficult to review.

This draft therefore starts a parallel, deliberately small **Diffusion LoRA Runtime**:

- dynamic LoRA only; no prefusion, LRU, or request-time download;
- adapters are registered and fully loaded at service startup;
- requests select registered adapters by name and scale only;
- one immutable fixed-shape bank supports weighted multi-LoRA composition;
- the scheduler separates batches by canonical composition;
- existing diffusion LoRA facilities and their models are not migrated in this PR.

MiniMax-H3 FL2VA Turbo v1.0 is the first model integration. Sampling steps, flow shifts, and the H3 execution plan remain request/pipeline concerns; the LoRA runtime only installs and activates weight updates.

## Design

Each supported model explicitly provides three pieces:

1. a loader that consumes its publication format and emits normalized low-rank updates;
2. a binding plan that allowlists components and maps logical targets to physical or packed modules;
3. an executor factory, with a default TP-aware `Wx + B(Ax)` linear executor and an extension point for genuinely different model math.

The H3 loader validates the LightX2V v1.0 rank/alpha contract, maps Diffusers keys to native H3 modules, restores the native FFN row order, binds Q/K/V to packed QKV, and rejects Ref2VA-only deployment.

Example deployment:

```bash
vllm serve MiniMaxAI/MiniMax-H3 \
  --omni \
  --task-type fl2va \
  --enable-diffusion-lora \
  --diffusion-lora \
    '{"name":"turbo","path":"lightx2v/Minimax-h3-Turbo"}'
```

Request selection is name-only:

```json
{
  "loras": [
    {"name": "turbo", "scale": 1.0}
  ]
}
```

## Current validation

- 22 targeted CPU and regression tests pass.
- The generic executor is checked against the exact two-adapter reference formula.
- Startup parsing, immutable registration, unknown-name rejection, batching identity, CLI propagation, and legacy-path conflict rejection are covered.
- The real LightX2V v1.0 checkpoint is parsed into 312 normalized H3 updates with rank 128 and alpha 128.
- Ruff, formatting, typos, SPDX, forbidden-import, and CUDA-call hooks pass; the new runtime package is mypy-clean.
- Four-H200 smoke tests pass with eager and regional compile under both USP4/Ring1 and TP4 at 768×1344, 107 frames, and 4 NFE. In the same server and with the same prompt/seed, the Base and Turbo requests produce different raw video and PCM-audio hashes (video SSIM about 0.50), confirming request-level activation. Compiled Turbo Stage 0 was 10.124 s with USP4 and 9.915 s with TP4; switching from Base to Turbo did not incur another compile pause.

This PR intentionally remains a draft while review and broader validation continue.

This is a smaller replacement direction for the broad experiment in #6017 and builds on the model-adapter motivation demonstrated by #2783. Thanks to everyone who reviewed and discussed those implementations.

### DIFF
--- a/docs/design/feature/diffusion_lora_runtime.md
+++ b/docs/design/feature/diffusion_lora_runtime.md
@@ -7,7 +7,8 @@ request-level LoRA selection. It is introduced in parallel with the legacy
 diffusion LoRA facilities. Existing models, examples, and request fields keep
 their legacy behavior until they are migrated explicitly.
 
-The first integration is MiniMax-H3 FL2VA Turbo. This initial scope supports:
+The first integration is MiniMax-H3 FL2VA, led by its Turbo deployment. This
+initial scope supports:
 
 - dynamic low-rank updates;
 - immutable startup registration;
@@ -75,6 +76,24 @@ admission. An empty or omitted composition selects the base model.
 The registry is not a default composition: registering an adapter makes it
 available but does not activate it automatically.
 
+Checkpoint format and adapter identity are independent. A model selects a
+decoder from checkpoint metadata and validates the tensors; it never selects a
+decoder from the user-provided deployment name. Multiple registered adapters
+may therefore share one format decoder while retaining distinct names, weight
+banks, and request scales.
+
+For example, a service can register two model-compatible adapters and compose
+them by name at request time:
+
+```bash
+--diffusion-lora '{"name":"fast","path":"/models/fast.safetensors"}' \
+--diffusion-lora '{"name":"style","path":"/models/style.safetensors"}'
+```
+
+```json
+{"loras":[{"name":"fast","scale":1.0},{"name":"style","scale":0.7}]}
+```
+
 ## Composition and execution
 
 For a base linear transform `W` and selected low-rank updates `(A_i, B_i)`, the
@@ -124,10 +143,12 @@ The first implementation has the following feature boundaries:
 
 ## MiniMax-H3 integration
 
-MiniMax-H3 owns the LightX2V Turbo v1.0 conversion. Its loader validates the
-rank-128, alpha-128 contract, maps Diffusers targets to native H3 modules,
-binds Q/K/V updates to packed QKV projections, and swaps the Diffusers FFN
-`lora_B` row order from `[value; gate]` to native `[gate; up]`.
+MiniMax-H3 owns both the LightX2V Turbo v1.0 conversion and normalization of
+the native H3 FL2VA fused-QKV layout. The Turbo path validates its rank-128,
+alpha-128 contract and swaps the Diffusers FFN `lora_B` row order from
+`[value; gate]` to native `[gate; up]`; the native path reads its declared rank
+and splits fused QKV updates into logical Q/K/V bindings. This format support
+does not endorse or register a specific third-party adapter.
 
 Sampling steps, flow shifts, task selection, and other recipe behavior remain
 pipeline/request concerns. Installing a Turbo adapter does not silently change
--- a/docs/design/feature/diffusion_lora_runtime.md
+++ b/docs/design/feature/diffusion_lora_runtime.md
@@ -1,0 +1,151 @@
+# Diffusion LoRA Runtime
+
+## Status and scope
+
+The Diffusion LoRA Runtime is a model-declared, startup-loaded backend for
+request-level LoRA selection. It is introduced in parallel with the legacy
+diffusion LoRA facilities. Existing models, examples, and request fields keep
+their legacy behavior until they are migrated explicitly.
+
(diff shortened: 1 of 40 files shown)


docs/design/feature/diffusion_lora_runtime.md
```diff
@@ -0,0 +1,151 @@
...
+MiniMax-H3 owns the LightX2V Turbo v1.0 conversion. Its loader validates the
+rank-128, alpha-128 contract, maps Diffusers targets to native H3 modules,
+binds Q/K/V updates to packed QKV projections, and swaps the Diffusers FFN
+`lora_B` row order from `[value; gate]` to native `[gate; up]`.
+
+Sampling steps, flow shifts, task selection, and other recipe behavior remain
+pipeline/request concerns. Installing a Turbo adapter does not silently change
+the execution plan.
+
+## Validation
```
hsliuustc0106, line 157: P3 (docs): the design page is thorough, but the user-facing surfaces are untouched: `docs/user_guide/diffusion_features.md` (LoRA column of the compatibility matrix; the batching note at :232 still describes only `lora_request`/`lora_scale`) and `recipes/MiniMaxAI/MiniMax-H3.md` have no mention of the new runtime. Since `--enable-diffusion-lora` and the `loras` request field are public contracts, updating those two before ready-for-review would keep users from mixing legacy and new fields.

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
```diff
@@ -559,6 +561,14 @@ class MiniMaxH3Pipeline(
     # uniform path available to partially constructed pipelines.
     _base_schedule_by_partition: ClassVar[Mapping[str, DMD2SigmaSchedule | None]] = {}
 
+    @property
+    def diffusion_lora_support(self):
+        """Expose FL2VA LoRA semantics without enabling Ref2VA adapters."""
+
+        if self.partition == "ref2va":
```
hsliuustc0106, line 568: P3: on a Ref2VA-only service started with `--enable-diffusion-lora`, this `return None` makes startup fail with the generic `MiniMaxH3Pipeline does not declare Diffusion LoRA Runtime support`, and the loader's clearer `Ref2VA-only` rejection in `MiniMaxH3TurboLoRALoader.__init__` becomes unreachable in production (its test constructs the loader directly). Returning the support object here and letting the loader raise would surface the partition-specific message.

vllm_omni/diffusion/worker/diffusion_worker.py
```diff
@@ -447,6 +449,11 @@ def init_lora_manager(self) -> None:
         if self.model_runner.pipeline is None:
             return
 
+        self.diffusion_lora_runtime = self.model_runner.diffusion_lora_runtime
+        if getattr(self, "diffusion_lora_runtime", None) is not None:
```
hsliuustc0106, line 453: P3: `load_model(load_format="dummy")` returns before `_init_diffusion_lora_runtime` (diffusion_model_runner.py:275), so a dummy-load service started with `--enable-diffusion-lora` comes up with `diffusion_lora_runtime=None` while the engine still accepts `loras` requests at admission (it parses registered names from config); they then fail per-request inside `_activate_lora_state` with the misleading "service did not enable" message. A guard here — `enable_diffusion_lora` set but runtime None → raise — would turn this into a startup failure.

vllm_omni/diffusion/worker/diffusion_model_runner.py
```diff
@@ -220,6 +221,37 @@ def _compile_transformer(self, attr_name: str) -> None:
...
+            name
+            for name in (
+                "enable_cpu_offload",
+                "enable_layerwise_offload",
+                "enable_distributed_layerwise_offload",
+            )
+            if getattr(self.od_config, name, False)
+        ]
+        if incompatible:
+            raise ValueError(f"Diffusion LoRA Runtime first release does not support offload; disable {incompatible}")
```
hsliuustc0106, line 240: P3/question: composition participates in both scheduler batch keys, but not in any Diffusion-KV cache identity, and `enable_diffusion_lora` + `diffusion_kv_mode=paged_scheduler` is not rejected here alongside offload. The design page (diffusion_lora_runtime.md:123) defers this to future prefix-publication work — but if paged prefix reuse can reach an integrated model, base and turbo requests could share KV blocks across compositions. Either reject the combination in this list now, or note why it is unreachable for the H3 integration.

PR #6476, opened 2026-08-22
[Feature] Support MiniMax-H3 Turbo LoRA with the legacy manager

mglyn:
## Motivation and scope

MiniMax-H3 urgently needs the published LightX2V Turbo LoRA, while its Diffusers checkpoint names and FFN layout do not match the native H3 transformer.

This is one of two alternative directions currently under consideration:

- **this PR** is the minimal H3-only patch on the existing `DiffusionLoRAManager` interface;
- #6473 introduces a separate model-declared Diffusion LoRA Runtime with startup registration and a cleaner long-term contract.

## Changes

- Add an optional model loader hook to the existing diffusion LoRA manager while preserving its generic PEFT fallback and public interface.
- Recognize the LightX2V MiniMax-H3 Turbo v1.0 safetensors contract (`key_format=minimax-h3-diffusers`, rank/alpha `128/128`).
- Map Diffusers transformer and token-refiner names to native H3 modules and restore the native fused FFN `[gate; up]` row order.
- Bind separate Q/K/V adapters to native packed QKV and fix global-vs-rank-local fused LoRA B slicing under TP.
- Reject Ref2VA use and enforce the published Turbo sampling contract: five sigma points (four denoiser evaluations), video flow shift 6, and audio flow shift 3.

The model-specific dynamic loader is parallel to the prefusion integrations added by #2783. This PR does not add another prefusion path.

## Validation

- 147 related CPU/regression tests pass, including checkpoint conversion, fallback, invalid metadata, packed QKV, TP fused slicing, lifecycle reuse, and binding completeness.
- The real v1.0 checkpoint loads and binds all 312 logical LoRA targets (624 A/B tensors) to 208 physical wrappers without missing pairs or activation shape mismatches.
- TP2 eager and TP2 regional `torch.compile` both complete 768×1344, 107-frame T2VA generation with H.264 video and AAC audio.
- At four NFE, the no-LoRA control is visibly degraded while Turbo restores coherent video and audio.

Performance spot check on the current minimal branch: 4× NVIDIA H200; Transformer sequence parallelism `USP4/Ring1` (no DiT tensor parallelism); VAE patch parallelism 4; text-encoder TP1; regional `torch.compile`; and default FlashAttention. All cases use the same 768×1344 FL2VA input, 107 frames/24 FPS, prompt, and seed. Times are median `stage_0_gen_ms` from five runs after two full-shape warmups.

| Case | LoRA execution | NFE | Stage 0 p50 | Five-run range |
|---|---|---:|---:|---:|
| Base reference | None | 49 | 68.388 s | 68.336–69.043 s |
| No-LoRA control | None | 4 | 8.967 s | 8.943–9.008 s |
| Turbo | Dynamic | 4 | 9.688 s | 9.639–10.231 s |

Dynamic Turbo at four NFE is **7.06× faster** than the 49-NFE reference, with **8.05%** overhead over the four-NFE no-LoRA control. These measurements exercise this PR's legacy dynamic LoRA wrapper path.

## I2V demo

https://github.com/user-attachments/assets/82d02c0f-35fd-4162-a5f4-3231bbd8b554

## Known limitations

- Dynamic execution only: no prefusion.
- No DLO support is added; this keeps the existing legacy backend behavior.
- Only one LoRA can be active at a time, so Turbo cannot be composed with another style or identity LoRA.
- The legacy request schema includes the adapter path. A client can therefore cause the server to resolve or download request-supplied LoRA weights through `get_adapter_absolute_path`; this inherited behavior may be unsuitable for an untrusted public endpoint and is not addressed here.
- Formal support is limited to LightX2V MiniMax-H3 Turbo v1.0 on FL2VA/T2VA.


mglyn:
@SamitHuang @hsliuustc0106

vllm_omni/diffusion/models/minimax_h3/lora.py
```diff
@@ -0,0 +1,144 @@
...
+
+    candidates = sorted(path.glob("*v1.0*.safetensors"))
+    if not candidates:
+        return None
+    if len(candidates) != 1:
+        raise ValueError(
+            "MiniMax-H3 Turbo v1.0 must resolve to exactly one safetensors "
+            f"file, found {[candidate.name for candidate in candidates]}"
+        )
+    return candidates[0]
```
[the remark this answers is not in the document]
mglyn: Thanks. The current revision intentionally constrains support to the exact published 4-step v1.0 native artifact. Metadata alone cannot distinguish the different step/packaging variants, so directory resolution now selects the declared filename and the H3 recipe documents that scope.

vllm_omni/diffusion/models/minimax_h3/lora.py
```diff
@@ -0,0 +1,144 @@
...
+    pairs: dict[str, set[str]] = {}
+    for name in checkpoint.keys():
+        if name.endswith(_LORA_A_SUFFIX):
+            raw_target = name[: -len(_LORA_A_SUFFIX)]
+            side = "a"
+        elif name.endswith(_LORA_B_SUFFIX):
+            raw_target = name[: -len(_LORA_B_SUFFIX)]
+            side = "b"
+        else:
+            raise ValueError(f"Unconsumed MiniMax-H3 Turbo tensor: {name!r}")
```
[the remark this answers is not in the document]
mglyn: Thanks. Keeping this strict is intentional because the loader supports one exact artifact. Unknown tensors may represent required bias or dense updates; silently ignoring them could produce an apparently successful but incorrect load. Generic fallback still applies when key_format does not match.

hsliuustc0106:
P3 (evidence): the performance table's provenance note says the numbers were collected before this minimal branch was split, but that branch's dynamic execution ran through its own modified `vllm_omni/diffusion/lora/layers/` stack (including a new `torch_linear.py` that does not exist on main), while this PR executes on main's unmodified upstream-derived wrappers. The loading semantics are indeed identical, but the "retains the same vLLM dynamic LoRA layer execution" sentence is not literally true, so the 3.56%-overhead figure may not transfer exactly. Since TP2 eager/compile generation already runs on this branch, one warm timing spot-check of `stage_0_gen_ms` (Base vs no-LoRA vs Turbo at 4 NFE) would let the table cite this branch directly.

hsliuustc0106:
@lishunyang12 @Isotr0py @princepride PTAL

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
```diff
@@ -1655,7 +1681,11 @@ def forward(self, request: DiffusionRequestBatch) -> DiffusionOutput:
         quality = sampling.quality
         logger.debug("MiniMax H3 request quality=%s", quality)
         extra = sampling.extra_args or {}
-        task = self._resolve_task(extra.get("task"), multi_modal_data)
+        task = self._resolve_task(
+            extra.get("task"),
+            multi_modal_data,
+            has_lora=sampling.lora_request is not None,
```
Gaohan123: **[P2] Check whether the Turbo adapter is actually active**

This treats any non-null `lora_request` as an active MiniMax-H3 Turbo adapter. However, `DiffusionLoRAManager.set_active_adapter()` explicitly deactivates all adapters when `lora_scale` is zero. A Ref2VA request carrying a LoRA request with `scale=0` will therefore run with no active LoRA, but still fail here with `MiniMax-H3 Turbo LoRA supports T2VA/FL2VA requests only`.

This also classifies checkpoints handled by the generic PEFT fallback as Turbo adapters, even though the model-specific loader returns `None` for those checkpoints.

Please base this restriction on the actually active adapter type/state rather than merely the presence of `lora_request`, and add regression coverage for Ref2VA with `lora_scale=0` and, if generic H3 PEFT adapters remain supported, the generic fallback path.

mglyn: fixed.

lishunyang12 (commented):
Review against current head `5dc0cb2d72b1c3de2f6fb3d3f2c669cfabb4114a`.

The model-specific conversion is deliberately scoped and the core mechanics look sound: strict v1.0 target validation, Diffusers-to-native name mapping, FFN row restoration, separate Q/K/V binding to packed QKV, and global-vs-rank-local LoRA-B slicing. I also reran the three affected CPU/regression test files on this head:

```text
141 passed
```

The real-checkpoint TP2/B300 validation from the previous head is available here:
https://lishunyang12.github.io/vllm-omni-rankings/scripts/minimax_h3_turbo_lora_pr6476_20260822/

### Correctness issue to address

**[P2] Keep Turbo adapter classification synchronized with the manager lifecycle.**

`MiniMaxH3Pipeline._turbo_lora_adapter_ids` records every ID recognized as Turbo, but `DiffusionLoRAManager.remove_adapter()` and LRU eviction do not clear or update that pipeline-owned set. The request schema also permits an explicit client-supplied `int_id`.

A concrete sequence is:

1. load Turbo with ID 7, so the pipeline records 7;
2. evict ID 7 from the manager;
3. load a generic PEFT adapter with the same ID 7;
4. `_has_active_turbo_lora()` still classifies it as Turbo and rejects a nonzero-scale Ref2VA request.

The latest fix correctly handles scale 0 and generic fallback before an ID has been recognized, but it does not cover eviction/reuse. Adapter kind should ideally live with the manager's registered-adapter state and be removed with it. A smaller fix could key the classification by both ID and canonical resolved path and replace/discard it on every load. Please add a regression covering `Turbo -> eviction -> generic same ID -> Ref2VA`.

### Additional improvements

1. **Validate the supported sampling contract when Turbo is active.** This PR formally supports one 4-step artifact, but a request can currently activate it with an arbitrary `num_inference_steps` and `flow_shift`. Rejecting unsupported values (at least steps != 4 or video flow shift != 6), or emitting a prominent warning, would prevent successful-but-invalid requests.

2. **Fail closed for a declared Turbo file with invalid metadata.** The loader checks `key_format` before the declared filename. Therefore an exact `minimax_h3_fl2v_turbo_4step_v1.0_768p_bf16.safetensors` file with missing/corrupt metadata returns `None` and falls into the generic PEFT path. If the declared file was selected, invalid metadata should produce a direct Turbo contract error; arbitrary non-H3 files can still use the fallback.

3. **Verify binding completeness, not only artifact completeness.** The loader proves that all 312 logical targets exist in the checkpoint, but the generic activation path can reset/skip missing or mismatched destination layers and still mark the adapter active. A post-binding invariant should ensure every normalized H3 target was bound to the expected physical/paked layer, including all Q/K/V slices.

4. **Rerun the short GPU transition test on this head.** The published E2E used `53f54af`; the current head adds adapter-state logic. A short `no LoRA -> Turbo -> no LoRA -> Turbo` sequence for both T2VA and FL2VA would close that validation gap and catch residual activation state.

The request-supplied path/download behavior is already documented as a known limitation. For any public endpoint, I would still recommend a startup allowlist/name-only selection or an explicit option disabling request-time adapter resolution.

mglyn:
Thanks for the detailed review. Fixed in `4f83ae9d3`: Turbo classification is now replaced on every real load, including eviction followed by same-ID reuse; the declared artifact fails closed on invalid metadata; activation verifies all 312 logical targets, including packed Q/K/V; and Turbo requests enforce five sigma points (four denoiser evaluations), video shift 6, and audio shift 3. The three related CPU suites now pass 147 tests.

I also reran the requested transition on 4× NVIDIA H200/USP4. Base and Turbo were each warmed once before the measured `Base → Turbo → Base → Turbo` sequence; every request used the same prompt and seed. The hashes below are decoded stream SHA256 values, rather than MP4 container hashes.

| Task/state | Stage 0 run 1 / run 2 | Video SHA256 (run 1 = run 2) | Audio SHA256 (run 1 = run 2) |
|---|---:|---|---|
| T2VA Base | 8.265 / 8.253 s | `e84bf659227a…` | `2ff80897b879…` |
| T2VA Turbo | 8.989 / 9.039 s | `80d6533c050b…` | `05acb1fa0811…` |
| FL2VA Base | 9.053 / 9.105 s | `9b65a96e68af…` | `a54a71cf23b5…` |
| FL2VA Turbo | 9.844 / 9.857 s | `f6146fa8cc5a…` | `9f0420c0b867…` |

All eight requests returned valid 768×1344, 107-frame H.264/AAC outputs. Each state reproduced exactly after the intervening state, while Base and Turbo hashes differed for both video and audio, so the result also rules out a no-op transition.

vllm_omni/diffusion/models/minimax_h3/lora.py
```diff
@@ -0,0 +1,168 @@
...
+        mapped_target = mapped_name.rsplit(".lora_", 1)[0]
+        if mapped_target.rsplit(".", 1)[-1] not in _TURBO_TARGETS:
+            raise ValueError(f"Unsupported MiniMax-H3 Turbo target: {raw_target!r}")
+        target_sides = pairs.setdefault(mapped_target, set())
+        if side in target_sides:
+            raise ValueError(f"Duplicate MiniMax-H3 Turbo tensor for {mapped_target}.{side}")
+        target_sides.add(side)
+
+        tensor = checkpoint.get_tensor(name)
+        if tensor.ndim != 2:
```
hsliuustc0106, line 106: This validates only `ndim` and the rank axis. A declared artifact with all 624 expected keys but an undersized outer dimension is accepted: vLLM zeros the destination and copies only the source extents, so the remainder stays zero, `_record_bound()` still records the target, and the name-only validator passes. I reproduced this with the real diffusion wrapper: undersized A/B tensors were accepted and marked active. Please preflight every target’s complete global A/B shape before mutating wrappers and cover undersized/oversized TP2 tensors.

vllm_omni/diffusion/lora/manager.py
```diff
@@ -635,6 +668,18 @@ def _activate_adapter(self, adapter_id: int, scale: float) -> None:
                 scale,
             )
 
+        if callable(binding_validator):
```
hsliuustc0106, line 660: This cleanup only covers the final validator. Any preceding `set_lora()` can raise after earlier layers were overwritten, leaving `_active_adapter_id` pointing to the previous adapter. Retrying that old ID/scale then takes the “already active” fast path and preserves mixed weights. I reproduced exactly this 1 → 2 failure and stale retry. Please make the whole binding loop plus validator transactional—restore the prior adapter or reset every wrapper and invalidate active state—and add a mid-loop failure regression.

recipes/MiniMaxAI/MiniMax-H3.md
```diff
@@ -759,10 +763,46 @@ balanced switch order.
...
+
+```bash
+export TURBO_DIR=/path/to/minimax-h3-turbo
+export TURBO_FILE=minimax_h3_fl2v_turbo_4step_v1.0_768p_bf16.safetensors
+hf download lightx2v/Minimax-h3-Turbo "${TURBO_FILE}" --local-dir "${TURBO_DIR}"
+export TURBO_LORA="${TURBO_DIR}/${TURBO_FILE}"
+```
+
+Add `--task-type fl2va --lora-backend peft --lora-path "${TURBO_LORA}"` to
+one of the FL2VA server commands above. `--lora-path` preloads the adapter;
```
hsliuustc0106: This includes the single-GPU `--enable-cpu-offload` command, but dynamic LoRA is not compatible with that lifecycle. The worker enables H3 offload before manager injection; vLLM stores `lora_a_stacked`/`lora_b_stacked` as ordinary tensor tuples, while `SequentialOffloadHook` moves only parameters and registered buffers. With the real wrapper I confirmed the base weight moves devices while A/B stay on CPU, so the first active Turbo forward will device-mismatch. Please stage/register these tensors with the component, or explicitly reject and document CPU offload, and add a regression.

hsliuustc0106:
Review and local-validation summary for current head `4f83ae9d`.

### Scope discussed

- [MiniMaxAI](https://huggingface.co/MiniMaxAI/MiniMax-H3) provides the base H3 model; [LightX2V](https://huggingface.co/lightx2v/Minimax-h3-Turbo) provides the supported Turbo LoRA weights. This PR integrates those third-party weights.
- LoRA itself does not imply distillation or fewer denoising steps. This specific Turbo adapter is trained for a few-step schedule: five sigma points produce four denoiser evaluations. Ordinary style or identity LoRAs still need the base schedule.
- This path is dynamic rather than prefused. Each active target computes the base projection plus two low-rank projections. H3 has 312 logical targets across 52 blocks, so each denoiser evaluation is slower, but reducing the PR reference from 49 evaluations to 4 dominates total runtime.

### Local verification

Environment: 2× NVIDIA L20X, TP2 eager/CUDNN attention, text-encoder TP2, VAE patch-parallel 2/tile, 1344×768 T2VA, 107 frames, fixed prompt/seed, one warmup plus three measured requests.

- Related CPU suite: **147 passed**.
- Real artifact: 624 BF16 tensors / 312 logical adapters, 1.2886 GiB payload.
- Adapter load: 2.356 s first in-process load; 0.195/0.188 s warm; 4.87 GiB phase-peak process RSS.

| Case | Stage-0 mean ± SD | Stage-0 p50 | Diffuse mean ± SD | Peak reserved VRAM/rank |
|---|---:|---:|---:|---:|
| main / no LoRA | 15.169 ± 0.054 s | 15.145 s | 10.196 ± 0.014 s | 75,598 MiB |
| PR / inactive LoRA | 15.621 ± 0.741 s | 15.245 s | 10.185 ± 0.017 s | 76,762 MiB |
| PR / Turbo | 16.568 ± 0.354 s | 16.385 s | 11.424 ± 0.006 s | 76,762 MiB |

Conclusions:

- Inactive wrappers preserve compute and output behavior: PR/no-LoRA diffuse time was -0.10% versus main, and all three MP4s were byte-identical to main at container, decoded-video, and decoded-audio levels.
- Active Turbo added 12.17% denoiser time. Stage-0 p50 overhead was **7.48%**, close to the reported 8.05% on different hardware/topology.
- Supporting LoRA added **1,164 MiB (+1.54%)** reserved VRAM per TP2 rank and roughly 3.3 GiB median aggregate process-tree RSS. Activation itself added no further peak because buffers were preallocated.
- All nine outputs decoded successfully as 1344×768, 107-frame H.264 plus stereo 32 kHz AAC. Turbo was deterministic, distinct from no-LoRA in both streams, and visibly coherent.
- I did not rerun the 49-evaluation reference; the local A/B focused on dynamic-wrapper parity and Turbo overhead at the required four evaluations.

### Outstanding blockers

1. [Validate complete model-facing A/B shapes before binding](https://github.com/vllm-project/vllm-omni/pull/6476#discussion_r3838587120). Current undersized tensors are silently zero-padded and still counted as bound.
2. [Make adapter activation transactional](https://github.com/vllm-project/vllm-omni/pull/6476#discussion_r3838587441). A mid-loop failure can leave mixed weights while manager state names the previous adapter.
3. [Handle or explicitly reject model-level CPU offload](https://github.com/vllm-project/vllm-omni/pull/6476#discussion_r3838587797). The base weights move to GPU while unregistered LoRA A/B tuples remain on CPU.

Recommended verdict remains **REQUEST_CHANGES** until these paths are addressed and covered by regressions.

mglyn:
Addressed the follow-up issues in commit 102e3718c:

1. The H3 loader now validates the complete global A/B shape for every supported Turbo target before wrapper mutation.
2. Adapter activation is transactional: any binding or model-validator failure resets every wrapper and invalidates the active state.
3. H3 Turbo now explicitly rejects model-level CPU offload, layerwise offload, and distributed layerwise offload, with the limitation documented.
4. The H3 loader converts fused fc1 into model-owned packed gate/up LoRA weights; the common manager no longer guesses the H3 TP layout.
5. Invalid Turbo step/shift values now consistently surface as OmniClientError request errors.

Validation: 41 focused CPU tests passed, changed-file pre-commit passed, and the real v1.0 artifact loaded as 312 logical adapters with 52 packed fc1 entries. A TP2 eager smoke on 2× NVIDIA H200 completed both Base and Turbo 768×1344, 107-frame H.264/AAC requests without skipped bindings or shape mismatches.

recipes/MiniMaxAI/MiniMax-H3.md
mglyn: Fixed in 102e3718c by explicitly rejecting the unsupported lifecycle before loading Turbo tensors. Model-level CPU offload, layerwise offload, and distributed layerwise offload are all covered by regressions and documented; the supported non-offloaded TP2 path was rerun successfully.

vllm_omni/diffusion/lora/manager.py
mglyn: Fixed in 102e3718c. The active ID is invalidated before the first mutation, and the complete bind loop plus model validator now runs transactionally. Any exception resets every wrapper. The regression covers successful adapter 1 activation, a mid-loop adapter 2 failure, and adapter 1 retry to prove the stale fast path is not taken.

vllm_omni/diffusion/models/minimax_h3/lora.py
mglyn: Fixed in 102e3718c. The H3 loader now checks exact global A/B dimensions for all six target kinds before LoRAModel construction or wrapper mutation, with undersized and oversized regressions. H3 fused fc1 is also converted into explicit packed gate/up weights instead of relying on a common-manager TP layout guess. The real v1.0 artifact and a TP2 Base/Turbo forward smoke both passed.

RuixiangMa:
LGTM

lishunyang12 (approved):

LeslieWylie:
Field report from a MiniMax-H3 deployment: there is one PEFT edge case worth guarding even if this PR intentionally supports only the published LightX2V `minimax-h3-diffusers` format.

We have seen converted LightX2V Turbo adapters in a fused/native-H3 PEFT layout with **208 physical targets** (52 blocks × `{qkv_proj,out_proj,fc1,fc2}`):

- `qkv_proj`: rank 384 (`A[384,5376]`, `B[21504,384]`) — three rank-128 Q/K/V adapters fused along rank/output;
- `out_proj`, `fc1`, `fc2`: rank 128;
- no `adaln_proj` / `final_layer`;
- `adapter_config.json` has only global `r=384`, `lora_alpha=384`, with no `rank_pattern` / `alpha_pattern` and no safetensors metadata.

We verified two sibling 4-step/8-step adapters with this exact structural contract. In these converted checkpoints the tensor payload already carries the effective scaling semantics; treating the global PEFT `r/alpha` as authoritative for every physical module can therefore silently mis-scale the rank-128 modules instead of failing loudly.

Suggestion for the generic PEFT fallback / validation path: derive and validate the **actual per-module rank from A/B tensor shapes**, and if it disagrees with global `r` without an explicit `rank_pattern`/model-owned conversion contract, reject with an actionable error (or route to a model-specific loader) rather than silently applying one global scale. At minimum, a regression test for mixed-rank packed/fused adapters would prevent this class of silent correctness bug.

This also interacts with packed QKV: after separate Q/K/V are fused, one physical wrapper legitimately has rank 3× the logical rank while the FFN/O-proj wrappers retain the original rank.

We also built a deterministic offline prefusion path for this converted H3 layout (208 targets rewritten into the native transformer) and can contribute a small synthetic fixture / rank-validation test if useful. I’m intentionally not proposing that vLLM-Omni support this opaque converted format as a public contract; the main ask is to make the generic fallback safe against mixed-rank metadata mismatch.

PR #5937, opened 2026-08-09
[Perf][Diffusion] MiniMax-H3: leaner VAE decode (autocast cache off, in-place revert, uint8 output)

brandneway:
### Purpose

显存快照分析显示 MiniMax-H3 的 VAE 解码峰值约 **11.75 GiB**(npu-smi 观察"解码时涨约 10G"),
其中三块可优化:

1. **autocast fp16 权重缓存 ~4.5 GiB**:`decode()` 在 fp16 autocast 下跑 fp32 权重的 VAE,
   autocast 默认把每个权重的 fp16 cast 副本缓存到整个 region 结束。但 tiled decode 每个 tile/chunk
   内每个权重只用一次,缓存零复用,纯白占;
2. **revert 链 ~4.8 GiB**:`processor.revert_tensor` 做反归一化时三步各新建一份全量(2.41 GiB)拷贝
   (Normalize 输出、clamp 输出、末尾 contiguous);
3. **后处理 ~250 ms**:`.float().cpu()` 下卡 2.41 GiB fp32,再在 CPU 上 clamp/×255 转 uint8。

### What this PR does

三个独立、可单独回退的优化:

- **A — autocast `cache_enabled=False`**(`pipeline.decode` + `interface.create_autocast_context`):
  关闭权重 cast 缓存,每个 tile 用完即释放。**省 ~4.5 GiB**。代价:每个 temporal chunk 重新 cast
  一遍权重(~9 ms × 6 chunk ≈ 55 ms),decode 是秒级,可忽略。
- **B — `_inplace_revert`**(`vae.decode_latent`):反归一化仍走 torchvision Normalize(1 份不可避免拷贝),
  但 clamp 改原地、去掉末尾 contiguous(downstream 只 slice/cast/D2H,不需要 contiguous)。
  **3 拷贝 → 1 拷贝,省 ~4.8 GiB**。
- **C — NPU/设备侧直出 uint8**(`pipeline.decode` + `post_process` + `video_api_utils`):
  decode 后直接 `clamp(0,1).mul(255).round().to(uint8)`,D2H 从 2.41 GiB fp32 降到 0.6 GiB uint8,
  且 mp4 编码器短路 uint8 不做 float 往返。**后处理 ~250 → ~31 ms(~8×)**。

实现注意:
- NPU 的 `inplace_sub` 对 `[1,3,1,1,1]` 广播有 op-plugin 对齐 bug(报 `8 and 3 cannot broadcast`),
[...]

PR #6477, opened 2026-08-22
[Core][Diffusion] Add bounded video output transports

NancyFyong:
## Summary

Adds bounded video delivery on top of the merged #6615 media contract, following RFC #6541:

- `/v1/videos/sync`: default bytes, base64, expiring URL, or same-host shared memory. The async job API rejects immediate-response modes.
- Consistent MP4/WebM policy (defaults: H.264/AAC and VP9/Opus), with deployment defaults and request encoder overrides. Encoding and publication stay off the event loop.
- Transactional multi-output publication and TTL cleanup. SHM is loopback-only, raw RGB uint8 and video-only; encoded outputs preserve audio.

This is presentation/transport work, not another typed-media model migration or a new streaming protocol.

## Review follow-ups

- **Worker encoder policy:** `preencode_mp4` receives the resolved deployment-plus-request codec/options, including deployment-only `crf=0`, through the shared H3 and WAN chunked-encoding paths. Real H.264 and HEVC encoding tests verify the selected codec; incompatible combinations fail before generation.
- **WebM audio:** both mux paths resample Opus-unsupported rates such as 44.1 kHz to 48 kHz, with real encode/decode coverage.
- **Cancellation and rollback:** URL/SHM publication is shielded and awaited before cleanup so a late writer cannot leave an orphan. SHM also handles repeated cancellation, second-output cancellation and publication failure.

## Performance

The review asked for default-path regression checks and **usable output**, not just a fast response or SHM handle. Measurements use the real HTTP route, fresh server/client processes, one warmup and three measured requests per case. Device postprocessing and worker pre-encoding are both disabled; MP4 uses H.264, CRF 18 and the ultrafast preset. Startup/warmup are excluded.

### Identical decoded frames

48 × 512 × 768 RGB frames, compressible content and random noise. Latency cells show **response headers / fully usable frames**, in ms; values are means. Usable time includes body transfer, URL download, decoding or SHM mapping and a full RGB read.

| Mode | Compressible latency | Noise latency | Noise server/client peak RSS (MiB) | Noise payload (MiB) |
| Base bytes | 35.3 / 132.1 | 74.8 / 223.9 | 1345 / 242 | 12.52 |
| PR bytes | 37.4 / 131.0 | 73.2 / 231.1 | 1329 / 236 | 12.52 |
| PR base64 | 40.0 / 139.4 | 132.0 / 312.5 | 1382 / 269 | 16.70 |
| PR URL | 35.3 / 134.3 | 77.1 / 240.7 | 1338 / 242 | 12.52 |
| PR SHM | 27.4 / 61.5 | 27.3 / 62.8 | 1308 / 183 | 54.00 raw |

RSS is the mean per-run peak, sampled at 5 ms; it includes shared pages and is not PSS. URL payload includes the artifact GET; SHM's 54 MiB is shared storage, not the HTTP handle size. Compressible encoded media is about 0.03 MiB, versus the same 54 MiB raw SHM allocation.

Selected noise-path costs: bytes encoding **70.6 ms**, client decode/read **146.4 ms**; URL publication **6.9 ms** and GET **22.5 ms**; base64 extraction **32.4 ms**; SHM publication **25.0 ms** and map/read **35.4 ms**. These components are not all additive.

Default noise bytes was **+3.3%** in this sample (base 213.5–230.8 ms; PR 227.9–236.7 ms). The overlapping, small samples do not establish either non-regression or a systematic slowdown. SHM avoids encoding/decoding and lowers measured client RSS, but trades compressed delivery for a larger raw allocation. URL is a lifecycle/storage option, not a demonstrated speedup.

### Real WAN2.2 workload

Two H20 GPUs, TP2, 512 × 512 × 17 frames, 20 steps, fixed seed. Mean [min–max] request-to-usable time; delivery starts when decoded frames are ready.

| Mode | Request → usable (s) | Frames ready → usable (ms) |
| --- | ---: | ---: |
| Base bytes | 12.72 [12.23–13.24] | 51.59 |
| PR bytes | 14.66 [12.81–17.28] | 52.95 |
| PR bytes, independent repeat | 12.48 [12.03–12.76] | 51.95 |
| PR SHM | 12.20 [11.14–13.28] | 39.89 |

**No stable end-to-end speedup is claimed.** SHM saves about 13 ms of delivery on a 12–15 s request; inference variability dominates. The slower first bytes run is retained, not replaced by the repeat. These shared-host, single-request loopback results are not throughput or remote-network claims.

In these measurements, encoded outputs match across base/PR transports; fixed-frame SHM matches source pixels exactly. H.264 is lossy, so raw-versus-encoded differences are expected. URL expiry, consumed/dropped SHM cleanup and output-shape checks passed.

Measurements compare base `2cbfd3e9` with runtime `51223a48` (also production-equivalent to `92adc541`). They predate the SHM cancellation fix and subsequent main synchronization; they are not presented as measurements of the latest head.

## Validation

After synchronizing with main in `a797df9c`: **993 passed, 3 skipped** in the broad suites, **120 engine/distributed tests passed**, and **all 3 WAN2.2 TP2 startup cases passed**. The skips are inherited HF-config/removed-API cases. The shared-encoder integration has **10 codec regressions**, with **8 expected failures against main**; real H.264/HEVC streams, per-request audio and transfer cleanup are covered. The SHM cancellation regressions remain passing. Applicable changed-file checks pass; baseline-normalized mypy has no new diagnostics, but is not globally clean. The unfiltered hook run was blocked by actionlint installation; it is not counted as a pass.

AI assistance: Pi coding agent assisted with implementation, tests, measurement scripts and this description; human review remains required.


vllm_omni/entrypoints/openai/serving_video.py
```diff
@@ -334,35 +362,62 @@ async def generate_videos(
...
+            # Skip encoding altogether: a same-host consumer maps the raw frames,
+            # so MP4 encoding and base64 would only add loss and copies.
+            from vllm_omni.diffusion.ipc import export_array_to_shm
+
+            from .video_api_utils import _coerce_video_to_uint8_frames
+
+            _t_shm_start = time.perf_counter()
+            video_data = [
+                VideoData(
+                    shm_handle=export_array_to_shm(_coerce_video_to_uint8_frames(video)),
```
SamitHuang, line 377: `export_array_to_shm` creates a POSIX segment that is only unlinked if some later caller runs `borrowed_shm_array`. A dropped HTTP response, a remote client that cannot map `/dev/shm`, or a crash between publish and release leaks a full-video segment until reboot. Do not expose this as an OpenAI-route `transport_mode` without a same-host gate and a TTL/sweeper (or unlink-on-disconnect) so `/dev/shm` cannot grow without bound.

vllm_omni/entrypoints/openai/serving_video.py
```diff
@@ -317,6 +320,31 @@ async def _run_and_extract(
...
+        """Store one encoded video and return a URL the client can fetch.
+
+        Prefers a URL published by the storage backend (a static server or object
+        store in front of it); otherwise points at this server's artifact route,
+        so URL mode needs no extra infrastructure.
+        """
+        from vllm_omni.entrypoints.openai import storage
+
+        storage_key = f"{uuid.uuid4().hex}.{output_format}"
+        await storage.STORAGE_MANAGER.save(video_bytes, storage_key)
```
SamitHuang, line 345: Sync URL artifacts are written under a UUID key and never registered in `VIDEO_STORE`. Default `FileBackend.file_ttl` is `None`, so these files accumulate forever, and `DELETE /v1/videos/{id}` cannot reclaim them. Require a TTL for URL mode (or record the keys and expire them) before this is a production transport; the existing TTL manager is unused on this path.

vllm_omni/entrypoints/openai/api_server.py
```diff
@@ -3743,6 +3748,32 @@ async def download_video(video_id: str) -> Response:
...
+        handle = await STORAGE_MANAGER.open(storage_key)
+    except ValueError:
+        # Raised for keys that try to escape the storage directory.
+        raise HTTPException(status_code=400, detail="Invalid artifact key") from None
+
+    if handle is None:
+        raise HTTPException(status_code=404, detail="Artifact not found")
+
+    if isinstance(handle, FileStorageHandle):
+        return FileResponse(path=handle.path, media_type="video/mp4", filename=storage_key)
```
SamitHuang, line 3771: `download_video_artifact` always serves `media_type="video/mp4"` even when `_store_video_artifact` writes `{uuid}.webm`. Clients and browsers will mis-handle WebM. Derive the type from the key suffix (or persist encoder settings) the way `/v1/videos/{id}/content` uses `job.media_type` — `media_type_for_format` already exists for this.

vllm_omni/diffusion/ipc.py
```diff
@@ -4,13 +4,29 @@
...
+
+Two read strategies exist, selected by
+:class:`~vllm_omni.diffusion.data.VideoOutputTransportConfig`:
+
+``copy`` (default)
+    The reader copies the payload out of the segment and unlinks it. Ownership
+    is trivially correct because the segment dies on read, at the cost of one
+    extra full host copy per hop.
+
+``shared_memory`` (zero-copy, opt-in)
```
SamitHuang, line 19: This docstring still says `VideoOutputTransportConfig` selects Hop-1 `copy` vs `shared_memory` unpack, but every engine unpack site still copies, and `transport_mode` now means the HTTP sink. Operators following this will think they enabled worker→engine zero-copy when they actually published raw frames on `/dev/shm` to every HTTP client. Update the docs to match the engine (always copy) and keep `shared_memory` as an explicit same-host consumer API, not a Hop-1 config switch.

SamitHuang (changes requested):
The device-side uint8 path is opt-in and the CPU/GPU gate tests plus the reported WAN RSS numbers address the original memory-pressure claim. The new HTTP sinks are not yet safe to ship as deployment-wide transport modes.

`transport_mode=shared_memory` publishes a POSIX segment on the OpenAI video route with no TTL and no same-host check: a dropped response or a remote client leaks `/dev/shm` for the lifetime of the node. `transport_mode=url` writes job-less files under default `file_ttl=None`, so disk grows without bound and `DELETE /v1/videos/{id}` cannot reclaim them. The artifact route also hardcodes `video/mp4` while this PR adds WebM.

Please fix those ownership/expiry bugs before merge. Also drop or rebase the Cosmos3 and MiniMax-H3 wirings if #5712 / #5937 land first, as already noted in the description. The ipc.py module docstring still describes `VideoOutputTransportConfig.transport_mode` as a Hop-1 unpack policy, which is not what the engine does.

vllm_omni/entrypoints/openai/serving_video.py
NancyFyong: Fixed in c6f63794. Shared-memory output now fails closed unless the API server is bound to a loopback host (127.0.0.0/8, ::1, or localhost), and every published response segment gets a configurable lease (`shared_memory_ttl_seconds`, default 300s). A single daemon sweeper unlinks unclaimed segments; normal `borrowed_shm_array` consumption still releases immediately. Tests cover non-loopback rejection, automatic expiry, early release, and zero `/dev/shm` growth across repeated runs.

vllm_omni/entrypoints/openai/serving_video.py
NancyFyong: Fixed in c6f63794 by reusing the existing storage TTL path. `StorageBaseManager` now declares whether it has an expiration policy, `LocalStorageTTLManager` opts in, and URL mode returns 503 before writing anything when expiration is absent. The built-in file backend is enabled with `VLLM_OMNI_SERVER_STORAGE__FILE_TTL`; tests assert the unsafe default manager writes no orphan file. The docs and out-of-tree backend contract were updated accordingly.

vllm_omni/entrypoints/openai/api_server.py
NancyFyong: Fixed in c6f63794. The artifact route derives MP4/WebM media type from the storage-key suffix through `media_type_for_format`. I also fixed the same issue in the raw `/v1/videos/sync` response and async job metadata/file extension, and added tests for both formats and all three paths.

vllm_omni/diffusion/ipc.py
NancyFyong: Fixed in c6f63794. The module docstring now states that Hop 1 always copies into engine-owned memory and unlinks on read. It describes the borrow helpers as separate explicit same-host APIs and no longer claims `VideoOutputTransportConfig.transport_mode` controls engine unpacking.

NancyFyong:
Self-review completed after the requested changes:

- Traced ownership for every URL and POSIX-SHM artifact. URL writes now require a backend expiration policy before the write; SHM publication is loopback-only and has a 300s configurable lease plus a single sweeper.
- Checked normal release, dropped-consumer expiry, and `/dev/shm` cleanup (three repeated IPC runs: segment-count delta 0).
- Checked MIME/container propagation for artifact downloads, `/v1/videos/sync`, and async video jobs (MP4 and WebM).
- Corrected the Hop-1 IPC documentation: engine unpack remains copy-and-unlink; the borrow API is separate.
- Ran 406 focused API/config/IPC tests (1 skipped), all green, plus mutation checks proving each of the four safety guards is covered.
- Ran the repository hooks on all touched files; all applicable hooks pass. The mypy result has zero new errors versus current `origin/main` (72 vs 72). No workflow files are touched, so actionlint remains skipped in this environment.
- Rechecked #5712 and #5937: both remain open, so the disclosed Cosmos3/MiniMax-H3 overlap has not yet been removed.

hsliuustc0106 (commented):
I do not expect to make too many changes to every related model, do we have a better or cleaner design

NancyFyong:
> I do not expect to make too many changes to every related model, do we have a better or cleaner design

Thank you for reply, I agree.

I plan to draft an RFC for a runner-owned device-output preparation contract:

1. **Canonical decoded-video descriptor**
   Pipelines return a canonical decoded-video descriptor containing the tensor,
   layout, value range, and whether any float-only consumer is still pending.

2. **Single device-postprocessing boundary**
   `DiffusionModelRunner` performs the single device-postprocessing step after
   `pipeline.forward()` and before D2H / IPC.

3. **Shared conversion, others untouched**
   Eligible videos are converted to `uint8` through the shared implementation,
   while audio, latent, image, and action outputs remain untouched.

4. **Model differences become declarative**
   Model-specific differences, such as MiniMax-H3's `[0, 1]` range, Cosmos3
   guardrails, WAN frame interpolation, and LingBot T2I, become declarative
   metadata or vetoes rather than duplicated control flow.

5. **Migrate existing pipelines**
   Pipelines such as LTX-2 and LingBot-Video would defer their current
   in-pipeline CPU conversion to this common boundary.
  
- **Reference implementation**: WAN can serve as the reference implementation.
- **Next steps**: After the RFC direction is agreed, I will reshape this PR or
  split the model migrations accordingly, rather than continuing with six
  independent forward / post-process branches.

NancyFyong:
> > I do not expect to make too many changes to every related model, do we have a better or cleaner design
> 
> Thank you for reply, I agree.
> 
> I plan to draft an RFC for a runner-owned device-output preparation contract:
> 
> 1. **Canonical decoded-video descriptor**
>    Pipelines return a canonical decoded-video descriptor containing the tensor,
>    layout, value range, and whether any float-only consumer is still pending.
> 2. **Single device-postprocessing boundary**
>    `DiffusionModelRunner` performs the single device-postprocessing step after
>    `pipeline.forward()` and before D2H / IPC.
> 3. **Shared conversion, others untouched**
>    Eligible videos are converted to `uint8` through the shared implementation,
>    while audio, latent, image, and action outputs remain untouched.
> 4. **Model differences become declarative**
>    Model-specific differences, such as MiniMax-H3's `[0, 1]` range, Cosmos3
>    guardrails, WAN frame interpolation, and LingBot T2I, become declarative
>    metadata or vetoes rather than duplicated control flow.
> 5. **Migrate existing pipelines**
>    Pipelines such as LTX-2 and LingBot-Video would defer their current
>    in-pipeline CPU conversion to this common boundary.
> 
> * **Reference implementation**: WAN can serve as the reference implementation.
> * **Next steps**: After the RFC direction is agreed, I will reshape this PR or
>   split the model migrations accordingly, rather than continuing with six
>   independent forward / post-process branches.

Hi, @hsliuustc0106 @SamitHuang ,RFC is here https://github.com/vllm-project/vllm-omni/issues/6541
we can discuss and I'll implement it in this pr

vllm_omni/entrypoints/openai/api_server.py
```diff
@@ -3542,6 +3549,7 @@ async def create_video_sync(
     raw_request.state.request_metadata = RequestResponseMetadata(request_id=request_id)
     started_at = time.perf_counter()
     try:
+        settings = handler._resolve_video_encoder(request)
```
hsliuustc0106: This route still calls `generate_video_bytes()` and returns raw bytes; the URL/shared-memory sink lives in `generate_videos()` but has no production caller. Thus `transport_mode=url` or `shared_memory` is ignored by both public video endpoints, so the loopback/TTL code is not exercised in the advertised server path. Please wire the sink into the public response contract and add route tests, or remove these modes from the server/CLI contract.

vllm_omni/diffusion/utils/media_utils.py
```diff
@@ -1,16 +1,140 @@
...
+    dropped in favour of the fallback's own defaults, because FFmpeg refuses
+    options from a different encoder family and would fail the whole encode.
+
+    Empty ``codec_options`` means "use the fast defaults for this codec". The
+    fallback follows ``output_format`` so we never hand a container a codec it
+    cannot hold.
+    """
+    if fallback is None:
+        fallback = default_video_codec_for_format(output_format)
+    requested = codec or fallback
```
hsliuustc0106: An explicit codec is only checked for encoder availability here, not compatibility with the selected container. `{"output_format":"webm","video_codec":"h264"}` resolves to H.264 and PyAV then raises `'webm' format does not support 'libx264' codec` instead of returning a valid artifact or client error. Please reject or normalize incompatible codec/format pairs, including when streaming forces MP4.

vllm_omni/entrypoints/cli/serve.py
```diff
@@ -638,6 +651,15 @@ def subparser_init(self, subparsers: argparse._SubParsersAction) -> TrackingArgu
             'MagCache: \'{"mag_threshold": 0.24, "mag_max_skip_steps": 5, "mag_retention_ratio": 0.1}\'. '
             "Calibration mode: add '\"mag_calibrate\": true'",
         )
+        omni_config_group.add_argument(
+            "--video-output-transport",
+            type=_json_object,
```
hsliuustc0106, line 684: `_json_object` is a new converter used by the parser below, but the docs hook executes this method with an `exec_globals` that does not define `_json_object`. The ReadTheDocs build is therefore failing with `NameError` while extracting CLI docs. Please add the converter to the hook globals and cover docs generation.

tests/config/test_omni_config.py
```diff
@@ -1089,6 +1089,63 @@ def test_diffusion_config_preserves_existing_coercion_hooks():
...
+
+    assert "video_output_transport" in projection_fields & diffusion_fields
+
+
+def test_omni_diffusion_config_normalizes_video_output_transport_mapping():
+    """The programmatic API accepts the same mapping that the CLI produces."""
+    from vllm_omni.diffusion.data import OmniDiffusionConfig, VideoOutputTransportConfig
+
+    cfg = OmniDiffusionConfig(
+        model="x",
```
hsliuustc0106: This test uses `model="x"`, which enters `OmniDiffusionConfig.__post_init__` and calls Hugging Face model resolution when HF offline mode is enabled (and `x` is not a valid local model). The same fixture is repeated in the new default-transport tests, so the tests fail before asserting transport behavior. Please use `model=None` or a local fixture, or patch `get_model_path`, so these CPU tests are hermetic.

vllm_omni/entrypoints/openai/api_server.py
```diff
@@ -2986,7 +2992,8 @@ async def _run_video_generation_job(
         updated_fields = {
             "status": VideoGenerationStatus.COMPLETED,
             "progress": 100,
-            "file_name": f"{video_id}.{job.file_extension}",
+            "file_name": f"{video_id}.{settings.output_format}",
```
hsliuustc0106: The file is still saved under the bare key `video_id`; only the metadata `file_name` gets the `.mp4`/`.webm` suffix. When `public_base_url` is configured, `STORAGE_MANAGER.open(video_id)` builds the redirect from that bare key, so async public URLs lose the container suffix and MIME-identifying filename, especially for WebM. Please save/open with the extension-bearing key or carry an explicit content type through the public-storage redirect.

benchmarks/diffusion/bench_device_postprocess.py
```diff
@@ -0,0 +1,194 @@
...
+def run_driver(rounds: int, batch: int, steps: int) -> None:
+    OUTPUT_DIR.mkdir(parents=True, exist_ok=True)
+    rows = []
+    for enable in (False, True):
+        for round_index in range(rounds):
+            env = dict(os.environ)
+            env["PYTHONPATH"] = str(REPO_ROOT) + os.pathsep + env.get("PYTHONPATH", "")
+            process = subprocess.run(
+                [
+                    sys.executable,
```
hsliuustc0106: This benchmark only compares flag-off/on inside the same PR checkout, and the documented run uses two fresh-process rounds. It does not provide a main-vs-PR MiniMax-H3 A/B, peak GPU memory, warm/cold separation, or audio correctness. Given the observed 1.43 GiB baseline transport and 11.561 s async D2H/SHM stall, please add the same-hardware/config comparison with at least three warm runs (startup reported separately) and video plus audio/PCM or latent hashes before treating the transport result as validated.

tests/entrypoints/test_serve.py
```diff
@@ -57,6 +60,30 @@ def test_serve_parser_accepts_deploy_config() -> None:
     assert args.get_explicit_kwargs_dict()["deploy_config"] == "/tmp/deploy.yaml"
 
 
+def test_serve_parser_accepts_video_output_transport() -> None:
```
hsliuustc0106, line 72: These tests stop at argparse and explicit-kwargs forwarding. Please add a startup regression covering omitted, false, and true through the two-worker diffusion bootstrap (or a lightweight equivalent), because the reported failure occurs after parsing and before readiness; parser/config tests cannot catch an EOF before the workers send their ready messages.

hsliuustc0106:
Local startup verification at PR head `135180d3` on 2x NVIDIA L20X (physical GPUs 1–2), using `gpu run`, the cached MiniMax-H3 FL2VA checkpoint, and the documented TP2/DLO controls: direct-Python no-flag health 210s; `{"enable_device_postprocess":false}` health 235s; `{"enable_device_postprocess":true}` health 215s. Both workers reached ready in all three cases; no EOFError or exit code 130 occurred. The first flagged attempt in this environment was malformed by nested-shell quoting and argparse correctly rejected it; the corrected rerun passed. This does not reproduce the reported other-node startup failure; if it remains reproducible, please attach the first worker stderr/traceback rather than only the parent EOFError.

docs/user_guide/diffusion/video_output_transport.md
```diff
@@ -0,0 +1,198 @@
+# Video Output Transport
```
hsliuustc0106: do we only to optimize video?

docs/user_guide/diffusion/video_output_transport.md
```diff
@@ -0,0 +1,198 @@
...
+
+Video payloads are orders of magnitude larger than image or audio payloads: a
```
hsliuustc0106: do we have a related feature design doc?

benchmarks/diffusion/bench_device_postprocess.py
```diff
@@ -0,0 +1,194 @@
...
+        "rss_delta_mib": round(peak_rss - baseline, 1),
+        "mp4_bytes": len(encoded),
+    }
+    print("RESULT " + json.dumps(result), flush=True)
+
+
+def run_driver(rounds: int, batch: int, steps: int) -> None:
+    OUTPUT_DIR.mkdir(parents=True, exist_ok=True)
+    rows = []
+    for enable in (False, True):
```
hsliuustc0106: Independent A/B at this head versus base `f446468c`, on the same 2x L20X pair with one warm-up plus three measured requests, found a 4.00x inferred transport reduction (float32 1464.75 MiB -> uint8 366.19 MiB), -34.3% process-tree RSS, -41.5% MP4 encoding time, and identical decoded video/AAC hashes. End-to-end/server latency averaged -6.8%, but PR variance was much higher (1.67 s vs 0.07 s baseline), while sampled GPU peak memory was essentially unchanged (-1.8%). Please add this methodology and uncertainty to the PR body and avoid presenting the e2e latency delta as a stable speedup.

vllm_omni/entrypoints/openai/api_server.py
NancyFyong: Fixed in 896fb6c6. `transport_mode` now defaults to `bytes`, so `/v1/videos/sync` keeps its existing raw MP4/WebM response. The same public route dispatches `base64`, `url`, and loopback `shared_memory` to `generate_videos()` and returns `VideoGenerationResponse` JSON. The async job endpoint rejects those immediate-response modes instead of ignoring them. Route tests exercise all four modes, URL retrieval, SHM borrow/release, and the safety rejections.

vllm_omni/diffusion/utils/media_utils.py
NancyFyong: Fixed in 896fb6c6. Encoder resolution now validates the selected codec against the effective container after hardware fallback. WebM + H.264 is rejected before generation with HTTP 400, and the fragmented streaming path is covered when it forces MP4 against an explicitly WebM-only codec.

vllm_omni/entrypoints/cli/serve.py
NancyFyong: Fixed in 896fb6c6. The MkDocs extraction globals now include `_json_object`; `tests/docs/test_generate_argparse.py` executes the statically extracted parser with `--video-output-transport`. I also ran `docs/mkdocs/hooks/generate_argparse.py` and confirmed the generated CLI page contains the option.

tests/config/test_omni_config.py
NancyFyong: Fixed in 896fb6c6. The new transport tests now construct `OmniDiffusionConfig(model=None)` rather than resolving `model="x"`/`"test"`. The config/API suite passes with HF offline mode.

vllm_omni/entrypoints/openai/api_server.py
NancyFyong: Fixed in 896fb6c6. Async jobs now compute one extension-bearing storage key (`<video_id>.mp4` or `.webm`) before generation and use it consistently for save, cleanup, open, delete, and public redirects. A route test verifies that a WebM job backed by `public_base_url` redirects to a URL ending in `<video_id>.webm`.

benchmarks/diffusion/bench_device_postprocess.py
NancyFyong: Reworked in 896fb6c6. The checked-in benchmark now compares an unmodified main checkout with the candidate in separate fresh processes, reports cold startup separately, requires at least one warmup plus three measured requests, samples process-tree RSS and `torch.accelerator` peak allocation, times encoding, and checks stable video plus optional audio/latent hashes. The PR body now contains the exact command and both the current-head WAN result and your independent MiniMax-H3 video/AAC result.

tests/entrypoints/test_serve.py
NancyFyong: Added `tests/e2e/online_serving/test_video_output_transport_startup.py` in 896fb6c6. It starts a real TP2 WAN diffusion server for omitted, explicit false, and explicit true, waits for `/health`, and checks the parent remains alive. All three cases passed locally in 590.83s with `/dev/shm` delta 0. I also separately repeated the three MiniMax-H3 TP2/DLO startup cases successfully; the automated regression uses WAN to avoid coupling this config-plumbing test to DLO's separate SHM lifecycle.

docs/user_guide/diffusion/video_output_transport.md
NancyFyong: For this PR, yes: the optimization scope is decoded video output. Image, audio, latent, and action contracts are unchanged. I made that explicit at the top of the guide. The outer media contract can be extended later, but this PR does not claim those modalities.

docs/user_guide/diffusion/video_output_transport.md
NancyFyong: Yes: #6541, `Stable Video Media Contract and Unified Pre-D2H Preparation`, is the related design RFC. The guide now links it directly and distinguishes this PR's current model-local implementation from the runner-owned follow-up design.

benchmarks/diffusion/bench_device_postprocess.py
NancyFyong: Thanks for running this. I added your setup, video/AAC hash result, and variance to the PR body and removed any stable latency claim. I also reran the final head `4d6967cf` against latest main `29e3165b` with one warmup plus three 40-step WAN requests: payload 370.195 -> 92.549 MiB, process-tree RSS 3697.6 -> 2692.9 MiB, encode 533.9 -> 458.2 ms, GPU peak effectively unchanged, and generation 65.299 -> 65.022 s. The body treats the generation delta as noise rather than a speedup.

vllm_omni/diffusion/postprocess/device_reduction.py
```diff
@@ -0,0 +1,97 @@
...
+    D2H copy carries uint8 instead of float. The result matches
+    ``VideoProcessor.postprocess_video(output_type="np")`` then the ``*255``
+    rounding done in the API server. Pass ``do_denormalize=False`` for VAEs that
+    already emit ``[0, 1]``.
+    """
+    if video.dim() != 5:
+        raise ValueError(f"expected a [B, C, F, H, W] video tensor, got shape {tuple(video.shape)}")
+
+    # Match the numpy path, which promotes to float before scaling.
+    frames = video.to(torch.float32)
```
RuixiangMa, line 145: On a near-full GPU, a request that previously succeeded could fail with torch.cuda.OutOfMemoryError and no fallback to the float path. Consider wrapping in try/except torch.cuda. OOM and returning the original float tensor as fallback.

vllm_omni/diffusion/models/hunyuan_video/pipeline_hunyuan_video_1_5.py
```diff
@@ -77,6 +83,14 @@ def get_hunyuan_video_15_post_process_func(od_config: OmniDiffusionConfig):
...
         if output_type == "latent":
             return video
+        if is_device_reduced(video):
+            # Already uint8 from the device-side reduction; Image.fromarray on
+            # uint8 matches the numpy_to_pil the float path would run.
+            frames = video.cpu().numpy()
+            if output_type == "np":
+                return frames
+            per_video = [[Image.fromarray(frame) for frame in single] for single in frames]
+            return per_video[0] if per_video else per_video
```
RuixiangMa: The float path flattens all batches into a list of PIL Images; the uint8 path returns only per_video[0]. The "np" path is correct (returns full [B, F, H, W, C]). Tests only cover B=1. Please verify B>1 works or add a guard.

vllm_omni/diffusion/models/lingbot_video/pipeline_lingbot_video.py
```diff
@@ -551,6 +560,11 @@ def _decode_latents(self, latents: torch.Tensor) -> torch.Tensor:
         with torch.autocast("cuda", dtype=autocast_dtype or torch.bfloat16, enabled=autocast_dtype is not None):
             decoded = self.vae.decode(vae_latents)
         frames = decoded[0] if isinstance(decoded, tuple) else decoded.sample
+        if reduce_to_uint8:
+            # LingBot denormalizes as clamp(-1,1) then (x+1)/2, which matches
+            # reduce_video_to_uint8_frames' (x*0.5+0.5).clamp(0,1).
+            frames = reduce_video_to_uint8_frames(frames)
+            return frames[0].cpu()
```
RuixiangMa: The float path returns [B, F, H, W, C]; the uint8 path returns frames[0]. The only batch-size guard is T2I-only, and T2I blocks reduction. Tests only cover B=1. Please verify B>1 works or add a guard.

hsliuustc0106:
fix conflicts

NancyFyong:
> fix conflicts

Thank you for your review，I‘ve fix conflicts.

hsliuustc0106:
Reviewed head `6583ac55fbb6ab724a9717b5705e7bcfde479b75`. Two findings:

- **[P1] Resample audio for WebM/Opus** — [media_utils.py:529](https://github.com/vllm-project/vllm-omni/blob/6583ac55fbb6ab724a9717b5705e7bcfde479b75/vllm_omni/diffusion/utils/media_utils.py#L529). The encoder receives the original audio sample rate, but Opus rejects 44.1 kHz, used by MAGI-2. Selecting WebM therefore fails after generation. Reproduced with real PyAV: MP4 succeeds at 44.1 kHz; WebM raises `avcodec_open2("libopus")`. Please resample to a supported rate in both mux paths and add 44.1 kHz coverage.

- **[P2] Wait for publication before cancellation cleanup** — [serving_video.py:556](https://github.com/vllm-project/vllm-omni/blob/6583ac55fbb6ab724a9717b5705e7bcfde479b75/vllm_omni/entrypoints/openai/serving_video.py#L556). Cancelling `save()` does not stop its background writer. Cleanup can delete a nonexistent artifact, then the writer creates it afterward, leaving an orphan until TTL expiry. Reproduced this ordering using the production storage classes and publication method with a delayed writer. Please retain and await the write task before rollback; the current cancellation test cancels only after saving finishes.

Validation: diff hygiene passed. CPU probes used Python 3.12 and PyAV 18.1.0. Targeted pytest collection was blocked by the available vLLM 0.28 environment versus this head's 0.29 requirement. GPU validation was not run.

NancyFyong:
@hsliuustc0106 Both findings from your review are fixed in `50e0afd1` and are still present at the current head `d56b3f52`. Both WebM mux paths now resample Opus-unsupported rates such as 44.1 kHz to 48 kHz, with real PyAV coverage for the array and iterator paths. URL publication now retains and shields the save task, waits for it before rollback on cancellation, and has a delayed-writer regression that verifies no artifact is left behind. I reran those regressions on the current head and all 3 passed. Could you please take another look when convenient?

hsliuustc0106:
Re-reviewed head `2d76c422827b8031810623ac381747661c116266`. Both previous findings are fixed. One new integration finding:

**[P2] Apply resolved encoder settings to worker pre-encoding** — [serving_video.py:469](https://github.com/vllm-project/vllm-omni/blob/2d76c422827b8031810623ac381747661c116266/vllm_omni/entrypoints/openai/serving_video.py#L469). With `preencode_mp4=true`, only request-local codec options reach the worker. Deployment options are ignored, and MiniMax-H3 still encodes H.264 regardless of the selected codec. A CPU API-boundary probe confirmed that deployment `video_codec_options={"crf":"0"}` resolves correctly but disappears from worker arguments, so the worker uses its defaults and silently changes output quality. Please forward the resolved policy and reject unsupported codec combinations before generation.

Validation: 219 targeted tests passed under vLLM 0.29, including the WebM resampling and delayed-writer cancellation regressions. Diff hygiene passed. GPU validation was not run.

hsliuustc0106:
Could you add performance evidence that isolates the benefit of #6477? The current sink benchmark establishes JSON size and consumer RSS, but does not measure delivery latency or exercise HTTP. The startup tests establish readiness rather than performance.

Please cover:

1. **Base vs PR, default MP4 bytes:** check that the existing `/v1/videos/sync` path does not regress.
2. **Bytes, base64, URL, and shared memory on the PR:** use the same fixed decoded frames through the real HTTP route. Measure from frames ready on the server until the client has usable frames, including artifact download for URL mode and decode/map plus reading the frames for the other modes. Report response latency separately so a fast URL/SHM handle response is not mistaken for completed delivery.
3. **One representative real-model workload:** report request-to-usable-output latency to show whether the isolated delivery savings matter end to end. Keep device postprocessing and `preencode_mp4` identical across comparisons so benefits from #6615 or worker pre-encoding are not attributed to #6477.

Please include the timing breakdown (encoding, publication/storage, transfer, client decode/map), peak server/client RSS, and correctness/cleanup checks. Use a compressible clip as well as the current random-noise input; compression affects the comparison substantially. Preserve exact base/head SHAs, environment, hardware/CPU affinity, dimensions, FPS, codec/options, and commands. Separate preparation and warm-up, then report at least three measured repetitions with variability.

This need not establish a throughput claim, but we should quantify the delivery benefit and default-path regression risk before merging. If the measured benefit is primarily memory or payload size rather than latency, please state that explicitly.

NancyFyong:
> Could you add performance evidence that isolates the benefit of #6477? The current sink benchmark establishes JSON size and consumer RSS, but does not measure delivery latency or exercise HTTP. The startup tests establish readiness rather than performance.
> 
> Please cover:
> 
> 1. **Base vs PR, default MP4 bytes:** check that the existing `/v1/videos/sync` path does not regress.
> 2. **Bytes, base64, URL, and shared memory on the PR:** use the same fixed decoded frames through the real HTTP route. Measure from frames ready on the server until the client has usable frames, including artifact download for URL mode and decode/map plus reading the frames for the other modes. Report response latency separately so a fast URL/SHM handle response is not mistaken for completed delivery.
> 3. **One representative real-model workload:** report request-to-usable-output latency to show whether the isolated delivery savings matter end to end. Keep device postprocessing and `preencode_mp4` identical across comparisons so benefits from [[Core][Diffusion] Add a typed pre-D2H video media contract #6615](https://github.com/vllm-project/vllm-omni/pull/6615) or worker pre-encoding are not attributed to [[Core][Diffusion] Add bounded video output transports #6477](https://github.com/vllm-project/vllm-omni/pull/6477).
> 
> Please include the timing breakdown (encoding, publication/storage, transfer, client decode/map), peak server/client RSS, and correctness/cleanup checks. Use a compressible clip as well as the current random-noise input; compression affects the comparison substantially. Preserve exact base/head SHAs, environment, hardware/CPU affinity, dimensions, FPS, codec/options, and commands. Separate preparation and warm-up, then report at least three measured repetitions with variability.
> 
> This need not establish a throughput claim, but we should quantify the delivery benefit and default-path regression risk before merging. If the measured benefit is primarily memory or payload size rather than latency, please state that explicitly.

sure，I'll add it then.

vllm_omni/entrypoints/openai/serving_video.py
```diff
@@ -414,8 +466,10 @@ async def _run_and_extract(
                     normalize_preencode_batch_frames(request.extra_params["preencode_batch_frames"])
                 except ValueError as exc:
                     raise HTTPException(status_code=HTTPStatus.BAD_REQUEST.value, detail=str(exc)) from exc
-            # Merge extra_params into extra_args
-            gen_params.extra_args.update(request.extra_params)
+            model_extra_params = without_video_output_overrides(request.extra_params)
+            if request.extra_params.get("preencode_mp4") and "video_codec_options" in request.extra_params:
+                model_extra_params["video_codec_options"] = request.extra_params["video_codec_options"]
```
SamitHuang: `without_video_output_overrides` strips `video_codec` and `video_codec_options`, then only request-local `video_codec_options` are copied back when `preencode_mp4` is set. `_resolve_video_output_settings` already has the merged deployment+request policy (`settings.codec`, `settings.codec_options`) used by API-side muxers, but those values never reach worker extra_args. With `preencode_mp4=true` the worker therefore ignores deployment `video_codec_options` (for example `crf`) and MiniMax-H3 still encodes H.264 regardless of the selected codec. Forward the resolved codec and options after stripping transport keys, reject incompatible combinations before generation, and replace `test_preencode_forwards_codec_options_but_not_other_output_settings` so it asserts worker extra_args match `settings` rather than locking in this drop.

tests/entrypoints/openai_api/test_video_server.py
```diff
@@ -188,6 +215,89 @@ def test_preencoded_video_bytes_preserve_metadata(mocker: MockerFixture):
...
+            "output_format": "mp4",
+        },
+    )
+    try:
+        asyncio.run(handler.generate_video_bytes(request, "preencoded-options"))
+        captured = engine.captured_sampling_params_list[0].extra_args
+        assert captured["preencode_mp4"] is True
+        assert captured["preencode_batch_frames"] == 5
+        assert captured["video_codec_options"] == {"preset": "ultrafast"}
+        assert "video_codec" not in captured
```
SamitHuang: This test currently requires `video_codec` to be absent from worker extra_args and only checks request-local `video_codec_options`. That encodes the silent policy split instead of the intended contract. Add a case where deployment `VideoOutputTransportConfig(video_codec_options={"crf": "0"})` (and a non-default `video_codec`) is the only source of encoder settings, then assert those resolved values appear in `extra_args` when `preencode_mp4` is true.

SamitHuang (changes requested):
Re-reviewed head 747949a320b116c51f6934ba116a23f7d4fbc7d2 against the earlier ownership/MIME/Hop-1 findings. Loopback SHM + TTL sweeper, URL expiration before write, suffix-derived media types, cancellation wait-before-rollback, and async rejection of immediate transports are still in place.

Two merge blockers remain:

1. Worker `preencode_mp4` still does not receive the resolved encoder policy. API-side encoding uses `settings.codec` / `settings.codec_options`; the worker only sees request-local `video_codec_options` when that key is present, so deployment `crf`/`video_codec` and MiniMax-H3 worker H.264 defaults silently diverge. This is the same P2 from the 2d76c42 review and is still true on this merge commit.
2. Isolated HTTP delivery evidence is still missing. The sink benchmark covers JSON size and child RSS, not `/v1/videos/sync` latency, URL download completion, SHM map+read, default bytes regression vs main, or a real-model request-to-usable-output breakdown. The PR description still says no throughput claim; that measurement set was requested before merge and has not landed on this head.

Do not merge until the resolved encoder settings reach the worker (with a regression that would fail today's extra_args capture) and the delivery timings are posted with base/head SHAs.

vllm_omni/entrypoints/openai/serving_video.py
```diff
@@ -478,47 +613,115 @@ async def generate_videos(
...
+                raise HTTPException(
+                    status_code=HTTPStatus.BAD_REQUEST,
+                    detail="transport_mode='shared_memory' does not support audio-bearing video output",
+                )
+            handles: list[VideoSharedMemoryHandle] = []
+            committed = False
+            try:
+                video_data = []
+                for index, video in enumerate(artifacts.videos):
+                    handle = await asyncio.to_thread(
```
lishunyang12: Shield and await SHM publication before cancellation cleanup, just like URL writes.

NancyFyong:
@lishunyang12 @hsliuustc0106 could you take another look at the fixes and updated performance section when you have a chance?

The SHM cancellation fix is in `23c46e36`: publication is awaited before rollback, including repeated cancellation and late publication failures. All eight new regression cases fail on the old head and pass with the fix. The resolved worker codec policy and WebM audio fixes are also included; the broader suite passed 752 tests with one skip.

The description now summarizes the real-HTTP base/PR and transport comparisons, plus the WAN2.2 workload. SHM reduces isolated delivery time, but we are not claiming a stable end-to-end speedup or that default bytes is regression-free—the noise case was 3.3% slower in this sample.

The main conflicts still need a separate sync; this request is to recheck the fixes and evidence, not to merge yet. Thanks!

AI assistance: Pi helped with implementation, tests, measurements and drafting this update.

SamitHuang:
Pls resolve the conflicts for further review, thx

vllm_omni/entrypoints/openai/serving_video_output_stream.py
```diff
@@ -460,16 +462,20 @@ async def _stream_video_bytes(
         progress: _SessionProgress,
     ) -> AsyncGenerator[tuple[bytes, dict[str, Any]], None]:
         """Yield encoded video bytes from diffusion streaming outputs."""
+        settings = resolve_video_output_settings(
+            self._engine_client,
+            request.extra_params,
+            low_latency=True,
+            force_output_format="mp4",
```
SamitHuang, line 469: `force_output_format="mp4"` only replaces the container. `resolve_video_output_settings` still applies deployment or `extra_params` `video_codec`, so a WebM override such as `libvpx-vp9` is then rejected as incompatible with MP4. The user guide says the fragmented stream stays MP4-only and independent of the artifact container; this path does not implement that. Ignore request/deployment WebM codecs when forcing MP4 (or map them to the MP4 default) and add a regression that the extra_params from `test_request_output_overrides_do_not_reach_model_extra_args` still start a streaming session.

SamitHuang (changes requested):
Re-reviewed head 2bfcb827d6914afc7ca0f756da1444a791c3e5f2 against the encoder-forwarding and HTTP-delivery findings.

Those two blockers are addressed. `preencode_mp4` now writes the resolved codec and options into worker extra_args, H3/WAN chunked sessions consume `video_codec`, and the tests assert deployment `crf=0` plus real H.264/HEVC encodes. WebM 44.1 kHz resampling, URL/SHM wait-before-rollback, loopback SHM, and URL expiration gating are still in place. The HTTP measurements no longer claim a stable end-to-end speedup.

One remaining merge issue: the MP4-only WebSocket path still resolves the artifact codec, so a documented WebM `video_codec` override is incompatible with `force_output_format="mp4"`.

vllm_omni/entrypoints/openai/serving_video_output_stream.py
NancyFyong: Thanks for review, done

PR #6481, opened 2026-08-22
[AR-Diffusion] Page in kernel-legal blocks so the default resolution can run

### LINKED ISSUE #6475: [Bug] AR-Diffusion paged KV ties block size to resolution, so the checkpoint's default 832x480 cannot run
### Summary

The AR-Diffusion paged KV cache uses one latent frame as one cache block. That
ties the KV block size to the output resolution, and FlashAttention's paged
kernel requires the block size to be a multiple of 16
(`vllm/v1/attention/backends/flash_attn.py`, `get_supported_kernel_block_sizes()
-> [MultipleOf(16)]`).

The consequence is that **the resolution the LingBot World v2 checkpoint ships
as its default cannot run the realtime path at all**:

```
832x480  ->  (480/16) x (832/16) = 30 x 52 = 1560 tokens per latent frame
1560 / 16 = 97.5                                  -> rejected
```

The failure surfaces as `Paged KV cache block size must be divisible by 16`,
which names neither the resolution nor the frame geometry that produced it, so
it does not point at anything the caller can act on.

### Where the coupling is

`vllm_omni/experimental/ar_diffusion/runner.py` passes the same value as both
the eviction unit and the paging unit:

```python
chunk_size=spec.tokens_per_frame,   # L118 -- eviction unit (a frame)
block_size=spec.tokens_per_frame,   # L132 -- paging unit (a kernel constraint)
```

These are independent concerns. AR-Diffusion evicts by frame; FlashAttention
tiles by 16 tokens. Binding them makes every resolution a kernel-compatibility
question, and the resolutions that do work only work by arithmetic accident.

### Reproduce

```python
# 832x480 -> 1560 tokens per latent frame
ARDiffusionKVCache(config, block_size=1560, ...)   # chunk_size=1560
```

Or end to end, with `--height 480 --width 832` against the AR-Diffusion engine.

### What I tried

I have a partial fix on a branch and would rather hand it over than open a PR,
because it changes something the RFC states publicly (see "Impact" below).

Decoupling the two turned up three places that treat a frame and a block as the
same object:

1. `ARDiffusionPagedForwardContext.num_current_video_blocks` asserted
   `seq_len % block_size == 0`. Once they differ, the count also depends on the
   offset the chunk starts at -- 30 tokens starting at position 30 span three
   16-token blocks, not two.
2. `video_block_table()` used `sink_blocks = spec.sink_chunks` directly, i.e.
   a frame count as a block count. The window has to be expressed in tokens and
   converted, rounding both ends up.
3. `ARDiffusionKVCache.__init__` sized scratch as
   `frames_per_block + declared_scratch_blocks`, again a frame count used as a
   block count.

Decoupling also exposes an aliasing bug that the aligned case hides: when a
chunk is not a whole number of blocks, the block holding the end of the history
also holds the start of the current chunk, so `history_block_ids +
current_video_block_ids` lists it twice. The kernel then reads those tokens
twice and drops the same number from the end, with correct shapes throughout.

### Result

Verified against dense attention over the exact K/V each sequence should see:

| case | result |
| --- | --- |
| 1440 tokens/frame, block = 1440 (today's aligned path) | `max\|err\| = 0.00e+00` |
| 320 tokens/frame, block = 320 (aligned) | `max\|err\| = 0.00e+00` |
| 1440 tokens/frame, block = 16 | `max\|err\| = 0.00e+00` |
| 320 tokens/frame, block = 16 | `max\|err\| = 0.00e+00` |
| **1560 tokens/frame, block = 16 -- committing path** | `max\|err\| = 0.00e+00` |
| **1560 tokens/frame, block = 16 -- scratch path** | **fails, ~4e-02** |

The 160 existing tests in `tests/diffusion/ar_diffusion` still pass, so the
aligned path is unchanged.

### What still blocks it

The committing path is correct because its tokens are written contiguously, so
there are no holes. The scratch path -- used by every non-committing denoise
step, which is four of the five forwards per generated block -- writes the
current chunk into a separate region and assumes the history ended on a block
boundary. With 1560 tokens per frame the history ends 8 slots into its last
block, and the kernel reads those 8 unwritten slots, shifting the whole
sequence.

Fixing it means copying the history's partial tail block (at most
`block_size - 1` tokens) into scratch before writing, which adds a copy and a
new invariant to a hot path. That seemed worth checking with you before doing.

### Impact on the RFC

The capacity formula published in #6227 is written for one-block-per-frame:

```
required_blocks(cap) = branches x (cap x (sink + window) + frames_per_block) + 2
```

Decoupling invalidates it, and the `+ frames_per_block + 2` constant term is
itself an artifact of the coarse granularity -- at 480x768 one block is
1.098 GiB. A finer block size would also reduce the rounding waste that term
compensates for.

### Alternatives

1. **Decouple** (what I tried). Correct in general, but needs the scratch-path
   fix above and changes the capacity formula.
2. **Validate and document.** Require `tokens_per_frame % 16 == 0` in the
   pipeline and reject early with a message naming the resolution and the token
   count. Does not fix 832x480, but stops it failing obscurely.
3. **An attention backend without the block-size constraint**, if one is
   acceptable on the target hardware.

Happy to finish (1) if that is the direction you want; the branch is ready to
hand over. Otherwise (2) is small and I can send it as a PR.

### Environment

vLLM 0.26.0, torch 2.11.0+cu130, 1x RTX PRO 6000 Blackwell (sm_120),
`robbyant/lingbot-world-v2-14b-causal-fast-diffusers`.

linzhenpl07:
Fixes #6475.

## What was wrong

The AR-Diffusion paged KV cache used one latent frame as one cache block, so
`runner.py` passed the same value as both the eviction unit and the paging
unit. These are independent concerns -- AR-Diffusion evicts by frame,
FlashAttention tiles by 16 tokens -- and binding them made every output
resolution a kernel-compatibility question.

The resolutions that work today work by arithmetic accident. The one the
LingBot World v2 checkpoint ships as its default does not:

    832x480 -> (480/16) x (832/16) = 1560 tokens per frame, 1560 % 16 = 8

It failed with `Paged KV cache block size must be divisible by 16`, which
names neither the resolution nor the frame geometry that produced it.

## What changed

**`runner.py`** picks the paging unit separately from the eviction unit. A
frame that is already a legal block stays one, so every resolution that runs
today keeps precisely the paging it has; anything else pages at 16. What counts
as legal is asked of the module that dispatches the kernel,
`supported_kernel_block_sizes()`, which answers `MultipleOf(16)`. FA4's
head-size-256 kernel pages at 128, but it is not reachable from this call path:
vLLM selects it only when the caller passes `supports_fa4_hd256=True`, which the
AR-Diffusion call does not, so that head size falls back to FA2.

**`paged_attention.py`** stops assuming a chunk is a whole number of blocks:

- `num_current_video_blocks` counts from the offset the chunk starts at. Thirty
  tokens starting at position thirty span three 16-token blocks, not two.
- `video_block_table()` places each block by the token positions it holds and
  keeps the blocks that hold a sink token or one of the window's recent
  tokens; the length it returns counts the slots actually written. When a
  chunk is not block-aligned, the block holding the end of the history also
  holds the start of the current chunk, and it is listed once. Listing it
  twice made the kernel read those tokens twice and drop as many from the end,
  with correct shapes throughout. Only the block-table indices of the sink and
  of the recent window are read, straight from the block manager's list without
  copying the table: the table keeps a null entry for every evicted position and
  grows with the session, so reading all of it would cost more on every tick of a
  long session. How the window is converted to blocks was itself
  wrong in an earlier revision of this branch; see *Fixed before merge*.
- `build_block_table()` derives the table width and `max_seq_len` from a block
  count, not a token count. Flooring a token count understated the capacity
  whenever the window was not a whole number of blocks: the width fell back on
  the live block count and began tracking how full the window was -- the shape
  churn the fixed width exists to prevent -- and `max_seq_len` could come out
  below the `kv_len` actually passed, by up to `block_size - 1`.

**The scratch path** is the interesting one. Four of the five forwards per
generated block are non-committing and write the current chunk into a separate
region, which assumed the history ended on a block boundary. It does not: at
1560 tokens per frame the history stops 8 slots into its last block on every
odd chunk, and since the kernel consumes the block table as one contiguous
run, it read those 8 unwritten slots as tokens and shifted everything after.

Rather than copy the history's partial tail block into scratch -- a cross-layer
copy on a hot path -- the chunk now simply begins where the history left off,
spilling its first few tokens into the free tail of the history's own last
block. That block is already this session's (blocks are allocated whole), the
slots hold nothing yet, and the committing forward overwrites the same slots
with clean K/V later, so the committed state is identical either way. No copy,
no new work per forward.

**`manager.py`** converts the capacity formula's terms from frames to blocks.
Each term counted frames and was used directly as a block count, which was the
identity only while the two were the same thing. Converting keeps the budget
the same number of *tokens* it always was.

## Fixed before merge

Separating the paging unit from the frame broke two things further down, both
on this branch only. `main` cannot run 832x480 at all, so neither reached it.

### The attention sink was indexed in frames

`remove_skipped_blocks` kept the first `sink_chunks` entries of the block table,
but `sink_chunks` counts frames and the table is indexed in blocks. The two were
interchangeable while a frame was a block. At 832x480 a frame spans 98 blocks,
so the configured 9-frame sink held 9 blocks -- 144 tokens instead of 14040 --
and eviction began inside frame 0, deleting from the region the sink exists to
protect. The sink end now rounds up to a block and the freed range ends on a
rounded-down boundary, so a straddling block stays protected; both are exact
when a frame is a whole number of blocks.

Measured on one RTX PRO 6000 (sm_120), bf16, 832x480, seed 0, same scene and
prompt, this change alone. Chroma high-frequency energy relative to each run's
own first ten ticks:

| session time | before | after |
| ---: | ---: | ---: |
| 15 s | 1.46x | 0.88x |
| 30 s | 1.94x | 0.97x |
| 45 s | 2.50x | 1.23x |
| 52 s | 4.47x | 1.24x |

This is the size of a regression the branch introduced, not an improvement over
`main`. The unfixed run is visually unusable by 45 s; the fixed run is still
coherent at 56 s, with mild magenta on the ground from 42 s, so long-session
degradation is not removed by this.

### The window read the wrong blocks once it slid

Reported in review on this PR. `video_block_table()` counted blocks off the ends
of the resident list and capped the table width at the window's token count
rounded up once. The sink and the recent window are separate token ranges, and
each can straddle a block edge on its own; when both did, the width was a block
short, the block holding the window's first tokens was the one left out, and the
returned length -- capped by a block count rather than counted from what was
written -- ran past the last written slot.

At 832x480 with the checkpoint's own 9-frame sink and 9-frame window, both
boundaries sit 8 tokens past an edge while their sum, 28080 tokens, is a whole
number of blocks. Once the window filled, every tick needed 1756 blocks against a
width of 1755: 8 or 16 of the window's tokens were never attended, and every
other tick read 8 unwritten slots.

Blocks are now placed by position, as described above. The width is the sink's
blocks plus the recent range's, where the recent range starts a whole number of
chunks in and so sits at most `block_size - gcd(chunk_size, block_size)` tokens
past an edge. That is zero for a chunk that is a whole number of blocks, so every
frame-aligned geometry keeps exactly the width it had; at 832x480 the width is
1756, exactly what the window needs. Whole blocks are still read, so each
boundary can contribute up to `block_size - 1` tokens the window has already let
go of; they are real K/V at their own positions.

### Action tokens after a partly written video block are refused

Action K/V is listed after the video blocks and the kernel reads the table as one
run, so a video run that ends part way through a block made the kernel read that
block's unwritten slots as the first action tokens and never reach the last ones.
On CPU the output came back NaN-poisoned and no longer matched dense attention,
on both the scratch and committing paths. This now raises. A frame that is a
whole number of blocks always ends on an edge, so only geometries that could not
page before this PR are refused.

## Verification

Against vllm 0.29.0, `tests/diffusion/ar_diffusion`,
`tests/diffusion/models/lingbot_world` and `tests/diffusion/models/dreamzero`
pass on this branch (427 passed) and merged into current `main` (470 passed,
3 skipped). Every existing test of the aligned path passes
unchanged, including the fixed-width test: a frame that is already a legal block
still pages as one block, and its window keeps the width it had.

New CPU regression tests, each comparing against dense attention over the
exact K/V the sequence should see:

| case | result |
| --- | --- |
| 1560 tokens/frame, block 16, non-committing path | matches to 1e-5 |
| 24 tokens/frame, block 16 (same remainder of 8), with and without action K/V | matches to 1e-5 |
| ragged window over 6 consecutive ticks | table width constant, `max_seq_len >= kv_len` throughout |
| 832x480 spec | builds a cache, 98 blocks per frame, eviction still by frame |
| sink at 1560 tokens/frame | eviction starts outside the 14040-token sink |
| attention after eviction: 24 tokens/frame with sink/window of (1, 2), (0, 2), (1, 3) chunks and a reset-at-boundary window, scratch and committing forwards, every unwritten slot poisoned | matches to 1e-5 |
| 832x480 with the checkpoint's 9-frame sink and window, 12 ticks | every slot read holds the token its position implies; width constant |
| action tokens after a partly written video block | refused on both paths |
| reading the window after 60 chunks, when the table is over ten times the window | no more block ids leave vLLM's block manager than the window's width |

The first row is the case this issue reported at ~4e-02. It runs on CPU
because the reference is dense attention rather than a kernel, so it is a
normal part of the suite rather than GPU-gated.

Every existing fixture holds `chunk_size == block_size`, so none of them
could reach this path; `chunk_size=24` against 16-token blocks reproduces
the shipped resolution's remainder of 8 at a size that runs quickly.

These tests are not vacuous. Reverting only the scratch-path change fails
exactly the two `commit_current=False` cases and the continuity assertion,
while both `commit_current=True` cases still pass -- the same split between
the committing and scratch paths the issue describes. Reverting the sink
conversion fails the sink test with `eviction starts at token 144, inside the
14040-token sink`. Reverting the window fix fails the three sliding-window
geometries, the 832x480 test and both action-refusal cases; the
reset-at-boundary case passes either way and guards the skipping of blocks a
reset has freed.

Copying the whole block table again fails the last row: 90 block ids read for a
6-block window. The count is taken at vLLM's `get_blocks`, so a copy is caught
even when only the window's entries are indexed afterwards.

The new tests run on the CPU reference path. The FlashAttention kernel reads
the block table the same way, but it is not exercised by them.

## Scope

An earlier revision of this branch also carried cross-session coalescing
primitives (`CoalescedPagedForward`, `batch_paged_metadata` and per-session
scratch regions). Nothing called them -- the AR-Diffusion runner serves one
request at a time -- so they were moved out of this PR. The commit history shows
them added and then removed; the diff against `main` does not include them.

## Behaviour changes

For a frame that is not a whole number of blocks, the block table is wider by
the room each window boundary can need -- 1756 blocks instead of 1755 at
832x480. Frame-aligned geometries keep their width.

Action tokens are refused after a video run that ends part way through a block,
as described above.

`test_budget_reduced_capacity_drives_runner_lru` had its byte figures updated
from 128/192 to 1,808/2,592. `tiny_spec` declares one token per frame, so it
used to page a single token at a time -- a block size no attention kernel
accepts. Paging at 16 makes each page sixteen times larger; the block *counts*
are unchanged, as is what the test is about.

### DIFF
--- a/vllm_omni/experimental/ar_diffusion/kv_cache/paged_attention.py
+++ b/vllm_omni/experimental/ar_diffusion/kv_cache/paged_attention.py
@@ -1,4 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM-Omni project
 """Paged self-attention helpers for AR-Diffusion KV reuse."""
 
 from __future__ import annotations
--- a/vllm_omni/experimental/ar_diffusion/kv_cache/paged_attention.py
+++ b/vllm_omni/experimental/ar_diffusion/kv_cache/paged_attention.py
@@ -7,11 +7,14 @@ from __future__ import annotations
 import itertools
 from collections.abc import Sequence
 from dataclasses import dataclass, field
-from typing import Any, ClassVar, NamedTuple
+from typing import TYPE_CHECKING, Any, ClassVar, NamedTuple
 
 import torch
 
 from vllm_omni.experimental.ar_diffusion.kv_cache.paged import compute_slot_mapping
+
+if TYPE_CHECKING:
+    from vllm.v1.attention.backend import MultipleOf
 
 _LAYER_IDX_TENSORS: dict[int, torch.Tensor] = {}
 
@@ -679,6 +682,29 @@ def _resolve_fa_version(head_size: int) -> int:
             version = 2
         _FA_VERSION_BY_HEAD_SIZE[head_size] = version
     return version
+
+
+def supported_kernel_block_sizes() -> list[int | MultipleOf]:
+    """Block sizes the kernel this module dispatches to will accept.
+
+    Same shape as vLLM's ``AttentionBackend.get_supported_kernel_block_sizes``
+    -- a plain int is that exact size, ``MultipleOf(b)`` is any positive
+    multiple of ``b`` -- but answered here rather than read off a backend,
+    because AR-Diffusion does not go through backend selection. It calls
+    ``flash_attn_varlen_func`` itself, choosing between vLLM's CUDA build and
+    ROCm's AITER a few lines below, and only this module knows which.
+
+    It lives next to that choice so there is one place to change. The
+    constraint is a property of the kernel, not of the card, and the kernels
+    reachable from here agree on 16 today: vLLM's CUDA FlashAttention, ROCm
+    AITER, and upstream ``flash_attn`` all advertise ``MultipleOf(16)``. Other
+    backends in the same tree do not -- ``hpc_attn`` accepts only 64, and
+    FlashInfer advertises pages of 128 or more solely on Blackwell -- so a
+    caller must treat this as data to be queried, never as the number 16.
+    """
+    from vllm.v1.attention.backend import MultipleOf
+
+    return [MultipleOf(16)]
 
 
 def _rocm_flash_attn_varlen_func():
--- a/vllm_omni/experimental/ar_diffusion/kv_cache/paged_attention.py
+++ b/vllm_omni/experimental/ar_diffusion/kv_cache/paged_attention.py
@@ -63,6 +63,10 @@ class ARDiffusionPagedForwardContext:
     seq_len: int
     commit_current: bool
     max_video_tokens: int
+    # Which scratch region this session's uncommitted chunk writes to. Sessions
+    # that may be coalesced into one forward must differ here or they overwrite
+    # each other's current K/V.
+    scratch_slot: int = 0
     current_video_block_ids: list[int] = field(default_factory=list)
     current_video_slot_mapping: torch.Tensor | None = None
     action_scratch_block_ids: list[int] = field(default_factory=list)
@@ -108,7 +112,9 @@ class ARDiffusionPagedForwardContext:
             positions = torch.arange(start, start + self.seq_len, dtype=torch.long)
             self.current_video_slot_mapping = compute_slot_mapping(table, positions, self.block_size).to(device=device)
         else:
-            self.current_video_block_ids = self.kv_cache.scratch_block_ids(self.kv_branch, 0, n_blocks)
+            self.current_video_block_ids = self.kv_cache.scratch_block_ids(
+                self.kv_branch, 0, n_blocks, slot=self.scratch_slot
+            )
             positions = torch.arange(self.seq_len, dtype=torch.long)
             self.current_video_slot_mapping = compute_slot_mapping(
                 self.current_video_block_ids,
@@ -135,6 +141,7 @@ class ARDiffusionPagedForwardContext:
             self.kv_branch,
             scratch_offset,
             action_blocks,
+            slot=self.scratch_slot,
         )
         positions = torch.arange(action_len, dtype=torch.long)
         self.action_slot_mapping = compute_slot_mapping(
--- a/vllm_omni/experimental/ar_diffusion/kv_cache/paged_attention.py
+++ b/vllm_omni/experimental/ar_diffusion/kv_cache/paged_attention.py
@@ -73,6 +73,17 @@ class ARDiffusionPagedForwardContext:
     action_slot_mapping: torch.Tensor | None = None
     query_len: int = 0
     kv_len: int = 0
+    # Tokens already committed when this forward began. Recorded rather than
+    # derived from the block count: the last block of the history is only
+    # partly written when a chunk is not a whole number of blocks, so
+    # blocks * block_size overstates it and would make attention read past
+    # what was written.
+    _history_tokens: int = 0
+    # Scratch blocks the current video tokens occupy. Not the same as
+    # len(current_video_block_ids): when the history ends mid-block, the first
+    # entry there is the history's own tail block, which is managed, not
+    # scratch. Action K/V is placed after this, so it must not count it.
+    _scratch_blocks_used: int = 0
     _allocated_video: bool = False
     _committed: bool = False
     _action_len: int = 0
@@ -89,13 +100,47 @@ class ARDiffusionPagedForwardContext:
         return int(self.kv_cache.block_size)
 
     @property
+    def chunk_size(self) -> int:
+        """Tokens in one chunk -- one latent frame. The eviction unit."""
+        return int(self.kv_cache.spec.chunk_size)
+
+    @property
+    def start_offset(self) -> int:
+        """Where in its first block this forward's tokens begin.
+
+        Non-zero exactly when the committed history does not end on a block
+        boundary, which is the normal case once a frame is not a whole number
+        of blocks: 1560 tokens per frame against 16-token blocks leaves the
+        history 8 slots into its last block on every odd-numbered chunk.
+        """
+        return int(self.adapter.num_computed_tokens) % self.block_size
+
+    @property
+    def max_video_blocks(self) -> int:
+        """Blocks the visible window spans, rounded up.
+
+        Rounding up is what keeps the block table a fixed shape: the window is
+        a token count, and flooring it would understate the capacity whenever
+        it is not a whole number of blocks.
+        """
+        return -(-self.max_video_tokens // self.block_size)
+
+    @property
     def num_current_video_blocks(self) -> int:
-        if self.seq_len % self.block_size != 0:
-            raise AssertionError(
-                "AR-Diffusion paged attention expects frame-aligned seq_len "
-                f"(multiple of block_size={self.block_size}), got {self.seq_len}"
-            )
-        return self.seq_len // self.block_size
+        """Blocks this forward's tokens occupy, counted from where they start.
+
+        A chunk need not be a whole number of blocks, so the count depends on
+        the offset it begins at, not only on its length: thirty tokens
+        starting at position thirty span three sixteen-token blocks, not two.
+        Rounding up leaves the tail of the last block unwritten, which is only
+        safe because nothing reads past ``kv_len`` -- see
+        :meth:`video_block_table`.
+
+        Both paths count the same way. The committing path writes straight
+        after the history; the scratch path is made to line up with it by
+        :meth:`ensure_video_slots`.
+        """
+        return -(-(self.start_offset + self.seq_len) // self.block_size)
 
     def ensure_video_slots(self, device: torch.device) -> None:
         """Allocate/write targets for the current video tokens, once per KV branch."""
@@ -103,6 +148,7 @@ class ARDiffusionPagedForwardContext:
             return
 
         n_blocks = self.num_current_video_blocks
+        self._history_tokens = int(self.adapter.num_computed_tokens)
         if self.commit_current:
             start = int(self.adapter.num_computed_tokens)
             self.kv_cache.allocate_token_slots(self.adapter, self.seq_len)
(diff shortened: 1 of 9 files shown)


linzhenpl07:
Verified end to end on hardware: 832x480 generates.

One session, three chunks, 1x RTX PRO 6000 Blackwell (sm_120), eager, bf16:

| chunk | latency |
| --- | --- |
| 0 | 9.38 s (includes warmup) |
| 1 | 3.33 s |
| 2 | 3.63 s |

Latent shape `[1, 16, 3, 60, 104]` throughout, all finite. That is 1560
tokens per latent frame paging at 16 -- the configuration that previously
could not construct a cache at all.

The whole `tests/diffusion/ar_diffusion` suite is 172 passed, 0 skipped with
a GPU present, which includes the eight kernel tests that skip on CPU.

To be clear about what this does and does not show: RTF at this resolution is
about 4.8, so 832x480 is not a realtime configuration. It is now a *runnable*
one, which is what the issue was about -- the checkpoint's own default no
longer fails with a message that names neither the resolution nor the frame
geometry.

vllm_omni/experimental/ar_diffusion/kv_cache/paged_attention.py
```diff
@@ -143,23 +222,47 @@ def ensure_action_slots(self, action_len: int, device: torch.device) -> None:
...
             visible_video_blocks = all_video_blocks
+            live_tokens = resident_history + self.seq_len
         else:
-            tail_blocks = max_video_blocks - sink_blocks
+            tail_blocks = max(max_video_blocks - sink_blocks, 0)
             visible_video_blocks = all_video_blocks[:sink_blocks]
             if tail_blocks:
                 visible_video_blocks += all_video_blocks[-tail_blocks:]
-        video_len = len(visible_video_blocks) * self.block_size
+            live_tokens = min(resident_history + self.seq_len, len(visible_video_blocks) * self.block_size)
```
david6666666, line 263: After window truncation, this counts the unwritten tail of a partial last page as live KV. With 24-token chunks, 16-token pages and a 5-page window, a 120-token prefix leaves only 72 written slots in the selected pages, but this returns 80. Please subtract the final-page padding and extend the ragged-window test to compare against dense attention after eviction, with unused slots poisoned.

linzhenpl07: Confirmed, and at the shipped geometry it was worse than this example. With your numbers the selected pages held 72 written slots against a returned 80. At 832x480, with the checkpoint's own 9-frame sink and 9-frame window, it fired on every tick once the window had filled: 8 or 16 of the window's tokens were dropped, and every other tick read 8 unwritten slots.

Fixed in d03bbc349. Blocks are now selected by the token positions they hold, read off the full block table after the chunk's own allocation, and the length counts the slots actually written. A partly written block that is not the last one listed raises instead of being read through.

As asked, there is now a dense comparison after eviction with every unwritten slot poisoned: `test_attention_after_eviction_reads_every_kept_token_and_nothing_unwritten`, next to the ragged-window test, on both the scratch and committing forward of every chunk. Its three sink/window geometries fail on the previous code; its reset-at-boundary case passes on both. The same contiguity requirement also covered action tokens placed after a partly written video block, which are now refused rather than read through.

vllm_omni/experimental/ar_diffusion/kv_cache/paged_attention.py
```diff
@@ -143,23 +222,47 @@ def ensure_action_slots(self, action_len: int, device: torch.device) -> None:
...
-        max_video_blocks = self.max_video_tokens // self.block_size
-        sink_blocks = int(self.kv_cache.spec.sink_chunks)
+        # Order-preserving union, not concatenation. When a chunk is not a
+        # whole number of blocks, the block holding the end of the history also
+        # holds the start of this chunk, so it appears in both lists. Reading
+        # it twice would feed those tokens to attention twice and drop the same
+        # number from the end -- silently, since the shapes stay right.
+        all_video_blocks = list(dict.fromkeys(self.history_block_ids + self.current_video_block_ids))
+        max_video_blocks = self.max_video_blocks
+        sink_blocks = -(-(int(self.kv_cache.spec.sink_chunks) * self.chunk_size) // self.block_size)
```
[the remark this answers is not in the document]
linzhenpl07: Confirmed; this is the same root cause as the length issue on line 263. The sink and the recent window each convert to blocks on their own, so rounding their summed token count once is a block short whenever both straddle an edge. With 24-token chunks, 16-token blocks, a 1-chunk sink and a 2-chunk window, the fifth chunk selected absolute blocks 0, 1, 5, 6 and 7 and dropped block 4, which holds tokens 72-79.

Fixed in d03bbc349. The width is now the sink's blocks plus the recent range's, with the recent range allowed its worst start offset of `block_size - gcd(chunk_size, block_size)`. That is zero for a chunk that is a whole number of blocks, so frame-aligned geometries keep their width, and at 832x480 it gives 1756 blocks, exactly what the window needs.

The stale tokens in the sink-straddling block are intended: whole blocks are read, so each boundary can contribute up to `block_size - 1` tokens the window has already let go of. They are real K/V at their own positions.

PR #6484, opened 2026-08-22
[Config] Remove diffusion_batch_size and use max_num_seqs instead.

LyxWxj:
Remove the redundant `diffusion_batch_size` plumbing and use `max_num_seqs` as the single diffusion batching configuration.

## Purpose

`diffusion_batch_size` duplicated the existing `max_num_seqs` configuration and could overwrite a user-provided value during Stage initialization. For example, `--max_num_seqs 4` could be reset to `1` because `diffusion_batch_size` defaulted to `1`.

This change:

- Removes `diffusion_batch_size` from `OmniBase`, `AsyncOmniEngine`, `StageRuntime`, Stage startup helpers, and diffusion clients.
- Preserves `max_num_seqs` through the existing stage configuration pipeline.
- Ensures `OmniDiffusionConfig.max_num_seqs` is consumed directly by the diffusion engine and request batch scheduler.
- Updates affected tests and call sites to use `max_num_seqs`.

## Test Plan

Targeted regression tests:

```bash
python -m pytest -q \
  tests/engine/test_async_omni_engine_stage_init.py::test_initialize_diffusion_stage_preserves_configured_max_num_seqs \
  tests/engine/test_async_omni_engine_stage_init.py::test_launch_diffusion_stage_replica_preserves_configured_max_num_seqs \
  tests/engine/test_single_stage_mode.py::TestSingleStageReplicaInitialization::test_initialize_local_diffusion_replica_registers_with_master

python -m pytest -q \
  tests/diffusion/test_inline_stage_diffusion_client.py::test_inline_client_requires_replica_id \
  tests/diffusion/test_multiproc_engine_concurrency.py::TestStageDiffusionClientErrorPropagation::test_initialize_client_requires_replica_id \
tests/diffusion/test_prompt_update.py::TestPromptUpdateExecution::test_runner_interaction_delegates_to_helios_pipeline
```

Additional checks:

```bash
python -m compileall -q vllm_omni tests
git diff --check
rg -n "diffusion_batch_size" vllm_omni tests --glob "*.py"
```

**vLLM Version:** `v0.26.0`

**vLLM-Omni Commit:** `77574f35c7c0716d3f309053cee579cda33d6006`

## Test Result

- Targeted regression tests: `6 passed`
- `compileall`: passed
- `git diff --check`: passed
- No remaining `diffusion_batch_size` references under `vllm_omni/` or `tests/`


LyxWxj:
cc@yJader

vllm_omni/entrypoints/omni_base.py
```diff
@@ -177,7 +177,6 @@ def __init__(
         # override the deploy YAML's ``async_chunk: true`` default.
         async_chunk = kwargs.get("async_chunk")
         output_modalities = kwargs.pop("output_modalities", None)
-        diffusion_batch_size: int = kwargs.pop("diffusion_batch_size", 1)
 
         if "log_requests" in kwargs:
```
hsliuustc0106, line 184: Now that `diffusion_batch_size` is no longer popped here, callers still passing it get a silent no-op: the kwarg flows through `**kwargs` into `_create_default_diffusion_stage_cfg`, which copies every kwarg into `engine_args` (config_factory.py), and `OmniDiffusionConfig.from_kwargs` drops it as an unknown field — so `diffusion_batch_size=8` silently reverts to `max_num_seqs=1` with no error.

Following the `log_requests` precedent two lines down, a guided `TypeError` ("`diffusion_batch_size` has been removed in Omni/AsyncOmni. Use `max_num_seqs`.") would make old scripts fail loudly instead of silently losing batching.

Side effect worth a conscious accept (not a change request): stage configs with `max_num_seqs > 1` on serial pipelines now fail fast at boot (`diffusion_engine.py` request-batch guard) where the old overwrite silently clamped them to 1 — correct, but it can break previously-booting configs.

tests/diffusion/test_diffusion_streaming_output.py
```diff
@@ -526,6 +525,7 @@ def __exit__(self, exc_type, exc, tb):
         runner = object.__new__(DiffusionModelRunner)
         runner.vllm_config = _make_vllm_config()
         runner.od_config = SimpleNamespace(
+            max_num_seqs = 1,
```
hsliuustc0106: The spaces around `=` here (`max_num_seqs = 1`) will fail the `ruff-format` pre-commit hook, which rewrites it to `max_num_seqs=1` (`ruff-check` E251 will not catch it — that rule is preview-gated). Worth bundling with the DCO sign-off push.

vllm_omni/entrypoints/omni_base.py
LyxWxj: Thanks for catching this. You are right that removing the `kwargs.pop("diffusion_batch_size", 1)` line can make legacy callers silently fall back to `max_num_seqs=1`, because the unknown kwarg is eventually dropped during diffusion config construction.

I will add an explicit compatibility guard in `OmniBase`, consistent with the existing `log_requests` handling:

`diffusion_batch_size has been removed in Omni/AsyncOmni. Use max_num_seqs instead.`

I will also add a regression test to ensure legacy usage fails loudly.

The stage-config behavior change is intentional: configured `max_num_seqs > 1` will now be honored, and serial pipelines that do not support request batching will fail fast rather than being silently clamped to 1.

tests/diffusion/test_diffusion_streaming_output.py
LyxWxj: Thanks for catching this. Fixed in the latest commit and bundled with the DCO sign-off update. The assignment is now formatted as `max_num_seqs=1` and passes the `ruff-format` requirement.

vllm_omni/entrypoints/omni_base.py
hsliuustc0106: Verified at 6c16c2c10: the `TypeError` guard fires before the kwarg reaches the stage-config path, and `test_removed_diffusion_batch_size_fails_loudly` pins it. Resolved.

tests/diffusion/test_diffusion_streaming_output.py
hsliuustc0106: Confirmed fixed at 6c16c2c10 — assignment is now `max_num_seqs=1`, and the pre-commit (ruff-format) lane is green on this head. Resolved.

hsliuustc0106:
P3 (not in the diff, so no inline anchor): `vllm_omni/deploy/wan2_2_ti2v.yaml:4` still carries `# Request-level batching uses scheduler max_num_seqs, not diffusion_batch_size.` — the clause names the removed knob, so the "no remaining `diffusion_batch_size` references under `vllm_omni/`" claim in the PR body doesn't quite hold. Suggest dropping the stale clause while this PR is open.

hsliuustc0106 (commented):
Re-review at 6c16c2c10: both round-1 threads are fixed and verified — the guided `TypeError` guard for legacy `diffusion_batch_size` callers (with `test_removed_diffusion_batch_size_fails_loudly`) and the ruff-format fix. One remaining P3, posted above: `vllm_omni/deploy/wan2_2_ti2v.yaml:4` still names the removed knob — a one-line comment cleanup. Otherwise LGTM.

LyxWxj:
> Re-review at [6c16c2c](https://github.com/vllm-project/vllm-omni/commit/6c16c2c1087ae5bcc09e65c8a38eef83fc767ce1): both round-1 threads are fixed and verified — the guided `TypeError` guard for legacy `diffusion_batch_size` callers (with `test_removed_diffusion_batch_size_fails_loudly`) and the ruff-format fix. One remaining P3, posted above: `vllm_omni/deploy/wan2_2_ti2v.yaml:4` still names the removed knob — a one-line comment cleanup. Otherwise LGTM.

Thanks for the re-review and for catching the stale comment. I removed the outdated diffusion_batch_size reference from vllm_omni/deploy/wan2_2_ti2v.yaml. The YAML configuration itself is unchanged. This fix is included in the latest commit.

hsliuustc0106 (approved):
lgtm

PR #6487, opened 2026-08-22
[Bugfix][Frontend] Return 503 for speech OOM

z2z23n0:
### What broke

When speech preprocessing raises `torch.OutOfMemoryError`, `OmniOpenAIServingSpeech.create_speech` falls through the generic exception handler. `create_error_response` defaults to HTTP 400, so a retryable server resource failure is rendered as a client-side bad request.

Unexpected non-`ValueError` speech failures take the same path and are also rendered as HTTP 400.

### Reproduction

Run the deterministic regression test, which injects the same exception raised by CUDA-backed reference audio encoding:

```bash
pytest -q \
  tests/entrypoints/openai_api/test_serving_speech.py::TestTTSMethods::test_create_speech_cuda_oom_returns_service_unavailable
```

Before this fix, the resulting error code is 400. With this fix, it is 503.

### Root cause

The generic speech exception handler created an `ErrorResponse` without an explicit error type or status. The API response adapter therefore used the default bad-request status.

### Fix

- Render `torch.OutOfMemoryError` as `ServiceUnavailableError` with HTTP 503.
- Render other unexpected speech failures as `InternalServerError` with HTTP 500.
- Keep `ValueError` as HTTP 400 and keep engine errors on the existing global handler path.
- Document HTTP 503 in the speech route OpenAPI response map.
- Add CPU L1 coverage for handler classification and HTTP response conversion.

## Test Plan

```bash
pytest -q \
  tests/entrypoints/openai_api/test_serving_speech.py::TestTTSMethods::test_create_speech_cuda_oom_returns_service_unavailable \
  tests/entrypoints/openai_api/test_serving_speech.py::TestTTSMethods::test_create_speech_unexpected_failure_returns_internal_server_error \
  tests/entrypoints/openai_api/test_serving_speech.py::test_api_server_create_speech_wraps_error_response_status
```

**vLLM Version:** `0.25.0` (incident GPU runtime, CUDA 13.0)

**vLLM-Omni Commit:** `0fc4496a` (incident runtime); patch based on `f09bdcbc7dc94827e76d4e9479674639be6ac442`

## Test Result

- Passed Python syntax compilation for all three changed files.
- Passed `git diff origin/main...HEAD --check`.
- Passed Ruff, formatting, typos, test CI marks, TTS adapter ratchet, SPDX headers, forbidden imports, and direct `torch.cuda` checks.


z2z23n0:
Self-reviewed the error handling and tests. Input errors still return 400, OOM returns 503, and unexpected failures return 500. I also checked that the API layer preserves the status code.

@vllm-omni-review-bot please take a look.

gcanlin (commented):
I think we should add the general error code 500 like

```
return self.create_error_response(
      f"Speech generation failed: {e}",
      err_type="InternalServerError",
      status_code=HTTPStatus.INTERNAL_SERVER_ERROR,
  )
```

z2z23n0:
> I think we should add the general error code 500 like
> 
> ```
> return self.create_error_response(
>       f"Speech generation failed: {e}",
>       err_type="InternalServerError",
>       status_code=HTTPStatus.INTERNAL_SERVER_ERROR,
>   )
> ```

@gcanlin  Thanks for reviewing! This PR already includes that change in the [generic exception handler](https://github.com/vllm-project/vllm-omni/blob/789b33f24bd1c6a6dbe8a322adb90c5d764319df/vllm_omni/entrypoints/openai/serving_speech.py#L3797-L3810): unexpected exceptions now return InternalServerError with HTTP 500. I also added a regression test that raises a RuntimeError and checks both the status code and error type.

gcanlin:
> > I think we should add the general error code 500 like
> > ```
> > return self.create_error_response(
> >       f"Speech generation failed: {e}",
> >       err_type="InternalServerError",
> >       status_code=HTTPStatus.INTERNAL_SERVER_ERROR,
> >   )
> > ```
> 
> @gcanlin Thanks for reviewing! This PR already includes that change in the [generic exception handler](https://github.com/vllm-project/vllm-omni/blob/789b33f24bd1c6a6dbe8a322adb90c5d764319df/vllm_omni/entrypoints/openai/serving_speech.py#L3797-L3810): unexpected exceptions now return InternalServerError with HTTP 500. I also added a regression test that raises a RuntimeError and checks both the status code and error type.

Sorry for missing. I mean could we only use 500 rather than adding 503? I'm not sure whether 503 is suitable for OOM. And vLLM also doesn't catch OOM to raise 503.

z2z23n0:
@gcanlin Thanks for clarifying. Updated in 57460ce63 to use the generic `InternalServerError` / HTTP 500 handler for OOM as well, and removed the 503 declaration.

The original issue was a CUDA OOM during reference-audio encoding being returned as HTTP 400. Using 500 fixes that misclassification without assuming the OOM is temporary. The OOM regression test now expects 500, and the API conversion test covers 400 and 500.

gcanlin (approved):
LGTM, thanks!

PR #6489, opened 2026-08-22
[Feature] Add Text-to-audio one-stage pipline for LTX 2.5

nagisa-kunhah:
## Purpose

Add the LTX-2.5 text-to-audio pipeline to vLLM-Omni.

This change provides the pipeline implementation, audio conditioning and
connector path, Diffusers checkpoint loading, model registration, the offline
and OpenAI-compatible audio generation entry points, and regression coverage.

The pipeline uses the registered `LTX2TextToAudioPipeline` class and produces
the audio-only output expected by `/v1/audio/generate`. The generated output is
stereo, 48 kHz PCM audio with the requested duration.

## Test Plan

**vLLM Version:** `0.29.0` (the version resolved by the checked-out vLLM
environment)

**vLLM-Omni Commit:** `a332a1067d1fd3edb7f5ac6b602f91aaa4b49a85`

Model used for the hardware run:
`/root/autodl-tmp/models/LTX-2.5-Diffusers`

The checkpoint is the LTX-2.5 Diffusers model used by the repository tests. The
run used one NVIDIA RTX PRO 6000 Blackwell Server Edition (96 GB), CUDA 13.0,
PyTorch `2.11.0+cu130`, BF16 weights, and CUDNN attention.

```text
python -m pytest -q -o addopts='' \
  tests/diffusion/models/ltx2/test_ltx2_audio_pipeline.py \
  tests/diffusion/models/ltx2/test_ltx25_pipeline.py \
  tests/diffusion/models/ltx2/test_ltx2_audio_transformer.py

python -m pytest -q -o addopts='' \
  --run-level full_model \
  tests/e2e/offline_inference/test_ltx25_audio.py \
  tests/e2e/online_serving/test_ltx25_audio.py

python -m pytest -q -o addopts='' \
  tests/e2e/accuracy/ltx/test_ltx25_official_similarity.py \
  -k 'connector_weight_index_is_authoritative or unwraps_flattened'

python /root/.codex/skills/run-vllm-omni-autodl/scripts/bench_ltx2_audio_matrix.py \
  --model /root/autodl-tmp/models/LTX-2.5-Diffusers \
  --output-dir /root/autodl-tmp/e2e-pipeline-full-20260912 \
  --mode eager --duration 1 --duration 2 --duration 5 --duration 10 \
  --duration 20 --steps 30 --repeats 2 --seed 20260912 \
  --attention-backend CUDNN_ATTN

python -m ruff check \
  vllm_omni/diffusion/models/ltx2 \
  tests/diffusion/models/ltx2 \
  tests/e2e/accuracy/ltx/test_ltx25_official_similarity.py
python -m ruff format --check \
  vllm_omni/diffusion/models/ltx2 \
  tests/diffusion/models/ltx2 \
  tests/e2e/accuracy/ltx/test_ltx25_official_similarity.py
git diff --check
```

### Pipeline E2E

The saved single-GPU offline generation matrix completed successfully for
30-step eager inference, with two seeds per requested duration:

| Requested audio | Mean latency | Second-run latency | Peak reserved memory | Output |
| ---: | ---: | ---: | ---: | --- |
| 1 s | 2.153 s | 2.049 s | 40,926 MiB | stereo, 48 kHz, 48,480 samples |
| 2 s | 2.140 s | 2.054 s | 40,930 MiB | stereo, 48 kHz, 96,480 samples |
| 5 s | 2.094 s | 2.010 s | 40,930 MiB | stereo, 48 kHz, 240,480 samples |
| 10 s | 2.146 s | 2.108 s | 40,930 MiB | stereo, 48 kHz, 480,480 samples |
| 20 s | 2.531 s | 2.500 s | 40,930 MiB | stereo, 48 kHz, 960,480 samples |

All ten outputs were finite and non-silent. The full result record and FP32
waveform artifacts are stored under
`/root/autodl-tmp/e2e-pipeline-full-20260912/`.

The original single-request smoke result is also retained:

| Case | Result |
| --- | ---: |
| Audio request | 2.0 s requested, 2.01 s generated |
| Latency | 1.015 s (one request, eight steps) |
| Peak memory reported by the runner | 40,928 MiB |
| Audio format | stereo, 48 kHz, 96,480 samples |
| Numerical validity | finite and non-silent |
| Output artifact | `/root/autodl-tmp/e2e-split-pipeline/seed_20260902.npy` |

The real local checkpoint was also served through the OpenAI-compatible API.
The request returned HTTP 200 and produced a valid WAV:

| Check | Result |
| --- | --- |
| Endpoint | `/v1/audio/generate` |
| Request | 2 s, 30 steps, seed 42 |
| Latency | 2.31 s end to end |
| WAV | 48 kHz, stereo, 16-bit PCM |
| Samples | 96,480 |
| Artifact | [e2e-pipeline-online-real.wav](https://github.com/user-attachments/files/32138400/e2e-pipeline-online-real-20260912.wav) |

The registered OpenAI-compatible path is covered by
`tests/e2e/online_serving/test_ltx25_audio.py`; it verifies HTTP success, WAV
header, stereo channel count, 48 kHz sample rate, the expected sample count,
finite samples, and non-empty audio.

The repository online test was executed at `--run-level full_model`, which uses
the real LTX-2.5 checkpoint rather than the core-model tiny-checkpoint fixture.
It passed the same format checks against the actual Diffusers checkpoint.

### Consistency with the official LTX implementation

The repository includes a pinned official-reference harness using Lightricks
LTX-2 commit `7954dcb0d986bdc36ef272564a9789ade07fcc65` and the matching
checkpoint artifacts. With identical prompts, seeds, BF16 settings, guidance
parameters, and PyTorch SDPA attention:

- tokenization, Gemma hidden states, audio projection, connector outputs,
  conditioning tensors, initial latents, and denoising checkpoints were
  bitwise identical;
- decoded waveforms were finite, stereo, and the expected duration;
- waveform differences were limited to the separate official and Diffusers
  audio decoder implementations. For 1 s and 5 s outputs, the measured
  cosine similarity was `>= 0.9999913` and relative L2 error was `<= 0.004437`.

These metrics validate the pipeline implementation and conditioning path. The
decoder-level differences are expected because the two runtimes use different
decoder kernel implementations; they are not used as a claim of bitwise audio
identity.

### Test command results

| Command | Result |
| --- | --- |
| LTX2 pipeline/transformer unit tests | 126 passed |
| Pinned-reference helper tests | 2 passed, 134 deselected |
| Offline real-checkpoint smoke | 1 passed |
| Online real-checkpoint smoke (`--run-level full_model`) | 1 passed |
| Direct real-checkpoint `/v1/audio/generate` smoke | passed |

### DIFF
--- a/tests/diffusion/models/ltx2/test_ltx2_audio_pipeline.py
+++ b/tests/diffusion/models/ltx2/test_ltx2_audio_pipeline.py
@@ -1,0 +1,703 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM-Omni project
+
+"""Unit tests for the unified LTX text-to-audio pipeline."""
+
+import json
+import math
+from types import SimpleNamespace
+
+import pytest
+import torch
+
+from vllm_omni.diffusion.models.interface import SupportAudioOutput
+from vllm_omni.diffusion.models.ltx2.ltx2_audio_runtime import LTXAudioRuntime
+from vllm_omni.diffusion.models.ltx2.ltx2_components import (
+    LTX2_T2A_COMPONENT_PROFILE,
+    LTX23_T2A_COMPONENT_PROFILE,
+    LTX25_T2A_COMPONENT_PROFILE,
+    create_audio_transformer_from_config,
+    get_ltx2_audio_post_process_func,
+    resolve_ltx_component_profile,
+)
+from vllm_omni.diffusion.models.ltx2.ltx2_guidance import LTXGuidancePlan, LTXGuidanceSpec
+from vllm_omni.diffusion.models.ltx2.ltx2_recipes import (
+    LTX2_T2A_RECIPE,
+    LTX23_T2A_RECIPE,
+    LTX25_T2A_RECIPE,
+    resolve_ltx_pipeline_recipe,
+)
+from vllm_omni.diffusion.models.ltx2.ltx2_request import (
+    LTX2AudioResourceLimits,
+    resolve_ltx_audio_num_frames,
+)
+from vllm_omni.diffusion.models.ltx2.pipeline_ltx2_audio import LTX2TextToAudioPipeline
+
+pytestmark = [pytest.mark.core_model, pytest.mark.diffusion, pytest.mark.cpu]
+
+
+@pytest.mark.parametrize(
+    ("version", "profile", "recipe", "transformer_subfolder"),
+    [
+        ("2", LTX2_T2A_COMPONENT_PROFILE, LTX2_T2A_RECIPE, "transformer"),
+        ("2.3", LTX23_T2A_COMPONENT_PROFILE, LTX23_T2A_RECIPE, "transformer"),
+        ("2.5", LTX25_T2A_COMPONENT_PROFILE, LTX25_T2A_RECIPE, "transformer_full"),
+    ],
+)
+def test_ltx_t2a_uses_one_pipeline_with_version_specific_full_profiles(version, profile, recipe, transformer_subfolder):
+    assert resolve_ltx_component_profile("text_to_audio", version) is profile
+    assert resolve_ltx_pipeline_recipe("text_to_audio", version) is recipe
+    assert profile.transformer_subfolder == transformer_subfolder
+    assert profile.dit_modules == ("transformer",)
+    assert profile.vae_modules == ("audio_vae",)
+    assert "vae" not in profile.vae_modules
+    assert not recipe.supports_cache_dit
+    assert recipe.request_guidance.audio.modality_scale == 1.0
+    assert LTXGuidancePlan.build(recipe.request_guidance).names == ("cond", "uncond", "ptb")
+
+
+def test_ltx_t2a_public_contract_is_audio_only():
+    assert SupportAudioOutput in LTX2TextToAudioPipeline.__mro__
+    assert LTX2TextToAudioPipeline.pipeline_kind == "text_to_audio"
+    assert LTX2TextToAudioPipeline.support_audio_output
+    assert not LTX2TextToAudioPipeline.support_image_input
+    assert not hasattr(LTX2TextToAudioPipeline, "support_video_output")
+    assert LTX2TextToAudioPipeline.dummy_run_num_frames == 9
+
+
+def test_ltx_t2a_runs_only_audio_connector_for_per_modality_projection():
+    class AudioConnector:
+        def __init__(self):
+            self.calls = []
+
+        def __call__(self, hidden_states, attention_mask):
+            self.calls.append((hidden_states, attention_mask))
+            return hidden_states + 1, torch.zeros_like(attention_mask)
+
+    audio_connector = AudioConnector()
+    connectors = SimpleNamespace(
+        config=SimpleNamespace(
+            per_modality_projections=True,
+            caption_channels=2,
+            audio_hidden_dim=4,
+        ),
+        audio_text_proj_in=torch.nn.Linear(6, 4, bias=True),
+        audio_connector=audio_connector,
+    )
+    pipe = object.__new__(LTXAudioRuntime)
+    torch.nn.Module.__init__(pipe)
+    object.__setattr__(pipe, "connectors", connectors)
+    prompt_embeds = torch.arange(24, dtype=torch.float32).reshape(2, 2, 2, 3)
+    attention_mask = torch.tensor([[0, 1], [1, 1]])
+
+    video, audio, output_mask = pipe._run_text_connectors(
+        prompt_embeds,
+        attention_mask,
+        padding_side="left",
+    )
+
+    variance = torch.mean(prompt_embeds**2, dim=2, keepdim=True)
+    normalized = prompt_embeds * torch.rsqrt(variance + 1e-6)
+    normalized = normalized.flatten(2, 3)
+    normalized = torch.where(attention_mask.bool().unsqueeze(-1), normalized, torch.zeros_like(normalized))
+    expected_projection = connectors.audio_text_proj_in(normalized * math.sqrt(2))
+    torch.testing.assert_close(audio_connector.calls[0][0], expected_projection)
+    torch.testing.assert_close(video, expected_projection + 1)
+    torch.testing.assert_close(audio, expected_projection + 1)
+    torch.testing.assert_close(output_mask, torch.ones_like(attention_mask))
+
+
+@pytest.mark.parametrize(
+    ("seconds", "frame_rate", "expected"),
+    [
+        (5.0, 24.0, 121),
+        (1.0, 24.0, 25),
+        (0.5, 24.0, 17),
+        (1.0, 25.0, 25),
+        (5.1, 24.0, 129),
+        (5.21, 24.0, 129),
+    ],
+)
+def test_ltx_t2a_duration_quantizes_up_to_legal_video_clock(seconds, frame_rate, expected):
+    assert (
+        resolve_ltx_audio_num_frames(
+            audio_length=seconds,
+            num_frames=None,
+            frame_rate=frame_rate,
+            default_num_frames=121,
+        )
+        == expected
+    )
+    assert (expected - 1) % 8 == 0
+
+
+@pytest.mark.parametrize(("seconds", "frame_rate"), [(0.5, 24.0), (5.1, 24.0)])
+def test_ltx_t2a_audio_length_does_not_resolve_shorter_than_requested(seconds, frame_rate):
+    num_frames = resolve_ltx_audio_num_frames(
+        audio_length=seconds,
+        num_frames=None,
+        frame_rate=frame_rate,
+        default_num_frames=121,
+    )
+
+    assert num_frames / frame_rate >= seconds
+
+
+def test_ltx_t2a_exact_num_frames_overrides_default_duration():
+    assert (
+        resolve_ltx_audio_num_frames(
+            audio_length=None,
+            num_frames=81,
+            frame_rate=24.0,
+            default_num_frames=121,
+        )
+        == 81
+    )
+
+
+def test_ltx_t2a_audio_resource_limits_accept_deployment_overrides():
+    limits = LTX2AudioResourceLimits.from_additional_config(
+        {
+            "ltx2_audio_limits": {
+                "max_duration_seconds": 12.5,
+                "max_latent_frames": 320,
+            }
+        }
+    )
+
+    assert limits.max_duration_seconds == 12.5
+    assert limits.max_latent_frames == 320
+
+
+@pytest.mark.parametrize(
+    ("additional_config", "error"),
+    [
+        ({"ltx2_audio_limits": {"max_duration_seconds": float("inf")}}, "finite and positive"),
+        ({"ltx2_audio_limits": {"max_duration_seconds": 0}}, "finite and positive"),
+        ({"ltx2_audio_limits": {"max_latent_frames": 0}}, "positive integer"),
+        ({"ltx2_audio_limits": {"max_latent_frames": 2.5}}, "positive integer"),
+        ({"ltx2_audio_limits": {"unknown": 1}}, "Unknown"),
+    ],
+)
+def test_ltx_t2a_audio_resource_limits_reject_invalid_config(additional_config, error):
+    with pytest.raises((TypeError, ValueError), match=error):
+        LTX2AudioResourceLimits.from_additional_config(additional_config)
+
+
+@pytest.mark.parametrize(
+    ("extra_args", "num_frames", "frame_rate", "is_dummy", "error"),
+    [
+        ({"audio_length": 5.0, "num_frames": 121}, 1, 24.0, False, "mutually exclusive"),
+        ({"audio_length": 20.11}, 1, 24.0, False, "max_duration_seconds"),
+        ({"audio_length": float("inf")}, 1, 24.0, False, "finite"),
+        ({"num_frames": 4097}, 1, 24.0, False, "duration"),
+        ({}, 9, 1.0, False, "frame_rate=24"),
+        ({}, 4097, 24.0, True, "duration"),
+    ],
+)
+def test_ltx_t2a_production_request_path_rejects_unsafe_shapes(
+    extra_args,
+    num_frames,
+    frame_rate,
+    is_dummy,
+    error,
+):
+    pipe = object.__new__(LTXAudioRuntime)
+    torch.nn.Module.__init__(pipe)
+    pipe.pipeline_recipe = SimpleNamespace(frame_rate=24.0, num_frames=121)
+    pipe._audio_resource_limits = LTX2AudioResourceLimits()
+    pipe._reject_video_options = lambda _sampling: None
+    pipe._resolve_request_inputs = lambda *_args, **_kwargs: pytest.fail("unsafe request reached normalization")
+    sampling = SimpleNamespace(
+        extra_args=extra_args,
+        num_frames=num_frames,
+        resolved_frame_rate=frame_rate,
+    )
+    request = SimpleNamespace(
+        sampling_params_list=[sampling],
+        is_dummy_run=lambda: is_dummy,
+    )
+
+    with pytest.raises(ValueError, match=error):
+        pipe._resolve_audio_request_inputs(request)
+
+
+def test_ltx_t2a_rejects_latent_budget_before_allocation():
+    pipe = object.__new__(LTXAudioRuntime)
+    torch.nn.Module.__init__(pipe)
+    pipe.audio_sampling_rate = 100
+    pipe.audio_hop_length = 1
+    pipe.audio_vae_temporal_compression_ratio = 1
+    pipe._audio_resource_limits = LTX2AudioResourceLimits(max_latent_frames=100)
+    pipe.prepare_audio_latents = lambda *_args, **_kwargs: pytest.fail("latent allocation must not run")
+    inputs = SimpleNamespace(num_frames=25, frame_rate=24.0)
+    prompt_context = SimpleNamespace(batch_size=1)
+
+    with pytest.raises(ValueError, match="latent frames"):
+        pipe._prepare_audio_state(inputs, prompt_context)
+
+
+@pytest.mark.parametrize(
+    ("audio_length", "num_frames", "frame_rate", "error"),
+    [
+        (1.0, 25, 24.0, "mutually exclusive"),
+        (0.0, None, 24.0, "positive"),
+        (1.0, None, 0.0, "frame_rate"),
+        (1.0, None, float("inf"), "finite"),
+        (float("nan"), None, 24.0, "finite"),
+        (None, 25.5, 24.0, "integer"),
+        (None, "25", 24.0, "integer"),
+        (None, True, 24.0, "integer"),
+        (None, float("nan"), 24.0, "integer"),
+        (None, float("inf"), 24.0, "integer"),
+        (None, 24, 24.0, "8 \\* k \\+ 1"),
+    ],
+)
+def test_ltx_t2a_rejects_invalid_duration_inputs(audio_length, num_frames, frame_rate, error):
+    with pytest.raises(ValueError, match=error):
+        resolve_ltx_audio_num_frames(
+            audio_length=audio_length,
+            num_frames=num_frames,
+            frame_rate=frame_rate,
+            default_num_frames=121,
+        )
+
+
+@pytest.mark.parametrize(
+    ("num_frames", "frame_rate", "error"),
+    [
+        (9, 1.0, "frame_rate=24"),
+        (4097, 24.0, "duration"),
+    ],
+)
+def test_ltx_t2a_runtime_rejects_unsafe_resolved_duration(num_frames, frame_rate, error):
+    limits = LTX2AudioResourceLimits()
+
+    with pytest.raises(ValueError, match=error):
+        limits.validate_resolved_duration(
+            num_frames=num_frames,
+            frame_rate=frame_rate,
+            expected_frame_rate=24.0,
+        )
+
+
+def test_ltx_t2a_runtime_rejects_requested_duration_above_limit():
+    with pytest.raises(ValueError, match="max_duration_seconds"):
+        LTX2AudioResourceLimits().validate_requested_duration(20.11)
+
+
+def test_ltx_t2a_duration_limit_below_minimum_clock_rejects_minimum_shape():
+    with pytest.raises(ValueError, match="max_duration_seconds"):
+        LTX2AudioResourceLimits(max_duration_seconds=0.1).validate_resolved_duration(
+            num_frames=9,
+            frame_rate=24.0,
+            expected_frame_rate=24.0,
+        )
+
+
+def test_ltx_t2a_default_resource_limits_accept_twenty_second_boundary():
+    limits = LTX2AudioResourceLimits()
+    num_frames = resolve_ltx_audio_num_frames(
+        audio_length=20.0,
+        num_frames=None,
+        frame_rate=24.0,
+        default_num_frames=121,
+    )
+
+    limits.validate_requested_duration(20.0)
+    limits.validate_resolved_duration(
+        num_frames=num_frames,
+        frame_rate=24.0,
+        expected_frame_rate=24.0,
+    )
+    limits.validate_latent_frames(512)
+    with pytest.raises(ValueError, match="max_duration_seconds"):
+        limits.validate_resolved_duration(
+            num_frames=num_frames + 8,
+            frame_rate=24.0,
+            expected_frame_rate=24.0,
+        )
+
+
+@pytest.mark.parametrize("latent_frames", [0, 513])
+def test_ltx_t2a_runtime_rejects_latent_shape_outside_budget(latent_frames):
+    with pytest.raises(ValueError, match="latent frames"):
+        LTX2AudioResourceLimits().validate_latent_frames(latent_frames)
+
+
+def test_ltx_t2a_checkpoint_metadata_selects_profile(tmp_path, monkeypatch):
+    from vllm_omni.diffusion.models.ltx2 import ltx2_audio_runtime
+
+    (tmp_path / "model_index.json").write_text(json.dumps({"model_version": "2.5"}))
+
+    def stub_components(pipe, od_config):
+        pipe.od_config = od_config
+        pipe.device = torch.device("cpu")
+
+    monkeypatch.setattr(ltx2_audio_runtime, "initialize_audio_pipeline_components", stub_components)
+    monkeypatch.setattr(LTX2TextToAudioPipeline, "setup_diffusion_pipeline_profiler", lambda *_args, **_kwargs: None)
+
+    od_config = SimpleNamespace(
+        model=str(tmp_path),
+        revision=None,
+        parallel_config=SimpleNamespace(tensor_parallel_size=1, sequence_parallel_size=1),
+        cache_backend="none",
+        additional_config=None,
+        enable_diffusion_pipeline_profiler=False,
+    )
+    pipe = LTX2TextToAudioPipeline(od_config=od_config)
+
+    assert pipe.model_version == "2.5"
+    assert pipe.component_profile is LTX25_T2A_COMPONENT_PROFILE
+    assert pipe.pipeline_recipe is LTX25_T2A_RECIPE
+    assert not hasattr(pipe, "audio_graph_runner")
+
+
+@pytest.mark.parametrize("cache_backend", ["cache_dit", "tea_cache"])
+def test_ltx_t2a_foundation_rejects_cache_backends(cache_backend):
+    od_config = SimpleNamespace(
+        parallel_config=SimpleNamespace(tensor_parallel_size=1, sequence_parallel_size=1),
+        cache_backend=cache_backend,
+        additional_config=None,
+    )
+
+    with pytest.raises(ValueError, match="does not support cache_backend"):
+        LTX2TextToAudioPipeline(od_config=od_config)
+
+
+@pytest.mark.parametrize(("tp_size", "sp_size"), [(2, 1), (1, 2)])
+def test_ltx_t2a_foundation_rejects_distributed_execution(tp_size, sp_size):
+    od_config = SimpleNamespace(
+        parallel_config=SimpleNamespace(
+            tensor_parallel_size=tp_size,
+            sequence_parallel_size=sp_size,
+        ),
+    )
+
+    with pytest.raises(ValueError, match="requires tensor_parallel_size=1"):
+        LTX2TextToAudioPipeline(od_config=od_config)
+
+
+def test_ltx_t2a_transformer_factory_projects_full_config_to_audio_only_model():
+    transformer = create_audio_transformer_from_config(
+        {
+            "in_channels": 128,
+            "num_attention_heads": 32,
+            "audio_in_channels": 4,
+            "audio_out_channels": 4,
+            "audio_num_attention_heads": 2,
+            "audio_attention_head_dim": 4,
+            "audio_cross_attention_dim": 8,
+            "caption_channels": 8,
+            "num_layers": 0,
+            "use_prompt_embeddings": False,
+        }
+    )
+
+    assert transformer.config.audio_in_channels == 4
+    assert not hasattr(transformer, "proj_in")
+
+
+def test_ltx_t2a_postprocess_emits_audio_payload_and_checkpoint_sample_rate(tmp_path):
+    vocoder_dir = tmp_path / "vocoder"
+    vocoder_dir.mkdir()
+    (vocoder_dir / "config.json").write_text(json.dumps({"output_sampling_rate": 48000}))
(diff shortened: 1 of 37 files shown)


hsliuustc0106:
@lishunyang12 @mglyn PTAL

hsliuustc0106:
I profiled the current head `d1b76a2f` on 1× NVIDIA L20X (BF16, TP=1, 30 steps). Please address the denoising launch/host overhead before marking this ready.

Nsight shows 45,137 GPU operations per 30-step request (1,504/step). Against the normal 1,334 ms denoising loop, actual CUDA kernels account for ~343 ms; GEMMs are only ~227 ms (17% of wall time). The main opportunity is launch replay/orchestration rather than replacing the cuBLASLt/NVJet GEMMs.

I built an opt-in local reference using regional dynamic compilation plus a worker-owned CUDA Graph around the complete audio Transformer call. It uses per-signature static buffers, separate graph pools, a bounded LRU, safe output cloning, capture fallback, and fail-fast restrictions for unvalidated TP/SP/offload/quantized/LoRA combinations.

| Audio | Baseline | Graph replay | Speedup |
|---|---:|---:|---:|
| 1.01 s | 1.622 ± 0.033 s | 0.639 ± 0.018 s | 2.54× |
| 5.01 s | 1.644 ± 0.031 s | 0.697 ± 0.001 s | 2.36× |
| 10.01 s | 1.699 ± 0.021 s | 0.895 ± 0.003 s | 1.90× |

The integrated 1-second path measured 0.627 ± 0.005 s versus 1.538 ± 0.029 s (2.45×), with +292 MiB peak VRAM (+0.70%) and a bitwise-identical waveform. A cache-size-3 stress run completed 6 captures and 3 evictions; the evicted 1-second signature recaptured with zero waveform error. The affected local suite is 515 passed / 1 skipped.

Could you add an equivalent opt-in, bounded worker-owned graph path (or another launch-reduction design) and include cold-capture latency, steady-state mean ± stddev, peak VRAM, and waveform/CLAP parity? Direct `torch.compile(mode="reduce-overhead")` was not sufficient here: CUDA Graph Trees crossed the warmup/request thread-local ownership boundary, so graph ownership needs to live in the diffusion worker execution path.

nagisa-kunhah:
Hi @hsliuustc0106, thank you for the detailed profiling results and the concrete CUDA Graph requirements. We propose adding an opt-in manual CUDA Graph around the complete LTX2 audio Transformer forward. 

Each diffusion worker will own one graph runner. On a cache miss, the runner will create static buffers, warm up and capture lazily on the request worker thread, then cache the graph in a bounded LRU. On a hit, it will copy current inputs into those buffers, replay the graph, and clone the static output before returning. Different signatures will share a runner-private CUDA graph pool and will never replay concurrently.

The cache key will describe only the input structure and Python control flow that determine whether a graph can be safely reused:

```python
@dataclass(frozen=True)
class LTX2AudioGraphKey:
    hidden_shape: tuple[int, ...]
    context_shape: tuple[int, ...]
    has_audio_attention_mask: bool
    has_perturbation_mask: bool
    stg_blocks: tuple[int, ...] | None
```

Tensor values such as prompt embeddings, latents, timestep, sigma, coordinates, and mask contents will not enter the key. They will be copied into static buffers before replay. Device and dtype are fixed runner invariants and will be validated explicitly before `copy_()`.

Each cached signature will own its graph, address-stable input buffers, and static output buffer:

```python
@dataclass
class LTX2AudioGraphEntry:
    graph: torch.cuda.CUDAGraph
    static_hidden_states: torch.Tensor
    static_context: torch.Tensor
    static_timestep: torch.Tensor
    static_sigma: torch.Tensor
    static_coords: torch.Tensor
    static_attention_mask: torch.Tensor | None
    static_perturbation_mask: torch.Tensor | None
    static_output: torch.Tensor
```

The runner will keep a bounded, configurable cache and one private CUDA graph pool. Capture failures will fall back to eager execution and be remembered per signature, so every denoise step does not retry the same failed capture. Graph entries will be cleared on worker sleep, model reload, and shutdown.

The CUDA Graph path will expose two configuration options:

```text
enable_ltx2_audio_cuda_graph
ltx2_audio_cuda_graph_max_entries
```

I'm woking on it now and may update this pr tomorrow. Thanks!

nagisa-kunhah:
update: 
I reran the LTX-2.5 audio E2E benchmark on one NVIDIA H20 with BF16, TP=1, 30 denoising steps, batch size 1, and `torch.compile` enabled. The measurements below are steady-state request latency (model initialization is excluded).

| Audio duration | PR baseline | Current CUDA Graph | Improvement |
|---:|---:|---:|---:|
| 1 s | 2.007 s | **1.649 s** | **17.8% faster** |
| 5 s | 2.221 s | **2.107 s** | **5.1% faster** |
| 10 s | 2.930 s | **2.847 s** | **2.8% faster** |

For an apples-to-apples comparison against the current compiled eager path, the current CUDA Graph implementation improves latency by:

| Audio duration | Compiled eager | CUDA Graph | Improvement |
|---:|---:|---:|---:|
| 1 s | 2.021 s | **1.649 s** | **18.4%** |
| 5 s | 2.161 s | **2.107 s** | **2.5%** |
| 10 s | 2.899 s | **2.847 s** | **1.8%** |

The graph runner behaved as expected in all cases:

- 2 graphs captured (startup dummy shape and request shape)
- 120/122 calls were cache hits in the final telemetry
- 0 capture failures
- 0 eager fallbacks
- Outputs were finite and had matching waveform statistics against eager

Peak reserved CUDA memory increased by approximately 0.25–0.31 GiB with CUDA Graphs.

Overall, the change is most effective for short (1 s) requests. For 5 s and 10 s requests, the absolute denoising work dominates and the improvement is smaller, but the graph path remains stable and slightly faster than both the PR baseline and the current eager control.

vllm_omni/diffusion/models/ltx2/ltx2_audio_cuda_graph.py
```diff
@@ -0,0 +1,441 @@
...
+    def _capture(self, **inputs: Any) -> LTX2AudioGraphEntry:
+        static_inputs = {
+            name: _static_copy(value) if isinstance(value, torch.Tensor) else value for name, value in inputs.items()
+        }
+        current_stream = torch.cuda.current_stream(self.device)
+        warmup_stream = torch.cuda.Stream(device=self.device)
+        warmup_stream.wait_stream(current_stream)
+        with torch.cuda.stream(warmup_stream), torch.no_grad():
+            for _ in range(3):
+                self._call_transformer(**static_inputs)
```
nagisa-kunhah: Warmup 3 times may be unnecessary.

vllm_omni/diffusion/models/ltx2/ltx2_audio_cuda_graph.py
```diff
@@ -190,6 +223,20 @@ def record_request_stats(self, before: Mapping[str, int]) -> dict[str, int]:
...
+        if self._active_request_generation is not None:
+            raise RuntimeError("LTX2 audio CUDA Graph request scopes cannot be nested")
+        before = self.stats_snapshot()
+        self._request_generation += 1
+        self._active_request_generation = self._request_generation
+        try:
+            yield
+        finally:
+            self.record_request_stats(before)
+            self._active_request_generation = None
```
nagisa-kunhah, line 223: This may block the implementation of continuous batching.

nagisa-kunhah:
Further plan:
- Request batching: Support same-shape request batches and group requests by resolved audio latent length instead of padding different durations together. This should improve throughput and GPU utilization.
- Cache-DiT tuning: Replace the global residual threshold with timestep- and layer-aware caching. Use stricter thresholds for sensitive layers, exclude the STG block from approximate caching, and keep the maximum continuous cached steps small.
- CUDA Graph coverage: Profile the denoising loop and consider capturing guidance and the Euler update together with the Transformer to reduce Python and launch overhead.
- Graph cache policy: Add shape-aware prewarming and tune max_entries based on the service’s common duration and batch-size distribution to reduce recapture and eviction overhead.

tests/model_tests/diffusion/test_alignment.py
```diff
@@ -32,6 +32,9 @@
     "WanDMDPipeline",
     "WanVACEPipeline",
     "LTX2TwoStagePipeline",
+    # Dedicated unit coverage projects full LTX checkpoints onto an audio-only
+    # graph; a tiny checkpoint fixture is not available yet.
+    "LTX2TextToAudioPipeline",
```
hsliuustc0106: Excluding the pipeline here leaves the full-pipeline weight-load path (connectors + audio_vae + vocoder + transformer through `AutoWeightsLoader`) covered only by hardware-gated e2e — that's the riskiest path of a model addition. Could you add a tiny audio-checkpoint fixture in a follow-up so the exclusion can be dropped? The dedicated unit tests cover the transformer-level loader but not the pipeline-level one.

vllm_omni/diffusion/models/ltx2/ltx2_audio_transformer.py
```diff
@@ -0,0 +1,428 @@
...
+                hidden_dtype=audio_hidden_states.dtype,
+            )
+            temb_prompt_audio = temb_prompt_audio.view(batch_size, -1, temb_prompt_audio.shape[-1])
+        else:
+            temb_prompt_audio = None
+        perturbation_kwargs = (attention_kwargs or {}).get("ltx_perturbation_kwargs", {})
+        for index, block in enumerate(self.transformer_blocks):
+            perturbation_mask = perturbation_kwargs.get("audio_self_attention_mask")
+            blocks = perturbation_kwargs.get("audio_self_attention_blocks")
+            routes_perturbation = getattr(block, "_routes_ltx2_audio_perturbation", False)
```
hsliuustc0106: `_routes_ltx2_audio_perturbation` is never set by any block (here or in the video model), so `routes_perturbation` is always False and the `if routes_perturbation:` branch at line 404 is dead — the `index not in blocks` drop below already produces the correct STG routing. The getattr/branch pair can be deleted.

vllm_omni/diffusion/models/ltx2/ltx2_audio_runtime.py
```diff
@@ -0,0 +1,728 @@
...
+    def __init__(self, *, od_config: OmniDiffusionConfig, prefix: str = "") -> None:
+        del prefix
+        parallel_config = getattr(od_config, "parallel_config", None)
+        tensor_parallel_size = int(getattr(parallel_config, "tensor_parallel_size", 1) or 1)
+        sequence_parallel_size = int(getattr(parallel_config, "sequence_parallel_size", 1) or 1)
+        if tensor_parallel_size != 1 or sequence_parallel_size != 1:
+            raise ValueError(
+                f"{type(self).__name__} requires tensor_parallel_size=1 and "
+                "sequence_parallel_size=1. Distributed execution is provided "
+                "by the LTX2 audio TP/SP extension."
```
hsliuustc0106: This message names an "LTX2 audio TP/SP extension" that doesn't exist anywhere in the repo, so users hitting TP>1 read it as a missing dependency. Suggest stating plainly that audio-only T2A is single-GPU (TP=1, SP=1) today.

vllm_omni/entrypoints/openai/protocol/audio.py
```diff
@@ -378,6 +378,16 @@ def validate_input(cls, v):
         default=0.0,
         description="Audio start time in seconds",
     )
+    num_frames: int | None = Field(
```
hsliuustc0106, line 381: `OpenAICreateAudioGenerateRequest` is a plain BaseModel, so extra fields are silently ignored: `audio_cfg_scale`/`audio_stg_scale`/`audio_rescale_scale`/`sigmas` sent to `/v1/audio/generate` are dropped without error even though the pipeline supports them via the chat extra-body path (`LTX_T2A_EXTRA_BODY_PARAMS`). Either forward them into `extra_args` like `num_frames`, or note in the recipe that tuning knobs are offline/chat-only.

hsliuustc0106:
Independent validation of head `7399355e` on 1× NVIDIA L20X (vllm 0.29.0, diffusers 0.40.0, transformers 5.14.1, torch 2.13.0+cu130, eager, model at a local LTX-2.5-Diffusers checkout):

- CPU unit suite: `test_ltx2_audio_pipeline.py` + `test_ltx25_pipeline.py` + `test_ltx2_audio_transformer.py` + `test_diffusers_loader.py` + `test_model_extras.py` + `test_serving_audio_generate.py` + the two CPU similarity tests — **287 passed, 0 failures**.
- Offline e2e `test_ltx25_audio.py` (`--run-level full_model`, real checkpoint): **passed** in 60s — also confirms the pipeline-level weight load against the real checkpoint that the alignment-suite exclusion leaves uncovered.
- Online e2e `test_ltx25_audio.py` (`vllm serve --omni` + `/v1/audio/generate`): **passed** in 78s, valid stereo 48 kHz WAV.

Note for reproducing: the runner venv needed `pytest-mock`, `pytest-asyncio`, and `torchmetrics` installed on top of the usual deps.

tests/model_tests/diffusion/test_alignment.py
nagisa-kunhah: Thanks! Fixed by removing it from EXCLUDED_MODELS and adding a tiny LTX2 text-to-audio checkpoint builder with common offline and online smoke-test coverage for the full pipeline initialization, weight-loading, and generation path.

vllm_omni/diffusion/models/ltx2/ltx2_audio_transformer.py
nagisa-kunhah: Thanks for catching this! Removed the dead getattr/branch pair and kept the direct block-index-based STG routing.

vllm_omni/entrypoints/openai/protocol/audio.py
nagisa-kunhah: Thanks! Added audio_cfg_scale, audio_stg_scale, audio_rescale_scale, and sigmas to /v1/audio/generate and forwarded them through extra_args, with tests covering the request parsing and forwarding behavior.

vllm_omni/diffusion/models/ltx2/ltx2_audio_runtime.py
nagisa-kunhah: Thanks! Updated the error message to state clearly that audio-only T2A currently requires tensor_parallel_size=1 and sequence_parallel_size=1, without referring to a nonexistent TP/SP extension.

hsliuustc0106:
@mglyn PTAL

vllm_omni/entrypoints/openai/protocol/audio.py
```diff
@@ -394,6 +404,22 @@ def validate_input(cls, v):
         le=1000,
         description="Number of inference steps",
     )
+    audio_cfg_scale: float | None = Field(
```
mglyn: Could we keep these LTX-specific controls out of the shared audio-generation request schema and route them through the same model-specific extras mechanism used by the video pipelines? `LTX_T2A_EXTRA_BODY_PARAMS` already declares the T2A contract. A generic `extra_args`/`extra_params` field here, normalized against the registered model extras where appropriate, would avoid making the common protocol depend on LTX. It would also keep aliases and options such as `audio_stg_blocks` available without adding a new public field for every model-specific knob.

vllm_omni/entrypoints/openai/serving_audio_generate.py
```diff
@@ -81,19 +84,33 @@ async def create_audio_generate(
...
+                audio_extra_args.update(
+                    {
+                        "audio_length": audio_length,
+                        "audio_start_in_s": audio_start,
+                        "audio_end_in_s": audio_end_in_s,
+                    }
+                )
+            if request.num_frames is not None:
+                audio_extra_args["num_frames"] = request.num_frames
+            for name in ("audio_cfg_scale", "audio_stg_scale", "audio_rescale_scale", "sigmas"):
```
mglyn: This tuple duplicates part of `LTX_T2A_EXTRA_BODY_PARAMS` and has already drifted from it (`audio_stg_blocks` and the guidance aliases are registered but not forwarded here). Could `/v1/audio/generate` merge a generic request extras mapping into `OmniDiffusionSamplingParams.extra_args`, following the existing video/chat diffusion path, and leave the accepted LTX names in `model_extras/ltx2.py` as the single model-specific declaration?

vllm_omni/diffusion/models/ltx2/ltx2_audio_runtime.py
```diff
@@ -0,0 +1,728 @@
...
+    local_files_only = os.path.exists(model)
+
+    pipeline.weights_sources = [
+        DiffusersPipelineLoader.ComponentSource(
+            model_or_path=model,
+            subfolder=profile.transformer_subfolder,
+            revision=revision,
+            prefix="transformer.",
+            fall_back_to_pt=True,
+            weight_name_patterns=(
```
mglyn, line 113: These production globs also match the unused cross-modal branch (`transformer_blocks.*.audio_to_video_attn.*`). In the pinned LTX-2.5 `transformer_full` checkpoint that is 576 tensors / about 2.26 GiB that the filtered iterator still materializes and hands to `load_weights()` only to be discarded. Please narrow the block patterns to the actual audio-only module tree (and add a regression using the production patterns); the loader unit test currently uses a narrower `audio_ff` pattern, so it does not catch this.

tests/diffusion/models/ltx2/test_ltx2_audio_transformer.py
```diff
@@ -0,0 +1,362 @@
...
+pytestmark = [pytest.mark.core_model, pytest.mark.diffusion, pytest.mark.cpu]
+
+
+@pytest.mark.parametrize("audio_cross_attn_mod", [False, True])
+@pytest.mark.parametrize("use_stg_mask", [False, True])
+def test_ltx_audio_block_matches_full_transformer_audio_branch(audio_cross_attn_mod, use_stg_mask):
+    """The projected block must preserve the full model's audio-only numerics."""
+    torch.manual_seed(0)
+    dim = 8
+    full_block = LTX2VideoTransformerBlock(
```
mglyn, line 53: This test is order-dependent. Running this file in isolation fails in `QKVParallelLinear` because the TP group is not initialized; it only works when another test has left process-global distributed state behind. Please initialize and tear down the model-parallel group in a fixture (or patch the parallel layer for this CPU unit) so this advertised test command is independently reproducible.

tests/diffusion/models/ltx2/test_ltx2_audio_pipeline.py
```diff
@@ -0,0 +1,721 @@
...
+@pytest.mark.parametrize(("tp_size", "sp_size"), [(2, 1), (1, 2)])
+def test_ltx_t2a_foundation_rejects_distributed_execution(tp_size, sp_size):
+    od_config = SimpleNamespace(
+        parallel_config=SimpleNamespace(
+            tensor_parallel_size=tp_size,
+            sequence_parallel_size=sp_size,
+        ),
+    )
+
+    with pytest.raises(ValueError, match="requires tensor_parallel_size=1"):
```
mglyn: Current head fails both parameterizations here: the assertion expects `requires tensor_parallel_size=1`, while the implementation now says `currently supports only tensor_parallel_size=1 ...`. Please update the assertion or restore matching wording; the PR test command is not green at this commit.

vllm_omni/diffusion/models/ltx2/ltx2_request.py
```diff
@@ -54,6 +55,122 @@ def guidance_scale(self) -> float:
 LTXCheckpointKind = Literal["distilled", "regular"]
 
 
+@dataclass(frozen=True)
+class LTX2AudioResourceLimits:
+    """Deployment bounds applied before LTX text-to-audio allocations."""
+
+    max_duration_seconds: float = 20.1
```
mglyn, line 62: `max_duration_seconds` currently is not an accepted request boundary. For example, the default `audio_length=20.1` passes `validate_requested_duration`, rounds up to 489 frames (20.375 s), and is then rejected by `validate_resolved_duration`; similarly, operator-provided limits generally expose a smaller undocumented request maximum. Could we define whether this setting limits requested or grid-aligned duration and make the validation/default/tests consistent with that contract?

vllm_omni/diffusion/models/ltx2/ltx2_audio_runtime.py
```diff
@@ -0,0 +1,728 @@
...
+        pipeline_name=type(pipeline).__name__,
+    )
+    pipeline.tokenizer = AutoTokenizer.from_pretrained(
+        model,
+        subfolder="tokenizer",
+        local_files_only=local_files_only,
+        revision=revision,
+    )
+    if profile.text_encoder_cls is None:
+        raise ImportError("LTX-2.5 requires Gemma4UnifiedForConditionalGeneration; install transformers>=5.10.1,<5.15.")
```
mglyn: This copied dependency hint has already drifted from both `requirements/common.txt` and the shared LTX initializer, which require `transformers>=5.13.0,<5.15`. Please use the same message (ideally a shared helper/constant) so users are not told that 5.10.1 is supported.

vllm_omni/diffusion/models/ltx2/ltx2_audio_runtime.py
```diff
@@ -0,0 +1,728 @@
...
+    with torch.device("cpu"):
+        pipeline.text_encoder = _load_component(
+            profile.text_encoder_cls,
```
mglyn, line 166: All `_load_component()` calls in this audio-only initializer inherit `_load_component`’s default `_LTX_COMPONENT_SUBFOLDERS`. The first prefetch is correctly audio-only, but if `from_pretrained_with_prefetch` enters its cache-healing retry, it will prefetch the full list, including the video VAE and latent upsampler that this runtime is designed to omit. Please pass `_LTX_AUDIO_COMPONENT_SUBFOLDERS` through these calls (or make the profile own its prefetch list) so the fallback path remains audio-only too.

vllm_omni/diffusion/models/ltx2/ltx2_audio_runtime.py
```diff
@@ -0,0 +1,728 @@
...
+        revision=revision,
+    )
+    _install_connector_attention(
+        pipeline.connectors,
+        preserve_learned_register_mask=profile.preserve_connector_attention_mask,
+    )
+    pipeline.audio_vae = _load_component(
+        AutoencoderKLLTX2Audio,
+        model,
+        "audio_vae",
```
mglyn, line 191: The LTX-2.5 connector is still loaded and placed in full even though `_run_text_connectors()` never executes its video side. In the pinned checkpoint the connector weights total about 5.91 GiB: the used `audio_connector` + `audio_text_proj_in` are about 1.47 GiB, while `video_connector` + `video_text_proj_in` leave about 4.44 GiB permanently resident but unreachable from T2A. Could the per-modality LTX-2.5 path construct/load an audio-only connector projection, or drop the video modules before `_place_aux_components()`? This is currently the largest avoidable video component in the purported audio-only graph.

tests/e2e/offline_inference/test_ltx25_audio.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+@pytest.fixture(scope="module", autouse=True)
+def require_ltx25_model_access() -> None:
+    if not os.path.isdir(MODEL):
+        skip_if_gated_repo_inaccessible(MODEL, filename="model_index.json")
+
+
+@hardware_test(res={"cuda": ["H100", "B200"]}, num_cards=1)
+def test_ltx25_text_to_audio_offline_generate() -> None:
```
mglyn, line 38: The checked-in real-checkpoint T2A coverage is only a smoke test (shape/sample rate/finite/non-zero). The PR body reports strong parity against a pinned official implementation, but neither `run_ltx25_reference.py` nor `test_ltx25_official_similarity.py` contains a T2A path, so those central correctness numbers are not reproducible from this change. Because this PR maintains a separate 425-line audio Transformer/runtime that can drift from the joint implementation while still producing valid audio, please add a pinned official T2A comparison with an explicit waveform/conditioning threshold, or at least check in the exact comparison harness used for the reported result.

vllm_omni/diffusion/models/ltx2/ltx2_audio_runtime.py
```diff
@@ -0,0 +1,728 @@
...
+            sigma_scalars = tuple(host_sigmas.tolist())
+            sigmas = host_sigmas.to(self.device)
+        timesteps = self._set_audio_sigmas(audio_scheduler, sigmas)
+        plan = self._guidance_plan
+        audio_coords = self.transformer.audio_rope.prepare_audio_coords(
+            audio_latents.shape[0],
+            padded_num_frames,
+            audio_latents.device,
+        )
+        audio_attention_mask = (
```
mglyn: This padding/mask path is currently unreachable in production: `__init__` rejects every `sequence_parallel_size != 1`, and `prepare_audio_latents()` only pads according to that SP size, so `padded_num_frames == original_num_frames` here. The corresponding test manufactures a padded state by calling `_run_audio_denoise()` directly, rather than exercising a valid request. Unless SP support is part of this PR, please remove the dormant padding/mask/`clear_audio_padding` machinery (including `preserve_sp_padded_audio_duration`) or clearly isolate it as follow-up work; carrying and testing an impossible execution branch makes the new runtime harder to maintain.

vllm_omni/diffusion/models/ltx2/ltx2_components.py
```diff
@@ -228,6 +251,9 @@ class LTXComponentProfile:
 )
 
 _COMPONENT_PROFILES: dict[tuple[str, str], LTXComponentProfile] = {
+    ("text_to_audio", "2"): LTX2_T2A_COMPONENT_PROFILE,
+    ("text_to_audio", "2.3"): LTX23_T2A_COMPONENT_PROFILE,
+    ("text_to_audio", "2.5"): LTX25_T2A_COMPONENT_PROFILE,
```
mglyn, line 256: The PR purpose, support table, recipe, real-checkpoint E2E, and reported official parity all scope this feature to LTX-2.5, but these entries silently enable production T2A for LTX-2 and 2.3 as well (with separate recipes/profiles). The tiny random-checkpoint matrix does not establish that those public combinations are correct. Could we either limit this PR to the advertised 2.5 profile or add/document real-checkpoint validation for 2/2.3? Otherwise this is extra support surface that can fail independently without users even knowing it is claimed.

vllm_omni/diffusion/models/ltx2/ltx2_audio_runtime.py
```diff
@@ -0,0 +1,728 @@
...
+            noise_scale=self.pipeline_recipe.phases[0].noise_scale,
+            dtype=prompt_context.positive_connector_audio_prompt_embeds.dtype,
+            device=self.device,
+            generator=inputs.generator,
+            latents=inputs.audio_latents,
+        )
+        return audio_latents, original_frames, padded_frames, latent_mel_bins
+
+    @staticmethod
+    def _set_audio_sigmas(scheduler, sigmas: torch.Tensor) -> torch.Tensor:
```
mglyn: This duplicates `ltx2_denoise._set_scheduler_sigmas()` exactly (including the private scheduler fields), while this file already imports `_official_ltx_sigmas` from the same module. Please reuse the shared helper (renaming/exporting it if desired) so scheduler-state setup has one implementation to maintain.

vllm_omni/diffusion/models/ltx2/ltx2_conditioning.py
```diff
@@ -341,7 +354,7 @@ def _prepare_prompt_context(
             prompt_embeds = torch.cat([negative_prompt_embeds, prompt_embeds], dim=0)
```
mglyn, line 354: This hook changes the conditioning path shared by the existing LTX-2.5 video pipelines. The PR test plan currently invokes `test_ltx25_official_similarity.py` only with `-k "connector_weight_index_is_authoritative or unwraps_flattened"`, so every actual video similarity case is deselected. Before merge, please run at least `test_ltx25_full_matches_official` for `full_one_stage` on both T2V and I2V and report the existing SSIM/PSNR/audio thresholds, to demonstrate that adding the audio-only override did not regress the established video paths.

mglyn (commented):
Thanks for the substantial T2A integration. I left several non-blocking suggestions around keeping model-specific request options in the existing extras mechanism, trimming unused video weights/components from the audio-only runtime, removing dormant or duplicated paths, making the reported parity reproducible, and running representative existing LTX-2.5 video accuracy cases. Please consider addressing these to reduce regression and maintenance risk.

vllm_omni/entrypoints/openai/protocol/audio.py
nagisa-kunhah: Thank you for pointing this out. The public Audio request now uses a generic `extra_params` field. LTX-specific parameters are declared and validated only by the model extras registry, so the shared protocol no longer needs model-specific fields.

vllm_omni/entrypoints/openai/serving_audio_generate.py
nagisa-kunhah: Thank you for catching this. The serving layer now reads the allowed fields from the registry and forwards them generically. This also covers `audio_stg_blocks` and the guidance aliases, removing the duplicated LTX parameter list.

vllm_omni/diffusion/models/ltx2/ltx2_audio_runtime.py
nagisa-kunhah: Thank you for finding this loading overhead. The audio-only loader patterns are now restricted and no longer match `audio_to_video_attn`. A regression test checks the actual glob behavior so unused video weights are not materialized.

tests/diffusion/models/ltx2/test_ltx2_audio_transformer.py
nagisa-kunhah: Thank you for the feedback. The test now isolates the TP global state and fixes the mask shape and parallel-weight initialization issues. Running `test_ltx2_audio_transformer.py` in isolation now passes: `14 passed`.

tests/e2e/offline_inference/test_ltx25_audio.py
nagisa-kunhah: We agree that a reproducible official-reference test is needed. Manual validation with real checkpoints currently shows LTX-2 cosine `0.99999985` and relative L2 `0.000904`, and LTX-2.3 cosine `0.9999904` and relative L2 `0.004748`. We will package the fixed revisions and thresholds into an automated E2E test.

vllm_omni/diffusion/models/ltx2/ltx2_components.py
nagisa-kunhah: We agree that a reproducible official-reference test is needed. Manual validation with real checkpoints currently shows LTX-2 cosine 0.99999985 and relative L2 0.000904, and LTX-2.3 cosine 0.9999904 and relative L2 0.004748. We will package the fixed revisions and thresholds into an automated E2E test.

vllm_omni/diffusion/models/ltx2/ltx2_audio_runtime.py
nagisa-kunhah: Addressed. The audio runtime now reuses the shared `_set_scheduler_sigmas()` helper and no longer maintains a second copy of the scheduler private-state logic.

vllm_omni/diffusion/models/ltx2/ltx2_audio_runtime.py
```diff
@@ -0,0 +1,719 @@
...
+            profile.vocoder_cls,
+            model,
+            "vocoder",
+            local_files_only=local_files_only,
+            dtype=dtype,
+            revision=revision,
+            prefetch_list=_LTX_AUDIO_COMPONENT_SUBFOLDERS,
+        )
+    except (TypeError, OSError, ValueError):
+        if profile.vocoder_fallback_cls is None or profile.vocoder_fallback_cls is profile.vocoder_cls:
```
lishunyang12, line 208: Extract the duplicated vocoder loading and fallback into a helper shared with the video initializer.

vllm_omni/diffusion/models/ltx2/ltx2_conditioning.py
nagisa-kunhah: Thanks, I ran `test_ltx25_full_matches_official` and the results are as follows:
| Case | SSIM mean | SSIM min | PSNR | Audio relative L2 | Audio cosine | Result |
| --- | ---: | ---: | ---: | ---: | ---: | --- |
| T2V | `0.997813` | `0.997247` | `45.441 dB` | `3.953e-09` | `1.000000000000` | Pass |
| I2V | `0.997928` | `0.997858` | `52.631 dB` | `8.578e-06` | `0.999999999963` | Pass |

However, there's a fix needed here:

https://github.com/vllm-project/vllm-omni/pull/6489/changes/6445624d6cea38b2e78ad4159e6d8fb95abe0d36#diff-8e9f94cf165ab2a337e626cb6f4a954696cdeb5451c8429f8051c8143e145d29R81-R84

Could you please take a look at this fix? Thanks!

vllm_omni/diffusion/models/ltx2/ltx2_audio_runtime.py
nagisa-kunhah: Thanks, I extract to `vllm_omni/diffusion/models/ltx2/ltx2_components.py`.

PR #5908, opened 2026-08-07
[Refactor][Diffusion] Move single-model layers into their model directories

congw729:
## Purpose

Complete a bounded part of JiusiServe/vllm-omni-project-manage#27 by moving model-specific layers to their owning model directories.

- Move `WanS2VRotaryPosEmbed` and `RotaryEmbeddingWanS2V` to `models/wan2_2/rope.py`, and `RMSNormVAE` to `models/wan2_2/norm.py`.
- Keep `RotaryEmbeddingWan` in the shared RoPE module: both Wan and Lingbot-World now use it. `RotaryEmbeddingWanS2V` inherits this shared implementation. Shared RoPE helpers and `RotaryEmbeddingS2VGrid` also retain their existing locations.
- Move the Bagel MoT package, including linear layers, RMSNorm and Triton ops, to `models/bagel/mot/`. Update model/test imports, benchmark and documentation paths, and the existing lint allowlist entry.

Synchronized with main at `6a03e45c340a60553884432ff4fc4a31a6219d55` on 2026-09-20 (PR head `73ca73037d89fe7c6712b1dad734a0b9ceed01d2`). The migration preserves this baseline's operator implementations, including NPU `index_select`, RMSNorm residual handling, vLLM 0.29.0 adaptation and CustomOp platform dispatch. The previous shared S2V indexing change is removed from the PR.

The moved internal import paths change without forwarding shims. Callers importing these internal modules must use their new locations. MoT's package-local tuning directory moves with the package; `VLLM_TUNED_CONFIG_FOLDER` remains available. CLI options, model registration names and operator signatures are unchanged.

## Current value and scope

Reassessed on 2026-09-11. The original scope has been narrowed because `RotaryEmbeddingWan` is now shared. The remaining migration has modest maintenance value: it makes ownership of Wan S2V/VAE and Bagel MoT code clearer while retaining genuinely shared components in `layers/`.

Existing model functionality works on main; this is a normal-priority maintenance cleanup with no performance improvement claim. The scope stays limited to relocation and reference updates.

### Latest refresh (2026-09-20)

Merged main `6a03e45c3` without conflicts. The migration scope is unchanged. This brings in #7426, which updates the realtime VAD test to expect `response.output_audio.*` events.

[...]

PR #3960, opened 2026-05-29
[Enhancement] Introduce High-Performance MoT (Mixture-of-Tokens) Kernels: Triton Implementation

princepride:
## Purpose

This PR is after PR #1328 rebase.

This PR introduces high-performance Mixture-of-Tokens (MoT) core operator primitives for vLLM-Omni, including MoTLinear and MoTRMSNorm. These operators are developed using Triton, following the proposals in RFC https://github.com/vllm-project/vllm-omni/issues/1146 and https://github.com/vllm-project/vllm-omni/issues/936.

## Test Plan

Commands used:

```bash
pytest -s -v \
  tests/e2e/offline_inference/test_bagel.py::test_bagel_text2text \
  tests/e2e/offline_inference/test_bagel.py::test_bagel_img2text \
  --run-level=advanced_model
```

BAGEL image quality and speed were also checked with the offline example:

```bash
CUDA_VISIBLE_DEVICES=0 VLLM_WORKER_MULTIPROC_METHOD=spawn \
PYTHONPATH=/proj-tango-pvc/users/zhipeng.wang/workspace/vllm-omni-pr1328-clean \
python3 examples/offline_inference/bagel/end2end.py \
  --model ByteDance-Seed/BAGEL-7B-MoT \
  --modality text2img \
  --prompts "A cute cat" \
  --steps 50 \
  --seed 42 \
  --output <output_dir>
```

For step 0-50 timing, the diffusion pipeline profiler was enabled during local A/B measurement and `BagelPipeline.bagel.generate_image` timing was used, excluding model loading, AR stage initialization, and VAE decode.

## Test Result

MoT RMSNorm kernel test:

```text
root@job-2f05f13a3e65b678-gvjxv:/proj-tango-pvc/users/zhipeng.wang/workspace/vllm-omni-pr1328-clean# pytest -s -v \
  tests/e2e/offline_inference/test_bagel.py::test_bagel_text2text \
  tests/e2e/offline_inference/test_bagel.py::test_bagel_img2text \
  --run-level=advanced_model
=========================================== test session starts ============================================
platform linux -- Python 3.12.13, pytest-9.0.3, pluggy-1.6.0 -- /usr/bin/python3
cachedir: .pytest_cache
rootdir: /proj-tango-pvc/users/zhipeng.wang/workspace/vllm-omni-pr1328-clean
configfile: pyproject.toml
plugins: anyio-4.13.0
collected 2 items                                                                                          

[...]

PR #6490, opened 2026-08-22
[Bugfix][Kernel] Fix MoT GEMM rejecting a one-sided text/vae bias

### LINKED ISSUE #6488: [Bug]: MoT fused GEMM rejects a one-sided bias, so MoT linear layers crash with their default bias flags
### Your current environment

<details>

```text
vLLM-Omni: main @ f09bdcbc7dc94827e76d4e9479674639be6ac442
vLLM:      v0.26.0
```

The rejection is a host-side assertion in `invoke_mot_gemm` that runs before the
Triton launch, so it does not depend on the accelerator, model, or torch version.
Reproduced on CPU.

</details>

### Your code version

<details>

```text
v0.26.0
```
</details>
<details>

```text
f09bdcbc7dc94827e76d4e9479674639be6ac442 (main)
```
</details>

### 🐛 Describe the bug

`invoke_mot_gemm` requires the text and VAE bias to be either both present or
both absent:

https://github.com/vllm-project/vllm-omni/blob/f09bdcbc7dc94827e76d4e9479674639be6ac442/vllm_omni/diffusion/layers/mot/ops/mot_gemm.py#L836-L839

But the two layers that call it expose `bias` and `vae_bias` as **independent**
constructor flags, and their defaults are exactly the rejected combination —
`bias=True, vae_bias=False`:

https://github.com/vllm-project/vllm-omni/blob/f09bdcbc7dc94827e76d4e9479674639be6ac442/vllm_omni/diffusion/layers/mot/mot_qkv_parallel_linear.py#L48-L49

https://github.com/vllm-project/vllm-omni/blob/f09bdcbc7dc94827e76d4e9479674639be6ac442/vllm_omni/diffusion/layers/mot/mot_row_parallel_linear.py#L42-L43

`vae_bias=False` registers `gen_exp.bias` as `None` while the parent class still
creates `self.bias` from `bias=True`:

https://github.com/vllm-project/vllm-omni/blob/f09bdcbc7dc94827e76d4e9479674639be6ac442/vllm_omni/diffusion/layers/mot/mot_qkv_parallel_linear.py#L113-L121

and both dispatchers hand that pair straight to the kernel:

https://github.com/vllm-project/vllm-omni/blob/f09bdcbc7dc94827e76d4e9479674639be6ac442/vllm_omni/diffusion/layers/mot/mot_qkv_parallel_linear.py#L181-L182

https://github.com/vllm-project/vllm-omni/blob/f09bdcbc7dc94827e76d4e9479674639be6ac442/vllm_omni/diffusion/layers/mot/mot_row_parallel_linear.py#L186-L187

So a MoT linear layer built with the constructor defaults raises
`AssertionError: Bias must be provided for both Text and VAE simultaneously, or
neither.` on its first gen-mode forward, inside the engine core. The mirror case
(`bias=False, vae_bias=True`) fails the same way. Two of the four combinations
the API offers are unusable, including the default one.

**Why it has not been hit yet.** The only in-tree consumer is Bagel, which
happens to pick a symmetric pair for every layer — `bias=True, vae_bias=True`
for `qkv_proj`, `bias=False` (and the `vae_bias=False` default) for `o_proj`:

https://github.com/vllm-project/vllm-omni/blob/f09bdcbc7dc94827e76d4e9479674639be6ac442/vllm_omni/diffusion/models/bagel/bagel_transformer.py#L458-L474

A second consumer that leaves `vae_bias` at its default — e.g. the MoT model
requested in #6471 — hits this immediately, and the failure mode (an assert in
the engine core on the first gen-mode step) gives no hint that the fix is to
flip a constructor flag.

The MoT kernel tests would not catch it either: `tests/diffusion/kernels/mot/`
contains only `gpu`-marked tests, and the one Buildkite step that walks that
directory selects `-m 'core_model and cpu'`, so both files are deselected in
every lane.

**Root cause.** The kernel carries a single `HAS_BIAS` constexpr for both
experts while `cur_bias_ptr` is chosen per expert at runtime:

https://github.com/vllm-project/vllm-omni/blob/f09bdcbc7dc94827e76d4e9479674639be6ac442/vllm_omni/diffusion/layers/mot/ops/mot_gemm.py#L678-L680

With one bias absent, the host substitutes the integer `0` for its pointer:

https://github.com/vllm-project/vllm-omni/blob/f09bdcbc7dc94827e76d4e9479674639be6ac442/vllm_omni/diffusion/layers/mot/ops/mot_gemm.py#L869-L870

and a runtime select between an integer and a pointer is not typeable in Triton,
so the assertion is what keeps that from reaching the compiler. Removing the
assertion alone would turn a clean error into a load from address 0 whenever a
VAE program ran with `HAS_BIAS=True`.

**CPU repro**

<details>
<summary>repro.py</summary>

```python
# The rejection is a host-side assert that runs before the Triton launch, so no
# GPU is needed. `tl` is stubbed only so the dtype mapping ahead of the assert
# can run on a machine without Triton; on a CUDA box, drop _stub_tl().
import types

import torch

from vllm_omni.diffusion.layers.mot.ops import mot_gemm

CONFIG = {"BLOCK_SIZE_M": 16, "BLOCK_SIZE_N": 32, "BLOCK_SIZE_K": 16,
          "GROUP_SIZE_M": 8, "num_warps": 4, "num_stages": 2}


class Launched(Exception):
    """Stands in for the Triton launch: reaching it means the host accepted the args."""


def _stub_tl():
    if getattr(mot_gemm.tl, "float32", None) is None:
        mot_gemm.tl = types.SimpleNamespace(
            float32="f32", float16="f16", bfloat16="bf16", int32="i32", int8="i8",
            float8e4nv="f8e4nv", float8e5="f8e5", float8e4m3fn="f8e4m3fn")

    class Fake:
        def __getitem__(self, grid):
            def launch(*a, **kw):
                raise Launched(kw)
            return launch

    mot_gemm.mot_unified_gemm_kernel = Fake()


_stub_tl()

M, K, N = 8, 16, 32
A = torch.randn(M, K, dtype=torch.bfloat16)
B_text = torch.randn(K, N, dtype=torch.bfloat16)
B_vae = torch.randn(K, N, dtype=torch.bfloat16)
C = torch.zeros(M, N, dtype=torch.bfloat16)
bias = torch.randn(N, dtype=torch.bfloat16)
text_indices = torch.arange(0, 4, dtype=torch.int32)
vae_indices = torch.arange(4, 8, dtype=torch.int32)


def run(bias_text, bias_vae, label):
    try:
        mot_gemm.invoke_mot_gemm(
            A=A, B_text=B_text, B_vae=B_vae, C=C,
            bias_text=bias_text, bias_vae=bias_vae,
            text_indices=text_indices, vae_indices=vae_indices,
            A_scale=None, B_text_scale=None, B_vae_scale=None,
            use_fp8_w8a8=False, use_int8_w8a8=False,
            use_int8_w8a16=False, use_int4_w4a16=False,
            A_per_channel_quant=False, B_per_channel_quant=False,
            config=dict(CONFIG))
    except Launched:
        print(f"  {label}: OK (reached kernel launch)")
    except AssertionError as exc:
        print(f"  {label}: AssertionError: {exc}")


print("bias combinations the MoT linear constructors can produce:")
run(bias, bias, "bias=True,  vae_bias=True ")
run(None, None, "bias=False, vae_bias=False")
run(bias, None, "bias=True,  vae_bias=False   <-- CONSTRUCTOR DEFAULTS")
run(None, bias, "bias=False, vae_bias=True ")
```

</details>

Output on `main @ f09bdcbc`:

```text
bias combinations the MoT linear constructors can produce:
  bias=True,  vae_bias=True : OK (reached kernel launch)
  bias=False, vae_bias=False: OK (reached kernel launch)
  bias=True,  vae_bias=False   <-- CONSTRUCTOR DEFAULTS: AssertionError: Bias must be provided for both Text and VAE simultaneously, or neither.
  bias=False, vae_bias=True : AssertionError: Bias must be provided for both Text and VAE simultaneously, or neither.
```

### Proposed fix

Make bias presence per expert instead of shared, so the kernel supports what the
layer API already offers:

1. Replace the single `HAS_BIAS: tl.constexpr` with `HAS_BIAS_TEXT` and
   `HAS_BIAS_VAE`, and have `_get_mot_pointers` return which expert a program
   serves so the epilogue can pick the right flag. All three cases stay
   compile-time constant, so the both-bias and no-bias paths generate exactly
   the code they do today.
2. On the host, resolve the two bias pointers together: when only one expert
   owns a bias, point the other expert at that same tensor so the runtime select
   stays typeable and the load stays in bounds, and let `HAS_BIAS_*` drop the
   value for the expert that has none.

That keeps `bias`/`vae_bias` genuinely independent rather than pushing the
constraint back onto callers, and it needs no extra allocation (the alternative,
materialising a zero bias vector for the missing expert, adds a per-call
allocation and a pointless epilogue add).

Happy to send a PR.

yashkgp:
## Purpose

Fixes #6488.

`invoke_mot_gemm` asserted that the text and VAE bias were either both present or
both absent. But `MoTQKVParallelLinear` and `MoTRowParallelLinear` expose `bias`
and `vae_bias` as **independent** constructor flags whose defaults —
`bias=True, vae_bias=False` — produce exactly the rejected pair, so a MoT linear
layer built with the default flags died with
`AssertionError: Bias must be provided for both Text and VAE simultaneously, or
neither.` on its first gen-mode forward, inside the engine core. Two of the four
combinations the API offers were unusable, including the default one. Bagel only
escapes it by happening to pick a symmetric pair for every layer.

The assertion was load-bearing rather than merely conservative: with one bias
absent the host substituted the integer `0` for its pointer, and the kernel
selects between the two bias pointers at runtime, so a select between an integer
and a pointer is not typeable in Triton. Dropping the assertion alone would have
turned a clean error into a load from address 0 on every VAE program.

So this makes bias presence per expert instead of shared:

* **Kernel** — the single `HAS_BIAS: tl.constexpr` becomes `HAS_BIAS_TEXT` and
  `HAS_BIAS_VAE`, and `_get_mot_pointers` now also returns `is_text_program` so
  the epilogue can pick the flag belonging to its own expert. All three cases are
  selected by `constexpr` conditions, so the both-bias and no-bias paths generate
  the same code as before; only the mixed case adds a select.
* **Host** — `_resolve_bias_pointers` resolves the two pointers together. When
  only one expert owns a bias, the other expert points at that same tensor, which
  keeps the runtime select typeable and the load in bounds, and `HAS_BIAS_*`
  drops the value for the expert that has none.

Pointing the bias-less expert at the other's tensor is what avoids the obvious
alternative — materialising a zero bias vector — which would add a per-call
allocation plus a pointless epilogue add.

No caller changes: both MoT linear layers already pass whatever pair they hold,
and that pair is now accepted.

Incidental: `mot_gemm.py` had no SPDX header, and the `check-spdx-header`
pre-commit hook adds one for any file the PR touches. That is the only unrelated
line in the diff.

## Test Plan

New CPU test, `tests/diffusion/kernels/mot/test_mot_gemm_bias.py`, marked
`[core_model, diffusion, cpu]` like its neighbours, so it is selected by the
existing ready lane (`pytest -sv tests/diffusion -m 'core_model and cpu'`). It
covers `_resolve_bias_pointers` for all four bias combinations plus the
mismatched-dtype guard, and asserts `invoke_mot_gemm` forwards the matching
`HAS_BIAS_TEXT` / `HAS_BIAS_VAE` for each. Everything asserted is host-side
argument resolution, so it needs no GPU.

```bash
pytest -sv tests/diffusion/kernels/mot/test_mot_gemm_bias.py -m 'core_model and cpu'
```

Regression check against the whole lane:

```bash
pytest -sv tests/diffusion -m 'core_model and cpu'
```

Versions used locally: vLLM `v0.26.0`, vLLM-Omni `main @ f09bdcbc` (CPU-only
macOS host).

## Test Result

**New test fails on `main`, passes here.** On `origin/main @ f09bdcbc` all 13
cases fail; the two one-sided ones fail with the production error:

```text
FAILED test_mot_gemm_bias.py::test_invoke_mot_gemm_forwards_per_expert_flags[True-False]
FAILED test_mot_gemm_bias.py::test_invoke_mot_gemm_forwards_per_expert_flags[False-True]
>       assert (bias_text is None) == (bias_vae is None), (
E       AssertionError: Bias must be provided for both Text and VAE simultaneously, or neither.
vllm_omni/diffusion/layers/mot/ops/mot_gemm.py:836: AssertionError
```

On this branch:

```text
13 passed, 14 warnings in 0.06s
```

**Bias combinations reaching the kernel** (repro from #6488):

| `bias` / `vae_bias` | `main @ f09bdcbc` | this PR |
|---|---|---|
| `True` / `True` | reaches launch | reaches launch |
| `False` / `False` | reaches launch | reaches launch |
| `True` / `False` *(defaults)* | **AssertionError** | reaches launch |
| `False` / `True` | **AssertionError** | reaches launch |

**Numerical check of the new epilogue (CPU).** This predates the GPU run below,
which supersedes it as evidence — keeping it because it independently checks the
index arithmetic against a from-scratch model rather than against Triton itself.
I re-implemented
`mot_unified_gemm_kernel` (`QUANT_TYPE=0`) line-for-line in PyTorch — routing,
grouped tile coordinates, masked GEMM, bias epilogue, masked store — and drove it
with the launch arguments `invoke_mot_gemm` actually produces, comparing against
`A[rows] @ B_expert + bias_expert`. Shapes are deliberately unaligned (`M=141`
with `BLOCK_SIZE_M=16`, so the last M-group is partial) to exercise the grouped
tile mapping too:

```text
M=141 K=48 N=96  BLOCK_M=16 GROUP_M=8  |text|=57 |vae|=84

bias=True,  vae_bias=True : HAS_BIAS_TEXT=True  HAS_BIAS_VAE=True   max_abs_err=7.63e-06  every_element_written_once=True  PASS
bias=False, vae_bias=False: HAS_BIAS_TEXT=False HAS_BIAS_VAE=False  max_abs_err=7.63e-06  every_element_written_once=True  PASS
bias=True,  vae_bias=False: HAS_BIAS_TEXT=True  HAS_BIAS_VAE=False  max_abs_err=7.63e-06  every_element_written_once=True  PASS
bias=False, vae_bias=True : HAS_BIAS_TEXT=False HAS_BIAS_VAE=True   max_abs_err=7.63e-06  every_element_written_once=True  PASS
```

`every_element_written_once` also confirms the grouped-tile mapping stays a
bijection over the output.

<details>
<summary>emulate_mot.py</summary>

```python
"""Emulate mot_unified_gemm_kernel (QUANT_TYPE=0) on CPU.

Triton has no macOS build, so the kernel body is re-implemented here
line-for-line from the Triton source (routing -> grouped tile coords ->
masked GEMM -> bias epilogue -> masked store) and driven by the *actual*
launch arguments invoke_mot_gemm produces. That exercises the real host-side
argument wiring and the real index arithmetic; only the arithmetic backend
differs.
"""
import types

import torch

from vllm_omni.diffusion.layers.mot.ops import mot_gemm

CONFIG = {"BLOCK_SIZE_M": 16, "BLOCK_SIZE_N": 32, "BLOCK_SIZE_K": 16,
          "GROUP_SIZE_M": 8, "num_warps": 4, "num_stages": 2}


class Captured(Exception):
    pass


def _install_stubs():
    mot_gemm.tl = types.SimpleNamespace(
        float32="fp32", float16="fp16", bfloat16="bf16", int32="i32", int8="i8",
        float8e4nv="f8e4nv", float8e5="f8e5", float8e4m3fn="f8e4m3fn")

    class FakeKernel:
        def __getitem__(self, grid):
            def launch(*args, **kw):
                raise Captured((grid, args, kw))
            return launch

    mot_gemm.mot_unified_gemm_kernel = FakeKernel()
    mot_gemm.triton = types.SimpleNamespace(cdiv=lambda a, b: -(-a // b))


_install_stubs()
cdiv = lambda a, b: -(-a // b)


def capture(A, B_text, B_vae, C, bias_text, bias_vae, text_idx, vae_idx):
    try:
        mot_gemm.invoke_mot_gemm(
            A=A, B_text=B_text, B_vae=B_vae, C=C,
            bias_text=bias_text, bias_vae=bias_vae,
            text_indices=text_idx, vae_indices=vae_idx,
            A_scale=None, B_text_scale=None, B_vae_scale=None,
            use_fp8_w8a8=False, use_int8_w8a8=False,
            use_int8_w8a16=False, use_int4_w4a16=False,
            A_per_channel_quant=False, B_per_channel_quant=False,
            config=CONFIG)
    except Captured as exc:
        return exc.args[0]
    raise RuntimeError("kernel was not launched")


def emulate(grid, args, kw, M):
    """Re-implementation of mot_unified_gemm_kernel for QUANT_TYPE=0."""
    # The first 25 params are positional; the constexpr flags arrive by keyword.
    (A, B_text, B_vae, C, p_bias_text, p_bias_vae, text_idx, vae_idx,
     M_text, M_vae, N, K, stride_am, stride_ak,
     sbkt, sbnt, sbkv, sbnv, stride_cm, stride_cn,
     p_a_scale, p_bt_scale, p_bv_scale, stride_scale_a, stride_scale_b) = args

    BM, BN, BK = kw["BLOCK_SIZE_M"], kw["BLOCK_SIZE_N"], kw["BLOCK_SIZE_K"]
    G = kw["GROUP_SIZE_M"]
    HAS_BIAS_TEXT = kw["HAS_BIAS_TEXT"]
    HAS_BIAS_VAE = kw["HAS_BIAS_VAE"]
    out = torch.full((M, N), float("nan"), dtype=torch.float32)
    touched = torch.zeros((M, N), dtype=torch.int32)

    num_pid_m_text = cdiv(M_text, BM)
    num_pid_n = cdiv(N, BN)
    g = grid(kw) if callable(grid) else grid
    n_programs = g[0]

    for pid in range(n_programs):
        # --- routing (mirrors _get_mot_pointers) ---
        if pid < num_pid_m_text * num_pid_n:
            cur_pid, cur_b, cur_bias = pid, B_text, p_bias_text
            cur_idx, M_limit, is_text = text_idx, M_text, True
        else:
            cur_pid = pid - num_pid_m_text * num_pid_n
            cur_b, cur_bias = B_vae, p_bias_vae
            cur_idx, M_limit, is_text = vae_idx, M_vae, False

        cur_num_pid_m = cdiv(M_limit, BM)
        num_pid_in_group = G * num_pid_n
        group_id = cur_pid // num_pid_in_group
        first_pid_m = group_id * G
        gsma = min(cur_num_pid_m - first_pid_m, G)
        pid_m = first_pid_m + (cur_pid % gsma)
        pid_n = (cur_pid % num_pid_in_group) // gsma

        offs_m_idx = pid_m * BM + torch.arange(BM)
        m_mask = offs_m_idx < M_limit
        rows = cur_idx[offs_m_idx.clamp(max=max(M_limit - 1, 0))].long()
        offs_n = pid_n * BN + torch.arange(BN)
        n_mask = offs_n < N

        # --- _core_standard_gemm ---
        acc = torch.zeros(BM, BN, dtype=torch.float32)
        for k in range(cdiv(K, BK)):
            offs_k = k * BK + torch.arange(BK)
            k_mask = offs_k < K
            a = torch.zeros(BM, BK, dtype=torch.float32)
            b = torch.zeros(BK, BN, dtype=torch.float32)
            for i in range(BM):
                if not m_mask[i]:
                    continue
                for j in range(BK):
                    if k_mask[j]:
                        a[i, j] = A[rows[i], offs_k[j]].float()
            for i in range(BK):
                if not k_mask[i]:
                    continue
                for j in range(BN):
                    if n_mask[j]:
                        b[i, j] = cur_b[offs_k[i], offs_n[j]].float()
            acc += a @ b

        # --- bias epilogue (the code under test) ---
        if HAS_BIAS_TEXT or HAS_BIAS_VAE:
            bias = torch.zeros(BN, dtype=torch.float32)
            for j in range(BN):
                if n_mask[j]:
                    bias[j] = cur_bias[offs_n[j]].float()
            if HAS_BIAS_TEXT and HAS_BIAS_VAE:
                acc = acc + bias.unsqueeze(0)
            elif HAS_BIAS_TEXT:
                acc = acc + torch.where(torch.tensor(is_text), bias.unsqueeze(0), torch.zeros(1))
            else:
                acc = acc + torch.where(torch.tensor(is_text), torch.zeros(1), bias.unsqueeze(0))

        # --- masked store ---
        for i in range(BM):
            if not m_mask[i]:
                continue
            for j in range(BN):
                if n_mask[j]:
                    out[rows[i], offs_n[j]] = acc[i, j]
                    touched[rows[i], offs_n[j]] += 1
    return out, touched


def reference(A, B_text, B_vae, bias_text, bias_vae, text_idx, vae_idx, M, N):
    ref = torch.zeros(M, N, dtype=torch.float32)
    ref[text_idx.long()] = A[text_idx.long()].float() @ B_text.float()
    if bias_text is not None:
        ref[text_idx.long()] += bias_text.float()
    ref[vae_idx.long()] = A[vae_idx.long()].float() @ B_vae.float()
    if bias_vae is not None:
        ref[vae_idx.long()] += bias_vae.float()
    return ref


torch.manual_seed(0)
# M deliberately not a multiple of BLOCK_SIZE_M, and the last M-group partial,
# so the grouped-tile bijection is exercised too.
M, K, N = 141, 48, 96
A = torch.randn(M, K, dtype=torch.float32)
Bt = torch.randn(K, N, dtype=torch.float32)
Bv = torch.randn(K, N, dtype=torch.float32)
perm = torch.randperm(M)
text_idx = perm[:57].sort().values.to(torch.int32)
vae_idx = perm[57:].sort().values.to(torch.int32)
bias = torch.randn(N, dtype=torch.float32)
bias2 = torch.randn(N, dtype=torch.float32)

cases = [
    ("bias=True,  vae_bias=True ", bias, bias2),
    ("bias=False, vae_bias=False", None, None),
    ("bias=True,  vae_bias=False", bias, None),
    ("bias=False, vae_bias=True ", None, bias2),
]
print(f"M={M} K={K} N={N}  BLOCK_M={CONFIG['BLOCK_SIZE_M']} "
      f"GROUP_M={CONFIG['GROUP_SIZE_M']}  |text|={len(text_idx)} |vae|={len(vae_idx)}")
print()
ok = True
for label, bt, bv in cases:
    C = torch.zeros(M, N, dtype=torch.float32)
    grid, args, kw = capture(A, Bt, Bv, C, bt, bv, text_idx, vae_idx)
    out, touched = emulate(grid, args, kw, M)
    ref = reference(A, Bt, Bv, bt, bv, text_idx, vae_idx, M, N)
    max_err = (out - ref).abs().max().item()
    every_once = bool((touched == 1).all())
    good = max_err < 1e-4 and every_once
    ok &= good
    print(f"{label}: HAS_BIAS_TEXT={str(kw['HAS_BIAS_TEXT']):5s} "
          f"HAS_BIAS_VAE={str(kw['HAS_BIAS_VAE']):5s}  "
          f"max_abs_err={max_err:.2e}  every_element_written_once={every_once}  "
          f"{'PASS' if good else 'FAIL'}")
print()
print("ALL PASS" if ok else "FAILURES PRESENT")
```

</details>

**No regressions in the lane.** `pytest tests/diffusion -m 'core_model and cpu'`
run on this branch and on a clean `origin/main` worktree, same interpreter, both
excluding two files that fail to import against vLLM 0.26.0 for unrelated reasons
(`FusedMoEFactory`-era API drift in `tests/diffusion/models/hunyuan_image3/`).
The macOS CPU baseline is not green, so the comparison is against the baseline
rather than against zero:

| | `origin/main` | this branch |
|---|---|---|
| failed | 169 | 169 |
| errors | 87 | 87 |

The two failure sets are byte-identical (same `md5`), i.e. no test changes state
because of this PR.

**Pre-commit gates** run locally on the changed files: `ruff check`,
`ruff format --check`, `check_spdx_header.py`, `check_forbidden_imports.py`,
`check_test_marks.py`, `mypy.py 3.10` — all clean.

## GPU verification

Verified on an A10G (`g5.2xlarge`, driver 595.91.07, torch `2.13.0+cu130`, Triton 3.7.1,
vLLM `0.27.1`). Full output in a comment below; the three results that matter:

1. **The layer-level bug reproduces and is fixed.** With Bagel's own shapes and
   `bias=True` / `vae_bias` left at its default, `origin/main` raises
   `AssertionError: Bias must be provided for both Text and VAE simultaneously, or
   neither.`; on this branch `MoTQKVParallelLinear` and `MoTRowParallelLinear` both
   produce correct output (`rel_err` 2.44e-03 / 2.49e-03 vs an `F.linear` reference).
2. **The kernel compiles and all four bias combinations are correct**, in bf16 and
   fp16, at an unaligned `M=141` (partial last M-group) and at `M=2048`.
3. **The symmetric paths are bit-for-bit unchanged.** For both-bias and no-bias the
   branch output is `torch.equal` to `origin/main`'s kernel on identical inputs, at
   both shapes and both dtypes — so the new branches do fold away at compile time as
   intended.

`pytest tests/diffusion/kernels/mot/` shows no regression: the 8 `*_und_mode` tests
fail identically on both trees when run in isolation (they are `text_indices is None`,
which delegates to `super().forward()` and never reaches this code), and everything
else passes on both, plus the 13 new tests here. The raw full-file counts differ
(main 7 failed / 55 passed, branch 8 failed / 67 passed) purely from test-order
sensitivity inside that already-broken `und_mode` set; details in the comment.

Unrelated pre-existing bug spotted in passing: those `und_mode` failures are all
`Cosine similarity too low: nan` on unmodified `main`. Not touched here — happy to
file it separately.


yashkgp:
Self-review as requested.

**Scope.** Two files: the bias epilogue + host bias-pointer resolution in `vllm_omni/diffusion/layers/mot/ops/mot_gemm.py`, and a new CPU test. No caller changes — both MoT linear layers already pass whatever `(bias, vae_bias)` pair they hold; the kernel just accepts it now.

**What I checked**

- **The assertion is genuinely load-bearing, not just cautious.** My first instinct was to delete it; that would have been wrong. With one bias absent the host passed the integer `0` for that pointer, and the kernel picks between the two bias pointers at runtime, so an int-vs-pointer select is not typeable in Triton. Deleting the assert alone converts a clean error into a load from address 0 on every VAE program. Hence the aliasing approach instead.
- **The both-bias and no-bias paths must not change.** Every new branch is selected by a `constexpr` condition (`HAS_BIAS_TEXT` / `HAS_BIAS_VAE`), so those two configurations should fold to the same code as before; only the mixed case adds a `tl.where`.
- **I deliberately avoided runtime flags.** My first version passed `has_bias_text`/`has_bias_vae` as runtime `int` kernel args and routed them through the existing `if` in `_get_mot_pointers` alongside `cur_stride_bk`. I backed that out: Triton's `equal_to_1` specialization would turn the `1` side into a `constexpr` while the `0` side stayed a runtime value, and I did not want to rely on cross-branch unification of a specialized and an unspecialized arg — a pattern I could not find exercised elsewhere in this kernel. The `constexpr` + `is_text_program` form has no such dependency.
- **Correctness of the epilogue, and that I did not break the tile mapping.** Covered below.
- **No behaviour change for Bagel**, the only current consumer: it passes `bias=True, vae_bias=True` for `qkv_proj` and `bias=False` for `o_proj`, i.e. both symmetric, both unaffected.

**Verification**

- New CPU test (13 cases) covering all four bias combinations at the `invoke_mot_gemm` boundary. **All 13 fail on `origin/main`** — the two one-sided ones with the exact production `AssertionError`. Marked `[core_model, diffusion, cpu]`, so the existing ready lane (`pytest tests/diffusion -m 'core_model and cpu'`) selects it.
- `pytest tests/diffusion -m 'core_model and cpu'` run on this branch and on a clean `origin/main` worktree with the same interpreter: failure sets are **byte-identical** (same `md5`, 169 failed / 87 errors both sides — the macOS CPU baseline is not green, so this is a baseline comparison, not a green run).
- Local pre-commit gates on the changed files: `ruff check`, `ruff format --check`, `check_spdx_header.py`, `check_forbidden_imports.py`, `check_test_marks.py`, `mypy.py 3.10`. mypy caught two real typing errors that ruff did not — both fixed.

**Known gap, stated plainly.** I do not have a CUDA host and Triton has no macOS build, so **I have not compiled this kernel.** To get past "the host wiring is right" I re-implemented `mot_unified_gemm_kernel` (`QUANT_TYPE=0`) line-for-line in PyTorch and drove it with the launch arguments `invoke_mot_gemm` actually produces (details block in the PR body). All four combinations match `A[rows] @ B_expert + bias_expert` at `max_abs_err=7.6e-06`, and every output element is written exactly once — which also confirms I did not disturb the grouped-tile bijection. That checks the arithmetic; it cannot catch a Triton codegen or typing error. Please let the CUDA lane compile it, or run `tests/diffusion/kernels/mot/test_mot_linear.py` on a GPU. I am trying to get GPU access to close this myself and will post results if I do.

**One routing note.** The bot attributed this to `docs/design/module/diffusion/diffusion_model_integration.md` (owners of `/vllm_omni/diffusion/models/`), but the change is in `vllm_omni/diffusion/layers/mot/ops/`. CODEOWNERS has no `/vllm_omni/diffusion/layers/` entry, so it falls through to the general `/vllm_omni/diffusion/` owners rather than the model-integration ones. @princepride authored these MoT kernels in #3960 and is in that group, so probably the best-placed reviewer here.

Also worth flagging separately (not touched in this PR): `tests/diffusion/kernels/mot/` contains only `gpu`-marked tests, and the one Buildkite step that walks that directory selects `-m 'core_model and cpu'`, so both existing files are deselected in every lane. That is why this bug survived. I left the pipeline alone since adding GPU CI time is a maintainer call, but happy to wire it up if you want it in this PR.

yashkgp:
Update: I got GPU access, so **the "kernel not compiled" caveat in the PR description is now closed.** I have updated that section of the description accordingly.

Environment: A10G (`g5.2xlarge`), driver 595.91.07, torch `2.13.0+cu130`, **Triton 3.7.1**, vLLM `0.27.1`, Python 3.13.

### 1. The layer-level bug reproduces on GPU, and the fix resolves it

The headline case from #6488 — `bias=True` with `vae_bias` left at its constructor default — with Bagel's own shapes (`K=3584`, `N=4608`, head 128, 28/4 heads, 2048 tokens, bf16):

```text
########## origin/main ##########
  MoTQKVParallelLinear     AssertionError: Bias must be provided for both Text and VAE simultaneously, or neither.

########## PR branch ##########
  MoTQKVParallelLinear     bias=True vae_bias=<default False>  rel_err=2.44e-03  PASS
  MoTRowParallelLinear     bias=True vae_bias=<default False>  rel_err=2.49e-03  PASS
```

Reference is `F.linear(x[text], weight, bias)` for text rows and `F.linear(x[vae], gen_exp.weight, None)` for VAE rows.

### 2. The kernel compiles, and all four combinations are numerically correct

Kernel-level, comparing against `A[rows] @ B_expert + bias_expert`, at an unaligned `M=141` (partial last M-group, so the grouped-tile path is exercised) and a realistic `M=2048`:

```text
=== bfloat16  M=141 K=48 N=96 ===
  bias=T  vae_bias=T                         rel_err=1.78e-03 PASS | main identical=True
  bias=F  vae_bias=F                         rel_err=1.78e-03 PASS | main identical=True
  bias=T  vae_bias=F  <- was AssertionError  rel_err=1.78e-03 PASS | main AssertionError (expected)
  bias=F  vae_bias=T  <- was AssertionError  rel_err=1.78e-03 PASS | main AssertionError (expected)
=== bfloat16  M=2048 K=1024 N=1536 ===
  bias=T  vae_bias=T                         rel_err=2.97e-03 PASS | main identical=True
  bias=F  vae_bias=F                         rel_err=2.94e-03 PASS | main identical=True
  bias=T  vae_bias=F  <- was AssertionError  rel_err=2.97e-03 PASS | main AssertionError (expected)
  bias=F  vae_bias=T  <- was AssertionError  rel_err=2.94e-03 PASS | main AssertionError (expected)
=== float16  M=141 K=48 N=96 ===   (all four PASS, rel_err <= 4.23e-04)
=== float16  M=2048 K=1024 N=1536 ===   (all four PASS, rel_err <= 3.69e-04)
ALL PASS
```

**`main identical=True` is `torch.equal`, not a tolerance check.** For both symmetric configurations the branch output is bit-for-bit identical to `origin/main`'s kernel on the same inputs, at both shapes and both dtypes. That is the direct evidence for the claim that the both-bias and no-bias paths are unchanged — the new branches fold away at compile time as intended.

### 3. Existing MoT GPU tests: no regression

`pytest tests/diffusion/kernels/mot/` — the `*_und_mode` tests fail, but **identically on both sides**, and they are unrelated to this change (`und` mode is `text_indices is None`, which delegates straight to `super().forward()`):

| | `origin/main` | PR branch |
|---|---|---|
| `und_mode` tests run in isolation | 8 failed | 8 failed |
| everything else | all pass | all pass + 13 new |

Worth being precise about one thing, because the raw full-file numbers look worse than they are: run as a whole file, main reports 7 failed / 55 passed and the branch 8 failed / 67 passed. The +13 passed is my new test file. The 7→8 is **not** a regression — it is test-order sensitivity within the already-broken `und_mode` set. Run those tests in isolation and both trees fail all 8, reproducibly (3/3 runs each). I checked this specifically rather than assuming it, since "one more failure than main" is exactly the kind of thing that deserves proof.

### Separate pre-existing bug I noticed while doing this

Those 8 `und_mode` failures are all `AssertionError: Cosine similarity too low: nan` — the und-mode path returns NaN on this setup, on unmodified `main`. It is unrelated to this PR and I have not touched it, but it is real and reproducible, and it survives because no Buildkite step selects `tests/diffusion/kernels/mot/` (the only lane that walks that directory filters on `-m 'core_model and cpu'` while both files are `gpu`-marked). Happy to open a separate issue for it if that is useful — say the word and I will, with the repro.

Teardown: the GPU instance has been terminated; no volumes or IAM resources left behind.

RuixiangMa:
LGTM

yashkgp:
Hey @RuixiangMa, @hsliuustc0106 

Can we mark this PR as ready to be merged ?

Gaohan123 (approved):
LGTM. Thanks

yashkgp:
Rebased onto current `main` (`ba3d48e1`) to clear the merge conflict. New head is `4038732`, still a single commit.

**What conflicted:** #5908 moved `vllm_omni/diffusion/layers/mot/` to `vllm_omni/diffusion/models/bagel/mot/` and added the same SPDX header this PR added. Apart from that move, main hasn't touched `mot_gemm.py`, and the one-sided-bias assert is still there (`models/bagel/mot/ops/mot_gemm.py:839`), so this fix is still needed.

**Resolution:** the kernel and host change is re-applied as-is at the new path. The diff is identical except for the SPDX lines, which main already carries. The only new line is the test import, `vllm_omni.diffusion.layers.mot.ops` → `vllm_omni.diffusion.models.bagel.mot.ops`, which would otherwise fail collection after the move. No other files are touched.

**Tests** (CPU-only macOS, **vLLM 0.28.0**; current main needs vLLM 0.30 since #7820):
- At `ca14e635f` (last main commit before #7820, MoT files identical to current main) plus this commit, with the full conftest: `pytest tests/diffusion/kernels/mot/test_mot_gemm_bias.py -m 'core_model and cpu'` gives `13 passed`. With `mot_gemm.py` reverted to main's version it gives `13 failed`, and the one-sided cases fail with the production `AssertionError: Bias must be provided for both Text and VAE simultaneously, or neither.` at `mot_gemm.py:839`.
- At current `main @ ba3d48e1` plus this commit I got the same result: 13 passed with the fix, 13 failed with main's file. For this run the top-level `vllm_omni/__init__` was bypassed because it needs vLLM 0.30; the test only exercises `mot_gemm.py`.
- `tests/diffusion/{kernels,models/bagel,layers,quantization,lora} -m 'core_model and cpu'` at `ca14e635f` gives the same failure set with and without this commit (pre-existing vLLM 0.28 / macOS import errors) plus the 13 new passes. pre-commit passes on both files.

The PR is still approved (@RuixiangMa, @Gaohan123) and ready to merge from my side. @Gaohan123, the existing `ready` label predates this push, so Buildkite won't run on the new head by itself. Could you remove and re-add `ready` if you'd like a fresh run before merging? For reference, the last run (build 14326, 08-29) had the CPU diffusion lane green including these 13 tests. Its only failure was an unrelated `EADDRINUSE` in the SanaVideo online-serving test.

PR #6492, opened 2026-08-22
[Recipe] MiniMax-H3 on A100-SXM4-40GB (TP4 + CPU offload / DLO)

CAICAIIs:
## Summary

Adds a deployment recipe for MiniMax-H3 on **NVIDIA A100-SXM4-40GB** (sm80), the
missing datacenter row in the hardware matrix (4090/5090/PRO 5000/6000/H20/DGX-Spark exist; A100 does not).

Three configurations are documented, all validated end-to-end (official 1344×768 reference video with audio, 480×256/96 frames, seed 0). Per-GPU HBM is reported as **denoising level / run peak**, because the two differ substantially:

| Path | Per-GPU HBM (denoising / worst sample) | End-to-end | `peak_memory_mb` (rank 0) |
|---|---|---|---|
| TP4 + CPU offload | 22.8 GiB / **33.0 GiB** (all ranks, reproduced) | 364.9 s | 32,098 |
| DLO rank-local (TP1 + DP4, shared-mmap) | 12.4 GiB / 33.3 GiB (intermittent single rank) | 1844.6 s | 19,108 |
| DLO AllGather (default) | 13.5 GiB / 34.0 GiB (intermittent single rank) | 1069.3 s | 19,724 |

Two measurement notes:

- the **peak** is a transient outside the denoising loop and leaves only ~5-7 GiB free on a 40 GiB
  card, so it — not the denoising level — is what to size against. The TP4 peak reproduces across
  runs; the DLO transients appeared in 2 of 3 measured runs (a 200 ms re-run peaked at 20.4 GiB on
  rank 0 and 13.8 GiB elsewhere), so the table lists them as worst samples rather than peaks.
- `peak_memory_mb` is the primary worker's (global rank 0) torch peak *reserved* device memory
  (`max_memory_reserved`, bytes/1024²); it is not summed across workers and is not comparable with
  the nvidia-smi column.

Key constraints documented:

- **sm80 → BF16 only**: NVFP4 and hardware FP8 are unavailable on A100 (FP8 tensor cores need sm89+).
- Model-level CPU offload via `--enable-cpu-offload` is the primary path. The vLLM-core offload flags
  (`--cpu-offload-gb`, `--offload-group-size`, ...) from the 4×L40S report are no-ops for diffusion
  and are omitted from the command; A/B measurements show identical memory behaviour with and
  without them.
- DLO rank-local (official TP1 shared-mmap topology, #6213) is the lower-HBM alternative and shares
  one checkpoint page-cache copy; DLO AllGather is ~1.7× faster for the same single request (each
  rank copies only its 1/group-size shard) but holds private shmem shards instead.
- Host RAM: 200 GiB minimum for either path, dominated by reclaimable page cache.
- `--vae-patch-parallel-size 4` is safe at this base: the rank-local tiling fallback
  (#6345, merged 2026-08-20) is present and exercised.
- TP8 with the same offload flags is untested; TP8 fully resident (no offload) does not fit on 40 GiB
  and aborts with an NCCL out-of-memory error.

Also links the new recipe from the main MiniMax-H3 recipe.

## Test Plan

- Validated on 8×A100-SXM4-40GB using 4 GPUs per command (`CUDA_VISIBLE_DEVICES=0,1,2,3`); TP8 was
  not exercised, and the recipe says so.
- Every configuration in the table completed end-to-end (`status=completed`, progress 100, valid
  h264 480×256 MP4 with AAC audio), with 1 Hz nvidia-smi sampling plus host-memory sampling and the
  terminal `/v1/videos/{id}` response captured for each run.
- The four-flag A/B, the rank-local DLO arm and the AllGather arm were re-run after review.
- The vLLM-Omni pipeline resolves the partition directory (FL2VA/Ref2VA) automatically; weights load
  from Hugging Face (`MiniMaxAI/MiniMax-H3`) or ModelScope (`MiniMax/MiniMax-H3`).


hsliuustc0106:
can you have a deep analysis of the data with good visualization
`DLO rank-local (TP1 + DP4, shared-mmap) | ~12.6 GiB peak | 2012.8 s | 19,108 MB
`
and why not try allgather + DLO?

CAICAIIs:
> **Update (2026-09-16):** the figures below are re-hosted (they were dropped from the branch as
> unreferenced assets) and relabeled — per-GPU HBM now shows denoising level vs run peak, and
> `peak_memory_mb` is rank 0's peak *reserved* MiB (not worker-aggregated). AllGather was measured
> as well: 1069.3 s vs 1844.6 s for the same request (~1.7x); details in my re-validation comment.

Thanks for the review. Both are addressed below, and the recipe has been updated with a corrected DLO validation block plus a "Why not AllGather?" section (commit `c7032e04`).

## 1. Deep analysis of `~12.6 GiB peak | 2012.8 s | 19,108 MB`

**~12.4 GiB per-GPU peak.** DLO rank-local (TP1 + DP4 + `--dlo-no-use-allgather`) keeps every weight in host memory via the loader's `checkpoint_mmap` plan, so the per-GPU footprint is activations + the double-buffered device slots, not weights:

| Stage | DLO (per GPU) | TP4 + offload (per GPU) |
|---|---|---|
| After model load (worker log) | 3.09 GiB | 10.79 GiB |
| Peak during denoising (live nvidia-smi) | **~12.4 GiB** (12,583–12,685 MiB) | ~22.3 GiB (22,827 MiB) |
| API `peak_memory_mb` (worker-aggregated) | 19,108 | 32,098 |

Breakdown: after load, 3.09 GiB ≈ encoder/VAE buffers + 2 shared device slots (max block 1.23 GiB BF16 each, `distributed_layerwise_backend.py`). The denoising peak adds the resident block + activations for the 480×256/96-frame packed sequence (~35K rows) — weights never land on the GPU. (MiB→GiB: 12,583–12,685 MiB = 12.3–12.4 GiB.) This replaces the recipe's earlier unbacked "~11–14 GiB" with a measured figure.

**2012.8 s** — this is the full end-to-end inference time reported by the API (`OmniTiming total=2012.78s`, engine time). Of that, the 49 denoising steps take ~1962 s at **~40 s/step** (elapsed-diff mean; range ~29–49 s/step, with co-tenant load spikes on this shared node), and the remaining ~51 s is VAE decode + audio generation + MP4 encode. The bottleneck is H2D streaming: 52 DiT blocks (~1.23 GiB max) are copied every step over PCIe — ≈50+ GiB/step of host→device traffic that double buffering can overlap but not eliminate. TP4+offload moves only ¼ per rank with stronger prefetch overlap, hence 5.62 s/step. Post-processing is negligible on both (0.12–0.13 s; MP4 encode 437–460 ms).

**19,108 MB.** `peak_memory_mb` is the worker-aggregated torch metric (`_extract_peak_memory_mb`), not a per-GPU nvidia-smi value — it sums process peaks plus allocator retention. It tracks the nvidia-smi trend (DLO ~40% below TP4's 32,098 MB).

Per-step latency curve (DLO vs TP4), HBM comparison, and E2E time:

DLO per-step latency

DLO vs TP4 HBM

DLO vs TP4 E2E

## 2. Why not AllGather + DLO?

`--dlo-use-allgather` (the default) shards host weights across the DP group (≈¼ per rank) and rebuilds each layer with `all_gather_into_tensor` — the throughput path: NVLink gathers are cheap here (all 8 A100s are NV12-connected through the NVSwitch) and DP>1 allows one request per rank per wave. We validated rank-local for two reasons:

1. **Single-request validation.** AllGather's payoff is DP concurrency, which only materializes under concurrent requests. For one 480×256 request it adds a synchronized collective per streamed block (every rank must enter each wave) without lowering HBM or latency.
2. **Host-memory footprint.** Rank-local TP1 shared-mmap keeps *one* reclaimable OS-page-cache copy of the checkpoint shared by all four replicas (each worker holds only two bounded pinned staging slots). AllGather pins ≈¼ of the model per rank — private, non-reclaimable, ≈1 full model across the group. For a "fit on a shared node" goal, rank-local is the more memory-friendly choice.

The recipe's new "Why not AllGather?" section notes that for concurrent-request throughput, `--dlo-use-allgather` is the recommended variant (drop `--dlo-no-use-allgather`; keep TP1+DP or TP-only, since HSDP+AllGather is rejected by the offloader to avoid double-sharding). Happy to add an AllGather-mode run as a follow-up if that would be useful.

recipes/MiniMaxAI/MiniMax-H3-A100-40GB.md
```diff
@@ -0,0 +1,186 @@
...
+> report in #5700 attributed its low VRAM to that flag combo, but the memory
+> savings there also came from `--enable-cpu-offload`; the extra flags only
+> emitted a vLLM-core warning (`offload_backend="auto"` with both UVA and
+> prefetch fields set).
+
+For Ref2VA, stop the server and restart with `/path/to/MiniMax-H3/Ref2VA`.
+With the official reference video (1344x768, includes an audio reference) the
+live per-GPU peak was ~22.3 GiB (nvidia-smi) during the 480x256/96-frame run —
+within 40 GiB with headroom for longer prompts or more references. The
+worker-aggregated torch peak reported by the API (`peak_memory_mb`) was
```
lishunyang12: Label this as rank 0's peak reserved MiB; workers aren't summed.

recipes/MiniMaxAI/MiniMax-H3-A100-40GB.md
```diff
@@ -0,0 +1,186 @@
...
+oriented path: NVLink gathers are cheap on this node (all eight A100s are
+NV12-connected through the NVSwitch), and DP>1 allows one request per rank per
+denoising wave. This recipe validates the rank-local path (`--dlo-no-use-allgather`)
+instead, for two reasons:
+
+1. **Single-request validation** — AllGather's payoff is DP concurrency (one
+   request per rank per wave), which only shows up with concurrent requests;
+   for one 480×256 request, AllGather adds a synchronized collective per
+   streamed block (every rank must enter each wave) without reducing HBM or
+   latency.
```
lishunyang12: Drop the latency guarantee; AllGather cuts per-rank H2D traffic even for one request.

CAICAIIs:
Thanks @lishunyang12 — both points fixed in `0b3bc5212`.

1. **rank 0 metric**: `peak_memory_mb` is `max_memory_reserved()` on the primary worker only —
   the recipe now says so.
2. **latency claim dropped**, replaced by the measurement: same request, AllGather 1069.3 s vs
   rank-local 1844.6 s (~1.7×), because each rank copies only 1/group-size instead of a full block.

Re-running also showed that both per-GPU "peaks" in the recipe were denoising-steady-state readings:
TP4 actually peaks at 33.7 GiB (reproduced across runs), so a 40 GiB card keeps ~7 GiB free, not
~17 GiB. The recipe now reports denoising level and worst sample separately.

(The three figures were dropped from the branch as unreferenced assets; the copies above now load
from a helper branch and carry the corrected labels.)

recipes/MiniMaxAI/MiniMax-H3-A100-40GB.md
CAICAIIs: Fixed in `0b3bc5212` — all three mentions now label `peak_memory_mb` as the primary worker's (global rank 0) torch peak *reserved* memory in MiB. Re-measured on the documented command: `peak_memory_mb = 32,098.0`. Details in my summary comment.

recipes/MiniMaxAI/MiniMax-H3-A100-40GB.md
CAICAIIs: Fixed in `0b3bc5212`, now with the measurement: same request, AllGather 1069.3 s vs rank-local 1844.6 s (~1.7x) — each rank copies only 1/group-size instead of a full block. The latency claim is gone; the section keeps only the host-memory difference as the reason to pick rank-local.

PR #6499, opened 2026-08-23
[Model][Frontend] MiniMax-H3: Parallelize MP4 response frame conversion

MosCloud:
## Purpose

This PR builds on [#6288](https://github.com/vllm-project/vllm-omni/pull/6288), which introduced direct-planar MP4 response encoding. It parallelizes per-frame conversion for non-streaming MiniMax-H3 video responses.

Each API process now owns one persistent conversion pool with 8 workers. A response uses `min(8, frame_count)` workers, submits work through a bounded FIFO queue, and yields frames in order. Serial and parallel conversion share the same frame builder, and the pool is closed during server shutdown. No CLI or request parameter is added; the legacy fallback and streaming fMP4 path are unchanged.

The fixed pool size is based on the validated `1/2/4/8/16/32` scan: 8 gave the best stable result, while 16 and 32 showed no stable gain. The implementation is generic, but validation is limited to MiniMax-H3 `FL2VA/t2va` on an Atlas A2 host with a 192-CPU Kunpeng 920 and 8 Ascend 910B4-1 NPUs, at `1344x768`, 24 FPS, and one request at a time.

## Test Plan

Focused checks:

```bash
python -m pytest -q \
  tests/entrypoints/openai_api/test_video_api_utils.py \
  tests/entrypoints/openai_api/test_video_server.py \
  tests/entrypoints/openai_api/test_api_server_guards.py
```

Ruff format and check were run on all seven files changed by the PR.

The current CPU comparison used BASE `a94b2580a193c8f0cd6d322bd69335f55f9d0401` and CANDIDATE `2ce21f2640cf540d97b943e1a5eaf8aefe2c0bcd`. Both replayed the same 124-frame MiniMax-H3 dump with `taskset 72-95`, default NUMA memory policy, libx264 `preset=ultrafast, threads=0`, one warmup, and 10 formal runs. The candidate reused one 8-worker converter across warmup and all formal runs, matching the service lifecycle.

The Atlas A2 service comparison used BASE `3d035bfa190e303f53d72e3baa10885f60abe682` and the earlier 8-worker CANDIDATE `3e24ea5ba0a498f2f9feab573104b66ddf8dbf55`: one discarded 5-second warmup per service, then three runs each at 5, 8.7, and 15 seconds with Laser attention, fixed seed, and one request at a time.

**vLLM Version:** `0.26.0+empty`

**vLLM-Omni Commit:** `2ce21f2640cf540d97b943e1a5eaf8aefe2c0bcd`

## Test Result

Focused tests: `158 passed, 15 warnings`. Ruff format and check passed.

Current fixed-dump CPU result:

| Path | MP4 encoding median [range] (ms) | Change |
| --- | ---: | ---: |
| BASE: latest `main`, serial | `4765.80` [4681.50, 4797.07] | - |
| CANDIDATE: persistent 8-worker pool | `1889.32` [1868.97, 1899.98] | `-60.357%` |

All CPU outputs matched the established golden SHA and passed ffprobe plus full video/audio decode.

Atlas A2 service MP4 response-encoding result:

| Requested duration | BASE median [range] (ms) | 8-worker CANDIDATE median [range] (ms) | Change |
| ---: | ---: | ---: | ---: |
| `5 s` | `4826.50` [4696.01, 4831.13] | `1916.88` [1914.95, 1919.35] | `-60.284%` |
| `8.7 s` | `7869.04` [7797.74, 7939.64] | `3184.63` [3167.23, 3190.22] | `-59.530%` |
| `15 s` | `13463.56` [13445.20, 13648.02] | `5443.59` [5436.20, 5444.34] | `-59.568%` |

All 20 service requests, including warmups, produced valid H.264/AAC MP4 output with matching hashes and no encoding fallback.


MosCloud:
Author self-review refreshed for submitted head `037f96c400a16e3252153bac2c72c2d3369312ae`.

I checked that:

- the request-local planar conversion lifecycle now has one owner in `video_frame_conversion.py`, while `video_api_utils.py` retains capability dispatch, mux integration, and public encoding wrappers;
- omitted workers preserve the exact PR1 baseline iterator, explicit `1` uses configured serial conversion without a pool, and `N >= 2` retains ordered bounded conversion clamped to the frame count;
- fallback, streaming, logging, audio/mux behavior, frame quantization, and executor cleanup retain their contracts; and
- the unsandboxed six-file focused suite passed with `217 passed, 15 warnings in 20.24s`, Ruff check/format passed, and submitted-head fixed-dump replays for all three states matched the established golden MP4 and passed route, pool, and full-decode checks.

The extraction changes code ownership without changing the encoding behavior or the measured optimization.

hsliuustc0106:
@lishunyang12 @fan2956 PTAL and test it in GPU/NPU seperately

hsliuustc0106:
Independent verification at PR head `838ec48305a70150d4906a8135f82fa374a364e8` against the documented CPU base `2e096788a6337b0bf7ac22c30bb9afa2afb8f3d4`.

Environment: 2-socket Intel Xeon Platinum 8480C (224 logical CPUs), Python 3.12.13, NumPy 2.3.5, PyAV 18.1.0. The benchmark is CPU-only, so GPUs were hidden. I used the fixed 124-frame 1344x768 input, 24 FPS, 32 kHz stereo audio, seed `20260817`, two excluded warmups, and six measured rounds per invocation. Outer order was counterbalanced as BASE → HEAD, then HEAD → BASE, giving 12 measurements per configuration.

| Configuration | Wall time, mean ± sample SD | Median [min, max] | Process CPU, mean ± sample SD |
| --- | ---: | ---: | ---: |
| BASE direct | 1909.3 ± 349.4 ms | 1896.2 [1447.9, 2437.6] ms | 3954.9 ± 463.6 ms |
| HEAD worker 1 | 1607.0 ± 161.0 ms | 1601.5 [1399.2, 1815.8] ms | 3533.5 ± 250.1 ms |
| HEAD worker 8 | 957.3 ± 79.7 ms | 940.5 [855.1, 1081.5] ms | 3670.3 ± 185.4 ms |

HEAD worker 8 reduced mean wall time by **40.43%** versus HEAD worker 1, with process CPU increasing by **3.87%**. The two outer-order blocks independently showed 39.05% and 41.57% reductions, so the direction remained stable despite host-load drift.

All formal BASE/worker-1/worker-8 outputs were byte-identical: 88,136,362 bytes with SHA-256 `848e238962c10f87b5dfbcd350edf5a9b5bbcc1db589b9579ee00a6d69159ca0`.

Process-level sampled peak RSS was 3,374,796/3,401,988 KB for the two BASE invocations and 3,364,600/3,364,660 KB for HEAD, a pooled HEAD-vs-BASE change of -0.70%. This is a whole-process comparison containing both benchmark variants, not a strict per-variant RSS measurement.

The focused CPU suite also passed: **213 passed** across the benchmark, video encoding/server, API guard, and CLI tests.

Result: the local A/B confirms a substantial worker-8 speedup with byte-identical output and no observed branch-level memory regression. It does not attempt to reproduce the exact 49.646% Kunpeng/Atlas A2 result: this host is x86 and experienced load averages from roughly 30 to 102 during the run. The full MiniMax-H3 Atlas A2/Ascend E2E matrix was not independently rerun here.

vllm_omni/entrypoints/openai/video_api_utils.py
```diff
@@ -611,38 +628,94 @@ def _resolve_audio_sample_rate(audio: AudioInput | None, audio_sample_rate: int
     return audio_sample_rate or DEFAULT_AUDIO_SAMPLE_RATE
 
 
+def _planar_scratch_dtype(common_dtype: np.dtype) -> np.dtype:
```
hsliuustc0106: can we move planar related into a seperate script?

hsliuustc0106: More specifically, please move the helpers from _planar_scratch_dtype through _iter_planar_video_frames (thread-local scratch, frame construction, bounded scheduling, and shutdown) into a dedicated internal module such as video_frame_conversion.py; keep capability dispatch and the public encoding wrappers here. That gives this request-local resource lifecycle one owner instead of expanding the general 905-line utility module.

vllm_omni/entrypoints/openai/video_api_utils.py
```diff
@@ -611,38 +628,94 @@ def _resolve_audio_sample_rate(audio: AudioInput | None, audio_sample_rate: int
...
+    worker_count = min(len(frames), _validate_frame_conversion_workers(worker_count))
+    if worker_count <= 1:
+        try:
+            for frame in frames:
+                yield _build_planar_video_frame(frame, common_dtype)
+        finally:
+            _clear_planar_channel_scratch()
+        return
+
+    executor = ThreadPoolExecutor(max_workers=worker_count, thread_name_prefix="video-planar")
```
hsliuustc0106: The per-request bound is solid, but there is no process-wide budget. True concurrent encodes or multiple API processes multiply workers, pending frames, scratch buffers, and codec threads. Please track service-level admission/backpressure and concurrency benchmarks before recommending more than one worker outside the documented single-request scope.

MosCloud: Agreed. The pool is request-local, so concurrent encodes or multiple API processes multiply workers, pending frames, scratch buffers, and codec threads. I updated the Design and Scope sections to make that boundary explicit and kept the `8 workers` recommendation limited to the validated MiniMax-H3 FL2VA/t2va Atlas A2 single-request setup. Service-level admission/backpressure and concurrency benchmarks should precede any broader recommendation.

vllm_omni/entrypoints/openai/video_api_utils.py
MosCloud: Done in `037f96c4`. I moved the planar conversion lifecycle, including scratch management, frame construction, the exact baseline iterator, the configured bounded scheduler, and shutdown, into `video_frame_conversion.py`. `video_api_utils.py` now retains capability dispatch, mux integration, and the public encoding wrappers, and the helper/lifecycle tests live in `test_video_frame_conversion.py`.

The unsandboxed focused suite passes with `217 passed, 15 warnings`, and submitted-head fixed-dump replays for omitted, explicit `1`, and explicit `8` all matched the established golden MP4 and passed route, pool, and full-decode checks.

hsliuustc0106:
@alex-jw-brooks PTAL

recipes/MiniMaxAI/MiniMax-H3-NPU.md
```diff
@@ -110,28 +110,220 @@ Do not add `--enforce-eager`. The first request includes regional
...
-### CPU MP4 response encoding (Atlas A2)
+### CPU MP4 response encoding: MiniMax-H3 FL2VA/t2va (Atlas A2)
 
-Non-streaming MP4 responses use one public automatic encoder. It checks the
-runtime frame shape, common dtype, and RGB channel-plane contiguity for every
-request; compatible inputs use direct planar PyAV frames, while unsupported
-inputs fall back to the legacy muxer before opening the PyAV container. No CLI
-flag, model declaration, or user configuration is required. Streaming fMP4
-output remains on its existing incremental path.
+This section records measured CPU response-encoding behavior for MiniMax-H3
```
alex-jw-brooks: Similar to the other PR - I think most of this should be removed, if a person is looking at this for a recipe, they likely just want brief guidance on how to get started with running things quickly, and will probably not read a long technical analysis.

It would be better to just give an optimal command with maybe 1-2 sentences explaining in my opinion. We definitely should not have internal details about the memory model and internal frame formats like gbr pixel format here

vllm_omni/entrypoints/openai/video_api_utils.py
```diff
@@ -731,17 +718,43 @@ def _encode_video_bytes(
...
-        video_codec_options=video_codec_options,
+    if requested_frame_conversion_workers is None:
+        return mux_av_video_audio_bytes(
+            video_frame_conversion._iter_baseline_planar_video_frames(frames, common_dtype),
+            width=frame_shape[1],
+            height=frame_shape[0],
+            audio_waveform=audio_np,
+            fps=float(fps),
+            # None occurs only without audio; the muxer does not consume the rate then.
+            audio_sample_rate=cast(int, effective_audio_sample_rate),
```
alex-jw-brooks: I think rather than casting here, `mux_av_video_autio_bytes` should be updated to reflect an optional sampling rate, correct?

I.e., the place where it is passed to (https://pyav.org/docs/develop/api/container.html#av.container.OutputContainer.add_stream is an optional), so we shouldn't force an invalid cast to int here

vllm_omni/entrypoints/openai/video_frame_conversion.py
```diff
@@ -0,0 +1,151 @@
...
+        _planar_scratch_local.channel = scratch
+    return cast(np.ndarray, scratch)
+
+
+def _clear_planar_channel_scratch() -> None:
+    if hasattr(_planar_scratch_local, "channel"):
+        delattr(_planar_scratch_local, "channel")
+
+
+def _build_planar_video_frame(frame: np.ndarray, common_dtype: np.dtype) -> av.VideoFrame:
```
alex-jw-brooks: This should use the same core as `_iter_baseline_planar_video_frames`, since it is just the single frame case, instead of recopying the entire algorithm

vllm_omni/entrypoints/openai/video_frame_conversion.py
```diff
@@ -0,0 +1,151 @@
...
+            pending.append(executor.submit(_build_planar_video_frame, frame, common_dtype))
+            if len(pending) == max_pending:
+                break
+
+        while pending:
+            converted_frame = pending.popleft().result()
+            try:
+                next_input = next(frame_iter)
+            except StopIteration:
+                pass
```
alex-jw-brooks: I think this is really opaque. If I understand correctly, this is just indexing into the list, but clamping right? If so, why not just do that directly without the iterator / stop iteration catch? E.g.,

```python
next_idx = min(max_pending, len(frames))
while pending:
    converted_frame = pending.popleft().result()
    if next_idx < len(frames):
        pending.append(executor.submit(_build_planar_video_frame, frames[next_idx], common_dtype))
        next_idx += 1
    yield converted_frame
```

vllm_omni/entrypoints/openai/video_frame_conversion.py
```diff
@@ -0,0 +1,151 @@
...
+from typing import TYPE_CHECKING, cast
+
+import numpy as np
+from numpy.typing import DTypeLike
+
+if TYPE_CHECKING:
+    import av
+
+
+_planar_scratch_local = threading.local()
```
alex-jw-brooks: NIt: we should try to avoid global state where possible, since this also only applies to some models, but is called on imports.

Also, given that this will simplify a lot after collapsing the planar frame building n=1 / n > 1 cases into one implementation and using list indexing instead of the iterator, this module will be pretty small. Can you fold it into `vllm_omni/entrypoints/openai/video_api_utils.py` as well to try to keep things more organized?

vllm_omni/entrypoints/openai/video_api_utils.py
```diff
@@ -693,10 +661,23 @@ def _encode_video_bytes(
...
+        None
+        if frame_conversion_workers is None
+        else video_frame_conversion._validate_frame_conversion_workers(frame_conversion_workers)
+    )
+    if requested_frame_conversion_workers is None:
+        frame_conversion_mode = "baseline_unconfigured"
+    elif requested_frame_conversion_workers == 1:
+        frame_conversion_mode = "configured_serial"
+    else:
+        frame_conversion_mode = "configured_parallel"
```
alex-jw-brooks: please do not use raw strings for the conversion mode. It's easy to break when refactoring

tests/entrypoints/openai_api/test_video_server.py
```diff
@@ -111,8 +116,75 @@ async def _generate_both_response_types():
...
-    assert "encoding_config" not in raw_encoder.call_args.kwargs
-    assert "encoding_config" not in base64_encoder.call_args.kwargs
+    assert raw_encoder.call_args.kwargs["frame_conversion_workers"] == workers
+    assert base64_encoder.call_args.kwargs["frame_conversion_workers"] == workers
+
+
+@pytest.mark.parametrize(
+    ("workers", "mode", "requested_workers", "effective_workers"),
+    [
+        (None, "baseline_unconfigured", "unset", "1"),
```
alex-jw-brooks: If I understand correctly, the resolution is that `baseline_unconfigured` and `configured_serial` both resolve to 1 effective workers, right?

Unless I am misunderstanding, we should not distinguish this in the code and maintain the 3 separate (str) states. Because all that should matter to control the behavior is the number of `effective_workers` for serial vs parallel, which will simplify the code a lot

vllm_omni/entrypoints/cli/serve.py
```diff
@@ -832,6 +843,16 @@ def subparser_init(self, subparsers: argparse._SubParsersAction) -> TrackingArgu
...
+        omni_config_group.add_argument(
+            "--video-response-frame-conversion-workers",
+            type=_video_response_frame_conversion_workers,
+            default=None,
+            help=(
+                "Opt into configurable CPU planar-frame conversion for non-streaming MP4 responses with a "
+                "positive worker count. Omitting this option preserves baseline serial conversion; configured "
+                "worker counts are clamped to each request's frame count."
+            ),
+        )
```
alex-jw-brooks: I do not think we should make this an API flag. If outputs are bit identical and it is expected that this will always be faster, then it will just cause people to deploy with less optimal settings until they read the recipe.

Rather than exposing this and having the pool be created per request, I think it would be cleaner to just detect a default that makes sense based on the concurrency experiments you ran, e.g., something like `min(os.cpu_count() // 4, 8)`, and using that as a persistent pool. Then we won't have to change the API, recreate/shutdown the pool per request, and it also should make this PR simpler

alex-jw-brooks (changes requested):
Some suggestions! Thanks for working on this. I think that this PR is much bigger than it needs to be, and that a lot redundancy can be removed, please see the review for suggestions.

I also think we should avoid an API flag here. If it is always expected to be faster, we should infer a reasonable threadpool count and just enable it by default instead of maintaining an extra path + api flag and relying on users to discover it to opt in

vllm_omni/entrypoints/openai/video_frame_conversion.py
MosCloud: Collapsed serial and parallel conversion onto the same _PlanarFrameConverter._build_frame implementation. There is no separate single-frame conversion algorithm now.

vllm_omni/entrypoints/openai/video_api_utils.py
MosCloud: Updated mux_av_video_audio_bytes to accept an optional sample rate and removed the invalid cast. Its existing 44.1 kHz default is resolved only when audio is present, with focused coverage for both audio and video-only calls.

vllm_omni/entrypoints/openai/video_frame_conversion.py
MosCloud: Removed module-level thread-local state. Scratch storage now belongs to the serving-owned converter instance, and the simplified implementation has been folded back into video_api_utils.py.

vllm_omni/entrypoints/openai/video_api_utils.py
MosCloud: The conversion lifecycle now has one serving-owned _PlanarFrameConverter, but the simplified implementation is kept in video_api_utils.py following the later review. The separate video_frame_conversion.py module and its dedicated test file have been removed.

vllm_omni/entrypoints/openai/video_api_utils.py
MosCloud: Removed the conversion-mode strings and the three-state routing. Request logs now report the selected encoding path and effective worker count directly.

vllm_omni/entrypoints/openai/video_frame_conversion.py
MosCloud: Replaced the iterator and StopIteration handling with direct list indexing and a bounded deque. The next task is submitted before the completed frame is yielded, while output remains FIFO.

tests/entrypoints/openai_api/test_video_server.py
MosCloud: Collapsed the baseline/configured states entirely. There is no CLI-controlled serial state now; the serving path uses one persistent pool and serial utility callers use the same frame builder without creating a pool.

vllm_omni/entrypoints/cli/serve.py
MosCloud: Removed the CLI and request-local pool lifecycle. Each API process now owns one persistent pool fixed at 8 workers, and each response uses min(8, frame_count). I kept 8 as a measured fixed value rather than deriving it from os.cpu_count, which may not reflect the service affinity or CPU quota. The pool is closed during server shutdown.

vllm_omni/entrypoints/openai/video_api_utils.py
MosCloud: The follow-up simplification changed the module boundary: after removing the CLI, three routing states, and request-local lifecycle, the remaining converter is small and has been folded back into video_api_utils.py as requested in the later review.

vllm_omni/entrypoints/openai/video_api_utils.py
MosCloud: The redesign removes per-request worker multiplication: concurrent encodes in one API process share a single persistent pool capped at 8 conversion threads. Per-request submission remains bounded, and multiple API processes each own their own process-local pool. The validation claim remains limited to one request at a time.

recipes/MiniMaxAI/MiniMax-H3-NPU.md
MosCloud: Removed the PR-specific recipe expansion. The optimization is now automatic and has no serving flag, so this PR no longer changes the MiniMax-H3 recipe.

vllm_omni/entrypoints/openai/video_api_utils.py
hsliuustc0106: The later folding-back supersedes the module split, and I'm fine with that endpoint: with the CLI flag, the three routing states, and the request-local lifecycle gone, the remaining converter is small enough that `video_api_utils.py` is the right home. Dropping my structural ask.

vllm_omni/entrypoints/openai/video_api_utils.py
hsliuustc0106: The final design answers this directly — one persistent process-local pool capped at 8 removes the per-request worker/buffer multiplication, and the single-request validation boundary stays documented. Resolved.

hsliuustc0106 (approved):
LGTM — both my threads are resolved: the planar conversion ended up simpler than the module split I asked for (folded back into `video_api_utils.py` once the CLI flag and routing states went away), and the persistent process-local pool capped at 8 answers the per-request multiplication concern. The single-request validation boundary stays documented.

hsliuustc0106 (approved):

alex-jw-brooks (approved):
Thanks for your contribution, LGTM :)

PR #6486, opened 2026-08-22
[Diffusion] Integrate no-AllGather DLO with Host Weight Runtime

hsliuustc0106:
## Summary

Integrates the final-layout Host Weight Runtime (HWR) contract from #6445 with
no-AllGather distributed layerwise offload (DLO).

- Adds opt-in `disabled` / `preferred` / `required` HWR configuration with
  zero-interaction gates when HWR or no-AllGather DLO is not selected.
- Restores exact final-layout DiT tensors on warm hits and skips ordinary DiT
  materialization.
- Publishes validated final-layout artifacts after canonical cold loading.
- Splits cold byte-changing finalization from shared cold/warm validation.
- Carries one process-local lease from loader to backend without making the
  model runner HWR-aware.
- Uses two bounded pinned host staging slots per rank and drains pending work
  before lease release.
- Uses a fresh canonical model after post-commit preferred-mode startup
  failure; required mode fails.
- Preserves checkpoint mmap when HWR is not selected and hardens filesystem
  artifacts after atomic rename, including overlayfs.

## Real MiniMax-H3 validation

All GPU cases used `gpu run`, eager mode, and `CUDNN_ATTN`.

### Published recipe defaults

The authoritative `examples/offline_inference/minimax_h3/dlo_lifecycle.py`
defaults are 1344x768, 5 seconds, 24 fps, and 2 steps.

| Mode | Parallel topology | Engine init (s) | First request (s) | Peak HBM (MB) | Output shape |
|---|---|---:|---:|---:|---|
| No offload | DP1 / TP2, encoder TP2, VAE TP2 | 73.23 | 17.848 | 77,218 | video `[124,768,1344,3]`, audio `[1,2,165600]` |
| DLO no-AllGather | DP2 / TP1, encoder TP1, VAE TP1 | 98.32 | 26.354 | 19,340 | video `[124,768,1344,3]`, audio `[1,2,165600]` |

DLO saved 57,878 MB HBM (74.96%, 3.99x lower) with a 1.477x first-request
latency cost. Both modes shut down with zero active children.

Matched warmed requests were then profiled with Nsight Systems. The measured
prompt/seed differed from warm-up while recipe shape and generation settings
remained fixed.

| Profile | Warm request (s) | Peak HBM (MB) |
|---|---:|---:|
| No offload DP1 / TP2 | 10.490 | 77,218 |
| DLO no-AllGather DP2 / TP1 | 20.073 | 19,338 |

[...]

PR #6501, opened 2026-08-23
[Diffusion] Add an FP8 Host Weight Runtime producer for DLO AllGather

baonudesifeizhai:

## Purpose

## Purpose

Implements Phase 1 of #6330 for online per-tensor FP8 with DLO AllGather.

- Produces canonical FP8 weights and scales from checkpoint tensors with bounded GPU scratch.
- Publishes them through the shared Host Weight Runtime instead of introducing a separate cache.
- Restores `HostWeightLease` / `HostWeightPlan` directly into the existing DLO AllGather consumer.
- Avoids ordinary full-model DiT materialization during cold production and warm reuse.
- Keeps canonical artifacts independent of DP/SP topology.

Scope is CUDA, MiniMax-H3 `Fp8PerTensorOnlineLinearMethod`, DiT weights,
TP=1, non-HSDP, and multi-rank DLO AllGather. Other quantization methods,
TP/HSDP, and cross-node sharing remain out of scope.

This builds on #6419 and #6486 and does not change the AllGather protocol.

## Test Plan

Focused HWR, final-layout, loader, and DLO CPU suites: 205 passed.
Ruff passed.

### MiniMax-H3 Ref2VA E2E

Four B300 GPUs, DP4/SP1, online per-tensor FP8, DLO AllGather,
1344x768, five seconds, three inference steps, and two overlapping
fixed-seed requests. Checkpoint pages were prewarmed; cold means a fresh
HWR artifact root.

| Run | Startup | vs. baseline | HWR result |
| --- | ---: | ---: | --- |
| Baseline | 140 s | — | disabled |
| Cold production | 143 s | +2.1% | `local_production` |
| Warm reuse | 99 s | -29.3% / 1.41x | `local_hit` |

Both requests produced identical decoded RGB-video and PCM-audio SHA256
values across baseline, cold, and warm. Cold and warm resolved the same
artifact identity; all four warm workers reported `local_hit`, with no
reproduction.

### Peak-resource smoke

MiniMax-H3 Ref2VA, DP2/SP2, same four-GPU environment:

| Run | Startup | Peak RSS | Peak PSS | Peak HBM |
| Baseline | 150 s | 243.9 GiB | 213.3 GiB | 99.0 GiB |
| Cold | 143 s | 211.4 GiB | 209.1 GiB | 15.4 GiB |
| Warm | 98 s | 210.6 GiB | 208.4 GiB | 16.9 GiB |

An artifact produced under DP2/SP2 was also reused under DP4/SP1 with
the same identity, local hits on every rank, and exact output parity.

The generic ordered-writer and source-digest optimizations overlap with
#6591; this PR contains no CUDA-registration or no-AllGather transport changes.

Relates to #6330.


baonudesifeizhai:
will consider https://github.com/vllm-project/vllm-omni/pull/6486

baonudesifeizhai:
Run | Startup | vs baseline | Peak RSS | Peak PSS | Peak HBM
-- | -- | -- | -- | -- | --
Baseline | 147s | — | 237.17 GiB | 213.83 GiB | 89.37 GiB
Cold | 155s | +8s / +5.4% | 240.45 GiB | 236.38 GiB | 16.91 GiB
Warm | 91s | −56s / −38.1% | 241.44 GiB | 239.07 GiB | 14.45 GiB

baonudesifeizhai:
Previously, we waited until the entire 31 GiB HWR artifact had been copied before releasing its file-backed pages. This change releases each tensor’s artifact pages immediately after that tensor has been copied.

The new flow is:

```text
mmap the tensor from the HWR artifact
        ↓
copy the portion required by the current rank
        ↓
write it into the long-lived pinned CPU DLO shard
        ↓
call MADV_DONTNEED to release the tensor’s artifact-backed pages
        ↓
replace the original model tensor with a placeholder
```

Previously, both of the following remained resident during initialization:

```text
~31 GiB of artifact-backed resident pages
+
the final pinned DLO shards
```

As a result, the warm-start peak PSS reached approximately 239 GiB.

With this change, artifact-backed pages are released incrementally as tensors are copied, instead of being retained until the entire model has been processed. Therefore, memory usage during initialization now consists primarily of:

```text
the small set of artifact pages currently being copied
+
the final pinned DLO shards
```

This reduces the warm-start peak PSS to approximately 208.4 GiB—a reduction of about 30.7 GiB, which closely matches the size of the released artifact-backed pages.
 
1. Reference video: `assets/r2va.mp4`
2. Topology: DP2 / SP2
3. GPUs: 4, 5, 6, and 7
4. Online per-tensor FP8 quantization
5. DLO AllGather
6. Three inference steps
7. Two concurrent requests
8. Identical video and audio hashes across baseline, cold, and warm runs
  
Run | Startup | Peak RSS | Peak PSS | Peak HBM
-- | -- | -- | -- | --
baseline | 150s | 243.94 GiB | 213.33 GiB | 98.98 GiB
cold | 143s | 211.39 GiB | 209.11 GiB | 15.39 GiB
warm | 98s | 210.57 GiB | 208.40 GiB | 16.91 GiB

baonudesifeizhai:
DP4/SP1 
Run | Startup | Peak RSS | Peak PSS | Peak HBM
-- | -- | -- | -- | --
baseline | 159s | 256.64 GiB | 175.29 GiB | 96.54 GiB
cold | 155s | 160.67 GiB | 158.11 GiB | 13.41 GiB
warm | 115s | 160.15 GiB | 157.74 GiB | 13.60 GiB

baonudesifeizhai:
MiniMax-H3 Ref2VA, online per-tensor FP8, DLO AllGather,
DP4/SP1 on four GPUs, two concurrent 3-step requests:

| Run | Startup | vs. baseline | Result |
| --- | ---: | ---: | --- |
| Baseline | 140 s | — | reference |
| Cold production | 143 s | +2.1% | exact video/audio parity |
| Warm local hit | 99 s | -29.3% / 1.41x | exact video/audio parity |

Cold production and warm reuse resolved the same artifact identity.
All four warm workers reported `local_hit`; no warm reproduction occurred.
Both requests produced byte-identical decoded video and audio hashes across
baseline, cold, and warm. Shutdown released the server port and removed the
temporary artifact.

baonudesifeizhai:
@vllm-omni-review-bot

baonudesifeizhai:
Self-review completed.

I checked:

- FP8 eligibility, source identity, bounded production, HWR resolution, lease/plan handoff, and DLO AllGather consumption;
- canonical artifact identity remains independent of DP/SP topology;
- cold production elects one builder and warm startup does not invoke the producer;
- file/tensor digests, durable publication, source guards, lease release, and shutdown ordering;
- unsupported configurations remain outside the Phase 1 path;
- focused CPU suites passed;
- DP4/SP1 baseline/cold/warm produced exact decoded video/audio parity;
- DP2/SP2 to DP4/SP1 cross-topology reuse produced the same identity and local hits on every rank.

I found no remaining correctness blocker for the Phase 1 scope. The generic
ordered-writer/source-digest changes overlap with #6591; this PR does not
include CUDA registration or no-AllGather transport changes.

baonudesifeizhai:
@vllm-omni-review-bot please review this PR.

hsliuustc0106:
do we have a follow-up PR working for no-allgather version with online fp8?

baonudesifeizhai:
Will split in to next pr for online fp8   no-allgather 

> do we have a follow-up PR working for no-allgather version with online fp8?

vllm_omni/diffusion/model_loader/host_weights/producers/final_layout_fp8.py
```diff
@@ -0,0 +1,287 @@
...
+    host_input: torch.Tensor
+    device_input: torch.Tensor
+    host_output: torch.Tensor
+    device_output: torch.Tensor
+    host_amax: torch.Tensor
+    pending_start: int | None = None
+    pending_rows: int = 0
+
+
+class FinalLayoutFP8Producer:
```
hsliuustc0106, line 73: [P2] The new FP8 runtime path ships with no tests. The producer, policy, preparation, eligibility gate, the DLO `runtime_lease_storage` sharding, and `release_tensor_pages` are all uncovered — the "205 passed" suites are the pre-existing BF16/HWR ones, and the only test change here is fixture isolation plus a copyright header.

The BF16 producer has a fully CPU-runnable mirror suite this can follow: `FinalLayoutBF16Producer` constructed directly in tests/diffusion/model_loader/test_final_layout_host_weights.py:965 (sharded production, determinism, restore rejection), `write_rows` in tests/host_weight_runtime/test_filesystem_store.py:136, loader cold/warm/fallback in tests/diffusion/model_loader/test_diffusers_loader.py:166, DLO lease consumption in tests/diffusion/offloader/test_distributed_layerwise_backend.py:748. The same shape needs no CUDA here: `runtime_fp8_requested` scope matrix + REQUIRED raise, policy validators, `FinalLayoutFP8ModelPreparation` prepare/activate round-trip, `split_tensor_targets_by_bytes` (incl. a weight/scale pair split across a shard boundary), `_tensor_ranges`, and the `as_strided` lease-slice path. Could you add the mirror suite in this PR?

vllm_omni/diffusion/data.py
```diff
@@ -795,9 +795,8 @@ class OmniDiffusionConfig:
     dlo_use_allgather: bool = True
     # Leading main-DiT blocks kept resident by distributed layerwise offload.
     dlo_resident_layers: int = 0
-    # Final-layout Host Weight Runtime policy. The loader only activates this
-    # for eligible no-AllGather DLO; all other configurations preserve their
-    # existing loader/storage path.
+    # Final-layout Host Weight Runtime policy. Representation-specific
```
hsliuustc0106, line 798: [P2] The eligibility contract this comment rewrites is still documented as the opposite in the design pages that own it. docs/design/feature/offloader/distributed_layerwise_offload.md says HWR backing "applies only to no-AllGather DLO" (:204-206), routes "DLO AllGather -> zero final-layout HWR interaction" in its eligibility tree (:234-238), and states "DLO AllGather never receives an HWR final-layout lease" (:289-292); docs/design/feature/host_weight_runtime.md:34-42 still lists FP8 producers and lease handoff to DLO as out of V1 scope; docs/design/module/host_weight_runtime.md:276-291 likewise. Since the preferred/required fallback semantics and the new eligibility rules are user-facing configuration behavior, those pages should move with the code in this PR.

vllm_omni/diffusion/model_loader/host_weights/producers/final_layout_fp8.py
```diff
@@ -0,0 +1,287 @@
...
+            slot.host_input[:rows].copy_(source[start : start + rows])
+            with torch.cuda.stream(slot.stream):
+                slot.device_input[:rows].copy_(slot.host_input[:rows], non_blocking=True)
+                low, high = slot.device_input[:rows].aminmax()
+                slot.host_amax.copy_(torch.maximum(low.abs(), high.abs()).float().reshape(1), non_blocking=True)
+                slot.ready.record(slot.stream)
+            slot.pending_rows = rows
+        for slot in slots:
+            amax = self._collect_amax(slot, amax)
+        probe = amax.to(device=self._device, dtype=source.dtype).reshape(1, 1)
```
hsliuustc0106: [P2] `scale` is produced on the current (default) stream but consumed by kernels on the per-slot pool streams with no ordering between them: `amax.to(...)` and the `scaled_fp8_quant` probe launch on the default stream here, then `_quantize_rows` passes `scale=scale` into `ops.scaled_fp8_quant` under `torch.cuda.stream(slot.stream)`. Pool streams are created with `cudaStreamNonBlocking`, so they do not implicitly order against the legacy default stream. Today the µs-scale probe plus the ms-scale pinned-host staging between the two sections makes the race practically unreachable (the default stream is otherwise idle during production), but any default-stream work landing during cold production turns it into silently wrong scales on every tensor. Recording the probe's completion — an event on the current stream with each slot stream `wait_event` before its first quant chunk, or simply running the probe on one of the slot streams — closes it.

vllm_omni/diffusion/model_loader/host_weights/producers/final_layout_fp8.py
```diff
@@ -0,0 +1,287 @@
...
+        generated_scales: dict[str, torch.Tensor] = {}
+
+        shards = split_tensor_targets_by_bytes(records, self._max_shard_bytes)
+        for index, shard in enumerate(shards, start=1):
+            specs = tuple(
+                TensorWriteSpec(record.name, tuple(record.tensor.shape), record.tensor.dtype, record.kind, record.role)
+                for record in shard
+            )
+            file_name = f"model-{index:05d}-of-{len(shards):05d}.safetensors"
+            with writer.open_tensor_file(file_name, specs) as output:
```
RuixiangMa, line 131: If a weight_scale record appears before its weight record (possible if nn.Module._parameters ordering differs), the scale falls through to _write_checkpoint_tensor, which attempts to read a non-existent checkpoint tensor. This currently works only because prepare() registers weight before weight_scale — an undocumented invariant.

vllm_omni/diffusion/model_loader/host_weights/producers/final_layout_fp8.py
```diff
@@ -0,0 +1,287 @@
...
+
+        return make_slot(), make_slot()
+
+    def _find_scale(
+        self,
+        source: torch.Tensor,
+        slots: tuple[_QuantSlot, _QuantSlot],
+        rows_per_chunk: int,
+    ) -> torch.Tensor:
+        amax = torch.zeros((1,), dtype=torch.float32)
```
RuixiangMa, line 230: Initializes amax = torch.zeros((1,)) with no epsilon guard. If all weights are zero, scale = 0/448 = 0 and subsequent quantization divides by zero. NaN in source propagates through aminmax → torch.maximum → scale with no detection.

vllm_omni/diffusion/model_loader/host_weights/runtime_fp8.py
```diff
@@ -0,0 +1,75 @@
...
+
+logger = init_logger(__name__)
+
+
+class RuntimeFP8UnavailableError(RuntimeError):
+    pass
+
+
+def _scope_reason(config: object, load_format: str, device: torch.device | None) -> str | None:
+    parallel = getattr(config, "parallel_config")
```
RuixiangMa: _scope_reason crashes if parallel_config is missing,use getattr(config, "parallel_config", None).

vllm_omni/diffusion/model_loader/host_weights/fp8_layout.py
```diff
@@ -0,0 +1,225 @@
...
+            )
+            layer.input_scale = None
+            layer._already_called_process_weights_after_loading = True
+
+    def activate_kernel_views(self) -> None:
+        from vllm.model_executor.utils import replace_parameter
+
+        for layer in self._layers:
+            method = layer.quant_method
+            replace_parameter(layer, "weight", layer.weight.t().data)
```
RuixiangMa, line 210: layer.weight.t().data shares storage with the HWR lease tensor. If process_weights_after_loading does in-place modification (e.g., Marlin reordering), it corrupts the immutable lease storage. Verify that online per-tensor FP8 never modifies weights in-place, or clone before calling.

vllm_omni/diffusion/model_loader/host_weights/producers/final_layout_fp8.py
```diff
@@ -0,0 +1,287 @@
...
+        probe = amax.to(device=self._device, dtype=source.dtype).reshape(1, 1)
+        return ops.scaled_fp8_quant(probe)[1]
+
+    @staticmethod
+    def _collect_amax(slot: _QuantSlot, current: torch.Tensor) -> torch.Tensor:
+        if not slot.pending_rows:
+            return current
+        slot.ready.synchronize()
+        slot.pending_rows = 0
+        return torch.maximum(current, slot.host_amax)
```
RuixiangMa: Allocates a new tensor on every call. I think it can use torch.maximum(current, slot.host_amax, out=current) for in-place accumulation.

vllm_omni/diffusion/model_loader/host_weights/runtime_fp8.py
```diff
@@ -0,0 +1,75 @@
...
+    dp = int(getattr(parallel, "data_parallel_size", 1) or 1)
+    sp = int(getattr(parallel, "sequence_parallel_size", 1) or 1)
+    if not getattr(config, "host_weight_runtime_root", None):
+        return "host_weight_runtime_root is not configured"
+    if not getattr(config, "enable_distributed_layerwise_offload", False):
+        return "distributed layerwise offload is disabled"
+    if not getattr(config, "dlo_use_allgather", True) or max(dp, sp) == 1:
```
RuixiangMa: This covers two distinct cases:
- not dlo_use_allgather → AllGather is disabled
- max(dp, sp) == 1 → AllGather is enabled but single rank

The message says "requires multi-rank DLO AllGather" which is misleading for the second case (AllGather is enabled, but only one rank). Consider splitting into two separate messages.

vllm_omni/diffusion/model_loader/host_weights/producers/final_layout_fp8.py
```diff
@@ -0,0 +1,287 @@
...
+        source = self._source(record.name)
+        rows_per_chunk = max(
+            1,
+            self._quant_chunk_bytes // (source.shape[1] * source.element_size()),
+        )
+        rows_per_chunk = min(rows_per_chunk, source.shape[0])
+        slots = self._make_slots(source, record.tensor.dtype, rows_per_chunk)
+        scale = self._find_scale(source, slots, rows_per_chunk)
+        self._quantize_rows(output, record.name, source, slots, rows_per_chunk, scale)
+        host_scale = scale.cpu()
```
RuixiangMa: Synchronous D2H copy of a 1-element tensor. Negligible for a scalar, but it stalls the host. Could use scale.to("cpu", non_blocking=True) with a synchronize later (e.g., before _release_source_pages).

vllm_omni/diffusion/model_loader/diffusers_loader.py
```diff
@@ -495,12 +497,15 @@ def load_model(
             if not is_offline:
                 load_device = device.type
                 offload_after_quant = True
-                logger.info(
```
RuixiangMa, line 498: These comments/logs are still relevant to the non-runtime fallback path (the elif not self._force_canonical_load: branch) that still exists in the code.

baonudesifeizhai: restored

vllm_omni/diffusion/model_loader/host_weights/producers/final_layout_fp8.py
baonudesifeizhai: Fixed by explicitly sorting FP8 production records before sharding. Each fp8_weight is now ordered immediately before its corresponding fp8_scale, independent of nn.Module._parameters registration order. The generated scale map is retained across shard boundaries, so weight/scale pairs split across files are also handled correctly.

vllm_omni/diffusion/model_loader/host_weights/fp8_layout.py
baonudesifeizhai: Fixed with an eligibility check before HWR resolution. The zero-copy lease-backed view is allowed only when the selected FP8 kernel inherits the base no-op post-load implementation, or when it uses the original Cutlass post-load implementation with fully 16-aligned weights. Marlin, Humming, unaligned Cutlass, and other kernels requiring post-load storage transforms fall back in preferred mode or fail in required mode. This preserves the zero-copy path without exposing immutable lease storage to in-place repacking.

vllm_omni/diffusion/model_loader/host_weights/producers/final_layout_fp8.py
```diff
@@ -0,0 +1,309 @@
...
+                        output.write_tensor(record.name, scale)
+                    elif record.role == "fp8_weight":
+                        scale_name = f"{record.name.removesuffix('.weight')}.weight_scale"
+                        generated_scales[scale_name] = self._write_fp8_weight(output, record)
+                    else:
+                        self._write_checkpoint_tensor(output, record)
+
+        if generated_scales:
+            raise ValueError(f"unwritten FP8 scales: {sorted(generated_scales)[:5]}")
+        return ProductionMetadata(
```
lishunyang12, line 146: Check source snapshots before and after FP8 production, before publishing the artifact.

hsliuustc0106 (approved):
All three open threads are resolved in de8cf24d — verified at this head: the deterministic weight-then-scale production order fixes the sharding mismatch, scale production is now event-ordered across the pool streams with a pinned non-blocking host copy (plus non-finite/zero amax guards), and the zero-copy lease is gated on storage-preserving kernels with the DLO/HWR design pages updated to match the real contract. The new 530-line producer test suite covers the gap I flagged. Remaining items are mechanical: rebase onto current main (branch conflicts right now), and it would be good to see the new FP8 tests ride a GPU lane on the rebased head.

baonudesifeizhai:
Mode | Startup Time | vs. Baseline
-- | -- | --
Baseline (HWR disabled) | 199s | —
Cold production | 207s | +8s / +4.0%
Warm local hit | 162s | -37s / -18.6% / 1.23× speedup

PR #6516, opened 2026-08-23
[Model] Support SenseNova-U1.5-8B-MoT and its distilled 8-step LoRA

### LINKED ISSUE #6471: [New Model]: SenseNova-U1.5-8B-MoT
### The model to consider.

https://huggingface.co/sensenova/SenseNova-U1.5-8B-MoT

### The closest model vllm-omni already supports.

Bagel?

### What's your difficulty of supporting the model you want?

_No response_

### Use case and motivation

_No response_

MrlixiangWE:
## Purpose

FIX #6471

SenseNova-U1.5-8B-MoT runs on the existing pipeline, but its distilled 8-step LoRA loaded
without error and changed no weight. This fixes that, and cuts denoise latency.

## Propose

1. `--lora-backend distill` was accepted and then warned `Pipeline does not support loading
   distilled LoRA weights for now`, generating without the adapter. Implement `load_lora_weights`
   on `SenseNovaU1Pipeline`, and raise when nothing matches instead of loading a no-op.
2. `_prepare_lora_delta` picked stacked mappings with `if param_name not in base_key` and did not
   stop at the first hit. `.qkv_proj` is a substring of `.qkv_proj_mot_gen`, so both rules fired
   and the delta came out twice the height of the parameter. Match on the tail and stop at the
   first hit. The fused names shipped by QwenImage and Wan2.2 do not overlap, so this selects the
   same shards for them; a test pins that.
3. Fold the adapter in fp32, so the low-rank product rounds to the parameter dtype once.
4. `Qwen3RMSNorm.forward` used an eager cast chain. Use `F.rms_norm`, which keeps the accumulation
   in fp32 through the weight multiply instead of rounding to bf16 first.
5. Annotate the four attributes `mypy` flags in `loader.py`; the file is now clean.
6. Add `recipes/SenseNova/SenseNova-U1.5.md`, and list U1.5 in supported-models and
   diffusion-features.

## Test Plan

```bash
cd tests && pytest diffusion/models/sensenova_u1/ diffusion/lora/test_loader.py -q
cd tests && pytest -m "core_model and cpu" -q

# 1x A800-80G, seed 42, one engine per arm, 1 warmup + 5 timed, medians
examples/offline_inference/text_to_image/text_to_image.py --model <U1.5> \
  --width 1024 --height 1024 --num-inference-steps 50 --cfg-scale 4.0
# and --num-inference-steps 8 --cfg-scale 1.0 --lora-path <8step> --lora-backend distill

vllm serve <U1.5> --omni --trust-remote-code [--lora-path <8step> --lora-backend distill]
```

### Latency, 1x A800-80G, seed 42, n=5 medians

| Case | before | after | change |
|---|---|---|---|
| 1024x1024, 50 steps, cfg 4.0 | 11726.9 ms | 10973.6 ms | **-6.42%** |
| 1536x1536, 50 steps, cfg 4.0 | 25674.1 ms | 23737.9 ms | **-7.54%** |
| 1024x1024, 8 steps, cfg 1.0, LoRA | 1114.4 ms | 1024.2 ms | **-8.09%** |

Peak VRAM unchanged, 34364 -> 34340 MiB. Five-run spread 0.3-3.7%.

### Distilled LoRA, 1024x1024, 8 steps, cfg 1.0

| Case | p50 | five-run range |
|---|---|---|
| main, `--lora-path` given | 1114.4 ms | 1108-1139 |
| main, no `--lora-path` | 1109.1 ms | 1106-1117 |
| this PR, `--lora-path` given | 1024.2 ms | 1023-1035 |
| this PR, no `--lora-path` | 1026.6 ms | 1018-1056 |

On main the first two rows differ by +0.48%: the adapter is dropped, so both runs use the base
weights. With this PR it is applied, 588 LoRA keys into 168 parameters, and still costs nothing
against the same-step control, -0.23%, inside the 1.16% and 3.70% run spreads.

### Accuracy of the RMSNorm change

Mean relative error against a float64 reference, same bf16/fp16 inputs:

| shape | dtype | eager chain | `F.rms_norm` |
|---|---|---|---|
| 4096x3584 | bf16 | 1.890e-3 | **1.409e-3** |
| 16384x4096 | bf16 | 1.886e-3 | **1.409e-3** |
| 16384x8192 | fp16 | 2.368e-4 | **1.761e-4** |

Lower in all 12 shape/dtype combinations tested: the eager path rounds to bf16 before the weight
multiply, `F.rms_norm` rounds once. CFG-parallel output is byte-identical to single GPU.

### Tests

| Suite | this PR | 3781214 |
|---|---|---|
| `tests/diffusion/models/` | 5 failed, 1867 passed, 4 errors | 5 failed, 1819 passed, 4 errors |
| `tests/diffusion/lora/` | 36 passed, 2 errors | 31 passed, 2 errors |
| L1 `core_model and cpu` | 137 failed, 9476 passed, 37 errors | 137 failed, 9423 passed, 37 errors |

Compared as failing-test-ID sets rather than counts: the two sides are identical in all three, at 7,
2 and 139 ids. The pass-count differences are the tests this PR adds. `sensenova_u1/` alone is 86
passed, and restoring the pre-PR implementation under these tests turns 20 of them red.

### Serving

| Server | text2text | img2text | text2img |
|---|---|---|---|
| full model | ok | ok | ok |
| + distilled LoRA | ok | ok | ok |

vLLM Version: 0.27.0
vLLM-Omni Commit: 378121467


MrlixiangWE:
Same prompt and seed, 1024x1024, on this branch.

| | 50 steps, cfg 4.0 | 8 steps, cfg 1.0, no LoRA | 8 steps, cfg 1.0, distilled LoRA |
|---|---|---|---|
| cat | cat_50step | cat_8step | cat_8step_lora |
| dog | dog_50step | dog_8step | dog_8step_lora |

10973.6 ms / 1026.6 ms / 1024.2 ms.

MrlixiangWE:
@hsliuustc0106 this PR is ready for review.

vllm_omni/diffusion/models/sensenova_u1/pipeline_sensenova_u1.py
```diff
@@ -1408,21 +1420,60 @@ def _run_denoising_loop(self, ns, caches, p, think_text="", is_it2i: bool = Fals
...
+        state_dict = _remap_state_dict_keys(
+            state_dict,
+            [(".lora_down.weight", ".lora_A.weight"), (".lora_up.weight", ".lora_B.weight")],
+        )
+        # Fold alpha and the low-rank product in fp32; scaling B in bf16 would
+        # round once before the matmul and once after.
+        state_dict = {k: v.to(torch.float32) for k, v in state_dict.items()}
+        state_dict = _apply_diffusers_lora_alpha_scaling(state_dict)
+
+        loaded_keys = self.load_lora_into_module(state_dict, self, prefix="")
```
hsliuustc0106: This breaks distilled LoRA with `--tensor-parallel-size 2`: both ranks load a 3,072-row local QKV weight, but `_prepare_lora_delta` builds a 6,144-row global delta and `params.add_` raises. Please slice the delta through vLLM's parameter weight-loader/TP metadata and add a TP=2 regression test.

tests/diffusion/models/sensenova_u1/test_sensenova_u1_rmsnorm.py
```diff
@@ -0,0 +1,79 @@
...
+    torch.manual_seed(0)
+    norm = Qwen3RMSNorm(hidden, eps=1e-6).to(dtype)
+    with torch.no_grad():
+        norm.weight.copy_(torch.rand(hidden, dtype=dtype) + 0.5)
+    x = torch.randn(7, hidden, dtype=dtype)
+    got = norm(x)
+    want = _reference(x, norm.weight, norm.variance_epsilon)
+    assert got.dtype == dtype
+    assert got.shape == x.shape
+    torch.testing.assert_close(got.to(torch.float32), want.to(torch.float32), atol=atol, rtol=atol)
```
hsliuustc0106, line 42: Restoring the pre-PR eager RMSNorm implementation still leaves all 14 tests in this file green, so the claimed low-precision rounding improvement is not regression-protected. Please add a deterministic bf16/fp16 case that distinguishes `F.rms_norm` from the legacy path, or compare both errors against the float64 reference.

vllm_omni/diffusion/models/sensenova_u1/pipeline_sensenova_u1.py
```diff
@@ -1408,21 +1420,60 @@ def _run_denoising_loop(self, ns, caches, p, think_text="", is_it2i: bool = Fals
...
+        if not loaded_keys:
+            raise ValueError(f"Distilled LoRA {paths[0]} matched no parameter of this pipeline.")
+        self.lora_loaded[adapter_name] = state_dict
```
hsliuustc0106: This keeps the full fp32 adapter state alive even though SenseNova has no unload path. The 814.9 MB BF16 file expands to 1.52 GiB of tensors and raised steady process-tree RSS by 1.95 GiB locally; keep only an adapter-name sentinel after one-way fusion unless unload is added.

hsliuustc0106 (changes requested):
Single-GPU validation passed (36 tests; full-model latency -11.7%), but TP=2 distilled LoRA startup crashes on both ranks.

hsliuustc0106:
I profiled one 1024×1024 step with Nsight Systems 2026.1.3 on an L20X, capturing only the third worker-side forward after the engine warmup and one user warmup.

| Path | Captured forward | Kernels | Kernel time | GPU busy | Launch API | Inter-launch gaps |
| Full, CFG 4 | 350.1 ms | 8,248 | 89.9 ms | 94.1 ms (26.9%) | 42.0 ms | 277.2 ms |
| No-LoRA control, CFG 1 | 254.3 ms | 6,372 | 56.7 ms | 60.3 ms (23.7%) | 31.8 ms | 194.7 ms |
| LoRA, CFG 1 (`a312a949`) | 285.4 ms | 6,372 | 56.9 ms | 58.2 ms (20.4%) | 34.6 ms | 225.9 ms |

Inter-launch gaps overlap GPU execution and are not additive to kernel time; they measure host intervals between consecutive launch APIs.

This is host-dispatch/kernel-launch limited, not device-compute limited:

- The two prefix forwards launch 4,204 kernels for only ~21.7 ms of GPU work, but take 143-171 ms on the host.
- CFG-1 denoise launches 1,877 kernels: ~34.1 ms GPU versus 71-76 ms host wall. CFG 4 nearly doubles this path to 3,753 kernels / 67.0 ms GPU.
- CFG-1 averages an 8.9 µs kernel, ~5 µs inside each launch API, and 31-35 µs between launches. The GPU is idle for 170-203 ms inside its active window.
- LoRA and no-LoRA CFG-1 have exactly the same kernel name/grid/block multiset, CUDA API counts, and memcpy signatures. The adapter adds no per-step kernel after fusion; the production gain comes from 8 steps + CFG 1.
- In the LoRA path, 2,119 copy/cast kernels spend 10.6 ms launching and only 5.6 ms executing; 900 elementwise-multiply kernels spend 4.3 ms launching and 2.9 ms executing.

Can you optimize the launch path before merge? The runtime currently reports that regional compilation is skipped because `SenseNovaU1ForCausalLM` has no `_repeated_blocks`. The first experiment I suggest is:

```python
_repeated_blocks = ["SenseNovaU1DecoderLayer"]
```

Please validate both UND/GEN specializations, resolutions, TP, and distilled LoRA, then provide the same-input A/B with:

1. kernel count and GPU-busy percentage;
2. latency mean ± stddev over at least three post-warmup runs;
3. output-quality parity with the same seed.

A successful change should materially reduce the 6,372 launches, especially copy/cast and elementwise kernels. If regional compile still graph-breaks or leaves the count unchanged, the next target should be CUDA Graph replay for stable resolution/sequence/CFG buckets. GEMM tuning alone has a low ceiling here because GEMMs are only about 11-15% of end-to-end wall time.

Nsight collection overhead was ~6 ms. The 9-11 second application timing printed during capture is invalid because `cudaProfilerStop` blocks while Nsight writes the report; the table uses timestamps inside the NVTX capture range.

MrlixiangWE:
Both suggestions are in. Taking the buckets first: they are not where the idle time is.

Corrected GPU busy — kernel time from the profiled run, wall clock from a clean one — is **92.0%** for text-to-image and **68.5%** for understanding/think. Replay over stable resolution/sequence/CFG buckets is worth at most ~7.5% there; the idle sits in AR decode, at 31.5%. Its attribution is 8.7% `cudaLaunchKernel`, 0.2% `cudaStreamSynchronize`, and **~86% inside no CUDA call at all** (Python/ATen dispatch), which rules out the `.item()` sync and selects CUDA Graphs.

The blocker was the cache. `DynamicCache.update` grows K/V with `torch.cat` on every step, so decode has no static shape. Padding to a bucket and masking the tail does give static shapes, but the mask costs more than the graph saves: 7.96-11.55 ms per step against 1.03 ms unmasked, at real decode shapes. A paged cache gives both — `flash_attn_varlen_func` takes the used length as a tensor (`seqused_k`) alongside a `block_table`, so the buffers stay bucket-sized and capturable while the kernel reads only the valid prefix, and one capture serves every length in the bucket.

## Whole PR, main -> this branch (1x A800-80G, seed 42, n=5, one session)

| Case | main | This branch | Change |
|---|---|---|---|
| think 1024x1024 | 17710.2 +/- 62.8 ms | 5424.2 +/- 10.1 ms | **-69.37%** |
| T2I 1024x1024, 8 steps, cfg 1.0, distilled LoRA | 1120.8 +/- 4.0 ms | 924.4 +/- 2.4 ms | **-17.52%** |
| T2I 1024x1024, 50 steps, cfg 4.0 | 11862.3 +/- 59.9 ms | 10560.7 +/- 26.5 ms | **-10.97%** |
| T2I 1536x1536, 50 steps, cfg 4.0 | 25779.6 +/- 48.3 ms | 23131.7 +/- 49.6 ms | **-10.27%** |

On main the distilled adapter is silently dropped, so that row is the user-facing before/after rather than a like-for-like speed comparison.

## Launch count, the criterion you set

Same input, one profiled generation of T2I 1024x1024, 8 steps, cfg 1.0, distilled LoRA.

| Measure | main | This branch | Change |
|---|---|---|---|
| `cudaLaunchKernel` | 24,244 | 7,782 | **-67.90%** |
| GPU kernels | 25,924 | 8,232 | **-68.25%** |
| Kernel time | 962.0 ms | 852.5 ms | -11.38% |
| GPU busy | 86.1% | 93.1% | +7.0 pt |

`cudaGraphLaunch` is 0 on both legs: the graph covers autoregressive decode, which text-to-image never runs. The reduction here is entirely the RoPE hoist, the uncompressed K/V and the dropped decode mask. Kernel count falls 68% while kernel time falls 11%, which matches your reading that these are ~9 us kernels whose cost is the launch, not the execution.

## Tensor parallelism (a312a94 -> this branch, 2x A800, SYS/PCIe, no NVLink)

| Case | Before | After | Change |
|---|---|---|---|
| think 1024x1024, TP=2 | 20678.3 +/- 28.8 ms | 5221.0 +/- 4.6 ms | **-74.75%** |
| T2I 1024x1024, 8 steps, cfg 1.0, distilled LoRA, TP=2 | 1234.7 +/- 25.9 ms | 1127.8 +/- 11.2 ms | **-8.66%** |

The baseline is a312a94 rather than main because this configuration is the one that crashed at startup before that commit. The T2I leg is bit-identical; think with the adapter loaded also captures and replays cleanly.

## Memory, same bytes with only `VLLM_OMNI_SENSENOVA_PAGED_DECODE` differing

| Measure | Off | On | Delta |
|---|---|---|---|
| Peak host RSS | 7345 MiB | 8146 MiB | +801 MiB, during capture |
| Resident after the run | 3589 MiB | 3612 MiB | **+23 MiB** |
| Peak VRAM, schedule pinned to the 8192 bucket | 35107 MiB | 35971 MiB | +864 MiB (+2.46%) |

## Everything I tried

| Change | Verdict | Evidence |
|---|---|---|
| Build the 3D RoPE tables once per forward | kept | 1,008 calls producing 17 distinct tables |
| Stop expanding K/V before the backend | kept | a +2.55% regression on its own; only pays once the mask is gone |
| Drop the all-zeros decode mask | kept | -5.66%, -8.19% together with the row above |
| Paged decode under a CUDA graph | kept | **-29.93% / -31.27%** on two prompts, same bytes, switch only, n=5 |
| `_repeated_blocks` regional compile (your first suggestion) | dropped | launches 19,974 -> 19,974, zero triton kernels, hits `recompile_limit (8)` and falls back to eager permanently, +2.9% |
| Removing the forced `sdpa_fallback` | dropped | init dies: the other backends invert a boolean mask and SenseNova passes an additive float one, so the line is load-bearing |
| Removing the `.item()` sync | dropped | 0.2% of the idle |
| Bucketing plus a mask | dropped | 7.96-11.55 ms per step against 1.03 ms unmasked |

## think, main vs this branch (same prompt, same seed)

think_main_vs_branch

think output text does change: paged FA2 and SDPA are equally accurate against a float64 reference (mean error identical to four significant figures at n = 271/512/900/1500/3000), but they agree with each other only to 7.63e-06, and in a 382-step argmax loop that eventually flips a token. All four T2I cases are bit-identical between a312a94 and this branch.

## Tests

`tests/diffusion/models/sensenova_u1/` — 84 passed, 16 new. Verified by reverting: dropping the `forward_und` branch turns exactly one test red, weakening the `flash_attn_varlen_func` capability probe turns one red, and restoring the Python index in `attend()` turns one red. Bucket size does not affect results — pinning the schedule to 512, 2048 or 8192 produces the same png — and growth plus re-capture is covered end to end with a subdivided schedule that triggers two consecutive growths.

hsliuustc0106:
The TP LoRA sharding, retained adapter state, and RMSNorm regression test are addressed at `e96b5cba`. Two issues remain from local re-verification:

- `pipeline_sensenova_u1.py:872`: `_generate_text` is the T2T/I2T decode loop, but it calls `_ar_step` without the decode context that `_generate_think` creates. The recipe therefore claims paged/CUDA-graph acceleration for image-to-text even though that path never reaches it. Please wire the context into `_generate_text` and add a path-level regression test, or narrow the documentation.
- `test_sensenova_u1_attention_gqa.py:185`: on one reserved L20X with PyTorch 2.13.0+cu129 and vLLM 0.27.0, the new CUDA suite reports `4 passed, 1 failed`; at `kv_len=512`, the maskless GQA path is closer to float64 in only 7/8 fixed-seed trials. The current 8/8 assertion does not hold across this supported CUDA stack. Please use an explicit aggregate/no-regression tolerance (or document and enforce the intended hardware/kernel support) and rerun the accuracy gate.

I stopped the full-model A/B at that failed feasibility gate, so the reported switch-only paged-decode speedup remains unconfirmed locally.

vllm_omni/diffusion/models/sensenova_u1/pipeline_sensenova_u1.py
MrlixiangWE: Fixed in a312a94 — the delta is written through the parameter's own weight loader, so each rank takes only its shard. `test_fused_delta_is_sharded_through_the_weight_loader` pins it on CPU, and TP=2 with the distilled LoRA now runs on two A800s.

tests/diffusion/models/sensenova_u1/test_sensenova_u1_rmsnorm.py
MrlixiangWE: You were right that the suite stayed green with the old implementation restored. `test_rmsnorm_is_closer_to_float64_than_the_cast_chain` now measures both implementations against the float64 reference and asserts the fused one is closer, so restoring the cast chain fails it.

vllm_omni/diffusion/models/sensenova_u1/pipeline_sensenova_u1.py
MrlixiangWE: Fixed in a312a94 — only an adapter-name sentinel survives the one-way fusion. `test_fusion_does_not_retain_the_adapter_tensors` asserts what is kept is neither a dict nor a tensor, and that a second load is still a no-op.

hsliuustc0106:
Rechecked `9e5bf5b7` locally. Both issues from my previous comment are fixed.

Environment: NVIDIA L20X, PyTorch 2.13.0+cu129, vLLM 0.27.0. All runs used one excluded warmup followed by three measured requests.

| Comparison | Baseline | Current | Delta |
|---|---:|---:|---:|
| Current head, graph off → on | 13.900 ± 0.112 s | 7.566 ± 0.131 s | **-45.57%** |
| `a312a949` → current, TP=1 | 36.835 ± 0.262 s | 7.566 ± 0.131 s | **-79.46%** |
| `a312a949` → current, TP=2 | 55.171 ± 0.329 s | 6.182 ± 0.293 s | **-88.79%** |

Timed NVML peak increased by 366 MiB (+1.01%) for TP=1 and 402 MiB/GPU (+2.02%) for TP=2. The real TP=2 distilled-LoRA path initialized and generated successfully on both ranks.

The updated suites also pass locally: 16 CPU tests and 5 CUDA tests. AR output diverges across implementations, so image hashes/SSIM are not parity metrics, but all compared images were visually sharp and prompt-faithful. This confirms the performance claims and keeps memory below the 5% regression gate.

vllm_omni/diffusion/models/sensenova_u1/paged_decode.py
```diff
@@ -0,0 +1,276 @@
...
+    def _forward(self):
+        return self.lm(
+            input_ids=self.input_ids,
+            indexes=self.indexes,
+            past_key_values=None,
+            use_cache=False,
+            paged_cache=self.cache,
+        ).logits
+
+    def _capture(self, key):
```
hsliuustc0106, line 337: In a fresh-cache feasibility run, engine readiness completed before this path was exercised. The first graph-enabled think request took 76.9 s, with the first capture occurring about 69 s after request start; subsequent post-warmup requests were 7.566 ± 0.131 s.

Can we warm a representative AR decode/capture before reporting readiness, or document and measure this cold first-request penalty separately? The current dummy diffusion warmup does not appear to cover paged decode.

vllm_omni/diffusion/models/sensenova_u1/paged_decode.py
```diff
@@ -0,0 +1,276 @@
...
+
+def _bucket_for(length: int) -> int:
+    for b in BUCKETS:
+        if length <= b:
+            return b
+    # Past the schedule, round up to a whole number of blocks.
+    return ((length + BLOCK_SIZE - 1) // BLOCK_SIZE) * BLOCK_SIZE
+
+
+class PagedDecodeCache:
```
hsliuustc0106, line 136: Can you document why this allocation remains model-local instead of using `DiffusionKVCacheManager` or `ARDiffusionModelRunner`?

`PagedDecodeCache` owns per-request tensors and an identity block table, so it bypasses scheduler admission, pooling, eviction, prefix reuse, and continuous-batching lifecycle. That is reasonable for the current `max_num_seqs=1` path, but please state the limitation and link a follow-up/deletion condition before this is treated as general vLLM paged-KV support.

hsliuustc0106:
image

vllm_omni/diffusion/models/sensenova_u1/paged_decode.py
MrlixiangWE: The tens of seconds is the language model's `torch.compile`, and where it lands depends on whether the startup warmup reaches that path. One session on an A800, inductor / triton / vLLM / CUDA JIT caches cleared for every arm, three requests each. "full" is `diffusion_compile_granularity=full` and "fixed" adds the warmup from 85acb30.

| Warmup | Compile | Readiness | First request | Steady | Above steady |
|---|---|---:|---:|---:|---:|
| active | default | 38.7 s | 6.12 s | 5.41 s | 0.71 s |
| active | full | 68.8 s | 13.13 s | 5.35 s | 7.77 s |
| active | full, fixed | 75.2 s | 10.44 s | 5.41 s | 5.03 s |
| **skipped** | full | 35.2 s | **45.67 s** | 5.48 s | **40.19 s** |
| skipped | default | 35.6 s | 8.78 s | 5.43 s | 3.35 s |

The fourth row is the shape you saw: readiness completes, then the first request pays tens of seconds. Your 76.9 s against a 7.566 s steady state is 69.3 s above steady, which matches the 69 s you measured to the first capture, and is 9.2x steady; that row is 40.19 s above steady, 7.3x. With the warmup active the same work is absorbed into readiness instead, so a run showing tens of seconds inside the request is one the warmup reached none of.

The default configuration compiles nothing at all: granularity is `regional` and `regionally_compile` returns early without `_repeated_blocks`, so the runner's "configured for lazy torch.compile" line is misleading.

85acb30 makes the dummy run warm the decode shape, which the text-to-image dummy request never touches: rows two and three, 7.77 s above steady before and 5.03 s after. It runs on both ranks under tensor parallelism.

A warmup cannot cover the capture half of your question: `_decode_context` builds the cache and the runner per request, so a startup capture is bound to buffers the first real request replaces — about 188 ms of every request rather than a one-off.

vllm_omni/diffusion/models/sensenova_u1/paged_decode.py
MrlixiangWE: Documented in the module docstring at 9e0561c, with the deletion condition: once the loop is driven by a scheduler-visible runner, or the pipeline serves more than one sequence per forward, this file goes and the manager takes over.

MrlixiangWE:
@hsliuustc0106 ready for review.

vllm_omni/diffusion/models/sensenova_u1/pipeline_sensenova_u1.py
```diff
@@ -709,8 +728,44 @@ def _apply_time_schedule(self, t, image_seq_len, timestep_shift):
         sigma = shift * sigma / (1 + (shift - 1) * sigma)
         return 1 - sigma
 
-    def _ar_step(self, next_token, t_idx, past_key_values):
+    def _decode_context(self, past_key_values):
```
hsliuustc0106, line 732: The paged path defaults on here and changes which SDPA kernel runs, but nothing compares end-to-end outputs with `VLLM_OMNI_SENSENOVA_PAGED_DECODE=0` — the unit tests pin kernel-level equivalence only and the recipe records just latency deltas. A same-seed A/B (think/text token ids identical; image similarity or latent hash with a stated tolerance) would close that gap, and the same pair can attribute how much of the -6-8% denoise win comes from this versus the RMSNorm/RoPE changes.

recipes/SenseNova/SenseNova-U1.5.md
```diff
@@ -0,0 +1,124 @@
+# SenseNova-U1.5
```
hsliuustc0106, line 1: recipes/README.md has no row for this recipe yet — model-addition policy requires each in-repo recipe to appear in the Available Recipes table; an entry mirroring the U1 row would do. Dropping TEMPLATE.md's `References` section is fine if you note the deviation in the PR body.

vllm_omni/diffusion/models/sensenova_u1/sensenova_u1_transformer.py
```diff
@@ -303,23 +314,18 @@ def __init__(self, config, layer_idx: int, quant_config=None, prefix: str = ""):
...
-        t_config = copy.deepcopy(config)
-        t_config.head_dim = config.head_dim // 2
-        self.rotary_emb = Qwen3RotaryEmbedding(config=t_config)
-
-        hw_config = copy.deepcopy(config)
-        hw_config.head_dim = config.head_dim // 4
-        hw_config.rope_theta = config.rope_theta_hw
-        hw_config.max_position_embeddings = config.max_position_embeddings_hw
-        self.rotary_emb_hw = Qwen3RotaryEmbedding(config=hw_config)
+        self.rotary_emb, self.rotary_emb_hw = _build_3d_rope(config)
```
hsliuustc0106: Now that the three RoPE tables are built once per forward on the model (:679) and threaded down via `position_embeddings`, these per-layer `Qwen3RotaryEmbedding` pairs are dead state: their only consumer is the `position_embeddings=None` fallback in `_project_and_rope`, which no in-tree caller reaches (`indexes` is asserted non-None). Dropping them removes a module tree per layer plus the config deepcopies at init.

vllm_omni/diffusion/models/sensenova_u1/pipeline_sensenova_u1.py
hsliuustc0106: Verified all five at be42af5 — the delta now routes through `_prepare_lora_delta` per stacked shard (with the loader test), the RMSNorm test asserts the fused path is closer to float64, only the adapter-name sentinel survives fusion, the cache-cleared cold-start table plus the 85acb30 warmup covers the readiness question, and 9e0561c documents the paged-cache scope and deletion condition. Thanks. Two things remain: my 08-27 A/B-parity comment is still open, and 0d68f32 adds ~460 lines of unreviewed 8B-MoT support to this PR — intentional, or should that land as a follow-up PR?

MrlixiangWE: Intentional. U1.5 shares `SenseNovaU1Pipeline` with U1 -- same
`model_type: neo_chat`, and `SenseNovaU1Config` already reads `use_pixel_head`
and `noise_scale_max_value`. Supporting it is making the distilled LoRA path
correct. The 460 lines:

| File | + | What |
|---|---:|---|
| tests (3 files) | 258 | tests |
| `recipes/SenseNova/SenseNova-U1.5.md` | 114 | recipe |
| `pipeline_sensenova_u1.py` | 67 | `load_lora_weights`, `stacked_params_mapping` |
| `vllm_omni/diffusion/lora/loader.py` | 14 | `_prepare_lora_delta` fix |
| `sensenova_u1_transformer.py` | 5 | `F.rms_norm` |
| two doc table rows | 2 | supported-models, diffusion-features |

Tests on the latest commit are running here; I will reply on each thread when
they finish, about six hours if nothing breaks.

vllm_omni/diffusion/models/sensenova_u1/paged_decode.py
```diff
@@ -0,0 +1,326 @@
...
+        device tensors, so a captured graph reads whatever they hold at replay
+        rather than whatever they held at capture.
+        """
+        flat_k = self.k[layer_idx].view(-1, self.kv_heads, self.head_dim)
+        flat_v = self.v[layer_idx].view(-1, self.kv_heads, self.head_dim)
+        flat_k.index_copy_(0, self.pos, key_bhsd[0, :, 0].unsqueeze(0))
+        flat_v.index_copy_(0, self.pos, value_bhsd[0, :, 0].unsqueeze(0))
+        heads, head_dim = query_bhsd.shape[1], query_bhsd.shape[3]
+        q = query_bhsd[0].transpose(0, 1).reshape(1, heads, head_dim)
+        out = _flash_varlen()(
```
RuixiangMa, line 278: All other call sites in the repo use keyword arguments exclusively. The paged-specific kwargs (seqused_k, block_table) are correctly passed as keywords, but the standard args (max_seqlen_q, cu_seqlens_q, max_seqlen_k) are positional. This works on the current vLLM version but risks silent breakage on version changes.

vllm_omni/diffusion/models/sensenova_u1/pipeline_sensenova_u1.py
```diff
@@ -1408,21 +1522,121 @@ def _run_denoising_loop(self, ns, caches, p, think_text="", is_it2i: bool = Fals
...
+                    for delta, shard_id in deltas:
+                        delta = delta.to(device=param.device, dtype=param.dtype)
+                        if shard_id is None:
+                            weight_loader(param, delta)
+                        else:
+                            weight_loader(param, delta, shard_id)
+                finally:
+                    param.data = saved
+            with torch.no_grad():
+                param.data.add_(buffer)
```
RuixiangMa, line 1648: param.data.add_(buffer) modifies in-place per parameter. If parameter N fails, parameters 1..N-1 are already fused. Low impact (load-time only, restart recovers), but consider a pre-pass or document the behavior.

vllm_omni/diffusion/models/sensenova_u1/paged_decode.py
```diff
@@ -0,0 +1,326 @@
...
+
+    def grow(self, length: int) -> bool:
+        """Move to the next bucket, keeping the tokens already stored."""
+        new_bucket = _bucket_for(length)
+        if new_bucket == self.bucket:
+            return False
+        old_k, old_v, old_len = self.k, self.v, self._length
+        nblocks = new_bucket // BLOCK_SIZE
+        shape = (nblocks, BLOCK_SIZE, self.kv_heads, self.head_dim)
+        self.k = [torch.zeros(shape, device=self.device, dtype=self.dtype) for _ in old_k]
```
hsliuustc0106: `grow()` mutates the live K/V addresses before bumping `generation`. I fault-injected a recoverable allocation failure between these assignments: K changed, V/bucket/generation did not, and the old graph key remained selected. A later request can therefore replay a graph captured against freed or reused storage. Please allocate and copy K, V, and the block table into locals, then commit them and increment `generation` only after the resize succeeds; add a regression test for this failure path.

vllm_omni/diffusion/models/sensenova_u1/pipeline_sensenova_u1.py
```diff
@@ -709,8 +728,66 @@ def _apply_time_schedule(self, t, image_seq_len, timestep_shift):
...
+        if past_key_values is None or not past_key_values.layers:
+            return None
+        layer0 = past_key_values.layers[0]
+        if layer0.keys is None:
+            return None
+        _, kv_heads, prefix, kv_head_dim = layer0.keys.shape
+        device = torch.device(self.device)
+        num_layers = len(self.language_model.model.layers)
+        decode = getattr(self, "_paged_decode", None)
+        if decode is None or not decode[0].reusable_for(
```
hsliuustc0106, line 767: This reuse decision ignores request-level PEFT state. `DiffusionWorker` calls `lora_manager.set_active_adapter(...)` before each request, while `LoRALinearLayer.apply()` uses Python control flow to gate the LoRA matmuls; CUDA capture freezes whichever branch ran during capture. A base capture can omit LoRA for the next adapted request, and an adapted capture can survive disable, adapter-switch, or scale changes. Please invalidate or rebuild graphs when adapter identity, scale, or buffer generation changes, or explicitly reject dynamic PEFT for SenseNova; add a base → LoRA → base sequential regression.

hsliuustc0106 (commented):
All five prior fix claims verified at be42af5 (delta sharded through `_prepare_lora_delta` with the loader test, RMSNorm float64-reference regression test, adapter-name sentinel only, cache-cleared cold-start table + 85acb30 warmup, paged-cache scope docstring), and the recipe is now linked from the README table. Two opens before approval from my side: the 08-27 paged-vs-eager A/B parity comment is still unanswered, and commit 0d68f32 adds ~460 lines of 8B-MoT support that postdates my review — please confirm whether it belongs in this PR or a follow-up.

MrlixiangWE:
Intentional — per-file breakdown at https://github.com/vllm-project/vllm-omni/pull/6516#discussion_r3877923430.

Parity numbers for the 08-27 comment are running; I will answer on that thread.

vllm_omni/diffusion/models/sensenova_u1/pipeline_sensenova_u1.py
MrlixiangWE: Same seed, `VLLM_OMNI_SENSENOVA_PAGED_DECODE` the only difference.

| case | on vs off |
|---|---|
| t2i, think off | byte-identical |
| t2i + 8-step distilled LoRA | byte-identical |
| t2i, think on | PSNR 39.88 dB, mean abs 1.31 |
| i2t, counting | byte-identical, 827 chars |
| i2t, describe | 774/770 chars, identical to char 261 |
| t2t | 945/936 chars, identical to char 466 |

Think decodes 257 tokens identically both ways, `max |Δlogit| = 0.72` over all of
them. Tracing the text loop the same way, on a separate run: counting is identical
end to end; describe and t2t part at a step whose top1-top2 margin is 0.000 and
0.125, against a `|Δlogit|` of 0.188 and 0.164 there.

Attribution, 1024x1024, 50 steps, cfg 4.0, think off, n=5 median, one run:

| | median | delta |
|---|---:|---:|
| main | 11859.5 ms | |
| branch, paged decode off | 10543.1 ms | -11.10% |
| branch, paged decode on | 10544.7 ms | +0.02% |

The denoise win is the RMSNorm / RoPE / K-V / mask changes, not paged decode.

vllm_omni/diffusion/models/sensenova_u1/sensenova_u1_transformer.py
MrlixiangWE: Fixed in 806626c. The fallback went with them, so `position_embeddings` is now
a required argument rather than a kwarg with a default.

vllm_omni/diffusion/models/sensenova_u1/paged_decode.py
MrlixiangWE: Fixed in 3149b71. The buffers, the copies and the block table are built into
locals and published together with the generation.
`test_a_failed_grow_leaves_the_cache_on_its_old_buffers` injects a failure into
the last allocation and asserts K, V, the block table, the bucket, the generation
and the length all held; reverting the publish turns exactly that test red and
nothing else.

vllm_omni/diffusion/models/sensenova_u1/pipeline_sensenova_u1.py
MrlixiangWE: Fixed in d7aae00. Reuse is off once `BaseLayerWithLoRA` wrappers exist under the
language model, so nothing captured crosses a request while the manager can bind,
rescale or reset slot 0. The distilled LoRA keeps the cross-request capture: it
fuses one-way and installs no wrapper, measured.
`test_no_capture_crosses_a_request_once_lora_wrappers_are_installed` is the
base -> LoRA -> base sequence, and TP=2 think and TP=2 distilled LoRA both run
clean on two A800s.

vllm_omni/diffusion/models/sensenova_u1/paged_decode.py
MrlixiangWE: Fixed in dcd1109. Every argument is a keyword now, and the capability probe
covers all seven instead of just `seqused_k` and `block_table` -- a wheel that
renamed one of the others would otherwise pass the probe and raise TypeError at
the first decode step.

vllm_omni/diffusion/models/sensenova_u1/pipeline_sensenova_u1.py
MrlixiangWE: Documented in e32770a. A pre-pass would have to hold every delta at once and
they are full rank: 168 parameters, 15.09 GiB for the 8B checkpoint against the
192 MiB one costs today. Fusion runs at load time, before the pipeline serves
anything, so a partially fused model is never reachable from a request.

recipes/SenseNova/SenseNova-U1.5.md
MrlixiangWE: Fixed in 8fb8683.

vllm_omni/diffusion/models/sensenova_u1/pipeline_sensenova_u1.py
hsliuustc0106: Delta round at e32770a — reviewed the four commits since my last pass (3149b71 one-step resize publish, d7aae00 LoRA-wrapper capture gating, dcd1109 keyword-only flash-attn args, e32770a fusion-failure doc); all clean, and the fault-injected `test_a_failed_grow_leaves_the_cache_on_its_old_buffers` is a nice touch. The A/B-parity ask is resolved for me: the same-seed table above plus 8fb86835's whole-loop decode comparison is exactly the durable coverage I wanted — please push that commit into this PR so the test and the recipe index are actually in-tree. Still outstanding from my 08-28 review: is 0d68f32's ~460-line 8B-MoT support intentional for this PR, or should it land as a follow-up?

MrlixiangWE: Intentional. This PR closes #6471, which asks for SenseNova-U1.5-8B-MoT support; the model already runs on the existing pipeline, and 0d68f32, the PR's first commit, is what makes its distilled LoRA correct, so it belongs here rather than in a follow-up. Per-file breakdown in https://github.com/vllm-project/vllm-omni/pull/6516#discussion_r3877923430.

8fb8683 is in the PR already, commit 17 of 25: `recipes/README.md` +1/-0 and the whole-loop decode test +53/-0.

MrlixiangWE:
@hsliuustc0106 0d68f32's title reads as new model support, which is the confusion. What the commit actually contains:

| file | + | what |
|---|---:|---|
| `test_sensenova_u1_distill_lora.py` | 124 | distilled LoRA fusion tests |
| `recipes/SenseNova/SenseNova-U1.5.md` | 114 | recipe |
| `test_sensenova_u1_rmsnorm.py` | 79 | RMSNorm tests |
| `pipeline_sensenova_u1.py` | 67 | `LoraLoaderMixin`, `stacked_params_mapping`, `load_lora_weights` |
| `test_loader.py` | 55 | regression for the `_prepare_lora_delta` substring bug |
| `loader.py` | 14 | `_prepare_lora_delta` matches on the tail and stops at the first hit |
| `sensenova_u1_transformer.py` | 5 | `F.rms_norm` |
| two doc table rows | 2 | supported-models, diffusion-features |

258 test, 116 doc, 86 implementation. The implementation is the LoRA path and the RMSNorm change — the model already ran on the existing pipeline before this PR, so there is no architecture code here to move.

MrlixiangWE:
@hsliuustc0106 On the ~460 lines you place after your review: my other guess is that you mean 6194f41 → be42af5 — why the PR went from 1661 to 2008 lines. That push was nine commits:

| commit | + | what |
|---|---:|---|
| 806626c | 16 | your comment: drop the per-layer rotary modules |
| 8fb8683 | 54 | your comments: whole-loop decode test, recipe index |
| b268901 | 3 | register the paged-decode env var in the inventory |
| 7226cd7 | 32 | grow the bucket schedule in steps past its last entry |
| a0ff6d5 | 61 | pin the write slot under capture |
| 3a7ff67 | 8 | re-measure the recipe figures |
| 1d9168d | 66 | **VRAM leak fix**: capture into the shared platform pool |
| e6c99fc | 120 | **VRAM leak fix**: reuse the cache and the runner across requests |
| be42af5 | 49 | **VRAM leak fix**: release those captures on sleep level 2 |

The last three are a VRAM leak this PR introduced. The decode runner was built per request, so every request captured its own graph and kept it, leaking around 40 MiB of device memory per request over 12 sequential think requests and never levelling off. 1d9168d moves capture to the shared platform pool that every other captured path in the tree uses; e6c99fc removes the leak by reusing the cache and the runner across requests, down to 0.2 MiB per request; be42af5 releases those captures on sleep level 2, which e6c99fc made necessary.

A regression this PR introduces belongs in this PR — splitting it out would leave this one leaking around 40 MiB of device memory per request.

vllm_omni/diffusion/models/sensenova_u1/pipeline_sensenova_u1.py
hsliuustc0106: Conceded — my framing was wrong on both counts: 0d68f32 was this PR's first commit (the #6471 support itself, predating my review), and the growth I attributed to it was actually the nine 6194f41→be42af5 commits answering review threads. With the scope question resolved and the A/B coverage question reduced to pushing 8fb86835, the only remaining item from my side is that push. Thanks for the detailed accounting.

hsliuustc0106: One more ask before approval: an e2e stage-attribution table in the same format #6714 and #6607 already use, so the paged-decode and LoRA claims are isolated rather than blended. The `VLLM_OMNI_SENSENOVA_PAGED_DECODE` switch makes this a clean A/B on the same head:

Conditions: 1×A800 80GB, same prompt + seed + resolution across columns, 2 warmups then N≥10 measured, P50 (P100). Baseline = main `<sha>`, candidates = head `<sha>` with the switch as shown.

| Metric | main eager (50-step) | PR paged=0 (50-step) | PR paged=1 (50-step) | PR paged=1 + LoRA (8-step) |
|---|---:|---:|---:|---:|
| Requested inference steps | 50 | 50 | 50 | 8 |
| `e2e_total_ms` | | | | |
| Stage 0 — AR prefill + decode (incl. think decode, tokens) | | | | |
| Stage handoff (e2e − stages) | | | | |
| Stage 1 — diffusion execution | | | | |
| Postprocess | | | | |
| MP4/PNG response encoding | | | | |
| Approx request → output | | | | |
| Sanity: output pixels / generated tokens | identical | identical | identical | identical |

main vs paged=0 proves the loading/fusion changes cost nothing at equal steps; paged=0 vs paged=1 isolates the paged-KV + CUDA-graph decode gain (and doubles as the scheduler-overhead measurement for the model-local-cache question); the LoRA column separates the step-reduction claim. A small companion table — requests that hit a new decode bucket and the max capture stall — would keep the mid-serving re-capture tail honest, since P50s hide it.

MrlixiangWE:
@hsliuustc0106 Stage attribution. One note on the shape first: `stage_metrics` reports a single stage named `diffusion` for this model, because `SenseNovaU1Pipeline`, which U1.5 shares with U1, runs its whole autoregressive loop inside one pipeline forward and the decode steps never reach the scheduler — the same fact behind the model-local cache. The AR/diffusion split below is therefore measured inside that stage, not across two of them.

Conditions: `sensenova/SenseNova-U1.5-8B-MoT`, 1×A800 80GB, TP=1, same prompt + seed 42 + 1024×1024 across columns, think on in every column, 2 warmups then N=10 measured, P50 (P100). Baseline = main `262ff14`, candidates = head `e32770a` with the switch as shown.

| Metric | main eager (50-step) | PR paged=0 (50-step) | PR paged=1 (50-step) | PR paged=1 + LoRA (8-step) |
|---|---:|---:|---:|---:|
| Requested inference steps | 50 | 50 | 50 | 8 |
| `e2e_total_ms` | 27.747 s (28.159) | 16.500 s (16.702) | 13.930 s (13.944) | 4.455 s (4.485) |
| Stage 0 — AR prefill + think decode | 16.281 s (16.681) | 6.283 s (6.461) | 3.604 s (3.612) | 3.593 s (3.603) |
| &nbsp;&nbsp;· AR prefill | 69.0 ms (78.0) | 38.9 ms (39.1) | 39.3 ms (39.5) | 39.0 ms (41.9) |
| &nbsp;&nbsp;· think decode | 16.146 s (16.545) | 6.213 s (6.391) | 3.534 s (3.539) | 3.522 s (3.533) |
| Stage 0 — diffusion execution | 11.460 s (11.492) | 10.251 s (10.271) | 10.317 s (10.335) | 0.854 s (0.877) |
| Stage handoff (e2e − stage_gen) | 1.5 ms (2.4) | 1.5 ms (2.1) | 1.3 ms (1.5) | 1.0 ms (1.3) |
| Postprocess | 0.0 ms | 0.0 ms | 0.0 ms | 0.0 ms |
| Sanity: output pixels | 1048576 | 1048576 | 1048576 | 1048576 |
| Sanity: think chars | 1233 | 1233 | 1233 | 1233 |

Isolated that way:

| Comparison | e2e | think decode | diffusion |
|---|---:|---:|---:|
| main → paged=0 (loading/fusion, RMSNorm, RoPE) | -40.53% | -61.52% | -10.56% |
| paged=0 → paged=1 (paged KV + CUDA graph) | -15.57% | -43.11% | +0.65% |

The earlier same-seed A/B ran with think off, where the decode loop this switch serves barely executes — hence +0.02% there against -15.57% here.

Stage handoff stays at 1.0–1.5 ms across all four, which is also the scheduler-overhead number for the model-local-cache question.

There is no PNG encoding row and no separate request → output row: the offline path returns the image in-process, so nothing encodes a response and `e2e_total_ms` is that number.

### Capture tail

Every capture timed individually, since P50s hide it. `paged=1`, 1 warmup + 5 measured per row.

| workload | requests | captures | bucket growths | max capture stall |
|---|---:|---:|---:|---:|
| T2I + think | 5 | 2 | 1 | 29.3 ms |
| edit, 2 images | 5 | 1 | 0 | 28.9 ms |
| image-to-text | 5 | 1 | 0 | 30.1 ms |

Three prefill lengths against the same buckets `(512, 1024, 2048, 4096, 8192)`, to reach the growth path rather than report a stall we never triggered:

| prefill | bucket growths | captures | max stall | trail |
|---|---:|---:|---:|---|
| 237 chars | 1 | 2 | 30.1 ms | `512 → 1024 at length 513` |
| 2007 chars | 1 | 1 | 34.0 ms | `1024 → 2048 at length 1025` |
| 4367 chars | 0 | 0 | — | starts in a bucket wide enough |

One re-capture costs 30–34 ms against a 14 s request. Five edit requests captured once between them, which is the cross-request reuse.

### The other dispatch entry points

`forward` dispatches three ways, and the table above covers `_forward_t2i` only. The other two, same head, switch as the only difference, 1 warmup + 5 measured:

| entry point | workload | paged=1 | paged=0 | delta | output identical |
|---|---|---:|---:|---:|---|
| `_forward_it2i` | edit, 2 images | 24.050 s | 30.158 s | -20.25% | image, not compared here |
| `_forward_text` | image-to-text | 7.263 s | 12.532 s | -42.04% | no — 2031 vs 2190 chars |
| `_forward_text` | text-to-text | 7.014 s | 12.640 s | -44.51% | no — 2750 vs 2789 chars |

The two text divergences are the greedy-argmax ties already measured on the same-seed A/B at https://github.com/vllm-project/vllm-omni/pull/6516#discussion_r3881056024: the first differing step has a top1-top2 margin of 0.000 and 0.125 against a `|Δlogit|` of 0.188 and 0.164.

vllm_omni/diffusion/models/sensenova_u1/pipeline_sensenova_u1.py
MrlixiangWE: Stage attribution. One note on the shape first: `stage_metrics` reports a single stage named `diffusion` for this model, because `SenseNovaU1Pipeline`, which U1.5 shares with U1, runs its whole autoregressive loop inside one pipeline forward and the decode steps never reach the scheduler — the same fact behind the model-local cache. The AR/diffusion split below is therefore measured inside that stage, not across two of them.

Conditions: `sensenova/SenseNova-U1.5-8B-MoT`, 1×A800 80GB, TP=1, same prompt + seed 42 + 1024×1024 across columns, think on in every column, 2 warmups then N=10 measured, P50 (P100). Baseline = main `262ff14`, candidates = head `e32770a` with the switch as shown.

| Metric | main eager (50-step) | PR paged=0 (50-step) | PR paged=1 (50-step) | PR paged=1 + LoRA (8-step) |
|---|---:|---:|---:|---:|
| Requested inference steps | 50 | 50 | 50 | 8 |
| `e2e_total_ms` | 27.747 s (28.159) | 16.500 s (16.702) | 13.930 s (13.944) | 4.455 s (4.485) |
| Stage 0 — AR prefill + think decode | 16.281 s (16.681) | 6.283 s (6.461) | 3.604 s (3.612) | 3.593 s (3.603) |
| &nbsp;&nbsp;· AR prefill | 69.0 ms (78.0) | 38.9 ms (39.1) | 39.3 ms (39.5) | 39.0 ms (41.9) |
| &nbsp;&nbsp;· think decode | 16.146 s (16.545) | 6.213 s (6.391) | 3.534 s (3.539) | 3.522 s (3.533) |
| Stage 0 — diffusion execution | 11.460 s (11.492) | 10.251 s (10.271) | 10.317 s (10.335) | 0.854 s (0.877) |
| Stage handoff (e2e − stage_gen) | 1.5 ms (2.4) | 1.5 ms (2.1) | 1.3 ms (1.5) | 1.0 ms (1.3) |
| Postprocess | 0.0 ms | 0.0 ms | 0.0 ms | 0.0 ms |
| Sanity: output pixels | 1048576 | 1048576 | 1048576 | 1048576 |
| Sanity: think chars | 1233 | 1233 | 1233 | 1233 |

Isolated that way:

| Comparison | e2e | think decode | diffusion |
|---|---:|---:|---:|
| main → paged=0 (loading/fusion, RMSNorm, RoPE) | -40.53% | -61.52% | -10.56% |
| paged=0 → paged=1 (paged KV + CUDA graph) | -15.57% | -43.11% | +0.65% |

The earlier same-seed A/B ran with think off, where the decode loop this switch serves barely executes — hence +0.02% there against -15.57% here.

Stage handoff stays at 1.0–1.5 ms across all four, which is also the scheduler-overhead number for the model-local-cache question.

There is no PNG encoding row and no separate request → output row: the offline path returns the image in-process, so nothing encodes a response and `e2e_total_ms` is that number.

### Capture tail

Every capture timed individually, since P50s hide it. `paged=1`, 1 warmup + 5 measured per row.

| workload | requests | captures | bucket growths | max capture stall |
|---|---:|---:|---:|---:|
| T2I + think | 5 | 2 | 1 | 29.3 ms |
| edit, 2 images | 5 | 1 | 0 | 28.9 ms |
| image-to-text | 5 | 1 | 0 | 30.1 ms |

Three prefill lengths against the same buckets `(512, 1024, 2048, 4096, 8192)`, to reach the growth path rather than report a stall we never triggered:

| prefill | bucket growths | captures | max stall | trail |
|---|---:|---:|---:|---|
| 237 chars | 1 | 2 | 30.1 ms | `512 → 1024 at length 513` |
| 2007 chars | 1 | 1 | 34.0 ms | `1024 → 2048 at length 1025` |
| 4367 chars | 0 | 0 | — | starts in a bucket wide enough |

One re-capture costs 30–34 ms against a 14 s request. Five edit requests captured once between them, which is the cross-request reuse.

### The other dispatch entry points

`forward` dispatches three ways, and the table above covers `_forward_t2i` only. The other two, same head, switch as the only difference, 1 warmup + 5 measured:

| entry point | workload | paged=1 | paged=0 | delta | output identical |
|---|---|---:|---:|---:|---|
| `_forward_it2i` | edit, 2 images | 24.050 s | 30.158 s | -20.25% | image, not compared here |
| `_forward_text` | image-to-text | 7.263 s | 12.532 s | -42.04% | no — 2031 vs 2190 chars |
| `_forward_text` | text-to-text | 7.014 s | 12.640 s | -44.51% | no — 2750 vs 2789 chars |

The two text divergences are the greedy-argmax ties already measured on the same-seed A/B at https://github.com/vllm-project/vllm-omni/pull/6516#discussion_r3881056024: the first differing step has a top1-top2 margin of 0.000 and 0.125 against a `|Δlogit|` of 0.188 and 0.164.

hsliuustc0106:
Ci failed

hsliuustc0106:
is there any CPU overhead for client e2e time? for the kernels, what's the optimization opportunity?

MrlixiangWE:
> Ci failed

I checked the buildkite log. The error is `[2026-08-30T23:07:03Z] nvidia-container-cli: initialization error: driver rpc error: timed out: unknown.` It looks like the CI container failed to start rather than a problem with my code. Could you retry it?

MrlixiangWE:
> is there any CPU overhead for client e2e time? for the kernels, what's the optimization opportunity?

All numbers below are re-measured, on 1x A800-80G with an Intel Xeon Gold 6336Y, TP=1, 1024x1024, seed 42, head `e32770a`.

## CPU overhead

2.31% of the AR decode window and 3.97% of the diffusion window at this head. Before the graph it was 37.29%.

Three arms, one session, each differing from the previous by one thing. B is not a switch, only the arm that separates the cache from the graph.

| Arm | KV cache | attention | CUDA graph |
|---|---|---|---|
| A `paged=0` | exact length | SDPA, flash backend | no |
| B | paged | `flash_attn_varlen` | no |
| C `paged=1` | paged | `flash_attn_varlen` | yes |

Profiled: think on, 8 steps. Kernel time from the profiled run, wall clock from a clean one; idle is the difference.

| Arm | clean wall | kernel time | idle | GPU busy | launches |
| A | 8259.5 ms | 5179.2 ms | 3080.3 ms | 62.71% | 206,149 |
| B | 8323.3 ms | 5136.6 ms | 3186.7 ms | 61.71% | 185,245 |
| C | 5196.9 ms | 5076.9 ms | 120.0 ms | 97.69% | 185,245 |

Kernel time is flat across the three arms, so the idle that disappears is host time and not device work. `think_chars` is 1233 in every run, and B and C issue the same launches.

Timing, think on, 50 steps, 2 warmups then n=10, P50:

| Arm | e2e | think decode | diffusion |
|---|---:|---:|---:|
| A | 17106.0 ms | 6500.4 ms | 10566.1 ms |
| B | 17142.5 ms | 6428.7 ms | 10639.9 ms |
| C | 14209.6 ms | 3485.1 ms | 10639.5 ms |

On the decode loop A -> B is -1.10% and B -> C is -45.79%: the paged cache buys the static shape, the graph buys the time.

Outside the model, wall - stage is 2.8-3.3 ms, 0.02% of e2e.

## Kernels

GEMM. Both windows are above 96% busy, so there is little left to take on the host side.

| window | GPU busy | GEMM | elementwise | attention | norm |
| AR decode, think on, 8 steps | 97.69% | 82.4% | 6.4% | 5.5% | 5.4% |
| diffusion, think off, 50 steps | 96.03% | 78.9% | 9.2% | 6.8% | 4.9% |

`ampere_bf16_s16816gemm_bf16_256x128_ldg8_f2f_stages_64x3_tn` is 7754.9 ms over 16,926 launches, 77.7% of diffusion kernel time.

vllm_omni/diffusion/worker/diffusion_worker.py
```diff
@@ -721,6 +721,12 @@ def sleep(self, level: int = 1) -> int:
                 self.model_runner.graph_runners.clear()
                 logger.info(f"[Worker {self.rank}] CUDA Graphs cleared.")
             model = self.model_runner.pipeline
+            # A pipeline may hold captures of its own across requests. Level 2
+            # discards the memory those were recorded against, so they have to
+            # go with `graph_runners`.
+            release = getattr(model, "release_captured_graphs", None)
```
hsliuustc0106: The cleanup is necessary, but this establishes a second graph-lifecycle protocol in the infrastructure worker: it clears `model_runner.graph_runners`, then duck-types an individual pipeline for a hidden method. `DiffusionModelRunner` owns model compilation, caches, and execution resources, so please expose one `model_runner.release_captured_graphs()` method that clears its runners and delegates internally, or register the paged runner with its graph registry. Otherwise every future pipeline that retains captures must discover this unwritten hook, and an omission will silently survive level-2 sleep.

hsliuustc0106:
Tested PR head `29c090d0c3c3ac609dff3ca50472b2beeff8c782` locally.

**Environment**

- 1x NVIDIA H200 (139 GiB, compute capability 9.0); physical GPU 1 was reserved for the run.
- Python 3.12.13; vLLM 0.28.0; torch 2.13.0+cu130; CUDA 13.0; diffusers 0.40.0; transformers 5.14.1; flashinfer 0.6.16.post3; BF16; TP=1.
- Model: `sensenova/SenseNova-U1.5-8B-MoT`; LoRA: `SenseNova-U1.5-8B-MoT-LoRA-8step.safetensors`; seed 42; 1024x1024. CUDA runs were executed through `gpu run --gpu-ids 1` with the PR worktree pinned in `PYTHONPATH`.

**Focused tests**

- `pytest -m "cpu and not cuda" tests/diffusion/models/sensenova_u1/ tests/diffusion/lora/test_loader.py -q`: **71 passed**, 41 deselected, 14 warnings, 5.95 s.
- `pytest -m "cpu and not cuda" tests/config/test_environment_variables.py tests/diffusion/test_diffusion_worker.py -q`: **13 passed**, 8 deselected, 14 warnings, 24.47 s.
- `pytest -m cuda tests/diffusion/models/sensenova_u1/test_sensenova_u1_attention_gqa.py tests/diffusion/models/sensenova_u1/test_sensenova_u1_paged_decode.py -q --tb=short`: **9 passed**, 23 deselected, 14 warnings, 102.13 s.

**Real-model smoke**

| Case | Result |
|---|---:|
| Base, 50 steps, cfg 4.0, think=false | image generated; 1024x1024 RGB; stage 9,017 ms; peak 34,384 MiB |
| PR, 8 steps, cfg 1.0, no LoRA control | image generated; stage 624.00 ms; 77.85 ms/step; peak 34,376 MiB |
| PR, 8 steps, cfg 1.0, distilled LoRA | image generated; **LoRA fused into 168 parameters**; stage 623.85 ms; 77.98 ms/step; peak 34,364 MiB |
| PR, 50 steps, cfg 4.0, think=true, `VLLM_OMNI_SENSENOVA_PAGED_DECODE=1` | image generated; stage 81,622 ms; peak 34,594 MiB |

The same-seed 8-step LoRA output differs from the no-LoRA control in 1,048,574/1,048,576 pixels (MAE 58.62), so the adapter is not a no-op. All generated PNGs were nonblank and had shape `(1024, 1024, 3)`.

**Online smoke**

The PR-source server reached `/health`; `img2text` returned a 1,921-character description for the generated image, and `text2text` returned the expected “Two plus two is four.” response. The server was shut down cleanly afterward.

The expected startup messages about the checkpoint lacking `configuration_neo_chat.py` appeared; generation was unaffected, consistent with the recipe note. This is a functional smoke, not a full multi-run image-quality benchmark.

@MrlixiangWE Could you add the H200 validation to `recipes/SenseNova/SenseNova-U1.5.md`, including the exact environment/commands, peak memory, and the 50-step and distilled-8-step timings? The recipe currently documents only the A800 measurements.

vllm_omni/diffusion/worker/diffusion_worker.py
MrlixiangWE: Fixed in 519c9b2, with the single runner method.

MrlixiangWE:
> Could you add the H200 validation to `recipes/SenseNova/SenseNova-U1.5.md`, including the exact environment/commands, peak memory, and the 50-step and distilled-8-step timings?

Added in be4c6b8 and dd52292.

PR #6519, opened 2026-08-23
[Bugfix] Warn that VLLM_TORCH_PROFILER_DIR was removed and is ignored

moisf56:
## Purpose

Addresses #5925.

`VLLM_TORCH_PROFILER_DIR` was removed from vLLM in vllm-project/vllm#33536 (v0.16.0), after profiling moved to `ProfilerConfig` in vllm-project/vllm#29912. Nothing reads it any more, in vLLM or in vLLM-Omni. Profiler endpoints are registered from `engine_args.profiler_config.profiler`, which matches upstream.

A stale value left in an environment or script is therefore silently inert. The gate at `vllm_omni/entrypoints/openai/api_server.py:541` had no `else` branch, so the server started normally, nothing was logged, and `/start_profile` returned 404 with nothing explaining why. As the reporter put it, the behavior was indistinguishable from success.

This adds a migration diagnostic: when the removed variable is set and no stage enables the profiler, log a warning naming the removal and pointing at `--profiler-config`. It is not a compatibility path and does not make the variable functional.

Two offline examples in this repo still gate on the removed variable (`examples/offline_inference/text_to_speech/cosyvoice3/end2end.py:152`, `examples/offline_inference/qwen2_5_omni/end2end.py:381`). Tracked as follow-up rather than folded in here, since they need a real opt-in flag rather than a rename.

`--profiler-config` works on the omni serve path and is listed under `vllm serve --help=all` or `--help=VllmConfig`. `docs/contributing/profiling.md` section 4 already documents it with a worked example; this adds a note there recording the removal.

Also adds parametrized coverage for `_should_enable_profiler_endpoints`, which had none.

## Test Plan

CPU-only, no GPU required.

```bash
pytest tests/entrypoints/openai_api/test_api_server_guards.py
```

New coverage:

- 13 parametrized cases for `_should_enable_profiler_endpoints`: missing and empty stage configs, `profiler_config` present but with no `profiler` key, attribute-style configs, and multi-stage scanning.
- One test driving `omni_run_server_worker` with `VLLM_TORCH_PROFILER_DIR` set and no `profiler_config`, asserting the warning is emitted and that `/start_profile` and `/stop_profile` are not registered.

**vLLM Version:** 0.27.0

**vLLM-Omni Commit:** b6028e2a2c9e8eeb6cdb9f6fabaf8acf17f1a003

## Test Result

`33 passed` (19 pre-existing + 14 new) at b6028e2a.

a2c8fc99 changes only the warning text, code comment, docstrings, and docs note, so runtime behavior is unchanged. The two `caplog` assertions still match the reworded message.

Lint and type gates checked against a `main` baseline rather than absolute counts:

| Gate | main | this branch |
|---|---|---|
| ruff check / ruff format | clean | clean |
| typos | clean | clean |
| markdownlint | 14 | 14, none added |
| mypy 3.12 | 5 | 5, none added |

Environment: Python 3.12.3, torch 2.13.0+cu132, Ubuntu 24.04.

### DIFF
--- a/docs/contributing/profiling.md
+++ b/docs/contributing/profiling.md
@@ -201,7 +201,7 @@ When `profiler_config.profiler` is set for a diffusion model, the server exposes
 - `POST /start_profile`
 - `POST /stop_profile`
 
-> **Note:** `VLLM_TORCH_PROFILER_DIR` does not enable these endpoints on the `--omni` path. vLLM-Omni registers them from `profiler_config` only, so setting that variable alone leaves `/start_profile` unregistered and logs a warning at startup. Use `--profiler-config` as shown below, or a `profiler_config` block in the stage YAML. `--profiler-config` is a nested config flag, so it does not appear in `vllm serve --help`.
+> **Note:** `VLLM_TORCH_PROFILER_DIR` was removed in vLLM v0.16.0 when profiling moved to `ProfilerConfig`, and nothing reads it any more. A stale value does not enable these endpoints; the server logs a migration warning at startup and leaves `/start_profile` unregistered. Use `--profiler-config` as shown below, or a `profiler_config` block in the stage YAML. `--profiler-config` is a nested config flag, so it is listed under `vllm serve --help=all` or `vllm serve --help=VllmConfig` rather than in the default `--help` output.
 
 ### Start the server
 
--- a/docs/contributing/profiling.md
+++ b/docs/contributing/profiling.md
@@ -201,6 +201,8 @@ When `profiler_config.profiler` is set for a diffusion model, the server exposes
 - `POST /start_profile`
 - `POST /stop_profile`
 
+> **Note:** `VLLM_TORCH_PROFILER_DIR` does not enable these endpoints on the `--omni` path. vLLM-Omni registers them from `profiler_config` only, so setting that variable alone leaves `/start_profile` unregistered and logs a warning at startup. Use `--profiler-config` as shown below, or a `profiler_config` block in the stage YAML. `--profiler-config` is a nested config flag, so it does not appear in `vllm serve --help`.
+
 ### Start the server
 
 Single-stage diffusion serving with torch profiler:
--- a/tests/entrypoints/openai_api/test_api_server_guards.py
+++ b/tests/entrypoints/openai_api/test_api_server_guards.py
@@ -42,6 +42,7 @@ from __future__ import annotations
 
 import asyncio
 import json
+import logging
 from contextlib import asynccontextmanager
 from types import SimpleNamespace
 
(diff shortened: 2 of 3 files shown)


moisf56:
###  Self-review

Scope: one `elif` in `omni_run_server_worker`, a doc note, and tests. No behavior change
when the profiler is already enabled, and the existing gate is untouched.

What I checked:

- The new branch only fires when VLLM_TORCH_PROFILER_DIR is set and no stage declares
  profiler_config.profiler, so existing setups are unaffected.
- Confirmed VLLM_TORCH_PROFILER_DIR is not referenced anywhere under vllm_omni/, so the
  variable really was inert on this path.
- Confirmed --profiler-config parses on the omni serve path, so the doc note is accurate.
- Tests: 33 passed in tests/entrypoints/openai_api/test_api_server_guards.py, 19 pre-existing
  plus 14 new. _should_enable_profiler_endpoints had no coverage before this.
- Lint and types checked against a main baseline rather than absolute counts: ruff and typos
  clean, markdownlint 14 on both, mypy 3.12 5 on both. Nothing new added.

One point a reviewer may want to weigh in on: the new lifespan test uses SimpleNamespace for
vllm_config, matching the three existing tests in that file. Happy to construct a real config
instead if that is preferred.

On the triage above: this PR touches docs/contributing/profiling.md, not
docs/design/module/profiling.md, so @gcanlin is likely not needed here.

hsliuustc0106:
This premise needs correcting before merge. `vllm_omni/entrypoints/openai/api_server.py:545` and the test docstring at `tests/entrypoints/openai_api/test_api_server_guards.py:502` say vLLM enables `/start_profile` from `VLLM_TORCH_PROFILER_DIR`, but that is false for this PR's vLLM 0.27 baseline. Upstream migrated profiling to `ProfilerConfig` / `--profiler-config` in [vllm#29912](https://github.com/vllm-project/vllm/pull/29912), then removed this variable and all of its loading logic in [vllm#33536](https://github.com/vllm-project/vllm/pull/33536). vLLM 0.16+ gates the profiler router on `app.state.args.profiler_config.profiler`, as Omni does. This is therefore a legacy-setting diagnostic, not an Omni/upstream mismatch.

Please either remove this compatibility branch, or frame it strictly as migration guidance for a setting removed upstream, e.g. "`VLLM_TORCH_PROFILER_DIR` was removed in vLLM v0.16 and is ignored; use `--profiler-config`." Update the PR body, code comment, test docstring, and documentation consistently. The docs should also point to `vllm serve --help=all` or `--help=VllmConfig`, where the flag is discoverable, rather than only saying it does not appear in help. The two remaining offline examples that key off this variable should be updated here or tracked explicitly as follow-up cleanup.

I reran the 14 targeted guard cases at frozen head `b6028e2a2c9e8eeb6cdb9f6fabaf8acf17f1a003` with vLLM 0.27.0; all passed. They verify the implementation, but the current wording locks in the obsolete contract.

moisf56:
Confirmed. VLLM_TORCH_PROFILER_DIR has zero occurrences in the vllm repo and is gone from
envs.py at v0.27.0. #33536's merge commit is an ancestor of v0.16.0 but not v0.15.1, so that
is exactly where it went. My framing was wrong.

Went with the reframe rather than removal, since two examples here still key off the variable,
so the stale contract is live in user scripts. Pushed a2c8fc99:

- Warning now says the variable was removed in vLLM v0.16.0 when profiling moved to
  ProfilerConfig, and points at --profiler-config.
- Code comment cites #29912 and #33536, framed as a migration diagnostic, not a compat path.
- Both test docstrings and the docs note updated to match.
- Docs now point at `vllm serve --help=all` / `--help=VllmConfig` instead of saying the flag is
  absent from help.
- PR body updated.

The two examples (cosyvoice3/end2end.py:152, qwen2_5_omni/end2end.py:381) need a real opt-in
flag rather than a rename, so I would rather track them as follow-up than fold them in here.
Say the word if you want them in this PR.

Thanks for rerunning the guard cases.

Also worth noting for #5925: the reporter was setting a variable vLLM had already removed.

moisf56:
Opened #6761 to track the two offline examples.

One note: the Test Plan says vLLM 0.27.0, which is what I tested against. Main has since
moved to 0.28.0, so if you want it re-verified on the current base I can rebase and rerun.

moisf56:
@hsliuustc0106 ready from my side whenever you have a slot

PR #6520, opened 2026-08-23
[Test][Doc] MiniCPM-o 4.5 encoder-cache regression coverage + caching

frank-2077:
## Purpose

Probe-first deliverable for the **audio-encoder reuse** slice of #5069, following the acceptance criteria set in [this comment](https://github.com/vllm-project/vllm-omni/issues/5069#issuecomment-<amy-2026-08-09>): verify the existing mm_hash / `EncoderCacheManager` behavior for repeated and mixed audio inputs first, and — since the generic cache turns out to cover MiniCPM-o audio correctly — limit the contribution to focused regression coverage + docs, with **no model-local cache**.

This PR contains no functional changes: tests and documentation only.

## What was verified before writing this PR

All experiments ran on current `main`, vLLM 0.26.0, with temporary print probes at `EncoderCacheManager.check_and_update_cache` and around the live `get_audio_hidden_states` call (reverted afterwards; not part of this PR). Requests were strictly sequential so hit/miss attribution is not blurred by batching. Hardware: 8x RTX 4090 probe profile (stage-0 TP=2, `enforce_eager`) — numbers are within-box only, not comparable to H100 baselines.

### 1. Attribution / call counts / ordering / latency / memory (cold, warm, mixed)

A, B, C = distinct 3 s audios; D = 7 s. One fresh server, default `mm_processor_cache_gb`:

| phase | request | cache check | encoder entered? | items encoded | encoder ms |
|---|---|---|---|---|---|
| cold | A alone | MISS | yes | 1 | 66 (first call; steady-state below) |
| warm | A again | **HIT** | **no** | — | — |
| cold | B alone | MISS | yes | 1 | 17 |
| mixed | [A (seen), C (new)] | HIT, MISS | yes ×1 | **1 (C only)** | 17 |
| warm-reorder | [C, A] (both seen) | HIT, HIT | no | — | — |
| warm-repeat | [A, C] again | HIT, HIT | no | — | — |
| cold | D alone (7 s) | MISS | yes | 1 | 23 |

- `mm_hash` is content-determined and stable — A's hash was identical across all four requests containing it; in the reordered request the two hashes tracked their items, not their positions.
- **Ordering**: the warm-repeat request (both items from cache, encoder never entered) produced byte-identical text to the mixed request (one item encoded live) — cached embeddings land in the right placeholder positions.
- Peak stage-0 memory moved 20 252 → 20 256 MiB over the whole workload; steady-state audio-encoder time is ~17 ms (3 s) / ~23 ms (7 s), computed redundantly per TP rank.
- Honest value note: at these audio lengths the reuse saves ~17–23 ms of TTFT — much smaller than the image-side win.

### 2. Finding: a cache hit is not numerically equivalent to a fresh encode

Loading only `apm` + `audio_projection_layer` (and separately `vpm` + resampler) standalone with checkpoint weights, and binding the *production* `get_audio_hidden_states` / `_process_image_input` methods (no code copy), shows that both encoders produce slightly different outputs for the **same item** depending on which other items shared its encoder batch — with a bitwise-clean determinism floor (same batch composition twice → identical):

| comparison (bf16, serving dtype) | max\|Δ\| | cosine |
|---|---|---|
| audio A alone vs A alone, re-run (floor) | 0 (bitwise) | 1.0000000 |
| audio A vs [A, B] same length | 0.125 | 0.99974 |
| audio A vs [A, B] longer companion | 0.125 | 0.99978 |
| audio A vs A zero-padded to B's length | 0.094 | 0.99979 |
| image X alone vs [X, X] (identical companion!) | 1.31 | 0.99917 |

Normalised against each tensor's own spread: Whisper tower up to 7.3σ (0.62σ on the final audio embedding); SigLIP tower up to 8.4σ (2.5σ on the final image embedding). fp32 shrinks everything by 2–4 orders of magnitude — this is bf16 accumulation-order sensitivity in the pad-to-batch-max design (the `[X, X]` row shows batch-size-dependent reduction order, not just padding), not a masking bug. Results are bit-identical across two independent torch builds/GPUs.

Because the encoder cache is content-addressed by mm_hash alone, it permanently serves whichever batch-composition variant was computed first. An image-side end-to-end check (prefix caching off; encoder cache primed with `[P, Q]` vs `P` alone; measured request byte-identical and submitted alone, batch of 1) flipped greedy text in **1 of 8** pairs, while the cached-solo vs fresh-solo control was 8/8 identical — so the cache mechanism itself is sound; what varies is which variant it stores. The audio-side end-to-end flip test has not been run (final-embedding delta is smaller there, 0.62σ vs 2.5σ).

This is modality-general and exists with prefix caching disabled, so it is **documented** here (design doc) and proposed for tracking as its own bounded issue rather than folded into this slice. The e2e regressions in this PR are deliberately designed **not** to depend on it: they compare paths that consume the identical cached embedding tensor, so they are deterministic in CI.

### 3. Finding: `mm_processor_cache_gb: 0` silently disables all encoder reuse

With `mm_processor_cache_gb: 0` **and** `enable_prefix_caching: false` (the shipped MiniCPM-o default), upstream vLLM replaces content mm_hashes with per-request counter IDs (`renderer0-mm-<N>-<modality>-<i>`; `vllm/renderers/base.py`, `_process_mm_uuids`), overriding even user-provided UUIDs. Measured effect: **100 % MISS including byte-identical repeats**, with the encoder re-entered every time and no warning logged. This is upstream-intended semantics ("both caches off ⇒ no reuse wanted"), but since every MiniCPM-o layout ships `enable_prefix_caching: false`, zeroing the processor cache to save host memory silently forfeits all cross-request encoder reuse. Now documented in the design doc + recipe.

## Changes

**Tests**

- `tests/core/test_mm_block_hash_minicpmo.py` (new, L1 `core_model + cpu`, 12 cases): mm_hash determinism/discrimination (content, modality kwarg, dtype, shape, sample rate); block-hash mm extra-key contract for the MiniCPM-o placeholder geometry — items straddling block boundaries carry deterministic negative in-block offsets, two items per block, same-item-different-offset disambiguation, interleaved image/audio packs keep prompt order; `OmniTensorPrefixCache` round-trip (merged hidden states on a prefix hit equal the full forward; stale-slot overwrite).
- `tests/e2e/offline_inference/test_minicpmo_4_5.py` (2 new tests, L4 `full_model + omni + cache`):
  - `test_audio_encoder_cache_warm_repeat_parity` — cold encode vs encoder-cache hit must produce identical greedy text;
  - `test_audio_encoder_cache_mixed_hit_partition_parity` — `[warm A, cold C]` vs its fully-cached byte-identical repeat must agree, guarding placeholder ordering and partial-hit partitioning.
- `tests/helpers/assertions.py`: `assert_omni_text_responses_identical` (shared cache-path parity assert, per the helpers convention).

**Docs**

- `docs/design/feature/prefix_caching.md`: two new caveat sections — the `mm_processor_cache_gb: 0` identifier-degradation trap, and the batch-composition dependence of cached encoder outputs.
- `recipes/OpenBMB/MiniCPM-o-4_5.md`: "Multimodal caching" note (do not zero `mm_processor_cache_gb` on stage 0; keep `enable_prefix_caching: false` until the #5069 parity work lands).

**CI**

- L1 tests are collected by the existing "Simple · Other Test" sweep (`tests/ -m 'core_model and cpu'`).
- The new e2e tests are collected by the existing nightly `full_model and H100 and omni` sweep — no nightly YAML change needed.
- `.buildkite/cuda/test-ready.yml`: added `tests/helpers/assertions.py` to the MiniCPM-o 4.5 Offline Test step's `source_file_dependencies`.

## Test Plan

```bash
# L1 — CPU only
cd tests
pytest -s -v core/test_mm_block_hash_minicpmo.py -m "core_model and cpu"

# L4 e2e — requires H100 (or A2) + openbmb/MiniCPM-o-4_5 weights cached
pytest -s -v e2e/offline_inference/test_minicpmo_4_5.py \
  -m "full_model and omni and cache" --run-level full_model
```

## Test Result

- L1: **12/12 passed** locally (CPU).
- L4 e2e: collected successfully (9 tests in the module, including the 2 new ones); not executed locally — the probe box's GPU layout differs from the CI single-H100 profile. The assertions mirror behavior measured directly by the attribution probe above.

## Related

- RFC #5069 (audio-encoder reuse slice; also corrects the RFC's "MRoPE correctness is the blocker" line — MiniCPM-o 4.5 does not use MRoPE, `uses_mrope() == False`; the actual blocker is output parity).
- Prior art: image-side attribution probe in the A.2 discussion; #5587 (streaming audio-encoder memory — deliberately not touched here).

### DIFF
--- a/tests/e2e/offline_inference/test_minicpmo_4_5.py
+++ b/tests/e2e/offline_inference/test_minicpmo_4_5.py
@@ -116,3 +116,76 @@ def test_video_to_audio(omni_runner, omni_runner_handler) -> None:
     video = generate_synthetic_video(24, 24, 20)["np_array"]
     request_config = {"prompts": get_question("video"), "videos": video, "modalities": ["audio"]}
     omni_runner_handler.send_omni_request(request_config)
+
+
+def _cache_probe_audio(phrase: str):
+    """Synthetic audio with a phrase-unique mm_hash.
+
+    Each distinct phrase yields distinct waveform bytes, so the encoder-cache
+    tests below control exactly which items are warm: they never collide with
+    the default-phrase audio used by the other tests in this module.
+    """
+    audio = generate_synthetic_audio(1, 1, 16000, phrase_text=phrase)["np_array"]
+    if len(audio.shape) == 2:
+        audio = audio.squeeze()
+    return (audio, 16000)
+
+
+@pytest.mark.full_model
+@pytest.mark.omni
+@pytest.mark.cache
+@hardware_test(res={"cuda": "H100", "npu": "A2"}, num_cards=1)
+@pytest.mark.parametrize("omni_runner", test_params, indirect=True)
+def test_audio_encoder_cache_warm_repeat_parity(omni_runner, omni_runner_handler) -> None:
+    """Repeating an identical audio request must reproduce the same text.
+
+    Regression for the audio half of the #5069 encoder-cache evaluation: the
+    first request encodes the audio cold; the byte-identical second request is
+    served from vLLM's content-addressed encoder cache (mm_hash) without
+    re-entering the Whisper encoder. Both paths consume the same cached
+    embedding tensor, so greedy text must match exactly. A divergence here
+    means cached encoder outputs are corrupted or misapplied on the hit path.
+    """
+    from tests.helpers.assertions import assert_omni_text_responses_identical
+
+    audio = _cache_probe_audio("encoder cache warm repeat probe")
+    request_config = {"prompts": get_question("audio"), "audios": audio, "modalities": ["text"]}
+    responses = [omni_runner_handler.send_omni_request(request_config) for _ in range(2)]
+    assert_omni_text_responses_identical(responses, context="audio warm repeat")
+
+
+@pytest.mark.full_model
+@pytest.mark.omni
+@pytest.mark.cache
+@hardware_test(res={"cuda": "H100", "npu": "A2"}, num_cards=1)
+@pytest.mark.parametrize("omni_runner", test_params, indirect=True)
+def test_audio_encoder_cache_mixed_hit_partition_parity(omni_runner, omni_runner_handler) -> None:
+    """Mixed hit/miss and fully-cached paths must agree, item order preserved.
+
+    Regression for the #5069 audio attribution probe. Sequence:
+      1. warm item A alone (fills its encoder-cache entry);
+      2. request [A, C]: A is a cache hit, C is encoded live — the encoder
+         receives only the missing item;
+      3. repeat [A, C] byte-identically: both items are now cache hits and the
+         encoder is not entered at all.
+    Requests 2 and 3 consume identical embedding tensors, so their greedy text
+    must match exactly. A divergence means cached embeddings were placed into
+    the wrong placeholder positions (ordering) or partial-hit partitioning
+    corrupted the batch.
+    """
+    from tests.helpers.assertions import assert_omni_text_responses_identical
+
+    audio_a = _cache_probe_audio("encoder cache mixed probe alpha")
+    audio_c = _cache_probe_audio("encoder cache mixed probe charlie")
+
+    # Step 1: make A warm so step 2 exercises the partial-hit partition.
+    omni_runner_handler.send_omni_request({"prompts": get_question("audio"), "audios": audio_a, "modalities": ["text"]})
+
+    mixed_config = {
+        "prompts": "Are these two audio clips the same?",
+        # Nested list: one prompt carrying two audio items (two placeholders).
+        "audios": [[audio_a, audio_c]],
+        "modalities": ["text"],
+    }
+    responses = [omni_runner_handler.send_omni_request(mixed_config) for _ in range(2)]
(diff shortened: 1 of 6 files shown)


tests/e2e/offline_inference/test_minicpmo_4_5.py
```diff
@@ -116,3 +116,76 @@ def test_video_to_audio(omni_runner, omni_runner_handler) -> None:
...
+    re-entering the Whisper encoder. Both paths consume the same cached
+    embedding tensor, so greedy text must match exactly. A divergence here
+    means cached encoder outputs are corrupted or misapplied on the hit path.
+    """
+    from tests.helpers.assertions import assert_omni_text_responses_identical
+
+    audio = _cache_probe_audio("encoder cache warm repeat probe")
+    request_config = {"prompts": get_question("audio"), "audios": audio, "modalities": ["text"]}
+    responses = [omni_runner_handler.send_omni_request(request_config) for _ in range(2)]
+    assert_omni_text_responses_identical(responses, context="audio warm repeat")
```
lishunyang12, line 154: Assert encoder-call counts for cold, warm, and mixed requests.

Issue #6435, opened 2026-08-21
[Bug]: Nightly CI, Qwen-Image, performance metrics regressed by more than 10% compared to the baseline in some scenarios


yenuo26:
### Your current environment

<details>
<summary>The output of <code>python collect_env.py</code></summary>

```text
CI env
```

</details>

### Your code version

<details>
<summary>The commit id or version of vllm</summary>

```text
0.27.0
```
</details>
<details>
<summary>The commit id or version of vllm-omni</summary>

```text
10906e3a6db947290581b88dc4b6f35bd2a26b41
```
</details>

### 🐛 Describe the bug

| Source | Model | Hardware | Test Suite | Test Scenario | Metric | Current Value | Baseline Value | Change | Status | Trend |
|--------|-------|----------|------------|---------------|--------|---------------|----------------|--------|--------|-------|
| Buildkite | Qwen/Qwen-Image | H100 | qwen_image | bench=512x512_steps20_high_concurrency, data=random, c=8, n=160 | test_qwen_image_single_device_step_execution_high_concurrency | e2e_latency_ms | 16309.4626 | 9656 | -68.90% | fail | 3 days |
| Buildkite | Qwen/Qwen-Image | H100 | qwen_image | bench=512x512_steps20_high_concurrency, data=random, c=8, n=160 | test_qwen_image_single_device_step_execution_high_concurrency | throughput_qps | 0.4799 | 0.8159 | -41.19% | fail | 3 days |
| Buildkite | Qwen/Qwen-Image | H100 | qwen_image | bench=512x512_steps20_high_concurrency, data=random, c=2, n=40 | test_qwen_image_single_device_step_execution_high_concurrency | e2e_latency_ms | 4181.4226 | 2992 | -39.75% | fail | 3 days |
| Buildkite | Qwen/Qwen-Image | H100 | qwen_image | bench=512x512_steps20_high_concurrency, data=random, c=4, n=80 | test_qwen_image_single_device_step_execution_high_concurrency | e2e_latency_ms | 8279.8693 | 5929.3 | -39.64% | fail | 3 days |
| Buildkite | Qwen/Qwen-Image | H100 | qwen_image | bench=512x512_steps20_high_concurrency, data=random, c=2, n=40 | test_qwen_image_single_device_step_execution_high_concurrency | throughput_qps | 0.4723 | 0.6714 | -29.66% | fail | 3 days |
[...]

PR #1593, opened 2026-03-02
[Feature][RL] Support batching for QwenImage in async mode

knlnguyen1802:
## Purpose
Fix https://github.com/vllm-project/vllm-omni/issues/1511
This PR aim to support batching for QwenImage in async mode
## Test Plan

<details>

```python
import asyncio
import time
from vllm_omni.entrypoints.async_omni import AsyncOmni
from vllm_omni.inputs.data import OmniDiffusionSamplingParams


async def run_unified_batch_mode(omni, prompts, label="batch"):
    """Run multiple prompts concurrently using AsyncOmni.generate()."""
    start = time.time()

    async def generate_one(prompt, idx, prefix):
        request_id = f"{prefix}-{idx}"
        async for output in omni.generate(
            prompt,
            request_id=request_id,
            sampling_params_list=[OmniDiffusionSamplingParams(
                num_inference_steps=10,
                width=512,
                height=512,
            )],
        ):
            # Save final image
            image = output.request_output.images[0]
            image.save(f"{prefix}_{idx}.jpg")
        return idx

    tasks = [asyncio.create_task(generate_one(prompt, i, label))
             for i, prompt in enumerate(prompts)]
    await asyncio.gather(*tasks)

    end = time.time()
    print(f"{label.capitalize()} mode ({len(prompts)} reqs) took: {end - start:.2f} seconds\n")


async def warmup_with_different_prompts(omni):
    """Warm up using a different set of prompts to avoid polluting timing."""
    warmup_prompts = [
        {"prompt": "a sunflower in a glass vase", "negative_prompt": "blurry"},
        {"prompt": "a rocket launching into space", "negative_prompt": "low detail"},
        {"prompt": "a small cottage in the snowy mountains", "negative_prompt": "foggy"},
        {"prompt": "a colorful parrot sitting on a tree branch", "negative_prompt": "low contrast"},
        {"prompt": "a steaming bowl of ramen noodles", "negative_prompt": "dark lighting"},
        {"prompt": "a classic car on a desert road", "negative_prompt": "modern"},
        {"prompt": "a neon-lit street in Tokyo at night", "negative_prompt": "low resolution"},
        {"prompt": "a dragon flying above a castle", "negative_prompt": "low detail"},
    ]

    print("Running warm-up with 8 different requests...")
    start = time.time()

    async def warmup_one(prompt, idx):
        async for _ in omni.generate(
            prompt,
            request_id=f"warmup-{idx}",
[...]

PR #6525, opened 2026-08-23
[BugFix] Qwen-image performance regressed - Avoid mapping diffusion_batch_size onto scheduler max_num_seqs

NumberWan:
Fix Qwen-image performance Regression

## Fixes

https://github.com/vllm-project/vllm-omni/issues/6435

## Summary

#5676 wrote `od_config.max_num_seqs = batch_size` on the **shared** diffusion stage init path. `batch_size` here is `diffusion_batch_size` (default **1**, not bound to `--max-num-seqs`). Qwen-Image nightly serve uses `--step-execution --max-num-seqs 8`, so capacity became 1 and HTTP concurrency 8 queued serially.

This PR **deletes those two assignments**. Scheduler capacity is only `--max-num-seqs` / stage YAML `max_num_seqs`. `diffusion_batch_size` stays a client knob.

Wan request-level batching of width 4 must set `--max-num-seqs 4` (or YAML). The default deploy file `vllm_omni/deploy/wan2_2_ti2v.yaml` now has `max_num_seqs: 4`. Replica-DP Wan YAML is unchanged (`max_num_seqs: 1`, one request per replica GPU).

There is **no** shared `max(CLI, batch_size)` helper.

## Nightly symptom (H100, from #6435)

`test_qwen_image_single_device_step_execution_high_concurrency`, `c=8`, `n=160`:

| Metric | Current | Baseline | Change |
|---|---:|---:|---:|
| `throughput_qps` | 0.4799 | 0.8159 | **−41.2%** |
| `e2e_latency_ms` | 16309 | 9656 | **+68.9%**  |

At `c=2/4/8`, QPS stayed ~flat (~0.47) while latency scaled nearly linearly (~4.2s → 8.3s → 16.3s). That is serial admission, not a gradual compute regression.

## RCA (1× L20X, relative only)

Same case: 512×512, 20 steps, `c=8`, `n=160`, `--step-execution --max-num-seqs 8`.

First slowing commit: `2a315e1a` — `[Perf] Support request-level batching for Wan2.2 pipelines` (**#5676**). Parent `d1e230c9` is still fast. Later days stay slow; **no second independent ≥15% drop**.

Daily checkpoint (same case):

| Day tip | QPS | Mean latency |
|---|---:|---:|
| 8/15 (still good) | ~0.70 | ~11.3s |
| 8/16 (first bad day) | ~0.53 | ~14.8s |
| 8/20 (issue tip) | ~0.51 | ~15.4s |

**A/B (4 runs each):**

| Variant | Median QPS | Median mean latency | vs parent |
|---|---:|---:|---:|
| Parent `d1e230c9` | **0.723** | **10.95s** | reference |
| Culprit `2a315e1a` | **0.502** | **15.59s** | QPS **−30.6%**, latency **+42.4%** |

CV ~2–3%. Delete only `od_config.max_num_seqs = batch_size` on the culprit → QPS **0.712** / **11.09s** (parent-level). Isolates the overwrite, not Wan pipeline files.

| Knob | Meaning | Qwen nightly |
|---|---|---|
| `--max-num-seqs` | scheduler in-flight | **8** |
| `diffusion_batch_size` | client request-batch width (#1593) | default **1** |

## Behavior change

Call sites that relied on `diffusion_batch_size=N` **alone** to raise scheduler in-flight width will now stay at CLI/YAML `max_num_seqs` (often 1). Pass `--max-num-seqs N` as well.

## Test plan

- [x] Unit tests: shared init `batch_size=4` leaves `max_num_seqs` at the CLI/YAML value (1 / 8 / `None`), not 4.
- [x] Qwen-Image step-execution sweep `c=1,2,4,8` (512×512 / 20 steps / `n=160`). Expect QPS to scale with `c`, not stay flat at ~0.47.
- [x] Wan request-level batching with `--max-num-seqs 4` (YAML or CLI). `diffusion_batch_size` alone must not raise scheduler capacity.


Gaohan123 (approved):
LGTM. Thanks

PR #6528, opened 2026-08-23
[Bugfix][MiniCPM-o] Prevent native duplex Stage-1 prompt overflow

Sy0307:
## What broke

Native MiniCPM-o 4.5 duplex sessions could crash Stage 1 after short streamed input segments. The observed failure was:

`ValueError: could not broadcast input array from shape (4111,) into shape (4096,)`

The Stage-1 EngineCore then exited and the replica was evicted. The failure was reproducible with a 600 ms segment followed by a 1200 ms segment on the combined duplex branch.

## Reproduction

Run the deterministic regression:

```bash
pytest -q tests/core/sched/test_omni_ar_scheduler_streaming.py -k native_duplex_prompt_reset
```

The H200 reproduction used the native duplex realtime helper with two turns:

```bash
python tests/e2e/online_serving/helpers/minicpmo_realtime_duplex_scenarios.py \
  --url ws://127.0.0.1:8106/v1/realtime?duplex=1 \
  --input-wav tests/assets/minicpmo_4_5/response_required_16k.wav \
  --turns 2 --first-turn-ms 600 \
  --turn-duration-ms 600 --turn-duration-ms 1200 \
  --chunk-ms 200 --realtime-input --validation-mode model-policy
```

## Root Cause

Native duplex Talker continuation keeps the previously computed codec history in the Stage-1 request and appends the next handoff condition. The scheduler admitted the next request without checking the combined prompt/history length against the Talker context limit (`max_model_len=4096`). This allowed a prompt of 4111 tokens to reach `GPUInputBatch.add_request()`.

## Fix

- Detect native-duplex Stage-1 updates before appending the next handoff.
- Compute the candidate prompt plus retained computed history.
- When the candidate reaches the context limit, use the existing `_replace_streaming_session()` path to clear stale prompt/output/KV bookkeeping and prefill the current handoff from zero.
- Keep one context position available for the next sampled codec token.
- Add a regression test for the 4111-to-4096 overflow shape.
- Include the combined duplex lifecycle, VAD, abort, segment-generation, and connector cleanup fixes already present on the working branch.

The deployment context limit remains 4096; this change does not hide the problem by increasing `max_model_len` or truncating a live KV sequence.

## Validation

- `229 passed, 1 skipped` for scheduler, MiniCPM-o bridge/Talker, connector, orchestrator, and segment-generation tests.
- `245 passed` for duplex protocol, handler, and request-scoped error handling tests.
- H200 native duplex protocol smoke passed.
- H200 short-segment runs produced audio and logged the expected reset, for example `current=4079 kept_output=16 next=18 max_model_len=4096`.
- H200 service remained healthy; no `4111`, broadcast-shape error, Stage-1 EngineCore exit, or replica eviction occurred after the fix.
- Environment: 1x NVIDIA H200 (143 GB), torch `2.11.0+cu130`, vLLM `0.27.0`.
- Ruff lint and format checks passed. Full pre-commit could not complete because the shared H200 environment cannot download actionlint Go dependencies; the failure was environmental, not a hook finding.
- The realtime helper emitted complete `response.done` audio, but its playback-ack send raced with session closure in this environment; the lower-level websocket protocol smoke passed.


Sy0307:
Superseded by signed PR #6529; this branch was closed because the DCO check required sign-off trailers on inherited commits.

PR #6529, opened 2026-08-23
[Bugfix][MiniCPM-o] Stabilize native duplex streaming and Stage-1 handoffs

Sy0307:
## Merge Status and Follow-ups (2026-09-05)

This integration branch merged into `main` on 2026-08-29 as `21dd4b33ac46e71479e839a9de3ed1b801046296`.

- It is the landed integration point for #5658/#5662/#5674/#5721; those source branches are provenance, not independent merge candidates.
- #6626 owns Talker context rollover; #6170 owns model-specific handoff/barge-in behavior.
- The remaining post-merge concern is bounding `assistant_response_snapshots` in long-lived sessions whose clients never send playback ACKs.
- Current-main coverage should keep stale-frame accounting, queue cleanup, segment-generation fencing, response/playback ordering, and multi-session behavior together.

## Scope

This PR is the integration-validation branch for the shared native-duplex scheduler and connector fixes:

- #5658: key streaming segment state by stage in the orchestrator.
- #5662: resynchronize the streaming-input counter after queue/status rewrites.
- #5674: fence stale async segment output and preserve valid queued work.
- #5721: sweep aborted requests from `skipped_waiting` before scheduling.

The Talker context rollover is owned by merged PR #6626 and is now part of the
base branch. This PR no longer carries a duplicate Stage-1 context reset; the
base provides the Talker `sliding_recompute` implementation and its model,
connector, and test coverage.

PR #6170 remains the owner of model-specific handoff failure isolation and
barge-in behavior. The remaining Realtime changes here cover the shared
response/playback lifecycle needed by this integration path.

## Fixes

- Drain `num_stale_output_tokens` and `async_tokens_to_discard` together when a
  stale frame belongs to both accounting domains, so a new segment first frame
  is not discarded.
- Capture the old segment generation before a queued streaming update mutates
  the request, and pass that generation to the boundary `save_async` task.
- Keep aborted queue entries, streaming counters, and per-stage segment state
  consistent across async scheduling transitions.
- Keep response turn identity synchronized when the bounded native
  auto-response closes before a model `turn_eos`, allowing the next model turn
  to create a new response.
- Validate playback response/item ownership and keep ACK_ONLY assistant
  snapshots response-local. A delayed ACK after a later user commit is rejected
  instead of appending old assistant text after the new user turn.

## Validation

- A800, vLLM `0.28.0`, three-stage MiniCPM-o 4.5, two-turn
  `response-required`: two `response.created`, two `response.done`, two
  `input_audio_buffer.committed`, two playback history commits, 23 audio
  deltas, zero errors, and zero stale audio deltas.
- Focused A800 regression tests: scheduler stale/generation, connector,
  orchestrator, protocol, attachment, and duplex handler paths pass.
- Ruff check and format pass.
- Stage-1 Talker context rollover and sliding recompute coverage are provided
  by merged PR #6626 in the base branch.

## Current Integration

- Current head: `e8123a93`.
- Base: main containing merged PR #6626.
- The duplicate scheduler context-reset block and its standalone regression
  test were removed in `e8123a93`.
- Required GitHub build, pre-commit, and DCO checks are expected to run on the
  new head; vendor Buildkite lanes are tracked separately.


Sy0307:
Self-review completed on head `776f5776`.

- Confirmed the branch integrates the four intended fixes: #5658, #5662, #5674, and #5721.
- Fixed the queued pre-boundary chunk false-drop introduced while hardening the #5674 watermark race.
- Added runtime capability validation so #6089 Nemotron (`supports_barge_in=false`) rejects hard server-VAD instead of advertising an unsupported interrupt contract.
- Restored `DuplexOverlapPolicy.AUTO` compatibility.
- Current-branch focused tests: `341 passed`.
- #6089 union compatibility surface: `337 passed, 1 skipped`.
- Real MiniCPM-o 4.5 weights loaded on H200 and protocol/overflow safety was exercised. The slow shared H200 did not complete the long native response before the bounded helper timeout; no EngineCore or prompt-overflow error occurred.

Sy0307:
Additional review status after the final H200 run:

- The four intended combo sources are #5658, #5662, #5674, and #5721; all four behavioral changes and their regression coverage are present in this head.
- A sender-side queued-task guard was corrected: enqueue-time generation filtering drops late callbacks, while already queued pre-boundary FIFO work is preserved.
- Shared server-VAD now rejects runtimes without `supports_barge_in`; `DuplexOverlapPolicy.AUTO` was restored for API compatibility.
- Final current-branch focused surface: `341 passed`; shared protocol/handler subset: `224 passed`.
- On a clean full-source H200 checkout using the real MiniCPM-o 4.5 checkpoint, all three stages loaded on idle GPU 1, protocol smoke passed, and the 600 ms + 1200 ms run produced real audio and repeated Stage-1 resets without prompt overflow or EngineCore failure.
- The slow H200 run did not reach natural response termination before the bounded session timeout; that remains an explicit validation limitation, not a claimed full response-lifecycle pass.
- PR #6089 needs a rebase/test expectation update for the newer server-VAD behavior; the shared production union is compatible, but the two PR heads are not textually conflict-free as-is.

Sy0307:
Removed the generic server-VAD scope from this PR as requested. The current head no longer adds `silero-vad`, `openai/vad.py`, server-VAD state/translator logic, server-VAD E2E drivers, or Buildkite dependencies. The framework keeps model-native turn taking and explicit control-plane barge-in only; model-specific VAD belongs in a model adapter such as a Qwen3-Omni integration.

The PR is now 23 files with `+838/-81`; test additions are `+582` lines. Minimal focused validation after reduction: `327 passed` for scheduler/connector/protocol/handler plus `88 passed, 1 skipped` for MiniCPM bridge/orchestrator tests, i.e. `415 passed, 1 skipped` across the selected surface.

Sy0307:
## Stack Clarification (2026-08-27)

This PR is tracked in RFC #6494 as the integration-validation branch for the core state stack, not as a second implementation of the model-specific work in #6170.

- Core sources are #5658, #5662, #5674, and #5721.
- #6170 owns MiniCPM-specific handoff failure isolation, Realtime lifecycle, and opt-in server-VAD/barge-in behavior.
- The generic server-VAD changes were removed from this branch.
- The latest visible checks are green, but the branch currently reports `CONFLICTING`/`DIRTY` and `REVIEW_REQUIRED`.

Next step is to validate the four core fixes as one stack, rebase onto the current `main`/#6170 integration point, and avoid merging duplicate model-specific changes.

vllm_omni/core/sched/omni_ar_scheduler.py
```diff
@@ -771,6 +781,30 @@ def _update_request_as_session(self, session: Request, update: StreamingUpdate)
...
+                logger.info(
+                    "Resetting native duplex Stage-1 prompt before context overflow: "
+                    "req=%s current=%s kept_output=%s next=%s max_model_len=%s",
+                    session.request_id,
+                    current_prompt_len,
+                    kept_output_len,
+                    next_prompt_len,
+                    max_model_len,
+                )
+                self._replace_streaming_session(session, update)
```
congw729: [P2] I noticed that the explicit prompt-replacement path releases the old KV/encoder cache before calling `_replace_streaming_session()`, while this overflow-reset path calls `_replace_streaming_session()` directly. Is retaining the old block allocation intentional here? Since the request token state and `num_computed_tokens` are reset to the new prompt, I wonder whether we should call `_release_replaced_streaming_prompt_cache()` first to keep the scheduler cache state aligned and release the nearly full old context.

congw729 (commented):
Thanks for putting these streaming fixes together! I left one question about the Stage-1 overflow reset path.

Sy0307:
Current head `d2a20632` is mergeable. GitHub build/pre-commit/DCO and docs checks are green; the required Buildkite lanes are running. This PR is intentionally the core-only integration stack for #5658, #5662, #5674, and #5721, with the model-specific #6170 work excluded. I replied to the Stage-1 overflow/cache question inline and the current focused validation is recorded in the PR. @tzhouam @gcanlin @yuanheng-zhao @NickCao @ywang96, please review once the Buildkite lanes complete.

vllm_omni/core/sched/omni_ar_scheduler.py
Sy0307: Thanks for catching this. This is addressed on the current head d2a20632: the native-duplex overflow path now calls _release_replaced_streaming_prompt_cache(session) immediately before _replace_streaming_session(...), matching the explicit replacement path. The regression covers both scheduler block release and encoder-cache release. I will keep this as part of the merge gate.

Sy0307:
CI update (2026-08-28): required Buildkite `vllm-omni` #14280, `vllm-omni-amd-ci` #11096, and `vllm-omni-intel-ci` #7506 are failing; NPU #6043 is still pending. GitHub build/pre-commit/DCO and docs checks remain green. Squash auto-merge is enabled, but review and all required CI gates must pass before it can merge. @tzhouam @gcanlin @yuanheng-zhao @NickCao @ywang96, please inspect the failing lane logs or re-run after the cause is addressed.

Sy0307:
Follow-up fix pushed in `2f39a954e`.

This closes the remaining multi-turn gaps found on head `263e01de`:

- drains `num_stale_output_tokens` and `async_tokens_to_discard` on the same stale frame, so the next segment's first frame is not dropped;
- snapshots the pre-stop segment generation before a queued streaming update mutates the request, avoiding the boundary double-advance;
- validates playback `response_id`/`item_id` ownership and stores assistant history snapshots per response for delayed ACKs;
- advances the serving model turn at the bounded auto-response close and lets the orchestrator bridge catch up to the next fence.

Real A800 validation with vLLM `0.28.0` and the three-stage MiniCPM-o 4.5 checkpoint now passes the full two-turn `response-required` lifecycle: `response.created=2`, `response.done=2`, `input_audio_buffer.committed=2`, `playback.acknowledged=2`, `history_committed=2`, `23` audio deltas, `0` errors, and `0` stale audio deltas. Focused validation also passed `214` handler tests, `51` protocol/attachment/actor tests, and `122` scheduler/connector/orchestrator tests. The existing standalone server-VAD failure-path test still times out independently and is outside this change.

Sy0307:
PTAK at this bugfix and can we merge it quickly? @amy-why-3459 @natureofnature

amy-why-3459:
Request changes — please do not merge this yet.

The four scheduler/connector/orchestrator fixes are real and the tests pin the right paths. Stage-1 reset before 4096, abort sweep of `skipped_waiting`, streaming-input counter resync, per-stage segment state, and the `save_async` generation watermark all match the crashes they describe. The KV release on the overflow path that @congw729 asked about is also there.

The problem is the second half. This branch is no longer the core-only stack that comes after #6170. `263e01de` / `2f39a954e` put Realtime lifecycle back in, and two of those changes fight the contract that already landed.

1. Do not restore `DuplexOverlapPolicy.AUTO`. #6170 left two modes: `listen_only` and explicit `barge_in_on_speech`. With `auto` out of the enum, `_normalize_overlap_policy("auto")` becomes `listen_only`. Putting `AUTO` back makes normalize keep `"auto"`, and `serving.py` has no AUTO arm — it falls through to the `overlap_short_ack_ms` / `overlap_barge_in_ms` heuristic and can barge in as `long_overlap_speech`. That is the third mode #6170 removed. Map `auto` → `listen_only` (or reject it). The new test that asserts normalize("auto") == "auto" is documenting the regression.

2. Late `playback.ack` recovery appends the previous assistant message at the current history tail. ACK_ONLY snapshots the text, then `register_history_item` + `truncate_history_item` `append`s it. `test_late_playback_ack_recovers_response_local_history_after_next_response_begins` allows this after the next `begin_response`. Once user-2 is already committed, assistant-1 lands after it and poisons later prompts / `conversation.item`. Register the pending item in `end_response` for ACK_ONLY. If the next user turn is already committed, reject the late ack or insert by response/turn — do not append.

Please also wait for the current Buildkite lanes. #14342 and the AMD/Intel/NPU jobs are still pending; the 8/28 required lanes failed. The overflow reset is a full Stage-1 session replace (codec history gone), not a live-KV trim — keep that wording in the description.

Sy0307:
Addressed both blockers in `379be9a92`.

1. Removed `DuplexOverlapPolicy.AUTO`. Legacy `"auto"` now normalizes to `listen_only`; the production-path regression asserts that long overlap speech stays `policy_listen_only` and cannot fall into `long_overlap_speech`.
2. ACK_ONLY now stages the response-owned assistant item in `end_response`. Each response carries its input-commit watermark; an ACK arriving after a later user commit returns `playback_ack_too_late`, and cancellation cannot append assistant A after user B. Timely ACK behavior is unchanged.

Validation on vLLM `0.28.0`: handler `217 passed, 1 deselected` (the separately known server-VAD timeout), protocol/attachment/actor `51 passed`, Ruff check/format passed. A fresh real A800 three-stage MiniCPM-o 4.5 two-turn `response-required` run also passed with `response.created=2`, `response.done=2`, `history_committed=2`, `23` audio deltas, `0` errors, and `0` stale deltas.

The Stage-1 overflow wording remains explicit: this is a full streaming-session replacement that drops prior codec history, not a live-KV trim. New CI is being triggered for this head.

Sy0307:
Merged current `main` containing #6626 and removed the overlapping Stage-1 context-reset implementation from this PR in `e8123a93`.

The duplicate `native_duplex_update` overflow reset and its standalone regression test are gone. Talker context rollover, `sliding_recompute`, prompt-window validation, receiver registration, and related Talker tests now come exclusively from merged #6626. The remaining #6529 diff is limited to its independent scheduler stale accounting, queue/counter, generation watermark, and Realtime response/playback lifecycle fixes.

Post-cleanup validation: A800 scheduler regression `22 passed`; Ruff check and format passed; PR-specific diff no longer contains MiniCPM Talker rollover files or `sliding_recompute` implementation.

Sy0307:
All previously raised code issues are addressed in `e8123a93`; the duplicate Stage-1 rollover was removed because merged #6626 now owns it, while the stale/generation and Realtime playback fixes remain in this PR. Focused tests and the real A800 two-turn `response-required` E2E pass; CUDA and Intel CI pass, while AMD/NPU are still failing and need separate investigation. Please re-review the current head when convenient.

vllm_omni/core/sched/omni_ar_scheduler.py
Sy0307: Resolved in `e8123a93`: the duplicate Stage-1 overflow reset was removed after #6626 merged, and Talker context rollover now comes from the base branch. The remaining PR changes cover independent scheduler and Realtime lifecycle fixes.

Sy0307:
@amy-why-3459 The requested changes are addressed in `379be9a92`, and the duplicate Stage-1 rollover was removed in `e8123a93` after #6626 merged. Focused tests and the real A800 two-turn `response-required` E2E pass; AMD/NPU CI failures remain separate vendor-lane issues. Please re-review the current head when convenient.

linyueqian (approved):
Reviewed the full diff at e8123a9 against merge-base 9a88d2d. The integration holds together well: the stale-frame double discard reported on the earlier head is properly fixed, with update_from_output draining `async_tokens_to_discard` together with `num_stale_output_tokens` on the same late frame so the next segment's first frame survives, and the regression test pins exactly that. The abort sweep now covering `skipped_waiting` closes a real stage-killing crash (upstream schedule() raises on a FINISHED_ABORTED request it admits), and the streaming-input counter resync closes the silent stage hang where has_work() read false with live requests parked in waiting. The per-stage keying of segment-boundary state in the orchestrator fixes genuine cross-stage contamination, and I verified no readers of the old flat fields remain and both `_duplex_output_context` call sites pass stage_id. The chunk adapter's segment-generation guard correctly prevents a late old-segment save_async from resurrecting the popped dedup watermark, with the boundary save carrying the generation captured before the queued update can mutate the request.

Validation: static review only, fork head fetched but never executed. CI at e8123a9: General 14383 and Intel 7595 green. AMD 11186 red is the known main-inherited FlashAttention tolerance failure in the diffusion suite (0.015625 vs 0.01, 2 of 4044+, no diffusion code touched here). NPU 6154's four failures all die at server boot on `ImportError: cannot import name 'constexpr_type' from 'triton.language.core'`, and main's scheduled nightly 6153 fails the same suites with the identical import error, so the NPU lane is environment-broken independent of this PR.

[suggestion] One non-blocking follow-up: `assistant_response_snapshots` in DuplexSession is only reclaimed when a history item is registered, deleted, or truncated, and the native auto-response path ends responses with commit_text=False without registering an item, so a long-lived session whose client never sends playback.ack retains one deep-copied text-plus-marks snapshot per turn. Worth bounding the dict or clearing superseded snapshots on turn completion in a follow-up.

PR #6526, opened 2026-08-23
[Model] Optimize MiniMax-H3 DLO component lifecycle

hsliuustc0106:
## Summary

- Retain reusable allocator blocks across MiniMax-H3 DLO encoder → DiT → VAE component boundaries while cached memory stays within 25% of device capacity and at least 5% remains physically free.
- Generalize `PinnedModuleStager` to stage one immutable module group with a shared copy stream/event while preserving parameter and buffer storage aliases, shapes, strides, offsets, and dtypes.
- Stage the encoder's non-block children from that immutable pinned CPU master, eliminating their GPU→CPU weight rematerialization.
- Force cache release and restore host masters on component/staging failure, missing telemetry, or an OOM retry. The ordinary DiT prefetch path is unchanged.

This remains a focused lifecycle change. Host Weight Runtime is now present on `main`, but this PR does not alter its selection, lease, staging-slot, or restore contracts.

## Review follow-up

Head `1843f05c` is rebased onto current `main` `29e3165b` and addresses all recent inline comments:

- `BaseException` during failure-path offload/cache cleanup is contained so forced release still runs and the original component failure is preserved.
- A repeated-lifecycle regression test covers `offload_to_cpu()` failing on the success path, forced cache release, and re-raise.
- The feature and module design docs now specify the 25% cached / 5% physical-free bounds, conservative telemetry fallback, OOM retry, memory tradeoff, and unchanged executor shutdown cleanup.

## Current matched L20X A/B

Both sides ran under one `gpu run` reservation on the same NVIDIA L20X (143,771 MiB, physical GPU 5) with fixed `cpunodebind=1,membind=0`, model/cache paths, prompt, seeds, and environment: Python 3.12.13, PyTorch 2.13.0+cu129, CUDA 12.9, driver 570.133.20, vLLM 0.27.0, and vLLM-Omni source at the listed SHAs.

Configuration: DP1/TP1, 1344×768, 5 seconds, 24 fps, `num_inference_steps=2` (one distilled denoising pass), `CUDNN_ATTN`, no DLO AllGather, zero resident layers, one excluded warmup, and three measured requests.

| Unprofiled metric | `main` (`29e3165b`) | This PR (`1843f05c`) | Delta |
|---|---:|---:|---:|
| Request latency, mean ± population stddev | 18.317 ± 0.267 s | 15.580 ± 0.032 s | **-14.94%** |
[...]

PR #6530, opened 2026-08-23
[Performance] Add per-component DLO offload policy

hsliuustc0106:
## Summary

- add `dlo_offload_components`, a boolean component map in the same style as per-component quantization configuration
- let users keep a discovered auxiliary component resident with `{"text_encoder": false}` while leaving unspecified components on their existing model-declared DLO lifecycle
- reject unknown component names and DiT paths; DLO continues to own DiT streaming

An empty map is the default and preserves current memory behavior. `true` permits existing offload behavior; it does not add offload support that a model has not declared. The optional `default` key handles unmatched encoders/VAEs.

## Dependency and scope

#6526 has merged into main. This draft is now rebased directly onto main (`780e8d40`) and consumes its encoder stager/cache lifecycle without copying or modifying allocator-cache retention, non-block staging, or the newer HWR host-registration path. The policy remains one DCO-signed commit.

This PR remains separate from #6486. It does not change DiT hooks, DLO collectives, VAE staging defaults, quantization, or CUDA compute.

## Usage

```python
omni = Omni(
    model="/path/to/model",
    enable_distributed_layerwise_offload=True,
    dlo_offload_components={"text_encoder": False},
)
```

CLI:

```bash
--dlo-offload-components '{"text_encoder": false}'
```

## Exact-recipe A/B

NVIDIA L20X, CUDA 13.0, Torch 2.13.0, vLLM 0.27.0. MiniMax-H3 FL2VA, DP1/TP1/SP1, DLO no-AllGather, 1344×768, 5 seconds, 24 fps, CUDNN_ATTN, request `num_inference_steps=2`. The current scheduler executes one denoising pass. One warm-up was excluded, followed by three measured requests.

| Text encoder | Measured requests | Mean | Peak HBM | Reserved plateau |
| offload `true` | 15.086 / 15.038 / 15.163 s | 15.096 ± 0.051 s | 22,722 MB | 23,825.7 MB |
| offload `false` | 14.022 / 14.054 / 14.142 s | 14.073 ± 0.051 s | 69,980 MB | 73,379.3 MB |

Keeping the whole text encoder resident saves 1.023 s (6.78%) in this paired run and adds 47,258 MB peak HBM / 49,554 MB reserved. Video `(124, 768, 1344, 3)` and audio `(1, 2, 165600)` hashes are byte-identical for every matching seed. Both configurations recover after an injected invalid request, have zero-spread three-request memory plateaus, and leave no child processes after shutdown.

## Nsight A/B

Same GPU and exact recipe; one warm-up followed by one captured measured request.

| Metric | offload `true` | offload `false` | Delta |
| --- | ---: | ---: | ---: |
| H2D payload | 130.856 GB | 79.197 GB | −51.658 GB |
| H2D device time | 2.392 s | 1.440 s | −0.953 s |
| Exposed H2D | 1.189 s | 0.272 s | −0.917 s |
| `encode_prompt` | 0.963 s | 0.050 s | −0.913 s |
| GPU request span | 15.566 s | 14.267 s | −1.300 s |
| CUDA kernel count | 311,426 | 311,426 | unchanged |
| CUDA kernel union | 12.555 s | 12.519 s | −0.036 s |

The option removes encoder H2D and associated host scheduling. It does not reduce CUDA compute; shared-host full-request timing outside the profiled phase remains noisy.

## Parallel topology validation

Paired `true`/`false` exact-recipe runs passed:

- DP1/TP1/SP1
- DP2/TP1/SP1
- DP1/TP2/SP1 with `text_encoder_tp_size=2`
- DP1/TP1/SP2
- DP4/TP1/SP1
- DP2/TP1/SP2
- DP1/TP4/SP1 with `text_encoder_tp_size=4`
- DP1/TP1/SP4

Every supported pair has byte-identical video/audio output, expected shapes, failure recovery, per-rank allocated/reserved/peak HBM, a zero-spread three-request reserved-memory plateau, and no remaining child processes.

Residency follows the model-owned encoder topology: +49,554 MB reserved on a TP1 encoder rank, +25,476 MB per TP2 rank, +13,506 MB per TP4 rank, and +0 MB on DP/SP parameter-free encoder peers. No-AllGather DP submissions are functional and serialized; these are not throughput-scaling claims.

Two requested mixed topologies are unsupported in both controls: DP2/TP2/SP1 and DP1/TP2/SP2. Ranks outside the partial-world TP2 encoder group hit the existing `assert self.cpu_group is not None` during MiniMax-H3 group construction; the component policy does not cause or resolve that startup failure.

## Reviewer-requested AllGather / encoder TP4 experiments

I also tested the two requested four-GPU topologies with DLO AllGather and `text_encoder_tp_size=4`. Each case used one same-prompt warm-up followed by three measured DP waves. Every measured wave contained two distinct prompts with the same seed, which detects cross-replica prompt collapse rather than hiding it behind different RNG seeds.

The runs used runtime head `79a0db1e`; the follow-up commit only documents these results and clarifies the CLI/help text.

| Topology | `true` / `false` completed | Policy A/B parity | Independent-prompt result | Resident reserved delta |
| --- | ---: | ---: | --- | ---: |
| DP2/TP2/SP1, encoder TP4 | yes / yes | byte-identical | unsupported: both requests byte-identical | +13,518 MB/rank |
| DP2/TP1/SP2, encoder TP4 | yes / yes | byte-identical | unsupported: both requests byte-identical | +13,470 MB/rank |

Both modes returned the expected video/audio shapes, recovered after an invalid asymmetric wave, held a zero-spread three-wave memory plateau on every rank, and left no child processes. Full-request timing was highly noisy and is not used as performance evidence.

These experiments show that the component policy preserves layouts and behaves identically with AllGather, but the requested encoder topology is not valid for independent DP requests. MiniMax-H3 currently creates one global encoder TP group and broadcasts rank 0's conditioning across the DiT world, so the second DP request receives the first request's prompt conditioning. Supporting this topology requires a separate replica-local encoder-group design; it is not a residency-policy change.

## Tests and checks

- 263 affected CPU tests passed on the rebased head
- Ruff, format, Markdown, SPDX, pytest-marker, forbidden-import, direct-CUDA, and diff-hygiene hooks passed
- the mypy hook reports the exact same 9 existing diagnostics on current main and this head; the new policy test has no mypy diagnostic
- exact-recipe DP1/TP1/SP1 resident-encoder smoke on tree `9d4396a3`: byte-identical pre-rebase hashes, 69,984 MB peak HBM, failure recovery, and no child processes
- InferMatrixCopilot Direct review completed on tree `9d4396a3` with no findings; final head `7db3a0cd` is a metadata-only DCO identity amend of the same tree

### DIFF
--- a/vllm_omni/engine/async_omni_engine.py
+++ b/vllm_omni/engine/async_omni_engine.py
@@ -1059,6 +1059,9 @@ class AsyncOmniEngine:
             "enable_distributed_layerwise_offload": kwargs.get("enable_distributed_layerwise_offload", False),
             "dlo_use_allgather": kwargs.get("dlo_use_allgather", True),
             "dlo_resident_layers": kwargs.get("dlo_resident_layers", 0),
+            "dlo_offload_components": (
+                {} if kwargs.get("dlo_offload_components") is None else kwargs["dlo_offload_components"]
+            ),
             "host_weight_runtime_mode": kwargs.get("host_weight_runtime_mode", "disabled"),
             "host_weight_runtime_root": kwargs.get("host_weight_runtime_root"),
             "dlo_host_registration_limit_gib": kwargs.get("dlo_host_registration_limit_gib", 0.0),
--- a/docs/design/feature/offloader/distributed_layerwise_offload.md
+++ b/docs/design/feature/offloader/distributed_layerwise_offload.md
@@ -29,6 +29,7 @@ Legend: ✅ supported, ⚠️ compatibility path or limited validation, ❌ unsu
 | **Other online quantization methods** | ❌ Rejected until runtime packing and scale layouts are validated. | ⚠️ Allowed through the ordinary loader; validation is method-specific. |
 | **Model-level or standard layerwise CPU offload** | ❌ Disabled because DLO takes priority. | ❌ Disabled because DLO takes priority. |
 | **Resident leading layers** | ❌ Rejected. | ✅ Requires eligible resident paths in the model's `OffloadPlan`. |
+| **Auxiliary component policy** | ✅ Encoders/VAEs may remain resident. | ✅ Encoders/VAEs may remain resident. |
 
 See [Parallelism compatibility](#parallelism-compatibility) and
 [Request and loading constraints](#request-and-loading-constraints) for the
@@ -419,6 +420,22 @@ failure.
 - **HSDP + DP or TP:** rejected independently by the diffusion parallel
   configuration.
 
+### Auxiliary component policy
+
+`dlo_offload_components` maps discovered encoder/VAE paths to booleans. A
+component set to `False` is materialized on the target device once and is not
+entered through the model's manual offload lifecycle. A `True` value permits
+the existing `OffloadPlan.on_demand_component_paths` and encoder block-hook
+behavior; it does not add support where the model declares none. The optional
+`default` key handles unmatched auxiliary components, and an empty map
+preserves the current policy.
+
+The policy is resolved after component discovery and before DLO mutates DiT
+storage. Unknown names, including DiT paths, are rejected. DLO continues to
+own DiT streaming regardless of this map, which avoids an invalid state where
+the loader has skipped ordinary DiT materialization but the backend is asked
+to keep that DiT resident.
+
 ## Request and loading constraints
 
 AllGather DP multi-concurrency requires:
@@ -429,6 +446,15 @@ AllGather DP multi-concurrency requires:
 
 The no-AllGather path does not impose these DLO-specific synchronized-wave
 requirements.
+
+MiniMax-H3 currently builds one global encoder TP group and broadcasts the
+resulting conditioning from global rank 0. Therefore an encoder TP group that
+spans multiple DP replicas is not valid for independent-prompt AllGather
+waves. Exact-recipe DP2/TP2 and DP2/SP2 experiments with encoder TP4 completed
+the collective schedule, but two distinct prompts with the same seed produced
+byte-identical video and audio in every measured wave. Enabling or disabling
+encoder offload produced paired-identical results, so this is a pre-existing
+model topology limitation rather than component-policy behavior.
 
 Direct checkpoint mmap can back either transfer path. It is currently limited
 to proven TP1, non-HSDP, non-online-quantized layouts. Other layouts use the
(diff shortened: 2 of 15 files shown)


vllm_omni/engine/async_omni_engine.py
```diff
@@ -1046,6 +1049,9 @@ def _create_default_diffusion_stage_cfg(kwargs: dict[str, Any]) -> list:
             "enable_distributed_layerwise_offload": kwargs.get("enable_distributed_layerwise_offload", False),
             "dlo_use_allgather": kwargs.get("dlo_use_allgather", True),
             "dlo_resident_layers": kwargs.get("dlo_resident_layers", 0),
```
hsliuustc0106, line 1061: does it only work for dit module?

hsliuustc0106 (commented):
does it work for DP2TP2 or DP2SP2 allgather with textencoder TP4? please provide more experiments

vllm_omni/engine/async_omni_engine.py
hsliuustc0106: No. This map controls only discovered auxiliary encoders and VAEs (text_encoder, video_vae, and audio_vae for MiniMax-H3). DiT remains owned by DLO, and DiT or unknown keys are rejected. I clarified the CLI help and docs in 3637fe54.

hsliuustc0106:
Ran the requested exact-recipe four-GPU A/B with DLO AllGather and `text_encoder_tp_size=4` (same-prompt warm-up, then three two-request waves with distinct prompts and identical seeds):

| Topology | `true` / `false` | Policy parity | Prompt isolation | Resident reserved delta |
| --- | --- | --- | --- | ---: |
| DP2/TP2/SP1 | completed / completed | byte-identical | failed: the two requests were byte-identical | +13,518 MB/rank |
| DP2/TP1/SP2 | completed / completed | byte-identical | failed: the two requests were byte-identical | +13,470 MB/rank |

Both modes returned the expected video/audio shapes, recovered after the invalid wave, had zero-spread three-wave memory plateaus on every rank, and left no child processes. Timings were noisy, so I am not using them as performance evidence.

The policy itself is not the cause: MiniMax-H3 currently creates one global encoder TP group and broadcasts rank 0's conditioning across the DiT world. Encoder TP4 therefore spans both DP replicas and collapses independent prompts in both the control and resident configurations. I documented this as an unsupported model topology in `3637fe54`; supporting it requires a separate replica-local encoder-group design.

PR #6531, opened 2026-08-24
[Quantization] Resolve ModelOpt algorithms from checkpoint metadata

linzhenpl07:
## Summary

- rebuild on-disk quantization configs through the mapping-based factory path
- preserve ModelOpt algorithm detection from `quant_algo` / `producer`
- compare an active config with the checkpoint's resolved algorithm rather than the generic `quant_method="modelopt"`
- cover NVFP4, FP8, serialized checkpoints, ignored layers, and AutoRound MXFP8 rebuilds

## Why

ModelOpt exports use the same `quant_method="modelopt"` label for different algorithms. The actual choice is stored in fields such as `quant_algo`. `resolve_quant_config_from_disk()` split the mapping into a method name plus kwargs before rebuilding it, bypassing the factory branch that distinguishes NVFP4 from FP8. A serialized NVFP4 checkpoint could therefore be rebuilt as the FP8 config or rejected as mismatched.

The fix keeps the complete disk mapping together whenever it is rebuilt or compared, so the existing factory performs the same algorithm detection for automatic loading as it does for an explicit mapping.

## Tests

```text
pytest -q tests/quantization/test_quant_config_from_disk.py
10 passed
```

This is also required for automatically loading the serialized LingBot World NVFP4 transformer used in #6227 experiments.

### DIFF
--- a/tests/quantization/test_quant_config_from_disk.py
+++ b/tests/quantization/test_quant_config_from_disk.py
@@ -1,0 +1,134 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+"""Tests for rebuilding a quantization config from a checkpoint's config.json.
+
+ModelOpt writes ``quant_method: "modelopt"`` for every algorithm it exports and
+distinguishes them with ``quant_algo``. Anything that selects a config class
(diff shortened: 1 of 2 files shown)


linzhenpl07:

PR #6532, opened 2026-08-24
[LingBot World] Enable serialized NVFP4 transformer loading

linzhenpl07:
## Summary

- enable quantization config plumbing through the LingBot World pipeline and transformer
- pass `quant_config` to supported vLLM parallel linear layers with checkpoint-aligned prefixes
- keep Q/K/V separate for quantized checkpoints so projection-specific ModelOpt global scales remain correct
- add a lightweight checkpoint-index dispatch auditor and CPU contract tests

## Why

LingBot World rejected every non-null quantization config, and its linear layers did not participate in vLLM's quantization dispatch. Simply removing those guards is not sufficient: the public ModelOpt NVFP4 export stores a different global scale for Q, K, and V, while the fused `QKVParallelLinear` path retains one merged scalar. On the public checkpoint those scales differ by up to 3.5x, so fusing the projections silently changes the model rather than merely reducing precision.

The unquantized path remains fused. The quantized path constructs separate Q/K/V projections matching the checkpoint layout and routes all supported parallel linears through the existing vLLM quantization kernels. Plain modulation `nn.Linear` layers remain unchanged.

## Validation

- checkpoint index audit: 340 quantized layers, 430 unquantized weights, and 1020 scale tensors, matching the model card
- real checkpoint dispatch selected `CutlassNvFp4LinearKernel`
- 768x480 single-session generation: BF16 eager 4.058 s, NVFP4 eager 3.348 s
- with the compiled-KV fix in #6463: BF16 compile 3.626 s, NVFP4 compile 2.845 s
- generated chunks remained finite in the real-checkpoint smoke test

## Tests

```text
167 CPU tests passed in the combined LingBot/dispatch suite
3 distributed CPU contract tests passed with a localhost Gloo group
```

Automatic reconstruction of the checkpoint's ModelOpt algorithm is submitted in #6531; without that companion change the same config can still be supplied explicitly.

Part of #6227's runtime optimization track.

### DIFF
--- a/tests/diffusion/models/lingbot_world/test_lingbot_world_attention.py
+++ b/tests/diffusion/models/lingbot_world/test_lingbot_world_attention.py
@@ -78,7 +78,12 @@ def _install_vllm_stubs() -> None:
             **kwargs,
         ) -> None:
             super().__init__()
-            del kwargs
+            # Record what the model hands the quantization dispatch. The stub
+            # cannot resolve a quant method, so tests assert on the plumbing:
+            # every quantizable projection must receive the config and a
+            # prefix rooted the way checkpoint exclusion lists are.
+            self.quant_config = kwargs.get("quant_config")
+            self.quant_prefix = kwargs.get("prefix", "")
             self.return_bias = return_bias
             self.weight = nn.Parameter(torch.empty(output_size, input_size))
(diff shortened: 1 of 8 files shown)


linzhenpl07:

PR #6533, opened 2026-08-24
[AR-Diffusion] Add session-owned streaming VAE decode

linzhenpl07:
## Summary

- add a session-owned `StreamingDecodeState` for Wan-family causal VAE decoders
- expose a narrow `SupportsStreamingDecode` lifecycle (`new_decode_state`, `decode_chunk`, `release`)
- keep the shared VAE module stateless while each AR-Diffusion session retains its own bounded temporal cache
- wire it into `LingBotWorldCausalDMDPipeline.post_decode()`, keyed by `request_id` and released
  from `close_ar_diffusion_session` / `reset_ar_diffusion_session`
- add a stepwise streaming example and CPU tests for chunk continuity, session isolation, causal
  frame geometry, bounded state, release semantics, and the non-streaming fallback

## Why

#6844 made one `generate()` one session: `post_decode()` emits one chunk per AR block and the
runner keys the AR session by `request_id`. Each block is decoded on its own, so the causal VAE
restarts per block and re-expands every block's opening latent frame to a single raw frame. As
that PR reports, an 81-frame request comes back as 63 frames and the served timeline does not line
up with an offline decode of the same latents.

The decoder can already decode incrementally. What it cannot do is carry its temporal cache
*between calls*: that cache is module-owned and cleared on every `decode()`. Sharing one module
across sessions therefore either restarts every chunk or lets sessions overwrite one another's
temporal context — both produce plausible but wrong video while every latency metric stays green.

This change moves ownership rather than rewriting the decoder: one state per session, threaded
through every committed chunk, released with the AR session.

## Where it plugs in

```mermaid
flowchart TD
    A["AsyncOmni.generate(request_id)"] --> B["ARDiffusionModelRunner.execute_stepwise"]
    B --> C["_bound_ar_session(session_id == request_id)<br/>binds runner-owned KV"]
    C --> D["prepare_encode"]
    D --> E["denoise_step"]
    E --> F["step_scheduler"]
    F -- "4 steps" --> E
    F --> G["post_decode()<br/>commit clean-x0 KV"]
    G --> H{"output_type"}
    H -- "latent" --> I["payload {latents}"]
    H -- "pixels" --> J["_decode_chunk_to_pixels(session_id=request_id)"]
    J --> K["WanStreamingDecoder.decode_chunk(latent, state)"]
    K <--> S[("StreamingDecodeState[request_id]<br/>one temporal cache per session")]
    K --> M["pixel chunk — 9 frames, then 12"]
    I --> N["streaming output"]
    M --> N
    N --> O{"request finished?"}
    O -- "no" --> D
    O -- "yes / aborted" --> P["_release_session()"]
    P --> Q["close_ar_diffusion_session(session_id)"]
    Q --> R["KV state and StreamingDecodeState released together"]
```

## Where the lifecycle lives

The pipeline holds one `StreamingDecodeState` per `request_id` and drops it from
`reset_ar_diffusion_session` / `close_ar_diffusion_session`. Those are the capability hooks
`ARDiffusionModelRunner._release_session()` already calls on reset, close, LRU eviction and a
failed forward, and `SupportsARDiffusionPipeline` documents them as the notification path for
model-owned non-KV state. So decoder state and KV state share one lifetime, `_bound_ar_session`
is untouched, and the model-neutral runner does not have to hold a model-specific temporal cache.

A chunk that fails mid-decode drops its own half-advanced cache rather than deferring to close:
the session's next chunk must not resume from a temporal context that was never completed.

VAEs that cannot stream keep the previous per-chunk decode behind a logged fallback. Tiled and
patch-parallel decode split a frame spatially and drive the module's own `feat_cache` across those
tiles, so threading a per-session cache through them is a separate design and is not attempted here.

## Correctness and measurements

From the CPU tests in this PR:

- chunking is bit-neutral against the same per-frame decode path run as one call
- interleaved sessions match their solo controls and do not mutate the module's own cache
- causal delivery geometry is 9 frames for the opening chunk and 12 for steady chunks, which is the
  frame timeline an offline whole-clip decode of the same latents produces; the fallback path
  delivers 9 per chunk, which is the frame loss this PR removes
- resident decoder state is bounded with session length, and a released state reports zero bytes

The equivalence control intentionally does not compare against `AutoencoderKLWan._decode`: its
whole-tensor `post_quant_conv` can choose a different BF16 kernel and measure that numerical
artifact instead of chunking. The control that isolates chunking is the same session decoded in a
single call through this same path.

Device measurements on the decoder itself, at 832x480 BF16 on one RTX PRO 6000: about 1801 MiB of
resident state per session, held flat across chunks, and 419.8 ms per steady chunk.

An independent end-to-end run on one H20 at this branch's `afc29c3d2`, reported in this PR's
comments, covers what the decoder alone cannot. Both image-to-video examples complete, where main
returns a short sequence, and measured against the native implementation at identical temporal
positions the streamed output scores higher on both metrics:

| Input | main | this PR |
| --- | --- | --- |
| Stonehenge | 0.6016 SSIM / 14.54 dB | 0.6164 SSIM / 15.40 dB |
| Lakeside tree | 0.5063 SSIM / 14.15 dB | 0.5097 SSIM / 15.05 dB |

Total latency there is 0.86-0.90% above main (70.55 s against 69.94 s) for a complete sequence
instead of a truncated one, and peak allocated VRAM is about 0.90 GiB lower.

Note which reference those figures use: the native implementation, not a one-shot decode of the
same latents, which is why they sit near 0.6 for main as well. #6844's 0.972 SSIM is the latter
comparison and is not re-measured here.

## Tests

```text
pytest -q tests/diffusion/ar_diffusion/test_streaming_decode.py          # 17 passed
pytest -q tests/diffusion/models/lingbot_world                           # 204 passed
```

No GPU end-to-end test is included: the served frame-count claim above belongs to #6844, and I
would rather add that test after running it than ship one that has never executed.

## What changed since the first review

The branch is rebased onto main with #6844, and the decoder now plugs into stepwise execution
instead of standing beside the deprecated tick entry point:

- the example no longer loads a VAE of its own; it drives one `generate()` with
  `step_execution=True` and `diffusion_streaming_output=True` and writes the streamed chunks
- the P1 review point about the missing latent rescale is now structural rather than fixed in the
  demo: the decode runs inside the pipeline, so it reuses the pipeline's own `_vae_latent_stats`
  and there is no second copy of that inversion to forget
- the P3 review point about releasing decoder state is likewise structural: release is driven by
  the AR session's close/reset notification, not by a caller's `finally`
- `release()` is declared on the `SupportsStreamingDecode` protocol, and
  `bytes_per_pixel_fp32` is a constructor argument, both as requested earlier

Part of #6227 Phase 1. Also unblocks #6841, which subclasses `WanStreamingDecoder` for
spatial-shard decode.

### DIFF
--- a/examples/offline_inference/diffusion/ar_diffusion_streaming_decode.py
+++ b/examples/offline_inference/diffusion/ar_diffusion_streaming_decode.py
@@ -151,10 +151,15 @@ def main(argv: Sequence[str] | None = None) -> int:
     ]
     plateaued = all(len(set(series[1:])) <= 1 for series in per_session_bytes if len(series) > 1)
 
-    # Isolation: each session's streamed output must equal its solo decode.
+    # Isolation: each interleaved session must equal a solo, single-call run
+    # through the same per-frame path.  Do not use ``vae._decode`` as the
+    # oracle: it applies post_quant_conv to the whole tensor and can select a
+    # different BF16 CUDA kernel, measuring that numerical artifact instead of
+    # whether chunking or session interleaving changed the result.
     isolated = True
     for session in range(args.sessions):
-        solo = vae._decode(latents[session], return_dict=False)[0]
+        solo_state = decoder.new_decode_state(f"solo-{session}")
+        solo = decoder.decode_chunk(latents[session], solo_state)
         streamed = torch.cat(emitted[session], dim=2)
         if streamed.shape != solo.shape or not torch.equal(streamed, solo):
             isolated = False
@@ -181,7 +186,7 @@ def main(argv: Sequence[str] | None = None) -> int:
     print(f"decoder state/session: {summary['resident_decoder_bytes_per_session'][0] / 2**20:.2f} MiB")
     print()
     print(f"bounded state (plateaued)            : {plateaued}")
-    print(f"streamed output == whole-clip decode : {isolated}")
+    print(f"streamed output == solo single call  : {isolated}")
     print(f"sessions produced distinct video     : {distinct}")
 
     if args.output_dir is not None:
--- a/examples/offline_inference/diffusion/ar_diffusion_streaming_decode.py
+++ b/examples/offline_inference/diffusion/ar_diffusion_streaming_decode.py
@@ -1,0 +1,197 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+"""Incremental decode for realtime AR-Diffusion sessions, end to end.
+
+Demonstrates the vertical path a realtime video session needs: several
+concurrent sessions, each advancing chunk by chunk, with frames emitted as
+soon as their chunk commits rather than after a whole-clip barrier.
+
+Runs on CPU with no checkpoint by default -- it builds a small Wan autoencoder
+from a config so the *protocol* can be exercised anywhere. Point ``--vae`` at a
+real checkpoint to run the shipped decoder instead; nothing else changes.
+
+What it shows
+
+* frames are delivered per chunk, so time-to-first-frame is one chunk's decode
+  rather than the whole session's;
+* resident decoder state plateaus -- the temporal cache is bounded by
+  construction, not by how long the session has run;
+* two interleaved sessions keep independent temporal context, checked against
+  a solo recording rather than assumed;
+* a session's opening chunk is shorter, because a causal decoder expands the
+  first latent frame to a single raw frame.
+
+Examples::
+
+    python examples/offline_inference/diffusion/ar_diffusion_streaming_decode.py \
+        --sessions 2 --chunks 6 --output-dir /tmp/stream-demo
+
+    python examples/offline_inference/diffusion/ar_diffusion_streaming_decode.py \
+        --vae <checkpoint> --height 480 --width 832 --sessions 1 --chunks 4
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import time
+from collections.abc import Sequence
+from pathlib import Path
+
+import torch
+
+from vllm_omni.experimental.ar_diffusion.streaming_decode import WanStreamingDecoder
+
+# A small Wan autoencoder: same causal structure as the shipped checkpoint,
+# narrower so the demo runs on a laptop CPU.
+_DEMO_VAE_CONFIG = {
+    "base_dim": 8,
+    "z_dim": 4,
+    "dim_mult": [1, 2],
+    "num_res_blocks": 1,
+    "temperal_downsample": [True],
+    "attn_scales": [],
+    "dropout": 0.0,
+}
+
+
+def parse_args(argv: Sequence[str] | None = None) -> argparse.Namespace:
+    parser = argparse.ArgumentParser(description=__doc__, formatter_class=argparse.RawDescriptionHelpFormatter)
+    parser.add_argument("--vae", default=None, help="VAE checkpoint or path. Omit for the built-in demo config.")
+    parser.add_argument("--sessions", type=int, default=2, help="Concurrent sessions, ticking round-robin.")
+    parser.add_argument("--chunks", type=int, default=6, help="Chunks each session generates.")
+    parser.add_argument("--latent-frames", type=int, default=3, help="Latent frames per chunk.")
+    parser.add_argument("--height", type=int, default=32, help="Output height in pixels.")
+    parser.add_argument("--width", type=int, default=32, help="Output width in pixels.")
+    parser.add_argument("--target-fps", type=float, default=16.0, help="Declared playout rate.")
+    parser.add_argument("--output-dir", type=Path, default=None, help="Write per-chunk frames and a summary here.")
+    parser.add_argument("--seed", type=int, default=0)
+    return parser.parse_args(argv)
+
+
+def build_vae(args: argparse.Namespace):
+    from diffusers import AutoencoderKLWan
+
+    if args.vae is not None:
+        return AutoencoderKLWan.from_pretrained(args.vae).eval()
+    torch.manual_seed(args.seed)
+    return AutoencoderKLWan(**_DEMO_VAE_CONFIG).eval()
+
+
+def spatial_ratio(vae) -> int:
+    return int(getattr(vae.config, "scale_factor_spatial", 8))
+
+
+@torch.no_grad()
+def main(argv: Sequence[str] | None = None) -> int:
+    args = parse_args(argv)
+    vae = build_vae(args)
+    for parameter in vae.parameters():
+        parameter.requires_grad_(False)
+    decoder = WanStreamingDecoder(vae)
+
+    ratio = spatial_ratio(vae)
+    latent_h, latent_w = args.height // ratio, args.width // ratio
+    if latent_h < 1 or latent_w < 1:
+        raise SystemExit(f"--height/--width must be at least the VAE spatial ratio ({ratio}).")
+
+    print(f"decoder causal convolutions : {decoder.num_causal_convs}")
+    print(f"output                      : {args.width}x{args.height} (latent {latent_w}x{latent_h})")
+    print(f"sessions                    : {args.sessions}, {args.chunks} chunks each")
+    print()
+
+    generator = torch.Generator().manual_seed(args.seed)
+    # One distinct latent stream per session: distinct inputs are what make the
+    # isolation check at the end discriminating rather than vacuous.
+    latents = [
+        torch.randn(1, vae.config.z_dim, args.chunks * args.latent_frames, latent_h, latent_w, generator=generator)
+        for _ in range(args.sessions)
+    ]
+
+    states = [decoder.new_decode_state(f"session-{i}") for i in range(args.sessions)]
+    emitted: list[list[torch.Tensor]] = [[] for _ in range(args.sessions)]
+    records: list[dict] = []
+    t_start = time.perf_counter()
+    first_frame_at: list[float | None] = [None] * args.sessions
+
+    for chunk_index in range(args.chunks):
+        for session in range(args.sessions):
+            window = slice(chunk_index * args.latent_frames, (chunk_index + 1) * args.latent_frames)
+            t0 = time.perf_counter()
+            frames = decoder.decode_chunk(latents[session][:, :, window], states[session])
+            latency = time.perf_counter() - t0
+            if first_frame_at[session] is None:
+                first_frame_at[session] = time.perf_counter() - t_start
+            emitted[session].append(frames)
+
+            record = {
+                "session": session,
+                "chunk_index": chunk_index,
+                "frames": int(frames.shape[2]),
+                "latency_s": latency,
+                "resident_decoder_bytes": states[session].nbytes(),
+            }
+            records.append(record)
+            print(
+                f"  session {session}  chunk {chunk_index}  "
+                f"{record['frames']:>2} frames  {latency * 1e3:7.1f} ms  "
+                f"decoder state {record['resident_decoder_bytes'] / 2**20:7.2f} MiB"
+            )
+            if args.output_dir is not None:
+                path = args.output_dir / f"session-{session}"
+                path.mkdir(parents=True, exist_ok=True)
+                torch.save(frames, path / f"chunk_{chunk_index:03d}.pt")
+        print()
+
+    wall = time.perf_counter() - t_start
+
+    # Boundedness: resident state must plateau, not grow with session length.
+    per_session_bytes = [
+        [r["resident_decoder_bytes"] for r in records if r["session"] == s] for s in range(args.sessions)
+    ]
+    plateaued = all(len(set(series[1:])) <= 1 for series in per_session_bytes if len(series) > 1)
+
+    # Isolation: each session's streamed output must equal its solo decode.
+    isolated = True
+    for session in range(args.sessions):
+        solo = vae._decode(latents[session], return_dict=False)[0]
+        streamed = torch.cat(emitted[session], dim=2)
+        if streamed.shape != solo.shape or not torch.equal(streamed, solo):
+            isolated = False
+    distinct = args.sessions < 2 or not torch.equal(torch.cat(emitted[0], dim=2), torch.cat(emitted[1], dim=2))
+
+    total_frames = sum(r["frames"] for r in records)
+    video_seconds = total_frames / args.target_fps
+    summary = {
+        "sessions": args.sessions,
+        "chunks_per_session": args.chunks,
+        "total_frames": total_frames,
+        "wall_s": wall,
+        "time_to_first_frame_s": first_frame_at,
+        "generated_fps": total_frames / wall if wall > 0 else None,
+        "rtf": wall / video_seconds if video_seconds > 0 else None,
+        "resident_decoder_bytes_per_session": [series[-1] for series in per_session_bytes],
+        "state_plateaued": plateaued,
+        "matches_solo_decode": isolated,
+        "sessions_produced_distinct_video": distinct,
+    }
+
+    print(f"time to first frame : {[f'{t:.3f}s' for t in first_frame_at]}")
+    print(f"total frames        : {total_frames} in {wall:.3f}s  ({summary['generated_fps']:.1f} fps)")
+    print(f"decoder state/session: {summary['resident_decoder_bytes_per_session'][0] / 2**20:.2f} MiB")
+    print()
+    print(f"bounded state (plateaued)            : {plateaued}")
+    print(f"streamed output == whole-clip decode : {isolated}")
+    print(f"sessions produced distinct video     : {distinct}")
+
+    if args.output_dir is not None:
+        args.output_dir.mkdir(parents=True, exist_ok=True)
+        (args.output_dir / "summary.json").write_text(json.dumps(summary, indent=2), encoding="utf-8")
+        (args.output_dir / "chunks.jsonl").write_text("".join(json.dumps(r) + "\n" for r in records), encoding="utf-8")
+        print(f"\nwrote {args.output_dir}")
+
+    return 0 if (plateaued and isolated and distinct) else 1
+
+
+if __name__ == "__main__":
+    raise SystemExit(main())
--- a/tests/diffusion/ar_diffusion/test_streaming_decode.py
+++ b/tests/diffusion/ar_diffusion/test_streaming_decode.py
@@ -1,0 +1,245 @@
+# SPDX-License-Identifier: Apache-2.0
+
+"""Streaming VAE decode against the real Wan autoencoder, on CPU.
+
+The autoencoder is built from a config rather than a checkpoint -- weights are
+random, only shapes and the temporal cache protocol matter -- so these run
+without a device or a download.
+
+The load-bearing test is equivalence: decoding a session chunk by chunk must
+produce exactly what decoding the whole clip produces. Everything else about
+streaming is worthless if that does not hold.
+"""
+
+from __future__ import annotations
+
+import pytest
+import torch
+
+from vllm_omni.experimental.ar_diffusion.streaming_decode import (
+    StreamingDecodeState,
+    SupportsStreamingDecode,
+    WanStreamingDecoder,
+)
+
+pytestmark = [pytest.mark.core_model, pytest.mark.cpu]
+
+# Small enough to decode on CPU in a test, same architecture as the shipped
+# checkpoint: base_dim 96 / dim_mult [1,2,4,4] would be far too slow, so the
+# widths are scaled down while the causal structure is unchanged.
+VAE_CONFIG = {
+    "base_dim": 8,
+    "z_dim": 4,
+    "dim_mult": [1, 2],
+    "num_res_blocks": 1,
+    "temperal_downsample": [True],
+    "attn_scales": [],
+    "dropout": 0.0,
+}
+LATENT_H = LATENT_W = 4
+
+
+@pytest.fixture(scope="module")
+def vae():
+    diffusers = pytest.importorskip("diffusers")
+    torch.manual_seed(0)
+    model = diffusers.AutoencoderKLWan(**VAE_CONFIG).eval()
+    for parameter in model.parameters():
+        parameter.requires_grad_(False)
+    return model
+
+
+@pytest.fixture(scope="module")
+def decoder(vae):
+    return WanStreamingDecoder(vae)
+
+
+def _latent(num_frames: int, *, seed: int = 0) -> torch.Tensor:
+    generator = torch.Generator().manual_seed(seed)
+    return torch.randn(1, VAE_CONFIG["z_dim"], num_frames, LATENT_H, LATENT_W, generator=generator)
+
+
+@torch.no_grad()
+def _decode_whole(vae, latent: torch.Tensor) -> torch.Tensor:
+    """The non-streaming reference: one call, cache cleared at both ends."""
+    return vae._decode(latent, return_dict=False)[0]
+
+
+# --------------------------------------------------------------------------
+# Equivalence: the property that makes streaming worth doing at all
+# --------------------------------------------------------------------------
+
+
+@torch.no_grad()
+@pytest.mark.parametrize("chunk_sizes", [(3, 3, 3), (1, 1, 1, 1, 1), (2, 4, 3), (9,)])
+def test_streaming_chunks_reproduce_the_whole_clip_decode(vae, decoder, chunk_sizes) -> None:
+    """Chunked decode must equal one-shot decode, for any chunking."""
+    total = sum(chunk_sizes)
+    latent = _latent(total)
+    reference = _decode_whole(vae, latent)
+
+    state = decoder.new_decode_state("s")
+    pieces = []
+    offset = 0
+    for size in chunk_sizes:
+        pieces.append(decoder.decode_chunk(latent[:, :, offset : offset + size], state))
+        offset += size
+    streamed = torch.cat(pieces, dim=2)
+
+    assert streamed.shape == reference.shape
+    torch.testing.assert_close(streamed, reference, rtol=0, atol=0)
+
+
+@torch.no_grad()
+def test_a_restarted_stream_does_not_match_a_continued_one(vae, decoder) -> None:
+    """Negative control: the equivalence test above must be discriminating.
+
+    Clearing state between chunks -- what the non-streaming path does on every
+    call -- has to produce a different result, otherwise the temporal cache
+    carries nothing and the assertion above is vacuous.
+    """
+    latent = _latent(6)
+    reference = _decode_whole(vae, latent)
+
+    restarted = torch.cat(
+        [
+            decoder.decode_chunk(latent[:, :, :3], decoder.new_decode_state("a")),
+            decoder.decode_chunk(latent[:, :, 3:], decoder.new_decode_state("b")),
+        ],
+        dim=2,
+    )
+    assert restarted.shape != reference.shape or not torch.equal(restarted, reference)
+
+
+# --------------------------------------------------------------------------
+# Cross-session isolation: the failure that timing metrics cannot see
+# --------------------------------------------------------------------------
+
+
+@torch.no_grad()
+def test_interleaved_sessions_keep_independent_temporal_context(vae, decoder) -> None:
+    """Two sessions ticking alternately must each match their solo decode.
+
+    A session that reads another's temporal cache produces a perfectly
+    plausible but wrong video, and every latency metric stays green, so this
+    has to be asserted against a recorded solo run rather than inferred.
+    """
+    latent_a = _latent(6, seed=1)
+    latent_b = _latent(6, seed=2)
+    solo_a = _decode_whole(vae, latent_a)
+    solo_b = _decode_whole(vae, latent_b)
+    assert not torch.equal(solo_a, solo_b), "negative control: the two sessions must differ"
+
+    state_a = decoder.new_decode_state("a")
+    state_b = decoder.new_decode_state("b")
+    out_a, out_b = [], []
+    for start in (0, 3):
+        window = slice(start, start + 3)
+        out_a.append(decoder.decode_chunk(latent_a[:, :, window], state_a))
+        out_b.append(decoder.decode_chunk(latent_b[:, :, window], state_b))
+
+    torch.testing.assert_close(torch.cat(out_a, dim=2), solo_a, rtol=0, atol=0)
+    torch.testing.assert_close(torch.cat(out_b, dim=2), solo_b, rtol=0, atol=0)
+
+
+@torch.no_grad()
+def test_the_module_cache_is_untouched_so_sessions_can_share_one_decoder(vae, decoder) -> None:
+    vae.clear_cache()
+    before = [entry for entry in vae._feat_map]
+    decoder.decode_chunk(_latent(3), decoder.new_decode_state("s"))
+    assert vae._feat_map == before, "streaming decode must not write the module-owned cache"
+
+
+# --------------------------------------------------------------------------
+# Causal geometry and boundedness
+# --------------------------------------------------------------------------
+
+
+@torch.no_grad()
+def test_opening_chunk_is_shorter_and_later_chunks_are_full_length(vae, decoder) -> None:
+    """(n - 1) * factor + 1 for the opening chunk, n * factor after it."""
+    factor = 2 ** sum(VAE_CONFIG["temperal_downsample"])
+    state = decoder.new_decode_state("s")
+    first = decoder.decode_chunk(_latent(3), state)
+    second = decoder.decode_chunk(_latent(3, seed=5), state)
+
+    assert first.shape[2] == (3 - 1) * factor + 1
+    assert second.shape[2] == 3 * factor
+    assert state.chunks_decoded == 2
+    assert state.frames_decoded == 6
+
+
+@torch.no_grad()
+def test_resident_state_does_not_grow_with_session_length(vae, decoder) -> None:
+    """Each causal convolution retains at most CACHE_T temporal slices."""
+    state = decoder.new_decode_state("s")
+    decoder.decode_chunk(_latent(3), state)
+    baseline = state.nbytes()
+    assert baseline > 0
+
+    for step in range(8):
+        decoder.decode_chunk(_latent(3, seed=step + 10), state)
+        assert state.nbytes() == baseline
+
+    assert sum(state.nbytes_by_device().values()) == baseline
+
+
+@torch.no_grad()
+def test_release_returns_the_session_to_its_pre_stream_state(vae, decoder) -> None:
+    state = decoder.new_decode_state("s")
+    decoder.decode_chunk(_latent(3), state)
+    assert state.started and state.nbytes() > 0
+
+    decoder.release(state)
+    assert state.nbytes() == 0
+    assert not state.started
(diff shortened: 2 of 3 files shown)


linzhenpl07:

examples/offline_inference/diffusion/ar_diffusion_streaming_decode.py
fhfuih, line 1: Seems this "example" script doesn't use any entrypoint at all (not even experimental `ARDiffusionSessionManager`) and is not an end-to-end script for user to run a model, but more like a script to test a feature component. Can you instead build on top of the existing `lingbot_world_v2_realtime.py` ?

linzhenpl07: Rebuilt on top of `lingbot_world_v2_realtime.py` -- same `AsyncOmni` + `ARDiffusionSessionManager` + JSONL event setup, kept close to verbatim. The only new step: each committed chunk's latent now goes straight into `WanStreamingDecoder` instead of being saved for later offline decode, carrying the causal VAE's temporal cache across chunks. Dropped the synthetic multi-session/no-checkpoint demo mode; `test_streaming_decode.py`'s tiny CPU config already owns equivalence/protocol correctness, so the example's job is just showing this end to end on the real model. Ran it on real hardware against the actual checkpoint (one RTX PRO 6000, 3 chunks): exit 0, every chunk finite, resident_decoder_bytes held at exactly 1888573440 bytes across all three chunks, time_to_first_frame was one chunk's generation + decode (7.92s) rather than the whole session's. Also spot-checked the decoded frames visually -- chunk 0's opening frame reproduces the i2v conditioning image and later frames continue coherently with no corruption.

examples/offline_inference/diffusion/ar_diffusion_streaming_decode.py
```diff
@@ -0,0 +1,295 @@
...
+            gen_started = time.perf_counter()
+            output = await session.next_chunk()
+            gen_elapsed = time.perf_counter() - gen_started
+            if len(output.images) != 1 or not isinstance(output.images[0], torch.Tensor):
+                raise RuntimeError("Expected one latent tensor from each realtime LingBot chunk.")
+            latent = output.images[0].detach().to(device=args.vae_device, dtype=vae_dtype)
+            metadata = consumer.chunk_metadata(output).to_dict()
+
+            decode_started = time.perf_counter()
+            frames = decoder.decode_chunk(latent, decode_state)
```
hsliuustc0106: [P1] Raw model-space latents are decoded without inverting the checkpoint's latent statistics — examples/offline_inference/diffusion/ar_diffusion_streaming_decode.py:247

The realtime tick path requires `output_type='latent'` and returns the raw generated latents; the pipeline's own decode path inverts the checkpoint normalization first — `(generated_latents * latent_std + latent_mean)` in `lingbot_world/pipeline.py:1277` (encode side `:797`). This example feeds `output.images[0]` straight into `decode_chunk`, whose docstring assumes rescaled input, so every chunk would decode to wrong video while latency/finite checks stay green. Fetch `vae.config.latents_mean/latents_std` and apply `latent * std + mean` (channel-shaped `[1,-1,1,1,1]`) before the call — or take the stats on `WanStreamingDecoder` and rescale inside `decode_chunk` so the contract can't be misused. If the visual spot-check already covered this, happy to be pointed at the rescale.

vllm_omni/experimental/ar_diffusion/streaming_decode.py
```diff
@@ -0,0 +1,225 @@
...
+    def declared_state_bytes(self, *, height: int, width: int, dtype: torch.dtype) -> int:
+        """Resident decoder bytes per session at this output shape.
+
+        Every cached tensor is ``[B, C_l, <= CACHE_T, H_l, W_l]`` with ``H_l``
+        and ``W_l`` fixed fractions of the output, so the total scales exactly
+        with ``height * width``. Callers that need the constant should measure
+        it once for their checkpoint; this helper scales a measured constant so
+        admission does not have to re-measure per resolution.
+        """
+        constant = getattr(self, "_bytes_per_pixel_fp32", None)
```
hsliuustc0106: [P3] `declared_state_bytes` always raises unless a caller sets a private attribute — vllm_omni/experimental/ar_diffusion/streaming_decode.py:218

The protocol's admission hook raises `RuntimeError` out of the box; it only succeeds after someone externally assigns `_bytes_per_pixel_fp32` (only the test does). Take the measured constant as a constructor argument (e.g. `WanStreamingDecoder(vae, bytes_per_pixel_fp32=...)`) or defer this method until the admission wiring lands.

examples/offline_inference/diffusion/ar_diffusion_streaming_decode.py
```diff
@@ -0,0 +1,295 @@
...
+                    "frame_shape": list(frames.shape),
+                    "finite": bool(torch.isfinite(frames).all()),
+                    "resident_decoder_bytes": decode_state.nbytes(),
+                    "metadata": metadata,
+                }
+            )
+            print(json.dumps(measurements[-1], sort_keys=True), flush=True)
+    finally:
+        try:
+            await manager.close_session(args.session_id)
```
hsliuustc0106: [P3] The reference consumer never releases the decode state — examples/offline_inference/diffusion/ar_diffusion_streaming_decode.py:270

The lifecycle here pairs decoder-state release with session close, and `SupportsStreamingDecode.release` exists precisely so a session runtime can release decoder state alongside KV state — but the demo's `finally` closes the session and engine without `decoder.release(decode_state)`. Add the release next to `close_session` so the example teaches the intended lifecycle.

examples/offline_inference/diffusion/ar_diffusion_streaming_decode.py
linzhenpl07: Confirmed, this was a real bug -- verified in code (lingbot_world/pipeline.py only inverts latents_mean/latents_std on the non-tick output_type="pixel" path; the tick path returns raw model-space latents) and fixed by applying the same rescale before decode_chunk. Re-ran end to end on a real GPU box against the actual checkpoint: same 3-chunk run, exit 0, all finite, resident_decoder_bytes unchanged at 1888573440 bytes across all three chunks. Visually the decoded frames are noticeably cleaner than the pre-fix run on the same input/seed -- the un-rescaled output had a smeared/dragged look this one doesn't, so this wasn't just a latent-statistics technicality.

vllm_omni/experimental/ar_diffusion/streaming_decode.py
linzhenpl07: Took the constructor-argument option: `WanStreamingDecoder(vae, bytes_per_pixel_fp32=...)`. Direct post-construction assignment still works too (the existing test uses that form unchanged), so this doesn't break anything -- it just gives callers who do want the constant a way to set it that can't be silently missed. Verified both forms produce the same declared_state_bytes() result.

examples/offline_inference/diffusion/ar_diffusion_streaming_decode.py
linzhenpl07: Added `decoder.release(decode_state)` ahead of `close_session` in the same finally chain, nested so a release failure still lets close_session/engine.shutdown() run. Re-ran end to end -- release ran with no exception, clean shutdown.

BruceLoveDecimal:
Tested this PR against the native implementation and main on a single NVIDIA H20 using two image-to-video examples; the results below cover latency, peak VRAM, output quality, and visual comparisons. And **end-to-end generation completed successfully for both examples**

1. Verified environment and hardware

| Item | Configuration |
|---|---|
| GPU | 1 × NVIDIA H20, 97,871 MiB reported memory |
| NVIDIA driver | 595.71.05 |
| OS / Python | Linux x86_64, kernel 5.15.0-130-generic / Python 3.12.3 |
| PyTorch / vLLM | 2.13.0 / 0.29.0 |
| Diffusers / Transformers | 0.40.0 / 5.14.1 |
| Native implementation | robbyant/lingbot-world-v2 @ `1895d300d8ac936401689b26389f51cbd36530eb` |
| Main | `bad50980c9e2679d4fa6352b0fbb121af9dfbb40` |
| PR #6533 | `afc29c3d230ffa32a29a28d92d3e8afcbeb5b3d1` |
| Recipe settings | 480 × 832, 81 requested frames, 4 DMD steps, flow shift 5, seed 42 |
| Execution | TP=1, eager, max_num_seqs=1, VAE tiling disabled; Omni KV memory fraction=0.6 |
| Precision | DiT/T5: BF16; native VAE: official FP32 default; Omni VAE: recipe BF16 |

The same preprocessed image files and checkpoint weights were used across groups. Native T5/VAE checkpoint keys were converted back to the official naming scheme and loaded with `strict=True`. Native used the vLLM-bundled FA2 kernel through a compatibility alias; Omni used its FA3 path. The native `max_area` upper bound was adjusted from 399360 to 399361 to avoid a floating-point floor issue and obtain the exact target resolution.

### 2. Performance and quality

Each group ran **one warm-up and three measured runs per input**, serially on the GPU. Latency is the mean ± standard deviation across measured runs, measured through CPU pixel availability; model initialization, file saving, video encoding, and network transport are excluded. Omni was tested through `AsyncOmni` with the in-process stepwise executor. Peak VRAM is the maximum **PyTorch CUDA allocated memory** across measured runs, including resident weights and the preallocated KV pool.

| Input | Implementation | Total latency (s) ↓ | Peak VRAM (GiB) ↓ |
|---|---|---:|---:|
| Stonehenge | Native | 88.294 ± 0.009 | 75.871 |
| Stonehenge | Main | 69.941 ± 0.008 | 84.133 |
| Stonehenge | PR #6533 | 70.569 ± 0.022 | 83.238 |
| Lakeside tree | Native | 88.539 ± 0.261 | 75.871 |
| Lakeside tree | Main | 69.943 ± 0.002 | 84.133 |
| Lakeside tree | PR #6533 | 70.547 ± 0.002 | 83.238 |

PR #6533 adds approximately **0.60–0.63 s (0.86–0.90%)** relative to main while restoring the complete requested output. Peak allocated VRAM is approximately **0.90 GiB lower** than main. Main produces an incomplete sequence, so its lower total latency should not be interpreted as equivalent completed work.

**Quality relative to native, evaluated at identical temporal positions:**

| Input | Implementation | SSIM ↑ | PSNR (dB) ↑ |
|---|---|---:|---:|
| Stonehenge | Main | 0.601557 | 14.5358 |
| Stonehenge | PR #6533 | 0.616397 | 15.4033 |
| Lakeside tree | Main | 0.506271 | 14.1518 |
| Lakeside tree | PR #6533 | 0.509697 | 15.0460 |

For a fair comparison, both candidates are evaluated on the same 63 native timestamps covered by main: 0–8, 12–20, 24–32, …, 72–80. No frame interpolation or duplication was applied. Metrics use uncompressed float32 RGB pixels in [0, 1]; SSIM uses a Gaussian window (sigma=1.5, population covariance), and PSNR is calculated from pooled pixel MSE.

For completeness, PR quality over its entire 81-frame output is **0.613714 SSIM / 15.3894 dB PSNR** for Stonehenge and **0.501416 / 14.9359 dB** for Lakeside tree.

### 3. Screenshots and videos

Screenshots compare **Native / Main / PR** from top to bottom at matched reference timestamps 0, 20, 44, and 80. Videos are encoded at 16 FPS for visual inspection only; encoded videos were not used for quality measurements. Main videos are shorter because their output is incomplete.

#### Stonehenge

image

**Native video**

https://github.com/user-attachments/assets/96703b1d-05e9-4af1-902c-516e91d97181

**Main video**

https://github.com/user-attachments/assets/89cdac74-8185-44dd-9f49-9a73851b95a0

**PR #6533 video**

https://github.com/user-attachments/assets/e36d2c81-2965-46b3-98c2-c597f8e747c3

#### Lakeside tree

image

**Native video**

https://github.com/user-attachments/assets/c266e295-d70c-43a9-82e3-d6a4547d6e28

**Main video**

https://github.com/user-attachments/assets/95ac9bdf-f1b9-4d97-bf1a-be977f4d583a

**PR #6533 video**

https://github.com/user-attachments/assets/967ecf63-4a6c-48f0-a1a5-3d12f9b354e9

vllm_omni/diffusion/models/lingbot_world/pipeline.py
```diff
@@ -1273,23 +1288,102 @@ def forward(self, req: DiffusionRequestBatch) -> DiffusionOutput:
...
+        vae_patch_parallel_size = int(
+            getattr(getattr(self.od_config, "parallel_config", None), "vae_patch_parallel_size", 1) or 1
+        )
+        if bool(getattr(self.vae, "use_tiling", False)):
+            reason = "VAE tiling is enabled"
+        elif vae_patch_parallel_size > 1:
+            reason = f"VAE patch parallelism is enabled (vae_patch_parallel_size={vae_patch_parallel_size})"
+        else:
+            try:
+                decoder = WanStreamingDecoder(self.vae)
```
tzhouam: **The per-session decoder cache is missing from session admission accounting.**

`ARDiffusionKVCacheSpec.model_owned_state_bytes_per_session` is documented as "a conservative upper bound for persistent per-session CUDA state outside runner-owned self/cross-attention KV", and `kv_cache/manager.py:284` subtracts `capacity * model_owned_state_bytes_per_session` from the budget before sizing the block pool.

LingBot still declares only `condition_bytes_per_session` at `pipeline.py:609` (~7 MiB), and this PR does not touch that declaration. But this decoder's cache is measured in this very PR at **1801 MiB per session** (module docstring line 27; `test_declared_state_bytes_scales_with_area_and_dtype` asserts 3602 MiB at fp32, which halves to 1801 at bf16) — roughly 250x the declared value, with `session_capacity=2`.

So two live stepwise sessions consume ~3.6 GiB that the KV sizing already handed to block pools: CUDA OOM partway through the second session's rollout, on a configuration the memory profiler reported as fitting.

The module even advertises `declared_state_bytes()` "for admission accounting" and `nbytes()` as "the quantity session admission has to account for" — neither is called from any production path (see the next comment).

The fix belongs at `pipeline.py:609`, which is outside this diff; anchoring here at the construction site.

vllm_omni/experimental/ar_diffusion/streaming_decode.py
```diff
@@ -0,0 +1,237 @@
...
+        """Resident decoder bytes per session at this output shape.
+
+        Every cached tensor is ``[B, C_l, <= CACHE_T, H_l, W_l]`` with ``H_l``
+        and ``W_l`` fixed fractions of the output, so the total scales exactly
+        with ``height * width``. Callers that need the constant should measure
+        it once for their checkpoint and pass it as ``bytes_per_pixel_fp32`` to
+        the constructor; this helper scales that measured constant so admission
+        does not have to re-measure per resolution.
+        """
+        constant = self._bytes_per_pixel_fp32
```
tzhouam, line 265: **`declared_state_bytes()` is unreachable in production, so the fix for the admission gap above can't just call it.**

The only production construction site is `pipeline.py:1315`, `WanStreamingDecoder(self.vae)` — no `bytes_per_pixel_fp32`. So `self._bytes_per_pixel_fp32 is None` for the only decoder that ever exists at runtime, and any real caller gets:

```
RuntimeError: declared_state_bytes needs a measured bytes-per-pixel constant for this checkpoint
```

Verified by constructing it exactly as the pipeline does.

The only exercise of this method is `test_streaming_decode.py:299`, which assigns the private `decoder._bytes_per_pixel_fp32` directly — precisely the "external attribute assignment after the fact" that the constructor argument's own docstring (line 150-154) says it exists to prevent. Net effect: the constructor argument is never supplied and this is dead API.

vllm_omni/experimental/ar_diffusion/streaming_decode.py
```diff
@@ -0,0 +1,237 @@
...
+
+        decoded_frames = []
+        for index in range(num_frames):
+            state.conv_idx[0] = 0
+            frame = self._vae.post_quant_conv(latent[:, :, index : index + 1])
+            # first_chunk marks the session's opening frame, not the call's:
+            # that is what makes chunk N + 1 continue chunk N rather than
+            # restart the causal expansion.
+            decoded_frames.append(
+                self._vae.decoder(
```
tzhouam: **This is the only VAE compute path in the repo that skips the autocast execution context.**

`OmniAutoencoderKLWan._execution_context()` wraps `encode()`, `decode()`, `decode_with_chunks()` and `tile_exec()` — every public compute entry point (`distributed/autoencoders/autoencoder_kl_wan.py`). It returns `current_omni_platform.create_autocast_context(...)` for fp16/bf16 params.

`decode_chunk` calls `self._vae.post_quant_conv` and `self._vae.decoder` directly, entering none of it. LingBot loads `DistributedAutoencoderKLWan`, which subclasses `OmniAutoencoderKLWan`, so the context does apply here.

On NPU this skips `torch.npu.amp.autocast` entirely (`platforms/npu/platform.py:353`), so streamed chunks are computed under a different numeric/dispatch context from the whole-clip decode this PR claims bit-equivalence with — and any future op autocast would promote to fp32 silently runs in low precision on this path only.

Suggest running the per-frame loop inside `self._vae._execution_context()`, or exposing a small public hook on the VAE for it.

tests/diffusion/ar_diffusion/test_streaming_decode.py
```diff
@@ -0,0 +1,305 @@
...
+    reference = _decode_whole(vae, latent)
+
+    restarted = torch.cat(
+        [
+            decoder.decode_chunk(latent[:, :, :3], decoder.new_decode_state("a")),
+            decoder.decode_chunk(latent[:, :, 3:], decoder.new_decode_state("b")),
+        ],
+        dim=2,
+    )
+    assert restarted.shape != reference.shape or not torch.equal(restarted, reference)
```
tzhouam: **This negative control is vacuous — it can never fail for the reason the docstring gives.**

```python
assert restarted.shape != reference.shape or not torch.equal(restarted, reference)
```

I ran it with this file's own `VAE_CONFIG`: with `temperal_downsample=[True]` (factor 2), restarting produces 5+5 = **10** frames while `_decode(6)` produces **11**.

```
reference.shape = (1, 3, 11, 8, 8)
restarted.shape = (1, 3, 10, 8, 8)
first disjunct (shape differs) = True
torch.equal arm ever evaluated? -> False
```

The first disjunct is always True, so `torch.equal` is never evaluated. The test passes purely because restarting changes the frame count — it would still pass against a decoder whose `feat_map` was write-only, which is exactly the failure the docstring says it guards against ("otherwise the temporal cache carries nothing and the assertion above is vacuous").

It also compares against `vae._decode`, the reference the neighbouring docstrings argue at length is not comparable.

A discriminating control would stream the same session with and without clearing state at the chunk boundary and compare the overlapping frames.

tests/diffusion/models/lingbot_world/test_pipeline_lingbot_world.py
```diff
@@ -2018,24 +2048,164 @@ def postprocess_video(self, video, *, output_type):
...
     state = _stepwise_state(num_frames=21)
     state.sampling.output_type = "np"
     fake = _FakeARState(state.request_id)
 
     with pipeline.bind_ar_diffusion_state(state.request_id, fake):
         outputs = _run_stepwise(pipeline, state)
 
-    # One decode per AR block, each seeing only that block's three latent frames.
-    assert [tuple(latents.shape) for latents in pipeline.vae.decode_inputs] == [(1, 16, 3, 2, 2)] * 2
     assert [output_type for _, output_type in processed] == ["np", "np"]
```
tzhouam, line 2177: **This deletion removes the only coverage of the stepwise latent rescale.**

The removed assertion was the one thing observing what actually reached the decoder on the stepwise path. The streaming path no longer routes through `_StubVAE.decode` (which appended to `decode_inputs`), and `_StubVAE.post_quant_conv = lambda latent: latent` (line 205) discards its input — so on the streaming branch nothing records anything.

`_decode_chunk_to_pixels` still does `(latents * latent_std + latent_mean).to(dtype=self.vae.dtype)` before streaming. A regression there — wrong stats, missing rescale, wrong dtype — now produces washed-out or saturated video with every test green.

Checked what survives: the identical assertion at line 2202 is inside `test_streaming_decode_falls_back_when_the_vae_cannot_stream`, i.e. the **fallback** path, and the value-level rescale assertion at line 1464 covers request mode. The streaming stepwise path has no rescale coverage at all.

Having the stub record the tensor it receives in `post_quant_conv` would restore it.

vllm_omni/experimental/ar_diffusion/streaming_decode.py
```diff
@@ -0,0 +1,237 @@
+# SPDX-License-Identifier: Apache-2.0
```
tzhouam, line 1: **Missing SPDX copyright header — CI will reject this.**

Verified by running the repo's own hook:

```
$ python tools/pre_commit/check_spdx_header.py <the three new/edited files>
The following files are missing the SPDX header:
  vllm_omni/experimental/ar_diffusion/streaming_decode.py
  tests/diffusion/ar_diffusion/test_streaming_decode.py
  examples/offline_inference/diffusion/ar_diffusion_streaming_decode.py
exit 1
```

This file and `tests/diffusion/ar_diffusion/test_streaming_decode.py` need:

```python
```

And `examples/offline_inference/diffusion/ar_diffusion_streaming_decode.py:2` carries the legacy text ("Copyright contributors to the vLLM project"), which the same hook rewrites and fails on.

vllm_omni/diffusion/models/lingbot_world/pipeline.py
```diff
@@ -1273,23 +1288,102 @@ def forward(self, req: DiffusionRequestBatch) -> DiffusionOutput:
...
+        if self._cached_streaming_decoder is not None:
+            return self._cached_streaming_decoder
+        decoder: WanStreamingDecoder | None = None
+        reason: str | None = None
+        vae_patch_parallel_size = int(
+            getattr(getattr(self.od_config, "parallel_config", None), "vae_patch_parallel_size", 1) or 1
+        )
+        if bool(getattr(self.vae, "use_tiling", False)):
+            reason = "VAE tiling is enabled"
+        elif vae_patch_parallel_size > 1:
```
tzhouam: **This branch is unreachable.**

`_validate_parallel_config(od_config)` runs at `pipeline.py:453`, the first statement of `__init__`, and raises `NotImplementedError("LingBot World v1 does not support VAE parallelism (vae_patch_parallel_size=N)")` for any `vae_patch_parallel_size > 1`.

A `LingBotWorldCausalDMDPipeline` therefore cannot exist with that value set, so neither the `getattr(getattr(self.od_config, "parallel_config", None), ...)` lookup above nor this `elif` can ever fire. It reads as a defensive check that maintainers will feel obliged to keep in sync with a constraint enforced elsewhere.

Suggest deleting both and keeping only the `use_tiling` check.

vllm_omni/diffusion/models/lingbot_world/pipeline.py
```diff
@@ -1273,23 +1288,102 @@ def forward(self, req: DiffusionRequestBatch) -> DiffusionOutput:
             stage_durations=self.stage_durations if hasattr(self, "stage_durations") else None,
         )
 
+    def _streaming_decoder(self) -> WanStreamingDecoder | None:
```
tzhouam, line 1354: **Nit: the tiling check is coarser than the condition it is standing in for.**

`AutoencoderKLWan._decode` only takes the tiled path when `width > tile_latent_min_width or height > tile_latent_min_height` (diffusers `autoencoder_kl_wan.py:1192`). This check treats the `use_tiling` flag alone as disqualifying, so `--vae-use-tiling` at a sub-threshold resolution needlessly falls back to per-chunk decode — silently losing a frame per block on a configuration that would never have tiled anyway.

Matching the diffusers threshold here would keep streaming in that case.

(I also considered whether caching the decision permanently could go stale if `use_tiling` changed later, but `use_tiling` is set once at load in `registry.py:445` and nothing flips it at runtime, so the caching itself looks fine.)

vllm_omni/diffusion/models/lingbot_world/pipeline.py
```diff
@@ -1273,23 +1288,102 @@ def forward(self, req: DiffusionRequestBatch) -> DiffusionOutput:
...
+        the rollout loses frames and continuity at each block boundary.
 
         The registered post-process hook returns a ``{"payload", "metadata"}``
         envelope untouched, so the pixel conversion it would do for a bare
         tensor has to happen here instead.
         """
         latent_mean, latent_std = self._vae_latent_stats(latents)
         vae_latents = (latents * latent_std + latent_mean).to(dtype=self.vae.dtype)
-        video = self.vae.decode(vae_latents, return_dict=False)[0]
+        decoder = None if session_id is None else self._streaming_decoder()
```
tzhouam: **The streaming branch drops the `vae.decode` profiler stage.**

`setup_diffusion_pipeline_profiler(profiler_targets=[..., "vae.decode", ...])` at `pipeline.py:554` resolves `vae.decode` and replaces it with a timing wrapper via `wrap_methods_by_paths` (`diffusion_pipeline_profiler.py:77`).

The fallback branch just above calls `self.vae.decode` and so is timed. The streaming branch reaches `self._vae.decoder(...)` through `WanStreamingDecoder.decode_chunk`, which is not the wrapped method. Since every `DiffusionOutput` from `post_decode` carries `stage_durations=self.stage_durations`, a streamed rollout with `enable_diffusion_pipeline_profiler` on reports **zero decode time** — the exact stage this PR is optimizing — and any perf regression in the streaming decoder is invisible to the profiler.

tests/diffusion/ar_diffusion/test_streaming_decode.py
```diff
@@ -0,0 +1,305 @@
...
+    torch.testing.assert_close(torch.cat(out_b, dim=2), solo_b, rtol=0, atol=0)
+
+
+# --------------------------------------------------------------------------
+# Cross-session isolation: the failure that timing metrics cannot see
+# --------------------------------------------------------------------------
+
+
+@torch.no_grad()
+def test_interleaved_sessions_keep_independent_temporal_context(vae, decoder) -> None:
```
tzhouam: **This duplicates `test_isolation_is_checked_against_each_session_decoded_alone` (line 152).**

Diffing the two bodies, they differ only in name, docstring, RNG seeds (1/2 vs 11/12) and line-wrapping. Same setup (two 6-frame latents), same solo streams, same "the sessions must differ" negative control, same `for start in (0, 3)` interleave, same per-session assertions.

That doubles the runtime of the slowest tests in this file while covering one property, and any future change to the isolation contract has to be edited in two places.

Suggest keeping one, parametrized over both seed pairs if both matter.

tests/diffusion/models/lingbot_world/test_pipeline_lingbot_world.py
```diff
@@ -2018,24 +2048,164 @@ def postprocess_video(self, video, *, output_type):
...
+    with pipeline.bind_ar_diffusion_state(state.request_id, fake):
+        _run_stepwise(pipeline, state)
+
+    pipeline.reset_ar_diffusion_session("req-reset")
+    assert pipeline._streaming_decode_states == {}
+    # Releasing a session that never decoded is a no-op, not an error.
+    pipeline.reset_ar_diffusion_session("req-reset")
+
+
+def test_streaming_decode_keeps_concurrent_sessions_isolated(monkeypatch) -> None:
```
tzhouam: **This test never interleaves the two sessions, so it doesn't test its own name.**

```python
for state in (first, second):
    with pipeline.bind_ar_diffusion_state(...):
        _run_stepwise(pipeline, state)
```

`req-a` runs to completion before `req-b` starts. The assertions that follow — `set(caches) == {"req-a", "req-b"}`, `caches["req-a"].feat_map is not caches["req-b"].feat_map`, and the per-session `first_chunk_flags` count — establish that the dict is keyed by request id and that each session opens its own causal expansion. None of them would catch chunk N of `req-b` reading `req-a`'s cache mid-rollout, which is the failure the docstring describes.

The pattern the runner actually exposes at `session_capacity=2` is A-chunk0, B-chunk0, A-chunk1, ... — exercised at the decoder level by `test_interleaved_sessions_keep_independent_temporal_context`, but nowhere at the pipeline level.

Driving the two `_run_stepwise` generators alternately would close that gap.

vllm_omni/experimental/ar_diffusion/streaming_decode.py
```diff
@@ -0,0 +1,237 @@
...
+        if not isinstance(counts, dict) or "decoder" not in counts:
+            raise TypeError("The autoencoder does not expose cached decoder convolution counts.")
+        return int(counts["decoder"])
+
+    def new_decode_state(self, session_id: str) -> StreamingDecodeState:
+        if not isinstance(session_id, str) or not session_id.strip():
+            raise ValueError("session_id must be a non-empty string.")
+        return StreamingDecodeState(session_id=session_id, feat_map=[None] * self.num_causal_convs)
+
+    def decode_chunk(self, latent: torch.Tensor, state: StreamingDecodeState) -> torch.Tensor:
```
tzhouam, line 191: **`decode_chunk` silently drops the `use_slicing` guarantee of the method it replaces.**

`AutoencoderKLWan.decode` splits on `self.use_slicing and z.shape[0] > 1`, decoding each sample with its own cleared cache, and `registry.py:443` sets `model.vae.use_slicing = od_config.vae_use_slicing` on this VAE. This wrapper does neither — it feeds `[B, C, 1, H, W]` slices through a single shared `state.feat_map`, which is correct only for `B == 1`.

This repo already treats the combination as a real hazard elsewhere: `distributed/autoencoders/autoencoder_kl_wan.py:87` raises for `use_slicing` with batch > 1 rather than silently proceeding.

LingBot currently enforces `num_reqs == 1` in `_parse_request`, so this is latent rather than live, but there's no `latent.shape[0] != 1` assertion to catch the day that changes.

tests/diffusion/ar_diffusion/test_streaming_decode.py
```diff
@@ -0,0 +1,305 @@
...
+
+
+def test_session_id_must_be_meaningful(decoder) -> None:
+    with pytest.raises(ValueError, match="session_id"):
+        decoder.new_decode_state("  ")
+
+
+def test_declared_state_bytes_scales_with_area_and_dtype(decoder) -> None:
+    # Measured for the shipped checkpoint: 37832 KiB of fp32 cache at 64x64.
+    decoder._bytes_per_pixel_fp32 = 37832 * 1024 / (64 * 64)
```
tzhouam: **This mutates a module-scoped fixture with no restore.**

The `decoder` fixture is `@pytest.fixture(scope="module")` (line 52), so `decoder._bytes_per_pixel_fp32 = ...` permanently changes the shared instance for the rest of the module. It happens to be the last test in file order today, so it passes — but under `pytest-randomly`, any reorder, or a `-k` selection, a later test asserting the default `None` behaviour would see the leaked constant.

Constructing a local `WanStreamingDecoder(vae, bytes_per_pixel_fp32=...)` here is both leak-free and the intended use of the constructor argument — which is otherwise never passed anywhere (see my comment on `streaming_decode.py:230`).

tzhouam (commented):
Reviewed at `afc29c3d2`, with the claims checked against a local checkout rather than read off the diff. Comments inline; the short version:

**Blocking**
- The per-session decoder cache — 1801 MiB by this PR's own measurement — is absent from `model_owned_state_bytes_per_session`, so AR-Diffusion admission sizes the KV pool as if it did not exist. Two concurrent stepwise sessions should OOM on a configuration the profiler cleared.
- `declared_state_bytes()`, the helper that would fix that, is unreachable: the one production construction site never passes `bytes_per_pixel_fp32`, so it raises. Confirmed by constructing it the way the pipeline does.
- The SPDX pre-commit hook fails on three files. Confirmed by running `tools/pre_commit/check_spdx_header.py`.

**Correctness**
- `decode_chunk` is the only VAE compute path that skips `OmniAutoencoderKLWan._execution_context()`, so streamed chunks run without autocast — which also undercuts the bit-equivalence claim.
- The streaming branch bypasses the wrapped `vae.decode`, so the `vae.decode` profiler stage reports zero for every streamed chunk.

**Test evidence**
The design here is sound — session-owned state released through the AR session's close/reset hooks is the right shape, and it matches how `SupportsARDiffusionPipeline` documents model-owned non-KV state. But the tests support it less than the description suggests: the negative control short-circuits on a frame-count difference and never evaluates its equality arm (reproduced: 11 frames vs 10), the pipeline-level "concurrent sessions" test never interleaves the sessions, the stepwise rescale assertion was deleted with nothing replacing it on the streaming path, and two decoder tests are identical apart from seeds.

I could not run the two suites cited in the description — my environment's vLLM predates current `main` — so those pass counts are unverified here, not disputed.

Rest are nits, marked as such.

vllm_omni/diffusion/models/lingbot_world/pipeline.py
```diff
@@ -1273,23 +1288,102 @@ def forward(self, req: DiffusionRequestBatch) -> DiffusionOutput:
...
+        return decoder
+
+    def _streaming_decode_state(self, decoder: WanStreamingDecoder, session_id: str) -> StreamingDecodeState:
+        state = self._streaming_decode_states.get(session_id)
+        if state is None:
+            state = decoder.new_decode_state(session_id)
+            self._streaming_decode_states[session_id] = state
+        return state
+
+    def _release_streaming_decode_state(self, session_id: str) -> None:
```
tzhouam, line 1391: **Nit: both branches do the same thing, and one is unreachable.**

A `StreamingDecodeState` only enters `_streaming_decode_states` via `_streaming_decode_state(decoder, ...)`, which requires a non-`None` decoder that is then stored in `_cached_streaming_decoder`. So by the time a state exists, `self._cached_streaming_decoder` is non-`None` and the `state.release()` branch is dead.

And `WanStreamingDecoder.release(state)` is a one-line `state.release()`, so the live branch is identical to the dead one anyway. The whole body reduces to:

```python
state = self._streaming_decode_states.pop(session_id, None)
if state is not None:
    state.release()
```

examples/offline_inference/diffusion/ar_diffusion_streaming_decode.py
```diff
@@ -0,0 +1,284 @@
...
+
+from __future__ import annotations
+
+import argparse
+import asyncio
+import json
+import math
+import time
+from collections.abc import Sequence
+from pathlib import Path
```
tzhouam, line 42: **These constants are hardcoded while `--model` is user-settable.**

`num_frames = (num_chunks * _FRAMES_PER_BLOCK - 1) * _TEMPORAL_COMPRESSION + 1` (line 181) goes straight into `OmniDiffusionSamplingParams`, and `--action-script` block length is validated against `_FRAMES_PER_BLOCK` (line 95).

The pipeline derives both quantities from the checkpoint instead — `transformer.config.num_frames_per_block` (`pipeline.py:598`) and `vae.config.scale_factor_temporal` (`pipeline.py:538`) — and caps at `_MAX_RAW_FRAMES = 117`.

They line up for the default checkpoint (10 x 3 blocks -> 117 frames exactly). Point `--model` at one with `num_frames_per_block != 3` or `scale_factor_temporal != 4` and the script builds an action script of the wrong per-block length and a `num_frames` inconsistent with the model — either a hard `ValueError` deep in `_parse_request`, or `matches_offline_frame_timeline: false` in summary.json with no indication that the constants, not the streaming decode, were at fault.

Reading them off the loaded model (or asserting them against it) would make the example safe to repoint.

tzhouam (approved):
LGTM

PR #6534, opened 2026-08-24
[Benchmark] Measure and overlap AR-Diffusion RGB delivery

linzhenpl07:
## Draft / stacked dependencies

This PR is intentionally a draft. It depends on:

- #6444 for the multi-session realtime harness
- #6533 for session-owned streaming VAE decode

The current diff temporarily includes both prerequisite branches. I will rebase it onto `main` after they land so that the final review contains only the vertical-path changes.

## Summary

- extend the realtime benchmark from latent readiness to actual RGB delivery
- add `DecodingSession` and report decoder time/share, delivered frame geometry, and resident decoder bytes per session
- add `OverlappedDecodingSession` with an explicit bounded generation lookahead
- overlap DiT generation for chunk N+1 with VAE decode for chunk N
- report outstanding generations instead of a `backpressure_s` metric that could never become nonzero in the caller-driven API

## Why

The existing harness ends when a latent chunk is ready, but users cannot display latent tensors. Its TTFF/RTF therefore omit VAE decode and do not describe RGB delivery. The omission becomes more important as TP reduces generation time while the single-device decoder remains a floor.

The first overlap implementation used `ensure_future()`, but synchronous decode immediately blocked the event loop, so the generation request was never sent until decode completed. The corrected path yields control after filling the lookahead, and the tests assert the actual event ordering rather than only the final duration.

`lookahead=1` is both the overlap window and the backpressure/memory bound: generation cannot run arbitrarily far ahead of the RGB consumer.

## Tests

```text
pytest -q \
  tests/diffusion/ar_diffusion/test_realtime_benchmark.py \
  tests/diffusion/ar_diffusion/test_streaming_decode.py
61 passed
```

## Follow-up validation

The CPU tests use a deterministic fake clock and assert ordering, bounds, frame accounting, and lifecycle behavior. End-to-end H200 generation/decode overlap remains a required hardware validation item under #6227; until that run is complete, `max(generation, decode)` is a target, not a claimed measured result.

Part of #6227 Phase 1.

### DIFF
--- a/benchmarks/ar_diffusion/README.md
+++ b/benchmarks/ar_diffusion/README.md
@@ -1,0 +1,97 @@
+# Realtime AR-Diffusion session benchmark
+
+Multi-session load harness for realtime AR-Diffusion serving: session arrivals,
+playout deadlines, and the metrics a streaming video deployment is actually
+gated on.
+
+## Load modes
+
+| Mode | Question it answers | Deadlines |
+|---|---|---|
+| `saturating` | How fast can this configuration go at all? Aggregate generated FPS, frames per GPU-second. | none |
+| `paced` | Does it keep up with playback? CPR, TTFC, worst-case per-chunk latency. | yes |
+
+Running only `saturating` produces a throughput number no deployment can use.
+Running only `paced` hides the ceiling, so a deadline miss cannot be attributed
+to insufficient capacity rather than to scheduling. Report both.
+
+## Load parameters
+
+Session **arrival rate** is the only valid open-loop axis. Per-session tick rate
+is not: ticks of one session are strictly ordered and state-dependent, so tick
+`k + 1` cannot be issued before chunk `k` returns and raising the rate only
+grows a queue. Concurrently active sessions are derived state, reported
+alongside every measurement as `peak_concurrent_sessions`.
+
+## Metrics
+
+Per session: `ttfc_s`, chunk latency P50/P95/P99, and
+
+```
+RTF = wall(K chunks) / (K * frames_per_chunk / target_fps)
+```
+
+Aggregate: `continuous_play_ratio` (CPR, the fraction of chunks delivered before
+their playout deadline, macro-averaged over sessions), `worst_case_chunk_latency_s`
+across all active sessions, `generated_fps`, and `frames_per_gpu_second` — the
+only unit comparable across replica counts, TP degrees and hardware.
+
+CPR is not redundant with RTF. RTF is a ratio of aggregates: a session can hold
+`RTF = 0.9` over a minute and still stutter visibly, because chunks arrive in
+bursts that individually miss their playout instants.
+
+Deadlines assume a declared playout grid. Playback starts once `--buffer-chunks`
+chunks are buffered; chunk `buffer_chunks - 1` anchors the grid and every later
+chunk is due one release period after its predecessor. Chunks inside the
+prebuffer have no deadline — their cost is start latency, which TTFC reports.
+
+## Model neutrality
+
+The chunk shape comes from the pipeline's declared `ARDiffusionKVCacheSpec`. No
+model name appears in this directory.
+
+One value is not derivable from the capability today: every frame count in the
+spec is in *latent* frames and nothing converts them to delivered frames, so the
+playout grid cannot be computed from the spec alone. Until the spec declares it,
+supply `--vae-temporal-factor` (default `1`, correct for a decoder that does not
+compress time).
+
+## Usage
+
+```bash
+# Ceiling
+python -m benchmarks.ar_diffusion.run_realtime_benchmark \
+    --model <checkpoint> --prompt "..." \
+    --num-sessions 1 --mode saturating --chunks 32 \
+    --vae-temporal-factor 4 \
+    --note "checkpoint=<...>" --note "hw=1xH200" --note "res=480x832" \
+    --note "steps=4" --output runs/n1.json
+
+# State concurrency N, execution concurrency 1
+python -m benchmarks.ar_diffusion.run_realtime_benchmark \
+    ... --num-sessions 2 --output runs/n2.json
+```
+
+At least one `--note` is required: a latency without model, checkpoint,
+resolution, denoising steps and hardware is not a result.
+
+## Reading N=1 against N=2
+
+With state concurrency `N` but execution concurrency 1, ticks serialize, so:
+
+```
+aggregate generated FPS(N)  ~=  aggregate generated FPS(1)
+per-session RTF(N)          ~=  N x RTF(1)
+```
+
+`compare_runs()` reports `generated_fps_ratio` and `per_tick_switching_cost_s`,
+the latter being latency beyond what pure queueing predicts. A measurable drop
+in aggregate FPS is per-tick switching cost — state bind/unbind, KV pool paging,
+conditioning rebuild — and that number is the baseline any cross-session
+batching work has to beat.
+
+## Tests
+
+`tests/diffusion/ar_diffusion/test_realtime_benchmark.py` drives the whole load
+model on a virtual clock against fake sessions: CPU only, no engine, no device,
+no checkpoint. `engine_binding.py` is the only module not covered.
--- a/benchmarks/ar_diffusion/README.md
+++ b/benchmarks/ar_diffusion/README.md
@@ -52,6 +52,26 @@ which TTFC reports. Walking cumulative frames rather than multiplying a constant
 period is what lets a deeper buffer grant real slack, and what keeps a causal
 decoder's shorter opening chunk from being credited with a full period.
 
+## The vertical path
+
+A session's tick returns latents, which nothing can watch. `DecodingSession`
+wraps a latent-producing session so each tick also decodes its chunk
+incrementally, which puts three things inside the measurement:
+
+- **TTFF rather than time-to-first-chunk.** The gate this RFC proposes is time
+  to first *frame*, so the decode has to be on the measured path.
+- **`decode_share`**, the fraction of chunk wall time spent decoding. With
+  generation and decode serialized this is the headroom overlapping them can
+  recover; once they overlap it is the number that has to shrink. Overlap work
+  without this baseline has nothing to beat.
+- **`resident_decoder_bytes_per_session`**, which admission needs and which no
+  other part of the run reports.
+
+Delivered frames then come from the decoder rather than from the profile, so
+the harness *measures* the temporal geometry instead of assuming it. Sessions
+that only produce latents keep working: the stage split is reported as `None`
+and the profile supplies the frame counts.
+
 ## Model neutrality
 
 The chunk shape comes from the pipeline's declared `ARDiffusionKVCacheSpec`. No
--- a/benchmarks/ar_diffusion/README.md
+++ b/benchmarks/ar_diffusion/README.md
@@ -41,20 +41,46 @@ CPR is not redundant with RTF. RTF is a ratio of aggregates: a session can hold
 bursts that individually miss their playout instants.
 
 Deadlines assume a declared playout grid. Playback starts once `--buffer-chunks`
-chunks are buffered; chunk `buffer_chunks - 1` anchors the grid and every later
-chunk is due one release period after its predecessor. Chunks inside the
-prebuffer have no deadline — their cost is start latency, which TTFC reports.
+chunks are buffered; from that instant the player consumes continuously, so
+
+```
+deadline(i) = t_ready(buffer_chunks - 1) + cumulative_frames(i) / target_fps
+```
+
+Chunks inside the prebuffer have no deadline — their cost is start latency,
+which TTFC reports. Walking cumulative frames rather than multiplying a constant
+period is what lets a deeper buffer grant real slack, and what keeps a causal
+decoder's shorter opening chunk from being credited with a full period.
 
 ## Model neutrality
 
 The chunk shape comes from the pipeline's declared `ARDiffusionKVCacheSpec`. No
 model name appears in this directory.
 
-One value is not derivable from the capability today: every frame count in the
-spec is in *latent* frames and nothing converts them to delivered frames, so the
-playout grid cannot be computed from the spec alone. Until the spec declares it,
-supply `--vae-temporal-factor` (default `1`, correct for a decoder that does not
+One conversion is not derivable from the capability today: every frame count in
+the spec is in *latent* frames and nothing converts them to delivered frames, so
+the playout grid cannot be computed from the spec alone. Supply
+`--vae-temporal-factor` (default `1`, correct for a decoder that does not
 compress time).
+
+**The conversion is not a multiplication.** A causal video decoder expands a
+session's first latent frame to one raw frame and every later one to the full
+factor, so `n` latent frames become `(n - 1) * factor + 1` raw frames the first
+time and `n * factor` every time after:
+
+| | latent frames | raw frames |
+|---|---|---|
+| chunk 0 | 3 | **9** |
+| chunk k > 0 | 3 | 12 |
+| K chunks | 3K | **12K − 3** |
+
+So `chunks x frames_per_chunk` over-counts every run. The profile carries
+`frames_per_first_chunk` alongside `frames_per_chunk`, and both `generated_fps`
+and `RTF` sum the per-chunk delivery. Pass `--non-causal-decoder` for a decoder
+that expands every latent frame identically.
+
+This is also why a single declared integer could not close the gap: the missing
+piece is a *mapping*, not a factor.
 
 ## Usage
 
--- a/benchmarks/ar_diffusion/__init__.py
+++ b/benchmarks/ar_diffusion/__init__.py
@@ -1,0 +1,2 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
--- a/benchmarks/ar_diffusion/decoding_session.py
+++ b/benchmarks/ar_diffusion/decoding_session.py
@@ -1,0 +1,102 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+"""Join a latent-producing realtime session to incremental decode.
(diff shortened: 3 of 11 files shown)


linzhenpl07:

linzhenpl07:
## H200 revalidation protocol: 16 FPS, multi-session, multi-resolution

This is the requested hardware validation for this draft. Please pin the exact tested tree (the current draft head is `d0ec229`; it contains the current #6444 and #6533 prerequisites) and return the raw JSON/event logs together with the final commit and diff.

### Scope and fixed workload

- Model: the same LingBot-World 2.0 causal-fast checkpoint for every run.
- DMD steps: 4; target playback: 16 FPS; latent frames/chunk: 3.
- Causal delivery geometry: opening chunk = **9 RGB frames**, steady chunk = **12 RGB frames**. The steady budget is therefore `12 / 16 = 0.750 s`.
- Core resolutions: `832x480`, `768x480`, `640x480`, `512x480`.
- For every resolution run `N = 1, 2, 3, 4` sessions. If N=4 passes, continue until the first failure.
- Generation TP: at least the intended deployment TP; preferably `1, 2, 4` (and 8 if the machine has 8 GPUs).
- H200 is Hopper/SM90, so the required baselines are BF16 eager and BF16 compile. Do **not** label a run NVFP4 unless actual native kernel dispatch is demonstrated; a checkpoint that merely loads via dequantization/fallback is not an H200 NVFP4 result.

### Freeze the environment first

```bash
mkdir -p runs/h200_16fps/meta
git rev-parse HEAD > runs/h200_16fps/meta/git_head.txt
git status --short > runs/h200_16fps/meta/git_status.txt
python -m pip freeze > runs/h200_16fps/meta/pip_freeze.txt
nvidia-smi > runs/h200_16fps/meta/nvidia_smi.txt
nvidia-smi topo -m > runs/h200_16fps/meta/topology.txt
```

Also record the absolute checkpoint path/hash, dtype, eager/compile, TP, visible GPUs, topology, `gpu_memory_fraction`, clocks/power limits, and neighboring GPU processes.

### Correctness gate

```bash
pytest -q tests/diffusion/ar_diffusion/test_streaming_decode.py
pytest -q tests/diffusion/ar_diffusion/test_realtime_benchmark.py
pytest -q tests/diffusion/ar_diffusion/test_session.py
pytest -q tests/diffusion/ar_diffusion/test_kv_cache.py
pytest -q tests/diffusion/ar_diffusion/test_paged_attention.py

python examples/offline_inference/diffusion/ar_diffusion_streaming_decode.py \
  --vae "$MODEL/vae" --device cuda --dtype bf16 \
  --height 480 --width 832 --sessions 2 --chunks 4 \
  --output-dir runs/h200_16fps/vae_smoke
```

The real-VAE smoke must produce `9/12/12/12`, bounded state, matching solo controls, and distinct videos for distinct sessions. Keep one `StreamingDecodeState` for the lifetime of each session. Recreating it per chunk incorrectly measures `9/9/9/...`. The solo oracle must use the same frame-wise streaming path in one call, not `AutoencoderKLWan._decode`, because whole-tensor BF16 `post_quant_conv` may select a different kernel and create a numerical artifact unrelated to session isolation.

### Measurement stages

Run each configuration in at least three fresh processes.

1. **Shared-VAE decoder:** for each resolution and N, measure eager and compile independently. Use one shared VAE module but one decode state per session. Report cycle time, service time/chunk, aggregate RGB FPS, per-session delivery RTF, state bytes/session, peak HBM, and finite output. A single shared decoder passes only when its measured N-session cycle is `<= 0.750 s`; do not estimate N-session capacity by multiplying the N=1 result.
2. **Generation:** for each `(resolution, TP, N)`, run both saturating and paced modes. Report per-session `delivery RTF`, not only compute RTF, plus CPR, aggregate latent-equivalent FPS, finite output, and peak HBM. Paced submission must advance by the frames actually delivered: chunk 1 is released after `9/16 s`, later chunks after `12/16 s`. A fixed 0.750 s delay for every chunk manufactures an opening deadline miss.
3. **Vertical path:** measure serial `gen -> decode`, then actual overlapped `DiT(N+1) || VAE(N)` with lookahead exactly 1. Report tick wall samples/P50/max, generation/decode/overlap time, overlap efficiency, `peak_outstanding_generations <= 1`, delivered RGB FPS/RTF, CPR, state bytes/session, and peak HBM.
4. **Backpressure:** after finding maximum realtime N=`K`, run `K+1`. Accept bounded queueing, admission rejection, or explicit FPS degradation. Do not drop a committed latent chunk: the following AR transition depends on it. RGB display-frame dropping is only valid when explicitly declared safe and when the bottleneck is downstream of model-state commitment.

Definitions:

```text
compute RTF  = this chunk's compute time / 0.750
delivery RTF = time between adjacent t_ready values for one session / 0.750
serial prediction = gen cycle + decode cycle
ideal overlap prediction = max(gen cycle, decode cycle)
```

Only `delivery RTF <= 1.0` is realtime; `<= 0.9` is the recommended headroom tier. CPR is macro-averaged per session so one starved session cannot be hidden by aggregate throughput.

### Important current limitation of this draft

The current #6534 tree contains `DecodingSession` and `OverlappedDecodingSession` plus CPU ordering/bounds tests, but `run_realtime_benchmark.py` still constructs raw latent sessions: it does **not** load a real VAE or wrap real LingBot sessions with either decoder wrapper, and its CLI does not yet expose the full multi-resolution LingBot sampling configuration.

Therefore, until an integration runner is added and archived with the result:

- generation and decoder may be reported as **separately measured**;
- `gen + dec` and `max(gen, dec)` must be labelled predictions;
- do not claim an end-to-end H200 overlap measurement merely because the wrappers import or the CPU tests pass.

Use this exact sentence when Stage 3 has not been wired:

> H200 generation and decoder were measured separately; end-to-end overlap was not measured.

Finally, the current LingBot horizon is 10 chunks. With six warm-up chunks this leaves only four steady samples/run, so publish raw samples, P50 and max. Do not present P95/P99 as stable SLOs; meaningful P99 requires extending the horizon and collecting at least 100 steady chunks per session.

Please return `meta/`, raw decoder results, generation results for both load modes, serial/overlap event logs if Stage 3 is wired, failure/OOM logs, and one summary table containing resolution, TP, N, mode, delivery RTF max, CPR, peak HBM, and pass/fail.

linzhenpl07:
## Stage 1.5: locate the bottleneck class before optimizing anything

Thanks for the 4xH200 run. The aggregate CSVs answer "did it hit 16 FPS" (no), but
they cannot answer "why", and the why now looks more interesting than the miss itself.

### What the returned data already says

Generation, N=1, TP=4, budget 0.750 s/chunk:

| resolution | eager delivery RTF | compile delivery RTF | shortfall |
|---|---:|---:|---:|
| 512x480 | 1.268 | 1.348 | 1.27x |
| 640x480 | 1.456 | 1.567 | 1.46x |
| 768x480 | 1.703 | 1.724 | 1.70x |
| 832x480 | 1.796 | 1.832 | 1.80x |

Decode is not the constraint: with `torch.compile` it reaches N=4 at every resolution
(worst delivery RTF 0.910 at 832x480). Generation is over budget at N=1.

### The part that needs a diagnosis, not a bigger multiplier

Estimating model FLOPs per chunk at 832x480 (4680 tokens x 4 DMD forwards, single
branch since `kv_branches` has one entry, 14B dense at 2 FLOPs/param/token, attention
ignored) gives 5.24e14 FLOPs. Against that:

| configuration | aggregate peak | ideal | measured | **MFU** |
|---|---:|---:|---:|---:|
| 1x RTX PRO 6000 Blackwell, TP=1 | 251 TF | 2.088 s | 4.051 s | **51.5%** |
| 4x H200, TP=4 | 3956 TF | 0.132 s | 1.374 s | **9.6%** |

15.8x the FLOPs, 2.95x the throughput. To make one session realtime at 832x480 the
MFU only has to go from 9.6% to **17.7%** — the missing 1.83x looks like it is already
in the machine and simply is not being used.

`torch.compile` shows the same split. On the single card it is a 1.119x gain at
768x480 (4.058 -> 3.626 s). On 4xH200 TP=4 it is a 1-2% **regression** (1.277 -> 1.293 s
at 768x480, 1.347 -> 1.374 s at 832x480). Whatever changed, it is not inherent to the
model or the shapes, because the same code path is healthy on one GPU.

Two variables changed together between those rows — TP=4 sharding and the H200
software stack — and the returned data cannot separate them.

Caveat: the MFU figures are derived from a parameter-count FLOP estimate, not measured.
Treat them as an order of magnitude. They are meant to motivate Test A, not to be cited.

---

## Test A — the discriminator (please run this first)

One run each at **TP=1** and **TP=2**, and re-confirm **TP=4**:

- resolution `832x480`, `N=1`, **eager**, bf16
- a handful of steady-state chunks is enough; **realtime is not the goal, MFU is**
- TP=1 will be roughly 4 s/chunk. That is expected and fine.
- memory is not a concern: 14B bf16 is ~28 GB and an H200 has 141 GB

Record per TP:

```text
tensor_parallel_size
steady-state chunk time (s)        # median over the measured chunks, excluding chunk 0
chunks measured
dtype, eager/compile, resolution
```

How it will be read:

| H200 TP=1 MFU | conclusion |
|---|---|
| ~50% | the loss is **TP sharding**; the H200 stack is fine |
| ~10% | the loss is the **H200 software stack**; unrelated to TP |
| in between | both contribute, in proportion |

This also finally pins `eff`. Today it can only be bounded, because both `eff` and the
H200 single-GPU latency are unknown and there is one equation. Assuming one H200 equals
one RTX PRO 6000 gives an upper bound of `eff = 0.753` at TP=4, falling to ~0.52 at
512x480. TP=1 plus TP=2 turns that into a measured curve, which also settles whether
TP=8 is worth trying (current extrapolation says 832x480 would still be RTF 1.08-1.17).

---

## Test B — profile one chunk (direction depends on Test A)

Do this **after** Test A, because the two outcomes call for looking at different things:
if TP=1 is also ~10%, the trace should be read for kernel selection and attention
backend; if TP=1 is ~50%, it should be read for all-reduce and graph breaks.

Configuration: `832x480`, `N=1`, `TP=4`, both eager and compile.

Use a profiler schedule that **skips the opening chunk and the compile cold start** —
the opening chunk is 9 frames rather than 12, and a cold start inside the window is
exactly what made `generated_fps` misleading in the returned CSVs:

```python
from torch.profiler import ProfilerActivity, profile, schedule

prof = profile(
    activities=[ProfilerActivity.CPU, ProfilerActivity.CUDA],
    schedule=schedule(wait=6, warmup=2, active=2, repeat=1),
    on_trace_ready=lambda p: p.export_chrome_trace(f"chunk_trace_rank{rank}.json"),
    record_shapes=True,
)
prof.start()
for _ in range(10):
    session.next_chunk()     # whatever drives one tick
    prof.step()              # once per chunk
prof.stop()
```

Please do not send the trace itself; it will be large. Six numbers are enough:

```text
1. gap ratio = 1 - (GPU busy / profiled wall)     <-- the single most important one
2. top 15 kernels by total device time, with call counts
3. total kernel launch count for the window
4. NCCL total device time and its share of the window
5. TORCH_LOGS=recompiles output for one N=1 run
6. whether the DiT is actually running under CUDA Graph
```

On (6): the deploy config carries `warmup_cudagraph: true`, but that flag has never been
confirmed to be in effect on the DiT side. If this workload is launch-bound and CUDA
Graph is not actually active, that alone could account for a large part of the gap.

A reduction script that computes 1-4 from a chrome trace (merging per-stream intervals
so overlapping streams are not double counted) can be provided if useful.

---

## Test C — raw data that Stage 1/2 asked for and did not come back

Independent of A and B, and cheap:

```text
meta/            git HEAD, git status, pip freeze, nvidia-smi
nvidia-smi topo -m                <-- important: rules out a non-NVLink topology as
                                      an explanation for eff = 0.753
generation rows for N > 1         <-- generation_aggregate.csv contains only N=1
per-chunk event logs (t_submit / t_ready / frames)
failure and OOM logs, if any
```

`nvidia-smi topo -m` matters more than it looks: if the four GPUs are not fully
NVLink-connected, part of the efficiency loss is topology rather than kernel efficiency,
and Test A would be read differently.

---

## One correction to the returned summary, and one caution

**Correction.** The statement that `torch.compile` makes generation ~1.6x slower does not
hold. It comes from `generated_fps`, which is total frames over total wall time and
therefore absorbs the compile cold start. `delivery_rtf`, measured from adjacent
`t_ready` deltas, is startup-insensitive and shows only a 1.3-7.7% regression. The
signature is that `1/fps_compile - 1/fps_eager` stays at 0.0876-0.1026 across a 1.63x
range of work — a constant additive cost, not a proportional slowdown. It works out to
roughly 10-11 s per run, which is the magnitude of one DiT compilation. The harness
appears not to prewarm.

**Caution.** In `generation_aggregate.csv` every `saturating` row is marked `pass=True`
while carrying delivery RTF 1.4-1.84 and an empty CPR, so the pass criterion for that
mode is evidently not checking RTF. Anyone skimming the table would read those rows as
"realtime achieved". Worth fixing or dropping the column before these numbers travel.

---

## What to record, and why each field matters

The previous round returned two aggregate CSVs. Everything below was missing, and
several conclusions could not be drawn as a result. The "why" column is there so that
if a field is impossible to capture, you can judge what an acceptable substitute is
rather than dropping it silently.

### Per run — identity and configuration

| field | why it matters | what breaks without it |
|---|---|---|
| `tensor_parallel_size` | every latency number has to be divided by it before it means anything | **this is what was missing last time.** With TP unknown, `eff` cannot be back-solved and no multi-GPU conclusion is interpretable |
| `num_replicas`, device list | distinguishes 1 replica x TP=4 from 2 x TP=2 at the same GPU count | two topologies with identical GPU counts produce very different per-session latency |
| dtype, and whether quantized | H200 is Hopper: no FP4. A "quantized" run that silently dequantizes is not an FP8 result | a mislabeled precision invalidates every speedup ratio derived from it |
| eager vs compile, and the actual `enforce_eager` value | the config flag and the effective mode have already disagreed once | compile/eager comparisons become meaningless |
| resolution, `N` sessions, target fps | the budget is `frames/fps`; 12/16 = 0.750 s | RTF cannot be computed at all |
| checkpoint path + revision/hash | ensures both sides measured the same weights | a different checkpoint silently explains any discrepancy |
| `gpu_memory_fraction` | sets the KV pool and therefore `resident_capacity` | session-loss counts cannot be attributed to capacity vs a bug |
| `git rev-parse HEAD` **and** `git status --short` | a dirty tree means the commit does not describe what ran | results cannot be reproduced or attributed to a PR |
| `pip freeze` (torch, vllm, vllm-omni, transformers) | kernel selection and compile behavior are version-dependent | a stack difference gets misread as a hardware or code difference |

### Per run — environment and hardware

| field | why it matters | what breaks without it |
|---|---|---|
| `nvidia-smi` (model, driver, clocks, power cap) | a power- or thermally-capped GPU produces low MFU for a reason unrelated to code | the entire MFU argument could be chasing a power cap |
| `nvidia-smi topo -m` | if the four GPUs are not fully NVLink-connected, part of the `eff` loss is topology | Test A would be read wrongly: topology loss looks identical to kernel inefficiency |
| other processes on those GPUs | contention inflates every latency silently | timings are not attributable |
| the startup log line naming the **resolved** attention backend | the env var is a request, not proof; it can fall back | a fallback backend is a plausible cause of low MFU and would be invisible |
| whether CUDA Graph is actually in effect on the DiT | `warmup_cudagraph: true` is set but has never been confirmed active | if launch-bound and graphs are off, that is likely the main finding, and it would be missed |

### Per run — timing discipline

| field | why it matters | what breaks without it |
|---|---|---|
| **per-chunk raw samples**, not only p50/max | a compile cold start inside the window is invisible in an aggregate | exactly how the "compile is 1.6x slower" claim arose |
| which chunk indices were counted | chunk 0 yields 9 frames, later chunks 12 (causal VAE geometry) | including chunk 0 understates throughput by ~25% for that chunk |
| whether prewarm ran, and for how many chunks | separates steady state from cold start | a fixed cost gets amortized into a per-chunk rate and reads as a slowdown |
| wall-clock start/end of the measured window | lets `sum(chunk times)` be checked against wall time | hidden stalls between chunks stay hidden |
| compute time and delivery time reported separately | they differ once N > 1 | mixing them produces a number that is neither |
| `t_submit` / `t_ready` / `frames` per chunk | this is the raw form everything else is derived from | no derived metric can be rechecked or recomputed |

### Pre-flight — check these **before** starting a long run

Cheap, and each one has already cost a re-run somewhere:

```text
[ ] TP, resolution, N, dtype, eager/compile echoed into the log at startup
[ ] resolved attention backend line captured
[ ] resident_capacity line captured (it explains session losses at N=3/4)
[ ] one throwaway chunk completes and is finite
[ ] chunk 0 produces 9 frames and chunk 1 produces 12 (confirms causal geometry
    and that decode state is per-session, not per-chunk)
[ ] nvidia-smi shows no other process on the target GPUs
```

### Post-run — run this before tearing the machine down

The point of this list is that every item is recoverable in seconds while the box is
up, and costs a full re-run once it is gone:

```text
[ ] meta/ written (git HEAD, git status, pip freeze, nvidia-smi, nvidia-smi topo -m)
[ ] per-chunk event log saved, not just the aggregate CSV
[ ] TP present as a column in every results file
[ ] both eager and compile captured for at least one identical configuration
[ ] TORCH_LOGS=recompiles output saved for one N=1 run
[ ] failure / OOM / session-loss logs saved, including the runs that failed
[ ] for Test A: TP=1, TP=2 and TP=4 all present at the same resolution
```

The last one deserves emphasis: **Test A is only meaningful as a set.** A TP=1 number
without TP=4 measured on the same box, same day, same stack does not discriminate
anything, because the comparison is against the local single-GPU baseline only by
assumption. Two of the three points, on the same machine, is what makes it a measurement.

---

## Summary of what is being asked

| test | cost | what it decides |
|---|---|---|
| **A** TP=1 and TP=2 at 832x480, N=1, eager | small | TP sharding vs H200 stack; measures `eff`; settles whether TP=8 is worth trying |
| **B** profile one chunk | medium | which kernels or which gaps hold the missing 1.83x |
| **C** meta, topology, N>1 rows, event logs | trivial | rules out topology; completes Stage 1/2 |

A is the one that matters most, and it is the cheapest of the three.

Issue #5588, opened 2026-07-30
[New Model]: Add microsoft/Mage-VL with end-to-end full-duplex multimodal streaming inference


Gaohan123:
### The model to consider.

- Model: https://huggingface.co/microsoft/Mage-VL
- Repository: https://github.com/microsoft/Mage/tree/8c94a0ac905167f40b05b09332b78752b7f9fbef/mage_vl
- Paper: https://arxiv.org/abs/2607.24904

Mage-VL is a 4B codec-native, proactive-streaming multimodal foundation model for image and video understanding. It combines Mage-ViT with a Qwen3-4B causal decoder and a lightweight cognition gate. The same checkpoint supports offline image/video understanding and event-gated causal streaming.

### The closest model vllm-omni already supports.

The closest model is AURA, which depends Qwen3-VL and other Qwen-based vision-language models supported by vLLM/vLLM-Omni. Mage-VL also relates to vLLM-Omni's streaming video and full-duplex serving work.

Mage-VL's language backbone is Qwen3-4B-Instruct-2507, but its visual path is different: it uses a custom codec-native Mage-ViT, variable-length visual tokens derived from I/P-frame structure, 3D RoPE, codec-window streaming memory, and a cognition gate for proactive event-triggered responses.

### What's your difficulty of supporting the model you want?

Basic offline image/video understanding is conceptually within upstream vLLM's multimodal model-support scope. This RFC intentionally focuses on the capability that makes Mage-VL especially relevant to vLLM-Omni: **end-to-end, full-duplex multimodal streaming inference**.

The main challenges are:

- **Codec-native visual preprocessing:** support traditional H.264/HEVC metadata (motion vectors and residual energy) and, where feasible, neural codec inputs such as DCVC-RT rate maps.
- **Variable-length visual tokenization:** preserve Mage-ViT's I-frame/P-frame token selection and 3D positional information instead of falling back to dense uniform frame sampling.
- **Stateful causal streaming:** incrementally ingest non-overlapping codec/video segments while retaining bounded recurrent streaming memory and recent codec context across the session.
- **Cognition-gate execution:** run the lightweight System-1 gate for every rolling window, remain silent for routine content, and invoke the full System-2 VLM only when a response-worthy event is detected.
- **Concurrent input and output:** continue accepting and encoding live video/image/text updates while a prior response is being generated or streamed, rather than forcing a strict `input -> response -> done` request boundary.
[...]

PR #6537, opened 2026-08-24
[model]Add experimental Mage-VL full-duplex adapter and offline integration

zyforsure:
## Purpose

Related to #5588.

Mage-VL requires codec-native image/video inference and a proactive streaming session contract. This PR adds an experimental Mage-VL integration around the shared full-duplex runtime and an offline entrypoint for Microsoft's reference implementation.

The checkpoint still uses custom Transformers remote code. Native vLLM model execution is intentionally not claimed here and remains follow-up work; the production-oriented Mage-VL WebSocket endpoint uses that verified Transformers path.

## Proposed changes

- Add `MageVLDuplexAdapter` with decoded-frame and opaque codec-window inputs.
- Isolate bounded causal visual history, pending queries, gate decisions, and event de-duplication per session.
- Support synchronous and asynchronous cognition-gate/generation callables.
- Reuse the shared duplex runtime for proactive responses, barge-in, and stale-output cancellation.
- Add one offline wrapper for image, frame-video, traditional codec-video, and event-gated streaming inference through the Microsoft Mage reference scripts.
- Validate the Mage checkout, media paths, arguments, and runtime dependencies with actionable errors.
- Add a Mage-VL recipe and CPU tests for codec commands, gate behavior, metadata preservation, de-duplication, and cross-session isolation.
- Rebase onto upstream `main` and add DCO sign-offs.

## Test plan

```bash
pytest tests/e2e/features/fullduplex/test_runtime.py \
  tests/e2e/features/fullduplex/test_core_contracts.py \
  tests/e2e/features/fullduplex/test_personaplex_adapter.py \
  tests/e2e/features/fullduplex/test_mage_vl_adapter.py \
  tests/e2e/features/fullduplex/test_mage_vl_serving.py -q

pytest tests/examples/offline_inference/test_mage_vl.py -q
```

### Reviewer follow-up validation (commit `8a2625f3f`)

- Deferred cognition-gate work is scheduled as an asyncio task, so the input
  reader remains available for `response.cancel`/barge-in while GPU inference runs.
- Added the shared `DuplexAdapter.on_close`/`flush` hooks and wired both runtimes
  to drain pending gate work and release per-session state on close.
- Temporary video materialization now occurs inside the cleanup-protected block;
  cancelled inference workers are drained before the shared GPU lock is released.
- Capabilities now advertise only `text` and encoded `video`, matching the
  Transformers backend contract; invalid `window_size > max_windows` is rejected.
- Removed the incomplete Qwen3-VL-based native registration.  The real Mage-VL
  vision configuration is not Qwen3-VL-compatible, so exposing that class through
  vLLM's global registry would make `vllm serve` fail or silently use the wrong
  vision tower. Native executor support remains explicitly out of scope.
- Replaced the copied Mage serving event loop with `MageVLDuplexRuntime`, a small
  subclass of the shared `DuplexRuntime`. Model-specific deferred-gate waiting and
  graceful gate draining now live in the Mage package while response lifecycle,
  cancellation, stale-output handling, and close semantics remain shared.
- Added serialized response transitions so the input reader and deferred gate
  watcher cannot race to create duplicate responses. Generic proactive adapters
  no longer receive Mage-specific flush/response-on-close behavior.
- Gated `--video-backend codec` at startup on the external `cv-preinfer` executable
  from `codec-video-prep>=0.2.5`, documented the dependency and frames fallback,
  and added CPU tests for both paths.

Targeted tests on the RTX 4090 host:

```text
Python 3.12.13, PyTorch 2.11.0+cu130, vLLM 0.27.0
NVIDIA GeForce RTX 4090, driver 595.84, 49,140 MiB

pytest tests/e2e/features/fullduplex/test_runtime.py \
       tests/e2e/features/fullduplex/test_core_contracts.py \
       tests/e2e/features/fullduplex/test_personaplex_adapter.py \
       tests/e2e/features/fullduplex/test_mage_vl_adapter.py \
       tests/e2e/features/fullduplex/test_mage_vl_serving.py -q
28 passed, 15 warnings in 0.52s
```

The suite includes a two-second synthetic gate and asserts that
`response.cancelled` is acknowledged in under 0.5 seconds, pinning that the
gate no longer blocks the sole input reader. The duplicate-session test now
consumes the WebSocket close frame and verifies close code 4429.

Repository checks completed locally:

- Pre-commit validation hooks: ruff, format, typos, markdownlint, mypy 3.10,
  SPDX, forbidden imports, CUDA API checks, and test marks passed.
- Python syntax compilation: passed for all changed Python files.
- `git diff --check`: passed.
- Commit includes the required DCO sign-off.

GitHub checks at head `8a2625f3f` all pass: build (Python 3.11), build
(Python 3.12), pre-commit, and DCO.

### Real RTX 4090 end-to-end inference

All rows below are real runs of the submitted `examples/offline_inference/mage_vl/end2end.py` wrapper against the cached `microsoft/Mage-VL` checkpoint. They are single-run functional timings, not benchmark numbers.

| Task | Input/configuration | Elapsed | Result |
| --- | --- | ---: | --- |
| Image | `dog.jpg`, 128 max tokens | 5.84 s | return code 0; generated a detailed Border Collie/rug description |
| Frame video | `soccer-broadcast.mp4`, 8 frames | 5.51 s | return code 0; identified the England/Argentina BBC Sport broadcast |
| Traditional codec video | same video, `--video-backend codec --codec-engine traditional` | 5.49 s | return code 0; described the BBC Sport commentator panel |
| Streaming gate | first four 4-second segments, 8 frames/segment, threshold 0.5 | 9.06 s | return code 0; four causal gate decisions, all `silence (p=0.00)` |
| Streaming response-path coverage | first 4-second segment, threshold 0.0, 128 max tokens | 10.22 s | return code 0; gate plus real text generation |

The threshold-0.0 run intentionally forces the response branch after computing the real gate score. It verifies generation and the end-to-end streaming path deterministically; it is not presented as a naturally triggered gate response. With the default 0.5 threshold, this sample's first four segments correctly record the model's actual `silence` decisions.

Minimal Mamba CUDA-kernel validation also passed after the CUDA extensions were linked consistently against CUDA 13:

```text
input/output: (1, 32, 256), bfloat16
device: NVIDIA GeForce RTX 4090
result: passed
selective_scan_cuda NEEDED: libcudart.so.13
```

### Prompts

- Image: `Describe this image in detail.`
- Video: `Describe this video.`
- Streaming: `Please describe the video content in detail based on the provided information.`

Image/video tasks also accept a custom prompt through `--question`.

### Public input media

The inputs are public files from Microsoft Mage commit `8c94a0ac905167f40b05b09332b78752b7f9fbef`; reviewers do not need access to the test server.

- Image: [`dog.jpg`](https://github.com/microsoft/Mage/blob/8c94a0ac905167f40b05b09332b78752b7f9fbef/mage_vl/assets/examples/dog.jpg), SHA-256 `164d8dfe707fb854e288ad2eea65c2db87e90af11f689c85502860eeaf3f4794`
- Video: [`soccer-broadcast.mp4`](https://github.com/microsoft/Mage/blob/8c94a0ac905167f40b05b09332b78752b7f9fbef/mage_vl/assets/examples/soccer-broadcast.mp4), SHA-256 `9d840e86558fe5a59bb743092d161228c63c687ab53588ee476a1cd46064125a`
- Video metadata: H.264, 960x540, 24 fps, 30.0 seconds, 3,914,911 bytes.

### Generated output

Image:

```text
The image depicts a dog sitting on a patterned rug. The dog appears to be a Border Collie, characterized by its thick, fluffy coat and distinctive facial markings. The dog's fur is primarily white with patches of black and brown, and it has a white blaze running down the center of its face.
```

Frame video:

```text
The video features a live broadcast of a soccer match between England and Argentina, with the final score displayed as England 1, Argentina 2. The broadcast is from BBC Sport.
```

Traditional codec video:

```text
The video captures a live sports broadcast, likely a football match, featuring a panel of four commentators. The commentators are positioned around a table, each holding a microphone with the BBC Sport logo.
```

Streaming, default threshold 0.5:

```text
[t=0.0-4.0s] gate=silence (p=0.00)
[t=4.0-8.0s] gate=silence (p=0.00)
[t=8.0-12.0s] gate=silence (p=0.00)
[t=12.0-16.0s] gate=silence (p=0.00)
```

Streaming response-path coverage, threshold 0.0:

```text
[t=0.0-4.0s] gate=response (p=0.00) -> The video features a man standing in a stadium, holding a yellow microphone with the BBC Sport logo on it. He is dressed in a black polo shirt and has a ring on his left hand. The background shows a large crowd of spectators in the stands, with the word 'ATLANTA' prominently displayed on a blue banner. The scoreboard in the bottom left corner of the screen shows the score as 'ENG 1 2 ARG FT', indicating that England has lost the match to Argentina with a final score of 1-2. The man appears to be providing commentary or analysis of the match, as he is speaking into the microphone
```

## Reproduction

Use the public Mage checkout and model ID; none of the private server paths are required:

```bash
git clone https://github.com/microsoft/Mage.git
git -C Mage checkout 76bec2bb3818863f470de7e867c2dc7f1d0bfd83

python examples/offline_inference/mage_vl/end2end.py \
  --mage-repo /path/to/Mage --task image \
  --model microsoft/Mage-VL \
  --image-path /path/to/Mage/mage_vl/assets/examples/dog.jpg \
  --question "Describe this image in detail."

python examples/offline_inference/mage_vl/end2end.py \
  --mage-repo /path/to/Mage --task video \
  --model microsoft/Mage-VL \
  --video-path /path/to/Mage/mage_vl/assets/examples/soccer-broadcast.mp4 \
  --video-backend frames --num-frames 8 \
  --question "Describe this video."

python examples/offline_inference/mage_vl/end2end.py \
  --mage-repo /path/to/Mage --task video \
  --model microsoft/Mage-VL \
  --video-path /path/to/Mage/mage_vl/assets/examples/soccer-broadcast.mp4 \
  --video-backend codec --codec-engine traditional --num-frames 8 \
  --question "Describe this video."

python examples/offline_inference/mage_vl/end2end.py \
  --mage-repo /path/to/Mage --task streaming \
  --model microsoft/Mage-VL \
  --video-path /path/to/Mage/mage_vl/assets/examples/soccer-broadcast.mp4 \
  --video-backend frames --num-frames 8 --cur-fps 1 \
  --segment-sec 4 --max-segments 4 --gate-threshold 0.5 \
  --max-new-tokens 64 --device cuda --attn-impl sdpa

# Deterministically exercise gate + response generation:
python examples/offline_inference/mage_vl/end2end.py \
  --mage-repo /path/to/Mage --task streaming \
  --model microsoft/Mage-VL \
  --video-path /path/to/Mage/mage_vl/assets/examples/soccer-broadcast.mp4 \
  --video-backend frames --num-frames 8 --cur-fps 1 \
  --segment-sec 4 --max-segments 1 --gate-threshold 0.0 \
  --max-new-tokens 128 --device cuda --attn-impl sdpa
```

## Runtime environment

- OS: Linux 6.8.0-136-generic, glibc 2.35.
- GPU: NVIDIA GeForce RTX 4090 (49,140 MiB); driver 595.84.
- Python: 3.11.15.
- PyTorch: 2.13.0+cu130; torchvision 0.28.0+cu130; CUDA 13.0; cuDNN 9.2.0.
- Transformers: 5.15.1.
- `mamba-ssm`: 2.3.2.post1; `causal-conv1d`: 1.6.2.post1, source-built for `sm_89` and linked to CUDA 13.
- Traditional codec helper: `cv-preinfer` from `codec-video-prep` 0.2.5.
- Mage runtime checkout: `76bec2bb3818863f470de7e867c2dc7f1d0bfd83`.
- Checkpoint: cached copy of the public `microsoft/Mage-VL` repository.

## Production WebSocket serving

This PR now includes a production-oriented Mage-VL WebSocket endpoint:

```text
WebSocket /v1/mage-vl/duplex
```

The serving layer provides bearer-token authentication, bounded concurrent sessions,
session-id validation, maximum message sizes, idle timeouts, graceful close,
disconnect cancellation/barge-in, per-session state cleanup, health/readiness endpoints,
and serialized access to the shared Transformers checkpoint. GPU inference runs in a
tracked worker and cancellation waits for the worker to drain before another request
uses the model, preventing cross-session CUDA races.

The endpoint accepts the shared duplex events (`input.append`, `input.commit`,
`response.create`, `response.cancel`, `playback.ack`, `close`) and emits
`session.created`, `response.created`, `response.delta`, `response.done`,
`response.cancelled`, and `error` events. The checked-in client is:

```bash
python examples/online_serving/mage_vl/duplex_client.py \
  --url ws://127.0.0.1:8090/v1/mage-vl/duplex \
  --auth-token "$MAGE_VL_API_KEY" \
  --video /path/to/segment.mp4 \
  --prompt "Describe this video segment in detail."
```

### RTX 4090 production endpoint result

The submitted endpoint was started with the real cached `microsoft/Mage-VL`
checkpoint on an NVIDIA RTX 4090 and exercised through the network client using
the public `soccer-broadcast.mp4` input:

```text
server: uvicorn on 127.0.0.1:8094
input: soccer-broadcast.mp4, H.264, 960x540, 24 fps, 30 seconds
SHA-256: 9d840e86558fe5a59bb743092d161228c63c687ab53588ee476a1cd46064125a
prompt: Describe this video segment in detail.
events: session.created -> response.created -> response.delta -> response.done
returncode: 0
elapsed: 6.353 s
after close: GET /ready => {"status":"ready","active_sessions":0}
```

Generated response:

```text
The video opens with a man in a black polo shirt holding a yellow BBC Sport microphone,
standing in a stadium filled with spectators. The scoreboard in the bottom left corner
shows the score as England 1, Argentina 2, with "FT" indicating the match has ended. The
man appears to be providing commentary or analysis of the game. The scene then transitions
to a wider shot revealing four men standing around a table with papers and a phone on it.
Each man is holding a yellow BBC Sport microphone, and they seem to be engaged in a
discussion about the match. The stadium in the background is packed with spectators, and
the scoreboard still
```

The server log showed successful WebSocket acceptance and no runtime error. The
session registry was empty after the client sent `close`, confirming connection
cleanup. This run used the shared-runtime refactor at `8a2625f3f`.


vllm_omni/model_executor/models/mage_vl/mage_vl.py
```diff
@@ -0,0 +1,60 @@
...
+                and hasattr(config, "text_config")
+            ):
+                return config
+            raise
+
+    def get_hf_processor(self, **kwargs: object) -> Any:
+        # Mage advertises a remote processor.  Its token expansion is identical
+        # to Qwen2/3-VL, so use the maintained vLLM processor implementation.
+        return self.ctx.get_hf_processor(
+            Qwen3VLProcessor,
```
linyueqian: [blocking] Native registration cannot load the real checkpoint: forcing Qwen3VLProcessor onto the remote MageVLProcessor checkpoint raises AttributeError (MageVLVideoProcessor has no attribute register_for_auto_class), and the mage_vl_vision config lacks all five of deepstack_visual_indexes/depth/in_channels/num_heads/num_position_embeddings. Base rejects this arch cleanly, so as shipped this is an error-quality regression. It also leaks beyond omni: arg_utils.py pushes every _OMNI_MODELS arch into vLLM's global ModelRegistry, so stock vllm serve is affected too. Reproduced on a 4x H200 host with the 11 G microsoft/Mage-VL checkpoint.

vllm_omni/experimental/fullduplex/mage_vl/adapter.py
```diff
@@ -0,0 +1,237 @@
...
+            text = str(data or "").strip()
+            if text:
+                state.pending_query = text
+            return
+
+        window = _coerce_window(modality, data)
+        state.windows.append(window)
+        if len(state.windows) > self._max_windows:
+            del state.windows[: len(state.windows) - self._max_windows]
+        await self._evaluate_gate(session)
```
linyueqian: [blocking] Awaiting the gate here blocks the sole input reader task (serving/server.py:51), so every subsequent client event queues behind the gate. Measured: a 2.0 s gate delays response.cancel by 2.0 s, which defeats barge-in. Run the gate in a task the reader does not await, or hand gating to the runtime loop.

vllm_omni/experimental/fullduplex/mage_vl/adapter.py
```diff
@@ -0,0 +1,237 @@
...
+
+        async for chunk in _iter_generate_result(self._generate(session, windows, query, gate)):
+            yield _as_output_chunk(chunk, self._output_modality)
+
+    async def on_barge_in(self, session: DuplexSession) -> None:
+        state = self._state_for(session)
+        state.pending_query = None
+        state.pending_gate = None
+
+    def close_session(self, session: DuplexSession) -> None:
```
linyueqian: [important] close_session exists here but is not part of the DuplexAdapter ABC, so the recipe's documented shared-DuplexRuntime path never calls it and per-session state accumulates (200 sessions leave 200 entries). PersonaPlex solved this with an on_close hook on the adapter plus a small DuplexRuntime subclass; promoting close_session into the ABC (or adopting the on_close pattern) fixes both this and the runtime fork below.

vllm_omni/experimental/fullduplex/mage_vl/serving/backend.py
```diff
@@ -0,0 +1,219 @@
...
+                # Retain the lock until it really exits so a replacement request
+                # cannot race the same shared model.
+                with contextlib.suppress(Exception):
+                    await worker
+                raise
+
+    def _gate_sync(self, windows: Sequence[MageVLCodecWindow]) -> float:
+        import torch
+
+        prepared = [self._prepare(window, _STREAMING_PROMPT) for window in windows]
```
linyueqian: [important] prepared is built outside any try/finally, so an exception between here and the cleanup leaks the temp MP4s it creates. Move the creation inside the try whose finally removes them.

vllm_omni/experimental/fullduplex/mage_vl/serving/backend.py
```diff
@@ -0,0 +1,219 @@
...
+        """Serialize GPU calls and drain worker threads before cancellation."""
+        async with self._inference_lock:
+            worker = asyncio.create_task(asyncio.to_thread(function, *args))
+            try:
+                return await asyncio.shield(worker)
+            except asyncio.CancelledError:
+                # Cancelling an asyncio wrapper cannot stop a CUDA worker thread.
+                # Retain the lock until it really exits so a replacement request
+                # cannot race the same shared model.
+                with contextlib.suppress(Exception):
```
linyueqian: [important] suppress(Exception) does not catch asyncio.CancelledError (a BaseException), so a second cancel can free _inference_lock while a worker is still live. Suppress CancelledError explicitly here or re-acquire the lock on the cancel path.

vllm_omni/experimental/fullduplex/mage_vl/serving/server.py
```diff
@@ -0,0 +1,374 @@
...
+                await self.cancel_response()
+            self.adapter.close_session(self.session)
+            self.session.state = DuplexState.CLOSED
+            self._closed = True
+
+    async def _start_response(self, emit: Emit) -> None:
+        await self.cancel_response()
+        self._response_task = asyncio.create_task(self._respond(emit))
+
+    async def cancel_response(self) -> None:
```
linyueqian: [suggestion] response.cancel with no in-flight task is a silent no-op, which drifts from the shared runtime's behavior of barging in unconditionally. Emitting an acknowledgment (or documenting the divergence) would keep clients portable between backends.

vllm_omni/experimental/fullduplex/mage_vl/adapter.py
```diff
@@ -0,0 +1,237 @@
...
+        self._generate = generate
+        self._window_size = window_size
+        self._max_windows = max_windows
+        self._output_modality = output_modality
+        # Adapter instances are shared by the serving runtime. Keep mutable
+        # cognition state per session so concurrent sessions cannot leak visual
+        # windows, queries, or event ids into one another.
+        self._session_state: dict[str, _MageVLSessionState] = {}
+
+    def capabilities(self) -> DuplexCapability:
```
linyueqian, line 89: [suggestion] capabilities advertises image and codec_window, but the backend only accepts {"video_base64": ...}; a client selecting by capability will send inputs that get rejected. Also window_size larger than the accumulated max windows silently disables proactive mode (adapter.py:138 never fires); a startup validation or a warning would surface both.

linyueqian (changes requested):
Reviewed at head d7f53561d against base 5de6f7c37, validated on a 4x H200 host with vLLM 0.27.0 and the real microsoft/Mage-VL checkpoint (11 G). Findings are inline; the verdict is request changes on three blockers: the registration and gate findings below, plus one with no single code anchor: two of the PR's own new tests fail as shipped (test_mage_vl_proactive_gate_speaks_once_for_event, test_mage_vl_rejects_duplicate_session_and_releases_disconnect). Both are marked core_model plus cpu, so the per-PR CI lane collects them and goes red; the PR body states the suites were not run. The two minicpmo_4_5 test_code2wav_batching failures in the same CPU job are inherited from base and are not this PR's problem.

Design direction, related to the close_session finding: PersonaPlex already solved this shape with an on_close adapter hook plus a small DuplexRuntime subclass, and its docstring says model policy belongs in the model package. This PR copy-pastes the whole runtime instead (serving/server.py:38-123 closely mirrors core/runtime.py:17-86). I would rather see the PersonaPlex pattern reused than a second runtime fork.

Validation notes: mamba_ssm does not build on that host (CUDA 13.0 toolkit vs torch cu129) and the upstream checkpoint's streammind_gate needs it for every task, so the real Mage forward was stubbed for the runtime checks; the registration and gate findings reproduce without it. The codec video backend also depends on an external cv-preinfer binary that ships nowhere I can find; worth documenting or gating.

zyforsure:
> Reviewed at head [d7f5356](https://github.com/vllm-project/vllm-omni/commit/d7f53561d8d263f9cf4ebbc858b157b501a086ac) against base [5de6f7c](https://github.com/vllm-project/vllm-omni/commit/5de6f7c37afdd189e010a46ba6b96fcc14f6388f), validated on a 4x H200 host with vLLM 0.27.0 and the real microsoft/Mage-VL checkpoint (11 G). Findings are inline; the verdict is request changes on three blockers: the registration and gate findings below, plus one with no single code anchor: two of the PR's own new tests fail as shipped (test_mage_vl_proactive_gate_speaks_once_for_event, test_mage_vl_rejects_duplicate_session_and_releases_disconnect). Both are marked core_model plus cpu, so the per-PR CI lane collects them and goes red; the PR body states the suites were not run. The two minicpmo_4_5 test_code2wav_batching failures in the same CPU job are inherited from base and are not this PR's problem.
> 
> Design direction, related to the close_session finding: PersonaPlex already solved this shape with an on_close adapter hook plus a small DuplexRuntime subclass, and its docstring says model policy belongs in the model package. This PR copy-pastes the whole runtime instead (serving/server.py:38-123 closely mirrors core/runtime.py:17-86). I would rather see the PersonaPlex pattern reused than a second runtime fork.
> 
> Validation notes: mamba_ssm does not build on that host (CUDA 13.0 toolkit vs torch cu129) and the upstream checkpoint's streammind_gate needs it for every task, so the real Mage forward was stubbed for the runtime checks; the registration and gate findings reproduce without it. The codec video backend also depends on an external cv-preinfer binary that ships nowhere I can find; worth documenting or gating.

I have changed my code according to your review, please proceed

vllm_omni/model_executor/models/mage_vl/mage_vl.py
zyforsure: Addressed in e1f548427: removed the incomplete Qwen3-VL-derived native registration and its global ModelRegistry exposure. The PR now explicitly uses the checkpoint-compatible Transformers remote-code backend and does not claim native vLLM execution.

vllm_omni/experimental/fullduplex/mage_vl/adapter.py
zyforsure: Addressed in bb33a3a8e and refined in 8a2625f3f: gate inference runs in adapter-owned tasks, while MageVLDuplexRuntime watches completion without blocking the input reader. The 2-second gate regression acknowledges response.cancel in under 0.5 seconds.

vllm_omni/experimental/fullduplex/mage_vl/adapter.py
zyforsure: Addressed in 8a2625f3f: removed close_session, retained the shared on_close hook, and replaced the copied serving loop with the small MageVLDuplexRuntime subclass. The recipe now uses that subclass and all disconnect/close paths release adapter session state.

vllm_omni/experimental/fullduplex/mage_vl/serving/backend.py
zyforsure: Addressed in e1f548427: prepared starts as an empty list and every _prepare call now occurs inside the try/finally cleanup region, so partial preparation is also removed on failure.

vllm_omni/experimental/fullduplex/mage_vl/serving/backend.py
zyforsure: Addressed in e1f548427: the cancellation drain explicitly suppresses both Exception and asyncio.CancelledError while the inference lock remains held, so repeated cancellation cannot release the GPU lock before the worker exits.

vllm_omni/experimental/fullduplex/mage_vl/serving/server.py
zyforsure: Addressed in bb33a3a8e and preserved by the shared runtime in 8a2625f3f: response.cancel always performs barge-in and emits response.cancelled, including when no response task is active.

vllm_omni/experimental/fullduplex/mage_vl/adapter.py
zyforsure: Addressed in e1f548427: capabilities now advertise only text/video, matching the production backend, and construction rejects window_size > max_windows. Tests pin both contracts.

linyueqian (commented):
Delta re-review at 8a2625f3. Every round-1 item is resolved, and the two structural ones got the strong version of the fix rather than the patch.

The native registration that could not load the real checkpoint is withdrawn entirely, mage_vl.py, its registry entries and the registration test are gone, which is the honest resolution given the processor mismatch. The forked serving runtime is gone too: server.py shrank by a hundred lines and the shared DuplexRuntime grew deferred-response support, a response lock, an explicit close with a graceful-close versus cancel distinction, and on_close/flush hooks on the adapter ABC, which is exactly the PersonaPlex direction round 1 asked for. The gate itself now runs in a background task scheduled off the input path with a wake queue for the serving loop, gate tasks are cancelled on barge-in and gathered on close, so the 2-second reader stall that defeated barge-in is structurally impossible rather than merely avoided.

The smaller items all landed: prepared moves inside the try so the temp MP4s cannot leak on an exception between creation and cleanup; CancelledError is suppressed explicitly where the CUDA worker is awaited; response.cancel now emits cancelled unconditionally through the shared runtime; capabilities advertise only text and video with a comment saying why; and video_backend codec now fails fast with an actionable error when cv-preinfer is missing, which also closes the undocumented-binary gap from my validation notes.

What remains is execution evidence: the two tests that failed as shipped in round 1 are still in the suite and per-PR CI has never run at this head, so I have added the ready label. Approval follows green lanes, with any red matched against the known inherited signatures.

linyueqian (approved):
Lanes at 8a2625f3: general, Intel and NPU green. The general lane collects the two tests that failed as shipped in round 1, so the lifecycle rework is now proven by execution, not just by reading. AMD fails only the CosyVoice3-TTS E2E step with exit 124, the same step-timeout that is currently failing identically on unrelated branches (I have seen it three times today across different PRs, all timeouts during compiled server bring-up, never an assertion); it has nothing to do with Mage-VL.

Approving. The round-2 comment covers the substance: the broken native registration was withdrawn rather than patched, the runtime fork was replaced by extending the shared DuplexRuntime, and every mechanical item landed. This is the shape the experimental fullduplex tree wants new adapters to take.

hsliuustc0106 (approved):

PR #6542, opened 2026-08-24
[perf] Add mask-free TRTLLM packed-padding path

### LINKED ISSUE #6358: [Bug][B200] MiniMax-H3 B200 default TRTLLM_ATTN rejects packed padding attn_mask
## Summary

On datacenter Blackwell (B200 / SM100), MiniMax-H3 auto-selects `TRTLLM_ATTN` as the diffusion attention default, then fails at the first packed DiT attention call with:

```
RuntimeError: Attention backend 'TRTLLM_ATTN' does not support attn_mask. Select a mask-capable backend.
```

This is a **backend-capability mismatch**, not an FP8 / quantization bug. It already fails on the BF16 baseline `Omni.generate()` before the quantized run starts.

Observed on:

```
tests/diffusion/models/minimax_h3/test_minimax_h3_quantization_quality.py::test_minimax_h3_quantization_quality[fp8_minimax_h3_fl2va]
```

Nightly job: **Diffusion · MiniMax H3 FP8 Quality Test with H100 · 2-GPU**. The test does not pin `--diffusion-attention-backend` / `DIFFUSION_ATTENTION_BACKEND`. The MiniMax-H3 perf job exports `FLASH_ATTN` and therefore does not hit this path.

Related: #5700 (H3 follow-up), #5779 (packed TRTLLM + H3 default).

## Motivation

`MiniMaxH3Pipeline` is marked `attention_mask_free=True` so that B200 + `head_dim=128` can default to `TRTLLM_ATTN` (see `vllm_omni/diffusion/model_metadata.py` and `vllm_omni/platforms/cuda/platform.py`). The comment says H3 alignment padding is a second packed document and that packed TRTLLM should consume `cu_seqlens` without a dense mask.

H3 packed attention only skips building `attn_mask` when the backend advertises:

- `supports_packed_mask_free()`, or
- `supports_prefix_kv_slicing`

(`vllm_omni/diffusion/models/minimax_h3/minimax_h3_transformer.py`, `_run_packed_attention`)

`FLASH_ATTN` on CUDA sets `supports_packed_mask_free()` to True, so H100 (FA default) never materializes the mask.

`TRTLLM_ATTN` does **not** override either capability (both stay `False` on `AttentionBackend`). It also leaves `supports_attention_mask()` as the default `False`. When `used < packed_total` (typical H3 alignment padding), H3 still builds a prefix mask:

```python
attn_mask = torch.arange(packed_total, device=q.device)[None] < used
```

Then `TRTLLM_ATTN` / the mask-capable gate rejects it. Local tree currently raises from `trtllm_attn.py` when a mask is present without a complete packed `extra` payload:

```text
TRTLLM_ATTN does not support attn_mask ... Select a mask-capable backend (e.g. CUDNN_ATTN / TORCH_SDPA).
```

## Repro

Unpinned MiniMax-H3 FL2VA generate on B200 (TP=2 is enough; SP is not required):

```python

omni = Omni(
    model="<MiniMax-H3 FL2VA dir>",
    enforce_eager=True,
    tensor_parallel_size=2,
    text_encoder_tp_size=2,
    vae_use_tiling=True,
)
omni.generate(...)  # fails in packed attention
```

Pytest equivalent:

```bash
pytest -sv tests/diffusion/models/minimax_h3/test_minimax_h3_quantization_quality.py \
  -k fp8_minimax_h3_fl2va -m "full_model and cuda" --run-level full_model
```

- **H100:** default `FLASH_ATTN` → packed mask-free path → pass.
- **B200 + no backend pin:** default `TRTLLM_ATTN` → H3 still emits `attn_mask` → fail.
- **B200 + `DIFFUSION_ATTENTION_BACKEND=FLASH_ATTN` (FA4) or `CUDNN_ATTN`:** expected to avoid this failure.

## Proposed Change

Pick one contract and make H3 + TRTLLM consistent (any one is enough):

1. **Preferred for the B200 default:** implement `TrtllmAttentionBackend.supports_packed_mask_free() -> True` (and keep consuming `cu_seqlens` / `max_seqlen` as #5779 intended), so H3 stops building the prefix mask.
2. If TRTLLM should accept this prefix mask, set `supports_attention_mask() -> True` and keep the packed trim path in `trtllm_attn.py`.
3. If TRTLLM must stay mask-incapable for non-packed layers, stop advertising MiniMax-H3 as `attention_mask_free` on Blackwell, or default H3 to `CUDNN_ATTN` / `FLASH_ATTN` until (1) lands.

Do not treat this as a quantization-test flake. Pinning `FLASH_ATTN` in the quality job is only a CI workaround.

## Validation

- [ ] Unpinned MiniMax-H3 FL2VA generate on B200 (BF16, then online FP8) no longer raises `does not support attn_mask`.
- [ ] Confirm server log still shows `Defaulting to diffusion attention backend TRTLLM_ATTN` if TRTLLM remains the Blackwell default.
- [ ] H100 `FLASH_ATTN` path still skips the mask (`supports_packed_mask_free`).
- [ ] `tests/diffusion/models/minimax_h3/test_minimax_h3_quantization_quality.py` and existing `tests/diffusion/attention/test_trtllm_attn.py` packed-mask cases.
- [ ] If TRTLLM stays mask-incapable, add a unit test that H3 does not pass `attn_mask` when `used < packed_total` under that backend.

bobboli:
### Problem

MiniMax-H3 packs each aligned attention input as `[real tokens, padding]`. #5779 made this layout correct for `TRTLLM_ATTN` by validating the prefix mask, trimming padded Q/K/V, and restoring the aligned output shape. However, the backend repeats that validation for every layer and denoise update. The mask reduction, prefix comparison, `nonzero` searches, and CUDA scalar reads serialize the stream immediately after Ulysses all-to-all.

The same mask also triggers the generic capability rejection reported in #6358 before the packed TRTLLM path can run on the #5543 branch.

### Root cause

The producer already knows the valid-token boundary on the host, but `AttentionMetadata` had no explicit way to publish that trusted boundary together with canonical device-side cumulative lengths. TRTLLM therefore had to reconstruct and validate it from CUDA tensors and `attn_mask` on every call.

### Fix

- Add `PackedPaddingMetadata` with host Q/KV lengths and canonical `[0, valid_length]` cumulative-length views.
- Let a packed producer omit `attn_mask` only when the selected backend advertises `supports_packed_mask_free()`.
- Make `TRTLLM_ATTN` validate the host contract and trim Q/K/V without reading CUDA scalars or materializing a mask.
- Use the host KV length for the SAGE block-size check on this path, avoiding another `.item()` synchronization.
- Advertise TRTLLM multi-document packed-varlen support for continuous batching, and omit empty padding documents from producer-generated `cu_seqlens`.
- Reject every nonempty `attn_mask` consistently while restoring the physical padded output shape for packed-padding inputs.

This does not add arbitrary-mask support to TRTLLM attention. The fast path is limited to a producer-owned, single-batch `[real, pad]` packing; the backend still checks host lengths, bounds, devices, dtypes, shapes, and maximum-length consistency.

Fixes #6358.

## Performance

Measured on 4x NVIDIA B300 SXM6 (267.7 GiB each), TP1/Ulysses4/Ring1, dense BF16 `TRTLLM_ATTN`, regional compile, and the MiniMax-H3 official starship workload at 1344x768, 243 frames, and 49 denoise updates. Each Nsight Systems capture followed one unprofiled warmup request.

The interval is from the end of the final Ulysses all-to-all kernel to the start of the main TRTLLM FMHA kernel:

| Path | Paired attention samples | Gap p50 | Gap p95 | Intermediate GPU activities | Intermediate D2H copies |
| Existing packed-mask validation | 9,797 | 499.683 us | 698.308 us | 244,925 | 97,970 |
| Packed-padding fast path | 9,680 | 0.352 us | 0.384 us | 0 | 0 |

existing packed-mask validation
packed-padding fast path

## Test Plan

**vLLM Version:** 0.26.0

**vLLM-Omni Commit:** c52493a5

Software used for the B300 runs: PyTorch 2.11.0+cu130, CUDA 13.0, and FlashInfer 0.6.16.post1.

```bash
CUDA_VISIBLE_DEVICES=0 FLASHINFER_DISABLE_VERSION_CHECK=1 \
python3 -m pytest -q tests/diffusion/attention/test_trtllm_attn.py

CUDA_VISIBLE_DEVICES='' FLASHINFER_DISABLE_VERSION_CHECK=1 \
python3 -m pytest -q \
  tests/diffusion/attention/test_trtllm_attn.py \
  tests/diffusion/models/minimax_h3/test_minimax_h3_contract.py \
  tests/diffusion/models/minimax_h3/test_minimax_h3_step_execution.py

pre-commit run --from-ref origin/main --to-ref HEAD
```

## Test Result

- B300 TRTLLM backend tests: `37 passed, 1 skipped`; the expected skip is the unsupported INT8 SAGE case on B300.
- CPU/mocked backend and MiniMax-H3 contract coverage: `159 passed, 10 skipped`; the skips are the intentionally hidden GPU kernel cases.
- Continuous-batch coverage includes an aligned non-final request and verifies strictly increasing packed boundaries.
- Real TRTLLM multi-document cross-attention matched per-document SDPA with relative L2 error `0.000041`.
- All diff-scoped pre-commit hooks passed, including Ruff, mypy 3.10, SPDX, forbidden-import, direct-`torch.cuda`, and pytest marker checks.
- 4x B300 E2E completed all 49 denoise updates and produced a valid 243-frame MP4. Its frame SHA256 exactly matches the previously validated packed-fastpath baseline.


yenuo26:
@bobboli Would you like to push this code into #5543 and have them merged together? Or should #5543 be merged first, and then this code be submitted as a bugfix afterwards?
If needed, I can give you push access to #5543.

bobboli:
I would keep #6542 separate and merge it first, then rebase #5543 on top. #5543’s generic mask-capability check otherwise rejects MiniMax-H3’s structural padding mask before TRTLLM is reached. This PR makes that path genuinely mask-free via producer-owned packed-padding metadata, so #5543 can validate the resulting contract without absorbing this implementation.

bobboli:
Self-review: checked API compatibility (retained `supports_packed_mask_free`), the single-request no-sync path, generic ragged/continuous-batch boundaries including an aligned non-final request, and non-mask rejection behavior. Focused CPU/GPU tests and diff-scoped pre-commit pass; the 4x B300 starship E2E produced 243 valid frames matching the prior baseline hash.

Gaohan123 (approved):
LGTM. Thanks

vllm_omni/diffusion/attention/backends/trtllm_attn.py
```diff
@@ -361,7 +467,13 @@ def forward_cuda(
         sage_kwargs: dict = {}
         # The SAGE kernel requires every KV sequence to contain at least one full
         # quantization block. Small auxiliary attention sites use the dense kernel.
-        use_sage = self.quant.enabled and bool(torch.all(seq_lens >= self.quant.k_block_size).item())
+        use_sage = False
+        if self.quant.enabled:
+            if known_min_kv_length is not None:
+                sage_lengths_supported = known_min_kv_length >= self.quant.k_block_size
+            else:
+                sage_lengths_supported = bool(torch.all(seq_lens >= self.quant.k_block_size).item())
```
hsliuustc0106, line 475: [P3] In the newly enabled multi-request step path, structural padding documents still count toward the SAGE length gate: any request whose alignment tail (1–63 rows at `_SEQ_ALIGN=64`) is shorter than `k_block_size` (default 16, so roughly a quarter of padded requests) fails `torch.all(...)` and silently drops the whole batch to dense attention. Single-request mode is immune because the tail is trimmed before this check (`known_min_kv_length=used`). Worth deriving the check from real-document lengths only (e.g. a producer-published minimum real KV length) now that TRTLLM accepts continuous batching.

vllm_omni/diffusion/attention/backends/trtllm_attn.py
```diff
@@ -266,12 +380,16 @@ def forward_cuda(
...
+            raise ValueError("packed_padding must be PackedPaddingMetadata")
+        if packed_padding is not None and not has_packed_metadata:
+            raise ValueError("PackedPaddingMetadata requires complete packed TRTLLM attention metadata")
+        if attn_mask is not None:
             raise ValueError(
-                "TRTLLM_ATTN does not support attn_mask (Wan sets one only under SP with "
-                "mask_sp_padding=True and a non-divisible seq len). Either set "
-                "parallel_config.mask_sp_padding=False, or select a mask-capable backend "
-                "(e.g. CUDNN_ATTN / TORCH_SDPA)."
+                "TRTLLM_ATTN does not support attn_mask. Represent structural suffix padding "
```
hsliuustc0106, line 390: [P3] The removed prefix-mask validation contract is no longer pinned by any test: `test_masked_layer_raises` covers `attn_mask` without packed metadata, and the two deleted tests covered the old accepted path. A one-line `pytest.raises` with complete packed `extra` plus an `attn_mask` would pin the new unconditional rejection — the contract #5543's generic mask-capability check will be rebased onto.

PR #4100, opened 2026-06-03
[Bugfix] Fix MOSS-TTS CI failures and make audio generation reproducible per sampling seed

zhangj1an:
PLEASE FILL IN THE PR DESCRIPTION HERE ENSURING ALL CHECKLIST ITEMS (AT THE BOTTOM) HAVE BEEN CONSIDERED.

## Latest Status [29 Aug 2026]
I'm back actively working on this PR. I will update PR description today and ping yueqian when im done

### Purpose 1: Fixes 3 CI bugs 

Closes #4087. 

| # | Symptom | Buildkite | 
|---|---------|-----------|
| 1 | `StageEngineCoreProc died during READY` (startup OOM) | [build 10912](https://buildkite.com/vllm/vllm-omni/builds/10912#019e8989-7b24-448a-adf1-13a553b4d71b) | 
| 2 | Engine-init OOM / `EOFError` from worker launch | [build 10952](https://buildkite.com/vllm/vllm-omni/builds/10952/canvas?sid=019e8c55-be43-4ee0-af30-8bde164554a4&tab=output) | 
| 3 | Online accuracy — empty transcript (`similarity=0.00`) | [build 10995](https://buildkite.com/vllm/vllm-omni/builds/10995/canvas?sid=019e8eb1-0845-428c-8bf8-6fac069739d1&tab=output) | 

### Purpose 2: For reproducibility, make sure the same request with the same seed always generates the same audio.

During audio generation, the model uses a random number generator (RNG) to pick each step's audio token. An RNG starts from a value called a seed, and the same seed makes it produce the same sequence of picks. So giving a request's RNG the same seed should yield the same audio every time. 

Previously, MOSS-TTS's talker did not pass the request's seed into its audio sampler. Instead, the sampling call fell back to PyTorch's global RNG, which the request's seed never sets. So the generated audio was not conditioned on the seed (the same request with the same seed could still come out different). We fixed it by giving each request its own RNG, seeded from the request's seed, so identical seeds reproduce the same audio stream. Added regression test for it. 

### Related Issues
Previously, purpose 2 contains duplicated fix for issues listed below, and is now removed. 

- [x] Cross-request audio corruption / the ability to handle **concurrent requests** proposed in https://github.com/vllm-project/vllm-omni/pull/3420#issuecomment-4600839626 by @Sy0307 was fixed in https://github.com/vllm-project/vllm-omni/pull/4415 by @yangyonggit.
[...]

PR #6543, opened 2026-08-24
[Bugfix] Honor request seed in MOSS-TTS adapters

### LINKED ISSUE #6480: [Bug]: MOSS-TTS online seed is inert — adapter reads the seed in build() before request.seed is applied
### Your current environment

Source-level ordering defect in `main` (verified at `37812146`).

### 🐛 Describe the bug

MOSS-TTS's online `seed` is inert: every request is synthesized with the deploy-config default (42) no matter what the client sends.

`Vevo2Adapter`-style adapters build `additional_information` in `build()`, and `MossTTSAdapter` reads the seed from the *sampling params* there:

```python
# vllm_omni/entrypoints/openai/tts_adapters/moss_tts.py:100-102
# MOSS reads the resolved seed at build time (it samples internally).
if sampling_params_list and getattr(sampling_params_list[0], "seed", None) is not None:
    tts_params["seed"] = [sampling_params_list[0].seed]
```

But `request.seed` is only copied into the sampling params **after** `build()` has already run:

| line in `serving_speech.py` | what happens |
|---|---|
| 2853 | `prepared = await adapter.build(request, sampling_params_list, ...)` |
| 2922 | `stage0_params.seed = request.seed` |

So at `build()` time `sampling_params_list[0].seed` still holds the deploy default. `vllm_omni/deploy/moss_tts.yaml:35` (and `moss_tts_nano.yaml:32`, `moss_tts_80gb.yaml:38`) set `seed: 42`, so the condition is always true and always writes `42`.

The later assignment cannot repair it: line 2920 `copy.deepcopy`s the sampling params, while the value the model actually reads was already captured into the prompt's `additional_information` at 2853.

**Net effect:** output is deterministic, but pinned to 42 — `"seed"` in the request body has no effect, and two requests with different seeds return byte-identical audio.

### Suggested fix

Read `request.seed` directly in `build()`, falling back to the sampling-params value as the deploy default:

```python
seed = request.seed
if seed is None and sampling_params_list:
    seed = getattr(sampling_params_list[0], "seed", None)
if seed is not None:
    tts_params["seed"] = [int(seed)]
```

### Context

Found while addressing review on #3422 — the Vevo2 adapter was about to copy this pattern, and the fix above is what shipped there. @linyueqian independently reached the same conclusion and asked me to file it. Not verified on hardware; the ordering above is from source, so a runtime confirmation (two requests, different seeds, compare bytes) would be worth doing before fixing.

JiataiWang:
Fixes #6480 for the reproduced MOSS-TTS-Nano path. Full-family model-side sampling is tracked separately in #4100 and #6034.

## What broke

At the adapter boundary, an explicit request `seed` was ignored whenever stage 0 had a default seed. `_MossTTSAdapterBase.build()` copied the stage default into MOSS model input before the shared speech-serving path applied `request.seed` to `SamplingParams`.

Both `MossTTSAdapter` and `MossTTSNanoAdapter` inherited that ordering defect. The end-to-end impact is narrower: Nano consumes `additional_information["seed"]`, so its online output was pinned to the deploy default (`42`). The full adapter now emits the correct contract too, but its current model consumers differ:

- Delay and Realtime still need model-side request-scoped generator plumbing (#4100 and #6034).
- Local already receives the request seed through `tts_local_seed` and its per-request generator path.

## Reproduction

With a stage default seed of `42`, build a MOSS request with `seed=0` or `seed=1234`. Before this fix, both adapter outputs still produced `tts_params["seed"] == [42]`. Issue #6480 also contains an end-to-end Nano reproduction with byte-identical audio before the fix and different outputs afterward.

The adapter regression is covered by:

```bash
pytest -s -v \
  tests/entrypoints/openai_api/test_tts_adapter.py::test_moss_tts_request_seed_overrides_stage_default \
  -m "core_model and cpu"
```

## Root cause

MOSS adapters build `additional_information` before the shared speech-serving path applies `request.seed` to `SamplingParams`. At build time, the adapter therefore saw only the deployment default and copied that value into the model input.

## Fix

Resolve the effective seed while building the MOSS model input: prefer `request.seed`, then fall back to the stage default when the request omits it. Tests cover both adapter contracts, the valid `seed=0` boundary, a nonzero request seed, and default fallback.

This does not change the pre-existing Nano concurrency limitation documented in the model: Nano currently seeds the global torch RNG, so distinct concurrent seeded requests are not isolated from one another.

## Test Plan

```bash
pytest -s -v tests/entrypoints/openai_api/test_tts_adapter.py -m "core_model and cpu"
pre-commit run --files \
  vllm_omni/entrypoints/openai/tts_adapters/moss_tts.py \
```

**vLLM Version:** `v0.26.1rc1.dev1131+ge8888b2d6` source (CPU-only local test environment)

**vLLM-Omni Commits:** `3555744e8db7e5ee458edfb5d3dd3529e96317f9`, `4fa68311`

## Test Result

- Before the source fix: 4 adapter regression cases failed with `[42]` instead of the request seed.
- After the fix and review clarification: `25 passed, 16 warnings in 1.49s`.
- All changed-file pre-commit hooks passed.

No model weights or GPU are required for this adapter-level regression.


JiataiWang:
Self-review:

- Traced the request through `_prepare_speech_generation` and confirmed MOSS consumes the seed from `additional_information` before the shared `SamplingParams` override runs.
- Kept the existing stage-default fallback; the only behavior change is that an explicit request seed now wins.
- Covered both full and Nano adapters, including `seed=0`, and ran the whole adapter test file (`25 passed`) plus changed-file pre-commit.

This is an adapter-level change; I did not run model-weight or GPU inference.

vllm_omni/entrypoints/openai/tts_adapters/moss_tts.py
```diff
@@ -97,9 +98,13 @@ async def build(
...
-        # MOSS reads the resolved seed at build time (it samples internally).
-        if sampling_params_list and getattr(sampling_params_list[0], "seed", None) is not None:
-            tts_params["seed"] = [sampling_params_list[0].seed]
+        # MOSS samples internally from additional_information. build() runs
+        # before the shared path applies request.seed to SamplingParams.
+        seed = request.seed
+        if seed is None and sampling_params_list:
+            seed = getattr(sampling_params_list[0], "seed", None)
+        if seed is not None:
+            tts_params["seed"] = [int(seed)]
```
linyueqian, line 107: [important] This lands for `moss_tts_nano` only. Of the four `stage_keys` on `MossTTSAdapter`, no full-family talker reads what is written here:

- `moss_tts` / `moss_tts_codec` resolve to `MossTTSDelayTalkerForGeneration`. `_sample_audio_codes` samples with an unseeded `torch.multinomial` at `modeling_moss_tts_talker.py:533` and `:588`, and the class never reads `additional_information["seed"]` nor the runner's `_omni_seed` kwarg. This is the production-recommended variant per `deploy/moss_tts.yaml:1`.
- `MossTTSRealtimeTalkerForGeneration` calls the `generate_frame` at `modeling_moss_tts_local.py:83`, which has no `generator` parameter at all, so its `_sample_token` call at `:133` draws from the global RNG.
- `moss_tts_local` / `moss_tts_local_codec` already honored `request.seed` before this PR, through `extra_args["tts_local_seed"]` (`serving_speech.py:2883`) into `gpu_model_runner.py:1877-1893` `_row_generator`, then `local_depth.py:202 generate_frame(generator=...)`.
- `moss_tts_nano` reads `_pick(info, "seed", None)` at `modeling_moss_tts_nano.py:318`. That is the one path this PR fixes.

No change requested in this hunk, the diff is a strict improvement. Flagging it because the PR description says "This affected both the full and Nano adapters", which is accurate at the adapter layer but not at the model layer, and that distinction decides whether #6480 can be closed.

tests/entrypoints/openai_api/test_tts_adapter.py
```diff
@@ -112,6 +116,44 @@ def test_moss_tts_applies_request_max_new_tokens(adapter_cls):
...
+    return asyncio.run(
+        adapter.build(
+            request,
+            [SamplingParams(seed=42)],
+            has_inline_ref_audio=False,
+        )
+    )
+
+
+@pytest.mark.parametrize("adapter_cls", [MossTTSAdapter, MossTTSNanoAdapter])
```
linyueqian, line 137: [nit] The `MossTTSAdapter` half of this parametrization asserts adapter output that no full-family consumer reads today (see the note on `moss_tts.py`). Worth keeping as an adapter contract test, but a one-line comment saying so would stop a later reader from taking it as evidence that the full family honors `seed` end to end.

linyueqian (commented):
Verified the ordering defect and the fix. `serving_speech.py:2811` calls `adapter.build()`, and `request.seed` only reaches the stage-0 `SamplingParams` at `serving_speech.py:2875-2883`, so `build()` could only ever observe the deploy default. The fix is correct and minimal, and the `seed=0` boundary is handled properly by the `is not None` check.

Reproduced your test result on a CPU-only env (vLLM 0.27.0) at `3555744e8`:

- head: `25 passed`
- with only the source hunk reverted to base: `4 failed, 21 passed` (`assert [42] == [0]`, `assert [42] == [1234]`)

So the new tests are genuine regression coverage, not just assertions of current behavior.

Also checked and clean: `int(seed)` is safe because `protocol/audio.py:147-153` bounds seed to `[0, 2**63-1]`, inside `torch.manual_seed`'s accepted range; `_conditioning_cache_salt` (`serving_speech.py:214-260`) does not fold `seed` in, so prefix-cache behavior is unchanged; neither `engine/serialization.py` nor the AR scheduler's `enable_update_additional_information` path touches the seed.

No blockers. Two scope notes inline, and two more here.

[nit] On nano, `seed` only delivers per-request reproducibility at `max_num_seqs=1`. `modeling_moss_tts_nano.py:318-330` sets the global torch RNG, and `deploy/moss_tts_nano.yaml:19` ships `max_num_seqs: 4`, so concurrent requests carrying distinct seeds now reset each other's stream mid-generation. The code already comments on the race, and it was not reachable before this PR because every request carried the same 42. Threading a per-request `torch.Generator` into `inference_stream` would be the real fix and belongs in its own change.

[nit] Worth resolving before #6480 is closed: the issue describes the symptom as "output is deterministic, but pinned to 42" with "two requests with different seeds return byte-identical audio". That holds for `moss_tts_nano`, but not for the Delay or Realtime families, where audio-code sampling was never seeded at all, so those paths were nondeterministic rather than pinned. Suggest closing #6480 as fixed for nano and tracking the remaining families separately.

vllm_omni/entrypoints/openai/tts_adapters/moss_tts.py
JiataiWang: Good catch — I overgeneralized the model-level effect from the shared adapter. I traced the consumers and agree: Nano is the end-to-end fix here; Delay/Realtime still need request-scoped generators, while Local already gets `tts_local_seed`. I tightened the PR description and linked the existing #4100/#6034 work instead of opening duplicate tracking. Kept this hunk as the shared adapter contract.

tests/entrypoints/openai_api/test_tts_adapter.py
JiataiWang: Added that comment in `4fa68311`: the full-adapter cases pin the adapter contract, while only Nano consumes the propagated seed end to end today. Re-ran the whole file (`25 passed`) and changed-file pre-commit.

linyueqian (approved):
Re-reviewed the delta `3555744e..4fa68311`. Both findings from the last round are resolved, the delta itself is a single comment line, and this is good to merge from my side.

The substantive one was the scope overclaim, and the fix is better than what I asked for. The description now opens with "Fixes #6480 for the reproduced MOSS-TTS-Nano path. Full-family model-side sampling is tracked separately in #4100 and #6034", and the "What broke" section separates the adapter-level defect from the model-level reality, naming Delay and Realtime as still needing request-scoped generator plumbing and Local as already receiving the seed through `tts_local_seed`. I checked the two tracking links rather than taking them on trust: #4100 is the MOSS-TTS reproducibility work and #6034 is "Use request-scoped RNG for MOSS-TTS-Realtime audio codes", which is precisely the Realtime gap I flagged. Linking those instead of opening duplicate trackers is the right call.

The test comment landed too, so a later reader will not mistake the full-adapter parametrization for evidence that the full family honours `seed` end to end.

Verified before approving: the PR merges cleanly against current main (`mergeable=true`, and `git merge-tree` agrees), and DCO, the 3.11 and 3.12 builds, docs and pre-commit are all green on `4fa68311`.

One thing for whoever merges: no Buildkite lane has reported on this head, since the PR carries no `ready` label. The change is confined to one adapter and its test, so the risk is low, but the label would give it a real CI pass first.

The two follow-ups I raised remain out of scope here and are correctly tracked elsewhere: per-request generator plumbing for the Delay and Realtime families, and the fact that on Nano a per-request seed is only meaningful at `max_num_seqs=1` while `deploy/moss_tts_nano.yaml` ships `max_num_seqs: 4`. That second one is worth a line on #6034 if it is not already captured, since concurrent requests carrying distinct seeds can still reset each other's stream.

Issue #4705, opened 2026-06-25
[New Model]: SenseNova-U1-8B-MoT-LoRAs


JaDonghao:
### The model to consider.

https://huggingface.co/sensenova/SenseNova-U1-8B-MoT-LoRAs

### The closest model vllm-omni already supports.

https://huggingface.co/sensenova/SenseNova-U1-8B-MoT

### What's your difficulty of supporting the model you want?

vLLM-Omni already supports the SenseNova-U1-8B-MoT model series. Does it also support SenseNova-U1-8B-MoT LoRAs? If yes, how can single-file SenseNova-U1 LoRAs (.safetensors) be converted into standard PEFT adapter directories compatible with vLLM-Omni? Could you provide corresponding technical support?

### Use case and motivation

The SenseNova-U1-8B-MoT model was trained on 2K resolution images. It produces severely degraded, practically unusable outputs at other resolutions. However, when deployed via vLLM-Omni, generating 2K images with this model takes over 50 seconds. Therefore, we urgently need to utilize the SenseNova-U1-8B-MoT-LoRA-8step-V1.0.safetensors file to boost generation performance.

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://vllm-omni.readthedocs.io), which can answer lots of frequently asked questions.

Roy214:
Hi, I would like to give it a try.

akshatvishu:
vLLM-Omni already supports `SenseNova-U1-8B-MoT` and PR #3971 adds inference-time LoRA loading when the adapter is in the documented PEFT format.

The released `SenseNova-U1-8B-MoT-LoRA-8step-V1.0.safetensors`  is not directly compatible with that path. It uses a single-file LoRA layout with `lora_down`, `lora_up` and `alpha` keys and also includes LoRA weights for `fm_modules.fm_head`.

To support it fully, I think we need:

1. A converter from the released single-file checkpoint to PEFT format.
2. Runtime LoRA support for `fm_modules.fm_head`.

This should be moderate scope: PR #3971 already covers the main LoRA path but the official checkpoint still needs format conversion, `fm_modules` LoRA handling, plus tests/docs.

PR #6544, opened 2026-08-24
[Bugfix] Fix diffusion LoRA silent no-op for aliased pipeline components

wisdom1710:
## Purpose

Fixes the silent failure mode where a diffusion LoRA loads and "activates" but applies to **zero layers**, producing base-identical output. Ref: #4705.

Two bugs in `DiffusionLoRAManager` caused this for `SenseNovaU1Pipeline` (and any pipeline with the same aliasing pattern):

**1. Component alias double-scan → registered layer names never match adapter keys.**

`SenseNovaU1Pipeline` exposes its denoiser twice: `self.transformer = self.language_model.model` (an alias matched by the default component list) and `_lora_components = ["language_model"]`. The manager scanned default components first, so every target layer was wrapped and registered under `transformer.layers.N...`, while adapter keys (after vLLM's `parse_fine_tuned_lora_name` strips `base_model.model.`) are `language_model.model.layers.N...`. `_get_lora_weights` then matched nothing, every layer fell back to `reset_lora(0)`, and all 168 layers silently ran as identity. Fix: after collecting candidate components, drop any component that is a nested submodule of another candidate, and only scan the outermost one.

**2. The static startup adapter was deactivated by the first ordinary request.**

`diffusion_worker` calls `set_active_adapter(req.sampling_params.lora_request, ...)` per request; requests without a LoRA pass `None`, which deactivated all adapters — including the static `--lora-path` one — immediately after warmup. Fix: when the request carries no LoRA request, fall back to the startup-configured static adapter instead of deactivating; explicit `scale≈0` requests still disable it for that request.

Also adds observability: per-component replacement counts at INFO, adapter activation counts, and a WARN when an adapter activates on zero LoRA layers (this exact silent failure mode).

## Test Plan

1. `pytest tests/diffusion/lora/ -q`
2. `ruff check` + `ruff format --check` on the changed file
3. Real-model verification on 2× RTX 4090D (TP=2): serve `SenseNova/SenseNova-U1.5-8B-MoT` with the official 8-step distilled LoRA (converted to a PEFT adapter dir) via `--lora-backend peft --lora-path`, then:
   - confirm startup logs show `replaced 168 layers` and `Activated adapter ... on 168/168 LoRA layers`;
   - same-seed 1024² 8-step `cfg_scale=1.0` text2img with vs. without the LoRA must differ (previously byte-identical);
   - verify 8-step output is sharp (DMD-distilled one-shot quality), not the undercooked baseline.

**vLLM Version:** 0.26.0
**vLLM-Omni Commit:** `71e9baae` (main, 2026-08-23)

## Test Result

- `tests/diffusion/lora/`: **31 passed**; 2 errors in `test_diffusion_lora.py::test_diffusion_model` and `test_images_generations_lora.py` — both are `advanced_model` e2e tests that download `Tongyi-MAI/Z-Image-Turbo` from Hugging Face, which is unreachable from the test machine (`Network is unreachable`). Unrelated to this change; needs CI.
- `ruff check` / `ruff format --check`: all passed.
- SenseNova-U1.5 8-step LoRA (2× 4090D, TP=2): 168/168 layers wrapped and activated; LoRA vs. baseline same-seed outputs now differ; 8-step output sharp. End-to-end ~6.8s for 1024² 8-step no-think vs ~63s for the 50-step think path.
- Note for reviewers: keeping a distilled 8-step LoRA active degrades high-step (50-step, cfg=4.0, think) output (over-sharpened/over-saturated). Per-request opt-out works: `"lora": {"name": "static", "path": ..., "scale": 0}`.

**BEFORE SUBMITTING:** read CONTRIBUTING.md and run the precheck-pr skill with the code agent for a self-check against project conventions.


wisdom1710:
Self-review checklist:
     - **Root cause verified**: Confirmed on real hardware (2× 4090D, SenseNova-U1.5-8B-MoT + PEFT LoRA) that the
   pipeline's `transformer` is a nested alias of `language_model.model`, causing the component scan to register layer
   names that never match adapter keys — LoRA reported "loaded" but bound to 0 layers. Also confirmed
   `set_active_adapter(None)` deactivated the static startup adapter.
     - **Fix scope**: Changes confined to `vllm_omni/diffusion/lora/manager.py`. Alias dedup keeps only outermost
   components during scan; `None` activation falls back to the static adapter. Added bind/activate count logs with a
   WARN on 0 hits so silent no-ops are visible.
     - **Tests**: `pytest tests/diffusion/lora/test_lora_manager.py test_loader.py test_base_linear.py` — 31 passed.
   `ruff check` and `ruff format --check` clean.
     - **Merge with main**: Resolved conflicts with the new `_record_bound`/binding_validator from #6476 by keeping
   both sides; refactored the bind loop into `_bind_adapter_weights() -> int` so the applied-layer count survives the
   refactor.
     - **End-to-end verification**: After the fix, logs show `replaced 168 layers` / `Activated adapter ... on
   168/168`; 8-step distilled-LoRA image generation (5.85s) produces byte-identical output to the pre-merge fixed
   build, confirming no behavior regression from the merge.
     - **Backward compatibility**: Verified high-step regression scenario is handled — requests can pass `"lora":
   {"scale": 0}` to disable the resident adapter when full-quality generation is needed.

wisdom1710:
Deployment write-up: SenseNova-U1.5 on 2× RTX 4090D — 5.85s per 1024² image, plus a silent LoRA failure we root-caused and fixed upstream #279https://github.com/OpenSenseNova/SenseNova-U1/discussions/279

amy-why-3459:
@SamitHuang PTAL

vllm_omni/diffusion/lora/manager.py
```diff
@@ -219,9 +226,14 @@ def set_active_adapter(self, lora_request: LoRARequest | None, lora_scale: float
         """Set the active LoRA adapter for the pipeline.
 
         Args:
-            lora_request: The LoRA request, or None to deactivate all adapters.
+            lora_request: The LoRA request, or None to use the static default
+                adapter (deactivating all adapters when no static default was
+                configured).
             lora_scale: The external scale for the LoRA adapter.
         """
+        if lora_request is None and self._static_lora_request is not None:
```
Gaohan123, line 248: **[P2] Preserve the startup scale when falling back to the static adapter**

When a request omits `lora_request`, this branch restores the startup adapter but keeps the request’s `lora_scale`, which defaults to `1.0`. The worker passes that value directly into this method, so an adapter initialized with `--lora-scale 0.25` is rebound at `1.0` on the first ordinary request. A startup scale of `0` also becomes `1.0`, unexpectedly enabling the adapter.

Please retain the startup scale alongside `_static_lora_request` and restore the intended default scale during fallback, while preserving explicit per-request overrides and the scale-zero opt-out. Add regression coverage for non-default and zero startup scales.

Gaohan123 (commented):
Please post some generation results in your PR description as comparison before and after the PR. Thanks

vllm_omni/diffusion/lora/manager.py
wisdom1710: Thanks for the careful review!
**P2 (startup scale lost in fallback)** — Fixed in the latest commit. The manager now records the startup scale alongside `_static_lora_request` (`_static_lora_scale`). During fallback, a request carrying the sampling-params default of 1.0 is treated as "unspecified" and the startup scale is restored; any other value — including an explicit per-request override and the scale-zero opt-out — is honored as-is. One caveat, documented in the docstring: an *explicit* per-request scale of 1.0 is indistinguishable from the default at the sampling-params layer, so it also resolves to the startup scale.
Regression coverage added in `tests/diffusion/lora/test_lora_manager.py`:
- `test_static_adapter_fallback_restores_startup_scale` — startup scale 0.25 survives plain requests; explicit 0.6 override wins; scale=0 still deactivates; next plain request re-activates at 0.25.
- `test_static_adapter_zero_startup_scale_stays_inactive` — `--lora-scale 0` is not re-enabled at 1.0 by a plain request; explicit non-zero override still opts in.
`pytest tests/diffusion/lora/test_lora_manager.py test_loader.py test_base_linear.py` — 34 passed; `ruff check` / `ruff format --check` clean.

**Generation comparison** 
## Before / after comparison
Same command, same seed (42), 8 steps, CFG 1.0, think off, 1024x1024 — only the code differs:
| | code | LoRA bound layers | result |
|---|---|---|---|
| **Before** | pre-PR `manager.py` | 0/168 (silent no-op) | undercooked, washed-out |
| **After** | this branch | 168/168 | sharp, fully rendered |

PROMPT="A serene Chinese ink-wash painting of misty Jiangnan water town at dawn, stone arch bridge over a canal, white-walled houses with dark tile roofs, soft morning light, a small boat with a fisherman, delicate brushwork, high detail"

Before (pre-PR code):
pr6544_before_oldcode
After (this branch):
pr6544_after_fixedcode

vllm_omni/diffusion/lora/manager.py
```diff
@@ -405,15 +431,43 @@ def _matches_target(module_name: str) -> bool:
...
+            if isinstance(component, nn.Module):
+                components.append((component_name, component))
+
+        # Pipelines may expose the same denoiser under several attributes
+        # (e.g. ``transformer`` aliasing ``language_model.model``). Scanning an
+        # inner alias first would register LoRA layers under names that adapter
+        # checkpoint keys (which follow the outer module path) can never match,
+        # so the adapter loads/activates but applies to zero layers. Keep only
+        # the outermost component for each module subtree.
+        outer_components: list[tuple[str, nn.Module]] = []
```
lishunyang12, line 447: Add a nested-alias test that verifies weights bind under the canonical component name.

Issue #4838, opened 2026-07-02
[Bug]: Voxtral TTS crashes with "expected scalar type BFloat16 but found Half" on T4 (compute capability 7.5)


Onkar17-73:
### Your current environment

<details>
<summary>The output of <code>python collect_env.py</code></summary>

```text
-2026-07-02 11:33:47--  https://raw.githubusercontent.com/vllm-project/vllm-omni/main/collect_env.py
Resolving raw.githubusercontent.com (raw.githubusercontent.com)... 185.199.109.133, 185.199.111.133, 185.199.108.133, ...
Connecting to raw.githubusercontent.com (raw.githubusercontent.com)|185.199.109.133|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 25236 (25K) [text/plain]
Saving to: ‘collect_env.py.1’

collect_env.py.1    100%[===================>]  24.64K  --.-KB/s    in 0.001s  

2026-07-02 11:33:47 (17.4 MB/s) - ‘collect_env.py.1’ saved [25236/25236]

Collecting environment information...
/kaggle/working/vllm-omni/vllm_omni/__init__.py:19: RuntimeWarning: Failed to import version from _version.py: No module named 'vllm_omni._version'
This typically happens in development mode before building.
Using fallback version 'dev'.
  from .version import __version__, __version_tuple__  # isort:skip # noqa: F401
INFO 07-02 11:34:04 [patch.py:252] NVFP4 W4A4 weight_scale NaN-clamp: installed.
==============================
        System Info
==============================
OS                           : Ubuntu 22.04.5 LTS (x86_64)
GCC version                  : (Ubuntu 11.4.0-1ubuntu1~22.04.3) 11.4.0
Clang version                : Could not collect
CMake version                : version 3.31.10
Libc version                 : glibc-2.35

==============================
       PyTorch Info
==============================
PyTorch version              : 2.11.0+cu130
Is debug build               : False
CUDA used to build PyTorch   : 13.0
ROCM used to build PyTorch   : N/A

==============================
      Python Environment
==============================
Python version               : 3.12.13 (main, Mar  4 2026, 09:23:07) [GCC 11.4.0] (64-bit runtime)
Python platform              : Linux-6.12.90+-x86_64-with-glibc2.35

==============================
       CUDA / GPU Info
==============================
Is CUDA available            : True
CUDA runtime version         : 12.8.93
CUDA_MODULE_LOADING set to   : 
[...]

PR #3253, opened 2026-04-29
[Bugfix][Qwen3TTS] Use float32 for code predictor on fp16-only GPUs

NickCao:
## Purpose

On GPUs without native bf16 (Turing SM 7.5, Volta SM 7.0), vLLM falls back to fp16 for bf16 models.  The code predictor small transformer then produces NaN hidden states because fp16 attention scores and SiLU-gated MLP intermediates overflow fp16 dynamic range, crashing torch.multinomial with a CUDA device-side assert.

Wrap the CodePredictorBaseModel forward pass in torch.amp.autocast with dtype=float32 when running in fp16.  Weights stay in fp16; only activations are cast per-op on the fly.  This extends the same float32 upcast pattern already used by RMSNorm and RoPE inside the code predictor to cover attention and MLP as well.

Also add dtype=torch.float32 to the three F.softmax calls in the sampling loop, matching the established convention in the vLLM main sampler (vllm/v1/sample/sampler.py, vllm/v1/sample/ops/ topk_topp_sampler.py).

Tested on g4dn.xlarge (T4, SM 7.5) with Qwen3-TTS-12Hz-0.6B-CustomVoice:
- Without fix: CUDA device-side assert on first request
- With fix: 2.32s WAV, 24kHz, valid audio

## Test Plan

On g4dn.xlarge (T4, SM 7.5):
```
  vllm serve Qwen/Qwen3-TTS-12Hz-0.6B-CustomVoic \
    --port 8091 --omni --enforce-eager --max-model-len 2048
```
```
curl http://localhost:8091/v1/audio/speech \
  -H "Content-Type: application/json" \
  -d '{"model": "Qwen/Qwen3-TTS-12Hz-0.6B-CustomVoice", "input": "Hello, this is a test.", "voice": "serena"}' \
  --output test.wav
```

## Test Result

Generates valid and legible wav.

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [x] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [x] The test plan. Please provide the test scripts & test commands. Please state the reasons if your codes don't require additional test scripts. For test file guidelines, please check the [test style doc](https://docs.vllm.ai/projects/vllm-omni/en/latest/contributing/ci/tests_style/)
- [x] The test results. Please paste the results comparison before and after, or the e2e results.
[...]

PR #6545, opened 2026-08-24
[Bugfix][Qwen3TTS] Talker hardcodes _embedding_dtype = torch.bfloat16, crashing every request on fp16-only GPUs (T4/sm75)

vedmalex:
## Summary

`Qwen3TTSTalkerForConditionalGeneration` reads the engine's configured dtype into a local `model_dtype`, uses it for the `_tts_pad_embed` buffer, and then hardcodes `self._embedding_dtype = torch.bfloat16` on the next line. On a GPU with no bf16 support, where vLLM forces `--dtype float16`, this makes the talker's per-request decode embeddings bfloat16 while the engine's `inputs_embeds` is float16, and the engine dies with `RuntimeError: index_copy_(): self and source expected to have the same dtype` on the first synthesis request.

## Where

`vllm_omni/model_executor/models/qwen3_tts/qwen3_tts_talker.py`, in `Qwen3TTSTalkerForConditionalGeneration.__init__` (line 446 at tag `v0.26.0`, still present on `main` as of 2026-08-24):

```python
model_dtype = getattr(vllm_config.model_config, "dtype", torch.bfloat16)
self.register_buffer(
    "_tts_pad_embed",
    torch.zeros(1, int(self.talker_config.hidden_size), dtype=model_dtype),
    persistent=False,
)
self._embedding_dtype = torch.bfloat16   # <-- ignores model_dtype computed above
```

`_embedding_dtype` is the dtype every per-request embedding is cast to in `preprocess_decode_batch` (the hook `OmniGPUModelRunner` calls), in the scalar `preprocess()` decode branch, and in `talker_mtp`. The mismatch surfaces in `vllm_omni/worker/gpu_model_runner.py` (`flush_decode_batch`), which allocates `inputs_embeds` from the *first* `req_embeds` it sees and then `index_copy_`s later ones into it:

```python
if inputs_embeds is None:
    inputs_embeds = torch.empty(
        (preprocess_input_ids.shape[0], req_embeds.shape[-1]),
        device=req_embeds.device,
        dtype=req_embeds.dtype,
    )
...
inputs_embeds.index_copy_(0, offsets_t, req_embeds)
```

## Reproduction

- GPU: NVIDIA Tesla T4 (compute capability 7.5, no bf16 support), Google Colab.
- `vllm-omni==0.26.0`.
- Models: `Qwen/Qwen3-TTS-12Hz-0.6B-Base` and `Qwen/Qwen3-TTS-12Hz-0.6B-CustomVoice`.
- `vllm serve <model> --trust-remote-code --omni --dtype float16` (`float16` forced because vLLM refuses bf16 below compute capability 8.0).
- Any `POST /v1/audio/speech` request.

```
File ".../vllm_omni/worker/gpu_model_runner.py", line 1723, in flush_decode_batch
    inputs_embeds.index_copy_(0, offsets_t, req_embeds)
RuntimeError: index_copy_(): self and source expected to have the same dtype,
              but got (self) Half and (source) BFloat16
```

Attention backend selected was `TRITON_ATTN` (not a FlashInfer/sm75 issue). This is an `EngineDeadError` — the server process dies for every subsequent request too. Observed on every request in two independent full runs, for `task_type` `CustomVoice` and `Base`, with and without `instructions`, with and without `ref_audio` — not specific to a task type, voice cloning, or the `instructions` field.

## Fix (this PR)

One line — use the `model_dtype` the same `__init__` already computes:

```diff
-        self._embedding_dtype = torch.bfloat16
+        self._embedding_dtype = model_dtype
```

The prefill path needs the same treatment: `Qwen3TTSPromptEmbedsBuilder.__init__` (`prompt_embeds_builder.py:332`) hardcodes `self._embedding_dtype = torch.bfloat16` too. Since the talker is its only constructor, this PR threads `model_dtype` through as a new keyword argument defaulting to `torch.bfloat16` (so any direct/test construction is unaffected). On bf16-capable hardware `model_dtype` already resolves to `bfloat16`, so both changes are no-ops there.

Two related hardcodes in the same class are deliberately left unchanged, since they're not what crashed here and may be intentional: `self.encoder.to(dtype=torch.bfloat16)` in `__init__` and again in `load_weights`. The encoder's `encode()` returns `torch.long` codec ids, so it does not feed the `_embedding_dtype`-typed path.

## Verification

Confirmed fixed on a real Colab Tesla T4 (2026-08-24), all three previously-crashing jobs now `PASSED` with real, independently re-verified waveform statistics (not silent garbage — checked peak/RMS/full-scale-fraction/distinct-sample-count against a human reference):

| Job | Model | RTF | peak | rms | full_scale_fraction |
| --- | --- | --: | --: | --: | --: |
| Base + `ref_audio` (voice clone) | `0.6B-Base` | 1.39 | 13406 | 1577.3 | 0.0 |
| CustomVoice, plain | `0.6B-CustomVoice` | 1.23 | 19517 | 2803.6 | 0.0 |
| CustomVoice + `instructions` | `0.6B-CustomVoice` | 1.12 | 24785 | 2964.9 | 0.0 |

Full context, including the two independent source-reads that located this and the reasoning for what was deliberately left unchanged: https://github.com/vedmalex/higgs-local-test/blob/main/docs/research/qwen3-tts-notes.md (section "Workaround B").

## Related

- #3253 (merged) — same model, same class of bug (a submodule a global fp16 override doesn't reach), different submodule (`qwen3_code_predictor.py`, not the talker's embedding dtype). Its fix is already present at `v0.26.0`; this crash happens with #3253 already installed, in an earlier code path.
- #4838 (open) — a different Omni TTS model (Voxtral TTS) failing on a T4 with the same error class ("Stage-1 has hardcoded `torch.bfloat16`..."). Not the same code path, but the same recurring pattern.

### DIFF
--- a/vllm_omni/model_executor/models/qwen3_tts/prompt_embeds_builder.py
+++ b/vllm_omni/model_executor/models/qwen3_tts/prompt_embeds_builder.py
@@ -300,6 +300,10 @@ class Qwen3TTSPromptEmbedsBuilder:
         ref_audio_artifact_cache_max_entries: Upper bound on the SHA1
             keyed LRU cache of per-ref-audio ``(ref_code, ref_spk_embedding)``
             artifacts. ``0`` disables the cache.
+        embedding_dtype: dtype the assembled prompt embeddings are produced
+            in. Must match the engine's configured model dtype, otherwise
+            the runner's ``inputs_embeds`` buffer and these embeddings
+            disagree (fatal on fp16-only GPUs forced to ``--dtype float16``).
     """
 
     def __init__(
@@ -317,6 +321,7 @@ class Qwen3TTSPromptEmbedsBuilder:
         encode_ref_audio_batch: Callable[..., list[torch.Tensor]],
         speaker_cache: Any | None = None,
         ref_audio_artifact_cache_max_entries: int = 256,
+        embedding_dtype: torch.dtype = torch.bfloat16,
     ):
         self._config = config
         self._talker_config = talker_config
@@ -329,7 +334,11 @@ class Qwen3TTSPromptEmbedsBuilder:
         self._tts_pad_embed_buffer = tts_pad_embed
         self._encode_ref_audio_batch_fn = encode_ref_audio_batch
         self._speaker_cache = speaker_cache
-        self._embedding_dtype = torch.bfloat16
+        # Prefill prompt embeddings must agree with the engine's configured
+        # dtype; the talker passes ``vllm_config.model_config.dtype`` in. The
+        # bfloat16 default only applies to callers that do not specify one
+        # (e.g. tests constructing the builder directly).
+        self._embedding_dtype = embedding_dtype
 
         self._text_tokenizer: Any | None = None
 
--- a/vllm_omni/model_executor/models/qwen3_tts/qwen3_tts_talker.py
+++ b/vllm_omni/model_executor/models/qwen3_tts/qwen3_tts_talker.py
@@ -443,7 +443,15 @@ class Qwen3TTSTalkerForConditionalGeneration(nn.Module):
             torch.zeros(1, int(self.talker_config.hidden_size), dtype=model_dtype),
             persistent=False,
         )
-        self._embedding_dtype = torch.bfloat16
+        # Talker embeddings are handed straight to the engine, which allocates
+        # ``inputs_embeds`` from the configured model dtype. Hardcoding
+        # bfloat16 here makes every embedding disagree with the engine on any
+        # GPU where bfloat16 is unavailable and ``--dtype float16`` is forced
+        # (Turing / sm75 and older), producing
+        # ``RuntimeError: index_copy_(): self and source expected to have the
+        # same dtype, but got (self) Half and (source) BFloat16`` in
+        # ``OmniGPUModelRunner.flush_decode_batch``. Follow the engine dtype.
+        self._embedding_dtype = model_dtype
 
         tokenizer_config = Qwen3TTSTokenizerV2Config.from_pretrained(
             self.model_path,
@@ -489,6 +497,7 @@ class Qwen3TTSTalkerForConditionalGeneration(nn.Module):
             tts_pad_embed=self._tts_pad_embed,
             encode_ref_audio_batch=self._encode_ref_audio_batch,
             speaker_cache=self._speaker_cache,
+            embedding_dtype=model_dtype,
         )
         self._load_custom_voice_profiles()
 


vedmalex:
## Self-review

What I checked before opening this PR:

- **Root cause, read from source at both `v0.26.0` and current `main`**: `Qwen3TTSTalkerForConditionalGeneration.__init__` computes `model_dtype = getattr(vllm_config.model_config, "dtype", torch.bfloat16)` correctly (used for the `_tts_pad_embed` buffer), then immediately overwrites `self._embedding_dtype = torch.bfloat16` on the next line, unconditionally. Confirmed by tracing every use of `_embedding_dtype` (`preprocess_decode_batch`, the scalar `preprocess()` decode branch, `talker_mtp`) that it is exactly what becomes `req_embeds` in `OmniGPUModelRunner.flush_decode_batch`, which then `index_copy_`s it into an `inputs_embeds` tensor allocated in the *engine's actual* dtype (`float16` on a T4).
- **Prefill path has the identical bug**: `Qwen3TTSPromptEmbedsBuilder.__init__` (`prompt_embeds_builder.py`) hardcoded the same `torch.bfloat16` literal, independent of the talker's `model_dtype`. Since the talker is `Qwen3TTSPromptEmbedsBuilder`'s only construction site, I threaded `model_dtype` through as a new `embedding_dtype` keyword argument defaulting to `torch.bfloat16` (so any direct/test construction elsewhere is unaffected) rather than changing the class's default.
- **What I deliberately left unchanged, and why**: `self.encoder.to(dtype=torch.bfloat16)` (`__init__` and `load_weights`) and the matching `.to(torch.bfloat16)` in `_encode_ref_audio_batch`. That path only produces `torch.long` codec ids consumed downstream — it doesn't feed `_embedding_dtype`-typed tensors, so it's out of scope for this specific crash. Flagging this explicitly in case a reviewer knows of a downstream fp16-only-GPU path I'm not seeing through the speech tokenizer/encoder.
- **On bf16-capable hardware this is a no-op**: `model_dtype` already resolves to `bfloat16` there, so both changed lines produce identical behavior to before.
- **GPU verification** (not just a read of the code): ran on a real Google Colab Tesla T4 (compute 7.5, no bf16 hardware support), `--dtype float16` forced. Before the fix: `RuntimeError: index_copy_(): ... Half ... BFloat16` on literally every request, for both `Qwen/Qwen3-TTS-12Hz-0.6B-Base` (task_type=Base, voice cloning) and `-0.6B-CustomVoice` (task_type=CustomVoice, plain and with `instructions`). After the fix: all three job types produced real synthesized speech — verified as real signal (not silence/clipping/a degenerate constant) via peak/RMS/full-scale-fraction/distinct-sample-count checks against a human-recorded reference, not just "no exception was raised." Numbers and methodology: https://github.com/vedmalex/higgs-local-test/blob/main/docs/research/qwen3-tts-notes.md
- Not tested: whether this also unblocks the `Base` task_type's reference-audio encoder path on other fp16-only hardware, or any interaction with quantized checkpoints — I don't have those to test against.

@FayeSpica, tagging you as the listed model owner in case there's context on why `_embedding_dtype` was hardcoded that I'm missing.

FayeSpica:
Thanks for the fix and the real GPU test.

**Root cause is correct.** The talker sent bfloat16 embeddings, but the engine buffers are float16. That mismatch is the crash.

**310P had the same fix.** The 310P patch already forces `_embedding_dtype` to float16. After your fix, that patch is no longer needed — could be cleaned up later. CC @zyz111222.

**Heads-up:** the docs build is failing on this PR. Need retry.

vllm_omni/model_executor/models/qwen3_tts/qwen3_tts_talker.py
```diff
@@ -443,7 +443,15 @@ def __init__(self, *, vllm_config: VllmConfig, prefix: str = ""):
...
-        self._embedding_dtype = torch.bfloat16
+        # Talker embeddings are handed straight to the engine, which allocates
+        # ``inputs_embeds`` from the configured model dtype. Hardcoding
+        # bfloat16 here makes every embedding disagree with the engine on any
+        # GPU where bfloat16 is unavailable and ``--dtype float16`` is forced
+        # (Turing / sm75 and older), producing
+        # ``RuntimeError: index_copy_(): self and source expected to have the
+        # same dtype, but got (self) Half and (source) BFloat16`` in
+        # ``OmniGPUModelRunner.flush_decode_batch``. Follow the engine dtype.
+        self._embedding_dtype = model_dtype
```
linyueqian, line 454: [important] No test pins the fixed invariant, and CI can never catch a re-hardcode: per-PR CI has no fp16-only GPU, and the existing unit tests bypass `__init__` entirely and pin `_embedding_dtype = torch.bfloat16` in their stubs (tests/model_executor/models/qwen3_tts/test_qwen3_tts_talker_preprocess.py:39, 88), which normalizes exactly the hardcode this PR removes. Smallest fix: `Qwen3TTSPromptEmbedsBuilder.__init__` is pure attribute assignment (no I/O, tokenizer is lazy), so a cheap CPU test can call the real `__init__` with SimpleNamespace stubs and assert `_embedding_dtype` follows the `embedding_dtype` argument; additionally parametrize the existing decode-path stub tests over float16 and assert the returned embeds match `_embedding_dtype`. That combination would have caught this bug and will catch a regression.

vllm_omni/model_executor/models/qwen3_tts/prompt_embeds_builder.py
```diff
@@ -317,6 +321,7 @@ def __init__(
         encode_ref_audio_batch: Callable[..., list[torch.Tensor]],
         speaker_cache: Any | None = None,
         ref_audio_artifact_cache_max_entries: int = 256,
+        embedding_dtype: torch.dtype = torch.bfloat16,
```
linyueqian, line 324: [suggestion] `embedding_dtype: torch.dtype = torch.bfloat16` keeps the hardcoded-bf16 footgun available to future callers, and the justifying comment at 339-340 ("tests constructing the builder directly") has no referent: the talker at qwen3_tts_talker.py:488 is the only in-repo `__init__` caller, and the tests build the class via `object.__new__` without calling `__init__`. Making the parameter required forces every future caller to choose the engine dtype explicitly; if the default stays, drop the tests claim from the comment.

linyueqian (commented):
Root cause is confirmed by static trace. `Qwen3TTSTalkerForConditionalGeneration.__init__` computes `model_dtype` from `vllm_config.model_config.dtype` (qwen3_tts_talker.py:440), uses it for the `_tts_pad_embed` buffer, then hardcoded `_embedding_dtype = torch.bfloat16` on the next line. `_embedding_dtype` types every decode-step embedding (`preprocess` at 699, `preprocess_decode_batch` at 849, `talker_mtp` at 1108). The consumer side is engine-dtype: `OmniGPUModelRunner` binds `inputs_embeds` to the persistent `self.inputs_embeds.gpu` buffer (gpu_model_runner.py:1623/1629) and `flush_decode_batch` does a strict `index_copy_` into it (1731), so under `--dtype float16` the first decode batch dies with Half vs BFloat16, exactly the reported T4 traceback.

The fix is minimal and covers the full embedding path: `_embedding_dtype = model_dtype` in the talker plus threading `embedding_dtype` into `Qwen3TTSPromptEmbedsBuilder` (its only in-repo constructor is the talker at qwen3_tts_talker.py:488). All builder consumers route through `self._embedding_dtype`, so no bf16 leaks remain on the fp16 embedding path. The only remaining bf16 hardcodes are the ref-audio codec encoder (talker 464, 974, 1051), whose output is `torch.long` codec ids, so leaving them is defensible. The 310P NPU patch already forces these exact two attributes to float16 in production, which is strong precedent that the fp16 embedding path is sound end to end; that subclass assigns after `super().__init__()` and is unaffected. On bf16 hardware `model_dtype` resolves to bfloat16, so mainline behavior is unchanged.

No blockers; two inline comments on test coverage and the constructor default. One CI note: the readthedocs check fails on this PR while passing on sibling open PRs, the diff touches no docs, so it looks transient and needs a re-trigger before merge.

Sowmya2400:
@linyueqian 

Is this PR still in progress? If so, could you please share an estimated timeline for when it might be merged?

vllm_omni/model_executor/models/qwen3_tts/qwen3_tts_talker.py
vedmalex: Done in 576340e, both halves of the suggestion plus one addition.

**Builder half (your suggestion).** `tests/model_executor/models/qwen3_tts/test_qwen3_tts_embedding_dtype.py` calls the real `Qwen3TTSPromptEmbedsBuilder.__init__` with `SimpleNamespace` stubs and asserts `_embedding_dtype` follows the `embedding_dtype` argument, parametrized over fp16 / bf16 / fp32, plus a test that omitting the argument raises (see the reply on the other thread for why it is now required).

**Decode half (your suggestion).** The stub helpers in `test_qwen3_tts_talker_preprocess.py` take `embedding_dtype` instead of hardcoding bfloat16, and the seven dtype-sensitive prefill/decode tests are parametrized over `[bfloat16, float16]`. Each now asserts the returned tensor's dtype against the configured one, not just its values: `out_embeds`, `past_hidden`, `text_step`, `mtp_inputs[1]`.

**Addition: the talker's own `__init__` is still unpinned by those two.** Neither test would catch someone putting `self._embedding_dtype = torch.bfloat16` back on qwen3_tts_talker.py:454 — the builder test exercises the builder, and the decode tests inject `_embedding_dtype` into a stub talker built with `__new__`. The real talker `__init__` is not cheap (it reads `Qwen3TTSTokenizerV2Config`, the encoder and an `AutoFeatureExtractor` from disk), so I pinned it statically instead: two AST tests assert that `__init__` derives `_embedding_dtype` from a name that itself reads `model_config`, and that the same name is passed as `embedding_dtype=` into the builder construction. Same source-level-contract approach as `tests/worker/test_capture_replay_contract.py`. If you would rather not have a source-shape assertion here, say so and I will drop those two — the other checks stand on their own.

**Teeth, verified by mutation** (each mutation applied alone, then reverted):

| mutation | result |
| --- | --- |
| `_embedding_dtype = torch.bfloat16` back in the talker | 2 failed |
| builder's `embedding_dtype` default restored | 1 failed |
| builder ignores the arg, hardcodes bfloat16 | 2 failed (fp16, fp32) |
| `embedding_dtype=` dropped from the builder call | 1 failed |
| `dtype = self._embedding_dtype` → `torch.bfloat16` in all three decode paths | 7 failed, all fp16 parametrizations, bf16 untouched |

Caveat on how that was run: this machine has no GPU and no vllm, so the suite ran against the real `vllm_omni` modules with `vllm` stubbed out at import. The assertions above are real; the environment is not CI's. CI is the run that counts.

vllm_omni/model_executor/models/qwen3_tts/prompt_embeds_builder.py
vedmalex: Taken as written: the parameter is now required, in 576340e.

You were right that the comment's justification had no referent — the tests build the class through `object.__new__`, so no direct-construction caller ever relied on that default. Rather than drop the claim and keep the default, I dropped the default: it is moved up among the required keyword-only arguments, and the comment now says why there is none.

```python
        encode_ref_audio_batch: Callable[..., list[torch.Tensor]],
        embedding_dtype: torch.dtype,
        speaker_cache: Any | None = None,
```

Checked before changing it that nothing else breaks: the only in-repo `__init__` caller is the talker at qwen3_tts_talker.py:488, and `_Qwen3TTSPromptEmbedsBuilder310P` subclasses the builder without overriding `__init__`, so it is constructed through that same call site. Every other reference in the repo is to the `estimate_prompt_len_from_additional_information` classmethod or to `object.__new__` in tests.

The missing-argument case is now a test rather than a promise (`test_builder_init_requires_embedding_dtype`); restoring the default makes it fail.

vedmalex:
Not stalled — the ball was on my side. Both review comments from @linyueqian are addressed in 576340e, and this push should also re-run the docs build that both @FayeSpica and @linyueqian read as transient (it fails here while passing on sibling PRs, and the diff still touches no docs).

**What changed since the review** (`git diff a75cd59..576340e --numstat`):

```
171   0  tests/model_executor/models/qwen3_tts/test_qwen3_tts_embedding_dtype.py
 76  38  tests/model_executor/models/qwen3_tts/test_qwen3_tts_talker_preprocess.py
  9   7  vllm_omni/model_executor/models/qwen3_tts/prompt_embeds_builder.py
```

- `embedding_dtype` is now a required keyword of `Qwen3TTSPromptEmbedsBuilder.__init__`, not a bfloat16 default — details and the caller audit in [that thread](https://github.com/vllm-project/vllm-omni/pull/6545#discussion_r4056098147).
- The fixed invariant is pinned by CPU-only tests, including the talker's own `__init__`, which neither suggested test would have covered — details and the mutation results in [that thread](https://github.com/vllm-project/vllm-omni/pull/6545#discussion_r4056097828).
- Behaviour on bf16 hardware is still unchanged; the shipped fix from a75cd59 is untouched.

@Sowmya2400 on timing: I can only speak for my side, which is ready now — the change is 2 source lines plus tests, `MERGEABLE`, and the T4 verification from the self-review above still holds. What it needs is an approving review; `linyueqian`'s pass said "no blockers" but was left as a comment, and the four requested reviewers (@tzhouam, @gcanlin, @Sy0307, @amy-why-3459) have not looked yet. Happy to rebase or split anything that makes that easier.

@FayeSpica on the 310P patch: I left it alone here. `_Qwen3TTSTalker310P` assigns `_embedding_dtype` after `super().__init__()`, so it still works and is now simply redundant on that platform; removing it is a separate change for whoever owns 310P (cc @zyz111222), and folding it into a bugfix PR would make this one harder to backport.

vedmalex:
Correction on the docs build, and a rebase.

I repeated in my previous comment that the readthedocs failure "looks transient." That was wrong, and I should not have passed it along without checking. It reproduced on the fresh build for 576340e ([build 34656343](https://app.readthedocs.org/projects/vllm-omni/builds/34656343/)), so I read the log instead of assuming.

**What actually fails.** The `pre_install` step, `python -m pip install --index-url https://download.pytorch.org/whl/cpu torch`:

```
Discarding .../typing_extensions-4.16.0-py3-none-any.whl: Requested typing-extensions>=4.10.0 ...
  has inconsistent Name: expected 'typing-extensions', but metadata has 'typing_extensions'
  Downloading typing_extensions-4.16.0.tar.gz (113 kB)
  Installing build dependencies: finished with status 'error'
      Looking in indexes: https://download.pytorch.org/whl/cpu
      ERROR: Could not find a version that satisfies the requirement flit_core<4,>=3.11 (from versions: none)
```

RTD's seeded pip rejects the `typing_extensions` wheel over the normalized-name check, falls back to the sdist, and the sdist's build dependency `flit_core` does not exist on the pytorch index — `--index-url` replaces PyPI rather than adding to it.

**Why only this PR.** It is not this PR's diff, and it is not luck: main already fixed it in 0637454, which adds `pip install --upgrade pip` and `--only-binary=:all:` to `.readthedocs.yml` — with a comment describing this exact failure. This branch was based on a4ea67a from August, 749 commits back, so it was still building docs with the old recipe. Sibling PRs pass because they sit on top of the fix.

So the remedy was a rebase, not a re-trigger. Force-pushed fa09467 (was 576340e), rebased onto current main, no conflicts. Re-verified on the rebased tree, not just assumed:

- the bug is still live on main — `self._embedding_dtype = torch.bfloat16` is still there at qwen3_tts_talker.py:471, so the fix is still needed after `[Rebase] Rebase to vLLM 0.29.0` (#7230) touched these files;
- the qwen3_tts + npu CPU tests give an identical set of results before and after the two commits (96 environment-related failures in my stubbed local harness, same list both ways — the real run is CI's);
- `ruff check`, `ruff format --check` and `tools/pre_commit/check_test_marks.py` are clean.

Nothing changed in the fix or the tests during the rebase; the diff against main is the same 4 files.

PR #6549, opened 2026-08-24
[Bugfix] Report engine-queued requests as waiting in vllm_omni:num_requests gauges

### LINKED ISSUE #6547: [Bug]: vllm_omni:num_requests_waiting stays 0 while requests queue inside the diffusion engine scheduler
### Your current environment

Not environment-specific. Logic bug in the pipeline-level Prometheus gauges, reproducible on any deployment serving a diffusion model from current main.

### 🐛 Describe the bug

`vllm_omni:num_requests_waiting` stays at ~0 and `vllm_omni:num_requests_running` overcounts when requests queue inside a diffusion stage engine.

**Repro**

1. `vllm serve <any diffusion model> --omni` (default scheduler, `max_num_seqs=1`)
2. Fire 10 concurrent generation requests
3. `curl localhost:8000/metrics | grep vllm_omni:num_requests`

Actual: `running=10, waiting=0`. Expected: `running=1, waiting=9` (the diffusion scheduler admits `max_num_running_reqs` at a time; the other 9 sit in its waiting deque).

**Why**

The frontend computes the gauges as `waiting = len(request_states) - running_counter` (see `_log_summary_and_cleanup` / `_process_single_result` in `vllm_omni/entrypoints/omni_base.py`). The counter is incremented by the orchestrator at dispatch time (`_register_running_request` in `vllm_omni/engine/orchestrator.py`), i.e. the moment a request is handed to a stage engine, not when the engine actually starts executing it. Requests queued inside the diffusion engine scheduler (`BaseScheduler._waiting` in `vllm_omni/diffusion/sched/base_scheduler.py`) are therefore counted as running. The engine-internal queue depth (`num_waiting_requests()`) never reaches the gauge formula - nothing in `vllm_omni/diffusion/` reports scheduler stats back to the client/orchestrator.

Diffusion stages are hit worst because their default concurrency is 1 and, unlike AR stages, they have no per-stage `vllm:num_requests_waiting{stage,replica}` fallback (diffusion never feeds the stat logger). AR stages have the same pipeline-gauge misattribution when their vLLM scheduler queues internally, but the per-stage `vllm:*` families at least expose the truth there.

**Impact**

Queue depth is invisible to autoscaling/alerting: a saturated diffusion server reports 0 waiting regardless of backlog.

Happy to submit a PR: plumb the diffusion scheduler queue depths back to the orchestrator over the existing ZMQ response channel, aggregate per (stage, replica), and fold the sum into the gauge formula so engine-queued requests count as waiting.

zetxqx:
## Purpose

Fixes #6547

`vllm_omni:num_requests_waiting` stayed at ~0 and `num_requests_running` overcounted whenever requests queued inside a stage engine. The gauges were computed as `waiting = len(request_states) - running_counter`, and the counter increments at orchestrator dispatch time, so requests sitting in the diffusion scheduler's waiting deque (default concurrency 1) were reported as running. A saturated diffusion server reported 0 waiting regardless of backlog.

This PR plumbs engine-internal queue depths back to the frontend gauge formula:

- Diffusion subprocess piggybacks `scheduler_stats` messages (num_waiting/num_running) on existing ZMQ message events: request arrival, each result send, abort, and error. No polling timer; messages are deduplicated on change.
- `StageDiffusionClient` stores the latest snapshot; the inline client exposes the same attribute as a property reading the scheduler directly, so both transports look identical to the orchestrator.
- The orchestrator aggregates per-(stage, replica) counts into a shared `_engines_waiting_counter` (`_note_engine_queue_waiting`), fed from the diffusion poll branch and from `SchedulerStats.num_waiting_reqs` on the AR branch, so AR pipelines are fixed too. Dead replicas reset their contribution to 0 on eviction.
- The frontend folds the sum into a single formula in `_publish_request_gauges`: `running = dispatched - queued_in_engines`, `waiting = (total - dispatched) + queued_in_engines`, clamped to `[0, dispatched]` to tolerate racy snapshots.

### DIFF
--- a/vllm_omni/engine/orchestrator.py
+++ b/vllm_omni/engine/orchestrator.py
@@ -387,6 +387,7 @@ class Orchestrator:
     # Class-level defaults so tests that bypass __init__ via object.__new__
     # don't AttributeError when transfer / counter emit paths access them.
     _running_counter: OmniRequestCounter | None = None
+    _engines_waiting_counter: OmniRequestCounter | None = None
     _transfer_emitter: Any = None
     _prom_metrics: Any = None
     _stat_logger: OmniPrometheusStatLogger | None = None
@@ -403,6 +404,7 @@ class Orchestrator:
         pd_config: dict[str, Any] | None = None,
         membership_controller: MembershipController | None = None,
         running_counter: OmniRequestCounter | None = None,
+        engines_waiting_counter: OmniRequestCounter | None = None,
         transfer_emitter: Any = None,
         prom_metrics: Any = None,
         log_stats: bool = False,
@@ -439,7 +441,13 @@ class Orchestrator:
             self._pd_bootstrap_addr = pd_config.get("bootstrap_addr")
             self._pd_prefill_engine_id = pd_config.get("prefill_engine_id")
         self.request_states: dict[str, OrchestratorRequestState] = {}
-        self._init_metrics_state(stage_pools, running_counter, transfer_emitter, log_stats=log_stats)
+        self._init_metrics_state(
+            stage_pools,
+            running_counter,
+            transfer_emitter,
+            engines_waiting_counter=engines_waiting_counter,
+            log_stats=log_stats,
+        )
 
         self._cfg_tracker = CfgCompanionTracker()
         self._stage_input_processors: dict[int, Any] = {}
@@ -485,6 +493,7 @@ class Orchestrator:
         stage_pools: list[StagePool],
         running_counter: OmniRequestCounter | None,
         transfer_emitter: Any,
+        engines_waiting_counter: OmniRequestCounter | None = None,
         log_stats: bool = False,
     ) -> None:
         """Wire up all metric-related orchestrator state.
@@ -510,6 +519,7 @@ class Orchestrator:
         from iteration_stats independently of scheduler gauges.
         """
         self._running_counter = running_counter
+        self._engines_waiting_counter = engines_waiting_counter
         self._transfer_emitter = transfer_emitter
 
         # Flat engine_idx ↔ (stage, replica) maps. The reverse map is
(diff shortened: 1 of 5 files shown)


zetxqx:
Author self-review (precheck-pr, full mode):

- Diff-scoped code-quality sweeps all clean: no new `**kwargs` string lookups, broad excepts, `Any` hints, `.clone()`/`deepcopy`, or event-loop blocking. The stats sends piggyback on existing async ZMQ message paths; the cross-thread scheduler reads are two int lengths (same access pattern the subprocess already used) and no lock is held across an await.
- Scope check: 8 files, all on the gauge data path (proc -> client -> orchestrator -> frontend formula + test). No unrelated refactors. Branch is based on current main.
- Regression test `test_engine_queued_requests_count_as_waiting` fails on main (old formula reports running=3/waiting=0 for the 3-dispatched/2-queued scenario) and passes here (running=1/waiting=2), asserting on real scraped Prometheus output.
- Behavior notes for reviewers: counts update at message events (arrival/result/abort/error), so between events the gauge is accurate to within +-1 request (exact at every event).
- Verified locally: compile + ruff on touched files. pytest runs in CI (macOS dev box here has no vllm install); repro steps are in #6547.

vllm_omni/engine/orchestrator.py
```diff
@@ -356,6 +333,7 @@ class Orchestrator:
     # Class-level defaults so tests that bypass __init__ via object.__new__
     # don't AttributeError when transfer / counter emit paths access them.
     _running_counter: OmniRequestCounter | None = None
+    _engines_waiting_counter: OmniRequestCounter | None = None
```
hsliuustc0106, line 390:
```suggestion
    _waiting_counter: OmniRequestCounter | None = None
```

zetxqx: Thanks for the review! I kept the `engines_` qualifier deliberately: "waiting" is overloaded here - the pipeline gauge computes `waiting = (total - dispatched) + engines_waiting`, where the first term is frontend-side waiting and this counter is only the engine-internal part. WDYT?

hsliuustc0106: Fair enough — since the pipeline gauge sums frontend-side waiting with this engine-internal counter, the `engines_` qualifier is doing real disambiguation work, not noise. Withdrawn.

hsliuustc0106 (approved):
LGTM — my naming suggestion was withdrawn after the rationale (the `engines_` qualifier disambiguates the two waiting quantities the pipeline gauge sums). Nothing outstanding from my side.

zetxqx:
Thanks @hsliuustc0106 ! I rebased the PR. https://github.com/vllm-project/vllm-omni/pull/6150 makes changes in the diffusion worker side. So there is less change in this PR. Can you take another look?

PR #6550, opened 2026-08-24
[Feature] Support MiniMax-H3 Turbo LoRA with DLO

mglyn:
## Purpose

Enable the MiniMax-H3 Turbo adapter from #6476 when distributed layerwise offload is active.

DLO continues to own and stream the base-block weights. Request-switchable LoRA A/B buffers remain resident on the compute device as sidecars, avoiding per-request DLO host-shard rebuilds. Buffer residency is preserved when the manager reallocates for a larger LoRA rank.

The default, non-DLO LoRA path is unchanged. The shared behavior change is limited to DLO + PEFT dynamic LoRA, a combination that previously left adapter buffers on CPU and could not execute against GPU activations. The tradeoff is fixed adapter HBM residency.

## Test Plan

**vLLM Version:** 0.27.0

**vLLM-Omni Commit:** d110dd669

- Run diffusion LoRA manager/layer tests and MiniMax-H3 LoRA, offload, and contract tests.
- Run changed-file pre-commit hooks.
- Validate H3 Base and Turbo with TP2 DLO no-AllGather.
- Validate DLO AllGather with TP1 x USP2, TP1 x DP2, and TP2 x DP2.
- For DP2, submit two concurrent identical requests so both DP replicas execute in the same wave.

## Test Result

- 174 targeted tests passed.
- All changed-file pre-commit hooks passed.
- TP2 DLO no-AllGather: Base and Turbo both produced valid 768x1344 H.264 video with AAC audio on 2x NVIDIA H200.
- TP1 x USP2 DLO AllGather: Turbo produced valid 1024x576 H.264 video with stereo AAC audio. Stage 0 was 20.24 s and peak usage was 20.80/20.89 GiB per GPU.
- TP1 x DP2 DLO AllGather: two concurrent Turbo requests completed in one DP wave at 28.16/28.06 s, with 19.17 GiB peak usage per GPU.
- TP2 x DP2 DLO AllGather: two concurrent Turbo requests completed in one DP wave at 24.70/24.68 s, with 17.33 GiB peak usage per GPU. Logs confirmed the strided DP groups `[0, 2]` and `[1, 3]`.
- Same-seed DP outputs had identical decoded audio. Decoded video was quality-equivalent but not bitwise identical (SSIM about 0.982). A no-LoRA Base control reproduced the same DP-rank behavior (SSIM 0.9806), while DP2 rank 0 matched DP1 bitwise, so this observation is not introduced by Turbo LoRA or adapter residency.

The existing orchestrator shutdown timeout warning remains observable after successful output generation; all workers exited and the GPU reservations were released.


hsliuustc0106:
allgather or non-allgather?

mglyn:
@hsliuustc0106 Both paths were validated.

- **No-AllGather:** TP2 + DLO, with rank-local TP-sharded LoRA sidecars.
- **AllGather:** TP1 x USP2; logs confirmed `checkpoint_mmap` AllGather over the SP group (`world_size=2`). The LoRA sidecars stay replicated per SP rank and do not participate in the DLO collective.

The Turbo adapter is rank 128 with 312 target modules:

| Item | Size |
|---|---:|
| Checkpoint tensor payload | 1.289 GiB |
| TP1 resident A/B buffers | 1.355 GiB/rank |
| TP2 resident A/B buffers | 0.911 GiB/rank |

TP1 is slightly larger than the checkpoint payload because the merged FFN projection keeps slice-local A buffers. In the TP1 x USP2 AllGather run, process GPU memory increased from about 5.24 GiB before manager initialization to at most 6.63 GiB after LoRA startup, consistent with the 1.355 GiB tensor calculation. The full 1024x576 Turbo request peaked at 20.80/20.89 GiB per NVIDIA H200, leaving about 3.1 GiB against a 24 GiB capacity target.

Keeping these buffers resident is intentional: DLO owns stable base-weight host shards, while the adapter is request-switchable. Rebuilding DLO shards on activation would couple request state to the host-weight plan; streaming A/B with every block would repeat H2D traffic at every layer and denoising step, and in AllGather mode could add another weight-transport lifecycle. Paying a fixed 0.91-1.36 GiB/rank once keeps request switching independent of DLO and leaves the collective restricted to base blocks.

hsliuustc0106:
> @hsliuustc0106 Both paths were validated.
> 
> * **No-AllGather:** TP2 + DLO, with rank-local TP-sharded LoRA sidecars.
> * **AllGather:** TP1 x USP2; logs confirmed `checkpoint_mmap` AllGather over the SP group (`world_size=2`). The LoRA sidecars stay replicated per SP rank and do not participate in the DLO collective.
> 
> The Turbo adapter is rank 128 with 312 target modules:
> 
> Item	Size
> Checkpoint tensor payload	1.289 GiB
> TP1 resident A/B buffers	1.355 GiB/rank
> TP2 resident A/B buffers	0.911 GiB/rank
> TP1 is slightly larger than the checkpoint payload because the merged FFN projection keeps slice-local A buffers. In the TP1 x USP2 AllGather run, process GPU memory increased from about 5.24 GiB before manager initialization to at most 6.63 GiB after LoRA startup, consistent with the 1.355 GiB tensor calculation. The full 1024x576 Turbo request peaked at 20.80/20.89 GiB per NVIDIA H200, leaving about 3.1 GiB against a 24 GiB capacity target.
> 
> Keeping these buffers resident is intentional: DLO owns stable base-weight host shards, while the adapter is request-switchable. Rebuilding DLO shards on activation would couple request state to the host-weight plan; streaming A/B with every block would repeat H2D traffic at every layer and denoising step, and in AllGather mode could add another weight-transport lifecycle. Paying a fixed 0.91-1.36 GiB/rank once keeps request switching independent of DLO and leaves the collective restricted to base blocks.

you need to validate DP > 1 cases

mglyn:
Validated DP > 1 with two concurrent requests in both TP1 x DP2 and TP2 x DP2 DLO AllGather topologies; both completed successfully. The same-seed comparison also exposed a pre-existing MiniMax-H3 video-VAE SDPA state leak, reproduced without active LoRA and tracked separately in #6566. The PR test results have been updated.

hsliuustc0106:
for the DP = 2 with allgather case, what's the host memory usage? did you take HWR as the runtime backend to reduce duplicated host memroy copies?

mglyn:
I measured the complete server process group via `/proc/*/smaps_rollup`; aggregate PSS is the primary host-memory metric because summed RSS double-counts shared pages. Both DP2 runs used two concurrent 1024x576, five-second Turbo requests in eager mode on NVIDIA H200:

- TP1 x DP2, DLO AllGather: startup peak PSS/RSS was 245.3/257.9 GiB; request-phase peak was 190.7/192.1 GiB.
- TP2 x DP2, DLO AllGather: startup peak PSS/RSS was 218.0/220.4 GiB; request-phase peak was 220.0/223.0 GiB.

These runs did not use HWR. #6486 has now merged support for BF16 HWR with **no-AllGather** DLO, but the current eligibility gate still rejects deployments with `lora_path`; AllGather also intentionally bypasses that BF16 HWR path. The open #6501 adds a separate, narrowly scoped HWR producer for online per-tensor FP8 + TP1 + AllGather, not this BF16 Turbo configuration.

Therefore HWR is not currently selectable for the exact Turbo + DP2 + AllGather combination tested here. Supporting dynamic PEFT sidecars with no-AllGather HWR would be a separate eligibility/validation follow-up.

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
```diff
@@ -586,8 +586,6 @@ def _load_diffusion_lora_adapter(
             offload_modes.append("model-level CPU offload (--enable-cpu-offload)")
         if getattr(od_config, "enable_layerwise_offload", False):
             offload_modes.append("layerwise offload (--enable-layerwise-offload)")
-        if getattr(od_config, "enable_distributed_layerwise_offload", False):
-            offload_modes.append("distributed layerwise offload (--enable-distributed-layerwise-offload)")
```
david6666666, line 590: why delete code here

hsliuustc0106:
> I measured the complete server process group via `/proc/*/smaps_rollup`; aggregate PSS is the primary host-memory metric because summed RSS double-counts shared pages. Both DP2 runs used two concurrent 1024x576, five-second Turbo requests in eager mode on NVIDIA H200:
> 
> * TP1 x DP2, DLO AllGather: startup peak PSS/RSS was 245.3/257.9 GiB; request-phase peak was 190.7/192.1 GiB.
> * TP2 x DP2, DLO AllGather: startup peak PSS/RSS was 218.0/220.4 GiB; request-phase peak was 220.0/223.0 GiB.
> 
> These runs did not use HWR. #6486 has now merged support for BF16 HWR with **no-AllGather** DLO, but the current eligibility gate still rejects deployments with `lora_path`; AllGather also intentionally bypasses that BF16 HWR path. The open #6501 adds a separate, narrowly scoped HWR producer for online per-tensor FP8 + TP1 + AllGather, not this BF16 Turbo configuration.
> 
> Therefore HWR is not currently selectable for the exact Turbo + DP2 + AllGather combination tested here. Supporting dynamic PEFT sidecars with no-AllGather HWR would be a separate eligibility/validation follow-up.

HWR is an experimental feature and will make more experiments and tests under no-allgather track first

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
mglyn: This was the explicit rejection for Turbo LoRA + DLO: the collected modes are passed as unsupported_offload_mode, and the H3 loader raises whenever it is non-None. This PR makes DLO supported by keeping the dynamic LoRA A/B sidecars resident on the compute device while DLO streams only the base weights, so DLO is removed from the unsupported list. CPU offload and regular layerwise offload remain rejected.

hsliuustc0106 (approved):

vllm_omni/diffusion/models/minimax_h3/pipeline_minimax_h3.py
hsliuustc0106: will it affect DLO without lora?

mglyn: 不会 这是之前的显示拒绝防护 既然支持了 DLO 可以安全去掉
DLO 未使用 LoRA 时 不会进入这个函数

PR #6555, opened 2026-08-24
[Test] Add MiniMax-H3 DLO DP2 T2VA smoke

david6666666:
## Summary
- Add the L2 MiniMax-H3 T2VA online-serving smoke test for two-card H100 DLO DP2.
- Exercise two concurrent /v1/videos/sync requests with num_inference_steps=4 and a 500 ms admission window.
- Register the test in .buildkite/cuda/test-ready.yml.

## Validation
- pre-commit: passed.
- pytest collection: 1 test collected.
- B300: 1 passed in 184.97 s using two cards and a local FL2VA partition; both concurrent requests returned valid 1344x768 MP4 video with audio.


tests/e2e/online_serving/test_minimax_h3_dlo_dp2_t2va.py
```diff
@@ -0,0 +1,116 @@
...
+    indirect=True,
+)
+def test_minimax_h3_dlo_dp2_t2va(omni_server: OmniServer, openai_client: OpenAIClientHandler) -> None:
+    """Validate one complete DLO all-gather DP2 wave with two T2VA jobs."""
+    with concurrent.futures.ThreadPoolExecutor(max_workers=2) as executor:
+        futures = [executor.submit(_run_t2va_request, openai_client, seed) for seed in (1101, 1102)]
+        videos = [future.result() for future in futures]
+
+    for video in videos:
+        assert_video_valid(video, width=WIDTH, height=HEIGHT, fps=FPS)
```
congw729, line 127: This T2VA test only validates the video stream. `assert_video_valid` uses OpenCV and does not check whether the MP4 contains an audio stream, so an audio-generation or muxing regression would still pass despite this test exercising `task=t2va`. Could we also assert that the returned MP4 has a non-empty audio stream?

congw729 (commented):
CI review: one coverage gap in the T2VA smoke assertion.

PR #6556, opened 2026-08-24
[Test] Add MiniMax-H3 DLO DP2 variants and Turbo LoRA L3 test

david6666666:
## Summary
- Add L3 two-card H100 DLO + DP2 online-serving cases for official FL2VA and Ref2VA request shapes.
- Add a two-card TP2 DLO + Turbo LoRA FL2VA case using merged support from PR #6550 and the Turbo artifact from PR #6476.
- Register the cases in `.buildkite/cuda/test-merge.yml`.
- Keep MiniMax-H3 request-specific conditioning local to each pure-DP replica while retaining DLO weight AllGather across the DP group.

## CI hang diagnosis and fix
Buildkite #14279 admitted both Ref2VA requests into one DP2 wave, but neither request reached denoising. Rank 1 timed out in c10d after 600 seconds and the executor reported `DLO DP collective wave timed out`.

The root cause was the MiniMax-H3 `_dit_rank_world()` helper using the complete worker WORLD for request-specific text/reference/VAE preparation. In pure DP, each rank owns an independent request; coupling those preprocessing paths through WORLD made rank 1 wait on the different Ref2VA request owned by rank 0 and could deadlock before the intended DLO weight AllGather. FL2VA happened to keep the two ranks aligned, while video-reference Ref2VA exposed the mismatch on H100.

Pure-DP request preparation now uses a logical single-rank replica group. Both ranks independently prepare and encode their own reference video/audio, and only the DLO offloader synchronizes across the DP group. The embedded-audio metadata broadcast now uses the same logical model group instead of bypassing the helper with WORLD.

The three server lifecycles remain isolated in fresh pytest processes so multiprocessing resource-tracker state cannot leak between cases.

## Validation
- B300, 2 GPUs: exact Ref2VA DLO DP2 concurrent wave passed after the fix, including the CI failure arrival order (seed 2102 before 2101).
  - Both ranks logged independent 2057-token Qwen presentations with one reference video.
  - Both requests completed in about 75-77 seconds and returned valid 1344x768, 24 FPS MP4 files with decodable audio.
- B300, 2 GPUs: DLO DP2 FL2VA concurrent wave passed.
- B300, 2 GPUs: DLO + Turbo LoRA TP2 no-AllGather passed; both TP ranks loaded the adapter and four denoiser evaluations completed.
- 147 MiniMax-H3 contract/parallel unit tests passed.
- Changed-file pre-commit hooks passed.
- `--diffusion-attention-backend` and `--enforce-eager` are not specified.

## Sampling note
- FL2VA and Ref2VA use `num_inference_steps=4`.
- The published Turbo artifact follows PR #6476: five requested sigma points produce four denoiser evaluations, so the LoRA request intentionally uses `num_inference_steps=5`.


tests/e2e/online_serving/test_minimax_h3_l3_variants.py
```diff
@@ -0,0 +1,267 @@
...
+def test_minimax_h3_dlo_dp2_fl2va(omni_server: OmniServer, openai_client: OpenAIClientHandler) -> None:
+    """Official first-frame FL2VA request over one complete DLO DP2 wave."""
+    for video in _run_dlo_wave(openai_client, _run_fl2va):
+        assert_video_valid(video, width=WIDTH, height=HEIGHT, fps=FPS)
+
+
+@pytest.mark.core_model
+@pytest.mark.advanced_model
+@pytest.mark.diffusion
+@pytest.mark.slow
```
congw729: These FL2VA/Ref2VA cases request video with audio, but `assert_video_valid` only inspects the video stream via OpenCV. A missing or broken audio stream would therefore go unnoticed. Could we add a lightweight assertion that each returned MP4 contains a non-empty audio stream? The same applies to the Turbo LoRA assertion below.

congw729 (commented):
CI review: one shared coverage gap in the MiniMax-H3 audio-video cases.

tests/e2e/online_serving/test_minimax_h3_l3_variants.py
david6666666: Addressed in 96b2775f and 980ef00c: all three returned MP4s now require a non-empty, decodable audio stream via PyAV. The Ref2VA request is single-item to avoid the DLO DP collective timeout observed in Buildkite; FL2VA remains the concurrent DP2 coverage.

david6666666:
Self-review: verified the L3 cases cover official FL2VA/Ref2VA DLO DP2 and Turbo LoRA with TP2 DLO no-AllGather. Audio streams are checked with PyAV, the Buildkite nested-quote failure is fixed, and B300 plus unit/pre-commit validation passed.

david6666666:
Follow-up update for this PR:

- Reverted the pure-DP request-preparation workaround in `pipeline_minimax_h3.py`. `_dit_rank_world()` and the Ref2VA audio metadata broadcast are back on the DiT WORLD group path.
- Added four separate L3 modules under `tests/e2e/online_serving/minimax/`: DLO/DP2 FL2VA, DLO/DP2 Ref2VA, Turbo LoRA, and the requested two-card FastH3 LoRA layout (text-encoder TP2, DiT HSDP2 + USP2, VAE patch/VPP2).
- Ref2VA is intentionally a single request after the revert so all ranks observe the same WORLD-group collective; FL2VA remains a concurrent DP2 wave.
- The FastH3 case follows [PR #6714](https://github.com/vllm-project/vllm-omni/pull/6714): load-time `--lora-path`, four-step T2VA at 1024x576, and no request-level `lora`/shift override. This PR branch predates the FastH3 loader, so the test is guarded and will skip until that support is present in the merge baseline.
- On 2x B300, the same FastH3 configuration produced a valid 1024x576, 24-fps MP4 with AAC audio. The sampled transient peak was 65,346 MiB; after compile/cache warm-up the resident peak was 77,978 MiB (~76.2 GiB per GPU). No H100 was available locally, so 80 GiB H100 compatibility is an inference with roughly 3.8 GiB measured margin.
- The Buildkite job now starts each lifecycle in a fresh pytest process, downloads optional artifacts only when their loader exists, and applies `timeout 35m` per case to avoid stale `resource_tracker` state and Ref2VA collective hangs.

Validation: changed-file pre-commit, compileall, YAML/Buildkite validation, the WORLD-group regression test (`1 passed`), and new-test collection (`4 collected`) passed locally. Full E2E execution requires the CI GPU/model environment.

david6666666:
Conflict-resolution follow-up:

- Merged the latest upstream/main into this PR and resolved the only textual conflict in the MiniMax-H3 parallel regression test, keeping the WORLD-group behavior assertion.
- Pushed merge commit 5007a06e0. The branch is now reported as MERGEABLE; remaining BLOCKED status is from pending checks, not a merge conflict.
- The merged main also contains the FastH3 loader from PR #6714, so the new two-card FastH3 L3 case is now active instead of source-guarded.
- Re-toggled the merge-test label to schedule CI for the resolved head.

david6666666:
Follow-up: removed `tests/diffusion/models/minimax_h3/test_minimax_h3_parallel.py` and `tests/e2e/online_serving/test_minimax_h3_l3_variants.py` per request. The PR now keeps only the four dedicated cases under `tests/e2e/online_serving/minimax/`, and CI was re-triggered after pushing commit 00920dbce.

david6666666:
Correction: restored `tests/diffusion/models/minimax_h3/test_minimax_h3_parallel.py` exactly to the latest main version; it is no longer changed by this PR. The legacy `tests/e2e/online_serving/test_minimax_h3_l3_variants.py` remains removed. The PR net diff now contains only the dedicated `minimax/` four-case suite and its CI job. Pushed commit 3b5a29218 and re-triggered merge-test.

david6666666:
Renamed the dedicated serving-test directory from `tests/e2e/online_serving/minimax` to `tests/e2e/online_serving/minimax_h3` and updated all Buildkite dependencies/commands. The four L3 tests are unchanged; collection still reports 4 tests. Pushed commit 78305ed06 and re-triggered merge-test.

david6666666:
Ref2VA DLO/DP2 hang root-cause analysis and fix:

- In [Buildkite #14458](https://buildkite.com/vllm/vllm-omni/builds/14458/canvas?sid=01a056df-f852-4a9a-89ce-66690cb9b302&tab=output), FL2VA completed successfully with two concurrent requests. Ref2VA stopped immediately after `RequestBatch admission wait done waiting=1 max_batch=2`; there was no OOM or worker exception.
- DLO + DP2 + AllGather enables DP multi-concurrency. The scheduler must issue a two-request wave so each DP rank selects one request and all ranks participate in the WORLD-group collectives used by MiniMax-H3 reference-video preparation. The old single-request Ref2VA test left one rank out of the collective, so the test hung.
- Fixed in commit `6fd53c7ee`: Ref2VA now uses the same `run_dlo_wave` two-request pattern as FL2VA. No MiniMax-H3 pipeline/code change was needed.

I also re-triggered the `merge-test` label CI for this fix.

david6666666:
Correction after validating the two-request wave in Buildkite #14463:

- The two-request test fix correctly changed admission from `waiting=1` to `waiting=2`, but Ref2VA still stopped before the first pipeline log. This confirms that the remaining issue is not the test wave itself.
- The earlier pure-DP request-preparation isolation from `8646ef3b` is required for this topology: direct WORLD collectives couple unrelated Ref2VA requests before DLO AllGather. Restored the minimal fix in commit `62834dc0e`; pure DP uses a logical local group for request-specific text/reference/VAE preparation, while model-parallel groups still use WORLD and DLO weight AllGather is unchanged.
- The Ref2VA test remains a two-request wave so both DP workers are active.

The old single-request test was invalid because DLO DP2 admission requires one request per DP rank. I re-triggered merge-test for the corrected code path.

david6666666:
Ref2VA DLO/DP2 hang follow-up (commit `bcacf3351`):

- Buildkite #14458 stopped after `waiting=1 max_batch=2`; #14463 and #14464 admitted two requests (`waiting=2`) but still emitted no `MiniMax H3 ref2va Qwen presentation` or denoise log. There was no OOM/worker exception in the captured logs; the stall is before the first pipeline log in the video-reference preprocessing path.
- The temporary production change in `pipeline_minimax_h3.py` has been reverted to upstream. The merge L3 case now uses two concurrent, official image-only Ref2VA requests, which exercises the same Ref2VA + DLO/DP2 scheduling/denoise path without launching two per-request ffmpeg/Qwen video preprocessors in the merge wave. Video-reference Ref2VA remains covered by the existing contract and accuracy suites.
- The shared request timeout is 600 seconds, so a future collective/preprocessing regression fails within the CI budget instead of occupying the 35-minute command timeout.

`merge-test` was re-toggled to run the updated four-case MiniMax-H3 job.

david6666666:
CI follow-up (Buildkite #14468, commit `bcacf3351`): the updated MiniMax-H3 L3 shard is green.

- DLO/DP2 FL2VA: 1 passed; two requests admitted (`waiting=2`, `max_batch=2`), both HTTP 200.
- DLO/DP2 Ref2VA: 1 passed; image-only two-request wave reached `MiniMax H3 ref2va Qwen presentation` (7195 tokens), completed denoise, and both requests returned HTTP 200. This no longer reproduces the #14458/#14464 stall.
- Turbo LoRA: 1 passed; TP2 + VAE patch parallel 2, load-time LoRA, HTTP 200.
- FastH3: 1 passed in 420.65s on H100 with the requested `--use-hsdp --hsdp-shard-size 2 --usp 2 --text-encoder-tp-size 2 --vae-patch-parallel-size 2`; 4 denoise forwards completed and the request returned HTTP 200.

FastH3 reported 61.3977 GiB model loading and 62.72 GiB process-scoped GPU memory after load on each worker, with no OOM; the CI helper only samples nvidia-smi before/after teardown, so it does not expose an exact in-request peak. The successful H100 run confirms this case fits the 80 GB cards.

The production `pipeline_minimax_h3.py` change remains reverted to upstream. The fix is test-side: use the official image-only Ref2VA path for the concurrent DLO/DP2 merge wave (video-reference preprocessing remains covered by the contract/accuracy suites) and bound request timeout to 600s. Buildkite: https://buildkite.com/vllm/vllm-omni/builds/14468

Gaohan123 (approved):
LGTM. Thanks

PR #6560, opened 2026-08-24
[Model] Add Cosmos3 mixed W8A8/W8A16 and W4A4/W4A16 denoising

wkutak:
## Purpose

This PR adds opt-in, denoising-step-aware mixed activation precision for
Cosmos3 diffusion models backed by serialized ModelOpt checkpoints.

The implementation:

- Uses the checkpoint-native quantized linear path for middle denoising steps
  and a dense A16 path for configured first and last steps.
- Supports W8A8/W8A16 scheduling for serialized tensorwise ModelOpt FP8
  checkpoints and W4A4/W4A16 scheduling for supported ModelOpt NVFP4
  checkpoints.
- Selects precision once at each scheduler-step boundary, so conditional,
  unconditional, and CFG-parallel transformer calls within one denoising step
  use the same precision.
- Keeps reasoner precision independently configurable as checkpoint-native or
  A16.
- Dynamically discovers compatible reasoner and generation linears instead of
  relying on model-size-specific layer counts.
- Reuses the live quantized weight representation. The A16 path dequantizes the
  current local weight partition and executes it with
  `torch.nn.functional.linear`.
- Reads an optional versioned policy from
  `quantization_config.runtime.diffusion_step_policy` in the authoritative
  `transformer/config.json`.
- Allows an explicit `cosmos3_mixed_precision` runtime object to override or
  disable a checkpoint policy without disabling checkpoint quantization.
- Fails closed for malformed policies, unsupported or mixed checkpoint
  formats, SmoothQuant FP8, non-tensorwise FP8 scales, incompatible FP8
  backends, and unsupported NVFP4 layouts.

When a checkpoint declares the following policy, the first and last three
denoising iterations use A16 and the middle iterations use the checkpoint's
native quantized path:

```json
{
  "quantization_config": {
    "quant_method": "modelopt",
    "quant_algo": "FP8",
    "runtime": {
      "diffusion_step_policy": {
        "schema_version": 1,
        "type": "first_last_n",
        "index_space": "denoising_loop_iteration",
        "scope": ["transformer"],
        "default_mode": "native",
        "first_steps": {"count": 3, "mode": "a16"},
        "last_steps": {"count": 3, "mode": "a16"},
        "overlap": "a16",
        "reasoner": "a16"
      }
    }
  }
}
```

Concrete example in HF Cosmos3 Super [repository](https://huggingface.co/nvidia/Cosmos3-Super/blob/fp8/transformer/config.json#:~:text=%22FP8_FP8%22%2C-,%22runtime%22%3A%20%7B,%7D%2C,-%22scale_block_quantize%22%3A) (fp8 branch).

Missing policy metadata preserves ordinary checkpoint-native execution. An
explicit runtime override takes precedence, while
`{"cosmos3_mixed_precision":{"enabled":false}}` disables only the schedule.

## Scope

This PR is limited to mixed-precision policy parsing, scheduler integration,
linear discovery, and native/A16 dispatch.

It intentionally does not add a mixed-precision dense-weight cache, pinned-host
staging, block prefetching, or alternative cache-placement modes. The A16 path
materializes its dense local weight directly from the live quantized weight
when invoked. Weight-caching and kernel-performance optimizations are deferred
to a separate PR.

This PR also does not introduce a custom W8A16 or W4A16 kernel. Attention and
other non-linear operations are unchanged.

## Test Plan

Run the focused mixed-precision suite:

```bash
python3 -m pytest -q -s \
  tests/diffusion/quantization/test_cosmos3_mixed_precision.py
```

The suite covers:

- Nested runtime configuration parsing, validation, and explicit disable.
- Checkpoint policy discovery and runtime-override precedence.
- Strict checkpoint format and policy-schema validation.
- First/last-step boundaries, overlap, and one-step schedules.
- Independent reasoner and generation precision.
- Native/A16 linear dispatch over live FP8 and NVFP4 weights.
- Canonical FP8 scale/layout validation and incompatible-backend rejection.
- NVFP4 layout and scale validation.
- Dynamic linear discovery.
- Scheduler-step forwarding and request-end state reset.
- Exact per-request precision traces.

Run Cosmos3-Nano I2V with the same fixed request for BF16, full W8A8, full
W8A16, and mixed W8A16/W8A8:

- 1280x720 output.
- 189 frames at 24 FPS.
- 50 denoising steps.
- Guidance scale 6.
- Flow shift 10.
- Full-range CFG.
- Seed 0.
- Three independent engine launches per policy.

For the mixed policy, verify the exact trace:

```text
A16 x3 / NATIVE x44 / A16 x3
```

Verify every retained video for dimensions, frame count, FPS, motion,
conditioning-frame alignment, request and asset hashes, checkpoint policy,
selected native backend, tensor-parallel configuration, and output hash.

**vLLM Version:** `v0.27.0`

**vLLM-Omni Git HEAD:** `bdc9a787734016f7cdb70a024cf7d5a36d467d22`

**Benchmark tracked-diff SHA-256:**
`02264e65917d56442e1519b3f746fd50f7620ddabdad131598689ae019ff169a`

**Benchmark image:**
`vllm-omni-cosmos3-mixed:h100-pcie-precision-bdc9a787-local`
(`sha256:bfa82a31386b8758d191df4cd1b94eab1ed95d4615db55b0d46ab4a94e7b903e`)

## Test Result

- Focused CUDA suite: `53 passed`.
- Ruff checks and formatting: passed.
- Python compilation: passed.
- Request and checkpoint-policy validation: passed.
- `git diff --check`: passed.
- All retained I2V videos passed shape, FPS, motion, conditioning alignment,
  policy, backend, and exact precision-trace validation.
- Each policy produced one deterministic MP4 SHA-256 value across its three
  repetitions.

### Cosmos3-Nano I2V performance

Measured on one NVIDIA H100 PCIe GPU with tensor parallel size 1 and concurrency
1. Each repetition used a fresh engine. Generation time is the wall-clock time
of the retained `Omni.generate` request; engine load time is excluded.

| Precision policy | Repetitions | Mean (s) | Median (s) | Sample stdev (s) | Min-max (s) | Speedup vs. BF16 |
| Full BF16 | 3 | 517.696 | 517.807 | 0.331 | 517.323-517.958 | 1.000x |
| Full W8A8 | 3 | 474.665 | 475.054 | 1.279 | 473.237-475.705 | 1.090x |
| Full W8A16 | 3 | 521.670 | 522.230 | 1.521 | 519.948-522.832 | 0.992x |
| Mixed W8A16 first/last 3, W8A8 middle 44 | 3 | 479.166 | 479.293 | 0.224 | 478.907-479.298 | 1.080x |

The mixed schedule retained 99.1% of full-W8A8 median throughput and was 8.0%
faster than BF16 for this request. Full W8A16 was approximately equal to BF16,
which is expected for the correctness-oriented dense A16 reference path in
this PR.

The measurements above cover the serialized ModelOpt FP8 W8A8/W8A16 strategy.
NVFP4 behavior is covered by the focused strategy and layout tests, but this PR
does not make an NVFP4 end-to-end performance claim.


wkutak:
Following are the examples of I2V generation with different precisions

BF16

https://github.com/user-attachments/assets/02ac1d61-8f4f-4259-acdf-9318e97d9c1e

MIXED 3+3 W8A16 / W8A8 middle

https://github.com/user-attachments/assets/a6de925f-a7b7-4be0-833d-0e40cdad4492

W8A8

https://github.com/user-attachments/assets/f355c8cf-f3d8-4790-8918-e9694c3835e0

W8A16

https://github.com/user-attachments/assets/98e03b6e-ee09-4903-aa01-8e408a96c9f4

Peyman13l (approved):

hsliuustc0106:
@david6666666 PTAL

vllm_omni/diffusion/models/cosmos3/mixed_precision/checkpoint.py
```diff
@@ -0,0 +1,168 @@
...
+def resolve_mixed_precision_config(
+    od_config: object,
+) -> tuple[Cosmos3MixedPrecisionConfig | None, PolicySource]:
+    """Resolve an explicit runtime override before the checkpoint default."""
+    override_present, override = Cosmos3MixedPrecisionConfig.resolve_additional_config(
+        getattr(od_config, "additional_config", None)
+    )
+    if override_present:
+        source: PolicySource = "additional_config" if override is not None else "additional_config_disabled"
+        return override, source
```
hsliuustc0106, line 54: [P2] Validate checkpoint compatibility for runtime overrides

Supplying `{"cosmos3_mixed_precision":{}}` returns here before `_validate_checkpoint_quantization()`, so an explicit schedule bypasses the advertised rejection of mixed-format checkpoints. `Cosmos3MixedPrecisionRuntime.install()` then silently skips unsupported quantized linears as long as it finds another compatible linear; for example, an FP8/MXFP8 configuration would only partially follow the requested A16 schedule. Please validate the effective quantization configuration for every enabled policy, preserve the explicit-disable escape hatch, and add an override-path regression test.

hsliuustc0106:
Adding a local Cosmos3-Nano FP8 A/B data point at head `3683fb4a24455982906efd02a2f8fff817ac86f5`, compared with base `6d84409a683aeef476ad2d47a839998ef2344b4b`. Both source snapshots were unmodified.

One NVIDIA L20X (SM90), TP=1, concurrency=1; 384×256, 17 frames at 24 FPS, 12 denoising steps, seed 0, guidance 6, flow shift 10, full CFG. Used the serialized FP8 `nvidia/Cosmos3-Nano` checkpoint and official I2V assets at revision `1af44289002ad8afe2567dd85e34fe5dea21e523`. BF16 activations, eager execution, no cache acceleration or CPU/layerwise offload; guardrails, sound and action generation disabled for both configurations.

| Configuration | Measured request times (s) | Mean ± sample SD (s) | Peak allocated GPU memory (GiB) |
|---|---|---|---|
| Base: native FP8 | 1.57845, 1.59131 | 1.58488 ± 0.00909 | 17.10741 |
| Head: first/last 3 steps A16, middle native, reasoner A16 | 1.93165, 1.92466 | 1.92816 ± 0.00494 | 17.10770 |

This is **21.7% higher latency** for mixed precision on this small workload. Each configuration had one excluded feasibility/warmup request followed by two measured requests on the same initialized engine, in A-then-B order on the same GPU with fixed CPU affinity. Timing covers synchronized `DiffusionEngine.step_streaming` preprocessing, generation and postprocessing; loading, startup warmup and output export are excluded.

Validation: both configurations produced finite, correctly shaped outputs and bitwise-identical raw outputs across their two measured repeats. The mixed generation trace was exactly `A16 ×3 / native ×6 / A16 ×3`, with matching conditional/unconditional precision and successful request-end reset. The focused mixed-precision suite passed locally: **52 passed**.

For quality, inspection of matching frames 0/4/8/12/16 found both clips visually plausible, with no obvious gross corruption unique to mixed precision and no clear quality winner. Their first frame is identical; later viewpoints and scene details diverge. Native FP8 shows more visible dust near the bend in the sampled frames, but this is not evidence of better quality. The clip is only 0.71 s while the prompt describes a 7 s driving/rockfall/braking sequence, so this does not establish temporal prompt adherence or quality equivalence/improvement. No full-BF16 quality reference was run.

This supplements the PR benchmark rather than reproducing it: 6/12 steps here use A16 (50%), versus 6/50 (12%) in the reported H100 workload, with different hardware, resolution and clip length. Both sides used vLLM 0.28.0 / PyTorch 2.13.0+cu129 / Transformers 5.14.1 / Diffusers 0.40.0 / FlashInfer 0.6.16.post3; the available vLLM 0.27 environment could not import the frozen source (`create_error_response` missing). The two warmed repeats and in-process timing also differ from the PR's three fresh-engine `Omni.generate` repetitions. These results do not verify or refute the reported H100 overhead or BF16 speedup.

hsliuustc0106:
Review summary for `3683fb4a24455982906efd02a2f8fff817ac86f5`:

- The actionable correctness finding is the [checkpoint-validation bypass with an explicit mixed-precision override](https://github.com/vllm-project/vllm-omni/pull/6560#discussion_r3953169317). For a mixed or unsupported checkpoint format, this can silently apply the schedule to only some linears, contrary to the intended fail-closed behavior. Please validate the effective checkpoint quantization before accepting an enabled override and add regression coverage, while preserving the explicit-disable escape hatch.
- The [local A/B data point](https://github.com/vllm-project/vllm-omni/pull/6560#issuecomment-5577444659) showed 21.7% higher latency on L20X at 384×256, 17 frames and 12 steps, with essentially unchanged peak allocated memory. The workload and A16 step fraction differ substantially from the PR benchmark, so this does not refute its reported H100 result; the small-overhead claim should remain scoped to that measured configuration.
- Sampled-frame inspection found no clear quality winner or obvious gross degradation unique to mixed precision. This short, single-prompt test does not establish quality equivalence or improvement.

The tested FP8 path otherwise passed the precision-transition, request-end reset and repeatability checks, and all 52 focused tests passed locally.

PR #6561, opened 2026-08-24
[Bugfix] Fix host-memory leak on aborted /v1/images/generations (#6462)

### LINKED ISSUE #6462: [Bug]: Host-memory leak in api_server frontend on aborted /v1/images/generations (diffusion) requests
### Your current environment

<details>

- Image: `quay.io/vllm/automation-vllm-omni:cuda-32248598484` → vllm-omni `0.26.1.dev12+rhaiv.0.g9629fa34a`, vllm `0.24.1`
- GPU: 1× NVIDIA L4 24 GB (AWS `g6.xlarge`), driver 595.91.07
- Model: `black-forest-labs/FLUX.2-klein-4B` (single-stage diffusion), served with `--ipc=host`
- Endpoint: `POST /v1/images/generations`

</details>

### 🐛 Describe the bug

Aborted image-generation requests leak host memory in the **`api_server` frontend process (PID 1)**, linearly and without being reclaimed after load stops.

This is a **separate bug from #6352**. #6352 is a sender-side leak in the `OmniARScheduler` + `SharedMemoryConnector` chunk-transfer path (audio/AR models), living in the **stage engine core process**. This one is in a **different process** (the frontend `api_server`) on a **different path** (diffusion/image, which uses neither `OmniARScheduler` nor the chunk-transfer adapter). Filing separately so they aren't conflated.

**Trigger:** client disconnect / request cancellation *mid-generation* (an *abort*), not failed/malformed requests. Completed requests do not leak.

### Observed behaviour

Monitoring `VmRSS` of the `api_server` process (PID 1) while driving `/v1/images/generations` and cancelling each request mid-generation:

| Phase | Requests | api_server RSS | Trend |
|---|---|---|---|
| Idle | — | stable (repeated reads identical) | flat |
| Completed image requests | 148 completed, 0 aborts | +3.9 MiB warmup then flat | plateaus — no leak |
| Aborted image requests (burst 1) | 360 aborts | **+74.2 MiB** | **linear, no plateau** |
| Aborted image requests (burst 2, repeat) | 357 aborts | **+72.7 MiB** | **linear, reproducible** |

- Steady-state slope ≈ **~210 KB/abort**, constant across the run.
- Growth is **retained after load stops** — idle RSS settles at the new higher level, it is not reclaimed.
- **Abort-specific:** completed image requests keep the frontend RSS flat; only aborts grow it.

### It is diffusion/image-specific, not a general frontend leak

Running the same abort burst against the **audio** endpoint (`/v1/audio/speech`, Qwen3-TTS) while monitoring the *same* `api_server` PID 1 shows only a one-time warmup (+56 MiB in the first 20 s from the concurrent-connection ramp) then a **flat** plateau for the remaining ~960 aborts — windowed steady-state slope ≈ 0.2 KB/abort ≈ 0. So the frontend does **not** leak per-abort on the audio path; the ~210 KB/abort linear growth is unique to the diffusion/image path.

Likely cause: when the client disconnects, the request handler is cancelled while the diffusion result / buffered image bytes are left unreclaimed in the frontend.

### How to reproduce

1. Serve a diffusion model, e.g. `black-forest-labs/FLUX.2-klein-4B`.
2. Fire valid `POST /v1/images/generations` requests and cancel each one shortly after generation starts (e.g. `asyncio.wait_for(..., timeout=1.0)` so the connection closes mid-flight → server records the abort).
3. Sample `VmRSS` of the `api_server` process (PID 1) over time.

Expected: RSS plateaus (as it does for completed requests / audio aborts). Actual: RSS grows ~210 KB per aborted request, linearly, and is retained.

### Note

Reported from downstream validation; happy to share the repro harness. No fix attached — reporting only.

zhang-keliang:
## Purpose

Fixes #6462 (host-memory leak in `api_server` on aborted `/v1/images/generations`).

Every request keeps `self._consumed_metric_messages[request_id]` — a per-request `set[int]` populated by `_handle_output_message` (`omni_base.py`) to de-dup metrics. The completion path (`_log_summary_and_cleanup`) pops both `request_states[rid]` and `_consumed_metric_messages[rid]`. The abort path — `AsyncOmni._abort` and the sync mirror `Omni.abort` — popped only `request_states`, so every aborted request retained one `set[int]` entry for the process lifetime.

Reporter's fingerprint matches: abort-only, diffusion-only (audio has its own dict with explicit `.pop()` sites in `serving_speech.py`), linear, retained after load stops. Related to #6413 / #6439 which fixed the executor-side abort leak (`_completed_outputs`); this covers the frontend side.

Fix: `_abort` (async) and `Omni.abort` (sync) now pop `_consumed_metric_messages[req_id]` in lockstep with `request_states`. Uses `getattr(self, "_consumed_metric_messages", None)` so existing subclasses / test doubles that don't initialise the dict still work.

## Test Plan

Added `tests/entrypoints/test_async_omni.py::test_abort_evicts_consumed_metric_messages_entry` alongside the existing `test_abort_handles_internal_request_mapping`. Uses the pre-existing `get_async_omni_instance` scaffold, seeds both dicts, calls `omni._abort([req_id])`, asserts both dicts drop the key. Fails on main, passes on this branch.

CPU-only tracemalloc reproduction (200 abort cycles × 40 metrics messages per request):

- Before fix: 71.1 KB retained (364 B/req); `_consumed_metric_messages` size = 200.
- After fix: 3232 B retained (16 B/req, all tracemalloc bookkeeping); dict empty.
- Completion control (both before and after): ~4 B/req, dict empty.

**vLLM Version:** N/A for this fix (pure frontend Python; no vLLM version bump)

**vLLM-Omni Commit:** off `3d035bf` (main at the time of the fix)

## Test Result

- Regression test passes on this branch (verified in isolation via a standalone driver that extracts `_abort` source, since the full pytest suite has heavy runtime deps not present in my minimal env).
- `ruff check` clean; `ruff format --check` clean (`ruff==0.14.10` per `.pre-commit-config.yaml`).
- No behavior change on the non-abort paths; no public API change.


NickCao:
Could `_consumed_metric_messages`, instead of being a dict keyed by request_id, be an attr attached to the request object itself? That way it can never leak.

zhang-keliang:
Adopted, thanks — moved the dedup set to `ClientRequestState.consumed_metric_message_ids`. Dropping the class-level `OmniBase._consumed_metric_messages` dict + its `setdefault` helper + the explicit dict pop in `_log_summary_and_cleanup`; the set now drops with the state via GC on both completion and abort. Net -14 LoC. New head 2868c44.

zhang-keliang:
Self-review:

- Scope: 4 files, +35/-16, all Python host-side. No CUDA / kernel / config touched.
- Root cause: `_consumed_metric_messages[rid]` populated on every stage metric emit but only popped in `_log_summary_and_cleanup`; abort skipped it → linear leak.
- After @NickCao's suggestion the set now lives on `ClientRequestState`, so completion and abort share one cleanup site (`request_states.pop(rid)`); no per-callsite dict maintenance.
- Regression test: `test_abort_drops_consumed_metric_message_ids_with_state` — populated set on the state → abort → state pops → set gone (also guards against re-introducing a class-level dict via `hasattr`).
- Ran locally: `pytest tests/entrypoints/test_async_omni.py tests/metrics/test_prometheus.py -m cpu` → 13 passed. `ruff check` and `ruff format --check` on all 6 modified files → clean (only pre-existing UP038 warnings on unrelated lines).
- No GPU / model-weight reproduction on this branch (host-side lifecycle only; leak measured by reporter with tracemalloc in `api_server` PID 1, and my CPU repro in the linked issue does the same). No perf benchmarks claimed.

hsliuustc0106:
resolve conflicts

zhang-keliang:
Rebased onto latest main and resolved the conflict — `main` had moved `prom_metrics` initialization earlier in `__init__` (post-0.28 rebase), and the new `test_same_finished_image_message_is_observed_exactly_once` in `test_emit_calls.py` stubbed the old class-level dict, so its fixture now seeds `request_states` with a per-request state instead. Branch is conflict-free (head 5abddb2); `pytest tests/metrics/test_emit_calls.py tests/entrypoints/test_async_omni.py tests/metrics/test_prometheus.py -m cpu` → 54 passed locally.

tests/entrypoints/test_async_omni.py
```diff
@@ -237,6 +237,31 @@ async def failing_abort_async(request_ids):
...
+    """Abort pops ``request_states[rid]`` and the per-request de-dup set
+    attached to it drops with the state (no class-level dict to maintain).
+    Regression guard for #6462 under the ``ClientRequestState``-attached
+    layout adopted in #6561 review — a stray class-level
+    ``_consumed_metric_messages`` re-introduction would resurface the leak.
+    """
+    from vllm_omni.entrypoints.client_request_state import ClientRequestState
+
+    async def run():
+        omni = get_async_omni_instance()
```
hsliuustc0106, line 352: [P2] Exercise the production leak path in this regression

This helper uses `object.__new__`, manually seeds the new field, and never runs `OmniBase.__init__` or a metric-bearing output through `_handle_output_message`. Reintroducing and populating the old request-keyed map in production could therefore still pass this test. Drive a metric message through the production handler, abort the request, and assert that no independent request-keyed metric state survives.

zhang-keliang: Done in d9d2b69 — the regression now drives a metric-bearing `OutputMessage` through `_handle_output_message` (asserting the handler records the de-dup entry on the state and de-duplicates a replay of the same message object), aborts via the public `abort()`, and then sweeps `vars(omni)` so no mapping on the instance still keys anything by the request id. A reintroduced request-keyed map populated by the production handler fails either the sweep or the `hasattr` guard.

tests/entrypoints/test_async_omni.py
```diff
@@ -237,6 +237,77 @@ async def failing_abort_async(request_ids):
...
+        # The production handler recorded the de-dup entry on the state...
+        assert id(msg) in state.consumed_metric_message_ids
+        # ...and de-duplicates a replay of the same message object.
+        omni._handle_output_message(msg)
+        assert len(state.consumed_metric_message_ids) == 1
+
+        assert not hasattr(omni, "_consumed_metric_messages")
+
+        await omni.abort("req-1")
+        assert rid not in omni.request_states
```
hsliuustc0106: P1 — Please rebase this regression test onto current `main`'s abort lifecycle. Since #6367, `AsyncOmni._abort()` keeps `request_states` registered, queues a terminal abort output, and relies on `generate()` cleanup to remove the state. Current `main` therefore conflicts with this test, and if this assertion is carried over it fails immediately after `await omni.abort("req-1")`. Please preserve the newer abort tests and assert that the per-request set is released after the normal `generate()`/`finally` cleanup, or update this test to the terminal-message contract.

hsliuustc0106 (commented):
I found one blocking integration issue in the regression test; see the inline comment. No additional high-confidence production-code findings.

tests/entrypoints/test_async_omni.py
zhang-keliang: Done — rebased onto current `main` (973d20f) and moved the test to the #6367 contract. Both new abort tests from `main` (`test_abort_enqueues_prefix_tokens_from_engine`, `test_abort_enqueues_synthetic_finished_when_engine_returns_empty`) are preserved. The regression now: drives a metric-bearing `OutputMessage` through `_handle_output_message` → `abort()` → asserts the state is *still* registered and the terminal abort output is on its queue (with the de-dup set living only on that state) → runs `generate()`'s cleanup via `_log_summary_and_cleanup` → asserts the state and the set are released together, with a `vars(omni)` sweep for any surviving request-keyed mapping. `pytest tests/entrypoints/test_async_omni.py tests/metrics/test_prometheus.py tests/metrics/test_emit_calls.py -m cpu` → 58 passed.

hsliuustc0106 (approved):
lgtm

PR #6563, opened 2026-08-24
[Diffusion] Add paged KV cache support for HunyuanImage3 DiT on NPU and GPU

Acerak01-fy:
## Purpose

- This PR consolidates the GPU implementation from #6658 (@zwhzzz0821) into this PR and enables HunyuanImage-3.0 DiT-stage end-to-end execution with Scheduler-managed paged KV cache on NVIDIA GPU and Ascend NPU.
- This work is part of RFC #5244 and builds on #6094 and #6102. The Scheduler owns logical KV allocation and release. The Worker owns physical KV tensors, native `BlockTables`, slot mappings, and attention metadata.
- Paged row construction and prefill/denoise phase activation are owned by the diffusion model runner. The Hunyuan model supplies its Q/K/V layout and `full_attn_spans`; it does not allocate Scheduler blocks or activate the Worker runtime.
- The first denoising step writes the complete allocated sequence. Later steps update only the timestep/image-token span. Subsequent piecewise attention calls read the persistent Worker-managed KV pages.
- The shared piecewise-attention implementation has two paths: a contiguous homogeneous-row fast path (direct output-buffer assembly) and an indexed fallback for heterogeneous rows. The fast path removes unnecessary per-segment output `index_copy_`/`ScatterUpdate` for the common Hunyuan CFG layout.
- On NVIDIA GPU, paged attention uses the native FA3 paged contract with TP, CFGP, strict Ulysses SP, and mixed causal/full piecewise attention.
- On Ascend NPU, paged attention uses Ascend-native FIA, `AscendBlockTables`, Ascend block geometry, and vLLM-Ascend's normal-layout `do_kv_cache_update`. 
- This PR targets request-level execution. Hunyuan paged step execution/continuous batching, imported AR-to-DiT KV, Ring SP, and AllGather-KV SP remain out of scope.

## Test Plan

- GPU E2E: run Dense/Paged comparisons on H200 with the same FA3 backend for TP4, TP2+SP2 (Ulysses), and TP2+CFGP2.
- GPU profile: after warmup, profile TP2+CFGP2 Dense and Paged with the same FA3 binary to separate attention-kernel cost, KV preparation, and work removed by paged output-buffer writes. Profiler time is excluded from E2E measurements.
- NPU unit tests: run the focused piecewise-attention and diffusion paged-attention adapter suites, including homogeneous fast path, heterogeneous fallback, row metadata, one-write-per-layer semantics, and native metadata validation.
- NPU E2E: run fresh-process Dense/Paged lanes for TP4/SP1, TP4/SP2, and TP4/CFGP2 with identical inputs, `request_execution` , 20 denoising-step warmup, and 20 measured denoising steps.
- NPU profile: collect warmed 3-step traces for Dense/Paged for all three topologies. 
- Pixel consistency: pair Dense/Paged outputs from the same sample round and report MAE, PSNR, SSIM-RGB, SSIM-luma, and exact-pixel ratio.

### NVIDIA GPU

The complete GPU Dense/Paged E2E and Nsight Systems analysis is available at:

https://github.com/vllm-project/vllm-omni/pull/6563#issuecomment-5449800888

The referenced comment covers H200 FA3 validation for TP4, TP2+SP2, and TP2+CFGP2, including latency, throughput, P99, attention-kernel cost, KV preparation cost, and piecewise output-buffer optimization.

### Ascend NPU

Validated on 8 Ascend 910B1 cards, CANN 9.1.0, the image's ATB `latest` runtime, Python 3.12.13, `torch 2.10.0+cpu`, and `torch_npu 2.10.0.post4`. The image contains core vLLM 0.28.0 and the independently versioned vLLM-Ascend package `0.19.1rc2.dev1690`. the run used 1024x1024 bfloat16 inference, `request_execution` , eager mode, 20 warmup denoising steps, and 20 measured denoising steps per lane.

#### NPU E2E

Every lane completed 5/5 measured inference samples and produced finite 1024x1024 RGB images. `Paged vs Dense` is `(Paged / Dense - 1) * 100%`; positive values mean Paged is slower. Worker memory is the reserved peak, including the paged KV pool.

| Topology | Dense median latency | Paged median latency | Paged vs Dense | Worker peak reserved (MiB), Dense / Paged |
|---|---:|---:|---:|---:|
| TP4 | 11.0637 s | 11.4155 s | +3.18% | 55,882 / 59,906 |
| TP4/SP2 | 10.4215 s | 10.5083 s | +0.83% | 36,424 / 48,526 |
| TP4/CFGP2 | 9.1029 s | 9.1775 s | +0.82% | 36,422 / 40,448 |

These are five-sample medians. The larger Paged Worker peak is expected from page-pool reservation and is not evidence of a leak.

### Pixel consistency on NPU

The following values are medians over five same-round Dense/Paged pairs from the E2E outputs. Both sides use the same prompt, seed, resolution, and denoising steps.

| Topology | Mean abs median (0-255) | PSNR median (dB) | SSIM-RGB median | SSIM-luma median | Exact-pixel median |
| TP4 | 4.6152 | 30.1405 | 0.957499 | 0.958187 | 1.56% |
| TP4/SP2 | 6.5412 | 26.7809 | 0.928255 | 0.928886 | 0.96% |
| TP4/CFGP2 | 5.8614 | 28.3284 | 0.939830 | 0.940480 | 0.95% |

#### attention operator

The device-task times below are averaged over four/eight ranks and normalized to one rank per measured denoising step. Call counts are the integer total for the three measured steps on one rank. Dense uses `FlashAttentionScore`; Paged uses native `FusedInferAttentionScore`. The different CANN contracts and shapes mean this is a path comparison, not a same-shape microbenchmark.

| Topology | Dense selector / calls (3-step total/rank) | Paged selector / calls (3-step total/rank) | Dense core (ms/rank/step) | Paged core (ms/rank/step) | Paged - Dense |
|---|---|---|---:|---:|---:|
| TP4 | `FlashAttentionScore` / 96 | `FusedInferAttentionScore` / 224 | 38.613 | 31.540 | -7.073 ms (-18.3%) |
| TP4/SP2 | `FlashAttentionScore` / 96 | `FusedInferAttentionScore` / 224 | 21.854 | 15.419 | -6.435 ms (-29.5%) |
| TP4/CFGP2 | `FlashAttentionScore` / 96 | `FusedInferAttentionScore` / 224 | 21.692 | 16.999 | -4.693 ms (-21.6%) |

Paged has 224 FIA calls over the three measured steps (`32 layers * (3 + 2 + 2)`), versus 96 Dense FA calls (`32 layers * 3`). Despite the finer granularity, the aggregate FIA device time is lower in all three traces.

#### data preparation and output assembly
The entries below are device-task milliseconds per rank per measured step. Index/gather and slice categories include some model-common work.

| Topology | Dense mask | Dense FA relayout | Dense Cat/Concat | Paged KV write | Paged IndexSelect/Gather | Paged output ScatterUpdate | Paged TensorMove + Slice | Paged Cat/Concat |
|---|---:|---:|---:|---:|---:|---:|---:|---:|
| TP4 | 2.026 | 3.033 | 1.007 | 1.825 | 10.035 | 0.000 | 5.251 | 0.832 |
| TP4/SP2 | 1.985 | 2.099 | 1.786 | 1.732 | 10.038 | 0.000 | 3.136 | 3.089 |
| TP4/CFGP2 | 0.921 | 2.254 | 0.806 | 1.114 | 4.997 | 0.000 | 1.849 | 0.120 |

The latest homogeneous fast path has zero active `aclnnInplaceIndexCopy_ScatterUpdate...` in all three Paged lanes. In the paired SP2 A/B profile, the old Paged path used 224 `ScatterUpdate` operations over three measured steps (9.381 ms/rank/step); the optimized path uses 0 operations and 0 ms while preserving the same FIA count and KV-write count. The replacement is contiguous row-major segment assembly, not deletion of logical output assembly.

#### NPU host timeline 

Profiler host time includes nested waits and must not be added to device time. The measured Free/synchronization differences are useful for attribution only:

| Topology | Dense Free (ms/rank/step) | Paged Free (ms/rank/step) | Delta | Dense -> Paged `StreamSynchronize` count / API ms |
|---|---:|---:|---:|---:|
| TP4 | 1.835 | 7.217 | +5.383 | 4.1 / 192.1 -> 11.4 / 281.5 |
| TP4/SP2 | 1.684 | 9.816 | +8.132 | 4.0 / 104.9 -> 11.4 / 159.1 |
| TP4/CFGP2 | 2.345 | 8.231 | +5.886 | 3.7 / 103.9 -> 10.4 / 153.7 |
#### NPU profile conclusion

Dense and Paged implement HunyuanImage3.0's mixed causal/full attention through different execution contracts. Dense materializes the attention mask and invokes one `FlashAttentionScore` for the complete sequence in each layer (96 calls over the three measured steps). Paged keeps KV in Scheduler-managed pages and invokes native `FusedInferAttentionScore` once per piecewise segment (96 calls on the first measured step and 64 on each later step, 224 calls in total).

For the measured shapes, the aggregate FIA device time is lower than the aggregate Dense FA time, so the observed E2E difference is not caused by the native attention kernel itself being slower. Paged removes Dense mask materialization and FA relayout, but adds one paged-KV cache write per layer per step, query/layout preparation, piecewise gather or slice/copy work, and more fine-grained host-visible segment launches. The homogeneous fast path removes the large output `ScatterUpdate`, but the remaining page and piecewise glue can still offset the FIA saving. 

## CC List

@hsliuustc0106 @asukaqaq-s @Bounty-hunter @Gaohan123 @zwhzzz0821 @fhfuih


vllm_omni/diffusion/diffusion_kv/config.py
```diff
@@ -42,3 +42,25 @@ def parse_diffusion_kv_cache_mode(value: object) -> DiffusionKVCacheMode:
 def is_scheduler_paged_kv_mode(mode: DiffusionKVCacheMode) -> bool:
     """Return whether a parsed cache mode uses Scheduler-owned paging."""
     return mode is DiffusionKVCacheMode.PAGED_SCHEDULER
+
+
+def validate_scheduler_paged_kv_step_execution(
```
Acerak01-fy: Is this over-defensive programming?

vllm_omni/diffusion/models/hunyuan_image3/hunyuan_image3_transformer.py
```diff
@@ -929,6 +935,18 @@ def __init__(
             paged_kv_cache_dtype=torch.bfloat16,
         )
 
+    def clear_legacy_prompt_kv_cache(self) -> None:
```
Acerak01-fy: Why do we need this

vllm_omni/diffusion/models/hunyuan_image3/pipeline_hunyuan_image3.py
```diff
@@ -900,12 +903,34 @@ def _restore_prompt_kv_cache(
             )
             mgr.image_kv_cache_lens = torch.cat(lens, dim=0).to(device=mgr.image_kv_cache_map[0].device)
 
+    def clear_legacy_prompt_kv_cache(self) -> None:
```
Acerak01-fy: Why is there code related to "profile" here?

Acerak01-fy (commented):
Try to simplify the code.

vllm_omni/diffusion/models/hunyuan_image3/pipeline_hunyuan_image3.py
```diff
@@ -475,8 +479,19 @@ def _validate_step_request(self, state: "StepRequestState") -> None:
         if sampling.num_outputs_per_prompt not in (None, 0, 1):
             raise ValueError("HunyuanImage3 step execution supports only one output per prompt.")
         parallel_config = self.od_config.parallel_config
-        if parallel_config.sequence_parallel_size > 1:
-            raise ValueError("HunyuanImage3 step execution does not support sequence parallel yet.")
+        if parallel_config.sequence_parallel_size > 1 and self._is_paged_kv_mode():
```
Acerak01-fy: Why do we still need this judgment when we're not supporting SP?

vllm_omni/platforms/npu/platform.py
```diff
@@ -77,8 +113,124 @@ def get_omni_generation_worker_cls(cls) -> str:
...
+    @classmethod
+    def configure_diffusion_vllm_config(cls, vllm_config: Any, od_config: Any) -> None:
+        """Use the block geometry required by Ascend's native paged kernel."""
+        if getattr(od_config, "diffusion_kv_mode", None) is None:
+            return
+        from vllm_omni.diffusion.diffusion_kv.config import DiffusionKVCacheMode
+
+        if od_config.diffusion_kv_mode is not DiffusionKVCacheMode.PAGED_SCHEDULER:
+            return
+        from vllm_ascend.attention.attention_v1 import AscendAttentionBackend
```
Acerak01-fy, line 121: Why is this particular writing style necessary?

vllm_omni/diffusion/data.py
```diff
@@ -1009,6 +1004,11 @@ def __post_init__(self):
             and self.diffusion_kv_max_rows_per_request is None
         ):
             raise ValueError("paged_scheduler requires diffusion_kv_max_rows_per_request to be set")
+        if self.diffusion_kv_mode is DiffusionKVCacheMode.PAGED_SCHEDULER and not self.step_execution:
```
Acerak01-fy: Why do we still need to verify `omni_config.py` when it has already been verified before?

vllm_omni/platforms/npu/platform.py
```diff
@@ -14,13 +15,127 @@
...
+    return (
+        selected.permute(0, 2, 1, 3)
+        .reshape(block_ids.numel(), cache.shape[2], block_size, head_size // last_dim, last_dim)
+        .permute(0, 1, 3, 2, 4)
+        .contiguous()
+        .reshape(block_ids.numel(), cache.shape[2] * head_size // last_dim, block_size, last_dim)
+    )
+
+
+def _pa_nz_to_logical_cache(
```
Acerak01-fy: Shouldn't this have been deleted already?

vllm_omni/diffusion/attention/backends/flash_attn.py
Acerak01-fy, line 277: Here, it needs to support both GPU and NPU. Will this conflict with the following forward_npu() method?

vllm_omni/diffusion/models/hunyuan_image3/pipeline_hunyuan_image3.py
```diff
@@ -485,6 +504,133 @@ def _validate_step_request(self, state: "StepRequestState") -> None:
...
+    def prepare_paged_attention_rows(
+        self,
+        states: list["StepRequestState"],
+    ) -> tuple[DiffusionPagedAttentionRow, ...]:
+        """Build native page-table rows in the exact DiT CFG batch order."""
+
+        if not self._is_paged_kv_mode():
+            raise RuntimeError("Hunyuan paged attention rows requested while diffusion_kv_mode is dense_legacy")
+        groups = self._split_step_groups(states)
+        if len(groups) != 1:
```
Gaohan123: [Blocking] This rejects a valid scheduler output whenever `max_num_seqs > 1` and a new request is admitted while another request is already running.

`BaseDiffusionScheduler.schedule()` can emit both `scheduled_new_reqs` and `scheduled_cached_reqs` in the same step. The model runner combines both sets into `states`, while `_step_group_key()` separates first-step requests from later-step requests. Therefore, a normal staggered arrival produces two groups here and fails before denoising.

Please either prepare and activate a separate paged-attention batch for each step group, or reject `paged_scheduler` with `max_num_seqs > 1` during configuration until multi-group execution is supported. A regression test with one cached request and one newly admitted request would also be needed.

vllm_omni/diffusion/models/hunyuan_image3/pipeline_hunyuan_image3.py
```diff
@@ -162,10 +163,13 @@ def _shift_full_attn_spans(
 
 
 def get_hunyuan_image_3_pre_process_func(od_config: OmniDiffusionConfig):
```
Gaohan123, line 163: [Blocking] This changes the existing HunyuanImage3 loading contract from an unconditional `trust_remote_code=True` to a configuration field whose default is `False`.

The HunyuanImage3 checkpoint uses custom Hugging Face configuration/model code, and the official loading example explicitly enables `trust_remote_code`. Most existing vLLM-Omni HunyuanImage3 recipes do not pass `--trust-remote-code`, because this integration previously enabled it internally. As a result, the documented default commands can now fail in `get_config()` before the pipeline starts.

Please either preserve the existing model-specific behavior, set the Hunyuan deploy/registry default explicitly, or update all affected recipes and add startup contract tests if requiring an explicit flag is an intentional breaking security change. Since this is unrelated to paged KV support, it would also be cleaner to handle it in a separate PR.

vllm_omni/diffusion/models/hunyuan_image3/pipeline_hunyuan_image3.py
```diff
@@ -347,7 +354,7 @@ class HunyuanImage3Pipeline(
     ]
 
     def __init__(self, od_config: OmniDiffusionConfig) -> None:
-        self.hf_config = get_config(od_config.model, trust_remote_code=True)
+        self.hf_config = get_config(od_config.model, trust_remote_code=od_config.trust_remote_code)
```
Gaohan123: This has the same compatibility issue as the preprocessing path above: the existing HunyuanImage3 recipes do not consistently set `trust_remote_code=True`, so the pipeline can now fail during initialization. Please keep both loading sites aligned with the intended and documented model-loading contract.

tools/pre_commit/check_spdx_header.py
```diff
@@ -3,21 +3,17 @@
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM-Omni project
 """Require SPDX license headers on source files.
```
Gaohan123, line 5: The modification in this file is unrelated to the feature

vllm_omni/platforms/npu/platform.py
```diff
@@ -14,13 +15,131 @@
...
+)
+
 _DIFFUSION_PACKED_MODULES_MAPPING = {
     "HunyuanImage3Pipeline": {
         "experts": ["experts.0.gate_up_proj", "experts.0.down_proj"],
     },
 }
 
 
+def _logical_cache_to_pa_nz(
```
Gaohan123: Please move tool functions into utils

vllm_omni/diffusion/models/hunyuan_image3/request_layout.py
```diff
@@ -200,6 +200,53 @@ def hunyuan_num_image_tokens(image_info: ImageInfo) -> int:
     return int(image_info.image_token_length) + int(image_info.add_timestep_token) + int(image_info.add_guidance_token)
 
 
+def _round_up_for_sequence_parallel(value: int, sequence_parallel_size: int) -> int:
+    if type(sequence_parallel_size) is not int or sequence_parallel_size <= 0:
+        raise ValueError(f"sequence_parallel_size must be a positive integer, got {sequence_parallel_size!r}")
+    return ((value + sequence_parallel_size - 1) // sequence_parallel_size) * sequence_parallel_size
+
+
+def hunyuan_paged_sequence_lengths(
```
Gaohan123: Wrap the processing logic out of model specific location

Gaohan123 (commented):
Why is the peak memory of paged version larger than original version?

Acerak01-fy:
> Why is the peak memory of paged version larger than original version?

The higher memory number is expected for this benchmark, but it is important to distinguish pool capacity from request-live usage.

  Paged mode preallocates a native physical KV block pool during Worker initialization. Our `peak_memory_mb_worker` metric is based on `max_memory_reserved()`, so it includes the preallocated KV pool, allocator high-water
  mark, BlockTables/metadata, and native attention workspace. The dense path has no persistent page pool and only materializes request-scoped dense KV, so a single request can show a lower reserved-memory value.

  Free pages are logically available for future requests and prefix reuse, but they are still physically reserved and must be included in device capacity planning.

  The benefit of paging is predictable allocation and reuse under concurrent or multi-request workloads, And also reduce the internal fragmentation of the memory.

vllm_omni/platforms/npu/platform.py
```diff
@@ -77,8 +196,142 @@ def get_omni_generation_worker_cls(cls) -> str:
...
+
+            touched_blocks = sorted({slot // kernel_block_size for slot in global_slots if slot >= 0})
+            compact_block_index = {block_id: index for index, block_id in enumerate(touched_blocks)}
+            local_slots = [
+                -1
+                if slot < 0
+                else compact_block_index[slot // kernel_block_size] * kernel_block_size + slot % kernel_block_size
+                for slot in global_slots
+            ]
+            plan = DiffusionPagedKVWritePlan(
```
RuixiangMa: These tensors are consumed at index_copy_ , PyTorch's index_copy_ requires int64 (long) index tensors on CUDA/NPU backends. Using int32 will raise RuntimeError.

vllm_omni/diffusion/models/hunyuan_image3/pipeline_hunyuan_image3.py
```diff
@@ -704,17 +902,103 @@ def _merge_step_model_inputs(
...
 
+        if align_paged_sp_rows:
+            assert paged_row_spans is not None
+            logical_query_lens, _, _, paged_prefix_lens, paged_valid_lens = paged_row_spans
+            aligned_prefix_len = max(paged_prefix_lens)
+            physical_query_len = aligned_prefix_len + max(
+                query_len - prefix_len
+                for query_len, prefix_len in zip(logical_query_lens, paged_prefix_lens, strict=True)
+            )
+            for key in ("position_ids", "image_mask"):
```
RuixiangMa: Surrounding code accesses model_kwargs["image_mask"] directly. If image_mask is not in model_kwargs (e.g., a model variant that doesn't emit it, or a merge path that skips it), this raises KeyError. If the merged value is None, image_mask.bool() raises AttributeError.

vllm_omni/diffusion/models/hunyuan_image3/hunyuan_image3_transformer.py
```diff
@@ -1069,6 +1088,181 @@ def _build_neg_ar_kv(
...
+        if self._injected_ar_kv is not None:
+            raise NotImplementedError(
+                "Hunyuan Scheduler paged KV does not support AR-stage KV injection; "
+                "the native cache must be populated by the DiT write itself."
+            )
+
+        batch_size = len(paged_query_lens)
+        if batch_size == 0:
+            raise ValueError("Paged Hunyuan attention received an empty batch")
+        if query_lens is None:
```
RuixiangMa: All tests pass query_lens explicitly. This path is unverified — max(paged_query_lens) could produce incorrect tensor shapes if the model expects a padded length.

vllm_omni/diffusion/models/hunyuan_image3/pipeline_hunyuan_image3.py
```diff
@@ -704,17 +902,103 @@ def _merge_step_model_inputs(
...
+                packed_indices.extend(range(row_offset, row_offset + prefix_len))
+                packed_indices.extend(
+                    range(
+                        row_offset + aligned_prefix_len,
+                        row_offset + aligned_prefix_len + current_len,
+                    )
+                )
+            identity_indices = list(range(len(logical_query_lens) * physical_query_len))
+            if packed_indices != identity_indices:
+                reference = merged.get("position_ids")
```
RuixiangMa: If position_ids is not a tensor (or not in merged), device is None, creating indices on CPU. Later restore_output calls restored.index_copy_(0, self.output_scatter_indices, output) where output may be on NPU, causing a device mismatch error.

vllm_omni/diffusion/attention/backends/utils/piecewise_attn.py
```diff
@@ -145,7 +145,7 @@ class PagedPiecewisePlan:
 
 
 PagedPiecewiseRunner = Callable[
```
hsliuustc0106, line 159: [P1] Same shared-seam collision as PR #6658 — agree a contract before either merges

#6658 redefines this alias to `[Q, K, V, meta, output_buffer | None]` with a contiguous view fast path, incompatible with this PR's `[Q, K|None, V|None, meta]` + `read_kv_from_cache`. Overlapping-region conflicts also land in `flash_attn.py::_forward_impl`'s paged branch and `paged_attention_adapter.py` (platform write plans + `paged_query_indices` + output scatter here vs. output buffer + native-metadata snapshot cloning there), plus two parallel Hunyuan integrations (runner-driven `prepare_paged_attention_rows()` vs. model-forward `_paged_attention_context`) and different `get_kv_cache_spec` name-canonicalization rules. Whichever PR lands second must semantically re-port its Hunyuan path rather than textually rebasing.

Suggest sequencing these two and extracting one seam contract first: runner protocol, row-builder ownership, and the plan/metadata extension point.

tools/pre_commit/check_spdx_header.py
hsliuustc0106: Agreed it's unrelated — and it points against current main's direction: the checker today deliberately treats the upstream vLLM line as stale and rewrites it to the vLLM-Omni line (that's why recently-touched files across the tree get converted — e.g. #6658 does exactly that). Accepting both lines here flips that policy inside a feature PR, and whichever of the two paged-KV PRs lands second churns the other's headers back. Suggest dropping this tool change from this PR entirely; new files already carry Omni headers, so nothing in the NPU work needs it.

vllm_omni/platforms/npu/platform.py
```diff
@@ -77,8 +196,142 @@ def get_omni_generation_worker_cls(cls) -> str:
...
+
+        op = getattr(getattr(torch.ops, "npu", None), "npu_scatter_pa_kv_cache", None)
+        schema = getattr(getattr(op, "default", None), "_schema", None)
+        if schema is None or "cache_mode" in str(schema):
+            return
+        device_operator = device_op.DeviceOperator
+        if getattr(device_operator, "_omni_cache_mode_compat", False):
+            return
+
+        device_operator.reshape_and_cache = classmethod(_reshape_and_cache_without_cache_mode)
```
hsliuustc0106: [P3] Confirm intended process-wide scope of this compat shim

Once installed, `DeviceOperator.reshape_and_cache` is replaced for every caller in the process (guarded by `_omni_cache_mode_compat`), and the 4-D shape gate above also matches standard `[T,H,D]/[N,B,H,D]` cache layouts, so non-diffusion scatter traffic sharing the worker would route through the PA_NZ roundtrip too. Benign today because diffusion workers own their process, but a one-line comment stating that assumption — or gating activation on the Scheduler-paged writer being active — would keep a future colocated stage from silently inheriting the roundtrip.

vllm_omni/diffusion/models/hunyuan_image3/pipeline_hunyuan_image3.py
Acerak01-fy: Fixed. I reject paged_scheduler with max_num_seqs > 1.

vllm_omni/diffusion/models/hunyuan_image3/pipeline_hunyuan_image3.py
Acerak01-fy: Fixed.

tools/pre_commit/check_spdx_header.py
Acerak01-fy: OK,I have deleted it.

vllm_omni/platforms/npu/platform.py
Acerak01-fy: Migration completed.

vllm_omni/diffusion/models/hunyuan_image3/request_layout.py
Acerak01-fy: For this part, I haven't made any changes yet because it involves a lot of unique semantics specific to the hunyuanimage model. Only a small portion of public organizations can be isolated.

vllm_omni/diffusion/attention/backends/flash_attn.py
```diff
@@ -23,6 +24,31 @@
 logger = init_logger(__name__)
 
 
+def _load_mindiesd_attention(api_name: str):
```
hsliuustc0106:
```suggestion
def _load_npu_attention(api_name: str):
```

vllm_omni/diffusion/models/hunyuan_image3/hunyuan_image3_transformer.py
```diff
@@ -883,8 +883,15 @@ def __call__(
 
 
 class ImageKVCacheManager(nn.Module):
```
hsliuustc0106: this class seems not HY-Image only, we may refactor it later

vllm_omni/diffusion/models/hunyuan_image3/pipeline_hunyuan_image3.py
```diff
@@ -485,6 +503,125 @@ def _validate_step_request(self, state: "StepRequestState") -> None:
         if getattr(self.od_config, "diffusion_kv_cache_skip_step_indices", None):
             raise ValueError("HunyuanImage3 step execution does not support diffusion KV cache step skips yet.")
 
+    def _is_paged_kv_mode(self) -> bool:
```
hsliuustc0106: many of the newly added functions are not model specific, we may extract a mixin later for better abstraction

vllm_omni/platforms/npu/platform.py
Acerak01-fy: Thanks for flagging this. I verified the behavior directly on an Ascend 910B1 with torch_npu 2.10.0rc2. Both int32 and int64 indices work for index_select and index_copy_ on the NPU.

  Also, only block_ids is consumed by index_select/index_copy_. local_slot_mapping is passed to npu_scatter_pa_kv_cache instead. The existing NPU E2E runs also exercise this write-back path successfully.

  I am therefore keeping block_ids as int32 to avoid adding a per-layer dtype conversion.

vllm_omni/diffusion/models/hunyuan_image3/pipeline_hunyuan_image3.py
Acerak01-fy: Thanks. image_mask is a required input for the supported HunyuanImage3 first-step gen_image path, so a model variant without image_mask is outside this model-specific integration contract.

  However, the previous isinstance check made the requirement look optional and could defer the failure. I have changed the SP alignment path to explicitly require both image_mask and position_ids to be tensors and raise a clear ValueError before alignment. A regression test now covers missing metadata.

vllm_omni/diffusion/models/hunyuan_image3/hunyuan_image3_transformer.py
Acerak01-fy: Agreed. The fallback was not valid for sequence parallel execution: paged_query_lens describes the global logical write length, while query_lens describes the local physical tensor length. Using max(paged_query_lens) can therefore produce an incorrect padded shape under SP.

  The production pipeline always supplies query_lens, so I removed the fallback and now raise a clear ValueError when the metadata is missing. I also added a regression test for this case.

vllm_omni/diffusion/models/hunyuan_image3/pipeline_hunyuan_image3.py
Acerak01-fy: Agreed. position_ids is required by the supported HunyuanImage3 paged Ulysses first-step path, so silently falling back to CPU indices was inappropriate.The code now validates that position_ids is a tensor before alignment and creates paged_query_indices directly on position_ids.device. The test also verifies that the generated indices remain on the same device.

vllm_omni/diffusion/models/hunyuan_image3/pipeline_hunyuan_image3.py
Acerak01-fy: Fixed.

vllm_omni/diffusion/models/hunyuan_image3/request_layout.py
Gaohan123: OK. We need a design in following PRs

Bounty-hunter:
It only support in `step_execution=True`?

Bounty-hunter:
It only support in `step_execution=True`?

vllm_omni/diffusion/attention/backends/utils/piecewise_attn.py
hsliuustc0106: Following up with the concrete contract worth agreeing on — and this PR's latest push makes the N-model framing real rather than hypothetical: the `max_num_seqs=1` gate explicitly defers the multi-request population (~11 step-execution pipelines become KV-reuse candidates as continuous batching and admission mature).

Proposal: everything outside `models/<name>/` stays model-blind (scheduler allocation, BlockTables, piecewise planning, platform write-plan hooks already are). Per-model cost shrinks to one row-builder file (CFG expansion, prefix source, write-span policy, SP padding rules, capability guards) behind one typed runner-invoked hook — this PR's `prepare_paged_attention_rows(states) -> tuple[DiffusionPagedAttentionRow, ...]` shape.

Also worth noting: this PR's K/V-None read-from-cache mode and #6658's output-buffer fast path are orthogonal — a single signature `[Q, K|None, V|None, meta, out|None]` accommodates both, so neither has to lose. Row-builder ownership is the actual decision, and the runner-driven shape here is the more general seam.

Practical path: if this lane lands first, make #6658 port onto it as a review condition (that port doubles as the generalization proof); then capture an "adopting scheduler-paged KV" doc + support-matrix entry so HiDream's author copies a doc instead of archaeology across two PRs.

vllm_omni/platforms/npu/platform.py
hsliuustc0106: Re-anchor note at bf871c31: this code moved to `platforms/npu/utils.py:61` (install site now `platform.py:114`) per the placement feedback — relocation reads clean. The scope point stands: the install still replaces `reshape_and_cache` process-wide, so a comment stating the diffusion-worker-process assumption (or gating activation on an active Scheduler-paged writer) would keep future colocated stages from silently inheriting the PA_NZ roundtrip.

vllm_omni/diffusion/attention/backends/utils/piecewise_attn.py
Acerak01-fy: I have now cherry-picked and re-based the GPU-compatible PR #6658. I have removed the redundant code for both NPU and GPU compatibility and unified the common paths. Merge the end-to-end adaptation of GPU and NPUs into this  PR, which significantly reduces the code volume.

Acerak01-fy:
> It only support in `step_execution=True`?

Now, it's only in `request_execution`.

vllm_omni/diffusion/models/hunyuan_image3/pipeline_hunyuan_image3.py
```diff
@@ -1829,18 +1956,42 @@ def _step_group_key(self, state: "StepRequestState") -> tuple[Any, ...]:
...
+            attention_mask = model_kwargs.get("attention_mask")
+            if not isinstance(attention_mask, torch.Tensor) or attention_mask.ndim != 4:
+                raise ValueError(f"Missing Hunyuan attention mask for paged request {state.request_id}.")
+            query_len = int(attention_mask.shape[-2])
+            seq_len = int(attention_mask.shape[-1])
+            full_attn_spans = model_kwargs.get("full_attn_spans")
+            segment_modes = None
+            if isinstance(full_attn_spans, list):
+                segment_modes = tuple(
+                    tuple(segment.mode for segment in build_segments(row_spans, 0, query_len))
```
Bounty-hunter: Why does the model/pipeline need to be aware of segment information here? I would expect the model to only pass full_attn_spans through, and let the attention backend build and validate the concrete segments.

vllm_omni/diffusion/worker/diffusion_model_runner.py
```diff
@@ -420,12 +397,26 @@ def get_kv_cache_spec(self) -> dict[str, KVCacheSpec]:
             raise RuntimeError("Model must be loaded before collecting Diffusion KV cache specs")
 
         cache_layers: dict[str, tuple[Attention, KVCacheSpec]] = {}
-        for layer_name, module in self.pipeline.named_modules():
+        cache_layer_paths: dict[str, str] = {}
```
Bounty-hunter: over-defensive? If we want to add this check, using cache_layers directly to detect duplicates should be enough.

Bounty-hunter (commented):
We need to collect profiling data to understand the root cause of the key performance changes:

1. Whether the attention operator itself shows any performance difference between page input and dense tensor input.
2. Whether the data preparation before attention is more or less expensive under page input. In theory, page input may introduce more read/write operations, while dense input may require an extra concat operation.

vllm_omni/platforms/npu/utils.py
```diff
@@ -0,0 +1,126 @@
...
+
+import torch
+
+_diffusion_paged_kv_write_plan: ContextVar[Any | None] = ContextVar(
+    "diffusion_paged_kv_write_plan",
+    default=None,
+)
+
+
+def _logical_cache_to_pa_nz(
```
Bounty-hunter: (1) Are you writing into the block and then reading it back and converting to pa_nz? If pa_nz is needed, shouldn't you write directly in pa_nz format at cache-write time?
(3) Is the pa_nz format strictly required, or would N,b,h,d be sufficient?

zwhzzz0821:
## Dense vs Paged performance and profile results

The Dense/Paged comparison was run on four Hopper GPUs with FlashAttention 3
enabled for both modes.

Test configuration:

- Model: `tencent/HunyuanImage-3.0-Instruct`
- Resolution: 1024 x 1024
- Denoising steps: 8
- One warmup request followed by 30 measured requests
- Request concurrency: 1
- Attention backend: FA3 for both Dense and Paged
- Topologies: TP4, TP2+SP2 (Ulysses), and TP2+CFGP2
- Every case completed 30/30 requests, with 0 failures

### E2E result

Negative deltas mean Paged is faster than Dense.

| Topology | Dense mean latency | Paged mean latency | Paged vs Dense | QPS, Dense -> Paged | P99, Dense -> Paged |
| TP4 | 2.2778 s | 2.2898 s | +0.53% | 0.4390 -> 0.4367 | 2.4113 -> 2.3662 s (-1.87%) |
| TP2+SP2 | 2.3946 s | 2.3577 s | -1.54% | 0.4176 -> 0.4241 | 2.6732 -> 2.4764 s (-7.36%) |
| TP2+CFGP2 | 2.1745 s | 2.1937 s | +0.88% | 0.4598 -> 0.4558 | 2.2574 -> 2.2912 s (+1.50%) |

The two regressions are below 1%, while TP2+SP2 is 1.54% faster with Paged
KV. There is no material E2E performance regression in the tested
configurations.

The pipeline profiler shows an expected local attention cost:

| Topology | Dense layer-0 self-attention | Paged layer-0 self-attention | Paged vs Dense |
|---|---:|---:|---:|
| TP4 | 2.2918 s | 2.2773 s | -0.63% |
| TP2+SP2 | 21.855 ms | 22.807 ms | +4.36% |
| TP2+CFGP2 | 2.1925 s | 2.1924 s | 0.00% |

This local difference is diluted by the rest of the pipeline and partially
offset by work that the Paged path removes.

### Pixel accuracy

Test target:

```text
pytest -o addopts= -s -v tests/e2e/accuracy/test_hunyuan_image3_pixel_accuracy.py
```

The deployment used `diffusion_kv_mode=paged_scheduler`, TP4,
`diffusion_kv_max_rows_per_request=2`, an 8-GiB KV-cache budget, and the same
1024 x 1024 text-to-image baseline for both execution paths.

| Execution path | Mean absolute difference (threshold) | P99 absolute difference (threshold) | SSIM (threshold) | PSNR (threshold) | Result |
| Online | 0.009110 (<= 0.030000) | 0.090196 (<= 0.300000) | 0.981924 (>= 0.970000) | 33.826607 dB (>= 30.000000 dB) | Passed |
| Offline | 0.009110 (<= 0.030000) | 0.090196 (<= 0.300000) | 0.981924 (>= 0.970000) | 33.826607 dB (>= 30.000000 dB) | Passed |

### Nsight Systems profile

I profiled TP2+CFGP2 because it has the largest repeatable E2E degradation.
Dense and Paged used the same FA3 binary and the same four GPUs. The profile
used two warmup requests followed by five captured requests, and both modes
completed 5/5.

The values below were obtained from CUDA kernels selected by exact NVTX ranges
and SQL across all four ranks. They are normalized by
`5 requests * 4 ranks`, so they represent kernel time per request per rank.
They must not be interpreted as a four-rank E2E wall-clock sum.

| CUDA operation | Exact kernel selector | Dense calls | Paged calls | Dense (ms/request/rank) | Paged (ms/request/rank) | Paged - Dense |
|---|---|---:|---:|---:|---:|---:|
| FA3 forward | `FlashAttnFwdSm90 (all forward variants)` | 10,880 | 10,880 | 55.828 | 64.922 | +9.093 |
| Write current K/V to page slots | `vllm::reshape_and_cache_flash_kernel` | 0 | 5,120 | 0.000 | 5.540 | +5.540 |
| Concat cached prompt K/V with current K/V | `CatArrayBatchedCopy (4-D) + vectorized cat gridX=2049` | 8,960 | 0 | 4.497 | 0.000 | -4.497 |
| Concat piecewise attention outputs | `CatArrayBatchedCopy_vectorized gridX=4224` | 5,120 | 0 | 2.604 | 0.000 | -2.604 |
| Combine FA3 Split-KV partial outputs | `flash::FlashAttnFwdCombine` | 0 | 640 | 0.000 | 0.148 | +0.148 |
| Prepare FA3 variable-length block schedule | `flash::prepare_varlen_num_blocks_kernel` | 0 | 140 | 0.000 | 0.013 | +0.013 |
| FA3/scheduler integer fills | `at::native::FillFunctor<int>` | 1,280 | 10,380 | 0.051 | 0.415 | +0.364 |


#### Attention operator

The FA3 main-kernel call count is identical:

```text
Dense: 11,520 calls
Paged: 11,520 calls
```

Therefore, the regression is not caused by calling attention more times.

The average main-kernel duration changes from 101.701 us for Dense to
118.268 us for Paged, an increase of 16.29%. This is the largest measured
cost. Dense FA3 reads one contiguous K/V tensor. Paged FA3 resolves K/V through
the block table and reads non-contiguous cache pages. The page-table
indirection and less-contiguous memory access make the Paged FA3 kernel slower.

#### Data preparation before attention

Dense spends 4.617 ms/request/rank concatenating prompt K/V and current K/V
before attention.

Paged removes that concat, but it performs the following work instead:

| Paged preparation item | Kernel time (ms/request/rank) |
|---|---:|
| Write current K/V into scheduler-assigned page slots | 5.191 |
| Input contiguity copy in `prepare_layer_context` | 2.013 |
| Piecewise metadata kernels | 0.047 |
| **Total** | **7.251** |

After accounting for the Dense concat that Paged removes, the net preparation
overhead is:

```text
7.251 - 4.617 = 2.634 ms/request/rank
```

The page write is required because the newly produced K/V must be placed into
the scheduler-assigned physical slots before paged attention can address it
through the block table. Dense also produces current K/V, but it can use the
result directly after concatenation and does not need to scatter it into the
paged cache layout.

#### Work removed by the Paged path

Dense spends 2.731 ms/request/rank concatenating piecewise attention outputs.
Paged writes each piece directly into the corresponding slice of the final
output buffer. Therefore, the profiled Paged output concat/scatter range has
zero kernel calls, saving the full 2.731 ms/request/rank.

This saving partially offsets:

- the +9.543 ms/request/rank FA3 main-kernel cost;
- the +2.634 ms/request/rank net preparation cost;
- 0.178 ms/request/rank of FA3 combine work; and
- 0.357 ms/request/rank of additional auxiliary fill work.

### Conclusion

The profile shows that the remaining Dense/Paged difference is not caused by
extra attention invocations. It comes primarily from:

1. the Paged FA3 kernel reading K/V through block-table indirection and
   non-contiguous pages; and
2. writing current K/V into scheduler-assigned page slots, plus an input
   contiguity copy.

Paged removes the Dense prompt/current-KV concat and the piecewise output
concat. The direct output-buffer write recovers 2.731 ms/request/rank, which
is why an observable operator-level regression becomes only a sub-1% E2E
difference in TP4 and TP2+CFGP2, and is fully offset in TP2+SP2.

The 30-request E2E numbers above are the performance source of truth. The
Nsight capture is used for root-cause attribution of the same FA3 Paged KV
data path; its per-kernel times must not be added directly to E2E wall time.

vllm_omni/diffusion/worker/diffusion_model_runner.py
zwhzzz0821: fixed

vllm_omni/diffusion/models/hunyuan_image3/pipeline_hunyuan_image3.py
zwhzzz0821: the pipeline now only passes full_attn_spans, while the attention backend builds and validates the concrete segments.

vllm_omni/diffusion/models/hunyuan_image3/hunyuan_image3_transformer.py
```diff
@@ -1,5 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
-# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM-Omni project
```
Gaohan123, line 2: Could we move the logic in the file into diffusion modelrunner? As it is framework logic rather than model specific logic

vllm_omni/diffusion/worker/diffusion_model_runner.py
Gaohan123, line 862: [P1] Clear model-owned legacy prompt KV after the profiling run

The startup profile runs before the physical paged cache and Worker adapter are initialized, so the profiling scope intentionally falls back to the dense attention path. For Hunyuan, the first denoising step stores prompt K/V tensors in each attention module's `image_kv_cache_map` and `image_kv_cache_lens`.

This cleanup only drops runner-owned state. The model still holds references to those tensors, so `gc.collect()` cannot release them. They are not cleared until the first real paged request reaches `_forward_paged()`.

As a result, automatic cache sizing counts temporary dense prompt KV as non-KV memory and allocates a permanently smaller page pool. With an explicit `kv_cache_memory_bytes`, the legacy tensors can remain resident while the fixed page pool is allocated, increasing the initialization peak and potentially causing an OOM.

Please explicitly clear the legacy prompt KV from all Hunyuan attention modules after profiling and before the profiling result is used to size or initialize the physical cache. Could we also add a regression test asserting that `image_kv_cache_map` and `image_kv_cache_lens` are both `None` after `profile_run()`?

vllm_omni/diffusion/models/hunyuan_image3/hunyuan_image3_transformer.py
zwhzzz0821: Since the dense mode is still supported, the model side still needs to determine whether to use dense or paged mode and perform the corresponding operations accordingly. Therefore, this part of the code is still necessary. We have also tried to move as much of the logic as possible into the framework.

Bounty-hunter:
@zwhzzz0821 show the accuracy test result for gpu

zwhzzz0821:
> @zwhzzz0821 show the accuracy test result for gpu

it listed in the report above

Bounty-hunter (approved):
LGTM

Gaohan123 (approved):
LGTM. Thanks

PR #6565, opened 2026-08-24
[Diffusion][LoRA] Validate per-module rank against PEFT metadata in the generic loader

LeslieWylie:
## Summary

Follow-up to my review note on #6476: the **generic** diffusion PEFT fallback can silently mis-scale adapters whose per-module rank does not match the single global `r`.

`DiffusionLoRAManager._load_adapter`'s generic path (`PEFTHelper.from_local_dir` + `LoRAModel.from_local_checkpoint`) derives one global `r`/`lora_alpha` from `adapter_config.json` and applies a uniform `scaling = lora_alpha / r` to every module (`LoRALayerWeights.from_config`). It never re-derives the per-module rank from the actual A/B tensors, and `vllm.lora.peft_helper.PEFTHelper` has no `rank_pattern`/`alpha_pattern` fields (they are silently dropped in `from_dict`).

So a checkpoint whose physical modules do not all share `r` — e.g. a fused-QKV adapter whose packed projection has rank `3*r` while the rest keep `r` — is loaded at the wrong rank and silently mis-scaled, with no error surfaced.

The model-specific path added in #6476 (`_load_diffusion_lora_adapter`) already handles the published H3 Turbo format correctly. This PR only hardens the **generic fallback**; model-specific loaders are unaffected.

## Change

`validate_generic_peft_ranks` in `vllm_omni/diffusion/lora/utils.py`, invoked only on the generic fallback branch. It derives the actual per-module rank from the A/B tensor shapes and:

- rejects a declared `rank_pattern`/`alpha_pattern` the generic loader cannot honor,
- rejects modules whose `lora_A`/`lora_B` ranks are mutually inconsistent (corrupt or packed/fused layout),
- rejects any module whose tensor rank disagrees with the global `r`,

each with an actionable error pointing at a model-specific loader or a canonical single-rank checkpoint.

It deliberately does **not** try to recover a per-module scale (`lora_alpha / actual_rank`): some distilled checkpoints bake the effective scale into the A/B tensors, so silently reinterpreting it could double-apply or drop it. Rejecting is the safe generic behavior; a model that understands its own contract supplies a loader.

## Tests

`tests/diffusion/lora/test_peft_rank_validation.py` — 9 tests, CPU, synthetic tensors (no real adapter):

- uniform rank == global `r` → accepted
- declared `rank_pattern` / `alpha_pattern` → rejected
- mixed physical rank, no pattern → actionable error naming the offending module
- `lora_A`/`lora_B` rank inconsistency → rejected
- packed per-slice ranks → accepted / rejected
- config reader detects declared patterns and tolerates a missing/unreadable config

Local run: 9/9 new pass, plus existing `test_lora_manager.py` / `test_base_linear.py` still pass.

## Field note

Exercised in a production deployment on Blackwell (SM120) with online FP8 (DiT + text encoder, BF16 VAE), tensor parallelism (TP2/TP4), and offline-merged Turbo weights. The generic-vs-model-specific split and this guard behave as expected there.


LeslieWylie:
Self-review completed.

I re-checked the PR specifically for scope, correctness, and regression risk:

- The validation is only invoked on the generic diffusion LoRA fallback path (`loaded is None`); model-specific loaders remain untouched.
- Per-module rank is derived from the actual LoRA A/B tensor shapes rather than inferred from the global PEFT `r`.
- A/B rank inconsistencies and mixed ranks that cannot be represented by the currently supported global PEFT metadata fail with actionable errors instead of silently selecting a scale.
- `rank_pattern` / `alpha_pattern` are intentionally rejected with guidance because the current `PEFTHelper` path does not preserve enough metadata to honor them correctly; this PR does not attempt to extend PEFT semantics.
- The implementation does not guess `alpha / actual_rank`, since converted/fused checkpoints may already encode effective scaling in their tensors.
- Model-owned packed/fused loaders, including the MiniMax-H3 work in #6476, are not affected.
- Tests use synthetic CPU tensors only and cover homogeneous ranks, mixed ranks, malformed A/B pairs, and unsupported PEFT patterns.
- Existing targeted LoRA/manager tests plus the new tests pass locally (`30 passed`).
- I also checked the diff for unrelated changes and internal deployment information; none is included.

The intended invariant is: the generic loader must not silently reinterpret ambiguous mixed-rank PEFT checkpoints.

LeslieWylie:
@vllm-omni-review-bot ready for review. I fixed the reported `ruff format` and `typos` failures and amended the commit with the required DCO sign-off; DCO is now green. The code change remains limited to the generic diffusion LoRA fallback and its CPU regression tests.

LeslieWylie:
Superseded by clean-history replacement #6994. An amend from a shallow checkout accidentally removed this branch's base ancestry, so GitHub closed the PR and would not reopen it after the ancestry was repaired. The replacement preserves the same scoped code change, tests, and DCO sign-off.

PR #6571, opened 2026-08-24
[Bugfix][Ascend NPU][Boogu] Replace complex RoPE with real-valued operations

JW-L-7:
## Summary

Replace the complex-valued rotary positional embedding path in Boogu-Image
with mathematically equivalent real-valued cosine and sine operations.

This prevents Ascend NPU inference from passing complex tensors to
`aclnnGather`, while preserving the behavior of the previous implementation
on CUDA and other supported platforms.

## Motivation

Boogu-Image previously generated complex-valued RoPE frequency tables and
gathered them using `torch.gather`.

On Ascend NPU, model execution failed because `aclnnGather` does not support
`DT_COMPLEX128` input tensors:

```text
Tensor self not implemented for DT_COMPLEX128
NPU function error: call aclnnGather failed
error code is 161002
```

The failure occurred in the Boogu rotary-position embedding path before the
transformer could complete inference.

## Changes

- Replace complex-valued RoPE frequency tables with separate real-valued
  cosine and sine tensors.
- Generate RoPE tables using:
  - `use_real=True`
  - `repeat_interleave_real=True`
- Replace complex multiplication based on `view_as_complex` and
  `view_as_real` with its equivalent real-valued formulation:

  ```text
  real = x_even * cos - x_odd * sin
  imag = x_even * sin + x_odd * cos
  ```

- Gather cosine and sine tensors independently, avoiding complex inputs to
  `torch.gather`.
- Propagate the real-valued RoPE representation through:
  - instruction refinement
  - image refinement
  - double-stream transformer blocks
  - single-stream transformer blocks
  - text-to-image and image-editing pipeline paths
- Use `float32` frequency tables on Ascend NPU and platforms without float64
  support.
- Preserve the previous higher-precision frequency-table behavior on
  platforms that support float64.
- Add regression tests for the real-valued RoPE implementation and its
  integration with the Boogu pipeline.

## Accuracy and Compatibility

The real-valued formulation is mathematically equivalent to the previous
complex multiplication:

```text
(a + ib)(cos + i sin)
  = (a * cos - b * sin) + i(a * sin + b * cos)
```

The implementations are not expected to be bitwise identical in every dtype
because decomposing complex multiplication into real-valued operations may
change floating-point rounding. In addition, Ascend NPU intentionally uses
float32 frequency tables to avoid unsupported and less broadly available
float64 and complex kernels.

A numerical comparison covering 3,932,160 RoPE output elements produced:

| Output dtype | Maximum absolute difference | Mean absolute difference | Different elements |
|---|---:|---:|---:|
| FP32 | 4.77e-7 | 2.26e-8 | 1,493,344 |
| FP16 | 0.001953125 | 4.23e-8 | 718 |
| BF16 | 0.0078125 | 1.43e-8 | 74 |

For FP16 and BF16, the maximum observed difference is approximately one
quantization step, and more than 99.98% of the output elements remain
bitwise identical.

Regression tests also compare the new implementation directly against the
previous complex-valued reference using `torch.testing.assert_close`.

The change does not modify:

- model weights
- checkpoint formats
- external serving APIs
- input or output tensor layouts
- the output dtype of the rotary embedding operation

For platform compatibility:

- Ascend NPU uses real-valued float32 frequency tables, avoiding unsupported
  complex gather operations.
- Platforms with float64 support preserve the previous higher-precision
  frequency-table generation behavior.
- The new path consists only of real-valued gather, slicing, multiplication,
  addition, subtraction, and stacking operations.
- The implementation no longer depends on complex gather support and uses
  standard real tensor operations available across the existing CUDA, ROCm,
  MUSA, XPU, and Ascend backend stacks.
- Ascend NPU and NVIDIA CUDA end-to-end inference were successfully validated.

### Unit tests

The modified Boogu transformer and pipeline test suites passed:

```text
61 passed
```

The test coverage includes:

- numerical equivalence with the previous complex-valued RoPE implementation
- cosine and sine frequency-table shapes
- repeated adjacent-channel frequency layout
- absence of complex tensors in the new frequency tables
- float32 frequency tables on Ascend NPU
- T2I transformer forward behavior
- TI2I transformer forward behavior
- propagation of real-valued frequencies through the Boogu pipeline

### Ascend NPU end-to-end validation

- The modified Boogu model loaded successfully.
- The inference service started successfully.
- Text-to-image inference completed successfully.
- The generated image was returned normally.
- The previous `DT_COMPLEX128` `aclnnGather` error no longer occurred.

### NVIDIA CUDA end-to-end validation

- The modified source was installed and loaded successfully.
- The Boogu model loaded successfully.
- The inference service started successfully.
- A 512x512 text-to-image request using 28 inference steps completed
  successfully.
- The response contained a valid PNG image.

### NVIDIA CUDA performance validation

All benchmark requests completed successfully with zero failures.

| Concurrency | Completed | Failed | Throughput | Mean latency | P50 | P99 | Peak memory |
|---:|---:|---:|---:|---:|---:|---:|---:|
| 1 | 10 | 0 | 0.1397 QPS | 7.157 s | 7.159 s | 7.228 s | 36,744 MB |
| 2 | 10 | 0 | 0.1405 QPS | 13.532 s | 14.239 s | 14.336 s | 36,744 MB |
| 4 | 10 | 0 | 0.1400 QPS | 24.310 s | 28.470 s | 28.700 s | 36,744 MB |

Benchmark configuration:

- Task: text-to-image
- Resolution: 512x512
- Inference steps: 28
- Requests per concurrency level: 10
- Warmup requests: 4
- Concurrency levels: 1, 2, and 4
- Negative prompts enabled


vllm_omni/diffusion/models/boogu_image/boogu_image_transformer.py
```diff
@@ -194,31 +207,49 @@ def __init__(
...
-        freqs_cis = []
-        freqs_dtype = torch.float32 if torch.backends.mps.is_available() else torch.float64
+    ) -> RotaryFrequencyTables:
+        freqs_real = []
+        # Ascend supports real-valued gather but float32 avoids routing the
+        # entire RoPE path through slower and less broadly supported float64
+        # kernels. Other platforms retain the previous precision where valid.
+        freqs_dtype = (
+            torch.float32
+            if current_omni_platform.is_npu() or not current_omni_platform.supports_float64()
```
hsliuustc0106: [P2] This drops the old explicit MPS fallback and regresses macOS/MPS runs.

The previous selection was `float32 if torch.backends.mps.is_available() else float64`. No platform in `vllm_omni/platforms/` overrides `supports_float64()` except MUSA, so on a Mac `current_omni_platform.supports_float64()` returns the interface default `True` and the tables are now built in float64 — then gathered and multiplied on MPS, where float64 ops are not supported. The model elsewhere still carries MPS-specific handling (`_get_freqs_real` moves `ids` to CPU for MPS), so MPS is clearly still an intended target.

Smallest fix: include MPS in the float32 branch, e.g. `if current_omni_platform.is_npu() or not current_omni_platform.supports_float64() or torch.backends.mps.is_available()` — or add an `is_mps()`/`supports_float64()` override to the platform interface so Wan/mammoth_moda2/omnigen2 (same idiom) benefit too.

hsliuustc0106 (commented):
Solid fix. I verified the real-valued formulation against the complex reference, the `supports_float64()` idiom (matches Wan/mammoth_moda2/omnigen2), and that diffusers' `get_1d_rotary_pos_embed` accepts `repeat_interleave_real` — plus the author numerics table and NPU/CUDA E2E evidence in the description. One P2 inline: dropping the old MPS→float32 branch regresses macOS, since `supports_float64()` defaults to True (only MUSA overrides it) while the code elsewhere still carries MPS-specific handling. Approve-able once that's addressed or explained.

vllm_omni/diffusion/models/boogu_image/boogu_image_transformer.py
JW-L-7: Thanks for catching this. I restored the previous MPS float32 fallback in `get_freqs_real` while retaining the NPU and platform capability checks.

I also added a regression test that simulates an MPS environment where the platform otherwise reports float64 support, and verifies that all generated RoPE cosine and sine frequency tables use float32.

The targeted real-valued RoPE tests pass: 4 passed.

JW-L-7:
## Self-review

### Scope

- The implementation changes are limited to the Boogu image model's rotary
  position embedding path and its associated tests.
- The complex-valued RoPE representation is replaced with an equivalent
  real-valued cosine/sine representation.
- No shared attention kernels or platform-wide behavior are modified.

### Correctness and accuracy

- Verified the real-valued RoPE formulation against the previous
  complex-valued reference implementation.
- Compared 3,932,160 RoPE output elements using FP32, FP16, and BF16.
- The observed maximum differences were:
  - FP32: `4.77e-7`
  - FP16: `0.001953125`
  - BF16: `0.0078125`
- For FP16 and BF16, the maximum difference is approximately one
  quantization step, and more than 99.98% of elements remain identical.

### Platform compatibility

- Ascend NPU end-to-end deployment and text-to-image inference were
  successfully verified.
- NVIDIA CUDA end-to-end deployment, inference, and serving benchmarks were
  successfully verified.
- CUDA-capable platforms continue to use float64 frequency-table generation
  where supported.
- The previous MPS float32 fallback has been retained.

### Tests

- The Boogu model unit tests passed in the NVIDIA validation environment.
- The targeted real-valued RoPE tests passed, including the new MPS regression
  coverage.
- Existing CI checks for Python 3.11, Python 3.12, pre-commit, DCO, and
  documentation have passed.

vllm_omni/diffusion/models/boogu_image/boogu_image_transformer.py
```diff
@@ -34,20 +34,33 @@
...
 
     Args:
         x: Query or key tensor of shape [B, S, H, D].
-        freqs_cis: Complex frequency tensor of shape [B, S, D // 2].
+        rotary_emb: Cosine and sine tensors, each shaped [B, S, D].
     """
-    x_rotated = torch.view_as_complex(x.float().reshape(*x.shape[:-1], x.shape[-1] // 2, 2))
-    freqs_cis = freqs_cis.unsqueeze(2)
-    x_out = torch.view_as_real(x_rotated * freqs_cis).flatten(3)
+    freqs_cos, freqs_sin = rotary_emb
```
hsliuustc0106, line 52: does it cause many bubles in profiling? otherwise, we can write some triton kernels to accelerate

JW-L-7: Thanks for raising this. I profiled both the previous complex-valued
implementation and the new real-valued implementation on an NVIDIA A800
using BF16 and the actual Boogu dimensions.

In eager mode, the real-valued path launches 10 kernels per call versus
3 kernels for the complex-valued path, and its isolated latency is
0.090–0.174 ms versus 0.033–0.055 ms.

However, the default vLLM-Omni deployment uses dynamic regional
torch.compile for the repeated Boogu transformer blocks. Under this path,
Inductor fuses the complete real-valued RoPE operation into a single
generated Triton kernel:

triton_poi_fused__to_copy_add_mul_slice_stack_sub_unsqueeze_view_*

The previous complex-valued path remains as 3 kernels because Inductor
reports that complex code generation is unsupported. In the isolated
compiled comparison, the real-valued path measured approximately
0.071 ms per call, compared with approximately 0.119 ms for the compiled
complex-valued path.

I also profiled an end-to-end 512x512, 28-step text-to-image request.
The fused real-valued RoPE kernels accounted for 89.40 ms in total,
or 1.58% of total GPU kernel time. The median adjacent launch gap was
approximately 1.15 us; the P95 gaps before and after the kernels were
approximately 1.76 us and 6.95 us, respectively.

Therefore, I did not observe significant bubbles in the default compiled
deployment. A separate hand-written Triton kernel does not appear
necessary because Inductor already generates a single fused Triton kernel.

vllm_omni/diffusion/models/boogu_image/boogu_image_transformer.py
hsliuustc0106: Fixed in aa9790b as suggested — MPS now joins the float32 branch with a dedicated regression test. Verified, thank you.

hsliuustc0106 (approved):
Delta verified: aa9790b adds MPS to the float32 branch exactly as suggested, with a proper regression test that mocks `torch.backends.mps.is_available` so it runs on Linux CI too. That resolves my P2; nothing outstanding from my side — the real-valued RoPE path, numerics evidence, and NPU/CUDA validation all stand.

hsliuustc0106:
please attach the acc/perf A/B test in both GPU/NPU

JW-L-7:
The following A/B tests were performed on an NVIDIA A800 and an Ascend 910B3,respectively.

Both used vLLM 0.28.0, BF16, the same Boogu weights, and the same core inference settings.

The GPU A/B used the exact PR base (`8e26dce1`) and head (`0fa34898`). The NPU test used the PR changes applied to `quay.io/ascend/vllm-omni:v0.28.0`.

### Accuracy

The operator tests used the actual Boogu RoPE dimensions. GPU results compare
the real-valued implementation with the previous complex-valued implementation;

NPU results validate the production real-valued RoPE path against an independently implemented, algebraically equivalent reference based on the standard 2×2 rotation matrix.

| Platform | Dtype | Max abs diff | Identical elements | `isclose` |
|---|---|---:|---:|---:|
| A800 | FP32 | `4.768e-7` | 60.77%–60.80% | 100% |
| A800 | FP16 | `9.766e-4` | 99.984%–99.985% | 100% |
| A800 | BF16 | `1.5625e-2` | 99.998% | 100% |
| Ascend 910B3 | FP32 | `4.768e-7` | 66.35%–66.37% | 100% |
| Ascend 910B3 | FP16 | `0.001953125` | 99.984%–99.987% | 100% |
| Ascend 910B3 | BF16 | `0.0078125` | 99.996%–99.998% | 100% |

For 10 paired GPU text-to-image requests with identical prompts and seeds:

| Metric | Result |
|---|---:|
| Successful requests | 10/10 for both base and head |
| Mean SSIM | `0.977749` |
| Mean PSNR | `38.3697 dB` |
| Mean absolute pixel difference | `1.74646/255` |

The generated images preserved the same subjects and composition; differences
were limited to fine texture, edges, and lighting details.

### NVIDIA A800 performance

Under `torch.compile`, the real-valued RoPE operator was `1.568x–1.908x`
faster than the previous complex-valued implementation.

| Metric | Base | Head | Change |
|---|---:|---:|---:|
| Mean latency | `7.09068 s` | `6.88956 s` | `-2.84%` |
| P90 latency | `7.17007 s` | `6.95289 s` | `-3.03%` |
| Throughput | `0.140711 QPS` | `0.144913 QPS` | `+2.99%` |
| Peak GPU memory | `37277 MiB` | `37277 MiB` | `0%` |

### Ascend 910B3 end-to-end validation

The NPU performance test used the same three prompts, seeds 42–51, two warm-up
requests, ten measured requests, 512x512 output, 28 inference steps,
`guidance_scale=4.0`, and concurrency 1 as the GPU A/B, repeated for three
rounds. The stability run retained the same inference settings and extended
the workload to 30 measured requests with seeds 1000–1029.

| Metric | Upstream complex | Real-valued PR |
|---|---:|---:|
| Service initialization | Blocked during warm-up | Passed |
| Blocking error | `aclnnGather` rejects `DT_COMPLEX128` | None |
| Performance requests | Unavailable | 30/30 passed |
| Mean latency | N/A | `14.113 s` |
| P50 / P90 / P99 latency | N/A | `14.103 / 14.178 / 14.312 s` |
| Throughput | N/A | `0.07082 QPS` |
| Peak NPU memory | N/A | `41,206 MiB` HBM |
| Stability run | Unavailable | 30/30 passed; `14.311 s` P99 |


The targeted Boogu tests passed on both platforms (`63 passed` on GPU and `63 passed` on NPU). 

The GPU A/B shows no accuracy, performance, or memory regression. 
The NPU validation confirms that the real-valued path resolves the Ascend compatibility blocker and runs accurately and stably end to end.

PR #6573, opened 2026-08-24
[Feature][Diffusion] Allow online INT8 quantization with DLO AllGather

brandneway:
DLO+AllGather rejected every online quantization method except per-tensor FP8, because only that finalized layout had been validated against DLO's dtype-flat shard / as_strided reconstruction protocol. Online INT8 and online MXFP8 are in the same layout family and are now allowed through the same path, so all three validated online methods (per-tensor FP8, INT8, MXFP8) can stack with DLO:

Online INT8

Int8OnlineLinearMethod (CUDA) leaves a transposed-view int8 weight plus a scalar fp32 scale -- the exact strided-view case the physical-order packing in _shard_and_pin already handles for online FP8.
NPUInt8OnlineLinearMethod (NPU) leaves a contiguous pre-transposed (K, N) int8 weight plus a 1-D fp32 per-channel scale -- plain, transportable dtypes with no exotic scale/packing layout.
Online MXFP8

NPUMxfp8OnlineLinearMethod (NPU) leaves a contiguous pre-transposed (K, N) fp8 weight plus a contiguous (K_groups/2, N, 2) e8m0 block scale -- plain 1-byte dtypes in ordinary contiguous layouts.
VllmMxfp8OnlineLinearMethod (vLLM kernel, XPU) leaves a contiguous (N, K) fp8 weight with its e8m0 scale stored as a transposed view over a contiguous (K/32, N) buffer -- the same strided-view family the physical-order packing already covers.
Every validated method keeps standard dtypes on the host shard, pinned staging, and the weight AllGather, so no transport changes are needed. The MXFP8 allowlist entry imports defensively, so the gate keeps working on vLLM builds without MXFP8 kernel support.

Also skip the AllGather allowlist when the effective DLO group size is one: no weight collective runs in that case (rank-local transfer), so an unvalidated online method must not be rejected there even with the default dlo_use_allgather=True.

Purpose
Allow the validated online quantization methods (per-tensor FP8, INT8, MXFP8) to combine with DLO AllGather, while keeping the gate fail-closed for unvalidated online methods.

Test Plan
pytest tests/diffusion/offloader/test_distributed_layerwise_backend.py -- byte-for-byte AllGather reconstruction of the finalized FP8/INT8/MXFP8 layouts;
pytest tests/diffusion/model_loader/test_diffusers_loader.py -- loader gate: online INT8/MXFP8 pass and use the ordinary loader before DLO sharding; group-size-one skips the gate; unvalidated methods are rejected;
pre-commit run --files <changed files>.

### DIFF
--- a/docs/design/feature/offloader/distributed_layerwise_offload.md
+++ b/docs/design/feature/offloader/distributed_layerwise_offload.md
@@ -27,6 +27,7 @@ Legend: ✅ supported, ⚠️ compatibility path or limited validation, ❌ unsu
 | **HSDP** | ❌ Rejected to avoid double-sharding parameters. | ⚠️ Limited end-to-end coverage. |
 | **Per-tensor online FP8 linears** | ✅ Ordinary loader finalizes weights and scales before DLO sharding. | ✅ Ordinary loader retains complete rank-local tensors. |
 | **Online INT8 linears** | ✅ Ordinary loader finalizes the int8 weight (contiguous or transposed view) and fp32 scale before DLO sharding. | ✅ Ordinary loader retains complete rank-local tensors. |
+| **Online MXFP8 linears** | ✅ Ordinary loader finalizes the fp8 weight (contiguous) and e8m0 block scale (contiguous or transposed view) before DLO sharding. | ✅ Ordinary loader retains complete rank-local tensors. |
 | **Other online quantization methods** | ❌ Rejected until runtime packing and scale layouts are validated. | ⚠️ Allowed through the ordinary loader; validation is method-specific. |
 | **Model-level or standard layerwise CPU offload** | ❌ Disabled because DLO takes priority. | ❌ Disabled because DLO takes priority. |
 | **Resident leading layers** | ❌ Rejected. | ✅ Requires eligible resident paths in the model's `OffloadPlan`. |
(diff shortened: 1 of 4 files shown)


tests/diffusion/offloader/test_distributed_layerwise_backend.py
```diff
@@ -245,6 +245,160 @@ def fake_allgather(output, local_shard, *, group):
...
+            def __init__(self):
+                super().__init__()
+                logical = torch.arange(1.0, 13.0).reshape(3, 4)
+                # NPU online int8: qweight.t().contiguous() -> (K, N), 1-D scale
+                self.npu_weight = nn.Parameter(logical.t().contiguous().to(torch.int8), requires_grad=False)
+                self.npu_weight_scale = nn.Parameter(torch.tensor([2.0, 3.0, 4.0]), requires_grad=False)
+                # CUDA online int8: Parameter(qweight.t().data) -> transposed
+                # view (stride (1, K)) over the (N, K) storage, scalar scale
+                self.cuda_weight = nn.Parameter((logical + 100).to(torch.int8).t(), requires_grad=False)
+                self.cuda_weight_scale = nn.Parameter(torch.tensor([0.5]), requires_grad=False)
```
hsliuustc0106, line 272: nit: the finalized CUDA online-INT8 scale is per-channel, not scalar — `Int8OnlineLinearMethod.process_weights_after_loading` quantizes via `ops.scaled_int8_quant(layer.weight, scale=None)`, which is vLLM's dynamic per-token branch returning an `(N, 1)` fp32 scale — so `torch.tensor([0.5])` (and the "scalar fp32 scale" wording in the PR description) mis-models the layout the protocol will actually see. Descriptive only: a contiguous (N, 1) scale flows through the same dtype-grouped packing as the 1-D NPU scale above, so the allowlist conclusion stands. Suggest `torch.tensor([[0.5], [0.6], [0.7]])` here and dropping "scalar" from the description.

david6666666:
GPU accuracy validation (base fbc93a0e08e4 vs PR head b6cd24f594f5) on 8x NVIDIA B300 (4 GPUs per run), using the same local MiniMax-H3 Ref2VA snapshot and:
`pytest -q tests/e2e/accuracy/minimax_h3/test_minimax_h3_i2va_ref2va_similarity.py --run-level full_model`

Both runs passed (2/2), with SSIM >= 0.97 and PSNR >= 34 dB:
- I2VA: base SSIM 0.972621 / PSNR 35.074173 dB; head 0.974463 / 36.098755 dB.
- Ref2VA: base and head SSIM 0.980582 / PSNR 42.651423 dB (the generated videos were byte-identical).

The new DLO loader/backend unit coverage also passes (5 targeted tests). Note that this accuracy test uses HSDP and does not enable DLO or online INT8/MXFP8 quantization, so it is a regression check for the default path rather than end-to-end coverage of the new DLO+online-quantization path.

docs/design/feature/offloader/distributed_layerwise_offload.md
```diff
@@ -459,8 +465,8 @@ Current source-level validation includes:
 - HSDP + DLO without AllGather acceptance at configuration level;
 - loader preflight fallback for TP, HSDP, online quantization, unknown custom
   loaders, missing keys, and shape/dtype mismatches;
-- ordinary-loader fallback for per-tensor online FP8 linears followed by DLO
-  sharding of finalized weights and scales;
+- ordinary-loader fallback for per-tensor online FP8, online INT8, and online
+  MXFP8 linears followed by DLO sharding of finalized weights and scales;
```
david6666666, line 469: [P2] Add a path-level check for the new allowlist

These checks are mock/unit-level, while the MiniMax-H3 accuracy test uses HSDP without DLO or online INT8/MXFP8. A runtime layout regression in the newly allowed AllGather path could therefore pass this suite. Please add a guarded GPU smoke for one supported method, or qualify the ✅ claim as source-level validation.

david6666666 (commented):
The loader gate and group-size-one handling look coherent, and the five added targeted tests pass. I left one non-blocking coverage comment below.

david6666666 (approved):
LGTM now

PR #6575, opened 2026-08-24
Update install_torchcodec_rocm.sh

adityaraj536:

## Test Plan

**vLLM Version:**

**vLLM-Omni Commit:**

## Test Result


congw729 (changes requested):
Thanks for contributing! I took a look at the change, and this PR currently only removes a blank line after `set -euo pipefail`; it does not change the script’s behavior or provide a meaningful formatting improvement. The PR description is also still the unfilled template, so the intended problem or motivation is unclear.

Could you please explain the intended fix and update the PR with the corresponding substantive change? If the blank-line removal was accidental, I think this PR can simply be closed. Happy to take another look if there is a larger change you meant to submit!

PR #5178, opened 2026-07-17
[Perf][Qwen3-TTS] Defer per-step save-path D2H and vectorize codec flatten

ShengleiFu:
## Purpose

Address W3 from the TTS refactor RFC #4855: remove the per-step host work on
the Qwen3-TTS async-chunk save path, which keeps the GPU idle under concurrency.

On each talker decode step the save path copied the frame to host
(`frame.cpu().tolist()`) and, on every emitted chunk, rebuilt the codebook-major
window with a Python double loop. Both run per request per step on the save
thread, so the device-to-host copy serializes the step loop as concurrency
rises. The sibling `qwen3_omni` save path already avoids this by accumulating
on-device frame tensors and flattening with a single vectorized transpose; this
PR brings `qwen3_tts` to the same shape.

This PR:

- Accumulates on-device frame tensors instead of `frame.cpu().tolist()`, so the
  host copy is deferred off the per-step path to one transfer per emitted chunk
  (matching `qwen3_omni` and the accumulator's declared `list[torch.Tensor]`
  type).
- Replaces the per-element Python flatten with a single
  `torch.stack(window).transpose(0, 1).reshape(-1)`, device-aligning the ref
  context before it is prepended.
- Adds save-path regression coverage: codebook-major layout and the all-zero
  frame skip, exercised through the real per-step append path.

Backward compatibility: behavior-preserving. Emitted `codes.audio` contents and
order, `left_context_size`, ref-context metadata, chunk cadence, and the
all-zero skip are unchanged; the full async-chunk suite (53 tests) passes.

Scope note: the `_extract_last_frame` emptiness guard (one `.any()` per step) is
kept — it mirrors `qwen3_omni`'s guard and the full-payload/token-only zero-row
filter. Removing it safely depends on whether the talker ever emits all-zero
frames mid-stream; deferred as a follow-up pending profiling.

Not a duplicate: #5070 batches the same class of per-step D2H for `mimo_audio`
(a different processor/file); #4745 is W4 (batched decode preprocess in the
talker). No open PR touches the `qwen3_tts` async-chunk save path.

## Test Plan

**vLLM Version:** vllm 0.25.0 (serving benchmark, on origin/main); unit suite also green under vllm 0.24.0

**vLLM-Omni Commit:** this branch (base `07957e12`)

[...]

PR #5221, opened 2026-07-20
[Perf][Engine] Event-driven orchestration loop (opt-in) — S1 of #4855

linyueqian:
## Purpose

S1 of the scheduling redesign in RFC #4855 (problem 3: the orchestrator coordination layer is poll-driven and keeps the GPU idle under load). This makes the orchestration loop **event-driven behind an opt-in flag** (`VLLM_OMNI_EVENT_DRIVEN_ORCH=1`, default off). The legacy poll loop is untouched when the flag is unset.

### What is slow today

The `Orchestrator` runs one coroutine that *serially* scans every stage replica with `asyncio.wait_for(client.get_output_async(), timeout=0.001)` and `asyncio.sleep(0.001)` when idle (`orchestrator.py`), and the serving-side final-output drain busy-polls `get_nowait()` + `sleep(0.001)` (`entrypoints/async_omni.py`). Two consequences:

- Every audio chunk crosses two-to-three 1 ms poll boundaries of pure jitter — a large fraction of the inter-chunk budget for a ~80 ms/chunk codec, and a direct contributor to streaming discontinuity at load.
- The `wait_for(..., 0.001)` emulation creates and cancels ~2 tasks per replica per millisecond, so the process spins at ~1 kHz even when completely idle.

The underlying stage client is vLLM's own `AsyncMPClient`, whose `get_output_async()` is natively awaitable and surfaces `EngineDeadError` through the same queue — upstream `AsyncLLM` consumes it with a plain `await`. This change restores that pattern for the omni orchestrator.

## Changes

- **`vllm_omni/engine/orchestrator.py`** — new `_orchestration_loop_event_driven()` (selected by the flag): one reader task per available LLM stage replica `await`s `client.get_output_async()` directly and feeds a single dispatch queue that the loop consumes **serially**, so routing/handling semantics are identical to the legacy loop; only the poll cadence is removed. Diffusion stages keep their nowait-poll contract via a per-pool poller task. Readers and pollers are reconciled against `available_replica_ids()` and client swaps on a 0.5 s wall-clock schedule, which covers elastic membership and replica eviction. The per-output handling is extracted into `_process_llm_stage_outputs()` and reused by **both** loops, so there is one code path for behavior. A one-line enabled-log announces the loop mode + reader/poller counts.
[...]

PR #6576, opened 2026-08-24
[Perf][Qwen3-TTS] Batch the async-chunk emptiness sync instead of syncing per decode step

linyueqian:
## Purpose

W3 of RFC #4855: remove the per-step device sync from the Qwen3-TTS async-chunk save path.

On every talker decode step the save path did two host round-trips per frame. `_extract_last_frame` called `frame.any().item()` to test whether the frame was all-zero padding, and the caller then did `frame.cpu().tolist()` before appending. Both run per request per step on the save thread, so they serialize the step loop as concurrency rises.

This continues @ShengleiFu's work in #5178, which was closed unmerged. That PR deleted the emptiness guard and filtered padding at chunk emit instead. **That approach cannot be rebased onto current main, and this PR does not use it** — see below.

### Why filtering at emit does not work any more

All-zero frames are not an edge case. The talker zeroes any frame whose layer-0 id is out of codebook range (`qwen3_tts_talker.py:1301`), which covers EOS, and the prefill path emits an all-zeros span outright (`:817`). #5178's own commit message measured it: *"~1.9% of mid-stream frames are all-zero (measured at c=128), so the guard is load-bearing and cannot simply be deleted."*

`len(code_prompt_token_ids[rid])` drives `compute_ramp_emit`, the `adjusted % chunk_size` gate, and `context_length`. Letting padding into that count moves every chunk boundary. #5178 compensated by recomputing `left_context_size` from the surviving frames, which worked only because Code2Wav then received an overlapping left-context window to trim against. **#5202 removed that window** — Code2Wav keeps per-request decoder state now — so the compensation no longer has anything to act on.

## Changes

Padding is still dropped *before* it reaches the accumulator, so `length` keeps meaning "real frames so far" and every gate sees exactly what it saw before. The sync is **batched** rather than removed.

- **`_extract_last_frame`** no longer inspects content. The frame goes to a per-request `pending_frames` list, on-device and unclassified. No sync on the step path.
- **`_commit_pending_frames`** resolves pending frames with one `.tolist()` over the stacked batch — a single sync covering all of them, with only the boolean mask crossing to the host.
- **`_emit_threshold`** is pure host arithmetic mirroring the emission gate (ramp cumulative, initial-chunk branch, steady-state modulo). The resolve runs only when the pending frames could change the decision: on finish, or when `len(committed) + len(pending)` could reach the threshold. Under-estimating costs an extra sync; over-estimating would delay a chunk, so the branches are kept in step with the gate directly below them.
- **At emit**, committed frames are already padding-free device tensors, so the codebook-major flatten is one `transpose(0,1).reshape(-1)` instead of a Python double loop over `num_quantizers x num_frames`, with a single `.cpu()` keeping the payload host-resident as before.

Measured on the resolve path: **5 resolves for 60 decode steps** at `codec_chunk_frames=25`, against 60 syncs before.

### Two behavior notes, called out deliberately

**Dynamic-IC load factor.** A request counts as active from its first frame rather than its first resolve. Frames now sit in `pending_frames` before committing, so counting only the committed side would under-report load by one for every request still inside its first chunk — including the request computing its own IC, which the old code always counted. `active` now unions both stores.

**Abort releases device frames.** `finish_requests` handles abort, error, and eviction, where no terminal chunk is sent and `cleanup_sender` never runs. That path already stranded `code_prompt_token_ids`; with host lists it leaked a little host memory, but the accumulators hold device tensors now, so it would hold GPU memory for the adapter's lifetime. The frame stores are released there. Only the frame stores — the chunk counters stay on `cleanup_sender`'s lifecycle, since calling it before a terminal send is explicitly unsupported.

## Test Plan

**vLLM Version:** 0.27.0

**vLLM-Omni Commit:** based on current main

- `pytest tests/model_executor/stage_input_processors/`
- `pytest tests/distributed/omni_connectors/` for the transfer-adapter cleanup path
- Mutation testing on each behavior this PR is responsible for
- Differential run of the padding property test against upstream's processor

## Test Result

**379 passed** in `tests/model_executor/stage_input_processors/`.

New coverage:

- A parametrized property test asserting padding is invisible to the emission schedule, across five chunk/initial-chunk/ramp configurations x 8 seeds.
- Boundary, padding-only, and terminal-padding cases.
- A sync-count test pinning that resolves scale with chunks rather than steps.
- The dynamic-IC load factor, which the padding tests cannot catch because both arms would shift together.
- Abort releasing the accumulators.

**Mutation-verified.** Each of these fails the test that covers it: counting padding toward `length`; resolving on every step; substituting a placeholder for the empty-finished sentinel; reverting the load-factor fix.

**Differential against main.** Running the padding property test against upstream's processor: main fails 3 of 40 cases, this PR passes 40. Main re-emits a window when a padding frame arrives while the accumulator sits exactly on a chunk boundary. That is pre-existing and this PR happens to remove it in those configurations; the test normalizes consecutive duplicate windows so it is asserting the shared contract rather than that incidental improvement.

### Honest caveats

- No e2e serving A/B on this branch yet. The claim here is "same emissions, fewer syncs", which is what the unit tests pin; the throughput benefit still wants a concurrency run before anyone counts on a number.
- The resolve heuristic is tuned for streams where padding is sparse (~2%). A pathological stream that is mostly padding would resolve more often — still correct, just closer to the old sync rate.
- The abort fix releases the frame stores but leaves the chunk counters to `cleanup_sender`, so aborted requests still retain small per-request counter entries until the adapter is torn down. That is pre-existing and unchanged.

Credit to @ShengleiFu for the original investigation and the c=128 padding measurement in #5178, which is what made the constraint here visible.


hsliuustc0106:
does it help compete with 
39a94479d71fceb5276e60464836c840

linyueqian:
Slope yes, floor no.

This removes an O(load) device sync from the save thread — that thread absorbs two syncs per frame per active request, and `.item()` drains the device while the talker and Code2Wav kernels are running. Chunk emission is serialized behind it, so under concurrency a backlogged save thread directly delays first audio. That is the regime where 118 ms → 397 ms happens.

It does not touch the ~49 ms floor. At 1 RPS the save thread is idle, so the syncs cost microseconds. The floor is the latency chain: talker prefill + one decode step + two connector hops + two 1 ms poll loops + Code2Wav decode. No amount of sync removal gets that to 27 ms.

Closing the floor needs different levers, and two of them are things we already documented and then did not ship:

- `deploy/qwen3_tts.yaml:87` says "CustomVoice / VoiceDesign are TTFA-optimal at 1", and the file ships `max_num_seqs: 64` on stage 1. The benchmark is CustomVoice.
- The header comment (lines 6-8) says stage 0 uses `max_num_batched_tokens=512` for first-chunk latency; line 62 ships `32768`. #4923 raised it and the comment stayed. `qwen3_tts_high_concurrency.yaml:56` still has 512.

Plus #5221, which replaces the two 1 ms poll loops with event-driven wakeups (measured -14.7% mean / -32.1% p99 TTFP at concurrency 8).

One thing worth confirming before we read too much into the comparison: what does their harness count as audible? With `initial_codec_chunk_frames: 1` our first chunk is a single frame, ~80 ms of audio. If they wait for a larger playable buffer the two numbers are not measuring the same event.

I would rather re-run their setup ourselves with the CustomVoice tuning applied than argue from their chart — otherwise we cannot tell whether the gap is the engine or our defaults.

PR #6579, opened 2026-08-24
[Frontend] Move non-streaming MP4 response encoding off the event loop

MosCloud:
## Purpose

Non-streaming MP4 responses currently run the complete PyAV/libx264 encode on the API server's asyncio event loop. In the validated 5-second workload, this blocks other API work for roughly the 4.7-second encoding interval.

This change moves complete raw MP4 and existing base64 response encoding to a handler-owned single-worker executor. An admission gate allows one native encode to be submitted at a time, cancellation and synchronous endpoint timeouts retain the slot until native codec work finishes, and shutdown gives active video tasks a five-second cancellation window before continuing handler and engine cleanup.

Streaming fMP4 and action-only responses keep their existing paths. There is no CLI option or request parameter. The mechanism is model- and platform-neutral, but it has currently been validated only with MiniMax-H3 `FL2VA`/`t2va` on the Atlas A2 configuration below.

## Test Plan

**Performance A/B stack:** vLLM `0.26.0+empty` with vLLM-Ascend `0.19.1rc2.dev1251+g905bbf372`

**vLLM-Omni revisions:** current review head `b14327b14656d4d9c23b24bb3cf5e9539d46a9f8`; A2 performance run `b10656285122b463a2bb868fdac35e9cb8cf7969`

- Run the complete video server CPU test file, Ruff check, Ruff format check, and the relevant repository static gates.
- Exercise both normal and stalled video-task cancellation before handler shutdown.
- Run a fixed-input CPU A/B benchmark with a 124-frame 1344x768 MiniMax-H3 dump, stereo 32 kHz audio, CPUs 72-95, one warmup, and five formal runs per side.
- Compare base `d150a4fde77d15d466102323a4048b0a8631d74c` and candidate `b10656285122b463a2bb868fdac35e9cb8cf7969` on one Atlas A2 host with 8x Ascend 910B4-1 NPUs. Use `FL2VA`/`t2va`, Laser attention, 1344x768 at 24 FPS, 124 frames, seed 1101, 50 requested steps/49 DiT forwards, one request at a time, one discarded 5-second warmup, and three formal 5-second requests per side.
- Validate every formal output with full decode and ffprobe, and monitor all ranks, devices, response headers, event-loop health probes, HBM, and shutdown state.
- Run one warmup and one formal 5-second MiniMax-H3 A2 smoke request after the shutdown fix, then terminate the API main process and verify complete worker, port, and device cleanup.

The performance A/B stack used driver 25.5.2, firmware 7.8.0.7.220, CANN 9.0.1, Python 3.12.13, PyTorch 2.10.0+cpu, torch_npu 2.10.0.post2, MindIE-SD 3.0.0, PyAV 18.0.0, and ffmpeg/ffprobe 4.4.2. The post-fix shutdown smoke used vLLM and vLLM-Ascend 0.28.0 with torch_npu 2.10.0.post4 on the same host.

## Test Result

- Complete video server CPU suite: `97 passed`.
- Current shutdown regression coverage: `3 passed`, covering registry cleanup plus normal and stalled cancellation.
- Ruff check, Ruff format check, SPDX, test-mark, forbidden-import, torch.cuda, and `git diff --check` gates passed.
- The post-fix A2 smoke returned HTTP 200 with a valid H.264/AAC MP4. Main-process SIGTERM completed with all eight workers exiting and the port and devices released. The final main rebase preserved both PR patches exactly under `git range-diff`.

Fixed-input CPU results (median, `n=5`):

| Metric | Inline | Dedicated executor | Change |
| --- | ---: | ---: | ---: |
| Maximum heartbeat gap | 4766.558 ms | 10.033 ms | -99.79% |
| Encode wall time | 4761.372 ms | 4702.553 ms | -1.24% |
| Process CPU time | 7094.849 ms | 7068.994 ms | -0.36% |

The CPU outputs were byte-identical.

Atlas A2 E2E results (formal median, `n=3`):

| Metric | Base | Candidate | Change |
| --- | ---: | ---: | ---: |
| Maximum `/health` response during each request | 4.350 s | 0.185 s | -95.74% |
| MP4 encoding | 4726.080 ms | 4816.380 ms | +1.91% |
| Stage 0 | 158465.635 ms | 160077.244 ms | +1.02% |
| Server E2E | 163.138 s | 164.909 s | +1.09% |

The `/health` result is the scheduling acceptance metric; MP4, Stage 0, and E2E values are diagnostic. All eight warmup/formal cases passed, and all six formal outputs had the same SHA-256 (`246100788ea4a839a3b0dc1a7a33405dd9b1f6252b6f6889cf8c70682f73c351`) across Base and Candidate. Each output decoded as 124-frame H.264 1344x768 video at 24 FPS with AAC stereo audio at 32 kHz. No rank failure, OOM, collective timeout, NaN/Inf, or backend fallback occurred.


recipes/MiniMaxAI/MiniMax-H3-NPU.md
```diff
@@ -133,6 +133,79 @@ time, 1344x768 at 24 fps, and 5, 8.7, and 15 second requests. This optimization
 only changes CPU MP4 response encoding; it does not change DiT execution or
 stage 0.
 
+#### Event-loop scheduling
+
+Complete non-streaming raw MP4 encoding runs in a dedicated single-worker
```
alex-jw-brooks: In my opinion, this is the wrong place for this summary for a couple of reasons.

- The delegation to avoid block the asyncio loop is platform independent, but being part of this recipe implies it's NPU specific 
- This is largely internal details for an optimization that should (at least in the longer term assuming compatible layouts) be model agnostic, and recipes should be more focused on how to do things than on how they work

Especially given that there's no user-facing configuration, I'd actually suggest removing it from the recipe and just ensuring things are well-documented in the PR description, and potentially adding a design doc on this to the public docs as a follow up

tests/benchmarks/benchmark_video_response_event_loop.py
```diff
@@ -0,0 +1,210 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM-Omni project
+"""Benchmark event-loop responsiveness for inline and offloaded MP4 encoding."""
```
alex-jw-brooks: Is there a concrete reason why we should add this script given that we will move to unblock the event loop in this PR? 

It is useful for validating this change, but I don't think it should be added since we likely won't use it again in the future

tests/entrypoints/openai_api/test_video_server.py
```diff
@@ -109,12 +113,311 @@ async def _generate_both_response_types():
...
+    path_logs = []
+
+    async def fake_run(*args, **kwargs):
+        return artifacts
+
+    def encode(video, *, fps, **kwargs):
+        encoder_threads.append(threading.get_ident())
+        return video_api_utils._encode_video_bytes(video, fps=fps, **kwargs)
+
+    monkeypatch.setattr(handler, "_run_and_extract", fake_run)
```
alex-jw-brooks: The encode path should already be dictated by the dimension ordering of `video`, right? If so, can avoid of all the monkey patching in `media_utils`?

IMO it would make more sense to just ensure that when running real encoding, we can run both paths without crashing, and ensure that the thread ID check passes. It will make this test much less brittle against refactors

tests/entrypoints/openai_api/test_video_server.py
```diff
@@ -109,12 +113,311 @@ async def _generate_both_response_types():
...
+        and "max_active_encodes=1" in message
+        for message in log_messages
+    )
+    timing_logs = [message for message in log_messages if message.startswith("Video response encoding (MP4")]
+    assert len(timing_logs) == 2
+    assert all(
+        "queue_inclusive=true" in message
+        and "execution=dedicated_executor" in message
+        and "max_active_encodes=1" in message
+        for message in timing_logs
```
alex-jw-brooks: I think this test should be removed. I don't think we should have dedicated unit tests for log messages, especially when the logger is monkeypatched, since this test will be broken by minor changes in logging levels or formats, refactors, etc.

tests/entrypoints/openai_api/test_video_server.py
```diff
@@ -109,12 +113,311 @@ async def _generate_both_response_types():
...
+
+@pytest.mark.asyncio
+async def test_action_only_raw_returns_before_executor_submission(monkeypatch):
+    handler = _make_video_handler()
+    action = object()
+    artifacts = _video_artifacts([{"action_only_output": True}], actions=[action])
+    submit_calls = []
+
+    async def fake_run(*args, **kwargs):
+        return artifacts
```
alex-jw-brooks: There is a lot of boilerplate in these tests, e.g., the same copied `fake_run` lambda, handler shutdown, etc. please refactor to use common fixtures. It will make things a lot less verbose, and also make it more clear what each test is actually checking

tests/benchmarks/benchmark_video_response_event_loop.py
```diff
@@ -0,0 +1,210 @@
...
+    try:
+        output = await handler._run_video_response_encoding(encode) if variant.offload else encode()
+        wall_ms = (time.perf_counter_ns() - wall_start) / 1_000_000
+        process_cpu_ms = (time.process_time_ns() - cpu_start) / 1_000_000
+    finally:
+        await asyncio.sleep(2 * heartbeat_interval_s)
+        stop_heartbeat.set()
+        await heartbeat_task
+
+    heartbeat_gaps_ms = [(current - previous) * 1000 for previous, current in zip(heartbeat_times, heartbeat_times[1:])]
```
hsliuustc0106: Could we remove this p99 claim or compute it from a fixed heartbeat schedule? Inline encoding blocks the event loop, so the heartbeat task emits only a few samples and one large post-encode gap, while the executor run emits a sample every interval. The p99 populations (and interpolation) therefore differ; max_heartbeat_gap_ms is the comparable metric here.

tests/entrypoints/openai_api/test_video_server.py
```diff
@@ -109,12 +113,311 @@ async def _generate_both_response_types():
...
+        handler.shutdown()
+
+    assert any(
+        "execution=dedicated_executor" in message
+        and "scope=non_streaming_mp4" in message
+        and "paths=raw_mp4,base64_handler" in message
+        and "max_active_encodes=1" in message
+        for message in log_messages
+    )
+    timing_logs = [message for message in log_messages if message.startswith("Video response encoding (MP4")]
```
hsliuustc0106: This test never queues on the semaphore: both encodes are awaited sequentially, and it only asserts the literal queue_inclusive=true label. It would pass even if _t_encode_start moved after await _run_video_response_encoding(...). Please add concurrent encodes with the first native encode blocked, then assert the second timing includes its wait.

tests/entrypoints/openai_api/test_video_server.py
```diff
@@ -199,34 +538,72 @@ async def get_supported_tasks(self):
...
+        record("storage_start")
 
     async def fake_get_vllm_config(engine_client):
         del engine_client
         return None
 
     async def fake_init_app_state(engine_client, state, args):
-        del engine_client, state, args
-        events.append("init_app_state")
+        del engine_client, args
```
alex-jw-brooks: I think that we need to reconsider the design of some of these tests. There is a lot of mocking going on here that is likely not needed, and in most cases it would be better to just test the entrypoint directly without so much patching, especially now that there are things being patched into the fake app init.

Can you please see if this can avoided, or add a `TODO: explore integration against actual entrypoint` comment? I can also help with this once I have time if it is urgent to land this PR quickly, since this is more of an issue in our testing framework imo. cc @hsliuustc0106

alex-jw-brooks (changes requested):
Thanks for adding this. Some suggestions, mostly around test redundancy and potentially removing some things from recipes / benchmarks that IMO don't belong there. :)

recipes/MiniMaxAI/MiniMax-H3-NPU.md
MosCloud: Removed the PR3 section from the MiniMax-H3 NPU recipe. There is no operator-facing setting, and the mechanism belongs to the model- and platform-neutral frontend path. The PR description now carries the design and the scoped MiniMax-H3/Atlas A2 evidence.

tests/benchmarks/benchmark_video_response_event_loop.py
MosCloud: Removed the checked-in benchmark script. It was a one-time validation tool for this change; the reproducible workload description and measured result remain in the PR description without adding a permanent benchmark surface.

tests/entrypoints/openai_api/test_video_server.py
MosCloud: Replaced the media_utils and route-logger monkeypatches with tiny real encodes for both supported layouts. The test only wraps the real encoder to record the worker thread ID, then asserts non-empty MP4 output and that encoding ran off the event-loop thread.

tests/entrypoints/openai_api/test_video_server.py
MosCloud: Removed the dedicated log-format test. Production logs remain, but tests no longer bind their wording or level. Queue-inclusive timing is now verified directly through the executor helper.

tests/entrypoints/openai_api/test_video_server.py
MosCloud: Added a shared video handler fixture with automatic shutdown plus common artifact and generation helpers. This removes the repeated fake_run setup and per-test shutdown blocks while keeping each test focused on one behavior.

tests/benchmarks/benchmark_video_response_event_loop.py
MosCloud: Removed the p99 claim everywhere in the PR. The inline and executor sample populations are not comparable, so maximum heartbeat gap is now the only heartbeat scheduling metric reported.

tests/entrypoints/openai_api/test_video_server.py
MosCloud: Moved timing ownership into _run_video_response_encoding, with the clock starting before semaphore acquisition. A concurrent test blocks the first native encode, starts a second encode, and uses a controlled clock to prove the second elapsed value includes its gate wait; it does not assert log text.

tests/entrypoints/openai_api/test_video_server.py
MosCloud: Split the previous test into focused handler shutdown, TaskRegistry cancellation, and entrypoint shutdown-order tests. The entrypoint test now constructs app state before startup and registers the tracked task after serve starts, rather than injecting handlers or work through fake_init_app_state. The remaining framework seams are marked with the requested TODO. The complete video server file passes: 97 passed.

MosCloud:
Self-review after addressing the review feedback:

- Reviewed the final diff against `main`: it is limited to the frontend executor/lifecycle implementation and focused video server tests. The MiniMax-H3 NPU recipe and one-time benchmark script are no longer part of the PR.
- Rechecked resource ownership and lifecycle: admission happens before executor submission, cancellation keeps the slot until native work finishes, tracked video tasks are cancelled before executor shutdown, and action-only/streaming paths remain unchanged.
- `tests/entrypoints/openai_api/test_video_server.py`: `97 passed` (15 existing deprecation warnings). Ruff check/format, SPDX, test-mark, forbidden-import, torch.cuda, and `git diff --check` gates passed.
- The exact local `mypy-3.10` hook remains non-clean on existing errors in the already modified entrypoint/test files. I checked the diagnostics against the final diff; none fall on lines added by this PR.
- I did not repeat the A2 E2E matrix for the review revision: it centralizes the existing queue-inclusive timing and changes tests/documentation, without changing native encoding or executor admission/cancellation behavior. The A2 run at `b1065628` remains the hardware evidence; the final revision is covered by the focused real-encoding CPU tests.

@alex-jw-brooks @hsliuustc0106, please take another look when convenient.

tests/benchmarks/benchmark_video_response_event_loop.py
hsliuustc0106: Removing the claim (and with it the non-comparable populations entirely) resolves this — max_heartbeat_gap_ms was the right metric to keep.

tests/entrypoints/openai_api/test_video_server.py
hsliuustc0106: That is exactly the test shape I asked for — clock before semaphore acquisition, concurrent encodes with the first blocked, gate wait proven included. Resolved.

hsliuustc0106 (approved):
LGTM — both my threads resolved: the p99 claim is gone (max_heartbeat_gap_ms is the comparable metric), and the queue-inclusive timing test now blocks the first encode and proves gate-wait inclusion with a controlled clock.

MosCloud:
Rebased this pull request onto the current main after the frame-conversion changes landed. The conflict was resolved by keeping the two concurrency domains separate: frame conversion continues to use the handler-owned 8-worker pool from main, while complete non-streaming MP4 response encoding uses a separate handler-owned single-worker executor with one active encode admitted at a time.

Shutdown now cancels and awaits tracked background video tasks before shutting down the handler-owned encoding and frame-conversion executors, preventing video tasks from submitting work after their handler resources have been closed. Raw MP4 and base64 non-streaming responses use the offloaded path, while streaming fMP4 and action-only responses remain unchanged.

The rebased commits are patch-equivalent to the previously reviewed revision. Focused Ruff, format, and diff checks passed, so the existing CPU A/B and Atlas A2 E2E evidence continues to cover the same response-encoding behavior.

tests/entrypoints/openai_api/test_video_server.py
hsliuustc0106: Delta check at 5005a00 (postdates my approval): verified — timing now starts before `gate.acquire()` inside `_run_video_response_encoding` and returns `(result, elapsed_ms)`, both call sites simplified, and the recipe internals section + benchmark script removals match the thread agreements. Good to go.

vllm_omni/entrypoints/openai/api_server.py
```diff
@@ -615,6 +615,7 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
             await shutdown_task
         finally:
             state = getattr(app, "state", None)
+            await VIDEO_TASKS.cancel_all()
```
linyueqian: [important] Bound this wait and let the rest of the `finally` run if it times out. Cancelling an in-flight job enters `AsyncOmni.generate()`'s cancellation handler, which awaits `abort_async(timeout=None)`. Only `AsyncOmniEngine.shutdown()` closes the RPC router that can release that wait, and it runs after this line. A dead or wedged orchestrator can therefore deadlock shutdown and leave stage workers running. Wrap the cancellation in `asyncio.wait_for(...)` and continue cleanup on timeout.

linyueqian (changes requested):
Reviewed at `5005a003`. Moving non-streaming video encoding off the event loop looks good, but shutdown can still hang indefinitely while cancelling active video requests. Details are inline.

Validation: static review of the complete diff from merge base `72a1ce48` and the cancellation, RPC, and shutdown paths. I did not execute the untrusted fork code.

vllm_omni/entrypoints/openai/api_server.py
MosCloud: Thanks for catching this shutdown deadlock path. Fixed in `b14327b1`: `VIDEO_TASKS.cancel_all()` is now bounded by a five-second `asyncio.wait_for`. If cancellation does not complete, the server logs a warning and continues video/speech handler and socket cleanup, allowing the enclosing engine shutdown to close the RPC router.

I added normal and stalled-cancellation variants to the server shutdown test, including the timeout warning and cleanup order. The three focused registry/shutdown cases pass. A MiniMax-H3 A2 smoke run also completed main-process SIGTERM with all eight workers exiting and the port and devices released.

linyueqian (approved):
Re-reviewed the delta at b14327b1. The round-three finding is resolved, and the test that came with it is better than what I asked for.

The shutdown wait is now bounded: cancel_all runs under asyncio.wait_for with a five second budget, and a timeout logs a warning and falls through instead of propagating, so the rest of the finally block still runs. That is what the finding needed, since the deadlock risk was never the cancellation itself but the fact that nothing after it could execute, including the engine shutdown that closes the RPC router the cancelled request is waiting on.

The regression test reproduces the actual failure rather than approximating it. It registers a video task that hangs forever inside its own CancelledError handler, which is the wedged-orchestrator shape from the finding, and parametrizes normal against timeout. In both cases it asserts the full shutdown ordering completes, from HTTP shutdown through video and speech shutdown to socket close and engine exit, and it asserts the warning appears only in the timeout case. That ordering assertion is the one that matters, because it is what proves the timeout path does not skip cleanup.

No new findings. The earlier suggestion about peak memory, where the gate is reached only after all frames are materialized, still stands as a follow-up and is not a merge blocker.

Validation: static review of the delta against the round-three finding, the bounded wait and its fall-through path, and the new test's fixtures and assertions. GitHub checks pass. Per-PR Buildkite has never run on this branch, so I have added the ready label and the approval waits on those lanes. No fork code was executed.

MosCloud:
> Re-reviewed the delta at [b14327b](https://github.com/vllm-project/vllm-omni/commit/b14327b14656d4d9c23b24bb3cf5e9539d46a9f8). The round-three finding is resolved, and the test that came with it is better than what I asked for.
> 
> The shutdown wait is now bounded: cancel_all runs under asyncio.wait_for with a five second budget, and a timeout logs a warning and falls through instead of propagating, so the rest of the finally block still runs. That is what the finding needed, since the deadlock risk was never the cancellation itself but the fact that nothing after it could execute, including the engine shutdown that closes the RPC router the cancelled request is waiting on.
> 
> The regression test reproduces the actual failure rather than approximating it. It registers a video task that hangs forever inside its own CancelledError handler, which is the wedged-orchestrator shape from the finding, and parametrizes normal against timeout. In both cases it asserts the full shutdown ordering completes, from HTTP shutdown through video and speech shutdown to socket close and engine exit, and it asserts the warning appears only in the timeout case. That ordering assertion is the one that matters, because it is what proves the timeout path does not skip cleanup.
> 
> No new findings. The earlier suggestion about peak memory, where the gate is reached only after all frames are materialized, still stands as a follow-up and is not a merge blocker.
> 
> Validation: static review of the delta against the round-three finding, the bounded wait and its fall-through path, and the new test's fixtures and assertions. GitHub checks pass. Per-PR Buildkite has never run on this branch, so I have added the ready label and the approval waits on those lanes. No fork code was executed.

THX for your comments and suggestions. Your suggestions on reducing the peak memory by optimizing the processing of the "all frames" is really good. It will be a key focus in the next stage.

linyueqian:
@alex-jw-brooks this is ready for another look. Your review from Aug 24 is still the open `CHANGES_REQUESTED` on this PR, and every point in it has been addressed since, so the branch is now blocked only on that dismissal.

Going through them against the current head `b14327b1`: the `tests/benchmarks/benchmark_video_response_event_loop.py` script is gone from the PR entirely, and so is the `recipes/MiniMaxAI/MiniMax-H3-NPU.md` summary, so the diff is down to four files (`api_server.py`, `serving_video.py`, `stores.py`, and the test module). The `media_utils` monkeypatching in the encode-path test is gone as well; the module is now a plain import and the test relies on the dimension ordering of `video` the way you suggested. The copied `fake_run` and per-test handler shutdown boilerplate is now a `video_handler` fixture plus `_make_video_handler`, `_video_artifacts`, and `_patch_video_artifacts` helpers, and the existing `test_raw_and_base64_encoders_receive_persistent_converter` was converted over to it. The standalone log-message test is removed; the only remaining `logger` patch is inside `test_server_worker_cancels_video_tasks_before_shutdown`, where the warning is asserted as behavior (it must fire on the stalled-cancellation parametrization and must not fire on the normal one) rather than as a log string.

Two review rounds landed after yours. The one that mattered was an unbounded `await VIDEO_TASKS.cancel_all()` in the shutdown `finally`, which could wedge if the RPC router closed only later; it is now wrapped in `asyncio.wait_for` with a 5.0s constant, the timeout is caught and logged, and the rest of the shutdown sequence still runs. The regression test asserts the full ordering `http_shutdown < video_task_cancelled < video_shutdown < speech_shutdown < socket_closed < engine_exit` in both the normal and the timeout case, which is what proves the timeout path does not skip cleanup.

Lanes at this head: general, Intel, and NPU green; AMD red only on `mi300_1: Diffusion / Model Test` with the SenseNova paged-decode ROCm flash-attn `ImportError` that fails identically on unrelated PRs and on main. This PR touches no diffusion code.

linyueqian:
The NPU red at `0da2d5c8` is not yours. Please do not spend time on it.

Your own NPU lane was green at `b14327b1` on 2026-09-03, with this same shutdown-cancellation code already in place. Nothing in your diff has changed since in a way that reaches these tests: you touch video serving, the server shutdown hook, `TaskRegistry.cancel_all`, and the video test file. What changed is the base. The four steps that fail are `Omni Model Test` and the three `MiniCPM-o 4.5` steps, while `Platforms Test`, `Diffusion NPU layer tests`, and the Qwen3-TTS step all pass in the same build, so this is not a generic server or shutdown break.

Cross-checking against other PRs supports that. NPU build 6570 on #7046 and 6575 on #6896 ran the identical MiniCPM-o steps in the same window and passed, and neither of their bases contains the eight main commits your base picked up. Yours is currently the only open PR whose base includes them.

Of those eight, four merged with only the general lane reporting and no NPU build at all: #5140, #6697, #6961, and #7016. #5140 centralizes Omni config resolution and #6697 moves model-specific helpers out of `serving_speech`, and both are in the path a multi-stage omni model like MiniCPM-o 4.5 depends on, so one of those two is the likely origin. I cannot read the Buildkite step logs without credentials, so I am giving you the isolation rather than the assertion.

On this PR: my approval at `b14327b1` still stands and the general lane is green at the current head. The remaining blocker is a `CHANGES_REQUESTED` from 2026-08-24 on commit `73cbeb41`, fifteen days and several heads ago, with two approvals stacked on top of it. That review needs dismissing or refreshing before this can merge.

herotai214:
@MosCloud — quick ask: could this wait until #5453 (P0.2 helper extract) merges?

#5453 is mid-rebase on `api_server.py` (helpers already moved out). Your ASGI/middleware hunk is still a valid `api_server.py` change, but landing #5453 first avoids another overlapping rebase on that file. Happy to help rebase if useful.

MosCloud:
> #5453 is mid-rebase on `api_server.py` (helpers already moved out). Your ASGI/middleware hunk is still a valid `api_server.py` change, but landing #5453 first avoids another overlapping rebase on that file.

Thanks for the heads-up. That ordering makes sense to us. Since #5453 moves the relevant helpers out of `api_server.py` and overlaps with the ASGI/shutdown lifecycle changes in #6579, letting it land first should allow us to adapt once against the final helper boundary and avoid repeated conflict resolution on the same file.

We are happy to hold #6579 for the requested 1-2 days. If #5453 lands within that window, we will rebase #6579 onto the latest `main`, adapt the lifecycle changes to the extracted helper structure, and rerun the focused validation. If it is still pending after that, we can ask the maintainers to decide the merge order once the review on #6579 is cleared.

Thanks for offering to help with the rebase.

MosCloud:
Rebased #6579 onto post-#5453 `main`. The only conflict was the constant block in `api_server.py`; the resolution keeps the extracted video constants in `video/generation/helpers.py` and `MAX_UINT32_SEED` in `diffusion.py`, while retaining the bounded video-task cancellation at the server lifecycle owner. The final PR footprint remains the same four files, and GitHub now reports the branch as mergeable.

Focused CPU validation loaded vLLM from the upstream `v0.29.0` source tag (`98dff2a8`) with the current PR sources: `122 passed`. Ruff check, Ruff format check, and `git diff --check` also passed. CI for the new head `da77b29d` is now queued.

@alex-jw-brooks, when convenient, could you refresh the outstanding review against this head?

vllm_omni/entrypoints/openai/api_server.py
```diff
@@ -191,6 +191,7 @@
 logger = init_logger(__name__)
 router = APIRouter()
 
+_VIDEO_TASK_CANCEL_TIMEOUT_S = 5.0
```
herotai214: Why do we declare here like that?
It looks weird to be placed between the APIRouters init;

We had VIDEO_SYNC_TIMEOUT_S imported from `vllm_omni.entrypoints.openai.video.generation.helpers`
Maybe it is a better place?

tests/entrypoints/openai_api/test_video_server.py
```diff
@@ -204,12 +474,103 @@ def isolated_video_backends(tmp_path, monkeypatch):
...
-    sock = FakeServerSocket()
+    sock = FakeServerSocket(events)
+
+    def record_warning(message, *args, **kwargs):
+        del kwargs
+        warning_messages.append(message % args)
+
+    monkeypatch.setattr(api_server.logger, "warning", record_warning)
+    if stalled_cancellation:
+        monkeypatch.setattr(api_server, "_VIDEO_TASK_CANCEL_TIMEOUT_S", 0.01)
```
herotai214: Should lives under vllm_omni.entrypoints.openai.video.generation.helpers instead of api_server?

MosCloud:
Rebased #6579 onto `main` at `64903b8e` after #7018 landed. The conflicts were in `serving_video.py` and `test_video_server.py`. The resolution preserves #7018 pre-encoded MP4 byte/metadata paths, while routing only legacy raw-frame MP4 encoding through PR3’s single-worker executor; PR3’s bounded shutdown ordering is unchanged. The final diff remains limited to the same four files.

Post-rebase validation: `129 passed` for `tests/entrypoints/openai_api/test_video_server.py`; Ruff check and format check passed; `git diff --check` passed. GitHub reports the branch as mergeable, and the new CI is queued.

@alex-jw-brooks, please refresh the outstanding review against head `acb01459` when convenient.

tests/entrypoints/openai_api/test_video_server.py
```diff
@@ -188,6 +186,275 @@ def test_preencoded_video_bytes_preserve_metadata(mocker: MockerFixture):
...
+        return artifacts
+
+    monkeypatch.setattr(handler, "_run_and_extract", fake_run)
+
+
+@pytest.mark.parametrize(
+    "video",
+    [
+        np.transpose(np.zeros((3, 2, 4, 4), dtype=np.uint8), (1, 2, 3, 0)),
+        np.zeros((2, 4, 6, 3), dtype=np.uint8),
```
linyueqian: [suggestion] The `legacy-fallback` parametrisation does not reach the legacy path. `_encode_video_bytes` passes `allow_strided_rgb_planes=frame_converter is not None and frame_converter.max_workers > 1` into `_direct_planar_fallback_reason`, and the handler fixture supplies the eight-worker converter, so the interleaved `(2, 4, 6, 3)` uint8 input is accepted by the direct-planar route just like the planar one; both ids exercise the same encoder path and the test would pass with the legacy fallback broken. Either build the handler for that case with a single-worker converter, or feed an input the planar route rejects (a non-uint8 dtype, or a strided view when strides are disallowed). This was already true at the head I approved, so it does not hold the merge.

linyueqian (approved):
Re-approving at `acb01459`, which is the head after the two rebases (post-#5453 and post-#7018). I read the `serving_video.py` delta against current `main` and it is the same design I approved at `b14327b1`, carried correctly across #7018: the non-streaming byte path and the base64 handler both submit the complete raw-frame encode to the dedicated single-worker executor through `_run_video_response_encoding`, pre-encoded `bytes`/`bytearray`/`memoryview` payloads bypass the encoder on both paths with their metadata preserved, mixed base64 batches keep output order by index, the admission slot is released from the native future's done callback via `call_soon_threadsafe` so a cancelled waiter cannot leak it, and `shutdown` clears the executor before joining it and then shuts the frame converter. Worker exceptions surface through the awaited future exactly as before. `api_server.py` adds only the bounded `VIDEO_TASKS.cancel_all()` ahead of handler shutdown, and `video/generation/helpers.py` is untouched.

On alex-jw-brooks' 2026-08-24 review, checked against this tree rather than the replies: the `media_utils` monkeypatching is gone (the route test wraps the real `_encode_video_bytes`), the standalone log-message test is gone, the repeated `fake_run` and handler setup are behind a shared fixture and two helpers, and the benchmark script is no longer in the diff. The one point still partial is entrypoint mocking in the shutdown test, which stubs app construction, the engine and HTTP serving and carries the requested integration TODO at line 666. Since that block predates all of this, it is for alex to lift or for a maintainer to dismiss; I am not asking for more test surgery here. One test-coverage nit from a second cross-model read is inline as a suggestion.

Two things before the merge rather than before the approve: no lane had run on this head (it was rebased with `ready` already on), so I cycled the label and 15265 is running; and `main` has moved 34 commits past the base with #7047 adding a test to the same `test_video_server.py`, so a rebase onto current `main` is needed for a clean merge. I will cycle `ready` again after that push and merge on green. Reviewed statically; fork head, no PR code executed.

tests/entrypoints/openai_api/test_video_server.py
MosCloud: Addressed in `b37e55cf`.

The parameterization now builds the handler with 8 frame-conversion workers for the direct-planar case and 1 worker for the legacy-fallback case. It also asserts the selected encoding route, so the interleaved HWC input is verified to take `legacy_fallback` instead of merely passing through the direct-planar path.

The rebase preserved the previously reviewed runtime patch. Final validation: `130 passed` for `tests/entrypoints/openai_api/test_video_server.py`; Ruff check/format and `git diff --check` passed.

linyueqian:
Sorry for the churn: #6759 landed on `main` twenty minutes after your rebase and it also edits `tests/entrypoints/openai_api/test_video_server.py`, so this branch is conflicting again. The conflict is test-only (both PRs add tests to the same file), the review is done and the approval for `b37e55cf` is registered on my side; one more rebase and I will cycle `ready` and merge on green. The other thing standing is alex-jw-brooks' 2026-08-24 CHANGES_REQUESTED, which predates every change discussed since; that needs their re-review or a maintainer dismissal, and I have flagged it to the committers.

linyueqian:
Correction to the note above: the conflicts are not test-only. #6759 changed `vllm_omni/entrypoints/openai/api_server.py` (the video DELETE path and the tracked-task cancellation before handler shutdown) and `vllm_omni/entrypoints/openai/serving_video.py` (job admission state), and both overlap with this PR's offload and shutdown hunks, so the rebase needs a real look at those two files rather than a mechanical one. The approach on each side is compatible: #6759 keeps jobs queued until scheduler admission and bounds the engine abort, this PR moves the encode off the loop and joins the encoding executor on shutdown; the merged `shutdown` should cancel tracked tasks, then shut the encoding executor, then the converter.

MosCloud:
Rebased onto current main at e9d27d87 after #6759.

This was not a mechanical conflict resolution. The merged lifecycle keeps #6759 queued-until-scheduler-admission and bounded DELETE abort behavior, while retaining the dedicated non-streaming response-encoding executor. Handler shutdown now has one canonical sequence: shut down and join the response-encoding executor, then shut down the frame-conversion pool. Server shutdown still bounds VIDEO_TASKS.cancel_all() at five seconds before handler shutdown.

Focused validation after the rebase: 134 passed for tests/entrypoints/openai_api/test_video_server.py, including the new #6759 queued/abort tests. Ruff check/format and git diff --check also passed.

linyueqian:
The re-fired lane at `e9d27d87` (build 15390) failed only `Simple · Engine&Entrypoints Test`, after 5m39s, which is the same signature as the `main` break tracked in #7626: this branch's base sits after `5dbae2b1` (#7608, the `consolidate_tensors` signature change) but before `99ff4f30` (#7413, which carries the snapshot-test fix), so the broken `tests/engine/test_output_metadata_snapshots.py` is in this tree and no build at this commit can pass that step. `main` has also moved under `vllm_omni/entrypoints/openai/api_server.py` since this head. One rebase onto current `main` (anything at or after `507cb1d8`) fixes both; a push onto a `ready` branch does not start a build by itself, so ping here and I will re-fire the lane. If the diff comes through unchanged the approval stands as is.

MosCloud:
@linyueqian Thanks for the diagnosis. I rebased the PR onto the latest `upstream/main` (`6bda549a`) and pushed the rewritten head `9a1d9615c`.

The rebase completed without unresolved conflicts, including the `api_server.py` changes. The final diff remains scoped to the video response lifecycle and route tests. Focused CPU tests passed (`551 passed, 7 skipped`), and Ruff, format, and diff checks are clean.

The GitHub checks are now running on the updated head. Please re-fire the Buildkite lane when convenient. The implementation intent is unchanged, so this should address the stale-base failure you identified.

PR #6580, opened 2026-08-24
[Bugfix][Diffusion] Release async outputs abandoned by consumer streams

rahul-steiger-nv:
## Purpose
Related to #6413 and a follow-up to #6439.
#6439 releases async diffusion outputs abandoned by aborted requests. Outputs can also be orphaned when the consumer stream disappears without an abort, leaving materialized image or video tensors cached indefinitely.
This PR tracks unclaimed output IDs per stream and releases them when the stream is missing, closed, cancelled, timed out, or the engine shuts down.

> **Dependency:** This PR is currently stacked on #6439 and will be rebased onto `main` after #6439 merges.

## Test Plan
Run the targeted CPU lifecycle tests:
```bash
python -m pytest -sv \
  tests/diffusion/test_diffusion_engine_cleanup.py \
  tests/diffusion/test_result_pump.py \
  -m "core_model and cpu"
```
Coverage includes missing and closed streams, cancellation during materialization, successful materialization, and normal completion without a consumer.
**vLLM Version:** 0.27.0
**vLLM-Omni Commit:** 26f9d128ba (based on #6439 at a91c94e717)
## Test Result
Targeted CPU lifecycle tests pass.
A broader `tests/diffusion` run reached an unrelated FlashAttention hardware error: `no kernel image is available for execution on the device`.

### DIFF
--- a/vllm_omni/diffusion/diffusion_engine.py
+++ b/vllm_omni/diffusion/diffusion_engine.py
@@ -1255,9 +1255,7 @@ class DiffusionEngine:
             # would be cached forever by the executor result pump (issue #6413).
             # Tell the executor to drop it before returning the aborted result.
             if runner_output is not None and runner_output.async_output_id is not None:
-                drop_output = getattr(self.executor, "drop_output", None)
-                if drop_output is not None:
-                    drop_output(runner_output.async_output_id)
+                self.executor.drop_output(runner_output.async_output_id)
             # Preserve runner-provided abort details when available.
             if runner_output is not None and runner_output.result is not None and runner_output.result.aborted:
                 return runner_output.result
--- a/vllm_omni/diffusion/diffusion_engine.py
+++ b/vllm_omni/diffusion/diffusion_engine.py
@@ -1251,6 +1251,13 @@ class DiffusionEngine:
             raise RuntimeError(f"Diffusion scheduler lost state for request {request_id}.")
 
         if state.status == DiffusionRequestStatus.FINISHED_ABORTED:
+            # An aborted request is never waited on, so a pending async output
+            # would be cached forever by the executor result pump (issue #6413).
+            # Tell the executor to drop it before returning the aborted result.
+            if runner_output is not None and runner_output.async_output_id is not None:
+                drop_output = getattr(self.executor, "drop_output", None)
+                if drop_output is not None:
+                    drop_output(runner_output.async_output_id)
             # Preserve runner-provided abort details when available.
             if runner_output is not None and runner_output.result is not None and runner_output.result.aborted:
                 return runner_output.result
--- a/vllm_omni/diffusion/diffusion_engine.py
+++ b/vllm_omni/diffusion/diffusion_engine.py
@@ -1352,6 +1352,13 @@ class DiffusionEngine:
             raise RuntimeError(f"Diffusion scheduler lost state for request {request_id}.")
 
         if state.status == DiffusionRequestStatus.FINISHED_ABORTED:
+            # An aborted request is never waited on, so a pending async output
+            # would be cached forever by the executor result pump (issue #6413).
+            # Tell the executor to drop it before returning the aborted result.
+            if runner_output is not None and runner_output.async_output_id is not None:
+                drop_output = getattr(self.executor, "drop_output", None)
+                if drop_output is not None:
+                    drop_output(runner_output.async_output_id)
             # Preserve runner-provided abort details when available.
             if runner_output is not None and runner_output.result is not None and runner_output.result.aborted:
                 return runner_output.result
--- a/vllm_omni/diffusion/diffusion_engine.py
+++ b/vllm_omni/diffusion/diffusion_engine.py
@@ -1356,9 +1356,7 @@ class DiffusionEngine:
             # would be cached forever by the executor result pump (issue #6413).
             # Tell the executor to drop it before returning the aborted result.
             if runner_output is not None and runner_output.async_output_id is not None:
-                drop_output = getattr(self.executor, "drop_output", None)
-                if drop_output is not None:
-                    drop_output(runner_output.async_output_id)
+                self.executor.drop_output(runner_output.async_output_id)
             # Preserve runner-provided abort details when available.
             if runner_output is not None and runner_output.result is not None and runner_output.result.aborted:
                 return runner_output.result
--- a/tests/diffusion/test_diffusion_engine_cleanup.py
+++ b/tests/diffusion/test_diffusion_engine_cleanup.py
@@ -334,3 +334,35 @@ def test_close_defers_resource_shutdown_until_worker_thread_stops() -> None:
     engine.executor.shutdown.assert_called_once()
     assert engine._shutdown_complete is True
     assert engine._loop_started is False
+
+def test_finalize_aborted_request_drops_pending_async_output() -> None:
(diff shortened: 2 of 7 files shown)


vllm_omni/diffusion/diffusion_engine.py
```diff
@@ -868,6 +794,30 @@ async def get_output_stream(self, request_id: str) -> AsyncGenerator[DiffusionOu
         try:
             while True:
                 output: DiffusionOutput = await queue.get()
+                async_output_id = getattr(output, "async_output_id", None)
```
fhfuih: `DiffusionOutput.async_output_id` is a real field (`__init__` sets the dict). `getattr(..., None)` only exists so `__new__`
  fakes/mocks don’t break; `_make_engine` already sets the dict. Use `output.async_output_id` directly.

fhfuih: Similar below for `self._unclaimed_async_outputs` etc.

vllm_omni/diffusion/diffusion_engine.py
```diff
@@ -868,6 +794,30 @@ async def get_output_stream(self, request_id: str) -> AsyncGenerator[DiffusionOu
...
+                if async_output_id is not None:
+                    fut = self.executor.wait_output_ready(async_output_id)
+                    try:
+                        output = await asyncio.wait_for(asyncio.wrap_future(fut), timeout=_ASYNC_OUTPUT_TIMEOUT)
```
fhfuih: On rebase onto current main, keep `_async_output_timeout()`

The wait you moved into get_output_stream still uses `_ASYNC_OUTPUT_TIMEOUT`. main already replaced that with `_async_output_timeout()`. Don’t restore the old constant when you unstack from #6439.

fhfuih (commented):
Thanks for identifying this problem and submit a PR.

So as I understand, you also move the waiting-for-output from `step_streaming` to `get_output_stream`. This also seems cleaner in terms of ownership (at least for me) because the stream yields materialized output, not a handle.

vllm_omni/diffusion/diffusion_engine.py
rahul-steiger-nv: I agree, updated!

vllm_omni/diffusion/diffusion_engine.py
rahul-steiger-nv: Sounds good! I will wait until #6439 merges and then use `_async_output_timeout()`

hsliuustc0106:
resolve conflicts

vllm_omni/diffusion/diffusion_engine.py
```diff
@@ -816,6 +827,9 @@ async def get_output_stream(self, request_id: str) -> AsyncGenerator[DiffusionOu
             with self._cv:
                 if self._out_streams.get(request_id) is queue:
                     self._out_streams.pop(request_id, None)
+                abandoned_ids = self._unclaimed_async_outputs.pop(request_id, set())
+            for async_output_id in abandoned_ids:
+                self.executor.drop_output(async_output_id)
 
     def async_add_req_and_stream_response(self, request: OmniDiffusionRequest) -> AsyncGenerator[DiffusionOutput, None]:
         request_id = self.add_request(request)
```
hsliuustc0106, line 926: [P1] Do not admit the request before its stream starts

`add_request()` runs before the returned async generator is first iterated. If a caller drops or closes that generator before its first `anext()`, the `get_output_stream()` body and its `finally` never run; the queue stays registered and the finished async output can still be cached indefinitely. Defer admission into an async-generator body (or another owner that closes before first iteration) and add an unstarted-stream regression.

vllm_omni/diffusion/diffusion_engine.py
```diff
@@ -806,6 +794,29 @@ async def get_output_stream(self, request_id: str) -> AsyncGenerator[DiffusionOu
...
+                    except (TimeoutError, asyncio.TimeoutError):
+                        self.executor.drop_output(async_output_id)
+                        describe = getattr(self.executor, "describe_pending_state", None)
+                        logger.error(
+                            "Timed out after %.0fs waiting for async output; executor state: %s",
+                            _ASYNC_OUTPUT_TIMEOUT,
+                            describe(async_output_id) if describe else "unavailable",
+                        )
+                        raise
+                    with self._cv:
```
hsliuustc0106: [P2] Retire this ID when the waiter completes exceptionally too

The ID is removed only after a successful await. If `OUTPUT_READY` completes the waiter with an exception, the pump has already popped it; generator cleanup then calls `drop_output()`, which registers a new placeholder for an output that already arrived, so it can never resolve. Retire the ID for every completed outcome and reserve `drop_output()` for cancellation or timeout where delivery may still be late; cover this with the real executor bookkeeping rather than a mock.

rahul-steiger-nv: The async output ID is now retired in a finally for every completed waiter outcome

vllm_omni/diffusion/diffusion_engine.py
rahul-steiger-nv: admission now happens inside the async-generator body

rahul-steiger-nv:
> resolve conflicts

I will resolve with a rebase once #6439 merges.

rahul-steiger-nv:
I will wait until #6439 merges.

vllm_omni/diffusion/diffusion_engine.py
```diff
@@ -894,6 +879,33 @@ async def get_output_stream(self, request_id: str) -> AsyncGenerator[DiffusionOu
...
             while True:
                 output: DiffusionOutput = await queue.get()
+                async_output_id = output.async_output_id
+                if async_output_id is not None:
+                    fut = self.executor.wait_output_ready(async_output_id)
+                    timeout = _async_output_timeout()
+                    try:
+                        output = await asyncio.wait_for(asyncio.wrap_future(fut), timeout=timeout)
+                    except asyncio.CancelledError:
+                        self.executor.drop_output(async_output_id)
```
lishunyang12: Skip drop_output for completed futures; cancellation after delivery recreates an orphaned waiter.

vllm_omni/diffusion/diffusion_engine.py
```diff
@@ -904,10 +916,20 @@ async def get_output_stream(self, request_id: str) -> AsyncGenerator[DiffusionOu
             with self._cv:
                 if self._out_streams.get(request_id) is queue:
                     self._out_streams.pop(request_id, None)
+                abandoned_ids = self._unclaimed_async_outputs.pop(request_id, set())
+            for async_output_id in abandoned_ids:
+                self.executor.drop_output(async_output_id)
 
-    def async_add_req_and_stream_response(self, request: OmniDiffusionRequest) -> AsyncGenerator[DiffusionOutput, None]:
+    async def async_add_req_and_stream_response(
```
lishunyang12, line 945: Update the admission test to iterate this generator before asserting preprocessing.

vllm_omni/diffusion/diffusion_engine.py
rahul-steiger-nv: updated

vllm_omni/diffusion/diffusion_engine.py
rahul-steiger-nv: updated

PR #6586, opened 2026-08-24
[Perf][Qwen3-TTS] Enable chunk ramp by default

Sy0307:
## Summary

- Enable `codec_chunk_ramp: [4, 4, 8, 16, 25]` by default for the Qwen3-TTS async deployment profiles.
- Apply the same default to `qwen3_tts.yaml` and `qwen3_tts_high_concurrency.yaml`.
- Add a configuration regression test for both profiles.

## Motivation

PR #5152 added the ramp as an opt-in feature to avoid the `1 -> 25` initial-to-steady chunk cliff. Our matched H200 run on the `seed-tts-eval` English workload (64 fixed requests, Stage 0/1 on separate H200 GPUs) showed the expected trade-off at concurrency 8:

| Configuration | p99 TTFP | p99 underrun | Request throughput |
| --- | ---: | ---: | ---: |
| Legacy `[1, 25, ...]` | 98 ms | 124 ms | 14.01 req/s |
| Ramp `[4, 4, 8, 16, 25]` | 160 ms | 68 ms | 14.10 req/s |

The ramp improves playback continuity without reducing measured throughput in this run, at the cost of a higher first-packet tail latency.

The default can be reverted in a deployment overlay with `codec_chunk_ramp: null`.

## Validation

- `git diff --check`
- YAML parsing verified for both modified deployment files
- H200 benchmark completed for baseline, ramp, initial chunk, and Stage 0/Stage 1 eager controls
- Full pytest was not runnable locally because the project dependency resolver currently rejects the existing Python 3.13 `transformers`/`qwen-asr` constraint combination; the targeted remote retry was unavailable due to intermittent SSH disconnects.

### DIFF
--- a/tests/config/test_config_factory.py
+++ b/tests/config/test_config_factory.py
@@ -1854,6 +1854,18 @@ class TestQwen3TTSPipeline:
         # Stage 1 uses its per-stage override
         assert stages[1].yaml_engine_args["model_arch"] == "Qwen3TTSCode2Wav"
 
+    @pytest.mark.parametrize(
+        "deploy_name",
+        ["qwen3_tts.yaml", "qwen3_tts_high_concurrency.yaml"],
+    )
+    def test_default_async_deploy_enables_chunk_ramp(self, deploy_name):
+        deploy = load_deploy_config(Path(get_deploy_config_path(deploy_name)))
+
+        assert deploy.async_chunk is True
+        assert deploy.connectors is not None
+        extra = deploy.connectors["connector_of_shared_memory"]["extra"]
+        assert extra["codec_chunk_ramp"] == [4, 4, 8, 16, 25]
+
     def test_subtalker_sampling_params_deep_merge_preserves_base_keys(self):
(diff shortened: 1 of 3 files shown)


linyueqian (approved):
lgtm

Sy0307:
Merge status: squash auto-merge is already enabled. The PR is approved and GitHub reports it as mergeable; GitHub checks, Intel/NPU/release/docs checks are green. The remaining blockers are required Buildkite `vllm-omni` #14239 and `vllm-omni-amd-ci` #11068, both failed. @linyueqian and maintainers, could you inspect or re-run those two lanes so auto-merge can proceed?

Sy0307:
Self-review on head `2c157fb`: this is a deployment/configuration change only. The default ramp is applied to the two Qwen3-TTS profiles, the config regression test covers both profiles, and the default-off override remains documented. The H200 A/B and targeted validation are recorded in the PR. The only remaining merge blockers are the failed required Buildkite lanes noted above.

PR #6485, opened 2026-08-22
[noissue] MiniCPM-o 4.5 Ascend qualification implementation

xiaojiu1414:
Team: noissue

This PR contains the reproducible MiniCPM-o 4.5 Ascend implementation
and official accuracy evidence for the competition qualification.

PR #6589, opened 2026-08-24
[noissue] MiniCPM-o 4.5 Ascend: Code2Wav execution-efficiency candidates (CFM 3-step + stage2 inductor compile)

xiaojiu1414:
Team: **noissue** (Gitee submission name: shijindashu)

## Summary

Execution-efficiency-only candidates for MiniCPM-o 4.5 on single-card Ascend 910C (Atlas A3), stacked on top of the qualification implementation in #6485 (`e3758833`). All changes are confined to scheduling / graph-mode / compilation configuration in `vllm_omni/deploy/minicpmo_4_5.yaml`. **No change to model behavior, sampling semantics, prompt/chat template, tokenizer, or the 24 kHz output rate.**

One variable per commit (branch history is the per-variable ledger):

| Commit | Change | Evidence |
|---|---|---|
| `9a3146b08` | Enable PIECEWISE cudagraph for stage 2 (Code2Wav) | Capture succeeded; replay padded batches tripped `seq_token_count_mismatch` (17/32 completed). Superseded by next commit. |
| `3e52826f7` | Keep stage-2 inductor compile, drop cudagraph replay | Same kernel fusion without graph padding; stable. |
| `47d770537` | Stack CFM refinement steps 4 -> 3 (`token2wav_n_timesteps: 3`) on top of the compile candidate | Single-variable prior evidence: median RTF 0.4214 vs baseline local median-of-3 0.4496 (diagnostic harness, N=32, first 2 warmups excluded). |

## Measured performance

All numbers are real runs from our `results/` archives; official-harness numbers come from the organizer environment.

- **Official harness** (organizer environment): `e3758833` scored **RTF 0.3843** and passed all four accuracy gates (see below). The stacked candidate `47d77053` has been submitted for official re-evaluation; expected ~0.358 by proportional scaling of the paired local A/B — **pending official write-back, treat as unconfirmed**.
- **Paired local A/B** (same day, same machine, official 32-prompt English Seed perf config, concurrency 1, 2 warmup requests excluded):

| Metric | e3758833 (local, median of 3 runs) | 47d77053 stack (paired run) |
|---|---:|---:|
| mean RTF | 0.4496 | 0.4138 (~8%) |
| median RTF | — | 0.3867 |
| TTFT p50 | — | 620 ms |

## Accuracy preservation

Official full-gate results at `e3758833` (organizer environment, all PASS):

| Benchmark | Score | Gate |
|---|---:|---|
| Daily-Omni | 78.11 | >= 77.5 |
| Video-MME | 69.56 | >= 67.0 |
| Seed-TTS Chinese CER | 1.055% | <= 1.56% relative gate |
| Seed-TTS ASV | 0.8772 | >= 0.689 |

Stacked candidate `47d77053`: official four-gate re-run pending submission write-back; local smoke (Seed English, N=10) WER 1.39% PASS. Full gates will be completed before final ranking submission per competition rules.

## Reproduce

- Image: `quay.io/ascend/vllm-omni:v0.25.0-a3`
- Hardware: single Ascend 910C, TP=1
- Config-only diff: see the three commits (final state touches only `vllm_omni/deploy/minicpmo_4_5.yaml`)
- Perf protocol: official 32-prompt Seed perf set, concurrency 1, first 2 warmup requests excluded, report p50/p90 separately; first-chunk and steady-state RTF reported separately

## Known issues (pre-existing at baseline, not introduced here)

- Stage-2 PIECEWISE cudagraph replay is not batch-shape-safe for this model (commit `9a3146b08` kept in-chain history for transparency; effectively disabled by `3e52826f7`).
- First measured request pays one-time torch.compile shape compilation (~55 s wall clock); a startup-warmup hardening candidate addressing this is validated separately and not part of this PR.
- Benchmark client teardown can abort with `corrupted size vs. prev_size` after results are persisted; present at baseline, untouched here.

## Relationship to #6485 (stacked PR)

The first ~20 commits of this branch are identical to the qualification implementation in #6485 (`e3758833`), included because this branch stacks directly on top of it and #6485 is not yet merged. **The new content of this PR is the last three perf commits (`9a3146b08`, `3e52826f7`, `47d770537`) plus one upstream-sync merge (`f7a95c102`, no source changes).** The net source diff vs `minicpm-challenge` beyond #6485 is config-only (`vllm_omni/deploy/minicpmo_4_5.yaml`). If #6485 merges first, this PR reduces to those three commits.

Note: upstream `a964efc55` (#6447) switched the CI perf Seed-TTS set to Chinese; performance numbers above are quoted from organizer-harness write-backs and our paired local A/B runs as labeled, not re-asserted against a specific prompt-language protocol.

### DIFF
--- a/vllm_omni/deploy/minicpmo_4_5.yaml
+++ b/vllm_omni/deploy/minicpmo_4_5.yaml
@@ -12,7 +12,7 @@ connectors:
       connector_get_sleep_s: 0.001
       connector_get_max_wait_first_chunk: 3000
       connector_get_max_wait: 300
-      codec_chunk_frames: 25
+      codec_chunk_frames: 50
       codec_left_context_frames: 3
       # Experimental performance candidate: fewer CFM refinement steps.
       # Keep this isolated from the stable accuracy-validated deployment.
--- a/vllm_omni/deploy/minicpmo_4_5.yaml
+++ b/vllm_omni/deploy/minicpmo_4_5.yaml
@@ -16,7 +16,7 @@ connectors:
       codec_left_context_frames: 3
       # Experimental performance candidate: fewer CFM refinement steps.
       # Keep this isolated from the stable accuracy-validated deployment.
-      token2wav_n_timesteps: 6
+      token2wav_n_timesteps: 4
 
 stages:
   - stage_id: 0
--- a/vllm_omni/deploy/minicpmo_4_5.yaml
+++ b/vllm_omni/deploy/minicpmo_4_5.yaml
@@ -14,6 +14,9 @@ connectors:
       connector_get_max_wait: 300
       codec_chunk_frames: 25
       codec_left_context_frames: 3
+      # Experimental performance candidate: fewer CFM refinement steps.
+      # Keep this isolated from the stable accuracy-validated deployment.
+      token2wav_n_timesteps: 8
 
 stages:
   - stage_id: 0
--- a/vllm_omni/deploy/minicpmo_4_5.yaml
+++ b/vllm_omni/deploy/minicpmo_4_5.yaml
@@ -103,3 +103,10 @@ platforms:
         max_num_batched_tokens: 8192
         compilation_config:
           cudagraph_mode: PIECEWISE
+      # Execution-efficiency candidate: Code2Wav (CFM + HiFT) previously ran
+      # fully eager while Thinker/Talker used piecewise graphs. Same math,
+      # graph-captured dispatch. Must re-pass accuracy gates before submission.
+      - stage_id: 2
+        enforce_eager: false
+        compilation_config:
+          cudagraph_mode: PIECEWISE
--- a/vllm_omni/deploy/minicpmo_4_5.yaml
+++ b/vllm_omni/deploy/minicpmo_4_5.yaml
@@ -109,4 +109,4 @@ platforms:
       - stage_id: 2
         enforce_eager: false
         compilation_config:
-          cudagraph_mode: PIECEWISE
+          cudagraph_mode: NONE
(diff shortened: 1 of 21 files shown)


vllm_omni/deploy/minicpmo_4_5.yaml
```diff
@@ -9,11 +9,14 @@ connectors:
...
-      connector_get_sleep_s: 0.01
+      connector_get_sleep_s: 0.001
       connector_get_max_wait_first_chunk: 3000
       connector_get_max_wait: 300
-      codec_chunk_frames: 25
+      codec_chunk_frames: 50
       codec_left_context_frames: 3
+      # Experimental performance candidate: fewer CFM refinement steps.
+      # Keep this isolated from the stable accuracy-validated deployment.
+      token2wav_n_timesteps: 3
```
congw729, line 19: From the current configuration, it looks like `token2wav_n_timesteps: 3` is placed in the shared connector section, so it may apply to CUDA/H100 as well as Ascend. Is this cross-platform behavior expected? My understanding is that reducing the CFM steps changes the generated-audio trajectory, while the PR describes this as an Ascend execution-efficiency candidate. If the optimization is intended only for Ascend, perhaps we could place it in an Ascend-specific deployment profile or override to avoid changing the validated defaults on other platforms.

vllm_omni/deploy/minicpmo_4_5.yaml
```diff
@@ -100,3 +103,10 @@ platforms:
         max_num_batched_tokens: 8192
         compilation_config:
           cudagraph_mode: PIECEWISE
+      # Execution-efficiency candidate: Code2Wav (CFM + HiFT) previously ran
+      # fully eager while Thinker/Talker used piecewise graphs. Same math,
+      # graph-captured dispatch. Must re-pass accuracy gates before submission.
```
congw729, line 108: From the configuration below, it looks like Stage 2 uses `enforce_eager: false` together with `cudagraph_mode: NONE`, which appears to enable compilation while disabling graph replay. Is that the expected execution mode here? The comment currently mentions “graph-captured dispatch,” so I wonder whether it could be updated to say “inductor compilation without graph replay” to avoid confusion during future profiling and maintenance.

congw729 (commented):
Thanks for putting together these MiniCPM-o performance experiments! I left two questions about the effective configuration scope and execution-mode description.

PR #6590, opened 2026-08-24
Feature/Irodori-TTS diffusion TTS support

blazingbhavneek:
# [Model] Add Irodori-TTS diffusion TTS support

Adds native [Irodori-TTS](https://github.com/Aratako/Irodori-TTS) inference to vLLM-Omni's diffusion engine, served over `/v1/audio/speech`. Benchmarked against the reference [Irodori-TTS-Server](https://github.com/Aratako/Irodori-TTS-Server).

## Purpose

Model-specific code is isolated under `vllm_omni/diffusion/models/irodori_tts/` (13 files, 7,018 of the diff's 8,403 inserted lines). Shared runtime changes are narrow: a registry entry, two guarded capability hooks in `diffusion_model_runner.py` (fused denoise step, batched request setup — pipelines that don't opt in keep the original per-request path), `extras` passthrough in stage config, and a generic diffusion-TTS adapter route in `serving_speech.py`. No other model is touched.

Not ported: SSE/streaming, word timestamps, voice storage, auto-chunking (30s request cap), non-WAV/PCM output, multi-candidate generation, advanced CFG/sway controls.

## Test Plan

**vLLM:** 0.26.0 (pinned release); local dev used vLLM `main` + 2 out-of-tree re-export shims, not part of this diff.
**vLLM-Omni:** rebased onto `main` @ `678aa258`.

8 server configs (2 models × 2 precisions, ours vs. official) benchmarked in one session on an RTX 5060 Ti 16GiB.

```bash
CUDA_VISIBLE_DEVICES="" python -m pytest -q \
  tests/diffusion/models/test_irodori_tts.py \
  tests/diffusion/models/irodori_tts/test_irodori_packed_batching.py \
  tests/diffusion/test_diffusion_model_runner.py \
  tests/entrypoints/openai_api/test_tts_adapter.py \

VLLM_OMNI_IRODORI_TEST_MODEL=<checkpoint-dir> python -m pytest -q -s \
  tests/diffusion/models/irodori_tts/test_irodori_batching_e2e.py
```

## Test Result

All 72 throughput cells and 24 parity renders completed with zero failures and exact sample counts. 187 unit/contract tests pass (2 pre-existing failures reproduce identically on bare `main`, unrelated to this diff).

**Speedup vs. official server**, single request, generated-audio-seconds/wall-second (`Irodori-TTS-v4-Small`; `v4.1-Small` tracks within run-to-run noise — the checkpoints differ only in `duration_predictor`):

| Precision | 4s | 10s | 30s |
|---|---:|---:|---:|
| FP32 | 2.31x | 2.08x | 2.07x |
| BF16 | 1.46x | 1.73x | 1.55x |

**Waveform parity** (correlation vs. official FP32 reference):

| Implementation | 4s | 10s | 30s |
|---|---:|---:|---:|
| vLLM-Omni FP32 | 0.9999 | 0.9980 | 0.9948 |
| vLLM-Omni BF16 | 0.9978 | 0.8595 | 0.8456 |
| Official BF16 | 0.9875 | 0.8834 | 0.9101 |

BF16 divergence at longer durations shows up identically on both implementations (rounding compounding under free-running continuation on a short benchmark prompt) — not a vLLM-Omni-specific gap. 60/60 concurrent HTTP renders (30 FP32 + 30 BF16) were bit-identical to their serial equivalent — no cross-request contamination under load.

**Multi-request batching** (offline `AsyncOmni`, the only route that can set `diffusion_batch_size` today), 8×4s requests, BF16:

| Batch | 1 | 2 | 4 | 8 |
|---|---:|---:|---:|---:|
| Speedup | 1.00x | 1.47x | 1.81x | 1.89x |

Saturates by batch 4 — the DiT step goes from 40.5% to 74.1% of this GPU's measured GEMM peak across that range, so the ceiling is remaining GPU headroom, not scheduling.

**Concurrency caveat — server args, not fixed in this PR:** `vllm serve` currently pins the diffusion stage to 1 request/step regardless of `--max-num-seqs`, a pre-existing shared-runtime gap. The numbers above are single-request and are what this PR actually ships. The fix (`max(batch_size, od_config.max_num_seqs or 1)` in `vllm_omni/engine/`) was verified working end-to-end (concurrency 8: 0.97x → 2.19x) but touches every diffusion model's startup, so it belongs in its own PR. Tracked at [Aratako/Irodori-TTS#39](https://github.com/Aratako/Irodori-TTS/issues/39).

## Next steps

- Land the server-args fix above in its own PR against `vllm_omni/engine/`.
- Root-cause a batch-8 GPU-utilization discrepancy noticed while rebasing (74.1% now vs. 88% in an earlier pre-rebase measurement on the same GPU); the batching ratio itself barely moved, so likely not a regression, but unconfirmed.
- Manual code review pass — see Disclaimer.

## Disclaimer

Most of this port (model, sampler, pipeline, batching) was AI-generated, with a human directing scope and reviewing results rather than authoring line-by-line. Test coverage (CPU contracts, runner dispatch contracts, GPU end-to-end batching/parity vs. the official server) is thorough, which is why confidence in correctness is reasonably high despite limited manual review — a manual read-through for style and further speed/memory improvements is planned as follow-up.


vllm_omni/diffusion/models/irodori_tts/pipeline_irodori_tts.py
```diff
@@ -0,0 +1,1041 @@
...
+    @staticmethod
+    def _condition_row(condition: IrodoriConditionState, index: int) -> IrodoriConditionState:
+        """Copy one request's rows out of a batched condition state.
+
+        The copy is deliberate: a view would keep the whole group's encoder
+        output alive for as long as the slowest request in it.
+        """
+
+        def row(value: torch.Tensor | None) -> torch.Tensor | None:
+            return None if value is None else value[index : index + 1].contiguous()
```
lishunyang12, line 677: Use clone here; contiguous slices still retain the entire batch's storage.

vllm_omni/diffusion/models/irodori_tts/model.py
```diff
@@ -0,0 +1,2490 @@
...
+        )
+        h, mlp_gate = self.mlp_adaln.forward_packed(
+            x,
+            cond_embed,
+            token_sequence_indices,
+        )
+        return x + self.dropout(mlp_gate * self.mlp(h))
+
+
+class DurationPredictor(nn.Module):
```
lishunyang12, line 1226: Move duration prediction and condition encoders into separate modules to keep this file manageable.

PR #6591, opened 2026-08-25
[Diffusion] Register HWR mmap for direct DLO H2D

hsliuustc0106:
## Summary

Adds PR3 on top of merged #6486:

- registers an exact final-layout HWR lease mmap for direct asynchronous H2D in no-AllGather DLO;
- retains PR2's bounded two-slot pinned staging path when registration is disabled, unsupported, or over budget;
- guarantees all-or-nothing registration, retryable unregistration, and unregister-before-lease-close teardown;
- restores worker shutdown ordering so the offloader is disabled before distributed state is destroyed;
- adds the representation-neutral writer/fsync/checksum and source-digest optimizations intentionally omitted from #6486.

AllGather and direct-checkpoint mmap behavior are unchanged.

## Design

```text
warm HWR hit
  -> backend takes HostWeightLease
  -> register lease.mapped_regions under pin_cpu_memory policy
     -> success: mmap tensor views -> existing rotating HBM buffers
     -> safe failure: existing two-slot pinned staging -> HBM buffers
  -> GPU kernels
```

`--dlo-host-registration-limit-gib` is an optional per-worker preflight ceiling over the complete page-aligned registration. Zero adds no ceiling. A rollback/unregistration failure is fatal and retains the registration plus lease; the backend never unmaps storage still owned by CUDA.

Registration is selected only for the final-layout HWR no-AllGather consumer. Checkpoint mmap may still require per-block loader transforms and remains staged. HWR-disabled, DLO-disabled, and AllGather configurations retain zero HWR registration interaction.

The generic HWR I/O changes are separate from transport semantics:

- ordered producers hash payload files and tensors while writing;
- payload fsync/checksum work overlaps later producer work;
- unordered producers retain a parallel readback checksum fallback;
- local canonical source shards are hashed in parallel and reuse node-local, metadata-validated digest records.

Manifest/READY publication still waits for every checksum and fsync.

## Validation

Current head: `49a48ae4`. GPU A/B evidence is frozen at `8984053c`; later commits harden teardown/error paths and stabilize tests without changing the successful registered or staged transfer loops.

### CPU and static

- Focused HWR/loader/offloader/worker/CLI suite: **238 passed**.
- CUDA-registration and DLO-backend regression files after the latest review fixes: **88 passed**.
- Changed-file pre-commit hooks passed (repository CI skips the existing Python 3.10 mypy hook).
- `git diff --check origin/main` passed.
- DCO, pre-commit, Python 3.11/3.12 wheel builds, docs, and release checks pass.

### Recipe-resolution one-GPU denoise A/B

Decision workload: MiniMax-H3 FL2VA, one L20X GPU, the documented `1344x768` 16:9 shape, four seconds, TP1/DP1/SP1, no-AllGather DLO, CUDNN attention, eager mode, and `dlo_resident_layers=0`. MiniMax-H3 counts sigma points, so `num_inference_steps=3` produces the requested **two actual DiT evaluations**; a value of 2 produces only one evaluation.

The same GPU ran staged -> registered and registered -> staged. Each engine received one excluded warmup, three measured requests, and one profiled request. A full-path feasibility request spent more than six minutes in single-GPU tiled VAE decode after the roughly ten-second DiT loop, so this diagnostic returns and hashes final video/audio latents without VAE decode. Prompt encoding, recipe-shape latent construction, all 52 DLO blocks, and both denoise evaluations remain active.

Request-level direction changed with order: staged -> registered measured `12.233 -> 12.126 s` (-0.9%), while registered -> staged measured `11.947 vs 11.488 s` (+4.0% for registration). Therefore this matrix shows **no repeatable request-level latency gain**; non-DiT temporal variation is larger than the exposed transport effect.

The DLO-specific profile is stable across both orders (four step observations per mode):

| Metric, per actual denoise step | Bounded staging | Registered mmap | Difference |
| --- | ---: | ---: | ---: |
| DLO H2D payload | 63.479 GiB | 63.479 GiB | unchanged |
| DLO H2D operations | 54 packed copies | 534 tensor copies | +480 |
| Summed DLO H2D device time | 1.251 s | 1.275 s | +1.9% |
| CPU `aten::copy_` inside DLO span | 0.979 s | 0.069 s | -0.910 s (-93.0%) |
| DLO copy-stream span | 4.757 s | 4.700 s | -57.7 ms (-1.2%) |

All four registered step spans were shorter by 45-81 ms. Compute was unchanged (9.427/9.417 s for two evaluations), engine-reported peak HBM was 10,794 MiB in both modes, and process-tree PSS was 131.06/127.06 GiB (-4.01 GiB). All six staged and six registered measured latent hashes matched exactly.

The attribution is per block: every denoise evaluation traverses 52 hooked blocks, and each hook prefetches the next block. Registration removes about 17.5 ms of CPU copy work per streamed block invocation on average, but only about 1.11 ms/block reaches the copy-stream critical path because high-resolution compute overlaps most staging work. The benefit therefore recurs per nonresident DiT layer on every denoise step; it is not one fixed operation at the step boundary.

### Small-shape transport stress A/B (`256x256`)

Controls: MiniMax-H3 FL2VA, the same 61.73 GiB warm HWR artifact, TP1/DP1/SP1, no-AllGather DLO, CUDNN attention, eager mode, 256x256, four seconds, `num_inference_steps=2` (one actual DiT evaluation under H3's sigma-point semantics), fixed prompt/seed, two warmups, ten measured waves, and one profiled wave. The engines were co-resident on disjoint GPUs and requests alternated AB/BA. All CUDA work ran under `gpu run`; a foreign-PID guard excluded contaminated attempts. This is a transfer-exposure stress test, not the recipe-resolution decision workload.

The staged control used a 1 MiB registration budget, selecting bounded pinned staging during preflight before any CUDA registration.

| Metric | Bounded staging | Registered mmap | Difference |
| --- | ---: | ---: | ---: |
| Mean request latency, 10 pairs | 4.5630 +/- 0.2433 s | 4.1279 +/- 0.1663 s | -435.1 ms (-9.5%) |
| Paired 95% CI |  |  | [-612.1, -258.1] ms; registered won 9/10 |
| Derived sequential throughput | 0.219 req/s | 0.242 req/s | +10.5% |
| CPU `aten::copy_`, profiled wave | 2.888 s | 1.883 s | -34.8% |
| H2D payload, profiled wave | 121.855 GiB | 121.855 GiB | unchanged |
| H2D device time, profiled wave | 2.878 s | 2.553 s | -11.3% |
| Compute-kernel time, profiled wave | 0.404 s | 0.402 s | unchanged |
| H2D copy count, profiled wave | 1,813 | 2,293 | +480 tensor-level direct copies |
| NCCL / measured NVLink traffic | none / 0 | none / 0 | unchanged |
| Peak HBM | 13,300 MiB | 13,300 MiB | unchanged |
| Sequential process-tree PSS smoke | 132.23 GiB | 128.23 GiB | -3.99 GiB |
| Video/audio hashes | reference | exact match, 10/10 | pass |

The small-shape result supports the mechanism when compute exposes transfer latency; it must not be generalized to recipe-resolution request latency. Registration removes recurrent mmap-to-private-pinned CPU staging but does **not** reduce H2D bytes or HBM. Profile durations are summed engine work and may overlap, so they are not wall time.

The registered worker logged 61.73 GiB in 13 ranges and unregistered before releasing its HWR lease. The PSS smoke was collected at `071c5acf`; later commits only removed private aliases and updated module design, with no transport-path change.

Registration is one-time startup work, not a startup optimization. The paired run observed 198.54/205.08 s staged/registered engine initialization, but startup order and shared host state were not controlled tightly enough to claim an exact penalty.

### Generic HWR startup-path A/B

Compared merged PR2 (`8360346a`) with this head on the same filesystem and NUMA node:

| Benchmark | Merged PR2 | PR3 | Result |
| --- | ---: | ---: | ---: |
| Publish 1 GiB as four ordered files, 3 reps | 2.873 s | 2.164 s | -24.7% wall time |
| Publication CPU time | 2.793 s | 2.302 s | -17.6% |
| Hash unchanged 1 GiB source, warm mean, 3 reps | 0.982 s | 0.0031 s | 321x faster |
| Hash 1 GiB source, first PR3 lookup | 0.982 s baseline | 0.258 s | -73.7% |
| MiniMax-H3 61.73 GiB / 13-shard source identity | 95.23 s direct hash | 7.57 ms warm lookup | exact revision match |

Synthetic source identities matched exactly. The real-model row measures PR2 direct hashing versus a metadata-validated PR3 warm digest; it is a cache-hit result, not a cold-cache comparison.

### TP2 feasibility and performance status

The earlier reserved-GPU TP2/DP1/SP1 feasibility run remains valid: both coordinate artifacts skipped ordinary transformer loading, each rank registered 30.86 GiB in seven ranges, inference completed with valid video/audio, peak HBM was 11,884 MiB/GPU, and both ranks unregistered before releasing their leases.

A repeated TP2 A/B was attempted at the current head, but an unrelated unreserved process entered GPU 3 after both ranks initialized and before the first warmup. The guard aborted and cleanup completed. No latency, memory, NCCL, or throughput number from that run is used, so this PR does not claim a repeated TP2 performance result.

## Scope

- BF16 final-layout HWR, no-AllGather DLO.
- No FP8 policy or producer changes.
- No AllGather HWR consumer.
- No checkpoint-mmap registration.
- No DP request orchestration.

Relates to #6414. Replaces the reusable registration/transport portion of closed #6240. Builds on #6486.


hsliuustc0106:
Self-review at `8984053c`: I traced config propagation, HWR lease handoff, registration capability/budget handling, direct and staged copy paths, partial rollback, retryable unregistration, and worker shutdown ordering. The review found one real gap—normal worker shutdown skipped offloader teardown—and fixed it before hardware validation.

Validation: 238 focused CPU tests and all changed-file hooks pass. On L20X TP1, three measured registered requests were byte-identical to staging and averaged 3.645 ± 0.006 s versus 4.882 ± 0.111 s, with unchanged 13,300 MiB HBM and 3.99 GiB lower process-tree PSS. On TP2, both coordinate artifacts skipped ordinary transformer loading, each rank registered 30.86 GiB in seven ranges, inference completed, and both ranks explicitly unregistered before releasing their leases. No task-owned GPU process remained.

I found no remaining correctness blocker. The draft still does not claim an Nsight H2D-payload profile or production throughput confidence; the TP2 cold population also reproduced the existing 30-second Orchestrator reap timeout after both workers had already disabled their backends and shut down.

hsliuustc0106:
> **Update:** the recipe-resolution follow-up in [this comment](https://github.com/vllm-project/vllm-omni/pull/6591#issuecomment-5404771864) supersedes any broad request-latency interpretation below. The `256x256` matrix is a transfer-exposure stress test; `num_inference_steps=2` also represents one actual MiniMax-H3 DiT evaluation under the current sigma-point semantics.

Required A/B validation is complete at frozen head `8984053c` and the PR description now contains the full experiment contract and results.

For TP1 no-AllGather DLO, ten fixed-seed staged/registered pairs produced:

- 4.5630 vs 4.1279 s mean request latency: **-435.1 ms (-9.5%)**, paired 95% CI `[-612.1, -258.1]` ms; registered won 9/10;
- exact video/audio parity for all ten pairs and unchanged 13,300 MiB peak HBM;
- CPU `aten::copy_` 2.888 -> 1.883 s (-34.8%);
- identical 121.855 GiB H2D payload, H2D device time 2.878 -> 2.553 s, and stable 0.404/0.402 s compute time;
- sequential PSS smoke 132.23 -> 128.23 GiB (-3.99 GiB).

This supports the intended mechanism: registration removes the recurrent mmap-to-private-pinned CPU copy; it does not reduce H2D bytes or HBM. Registration remains one-time startup work, so no startup-latency improvement is claimed.

The generic PR3 paths also measured 2.873 -> 2.164 s (-24.7%) for publishing a 1 GiB ordered artifact and 0.982 s -> 3.1 ms (321x) for a warm 1 GiB source digest. A real 61.73 GiB MiniMax-H3 source identity changed from 95.23 s direct hashing to a 7.57 ms metadata-validated warm lookup with the exact same revision.

A repeated TP2 performance A/B is intentionally **not** claimed: an unrelated unreserved process entered reserved GPU 3 after both ranks initialized and before the first warmup. The guard aborted, and both ranks unregistered, released leases, and shut down cleanly. The earlier TP2 functional run remains the TP2 correctness evidence; no result from the intruded attempt is used.

hsliuustc0106:
Recipe-resolution follow-up at `8984053c` changes the latency interpretation of my earlier small-shape comment.

First, MiniMax-H3's `num_inference_steps` currently counts sigma points: `2` produces one actual DiT evaluation. For the requested two denoise evaluations, this follow-up uses `num_inference_steps=3`. The PR description has been corrected accordingly.

I ran a same-GPU, order-reversed staged/registered diagnostic at the recipe's documented `1344x768` 16:9 shape, four seconds, TP1/DP1/SP1, `dlo_resident_layers=0`, with one warmup, three measurements, and one profile per engine. A full-path feasibility request spent more than six minutes in single-GPU tiled VAE decode after the roughly ten-second DiT loop, so the decision diagnostic returns and hashes final latents without VAE decode; it retains prompt encoding, recipe-shape construction, all 52 DLO blocks, and both denoise evaluations.

The request-level direction reversed with order:

- staged -> registered: `12.233 -> 12.126 s` (-0.9%);
- registered -> staged: `11.947 vs 11.488 s` (+4.0% for registration).

Therefore I do **not** claim a repeatable request-level latency gain at recipe resolution. The DLO-specific profile is stable across both orders, however:

| Per actual denoise step | Staged | Registered |
| --- | ---: | ---: |
| DLO H2D payload | 63.479 GiB | 63.479 GiB |
| CPU `aten::copy_` inside DLO span | 0.979 s | 0.069 s (-93.0%) |
| DLO copy-stream span | 4.757 s | 4.700 s (-57.7 ms) |
| DLO H2D operations | 54 packed | 534 tensor-level |
| Summed DLO H2D device time | 1.251 s | 1.275 s |

All four registered step spans were 45-81 ms shorter. Two-step compute and peak HBM were unchanged; process-tree PSS fell 4.01 GiB; all six staged and six registered measured latent hashes matched exactly.

The architectural attribution is **per streamed DiT block**: each of the 52 hooks prefetches the next block, and the sequence repeats on every denoise evaluation. On average, registration removed about 17.5 ms of CPU copy work per block invocation, but only about 1.11 ms/block was exposed on the DLO stream because high-resolution compute overlaps most of it. The earlier `256x256` result remains useful as a transfer-exposure stress test, not as the recipe-resolution serving claim.

hsliuustc0106:
DP2/TP1 no-AllGather host-memory matrix at `90506c6e` (same host and physical GPUs 4/5 for every accepted case; CUDNN attention, eager, `dlo_resident_layers=0`, fixed 256x256 / 4-second T2VA requests, `num_inference_steps=2`, two concurrent requests).

The cold `preferred` startup is excluded from the measurements; it only populated the warm 61.73 GiB HWR artifact.

| Case | Backing / transport | Host staging | Peak process-tree PSS | Peak Private_Dirty |
| No HWR | checkpoint mmap | 2 x 1.231 GiB slots per worker | 1.656 GiB | 1.265 GiB |
| HWR staged | warm HWR hit + 1 MiB registration budget fallback | 2 x 1.231 GiB slots per worker | 1.661 GiB | 1.252 GiB |
| HWR registered | warm HWR hit + direct H2D | no staging buffers allocated | 1.624 GiB | 1.235 GiB |

The registered case logged 61.73 GiB in 13 ranges on each worker, completed both requests, unregistered both mappings before lease release, and left no active children. The staged case also completed and shut down cleanly.

This confirms that HWR staging does not materially change PSS relative to the existing checkpoint-mmap baseline. Registration removes the configured staging capacity (about 4.925 GiB across two DP workers), but this run does not support claiming a 4.9 GiB process-PSS reduction: registered PSS was about 1.9% below no HWR. `smaps_rollup` Locked stayed at roughly 16 KiB in all cases, so it is not a sufficient metric for CUDA-registered physical host pages.

vllm_omni/diffusion/offloader/cuda_host_registration.py
```diff
@@ -0,0 +1,244 @@
...
+        return str(error)
+    if isinstance(message, bytes):
+        return message.decode(errors="replace")
+    return str(message)
+
+
+def _consume_last_cuda_error(expected_error: int) -> None:
+    """Clear one handled CUDA Runtime error before returning to PyTorch."""
+    try:
+        get_last_error = ctypes.CDLL(None).cudaGetLastError
```
hsliuustc0106: `_consume_last_cuda_error` resolves `cudaGetLastError` through `ctypes.CDLL(None)`, but libcudart is usually loaded with local scope in pip-wheel deployments, so this lookup can raise `AttributeError` — and `_handled_error_message` then converts that into `HostRegistrationCleanupError`. Since that is a `HostWeightRegistrationError` subclass, `create()` re-raises it at the bare `raise` after a **clean** rollback, and `_try_register_hwr_mmap` classifies it as rollback-failed → fatal startup instead of the documented staging fallback. Concrete trigger: pinning ~62 GiB under host-memory pressure fails mid-way with `cudaErrorMemoryAllocation`, rollback unregisters everything cleanly, then message-building fails to clear the pending error (the `pending_error not in (0, expected)` strictness also escalates any pre-existing stray CUDA error). The register-failure tests monkeypatch `_consume_last_cuda_error` away and the dedicated test stubs `ctypes.CDLL`, so this path has never run against a real runtime. Smallest fix: prefer `runtime.cudaGetLastError` (the same cudart handle already supplying `cudaGetErrorString`) with `CDLL(None)` as fallback; optionally, when rollback completed with zero rollback errors, demote to plain `HostRegistrationError` so the documented fallback applies.

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
```diff
@@ -916,6 +980,126 @@ def offload_resident_layers(self) -> None:
...
+
+    def _release_registered_mmap(self) -> None:
+        """Release every platform registration before closing the HWR lease."""
+        registration = self._host_registration
+        if registration is None:
+            return
+        errors = registration.close()
+        if errors:
+            logger.error("HWR mmap unregistration failed; retaining lease mappings for retry: %s", errors[:3])
+            raise HostRegistrationCleanupError(f"failed to unregister {len(errors)} HWR mmap range(s)")
```
hsliuustc0106, line 1139: `_release_registered_mmap` correctly retains the failed registration plus the open lease, but nothing keeps that pair alive: in a preferred-mode warm startup, an unregistration failure propagates through `enable()`, and `enable_offload_backend`'s fresh retry (`offloader/__init__.py` `backend = get_offload_backend(...)`) rebinds `backend`, dropping the old backend object. Its `HostWeightLease.__del__` (`lease.py:138`) then closes the lease and unmaps pages CUDA still has registered — violating the stated "backend never unmaps storage still owned by CUDA" invariant. Consequence is a pinned-page leak until process exit rather than corruption (pages stay pinned by the driver), and it requires a persistently failing `cudaHostUnregister`, so this is hardening rather than a blocker. Suggestion: retain `(registration, lease)` in a module-level list in both cleanup-failure retention branches so neither is finalizer-closed while a registration is live.

hsliuustc0106 (commented):
Self-review at final head `90506c6e`, findings in the two inline threads:

- **P2** — `_consume_last_cuda_error` resolves `cudaGetLastError` through `ctypes.CDLL(None)`, which can fail to resolve on pip-wheel deployments; that escalates a *cleanly rolled back* `cudaHostRegister` failure (e.g. pinning ~62 GiB under host-memory pressure) into fatal startup instead of the documented staging fallback. Fix: prefer `runtime.cudaGetLastError` and demote to a plain `HostRegistrationError` after a clean rollback.
- **P3** — the unregistration-failure lease retention can be defeated by GC in the preferred-mode fresh-retry path (`HostWeightLease.__del__` unmaps pages CUDA still has registered); retain `(registration, lease)` at module level in the failure branches.

Everything else verified: registration lifecycle (per-file page coalescing, budget pre-check, rollback-with-retention, retryable close), staging-fallback selection before allocation, unregister-before-lease-close guard, worker teardown ordering, config validation on all surfaces, and zero registration interaction for checkpoint-mmap / AllGather / HWR-disabled. The writer/fsync/checksum and source-digest changes re-applied here are byte-identical to the #6486 pre-merge review of the same code.

vllm_omni/diffusion/offloader/distributed_layerwise_backend.py
hsliuustc0106: Fixed in `49a48ae4`. A process-local safety registry now owns every live `(registration, lease)` pair in addition to the backend. Failed unregistration leaves the pair retained across backend GC/fresh retry; a successful retry removes it before lease close. Added coverage for backend drop and retry cleanup.

vllm_omni/diffusion/offloader/cuda_host_registration.py
hsliuustc0106: Fixed in `49a48ae4`. Error consumption now prefers `runtime.cudaGetLastError` and uses `CDLL(None)` only as a compatibility fallback. Failure to resolve or reconcile that error state is a plain `HostRegistrationError`, so a cleanly rolled-back registration selects bounded staging; only live ranges that fail to unregister remain cleanup-fatal. Added runtime-handle, global-fallback, and clean-rollback regressions.

baonudesifeizhai:
Independent MiniMax-H3 Ref2VA B300 A/B</h3><p>I ran an independent A/B of the MiniMax-H3 DLO component-lifecycle change.</p><h4>Environment</h4><ul><li>GPU: 1× NVIDIA B300 SXM6 AC, 275,040 MiB</li><li>Driver: 610.43.02</li><li>Python: 3.13.12</li><li>PyTorch: 2.13.0+cu130</li><li>CUDA: 13.0</li><li>vLLM: 0.27.0</li><li>Model: MiniMax-H3 Ref2VA, node-local ext4 storage</li><li>Base: <code dir="ltr">29e3165bb938088d007cad9094cf66154c8d88c1</code></li><li>PR: <code dir="ltr">1843f05ced2fa1a4d43a0647dec7f48709c0c7da</code></li></ul><p>Both sides selected:</p><pre dir="ltr"><code>DLO host-weight plan active (rank-local, checkpoint_mmap)</code></pre><h4>Server command</h4><p>The same command was run from each checkout:</p><pre dir="ltr"><code>PYTHONPATH=&lt;CHECKOUT&gt; \
CUDA_VISIBLE_DEVICES=0 \
VLLM_WORKER_MULTIPROC_METHOD=spawn \
VLLM_OMNI_VIDEO_SYNC_TIMEOUT=14400 \
vllm serve /path/to/MiniMax-H3/Ref2VA \
  --omni \
  --host 127.0.0.1 \
  --port 18091 \
  --trust-remote-code \
  --num-gpus 1 \
  --usp 1 \
  --ring 1 \
  --text-encoder-tp-size 1 \
  --vae-patch-parallel-size 1 \
  --vae-parallel-mode tile \
  --vae-use-tiling \
  --diffusion-attention-backend CUDNN_ATTN \
  --dlo-no-use-allgather \
  --dlo-resident-layers 0 \
  --enforce-eager \
  --stage-init-timeout 1800 \
  --init-timeout 1800</code></pre><h4>Request</h4><ul><li>Task: Ref2VA</li><li>Resolution: 1344×768</li><li>Duration/FPS: 5 seconds at 24 FPS</li><li>Denoising steps: 2</li><li>Seed: 4242</li><li>Flow shift: 12</li><li>Audio flow shift: 3</li><li>Protocol: one excluded warmup followed by three sequential measured requests in the same server process</li></ul><pre dir="ltr"><code>curl --fail-with-body -sS \
  -o output.mp4 \
  -X POST http://127.0.0.1:18091/v1/videos/sync \
  -F 'prompt=Preserve the people, composition, motion, and synchronized ambience from the reference video.' \
  -F 'width=1344' \
  -F 'height=768' \
  -F 'fps=24' \
  -F 'num_inference_steps=2' \
  -F 'flow_shift=12' \
  -F 'seed=4242' \
  -F 'extra_params={"task":"ref2va","duration":5.0,"aspect_ratio":"16:9","audio_flow_shift":3.0}' \
  -F 'input_references=@r2va.mp4;type=video/mp4'</code></pre><h4>Results</h4><div><div><div>
Metric | Base | PR | Delta
-- | -- | -- | --
Client E2E runs | 30.826 / 31.023 / 30.792 s | 29.776 / 29.873 / 29.792 s | —
Client E2E mean ± population stddev | 30.880 ± 0.102 s | 29.814 ± 0.043 s | -3.45%
Server-side E2E mean ± population stddev | 30.314 ± 0.102 s | 29.266 ± 0.040 s | -3.46%
Sampled peak GPU memory | 23,554 MiB | 32,006 MiB | +8,452 MiB (+35.88%)

</div></div></div><p>GPU memory was sampled from <code dir="ltr">nvidia-smi memory.used</code> every 100 ms during the three measured requests.</p><p>All eight requests, including warmups, returned HTTP 200. Both servers completed normal shutdown with no <code dir="ltr">ERROR</code>, traceback, or OOM.</p><p>The generated output was 124 frames at 1344×768 with stereo 32 kHz audio. Decoded media hashes matched between base and PR:</p><pre dir="ltr"><code>Video RGB24 SHA256:
145cbbe9dc62806f5f2a0253572d6d7cfb7f969a24b83cd79d1be38f27117c6a

Audio PCM-f32le SHA256:
098ae9ac130e5e3efbcf0b5ec67c4350964c30b7360b2ee629ed400a5dcf0b12</code></pre><p>This Ref2VA B300 result is complementary to the FL2VA measurements in the PR: it shows a reproducible ~3.45% warm-request E2E improvement with identical decoded output, at the cost of a higher retained-memory plateau.</p><p>Targeted tests also passed:</p>

hsliuustc0106:
> Independent MiniMax-H3 Ref2VA B300 A/B
> 
> I ran an independent A/B of the MiniMax-H3 DLO component-lifecycle change.
> 
> #### Environment
> * GPU: 1× NVIDIA B300 SXM6 AC, 275,040 MiB
> * Driver: 610.43.02
> * Python: 3.13.12
> * PyTorch: 2.13.0+cu130
> * CUDA: 13.0
> * vLLM: 0.27.0
> * Model: MiniMax-H3 Ref2VA, node-local ext4 storage
> * Base: `29e3165bb938088d007cad9094cf66154c8d88c1`
> * PR: `1843f05ced2fa1a4d43a0647dec7f48709c0c7da`
> 
> Both sides selected:
> 
> ```
> DLO host-weight plan active (rank-local, checkpoint_mmap)
> ```
> 
> #### Server command
> The same command was run from each checkout:
> 
> ```
> PYTHONPATH=<CHECKOUT> 
> CUDA_VISIBLE_DEVICES=0 
> VLLM_WORKER_MULTIPROC_METHOD=spawn 
> VLLM_OMNI_VIDEO_SYNC_TIMEOUT=14400 
> vllm serve /path/to/MiniMax-H3/Ref2VA 
> --omni 
> --host 127.0.0.1 
> --port 18091 
> --trust-remote-code 
> --num-gpus 1 
> --usp 1 
> --ring 1 
> --text-encoder-tp-size 1 
> --vae-patch-parallel-size 1 
> --vae-parallel-mode tile 
> --vae-use-tiling 
> --diffusion-attention-backend CUDNN_ATTN 
> --enable-distributed-layerwise-offload 
> --dlo-no-use-allgather 
> --dlo-resident-layers 0 
> --enforce-eager 
> --stage-init-timeout 1800 
> --init-timeout 1800
> ```
> 
> #### Request
> * Task: Ref2VA
> * Resolution: 1344×768
> * Duration/FPS: 5 seconds at 24 FPS
> * Denoising steps: 2
> * Seed: 4242
> * Flow shift: 12
> * Audio flow shift: 3
> * Protocol: one excluded warmup followed by three sequential measured requests in the same server process
> 
> ```
> curl --fail-with-body -sS 
> -o output.mp4 
> -X POST http://127.0.0.1:18091/v1/videos/sync 
> -F 'prompt=Preserve the people, composition, motion, and synchronized ambience from the reference video.' 
> -F 'width=1344' 
> -F 'height=768' 
> -F 'fps=24' 
> -F 'num_inference_steps=2' 
> -F 'flow_shift=12' 
> -F 'seed=4242' 
> -F 'extra_params={"task":"ref2va","duration":5.0,"aspect_ratio":"16:9","audio_flow_shift":3.0}' 
> -F 'input_references=@r2va.mp4;type=video/mp4'
> ```
> 
> #### Results
> Metric	Base	PR	Delta
> Client E2E runs	30.826 / 31.023 / 30.792 s	29.776 / 29.873 / 29.792 s	—
> Client E2E mean ± population stddev	30.880 ± 0.102 s	29.814 ± 0.043 s	-3.45%
> Server-side E2E mean ± population stddev	30.314 ± 0.102 s	29.266 ± 0.040 s	-3.46%
> Sampled peak GPU memory	23,554 MiB	32,006 MiB	+8,452 MiB (+35.88%)
> GPU memory was sampled from `nvidia-smi memory.used` every 100 ms during the three measured requests.
> 
> All eight requests, including warmups, returned HTTP 200. Both servers completed normal shutdown with no `ERROR`, traceback, or OOM.
> 
> The generated output was 124 frames at 1344×768 with stereo 32 kHz audio. Decoded media hashes matched between base and PR:
> 
> ```
> Video RGB24 SHA256:
> 145cbbe9dc62806f5f2a0253572d6d7cfb7f969a24b83cd79d1be38f27117c6a
> ```
> 
> `Audio PCM-f32le SHA256:
> 098ae9ac130e5e3efbcf0b5ec67c4350964c30b7360b2ee629ed400a5dcf0b12`
> 
> This Ref2VA B300 result is complementary to the FL2VA measurements in the PR: it shows a reproducible ~3.45% warm-request E2E improvement with identical decoded output, at the cost of a higher retained-memory plateau.
> 
> Targeted tests also passed:

PR #6526 merged

Issue #6495, opened 2026-08-22
[New Model]: Audio8/Audio8-TTS-Preview-0.1b (FalconH1 hybrid slow AR + DualAR fast branch)


linyueqian:
### The model to consider.

https://huggingface.co/Audio8/Audio8-TTS-Preview-0.1b

Tracked separately from #5562 (the 0.6b) because despite sharing `model_type: "arktts"` and `architectures: ["ArkttsModel"]`, the two checkpoints do not share a slow branch, a tokenizer, or a license.

### The closest model vllm-omni already supports.

Two different models, for two different halves:

- **Fast branch and codec:** `fishaudio/s2-pro`, via `vllm_omni/model_executor/models/fish_speech/`. Identical DualAR fast head and codec shape (10 codebooks x 4096, 2048 samples per frame at 44.1 kHz, shipped as `codec.pth`).
- **Slow branch:** `nvidia/Nemotron-Labs-*` VoiceChat, via `nemotron_voicechat_thinker.py`. That is the in-tree precedent for a hybrid Mamba backbone on vLLM paged execution: it declares `IsHybrid`/`HasInnerState` and wires `MambaStateShapeCalculator.mamba2_state_shape`.

### What's your difficulty of supporting the model you want?

The slow AR branch is not a transformer. `modeling_arktts.py` builds it as `transformers.FalconH1Model` from a `FalconH1Config` derived from the arktts config, and generation maintains a `FalconHybridMambaAttentionDynamicCache` (attention KV plus Mamba SSM state). So this stage needs vLLM's hybrid state cache, not ordinary paged KV.

Differences from the 0.6b, from `config.json`:

| | 0.6b (#5562) | 0.1b |
|---|---|---|
| Slow branch | dense GQA transformer, 24L, dim 896, 14 heads / 2 KV | `FalconH1Model`, hybrid Mamba2 + attention, 24L, dim 512, 8 heads / 2 KV |
| Slow cache | ordinary paged KV | `FalconHybridMambaAttentionDynamicCache` |
| Mamba2 config | n/a | `mamba_d_ssm 768`, `mamba_n_heads 24`, `mamba_d_state 64`, `mamba_d_conv 4`, `mamba_chunk_size 128` |
| Fast branch | 4L, dim 896, 14 heads / 2 KV | 4L, dim 512, 8 heads / 2 KV |
| Codec | `codec.pth`, 10 x 4096, 44.1 kHz | same |
[...]

PR #6593, opened 2026-08-25
[Model] Add Audio8 TTS Preview 0.1B (FalconH1 hybrid DualAR)

NancyFyong:
Adds `Audio8/Audio8-TTS-Preview-0.1b`, the second checkpoint of the arktts DualAR TTS family. Addresses #6495.

It shares `model_type: "arktts"`, the 44.1 kHz codec (byte-identical `codec.pth`), the Fast AR, RAS sampling, the prompt format and the streaming pipeline with the 0.6b — but the Slow AR backbone is different:

- **0.6b** Slow AR = Qwen2 (pure attention)
- **0.1b** Slow AR = FalconH1 (Mamba2 + attention parallel hybrid, with muP multipliers)

## Ownership and license — please read

- #6495 was filed by @linyueqian, and @BruceLoveDecimal later said I'm welcome to take it and continue. I built this from the 0.6b work; happy to hand the branch over, co-author, or close in favour of someone else's version.
- The 0.1b ships under the **Audio8 Community License v1.0**, not Apache-2.0 like the 0.6b. Whether vllm-omni wants to carry a model under that license is a maintainer decision; this PR does not presume the answer.

## Rebased onto the merged 0.6b

#6157 (the 0.6b) is now merged, so this is a single 0.1b commit rebased onto current `main` (no longer stacked). The diff below is the net 0.1b delta.

## What's in it

- Backbone-aware `Audio8TTSConfig`: `slow_backbone="falcon_h1"` builds a `FalconH1Config` (annotated with the DualAR fields the backbone itself lacks); the 0.6b Qwen2 path is untouched.
- `Audio8TTS01BSlowARForConditionalGeneration`: a FalconH1 Slow AR with its own 4097-wide semantic head, loading the backbone through vLLM's own FalconH1 weight loader (so qkv/gate_up/mamba fusion is correct by construction).
- The `arktts` pipeline is now a resolver keyed on `slow_backbone`; new `deploy/audio8_tts_01b.yaml`.
- Shared codec, Fast AR, RAS sampling and `prompt_utils` reused unchanged — the prompt ids are byte-identical to the 0.1b reference processor.
- Registry entry, shared TTS adapter stage key, supported-models row, benchmark entry.

Two integration points are specific to wrapping vLLM's FalconH1 as a sub-backbone; both showed up on the first real GPU run and are now fixed:

- vLLM's `FalconH1Model` reads `vllm_config.model_config.hf_config` directly (unlike `Qwen2Model`, which calls `get_text_config()`), so the Slow AR hands it a config view whose `hf_config` is the FalconH1 `text_config`.
- The Slow AR declares `IsHybrid`/`HasInnerState` and forwards the mamba state shape/dtype/copy classmethods to `FalconH1ForCausalLM` (through the same view), so vLLM sizes the Mamba2 state cache. This follows the `nemotron_voicechat` precedent cited in #6495.

## Testing

- 6 CPU unit tests (config for both variants, `slow_backbone` resolver routing, registry mapping, adapter detection, Fast AR weight remap) plus offline and online e2e (`advanced_model`, real weights).
- Validated end to end on 1x H20 against current main: the offline e2e passes at `advanced_model`, and manual runs produce clean 44.1 kHz speech for English, Chinese, Japanese, a long English passage, and zero-shot voice cloning.

### Measured on 1x H20 (warm, `enforce_eager` per the shipped deploy)

Single-request, text-only:

| language | audio | RTF |
| --- | --- | --- |
| English | 5.4 s | 0.83 |
| Chinese | 5.0 s | 0.83 |
| Japanese | 5.0 s | 0.82 |
| English (long) | 13.5 s | 0.82 |
| voice clone | 4.4 s | 1.53 |

Offline batch throughput (same prompt, default deploy `max_num_seqs: 4`):

| concurrency | req/s | audio s / wall s | aggregate RTF |
| --- | --- | --- | --- |
| 1 | 0.22 | 1.2 | 0.86 |
| 4 | 0.77 | 3.4 | 0.30 |
| 8 | 0.94 | 4.0 | 0.25 |

The 0.1b is slower per request than the 0.6b (which reaches ~0.19 RTF): it runs `enforce_eager` (full-model CUDA-graph capture over the per-step Mamba2 state advance isn't validated yet) and the Mamba2 selective-scan/causal-conv kernels carry launch overhead. Batching recovers most of it (RTF 0.86 → 0.25). Voice cloning is higher because it encodes the reference audio each request.

## Audio samples (vLLM-Omni, 1x H20)

Offline output for five prompts (mono PCM_16, 44.1 kHz). WAVs and a drag-and-drop guide are on the fork release [`audio8-tts-01b-samples`](https://github.com/NancyFyong/vllm-omni/releases/tag/audio8-tts-01b-samples); inline players below are added by dragging each file onto its placeholder in the browser editor.

| sample | text | audio |
| --- | --- | --- |
| English | "Welcome to Audio8 TTS, a compact multilingual text to speech model." | 5.4 s |
| Chinese | "欢迎使用 Audio8 语音合成，这是一个轻量的多语种模型。" | 5.0 s |
| Japanese | "Audio8 音声合成へようこそ。これは軽量な多言語モデルです。" | 5.0 s |
| English (long) | long passage | 13.5 s |
| voice clone | reference = the English sample above (no third-party voice) | 4.4 s |

The 0.1b's official HuggingFace transformers reference code imports `FalconHybridMambaAttentionDynamicCache`, which current transformers has removed, so an A/B against the HF reference (as done for the 0.6b) could not be produced on this stack — the vllm-omni path runs because it builds on vLLM's own maintained FalconH1 rather than the checkpoint's remote code.

## Known gaps

- `enforce_eager` is on by default for Stage 0; a piecewise-graph Mamba run could close the per-request gap but hasn't been measured.
- Voice-clone streaming and multi-language output were exercised manually, not in CI. CI marks are `core_model` (CPU) with e2e at `advanced_model`.


BruceLoveDecimal:
@NancyFyong No worries—please feel free to claim the issue and continue with the 0.1B integration. Since you’ve already done substantial work on the 0.6B PR, it makes perfect sense for you to carry this through to 0.1B as well

hsliuustc0106:
This PR touches **vllm_omni/model_executor/ tests/model_executor/ examples/online_serving/ vllm_omni/entrypoints/ benchmarks/tts/** (42 files). Based on CODEOWNERS coverage of the changed files, the most-related reviewers appear to be:

>  @linyueqian @gcanlin @tzhouam

linyueqian (commented):
Ran GPU e2e on this at `c9f03e5e4` (1x L20X 141GB, vLLM 0.27.0 / torch 2.13.0+cu129 / transformers 5.14.1, real `Audio8/Audio8-TTS-Preview-0.1b` weights). Thanks for putting it up as a draft with the e2e gap called out; that gap turns out to be the whole story, so here is what it looks like from a live run.

Since the PR ships no e2e test for the 0.1b, I wrote one mirroring `tests/e2e/offline_inference/test_audio8_tts.py` against `audio8_tts_01b.yaml`.

## Blocking

### 1. The 0.1b Slow AR never constructs. `vllm_omni/model_executor/models/audio8_tts/audio8_tts_falcon_slow_ar.py:83`

Every boot dies at model init:

```
File "vllm_omni/model_executor/models/audio8_tts/audio8_tts_falcon_slow_ar.py", line 83, in __init__
  self.model = FalconH1Model(vllm_config=vllm_config, prefix=maybe_prefix(prefix, "model"))
File "vllm/model_executor/models/falcon_h1.py", line 425, in __init__
  self.vocab_size = config.vocab_size
AttributeError: 'Audio8TTSConfig' object has no attribute 'vocab_size'
```

`FalconH1Model.__init__` reads `vllm_config.model_config.hf_config` **directly** (`falcon_h1.py:418`). `Qwen2Model` reads `hf_config.get_text_config()` (`qwen2.py:343`), which is why the 0.6b path works. So the backbone receives the outer `Audio8TTSConfig`, which takes `vocab_size` as a ctor argument and never assigns it.

This is the module's stated design assumption rather than a slip. The comment at `audio8_tts_falcon_slow_ar.py:57` reads "For 0.1b, `get_text_config()` returns a FalconH1Config annotated with the DualAR fields the parent reads", and `FalconH1Model` is the one consumer that does not call it.

Smallest fix, matching what is already in tree: shallow-copy `vllm_config` and `model_config`, set `hf_config` / `hf_text_config` to `self.text_config`, and pass that to the backbone. See `higgs_audio_v3/higgs_audio_v3_talker.py:152-158` and `moss_tts/modeling_moss_tts_talker.py:820-826`.

### 2. Hybrid-Mamba interfaces are not declared. `audio8_tts_falcon_slow_ar.py:44`

With finding 1 patched, startup dies one step later, in KV cache spec construction:

```
File "vllm/v1/worker/gpu_model_runner.py", line 7827, in get_kv_cache_spec
  if spec := attn_module.get_kv_cache_spec(self.vllm_config):
File "vllm/model_executor/layers/mamba/abstract.py", line 65, in get_kv_cache_spec
  assert mamba_block_size is not None
AssertionError
```

`Audio8TTS01BSlowARForConditionalGeneration` inherits only from the Qwen2 Slow AR, so it declares neither `IsHybrid` / `HasInnerState` nor `get_mamba_state_shape_from_config` / `get_mamba_state_dtype_from_config`. Measured on this head:

| class | `is_hybrid` |
| --- | --- |
| `Audio8TTS01BSlowARForConditionalGeneration` | **False** |
| `FalconH1ForCausalLM` | True |
| `NemotronVoiceChatThinkerForConditionalGeneration` | True |

`ModelRegistry` derives `is_hybrid` from `getattr(model, "is_hybrid", False)`, and `platforms/interface.py:644` gates `_align_hybrid_block_size` on it. That call is the only thing that sets `cache_config.mamba_block_size`, hence the assertion.

Both in-tree hybrid siblings declare the interfaces and delegate the state classmethods: `nemotron_voicechat/nemotron_voicechat_thinker.py:135-172` and `audex/audex_omni.py:287-310`. One caveat when you add them: `FalconH1ForCausalLM.get_mamba_state_shape_from_config` also reads `.hf_config` directly (`falcon_h1.py:555`), so it needs the same sub-config shim as finding 1.

### 3. Stage-0 EngineCore dies during generation

With findings 1 and 2 patched locally using the in-tree pattern above, the model does load (343 weights, mamba page sizing succeeds, attention block size aligned to 208 tokens), and then:

| Path | Result |
| --- | --- |
| Offline, single request, `advanced_model` | passed, 4.10 s, 44.1 kHz, codec decoded |
| Offline, 4 concurrent (`max_num_seqs: 4`, the shipped default) | **stage-0 core dies**, `OmniEngineDeadError`; reproduced twice, including under `CUDA_LAUNCH_BLOCKING=1` |
| Online `vllm serve --omni`, single `/v1/audio/speech` | **HTTP 500** `EngineDeadError` after 4.9 s |

The core process exits with no Python traceback and no kernel signal in `dmesg`. These two patches are mine, so the final shape of this one may differ once you fix 1 and 2 properly. Flagging it because it means the path needs more than config plumbing.

## Test coverage

The 6 new CPU tests pass (43 across the whole Audio8 suite, matching the PR body), but they all assert **metadata**: config shapes, resolver routing, registry entry, adapter stage key, weight remap. None constructs `Audio8TTS01BSlowARForConditionalGeneration`, which is why both blockers survived a green suite.

`tests/model_executor/models/audio8_tts/test_audio8_tts_01b.py:68` asserts `get_text_config()` returns a `FalconH1Config`. That pins the producer side of exactly the contract finding 1 shows the consumer does not honor. A single test that instantiates the model class under a dummy `VllmConfig` would have caught both.

There is also no e2e test for the 0.1b: `tests/e2e/{offline_inference,online_serving}/test_audio8_tts.py` stay hardcoded to `audio8_tts.yaml` and the 0.6b, so nothing in CI exercises this path even at `advanced_model`.

## Suggestion

`compute_logits` (`audio8_tts_falcon_slow_ar.py:135-145`) scatters the compact `[B, 4097]` semantic head into a full `[B, 69633]` `-inf` tensor so the inherited `_compact_semantic_logits` can slice 4097 columns back out. The parent's own docstring at `audio8_tts_slow_ar.py:283` calls avoiding that full-vocabulary tensor "the difference between a cheap and an expensive AR step". Hoisting a `_semantic_logits()` hook onto the parent would let the 0.1b return `compact` directly and skip the round trip.

## Checked and fine

- The released `config.json` does carry `slow_backbone: "falcon_h1"`, so the resolver routes correctly. Engine log confirms `Resolved architecture: Audio8TTS01BSlowARForConditionalGeneration`. It also carries `num_key_value_heads: 2`, so the #6424 NaN-logits trap does not apply here.
- muP multipliers reach the config correctly through `kwargs.get()` (`embedding_multiplier` 0.10888671875, `lm_head_multiplier` 0.078125).
- `enforce_eager: true` on stage 0 is well justified for the Mamba2 state path.
- The 0.6b findings I raised on #6157 are fixed in this stack: the tokenized helper moved to `tests/helpers/client.py` with no `.request_output` hop, and the concurrency test now uses 4 distinct prompts with `assert_distinct_outputs`.

## Not verified

Audio **correctness** is still open. The single-request offline test passes duration and sample-rate assertions, but I never got a waveform out to transcribe because both the online path and a standalone generator hit finding 3. Voice cloning is untested. Whether `semantic_output` should be scaled by `lm_head_multiplier` (the code says the reference does not) is unchecked against `modeling_arktts.py`.

On ownership and the Audio8 Community License: both are maintainer calls, not review findings. No opinion from me on either beyond noting they are still open.

NancyFyong:
Thanks for the live run — that was exactly the gap, and all three findings were on the money. They're fixed now and I re-ran the whole thing on current main.

**1 & 2 (config view + hybrid interfaces).** Fixed the way you pointed at. The Slow AR now inherits `HasInnerState`/`IsHybrid`, and a small module-level `_falcon_config_view()` shallow-copies `vllm_config`/`model_config` and swaps `hf_config`/`hf_text_config` to the FalconH1 text config. So `FalconH1Model.__init__` finds `vocab_size` and the mamba-state classmethods get the mamba dims — that's your caveat about `get_mamba_state_shape_from_config` reading `.hf_config` directly, so those classmethods go through the same view rather than the raw config.

**3 (core death at concurrency / online 500).** Doesn't reproduce once 1 & 2 are done properly. Offline single + 4-concurrent (the shipped `max_num_seqs: 4`) and online single `/v1/audio/speech` + streaming all pass at `advanced_model` with real weights on 1x H20. The online path runs the ASR/HNR asserts too, so there's intelligible audio coming out now, including voice cloning — that answers the correctness bit you couldn't get to.

**Tests.** Added the two e2e files (offline + online, mirroring the one you wrote) so the path actually runs in CI at `advanced_model`, plus two CPU regressions aimed squarely at your "nothing constructs the class" point: one pins `is_hybrid` and the mamba classmethods, the other pins that the config view hands FalconH1 a config carrying `vocab_size`. Both are mutation-checked — drop the interface or revert the shim and they go red.

On the `compute_logits` suggestion: agreed, the full-vocab scatter does an extra `[B, 69633]` alloc per step that the parent's own docstring flags as the expensive path. I left it as-is to keep this diff additive, but I'm happy to hoist a `_semantic_logits()` hook onto the parent so 0.1b can return the compact tensor directly — say the word and I'll fold it in, otherwise I'll file it as a follow-up.

Rebased down to a single commit on current main now that the 0.6b has merged. Thanks again for digging in.

vllm_omni/model_executor/models/audio8_tts/audio8_tts_falcon_slow_ar.py
```diff
@@ -0,0 +1,282 @@
...
+        self.make_empty_intermediate_tensors = getattr(self.model, "make_empty_intermediate_tensors", None)
+
+        # Summed multi-codebook input embedding table (identical to 0.6b).
+        self.codebook_embeddings = nn.Embedding(
+            self._codebook_size * self._num_codebooks,
+            self.text_config.hidden_size,
+        )
+
+        # Fast AR: separate compilation context, same as the 0.6b Slow AR.
+        fast_ar_compilation = copy.copy(vllm_config.compilation_config)
```
lishunyang12: Extract the shared Fast AR initialization into a helper used by both backbones.

NancyFyong: Done — moved it to `_init_fast_ar` on the parent, both backbones call that now. Re-ran the 0.6b and 0.1b offline e2e to be safe, both still pass. Thanks!

NancyFyong:
@lishunyang12 @linyueqian  could you PTAL when you have time?

PR #6598, opened 2026-08-25
[Bugfix] Handle DoS Overflow Cases

alex-jw-brooks:
## Purpose
Currently, we do not set bounds many of the `int` fields, which can actually be used to maliciously kill the server. There are a couple of cases here that are worth pointing out:
- Explicit params, e.g., `seed`, which we can easily validate and reject on the frontend based on dtype, similar to the way vLLM does (e.g., [here](https://github.com/vllm-project/vllm/blob/d3e2888c7588fe3a7be93606c7c626dbfd304d2e/vllm/entrypoints/openai/chat_completion/protocol.py#L1086)).
- Other cases that are less straightforward to validate, e.g., in extra params.

Because of this, we should both validate the params to reject bad values cleanly as validation errors, also tightening bounds where needed (e.g., `num_inference_steps` and `max_new_tokens` should be positive), but we should also catch overflow errors in the orchestrator to prevent engine death for more implicit cases, e.g., huge values maliciously passed through something like `stop_token_ids`.

Examples of exploiting overflow behaviors to kill the server:

### Out of range seed can be used to kill multistage diffusion models
Start the server:
```bash
vllm serve ByteDance-Seed/BAGEL-7B-MoT --omni --port 8093 --omni
```
Then try to send a huge value (>  2**63 - 1) for the seed, like this:
```bash
curl -sS http://127.0.0.1:8093/v1/images/generations \
    -H 'Content-Type: application/json' \
    -d '{"prompt":"a cat","num_inference_steps":1,"seed":1000000000000000000000000000000}'
```

This will kill the engine and ultimately bring down the server.
```
...
(APIServer pid=2921201) ERROR 08-25 04:00:57 [async_omni.py:947] [AsyncOmni] Engine dead: can't serialize ints < -2**63 or > 2**64 - 1
...
(APIServer pid=2921201) OverflowError: can't serialize ints < -2**63 or > 2**64 - 1
(APIServer pid=2921201) ERROR 08-25 04:00:57 [api_server.py:2062]     waiter = self._router.register(key)
(APIServer pid=2921201) ERROR 08-25 04:00:57 [api_server.py:2062]              ^^^^^^^^^^^^^^^^^^^^^^^^^^
(APIServer pid=2921201) ERROR 08-25 04:00:57 [api_server.py:2062]   File "/home/alex-jw-brooks/vllm-omni/vllm_omni/engine/rpc_result_router.py", line 45, in register
(APIServer pid=2921201) ERROR 08-25 04:00:57 [api_server.py:2062]     raise RuntimeError(self._terminal_error.error)
(APIServer pid=2921201) ERROR 08-25 04:00:57 [api_server.py:2062] RuntimeError: can't serialize ints < -2**63 or > 2**64 - 1
(APIServer pid=2921201) INFO:     127.0.0.1:43570 - "POST /v1/images/generations HTTP/1.1" 500 Internal Server Error
(APIServer pid=2921201) INFO:     Shutting down
(APIServer pid=2921201) INFO:     Waiting for application shutdown.
(APIServer pid=2921201) INFO:     Application shutdown complete.
(APIServer pid=2921201) INFO:     Finished server process [2921201]
(APIServer pid=2921201) INFO 08-25 04:00:58 [omni_base.py:724] [AsyncOmni] Shutting down
(APIServer pid=2921201) INFO 08-25 04:00:58 [async_omni_engine.py:1881] [AsyncOmniEngine] Shutting down Orchestrator
```

### (Less Explicit) Maliciously Large Stop Tokens IDs
In the previous example, `seed` can be validated easily while parsing, which this PR adds. However, we still do need to catch the `OverflowError` generically on the orchestrator for cases where we may pass large numerics that are serialized as nested parts of the request. For example...
Start the server:
```
vllm serve Qwen/Qwen3-Omni-30B-A3B-Instruct --omni --port 8123
```
Then try to send a request with a giant value for `stop_token_ids`, like this:
```
curl -sS http://127.0.0.1:8123/v1/chat/completions \
    -H 'Content-Type: application/json'   \
    -d '{"messages":[{"role":"user","content":"hi"}],"stop_token_ids":[1000000000000000000000000000000]}'
```
Will crash the server in a similar manner:
```
(APIServer pid=2928108) ERROR 08-25 04:07:27 [api_server.py:1234] RuntimeError: can't serialize ints < -2**63 or > 2**64 - 1
(APIServer pid=2928108) INFO:     127.0.0.1:41026 - "POST /v1/chat/completions HTTP/1.1" 500 Internal Server Error
(APIServer pid=2928108) INFO:     Shutting down
(APIServer pid=2928108) INFO:     Waiting for application shutdown.
(APIServer pid=2928108) INFO:     Application shutdown complete.
(APIServer pid=2928108) INFO:     Finished server process [2928108]
(APIServer pid=2928108) INFO 08-25 04:07:30 [omni_base.py:724] [AsyncOmni] Shutting down
(APIServer pid=2928108) INFO 08-25 04:07:30 [async_omni_engine.py:1881] [AsyncOmniEngine] Shutting down Orchestrator
(omni) [alex-jw-brooks@h100 ~]$ 
```

## Test Plan
- Add regression test for ensuring orchestrator doesn't explode when it sees an `OverflowError` that currently fails on `main`
- Add lt/gt bounds to protocol values where applicable, similar to what vLLM already has in the upstream
- Add error handling for `OverflowError` to the Orchestrator and ensure that the cases above don't crash the server

## Test Result
Tests pass. Example for the respective crash cases:

Bagel with a large seed will hit our earlier validation and points to the seed specifically.:
```
{"error":{"message":"1 validation error:\n  {'type': 'less_than_equal', 'loc': ('body', 'seed'), 'msg': 'Input should be less than or equal to 9223372036854775807', 'input': 1000000000000000000000000000000, 'ctx': {'le': 9223372036854775807}}","type":"Bad Request","param":"body.seed","code":400}}
```

Qwen3Omni out of range tokens hits the except around the orchestrator, which is less ideal since it's less specific, but at least prevents engine death:
```
{"error":{"message":"Request contains a value that cannot be serialized: can't serialize ints < -2**63 or > 2**64 - 1","type":"BadRequestError","param":null,"code":400}}
```

### DIFF
--- a/vllm_omni/entrypoints/openai/protocol/audio.py
+++ b/vllm_omni/entrypoints/openai/protocol/audio.py
@@ -1,8 +1,14 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM-Omni project
 import math
 from typing import Any, Literal
 
 import numpy as np
 from pydantic import AliasChoices, BaseModel, Field, field_validator, model_validator
+
+# Bound int request fields to avoid overflow issues.
+_INT64_MIN = -(2**63)
+_INT64_MAX = 2**63 - 1
 
 _MAX_EMBEDDING_DIM = 8192
 
@@ -142,6 +148,8 @@ class OpenAICreateSpeechRequest(BaseModel):
     )
     max_new_tokens: int | None = Field(
         default=None,
+        ge=1,
+        le=_INT64_MAX,
         description="Maximum tokens to generate",
     )
     seed: int | None = Field(
@@ -154,6 +162,7 @@ class OpenAICreateSpeechRequest(BaseModel):
     initial_codec_chunk_frames: int | None = Field(
         default=None,
         ge=0,
+        le=_INT64_MAX,
         description="Per-request initial chunk size override. If null, computed dynamically based on server load.",
     )
     non_streaming_mode: bool | None = Field(
@@ -365,10 +374,14 @@ class OpenAICreateAudioGenerateRequest(BaseModel):
     )
     num_inference_steps: int | None = Field(
         default=None,
+        ge=1,
+        le=_INT64_MAX,
         description="Number of inference steps",
     )
     seed: int | None = Field(
         default=None,
+        ge=_INT64_MIN,
+        le=_INT64_MAX,
         description="Random seed for reproducibility",
     )
 
@@ -413,8 +426,8 @@ class SpeechBatchItem(BaseModel):
     ref_audio: str | None = None
     ref_text: str | None = None
     x_vector_only_mode: bool | None = None
-    max_new_tokens: int | None = None
-    initial_codec_chunk_frames: int | None = Field(default=None, ge=0)
+    max_new_tokens: int | None = Field(default=None, ge=1, le=_INT64_MAX)
+    initial_codec_chunk_frames: int | None = Field(default=None, ge=0, le=_INT64_MAX)
     non_streaming_mode: bool | None = None
 
 
@@ -433,8 +446,8 @@ class BatchSpeechRequest(BaseModel):
     ref_audio: str | None = None
     ref_text: str | None = None
     x_vector_only_mode: bool | None = None
-    max_new_tokens: int | None = None
-    initial_codec_chunk_frames: int | None = Field(default=None, ge=0)
+    max_new_tokens: int | None = Field(default=None, ge=1, le=_INT64_MAX)
+    initial_codec_chunk_frames: int | None = Field(default=None, ge=0, le=_INT64_MAX)
     non_streaming_mode: bool | None = None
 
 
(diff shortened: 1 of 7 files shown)


lishunyang12 (approved):

vllm_omni/entrypoints/openai/protocol/audio.py
```diff
@@ -1,9 +1,15 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM-Omni project
 import math
 from typing import Any, Literal
 
 import numpy as np
 from pydantic import AliasChoices, BaseModel, Field, field_validator, model_validator
 
+# Bound int request fields to avoid overflow issues.
+_INT64_MIN = -(2**63)
```
hsliuustc0106, line 10: nit: `_INT64_MIN`/`_INT64_MAX` are now defined identically in audio.py, images.py, and videos.py — consider hoisting them to one shared module so the bounds can't drift apart.

vllm_omni/entrypoints/openai/protocol/audio.py
```diff
@@ -142,18 +148,21 @@ def _normalize_voice(cls, voice: str | dict[str, str]) -> str:
...
         default=None,
+        ge=1,
+        le=_INT64_MAX,
         description="Maximum tokens to generate",
     )
     seed: int | None = Field(
         default=None,
-        ge=0,
-        le=2**63 - 1,
+        ge=_INT64_MIN,
```
hsliuustc0106, line 157: This widening is user-visible on `/v1/audio/speech` (`seed=-1` used to 422 — see the `seed_negative` → `seed_below_min` test change), so worth a line in the PR body noting negative seeds are now accepted for parity with upstream vLLM chat seed semantics; only the commit message records the intent today.

PR #6600, opened 2026-08-25
[Perf][TTS] Add deadline-aware audio scheduling

Sy0307:
## Summary

Add an opt-in playback-deadline policy for async-chunk TTS stages and fix duplicate terminal codec-frame emission.

The existing `active_stream_window` uses FIFO promotion. With `active_stream_policy: edf`, the final audio generation stage estimates each request's playback horizon from emitted PCM and reselects the most urgent ready streams at each chunk boundary. Streams close to buffer exhaustion are protected from first-audio backlog.

The default remains unchanged: `active_stream_window: 0` and FIFO behavior.

## What Changed

- Add `active_stream_policy: fifo|edf` validation in `OmniChunkTransferAdapter`.
- Track per-request first-audio and playback deadlines using monotonic server-side timestamps.
- Restrict deadline accounting to `generation` stages with `engine_output_type=audio`.
- Re-select up to `K=active_stream_window` ready audio streams by deadline.
- Protect established streams whose estimated playback horizon is within the configured safety margin.
- Add an operator note for using EDF with a codec chunk ramp.
- Fix non-ramp Qwen3-TTS terminal handling: when the previous chunk ended exactly on a boundary, terminal emits an empty finish marker instead of resending the last chunk. Tail frames are flushed exactly once.

## Configuration

```yaml
active_stream_window: 2
connectors:
  connector_of_shared_memory:
    extra:
      codec_chunk_ramp: [4, 4, 8, 16, 25]
      active_stream_policy: edf
      playback_deadline_first_chunk_ms: 100
      playback_deadline_safety_margin_ms: 250
```

The deadline is a server-side estimate. `/v1/audio/speech` does not currently provide client playback acknowledgements, so network and client jitter are not directly fed back into scheduling.

## Validation

- `174 passed` across Qwen3-TTS async-chunk, incremental Code2Wav, and Code2Wav tests on the H200 vLLM 0.27 environment.
- `ruff check` passed.
- `ruff format --check` passed.
- Added tests for exact-boundary terminal output, tail-only flush, frame uniqueness, EDF ordering, K>1 selection, protection, timing config validation, cleanup, and scheduler audio accounting.

## H200 End-to-End A/B

Workload: Qwen3-TTS 1.7B CustomVoice, same model/config/GPU, `K=2`, ramp `[4,4,8,16,25]`, 512 requests per run, 3 repetitions, fixed `max_new_tokens=26` so every request generated exactly 2 seconds of audio. FIFO and EDF had identical total audio duration.

| Concurrency | Metric | FIFO | EDF | Delta |
|---:|---|---:|---:|---:|
| 64 | Request throughput | 30.10 req/s | 34.73 req/s | +15.4% |
| 64 | Audio throughput | 60.19x | 69.46x | +15.4% |
| 64 | p95 Audio TTFP | 2173 ms | 1061 ms | -51.2% |
| 64 | p95 E2E | 2216 ms | 2186 ms | -1.4% |
| 64 | p99 underrun | 0 ms | 42.5 ms | within 100 ms budget |
| 128 | Request throughput | 32.50 req/s | 35.01 req/s | +7.7% |
| 128 | Audio throughput | 65.01x | 70.02x | +7.7% |
| 128 | p95 Audio TTFP | 4309 ms | 3084 ms | -28.4% |
| 128 | p95 E2E | 4354 ms | 4252 ms | -2.3% |
| 128 | p99 underrun | 0 ms | 42.0 ms | within 100 ms budget |

The throughput increase is workload-dependent and is not the feature guarantee. It may come from improved phase/batch alignment after EDF reordering; the intended contract is no throughput regression plus lower first-audio latency and bounded playback gaps.

## Limitations / Follow-ups

- No client playback ACK path yet; the deadline is based on server-side output emission.
- `K` remains static and requires tuning per GPU class.
- Multi-replica deployments still require sticky per-stream routing.
- This does not implement true suffix Transformer KV incrementality; that remains separate from scheduling.
- A longer Poisson benchmark and cross-GPU validation are still needed before enabling EDF by default.

## Local Validation

```bash
pytest -q tests/model_executor/stage_input_processors/test_qwen3_tts_async_chunk.py \
  tests/model_executor/models/qwen3_tts/test_qwen3_tts_incremental_decode.py \
  tests/model_executor/models/qwen3_tts/test_qwen3_tts_code2wav.py
```

Prerequisites: vLLM-Omni dependencies and the Qwen3-TTS checkpoint for E2E validation.

### DIFF
--- a/docs/user_guide/examples/online_serving/text_to_speech.md
+++ b/docs/user_guide/examples/online_serving/text_to_speech.md
@@ -457,6 +457,7 @@ python qwen3_tts/gradio_fastrtc_demo.py --api-base http://localhost:8000
 - With async chunking, Qwen3-TTS Base voice cloning sends the full reference context in the first Code2Wav packet, then caches that prefix on the Code2Wav stage for follow-up chunks in the same request.
 - `vllm_omni/deploy/qwen3_tts.yaml` is the default deploy config (loaded by HF `model_type`); per-stage runtime overrides are available via `--stage-N-<field> <value>`.
 - Under vocoder-bound overload (single-stream `rtf_p99 ≥ 1` at the target concurrency), set `active_stream_window: 2` at the top of the deploy yaml to cap simultaneously active Stage 1 streams. Off by default; trades TTFP for streaming continuity. See [#3592](https://github.com/vllm-project/vllm-omni/pull/3592) for the mechanism and tradeoff numbers.
+- For a bounded active-stream deployment, `active_stream_policy: edf` can prioritize streams whose server-estimated playback buffer is closest to exhaustion. Configure it under the connector's `extra` block and use a `codec_chunk_ramp`; this is opt-in, estimates playback from server-side PCM emission (no client playback acknowledgements), and targets first-audio latency and continuity rather than single-request completion latency.
 
 ---
 
--- a/tests/core/sched/test_omni_generation_scheduler_playback_deadline.py
+++ b/tests/core/sched/test_omni_generation_scheduler_playback_deadline.py
@@ -1,0 +1,39 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+
+from types import SimpleNamespace
+
+import pytest
+import torch
+
+from vllm_omni.core.sched.omni_generation_scheduler import OmniGenerationScheduler
+
+pytestmark = [pytest.mark.core_model, pytest.mark.cpu]
+
+
+def test_generation_scheduler_records_audio_output_for_edf(mocker):
+    scheduler = OmniGenerationScheduler.__new__(OmniGenerationScheduler)
+    scheduler.chunk_transfer_adapter = SimpleNamespace(
+        playback_deadline_enabled=True,
+        record_audio_output=mocker.Mock(),
+    )
+    multimodal_output = {
+        "model_outputs": torch.zeros(24000),
+        "sr": torch.tensor(24000),
+    }
+
+    scheduler._record_playback_deadline_output("req", multimodal_output)
+
+    scheduler.chunk_transfer_adapter.record_audio_output.assert_called_once_with("req", 24000, 24000)
+
+
+def test_generation_scheduler_skips_audio_accounting_when_edf_is_disabled(mocker):
+    scheduler = OmniGenerationScheduler.__new__(OmniGenerationScheduler)
+    scheduler.chunk_transfer_adapter = SimpleNamespace(
+        playback_deadline_enabled=False,
+        record_audio_output=mocker.Mock(),
+    )
+
+    scheduler._record_playback_deadline_output("req", {"model_outputs": torch.zeros(24000)})
+
+    scheduler.chunk_transfer_adapter.record_audio_output.assert_not_called()
--- a/tests/distributed/omni_connectors/test_chunk_transfer_adapter.py
+++ b/tests/distributed/omni_connectors/test_chunk_transfer_adapter.py
@@ -113,6 +113,7 @@ def build_adapter(monkeypatch, mocker: MockerFixture):
         *,
         stage_id: int = 1,
         model_mode: str = "ar",
+        engine_output_type: str | None = None,
         max_num_seqs: int = 2,
         active_stream_window: int = 0,
         connector_extra: dict | None = None,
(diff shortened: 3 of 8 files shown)


docs/user_guide/examples/online_serving/text_to_speech.md
```diff
@@ -457,6 +457,7 @@ python qwen3_tts/gradio_fastrtc_demo.py --api-base http://localhost:8000
 - With async chunking, Qwen3-TTS Base voice cloning sends the full reference context in the first Code2Wav packet, then caches that prefix on the Code2Wav stage for follow-up chunks in the same request.
```
hsliuustc0106, line 457: does it only work for a single model? if so, I suggest not to add this optimization. If not, please check other important tts model with audio output and add the user guide and feature design docs

Sy0307:
Self-review on head `5740ca8`: I checked that the default FIFO behavior remains unchanged, EDF selection is limited to generation stages with audio output, deadline state is cleaned up per request, and the Qwen3-TTS terminal boundary path emits the final marker without duplicating the last frame. The focused Qwen3-TTS/scheduler/connector validation reported 174 passed, with ruff and formatting checks passing. The generic scheduling path is intended for audio-producing generation stages; the exact-boundary terminal fix is Qwen3-TTS-specific. This PR is not merge-ready yet because GitHub reports `CONFLICTING` and `REVIEW_REQUIRED`, and cross-model validation/docs remain to be addressed.

docs/user_guide/examples/online_serving/text_to_speech.md
Sy0307: The scheduling policy is generic for audio-producing generation stages, while the exact-boundary terminal fix is Qwen3-TTS-specific. The current validation is Qwen3-focused, so I agree that cross-model validation and user-facing documentation are needed before merging this as a general optimization. I will treat that scope clarification as a prerequisite rather than overclaiming support.

linyueqian (approved):
Reviewed the full diff at 5740ca8e against merge base 072bfc02. Both halves hold up, and the terminal-frame fix is better motivated than the description suggests.

The duplicate terminal emission fix is correct, and the counter swap that carries it is the substantive part. Moving from `put_req_chunk` to `ramp_chunk_count` looks like an incidental rename, but the two have different lifetimes: `ramp_chunk_count` is popped at segment boundaries in the same block that pops `code_prompt_token_ids`, while `put_req_chunk` lives for the whole request. Since `length` is derived from `code_prompt_token_ids`, the new `emitted_frames` arithmetic is now dimensionally consistent with the value it is compared against, where the old `first_chunk` test compared a request-lifetime counter against a segment-local length. That silently mis-set `first_chunk` on every segment after the first for multi-segment streaming requests, so this change quietly fixes a second defect beyond the exact-boundary duplicate the title describes. It also matches how the ramp branch immediately above already derives its `chunk_index`. Worth calling out in the description so it is not lost as a refactor.

On the EDF path I specifically checked whether the feature can actually turn on, because `playback_deadline_enabled` gates on `worker_type` and `engine_output_type`, neither of which appears anywhere in `qwen3_tts.yaml` or in any other deploy YAML in the tree. They come from the pipeline registry instead: `QWEN3_TTS_PIPELINE` declares Stage 1 as `LLM_GENERATION` with `engine_output_type="audio"`, so uncommenting `active_stream_policy: edf` in the deploy config does take effect for this model. The deadline model itself is sound, with `max(deadline, now)` correctly restarting the horizon for a starved stream rather than crediting it for audio it has already played through, and the two-level sort key keeps a near-underrun stream from being preempted by a backlog of overdue first-audio deadlines.

The default path is untouched: `active_stream_policy` defaults to `fifo`, EDF additionally requires a positive `active_stream_window`, and the configuration errors are raised at construction so a bad deploy config fails loudly at startup rather than degrading at runtime. The documentation is honest about estimating playback from server-side PCM emission with no client acknowledgements.

Both inline notes are non-blocking observations about configuration ergonomics and queue-promotion symmetry rather than defects in the mechanism.

Validation: static review only. I did not execute the fork head, run the new tests, or measure any latency or continuity claim, so the first-audio and underrun behaviour rests on your reported numbers. I confirmed the counter lifetimes, the registry-supplied stage attributes, and the configuration validation by reading the code at the reviewed head. Note this is a performance feature, so merging is still gated by the current release freeze.

Sy0307:
Continue to foucs on this PR.

